PR #9300, opened 2021-05-27
Fix #9281: acquire a company uses special bookkeeping to make you rich

### LINKED ISSUE #9281: Company Takeover Maths not correct.
Not sure if bug but here is the scenario

1) Company A buys out company B
2) All the loans and all the balance should get added to company A finances.

However what is happening is
1) Company B:- (Balance - Loan amount) is added to company A finance.
2) Company B:- Loan amount is added as is to company A finance despite above money recovery.
3) Net result:- Company A Loan = Loan of A + Loan of B and Balance of A = Balance of A + (Balance of B - Loan of B)

This does not make sense. Screenshots attached showing pre and post finance details.
premerge_OpenTTD_1 10 1
postmerge_OpenTTD_1 10 1

version 1.10.1

TrueBrain:
Fixes #9281

## Motivation / Problem

Company buy-out is a bit weird. See #9281 for my analysis on the subject.

## Description

```
When you buy-out a company, you got your shares back. This is
based on company-value, which includes values for the vehicles etc.
In other words, you not only got the vehicles, but you also got
paid to get them back.

Additionally, you also got the loan of the company, but not the
money for the loan (as that is subtracted from the company-value).

Solve this by changing the rules of a buy-out: don't sell your
shares, get the loan AND the balance and get the infrastructure.
```

## Limitations

This might surprise existing players, as their current tricks to get rich fast no longer work.

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -324,8 +324,11 @@ void ChangeOwnershipOfCompanyItems(Owner old_owner, Owner new_owner)
 		Backup<CompanyID> cur_company2(_current_company, FILE_LINE);
 		const Company *c = Company::Get(old_owner);
 		for (i = 0; i < 4; i++) {
-			cur_company2.Change(c->share_owners[i]);
-			if (_current_company != INVALID_OWNER) {
+			if (c->bankrupt_value == 0 && c->share_owners[i] == new_owner) {
+				/* You are the one buying the company; so don't sell the shares back to you. */
+				Company::Get(new_owner)->share_owners[i] = COMPANY_SPECTATOR;
+			} else if (c->share_owners[i] != INVALID_OWNER) {
+				cur_company2.Change(c->share_owners[i]);
 				/* Sell the shares */
 				CommandCost res = DoCommand(0, old_owner, 0, DC_EXEC | DC_BANKRUPT, CMD_SELL_SHARE_IN_COMPANY);
 				/* Because we are in a DoCommand, we can't just execute another one and
@@ -1986,6 +1989,9 @@ static void DoAcquireCompany(Company *c)
 
 	if (c->bankrupt_value == 0) {
 		Company *owner = Company::Get(_current_company);
+
+		/* Get both the balance and the loan of the company you just bought. */
+		SubtractMoneyFromCompany(CommandCost(EXPENSES_OTHER, -c->money));
 		owner->current_loan += c->current_loan;
 	}
 


James103:
Theoretically, this could mean that you could bankrupt yourself by buying out a company that was already going to bankrupt soon, as you could take on their debt causing your company's net worth to fall below negative the max loan. Are there any real examples of that happening?

rubidium42:
I agree with not repaying shares, as you "paid" that money to get the shares from some other share holder.

The other change with getting the bank balance and repaying the loan definitely sounds right to me. However, I agree with James that we might want to make it slightly less dangerous for players. After all, if you buy a company for cheap with huge debts, then you are in real problems if your cash flow is not enough. This command would then be the only command which allows you to get into a negative bank balance.

It can be solved by adding/"subtracting" the loan and bank balance to the cost of the merger. Something along the lines of https://gist.github.com/rubidium42/6d4ce350095f2fcb4435feb3ac4e581f will prevent you from getting broke attempting to buy out a company with a lot of debt. The difference does end up in the EXPENSES_OTHER category of the finance window, so that might have some repercussions somewhere else.

TrueBrain:
> 
> It can be solved by adding/"subtracting" the loan and bank balance to the cost of the merger. Something along the lines of https://gist.github.com/rubidium42/6d4ce350095f2fcb4435feb3ac4e581f will prevent you from getting broke attempting to buy out a company with a lot of debt. The difference does end up in the EXPENSES_OTHER category of the finance window, so that might have some repercussions somewhere else.

Mostly as reminder to myself, the above patch uses the company balance too soon: a company can have shares in other companies, which have to be sold off first and can alter the company balance. It simply means there is some more refactoring to do.
I also think that a solution like this would make it really unclear if you are short in money how it got to that amount, so I think we need to split this up into two errors .. 1 about not having the money to buy the shares, and 1 about not having the money to buy off the company.
I will thinker with this some more soon :)

rubidium42:
> .... a company can have shares in other companies, which have to be sold off first and can alter the company balance...

Why do the shares in other companies need to be sold? Aren't you buying all assets, including shares the other company has?
Granted, when thinking about it the problem still holds for the shares the company you buy out has in your company. Though that quickly gets into a weird infinite feedback loop with valuations and the likes, which might be exploited in some other manner.
So maybe the "easier" solution would be that you may only buy/have shares of companies that do not have a share in you? Though possibly not really part of this PR.

TrueBrain:
> Why do the shares in other companies need to be sold? Aren't you buying all assets, including shares the other company has?
> Granted, when thinking about it the problem still holds for the shares the company you buy out has in your company. Though that quickly gets into a weird infinite feedback loop with valuations and the likes, which might be exploited in some other manner.
> So maybe the "easier" solution would be that you may only buy/have shares of companies that do not have a share in you? Though possibly not really part of this PR.

I am happy you went through these motions ;)

We have to sell the shares, as indeed otherwise you all of a sudden can have shares to your own company. Not allowing to buy shares in a company that has shares in yours is rather odd, as that would mean the best way to prevent any players from buying your company if you go bankrupt is ensuring you have 25% shares in them .. which becomes really hard to explain to players :P

The other issue is recursion: it otherwise means that buying company A can also leads you to buy company B, C, D, etc, because your total shares are now 100%. This becomes .. "difficult", especially to check if you have the money to buy all those .. and explaining that to the user will be a field-day.

So no, I like the current system, where it simply sells the shares the company that is being acquired has. Simple, efficient, a lot less explaining to the user :D So yeah, if anything, not for this PR :D

TrueBrain:
Okay, after some fiddling, I figured out this is a much deeper rabbit hole than I would like. So I went back to an easier solution.

The only thing this PR now changes is that instead of selling back your shares, you get the balance of the company you bought. This means you can get a loan over the max-loan after taking over a company (but that was already the case, so not a regression of this PR).

If we want to fix this properly, we really need to fix several UI elements. For example, when buying shares it is unclear what it will cost you or how it comes up with the number. You need to understand it is a % of the company value, but that is not clear. Next, when you buy 100%, we could use a small GUI that tells you what is going to happen: shares being sold, paying off the loan, what is left over for you. And that you can accept that or not.

Personally I have no problem with that you can buy a company that is nearly bankrupt; it might have a lot of infrastructure you plan to sell, for example.

But all this work is a lot more than the effort I am currently willing to spend on this PR, so I focused on addressing the issue presented, and that is a much smaller and easier fix :)

rubidium42 (approved):

PR #9302, opened 2021-05-27
Change: split check and update callbacks for settings

rubidium42:
## Motivation / Problem

In the settings there is currently one callback that does both check whether a change is allowed, and it updates when a change is performed. This means that one first needs to set the new value, then call that callback and when that fails the value needs to be reset again.
Similarly there is no useful callback for strings, as the callback will be called with `0` as parameter and that cannot be used.

The goal was to be able to validate the server and client name when they are entered from the console, and not be in a situation where the old name is lost once the validation of the name is happening.

## Description

Split the validation/check/pre change callback from the update/post change callback.
The pre-change callback will be of type `bool Func(std::string &new_value)` or `bool Func(int32 &new_value)` which allows the pre change to update the value if needed. For example removing/clearing a bit that should never be set, or more useful for strings trimming white space or replacing * with an empty string.
The post-change callback will be of type `void Func(const std::string &new_value)` or `void Func(int32 new_value)`.

The effect of splitting these changes is that it becomes way clearer on which settings there are actual checks, and for which there is merely a post-change callback.

As added bonus with this precheck it becomes impossible to clear the server and client name via the console or the UI. The only viable path is clearing it in the configuration, but that should only be feasible for the dedicated server and there are guards in place to replace the empty names with something.

## Limitations

None that I know of.

## Checklist for review

### DIFF
--- a/src/network/network_client.cpp
+++ b/src/network/network_client.cpp
@@ -1344,27 +1344,22 @@ bool NetworkValidateClientName()
 }
 
 /**
- * Send the server our name.
+ * Send the server our name as callback from the setting.
+ * @param newname The new client name.
  */
-void NetworkUpdateClientName()
+void NetworkUpdateClientName(const std::string &client_name)
 {
 	NetworkClientInfo *ci = NetworkClientInfo::GetByClientID(_network_own_client_id);
-
 	if (ci == nullptr) return;
-	/* There is no validation on string settings, it is actually a post change callback.
-	 * This method is called from that post change callback. So, when the client name is
-	 * changed via the console there is no easy way to prevent an invalid name. Though,
-	 * we can prevent it getting sent here. */
-	if (!NetworkValidateClientName()) return;
 
 	/* Don't change the name if it is the same as the old name */
-	if (_settings_client.network.client_name.compare(ci->client_name) != 0) {
+	if (client_name.compare(ci->client_name) != 0) {
 		if (!_network_server) {
-			MyClient::SendSetName(_settings_client.network.client_name.c_str());
+			MyClient::SendSetName(client_name.c_str());
 		} else {
 			/* Copy to a temporary buffer so no #n gets added after our name in the settings when there are duplicate names. */
 			char temporary_name[NETWORK_CLIENT_NAME_LENGTH];
-			strecpy(temporary_name, _settings_client.network.client_name.c_str(), lastof(temporary_name));
+			strecpy(temporary_name, client_name.c_str(), lastof(temporary_name));
 			if (NetworkFindName(temporary_name, lastof(temporary_name))) {
 				NetworkTextMessage(NETWORK_ACTION_NAME_CHANGE, CC_DEFAULT, false, ci->client_name, temporary_name);
 				ci->client_name = temporary_name;
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -1343,37 +1343,28 @@ static bool InvalidateShipPathCache(int32 p1)
 	return true;
 }
 
-static bool UpdateClientName(int32 p1)
+/**
+ * Replace a passwords that are a literal asterisk with an empty string.
+ * @param newval The new string value for this password field.
+ * @return Always true.
+ */
+static bool ReplaceAsteriskWithEmptyPassword(std::string &newval)
 {
-	NetworkUpdateClientName();
-	return true;
-}
-
-static bool UpdateServerPassword(int32 p1)
-{
-	if (_settings_client.network.server_password.compare("*") == 0) {
-		_settings_client.network.server_password.clear();
-	}
-
-	NetworkServerUpdateGameInfo();
-	return true;
-}
-
-static bool UpdateRconPassword(int32 p1)
-{
-	if (_settings_client.network.rcon_password.compare("*") == 0) {
-		_settings_client.network.rcon_password.clear();
-	}
-
+	if (newval.compare("*") == 0) newval.clear();
 	return true;
 }
 
 static bool UpdateClientConfigValues(int32 p1)
 {
+	UpdateClientConfigValues();
+	return true;
+}
+
+/** Update the game info, and send it to the clients when we are running as a server. */
+static void UpdateClientConfigValues()
+{
 	NetworkServerUpdateGameInfo();
 	if (_network_server) NetworkServerSendConfigUpdate();
-
-	return true;
 }
 
 /* End - Callback Functions */
@@ -2069,8 +2060,10 @@ bool SetSettingValue(const StringSettingDesc *sd, std::string value, bool force_
 void StringSettingDesc::ChangeValue(const void *object, std::string &newval) const
 {
 	this->MakeValueValid(newval);
+	if (this->pre_check != nullptr && !this->pre_check(newval)) return;
+
 	this->Write(object, newval);
-	if (this->proc != nullptr) this->proc(0);
+	if (this->post_callback != nullptr) this->post_callback(newval);
 
 	if (_save_config) SaveToConfig();
 }
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -807,99 +807,56 @@ const StringSettingDesc *SettingDesc::AsStringSetting() const
 /* Begin - Callback Functions for the various settings. */
 
 /** Reposition the main toolbar as the setting changed. */
-static bool v_PositionMainToolbar(int32 p1)
+static void v_PositionMainToolbar(int32 new_value)
 {
 	if (_game_mode != GM_MENU) PositionMainToolbar(nullptr);
-	return true;
 }
 
 /** Reposition the statusbar as the setting changed. */
-static bool v_PositionStatusbar(int32 p1)
+static void v_PositionStatusbar(int32 new_value)
 {
 	if (_game_mode != GM_MENU) {
 		PositionStatusbar(nullptr);
 		PositionNewsMessage(nullptr);
 		PositionNetworkChatWindow(nullptr);
 	}
-	return true;
-}
-
-static bool PopulationInLabelActive(int32 p1)
-{
-	UpdateAllTownVirtCoords();
-	return true;
-}
-
-static bool RedrawScreen(int32 p1)
-{
-	MarkWholeScreenDirty();
-	return true;
 }
 
 /**
  * Redraw the smallmap after a colour scheme change.
  * @param p1 Callback parameter.
- * @return Always true.
  */
-static bool RedrawSmallmap(int32 p1)
+static void RedrawSmallmap(int32 new_value)
 {
 	BuildLandLegend();
 	BuildOwnerLegend();
 	SetWindowClassesDirty(WC_SMALLMAP);
-	return true;
 }
 
-static bool InvalidateDetailsWindow(int32 p1)
-{
-	SetWindowClassesDirty(WC_VEHICLE_DETAILS);
-	return true;
-}
-
-static bool StationSpreadChanged(int32 p1)
+static void StationSpreadChanged(int32 p1)
 {
 	InvalidateWindowData(WC_SELECT_STATION, 0);
 	InvalidateWindowData(WC_BUILD_STATION, 0);
-	return true;
 }
 
-static bool InvalidateBuildIndustryWindow(int32 p1)
+static void CloseSignalGUI(int32 new_value)
 {
-	InvalidateWindowData(WC_BUILD_INDUSTRY, 0);
-	return true;
-}
-
-static bool CloseSignalGUI(int32 p1)
-{
-	if (p1 == 0) {
+	if (new_value == 0) {
 		DeleteWindowByClass(WC_BUILD_SIGNAL);
 	}
-	return true;
 }
 
-static bool InvalidateTownViewWindow(int32 p1)
-{
-	InvalidateWindowClassesData(WC_TOWN_VIEW, p1);
-	return true;
-}
-
-static bool DeleteSelectStationWindow(int32 p1)
-{
-	DeleteWindowById(WC_SELECT_STATION, 0);
-	return true;
-}
-
-static bool UpdateConsists(int32 p1)
+static void UpdateConsists(int32 new_value)
 {
 	for (Train *t : Train::Iterate()) {
 		/* Update the consist of all trains so the maximum speed is set correctly. */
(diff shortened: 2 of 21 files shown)


src/settings.cpp
```diff
@@ -2026,12 +2021,12 @@ uint GetCompanySettingIndex(const char *name)
...
  * @note Strings WILL NOT be synced over the network
  */
-bool SetSettingValue(const StringSettingDesc *sd, const char *value, bool force_newgame)
+bool SetSettingValue(const StringSettingDesc *sd, std::string value, bool force_newgame)
 {
 	assert(sd->save.conv & SLF_NO_NETWORK_SYNC);
 
-	if (GetVarMemType(sd->save.conv) == SLE_VAR_STRQ && strcmp(value, "(null)") == 0) {
-		value = nullptr;
+	if (GetVarMemType(sd->save.conv) == SLE_VAR_STRQ && value.compare("(null)") == 0) {
```
TrueBrain, line 1893: just curious: `value == "(null)"` doesn't work for std::string?

src/settings.cpp
```diff
@@ -460,34 +460,28 @@ int32 IntSettingDesc::Read(const void *object) const
 }
 
 /**
- * Set the string value of a setting.
- * @param object The object the setting is to be saved in.
- * @param str The string to save.
+ * Validate the value of a setting.
```
TrueBrain: You cannot commit on commit messages in GitHub, so just a random place, sorry:

`Codechange: add helper functions to write an string setting value`

Is it me, or is this more like "Splitting Write_ValidateSetting into two functions" kinda thing?

btw, it is `a string`, not `an string` :)

src/settings.cpp
```diff
@@ -398,8 +398,16 @@ size_t BoolSettingDesc::ParseValue(const char *str) const
  */
 void IntSettingDesc::Write_ValidateSetting(const void *object, int32 val) const
 {
-	void *ptr = GetVariableAddress(object, &this->save);
+	this->Validate(val);
```
TrueBrain: Why do you here choose to keep `Write_ValidateSetting` and in the commit before replace all instances with a call to these two functions?

src/network/network_client.cpp
```diff
@@ -1344,27 +1344,22 @@ bool NetworkValidateClientName()
...
 	if (ci == nullptr) return;
-	/* There is no validation on string settings, it is actually a post change callback.
-	 * This method is called from that post change callback. So, when the client name is
-	 * changed via the console there is no easy way to prevent an invalid name. Though,
-	 * we can prevent it getting sent here. */
-	if (!NetworkValidateClientName()) return;
 
 	/* Don't change the name if it is the same as the old name */
-	if (_settings_client.network.client_name.compare(ci->client_name) != 0) {
+	if (client_name.compare(ci->client_name) != 0) {
```
TrueBrain, line 1356: Similar as earlier: `client_name != ci->client_name`? Or does that do something else?

I see more of these cases throughout this PR, so will await the answer before I mark the other places :D

src/settings.cpp
```diff
@@ -2060,8 +2041,10 @@ bool SetSettingValue(const StringSettingDesc *sd, std::string value, bool force_
 void StringSettingDesc::ChangeValue(const void *object, std::string &newval) const
 {
 	this->Validate(newval);
+	if (this->pre_check != nullptr && !this->pre_check(newval)) return;
```
TrueBrain, line 1911: Should the internal `Validate` happen before or after the `pre_check`? I ask as your doxygen comment says `pre_check` can change the value, possibly making it fail `Validate`? Not sure which one is more important :)

src/table/settings.ini
```diff
@@ -5,11 +5,12 @@
 ;
 
 [pre-amble]
-/* Begin - Callback Functions for the various settings */
+/* Begin - Callback Functions for the various settings; single line wrapper implementations are okay. */
```
TrueBrain: Is this really a good call? It now also lacks doxygen comments, and adding them here makes this really noisy? (honest question, to be clear :D)

src/table/settings.ini
```diff
@@ -5,11 +5,12 @@
...
 static bool v_PositionMainToolbar(int32 p1);
 static bool v_PositionStatusbar(int32 p1);
 static bool PopulationInLabelActive(int32 p1);
 static bool RedrawScreen(int32 p1);
+static void RedrawScreen(const std::string &newval) { RedrawScreen(0); }
```
TrueBrain: This looks weird, but `RedrawScreen` ofc ignores `p1`. I do wonder if we should comment on that or not? Similar with `UpdateClientConfigValues`.

src/settings.cpp
```diff
@@ -991,30 +924,24 @@ static bool TrainAccelerationModelChanged(int32 p1)
...
 	SetWindowClassesDirty(WC_VEHICLE_DETAILS);
-
-	return true;
 }
 
 /**
  * This function updates the train acceleration cache after a steepness change.
- * @param p1 Callback parameter.
- * @return Always true.
+ * @param new_value Callback parameter.
```
TrueBrain: Not sure if you want to do this in this PR, but doxygen comments on these callbacks are a bit touch-and-go. Some don't have them, some don't mention `new_value`, and some do.

Either way, maybe also fix `Callback parameter` to be a bit more descriptive like `New value of the setting` or what-ever :)

src/settings.cpp
```diff
@@ -1188,15 +1045,14 @@ static bool MaxNoAIsChange(int32 i)
...
 }
 
 /**
  * Check whether the road side may be changed.
- * @param p1 unused
+ * @param new_value unused
  * @return true if the road side may be changed.
  */
-static bool CheckRoadSide(int p1)
+static bool CheckRoadSide(int &new_value)
```
TrueBrain: Sometimes the signature changes :D (was already the case in the old code ofc). But `int` should be `int32` I would think?

src/settings.cpp
```diff
@@ -1080,81 +988,32 @@ static bool ZoomMinMaxChanged(int32 p1)
...
 /**
  * Update any possible saveload window and delete any newgrf dialogue as
  * its widget parts might change. Reinit all windows as it allows access to the
  * newgrf debug button.
- * @param p1 unused.
- * @return Always true.
+ * @param new_value unused.
  */
-static bool InvalidateNewGRFChangeWindows(int32 p1)
+static void InvalidateNewGRFChangeWindows(int32 new_value)
```
TrueBrain, line 1015: I do wonder, and might be out of scope of this PR .. all these callbacks are now "somewhere" in the code, most in `settings.cpp`, but not all .. shouldn't we just put them in `settings_callbacks.cpp` or something? Just a random thought.

src/table/settings.ini
```diff
@@ -6,47 +6,49 @@
...
+static void InvalidateTownViewWindow(int32 new_value) { InvalidateWindowClassesData(WC_TOWN_VIEW, new_value); }
+static void DeleteSelectStationWindow(int32 new_value) { DeleteWindowById(WC_SELECT_STATION, 0); }
+static void UpdateConsists(int32 new_value);
+static void TrainAccelerationModelChanged(int32 new_value);
+static void RoadVehAccelerationModelChanged(int32 new_value);
+static void TrainSlopeSteepnessChanged(int32 new_value);
+static void RoadVehSlopeSteepnessChanged(int32 new_value);
+static void DragSignalsDensityChanged(int32 new_value) { InvalidateWindowData(WC_BUILD_SIGNAL, 0); }
+static void TownFoundingChanged(int32 new_value);
+static void DifficultyNoiseChange(int32 new_value);
```
TrueBrain, line 20: The more I look at this (ever growing and by now unreadable) list, the more I wonder if we can't do better. Maybe moving it to a header-file would already help out .. dunno. But now with all these extra (inline) entries, and I fully understand why you did that btw, it is now .. bloaty :D

Not saying you should do anything with it, just remarking it feels off

src/table/settings.ini
```diff
@@ -6,47 +6,49 @@
...
+static void InvalidateNewGRFChangeWindows(int32 new_value);
+static void InvalidateIndustryViewWindow(int32 new_value) { InvalidateWindowClassesData(WC_INDUSTRY_VIEW); }
+static void InvalidateAISettingsWindow(int32 new_value) { InvalidateWindowClassesData(WC_AI_SETTINGS); }
+static void RedrawTownAuthority(int32 new_value) { SetWindowClassesDirty(WC_TOWN_AUTHORITY); }
+static void InvalidateCompanyInfrastructureWindow(int32 new_value) { InvalidateWindowClassesData(WC_COMPANY_INFRASTRUCTURE); }
+static void InvalidateCompanyWindow(int32 new_value) { InvalidateWindowClassesData(WC_COMPANY); }
+static void ZoomMinMaxChanged(int32 new_value);
+static void SpriteZoomMinChanged(int32 new_value);
+static void MaxVehiclesChanged(int32 new_value);
+static void InvalidateShipPathCache(int32 new_value);
```
TrueBrain, line 33: one thing I do not like one bit with this pre/post, is that it is very difficult to spot what functions are pre, and which are post. It is in the nuance:
For ints, it is a single `&`.
For strings, it is a `const` marker.
This is really hard to read. And the naming being all over the place doesn't help. Maybe if we already order it based on pre/post, that could help, but not sure.

Again not sure if that is something for this PR, but this PR does make it worse than it was :) Some kind of naming convention might seriously help here.

Just some examples of what frustrates:
`MaxNoAIsChange` vs `ZoomMinMaxChanged` vs `UpdateMaxHeightLevel`
`CanUpdateServiceInterval` vs `CheckDynamicEngines` vs `ReplaceAsteriskWithEmptyPassword`

It is often not even clear they are settings callbacks, and feel like anything can use them (which might even be true).

TrueBrain (commented):
I fully understand some comments might be out-of-scope of this PR, but they also might be easy to fix while at it. I leave that up to you :)

src/settings.cpp
rubidium42: It would work. Though I have noticed in myself that I make quite a few errors changing these from `compare` to `==`, so I rather do the safe thing in changing this code.

src/settings.cpp
rubidium42: As there are a few places where it is still called. For example in old save game conversion where you do want the new setting to be at least in the valid range, or in LoadSettings where it is used to validate the data in the save game.

src/network/network_client.cpp
rubidium42: It would function the same, but this change is (now) easier for me.

src/settings.cpp
rubidium42: In my opinion this is the order to use, as I would assume that the **check** for a setting does not make the setting invalid. Similarly the `Validate` can make what has been checked invalid. Take for example the setting "display_opt". That uses a bitmask with values from DisplayOptions which does not have a value for bit 4. So practically the check could just remove bit 4 from the setting, but if the value would be 256 then the pre check would not remove bit 4 but the validation clamps it to 255 which has bit 4 set.

src/table/settings.ini
rubidium42: Well, they were (mostly) already undocumented. Furthermore declaring it here and then in another place provide the implementation adds another 5 lines without making it any clearer.

To be honest, all there one liner functions here are just pointless in any case. I have attempted to use lambdas to get rid of them, and just use something like `[](newval) { MarkWholeScreenDirty(); }` in the configurations below, as that is probably clearer over all but I failed.

I can remove the implementation from these lines and add them back to settings.cpp, but I'm not going to document them all. That's out of scope for this PR.

src/table/settings.ini
rubidium42: Good question.

src/settings.cpp
rubidium42: The tables are loaded/declared from this C++ file, so putting the callback in a separate file means also adding them to a header. Which would mean moving everything from the preamble to some header.
Or define the settings tables to be not static, but then everything outside of this C++ file might get tempted to mess with them directly.

src/table/settings.ini
rubidium42: I found it bloaty where it was right now. Defining it here and then declaring it later.

src/table/settings.ini
rubidium42: The easiest way to distinguish between 'pre' and 'post' is by the return parameter. When it is a `bool`, then it is a pre check callback. When it is a `void` it is a post update callback. Previously you had no clue at all, so that's already definitely an improvement ;)

I agree the naming is inconsistent, though changing naming and/or moving things around makes reviewing more difficult.

src/settings.cpp
TrueBrain: what do you mean with "quite a few errors"? What is the problem that occurs?

src/settings.cpp
TrueBrain: I was thinking the other way around: your name doesn't fit, so `Validate` fails, but it is mostly full of whitespace, so `pre_check` would remove that :P I am fine either way, was just making sure there was a reason you went for this order.

src/table/settings.ini
TrueBrain: In general we have the rule: document what you add, so that might work here too. But I agree that most of these are bullshit functions anyway. Lambdas would be nice .. but yeah, for another PR :)

src/settings.cpp
TrueBrain: I think I mentioned later to have a header file for them too yeah .. but not for this PR :)

src/table/settings.ini
TrueBrain: As said, doesn't have to be for this PR, but might be a good cleanup to do somewhere soon. Similar with the other related comments .. I also noticed I find it rather difficult to add stuff because of this.

That said, in general, the settings.ini is messy .. very hard to find anything, as it is just an endless list of shit .. but one thing at the time :)

src/settings.cpp
```diff
@@ -460,34 +460,28 @@ int32 IntSettingDesc::Read(const void *object) const
 }
 
 /**
- * Set the string value of a setting.
- * @param object The object the setting is to be saved in.
- * @param str The string to save.
+ * Validate the value of a setting.
+ * @param str The string to validate.
```
TrueBrain: As by IRC conversation, this might not be the most clear name for what this does. At least we need to extend the doxygen comment to reflect it changes the value.

TrueBrain (approved):
"It seems fine" :D

PR #9304, opened 2021-05-29
Codechange: rename str_validate to StrMakeValid(InPlace)

TrueBrain:
## Motivation / Problem

One too many times I fall into the trap that `str_validate` changes the string to become valid, instead of telling me if it is valid. So let's stop with annoying the developer, and use a name that is more explicit.

## Description

`ValidateString` used to do `str + strlen(str) + 1`. I do not know why it was doing the `+ 1`, as that always puts it outside the string. So while at it, I changed that to `str + strlen(str)` and changed all existing manual `str + strlen(str)` into using the wrapper function.

## Checklist for review

### DIFF
--- a/src/script/api/script_object.cpp
+++ b/src/script/api/script_object.cpp
@@ -283,7 +283,7 @@ ScriptObject::ActiveInstance::~ActiveInstance()
 {
 	char buffer[64];
 	::GetString(buffer, string, lastof(buffer));
-	::str_validate(buffer, lastof(buffer), SVS_NONE);
+	::StrMakeValidInPlace(buffer, lastof(buffer), SVS_NONE);
 	return ::stredup(buffer);
 }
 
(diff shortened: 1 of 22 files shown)


src/script/api/script_object.cpp
```diff
@@ -312,7 +312,7 @@ ScriptObject::ActiveInstance::~ActiveInstance()
 	if (!StrEmpty(text) && (GetCommandFlags(cmd) & CMD_STR_CTRL) == 0) {
 		/* The string must be valid, i.e. not contain special codes. Since some
 		 * can be made with GSText, make sure the control codes are removed. */
-		::str_validate(const_cast<char *>(text), text + strlen(text), SVS_NONE);
+		::StrMakeValidInPlace(const_cast<char *>(text), SVS_NONE);
```
rubidium42: The `const_cast<char *>` is not needed anymore, since the `StrMakeValidInPlace` without `last`'s `str` parameter is `const char *`.

src/string.cpp
```diff
@@ -246,49 +246,48 @@ static void str_validate(T &dst, const char *str, const char *last, StringValida
...
 	*dst = '\0';
 }
 
 /**
- * Scans the string for valid characters and if it finds invalid ones,
- * replaces them with a question mark '?' (if not ignored)
- * @param str the string to validate
- * @param settings the settings for the string validation.
+ * Scans the string for invalid characters and replaces then with a
+ * question mark '?' (if not ignored).
```
rubidium42, line 264: Maybe mention in this comment that this variant must only be called when you are certain the string in '\0' terminated, and that otherwise the other variant must be used instead.

rubidium42 (approved):

PR #9306, opened 2021-05-29
Codechange: use setting name instead of index for CmdChange(Company)Setting

TrueBrain:
## Motivation / Problem

I am working towards splitting up `settings.ini` into several files. But I find these gems that requires addressing first.

## Description

These two especially I wanted to address:

- There are now constraints on settings.ini you might not expected. For example, conditional settings always have to come last, as otherwise they would influence the index.
- Adding additional SettingTables to read from would mean several functions need to be aware of this. Now a far few less.

And as bonus:
- The code becomes a lot simpler because of this.

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -104,17 +104,6 @@ const SettingDesc *GetSettingDescription(uint index)
 }
 
 /**
- * Get the setting at the given index into the company settings table.
- * @param index The index to look for.
- * @return The setting at the given index, or nullptr when the index is invalid.
- */
-static const SettingDesc *GetCompanySettingDescription(uint index)
-{
-	if (index >= _company_settings.size()) return nullptr;
-	return _company_settings.begin()[index].get();
-}
-
-/**
  * Groups in openttd.cfg that are actually lists.
  */
 static const char * const _list_group_names[] = {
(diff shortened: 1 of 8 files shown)


src/settings.cpp
```diff
@@ -1709,20 +1698,65 @@ void IntSettingDesc::ChangeValue(const void *object, int32 newval) const
...
  * @param p2 the new value for the setting
  * The new value is properly clamped to its minimum/maximum when setting
- * @param text unused
+ * @param text the name of the setting to change
  * @return the cost of this operation or an error
  * @see _settings
  */
 CommandCost CmdChangeSetting(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const char *text)
 {
-	const SettingDesc *sd = GetSettingDescription(p1);
```
rubidium42, line 1725: Is text guaranteed to be non-`nullptr`? If not, `strcmp` on `nullptr` is undefined behaviour. 
Yes, from how we call it, but I'm rather thinking about a bad actor trying to mess things up.

src/settings.cpp
```diff
@@ -1741,15 +1775,16 @@ CommandCost CmdChangeSetting(TileIndex tile, DoCommandFlag flags, uint32 p1, uin
...
+ * @param p1 unused
  * @param p2 the new value for the setting
  * The new value is properly clamped to its minimum/maximum when setting
- * @param text unused
+ * @param text the name of the company setting to change
  * @return the cost of this operation or an error
  */
 CommandCost CmdChangeCompanySetting(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const char *text)
 {
-	const SettingDesc *sd = GetCompanySettingDescription(p1);
```
rubidium42, line 1752: Same comment as with CmdChangeSetting.

src/settings.cpp
```diff
@@ -1709,20 +1698,65 @@ void IntSettingDesc::ChangeValue(const void *object, int32 newval) const
...
+	for (auto &sd : _settings) {
+		if (!SlIsObjectCurrentlyValid(sd->save.version_from, sd->save.version_to)) continue;
+		const char *short_name = strchr(sd->name, '.');
+		if (short_name != nullptr) {
+			short_name++;
+			if (strcmp(short_name, name) == 0) return sd.get();
+		}
+	}
+
+	if (strncmp(name, "company.", 8) == 0) name += 8;
```
rubidium42, line 1737: This should go in the `GetCompanySettingFromName` function. It is only relevant for company settings, but more importantly the calls to DoCommand use `company.engine_renew` and so on.

rubidium42 (approved):

PR #9307, opened 2021-05-29
Fix: artificially limit heightmap sizes to something reasonable

rubidium42:
## Motivation / Problem

LGTM:
`if ((uint64)width * height >= (size_t)-1) {`
Comparison is always false because ... * ... <= 18446744065119617024.

Considering 18446744065119617023 a valid size is well outside of the realm of sanity for any of our players computers. So, I added some more sane limits.

If someone pushes a PNG or BMP with just under size_t dimensions into BaNaNaS and anyone trying to load that heightmap, it will crash their game when trying to allocate memory. After all, make the size so you want to allocate slightly less than (size_t)-1 and that allocation will fail as there is guaranteed not enough memory, as OpenTTD will already be using more than 1 byte.

## Description

Introduce a limit on the maximum side length of an heightmap image to be at most 2 times `MAX_MAP_SIZE` (so currently 8192), and a limit on the maximum amount of bytes to allocated for a heightmap set at 1 gigabyte.
With a 24bpp BMP that 1 GB allocation limit would result in an image with ~256 million pixels, so that limit won't be practically reached yet with 8192 maximum side lengths. However, when the maximum map size is changed, or at least the maximum size along one edge (such as in JGRPP) the size length limit becomes larger and this gets into play.

## Limitations

See description; you can't load gigapixel size height maps anymore.

## Checklist for review

### DIFF
--- a/src/heightmap.cpp
+++ b/src/heightmap.cpp
@@ -23,6 +23,40 @@
 #include "safeguards.h"
 
 /**
+ * Maximum number of pixels for one dimension of a heightmap image.
+ * Do not allow images for which the longest side is twice the maximum number of
+ * tiles along the longest side of the (tile) map.
+ */
+static const uint MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS = 2 * MAX_MAP_SIZE;
+
+/*
+ * Maximum size in pixels of the heightmap image.
+ */
+static const uint MAX_HEIGHTMAP_SIZE_PIXELS = 256 << 20; // ~256 million
+/*
+ * When loading a PNG or BMP the 24 bpp variant requires at least 4 bytes per pixel
+ * of memory to load the data. Make sure the "reasonable" limit is well within the
+ * maximum amount of memory allocatable on 32 bit platforms.
+ */
+static_assert(MAX_HEIGHTMAP_SIZE_PIXELS < UINT32_MAX / 8);
+
+/**
+ * Check whether the loaded dimension of the heightmap image are considered valid enough
+ * to attempt to load the image. In other words, the width and height are not beyond the
+ * #MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS limit and the total number of pixels does not
+ * exceed #MAX_HEIGHTMAP_SIZE_PIXELS. A width or height less than 1 are disallowed too.
+ * @param width The width of the to be loaded height map.
+ * @param height The height of the to be loaded height map.
+ * @return True iff the dimensions are within the limits.
+ */
+static inline bool IsValidHeightmapDimension(size_t width, size_t height)
+{
+	return (uint64)width * height <= MAX_HEIGHTMAP_SIZE_PIXELS &&
+		width > 0 && width <= MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS &&
+		height > 0 && height <= MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS;
+}
+
+/**
  * Convert RGB colours to Grayscale using 29.9% Red, 58.7% Green, 11.4% Blue
  *  (average luminosity formula, NTSC Colour Space)
  */
@@ -146,8 +180,7 @@ static bool ReadHeightmapPNG(const char *filename, uint *x, uint *y, byte **map)
 	uint width = png_get_image_width(png_ptr, info_ptr);
 	uint height = png_get_image_height(png_ptr, info_ptr);
 
-	/* Check if image dimensions don't overflow a size_t to avoid memory corruption. */
-	if ((uint64)width * height >= (size_t)-1) {
+	if (!IsValidHeightmapDimension(width, height)) {
 		ShowErrorMessage(STR_ERROR_PNGMAP, STR_ERROR_HEIGHTMAP_TOO_LARGE, WL_ERROR);
 		fclose(fp);
 		png_destroy_read_struct(&png_ptr, &info_ptr, nullptr);
@@ -255,8 +288,7 @@ static bool ReadHeightmapBMP(const char *filename, uint *x, uint *y, byte **map)
 		return false;
 	}
 
-	/* Check if image dimensions don't overflow a size_t to avoid memory corruption. */
-	if ((uint64)info.width * info.height >= (size_t)-1 / (info.bpp == 24 ? 3 : 1)) {
+	if (!IsValidHeightmapDimension(info.width, info.height)) {
 		ShowErrorMessage(STR_ERROR_BMPMAP, STR_ERROR_HEIGHTMAP_TOO_LARGE, WL_ERROR);
 		fclose(f);
 		BmpDestroyData(&data);
@@ -295,6 +327,8 @@ static void GrayscaleToMapHeights(uint img_width, uint img_height, byte *map)
 {
 	/* Defines the detail of the aspect ratio (to avoid doubles) */
 	const uint num_div = 16384;
+	/* Ensure multiplication with num_div does not cause overflows. */
+	static_assert(num_div <= std::numeric_limits<uint>::max() / MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS);
 
 	uint width, height;
 	uint row, col;


James103:
> See description; you can't load gigapixel size height maps anymore

A workaround is to use an external image processing software (such as GIMP or Paint.net) to scale/resize and crop the input height map to whatever playing area you want and at the size you want.

src/heightmap.cpp
```diff
@@ -255,8 +295,8 @@ static bool ReadHeightmapBMP(const char *filename, uint *x, uint *y, byte **map)
 		return false;
 	}
 
-	/* Check if image dimensions don't overflow a size_t to avoid memory corruption. */
-	if ((uint64)info.width * info.height >= (size_t)-1 / (info.bpp == 24 ? 3 : 1)) {
+	/* 1 byte per pixel for the map, and 3 or 1 bytes per pixel for loading the BMP data. */
```
TrueBrain: What do you mean "for the map"?

src/heightmap.cpp
```diff
@@ -22,6 +22,47 @@
 
 #include "safeguards.h"
 
+/**
+ * Maximum number of pixels for one dimension of a heightmap image.
+ *
+ * At some point the number of pixels in a heightmap becomes larger compared to
+ * the pixels that will be taken for the map. With this limit already 75% of the
+ * pixels would be discarded for the largest image and the largest map size.
```
TrueBrain: I read this 5 times now, and I am still not sure what it says.

"that will be taken for the map" is not something I understand :(

I also don't understand what "75% of the pixels would be discarded" means.

Sorry ..

src/heightmap.cpp
```diff
@@ -22,6 +22,47 @@
...
+/*
+ * Maximum size, in bytes, of the heightmap to consider loading.
+ *
+ * The theoretical maximum size of an image is about 16 exabytes for PNG and
+ * about 4 exabytes for a BMP image. On a 32 bit platform this will overflow
+ * size_t which could lead to memory corruption.
+ * Even on a 64 bit platform that is way larger than feasible; we could allow it
+ * on those platforms, but then the memory allocation that is done will almost
+ * certainly fail causing a crash to the desktop instead of a nice user error.
+ * So, artificially limit the maximum size of a hightmap image to allocate one
```
TrueBrain: `hightmap` missing an `e`

src/heightmap.cpp
```diff
@@ -22,6 +22,47 @@
...
+ * gigabyte here.
+ */
+static const uint MAX_HEIGHTMAP_SIZE_IN_BYTES = 1 << 30;
+
+/**
+ * Check whether the loaded dimension of the heightmap is considered valid enough
+ * to attempt to load. In other words, the width and height are not beyond the
+ * #MAX_HEIGHTMAP_SIDE_LENGTH_IN_PIXELS limit and the number of bytes that would
+ * be allocated for the table does not exceed MAX_HEIGHTMAP_SIZE_IN_BYTES.
```
TrueBrain: `table`? What table?

src/heightmap.cpp
rubidium42: The `byte **map` parameter of this function that is MallocT-ed about 20 lines below here.

src/heightmap.cpp
rubidium42: Good point. The comment started out somewhere else, so now `table` does indeed not make much sense anymore. It is the (height)`map` table that gets filled in the ReadHeightmap* functions.

src/heightmap.cpp
rubidium42: If the image is 8192x8192 pixels, and the size of the map is 4096x4096 then only 25% of the pixels of the image end up being used for the height of the map. The other 75% are discarded.
I'll try to figure out some better wording tomorrow when I've hopefully have somewhat forgotten what I have written, so it becomes clearer what is unclear to others. But I guess map being used interchangeably for the map-array and height-map, together with a parameter `map` in the ReadHeightmap* functions will cause a lot of confusion and headaches.

src/heightmap.cpp
TrueBrain: Okay, so this does a bit more than I expected.

Basically, for PNGs you only say: it allocated the "map" only, so as long as that is in bounds, I am fine. You forgive the PNG buffer basically.
For BMPs, you say: we will also allocated based on bpp, so that also has to fit in this value.

Do I get that right?

src/heightmap.cpp
```diff
@@ -22,6 +22,47 @@
 
 #include "safeguards.h"
 
+/**
+ * Maximum number of pixels for one dimension of a heightmap image.
+ *
+ * At some point the number of pixels in a heightmap becomes larger compared to
+ * the pixels that will be taken for the map. With this limit already 75% of the
+ * pixels would be discarded for the largest image and the largest map size.
+ */
```
TrueBrain, line 29: `Be reasonable in how much bigger the heightmap can be compared to the biggest map size.`

That alone for me would be sufficient, I think :) Also, I think times two is a bit low .. maybe times 4?

src/heightmap.cpp
```diff
@@ -22,6 +22,47 @@
...
+ * The theoretical maximum size of an image is about 16 exabytes for PNG and
+ * about 4 exabytes for a BMP image. On a 32 bit platform this will overflow
+ * size_t which could lead to memory corruption.
+ * Even on a 64 bit platform that is way larger than feasible; we could allow it
+ * on those platforms, but then the memory allocation that is done will almost
+ * certainly fail causing a crash to the desktop instead of a nice user error.
+ * So, artificially limit the maximum size of a hightmap image to allocate one
+ * gigabyte here.
+ */
+static const uint MAX_HEIGHTMAP_SIZE_IN_BYTES = 1 << 30;
```
TrueBrain: `Ensure memory is within reason while loading a heightmap.`

Not sure if the rest really added value for me. It kinda only confused, as it sounded like it is much more than it really is, if I understand you correctly now.

Some more thinking, basically this comment tells me why it isn't a really big number. It doesn't tell me anything why this (relative) small number. This really confused me, as I was looking for a link; but this was just your train of thought :) I think that is fine in a PR description, but in a comment it really doesn't add much. I guess it really is a "1GB is enough for everyone" comment :)

src/heightmap.cpp
```diff
@@ -146,8 +187,7 @@ static bool ReadHeightmapPNG(const char *filename, uint *x, uint *y, byte **map)
 	uint width = png_get_image_width(png_ptr, info_ptr);
 	uint height = png_get_image_height(png_ptr, info_ptr);
 
-	/* Check if image dimensions don't overflow a size_t to avoid memory corruption. */
-	if ((uint64)width * height >= (size_t)-1) {
+	if (!IsValidHeightmapDimension(width, height)) {
```
TrueBrain, line 183: I think `bytes_per_pixel = 1` is underestimating the usage for PNGs. Maybe default to 3, to get it a bit more realistic?

src/heightmap.cpp
TrueBrain: Checked all heightmaps on BaNaNaS. Biggest is 8kx8k. So anything equal or above that is fine from that point of view :)

src/heightmap.cpp
rubidium42: I've just removed the logic for "bytes" and limited pixels to 256M instead of potentially allowing 512M when using grayscale.

src/heightmap.cpp
rubidium42: Yeah, mostly because BMP allocation goes via MallocT and thus "crashes" the game. Whereas PNG allocation goes via malloc and just returns a PNG error. I have removed the distinction for number of bytes to allocate though.

TrueBrain (approved):

src/heightmap.cpp
rubidium42: More like: "why are you attempting to load a more than 256 million pixel image as a heightmap? You would be discarding a lot of pixels in the original image!"

The maximum number of tiles for vanilla us is 16 million. Since the mapping from image heightmap to tile map iterates over the tiles and then gets the height data from the associated image pixel, any pixel that is not mapped from a tile will not even be used. So, that would mean that more than 95% of the pixels will only be loaded but never used in creating the heights of the tile map.
Practically the new current side length limit of 8kx8k would be 64 million pixels would not need the 256M limit. However, the code that existed was explicitly there to prevent overflows when loading really large heightmaps, so if in the future we were to increase the maximum number of tiles along an edge, that overflowing issue might return. As such the limit on the total number of pixels. That limit would already be used by JGRPP, so it's not actually pointless right now.

PR #9308, opened 2021-05-29
Codechange: move casting a "const char *" to "char *" to the caller

TrueBrain:
## Motivation / Problem

With #9306 I found out that the fact that `StrMakeValidInPlace` says `str` is a `const char*`, but really is changing the string, can have some nasty side-effects.

## Description

Play nice, don't pretend we are not going to change something we are. Let the caller figure out if casting from `const char*` to `char *` is a safe thing to do.

In one case (`ScriptObject::DoCommand`) it would result in very ugly code to let the caller do it, so instead duplicate the value and make the changes if needed.

## Limitations

I really wonder if some of these casts are in fact safe. But this code has been running for years now, so I guess it is?

## Checklist for review

### DIFF
--- a/src/script/api/script_object.cpp
+++ b/src/script/api/script_object.cpp
@@ -309,10 +309,11 @@ ScriptObject::ActiveInstance::~ActiveInstance()
 		return false;
 	}
 
-	if (!StrEmpty(text) && (GetCommandFlags(cmd) & CMD_STR_CTRL) == 0) {
+	std::string command_text = text == nullptr ? std::string{} : text;
+	if (!command_text.empty() && (GetCommandFlags(cmd) & CMD_STR_CTRL) == 0) {
 		/* The string must be valid, i.e. not contain special codes. Since some
 		 * can be made with GSText, make sure the control codes are removed. */
-		::StrMakeValidInPlace(text, SVS_NONE);
+		command_text = ::StrMakeValid(command_text, SVS_NONE);
 	}
 
 	/* Set the default callback to return a true/false result of the DoCommand */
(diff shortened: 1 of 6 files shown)


src/string.cpp
```diff
@@ -266,7 +266,7 @@ void StrMakeValidInPlace(char *str, const char *last, StringValidationSettings s
  * otherwise use StrMakeValidInPlace(str, last, settings) variant.
  * @param str The string (of which you are sure ends with '\0') to validate.
  */
-void StrMakeValidInPlace(const char *str, StringValidationSettings settings)
+void StrMakeValidInPlace(char *str, StringValidationSettings settings)
 {
 	/* We know it is '\0' terminated. */
 	StrMakeValidInPlace(const_cast<char *>(str), str + strlen(str), settings);
```
frosch123: this cast is no longer needed?

src/script/api/script_object.cpp
```diff
@@ -309,10 +309,15 @@ ScriptObject::ActiveInstance::~ActiveInstance()
 		return false;
 	}
 
-	if (!StrEmpty(text) && (GetCommandFlags(cmd) & CMD_STR_CTRL) == 0) {
+	/* We might modify the text a bit later, so make a copy first. */
+	std::string command_text = {};
```
frosch123: The "= {}" is not needed, and it annoys me :) sorry

src/script/api/script_object.cpp
```diff
@@ -328,7 +333,7 @@ ScriptObject::ActiveInstance::~ActiveInstance()
 	if (!estimate_only && _networking && !_generating_world) SetLastCommand(tile, p1, p2, cmd);
 
 	/* Try to perform the command. */
-	CommandCost res = ::DoCommandPInternal(tile, p1, p2, cmd, (_networking && !_generating_world) ? ScriptObject::GetActiveInstance()->GetDoCommandCallback() : nullptr, text, false, estimate_only);
+	CommandCost res = ::DoCommandPInternal(tile, p1, p2, cmd, (_networking && !_generating_world) ? ScriptObject::GetActiveInstance()->GetDoCommandCallback() : nullptr, text == nullptr ? nullptr : command_text.c_str(), false, estimate_only);
```
frosch123: This looks scary to me. Maybe:
```
command_text.empty() ? nullptr : command_text.c_str()
```

frosch123 (dismissed):

frosch123 (approved):

PR #9310, opened 2021-05-29
Codechange: [Network] Use more std::string in server code

rubidium42:
## Motivation / Problem

Partial conversion of C-string to std::string is not done.

## Description

Change some internal code related to kicking/banning, rcon and client names to use std::string over C-strings.

## Limitations

None

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -1719,7 +1719,7 @@ bool NetworkFindName(char *new_name, const char *last)
  * @param new_name the new name for the client
  * @return true iff the name was changed
  */
-bool NetworkServerChangeClientName(ClientID client_id, const char *new_name)
+bool NetworkServerChangeClientName(ClientID client_id, const std::string &new_name)
 {
 	/* Check if the name's already in use */
 	for (NetworkClientInfo *ci : NetworkClientInfo::Iterate()) {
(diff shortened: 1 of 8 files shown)


src/network/network_server.cpp
```diff
@@ -429,12 +429,12 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::SendCompanyInfo()
...
  * @param reason In case of kicking a client, specifies the reason for kicking the client.
  */
-NetworkRecvStatus ServerNetworkGameSocketHandler::SendError(NetworkErrorCode error, const char *reason)
+NetworkRecvStatus ServerNetworkGameSocketHandler::SendError(NetworkErrorCode error, const std::string &reason)
 {
 	Packet *p = new Packet(PACKET_SERVER_ERROR);
 
 	p->Send_uint8(error);
-	if (reason != nullptr) p->Send_string(reason);
+	if (!reason.empty()) p->Send_string(reason);
```
TrueBrain, line 437: We do what now? Oof .. nothing to fix in this PR, but sending fields like this optional is .. euh .. weird.

TrueBrain (approved):

src/network/network_server.cpp
rubidium42: Agreed, it is weird. But the other side in network_client.cpp is even weirder. And... it is part of the happy 8 (previously 6) packets that must remain in order and mostly retain their format. So this is mostly to not break that.

PR #9314, opened 2021-05-30
Codechange: let the setting name be std::string

rubidium42:
## Motivation / Problem

Conversion of C-string to std::string is not completed yet. And since we're heavily refactoring settings, lets tackle one there too.

## Description

Introduce helper functions for determining a string starts or ends with a given prefix/suffix.
Introduce helper function for passing std::string as SetDParamStr for error messages.
Rewrite some internal code to make use of those.

Sort of feature: when doing something like 'set engine_renew on', instead of it saying 'engine_renew = on' it will now mention the fully qualified name, i.e. 'company.engine_renew'.

## Limitations

Left script and gamelog code alone... something with rabbit holes ;)

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -1235,7 +1235,7 @@ static void HandleOldDiffCustom(bool savegame)
 		if (has_no_town_council_tolerance && name == "town_council_tolerance") continue;
 
 		std::string fullname = "difficulty." + name;
-		const SettingDesc *sd = GetSettingFromName(fullname.c_str());
+		const SettingDesc *sd = GetSettingFromName(fullname);
 
 		/* Some settings are no longer in use; skip reading those. */
 		if (sd == nullptr) {
@@ -1705,22 +1705,19 @@ void IntSettingDesc::ChangeValue(const void *object, int32 newval) const
  * @return Pointer to the setting description of setting \a name if it can be found,
  *         \c nullptr indicates failure to obtain the description.
  */
-static const SettingDesc *GetSettingFromName(const char *name, const SettingTable &settings)
+static const SettingDesc *GetSettingFromName(const std::string_view name, const SettingTable &settings)
 {
 	/* First check all full names */
 	for (auto &sd : settings) {
 		if (!SlIsObjectCurrentlyValid(sd->save.version_from, sd->save.version_to)) continue;
-		if (strcmp(sd->name, name) == 0) return sd.get();
+		if (sd->name == name) return sd.get();
 	}
 
 	/* Then check the shortcut variant of the name. */
+	std::string short_name_suffix = std::string{ "." }.append(name);
 	for (auto &sd : settings) {
 		if (!SlIsObjectCurrentlyValid(sd->save.version_from, sd->save.version_to)) continue;
-		const char *short_name = strchr(sd->name, '.');
-		if (short_name != nullptr) {
-			short_name++;
-			if (strcmp(short_name, name) == 0) return sd.get();
-		}
+		if (StrEndsWith(sd->name, short_name_suffix)) return sd.get();
 	}
 
 	return nullptr;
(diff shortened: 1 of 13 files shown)


src/settings.cpp
```diff
@@ -1749,9 +1744,10 @@ void GetSettingSaveLoadByPrefix(const char *prefix, std::vector<SaveLoad> &savel
  * @return Pointer to the setting description of setting \a name if it can be found,
  *         \c nullptr indicates failure to obtain the description.
  */
-static const SettingDesc *GetCompanySettingFromName(const char *name)
+static const SettingDesc *GetCompanySettingFromName(std::string_view name)
 {
-	if (strncmp(name, "company.", 8) == 0) name += 8;
+	static const std::string_view company_prefix = "company.";
```
TrueBrain, line 1747: Can a string be stored in a `string_view`? Where is the real string stored?

Shouldn't this just be a `std::string`?

src/string.cpp
```diff
@@ -358,6 +358,33 @@ void StrTrimInPlace(std::string &str)
...
+ * Check whether the given string starts with the given prefix.
+ * @param str    The string to look at.
+ * @param prefix The prefix to look for.
+ * @return True iff the begin of the string is the same as the prefix.
+ */
+bool StrStartsWith(const std::string_view str, const std::string_view prefix)
+{
+	size_t prefix_len = prefix.size();
+	if (str.size() < prefix_len) return false;
+	return str.compare(0, prefix_len, prefix, 0, prefix_len) == 0;
```
TrueBrain, line 371: I am staring at this, and think: sure? I really have no clue if this is doing the right thing :D

We really need unit-tests for these kind of things .. :P

src/settings.cpp
rubidium42: Yes.  string_view is like span but for string-like things, and in this case it is initialized with a C-string instead of converting the C-string to a std::string and then to a std::string_view. The advantage of using string_view here is that information of the length that is available at compile time does not get lost, compared to using a `const char *`, which makes later size checks/comparisons cheaper. And that calls to `StrStartsWith` do not need to (implicitly) convert data.

src/string.cpp
rubidium42: Or even better, use library quality code such as https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with

If the comment is about the compare, cppreference says "`constexpr int compare(size_type pos1, size_type count1, basic_string_view v, size_type pos2, size_type count2) const;` (3)
...
 3) Equivalent to `substr(pos1, count1).compare(v.substr(pos2, count2))`."

src/settings.cpp
TrueBrain: Sometimes it seems you and I are not talking the same language ;) You do realise I reviewed enough of your PRs to know what `std::string_view` does right? Not what I was asking about ;)

What I am asking, is that `string_view` points to (but does not contain) a string. So is it safe to assign a value to `string_view` like this, or do we need to store it in a proper variable first? For `std::span` I have been told it is UB, so is it a similar problem with `string_view`?

src/string.cpp
TrueBrain: No, the comment is not about the compare. The comment is that I have no clue if this is doing the right thing under all circumstance, as that is near impossible to figure out based on the code. For that, unit-tests would be very welcome.

src/settings.cpp
TrueBrain: To answer my own question: "it depends". https://en.cppreference.com/w/cpp/string/basic_string_view gives a nice example of good and bad usage, which can be best seen here: https://godbolt.org/z/c6YcG1Tzj . Annoyingly, GCC doesn't complain about this at all. Clang does, so that is something at least :)

So from my understanding, your usage is fine, as the c-string is in the `.data`, and as such remains valid. So the view around it remains valid too. If it would be anything that requires any runtime setup, it will not be valid.

Such a thin line to walk .. parts of C++ that still buzz my ears :)

src/string.cpp
TrueBrain: Right, as mentioned, reviewing this by code was really difficult to estimate if this would do the right thing. So I made a mini-unit-text:

https://godbolt.org/z/bxErvPY3o

It seems to be doing what I expect, but let me know if I forgot any test-cases.

TrueBrain (approved):

PR #9315, opened 2021-05-30
Codechange: use std::string with network's ContentInfo

rubidium42:
## Motivation / Problem

C-string to std::string conversion.

## Description

Replace custom allocation of tags/dependencies arrays with vectors, use proper initializations and use std::string instead of string buffers.

## Limitations

Maximum sizes for the fields are magic numbers. However, since they are only used ones as parameter to `Recv_String`, I'm not certain how much benefit adding some constants with those values is going to add. Then you'd need to document that it is to limit the data received from the content server, but... the only place where they are used already implies it.

## Checklist for review

### DIFF
--- a/src/network/core/tcp_content_type.h
+++ b/src/network/core/tcp_content_type.h
@@ -26,6 +26,7 @@ enum ContentType {
 	CONTENT_TYPE_GAME          = 9, ///< The content consists of a game script
 	CONTENT_TYPE_GAME_LIBRARY  = 10, ///< The content consists of a GS library
 	CONTENT_TYPE_END,               ///< Helper to mark the end of the types
+	INVALID_CONTENT_TYPE       = 0xFF, ///< Invalid/uninitialized content
 };
 
 /** Enum with all types of TCP content packets. The order MUST not be changed **/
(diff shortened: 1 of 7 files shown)


src/network/network_content.cpp
```diff
@@ -70,9 +70,9 @@ bool ClientNetworkContentSocketHandler::Receive_SERVER_INFO(Packet *p)
 	ci->dependencies = MallocT<ContentID>(ci->dependency_count);
 	for (uint i = 0; i < ci->dependency_count; i++) ci->dependencies[i] = (ContentID)p->Recv_uint32();
 
-	ci->tag_count = p->Recv_uint8();
-	ci->tags = MallocT<char[32]>(ci->tag_count);
-	for (uint i = 0; i < ci->tag_count; i++) p->Recv_string(ci->tags[i], lengthof(*ci->tags));
+	uint tag_count = p->Recv_uint8();
+	ci->tags.reserve(tag_count);
```
TrueBrain, line 74: is `reserve` really useful? (honest question, to be clear)

src/network/network_content.cpp
```diff
@@ -70,9 +70,9 @@ bool ClientNetworkContentSocketHandler::Receive_SERVER_INFO(Packet *p)
...
+	for (uint i = 0; i < tag_count; i++) ci->tags.push_back(p->Recv_string(32));
```
TrueBrain: would you mind making `32` into an enum value, like the rest, while here? Would be appreciated :)

src/network/network_content.cpp
```diff
@@ -56,10 +56,10 @@ bool ClientNetworkContentSocketHandler::Receive_SERVER_INFO(Packet *p)
...
 	ci->filesize = p->Recv_uint32();
 
-	p->Recv_string(ci->name, lengthof(ci->name));
-	p->Recv_string(ci->version, lengthof(ci->version));
-	p->Recv_string(ci->url, lengthof(ci->url));
-	p->Recv_string(ci->description, lengthof(ci->description), SVS_REPLACE_WITH_QUESTION_MARK | SVS_ALLOW_NEWLINE);
+	ci->name        = p->Recv_string(32);
+	ci->version     = p->Recv_string(16);
+	ci->url         = p->Recv_string(96);
+	ci->description = p->Recv_string(512, SVS_REPLACE_WITH_QUESTION_MARK | SVS_ALLOW_NEWLINE);
```
TrueBrain: Oof, more of those magic numbers. Seems this part of the code really collected them ... Well, just see if you can/want to do something with them, or leave them, what-ever ;)

src/network/core/tcp_content_type.h
```diff
@@ -26,6 +26,7 @@ enum ContentType {
 	CONTENT_TYPE_GAME          = 9, ///< The content consists of a game script
 	CONTENT_TYPE_GAME_LIBRARY  = 10, ///< The content consists of a GS library
 	CONTENT_TYPE_END,               ///< Helper to mark the end of the types
+	CONTENT_TYPE_INVALID       = 0xFF, ///< Invalid/uninitialized content
```
TrueBrain: Code convention, and I am not sure I agree with it, but we do it everywhere, says `INVALID` is a prefix, not postfix. so `INVALID_CONTENT_TYPE`.
Not everyone got that memo btw, so there are some strays, but .. yeah .. you can see it with `CONTENT_ID` a bit lower too :)

src/network/network_content.cpp
rubidium42: From cppreference: "Correctly using reserve() can prevent unnecessary reallocations, but inappropriate uses of reserve() (for instance, calling it before every push_back() call) may actually increase the number of reallocations...". Long story short, here we tell the vector: "heh, we know we need at least this many elements so you can prepare for that", and then the vector does not need to use its (potential) exponential growth algorithm where it doubles the size of the vector each time. So if there are say 9 tags, it would first make a vector of 2, then reallocate to 4, then reallocate to 8 and finally reallocate to 16. Since each reallocate might copy, that is potentially 14 unnecessary copies of data and leaving room for 7 tags that we know will never be filled.

src/network/network_content.cpp
rubidium42: I have been pondering doing that, but then I'd "need" to add them to the massive list of "enum" values for the other sizes of things in the network and only use them once. 
So I was somewhat in doubt whether that actually makes things much clearer. As I would replace the 32 here with NETWORK_CONTENT_TAG_LENGTH, but that could already be inferred by it reading a string from the packet into the tag list of content.

TrueBrain: Well, in general magic numbers are bad ofc, I think it is more a question if we want to address that here and now :) Even adding it to the big list and using it only once has benefits: it is clear that the length is only used in one place. So if someone extends it to a higher number, it is very clear that there are no lingering problems because you missed a "32 - 1" situation etc :D

So if you don't mind, I would really appreciate it, but I also fully understand if it is "future work" :)

TrueBrain (approved):

Issue #83, opened 2006-03-25
openttd crashes when joining network game


DorpsGek:
**_gradator_ opened the ticket and wrote:**

> Hello,
> 
> Yesterday I joined the game running on "Official OpenTTD Server # 1" without any problems.
> 
> This morning it appears that I have been disconnected (game finished ?) in anyway the game
> returned to the main menu.
> 
> Now when I want to join the game (as spectator or player) which is running on this server,
> my client connects to, download all data needed, and crashes a few seconds later (I can even
> scroll the world during this time).
> 
> It is fully reproductible.
> 
> here are the debug log (with the game running with ./openttd -d 9999 :
> 
> dbg: load sprite 1420
> dbg: load sprite 1317
> dbg: load sprite 1433
> dbg: load sprite 1319
> dbg: [NET] Received FRAME 1885715
> dbg: [NET] Received FRAME 1885716
> dbg: [NET] Received FRAME 1885717
> dbg: [NET] Received FRAME 1885718
> openttd: pool.h:51: GetItemFromPool: Assertion \`index < pool->total_items' failed.
> Aborted
> 
> It always happen in a frame between 1880000 to 1900000.
> 
> 
> Sylvain

**Reported version**: _0.4.6_
**Operating system**: _Linux_

---

DorpsGek:
**_Celestar_ wrote:**

> Possibly a dublicate. Will check

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/83#comment165_

DorpsGek:
**_Celestar_ closed the ticket.**

**Reason for closing: _Duplicate_**

> 

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/83_

PR #9325, opened 2021-05-31
Feature: Ctrl-Click on vehicle list in GB_SHARED_ORDERS opens order window

btzy:
## Motivation / Problem

To figure out which vehicle group we want (e.g. to clone) when looking at a vehicle list (e.g. at a station), we need to look at the order list.  However, it takes 3 clicks (from "Bardingstone Heights - 48 trains" to "Train #83 (Orders)"), which is a bit unsatisfactory:

image

### Ctrl-Click on vehicle list window

This PR implements Ctrl-Click in the vehicle list window.  Ctrl-Clicking on a vehicle in the vehicle list window (when "Group by" is set to "Shared orders") will open the order list of the first vehicle sharing those orders.

### Issues with group GUI window

There is a slight change in behaviour for the group GUI window:

image

Ctrl-Clicking on a vehicle in the group window used to select the group in which the vehicle is in, but now it doesn't do that (when "Group by" is set to "Shared orders").  Well I think that was previously a bug - since vehicles from different groups could share orders, there may not be a unique group to select.

### Issues with shared order list window

The shared orders window can't be grouped by shared orders, and it is pointless to do so anyway.  But it still makes sense to want to open the order window to see the list of orders.  Hence this PR adds a button at the top right corner to open the order list.  The button is in a similar position to the the "Orders" button in the timetable window.

image

## Unanswered questions

Should the order window have a title that feels more like the orders are shared, such as "Shared orders of Train #83 and 9 other vehicles" instead of "Train #83 (Orders)"?  Since with shared orders, we no longer think of the order list as a property of a single vehicle, but instead logically each vehicle subscribes to some (shared) order list.  Apart from the "Skip" button (and possibly the "Timetable" button which I'm not too familiar with), everything else pertains to things that are shared.

## Note

With the addition of this PR, it is now possible to open the orders window without first opening the associated vehicle window.  When closing a vehicle window, the associated orders window is automatically closed too, so it seems that previously such a situation could never happen.  This PR doesn't seem to break anything though.  But it does raise the question of whether we should head towards further decoupling between the order window and the vehicle window in the future, at least when the vehicle has shared orders.

In any case, this PR can be merged without answering the unanswered questions yet, so it isn't a blocking issue.

## Limitations

Not that I know of.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -1843,10 +1843,14 @@ public:
 						const Vehicle *v = vehgroup.vehicles_begin[0];
 						/* We do not support VehicleClicked() here since the contextual action may only make sense for individual vehicles */
 
-						if (vehgroup.NumVehicles() == 1) {
-							ShowVehicleViewWindow(v);
+						if (_ctrl_pressed) {
+							ShowOrdersWindow(v);
 						} else {
-							ShowVehicleListWindow(v);
+							if (vehgroup.NumVehicles() == 1) {
+								ShowVehicleViewWindow(v);
+							} else {
+								ShowVehicleListWindow(v);
+							}
 						}
 						break;
 					}
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -1367,7 +1367,13 @@ void ChangeVehicleViewWindow(VehicleID from_index, VehicleID to_index)
 static const NWidgetPart _nested_vehicle_list[] = {
 	NWidget(NWID_HORIZONTAL),
 		NWidget(WWT_CLOSEBOX, COLOUR_GREY),
-		NWidget(WWT_CAPTION, COLOUR_GREY, WID_VL_CAPTION),
+		NWidget(NWID_SELECTION, INVALID_COLOUR, WID_VL_CAPTION_SELECTION),
+			NWidget(WWT_CAPTION, COLOUR_GREY, WID_VL_CAPTION),
+			NWidget(NWID_HORIZONTAL),
+				NWidget(WWT_CAPTION, COLOUR_GREY, WID_VL_CAPTION_SHARED_ORDERS),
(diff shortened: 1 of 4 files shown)


src/vehicle_gui.cpp
```diff
@@ -1651,14 +1663,17 @@ struct VehicleListWindow : public BaseVehicleListWindow {
 		/* Set up the window widgets */
 		this->GetWidget<NWidgetCore>(WID_VL_LIST)->tool_tip = STR_VEHICLE_LIST_TRAIN_LIST_TOOLTIP + this->vli.vtype;
 
+		NWidgetStacked* nwi = this->GetWidget<NWidgetStacked>(WID_VL_CAPTION_SELECTION);
```
TrueBrain: Minor nitpick in case we need to rebase: `NWidgetStacked *nwi`

TrueBrain (dismissed):
Sorry it took a bit of time to get to this PR.

I have nothing against this PR, the motivation is reasonable, and it seems to help players, without hurting people that are used to their workflow. Basically, I do not see a reason why not.
I really do appreciate your lengthy explaining why you did what you did, that is very helpful. Tnx for that!

> Should the order window have a title that feels more like the orders are shared, such as "Shared orders of Train #83 and 9 other vehicles" instead of "Train #83 (Orders)"? Since with shared orders, we no longer think of the order list as a property of a single vehicle, but instead logically each vehicle subscribes to some (shared) order list. Apart from the "Skip" button (and possibly the "Timetable" button which I'm not too familiar with), everything else pertains to things that are shared.

Yeah, this has been annoying me for a long time. I think https://github.com/OpenTTD/OpenTTD/discussions/9551 is related. I do not think changing the caption now is a good idea, exactly for the reasons you mention (skip and timetable). But we really should address this whole Order window, as it is getting a bit out of hand :D But I will take my ideas to the discussion for that :)

src/vehicle_gui.cpp
btzy: Fixed.

btzy:
@TrueBrain Thanks for your review!  I've rebased and fixed the nit.  It does seem that rebasing dismisses your approving review...

TrueBrain (approved):

PR #9330, opened 2021-06-02
Fix #9329: [CMake] FindVersion.cmake relied on internal cmake variable

### LINKED ISSUE #9329: No rule to make target 'generated/ottdres.rc' when Cross-compiling with CMake + MINGW
I'm unable to cross-compile OpenTTD (master) inside a MINGW environment using CMake due to the aforementioned error.

Things I did:
 * Follow the steps mentioned in: https://github.com/OpenTTD/OpenTTD/pull/8350
 * `cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=~/toolchain-mingw32.cmake -DHOST_BINARY_DIR=native/ ..`
* `cmake --build . --target openttd`

In the toolchain file I've everything set up correctly, as I used it before to build other projects. Either there is something I didn't do previously (probably, since openttd cross-compilation with cmake is not well documented), or there is something missing in CMake files.

glx22:
## Motivation / Problem
`FindVersion.cmake` is ran via `cmake -P` and `WIN32` is only set on windows systems.
When cross-compiling for windows from linux, `WIN32` is set via the toolchain so configuration works, but `cmake -P` doesn't have it at build time.

## Description
Explicitly tell FindVersion.cmake it's targeting windows.

## Checklist for review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ add_custom_target(find_version
                 -DREV_MAJOR=${PROJECT_VERSION_MAJOR}
                 -DREV_MINOR=${PROJECT_VERSION_MINOR}
                 -DREV_BUILD=${PROJECT_VERSION_PATCH}
+                -DWINDOWS=${WIN32}
                 -P "${CMAKE_SOURCE_DIR}/cmake/scripts/FindVersion.cmake"
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
         BYPRODUCTS ${GENERATED_SOURCE_FILES}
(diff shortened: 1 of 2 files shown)


LordAro:
Worth a comment explaining why there's a WINDOWS car ?

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ add_custom_target(find_version
                 -DREV_MAJOR=${PROJECT_VERSION_MAJOR}
                 -DREV_MINOR=${PROJECT_VERSION_MINOR}
                 -DREV_BUILD=${PROJECT_VERSION_PATCH}
+                -DWINDOWS=${WIN32}
```
PeterN, line 80: Does `-DWIN32=${WIN32}` work instead, and remove the need for the second change?

glx22: Yes, that works too.

michicc (approved):

PR #9336, opened 2021-06-04
Fix: [Script] Ensure the saved script strings are properly validated and terminated when being read from the save game

rubidium42:
## Motivation / Problem

Just set the string data length to 255 and fill it will all non-zeros. When buf[255] is also non-zeros, this will read beyond the bounds when making the string. Similarly the maximum value of _script_sl_byte can only be 255 though 256 bytes are allocated for the buffer.

## Description

Allocate just enough memory for a buffer of _script_sl_byte long.
Perform a string validation on the string coming from the save game ensuring termination and valid Utf8 characters

## Limitations

Potentially a game script that put some non-Utf8 characters in their strings would get different information when loading compared to what they saved. However, a string should be a valid string. If you want something else stored, use an array of integers.

## Checklist for review

### DIFF
--- a/src/script/script_instance.cpp
+++ b/src/script/script_instance.cpp
@@ -569,8 +569,9 @@ bool ScriptInstance::IsPaused()
 
 		case SQSL_STRING: {
 			SlObject(nullptr, _script_byte);
-			static char buf[256];
+			static char buf[std::numeric_limits<decltype(_script_sl_byte)>::max()];
 			SlArray(buf, _script_sl_byte, SLE_CHAR);
+			StrMakeValidInPlace(buf, buf + _script_sl_byte);
 			if (vm != nullptr) sq_pushstring(vm, buf, -1);
 			return true;
 		}


src/script/script_instance.cpp
```diff
@@ -569,8 +569,9 @@ bool ScriptInstance::IsPaused()
 
 		case SQSL_STRING: {
 			SlObject(nullptr, _script_byte);
-			static char buf[256];
+			static char buf[std::numeric_limits<decltype(_script_sl_byte)>::max()];
```
LordAro, line 572: This feels real dirty. Is `lengthof` not good enough?

rubidium42: No, `lengthof` is giving you the numer of elements that are in an array. Here I am not looking for the number of elements in the array, but I am rather looking for the maximum value of `_script_sl_byte` as that's the maximum number of bytes that will be used of the buffer.

Practically `255` or `UINT8_MAX` would suffice, but those will break as soon as someone decides `_script_sl_byte` is to be a UINT16. Ofcourse it could just test whether `_script_sl_byte` is less or equal to `lengthof(buf)`, but that will in its turn cause complaints about dead code since currently there is no way for `_script_sl_byte` to be larger than 255. Only once the type is made larger that becomes a possibility, and only then that code makes any sense.
This way the allocated buffer will always be large enough. Well, for uint16. For uint32 it will probably crash quickly due to allocating too much on the stack. The alternative would be to use `255` or `UINT8_MAX` and then some `static_assert` to ensure the value for `_script_sl_byte` is always within bounds, but that would likely include something similar to what I have written here.

michicc (approved):

PR #9338, opened 2021-06-06
Several code refactors of the SaveLoad code

TrueBrain:
Depends on #9335

## Motivation / Problem

When you start fiddling with the SaveLoad code, you find many odd things. And sometimes they annoy me enough that you get this collection of commits.

## Description

First commit is #9335, after that:

- First patch is some code hygiene
- Second patch too
- Third: `SL_DEQEUE` suggested to support signed values, but they were silently cast to unsigned .. which, with conversion code like `SLE_FILE_I8 | SLE_VAR_I32` might not work as you expect. Luckily, nobody was using this code for signed values, but let's make sure we are ahead of this issue.
- Forth: `SlList` is very similar to `SlDeque`, but a copied the code. So, together with the third commit, I changed the `SlDequeHelper` into a more general one, where anything supporting `begin()` and `end()` will work on.
- Fifth: `SL_LST` is a list of references, which the name did not tell you. The way reference are done is already very error-prone, and this wasn't helping. So I renamed it to `SL_REFLST` to be more obvious of its function.
- Sixth: cheat-chunk was weird. Just that. Weird. Made it slightly less weird.

I could have made 4 PRs out of this too, so if this is too hard to review, let me know, and I will split it up :)

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -1218,6 +1218,31 @@ static void *IntToReference(size_t index, SLRefType rt)
 }
 
 /**
+ * Handle conversion for references.
+ * @param ptr The object being filled/read.
+ * @param conv VarType type of the current element of the struct.
+ */
+void SlSaveLoadRef(void *ptr, VarType conv)
+{
+	switch (_sl.action) {
+		case SLA_SAVE:
+			SlWriteUint32((uint32)ReferenceToInt(*(void **)ptr, (SLRefType)conv));
+			break;
+		case SLA_LOAD_CHECK:
+		case SLA_LOAD:
+			*(size_t *)ptr = IsSavegameVersionBefore(SLV_69) ? SlReadUint16() : SlReadUint32();
+			break;
+		case SLA_PTRS:
+			*(void **)ptr = IntToReference(*(size_t *)ptr, (SLRefType)conv);
+			break;
+		case SLA_NULL:
+			*(void **)ptr = nullptr;
+			break;
+		default: NOT_REACHED();
+	}
+}
+
+/**
  * Template class to help with list-like types.
  */
 template <template<typename, typename> typename Tstorage, typename Tvar, typename Tallocator = std::allocator<Tvar>>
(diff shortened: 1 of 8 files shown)


src/saveload/saveload.cpp
```diff
@@ -1210,133 +1218,112 @@ static void *IntToReference(size_t index, SLRefType rt)
...
+
+		SlStorageT *l = (SlStorageT *)storage;
 
 		switch (_sl.action) {
 			case SLA_SAVE: {
 				SlWriteUint32((uint32)l->size());
 
-				typename SlDequeT::iterator iter;
+				typename SlStorageT::iterator iter;
 				for (iter = l->begin(); iter != l->end(); ++iter) {
```
LordAro: this could be a range-for ? Would avoid the awkward `&(*iter)`

src/saveload/saveload.cpp
```diff
@@ -1437,7 +1433,7 @@ size_t SlCalcObjMemberLength(const void *object, const SaveLoad &sld)
 		case SL_REF:
 		case SL_ARR:
 		case SL_STR:
-		case SL_LST:
+		case SL_REFLST:
```
LordAro, line 1436: Since you're renaming & not preserving length, why not just go for the full `SL_REFLIST` ?

src/saveload/saveload.h
```diff
@@ -497,7 +497,7 @@ enum SaveLoadType : byte {
 	SL_REF         =  1, ///< Save/load a reference.
 	SL_ARR         =  2, ///< Save/load an array.
 	SL_STR         =  3, ///< Save/load a string.
-	SL_LST         =  4, ///< Save/load a list.
+	SL_REFLST      =  4, ///< Save/load a reference list.
```
LordAro: Since you're renaming & not preserving length anyway, why not just go for the full `SL_REFLIST` ?

TrueBrain: I kept the "short" notation for all of them .. but if we don't mind tables being (even more) misaligned, yeah, lets go for more descriptive names.

But just so to check, it does mean that the SLG name will also be `SLG_REF_LIST`, and that tables alignment will be even worse. You okay with that?

src/saveload/saveload.cpp
```diff
@@ -1217,6 +1217,67 @@ static void *IntToReference(size_t index, SLRefType rt)
...
+	typedef Tstorage<Tvar, Tallocator> SlStorageT;
+public:
+	/**
+	 * Internal templated helper to return the size in bytes of a list-like type.
+	 * @param storage The storage to find the size of
+	 * @param conv VarType type of variable that is used for calculating the size
+	 */
+	static size_t SlCalcLen(const void *storage, VarType conv)
+	{
+		const SlStorageT *l = (const SlStorageT *)storage;
```
rubidium42: I'd use `static_cast` so the compile has some chance to attempt to type check it, if it has enough information to deduce it. With a C-style cast like it is now it will definitely not do it.
Though the chance of it actually happening would probably be quite low as the type information of the `storage` has probably been "lost".

src/saveload/saveload.cpp
```diff
@@ -1217,6 +1217,67 @@ static void *IntToReference(size_t index, SLRefType rt)
...
+
+		int type_size = 4;
```
rubidium42: This feels kind of a "random" name. It's not really the type size, but it is already super unclear in the original code due to the bad comment.
I know it's the size in (8 bit) bytes of the uint32 size field that is being stored in the save game. But naming that clearly and concisely is a bit weird; the obvious one would be `size_size`, i.e. the size of the size, but that makes no real sense. Similarly `length_size`, but it's the `sizeof` of the result of `size()` that is written.

Maybe `return sizeof(uint32) + l->size() * item_size;`? That would at least tell you there a `uint32` + `size()` times something of `item_size`, and implicitly I'd reckon that would imply "a size field + all the elements".

src/saveload/saveload.cpp
```diff
@@ -1217,6 +1217,67 @@ static void *IntToReference(size_t index, SLRefType rt)
...
+	 * Internal templated helper to save/load a list-like type.
+	 * @param storage The storage being manipulated.
+	 * @param conv VarType type of variable that is used for calculating the size.
+	*/
+	static void SlSaveLoad(void *storage, VarType conv)
+	{
+		SlStorageT *list = (SlStorageT *)storage;
+
+		switch (_sl.action) {
+			case SLA_SAVE: {
```
rubidium42: Unneeded brackets, especially as you remove them in the next commit.

src/saveload/saveload.cpp
```diff
@@ -1228,125 +1253,103 @@ class SlStorageHelper {
...
-			PtrList::iterator iter;
-			for (iter = l->begin(); iter != l->end(); ++iter) {
-				void *ptr = *iter;
-				SlWriteUint32((uint32)ReferenceToInt(ptr, conv));
-			}
-			break;
-		}
-		case SLA_LOAD_CHECK:
-		case SLA_LOAD: {
-			size_t length = IsSavegameVersionBefore(SLV_69) ? SlReadUint16() : SlReadUint32();
```
rubidium42, line 1264: I think save games before SLV_69 are broken by this. The new code always reads 32 bits.
The same then also holds for the SlCalc(List)Len.

Since SLE_CONDLST was added in SLV_57, opntitle.dat just works fine. PSG-40 is "broken" (unexpected end of chunk) in this PR, with 1.11.0 it's not broken.

src/saveload/saveload.h
```diff
@@ -497,7 +497,7 @@ enum SaveLoadType : byte {
 	SL_REF         =  1, ///< Save/load a reference.
 	SL_ARR         =  2, ///< Save/load an array.
 	SL_STR         =  3, ///< Save/load a string.
-	SL_LST         =  4, ///< Save/load a list.
+	SL_REFLIST     =  4, ///< Save/load a reference list.
```
rubidium42: Maybe `list of #SLREF elements` is clearer?

src/saveload/saveload.cpp
TrueBrain: You spend a lot of words on this :D

Either way, this is moved code; I tend to limit what I change in them until someone wants it changed :P

src/saveload/saveload.cpp
TrueBrain: Again, moved code. This rabbit-hole is already rather deep, so I really want to scope it at some point. I don't want to second-guess every decision ever made in our code-past :D

src/saveload/saveload.cpp
TrueBrain: So they are removed in the next commit :)

Again, moved code :D Not going to resolve this one, as you pointed out, fixed in next commit :)

src/saveload/saveload.cpp
TrueBrain: wauw, that took way to long to decypher what you said here ;)

Seems in all places the length-thing did survive except for SLA_LOAD. Ironical :)

`SlCalc` etc does have the right code. But, ironically, `SlCalc` doesn't need it, as that is only used for writing. But not going to change everything all at once ;)

TrueBrain: Should be fixed; can you try the savegame that was failing?

I have a wide collection, but clearly not one that triggered this issue :D

src/saveload/saveload.cpp
TrueBrain: For now I settled on a comment. Soon (tm) the line is going to be removed anyway, as it is stupid that we spend 4 bytes on a length-indicator and not use a gamma, where 80% of the rest of the code just uses a gamme for length-indicators .. ugh .. SaveLoad code :P So a comment will have to do for now :)

src/saveload/saveload.cpp
```diff
@@ -1217,141 +1218,144 @@ static void *IntToReference(size_t index, SLRefType rt)
...
+		assert(cmd == SL_VAR || cmd == SL_REF);
 
-		int type_size = 4;
-		/* Each entry is saved as type_size bytes, plus type_size bytes are used for the length
-		 * of the list */
-		return l->size() * SlCalcConvFileLen(conv) + type_size;
+		const SlStorageT *list = static_cast<const SlStorageT *>(storage);
+
+		int type_size = 4; // Size of the length of the list.
+		int item_size = cmd == SL_VAR ? SlCalcConvFileLen(conv) : (IsSavegameVersionBefore(SLV_69) ? 2 : 4);
```
rubidium42: This seems to be used only for saving (`	assert(_sl.action == SLA_SAVE);` does not fail), so the version check is not needed.

src/saveload/saveload.cpp
```diff
@@ -1217,141 +1218,144 @@ static void *IntToReference(size_t index, SLRefType rt)
...
 
-				/* Load each value and push to the end of the deque */
+				/* Load each value and push to the end of the storage. */
 				for (size_t i = 0; i < length; i++) {
-					T data;
-					SlSaveLoadConv(&data, conv);
-					l->push_back(data);
+					Tvar data;
+					SlSaveLoadMember(cmd, &data, conv);
+					list->push_back(data);
```
rubidium42: Could be something like `Tvar &data = list->emplace_back();` and then the `list->push_back(data);` is not needed anymore. Less code and less copying, as it will immediately write the data in the element in the storage instead of first on the stack and then copying it to the storage.

rubidium42 (approved):

PR #9339, opened 2021-06-07
Codechange: add and use the ability to store (lists of) structs in savegames

TrueBrain:
## Motivation / Problem

In a few places (stations, vehicles, towns, companies, ..) the SaveLoad code needs to handle either structs in structs, or lists of structs in structs.

This is all done manually, and not only hard to read, also easy to mess up.

## Description

Introduce `SL_STRUCT` and `SL_STRUCTLIST` to embed a struct in another struct.

For example, for Vehicles it used to have a special `SL` command to embed the common vehicle structure. This meant that SaveLoad needed to know about Vehicles.

Now, the common vehicle sub-struct is called `SlVehicleCommon` and to save it you can use:
```
SLEG_STRUCT(SlVehicleCommon)
```

This works with structs in structs in structs in structs in structs in ... just fine too.

NOTE: I made several commits to introduce `SL_STRUCT` in the rest of the code, as I am hoping that makes reviewing easier. 
NOTE 2: I left a lot of odd constructs in there, just to make review easier. The code really could use a cleanup, but .. I am trying not to do everything at once, also for my own sanity :D
NOTE 3: there should be no difference before or after this PR in respect to what is written/loaded from disk.

## Limitations

- I tested this with ~500 savegames, ranging from TTDp till 1.11.2, validating that they load and run. Nothing that wasn't broken before this PR is broken now. So I think there are no regressions, but with this many changes in the SaveLoad code, that cannot be guaranteed.

## Checklist for review

### DIFF
--- a/src/saveload/saveload.h
+++ b/src/saveload/saveload.h
@@ -400,6 +400,73 @@ struct ChunkHandler {
 /** A table of ChunkHandler entries. */
 using ChunkHandlerTable = span<const ChunkHandler>;
 
+/** A table of SaveLoad entries. */
+using SaveLoadTable = span<const struct SaveLoad>;
+
+/** Handler for saving/loading an object to/from disk. */
+class SaveLoadHandler {
+public:
+	virtual ~SaveLoadHandler() {}
+
+	/**
+	 * Save the object to disk.
+	 * @param object The object to store.
+	 */
+	virtual void Save(void *object) const {}
+
+	/**
+	 * Load the object from disk.
+	 * @param object The object to load.
+	 */
+	virtual void Load(void *object) const {}
+
+	/**
+	 * Similar to load, but used only to validate savegames.
+	 * @param object The object to load.
+	 */
+	virtual void LoadCheck(void *object) const {}
+
+	/**
+	 * A post-load callback to fix #SL_REF integers into pointers.
+	 * @param object The object to fix.
+	 */
+	virtual void FixPointers(void *object) const {}
+
+	/**
+	 * Get the description of the fields in the savegame.
+	 */
+	virtual SaveLoadTable GetDescription() const = 0;
+};
+
+/**
+ * Default handler for saving/loading an object to/from disk.
+ *
+ * This handles a few common things for handlers, meaning the actual handler
+ * needs less code.
+ *
+ * Usage: class SlMine : public DefaultSaveLoadHandler<SlMine, MyObject> {}
+ *
+ * @tparam TImpl The class initializing this template.
+ * @tparam TObject The class of the object using this SaveLoadHandler.
+ */
+template <class TImpl, class TObject>
+class DefaultSaveLoadHandler : public SaveLoadHandler {
+public:
+	SaveLoadTable GetDescription() const override { return static_cast<const TImpl *>(this)->description; }
+
+	virtual void Save(TObject *object) const {}
+	void Save(void *object) const override { this->Save(static_cast<TObject *>(object)); }
+
+	virtual void Load(TObject *object) const {}
+	void Load(void *object) const override { this->Load(static_cast<TObject *>(object)); }
+
+	virtual void LoadCheck(TObject *object) const {}
+	void LoadCheck(void *object) const override { this->LoadCheck(static_cast<TObject *>(object)); }
+
+	virtual void FixPointers(TObject *object) const {}
+	void FixPointers(void *object) const override { this->FixPointers(static_cast<TObject *>(object)); }
+};
+
 /** Type of reference (#SLE_REF, #SLE_CONDREF). */
 enum SLRefType {
 	REF_ORDER          =  0, ///< Load/save a reference to an order.
@@ -501,10 +568,12 @@ enum SaveLoadType : byte {
 	SL_REFLIST     =  4, ///< Save/load a list of #SL_REF elements.
 	SL_DEQUE       =  5, ///< Save/load a deque of #SL_VAR elements.
 	SL_STDSTR      =  6, ///< Save/load a \c std::string.
+	SL_STRUCT      =  7, ///< Save/load a struct.
+	SL_STRUCTLIST  =  8, ///< Save/load a list of structs.
 	/* non-normal save-load types */
-	SL_WRITEBYTE   =  8,
-	SL_VEH_INCLUDE =  9,
-	SL_ST_INCLUDE  = 10,
+	SL_WRITEBYTE   =  9,
+	SL_VEH_INCLUDE = 10,
+	SL_ST_INCLUDE  = 11,
 };
 
 typedef void *SaveLoadAddrProc(void *base, size_t extra);
@@ -519,10 +588,8 @@ struct SaveLoad {
 	size_t size;                    ///< the sizeof size.
 	SaveLoadAddrProc *address_proc; ///< callback proc the get the actual variable address in memory
 	size_t extra_data;              ///< extra data for the callback proc
+	SaveLoadHandler *handler;       ///< Custom handler for Save/Load procs.
 };
-
-/** A table of SaveLoad entries. */
-using SaveLoadTable = span<const SaveLoad>;
 
 /**
  * Storage of simple variables, references (pointers), and arrays.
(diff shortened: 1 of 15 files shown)


src/saveload/cheat_sl.cpp
```diff
@@ -14,37 +14,60 @@
...
  * Load the cheat values.
  */
 static void Load_CHTS()
 {
-	Cheat *cht = (Cheat*)&_cheats;
-	size_t count = SlGetFieldLength() / 2;
+	size_t count = SlGetFieldLength();
 	/* Cannot use lengthof because _cheats is of type Cheats, not Cheat */
-	if (count > sizeof(_cheats) / sizeof(Cheat)) SlErrorCorrupt("Too many cheat values");
+	if (count / 2 > sizeof(_cheats) / sizeof(Cheat)) SlErrorCorrupt("Too many cheat values");
```
rubidium42: Maybe you could use `count > lengthof(_cheats_desc)`? Then you could also remove the nasty dummies.

Having said that, this commit does not seem to be part/mentioned in this PR. Are you trying to cheat it in? ;)
Or did you want to consider using SLE_STRUCT for each of the Cheats?

src/saveload/saveload.h
```diff
@@ -399,6 +399,27 @@ struct ChunkHandler {
 /** A table of ChunkHandler entries. */
 using ChunkHandlerTable = span<const ChunkHandler>;
 
+/** A table of SaveLoad entries. */
+using SaveLoadTable = span<const struct SaveLoad>;
+
+class SaveLoadHandler {
```
rubidium42, line 407: I guess the draftiness of this PR means that the comments here are still pending / that you're waiting on glx's work, right?

src/saveload/town_sl.cpp
```diff
@@ -113,7 +113,89 @@ void UpdateHousesAndTowns()
...
+			SlObject(&t->received[i], this->GetDescription());
+		}
+	}
+};
+
+class SlTownAcceptanceMatrix : public SaveLoadHandler {
+public:
+	SaveLoadTable GetDescription()
+	{
+		/* Here due to private member vars. */
```
rubidium42: Is this still needed? I can't see `GetTileMatrixDesc` being made `friend` anywhere, so it seems to be not needed anymore.

src/saveload/saveload.h
```diff
@@ -692,11 +690,8 @@ struct SaveLoad {
  */
 #define SLE_CONDNULL(length, from, to) {SL_ARR, SLE_FILE_U8 | SLE_VAR_NULL, length, from, to, 0, nullptr, 0, nullptr}
 
-/** Translate values ingame to different values in the savegame and vv. */
-#define SLE_WRITEBYTE(base, variable) SLE_GENERAL(SL_WRITEBYTE, base, variable, 0, 0, SL_MIN_VERSION, SL_MAX_VERSION, 0)
-
-#define SLE_VEH_INCLUDE() {SL_VEH_INCLUDE, 0, 0, SL_MIN_VERSION, SL_MAX_VERSION, 0, [] (void *b, size_t) { return b; }, 0}
-#define SLE_ST_INCLUDE() {SL_ST_INCLUDE, 0, 0, SL_MIN_VERSION, SL_MAX_VERSION, 0, [] (void *b, size_t) { return b; }, 0}
+/** Only write byte during saving; never read it during loading. */
```
rubidium42: The comment is misleading as the byte is read, just separately using `SlReadByte` to ensure the right (sub)class is constructed. Now it reads like: "we write a byte to the save game, that we never need", and I start wondering. Why write a byte to the save game that is never read?

src/saveload/cheat_sl.cpp
TrueBrain: This commit message making this change literal is called: `Codechange: rework CHTS-chunk save/load to be more like the others` :P I have to see if I have enough work to warrant another "mixed-fixes-for-stupid-code" PR .. but that is also why this PR is still a draft :)

Either way, this was existing code, I really did not parse yet what it said or tries to say. Just moved a line of code from one above to the one below.

What do you mean with "nasty dummies"? What "nasty" and what "dummies" is that referring to?

src/saveload/saveload.h
TrueBrain: it means a lot of things, but most that this is not the final form yet, and that I didn't do a pass myself on any coding guideliness :) I mostly created the PR so I could collect opinions on the way of doing stuff.

src/saveload/saveload.h
TrueBrain: I am happy to read that you read it like that, as that is exactly what it does. And sure, we have an intended usage of it, but that is just that: how we intend someone to use it .. but nothing is enforcing that.

It really is a shitty mechanism that is very easy to misunderstand and fuck up. So I am more looking into how to remove it completely ;)

That said, I can add a line of doxygen comment to explain the intended use of it .. but at least you now understood correctly what it did :) The initial comment was truly misleading ;)

src/saveload/cheat_sl.cpp
rubidium42: For the dummies, take a look at: https://github.com/OpenTTD/OpenTTD/blob/1e564b333fa2d7bde2f21d7a05bb3e81305185c5/src/cheat_type.h#L31 further on line 33 and 36. That, together with the comment around line 23 made them nasty to me.

src/saveload/vehicle_sl.cpp
```diff
@@ -812,66 +878,98 @@ SaveLoadTable GetVehicleDescription(VehicleType vt)
...
-		_roadveh_desc,
-		_ship_desc,
-		_aircraft_desc,
-		_special_desc,
-		_disaster_desc,
-		_common_veh_desc,
-	};
+	void Save(Vehicle *v) const override { this->GenericSaveLoad(v); }
+	void Load(Vehicle *v) const override { this->GenericSaveLoad(v); }
+	void FixPointers(Vehicle *v) const override { this->GenericSaveLoad(v); }
```
rubidium42, line 953: I'm seeing these three return over and over again, might it be worth it to have an intermediate "super" class with virtual GenericSaveLoad, or maybe even a templated intermediate super class that requires the implementing sub class to provide a GenericSaveLoad? The same is already done with description, and it means less duplicating of these three lines all over the save load code.

src/saveload/saveload.h
```diff
@@ -400,6 +400,68 @@ struct ChunkHandler {
...
+	virtual SaveLoadTable GetDescription() const = 0;
+};
+
+/**
+ * Default handler for saving/loading an object to/from disk.
+ *
+ * This handles a few common things for handlers, meaning the actual handler
+ * needs less code.
+ */
+template <class TOwn, class TObject>
```
rubidium42: Doxygen misses for these parameters.

rubidium42: TOwn -> TInstance or TImpl? Now it's a bit confusing w.r.t. Towns.

src/saveload/vehicle_sl.cpp
TrueBrain: Later PRs mostly remove this; it is simply as a stop-gap. As I knew if I would to copy/paste the code and create identical functions I would get the complaints I have duplicated code :P This is the best I could do in the chain of PRs :)  Adding an abstraction now would just be a waste of my time, basically, as I would remove it 1 PR later :P I am lazy :D

But even after all this work there is plenty of room to refactor the code to have less code. But I left this out of this set of PRs, as reviewing this is already difficult enough :)

(just to confirm for myself, after #9322:
```
$ grep -R -n GenericSaveLoad ../src/ | wc -l
0
```
)

src/saveload/station_sl.cpp
```diff
@@ -374,89 +374,118 @@ static void Ptrs_STNS()
...
+		SlObject(bst, this->GetDescription());
+	}
+
+	void Save(BaseStation *bst) const override { this->GenericSaveLoad(bst); }
+	void Load(BaseStation *bst) const override { this->GenericSaveLoad(bst); }
+	void FixPointers(BaseStation *bst) const override { this->GenericSaveLoad(bst); }
 };
 
-static OldPersistentStorage _old_st_persistent_storage;
+class SlStationCommon : public DefaultSaveLoadHandler<SlStationCommon, BaseStation> {
```
rubidium42: Maybe SlNormalStation as now it sort-of implies there might be a sub class?

src/saveload/station_sl.cpp
```diff
@@ -374,89 +374,118 @@ static void Ptrs_STNS()
...
-static const SaveLoad _base_station_desc[] = {
-	      SLE_VAR(BaseStation, xy,                     SLE_UINT32),
-	      SLE_REF(BaseStation, town,                   REF_TOWN),
-	      SLE_VAR(BaseStation, string_id,              SLE_STRINGID),
-	     SLE_SSTR(BaseStation, name,                   SLE_STR | SLF_ALLOW_CONTROL),
-	      SLE_VAR(BaseStation, delete_ctr,             SLE_UINT8),
-	      SLE_VAR(BaseStation, owner,                  SLE_UINT8),
-	      SLE_VAR(BaseStation, facilities,             SLE_UINT8),
-	      SLE_VAR(BaseStation, build_date,             SLE_INT32),
+class SlStationBase : public DefaultSaveLoadHandler<SlStationBase, BaseStation> {
```
rubidium42, line 508: I'd note in the comment for this one that SlWaypoint also makes use of this.

rubidium42 (approved):

PR #9341, opened 2021-06-09
Add: '-X' option to ignore global folders in the search path

TrueBrain:
## Motivation / Problem

Imagine this: you have 400 NewGRFs and 4 basesets in your `~/.local/share/openttd` folder. You want to run some automated tests for OpenTTD that has 10 NewGRFs and 1 baseset. Owh no, it still looks all those other NewGRFs too!

But wait, we have `-c` you say! Yeah .. that doesn't work for the global folders.

Okay okay .. so I just rename that `~/.local/share/openttd` folder every time .. well .. I am crazy, but not that crazy.

## Description

So I present you with an alternative: `-X`. This setting completely ignores global folders and only use the ones you gave with `-c`. Nice.

```
This is extreme useful for automated testing. Without this, OpenTTD
will always look in your personal-dir (like ~/.local/share/openttd
or %USER%\Documents\OpenTTD). For most users this is exactly what
we want, that there is a shared place for all their files.

However, for automated testing this is rather annoying, as your
local development files influence the automated test. As such,
'-X' counters this, and only gives the local folders. This is
especially useful in combination with '-x' and '-c'.
```

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -197,6 +197,7 @@ static void ShowHelp()
 		"  -M music_set        = Force the music set (see below)\n"
 		"  -c config_file      = Use 'config_file' instead of 'openttd.cfg'\n"
 		"  -x                  = Never save configuration changes to disk\n"
+		"  -X                  = Don't use global folders to search for files\n"
 		"  -q savegame         = Write some information about the savegame and exit\n"
 		"\n",
 		lastof(buf)
@@ -506,6 +507,7 @@ static const OptionData _options[] = {
 	 GETOPT_SHORT_VALUE('G'),
 	 GETOPT_SHORT_VALUE('c'),
 	 GETOPT_SHORT_NOVAL('x'),
+	 GETOPT_SHORT_NOVAL('X'),
 	 GETOPT_SHORT_VALUE('q'),
 	 GETOPT_SHORT_NOVAL('h'),
 	GETOPT_END()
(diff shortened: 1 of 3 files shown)


michicc (dismissed):

src/openttd.cpp
```diff
@@ -650,7 +654,7 @@ int openttd_main(int argc, char *argv[])
 		 *
 		 * The next two functions are needed to list the graphics sets. We can't do them earlier
 		 * because then we cannot show it on the debug console as that hasn't been configured yet. */
-		DeterminePaths(argv[0]);
+		DeterminePaths(argv[0], false);
```
rubidium42: Shouldn't this be `only_local_path`? Otherwise `openttd -X -h` lists base sets that `openttd -X -I/S/M ...` cannot find

rubidium42:
Shouldn't docs/openttd.6 (the man page) be updated with this new parameter?

TrueBrain:
> 
> 
> Shouldn't docs/openttd.6 (the man page) be updated with this new parameter?

lol, we have that file? Is that still in use ...

Either way, that file hasn't been updated in a while, and I have no clue how I would be to update it. Even worse, it uses different ways to describe the same settings than `openttd.cpp` .. so I think it should just be overhauled completely. But 0 clue how to do that, and I really do not feel like spending time on it :P (sorry, but at least I am honest :D)

rubidium42 (approved):

PR #9342, opened 2021-06-09
Change: allow pause/unpause console command in single player too

TrueBrain:
## Motivation / Problem

For some weird reason, `pause`/`unpause` console command only works in network games. I couldn't think of a single reason for that to make sense.

Mainly I needed this to ensure games are unpaused after game-start for automated testing. Now in `game_start.scr` I can do `unpause`, and I am sure the game is not paused. Sweet.

## Description

Fixed the weird limitation.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -160,11 +160,24 @@ DEF_CONSOLE_HOOK(ConHookNoNetwork)
 	return CHR_ALLOW;
 }
 
+/**
+ * Check if are either in singleplayer or a server.
+ * @return True iff we are either in singleplayer or a server.
+ */
+DEF_CONSOLE_HOOK(ConHookServerOrNoNetwork)
+{
+	if (_networking && !_network_server) {
+		if (echo) IConsoleError("This command is only available to a network server.");
+		return CHR_DISALLOW;
+	}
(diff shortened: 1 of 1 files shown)


src/console_cmds.cpp
```diff
@@ -160,6 +160,26 @@ DEF_CONSOLE_HOOK(ConHookNoNetwork)
 	return CHR_ALLOW;
 }
 
+/**
+ * Check whether we are in singleplayer mode or are the server.
```
rubidium42: The name of the hook and content of the comment do not match with the implementation. It does not check whether its the server or not in network, it also checks for being in the menu.
I don't really have an useful alternative name for the hook though, unless making it super explicit like "ConHookServerOrSinglePlayerOrEditor".

TrueBrain: Been fiddling with this, but basically I think there is no right answer.

Updated the comment to better reflect reality, but as far as the function-name goes .. I think this name is already confusing enough, adding more letters to it doesn't help :) So yeah, I also don't have any useful alternatives, so lets go with this :)

TrueBrain: Nevermind, found a clean solution :)

rubidium42 (approved):

PR #9343, opened 2021-06-09
Add: adhere the autosave_on_exit setting for Null videodriver

TrueBrain:
## Motivation / Problem

Our scripting stuff is rarely useful, but I now had a use, it wasn't supplying. 

For automated testing, I want to load an (old) savegame, run for a few ticks, and save&exit, so I could compare it with later results.

But .. there is no way to do this "run for a few ticks, and save&exit" currently. I could see two possible solutions:

1) have a console command that "delayed executes" a command, so I can say: "execute in 200 ticks: save & exit".
2) have a script that is called just after the game is stopping but before anything is shut down.
3) (tnx to @glx22) make sure `autosave_on_exit` also works for the null driver.

## Description

I went with 3), as 2) felt weird and dirty. 1) would be ideal, but is a lot more work.

```
This is especially useful for automated-testing, to make a save
when the game quits while using "-vnull:ticks=N".
```

## Checklist for review

### DIFF
--- a/src/video/null_v.cpp
+++ b/src/video/null_v.cpp
@@ -10,6 +10,7 @@
 #include "../stdafx.h"
 #include "../gfx_func.h"
 #include "../blitter/factory.hpp"
+#include "../saveload/saveload.h"
 #include "../window_func.h"
 #include "null_v.h"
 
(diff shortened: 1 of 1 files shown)


glx22:
Can't use the save at exit setting  https://github.com/OpenTTD/OpenTTD/blob/12de256b72fb7442cd84c6a0e0c6a61636255a6b/src/openttd.cpp#L817 ?

TrueBrain:
> 
> 
> Can't use the save at exit setting
> 
> https://github.com/OpenTTD/OpenTTD/blob/12de256b72fb7442cd84c6a0e0c6a61636255a6b/src/openttd.cpp#L817
> ?

Completely forgot about that setting, yes, that works too :) Still needs a minor change ;)

Nice, tnx :)

bin/scripts/game_exit.scr.example
```diff
@@ -0,0 +1 @@
+save on_exit
```
glx22: Left-over file I think.

rubidium42 (approved):

PR #9344, opened 2021-06-09
Change: Reduce real sprite groups if possible.

PeterN:
## Motivation / Problem

Many NewGRF features using Act 3/2/1 chains don't require loading/loaded stages (which are determined by the `RealSpriteGroup`), however by spec they are still created, and evaluated on every chain resolve.

For features that do use loading/loaded stages, it is also fairly common for the same sprite or callback result to be returned.

## Description

This change skips creating a RealSpriteGroup if there is only one result, or if all results are the same.
It also deduplicates the different ResolveReal() resolver function for resolvers which don't care about loading/loaded stages.

## Limitations

The deduplication does not take account of the order of checking loading/loaded results which was previously inconsistent, therefore it's possible that if loading/loaded stages were supplied for a feature that does not expect it, a different result will now be used. But if this is the case then I think NewGRF is out-of-spec anyway.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -109,7 +109,7 @@ public:
 	int skip_sprites;         ///< Number of pseudo sprites to skip before processing the next one. (-1 to skip to end of file)
 
 	/* Currently referenceable spritegroups */
-	SpriteGroup *spritegroups[MAX_SPRITEGROUP + 1];
+	const SpriteGroup *spritegroups[MAX_SPRITEGROUP + 1];
 
 	/** Clear temporary data before processing the next file in the current loading stage */
 	void ClearDataForNextFile()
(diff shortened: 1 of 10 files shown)


michicc:
Different result means at worst a different real sprite, right? A different CB chain would be worse, but we never promised bug-for-bug compatibility, so I'd consider a wrong sprite for an out-of-spec GRF quite acceptable.

PeterN:
In theory this can also affect callback results as a plain action 2 can return callback results (at least our code supports it, whether the spec intends that is another matter)... but:

In fact after checking the NewGRF spec it seems we did it wrong for railtypes and roadtypes (which is fun, because it is our spec) as the code tested loading first then loaded, when the spec says loaded must be 1 and loading must be 0.

michicc (approved):

PR #9346, opened 2021-06-09
Feature: [NewGRF] Maximum curve speed modifier for rail vehicles

vituscze:
## Motivation / Problem

Prior to https://github.com/OpenTTD/OpenTTD/commit/0f91cb04791b329c5e457588a39d1e7a43e43136, rail type's `curve_speed` essentially worked as a vehicle property. A rail vehicle could set its curve speed advantage to an arbitrary value by choosing the proper "home" rail type and the advantage would apply when traveling on compatible rail types.

Currently, the only way a rail vehicle can control its curve speed bonus is via the tilt property, which is a flat 20% bonus. There are no other ways in which the original behavior can be recovered. NUTS (and possibly other NewGRFs) relied on that behavior quite heavily.

## Description

This PR fixes the stated problem by introducing a new rail vehicle property: `curve_speed_mod` (0x2E). A 16 bit signed fixed-point binary number with 8 fractional bits, ranging from -128 to 127.996. 0 represents no bonus (default behavior), positive numbers increase maximum curve speed, negative numbers decrease. As an example, the value 1.0 (0x0100) would double the vehicle's speed in curves.

The way this property works is analogous to tilt. Each part of the consist has its own `curve_speed_mod` and the curve speed advantage of the entire consist is given by its weakest link.

Moreover, NewGRFs can also affect `curve_speed_mod` via callback 0x36. The resulting 15 bits represent values from -64 to 63.996. To support negative modifiers, `GetEngineProperty` and `GetVehicleProperty` now optionally allow negative numbers (existing uses are unaffected).

This change also expands the possible design space compared to the original behavior: NewGRFs can provide powerful trains that travel slowly in curves; wagons that, when refitted to fragile cargo, must travel slowly in curves; and so on.

NML support: https://github.com/OpenTTD/nml/pull/222

## Limitations

No known limitations.

## Checklist for review

### DIFF
--- a/src/engine_type.h
+++ b/src/engine_type.h
@@ -59,6 +59,7 @@ struct RailVehicleInfo {
 	byte tractive_effort;           ///< Tractive effort coefficient
 	byte air_drag;                  ///< Coefficient of air drag
 	byte user_def_data;             ///< Property 0x25: "User-defined bit mask" Used only for (very few) NewGRF vehicles
+	int16 curve_speed_mod;          ///< Modifier to maximum speed in curves (fixed-point binary with 8 fractional bits)
 };
 
 /** Information about a ship vehicle. */
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1333,6 +1333,10 @@ static ChangeInfoResult RailVehicleChangeInfo(uint engine, int numinfo, int prop
 				break;
 			}
 
+			case PROP_TRAIN_CURVE_SPEED_MOD: // 0x2E Curve speed modifier
+				rvi->curve_speed_mod = buf->ReadWord();
+				break;
+
 			default:
 				ret = CommonVehicleChangeInfo(ei, prop, buf);
 				break;
--- a/src/newgrf_engine.cpp
+++ b/src/newgrf_engine.cpp
@@ -1177,16 +1177,23 @@ uint16 GetVehicleCallbackParent(CallbackID callback, uint32 param1, uint32 param
 
 
 /* Callback 36 handlers */
-uint GetVehicleProperty(const Vehicle *v, PropertyID property, uint orig_value)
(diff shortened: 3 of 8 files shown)


vituscze:
Together with @V453000, we've also prepared a modified version of NUTS and a savegame that showcases positive and negative curve speed modifiers as well as curve speed mod callbacks.

savegame: [curve_test.zip](https://github.com/OpenTTD/OpenTTD/files/6626896/curve_test.zip)
NewGRF: https://dev.openttdcoop.org/attachments/download/9875/nuts-curve_speed_mod.grf

andythenorth:
Curve speed -99% https://en.wikipedia.org/wiki/4-14-4

V453000:
I went a bit further and replaced all TRAIN_FLAG_TILT with increased curve_speed_mod. This is so much cleaner and I can't wait when I get the courage to finetune the values further.

Not sure if it's helpful at all (probably not), but I'll provide it here too.

NewGRF: https://dev.openttdcoop.org/attachments/download/9876/nuts-curve_speed_mod.grf

https://github.com/V453000/NUTS/commit/1b6e15cb014d7ddba15460080a4480b8d4670c11

michicc:
Makes sense to me and the code looks clean, but I might miss some intimate NewGRF knowledge to notice any problems.

TrueBrain (approved):
I think every dev by now has said: yeah, this looks fine, just the NewGRF part I don't know about :D So maybe we should just bite the bullet and go for it .. if these three gentlemen in this thread can't figure out a way to extend the NewGRF specs nicely, who can! :D

Guess the most work is making sure https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status is updated properly, which is mainly to track if we truly updated everything involved here.

michicc:
Spec state page updated, will write NFO wiki part.

PR #9352, opened 2021-06-11
Codechange: use the fmt library for simpler debug formats

rubidium42:
## Motivation / Problem

The format codes for `printf` are complicated and annoying to work with. When changing something from C-string to std::string you need to explicitly add `.c_str()`, but also for some types you need to "inject" `PRINTF_SIZE`.

Preferably we would use something in the standard library, but formatting has landed in C++20 so... almost there but not quite yet. There is however the "fmt" library that has the same API as C++20 just in a different namespace, so we can use that as a stop-gap measure until we go to C++20. See https://fmt.dev/7.1.3/api.html#compatibility-with-c-20-std-format

## Description

The vendoring of the "fmt" library and full conversion of all DEBUG statements to Debug.

## Limitations

* ~9000 lines of extra code due to the vendoring of the library.
* There is a set of `fmt::printf` functions, however they do not support things like `%ld` so using that as a migration step is out of the question.
* The error messages on wrong formats can be very unhelpful, though it does have compile time validation of the format strings.
* There are a lot of inconsistencies in the debug messages. Those are not in scope here, unless someone wants to patch that on top of this. That I'll gladly accept.

## Checklist for review

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -100,46 +100,43 @@ char *DumpDebugFacilityNames(char *buf, char *last)
 
 /**
  * Internal function for outputting the debug line.
- * @param dbg Debug category.
- * @param buf Text line to output.
+ * @param level Debug category.
+ * @param message The message to output.
  */
-void debug_print(const char *dbg, const char *buf)
+void DebugPrint(const char *level, const std::string &message)
 {
 	if (_debug_socket != INVALID_SOCKET) {
-		char buf2[1024 + 32];
-
-		seprintf(buf2, lastof(buf2), "%sdbg: [%s] %s\n", GetLogPrefix(), dbg, buf);
+		std::string msg = fmt::format("{}dbg: [{}] {}\n", GetLogPrefix(), level, message);
 		/* Sending out an error when this fails would be nice, however... the error
 		 * would have to be send over this failing socket which won't work. */
-		send(_debug_socket, buf2, (int)strlen(buf2), 0);
+		send(_debug_socket, msg.c_str(), (int)msg.size(), 0);
 		return;
 	}
-	if (strcmp(dbg, "desync") == 0) {
+	if (strcmp(level, "desync") == 0) {
 		static FILE *f = FioFOpenFile("commands-out.log", "wb", AUTOSAVE_DIR);
 		if (f == nullptr) return;
 
-		fprintf(f, "%s%s\n", GetLogPrefix(), buf);
+		fprintf(f, "%s%s\n", GetLogPrefix(), message.c_str());
 		fflush(f);
 #ifdef RANDOM_DEBUG
-	} else if (strcmp(dbg, "random") == 0) {
+	} else if (strcmp(level, "random") == 0) {
 		static FILE *f = FioFOpenFile("random-out.log", "wb", AUTOSAVE_DIR);
 		if (f == nullptr) return;
 
-		fprintf(f, "%s\n", buf);
+		fprintf(f, "%s\n", message.c_str());
 		fflush(f);
 #endif
 	} else {
-		char buffer[512];
-		seprintf(buffer, lastof(buffer), "%sdbg: [%s] %s\n", GetLogPrefix(), dbg, buf);
+		std::string msg = fmt::format("{}dbg: [{}] {}\n", GetLogPrefix(), level, message);
 #if defined(_WIN32)
 		wchar_t system_buf[512];
-		convert_to_fs(buffer, system_buf, lengthof(system_buf));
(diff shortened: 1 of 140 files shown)


glx22:
Ok, for targets using `vcpkg` you can add `fmt` to the install list. For MinGW it's also `fmt`. Don't know for Linux and Emscripten.

LordAro:
Isn't it a header-only library anyway? Might as well just vendor it

rubidium42:
> Ok, for targets using `vcpkg` you can add `fmt` to the install list. For MinGW it's also `fmt`. Don't know for Linux and Emscripten.

I tried looking for that, but couldn't find it in the source code. So I guess it is in some helper repository. In any case, I have not bothered with that as the decision for "vendoring" it or not needs to be made first. If we are going to vendor it, then adding the library to those lists has been quite futile.

> Isn't it a header-only library anyway? Might as well just vendor it

You can use it as header-only library or as normal library. For the former you need to add an extra define before the include, though that should not be a problem.

src/debug.h
```diff
@@ -12,9 +12,11 @@
 
 #include "cpu.h"
 #include <chrono>
+#define FMT_HEADER_ONLY
+#include "3rdparty/fmt/format.h"
```
TrueBrain, line 15: as I would expect more places to start using `fmt`, maybe we should make a `core/fmt_type.hpp` or what-ever that does this header-dance for us? Just to make sure we don't screw that up over the years :P

src/debug.h
```diff
@@ -28,11 +30,13 @@
...
- * Output a line of debugging information.
- * @param name Category
- * @param level Debugging level, higher levels means more detailed information.
+ * Ouptut a line of debugging information.
+ * @param name The category of debug information.
+ * @param level The maximum debug level this message should be shown at. When the debug level for this category is set lower, then the message will not be shown.
+ * @param format_string The formatting string of the message.
  */
-#define DEBUG(name, level, ...) if ((level) == 0 || _debug_ ## name ## _level >= (level)) debug(#name, __VA_ARGS__)
+#define Debug(name, level, format_string, ...) if ((level) == 0 || _debug_ ## name ## _level >= (level)) DebugPrint(#name, fmt::format(FMT_STRING(format_string), ## __VA_ARGS__))
```
TrueBrain, line 37: Maybe not for this PR, but I guess we could remove this macro and have a `_debug_level[]` instead of this. Can be a `std::map` I guess. Dunno if it is a good idea, but less macros can be a good thing :D

src/debug.cpp
```diff
@@ -100,67 +100,46 @@ char *DumpDebugFacilityNames(char *buf, char *last)
 
 /**
  * Internal function for outputting the debug line.
- * @param dbg Debug category.
- * @param buf Text line to output.
+ * @param level Debug category.
+ * @param message The message to output.
  */
-static void debug_print(const char *dbg, const char *buf)
+void DebugPrint(const char *level, const std::string &message)
```
TrueBrain, line 106: for some reason I was expecting a `std::string_view` :D

TrueBrain (commented):
This would make life so much better in many ways (how often I get warnings because I used `%d` instead of `%lu` .. I cannot even start to tell you about).

PS: I vote to vendor this library; it is small enough after all.
PPS: your description is pretty much out-of-date :P

src/debug.h
michicc: Alternatively, the define could be in `stdafx.h`, then it is always set whenever `fmt` is used.

src/debug.cpp
rubidium42: That requires some more work further down the line to do nicely.

src/debug.h
rubidium42: Good, so there is some consensus to putting in in `stdafx.h` ;)

src/debug.h
rubidium42: For performance reasons I would always leave this as a macro, although the macro might potentially be simplified a bit. From a maintenance standpoint having all the different names with their levels in a `std::map` would definitely be nice, though I'm not so sure about the performance implications as the lookup for a map are not negligible. That could be researched, but is outside of the scope of this PR.

With the macro the resolving of the parameters only needs to be done if the debug message is going to be printed. With a inline templated function the parameter needs to be resolved even when the debug message is not going to be printed. See e.g. https://godbolt.org/z/6KcKac7j8 for the subtle difference in the generated code, though the effects go from "almost never executed, unless the user cranks the debug level up" to "executed every loop" making unused DEBUGs much more expensive.

TrueBrain (approved):

PR #9355, opened 2021-06-12
Codechange: Use dynamic string list for contents of land info window.

michicc:
## Motivation / Problem

Land info window is using a fixed char array for storing the window's contents. As a safety margin to account for any possibilities, the buffer is a lot larger than the typical contents needs.

## Description

Instead of the fixed array that is larger than needed, store the text in a dynamic StringList. While we're at it, remove the remaining c-style string bits.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -32,6 +32,9 @@
 
 #include "table/strings.h"
 
+#include <sstream>
+#include <iomanip>
+
 #include "safeguards.h"
 
 /** Method to open the OSK. */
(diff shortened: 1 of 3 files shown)


src/misc_gui.cpp
```diff
@@ -97,21 +92,19 @@ class LandInfoWindow : public Window {
...
-			if (i == 0) size->height += 4;
-		}
-
-		if (!StrEmpty(this->landinfo_data[LAND_INFO_MULTICENTER_LINE])) {
-			uint width = GetStringBoundingBox(this->landinfo_data[LAND_INFO_MULTICENTER_LINE]).width + WD_FRAMETEXT_LEFT + WD_FRAMETEXT_RIGHT;
-			size->width = std::max(size->width, std::min(300u, width));
-			SetDParamStr(0, this->landinfo_data[LAND_INFO_MULTICENTER_LINE]);
-			size->height += GetStringHeight(STR_JUST_RAW_STRING, size->width - WD_FRAMETEXT_LEFT - WD_FRAMETEXT_RIGHT);
+		for (size_t i = 0; i < this->landinfo_data.size(); i++) {
+			if (i == this->landinfo_data.size() - 1 && this->last_is_multicenter) {
```
rubidium42: What is the reason you have chosen to introduce `last_is_multicenter` and some fudging with the element being the last in the vector instead of having a `cargo_acceptance` string where you put `line.str()` in and check the string for emptiness like was done before?

michicc: Because I didn't think of it. It seems a lot nicer, though. Expect an update soon.

rubidium42 (approved):

PR #9367, opened 2021-06-13
Fix #9358: TarScanner no longer ignores empty files

### LINKED ISSUE #9358: Bug Report - tar loader ignores empty files. An AI that requires such file fails to start
## Version of OpenTTD

OpenTTD 1.11.2 and Windows 10
## Expected result

.tar loader shouldn't ignore empty files.
## Actual result

.tar loader ignores an empty file which is required file for the AI to start without erroring out.
## Steps to reproduce

The AI named Mungo, version 5 has an empty file, but it's required for the AI to start.
See conversation https://webster.openttdcoop.org/index.php?channel=openttd&date=1623369600#1623423363

SamuXarick:
## Motivation / Problem
An AI requires a file with 0 bytes to be able to start, but the tar scanner skips empty files, causing the AI to fail with a "file not found" error.
See #9358 

## Description
The problem is solved by allowing tar scanner to add empty files.

## Limitations
It was suggested to fix it this way, but there's some uncertainty about it. I'm not familiar with this code.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -541,9 +541,6 @@ bool TarScanner::AddFile(const std::string &filename, size_t basepath_length, co
 		switch (th.typeflag) {
 			case '\0':
 			case '0': { // regular file
-				/* Ignore empty files */
-				if (skip == 0) break;
-
 				if (strlen(name) == 0) break;
 
 				/* Store this entry in the list */


LordAro (approved):
Probably fine. Commit might want rewording on squash

PR #9368, opened 2021-06-13
Codechange: Remove last FOR_XXX macros

glx22:
## Motivation / Problem
I finally finished removing `FOR_XXX` macros.

(Now Intellisense is more usable)

## Description
- Replaced manually allocated `HeightMap::h` with a vector
- Used a span for `_sorted_cargo_specs` sub-range of standard `CargoSpec`
- Introduced an iterator for bitset (and used it)

## Checklist for review

### DIFF
--- a/src/cargotype.cpp
+++ b/src/cargotype.cpp
@@ -149,8 +149,8 @@ SpriteID CargoSpec::GetCargoIcon() const
 	return sprite;
 }
 
-std::vector<const CargoSpec *> _sorted_cargo_specs; ///< Cargo specifications sorted alphabetically by name.
-uint8 _sorted_standard_cargo_specs_size;            ///< Number of standard cargo specifications stored in the _sorted_cargo_specs array.
+std::vector<const CargoSpec *> _sorted_cargo_specs;   ///< Cargo specifications sorted alphabetically by name.
+span<const CargoSpec *> _sorted_standard_cargo_specs; ///< Standard cargo specifications sorted alphabetically by name.
 
 /** Sort cargo specifications by their name. */
 static bool CargoSpecNameSorter(const CargoSpec * const &a, const CargoSpec * const &b)
@@ -196,13 +196,16 @@ void InitializeSortedCargoSpecs()
 	/* Sort cargo specifications by cargo class and name. */
 	std::sort(_sorted_cargo_specs.begin(), _sorted_cargo_specs.end(), &CargoSpecClassSorter);
 
+	/* Count the number of standard cargos and fill the mask. */
 	_standard_cargo_mask = 0;
-
-	_sorted_standard_cargo_specs_size = 0;
+	uint8 nb_standard_cargo = 0;
 	for (const auto &cargo : _sorted_cargo_specs) {
 		if (cargo->classes & CC_SPECIAL) break;
-		_sorted_standard_cargo_specs_size++;
+		nb_standard_cargo++;
 		SetBit(_standard_cargo_mask, cargo->Index());
 	}
+
+	/* _sorted_standard_cargo_specs is a subset of _sorted_cargo_specs. */
+	_sorted_standard_cargo_specs = { _sorted_cargo_specs.data(), nb_standard_cargo };
 }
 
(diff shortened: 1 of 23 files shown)


src/cargotype.cpp
```diff
@@ -198,11 +198,13 @@ void InitializeSortedCargoSpecs()
...
-	_sorted_standard_cargo_specs_size = 0;
+	uint8 sorted_standard_cargo_specs_size = 0;
 	for (const auto &cargo : _sorted_cargo_specs) {
 		if (cargo->classes & CC_SPECIAL) break;
-		_sorted_standard_cargo_specs_size++;
+		sorted_standard_cargo_specs_size++;
 		SetBit(_standard_cargo_mask, cargo->Index());
 	}
+
+	_sorted_standard_cargo_specs = { _sorted_cargo_specs.data(), sorted_standard_cargo_specs_size };
```
TrueBrain: This is a vector, so you should just be able to convert that to a `span<>`? Not really sure why the size needed to be calculated first?

src/core/bitmath_func.hpp
```diff
@@ -345,20 +345,55 @@ static inline T ROR(const T x, const uint8 n)
...
+	};
+
+	BitIterator(Tbitset bitset) : bitset(bitset) {}
+	Iterator begin() { return Iterator(this->bitset); }
+	Iterator end() { return Iterator(static_cast<Tbitset>(0)); }
+	bool empty() { return this->begin() == this->end(); }
+
+private:
+	Tbitset bitset;
+};
```
TrueBrain, line 371: Sadly, I have not enough knowledge to rate this part of the PR in any way :D @LordAro , mind giving this a look? Tnx!

src/core/bitmath_func.hpp
```diff
@@ -320,31 +320,6 @@ static inline T ROR(const T x, const uint8 n)
 	return (T)(x >> n | x << (sizeof(x) * 8 - n));
 }
 
-/**
```
TrueBrain, line 323: bit weird commit message .. `codechange` .. `remove`... :D

Maybe something like:

`Cleanup: remove now unused FOR_EACH_SET_BIT_EX macro`
?

src/cargotype.cpp
glx22: Because we want a subset of the vector (most of the time it will be full vector though)

TrueBrain: Ah, the `break` there .. right .. that is sneaky :D

I think this deserves a comment somewhere :P

src/core/bitmath_func.hpp
LordAro: Dunno much either. Part of me wonders whether std::bitset might be useable somehow (though may well require a fair bit more refactoring to do so)

rubidium42: Though bitset doesn't have iterators (or rather begin/end), so can it iterate over them? With my naive check you can't, and I think that was the whole point of making this iterator.

src/core/bitmath_func.hpp
```diff
@@ -345,20 +345,55 @@ static inline T ROR(const T x, const uint8 n)
...
- * @param bitset_value The value which we check for set bits.
- */
-#define FOR_EACH_SET_BIT(bitpos_var, bitset_value) FOR_EACH_SET_BIT_EX(uint, bitpos_var, uint, bitset_value)
+ /**
+ * Iterable ensemble of each set bit in a value.
+ * @tparam Tbitpos Type of the position variable.
+ * @tparam Tbitset Type of the bitset value.
+*/
+template <typename Tbitpos = uint, typename Tbitset = uint>
+struct BitIterator {
```
rubidium42: The name of the class does not imply that it only considers set bits, nor that it actually returns the position and not be bit value.

TrueBrain:
I have no more comments. @rubidium42 : this okay with you too?

PS: really cool finally these really old macros are goneeeeeeee. Nice job @glx22 :D

TrueBrain (approved):
I am going to assume the silence is a yes :)

PR #9370, opened 2021-06-13
Fix a99ac62: fmt's include of cassert breaks our assert logic

rubidium42:
## Motivation / Problem

Including `<cassert>` breaks our asserts and that could lead to e.g. the pool checks to fail as in some cases the side effects of asserts are ran, and in others they are not.

## Description

Remove `<cassert>` from being included in "format-inl.h" and add a comment there so we are unlikely to re add it when updating fmt.

## Limitations

None as far as I know.

## Checklist for review

### DIFF
--- a/src/3rdparty/fmt/format-inl.h
+++ b/src/3rdparty/fmt/format-inl.h
@@ -8,7 +8,7 @@
 #ifndef FMT_FORMAT_INL_H_
 #define FMT_FORMAT_INL_H_
 
-#include <cassert>
+/* Do not include cassert as that breaks our own asserts. */
 #include <cctype>
 #include <climits>
 #include <cmath>


PeterN (approved):

TrueBrain:
To leave a bit of background:

We define 3 ways of compiling OpenTTD (in relation to asserts):
- Debug builds: asserts enabled, `NDEBUG` not defined
- Pre-release builds (RCs, beta, ..): asserts enabled, `NDEBUG` defined
- Release builds: asserts disabled and `NDEBUG` defined

Now there are all kind of things at play here, but the important thing to know: with `Release` and `RelWithDebInfo` as build-type, CMake automatically adds `NDEBUG`.
Also, `cassert` will noop `assert()` if `NDEBUG` is set.

To allow this "in-between" state for pre-releases, we have `WITH_ASSERT`. If this is enabled, no matter if `NDEBUG` is set or not, `assert` will be defined. We do this in `stdafx.h`.

Now the final pieces of this puzzle: `#include <cassert>` includes `assert.h` which does it at-most to ignore any header-validation and say: fuck you guys, if you include me, I will take over `assert()`.

Normally, the flow is:
- Every `.cpp` file includes `stdafx.h`
- `stdafx.h` includes `cassert` first
- `stdafx.h` fixes assert in case of `NDEBUG` and `WITH_ASSERT`

Now with fmt it breaks, because:
- A `.cpp` file includes `stdafx.h` and `debug.h`
- `stdafx.h` includes `cassert` first
- `stdafx.h` fixes assert in case of `NDEBUG` and `WITH_ASSERT`
- `debug.h` file includes fmt-library
- fmt-library includes `cassert` again
- `cassert` takes over our `assert()`

This means the game thinks it runs with asserts, but `assert()` doesn't do anything for files using `debug.h` but does something for files that do not. This not only gives compiler warnings, but also triggers an `assert()` at some point, as assert-only code is executed in one file but not in another.

In other words: NEVER include `cassert` after `stdafx.h` or we will run into trouble. That also goes for 3rdparty libraries.

PR #9371, opened 2021-06-13
Assorted C-string -> std::string conversions in network

rubidium42:
## Motivation / Problem

The ongoing conversion of C-string to std::string.

## Description

All kinds of small, but scattered in networking code, changes to go from C-strings to std::string.
* Use fmt in the network address creation.
* Use std::string is IsInNetmask; https://godbolt.org/z/n8TTM5osz is a simple sort of unit test for this change.
* Let GetHostName/GetClientIP directly return the std::string reference instead of the C-string, to prevent converting is back to a std::string later on (e.g. in IsInNetmask), but also in other locations.
* Let NetworkError just return a reference to its std::string.
* Use fmt in HTTP request creation and pass the host as std::string reference.
* Use std::string_view for network compatability checks; https://godbolt.org/z/sj9Ya3q6W is a simple unit test for this change.
* Use std::string for returning the client name.
* Use std::string for (temporarily) storing the name of the downloading content.

## Limitations

Contains #9372.

## Checklist for review

### DIFF
--- a/src/network/core/tcp_http.cpp
+++ b/src/network/core/tcp_http.cpp
@@ -32,7 +32,7 @@ static std::vector<NetworkHTTPSocketHandler *> _http_connections;
  * @param depth    the depth (redirect recursion) of the queries
  */
 NetworkHTTPSocketHandler::NetworkHTTPSocketHandler(SOCKET s,
-		HTTPCallback *callback, const char *host, const char *url,
+		HTTPCallback *callback, const std::string &host, const char *url,
 		const char *data, int depth) :
 	NetworkSocketHandler(),
 	recv_pos(0),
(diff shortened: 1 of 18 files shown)


src/network/core/tcp_http.cpp
```diff
@@ -42,19 +42,16 @@ NetworkHTTPSocketHandler::NetworkHTTPSocketHandler(SOCKET s,
...
-	char *buffer = AllocaM(char, bufferSize);
-
 	Debug(net, 5, "[tcp/http] Requesting {}{}", host, url);
+	std::string request;
 	if (data != nullptr) {
-		seprintf(buffer, buffer + bufferSize - 1, "POST %s HTTP/1.0\r\nHost: %s\r\nUser-Agent: OpenTTD/%s\r\nContent-Type: text/plain\r\nContent-Length: %d\r\n\r\n%s\r\n", url, host, GetNetworkRevisionString(), (int)strlen(data), data);
+		request = fmt::format("POST {} HTTP/1.0\r\nHost: {}\r\nUser-Agent: OpenTTD/{}\r\nContent-Type: text/plain\r\nContent-Length: {}\r\n\r\n{}\r\n", url, host, GetNetworkRevisionString(), strlen(data), data);
 	} else {
-		seprintf(buffer, buffer + bufferSize - 1, "GET %s HTTP/1.0\r\nHost: %s\r\nUser-Agent: OpenTTD/%s\r\n\r\n", url, host, GetNetworkRevisionString());
+		request = fmt::format("GET {} HTTP/1.0\r\nHost: {}\r\nUser-Agent: OpenTTD/{}\r\n\r\n", url, host, GetNetworkRevisionString());
```
TrueBrain, line 50: despite the fact that this file clearly has `debug.h` included, I wonder if we should include `fmt` too, just in case someone removes `debug.h`. I often wonder this, about when we do add an include and when not. It is a bit hit and go :)

src/network/network_server.cpp
```diff
@@ -381,9 +379,7 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::SendCompanyInfo()
 	}
 
 	for (NetworkClientSocket *csi : NetworkClientSocket::Iterate()) {
-		char client_name[NETWORK_CLIENT_NAME_LENGTH];
-
-		((ServerNetworkGameSocketHandler*)csi)->GetClientName(client_name, lastof(client_name));
+		std::string client_name = ((ServerNetworkGameSocketHandler*)csi)->GetClientName();
```
TrueBrain: `static_const` while at it?

src/network/core/tcp_http.cpp
rubidium42: I rather do not add more `#include`s than needed. Even though you then "obfuscate" what you're doing, but otherwise I would have to add `#include <string>` about everywhere as well.

TrueBrain (dismissed):
YOLO!

TrueBrain (approved):
Sure!

src/settings.cpp
```diff
@@ -1272,7 +1272,7 @@ static void AILoadConfig(IniFile *ini, const char *grpname)
 				continue;
 			}
 		}
-		if (item->value.has_value()) config->StringToSettings(item->value->c_str());
+		if (item->value.has_value()) config->StringToSettings(*item->value);
```
rubidium42, line 1275: I'd rather use `item->value.value()` but... https://twitter.com/timur_audio/status/1184424350105161730?lang=nl

PR #9372, opened 2021-06-13
Fix: [Network] Determining GetNetworkRevisionString could overflow and underflow its buffer

rubidium42:
## Motivation / Problem

Depending on the content of `_openttd_revision` and `_openttd_revision_hash`, `GetNetworkRevisionString()` could overflow and underflow its buffer for non-tagged releases.
Practically it should not trigger if revision and revision_hash are set correctly by the build process.

In any case:
* when `_openttd_revision_hash` is less than 10 characters, it will read beyond the bounds.
* when `_openttd_revision` does not have a dash `-`, `hashofs` is effectively the negative hash. By "luck" it gets converted to `size_t` in the following if so `hashofs` get set to something more sane but if that were not to happen, `strecpy` would start at `nullptr`.
* when `_openttd_revision`'s length is less than `githash_suffix`'s length, then `hashofs` becomes negative and `strecpy` would start writing before the buffer.

So, when the revision and hash are, for whatever reason, not filled there are reads and writes beyond the bounds of buffers.

## Description

Rewrite the code to use `std::string` and perform the appropriate checks on sizes.
See https://godbolt.org/z/bbP3fxdxE for the behavior (old and new). When adding `-fsanitize=address` the example will be broken on the old version of the code.

## Limitations

None, except the output differing in cases of under- and overflow.
Consideration can be made whether to backport it or not; problem is that it uses {fmt}, so some other manner of safely getting the githash should be devised in case it is going to be backported.

## Checklist for review

### DIFF
--- a/src/network/core/game_info.cpp
+++ b/src/network/core/game_info.cpp
@@ -36,45 +36,36 @@ NetworkServerGameInfo _network_game_info; ///< Information about our game.
 
 /**
  * Get the network version string used by this build.
- * The returned string is guaranteed to be at most NETWORK_REVISON_LENGTH bytes.
+ * The returned string is guaranteed to be at most NETWORK_REVISON_LENGTH bytes including '\0' terminator.
  */
 const char *GetNetworkRevisionString()
 {
-	/* This will be allocated on heap and never free'd, but only once so not a "real" leak. */
-	static char *network_revision = nullptr;
+	static std::string network_revision;
 
-	if (!network_revision) {
-		/* Start by taking a chance on the full revision string. */
-		network_revision = stredup(_openttd_revision);
-		/* Ensure it's not longer than the packet buffer length. */
-		if (strlen(network_revision) >= NETWORK_REVISION_LENGTH) network_revision[NETWORK_REVISION_LENGTH - 1] = '\0';
-
-		/* Tag names are not mangled further. */
+	if (network_revision.empty()) {
+		std::string network_revision = _openttd_revision;
 		if (_openttd_revision_tagged) {
-			Debug(net, 3, "Network revision name: {}", network_revision);
-			return network_revision;
-		}
+			/* Tagged; do not mangle further, though ensure it's not too long. */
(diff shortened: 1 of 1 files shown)


src/network/core/game_info.cpp
```diff
@@ -36,45 +36,38 @@ NetworkServerGameInfo _network_game_info; ///< Information about our game.
...
-		if (_openttd_revision_tagged) {
-			Debug(net, 3, "Network revision name: {}", network_revision);
-			return network_revision;
+		if (!_openttd_revision_tagged) {
+			/* Prepare a prefix of the git hash. */
+			assert(_openttd_revision_modified < 3);
+			std::string githash_suffix = fmt::format("-{}{}", "gum"[_openttd_revision_modified], _openttd_revision_hash);
+			if (githash_suffix.size() > GITHASH_SUFFIX_LEN) githash_suffix.resize(GITHASH_SUFFIX_LEN);
+
+			/* Where did the hash start in the original string? Overwrite from that position, unless that would go past end of packet buffer length. */
```
TrueBrain: `packet buffer length`, hihi, guess that is a really old comment copy/pasted around :P

src/network/core/game_info.cpp
```diff
@@ -36,45 +36,38 @@ NetworkServerGameInfo _network_game_info; ///< Information about our game.
...
+			size_t hash_end = network_revision.find_last_of('-');
```
TrueBrain, line 57: Where your version differs from the original is at this point. Say, I have a really long branch name:

`more_string_in_networdfsfsdk`
The `_openttd_revision` would become something like `20210613-more_string_in_networdfsfsdk-m34ed1eeee4`. This is cut to `20210613-more_string_in_networdfsfs` (or something, didn't really check, but without the `-`). And now the "hash" is removed, leaving `20210613-`. This means the result is `20210613-m34ed1eeee4` instead of `20210613-more_string-m34ed1eeee4`.

I am not 100% sure if `_openttd_revision` allows this to start with, but really long branch names are pretty common.

In your godbolt you can see this problem in action with the above string. Basically, you were missing the case where the second `-` was passed the `NETWORK_REVISION_LENGTH` :)

TrueBrain (approved):

PR #9373, opened 2021-06-14
Cleanup: nested_array should not be accessed directly.

PeterN:
## Motivation / Problem

Window::nested_array is specifically documented with `Do not access directly, use #Window::GetWidget() instead.`

## Description

Replace accesses with GetWidget() as documented.

## Checklist for review

### DIFF
--- a/src/airport_gui.cpp
+++ b/src/airport_gui.cpp
@@ -485,7 +485,7 @@ public:
 				break;
 
 			case WID_AP_AIRPORT_LIST: {
-				int num_clicked = this->vscroll->GetPosition() + (pt.y - this->nested_array[widget]->pos_y) / this->line_height;
+				int num_clicked = this->vscroll->GetPosition() + (pt.y - this->GetWidget<NWidgetBase>(widget)->pos_y) / this->line_height;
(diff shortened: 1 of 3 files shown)


TrueBrain (approved):
Guess one more entry is violating the description, but using `GetWidget` won't work there :P

```
src/network/network_gui.cpp:         w->nested_array[WID_NG_HEADER] = this;
```

TrueBrain:
And as @LordAro doesn't appreciate me asking this on IRC: can't we make `nested_array` private to prevent this in the future? :)

(sorry @LordAro , just teasing you :P)

PeterN:
> Guess one more entry is violating the description, but using `GetWidget` won't work there :P
> 
> ```
> src/network/network_gui.cpp:         w->nested_array[WID_NG_HEADER] = this;
> ```

Yes, this one is rather "internal widget code" that happens to be in a standard GUI file.

PeterN:
> And as @LordAro doesn't appreciate me asking this on IRC: can't we make `nested_array` private to prevent this in the future? :)

Not without some major refactoring, a lot of this code is very C-style and isn't part of the Window class.

PR #9374, opened 2021-06-14
Change: make lists in savegames consistent

TrueBrain:
## Motivation / Problem

We have a bunch of different lists, which are all done slightly different.

- We have a "fixed-length" list, where the code dictates the length. No validation is done if that same value was also used when writing to disk, which means we have plenty of savegame bumps for cases where it did.
- We have lists like `SL_REFLIST` that have a 32bit length indicator.
- We have all kind of custom stuff with `SL_STRUCTLIST`, where sometimes the length is stored in the parent blob, saved/loaded to a global. But sometimes totally different solutions are used.

This makes self-described savegames really hard (the ultimate goal of this set of PRs, see #9322), but also means savegame-readers need to be really aware of what they are reading, and support a lot of different savegame-versions to do the right thing.

## Description

So .. as you might expect, I set out to fix that problem. Now the rules are simple:

- Any list, no matter who/what you are, has a gamma before the list to indicate how many entries are in the list.

In most cases this means you know the length. For `SL_STRUCT` this is not the case, and as such you need to know what struct is stored to figure out the length. This will be addressed in a later PR.

As using `SL_STRUCT` is normally only useful if it is an optional field (why else bother making it in a new struct, and not just embed it in the parent?), internally it is a `SL_STRUCTLIST` of either 0 or 1 in length. So if it is empty (nothing saved), the length-field is 0. Otherwise, length-field is 1.

## Limitations

- A lot of testing is involved, see #9339 and #9322 for details. So no known issues or limitations.
- Other solutions have been considered, instead of a length-field, like for example a "type", where the last entry is an empty one indicating "end". But given the current state of the code, that is vastly more work.

## Checklist for review

### DIFF
--- a/src/saveload/company_sl.cpp
+++ b/src/saveload/company_sl.cpp
@@ -372,8 +372,19 @@ public:
 
 class SlCompanyOldEconomy : public SlCompanyEconomy {
 public:
-	void GenericSaveLoad(CompanyProperties *c) const
+	void Save(CompanyProperties *c) const override
 	{
+		SlSetStructListLength(c->num_valid_stat_ent);
+		for (int i = 0; i < c->num_valid_stat_ent; i++) {
+			SlObject(&c->old_economy[i], this->GetDescription());
+		}
+	}
+
+	void Load(CompanyProperties *c) const override
+	{
+		if (!IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH)) {
+			c->num_valid_stat_ent = (uint8)SlGetStructListLength(UINT8_MAX);
+		}
 		if (c->num_valid_stat_ent > lengthof(c->old_economy)) SlErrorCorrupt("Too many old economy entries");
 
 		for (int i = 0; i < c->num_valid_stat_ent; i++) {
@@ -381,10 +392,7 @@ public:
 		}
 	}
 
-	void Save(CompanyProperties *c) const override { this->GenericSaveLoad(c); }
-	void Load(CompanyProperties *c) const override { this->GenericSaveLoad(c); }
-	void LoadCheck(CompanyProperties *c) const override { this->GenericSaveLoad(c); }
-	void FixPointers(CompanyProperties *c) const override { this->GenericSaveLoad(c); }
+	void LoadCheck(CompanyProperties *c) const override { this->Load(c); }
 };
 
 class SlCompanyLiveries : public DefaultSaveLoadHandler<SlCompanyLiveries, CompanyProperties> {
@@ -395,12 +403,33 @@ public:
 		SLE_CONDVAR(Livery, colour2, SLE_UINT8, SLV_34, SL_MAX_VERSION),
 	};
 
-	void GenericSaveLoad(CompanyProperties *c) const
+	/**
+	 * Get the number of liveries used by this savegame version.
+	 * @return The number of liveries used by this savegame version.
+	 */
+	size_t GetNumLiveries() const
 	{
-		int num_liveries = IsSavegameVersionBefore(SLV_63) ? LS_END - 4 : (IsSavegameVersionBefore(SLV_85) ? LS_END - 2: LS_END);
+		if (IsSavegameVersionBefore(SLV_63)) return LS_END - 4;
+		if (IsSavegameVersionBefore(SLV_85)) return LS_END - 2;
+		if (IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH)) return LS_END;
+		/* Read from the savegame how long the list is. */
+		return SlGetStructListLength(LS_END);
+	}
+
+	void Save(CompanyProperties *c) const override
+	{
+		SlSetStructListLength(LS_END);
+		for (int i = 0; i < LS_END; i++) {
+			SlObject(&c->livery[i], this->GetDescription());
+		}
+	}
+
+	void Load(CompanyProperties *c) const override
+	{
+		size_t num_liveries = this->GetNumLiveries();
 		bool update_in_use = IsSavegameVersionBefore(SLV_GROUP_LIVERIES);
 
-		for (int i = 0; i < num_liveries; i++) {
+		for (size_t i = 0; i < num_liveries; i++) {
 			SlObject(&c->livery[i], this->GetDescription());
 			if (update_in_use && i != LS_DEFAULT) {
 				if (c->livery[i].in_use == 0) {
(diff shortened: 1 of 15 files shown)


src/saveload/company_sl.cpp
```diff
@@ -395,12 +403,20 @@ class SlCompanyLiveries : public DefaultSaveLoadHandler<SlCompanyLiveries, Compa
...
-		int num_liveries = IsSavegameVersionBefore(SLV_63) ? LS_END - 4 : (IsSavegameVersionBefore(SLV_85) ? LS_END - 2: LS_END);
+		SlSetStructListLength(LS_END);
+		for (int i = 0; i < LS_END; i++) {
+			SlObject(&c->livery[i], this->GetDescription());
+		}
+	}
+
+	void Load(CompanyProperties *c) const override
+	{
+		size_t num_liveries = IsSavegameVersionBefore(SLV_63) ? LS_END - 4 : IsSavegameVersionBefore(SLV_85) ? LS_END - 2: IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH) ? (int)LS_END : SlGetStructListLength(LS_END);
```
rubidium42: Is the cast to int really necessary?

Maybe create a helper function that returns the liveries? That makes it quite a bit more legible overall.
```
static size_t GetNumberOfLiveriesInTheSavegame() {
    if (IsSavegameVersionBefore(SLV_63)) return LS_END - 4;
    if (IsSavegameVersionBefore(SLV_85)) return LS_END - 2;
    if (IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH)) return LS_END;
    return SlGetStructListLength(LS_END);
}
```

src/saveload/station_sl.cpp
```diff
@@ -369,7 +383,7 @@ class SlStationGoods : public DefaultSaveLoadHandler<SlStationGoods, BaseStation
 			memcpy(st->airport.psa->storage, _old_st_persistent_storage.storage, sizeof(_old_st_persistent_storage.storage));
 		}
 
-		uint num_cargo = IsSavegameVersionBefore(SLV_55) ? 12 : IsSavegameVersionBefore(SLV_EXTEND_CARGOTYPES) ? 32 : NUM_CARGO;
+		size_t num_cargo = IsSavegameVersionBefore(SLV_55) ? 12 : IsSavegameVersionBefore(SLV_EXTEND_CARGOTYPES) ? 32 : IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH) ? (size_t)NUM_CARGO : SlGetStructListLength(NUM_CARGO);
```
rubidium42: Maybe a helper function here as well to improve legibility?

src/saveload/town_sl.cpp
```diff
@@ -124,14 +124,15 @@ class SlTownSupplied : public DefaultSaveLoadHandler<SlTownSupplied, Town> {
...
+		SlSetStructListLength(NUM_CARGO);
 		for (CargoID i = 0; i < NUM_CARGO; i++) {
 			SlObject(&t->supplied[i], this->GetDescription());
 		}
 	}
 
 	void Load(Town *t) const override
 	{
-		uint num_cargo = IsSavegameVersionBefore(SLV_EXTEND_CARGOTYPES) ? 32 : NUM_CARGO;
+		size_t num_cargo = IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH) ? (IsSavegameVersionBefore(SLV_EXTEND_CARGOTYPES) ? 32 : NUM_CARGO) : SlGetStructListLength(NUM_CARGO);
```
rubidium42: The previously mentioned helper for stations can probably be reused here as well, right?

src/saveload/saveload.cpp
```diff
@@ -1043,45 +1043,28 @@ static void SlStdString(void *ptr, VarType conv)
 }
 
 /**
- * Return the size in bytes of a certain type of atomic array
- * @param length The length of the array counted in elements
- * @param conv VarType type of the variable that is used in calculating the size
+ * Internal function to save/Load a list of SL_VARs.
+ * SlCopy() and SlArray() are very similar, with the exception of the header.
```
rubidium42, line 1047: Just for confirmation:
SlCopy is "meant" for anything that has a dynamic size?
SlArray is "meant" for anything that has a fixed size?
Or am I missing something else here? I can't really figure out from the comments of either SlCopy or SlArray.
And I'm seeing SlCopies with fixed sizes, so I guess what I'm thinking is wrong. In any case, I guess it needs to be clearer whether when to use SlCopy or SlArray.

TrueBrain: Will update the comments, but basically:

`SlCopy` you can use from outside `saveload.c`, `SlArray` is only usable via an `SL_ARR`. The only reason for the difference is that one has a length-field and the other doesn't.

src/saveload/station_sl.cpp
```diff
@@ -344,15 +360,25 @@ class SlStationGoods : public DefaultSaveLoadHandler<SlStationGoods, BaseStation
 	}
 #endif
 
+	/**
+	 * Get the number of cargoes used by this savegame version.
+	 * @return The number of cargoes used by this savegame version.
+	 */
+	size_t GetNumCargo() const {
```
rubidium42: `{` on newline.

src/saveload/town_sl.cpp
```diff
@@ -122,16 +122,28 @@ class SlTownSupplied : public DefaultSaveLoadHandler<SlTownSupplied, Town> {
 		SLE_CONDVAR(TransportedCargoStat<uint32>, new_act, SLE_UINT32, SLV_165, SL_MAX_VERSION),
 	};
 
+	/**
+	 * Get the number of cargoes used by this savegame version.
+	 * @return The number of cargoes used by this savegame version.
+	 */
+	size_t GetNumCargo() const {
```
rubidium42: `{` on newline.

src/saveload/company_sl.cpp
```diff
@@ -395,12 +403,32 @@ class SlCompanyLiveries : public DefaultSaveLoadHandler<SlCompanyLiveries, Compa
 		SLE_CONDVAR(Livery, colour2, SLE_UINT8, SLV_34, SL_MAX_VERSION),
 	};
 
-	void GenericSaveLoad(CompanyProperties *c) const
+	/**
+	 * Get the number of liveries used by this savegame version.
+	 * @return The number of liveries used by this savegame version.
+	 */
+	size_t GetNumLiveries() const {
```
rubidium42: `{` on newline.

rubidium42 (approved):

PR #9375, opened 2021-06-14
Change: make as many savegame chunks a CH_ARRAY as possible

TrueBrain:
## Motivation / Problem

Weird goal, I know right? But hear me out.

#9322 sets out to make to make our savegame self-descriptive (see motivation for that there). But in order to do that, we need to make all chunks a table.

The first step in making anything a table, is to first make it an array. And so we come to the motivation of this PR: to make that possible.

## Description

Most `CH_RIFF`s in fact are just an array of length 1. So it might seem silly at first, but that is exactly what this PR does: make those chunks into an array of length 1.

The exception was the `GLOG` chunk. That required some more love and attention.

After this PR, only the map-chunks are `CH_RIFF`. Those could be made into `CH_ARRAY`, but that is not really useful, as they cannot become self-describing for now anyway.

Lastly, the chunk table tells how chunks are stored on disk. Chunks that are no longer stored were still marked by their old type. This was often confusing. So I sneaked that change in here, to mark those as `CH_READONLY`, making it more clear that it is not by accident.

## Checklist for review

### DIFF
--- a/src/saveload/animated_tile_sl.cpp
+++ b/src/saveload/animated_tile_sl.cpp
@@ -18,13 +18,17 @@
 
 extern std::vector<TileIndex> _animated_tiles;
 
+static const SaveLoad _animated_tile_desc[] = {
+	 SLEG_VECTOR(_animated_tiles, SLE_UINT32),
+};
+
 /**
  * Save the ANIT chunk.
  */
 static void Save_ANIT()
 {
-	SlSetLength(_animated_tiles.size() * sizeof(_animated_tiles.front()));
-	SlCopy(_animated_tiles.data(), _animated_tiles.size(), SLE_UINT32);
+	SlSetArrayIndex(0);
+	SlGlobList(_animated_tile_desc);
 }
 
 /**
@@ -45,10 +49,17 @@ static void Load_ANIT()
 		return;
 	}
 
-	uint count = (uint)SlGetFieldLength() / sizeof(_animated_tiles.front());
-	_animated_tiles.clear();
-	_animated_tiles.resize(_animated_tiles.size() + count);
-	SlCopy(_animated_tiles.data(), count, SLE_UINT32);
+	if (IsSavegameVersionBefore(SLV_RIFF_TO_ARRAY)) {
+		size_t count = SlGetFieldLength() / sizeof(_animated_tiles.front());
+		_animated_tiles.clear();
+		_animated_tiles.resize(_animated_tiles.size() + count);
+		SlCopy(_animated_tiles.data(), count, SLE_UINT32);
+		return;
+	}
+
+	if (SlIterateArray() == -1) return;
+	SlGlobList(_animated_tile_desc);
+	if (SlIterateArray() != -1) SlErrorCorrupt("Too many ANIT entries");
 }
 
 /**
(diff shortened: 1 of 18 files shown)


src/saveload/animated_tile_sl.cpp
```diff
@@ -18,13 +18,17 @@
 
 extern std::vector<TileIndex> _animated_tiles;
 
+static const SaveLoad _animated_tile_desc[] = {
```
rubidium42, line 21: In the commit message: practicle -> practical

src/saveload/animated_tile_sl.cpp
```diff
@@ -45,18 +49,25 @@ static void Load_ANIT()
 		return;
 	}
 
-	uint count = (uint)SlGetFieldLength() / sizeof(_animated_tiles.front());
-	_animated_tiles.clear();
-	_animated_tiles.resize(_animated_tiles.size() + count);
-	SlCopy(_animated_tiles.data(), count, SLE_UINT32);
+	if (IsSavegameVersionBefore(SLV_RIFF_TO_ARRAY)) {
+		uint count = (uint)SlGetFieldLength() / sizeof(_animated_tiles.front());
```
rubidium42: YIKYOMI, but count can be `size_t` as everything using it accepts `size_t`

src/saveload/saveload.cpp
```diff
@@ -1444,6 +1445,48 @@ static void SlDeque(void *deque, VarType conv)
...
+		default: NOT_REACHED();
+	}
+}
+
+/**
+ * Save/load a std::vector.
+ * @param vector The std::vector being manipulated
+ * @param conv VarType type of variable that is used for calculating the size
+ */
+static void SlVector(void *vector, VarType conv)
```
rubidium42, line 1482: I don't think it was meant that this got added in this commit.

TrueBrain: It was, as this is the first use. I can split it off in a new commit that only adds it, but doesn't use it. Amount of code was low enough that I personally didn't care :)

src/saveload/gamelog_sl.cpp
```diff
@@ -15,147 +15,317 @@
...
+		if (lc->ct != GLCT_GRFBUG) return;
+		SlObject(lc, this->GetDescription());
+	}
+
+	void Save(LoggedChange *lc) const override { this->GenericSaveLoad(lc); }
+	void Load(LoggedChange *lc) const override { this->GenericSaveLoad(lc); }
+	void LoadCheck(LoggedChange *lc) const override { this->GenericSaveLoad(lc); }
 };
 
-static_assert(lengthof(_glog_desc) == GLCT_END);
```
rubidium42, line 92: Would it make sense to add something similar to `SlGamelogAction::description` but with a `+ 1`? Alternatively you could say that if someone adds a game log action but do not check it survives a save-load, then the game log action is pointless and should not be added.

rubidium42 (approved):

src/saveload/gamelog_sl.cpp
TrueBrain: Exactly the last part. The assert was there as the length of the list was important. That is no longer the case. And especially as you can now view a savegame, I think we have to assume the developer validates it :D

PR #9376, opened 2021-06-16
Feature: Show the number of clients and companies in the online players window

telk5093:
## Motivation / Problem
Thanks to @TrueBrain, we have a new Online players window by 526635942451479bee66e9eb61c50f91ae48a7dd
It is nice, but hard to know how many clients and companies at once during playing a multiplayer game.
So I added an additional line to display the total number of clients and companies.

## Description
This feature adds the number of clients and companies at the bottom of the online players window like screenshot below:
image

## Limitations
 * There is a problem with displaying number of clients/companies in RTL languages, such as Arabic.
 I have no idea whether my code is wrong, or it will be fixed when Arabic and RTL language translators make a translation.
 image

 * I'm not good at C++ and OpenTTD's code so please let me know if any problems or suggestions. Thanks beforehand!

## Checklist for review


src/widgets/network_widget.h
```diff
@@ -106,6 +106,7 @@ enum ClientListWidgets {
 	WID_CL_MATRIX,                     ///< Company/client list.
 	WID_CL_SCROLLBAR,                  ///< Scrollbar for company/client list.
 	WID_CL_COMPANY_JOIN,               ///< Used for QueryWindow when a company has a password.
+	WID_CL_CLIENT_NUMBER,              ///< Client list.
```
rubidium42: Maybe WID_CL_CLIENT_COMPANY_COUNT would be a better name, as it doesn't contain the number of the client. It's also not the client list. Please also amend the name of the string in english.txt.

src/network/network_gui.cpp
```diff
@@ -2371,6 +2372,16 @@ struct NetworkClientListWindow : Window {
 
 				break;
 			}
+
+			case WID_CL_CLIENT_NUMBER: {
+				/* Number of clients */
+				int text_y_offset = std::max(0, ((int)this->line_height - (int)FONT_HEIGHT_NORMAL) / 2);
```
rubidium42: `this->line_height` is set to the maximum of some height and `FONT_HEIGHT_NORMAL`, so that value will always be a positive number. So no need for `int` or `std::max` here.
Having said that, take a look at `SetDataTip` and how it's used with `NWidget(WWT_TEXTBTN, ...`. I think by adding a `SetDataTip(STR_NETWORK_CLIENT_LIST_CLIENT_NUMBER, <some to be defined string for the tooltip>)` to line 1634, and putting the `SetDParam` lines below in case of `SetStringParameters` around line 2040, you do not need all this custom calculation and drawing.

src/network/network_gui.cpp
```diff
@@ -1628,10 +1628,11 @@ static const NWidgetPart _nested_client_list_widgets[] = {
...
 		NWidget(NWID_HORIZONTAL),
 			NWidget(WWT_MATRIX, COLOUR_GREY, WID_CL_MATRIX), SetMinimalSize(180, 0), SetResize(1, 1), SetFill(1, 1), SetMatrixDataTip(1, 0, STR_NULL), SetScrollbar(WID_CL_SCROLLBAR),
-			NWidget(NWID_VERTICAL),
-				NWidget(NWID_VSCROLLBAR, COLOUR_GREY, WID_CL_SCROLLBAR),
-				NWidget(WWT_RESIZEBOX, COLOUR_GREY),
-			EndContainer(),
+			NWidget(NWID_VSCROLLBAR, COLOUR_GREY, WID_CL_SCROLLBAR),
+		EndContainer(),
+		NWidget(NWID_HORIZONTAL),
+			NWidget(WWT_TEXTBTN, COLOUR_GREY, WID_CL_CLIENT_NUMBER), SetFill(1, 0), SetMinimalTextLines(1, 0), SetResize(1, 0), SetAlignment(SA_RIGHT),
```
rubidium42: Shouldn't this be a `WWT_LABEL` or maybe `WWT_TEXT`? It's not a button you are supposed to click on, right?

rubidium42 (commented):
> * There is a problem with displaying number of clients/companies in RTL languages, such as Arabic.
>   I have no idea whether my code is wrong, or it will be fixed when Arabic and RTL language translators make a translation.

It's expected that it looks weird due to all kinds of different rules for ordering different things. There's nothing you can do about it, however the translators can inject special characters to ensure the string is shown correctly when translated.

It's a nice first attempt, though there are some things that could be improved to make it clearer and simpler overall.

src/network/network_gui.cpp
```diff
@@ -1628,10 +1628,11 @@ static const NWidgetPart _nested_client_list_widgets[] = {
 		EndContainer(),
 		NWidget(NWID_HORIZONTAL),
 			NWidget(WWT_MATRIX, COLOUR_GREY, WID_CL_MATRIX), SetMinimalSize(180, 0), SetResize(1, 1), SetFill(1, 1), SetMatrixDataTip(1, 0, STR_NULL), SetScrollbar(WID_CL_SCROLLBAR),
-			NWidget(NWID_VERTICAL),
-				NWidget(NWID_VSCROLLBAR, COLOUR_GREY, WID_CL_SCROLLBAR),
-				NWidget(WWT_RESIZEBOX, COLOUR_GREY),
```
glx22, line 1633: I think it's probably better to keep scroll bar and resize box in the same container. And group the matrix and your new widget, it should help to align everything.

telk5093: Oh my, why I haven't thought like that! Thanks and I'll apply.

src/network/network_gui.cpp
telk5093: It was my intend to draw the container of the count of clients/companies as a box shape.
Anyway I changed ``WWT_TEXTBTN`` into ``WWT_TEXT`` now, though it looks weird for me.
Please let me know if there is a better idea.

src/network/network_gui.cpp
telk5093: Thanks for details, it was a beating my head against a stone wall 😛

telk5093:
Now it looks like this:
image

src/network/network_gui.cpp
rubidium42: It definitely looks different, though I think it looking different is not a bad thing as it makes it more clear it is not one of the lines in the table, but rather a footer. With the text button it looked like it was part of the table.
Having said that, I think it looks better when the string is aligned in the center (`SA_CENTER`), as the right alignment looks weird to me. Though, that can also be my bad taste. Maybe some others have an opinion they want to share about the looks of the button and alignment of the text.

embeddedt: I agree that right-alignment looks strange here. It should either be centered or left-aligned, IMO. Left alignment probably makes more sense, since this is not a button.

telk5093: Okay, let's do them.

**Left:**
image

**Centre:**
image

I like whatever and so I can't decide it.

glx22: Center is not bad, left is OK too, both are better than right. If I have to decide I'd say center.
BTW it seems the matrix is longer than the scroll bar.

TrueBrain: Center looks really nice in my book :D

telk5093: OK, now make it centered.

TrueBrain (approved):

PR #9377, opened 2021-06-17
Cleanup: nullptr and boolean literals instead of NULL, 0 or 1

rubidium42:
## Motivation / Problem

Inconsistent `nullptr` usage, or lack thereof in some places where it still used `NULL` or `0`.
Inconsistent boolean usage, where it would use `0` or `1` to denote `false` or `true`.

## Description

Mostly the result of some experimentation with clang-tidy.
Next to `nullptr` and boolean literals, also some simplifications of boolean logic, e.g. `e?true:false` -> `e != 0`.
Also a duplicate set of lines between both `base_station_base.h` and `station_base.h`  was removed from the latter, as the latter already includes the former.

## Limitations

Have checked only code that was compiled by clang on my computer, so there might be other places.

## Checklist for review


LordAro:
I feel like squirrel was left alone on purpose during the 'main' NULL -> nullptr conversion. Not that it particularly matters, it's so custom at this point that any upstream changes are infeasible anyway...

TrueBrain (approved):

PR #9378, opened 2021-06-17
Codechange: convert news C-strings to std::string

rubidium42:
## Motivation / Problem

Continue the C-string to std::string conversion.

## Description

Change `free_data` from something that is `malloc`ed/`free`d to something that is allocated with `new` and `delete`. Type this so it becomes easier to spot all types of information that get passed into there by finding the appropriate sub classes. In doing so replace `strdup`ed strings with strings in an allocated sub class.

## Limitations

Practically all this information could be put in sub classes of NewsItem, but that is something that should be done at a later point.

## Checklist for review

### DIFF
--- a/src/news_type.h
+++ b/src/news_type.h
@@ -135,14 +135,11 @@ struct NewsItem {
 	uint32 ref1;                 ///< Reference 1 to some object: Used for a possible viewport, scrolling after clicking on the news, and for deleting the news when the object is deleted.
 	uint32 ref2;                 ///< Reference 2 to some object: Used for scrolling after clicking on the news, and for deleting the news when the object is deleted.
 
-	const NewsAllocatedData *data; ///< Custom data for the news item that have to be deallocated (deleted) when the news item has reached its end.
-
-	~NewsItem()
-	{
-		delete this->data;
-	}
+	std::unique_ptr<const NewsAllocatedData> data; ///< Custom data for the news item that will be deallocated (deleted) when the news item has reached its end.
 
 	uint64 params[10]; ///< Parameters for string resolving.
+
+	NewsItem(StringID string_id, NewsType type, NewsFlag flags, NewsReferenceType reftype1, uint32 ref1, NewsReferenceType reftype2, uint32 ref2, const NewsAllocatedData *data);
 };
 
 /** Container for a single string to be passed as NewsAllocatedData. */
--- a/src/news_type.h
+++ b/src/news_type.h
@@ -115,6 +115,12 @@ struct NewsTypeData {
 	NewsDisplay GetDisplay() const;
 };
 
+/** Container for any custom data that must be deleted after the news item has reached end-of-life. */
+struct NewsAllocatedData {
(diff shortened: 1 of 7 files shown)


src/news_type.h
```diff
@@ -115,6 +115,10 @@ struct NewsTypeData {
 	NewsDisplay GetDisplay() const;
 };
 
+/** Container for any custom data that must be deleted after the news item has reached end-of-life. */
+struct NewsAllocatedData {};
```
LordAro: Is there a way we can force this type to never be used directly? (Which is presumably undesirable?)

src/news_type.h
```diff
@@ -129,23 +133,29 @@ struct NewsItem {
 	uint32 ref1;                 ///< Reference 1 to some object: Used for a possible viewport, scrolling after clicking on the news, and for deleting the news when the object is deleted.
 	uint32 ref2;                 ///< Reference 2 to some object: Used for scrolling after clicking on the news, and for deleting the news when the object is deleted.
 
-	void *free_data;             ///< Data to be freed when the news item has reached its end.
+	const NewsAllocatedData *data; ///< Custom data for the news item that have to be deallocated (deleted) when the news item has reached its end.
```
TrueBrain: Didn't you just basically make a `std::unique_ptr` here? From a quick (but not detailed :P) look, that might help you out too, and avoids the "weird" dtor :)

src/news_type.h
rubidium42: We could define all "default" constructors as protected and then you must make a sub class of it. On the other hand, why make a new instance of this when you can pass nullptr as easily?
So I'm not quite sure whether all the guarding against such a thing is worth it. I just added it for a tiny bit more type safety.

src/news_type.h
```diff
@@ -115,6 +115,10 @@ struct NewsTypeData {
 	NewsDisplay GetDisplay() const;
 };
 
+/** Container for any custom data that must be deleted after the news item has reached end-of-life. */
+struct NewsAllocatedData {};
+
```
TrueBrain, line 122: I think you have to make the dtor virtual now, as this is most likely UB :)

https://godbolt.org/z/axTW5sMEx

as example. To quote https://en.cppreference.com/w/cpp/memory/unique_ptr:

`If T is a derived class of some base B, then std::unique_ptr<T> is implicitly convertible to std::unique_ptr<B>. The default deleter of the resulting std::unique_ptr<B> will use operator delete for B, leading to undefined behavior unless the destructor of B is virtual.`

I can be completely wrong :) Running valgrind with your PR will give you the true answer here I am sure :P

Edit: from what I can tell, with the current way of doing stuff (read: using `std::string`), it works. Not 100% sure why, but that might be the UB part, I guess :P What doesn't work if you add something like another `std::unique_ptr` to one of the data classes. Example: https://godbolt.org/z/17f5v8a1n

Sorry, I was having some fun with this, trying to learn and understand what C++ is actually doing. Can't say I got really far, but .. yeah .. :P

TrueBrain (approved):

PR #9379, opened 2021-06-17
Change: prevent palette updates during copying to the video driver 

TrueBrain:
## Motivation / Problem

CLang's ThreadSanitizer complained a lot about our way of doing palettes. Turns out, depending on the video driver, it is also a bit of a mess how it is used.

## Description

This PR does two things:

1) first, it fixes the discrepancy between the video-drivers. Although not strictly needed to use `_local_palette` for the ones that weren't using it, as the `_cur_palette` is only used in that function, having everything the same does make the next commit a lot easier.

2) set a mutex-guard on `_cur_palette` access, so only one thread is accessing it at any given time.

This is a bit of an extreme solution in the sense that it possibly could be implemented in a more lightweight variant. `count_dirty` is used as event to notify the draw-thread a new palette is available. This means that guarding this would be sufficient, but it is also less clear that this is the case. So I didn't go for this solution, but it is something we can talk about.

## Limitations

- There are other places accessing `_cur_palette`, but they seem to all be done from the game-thread, which is all perfectly fine ofc.

## Checklist for review

### DIFF
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -51,6 +51,8 @@ static byte _stringwidth_table[FS_END][224]; ///< Cache containing width of ofte
 DrawPixelInfo *_cur_dpi;
 byte _colour_gradient[COLOUR_END][8];
 
+static std::recursive_mutex _palette_mutex; ///< To coordinate access to _cur_palette.
+
 static void GfxMainBlitterViewport(const Sprite *sprite, int x, int y, BlitterMode mode, const SubSprite *sub = nullptr, SpriteID sprite_id = SPR_CURSOR_MOUSE);
 static void GfxMainBlitter(const Sprite *sprite, int x, int y, BlitterMode mode, const SubSprite *sub = nullptr, SpriteID sprite_id = SPR_CURSOR_MOUSE, ZoomLevel zoom = ZOOM_LVL_NORMAL);
 
@@ -1197,6 +1199,7 @@ void DoPaletteAnimations();
 
 void GfxInitPalettes()
 {
+	std::lock_guard<std::recursive_mutex> lock(_palette_mutex);
 	memcpy(&_cur_palette, &_palette, sizeof(_cur_palette));
 	DoPaletteAnimations();
 }
@@ -1212,6 +1215,8 @@ void GfxInitPalettes()
  */
 bool CopyPalette(Palette &local_palette, bool force_copy)
 {
+	std::lock_guard<std::recursive_mutex> lock(_palette_mutex);
+
 	if (!force_copy && _cur_palette.count_dirty == 0) return false;
 
 	local_palette = _cur_palette;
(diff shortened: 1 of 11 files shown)


src/gfx.cpp
```diff
@@ -1201,6 +1201,30 @@ void GfxInitPalettes()
...
+bool CopyPalette(Palette &local_palette, bool force_copy)
+{
+	if (!force_copy && _cur_palette.count_dirty == 0) return false;
+
+	local_palette = _cur_palette;
+	_cur_palette.count_dirty = 0;
+
+	if (force_copy) {
+		local_palette.first_dirty = 0;
+		local_palette.count_dirty = 254;
```
rubidium42: Minor nitpick in case you're going to rebase this: seems you put the fix to 256 in the wrong commit.

rubidium42 (commented):
Though functionally the behaviour is the same, you add a `static Palette _local_palette` to some drivers and some drivers have a class variable for it. Might it not make sense to make `Palette local_palette` a variable of the video driver class, making everything even more consistent? That way any new driver will have the local palette making it more likely that someone doesn't access _cur_palette.
You could consider making the `CopyPalette` a function of the video driver, so it becomes the "problem" of the (generic) video driver to fill the local palette in a safe manner.
Although, alternatively screenshot could just copy the palette as well, and just remove `_cur_palette` from `gfx_func.h` all together. That would make it harder for new video drivers to access it; they would simply fail to compile, instead of function but be thread unsafe.

TrueBrain:
> You could consider making the `CopyPalette` a function of the video driver, so it becomes the "problem" of the (generic) video driver to fill the local palette in a safe manner.

I rather have the functions locally next to each other, so it is a lot easier to spot what functions are cooperating in the lock. Moving `CoypPalette` would be a lot of distance between those 2 functions, meaning another global mutex, etc etc. The code doesn't really get better because of it ;)
The problem in reality is more that `Palette` should be a class with a `Copy` function, etc etc, but that is a rather large code-refactor .. not for now :D

> Although, alternatively screenshot could just copy the palette as well, and just remove `_cur_palette` from `gfx_func.h` all together. That would make it harder for new video drivers to access it; they would simply fail to compile, instead of function but be thread unsafe.

That would be future work, and has to be considered carefully. There is no real reason to do this, other than to avoid someone accessing `_cur_palette` from the video driver. But the biggest problem here is that 1 blitter also accesses the `_cur_palette` which would also need to be addressed in that case. So this would involve a bit more work :D

TrueBrain:
> 
> Though functionally the behaviour is the same, you add a `static Palette _local_palette` to some drivers and some drivers have a class variable for it. Might it not make sense to make `Palette local_palette` a variable of the video driver class, making everything even more consistent? That way any new driver will have the local palette making it more likely that someone doesn't access _cur_palette.

After a bit of fiddling with this: this is not as simple as it looks on the surface. Although michi and I did clean up the win32, cocoa and sdl2 driver, we didn't do allegro and sdl1 (yet?). And guess what? THEY ARE UGLY :P For both I have to refactor the video driver quiet a bit. It starts with 1 function that is `static` while accessing the palette .. and that turns into changing a lot of other functions to `static`. But that also means some of the other things done in the other videodrivers should be done, otherwise future-us will hurt even more, and ... down the rabbit hole we go. So I pressed the panic button and resurfaced :P

So no, I am not going to do this for this PR if you don't mind, and leave it as it is. It is still a good idea, but requires more work than fits in this PR scope :D

michicc (approved):

PR #9381, opened 2021-06-17
Fix: don't propagate shift/ctrl state till next game-tick

TrueBrain:
## Motivation / Problem

Another bug found with ThreadSanitizer. It is a bit unlikely someone really runs into this, but what we do is also not correct. So I fixed it anyway.

Heavily inspired by https://github.com/JGRennison/OpenTTD-patches/commit/59daa573b5dc335e9adcdc365bd65fabcb8f145f

## Description

```
When the game-loop is very slow, it was easily possible to start
the loop with _shift_pressed being false, but end with
_shift_pressed being true. This doesn't hurt the game as such,
but for the user this can be very weird: I pressed "Buy Vehicle",
pressed shift a bit later, and I still get a cost indication.
```

## Checklist for review

### DIFF
--- a/src/video/video_driver.cpp
+++ b/src/video/video_driver.cpp
@@ -116,21 +116,6 @@ void VideoDriver::Tick()
 		/* Avoid next_draw_tick getting behind more and more if it cannot keep up. */
 		if (this->next_draw_tick < now - ALLOWED_DRIFT * this->GetDrawInterval()) this->next_draw_tick = now;
 
-		/* Keep the interactive randomizer a bit more random by requesting
-		 * new values when-ever we can. */
-		InteractiveRandom();
-
-		this->InputLoop();
-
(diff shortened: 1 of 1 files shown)


michicc:
There isn't really much threading left with this, is it?

TrueBrain:
> 
> 
> There isn't really much threading left with this, is it?

Well, yes and no :) Those 2 functions are basically a no-op .. just some shuffling around of a few variables. So nothing really changes with this PR on that balance.
But in general, there is a way too much inside the game-state-lock in my opinion. But .. that needs more work and attention to detail to fix.

The most important things for now is that the `Paint` and taking the video-lock are outside the game-state-lock. The first is rather expensive on non-OpenGL .. the last is expensive with vsync on. So there is still some benefit, luckily enough :D

michicc (approved):

PR #9382, opened 2021-06-17
Fix: on startup, NewGRF scan could case race-condition

TrueBrain:
## Motivation / Problem

NewGRF scan could trigger ThreadSanitizer (rightfully), because it tried to access a variable that wasn't fully prepared to be accessed yet.

## Description

```
Creating a thread was not thread-safe. The irony.

The video-driver has a function GameLoopPause() which first checks
if the thread is the game-thread or not. For this it needs access
to this->game_thread. This variable is set in StartNewThread().

However, due to timing, it is well possible GameLoopPause() is
called from the thread well before this->game_thread is assigned.

And so we have a race-condition!

Simply solve this by preventing a thread to start till we are
done with our bookkeeping.
```

## Checklist for review

### DIFF
--- a/src/thread.h
+++ b/src/thread.h
@@ -14,6 +14,7 @@
 #include "crashlog.h"
 #include <system_error>
 #include <thread>
+#include <mutex>
 
 /**
  * Sleep on the current thread for a defined time.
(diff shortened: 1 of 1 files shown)


src/thread.h
```diff
@@ -14,6 +14,8 @@
 #include "crashlog.h"
 #include <system_error>
 #include <thread>
+#include <mutex>
+#include <condition_variable>
```
rubidium42: `std::lock_guard` is in `<mutex>` and so is `std::mutex`. Where do we need `<condition_variable>`?

src/thread.h
```diff
@@ -46,7 +48,17 @@ inline bool StartNewThread(std::thread *thr, const char *name, TFn&& _Fx, TArgs&
 {
 #ifndef NO_THREADS
 	try {
+		static std::mutex cv_m;
```
rubidium42: It's a bit weird to suggest the variable has something to do with conditional variables, when it does not. Maybe just call it `mutex` or `thread_start_mutex`?

TrueBrain: I think both lines of code show what I initially was building :P But I aint doing that till C++20! Fuck that shit ;)

rubidium42 (approved):

PR #9389, opened 2021-06-20
Fix bf500c3: reduce the amount of vars to track by using std::string_view

### LINKED ISSUE #9386: Building on MacOS fails with "bus error: 10"
## Version of OpenTTD
Every commit after bf500c39c9b6c7359204c5454aefa0dc7f51dae5

## Expected result
Being able to compile on MacOS with `RelWithDebInfo`.

## Actual result
Compiling on MacOS fails with "Bus error: 10" when using `RelWithDebInfo`.

## Steps to reproduce
Build on MacOS. Or see our [nightly](https://github.com/OpenTTD/OpenTTD/actions/runs/950837085) .. it fails every night.

TrueBrain:
Fixes #9386.
Alternative to #9390. Closes #9390.

## Motivation / Problem

MacOS said: BOOOO
GCC said: fuck this shit

See #9386 for details and triage. But basically, since we changed a `const char *` into `std::string`, more than one compiler went apeshit on us. I present a (possibly temporary) solution: replace it with `const std::string_view`.

## Description

I do not really know why compilers start to track the `std::string` like this, or let alone why it blows up optimizations (GCC, from what I understand, tracks 50M vars before giving up .. that sounds like an insane alot for the few settings we have). Either way, on a hunch I replaced `std::string` with `const std::string` .. no change. So what about `const std::string_view`? Fixes the issue. Riggghhtttttt....

So to be clear, I have no clue why. I do know it is (most likely) a save operation, as the string itself will be stored in the data-segment, and the `string_view` will be a thin wrapper around that.

Sadly, this means in a few places where we were replacing `const char *` with `std::string`, I revert that situation a bit, as I cheeky use `.data()` to get this compiling again. By default a `std::string_view` cannot be casted to a `std::string`. But I mostly think this indicates places where we should change the `std::string` to `std::string_view`  instead of the other way around. Either way, I pushed this outside the scope of this PR, to keep this "fix" to its core element.

The commit message:
```
GCC complaints with:
  note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without

clang on MacOS just crashes (with bus error 10, read: out of memory).

This started to happen after introducing std::string to each setting.
As these are static strings that never change, there is no need to
use std::string. Instead, we can use std::string_view for this. This
seems to make both compilers really happy (as they basically still
only store a "const char *" as it was before this commit).
```

## Limitations

- I have 0 clues why this fixes the issue, or why the original code was an issue in the first place. This by no means tells us the root-problem. It just seems to fix the issue so everything is happy again. Any insights to what is going on here exactly would be greatly appreciated.

## Checklist for review

### DIFF
--- a/src/script/api/script_gamesettings.cpp
+++ b/src/script/api/script_gamesettings.cpp
@@ -37,7 +37,7 @@
 
 	if ((sd->flags & SF_NO_NETWORK_SYNC) != 0) return false;
 
-	return ScriptObject::DoCommand(0, 0, value, CMD_CHANGE_SETTING, sd->name.c_str());
+	return ScriptObject::DoCommand(0, 0, value, CMD_CHANGE_SETTING, sd->name.data());
 }
 
 /* static */ bool ScriptGameSettings::IsDisabledVehicleType(ScriptVehicle::VehicleType vehicle_type)
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -318,13 +318,13 @@ size_t IntSettingDesc::ParseValue(const char *str) const
 	if (end == str) {
 		ErrorMessageData msg(STR_CONFIG_ERROR, STR_CONFIG_ERROR_INVALID_VALUE);
 		msg.SetDParamStr(0, str);
-		msg.SetDParamStr(1, this->name);
+		msg.SetDParamStr(1, this->name.data());
 		_settings_error_list.push_back(msg);
 		return this->def;
 	}
 	if (*end != '\0') {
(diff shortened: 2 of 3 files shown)


TrueBrain:
On request tested to keep `name` a `std::string` but change the parameter type to `std::string_view`. That doesn't do enough to resolve the issue. The `name` has to become `std::string_view` before the problem is resolved.

TrueBrain:
This is not the core issue, just a way to make it hide again. Let's not go for this solution, but try to find a proper solution.

PR #9395, opened 2021-06-23
Change #9188: Netsave now behaves like autosave

Taschi120:

PR #9397, opened 2021-06-24
Feature: Persistant rotation of numbered autosave after restart

### LINKED ISSUE #9396: Rotation of numbered autosaves is confusing
## Version of OpenTTD
master, revision 37116a7, any OS

## Expected result
When the game makes a new autosave, it should
a) check if there is a free "autosave slot" and use it
b) if no autosave slots are free, overwrite the oldest autosave

## Actual result
The first autosave after starting OpenTTD will always be called "autosave0", followed by "autosave1" and so on. When exiting and restarting OpenTTD, the next autosave will again be called "autosave0" because the counter resets.

In some cases, this might end up actually overwriting the *newest* autosave first, which is highly unintuitive and unexpected.

## Steps to reproduce
Start the executable, start a new game, fast-forward until it has created a couple of autosaves. Exit & restart the executable, start a new game, fast-forward until a new autosave is created. Check file creation dates in the autosave folder.

Note that the "keep_all_autosave" settings has to be "false", which it is by default.

glx22:
Fixes #9396

## Motivation / Problem
Autosave numbering always starts from 0 when OpenTTD is started.

## Description
Retrieve the number of the most recent autosave and use it as a base for the next one.

## Limitations
If more than one OpenTTD instance is running, and if they share the same `autosave` directory, they will still overwrite each other savegames.

## Checklist for review

### DIFF
--- a/src/fios.cpp
+++ b/src/fios.cpp
@@ -22,6 +22,7 @@
 #include <sys/stat.h>
 #include <functional>
 #include <optional>
+#include <charconv>
 
 #ifndef _WIN32
 # include <unistd.h>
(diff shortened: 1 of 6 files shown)


Taschi120:
Note that this PR and PR #9395 have conflicts.

src/fios.cpp
```diff
@@ -745,3 +746,49 @@ void ScanScenarios()
...
+	/* Find the number for the most recent save, if any. */
+	if (list.begin() != list.end()) {
+		SortingBits order = _savegame_sort_order;
+		_savegame_sort_order = SORT_BY_DATE | SORT_DESCENDING;
+		std::sort(list.begin(), list.end());
+		_savegame_sort_order = order;
+
+		std::string_view name = list.begin()->title;
+		std::from_chars(name.data() + this->prefix.size(), name.data() + name.size(), this->number);
+	}
```
TrueBrain, line 784: This is a lot of code for in a ctor. Somehow that triggers me. But I do not have any argument for it :P So it might be bullocks. Opinions? :D

Taschi120: Well, the thing has to be initialized *somewhere* and where else should it be, if not in the constructor? And the code seems fairly specific to this one particular problem, so it's not like refactoring anything into separate functions would have much of an advantage.

src/fios.cpp
```diff
@@ -745,3 +746,49 @@ void ScanScenarios()
...
+{
+	static std::string _prefix;
+	static std::optional<std::string> _autosave_path;
+	if (!_autosave_path) _autosave_path = FioFindDirectory(AUTOSAVE_DIR);
+
+	/* Callback for FiosFileScanner. */
+	static fios_getlist_callback_proc *proc = [](SaveLoadOperation fop, const std::string &file, const char *ext, char *title, const char *last) {
+		if (strcasecmp(ext, ".sav") == 0 && StrStartsWith(file, _prefix)) return FIOS_TYPE_FILE;
+		return FIOS_TYPE_INVALID;
+	};
```
TrueBrain, line 765: Suggestions to make it a bit more clear what `_prefix` is doing and that it is static:

```
	static std::optional<std::string> _autosave_path;
	if (!_autosave_path) _autosave_path = FioFindDirectory(AUTOSAVE_DIR);

	static std::string _prefix; ///< Static as the lambda needs access to it.

	/* Callback for FiosFileScanner. */
	static fios_getlist_callback_proc *proc = [](SaveLoadOperation fop, const std::string &file, const char *ext, char *title, const char *last) {
		if (strcasecmp(ext, ".sav") == 0 && StrStartsWith(file, _prefix)) return FIOS_TYPE_FILE;
		return FIOS_TYPE_INVALID;
	};

	/* Prefix to check in the callback. */
	_prefix = *_autosave_path + this->prefix;
```

? Dunno :)

glx22:
Function names are not the best.

TrueBrain (approved):

PR #9398, opened 2021-06-24
Fix: [Squirrel] Enforce comma usage in function calls

glx22:
## Motivation / Problem
Squirrel errors for `call(a, b, c,)` with `expression expected, found ')'`, but accepts `call(a b c)`.
I think it's inconsistent, and it should enforce `,` usage.

## Description
Added a check to enforce `,` in function calls. And found 2 missing ones in regression test :)

I tried to respect Squirrel coding style.

## Limitations
Existing AIs may fail to compile if they miss some `,` in function calls.

## Checklist for review

### DIFF
--- a/regression/regression/main.nut
+++ b/regression/regression/main.nut
@@ -1702,9 +1702,9 @@ function Regression::Vehicle()
 	print("  GetWagonAge():        " + AIVehicle.GetWagonAge(17, 0));
 	print("  GetWagonEngineType(): " + AIVehicle.GetWagonEngineType(17, 1));
 	print("  GetWagonAge():        " + AIVehicle.GetWagonAge(17, 1));
-	print("  GetWagonEngineType(): " + AIVehicle.GetWagonEngineType(17 2));
+	print("  GetWagonEngineType(): " + AIVehicle.GetWagonEngineType(17, 2));
 	print("  GetWagonAge():        " + AIVehicle.GetWagonAge(17, 2));
-	print("  GetWagonEngineType(): " + AIVehicle.GetWagonEngineType(17 3));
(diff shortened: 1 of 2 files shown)


TrueBrain:
For all the logical reasons this sounds like an extremely bad idea without implementing some versioning / compatibility system :p

LordAro:
Agreed with TB - breaking existing AIs isn't worth it. They're broken enough as it is ;)

Might be worth documenting somewhere though...

TrueBrain:
Or make it a developer warning in case it is by accident :)

glx22:
Indeed compatibility is the biggest issue. And limiting the extra check to >1.12 is quite impossible.

TrueBrain:
Each AI tells his compatibility version, so possible it is for sure. The question is, is it worth all that code to basically just scratch an itch you are having ;)

Personally, but this is just my opinion, I agree with LordAro: we should just document this behaviour and leave it for what it is. And documenting it on the wiki alone would be sufficient for me.

glx22:
I know API version is in `ScriptInstance`, but it's not accessible from `SQCompiler` (without changing stuff only for that), then comparing versions needs extra work because it's a string, not a number.

So I agree, it's probably better to discard this PR.

rubidium42:
Would it be feasible to make it a warning? I'm not sure whether there's infrastructure for that in the code, but it would at least nudge script developers to not add new cases.

TrueBrain:
Friendly poke, is a warning possible? Or should we close this PR?

glx22:
There's no warning support in `SQCompiler` and errors are handed via exceptions.
So adding warning support just for one use is probably too much work.
Let's just close the PR.

PR #7922, opened 2020-01-09
Improved logic of sharing industry production between stations

ldpl:
Since the dawn of time Industries in TTD only share cargo between two stations max. This behavior doesn't seem to be based on any logic and leads to some issues in mp as it can be easily abused. 

So this patch:
1. Removes the limit of 2 stations.
2. Shares cargo between companies first and then between stations of each company separately to prevent exploiting it by spaming station.

Shares are still based on station rating like it was before and, in fact, this patch doesn't change behaviour at all if there are less than 3 stations around (except for negligible rounding losses).

I didn't add a new setting for this change as IMO it only fixes exploits/limitations and doesn't change the gameplay logic. So I can't imagine why would anyone ever want to opt-out of it.

PR #9405, opened 2021-06-29
Reduce allocations in string handling

fsimonis:
## Motivation / Problem

I often use OpenTTD to get accustomed with new tools. This time I used heaptrack to analyse and locate allocations in the code.

There are some places in string handling that reallocate temporary char buffers per call. As OpenTTD has to format quite a few strings per frame, this accumulates quickly.

## Description

_edit: skip this part_

> Proxy functions often follow this pattern:
> 1. Allocate char buffer
> 2. Populate char buffer with a string
> 3. Process the string
> I marked these buffers as static to prevent unnecessary allocations. The content will simply be overwritten by the population function.

Next is the `FormatString` function. Currently, this function uses a stack to keep track of the remaining string.
This is not necessary as this stack can never reach 2 values. (Correct me if I am wrong.) Hence using a pointer suffices and prevents allocations altogether.

I used the first 30s of [this savegame](https://www.tt-forums.net/viewtopic.php?f=60&t=49307) to analyse the differences. I am using the sdl(_opengl) drivers.
The overall temporary allocations are down by ~90% and FormatString now doesn't allocate at all.
`ViewportDrawStrings` + Callers is now responsible for 26% of total temporary allocation (48k), which was 67% (413k).

Temp allocation hotspots after this PR are:
* `Layouter::GetCachedParagraphLayout` ~ 25%
* `ViewportSortParentSpritesSSE41`~ 34%
Main allocation hotspot is `ViewportSortParentSpritesSSE41` with ~85% of all allocations.

I am not sure how much this translates to FPS improvement. The 512 samples are not capable of smoothening out the frame time jitter on my dev system. Maybe a reviewer can give this a try.

## Limitations

_none_

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -800,15 +800,14 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 	WChar b = '\0';
 	uint next_substr_case_index = 0;
 	char *buf_start = buff;
-	std::stack<const char *, std::vector<const char *>> str_stack;
-	str_stack.push(str_arg);
+	const char * next_str_arg = str_arg;
 
 	for (;;) {
-		while (!str_stack.empty() && (b = Utf8Consume(&str_stack.top())) == '\0') {
-			str_stack.pop();
+		if (next_str_arg != nullptr && (b = Utf8Consume(next_str_arg)) == '\0') {
+			next_str_arg = nullptr;
 		}
-		if (str_stack.empty()) break;
-		const char *&str = str_stack.top();
+		if (next_str_arg == nullptr) break;
+		const char *str = next_str_arg;
 
 		if (SCC_NEWGRF_FIRST <= b && b <= SCC_NEWGRF_LAST) {
 			/* We need to pass some stuff as it might be modified; oh boy. */
@@ -915,13 +914,13 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 
 			case SCC_NEWGRF_STRINL: {
 				StringID substr = Utf8Consume(&str);
-				str_stack.push(GetStringPtr(substr));
+				next_str_arg = GetStringPtr(substr);
 				break;
 			}
 
 			case SCC_NEWGRF_PRINT_WORD_STRING_ID: {
 				StringID substr = args->GetInt32(SCC_NEWGRF_PRINT_WORD_STRING_ID);
-				str_stack.push(GetStringPtr(substr));
+				next_str_arg = GetStringPtr(substr);
 				case_index = next_substr_case_index;
 				next_substr_case_index = 0;
 				break;


LordAro:
Intriguing! My immediate thought - will making those string buffers `static` break thread safety?

TrueBrain:
Hihi, using OpenTTD as a benchmark application .. things I never expected people to do :D Nice!

Question for you: I do not fully understand why that tool considers a stack variable "an allocation"; as that shouldn't be an allocation at all. At least, from what I see in the assembly code, these `buffer`s don't end up on the heap, but stay on the stack. So the stack is just increased with, in this case, 2048, and the `buffer` sits there on the stack. That is not an "allocation" in the classical sense, as it doesn't happen on the heap.

So I wonder if that tool sees something I am not seeing (very much possible :D), or if some compilers with some optimization settings move the `buffer` to the heap or something? Do you have more detail about what is going on exactly here? I am really curious if it is really on the heap what is going on, as that is not something I would expect to happen here :D

Any insights would be appreciated :)

PS: btw, not saying it is such a great idea to have such large variables on the stack, but that is not the subject of this PR :D

fsimonis:
> Intriguing! My immediate thought - will making those string buffers static break thread safety?

Are there multiple drawing threads? If that is the case, then yes.

> Question for you: I do not fully understand why that tool considers a stack variable "an allocation"; as that shouldn't be an allocation at all. At least, from what I see in the assembly code, these buffers don't end up on the heap, but stay on the stack. So the stack is just increased with, in this case, 2048, and the buffer sits there on the stack. That is not an "allocation" in the classical sense, as it doesn't happen on the heap.

This made me scratch my head at first, but could remember some similar situation I once ran into.
_edit: This is in fact another issue._

> These buffers are in fact allocated on the heap because of this:
> https://github.com/OpenTTD/OpenTTD/blob/98e653dacc98d308e440503c9cf9bba01933d1b4/src/gfx_func.h#L85
> 
> `DRAW_STRING_BUFFER` is in fact not a compile time constant. There is a gcc extension that allows to use this syntax also with runtime values which allocates the buffer on the heap instead.
> So it would actually be very interresting to turn the constant into a compile-time constant and see how this perfoms:
> ```diff
> -static const int DRAW_STRING_BUFFER = 2048;
> +constexpr int DRAW_STRING_BUFFER = 2048;
> ```
> 
> One can disallow compiler extensions in CMake using [CXX_EXTENSIONS](https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html#prop_tgt:CXX_EXTENSIONS).

vituscze:
`DRAW_STRING_BUFFER` is a constant-expression as far as C++ is concerned. In fact, you can even do:
```cpp
constexpr int x = DRAW_STRING_BUFFER;
```

Using `static const int` as array size produces the usual stack "allocation" in both gcc and clang: subtracting and then adding a fixed number to the stack pointer.

fsimonis:
@vituscze Hi, I did some more digging regarding this and in fact you are right. Thanks for pointing that out!

I also checked the output from heaptrack again and the buffers are in fact not allocated on the heap. I simply misread the information presented by it. Part of getting used to a new tool, I guess.

@LordAro I removed the commit making the buffers static, so no worries about thread safety there.

The allocation reduction stays identical though, so the rest of the claims remain valid.

ldpl:
> Next is the `FormatString` function. Currently, this function uses a stack to keep track of the remaining string.
> This is not necessary as this stack can never reach 2 values. (Correct me if I am wrong.) 

Doesn't look that way to me. Have you tested with something that actually uses SCC_NEWGRF_STRINL or SCC_NEWGRF_PRINT_WORD_STRING_ID? No idea what uses it though but to me it looks like it should break with this patch as it will throw away any leftover symbols after encountering those string codes.

What can probably be done though is postponing stack/vector allocation until it's actually required like how I did it in #7922

fsimonis:
@ldpl Hmm, right.

One could also get rid of the stack by rewriting the function so that it operates on a single formatting string and recursively calls itself for these formatting codes. This makes the code cleaner, reduces the allocations and remains backwards compatible.
I'll look into that.

fsimonis:
Will close this PR. The required refactoring won't have a lot to do with this.

PR #9017, opened 2021-04-11
Feature: play easier together with friends from behind home routers

TrueBrain:
Integrated #9234. Closes #9234.

~Test binaries can be found here: https://www.openttd.org/downloads/openttd-branches/pr9017/latest.html
See below how to help out testing.~

~If you build yourself, make sure to name your branch "pr9017". Otherwise your revision will not match anyone using the binaries above.~

**!!NOTE!!** Currently the Game Coordinator is not working with this PR, while I rebuild this PR from the ground up. See https://gist.github.com/TrueBrain/75c016cdb90efa18cb79f3f6bae9af25 for details.

## Motivation / Problem

Currently, to host a network server you really need to know a few things about routers and networking. This is way to advanced for the average Joe, and this becomes very clear when looking at Steam Discussions.
For Rich Presence (Discord, Steam, ..) we need "invite codes" so people can join each others game.

So, can we do this? I wrote a gist to explain the process of getting to this PR: https://gist.github.com/TrueBrain/f5685dc0fdbd012aac48ea2a36158198

But in case you are not into reading, let me summarize:

- Every invite-only / public server creates a persistent TCP connection to a Game Coordinator (replacing the master-server)
- Every server gets assigned an unique "join key" (called an "invite code" in the UI)
- Server listing returns only the "join key" of a server
- A client can join any server based on this "join key" (without even knowing the IP of the server), either via listing or directly inputting it
- The Game Coordinator will try to connect the server and client. It tries 3 methods: direct connect, STUN, and if all fails, TURN (see gists for details)

There are now three kind of servers:
- Local servers: do not connect to the Game Coordinator. "Search LAN" does work for these servers, as do people with the public IP and if port-forwarding is setup correctly. Just the old "non-advertised" servers.
- Invite-only servers: the owner of the server needs to give the invite-code to someone before he can join.
- Public servers are the old advertised servers.

## Description

See the individual commits and the gist above. Not much more to add.

## Limitations

- Server info is delayed by 30s (by design). You can refresh individual servers if you want up-to-date information.

## Testing

If you want to help out testing, great!

- Start the client by downloading it here: https://www.openttd.org/downloads/openttd-branches/pr9017/latest.html (or build it yourself, make sure you name your branch `pr9017`!)
- Start a server, make it public or invite-only
- Ask someone to join your server

Notice how you didn't have to setup firewalls, port-forwarding, etc?

Let me know if you find any problems, etc.

## Checklist for review

[...]

PR #9406, opened 2021-06-29
Change: [strgen] 'warnings' for translations are now 'infos' unless invoked with -w or -t

glx22:
## Motivation / Problem
A change in gcc-problem-matcher made it pickup strgen warnings for translations. These are usually handled by eints later.

## Description
Use `info` for translation warnings, unless `-w` or `-t` switches are used.
This is done by abusing `_show_todo`.
Tested on #9017 (see output [here](https://github.com/glx22/OpenTTD/runs/2944908462?check_suite_focus=true)).
I also tested locally to be sure `english.txt` still always generate `warning` and not `info`.

## Checklist for review

### DIFF
--- a/src/strgen/strgen.cpp
+++ b/src/strgen/strgen.cpp
@@ -46,7 +46,11 @@ void CDECL strgen_warning(const char *s, ...)
 	va_start(va, s);
 	vseprintf(buf, lastof(buf), s, va);
 	va_end(va);
-	fprintf(stderr, LINE_NUM_FMT("warning"), _file, _cur_line, buf);
+	if (_show_todo > 0) {
+		fprintf(stderr, LINE_NUM_FMT("warning"), _file, _cur_line, buf);
+	} else {
+		fprintf(stderr, LINE_NUM_FMT("info"), _file, _cur_line, buf);
+	}
 	_warnings++;
 }
 
--- a/src/strgen/strgen_base.cpp
+++ b/src/strgen/strgen_base.cpp
@@ -818,8 +818,12 @@ void StringReader::ParseFile()
 	char buf[2048];
 	_warnings = _errors = 0;
 
-	_translation = this->master || this->translation;
+	_translation = this->translation;
 	_file = this->file;
+
+	/* Abusing _show_todo to replace "warning" with "info" for translations. */
+	_show_todo &= 3;
+	if (!this->translation) _show_todo |= 4;
 
 	/* For each new file we parse, reset the genders, and language codes. */
 	MemSetT(&_lang, 0);


src/strgen/strgen.cpp
```diff
@@ -46,7 +46,11 @@ void CDECL strgen_warning(const char *s, ...)
 	va_start(va, s);
 	vseprintf(buf, lastof(buf), s, va);
 	va_end(va);
-	fprintf(stderr, LINE_NUM_FMT("warning"), _file, _cur_line, buf);
+	if (_show_todo > 0) {
+		fprintf(stderr, LINE_NUM_FMT("warning"), _file, _cur_line, buf);
+	} else {
+		fprintf(stderr, LINE_NUM_FMT("info"), _file, _cur_line, buf);
+	}
```
TrueBrain, line 53: why would the classification change depending on a setting? I am not sure I follow that logic. For who will it become a warning that doesn't think info is what he is looking for?

TrueBrain (commented):
This commit message is way to specific :P It is okay to use extra lines in the commit message, like:

```
Change: [strgen] Change warnings for translations into infos

Unless invoked with -w ("show warnings") or -t ("show todos")

Eints normally fixes the warnings after a Pull Request, so it is not
really useful information for the developer to see as a warning.
```

At least, I am guessing that is what -w and -t do, please feel free to correct where needed :P

src/strgen/strgen.cpp
glx22: I'm not sure to fully understand your question.
`-t` sets bit 0 of `_show_todo`, `-w` sets bit 1, and now english.txt sets bit 2.
So if any bit is set it's a warning, else it's an info.

TrueBrain: No, I am wondering why this if-statement exists in the first place. Why would it be a "warning" in some cases, and "info" in others. That always feels a bit weird, and often tells there is something else being a bit off. I haven't checked how other places do it. I just don't get why the log-level would change depending on `-t` for example.

TrueBrain: Okay, a bit closer look, I do not think this is the right approach. There are plenty of warnings that really are warnings, and never infos. Like invalid UTF-8, for example. What if we just add a new command like `strgen_translation_warning`, in that case I could understand this if-statement a lot more :)

Edit: or better yet, just don't output anything if `-w` and `-t` aren't set for translation errors? :D

TrueBrain (approved):

src/strgen/strgen.cpp
TrueBrain: Nevermind all this, it is already messy as fuck. So it is not fair to ask from you to fix "everything" just to fix a bug :P So lets just do this.

PR #9408, opened 2021-07-03
Codechange: replace InjectDParam/ShiftParameters

rubidium42:
## Motivation / Problem

ShiftParameters, and due to that InjectDParam use MemMoveT to move the parameters around. Since this only updates the parameters, but not the types associated with the parameters it cannot be used when types are used.

While doing this also spotted a declaration of a function that was not defined, so that will be removed as well.

## Description

Just set the right parameter indices in one subsidy function to be able to get rid of both ShiftParameters and InjectDParam.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -84,16 +84,6 @@ int64 StringParameters::GetInt64(WChar type)
 }
 
 /**
- * Shift all data in the data array by the given amount to make
- * room for some extra parameters.
- */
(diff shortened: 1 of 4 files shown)


TrueBrain (approved):
The longer I look at this, the more nausea I get .. yeah, let's burn this with fire, shall we?

PR #9409, opened 2021-07-03
Fix cdb3dd049: Loading of GOAL chunks

glx22:
## Motivation / Problem
Loading of pre-table chunks must use vector returned by `SlCompatTableHeader` not the `Saveload` array.

## Description
Use the right thing.

## Checklist for review


TrueBrain (approved):
Bit weird commit message, but otherwise looks good. Tnx for fixing it :)

The commit message suggests loading doesn't work currently, but that is not true. It will only fail if things get changed. So I suggest something like:

`Fix cdb3dd049: GOAL chunk was not using the header for loading`

Or something.

PR #9410, opened 2021-07-03
Fix: Inconsistent error message when overbuilding canal on a lock/ship depot tile

SamuXarick:
## Motivation / Problem
This is a delicate issue.
It originated when my AI was overbuilding a canal on the entry/exit of a lock tile. If the waterclass was canal, I would get a 'Can't build canals here... ... already built' error message, but if the waterclass was sea, I would get 'Can't build canals here... Building must be demolished first'.

https://github.com/OpenTTD/OpenTTD/blob/d38ad7d80ca592612ce7bc98e770e40ae778fd04/src/water_cmd.cpp#L478-L485

First I discovered that `IsTileType(current_tile, MP_WATER)` can't be used here, due to Ship Depots and Locks also being tiles of the same type `MP_WATER `as Canals. I changed to `IsWaterTile(current_tile)` and I thought that would be enough to fix the problem, but it wasn't.

Now I was getting 'Can't build canals here... Must demolish canal first'.

I found that `DC_FORCE_CLEAR_TILE` was in the way. It would end up doing these checks:
https://github.com/OpenTTD/OpenTTD/blob/d38ad7d80ca592612ce7bc98e770e40ae778fd04/src/landscape.cpp#L696-L702

I still tried adding some extra checks to line 698, to exclude locks and ship depots, but that would mean I was going against the purpose of `DC_FORCE_CLEAR_TILE`.

Went back to the origin of this flag, at line 482 of water_cmd.cpp. Decided to remove the flag, and that finally solved the issue.

Now, the error message is consistent, regardless of waterclass: 'Can't build canals here... Building must be demolished first'. They reach `ClearTile_Water` to get the error string.

There is however, behaviour changes when experimenting with tiles of type `MP_OBJECT`. I Loaded NewGRF 'OpenGFX+ Landscape 1.1.2'. When we have a canal, and a rocky land on the canal, overbuilding the canal no longer comes up with an error message saying 'Can't build canals here... ... Must demolish canal first'. Now it overbuilds with no error. Also, if we build a canal on top of a rocky land on sea, the price is also different. It doesn't include the price of clearing sea.

## Checklist for review

### DIFF
--- a/src/water_cmd.cpp
+++ b/src/water_cmd.cpp
@@ -475,10 +475,11 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 			return_cmd_error(STR_ERROR_FLAT_LAND_REQUIRED);
 		}
 
-		/* can't make water of water! */
-		if (IsTileType(current_tile, MP_WATER) && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
-
 		bool water = IsWaterTile(current_tile);
+
+		/* can't make water of water! */
+		if (water && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
+
 		ret = DoCommand(current_tile, 0, 0, flags, CMD_LANDSCAPE_CLEAR);
 		if (ret.Failed()) return ret;
 
--- a/src/water_cmd.cpp
+++ b/src/water_cmd.cpp
@@ -479,7 +479,7 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 		if (IsTileType(current_tile, MP_WATER) && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
 
 		bool water = IsWaterTile(current_tile);
-		ret = DoCommand(current_tile, 0, 0, flags | DC_FORCE_CLEAR_TILE, CMD_LANDSCAPE_CLEAR);
+		ret = DoCommand(current_tile, 0, 0, flags, CMD_LANDSCAPE_CLEAR);
 		if (ret.Failed()) return ret;
 
 		if (!water) cost.AddCost(ret);


SamuXarick:
Apparently, this also fixes inconsistent error messages when overbuilding on dock-water-parts and buoys on sea/canal.

TrueBrain:
Wow. This PR is really hard to read .. "rambling by a madman" .. "a day in the life of Samu" .. please do not do this to us. It makes it really hard to understand what you are trying to fix.

Additionally, this PR is basically: I fiddled till it worked. And that brings me to: a broken clock is even right twice a day. So let's dive in, see what happened, and what this change actually does.

Regarding `DC_FORCE_CLEAR_TILE`:
In 2010, https://github.com/OpenTTD/OpenTTD/commit/0e250f2bdfc8eb7f435f59398766f7774e31610c makes the change that building canals over seas also charge you for the price of removing the sea first, because .. logic? the reasoning is never explained, so I have no clue.
In 2013, https://github.com/OpenTTD/OpenTTD/commit/00530f4ccfd6c076b9ca0d4996ce4a185b5a0311 makes the change to remove any water-based costs when building canals, with the reasoning: you are not going to drain the sea first, just to fill it back up again.

Those two commits kinda contradict each other. Given the second negates the use for `DC_FORCE_CLEAR_TILE`, the change in this PR is fine as far as I can tell.

As for the `IsTileType(current_tile, MP_WATER)`, I think this is simply forgotten to be changed over the years into `IsWaterTile`, as that clearly is the intention of that line (given the comment above). It dates back to the first revision we can track, in a time where `MP_WATER` truly meant: this is water. So the change in this PR is fine there too, as the comment clearly states the intention is to only prevent making water into water.

But, that said, when we dive a bit deeper in what that if-statement does, we notice the function has changed so drastically, it no longer really means what it used to. Take this change for example, https://github.com/OpenTTD/OpenTTD/commit/a12c748f5bdab0619e2ac8008aaf9324fe2a6419, here it was added that using CTRL when placing canals meant placing sea. For that this check is useful, to prevent water being made to water, to nudge the player in the right direction.
However, this functionality no longer exists. Only in the SE you can place sea, and even that is limited. More important, if you place sea over sea you get "Canal already build" error. Which is weird.
Edit: basically the only thing this check does these days, is prevent you from building canals on things owned by `OWNER_WATER` (or sea). Read: rivers and sea.

~~So I suggest we completely remove this if-statement. That brings building sea on-par with building rivers: if you build sea over sea, or river over river, no error is given.~~

Finally, this PR really should be two commits: one removing the `DC_FORCE_CLEAR_TILE`, indicating it didn't do anything anymore. The second removing the if-statement, stating it no longer does anything.

@SamuXarick : Are you up for that? Or are you like: way over my head, please take care of it?

Edit: I suggest two commits with messages something like this:

`DC_FORCE_CLEAR_TILE` remove commit:
```
Fix: reduce cost of building canals over objects on sea

It is not like we will drain the sea first, to put water back in it after.
Besides, the cost for draining the sea isn't calculated for all other cases either.
```

Changing to `IsWaterTile`:
```
Fix: wrong error message when building canals over ship depots / locks

IsTileType() also considers ship depots and locks water. IsWaterTile() does the right thing.
```

src/water_cmd.cpp
```diff
@@ -475,11 +475,12 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 			return_cmd_error(STR_ERROR_FLAT_LAND_REQUIRED);
 		}
 
+		bool water = IsWaterTile(current_tile);
+
 		/* can't make water of water! */
-		if (IsTileType(current_tile, MP_WATER) && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
+		if (water && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
```
TrueBrain, line 481: Please completely remove this if-statement. It has no reason to exist anymore.

TrueBrain: @SamuXarick points out I am wrong, and it should exist. But indeed, it should at least be `IsWaterTile` instead of the `MP_WATER` check :)

TrueBrain (approved):

PR #9411, opened 2021-07-03
Replace master-server (UDP-based) with Game Coordinator (TCP-based)

TrueBrain:
## Motivation / Problem

This is a part of #9017.

Currently OpenTTD communicates over UDP with the master-server to register itself, and the master-server uses UDP to polls the server for game-state. This has several disadvantages, as we noticed over the years:

- The users that do manage to setup a port-forwarding, often only manage to do this for TCP. Doing it for UDP too, is often either difficult in routers, or just not something they understand. As such, people have a lot of trouble setting up portforwarding in such a way the master-server can see the server. But to make matters worse, without UDP port-forwarding, people can join a server via the direct IP. This is frustrating to users.
- UDP is, by definition, unreliable. So we have a lot of code to retry several times and trying to recover.
- UDP has a hard MTU-size, as fragmentation is impossible. We try to be on the safe-side with the MTU we picked, but there are known scenarios where you cannot host OpenTTD because the connection doesn't allow such large UDP packets to travel through it. (mostly GRE links etc) Explaining this to users is very very difficult.
- #9017 works towards adding STUN support. For this we need a persistent connection between something on the openttd.org infra and the server, so we can push messages to the server of which we know they arrive safely. In theory we could use UDP for this, and with some smart keep-alive etc we can keep a bi-direction connection alive .. but it is tricky. More-over, most infrastructure, also on our hosting side, is more aimed at TCP connections than UDP connections, and several issues are to be expected.
- Pretty sure I miss a few arguments here.

In total, it became pretty clear we should just stop using UDP, and switch completely to TCP. Well, I say completely .. local LAN discovery is and always will be UDP based. That is just how multicasts work ;)

To be clear, this doesn't mean OpenTTD will always be using TCP. It is surely an option to use some library like https://github.com/ValveSoftware/GameNetworkingSockets that does resilient connections over UDP. Either way, we first have to change to a single way of doing stuff, before we can think about changing it all to something else. And this is far out of scope of this PR.

## Description

This removes all master-server communication and all UDP-related code from OpenTTD, and replaces it with a TCP-based Game Coordinator code.
Again, the local LAN discovery still uses UDP and remains.

Why the change of name? To keep it clear that this is a completely different component. And in the future the Game Coordinator will be tasked with many other tasks, like join-keys, STUN, TURN, etc. It much more takes on the role of a coordinator, and no longer "just" lists the servers.
In the background, master-server and Game Coordinator communicate with each other, so the server-listing will be a combination of both for 1.12+. 1.11 will only see 1.11 and older servers.

The way it works is like this (and yes, this is documented in `network_coordinator.cpp`, just to beat LordAro to the punch :P):

```cpp
/**
 * Game Coordinator communication.
 *
 * For servers:
 *  - Server sends CLIENT_REGISTER.
 *  - GC probes server to check if it can directly connect.
 *  - GC sends SERVER_REGISTER_ACK with type of connection.
 *  - Server sends every 30 seconds CLIENT_UPDATE.
 *
 * For clients (listing):
 *  - Client sends CLIENT_LISTING.
 *  - GC returns the full list of public servers via SERVER_LISTING (multiple packets).
 */
```

Basically this means as soon as the client clicks "Search Internet", it gets the full list of servers, including their details. It no longer has to send a flood of UDP packets to all servers.

`Private` is renamed to `Local`, as there is no such thing as `Private` .. a server is either accessible locally (or by the IP if you happen  to know that), or publicly listed. This change makes the later addition of `Invite Only` a lot more sensible to the user.

Future PRs will extend this functionality with invite-codes, STUN, TURN, ... But, baby-steps.

## Limitations

Currently there are a few rough edged, but further PRs will address those.  For example:

- The `CLIENT_UPDATE` sends all the NewGRF data every time. But as they can never change, this is a waste of bandwidth.
- Currently servers can only register via either IPv4 or IPv6, not both. This is because future PR will use the STUN requests to discover the IPs. I could strictly seen already add this in this PR, but that would increase the change drastically. And solving it differently is rather wasting my time a lot, as I have to create something I have to remove a few PRs later.
- Currently servers registering via IPv6 will be told that they do not have their firewall open. This might not be true, but the Game Coordinator can't query IPv6 yet.
- Currently servers using 1.11.2 or earlier are not showing up in the server listing. This will be addressed before 1.12.

## Checklist for review

### DIFF
--- a/src/network/core/config.h
+++ b/src/network/core/config.h
@@ -12,8 +12,6 @@
 #ifndef NETWORK_CORE_CONFIG_H
 #define NETWORK_CORE_CONFIG_H
 
-/** DNS hostname of the masterserver */
-static const char * const NETWORK_MASTER_SERVER_HOST            = "master.openttd.org";
 /** DNS hostname of the Game Coordinator server */
 static const char * const NETWORK_COORDINATOR_SERVER_HOST       = "coordinator.openttd.org";
 /** DNS hostname of the content server */
@@ -22,10 +20,7 @@ static const char * const NETWORK_CONTENT_SERVER_HOST           = "content.opent
 static const char * const NETWORK_CONTENT_MIRROR_HOST           = "binaries.openttd.org";
 /** URL of the HTTP mirror system */
 static const char * const NETWORK_CONTENT_MIRROR_URL            = "/bananas";
-/** Message sent to the masterserver to 'identify' this client as OpenTTD */
-static const char * const NETWORK_MASTER_SERVER_WELCOME_MESSAGE = "OpenTTDRegister";
 
-static const uint16 NETWORK_MASTER_SERVER_PORT      = 3978;           ///< The default port of the master server (UDP)
 static const uint16 NETWORK_COORDINATOR_SERVER_PORT = 3976;           ///< The default port of the Game Coordinator server (TCP)
 static const uint16 NETWORK_CONTENT_SERVER_PORT     = 3978;           ///< The default port of the content server (TCP)
 static const uint16 NETWORK_CONTENT_MIRROR_PORT     =   80;           ///< The default port of the content mirror (TCP)
@@ -54,7 +49,6 @@ static const uint16 COMPAT_MTU                      = 1460;           ///< Numbe
 static const byte NETWORK_GAME_ADMIN_VERSION        =    1;           ///< What version of the admin network do we use?
 static const byte NETWORK_GAME_INFO_VERSION         =    4;           ///< What version of game-info do we use?
 static const byte NETWORK_COMPANY_INFO_VERSION      =    6;           ///< What version of company info is this?
-static const byte NETWORK_MASTER_SERVER_VERSION     =    2;           ///< What version of master-server-protocol do we use?
 static const byte NETWORK_COORDINATOR_VERSION       =    1;           ///< What version of game-coordinator-protocol do we use?
 
 static const uint NETWORK_NAME_LENGTH               =   80;           ///< The maximum length of the server name and map name, in bytes including '\0'
--- a/src/network/core/config.h
+++ b/src/network/core/config.h
@@ -14,6 +14,8 @@
 
 /** DNS hostname of the masterserver */
 static const char * const NETWORK_MASTER_SERVER_HOST            = "master.openttd.org";
+/** DNS hostname of the Game Coordinator server */
+static const char * const NETWORK_COORDINATOR_SERVER_HOST       = "coordinator.openttd.org";
 /** DNS hostname of the content server */
 static const char * const NETWORK_CONTENT_SERVER_HOST           = "content.openttd.org";
 /** DNS hostname of the HTTP-content mirror server */
@@ -23,14 +25,15 @@ static const char * const NETWORK_CONTENT_MIRROR_URL            = "/bananas";
 /** Message sent to the masterserver to 'identify' this client as OpenTTD */
 static const char * const NETWORK_MASTER_SERVER_WELCOME_MESSAGE = "OpenTTDRegister";
 
-static const uint16 NETWORK_MASTER_SERVER_PORT    = 3978;         ///< The default port of the master server (UDP)
-static const uint16 NETWORK_CONTENT_SERVER_PORT   = 3978;         ///< The default port of the content server (TCP)
-static const uint16 NETWORK_CONTENT_MIRROR_PORT   =   80;         ///< The default port of the content mirror (TCP)
-static const uint16 NETWORK_DEFAULT_PORT          = 3979;         ///< The default port of the game server (TCP & UDP)
-static const uint16 NETWORK_ADMIN_PORT            = 3977;         ///< The default port for admin network
-static const uint16 NETWORK_DEFAULT_DEBUGLOG_PORT = 3982;         ///< The default port debug-log is sent to (TCP)
+static const uint16 NETWORK_MASTER_SERVER_PORT      = 3978;           ///< The default port of the master server (UDP)
+static const uint16 NETWORK_COORDINATOR_SERVER_PORT = 3976;           ///< The default port of the Game Coordinator server (TCP)
+static const uint16 NETWORK_CONTENT_SERVER_PORT     = 3978;           ///< The default port of the content server (TCP)
+static const uint16 NETWORK_CONTENT_MIRROR_PORT     =   80;           ///< The default port of the content mirror (TCP)
+static const uint16 NETWORK_DEFAULT_PORT            = 3979;           ///< The default port of the game server (TCP & UDP)
+static const uint16 NETWORK_ADMIN_PORT              = 3977;           ///< The default port for admin network
+static const uint16 NETWORK_DEFAULT_DEBUGLOG_PORT   = 3982;           ///< The default port debug-log is sent to (TCP)
 
-static const uint16 UDP_MTU                       = 1460;         ///< Number of bytes we can pack in a single UDP packet
+static const uint16 UDP_MTU                         = 1460;           ///< Number of bytes we can pack in a single UDP packet
 /*
  * Technically a TCP packet could become 64kiB, however the high bit is kept so it becomes possible in the future
  * to go to (significantly) larger packets if needed. This would entail a strategy such as employed for UTF-8.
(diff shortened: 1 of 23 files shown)


src/network/core/tcp_coordinator.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "../../debug.h"
+#include "tcp_coordinator.h"
+
+#include "../../safeguards.h"
+
+/**
+ * Handle the given packet, i.e. pass it to the right
+ * parser receive command.
+ * @param p the packet to handle
+ * @return true if we should immediately handle further packets, false otherwise
```
rubidium42: Missing trailing dots?

src/network/core/tcp_coordinator.cpp
```diff
@@ -0,0 +1,73 @@
...
+		case PACKET_COORDINATOR_CLIENT_UPDATE:         return this->Receive_CLIENT_UPDATE(p);
+
+		default:
+			Debug(net, 0, "[tcp/coordinator] Received invalid packet type {}", type);
+			return false;
+	}
+}
+
+/**
+ * Receive a packet at TCP level
```
rubidium42: Missing trailing dot?

src/network/core/tcp_coordinator.h
```diff
@@ -0,0 +1,110 @@
...
+
+#ifndef NETWORK_CORE_TCP_COORDINATOR_H
+#define NETWORK_CORE_TCP_COORDINATOR_H
+
+#include "os_abstraction.h"
+#include "tcp.h"
+#include "packet.h"
+#include "game_info.h"
+
+/** Enum with all types of TCP Game Coordinator packets. The order MUST not be changed **/
```
rubidium42: Missing trailing dot?

src/network/network_coordinator.cpp
```diff
@@ -0,0 +1,227 @@
...
+		std::string connection_type;
+		switch (_network_server_connection_type) {
+			case CONNECTION_TYPE_ISOLATED: connection_type = "Remote players can't connect (your server doesn't allow remote connections)"; break;
+			case CONNECTION_TYPE_DIRECT:   connection_type = "Public"; break;
+
+			case CONNECTION_TYPE_UNKNOWN: // Never returned from Game Coordinator.
+			default: connection_type = "Unknown"; break; // Should never happen, but don't fail if it does.
+		}
+
+		Debug(net, 3, " ");
```
rubidium42: Won't an empty string work equally well? Now it looks a bit weird just printing a space.
On a side note I am wondering the use of an empty line in the debug/log output; it makes logs longer and will not help as much as the lines with all dashes used below.

src/network/network_coordinator.cpp
```diff
@@ -0,0 +1,227 @@
...
+	if (this->sock == INVALID_SOCKET) {
+		static std::chrono::steady_clock::time_point last_attempt = {};
+
+		/* Don't auto-reconnect when we are not a server. */
+		if (!_network_server) return;
+		/* Don't reconnect if we are connecting. */
+		if (this->connecting) return;
+		/* Throttle how often we try to reconnect. Otherwise after an outage
+			* all servers reconnect at the same time, potentially overwhelming
+			* the Game Coordinator. */
```
rubidium42: Indenting issue of the comment. Last tab to space.

src/network/network_coordinator.cpp
```diff
@@ -0,0 +1,227 @@
...
+		if (std::chrono::steady_clock::now() < last_attempt + std::chrono::seconds(1) * last_attempt_backoff) return;
+
+		last_attempt = std::chrono::steady_clock::now();
+		/* Delay reconnecting with up to 32 seconds. */
+		if (last_attempt_backoff < 32) {
+			last_attempt_backoff *= 2;
+		}
+
+		Debug(net, 1, "Connection with Game Coordinator lost; reconnecting ...");
```
rubidium42: Excluding lang all (55) but two cases of "..." in/near strings did not have a space in front. For consistency that space should not be here either I guess.
`grep -R '\.\.\.' .|grep '"'|grep -v /lang/`

src/network/network_coordinator.cpp
```diff
@@ -0,0 +1,227 @@
...
+		this->Register();
+		return;
+	} else if (last_attempt_backoff != 1) {
+		last_attempt_backoff = 1;
```
rubidium42: The result of this case is that `last_attempt_backoff` is 1, so maybe just remove the whole if as the check and conditional might be more expensive than just writing it. It also gives slightly simpler code. Unless this case gets more complex in a later step, then it is fine.

src/network/network_coordinator.h
```diff
@@ -0,0 +1,51 @@
...
+class ClientNetworkCoordinatorSocketHandler : public NetworkCoordinatorSocketHandler {
+private:
+	std::chrono::steady_clock::time_point next_update; ///< When to send the next update (if server and public).
+
+protected:
+	bool Receive_SERVER_ERROR(Packet *p) override;
+	bool Receive_SERVER_REGISTER_ACK(Packet *p) override;
+
+public:
+	bool connecting; ///< Are we connecting to the Game Coordinator?
```
rubidium42, line 45: Sounds like some internal state to me that others should not touch directly, but rather via some accessor. Are there any external functions that should be able to write this field?

src/network/core/tcp_coordinator.h
```diff
@@ -95,6 +97,33 @@ class NetworkCoordinatorSocketHandler : public NetworkTCPSocketHandler {
...
+	 *
+	 * @param p The packet that was just received.
+	 * @return True upon success, otherwise false.
+	 */
+	virtual bool Receive_CLIENT_LISTING(Packet *p);
+
+	/**
+	 * Game Coordinator replies with a list of all public servers. Multiple
+	 * of these packets are received after a request till all servers are
+	 * send over. Last packet will have server count of 0.
```
rubidium42: 'are sent over'? Must be checked with someone else as I'm not 100% sure.

src/network/core/tcp_coordinator.h
```diff
@@ -95,6 +97,33 @@ class NetworkCoordinatorSocketHandler : public NetworkTCPSocketHandler {
...
+	 *
+	 *  uint16  Amount of public servers in this packet
```
rubidium42: Missing trailing dot?

src/network/network_gui.cpp
```diff
@@ -448,7 +462,7 @@ class NetworkGameWindow : public Window {
 	}
 
 public:
-	NetworkGameWindow(WindowDesc *desc) : Window(desc), name_editbox(NETWORK_CLIENT_NAME_LENGTH), filter_editbox(120)
+	NetworkGameWindow(WindowDesc *desc) : Window(desc), name_editbox(NETWORK_CLIENT_NAME_LENGTH), filter_editbox(120), searched_internet(false)
```
rubidium42: Wasn't the consensus to use `bool searched_internet = false; ///< Did we ever...` instead?

src/network/core/config.h
```diff
@@ -25,7 +23,6 @@ static const char * const NETWORK_CONTENT_MIRROR_URL            = "/bananas";
 /** Message sent to the masterserver to 'identify' this client as OpenTTD */
 static const char * const NETWORK_MASTER_SERVER_WELCOME_MESSAGE = "OpenTTDRegister";
```
rubidium42: This constant might be something else to remove based on the name and comment.

src/network/network_coordinator.cpp
TrueBrain: Both look equally weird, so I have been going back and forth between those 2 solutions :P But I guess now I will go back to the other one :D

src/network/network_coordinator.h
TrueBrain: This is basically the same as how `network_content.cpp` does this.

The main issue is that there is a `TCPConnector` which sets this value. So it has to be public. Or with friends ofc. But in general I tried not to walk too far from how other parts of the network code did this, to not create yet-another-way-of-doing-shit :P For sure it could use some code-cleanup. But not for this PR, I suggest :)

embeddedt:
Another argument for dropping UDP support: ngrok (and several other reverse-tunneling services) only work with TCP, so this is a very welcome change. :+1:

TrueBrain:
I think this is ready. Game Coordinator is deployed, and talks (two-way) with the Master Server. So anything listed in the Master Server shows up via the Game Coordinator, and the other way around.

Of course you cannot join those servers that are exchanged, as they for sure are of a different revision, but that is not really the point here ;)

It could use some testing of some people, just registering their server, see if they can find it, etc etc. But .. I think this is ready :)

src/network/core/tcp_coordinator.h
```diff
@@ -0,0 +1,138 @@
...
+#ifndef NETWORK_CORE_TCP_COORDINATOR_H
+#define NETWORK_CORE_TCP_COORDINATOR_H
+
+#include "os_abstraction.h"
+#include "tcp.h"
+#include "packet.h"
+#include "game_info.h"
+
+/** Enum with all types of TCP Game Coordinator packets. The order MUST not be changed. **/
+enum PacketCoordinatorType {
```
rubidium42, line 27: I think it would be useful to add some extra documentation as here there are two concepts getting intermingled. You have the game server, game client and game coordinator, and the protocol server and client. The game coordinator being the server in the protocol is right, though you get things like CLIENT_REGISTER which is the server registering to the coordinator. What about:

There are three types of entities in this protocol. The game coordinator to which game servers register and game clients request the list of servers, the game server that host the games that are played and the game clients that are the end users playing the games. In this protocol the game coordinator is seen as the server (everything connects to the game coordinator) and as such packets sent by the game coordinator have SERVER in the type name. The game server and game client connect to the game server and as such packets sent by them have the CLIENT in the type name.

Next to that I would prefix server and client in the comments below with game to make the distinction between the protocol client/server and the game client/server.

rubidium42 (approved):

PR #9413, opened 2021-07-04
Codechange: Introduce and use ChunkHandler sub-classes

glx22:
## Motivation / Problem
`ChunkHandler` uses pointers to functions to determine what to do on saving, loading, load checking, pointer fixing.
That's a legacy of its C origin.

## Description
Move `ChunkHandler` to C++ world, and benefit from inheritance.

Diff is easier to read with "Hide whitespace changes" enabled.

## Checklist for review

### DIFF
--- a/src/saveload/airport_sl.cpp
+++ b/src/saveload/airport_sl.cpp
@@ -14,28 +14,36 @@
 
 #include "../safeguards.h"
 
-static void Save_APID()
-{
-	Save_NewGRFMapping(_airport_mngr);
-}
+struct APIDChunkHandler : ChunkHandler {
+	APIDChunkHandler() : ChunkHandler('APID', CH_TABLE) {}
 
-static void Load_APID()
-{
-	Load_NewGRFMapping(_airport_mngr);
-}
+	void Save() const override
+	{
+		Save_NewGRFMapping(_airport_mngr);
+	}
 
-static void Save_ATID()
-{
-	Save_NewGRFMapping(_airporttile_mngr);
-}
+	void Load() const override
+	{
+		Load_NewGRFMapping(_airport_mngr);
+	}
+};
 
-static void Load_ATID()
-{
-	Load_NewGRFMapping(_airporttile_mngr);
-}
+struct ATIDChunkHandler : ChunkHandler {
+	ATIDChunkHandler() : ChunkHandler('ATID', CH_TABLE) {}
 
-static const ChunkHandler ATID{ 'ATID', Save_ATID, Load_ATID, nullptr, nullptr, CH_TABLE };
-static const ChunkHandler APID{ 'APID', Save_APID, Load_APID, nullptr, nullptr, CH_TABLE };
+	void Save() const override
+	{
+		Save_NewGRFMapping(_airporttile_mngr);
+	}
+
+	void Load() const override
+	{
+		Load_NewGRFMapping(_airporttile_mngr);
+	}
+};
+
+static const ATIDChunkHandler ATID;
+static const APIDChunkHandler APID;
 static const ChunkHandlerRef airport_chunk_handlers[] = {
 	ATID,
 	APID,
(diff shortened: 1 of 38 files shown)


src/saveload/airport_sl.cpp
```diff
@@ -14,29 +14,39 @@
...
-
-static const ChunkHandler airport_chunk_handlers[] = {
-	{ 'ATID', Save_ATID, Load_ATID, nullptr, nullptr, CH_TABLE },
-	{ 'APID', Save_APID, Load_APID, nullptr, nullptr, CH_TABLE },
+struct APIDChunkHandler : ChunkHandler {
+	APIDChunkHandler() : ChunkHandler('APID', CH_TABLE) {}
+
+	void Save() const override
+	{
+		Save_NewGRFMapping(_airport_mngr);
```
rubidium42: Might not be really in scope here, but all these NewGRF mappings are essentially the same except for the ID and the reference table. May it be useful to just have one NewGRFMappingChunkHandler with chunk ID and a reference to the table?

src/saveload/economy_sl.cpp
```diff
@@ -76,41 +88,49 @@ static const SaveLoad _cargopayment_desc[] = {
...
 
+static const CAPYChunkHandler CAPY;
+static const PRICChunkHandler PRIC;
+static const CAPRChunkHandler CAPR;
+static const ECMYChunkHandler ECMY;
+static const ChunkHandlerRef economy_chunk_handlers[] = {
+	CAPY,
+	PRIC,
+	CAPR,
+	ECMY,
```
rubidium42: Stray newline?

src/saveload/saveload.h
```diff
@@ -401,15 +400,23 @@ enum ChunkType {
...
-	ChunkSaveLoadProc *load_check_proc; ///< Load procedure for game preview.
 	ChunkType type;                     ///< Type of the chunk. @see ChunkType
+
+	ChunkHandler(uint32 id, ChunkType type) : id(id), type(type) {}
+
+	virtual ~ChunkHandler() {}
+
+	virtual void Save() const { NOT_REACHED(); }
+	virtual void Load() const = 0;
+	virtual void FixPointers() const {}
```
rubidium42, line 428: Save and Load are pretty self-explanatory, though maybe it is worth to explain in some documentation what FixPointers and LoadCheck are doing, and especially when they should be implemented and when implementing them is a waste of effort.
For Save it might be useful to mention that if the chunk is only needed for loading old savegames and it should not be added to the savegame, then it can be left unimplemented but CH_READONLY must be set as chunk type.

src/saveload/airport_sl.cpp
glx22: Ah yes, that's one of the thing this PR allows.

src/saveload/economy_sl.cpp
glx22: Oups, somehow a conflict fixing failure :)

src/saveload/saveload.h
glx22: Based on the old `load_check_proc` comment, maybe `LoadCheck()` should be renamed to `LoadPreview()`.

rubidium42: I would leave it as `LoadCheck()` for now. Otherwise you would need to rename everything related to it as well, so something for a future PR.

src/saveload/saveload.h
```diff
@@ -401,15 +400,45 @@ enum ChunkType {
...
+	/**
+	 * Load the chunk.
+	 * Must be overriden.
+	 */
+	virtual void Load() const = 0;
+
+	/**
+	 * Fix the pointers.
+	 * Pointers are saved using the index of the pointed object.
+	 * On load, pointers are filled with indices and need to be fixed to point the the real object.
```
rubidium42: `the the` -> `to the` on this line.
`overriden` -> `overridden` in all these functions.

rubidium42 (dismissed):

rubidium42 (approved):

PR #9414, opened 2021-07-05
Codechange: Split off settings.cpp into several separate files

rubidium42:
## Motivation / Problem

Potential workaround for the issue #9386. By splitting the settings table and saveload code off from the main settings file, the settings become more manageable in file size. It also provides a better separation of concerns.

## Description

As the motivation already mentioned. Move the table/settings.h include from settings.cpp to a new settings_table.cpp and move the saveload specific code to saveload/settings_sl.cpp.

Review hint: `git diff --color-moved` as the vast majority is just moving code around. Most work was in reducing the includes though.

## Limitations

This won't necessarily work forever with adding settings, though... one could then split of one of the tables into a separate cpp file, or potentially one for each of the tables. In that case settings_table.h does not need to be amended, just settings_table.cpp needs to be split.

## Checklist for review

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -338,6 +338,8 @@ add_files(
     settings_gui.cpp
     settings_gui.h
     settings_internal.h
+    settings_table.h
+    settings_table.cpp
     settings_type.h
     ship.h
     ship_cmd.cpp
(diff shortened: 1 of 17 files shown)


glx22:
Possible fix for VS2017 regression fail (div by 0 in AfterLoad() of intro game) in https://github.com/OpenTTD/OpenTTD/compare/master...glx22:pr/9414

I think it was working fine before the split because everything was in the same translation unit.

TrueBrain (approved):
Lets see how long it takes before it comes back ...

PR #9415, opened 2021-07-06
Fix 39e90ec: Integers for scripts are 64bit, but saved as 32bit

glx22:
## Motivation / Problem
In https://github.com/OpenTTD/OpenTTD/commit/39e90ec6e01f56cea5923de2da05d7467bef20ab, integer for scripts was made 64bits always.
But before this change they were always saved as 32bits, mainly for compatibility. 

## Description
Now they are saved as 64bit. A savegame bump is of course required.

## Limitations
May need some documentation fixes and updates too.

## Checklist for review

### DIFF
--- a/src/saveload/saveload.h
+++ b/src/saveload/saveload.h
@@ -335,6 +335,7 @@ enum SaveLoadVersion : uint16 {
 	SLV_RIFF_TO_ARRAY,                      ///< 294  PR#9375 Changed many CH_RIFF chunks to CH_ARRAY chunks.
 
 	SLV_TABLE_CHUNKS,                       ///< 295  PR#9322 Introduction of CH_TABLE and CH_SPARSE_TABLE.
+	SLV_SCRIPT_INT64,                       ///< 296  PR#9415 SQInteger is 64bit but was saved as 32bit.
 
 	SL_MAX_VERSION,                         ///< Highest possible saveload version
 };
--- a/src/script/script_instance.cpp
+++ b/src/script/script_instance.cpp
@@ -364,8 +364,8 @@ static const SaveLoad _script_byte[] = {
 			SQInteger res;
 			sq_getinteger(vm, index, &res);
 			if (!test) {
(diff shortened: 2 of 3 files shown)


src/saveload/saveload.h
```diff
@@ -335,6 +335,7 @@ enum SaveLoadVersion : uint16 {
 	SLV_RIFF_TO_ARRAY,                      ///< 294  PR#9375 Changed many CH_RIFF chunks to CH_ARRAY chunks.
 
 	SLV_TABLE_CHUNKS,                       ///< 295  PR#9322 Introduction of CH_TABLE and CH_SPARSE_TABLE.
+	SLV_AI_INT64,                           ///< 296  PR#9415 SQInteger is 64bit but was saved as 32bit.
```
rubidium42: Should probably be `SLV_SCRIPT_INT64` due to it being more generic than just AIs as the commit message already mentions.

src/script/script_instance.cpp
```diff
@@ -364,8 +364,8 @@ static const SaveLoad _script_byte[] = {
 			SQInteger res;
 			sq_getinteger(vm, index, &res);
 			if (!test) {
-				int value = (int)res;
-				SlCopy(&value, 1, SLE_INT32);
+				int64 value = (int64)res;
```
rubidium42, line 367: I think the cast should not be needed as SQInteger is already defined at `__int64`. Explicitly having a separate variable of the right type is definitely a good thing.

rubidium42 (commented):
I haven't found any documentation mentioning saving 32 bits values, so it's safe on that front. However, the links (2) in the documentation of ScriptController to the AI:Save/Load page on the wiki are broken. I guess they should now be https://wiki.openttd.org/en/Development/Script/Save%20and%20Load. That page does not seem to mention the values being 32 bits either, so it doesn't need to be updated.

rubidium42 (approved):

PR #9416, opened 2021-07-07
Codechange: Simplify FormatString

fsimonis:
## Motivation / Problem

`FormatString` uses a stack to handle formatting of inlined strings.
This is overly complex and leads to unnecessary allocations in the drawing code of OpenTTD, 

## Description

Based on #9405, this PR simplifies `FormatString` by removing its stack and replacing it by recursion for the format codes  `SCC_NEWGRF_STRINL` and `SCC_NEWGRF_PRINT_WORD_STRING_ID`.
This change removes the use of temporary allocations in this function.
With this change, the overall temporary allocations are down by ~90% for the test case mentioned in #9405.

I am not sure why, but `SCC_NEWGRF_PRINT_WORD_STRING_ID` got triggered repeatedly in world generation.

## Limitations

I was unable to test the changes for the affected format code `SCC_NEWGRF_STRINL` as I couldn't find a NewGrf which uses it.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -926,7 +926,8 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 
 			case SCC_NEWGRF_STRINL: {
 				StringID substr = Utf8Consume(&str);
-				buff = FormatString(buff, GetStringPtr(substr), args, last, case_index, game_script, dry_run);
+				buff = FormatString(buff, GetStringPtr(substr), args, last, next_substr_case_index, game_script, dry_run);
+				next_substr_case_index = 0;
 				break;
 			}
 
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -932,8 +932,7 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 
 			case SCC_NEWGRF_PRINT_WORD_STRING_ID: {
 				StringID substr = args->GetInt32(SCC_NEWGRF_PRINT_WORD_STRING_ID);
-				buff = FormatString(buff, GetStringPtr(substr), args, last, case_index, game_script, dry_run);
-				case_index = next_substr_case_index;
+				buff = FormatString(buff, GetStringPtr(substr), args, last, next_substr_case_index, game_script, dry_run);
 				next_substr_case_index = 0;
 				break;
 			}
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -818,8 +818,7 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 	uint next_substr_case_index = 0;
 	char *buf_start = buff;
 	const char *current_str_arg = str_arg;
-	for (;;) {
-		if ((b = Utf8Consume(&current_str_arg)) == '\0') break;
+	while ((b = Utf8Consume(&current_str_arg)) != '\0') {
 		const char *&str = current_str_arg;
 
 		if (SCC_NEWGRF_FIRST <= b && b <= SCC_NEWGRF_LAST) {
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -817,7 +817,7 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 	WChar b = '\0';
 	uint next_substr_case_index = 0;
 	char *buf_start = buff;
-	const char* current_str_arg = str_arg;
+	const char *current_str_arg = str_arg;
 	for (;;) {
 		if ((b = Utf8Consume(&current_str_arg)) == '\0') break;
 		const char *&str = current_str_arg;
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -34,7 +34,6 @@
 #include "debug.h"
 #include "game/game_text.hpp"
 #include "network/network_content_gui.h"
-#include <stack>
 
 #include "table/strings.h"
 #include "table/control_codes.h"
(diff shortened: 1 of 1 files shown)


glx22:
> I am not sure why, but SCC_NEWGRF_PRINT_WORD_STRING_ID got triggered repeatedly in world generation.

It's because some industries have an integrated station, and that implies a call to get the station name.

src/strings.cpp
```diff
@@ -933,13 +927,13 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
-				str_stack.push(GetStringPtr(substr));
+				buff = FormatString(buff, GetStringPtr(substr), args, last, case_index, game_script, dry_run);
 				break;
 			}
 
 			case SCC_NEWGRF_PRINT_WORD_STRING_ID: {
 				StringID substr = args->GetInt32(SCC_NEWGRF_PRINT_WORD_STRING_ID);
-				str_stack.push(GetStringPtr(substr));
+				buff = FormatString(buff, GetStringPtr(substr), args, last, case_index, game_script, dry_run);
 				case_index = next_substr_case_index;
```
rubidium42: This seems wrong now as this `case_index` was what was used to format the string, and now it's using an old value. Other calls seems to be passing next_substr_case_index.

src/strings.cpp
```diff
@@ -818,15 +817,10 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 	WChar b = '\0';
 	uint next_substr_case_index = 0;
 	char *buf_start = buff;
-	std::stack<const char *, std::vector<const char *>> str_stack;
```
rubidium42, line 821: What about the dry-run FormatString calls that are now going to be added to the processing. You call it, rightfully, with dry_run... however that means that when we alread did the dry run after this has run for the main string, you will still call it for all the sub strings in the STRINL and PRINT_WORD_STRING_ID cases. Note that most other calls to FormatString in this function declare their own set of parameters to work on that do not have type information.
So the question is: are those allocations really that expensive that it warrants doing multiple dry runs for the same string?
It might be simplifying things code wise and seemingly allocation wise, but does that not potentially increase the computational complexity of this?

src/strings.cpp
```diff
@@ -818,15 +817,10 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
-	str_stack.push(str_arg);
-
+	const char* current_str_arg = str_arg;
```
rubidium42: Coding style: `const char* current_str_arg` -> `const char *current_str_arg`.
`char* foo, bar` gives you a `char *foo` and `char bar`, so the `*` belongs to the variable not the type which is why our coding style puts the `*` before the variable instead of after the type.

src/strings.cpp
rubidium42: With these changes `SCC_SET_CASE` changing `next_substr_case_index` while processing the substring does not influence any later strings, although in the past it did.
I do not know enough about the NewGRF text stack to say anything useful the effects of that. Are NewGRFs depending on this behaviour? Or is the behaviour maybe just plain wrong as it is now? I do not know. I do know it gets changed, and that makes me feel uncomfortable.

src/strings.cpp
```diff
@@ -818,15 +817,10 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
 	for (;;) {
-		while (!str_stack.empty() && (b = Utf8Consume(&str_stack.top())) == '\0') {
-			str_stack.pop();
-		}
-		if (str_stack.empty()) break;
-		const char *&str = str_stack.top();
+		if ((b = Utf8Consume(&current_str_arg)) == '\0') break;
```
rubidium42: `for (;;) { if (expr) break;` -> `while (!expr) {`? It seems weird to not set a condition for the loop and then in the first step define a simple if to break out of the loop.

src/strings.cpp
```diff
@@ -818,15 +817,10 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
+		const char *&str = current_str_arg;
```
rubidium42, line 822: This seems to be some messing about so the value in the stack gets updated properly. However, now there is only one value `current_str_arg`, so can't `const char *current_str_arg = str_arg` become `const char *str = str_arg`, remove this line and pass `str` instead of `current_str_arg` to `Utf8Consume`? That seems to simplify this complicated construct a lot.

src/strings.cpp
fsimonis: I left it like that to reduce the diff for the review, but I agree that this really looks silly.

src/strings.cpp
fsimonis: I was wondering about that as well, so thanks for bringing it up.

The use-case here looks actually very similar to the one of the `SCC_INDUSTRY_NAME`, which passes `next_substr_case_index` to the call to `FormatString` and then resets it to 0.
I'll change this.

https://github.com/OpenTTD/OpenTTD/blob/e99d594d269b38c4e144dd73ed48d9655531756a/src/strings.cpp#L1372-L1388

src/strings.cpp
fsimonis: `str` is used 26 times in the function, thus changing it will introduce a lot of noise in this PR.
I rather change this in a separate clean-up PR.

LordAro: Keep it as a separate commit, and I'd have no issues with it being in this PR

2TallTyler:
@fsimonis Are you still interested in finishing this PR?

fsimonis:
@2TallTyler let's close this PR.

If I get around to this again, I'll start from a current state.

Issue #9407, opened 2021-06-30
[Bug]: CmdFoundTown may cause multiplayer desyncs due to not initialising Town::stations_near


JGRennison:
### Version of OpenTTD

master

### Expected result

No multiplayer desyncs or incorrect caches

### Actual result

CmdFoundTown may be called by the player in game, in the scenario editor, or by a GS.

In the event that a town is founded in range of existing stations (i.e. where one or more new town house tiles are within the station's catchment area), the town's stations_near will be incorrectly empty, instead of containing the stations within range.
In the event that this occurs in multiplayer, players which join after the town founding event will not have an incorrect cache and this may result in a state divergence. Most notably this is used by StationFinder::GetStations.

### Steps to reproduce

1. Found a town near to an existing station
2. Note that Town::stations_near is incorrect, CheckCaches can be used to verify this

glx22:
The issue is https://github.com/OpenTTD/OpenTTD/blob/a7fabe497cd5b9f3b9e6b813b209d665d808d789/src/town_cmd.cpp#L2313-L2318 in `MakeTownHouse()` (added with https://github.com/OpenTTD/OpenTTD/commit/8b1b3fd0f91fd30d09d3acaa65bd9a7d23d966cc) being skipped because https://github.com/OpenTTD/OpenTTD/blob/a7fabe497cd5b9f3b9e6b813b209d665d808d789/src/town_cmd.cpp#L1993 in `CmdFoundTown()` (~~added with https://github.com/OpenTTD/OpenTTD/commit/ffec9b41e8d2efb539d5e24593249bc60baffd13~~ present since at least r1)

TrueBrain:
Tnx for the analysis @glx22 , that really helped as lot :) @JGRennison : I want for a slightly different solution than your commit, as I explain in #9526. Let me know if you foresee any issue with that.

PR #9418, opened 2021-07-08
Codechange: Don't explicitly unset _generating_world outside of genworld.cpp

glx22:
## Motivation / Problem
While trying to understand the cause for #9407, I noticed some 
```
...
_generating_world = true;
DoStuff();
_generating_world = false;
...
```
They are relatively safe (mainly used in GUI code), but I think it can (and should) be done safer.
The only place `_generating_world` can be explicitly set to `false` is in `_GenerateWorld()`, ie during the real generation.

## Description
I just followed the idea of https://github.com/OpenTTD/OpenTTD/commit/ffec9b41e8d2efb539d5e24593249bc60baffd13, and used `Backup` to replace all direct writes to `_generating_world` from outside `_GenerateWorld()`.
One place was already doing a proper backup, but without using `Backup`, so I updated it too.

## Checklist for review

### DIFF
--- a/src/industry_gui.cpp
+++ b/src/industry_gui.cpp
@@ -608,9 +608,9 @@ public:
 			ShowErrorMessage(STR_ERROR_CAN_T_GENERATE_INDUSTRIES, STR_ERROR_MUST_FOUND_TOWN_FIRST, WL_INFO);
 		} else {
 			extern void GenerateIndustries();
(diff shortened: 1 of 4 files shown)


LordAro:
Of course, the better solution would not be relying on `_generating_world` all over the place...

TrueBrain (approved):

PR #9421, opened 2021-07-08
Codechange: split settings.ini over several files

### LINKED ISSUE #9386: Building on MacOS fails with "bus error: 10"

TrueBrain:
Fixes #9386.

## Motivation / Problem

MacOS build currently fails because our setting tables are too big. But we have been willing to split the settings table up for a while now. So a good enough reason to make that happen.

## Description

```
This reduced the load on compilers, as currently for example MacOS
doesn't like the huge settings-tables.

Additionally, nobody can find settings, as the list is massive and
unordered. By splitting it, it becomes a little bit more sensible.
```

I split the settings based on the group they are in the settings. There is a lot more to do here, but I wanted to keep an easy-to-review PR. If you use a coloured-diff, you will see that things are only moved:

```bash
git fetch https://github.com/TrueBrain/OpenTTD settings-split-table
git show FETCH_HEAD --color-moved=zebra
```

`settings.ini` is removed completely, meaning patches will have a nice conflict on it, making it more clear for patches they should be moving their settings.

Future-PRs can address stuff like:
- Make flags default per settings-file and reduce the amount of things being repeated
- Force settings-files to read/write into a given struct (instead of repeating that every time)
- Move some settings around; there are some strange choices currently, for example I am not sure `pf.reverse_at_signal` should be a pathfinding setting. And there are more similar weird ones.
- Many more small things.

But, this PR is just to get that car moving. I will not be addressing these things above any time soon, so if you want to, feel free to go for it :)

## Checklist for review


src/table/settings/driver_settings.ini
```diff
@@ -0,0 +1,148 @@
+; This file is part of OpenTTD.
+; OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+; OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+; See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+;
+
+; Driver settings as stored in the main configuration file ("openttd.cfg").
```
rubidium42: The actual driver settings are in misc_settings, and these settings are in my opinion not really driver related. It's currently the sound equivalent of gui.

src/table/settings/pathfinding_settings.ini
```diff
@@ -0,0 +1,615 @@
...
+cat      = SC_EXPERT
+
+[SDT_VAR]
+var      = pf.wait_twoway_signal
+type     = SLE_UINT8
+def      = 41
+min      = 2
+max      = 255
+cat      = SC_EXPERT
+[SDT_VAR]
```
rubidium42, line 145: Newline gone missing?

src/table/settings/pathfinding_settings.ini
```diff
@@ -0,0 +1,615 @@
...
+
+[SDT_VAR]
+var      = pf.npf.maximum_go_to_depot_penalty
+type     = SLE_UINT
+from     = SLV_131
+def      = 20 * NPF_TILE_LENGTH
+min      = 0
+max      = 1000000
+cat      = SC_EXPERT
+[SDT_BOOL]
```
rubidium42, line 319: Missing newline?

src/table/settings/world_settings.ini
```diff
@@ -0,0 +1,544 @@
...
+var      = game_creation.se_flat_world_height
+type     = SLE_UINT8
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
+def      = 1
+min      = 0
+max      = 15
+str      = STR_CONFIG_SETTING_SE_FLAT_WORLD_HEIGHT
+strval   = STR_JUST_COMMA
+cat      = SC_BASIC
+[SDT_VAR]
```
rubidium42, line 239: Missing newline?

src/table/settings/driver_settings.ini
TrueBrain: I am open for other suggestions. But neither `music`, nor `sound` covers it, and making 2 tables out of this feels wrong. That is why I came to `driver`, but if you have a better idea, I am all ears.

Edit: one can also move, not in this PR, the driver-settings from misc to this file, as I kinda want `misc` to die off anyway. But that is not for this PR. But maybe with those settings moved you see it as less of a problem? I dunno .. naming these kind of things is hard ;)

Edit 2: some more thinking later, if we in a later PR move things like `fullscreen` and `allow_hdpi` and `support_8bpp` etc to this file too, I think it is a fine name for all these driver settings. But that is just my opinion here :D I am still open for better suggestions ;)

Xaroth (dismissed):

src/table/settings/driver_settings.ini
michicc: It really depends if you want `driver` to be taken like the rest of the source code uses it or not. Right now, all settings are UI settings that don't relate to anything the source otherwise calls driver.

Our terminology is probably not very common though :)

TrueBrain: @michicc mentioned on IRC that the 20 year dated name for sound/music is `multimedia`, and I like that name. At least solved this problem; we can always revisit where things go etc at a later date :) This is just a start after all ;)

glx22 (approved):

PR #9423, opened 2021-07-10
Fix 11ab3c4: [NewGRF] Overflow when determining cargo mask for string code 9A 1E

rubidium42:
## Motivation / Problem

From coverity's scan:
```
1453026 Bad bit shift operation
The operation may have an undefined behavior or yield an unexpected result.

In RemapNewGRFStringControlCode(unsigned int, char *, char **, char const**, long long *, unsigned int, bool): 
A bit shift operation has a shift amount which is too large or has a negative value.
```

## Description

Make the 1 that is shifted 64 bits, so shifting with a cargo type > 32 will not result in a 0 value for the cargo mask to look up.

## Limitations

I haven't tested it as I have no idea which NewGRF uses this string code. Making a NewGRF for this is a bit out of my league.

## Checklist for review


TrueBrain (approved):
Sure.

LordAro (approved):
Sure.

PR #9428, opened 2021-07-10
Feature: raise the maximum NewGRF limit to 255

TrueBrain:
## Motivation / Problem

Every limit is one limit too many! WE WANT MORE NEWGRFS! MOAR! MMOOAAARRRR!

## Description

As the Game Info now travels via TCP, we can increase the MTU limit on both sides to allow more NewGRFs. So .. I did exactly that.

I tested it with up to 209 NewGRFs .. that was all I had on disk :P

image

Picture to proof it!

## Limitations

- 255 isn't enough. But .. a problem for another day ;)

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -9853,6 +9853,8 @@ void LoadNewGRF(uint load_index, uint num_baseset)
 				num_non_static++;
 			}
 
+			num_grfs++;
+
 			LoadNewGRFFile(c, stage, subdir, false);
 			if (stage == GLS_RESERVE) {
 				SetBit(c->flags, GCF_RESERVED);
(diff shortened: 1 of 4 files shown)


James103:
Why not just download at least 46 more NewGRFs from BaNaNaS and load them in addition to what you have already?

src/newgrf.cpp
```diff
@@ -80,7 +80,7 @@ static uint32 _ttdpatch_flags[8];
 GRFLoadedFeatures _loaded_newgrf_features;
 
 static const uint MAX_SPRITEGROUP = UINT8_MAX; ///< Maximum GRF-local ID for a spritegroup.
-static const uint MAX_GRF_COUNT = 128; ///< Maximum number of NewGRF files that could be loaded.
+static const uint MAX_GRF_COUNT = 255; ///< Maximum number of NewGRF files that could be loaded.
```
rubidium42: The base set loads two NewGRFs, so practically you can now load only 253 NewGRFs in a game. Furthermore, it won't allow static NewGRFs to be loaded (for fonts or so) when the server has 253 or more NewGRFs set.

rubidium42 (approved):
Next rabbit hole please ;)

PR #9430, opened 2021-07-11
Fix: ensure no more than the allowed number of NewGRFs are loaded from the configuration

rubidium42:
## Motivation / Problem

When loading OpenTTD with a configuration that has more than NETWORK_MAX_GRF_COUNT NewGRFs configured bad things might happen.

The UI limits the user from entering more than that amount, but loading the configuration does not perform such sanity checks.

With older versions, including 1.11.2, this can cause an assertion or when assertions are disabled a buffer overrun when advertising to the master server.

With versions since the packet rewrite this causes only the assertion as the buffer will automatically resize if needed. However, the packets might be interpreted incorrectly by others, especially the UDP packets.

With any, but more likely for #9428, this could lead to an overflow of the number of NewGRFs in the packet, meaning the clients see 0 NewGRFs when the server actually has 256.

## Description

The issue is solved by limiting the number of non-static NewGRFs that are loaded from the configuration file.

## Limitations

Saving the configuration will now remove any NewGRFs over the limit from the configuration, so going back and forth between stable and post #9428 might loose NewGRFs in the configuration if you have set more than 62.

This change doesn't magically fix the problem of loading valid post #9428 settings files in 1.11. That requires this PR to be backported.

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -956,6 +956,7 @@ static GRFConfig *GRFLoadConfig(IniFile &ini, const char *grpname, bool is_stati
 
 	if (group == nullptr) return nullptr;
 
+	uint num_grfs = 0;
 	for (item = group->item; item != nullptr; item = item->next) {
 		GRFConfig *c = nullptr;
 
@@ -1030,8 +1031,14 @@ static GRFConfig *GRFLoadConfig(IniFile &ini, const char *grpname, bool is_stati
 			continue;
 		}
 
-		/* Mark file as static to avoid saving in savegame. */
-		if (is_static) SetBit(c->flags, GCF_STATIC);
+		if (is_static) {
+			/* Mark file as static to avoid saving in savegame. */
+			SetBit(c->flags, GCF_STATIC);
+		} else if (++num_grfs > NETWORK_MAX_GRF_COUNT) {
+			/* Check we will not load more non-static NewGRFs than allowed. This could trigger issues for game servers. */
+			ShowErrorMessage(STR_CONFIG_ERROR, STR_NEWGRF_ERROR_TOO_MANY_NEWGRFS_LOADED, WL_CRITICAL);
+			break;
+		}
 
 		/* Add item to list */
 		*curr = c;


src/newgrf_config.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include "safeguards.h"
 
+static_assert(MAX_GRF_COUNT >= NETWORK_MAX_GRF_COUNT + NUM_BASESET_NEWGRFS);
```
TrueBrain: These asserts I always consider "annoying future-us" :) Mainly, as I think it is not needed.

The `MAX_GRF_COUNT` is flexible, and can be any value. So we can just use the other constants to set a valid value for it. I would define it something like:

`static const uint MAX_GRF_COUNT =NETWORK_MAX_GRF_COUNT + NUM_BASESET_NEWGRFS + MAX_STATIC_GRF_COUNT;`

Where the last const is something like 16 or what-ever .. just some safe value. That means that nobody ever again has to worry about this value :)

TrueBrain (commented):
> The second issue is solved by limiting the number of NewGRFs that are loaded from the configuration using the now accessible constants.

Although strictly true, you are overselling the solution here. Assuming it won't take months to approve my PR to increase the limit, this only "solves" the issue for 1 or 2 commits :P So we only solve it if we in the future ever increase the limit again, I guess :) And of course for people manually adding more entries in the ini-files.

It still leaves the issue that if you start a server in 1.11 with a configuration from 1.12 you could crash the game. But we have had this more often, like with a zoom setting from 1.10 to 1.11. Not sure those problems are solveable without releasing a 1.11.3 .. and I am not sure that is useful :D

src/newgrf_config.cpp
rubidium42: Except that introducing a `MAX_STATIC_GRF_COUNT` implies that you should also check for that limit, which is currently not done. Also 16 might be too limiting as currently it is effectively 126.

TrueBrain: I was using just any random value :) If you want it bigger, go for it. No need to nitpick like that ;)

And yeah, checking on it would be nice .. but if you don't, it isn't a regression, as we currently also don't check for it. It just explains why `NETWORK_MAX_GRF_COUNT` is 62, and `MAX_GRF_COUNT` is 128. It is just making more visible where those numbers are deduced from.

Edit: owh, here is an idea, if we don't want to add a check but want the name to match what it does, call it `AT_LEAST_THIS_MANY_STATIC_GRF_COUNT` or something ;) :D

rubidium42:
> > The second issue is solved by limiting the number of NewGRFs that are loaded from the configuration using the now accessible constants.
> 
> Although strictly true, you are overselling the solution here. Assuming it won't take months to approve my PR to increase the limit, this only "solves" the issue for 1 or 2 commits :P So we only solve it if we in the future ever increase the limit again, I guess :) And of course for people manually adding more entries in the ini-files.

The issue remains after merging #9428, and there it is actually more silent. A server with 256 NewGRFs will advertise 0, while sending 256 identifiers. Clients will join the game without the appropriate NewGRFs and yuck.
Since the configuration is very human-editable, it should be considered unsafe and made safe for consumption by OpenTTD. That is exactly what this PR does.

It is true though that with #9428 it becomes easier for an end user to trigger this issue in older versions of OpenTTD, but #9428 was only the cause that triggered me to think about it.

TrueBrain (approved):

src/newgrf_config.cpp
rubidium42: Better idea was to just abolish that completely, i.e. #9431.

PR #9432, opened 2021-07-11
Feature: join servers based on their invite code

TrueBrain:
!!WARNING!!
This is currently only working with staging. See `Testing` how to test this PR!

## Motivation / Problem

Currently joining a server requires you to have either of these two:

- Know the hostname + port of the server.
- Find the server in the (huge) list of public servers.

This is often annoying if you want your friend to join the same server.

Additionally, this allows for several other future additions, that without a system like this would be impossible:
- "invite only" games: games that are open but not listed. Via an invite code you can join such games.
- STUN / TURN support, where you no longer need to port-forward in order to play.

## Description

There are now two type of connection-strings:
- server address. This connection-string is of the form `hostname:port`.
- invite code. This connection-string always starts with a `+`.

The `+` of the connection-string to indicate invite codes is part of the network protocol, and the Game Coordinator can either give a server address or an invite code per server. This is done for backwards compatibility, allowing you to still see servers from 1.11 and before. These are always server address based, ofc.

As using an invite code means you do not know the `hostname:port` of the server, several packets are needed to talk to the Game Coordinator to discover this information. The idea is very simple:

1) you tell the GC you want to connect to an indicate invite code
2) the GC tells you how to reach that server, currently the only supported thing is a "direct-ip", where it gives you the `hostname:port` of the server. (in the future, STUN and TURN will be added)
3) you connect to that address.
4) you tell the GC the outcome: could you connect or not.
5) the GC might offer an alternative way of connecting to the server.
6) if all methods are exhausted, the GC tells you there is no way for you to connect to the server.

## Testing

This is deployed on staging as of now, and you can test that by starting OpenTTD with:

`OTTD_COORDINATOR_CS="coordinator.openttd.org:4976" ./openttd`

It will NOT (!) work on production, so starting this PR unmodified will result in a non-functional multiplayer experience!

## Checklist for review

### DIFF
--- a/src/network/core/address.cpp
+++ b/src/network/core/address.cpp
@@ -414,17 +414,22 @@ void NetworkAddress::Listen(int socktype, SocketList *sockets)
 }
 
 /**
- * Convert a string containing either "hostname" or "hostname:ip" to a
- * ServerAddress, where the string can be postfixed with "#company" to
+ * Convert a string containing either "hostname", "hostname:port" or invite code
+ * to a ServerAddress, where the string can be postfixed with "#company" to
  * indicate the requested company.
  *
  * @param connection_string The string to parse.
  * @param default_port The default port to set port to if not in connection_string.
- * @param company Pointer to the company variable to set iff indicted.
+ * @param company Pointer to the company variable to set iff indicated.
  * @return A valid ServerAddress of the parsed information.
  */
 /* static */ ServerAddress ServerAddress::Parse(const std::string &connection_string, uint16 default_port, CompanyID *company_id)
 {
+	if (StrStartsWith(connection_string, "+")) {
+		std::string_view invite_code = ParseCompanyFromConnectionString(connection_string, company_id);
+		return ServerAddress(SERVER_ADDRESS_INVITE_CODE, std::string(invite_code));
+	}
+
 	uint16 port = default_port;
 	std::string_view ip = ParseFullConnectionString(connection_string, port, company_id);
 	return ServerAddress(SERVER_ADDRESS_DIRECT, std::string(ip) + ":" + std::to_string(port));
--- a/src/network/core/address.cpp
+++ b/src/network/core/address.cpp
@@ -10,6 +10,7 @@
 #include "../../stdafx.h"
 
 #include "address.h"
+#include "../network_internal.h"
 #include "../../debug.h"
 
 #include "../../safeguards.h"
@@ -410,4 +411,21 @@ void NetworkAddress::Listen(int socktype, SocketList *sockets)
 	socklen_t addr_len = sizeof(addr);
 	getpeername(sock, (sockaddr *)&addr, &addr_len);
 	return NetworkAddress(addr, addr_len).GetAddressAsString();
+}
+
+/**
+ * Convert a string containing either "hostname" or "hostname:ip" to a
+ * ServerAddress, where the string can be postfixed with "#company" to
+ * indicate the requested company.
+ *
+ * @param connection_string The string to parse.
+ * @param default_port The default port to set port to if not in connection_string.
+ * @param company Pointer to the company variable to set iff indicted.
+ * @return A valid ServerAddress of the parsed information.
+ */
+/* static */ ServerAddress ServerAddress::Parse(const std::string &connection_string, uint16 default_port, CompanyID *company_id)
+{
+	uint16 port = default_port;
+	std::string_view ip = ParseFullConnectionString(connection_string, port, company_id);
+	return ServerAddress(SERVER_ADDRESS_DIRECT, std::string(ip) + ":" + std::to_string(port));
 }
--- a/src/network/core/tcp.h
+++ b/src/network/core/tcp.h
@@ -82,10 +82,12 @@ private:
 		RESOLVING,  ///< The hostname is being resolved (threaded).
 		FAILURE,    ///< Resolving failed.
 		CONNECTING, ///< We are currently connecting.
+		CONNECTED,  ///< The connection is established.
 	};
 
 	std::thread resolve_thread;                         ///< Thread used during resolving.
 	std::atomic<Status> status = Status::INIT;          ///< The current status of the connecter.
+	std::atomic<bool> killed = false;                   ///< Whether this connecter is marked as killed.
 
 	addrinfo *ai = nullptr;                             ///< getaddrinfo() allocated linked-list of resolved addresses.
 	std::vector<addrinfo *> addresses;                  ///< Addresses we can connect to.
@@ -101,7 +103,7 @@ private:
 	void OnResolved(addrinfo *ai);
 	bool TryNextAddress();
 	void Connect(addrinfo *address);
-	bool CheckActivity();
+	virtual bool CheckActivity();
 
 	/* We do not want any other derived classes from this class being able to
 	 * access these private members, but it is okay for TCPServerConnecter. */
@@ -125,15 +127,25 @@ public:
 	 */
 	virtual void OnFailure() {}
 
+	void Kill();
+
 	static void CheckCallbacks();
 	static void KillAll();
 };
 
 class TCPServerConnecter : public TCPConnecter {
+private:
+	SOCKET socket = INVALID_SOCKET; ///< The socket when a connection is established.
+
+	bool CheckActivity() override;
+
 public:
 	ServerAddress server_address; ///< Address we are connecting to.
 
 	TCPServerConnecter(const std::string &connection_string, uint16 default_port);
+
+	void SetConnected(SOCKET sock);
+	void SetFailure();
 };
 
 #endif /* NETWORK_CORE_TCP_H */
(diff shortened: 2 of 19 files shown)


TrueBrain:
Okay, I deployed a compatible Game Coordinator on staging; from what I can see it is doing what it should be doing, but it could of course use a bit of testing :)

Tnx to #9429 this became a lot easier, you just have to set the env-variable `OTTD_COORDINATOR_CS` to `coordinator.openttd.org:4976`.

glx22:
Invite codes works for me with IPv6 or IPv4 (not both at the same time of course, current limitation of GC ;) )

src/network/core/address.cpp
```diff
@@ -411,3 +412,25 @@ void NetworkAddress::Listen(int socktype, SocketList *sockets)
...
+ * indicate the requested company.
+ *
+ * @param connection_string The string to parse.
+ * @param default_port The default port to set port to if not in connection_string.
+ * @param company Pointer to the company variable to set iff indicted.
+ * @return A valid ServerAddress of the parsed information.
+ */
+/* static */ ServerAddress ServerAddress::Parse(const std::string &connection_string, uint16 default_port, CompanyID *company_id)
+{
+	if (!connection_string.empty() && connection_string[0] == '+') {
```
rubidium42: `if (StrStartsWith(connection_string, "+")) {`?

src/network/core/address.cpp
```diff
@@ -411,3 +412,25 @@ void NetworkAddress::Listen(int socktype, SocketList *sockets)
...
 }
+
+/**
+ * Convert a string containing either "hostname", "hostname:port" or invite code
+ * to a ServerAddress, where the string can be postfixed with "#company" to
+ * indicate the requested company.
+ *
+ * @param connection_string The string to parse.
+ * @param default_port The default port to set port to if not in connection_string.
+ * @param company Pointer to the company variable to set iff indicted.
```
rubidium42: s/indicted/indicated/ ? I highly doubt 'indicted' is the word you meant to use as it is only used in law settings.

src/network/core/tcp_coordinator.h
```diff
@@ -85,6 +95,8 @@ class NetworkCoordinatorSocketHandler : public NetworkTCPSocketHandler {
 	/**
 	 * Game Coordinator acknowledges the registration.
 	 *
+	 *  string  Invite code that can be used to join this server.
+	 *  string  Secret that belongs to the invite code (only needed if reusing the invite code on next CLIENT_REGISTER).
```
rubidium42: s/CLIENT/SERVER/?

src/network/network_coordinator.cpp
```diff
@@ -229,6 +343,118 @@ void ClientNetworkCoordinatorSocketHandler::GetListing()
...
 }
 
+/**
+ * Join a server based on an invite code.
+ * @param invite_code The invite code of the server to connect to.
+ * @param connecter The connecter of the request.
+ */
+void ClientNetworkCoordinatorSocketHandler::ConnectToServer(const std::string &invite_code, TCPServerConnecter *connecter)
+{
+	assert(invite_code[0] == '+');
```
rubidium42: `assert(StrStartsWith(invite_code, "+"));`?
Mostly for the even rarer case of the invite_code being empty when getting here.

src/network/network_coordinator.cpp
```diff
@@ -229,6 +343,118 @@ void ClientNetworkCoordinatorSocketHandler::GetListing()
...
+
+	if (this->connecter_pre.find(invite_code) != this->connecter_pre.end()) {
+		/* If someone is hammering the refresh key, he can set out two
```
rubidium42: We eradicated the (s)he from the code comments, also is `set out` what you meant? 
So something like `..., one can sent out ...`?

src/network/network_coordinator.cpp
```diff
@@ -149,6 +202,58 @@ bool ClientNetworkCoordinatorSocketHandler::Receive_GC_LISTING(Packet *p)
...
+
+	/* Find the connecter based on the invite code. */
+	auto connecter_it = this->connecter_pre.find(invite_code);
+	if (connecter_it == this->connecter_pre.end()) {
+		this->CloseConnection();
+		return false;
+	}
+
+	/* Now store it based on the token. */
+	this->connecter[token] = connecter_it->second;
```
rubidium42, line 218: Can't the tokens and invite_codes be in the same map? After all, the invite codes always start with `+` so with little effort one could prevent tokens to start with `+` and they could use the same space. After all, in both cases its just a mapping of a string to the connecter.

src/network/core/tcp_coordinator.h
```diff
@@ -23,15 +23,22 @@
...
-	PACKET_COORDINATOR_END,             ///< Must ALWAYS be on the end of this list!! (period).
+	PACKET_COORDINATOR_GC_ERROR,              ///< Game Coordinator indicates there was an error.
+	PACKET_COORDINATOR_SERVER_REGISTER,       ///< Server registration.
+	PACKET_COORDINATOR_GC_REGISTER_ACK,       ///< Game Coordinator accepts the registration.
+	PACKET_COORDINATOR_SERVER_UPDATE,         ///< Server sends an set intervals an update of the server.
+	PACKET_COORDINATOR_CLIENT_LISTING,        ///< Client is requesting a listing of all public servers.
+	PACKET_COORDINATOR_GC_LISTING,            ///< Game Coordinator returns a listing of all public servers.
+	PACKET_COORDINATOR_CLIENT_CONNECT,        ///< Client wants to connect to a server based on an invite code.
+	PACKET_COORDINATOR_GC_CONNECTING,         ///< Game Coordinator informs the client of the token assigned to the connection attempt.
+	PACKET_COORDINATOR_SERCLI_CONNECT_FAILED, ///< Client/server tells the Game Coordinator the current connection attempt failed.
```
rubidium42, line 37: Won't it be cleaner/clearer to just have a separate packet for a server failure? Maybe you want to sent other information in such cases? The current code asserts on not being a server before sending this packet.

src/network/core/tcp.h
```diff
@@ -82,10 +82,12 @@ class TCPConnecter {
 		RESOLVING,  ///< The hostname is being resolved (threaded).
 		FAILURE,    ///< Resolving failed.
 		CONNECTING, ///< We are currently connecting.
+		HAS_RESULT, ///< Our connecter is finished and "result_socket" contains the result (TCPServerConnecter only).
```
rubidium42: `HAS_RESULT` is a quite non-descript status name. The name itself doesn't give much of a clue.

Maybe GC_FINISHED? As in: "The Game Coordinator has finished its connection attempts, and set the socket to use which could be INVALID_SOCKET in case of a failure."
Potentially prefixing the others (except INIT) with DC_... to make it clearer those are used for Direct (Connect).

Alternatively add a CONNECTED state instead, with a SetFailure() and SetConnected(SOCKET socket) so the code with SetResult(INVALID_SOCKET) can become SetFailure().
In CheckActivity for TCPServerConnector you can then read the status; with CONNECTED -> OnSuccess, FAILURE -> OnFailure, rest: return false. I would even set the state to CONNECTING after calling ConnectToServer, so when debugging the status shows CONNECTING instead of INIT.

src/network/network_coordinator.cpp
TrueBrain: Is there a benefit of having them in the same map?

The reason I used two, is because it makes debugging slightly easier; if something is stuck in `connecter_pre` it means something very different then when it is stuck in `connecter`. It is more about the clarity of what the bookkeeping is doing.

So, can it? Sure. But do we want to?

src/network/core/tcp_coordinator.h
TrueBrain: This name is picked not so much for this PR, but for the future STUN PR. I didn't want to rename the packet, so I already sneaked in how STUN is going to work :)

Basically, this packet is send back as a response to the GC asking you to connect to somewhere. For direct-connect, what this PR only support, it is only the client that can do this. But with STUN, the GC sends both the server and client a "please connect here" request. They both answer with a "failure" when they cannot. They are designed to be identical in payload, as for STUN it doesn't matter who is server and who is client. As such, they will always be identical. And so I put them in a single packet.

But I agree in this PR it looks a bit weird, as it is never used from the server point of view. But renaming a packet is also a bit meh in the long run. A bit of a balance here, I guess. But I rather keep this name so it makes sense after the STUN PR. Unless you really want me to, I can rename it `CLIENT` here to rename it back into `SERCLI` with the STUN PR :)

Other suggestions are of course welcome, but having this in two packets will be more confusing imo :)

src/network/network_coordinator.cpp
rubidium42: The main benefit would be fewer maps to maintain, though that benefit might be outweighed by the debugging benefit of having several maps. So it's somewhat of an open question where I'm agnostic to the answer.

TrueBrain: In that case, given the key for both maps really indicate something completely different, I rather keep two maps :) Pretty sure that will help future-us :D

rubidium42 (approved):

PR #9434, opened 2021-07-11
Add: allow setting your server visibility to "invite-only"

TrueBrain:
## Motivation / Problem

As you can join based on invite codes now, it allows for servers to be registered but not public. In other words: invite-only. If you know the invite code you can join, but otherwise it won't be visible in the server-listing.

This might not seem all that useful now, as you could as well give someone your IP+port. But invite-codes are shorter, and where IPs can change, invite-codes do not.

And of course, if STUN lands, it becomes even more clear why this is useful :)

## Description

```
In this mode you do register to the Game Coordinator, but your
server will not show up in the public server listing. You can give
your friends the invite code of the server with which they can
join.
```

## Checklist for review

### DIFF
--- a/src/table/settings/network_settings.ini
+++ b/src/table/settings/network_settings.ini
@@ -9,14 +9,18 @@
 [pre-amble]
 static void UpdateClientConfigValues();
 
+static std::initializer_list<const char*> _server_game_type{"local", "public", "invite-only"};
+
 static const SettingVariant _network_settings_table[] = {
 [post-amble]
 };
 [templates]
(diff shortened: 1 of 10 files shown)


src/table/settings/network_settings.ini
```diff
@@ -159,10 +163,16 @@ flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_NETWORK_ONLY
 def      = true
 cat      = SC_EXPERT
 
-[SDTC_BOOL]
-var      = network.server_advertise
```
rubidium42, line 163: docs/multiplayer.md references this setting, so that requires some updating.

rubidium42 (approved):

PR #9435, opened 2021-07-11
Fix #9433: Use world generation seed from openttd.cfg if provided

### LINKED ISSUE #9433: [Bug]: World generation seed from openttd.cfg is ignored
### Version of OpenTTD

master @ commit f4dd2d8, Windows 10 x64

### Expected result

Setting the parameter "generation_seed" in the "game_creation" section of openttd.cfg should lead to OpenTTD using the provided seed for game creation.

### Actual result

World generation ignores the "generation_seed" parameter. 

### Steps to reproduce

1. Open openttd.cfg.
2. Set "generation_seed" to any uint32 value, e. g. 42
3. Start OpenTTD
4. Start a new game
5. Check world generation seed with the "get_seed" command in the in-game console. It will most likely not be 42.
6. Exit OpenTTD

Taschi120:
## Motivation / Problem
Fix #9433 

## Description

Code which overwrites user settings removed

## Limitations

none that I'm aware of

## Checklist for review

looks fine to me

### DIFF
--- a/src/genworld.cpp
+++ b/src/genworld.cpp
@@ -93,7 +93,9 @@ static void _GenerateWorld()
 		_generating_world = true;
 		if (_network_dedicated) Debug(net, 3, "Generating map, please wait...");
 		/* Set the Random() seed to generation_seed so we produce the same map with the same seed */
-		if (_settings_game.game_creation.generation_seed == GENERATE_NEW_SEED) _settings_game.game_creation.generation_seed = _settings_newgame.game_creation.generation_seed = InteractiveRandom();
+		if (_settings_game.game_creation.generation_seed == GENERATE_NEW_SEED) {
+			_settings_game.game_creation.generation_seed = InteractiveRandom();
+		}
 		_random.SetSeed(_settings_game.game_creation.generation_seed);
 		SetGeneratingWorldProgress(GWP_MAP_INIT, 2);
 		SetObjectToPlace(SPR_CURSOR_ZZZ, PAL_NONE, HT_NONE, WC_MAIN_WINDOW, 0);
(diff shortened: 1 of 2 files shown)


TrueBrain:
There are a few more places that overwrite `_settings_newgame.game_creation.generation_seed`, most related to Network games. They might need some love too, as otherwise starting a network game results in a similar issue ;)

Taschi120:
Oh, yeah, you're entirely right.

TrueBrain:
The more I think about this, the more I think `generation_seed` shouldn't be stored in `openttd.cfg`. It makes very little sense to store it there to me. Let me explain my thinking:

You can use `-G` on startup to force a seed. This I can fully understand, as that is something people sometimes want to do.

Other than that .. there are very few usecases where you want to force a seed. I can understand that you want to enter a seed number when you start a game .. but other than that .. it should always be a random value.

Why storing it in `openttd.cfg` is weird to me: if you ever by accident put a number in there that is not `GENERATE_NEW_SEED`, and looking at this PR many people already have that now in their `openttd.cfg`, you will always get the same map over and over and over and over and over again, without clear feedback why this is the case.

So the more I think about it, the less sense it all makes.

Either way, that does result in a real problem with this PR: as many people have non-`GENERATE_NEW_SEED` in their `openttd.cfg`, this PR would make my example a reality: maps will be very static all of a sudden for a lot of people :)

Maybe we should approach this problem differently, and put the seed in the "New Game" menu as many games do. And a "randomize" next to it. And always randomize it on opening the GUI. That way people who want a fixed value can fill it in there.
And we should stop storing it in `openttd.cfg`  in my opinion :)

But that is just one opinion; I would love to hear others :)

Taschi120:
Yes, I absolutely forgot to think about what happens to existing openttd.cfg files so I will retract this PR for now.

Personally I would also like a UI element for manually setting a seed. For implementing that, the "troublemakers" from #9433 will still need to be removed.

PR #9443, opened 2021-07-15
Feature: allow the use of STUN to connect client and server together 

TrueBrain:
## Motivation / Problem

STUN is a method of connecting a client with a server without either of them using port-forwarding while either or both can be behind one or several NATs. In other words: it is so much easier to connect to a server.

This PR (finally) allows this method of connecting. The Game Coordinator is the broker between the two to make this a reality.

## Description

```
This method doesn't require port-forwarding to be used, and works for
most common NAT routers in home setups. But, for sure it doesn't work
for all setups, and not everyone will be able to use this.
```

## Checklist for review


src/network/core/tcp_connect.cpp
```diff
@@ -24,10 +24,14 @@
 static std::vector<TCPConnecter *> _tcp_connecters;
 
 /**
- * Create a new connecter for the given address
- * @param connection_string the address to connect to
+ * Create a new connecter for the given address.
+ * @param connection_string the address to connect to.
```
rubidium42: Capitalize the `the`.

src/network/core/tcp_coordinator.h
```diff
@@ -215,6 +219,50 @@ class NetworkCoordinatorSocketHandler : public NetworkTCPSocketHandler {
 	 */
 	virtual bool Receive_GC_DIRECT_CONNECT(Packet *p);
 
+	/**
+	 * Game Coordinator requests the client/server to do a STUN request to the
+	 * STUN server. Important is to remember the local port these STUN requests
+	 * are send from, as this will be needed for later conenctions too.
```
rubidium42: `sent`

src/network/core/tcp_coordinator.h
```diff
@@ -215,6 +219,50 @@ class NetworkCoordinatorSocketHandler : public NetworkTCPSocketHandler {
...
+	 */
+	virtual bool Receive_GC_STUN_REQUEST(Packet *p);
+
+	/**
+	 * Client/server informs the Game Coordinator the result of a STUN request.
+	 *
+	 *  uint8   Game Coordinator protocol version.
+	 *  string  Token to track the current connect request.
+	 *  uint8   Interface number, as given during STUN request.
+	 *  uint8   Connection result (0 = failed, 1 = success).
```
rubidium42: Given two values, shouldn't it be a `bool` denoting `Connection successful`

src/network/network_coordinator.cpp
```diff
@@ -263,6 +304,49 @@ bool ClientNetworkCoordinatorSocketHandler::Receive_GC_DIRECT_CONNECT(Packet *p)
...
+	auto family_it = stun_it->second.find(family);
+	if (family_it == stun_it->second.end()) return true;
+
+	/* Ensure all other pending connection attempts are killed. */
+	if (this->game_connecter != nullptr) {
+		this->game_connecter->Kill();
+		this->game_connecter = nullptr;
+	}
+
+	/* We now mark the connection as close, but we do not really close the
```
rubidium42: First `close` -> `closed`?

src/network/network_coordinator.h
```diff
@@ -42,6 +49,7 @@ class ClientNetworkCoordinatorSocketHandler : public NetworkCoordinatorSocketHan
 	std::chrono::steady_clock::time_point next_update; ///< When to send the next update (if server and public).
 	std::map<std::string, TCPServerConnecter *> connecter; ///< Based on tokens, the current connecters that are pending.
 	std::map<std::string, TCPServerConnecter *> connecter_pre; ///< Based on invite codes, the current connecters that are pending.
+	std::map<std::string, std::map<int, std::unique_ptr<ClientNetworkStunSocketHandler>>> stun_handlers; ///< All pending STUN handlers.
```
rubidium42: Maybe document this is a token -> family -> SocketHandler mapping?

src/network/network_coordinator.h
```diff
@@ -33,6 +34,12 @@
  *        - Send the client a GC_CONNECT with the peer address.
  *        - a) Client connects, client sends CLIENT_CONNECTED to Game Coordinator.
  *        - b) Client connect fails, client sends CLIENT_CONNECT_FAILED to Game Coordinator.
+ *    2) STUN?
```
rubidium42, line 39: Maybe a small explanation of what STUN is (supposed) to be, not necessarily here though. Something along the lines (if explained it correctly) would probably suffice:
Both the game server and game client connect to the turn server, after which the turn server tells the game client to connect to the game server's public IP and port that was used to connect to the turn server. This trickery allows two computers that are both behind a NAT to connect to each other without an intermediate server except for setting up the connection.

src/network/network_coordinator.cpp
```diff
@@ -263,6 +304,49 @@ bool ClientNetworkCoordinatorSocketHandler::Receive_GC_DIRECT_CONNECT(Packet *p)
...
+	this->stun_handlers[token][AF_INET] = ClientNetworkStunSocketHandler::Stun(token, AF_INET);
+	return true;
+}
+
+bool ClientNetworkCoordinatorSocketHandler::Receive_GC_STUN_CONNECT(Packet *p)
+{
+	std::string token = p->Recv_string(NETWORK_TOKEN_LENGTH);
+	uint8 tracking_number = p->Recv_uint8();
+	uint8 family = p->Recv_uint8();
+	std::string host = p->Recv_string(NETWORK_HOSTNAME_PORT_LENGTH);
```
rubidium42, line 321: https://en.wikipedia.org/wiki/STUN mentions that the IP address and port should be obfuscated (in their example by a XOR) so deep inspection stuff does (likely) not mess with the data. Might that be a wise strategy to employ here? Maybe not XOR but some other method? Given everything's a character, maybe add 10 to every character or so?

TrueBrain: lol ... I wouldn't even know why a NAT would waste CPU cycles on that, but sure.

Most changes are a bit tricky, as it needs to remain a string .. so I will just reverse the hostname. That should be sufficient I hope :D

TrueBrain: ah, no, this is not needed at all. Pfew, saves me from weird code.

Normally what STUN does, is return to the client itself what his public IP:port is for that socket. For OpenTTD I opt'd not to do that, as that would mean we need another round-trip from the client to the Game Coordinator to tell about that public IP:port. I found that wasteful and potentially allows OpenTTD to be used for DoS attacks. So, instead, the STUN server tells the GC about the public IP:port, and this is never told back to the client.

The client does get back the IP:port of its peer, but no NAT will rewrite that. So there is no need to mitigate this in our setup.

TrueBrain:
Deployed on staging. Ready for testing :)

docs/game_coordinator.md
```diff
@@ -0,0 +1,62 @@
...
+assigns the server an unique code, called an `invite code`.
+
+When a client wants to join a server, it asks the Game Coordinator to help
+with this by giving it the `invite code` of the server. The Game Coordinator
+will, in this order, attempt several ways to connect the client and server
+together:
+
+## 1) Via direct IPv4 / IPv6
+
+Upon registration, the Game Coordinator probed the server to see if based on
```
rubidium42: s/probed/probes/

docs/game_coordinator.md
```diff
@@ -0,0 +1,62 @@
...
+his public IPv4 and/or IPv6, it allows other people from the Internet to
```
rubidium42: s/his/their/

docs/game_coordinator.md
```diff
@@ -0,0 +1,62 @@
...
+connect to the game port. If successful, a client is first asked to try this.
```
rubidium42: The whole sentence is a bit (over?) complicated. Maybe:
Just after registration the Game Coordinator probes the server, using its public IP address, to see if it allows players to connect directly. If successful, clients are asked to try this method first.

docs/game_coordinator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- If the server is IPv4 only, the client is only asked to connect via IPv4.
+- If the server is IPv6 only, the client is only asked to connect via IPv6.
+- If the server is both IPv4 and IPv6, the client is asked to connect to to
+  one of those first. If that fails, it is asked to connect to the other.
+  Wether it tries IPv4 or IPv6 first, strongly depends on several network
```
rubidium42: s/Wether/Whether/

rubidium42 (approved):

PR #9447, opened 2021-07-17
Feature: allow the use of TURN to connect client and server together

TrueBrain:
Note: this is currently only deployed on staging. To test it out:
- Host your server on a port that you do NOT forward (you do not want to have a Public server).
- We "disable" STUN by using a server that doesn't exist ("localhost"). This will give a connection error, that is fully expected.
- Start the game like: `OTTD_COORDINATOR_CS="coordinator.openttd.org:4976" OTTD_STUN_CS="localhost" ./openttd`
- Host a server, join as client .. there is a 2 second delay till it switched  to TURN. This is because it is expecting STUN, but as we "disabled" that, it will fail. There is a 2 second timeout on waiting for STUN, hence the delay.

## Motivation / Problem

As final piece of #9017, this introduces TURN support in the OpenTTD client. This is meant to allow connecting those clients and servers that we cannot connect in any other way.

With this, it should be possible in nearly all situations for player A to play with player B, without any configuration in their network. The situations where it won't be possible are most properly extremes: those who do not have (stable) Internet, or those that have very poor peering and cannot connect to our TURN servers.

## Description

```
TURN is a last resort, used only if all other methods failed.
TURN is a relay approach to connect client and server together, where
openttd.org (by default) is the middleman.

It is very unlikely either the client or server cannot connect to
the STUN server, as they are both already connected to the Game
Coordinator. But in the odd case it does fail, estabilishing the
connection fails without any further possibility to recover.
```

For OpenTTD TURN is expensive, as we are paying for the relaying (bandwidth cost), and as such the Game Coordinator does its at-most to find any other means first.

Additional, our TURN-servers are not off-the-shelves TURN-servers, but instead they validate the protocol they are relaying. This avoids us being used as TURN-server for any other purpose other than relaying OpenTTD.

Future work extends TURN-servers to allow Websockets, meaning we can provide full network support for our Emscripten builds. But this is kept out of this PR for now.

## Checklist for review

### DIFF
--- a/docs/game_coordinator.md
+++ b/docs/game_coordinator.md
@@ -62,3 +62,22 @@ server can continue to talk to each other.
 Some NAT gateways do not allow this method; for those this attempt will fail,
 and this also means that it is not possible to create a connection between
 the client and server.
+
+## 3) Via TURN
+
+As a last resort, the Game Coordinator can decide to connect the client and
+server together via TURN. TURN is a relay service, relaying the messages
+between client and server.
+
+As the client and server can already connect to the Game Coordinator, it is
+very likely this is successful.
+
+It is important to note that a relay service has full view of the traffic
+send between client and server, and as such it is important that you trust
+the relay service used.
+For official binaries, this relay service is hosted by openttd.org. The relay
+service as hosted by openttd.org only validates it is relaying valid OpenTTD
+packets and does no further inspection of the payload itself.
+Although in our experience most patch-packs also use the services as offered
+by openttd.org, it is possible they use different services. Please be mindful
+about this.
--- a/src/network/core/tcp_turn.h
+++ b/src/network/core/tcp_turn.h
@@ -1,0 +1,79 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file tcp_turn.h Basic functions to receive and send TCP packets to/from the TURN server.
+ */
+
+#ifndef NETWORK_CORE_TCP_TURN_H
+#define NETWORK_CORE_TCP_TURN_H
+
+#include "os_abstraction.h"
+#include "tcp.h"
+#include "packet.h"
+#include "game_info.h"
+
+/** Enum with all types of TCP TURN packets. The order MUST not be changed. **/
+enum PacketTurnType {
+	PACKET_TURN_TURN_ERROR,     ///< TURN server is unable to relay.
+	PACKET_TURN_SERCLI_CONNECT, ///< Client or server is connecting to the TURN server.
+	PACKET_TURN_TURN_CONNECTED, ///< TURN server indicates the socket is now being relayed.
+	PACKET_TURN_END,            ///< Must ALWAYS be on the end of this list!! (period)
+};
+
+/** Base socket handler for all TURN TCP sockets. */
+class NetworkTurnSocketHandler : public NetworkTCPSocketHandler {
+protected:
+	bool ReceiveInvalidPacket(PacketTurnType type);
+
+	/**
+	 * TURN server was unable to connect the client or server based on the
+	 * token. Most likely cause is an invalid token or the other side that
+	 * hasn't connected in a reasonable amount of time.
+	 *
+	 * @param p The packet that was just received.
+	 * @return True upon success, otherwise false.
+	 */
+	virtual bool Receive_TURN_ERROR(Packet *p);
+
+	/**
+	 * Client or servers wants to connect to the TURN server (on request by
+	 * the Game Coordinator).
+	 *
+	 *  uint8   Game Coordinator protocol version.
+	 *  string  Token to track the current TURN request.
+	 *
+	 * @param p The packet that was just received.
+	 * @return True upon success, otherwise false.
+	 */
+	virtual bool Receive_SERCLI_CONNECT(Packet *p);
+
+	/**
+	 * TURN server has connected client and server together and will now relay
+	 * all packets to each other. No further TURN packets should be send over
+	 * this socket, and the socket should be handed over to the game protocol.
+	 *
+	 *  string  Hostname of the peer. This can be used to check if a client is not banned etc.
+	 *
+	 * @param p The packet that was just received.
+	 * @return True upon success, otherwise false.
+	 */
+	virtual bool Receive_TURN_CONNECTED(Packet *p);
+
+	bool HandlePacket(Packet *p);
+public:
+	/**
+	 * Create a new cs socket handler for a given cs.
+	 * @param s  the socket we are connected with.
+	 * @param address IP etc. of the client.
+	 */
+	NetworkTurnSocketHandler(SOCKET s = INVALID_SOCKET) : NetworkTCPSocketHandler(s) {}
+
+	bool ReceivePackets();
+};
+
+#endif /* NETWORK_CORE_TCP_TURN_H */
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -2699,3 +2699,105 @@ void ShowNetworkCompanyPasswordWindow(Window *parent)
 
 	new NetworkCompanyPasswordWindow(&_network_company_password_window_desc, parent);
 }
+
+/**
+ * Window used for asking the user if he is okay using a TURN server.
+ */
+struct NetworkAskRelayWindow : public Window {
+	std::string connection_string; ///< The TURN server we want to connect to.
+	std::string token;             ///< The token for this connection.
+
+	NetworkAskRelayWindow(WindowDesc *desc, Window *parent, const std::string &connection_string, const std::string &token) : Window(desc), connection_string(connection_string), token(token)
+	{
+		this->parent = parent;
+		this->InitNested(0);
+	}
+
+	void UpdateWidgetSize(int widget, Dimension *size, const Dimension &padding, Dimension *fill, Dimension *resize) override
+	{
+		if (widget == WID_NAR_TEXT) {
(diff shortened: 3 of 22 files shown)


docs/game_coordinator.md
```diff
@@ -62,3 +62,12 @@ server can continue to talk to each other.
 Some NAT gateways do not allow this method; for those this attempt will fail,
 and this also means that it is not possible to create a connection between
 the client and server.
+
+## 3) Via TURN
+
+As a last resort, the Game Coordinator can decide the connect the client and
+server together via TURN. TURN is a relay service, where (by default)
```
rubidium42: The "by default" kind of implies you can configure it as client, but I guess that's not the case though I wonder how a client can configure that. 
Maybe "TURN is a relay service where the Game Coordinator determines the server that will perform the relaying. For the default Game Coordinator a server from openttd.org is performing that duty."

Next to that it should probably be made clear that we are actually decoding the protocol with the implications of that. "OpenTTD's TURN servers will decode the data streams to validate the packets to ensure this service is only used for OpenTTD games. This means that it is technically possible for the TURN server to extract any information passed such as private chat messages."

Maybe we should, because of that potential, also consider having a setting for a server/client to opt out of the relaying feature. As it might be "scary" to send that information over OpenTTD's infrastructure, especially if that happens transparently in the protocol negotiation. I would probably make the default to be opted in though, as we (hopefully) are not going to do anything nefarious with those game streams.

src/network/core/tcp_turn.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+/** Base socket handler for all TURN TCP sockets. */
+class NetworkTurnSocketHandler : public NetworkTCPSocketHandler {
+protected:
+	bool ReceiveInvalidPacket(PacketTurnType type);
+
+	/**
+	 * TURN server was unable to connect the client or server based on the
+	 * ticket. Most likely cause is an invalid ticket or the other side that
```
rubidium42: Shouldn't the 'ticket' in this comment be 'token', as is used in almost all other packets?

docs/game_coordinator.md
TrueBrain: First part, will see how I can reword that.

Second part: we do not decode the messages. We only validate it is an OpenTTD packet, and relay it. There is an important difference here. We merely check the header is intact, and the length is equal to the packet. There is no inspection of the packet itself. If in the future we would add encryption, it would just keep on working, basically ;) So I will add something to this extend, but not in your wording, as that suggests we do DPI; we do not :P

Last part: I guess adding a setting is possible. Or we can ask the first time, solves any opt-in/opt-out discussion .. and maybe notify the user every time a connection via TURN is created. But mostly, this shows we need encryption for OpenTTD. That way, TURN server wouldn't be able to see what is going on; just that they are OpenTTD packets ;)

glx22: `... can decide the connect the client ...`, first `the` should be a `to`

rubidium42: From Wikipedia about "Deep packet inspection": `DPI functionality is invoked when a device looks or takes other action based on information beyond Layer 3 of the OSI model. DPI can identify and classify traffic based on a signature database that includes information extracted from the data part of a packet, allowing finer control than classification based only on header information.` Layer 3 of the OSI model is the TCP/UDP layer, OSI Layer 4 would probably be the packets defined in OpenTTD and given you take action based on the apparent validity of the packet (even if that is only the type and especially length), it is a decision on OSI Layer 4 and by definition it is some sort of Deep Packet Inspection.

So it definitely is going to decode the "packet" layer given the documentation. We say we are not going to actually do anything with the content of the packets, though there is nothing we can really do to prove that.

Yes, encryption could help, but how are you going to prove that? How does the client know if the server is who the connection says it is? Let me guess, the Game Coordinator is going to arrange that? OpenTTD is in control of that too, so it could easily intercept all traffic and decrypt it and then encrypt it with another key that the Game Coordinator provided to the client.
Or maybe the public DNS of the server, with all encryption key information in the DNS system? Though how does the client know that information? That is provided by the Game Coordinator, right? So that could also be messed with.
Yes, it could be safe in case the client is told beforehand what the server's encryption key is (assuming public key encryption for the first part), and of course the server is told the clients encryption key. Basically the whole point is that there does not seem to be a way to prove the server is who the server says to be, and as such there is no way to prove there is no MITM attack and as such you as a player/server have to believe the hosters of the TURN servers to not look at the data streams.

TrueBrain: > 
> Yes, encryption could help, but how are you going to prove that? How does the client know if the server is who the connection says it is? Let me guess, the Game Coordinator is going to arrange that? OpenTTD is in control of that too, so it could easily intercept all traffic and decrypt it and then encrypt it with another key that the Game Coordinator provided to the client.
> Or maybe the public DNS of the server, with all encryption key information in the DNS system? Though how does the client know that information? That is provided by the Game Coordinator, right? So that could also be messed with.
> Yes, it could be safe in case the client is told beforehand what the server's encryption key is (assuming public key encryption for the first part), and of course the server is told the clients encryption key. Basically the whole point is that there does not seem to be a way to prove the server is who the server says to be, and as such there is no way to prove there is no MITM attack and as such you as a player/server have to believe the hosters of the TURN servers to not look at the data streams.

I see you went through the motions of "what is possible" ;)

Just so you know, there is a proof-of-concept PR that delves into most of this, explaining how it could work. This of course would be an E2E implementation, and the only requirements would be that the L4 packet definition remains intact (so we can relay it correctly). Before you reinvent the wheel in your head, it might be a good read (there are several gists attached to it, working out some use-cases etc).

Either case, not something for this PR ofc :)

TrueBrain:
After fiddling with several solution for the "people should know their session is being relayed" part, I settled on the following solution:

When you join a server, and the session turns out to be relayed, you are being told about this via an info-popup. In this popup is the address of the relay service. For our binaries that will read `turn.openttd.org` or similar. It also includes a link to a page (which I have yet to create) to explain what is going on, in language non-techies can understand.

I did thinker a bit with adding a "You are about to connect to `turn.openttd.org`; are you sure?" dialog, but came to the realisation that most people will not understand what it is asking them (and just always click 'yes'). Such windows in general are not all that useful to add.
I think it is more useful to tell people they are being relayed, and on the page it points to explain if they don't want to, they should disconnect from the server. After all, there is no other way we can connect them otherwise, so there is not much difference. But this way we have a small opportunity to educate our users with a bit more room than just one or two sentences as we would have in-game.

TrueBrain:
> 
> When you join a server, and the session turns out to be relayed, you are being told about this via an info-popup. In this popup is the address of the relay service. For our binaries that will read `turn.openttd.org` or similar. It also includes a link to a page (which I have yet to create) to explain what is going on, in language non-techies can understand.


Minor issue with this approach .. I can only show the warning when it happens, but if you are joining, all warnings are removed just before you enter the game .. so you see this popup for a short moment for it to vanish into thin air.

Also, it is a fucking nag screen .. I need a better approach for this :D Maybe only show it once per TURN server, and ask to "accept" the relay server.

Edit: what nags me a bit about this, if someone wants to do bad, he can create his own client, host his own GC and TURN, force TURN on every connection, and remove what-ever-we-add-to-tell-the-user. So maybe this is just all complete non-sense anyway, and it is just bikeshedding.

Maybe something like this works:

```
We cannot connect you directly to the server. We would like to relay your session via 'relay.openttd.org'.

<no> <yes, this once> <yes, always>
```

I guess we need something like this.

2TallTyler:
Out of curiosity, why is it so important to inform the user that they are being relayed through an official OpenTTD server? Simply a privacy disclosure, or something else?

TrueBrain:
Exactly that. The user might expect that the connection between them and the server is private, but without any indication or notification is can now be relayed through our infra.

I am just not sure if and how to make the user a participant in this story .. it is tricky to do without being very technical about it :)

TrueBrain:
This is what I came up with:

image
image

It feels smooth and looks to be working well. So I think we have to go for this for now.

TrueBrain:
On popular request, increased the time you see the above red window from ~8 seconds to ~18 seconds. Hopefully that solves the "I didn't have time to read the window" comments I have gotten :) No clue what a good value is, as to me 8 seconds already sounded like a long time. But it is now even longer! :D

src/settings_type.h
```diff
@@ -290,6 +290,7 @@ struct NetworkSettings {
 	bool        reload_cfg;                               ///< reload the config file before restarting
 	std::string last_joined;                              ///< Last joined server
 	bool        no_http_content_downloads;                ///< do not do content downloads over HTTP
+	uint8       use_relay_service;                        ///< Use relay service? (0 = never, 1 = ask, 2 = allow)
```
rubidium42: I'd rather see an enum here just like is done with `ServerGameType``.

src/network/network_gui.cpp
```diff
@@ -2699,3 +2699,108 @@ void ShowNetworkCompanyPasswordWindow(Window *parent)
...
 }
+
+/**
+ * Window used for asking the user if he is okay using a TURN server.
+ */
+struct NetworkAskRelayWindow : public Window {
+	std::string connection_string; ///< The TURN server we want to connect to.
+	std::string token;             ///< The token for this connection.
+
+	NetworkAskRelayWindow(WindowDesc *desc, Window *parent, const std::string &connection_string, std::string &token) : Window(desc)
```
rubidium42: Why isn't token `const`?

src/network/network_gui.cpp
```diff
@@ -2699,3 +2699,108 @@ void ShowNetworkCompanyPasswordWindow(Window *parent)
...
+	{
+		this->parent = parent;
+		this->connection_string = connection_string;
+		this->token = token;
```
rubidium42: Maybe use C++ style initialisation?

src/network/network_gui.cpp
```diff
@@ -2699,3 +2699,108 @@ void ShowNetworkCompanyPasswordWindow(Window *parent)
...
+	WDF_MODAL,
+	_nested_network_ask_relay_widgets, lengthof(_nested_network_ask_relay_widgets)
+);
+
+/**
+ * Show a modal confirmation window with "no" / "yes, once" / "yes, always" buttons.
+ * @param connection_string The relay server we want to connect to.
+ * @param token The token for this connection.
+ */
+void ShowNetworkAskRelay(const std::string &connection_string, std::string &token)
```
rubidium42: Token should probably be `const` as I doubt you intend to change it here.

src/network/network_gui.cpp
TrueBrain: It just wants to be special, okay? Nothing wrong with that! :P

rubidium42 (approved):

PR #8285, opened 2020-07-30
Rewrite `OverflowSafeInt` to avoid undefined behaviour

techgeeknz:
The current implementation of `OverflowSafeInt` does not handle the maximally-negative boundary condition very well. This rewrite uses an unsigned type to calculate and represent absolute values, to avoid invoking undefined behaviours such as `abs(INT64_MIN) < 0` and `-INT64_MIN == INT64_MIN`. Fixes #8284.

PR #9451, opened 2021-07-19
Fix potential undefined behaviour when using OverflowSafeInt

### LINKED ISSUE #8284: Negating or taking absolute values of maximally-negative OverflowSafeInt may result in undefined behaviour
As mentioned in https://github.com/OpenTTD/OpenTTD/pull/7924, `OverflowSafeInt` can produce incorrect results when the value underflows. This appears to be the result of undefined behaviour caused by taking the inverse or absolute value of T_MIN; the magnitude of which cannot usually be represented as a positive value of the same integer type.

I am not yet familiar with OpenTTD's continuous integration and regression testing system; however, I have an idea to work around the issue using typecasts to unsigned integers where appropriate.

LordAro:
Fixes #8284
Closes #8285 

## Motivation / Problem
OverflowSafeInt can actually underflow.

## Description
Changes ~completely copied from~ inspired by JGR's work:
https://github.com/JGRennison/OpenTTD-patches/commit/6821c0e9453accba456d5484604f4ed8fd8c94ec
https://github.com/JGRennison/OpenTTD-patches/commit/bb8d2c3fe014c9459780db3e93ca4e182c87f4de
https://github.com/JGRennison/OpenTTD-patches/commit/609f37cef953b2e95ed21216eaa51928aa688d1e
https://github.com/JGRennison/OpenTTD-patches/commit/db0e25f7cd8740355a1183f7399641f3d277cd80

But mildly neatened and modernised.

## Checklist for review

### DIFF
--- a/src/core/overflowsafe_type.hpp
+++ b/src/core/overflowsafe_type.hpp
@@ -39,16 +39,33 @@ public:
 	/**
 	 * Safe implementation of addition.
 	 * @param other the amount to add
-	 * @note when the addition would yield more than T_MAX (or less than T_MIN),
-	 *       it will be T_MAX (respectively T_MIN).
+	 * @note when the addition would yield more than T_MAX, it will be T_MAX.
 	 */
 	inline OverflowSafeInt& operator += (const OverflowSafeInt& other)
 	{
-		if ((T_MAX - abs(other.m_value)) < abs(this->m_value) &&
-				(this->m_value < 0) == (other.m_value < 0)) {
-			this->m_value = (this->m_value < 0) ? T_MIN : T_MAX ;
+		if (this->m_value > 0 && other.m_value > 0 && (T_MAX - other.m_value) < this->m_value) {
+			this->m_value = T_MAX;
+		} else if (this->m_value < 0 && other.m_value < 0 && (this->m_value == T_MIN || other.m_value == T_MIN || ((T_MAX + this->m_value) + other.m_value < (T_MIN + T_MAX)))) {
+			this->m_value = T_MIN;
 		} else {
 			this->m_value += other.m_value;
+		}
+		return *this;
+	}
+
+	/**
+	 * Safe implementation of subtraction.
+	 * @param other the amount to subtract
+	 * @note when the subtraction would yield less than T_MIN, it will be T_MIN.
+	 */
+	inline OverflowSafeInt& operator -= (const OverflowSafeInt& other)
+	{
+		if (this->m_value > 0 && other.m_value < 0 && (T_MAX + other.m_value) < this->m_value) {
+			this->m_value = T_MAX;
+		} else if (this->m_value < 0 && other.m_value > 0 && (T_MAX + this->m_value) < (T_MIN + T_MAX) + other.m_value) {
+			this->m_value = T_MIN;
+		} else {
+			this->m_value -= other.m_value;
 		}
 		return *this;
 	}
@@ -57,7 +74,6 @@ public:
 	inline OverflowSafeInt  operator +  (const OverflowSafeInt& other) const { OverflowSafeInt result = *this; result += other; return result; }
 	inline OverflowSafeInt  operator +  (const int              other) const { OverflowSafeInt result = *this; result += (int64)other; return result; }
 	inline OverflowSafeInt  operator +  (const uint             other) const { OverflowSafeInt result = *this; result += (int64)other; return result; }
-	inline OverflowSafeInt& operator -= (const OverflowSafeInt& other)       { return *this += (-other); }
 	inline OverflowSafeInt  operator -  (const OverflowSafeInt& other) const { OverflowSafeInt result = *this; result -= other; return result; }
 	inline OverflowSafeInt  operator -  (const int              other) const { OverflowSafeInt result = *this; result -= (int64)other; return result; }
 	inline OverflowSafeInt  operator -  (const uint             other) const { OverflowSafeInt result = *this; result -= (int64)other; return result; }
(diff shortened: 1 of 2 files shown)


src/core/overflowsafe_type.hpp
```diff
@@ -39,25 +39,41 @@ class OverflowSafeInt
...
-	 * @note when the addition would yield more than T_MAX (or less than T_MIN),
-	 *       it will be T_MAX (respectively T_MIN).
+	 * @note when the addition would yield more than T_MAX, it will be T_MAX.
 	 */
 	inline OverflowSafeInt& operator += (const OverflowSafeInt& other)
 	{
-		if ((T_MAX - abs(other.m_value)) < abs(this->m_value) &&
-				(this->m_value < 0) == (other.m_value < 0)) {
-			this->m_value = (this->m_value < 0) ? T_MIN : T_MAX ;
+		if ((this->m_value > 0) && (other.m_value > 0) && (T_MAX - other.m_value) < this->m_value) {
```
TrueBrain: lots of unneeded `()`. Not sure that is a good or bad thing, just an observation at this point :D

src/core/overflowsafe_type.hpp
```diff
@@ -39,25 +39,41 @@ class OverflowSafeInt
 	/**
 	 * Safe implementation of addition.
 	 * @param other the amount to add
-	 * @note when the addition would yield more than T_MAX (or less than T_MIN),
-	 *       it will be T_MAX (respectively T_MIN).
+	 * @note when the addition would yield more than T_MAX, it will be T_MAX.
```
TrueBrain, line 51: this commit message is a bit weird. `Fix: OverflowSafeInt behaviour with underflows `. and there really isn't much more context what this commit is trying to address, except "something" with underflows :) Maybe spend a few words on it in the commit message, or reword it to make it a bit more clear what was happening here?

The PR also says "potential" undefined behaviour .. which always makes me wonder: is it really potential? :D

Mainly I am trying to understand what this commit fixes, and it is really difficult to see :P Math hard .. :D

LordAro: "potential" comes from the fact that you only get undefined behaviour if you over/underflow. Not at all in 'normal' usage

TrueBrain:
I am really certain I do not understand this code, so I just brute-forced it what I could think of:

https://godbolt.org/z/KE3qc1cqq

If you think I missed cases, feel free to add them and post the new link ;)

Seems all cases I could come up with work fine.

Edit: just to smoketest the compiler built-in variants: https://godbolt.org/z/5T8bPeYKe . Also shows no issues for the cases I came up with.

src/core/overflowsafe_type.hpp
TrueBrain: Suggestion:

```
Fix: OverflowSafeInt's underflow protection doesn't work in all cases

For example: INT32_MIN - 1 != INT32_MIN, while it would be expected
to be INT32_MIN.
```

rubidium42:
> Edit: just to smoketest the compiler built-in variants: https://godbolt.org/z/5T8bPeYKe . Also shows no issues for the cases I came up with.

Maybe guard against the built-in variant being used when min/max are not the min/max of the type? For example, the following works with the original code, but fails with the built-in variant.
```
OverflowSafeInt<int8_t, 5, 3> c = 3;
assert(c + 3 == 5);
```

LordAro:
Is it worth supporting that though? We can get rid of the extra 2 tparams with std::numeric_limits these days

rubidium42:
> Is it worth supporting that though? We can get rid of the extra 2 tparams with std::numeric_limits these days

Knowing that std::numeric_limits exists was kind of the reason I was trying this as I (naively) assumed that it might have worked. After all, why have those parameters. Though removing the parameters seems another viable solution to the problem.

TrueBrain:
I think using `numeric_limits` would be great. I just worry about the depth of this rabbithole you found. Also:

https://github.com/OpenTTD/OpenTTD/blob/d1cf56609defb1ea8bd5d642f6df1bccb000d7d6/src/script/api/script_company.cpp#L143-L145

Can we address those 2 instances while at it? Seems it should just be `OverflowSafeInt32` :)

src/core/overflowsafe_type.hpp
```diff
@@ -34,7 +34,7 @@ class OverflowSafeInt
 
 	inline OverflowSafeInt& operator = (const OverflowSafeInt& other) { this->m_value = other.m_value; return *this; }
 
-	inline OverflowSafeInt operator - () const { return OverflowSafeInt(-this->m_value); }
+	inline OverflowSafeInt operator - () const { return OverflowSafeInt(this->m_value == T_MIN ? T_MAX : -this->m_value); }
```
TrueBrain: Re commit message `Fix: OverflowSafeInt negation not handling INT64_MIN `
Here too the commit message could use a bit more detail what is going on, for future us. Something like:

```
Fix: OverflowSafeInt negation not handling INT64_MIN

INT64_MIN negated is above INT64_MAX, and would overflow.
Instead, when negating INT64_MIN make it INT64_MAX.
This does mean that -(-(INT64_MIN)) != INT64_MIN.
```

Or something along those lines

TrueBrain (dismissed):
For future-us: the mistake made here is that `-T_MAX != T_MIN`, but the code did assume it. In essence, that is what broke underflow protection, and what this PR addresses.

TrueBrain (dismissed):

src/core/overflowsafe_type.hpp
```diff
@@ -138,74 +136,81 @@ class OverflowSafeInt
...
-template <class T> inline OverflowSafeInt<T> operator / (byte  a, OverflowSafeInt<T> b) { return (OverflowSafeInt<T>)a / (int)b; }
+template <class T> inline constexpr OverflowSafeInt<T> operator + (const byte  a, const OverflowSafeInt<T> b) { return b + (uint)a; }
+template <class T> inline constexpr OverflowSafeInt<T> operator - (const byte  a, const OverflowSafeInt<T> b) { return -b + (uint)a; }
+template <class T> inline constexpr OverflowSafeInt<T> operator * (const byte  a, const OverflowSafeInt<T> b) { return b * (uint)a; }
+template <class T> inline constexpr OverflowSafeInt<T> operator / (const byte  a, const OverflowSafeInt<T> b) { return (OverflowSafeInt<T>)a / (int)b; }
 
 typedef OverflowSafeInt<int64> OverflowSafeInt64;
 typedef OverflowSafeInt<int32> OverflowSafeInt32;
 
+/* Some basic "unit tests". Also has the bonus of confirming that constexpr is working */
```
TrueBrain: If we are nitpicking anyway, mind adding dots at the end of sentences?

LordAro: I was following the convention of the current file :p

TrueBrain (approved):
Nice.

PR #9455, opened 2021-07-22
Fix #9440: negative cargo payments not being handled right

### LINKED ISSUE #9440: [Bug]: Inflation breaks negative cargo payments
### Version of OpenTTD

OpenTTD 1.11.2, Windows 10

### Expected result

My NewGRF industry set, Industries of the Caribbean, uses negative cargo rates set in the NML properties (not the profit callback, [where a known issue exists](https://github.com/OpenTTD/nml/pull/212)).

When Inflation is disabled, this works as intended. A cargo delivery like that in the savegame attached below costs the company about -£50.

### Actual result

When Inflation is enabled, that -£50 cargo delivery instead gives £205 million.

### Steps to reproduce

A user provided this savegame illustrating the issue.
[IOTC_inflation_bug.zip](https://github.com/OpenTTD/OpenTTD/files/6809096/IOTC_inflation_bug.zip)

When a new map is started with Inflation disabled and the same set of NewGRFs, the profit works as intended.

rubidium42:
## Motivation / Problem

Fixes #9440; negative cargo payments causing huge payments.

## Description

```
Cargo payments were stored as unsigned integer, but cast to int64 during
application of inflation. However, then being multiplied with a uint64
making the result uint64. So in the end the payment that should have been
negative becomes hugely positive.
```

## Limitations

First and major question is whether it is wanted that cargo payments can be negative. The specifications do not seem to be very clear about it, though the custom cargo calculation allowing negative factors might imply that the cargo payments themselves can be negative.

So, this makes it possible and seemingly work for the graphs and actual payment, but I'm far from certain it should be this way. So closing this PR and #9440 as not-a-bug would be a valid solution as well.

## Checklist for review

### DIFF
--- a/src/cargotype.h
+++ b/src/cargotype.h
@@ -61,7 +61,7 @@ struct CargoSpec {
 	uint8 rating_colour;
 	uint8 weight;                    ///< Weight of a single unit of this cargo type in 1/16 ton (62.5 kg).
 	uint16 multiplier;               ///< Capacity multiplier for vehicles. (8 fractional bits)
-	uint32 initial_payment;          ///< Initial payment rate before inflation is applied.
+	int32 initial_payment;           ///< Initial payment rate before inflation is applied.
 	uint8 transit_days[2];
 
 	bool is_freight;                 ///< Cargo type is considered to be freight (affects train freight multiplier).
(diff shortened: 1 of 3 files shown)


michicc (approved):
Yes. For whatever it counts.

PR #9457, opened 2021-07-24
Feature: [Linkgraph] Prioritize faster routes for passengers and express cargo

nchappe:
## Motivation / Problem

Cargodist does not optimize for travel time but for distance, which means passengers sometimes take (for instance) a slow but short bus route instead of a faster train route to their destination. I think it would be more intuitive if cargodist prioritized the fastest paths, at least for passengers and express cargo (goods, food). Since cargo payment depends on travel time, this may also increase income for some networks.

## Description

This PR adds a member variable `travel_time_sum` to `LinkGraph::Edge`. When a vehicle traverses a link, the product of its travel time and its capacity is added to this variable, so that `travel_time_sum / capacity` gives *the average travel time of a cargo unit traversing this edge in a saturated network*. Just like capacities, the value is halved every 256 days, so that older travel times weigh less in the average.

This average travel time is then used in the MCF solver in place of the distance for cargo of classes `CC_PASSENGERS` and `CC_EXPRESS`.

## Limitations

This change could be hidden behind a default-disabled setting for better backward compatibility.

## Checklist for review

### DIFF
--- a/src/linkgraph/mcf.cpp
+++ b/src/linkgraph/mcf.cpp
@@ -284,12 +284,21 @@ void MultiCommodityFlow::Dijkstra(NodeID source_node, PathVector &paths)
 				capacity /= 100;
 				if (capacity == 0) capacity = 1;
 			}
-			/* punish in-between stops a little */
+			/* Prioritize the fastest route for passengers, mail and express cargo,
+			 * and the shortest route for other classes of cargo.
+			 * In-between stops are punished with a 1 tile or 1 day penalty. */
+			bool express = IsCargoInClass(this->job.Cargo(), CC_PASSENGERS) ||
+				IsCargoInClass(this->job.Cargo(), CC_MAIL) ||
+				IsCargoInClass(this->job.Cargo(), CC_EXPRESS);
 			uint distance = DistanceMaxPlusManhattan(this->job[from].XY(), this->job[to].XY()) + 1;
+			/* Compute a default travel time from the distance and an average speed of 1 tile/day. */
+			uint time = (edge.TravelTime() != 0) ? edge.TravelTime() + DAY_TICKS : distance * DAY_TICKS;
+			uint distance_anno = express ? time : distance;
+
 			Tannotation *dest = static_cast<Tannotation *>(paths[to]);
-			if (dest->IsBetter(source, capacity, capacity - edge.Flow(), distance)) {
+			if (dest->IsBetter(source, capacity, capacity - edge.Flow(), distance_anno)) {
 				annos.erase(dest);
-				dest->Fork(source, capacity, capacity - edge.Flow(), distance);
+				dest->Fork(source, capacity, capacity - edge.Flow(), distance_anno);
 				dest->UpdateAnnotation();
 				annos.insert(dest);
 			}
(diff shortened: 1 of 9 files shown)


src/saveload/saveload.h
```diff
@@ -337,6 +337,7 @@ enum SaveLoadVersion : uint16 {
 
 	SLV_TABLE_CHUNKS,                       ///< 295  PR#9322 Introduction of CH_TABLE and CH_SPARSE_TABLE.
 	SLV_SCRIPT_INT64,                       ///< 296  PR#9415 SQInteger is 64bit but was saved as 32bit.
+	SLV_LINKGRAPH_TRAVEL_TIME,              ///< 297          Store travel time in the linkgraph.
```
LordAro: Can be filled in now

LordAro (commented):
Strange mismatch between `uint64` and `uint32` (and `uint`). Sure that these are all correct?

nchappe:
Thanks for the comments. There were indeed some typing inconsistencies with travel times, now they are all uint32. As for `LinkGraph::BaseEdge::travel_time_sum`, it has to be uint64 because it's basically `travel_time * capacity`, which would not fit in a uint32.

src/linkgraph/mcf.cpp
```diff
@@ -284,12 +284,19 @@ void MultiCommodityFlow::Dijkstra(NodeID source_node, PathVector &paths)
...
 			}
-			/* punish in-between stops a little */
+			/* Prioritize the fastest route for passengers and express cargo,
+			 * and the shortest route for other classes of cargo.
+			 * In-between stops are punished with a 1 tile or 1 day penalty. */
 			uint distance = DistanceMaxPlusManhattan(this->job[from].XY(), this->job[to].XY()) + 1;
+			/* Compute a default travel time from the distance and an average speed of 1 tile/day. */
+			uint time = (edge.TravelTime() != 0) ? edge.TravelTime() + DAY_TICKS : distance * DAY_TICKS;
+			uint distance_anno = (IsCargoInClass(this->job.Cargo(), CC_PASSENGERS) ||
+				IsCargoInClass(this->job.Cargo(), CC_EXPRESS)) ? time : distance;
```
michicc: I'd include `CC_MAIL` for time as well.

nchappe: This makes sense, I added the CC_MAIL class.

I wonder if mail, passengers, and maybe fruits could be added to the CC_EXPRESS class? A comment in cargo_type.h states that CC_EXPRESS is "also possible for passengers", and it would simplify this time/distance logic.

michicc: No, changing the pre-defined default cargoes is a bad idea as a heap of NewGRFs depend on it.

LordAro: Comment above needs updating too

LordAro: Probably best to pull out the ternary condition into its own variable too

nchappe: I've updated the forgotten comment and moved the ternary.

michicc (approved):
My mini test-game confirms that it works in some way (😄), and I like the behaviour more.

Test setup: A slow direct bus route compared to a longer but faster train route with one intermediate change. Unmodified, pax are perpetually waiting for full buses while with this PR, the trains carry the main load.

Somebody will cry wolf that there carefully timed network is completely disrupted, of course.

Issue #9458, opened 2021-07-27
[Bug]: desync after joining server


ecolortest:
### Version of OpenTTD

1.11.2, macos 10.13.6

### Expected result

should not desync after joining server

### Actual result

get desync from server after 30 second joining the server

### Steps to reproduce

1. join reddit.com/r/OpenTTD 1 - Vanilla
2. after joining, 30 seconds later then desync

James103:
I joined the server you mentioned, waited for a couple minutes, and did not desync in the meantime. The server has probably reset to a new game", making this issue invalid (expired / out of date).

"Reddit servers reset approximately every 22 hours if running with default 1950-2050 games at 1x time speed.

ecolortest:
> I joined the server you mentioned, waited for a couple minutes, and did not desync in the meantime. The server has probably reset to a new game", making this issue invalid (expired / out of date).
> 
> "Reddit servers reset approximately every 22 hours if running with default 1950-2050 games at 1x time speed.

its was desync, not an world reset

James103:
As of the time of writing, the server `reddit.com/r/OpenTTD 1 - Vanilla` has [permanently closed](https://www.reddit.com/r/openttd/comments/orze4l/all_good_things_and_their_endings/), making this issue invalid unless it can be reproduced on a different server.

TrueBrain:
Sadly, without a savegame, this is really difficult for us to debug. Let alone confirm the problem is fixed :)

If this happens in the future, we would greatly appreciate a savegame (the game makes a `netsave` emergency save on desync), so at least we have something to work with. Ideally, we also have the savegame from the server, but that might be harder to obtain :)

In short, I really do appreciate you bringing this to our attention, but as it is, this is not actionable :( Sorry .. but in positive news, recently we did find a desync bug with MacOS clients .. hopefully that was causing this issue too :)

PR #9461, opened 2021-08-03
Fix 68f2213: Don't use GetPoolSize() for end of pool iterator

glx22:
## Motivation / Problem
`GetPoolSize()` returned value can change during the pool iteration, so using it for `end()` was not a good idea.

## Description
Use a constant (max size of the pool) for `end()`, and set `index` to this value whenever it reaches `GetPoolSize()`.

## Checklist for review

### DIFF
--- a/src/core/pool_type.hpp
+++ b/src/core/pool_type.hpp
@@ -160,7 +160,11 @@ struct Pool : PoolBase {
 
 	private:
 		size_t index;
-		void ValidateIndex() { while (this->index < T::GetPoolSize() && !(T::IsValidID(this->index))) this->index++; }
+		void ValidateIndex()
+		{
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):
Might solve #9458, but definitely solves a desync via the Vehicle tick counter creating effect vehicles that might not get iterated over.
This as the range based for loop gets `end`, i.e the original pool size (or actually the highest ID ever assigned), and iterates up to there. If the game already ran a while that highest ID might be significantly higher than the current highest ID, so a joining client would have a lower highest ID, and thus a lower pool size and thus a lower 'value' for `end()`. This means that in the existing clients the effect vehicle created in the vehicle tick gets iterated over, whereas it does not in the newly joined client causing a divergence in state and eventual desync. Especially as the effect vehicle will be removed earlier, and all assigned vehicle IDs may be off after that event.

PR #9467, opened 2021-08-10
Feature: make "join game" button join the game, instead of first showing a lobby window

TrueBrain:
## Motivation / Problem

(part of commit message)
```
Nobody really paid attention to the lobby window, and it completely
missed its purpose. Most people don't even wait for companies to
show up, but just hit "New Company".
This in turn means people create a lot of unneeded companies, while
they "just want to watch the game" or join another company.
```

## Description

(part of commit message)
```
Instead, "Join Game" now just joins the game as spectators.
```

A future PR will address the gap that now exists between: I joined a server, now how to start playing?

## Checklist for review


src/network/network_client.cpp
```diff
@@ -595,39 +589,7 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::Receive_SERVER_COMPANY_INFO(Pa
...
-		}
-		company_info->ai               = p->Recv_bool();
-
-		company_info->clients = p->Recv_string(NETWORK_CLIENTS_LENGTH);
-
-		SetWindowDirty(WC_NETWORK_WINDOW, WN_NETWORK_WINDOW_LOBBY);
-
-		return NETWORK_RECV_STATUS_OKAY;
-	}
+	/* Unused, but this packet is part of the "this will never change" packet group. */
```
rubidium42, line 594: You can remove the whole function. The client should only get it after requesting it, so it should not expect it. So, it's perfectly fine to just let the default implementation of tcp_game.cpp handle this packet. So consider it an invalid packet.

src/window_type.h
```diff
@@ -25,7 +25,6 @@ enum WindowNumberEnum {
 	WN_CONFIRM_POPUP_QUERY_BOOTSTRAP, ///< Query popup confirm for bootstrap.
 
 	WN_NETWORK_WINDOW_GAME = 0,     ///< Network game window.
-	WN_NETWORK_WINDOW_LOBBY,        ///< Network lobby window.
```
rubidium42, line 28: The removal of a window number enum or a set of widgets technically breaks the GSWindow API. Though practically the GS can't be running when the lobby window is open, so... I guess it's not an issue. Also wonderful that the join/download window numbers are exposed to GSes, which will never be able to highlight (or close) them.

rubidium42 (commented):
Can't the server side be simplified as well? Practically not implementing `Receive_CLIENT_COMPANY_INFO` (and letting tcp_game's implementation take over) would be possible since the lobby should never be shown without going through the game list which prevents connecting to a server with the wrong version, right?
Alternatively sending SERVER_COMPANY_INFO (`SendCompanyInfo`) gets heavily simplified by just sending a packet with that type and a new version number and a `false` boolean. Then the protocol is not closing the client's connection, but the client gets nothing except an unexpected version which is to be expected when you send a newer version. Although that might ofcourse cause whining from some people that you cannot acquire that information anymore, but alas... they should be using the admin protocol instead I guess.
I can imagine postponing this can of worms to a separate PR though.

src/network/network_client.cpp
TrueBrain: I wanted to keep the function around for visibility. It is also not empty, and as we have seen the other day, the pre-conditions can be useful for future-us. So I think we should leave it for now, as older servers can strictly seen send this command to us for what-ever reason.

Edit: I created #9475 to follow up on this.

src/window_type.h
TrueBrain: Also, we always told GS authors to use the enums supplied, and not the internal values.

Edit: this means worst case a GS breaks because they used this enum .. not sure I care :D
Edit 2: on request by glx, added it to the gs-changelog.

TrueBrain:
> 
> 
> Can't the server side be simplified as well? Practically not implementing `Receive_CLIENT_COMPANY_INFO` (and letting tcp_game's implementation take over) would be possible since the lobby should never be shown without going through the game list which prevents connecting to a server with the wrong version, right?
> Alternatively sending SERVER_COMPANY_INFO (`SendCompanyInfo`) gets heavily simplified by just sending a packet with that type and a new version number and a `false` boolean. Then the protocol is not closing the client's connection, but the client gets nothing except an unexpected version which is to be expected when you send a newer version. Although that might ofcourse cause whining from some people that you cannot acquire that information anymore, but alas... they should be using the admin protocol instead I guess.
> I can imagine postponing this can of worms to a separate PR though.

As with the other comment, these packets are part of the "we never going to change this" group. I am in no mood to look into that any further :P So possibly you are right .. but yeah, not for this PR :)

Edit: I created #9475 to address this.

src/network/core/tcp_game.h
```diff
@@ -38,7 +38,7 @@ enum PacketGameType {
 	PACKET_CLIENT_JOIN,                  ///< The client telling the server it wants to join.
 	PACKET_SERVER_ERROR,                 ///< Server sending an error message to the client.
 
-	/* Packets used for the pre-game lobby. */
+	/* Packets used for the pre-game lobby (unused, but remain for backward/forward compatibility). */
 	PACKET_CLIENT_COMPANY_INFO,          ///< Request information about all companies.
```
frosch123, line 42: maybe add a "_UNUSED" to the constants themself?

src/script/api/game_changelog.hpp
```diff
@@ -21,6 +21,14 @@
  * \li GSNewGRF
  * \li GSNewGRFList
  *
+ * API removals:
+ * \li GSWindow::NetworkStartServerWidgets::WID_NSS_SPECTATORS_LABEL
```
frosch123: The GUI changes with every version. GSWindow explicitly states that it has no version guarantee. We never document GUI changes here.

src/network/core/tcp_game.h
TrueBrain: To give that change a bit more visibility, I did that in https://github.com/OpenTTD/OpenTTD/pull/9475 .

frosch123 (approved):

PR #9469, opened 2021-08-10
Feature: servers can now set a Message Of The Day

TrueBrain:
Includes #9467

Draft, because it will most likely be superseded by #9474, reaching the same goal in a different way.

## Motivation / Problem

Now we no longer have a lobby window, it is a bit unclear for new players what they should be doing to start playing. So I wanted to add something that points this out.

It was suggested to add a MOTD for this, and that sounds like an excellent idea. I tried a few things, see what I liked most, and came up with this.

## Description

- Synchronize server-name, and put that in the "Online Players" window too.
- On join, show a MOTD window, which contains:
  - server-name in caption.
  - MOTD in body (which the server-owner can change).
  - As footer, when you are (still) spectator, a hint how to get playing.

When the MOTD is changed when you are connected, you are shown the MOTD again. The footer will be removed if you are no longer a spectator. Not sure this is wanted behaviour, but it sounds like the right thing to do to me.

Additionally, I decided MOTDs should never be empty. It just feels better to force server-owners to write something nice :) But I can be convinced otherwise!

image
image
image

PS: suggestions for the default MOTD (`Please enjoy your stay.`) are welcome :)

## Limitations

- There is a weird bug when you have a very short MOTD, `GetStringHeight` returns 0. I have no clue why, but I also moved it out of this PR to fix that, as I do not think it is a bug in this PR.
- Ideally I would like that the MOTD accepts `{}` codes, so you can make it fancy and colourful. But .. that seems neither easy and also a bit tricky. So I didn't do it for this PR, but it might be future-work.

## Checklist for review


TrueBrain:
.

src/lang/english.txt
```diff
@@ -2236,6 +2218,7 @@ STR_NETWORK_ERROR_TIMEOUT                                       :{WHITE}Connecti
 STR_NETWORK_ERROR_SERVER_ERROR                                  :{WHITE}A protocol error was detected and the connection was closed
 STR_NETWORK_ERROR_BAD_PLAYER_NAME                               :{WHITE}Your player name has not been set. The name can be set at the top of the Multiplayer window
 STR_NETWORK_ERROR_BAD_SERVER_NAME                               :{WHITE}Your server name has not been set. The name can be set at the top of the Multiplayer window
+STR_NETWORK_ERROR_BAD_SERVER_MOTD                               :{WHITE}The Message Of The Day has not been set. The name can be set in the Start Server window
```
```suggestion
STR_NETWORK_ERROR_BAD_SERVER_MOTD                               :{WHITE}The Message Of The Day has not been set. The message can be set in the Start Server window
```

ldpl:
I'm very much against this feature the way it is right now. Though I'm just looking at UI and PR description like normal users and didn't dig too deep into the code to see if some of these limitations can be hacked around.

First of all, the whole premise is wrong. You're trying to solve bad UI by adding more bad UI. Go fix mp window, don't add kludges in other places.

And what is this window even supposed to be? Is it MOTD or is it welcome message? MOTD it's a weird thing to forcefully pop up on login, especially since it doesn't even change daily :p 

Welcome message can be a useful feature but the way it is atm it's not usable because:
- One line is not enough. Neither is one colour.
- Uses same server name in the header as network lobby window. And many decent servers use that to show additional info that doesn't make sense in a window header (just look at the server list).
- No way to set it from GS or disable it to show custom welcome popup.

So, basically, as I see atm it can only be used to put some random junk in that message. If I can't find a way to hack around it I may be forced to show something like "Sorry for this stupid popup but OpenTTD forces us to show it, please read the real welcome message that follows" there.

ldpl:
Some real welcome messages for example:
Screenshot from 2021-08-13 15-01-18
Screenshot from 2021-08-13 14-54-01
Screenshot from 2021-08-13 14-53-10
Screenshot from 2021-08-10 21-50-47

nielsmh:
What I think would *actually* be useful would be to store the MOTD in a text file by itself, first and foremost.
The client could store a hash of the MOTD somewhere so it might not display it repeatedly if the player keeps joining the same server.

Bonus feature would be to have an in-game administration UI to let the server owner edit the MOTD. That would require implementing a multi-line text editor UI.
Also see #7786 for a window that supports hyperlinks and long, scrollable text.

TrueBrain:
@ldpl 
> 
>     * One line is not enough. Neither is one colour.

We can demand a golden palace from day 1, but sometimes all we get is a shed. Either way, this is an opinion, and that is fine.

@ldpl 
> 
>     * Uses same server name in the header as network lobby window. And many decent servers use that to show additional info that doesn't make sense in a window header (just look at the server list).

Personally, I do not think this really is an issue. We could limit the text, but I would also hardly call it a show-stopper. But I am open for alternatives.

@ldpl 
> 
>     * No way to set it from GS or disable it to show custom welcome popup.

No. And in my opinion we should never allow GS to change MOTD. GS has plenty of ways to communicate with the user, no need to add yet-another-one on top of that. This is not what a MOTD is for.

@nielsmh 
> 
> 
> What I think would _actually_ be useful would be to store the MOTD in a text file by itself, first and foremost.
> The client could store a hash of the MOTD somewhere so it might not display it repeatedly if the player keeps joining the same server.
> 
> Bonus feature would be to have an in-game administration UI to let the server owner edit the MOTD. That would require implementing a multi-line text editor UI.
> Also see #7786 for a window that supports hyperlinks and long, scrollable text.

Not sure the word "actually" was really needed here, but yeah, I started out with trying to do exactly that ;) But there are more than a few problems, as it turns out :)

The main thing I found out, that adding a feature that cannot be controlled from in-game, is a weird thing to do. With that I mean it didn't feel right that you could only modify the MOTD via a text-editor.

So I wanted to make it editable from the interface (you called it a bonus, but I think that is pretty mandatory :P). But as you mentioned, that is currently not possible .. and adding it doesn't seem easy. So I tried to allow `{}` and stuff .. but that also turns out to be non-trivial. Maybe someone else has ideas (hence me mentioning it in Limits :P), as that would bridge the gap. Would also allow colours etc, which I would really like, but .. yeah .. couldn't find a way to get it to work. Tried to follow how it works for GSes, but got lost in the maze somewhere ...

That is how I got to this result, as a stop-gap, allowing for plenty of future-work. Rome doesn't have to be build in one day ofc :)

If you have some suggestions how to bridge the gap, I am all ears (and again, why I mention these things in limitations :D). Would be nice if we can allow `{}`, but I really struggled there.

And yes, when is your next holiday to finish #7786? That would be really useful and cool :D :D

ldpl:
> We can demand a golden palace from day 1, but sometimes all we get is a shed. Either way, this is an opinion, and that is fine.

It's one thing to start small and another to add a useless popup that exists only to kludge an issue in some other place. Welcome message is the face of the server, it doesn't get a second chance to introduce itself to the player. And you're trying to brush it off with some one-liner.

> > ```
> > * Uses same server name in the header as network lobby window. And many decent servers use that to show additional info that doesn't make sense in a window header (just look at the server list).
> > ```
> 
> Personally, I do not think this really is an issue. We could limit the text, but I would also hardly call it a show-stopper. But I am open for alternatives.

It's not a big issue and, tbh, it's an issue with the network lobby that servers need all that junk. But until it's addressed I guess I'd suggest adding a second setting: lobby name for lobby and short name for other UI.

> > ```
> > * No way to set it from GS or disable it to show custom welcome popup.
> > ```
> 
> No. And in my opinion we should never allow GS to change MOTD. GS has plenty of ways to communicate with the user, no need to add yet-another-one on top of that. This is not what a MOTD is for.

"Plenty" of ways but no good one :p  Welcome message that you for some reason keep calling MOTD is a unique window, allowing GS modification is to allow maximizing its utility. For example, look at CityMania welcome window, it actually calculates the exact time until server restart, super useful feature. Oh, and btw, even in its current form GS are probably able to change it since they can change settings, there is just no point.

P.S. MOTD: https://i.imgur.com/nlPHUzO.png

TrueBrain:
Going to close this PR for now. Most developers agree that having a MOTD would be valuable additional to OpenTTD, but also that not having multiline support sucks. As that requires more groundwork first, it is unlikely this Pull Request is useful by the time that is done. Additionally, I will not actively be pursuing adding MOTD for now, and I hate having dangling PRs on my name I have no further interest in :D

The goal of having a place to hint to players they are spectator and should join a company has been addressed in others ways, at least, so we hope. We will find out with our first beta :D

PR #9474, opened 2021-08-13
Add: [InterfaceTutorial] show tutorial the first time you join any server 

TrueBrain:
Closes #9469.

## Motivation / Problem

With #9467 we can leave new players a bit lost in where to go next after joining a server. So after some talking to @nielsmh on IRC, we figured a tutorial would really help. Not only here, but on many more places in the game.

This is as alternative to #9469, which addresses the same issue, but via a MOTD. A tutorial might be more to the point, as it means the MOTD window doesn't share more than one purpose. Additionally, MOTD needs more work around text-editing / styling in OpenTTD before it becomes truly useful.

## Description

UPDATE: to make it more clear, renamed everything to "Interface Tutorial", as these tutorials are only meant for the interface parts. Not for gameplay. The caption still reads "Tutorial", as this is more a developers mindset, than something the user needs to know.

To start somewhere, this PR introduces a small tutorial framework for the multiplayer only. Over time we can extend this for other parts, but this gives us the opportunity to explore this a bit before going all-out with tutorials. In-game tutorials should still be done by GSes, but this covers the parts that a GS cannot cover. Basically, "how to use our interface" stuff.

Tutorials are only shown once, the first time you trigger it. In this case, only the first time you join any server. Never again after that. This information is saved in `private.cfg`, so people are less likely to give their tutorial status to other people when copying files.

I kept the framework really small, while still making it already a tiny bit generic. For example, future-work could add that "Close" can also be "Next", and that it chains several steps after each other. Think: "Here you can build trains" -> next -> "And here you can build road", etc. But this for now are only ideas.
It uses highlights to indicate where the user should look, just like a GS Can.

The text for this tutorial needs work, so suggestions are very welcome. Also (constructive) feedback on what you think about this tutorial system is welcome.

image

*!!!* Despite what the screenshot says, the text currently is:
```
You always first join a server as spectator.
As a spectator, you can look around, but not build.
To start playing, use the "Online Players" window to either create your own company or join a friend's.
```

## Checklist for review

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -205,6 +205,9 @@ add_files(
     ini.cpp
     ini_load.cpp
     ini_type.h
+    interface_tutorial_gui.cpp
+    interface_tutorial_gui.h
+    interface_tutorial_type.h
     intro_gui.cpp
     landscape.cpp
     landscape.h
--- a/src/interface_tutorial_gui.cpp
+++ b/src/interface_tutorial_gui.cpp
@@ -1,0 +1,136 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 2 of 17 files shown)


James103:
To improve grammar, the second line of the proposed tutorial can be changed from
> As spectator you can look around, but not build.

To:
> As a spectator, you can look around, but not build.

TrueBrain:
Going to close this PR for now. We agree that such interface tutorial can be useful, but also that we should do it everywhere in that case. And that is a state we will most likely never reach. So leaving this behind for future-us, but not actively pursuing this.

PR #9475, opened 2021-08-14
Remove: COMPANY_INFO packets and related code

### LINKED ISSUE #7268: Suggestion: Option not to disclose server information when password-protected
When a server is password-protected, information about it (list of companies, NewGRF extensions, etc.) are still available through queries.

It would be nice if an option in the server configuration allowed to avoid sending that information through queries, thus non giving away information to not-authenticated clients.

TrueBrain:
## Motivation / Problem

#9467 asked if we could remove the `COMPANY_INFO` packets, as they are now unused (by our clients).

## Description

To get a bit of visual what it involves, I created this PR. I have several reservations whether this is a good idea or not.

- This also removes any other (custom) client to get company info. But to counter that, #7268 has a good point that showing company info on a private server is a bit weird .. and not really what you expect in 2021. So removing this would also work towards resolving #7268.
- `COMPANY_INFO` is part of the DO NOT CHANGE THIS .. which makes me wonder if there is anything I am overlooking, why this would be bad. I noticed I failed to estimate the impact of this.
- Currently you can gather this information from every server in the public listing. That might not have been the intention of this packet, but this removes that ability. It has to be noted that the admin port still has the ability to get the company info, but this of course requires authorization.
- Servers might use this on a banner to show who is on their server, how many trains, etc. They really should be using the admin port for that, but one cannot guarantee someone is using this.

We can also do everything "in between". Only remove the client-side, only remove the server-side, leave it alone.

I am rather conflicted about this, but I just present this PR so we can have a talk.

Edit: the more I think about it, the more I think all cases I mention where people might be using this packet, are illegal use of the packet. Additionally, #7268 really points out this packet should not exist in this form. So I do think we should remove it. Means we can also safely close #7268, as with this PR (combined with all the other work done on multiplayer) it is sufficiently addressed in my opinion.

## Checklist for review

### DIFF
--- a/src/network/core/core.h
+++ b/src/network/core/core.h
@@ -71,8 +71,6 @@ public:
 	 * Reopen the socket so we can send/receive stuff again.
 	 */
 	void Reopen() { this->has_quit = false; }
-
-	void SendCompanyInformation(Packet *p, const struct Company *c, const struct NetworkCompanyStats *stats, uint max_len = NETWORK_COMPANY_NAME_LENGTH);
 };
 
 #endif /* NETWORK_CORE_CORE_H */
--- a/src/network/core/tcp_game.cpp
+++ b/src/network/core/tcp_game.cpp
@@ -73,8 +73,6 @@ NetworkRecvStatus NetworkGameSocketHandler::HandlePacket(Packet *p)
 		case PACKET_SERVER_ERROR:                 return this->Receive_SERVER_ERROR(p);
 		case PACKET_CLIENT_GAME_INFO:             return this->Receive_CLIENT_GAME_INFO(p);
 		case PACKET_SERVER_GAME_INFO:             return this->Receive_SERVER_GAME_INFO(p);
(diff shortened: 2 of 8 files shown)


LordAro (approved):
Shrug.

frosch123 (approved):

YorVeX:
Is it intended that now the response to a PACKET_UDP_CLIENT_FIND_SERVER request is an empty packet of type PACKET_UDP_SERVER_RESPONSE with the packet version field also defaulting to 0, making all clients which had previously implemented parsing that response now think they are getting an outdated packet version? To me it was really confusing. So I think this code in network_udp.cpp is just a leftover and should be removed:
image

I don't see the point in sending an empty response (with an "older" packet version than in previous implementation), just don't respond to that packet at all if you don't want people to be able to get this information anymore through a public UDP request. 

Though I don't really understand what the problem with it was in the first place or why it would have been "illegal" use, there is so many game server implementations including big ones like Half-Life/[Source engine](https://developer.valvesoftware.com/wiki/Server_queries) that give back basic server data in UDP responses.

PR #9479, opened 2021-08-14
Add: open Online Players GUI on starting/joining a server

TrueBrain:
## Motivation / Problem

Basically everyone says: we should open the Online Players GUI when you join a server.

So .. here is the code to make that happen :)

## Description

Needed slightly more code, as by the time the window opens, you haven't received `ClientInfo` yet. So also make sure the GUI doesn't crash because of that.

## Checklist for review

### DIFF
--- a/src/network/network_client.cpp
+++ b/src/network/network_client.cpp
@@ -868,6 +868,8 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::Receive_SERVER_MAP_DONE(Packet
 	/* Say we received the map and loaded it correctly! */
 	SendMapOk();
 
+	ShowClientList();
+
 	/* New company/spectator (invalid company) or company we want to join is not active
 	 * Switch local company to spectator and await the server's judgement */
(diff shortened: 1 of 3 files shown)


src/network/network_client.cpp
```diff
@@ -868,6 +868,8 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::Receive_SERVER_MAP_DONE(Packet
 	/* Say we received the map and loaded it correctly! */
 	SendMapOk();
 
+	ShowClientList();
```
frosch123, line 871: Shuold this be here? Or in the SPECTATOR case below only?
Thinking of cases
* Starting a non-dedicated server yourself.
* Joining a company directly via command line or other options.

TrueBrain: As talked over on IRC:

We could either open it only when you join as spectator, or always. We currently do always, but it is a bit of a toss-up which is better.

frosch123 (approved):

PR #9497, opened 2021-08-18
Fix 659989af45: Set appropriate town window dirty when building/removing airports.

J0anJosep:
## Motivation / Problem

When building or removing airports and airport noise setting is on, the invalidated town window is not right.

The noise generated by an airport is applied to the nearest town to the airport and not to the town associated to the station (see commit 659989af45).

## Description

Invalidate the town window where the noise is really applied to.

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -2359,7 +2359,7 @@ CommandCost CmdBuildAirport(TileIndex tile, DoCommandFlag flags, uint32 p1, uint
 		InvalidateWindowData(WC_STATION_VIEW, st->index, -1);
 
 		if (_settings_game.economy.station_noise_level) {
-			SetWindowDirty(WC_TOWN_VIEW, st->town->index);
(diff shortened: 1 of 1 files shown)


LordAro (dismissed):

src/station_cmd.cpp
```diff
@@ -2436,7 +2437,7 @@ static CommandCost RemoveAirport(TileIndex tile, DoCommandFlag flags)
 		InvalidateWindowData(WC_STATION_VIEW, st->index, -1);
 
 		if (_settings_game.economy.station_noise_level) {
-			SetWindowDirty(WC_TOWN_VIEW, st->town->index);
+			SetWindowDirty(WC_TOWN_VIEW, nearest->index);
```
michicc: I would probably actually move this block to the earlier `DC_EXEC` section to avoid having `nearest` spread out over the function. `SetWindowDirty` isn't doing anything by itself, so the location of the call shouldn't matter at all.

LordAro (approved):

PR #9499, opened 2021-08-18
Fix 2f9c48b: [Linkgraph] Delete links served only by stopped or crashed vehicles

nchappe:
## Motivation / Problem

Links served only by stopped (or crashed) vehicles are not deleted from the linkgraph, so cargoes routed through such a link pile up waiting for vehicles that will never come.

## Description

This PR adds a simple check for stopped and crashed vehicles in the link refresher.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -3688,8 +3688,11 @@ void DeleteStaleLinks(Station *from)
 					auto iter = vehicles.begin();
 					while (iter != vehicles.end()) {
 						Vehicle *v = *iter;
-
-						LinkRefresher::Run(v, false); // Don't allow merging. Otherwise lg might get deleted.
+						/* Do not refresh links of vehicles that have been stopped in depot for a long time. */
+						if (!v->IsStoppedInDepot() || static_cast<uint>(_date - v->date_of_last_service) <=
+								LinkGraph::STALE_LINK_DEPOT_TIMEOUT) {
+							LinkRefresher::Run(v, false); // Don't allow merging. Otherwise lg might get deleted.
+						}
 						if (edge.LastUpdate() == _date) {
 							updated = true;
 							break;
--- a/src/linkgraph/linkgraph.h
+++ b/src/linkgraph/linkgraph.h
@@ -447,6 +447,9 @@ public:
 	/** Minimum effective distance for timeout calculation. */
 	static const uint MIN_TIMEOUT_DISTANCE = 32;
 
+	/** Number of days before deleting links served only by vehicles stopped in depot. */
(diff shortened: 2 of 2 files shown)


michicc:
The commit reference in the commit message feels wrong as this PR only touches the actual code from that commit very indirectly.

I'd also not call it a Fix but a Change, because immediately deleting links with only stopped or crashed vehicles is not as clear cut as it seems. Such a state might be only temporary e.g. because the player stopped all vehicles on a route for some track rework or because they need a few minutes to rebuild a crashed aircraft, and they will not be pleased that all cargo is rerouted.

Ideally, this check would include some kind of timeout, i.e. only delete the link if the state is persisting for some to-be-decided number of ticks/days. No idea if this would be an easy change or cause a complete rework of the link graph, though 😄 .

nchappe:
Thanks for the feedback, I pushed a more conservative version of this PR that focuses on vehicles stopped in depot for a long time. A vehicle in depot is now excluded from the list of vehicles serving a link if it has been more than 1024 days + the total duration of its order list since it last served an order. A delay of 1024 days *after entering the depot* would be safer, but this value is not stored anywhere.

As you pointed out there is always the possibility of a track rework that lasts three years and causes cargo to be re-routed to low-capacity secondary paths, I hope 1024 days of timeout is enough to make this drawback acceptable.

michicc:
I guess you could bikeshed endlessly over the exact time, but 1024 days seems plenty enough for me.

nchappe:
I just found `Vehicle::date_of_last_service`, which is exactly what I needed, so I pushed a cleaner patch.

src/station_cmd.cpp
```diff
@@ -3688,8 +3688,11 @@ void DeleteStaleLinks(Station *from)
 					auto iter = vehicles.begin();
 					while (iter != vehicles.end()) {
 						Vehicle *v = *iter;
-
-						LinkRefresher::Run(v, false); // Don't allow merging. Otherwise lg might get deleted.
+						/* Do not refresh links of vehicles that have been stopped in depot for a long time. */
+						if (!(v->IsStoppedInDepot() && static_cast<uint>(_date - v->date_of_last_service) >
+								LinkGraph::STALE_LINK_DEPOT_TIMEOUT)) {
```
michicc:
```suggestion
						if (!v->IsStoppedInDepot() || static_cast<uint>(_date - v->date_of_last_service) <=
								LinkGraph::STALE_LINK_DEPOT_TIMEOUT) {
```

I think applying some De-Morgan to the condition would make it more reasbale.

nchappe: Done :)

michicc (approved):
Sorry for the pedantry.

PR #9502, opened 2021-08-21
Fix #9501: [Network] crash when more than one game-info query was pending

### LINKED ISSUE #9501: [Crash]: Crashes when Multiplayer Opened
### Version of OpenTTD

1.12.0 - Beta 2  Windows

### Steps to reproduce

1.Opened the Game
2.Pressed Multiplayer 
2.5.Pressed Ok on  "Assertion failed at line 146 of D:\a\OpenTTD\src\network\network_client.cpp:
ClientNetworkGameSocketHandler::my_client == nullptr"
Errorcrash
3.Fatal Application Failure
Errorcrash2

### Upload crash files

[crashDatathingsMatu.zip](https://github.com/OpenTTD/OpenTTD/files/7022914/crashDatathingsMatu.zip)
All of the Crash files

TrueBrain:
## Motivation / Problem

When you have more than one query-for-game-info pending, the game crashes. This is easiest reproduced by starting a local server, and adding `127.0.0.1` and `127.0.0.2` to your server-list. Close the game, restart it, click Multiplayer. Now two queries will be send out at the exact same time.

## Description

We piggy-backed on the client code to, after establishing a TCP connection, query the server. The client code only has a very weird and annoying to follow singleton called `my_client`. As it turns out, when it is being asked to query two different servers at the same time, it panics.

To resolve the situation in a more robust way, I cut out the game-info query part from the client code, and moved it to `network_query`. This part is now responsible solely for querying the game-info of a server, and can have as many parallel connection as needed. It self-registers and self-destroys.

As an additional bonus, this means we can also handle errors from servers in a bit nicer way for when a query fails. This is best seen about the code removed from the client code.

As the protocol reuses other packages to indicate issues, the server can send back 4 different packets depending on the situation on a game-info request:
- `game-info`, what we are after.
- `error`, when something weird went wrong. For example, when it never saw your game-info request packet .. it will hit a timeout and close your connection. But this is really unlikely to ever happen.
- `banned`, when your IP is banned (you won't be able to query the server in that case either; only seems fair).
- `full`, which is already a bug on our bugtracker, but currently at a very early moment in time we validate if the server has room for you. If not, you get this packet back, and refresh fails. This of course has to be resolved, but as for now, this is still happening.

## Checklist for review

### DIFF
--- a/src/network/network_query.cpp
+++ b/src/network/network_query.cpp
@@ -1,0 +1,145 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file network_query.cpp Query part of the network protocol. */
+
+#include "../stdafx.h"
+#include "core/game_info.h"
+#include "network_query.h"
+#include "network_gamelist.h"
+#include "../error.h"
+
+#include "table/strings.h"
+
+#include "../safeguards.h"
+
+std::vector<std::unique_ptr<QueryNetworkGameSocketHandler>> QueryNetworkGameSocketHandler::queries = {};
+
+NetworkRecvStatus QueryNetworkGameSocketHandler::CloseConnection(NetworkRecvStatus status)
+{
+	assert(status != NETWORK_RECV_STATUS_OKAY);
+	assert(this->sock != INVALID_SOCKET);
+
+	return status;
+}
+
+/**
+ * Check the connection's state, i.e. is the connection still up?
+ */
+bool QueryNetworkGameSocketHandler::CheckConnection()
+{
+	std::chrono::steady_clock::duration lag = std::chrono::steady_clock::now() - this->last_packet;
+
+	/* If there was no response in 5 seconds, terminate the query. */
+	if (lag > std::chrono::seconds(5)) {
+		this->CloseConnection(NETWORK_RECV_STATUS_CONNECTION_LOST);
+		return false;
+	}
+
+	return true;
+}
+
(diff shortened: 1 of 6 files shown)


src/network/network_query.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * Check if any query needs to send or receive.
+ */
+/* static */ void QueryNetworkGameSocketHandler::SendReceive()
+{
+	for (auto it = QueryNetworkGameSocketHandler::queries.begin(); it != QueryNetworkGameSocketHandler::queries.end(); /* nothing */) {
+		if (!(*it)->Receive()) {
+			QueryNetworkGameSocketHandler::queries.erase(it);
```
glx22: Missing update of `it` after `erase` ?

src/network/network_query.cpp
```diff
@@ -0,0 +1,147 @@
...
+			continue;
+		}
+		if (!(*it)->CheckConnection()) {
+			QueryNetworkGameSocketHandler::queries.erase(it);
```
glx22: Same here.

src/network/network_query.h
```diff
@@ -0,0 +1,59 @@
...
+
+/** @file network_query.h Query part of the network protocol. */
+
+#ifndef NETWORK_QUERY_H
+#define NETWORK_QUERY_H
+
+#include "network_internal.h"
+
+/** Class for handling the client side of quering a game server. */
+class QueryNetworkGameSocketHandler : public ZeroedMemoryAllocator, public NetworkGameSocketHandler {
```
rubidium42, line 16: Is the `ZeroedMemoryAllocator` really needed here? Only the `ClientNetworkGameSocketHandler` uses it but not the `ServerNetworkGameSocketHandler` so it is not something the `NetworkGameSocketHandler` intrinsicly needs. Since this has only one instance variable that gets in the constructor, I think there is no use for it.

TrueBrain: `ServerNetworkGameSocketHandler` also zero's the object, but this is done by the pool (`Tzero` is set to `true`). So there too `Calloc` is used. In other words: it currently is consistent in zero'ing the object on creation.

Of course these days we can just set the initial values of variables, and possibly get ride of this whole `ZeroedMemoryAllocator`, but that might be best for another PR. For now, I suggest we leave it consistent, and zero the object for all children using `NetworkGameSocketHandler`.

rubidium42 (approved):

Issue #9490, opened 2021-08-17
[Bug]: Refreshing a full server gives an error


ldpl:
### Version of OpenTTD

12.0-beta1

### Expected result

Update server info. I'd like to know if some clients happen to disconnect ;)

### Actual result

Shows "server is full" error message and closes network lobby.

### Steps to reproduce

0. Open multiplayer lobby
1. Find a full server (maxed out on clients) and click on it
2. Click "Refresh server"

TrueBrain:
During `accept()` it is validated if there is still room on the server. This obviously is the wrong place, and it should be done a bit later, always allowing for a `refresh` to happen.

PR #9506, opened 2021-08-23
Fix: [Network] show query errors in the server listing instead of error popup

TrueBrain:
Includes #9502.

## Motivation / Problem

While investigating solutions for #9490, and with the introduction of #9502, I realised how annoying the red popups are when querying servers. It makes much more sense (to me) to show the server status in the info dialog we have for it, similar to "offline".

## Description

As #9502 moved all the query code to its own file, it is a lot easier to add this. Making this PR pretty straight-forward.

image
image

And similar for "Server too old".

To be clear, this is not the final solution for #9490, but at least makes it less icky.

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -407,7 +407,7 @@ protected:
 		DrawString(nwi_name->pos_x + WD_FRAMERECT_LEFT, nwi_name->pos_x + nwi_name->current_x - WD_FRAMERECT_RIGHT, y + text_y_offset, cur_item->info.server_name, TC_BLACK);
 
 		/* only draw details if the server is online */
-		if (cur_item->online) {
+		if (cur_item->status == NGLS_ONLINE) {
 			const NWidgetServerListHeader *nwi_header = this->GetWidget<NWidgetServerListHeader>(WID_NG_HEADER);
 
 			if (nwi_header->IsWidgetVisible(WID_NG_CLIENTS)) {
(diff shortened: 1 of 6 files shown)


src/network/network_gui.cpp
```diff
@@ -637,10 +637,20 @@ class NetworkGameWindow : public Window {
...
 
-			DrawString(r.left + WD_FRAMERECT_LEFT, r.right - WD_FRAMERECT_RIGHT, r.top + detail_height + 4, STR_NETWORK_SERVER_LIST_SERVER_OFFLINE, TC_FROMSTRING, SA_HOR_CENTER); // server offline
+			StringID message = INVALID_STRING_ID;
+			switch (sel->status) {
+				case NGLS_OFFLINE: message = STR_NETWORK_SERVER_LIST_SERVER_OFFLINE; break;
+				case NGLS_FULL: message = STR_NETWORK_SERVER_LIST_SERVER_FULL; break;
+				case NGLS_BANNED: message = STR_NETWORK_SERVER_LIST_SERVER_BANNED; break;
+				case NGLS_TOO_OLD: message = STR_NETWORK_SERVER_LIST_SERVER_TOO_OLD; break;
+
+				case NGLS_ONLINE: break;
```
rubidium42: The break isn't really needed as it can't be reached (unless there's a concurrency issue).

rubidium42 (dismissed):

src/network/network_gui.cpp
LordAro: I would go the other way and remove the default case - let compiler warn you when not all cases are handled
(and then put the `NOT_REACHED` in `NGLS_ONLINE` )

TrueBrain: The absolute worst part .. I intended to do what @LordAro mentions .. but somewhere my brain said: fuck this, and went his own way ;)

rubidium42 (approved):

PR #9508, opened 2021-08-23
Fix #9490: [Network] a full server couldn't be queried either

### LINKED ISSUE #9490: [Bug]: Refreshing a full server gives an error

TrueBrain:
## Motivation / Problem

#9490 shows that querying a full server wasn't possible. In most cases there isn't really a reason to not allow this.

## Description

```
You can now still query a full server, as long as the maximum
amount of allowed connections isn't reached. This means that as
long as there are not 255 clients connected to a server, you can
always connect to query.
```

Basically, you are now allowed to do a `GAME_INFO` packet despite the server being full. `CLIENT_JOIN` is the first packet checking if the server is really full (and drops the connection if it is). All other packets require that `CLIENT_JOIN` was successful, and otherwise they return a `NOT_EXPECTED` error.

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -305,7 +305,7 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::CloseConnection(NetworkRecvSta
 /* static */ bool ServerNetworkGameSocketHandler::AllowConnection()
 {
 	extern byte _network_clients_connected;
-	bool accept = _network_clients_connected < MAX_CLIENTS && _network_game_info.clients_on < _settings_client.network.max_clients;
+	bool accept = _network_clients_connected < MAX_CLIENTS;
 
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):
Still leaves the concurrency issue open that clients_on is checked upon join but only incremented much later, so multiple clients could pass this "gate" before the count in incremented thus essentially allowing more clients to join than the configured maximum.
It's fine to move that to a separate story though.

PR #9511, opened 2021-08-23
Feature: Add selected toolbar buttons to MacBook Pro Touch Bar

ddebruijne:
## Motivation / Problem
I've wanted to implement some functionality on the MacBook Pro's touch bar, and noticed OpenTTD was not using it yet.
 
As this was mainly a personal coding exercise, I don't currently have my expectations set on having this change merged into the mainline branch, but I figured to send in a PR anyways in case the maintainers do want to have it integrated :)
 
## Description
This implements selected in-game toolbar buttons on the MacBook Pro's Touch Bar, the included commit adds the following buttons:
	•	Pause
	•	Fast-Forward
	•	Zoom in/out
	•	Most building windows
 
With this change players could enjoy having these shortcuts on a familar place on the keyboard which is also used by the rest of the OS.
 
openttd_touchbar
 
## Limitations
The player can't customize the touch bar yet. In this commit, it is fixed to instant actions or windows that can replace each other.
 
## Checklist for review


orudge:
The concept seems good to me, but unfortunately the graphics aren't working on my Mac:

touchbar

The buttons do do what they say they should do, however. :)

I suspect some tidying up of parts of the code would be necessary too.

michicc:
It is a bit niche and we don't really have any properly active OSX devs, so the chances aren't that high.

Nevertheless, if it should be included, it would definitely need a pass through https://wiki.openttd.org/en/Development/Coding%20style

ddebruijne:
Thanks for taking a look!
Text appearing instead of the sprite is a fallback in place when it can't generate the button sprite. I'll have a look to see where that could go wrong and patch up the style issues in the process.

ddebruijne:
Looks like the checker didn't like that I made a separate commit - do they need to be squashed?

@orudge I've made a small tweak to the objective-c file which _may_ fix it, but it's quite tough to test as I can't repro what you are seeing. I saw from your photo that you were generating a DMG which I couldn't get working on my machine due to signing issues I believe, maybe that could be related?

In any case, if this doesn't fix it, I'd rather close the PR and not integrate it. If the sprites don't work, it kinda defeats the point of the implementation in my opinion, and is not clean. 😄

michicc:
We would indeed squash a cleanup commit, especially as the commit checker checks each commit individually:

Details: https://github.com/OpenTTD/OpenTTD/runs/3412939438?check_suite_focus=true

orudge:
I'll test it when I get a chance (quite busy at the moment), and if it's still not working, might be able to do some debugging myself.

orudge:
OK, looks like this is now working for me. :)

ddebruijne:
Thanks for checking, appreciate it! 
Squashed the commits and rebased to latest, so hopefully the tester should be pleased with it now too :)

src/gfx.cpp
```diff
@@ -1059,18 +1059,23 @@ void DrawSprite(SpriteID img, PaletteID pal, int x, int y, const SubSprite *sub,
...
+ * @param mode The settings for the blitter to pass.
+ * @param sub Whether to only draw a sub set of the sprite.
+ * @param zoom The zoom level at which to draw the sprites.
+ * @param dst Optional parameter for a different blitting destination.
  * @tparam ZOOM_BASE The factor required to get the sub sprite information into the right size.
  * @tparam SCALED_XY Whether the X and Y are scaled or unscaled.
  */
 template <int ZOOM_BASE, bool SCALED_XY>
-static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mode, const SubSprite * const sub, SpriteID sprite_id, ZoomLevel zoom)
+static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mode, const SubSprite * const sub, SpriteID sprite_id, ZoomLevel zoom, DrawPixelInfo *dst = nullptr)
```
michicc: Use `const DrawPixelInfo`, then you don't need to remove the const below.

src/gfx.cpp
```diff
@@ -1185,6 +1190,44 @@ static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mo
...
+/**
+ * Draws a sprite to a new RGBA buffer (see Colour union) instead of drawing to the screen.
+ * Free the memory after use.
+ *
+ * @param spriteId The sprite to draw.
+ * @return Pixel Buffer
+ */
+const uint32* DrawSpriteToRgbaBuffer(SpriteID spriteId)
+{
+	// Gather information about the sprite to write.
```
michicc: Code style: Use `/* */` for comments except if the comment is at the of a statement line. (More instances of this in this function).

src/gfx.cpp
```diff
@@ -1185,6 +1190,44 @@ static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mo
...
+	const SpriteID real_sprite = GB(spriteId, 0, SPRITE_WIDTH);
+	const Sprite *sprite = GetSprite(real_sprite, ST_NORMAL);
+
+	// Prepare new DrawPixelInfo - Normally this would be the screen but we want to draw to another buffer here.
+	// Normally, pitch would be scaled screen width, but in our case our "screen" is only the sprite width wide.
+	DrawPixelInfo dpi;
+	dpi.dst_ptr = MallocT<uint32>(sprite->width * sprite->height);
```
michicc: This will break when using any 8bpp blitter (like e.g. `8bpp-optimized`), as they output byte-sized palette indices. By default, MacOS will always use one of the 32bpp blitters, but it is still possible for a user to override this.

As such, this situation should be gracefully handled, especially as later on somebody else might trust the function name and use it for something else not specific to MacOS, not realising this limitation.

src/gfx.cpp
```diff
@@ -1185,6 +1190,44 @@ static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mo
...
 }
 
+/**
+ * Draws a sprite to a new RGBA buffer (see Colour union) instead of drawing to the screen.
+ * Free the memory after use.
+ *
+ * @param spriteId The sprite to draw.
+ * @return Pixel Buffer
+ */
+const uint32* DrawSpriteToRgbaBuffer(SpriteID spriteId)
```
michicc: Use `std::unique_ptr` to clearly indicate who owns the return memory. Would require changing MallocT to new[] or adding a custom deleter.

src/gfx.cpp
```diff
@@ -1059,18 +1059,23 @@ void DrawSprite(SpriteID img, PaletteID pal, int x, int y, const SubSprite *sub,
...
  * @tparam SCALED_XY Whether the X and Y are scaled or unscaled.
  */
 template <int ZOOM_BASE, bool SCALED_XY>
-static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mode, const SubSprite * const sub, SpriteID sprite_id, ZoomLevel zoom)
+static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mode, const SubSprite * const sub, SpriteID sprite_id, ZoomLevel zoom, const DrawPixelInfo *dst = nullptr)
 {
 	const DrawPixelInfo *dpi = _cur_dpi;
+	if (dst != nullptr) {
+		dpi = dst;
+	}
```
TrueBrain: `dpi = (dst != nullptr) ? dst : _cur_dpi`
maybe?
Makes it slightly easier to read, all those 3 letter variables are hard to keep apart :)

TrueBrain (commented):
A very small nitpick. Doesn't need fixing, but if there is a need for rebase, might be nice to tackle this too.

Code-wise PR looks good to me. @michicc : do you have any additional comments?

TrueBrain (dismissed):

src/gfx.cpp
```diff
@@ -1185,6 +1190,50 @@ static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mo
 	BlitterFactory::GetCurrentBlitter()->Draw(&bp, mode, zoom);
 }
 
+/**
+ * Draws a sprite to a new RGBA buffer (see Colour union) instead of drawing to the screen.
+ * Returns a nullpointer if a 8bpp blitter is being used.
```
James103: No other occurrence of `nullpointer` exists in the rest of the code base.
Remove this line, and instead replace `@return Pixel Buffer` with `@return Pixel buffer, or nullptr if an 8bpp blitter is being used.`

src/gfx.cpp
```diff
@@ -1185,6 +1190,50 @@ static void GfxBlitter(const Sprite * const sprite, int x, int y, BlitterMode mo
...
+ *
+ * @param spriteId The sprite to draw.
+ * @return Pixel Buffer
+ */
+std::unique_ptr<uint32[]> DrawSpriteToRgbaBuffer(SpriteID spriteId)
+{
+	Blitter *blitter = BlitterFactory::GetCurrentBlitter();
+	if (!blitter->Is32BppSupported()) {
```
michicc: If we are into nitpicking: Maybe single line? (Feel free to ignore)

src/video/cocoa/cocoa_wnd.mm
```diff
@@ -405,6 +405,87 @@ - (instancetype)initWithContentRect:(NSRect)contentRect styleMask:(NSUInteger)st
...
+- (nullable NSTouchBar *)makeTouchBar
+{
+	NSTouchBar *bar = [ [ NSTouchBar alloc ] init ];
+	bar.delegate = self;
+	bar.defaultItemIdentifiers = touchBarButtonIdentifiers;
+
+	return bar;
+}
+
+-(NSImage*)generateImage:(int)spriteId
```
michicc: `NSImage *` (missing space)

src/video/cocoa/cocoa_wnd.mm
```diff
@@ -405,6 +405,87 @@ - (instancetype)initWithContentRect:(NSRect)contentRect styleMask:(NSUInteger)st
...
+	/* Copy the sprite to the NSBitmapImageRep image buffer */
+	const Colour *src = (const Colour *)buffer.get();
+	for (int y = 0; y < fullspr->height; y++) {
+		for (int x = 0; x < fullspr->width; x++) {
+			NSUInteger pixel[4];
+			pixel[0] = src->r;
+			pixel[1] = src->g;
+			pixel[2] = src->b;
+			pixel[3] = src->a;
+			[bitmap setPixel:pixel atX:x y:y];
```
michicc: Missing spaces after/before `[` and `]`.

src/video/cocoa/cocoa_wnd.mm
```diff
@@ -405,6 +405,87 @@ - (instancetype)initWithContentRect:(NSRect)contentRect styleMask:(NSUInteger)st
...
+
+			src += 1;
+		}
+	}
+
+	/* Finally, convert the NSBitmapImageRep we created to a NSimage we can put on the button and clean up. */
+	NSImage* outImage = [ [ NSImage alloc ] initWithSize:NSMakeSize(fullspr->width, fullspr->height) ];
```
michicc: `NSImage *outImage` (position of space)

michicc (commented):
If we are into nitpicking, I have some nits, too. (Nothing in any way serious.)

ddebruijne:
Thanks for the feedback! I'll try to address these points this week 👍🏻

src/video/cocoa/cocoa_wnd.mm
LordAro: Missed this one ;)

LordAro: ...And `pixel[0]` shouldn't have spaces

TrueBrain (approved):

PR #9513, opened 2021-08-27
Change: Use colourblind-friendly gradient for linkgraph

### LINKED ISSUE #9168: Linkgraphs are not colourblind-friendly
## Version of OpenTTD
1.11.1 and earlier.

## Expected result
[According to the NHS](https://www.nhs.uk/conditions/colour-vision-deficiency/), about 1 in 12 men and 1 in 200 women are red/green colourblind.

The linkgraph colours which show if a route is overloaded, saturated, or unused should use colours which can be easily distinguished.

## Actual result
rhd

Linkgraphs use a red-green colour scale. Additionally, the green "saturated" colour matches the temperate grass and is quite difficult to see. This is a separate issue, but is worth considering if we're discussing a colour change.

## Steps to reproduce
1. Start a game with cargo distribution enabled for passengers (symmetric is fine).
2. Connect two towns with a bus route.
3. Open the Cargo Flow Legend from the map icon in the main toolbar, and enable Passengers for your company.

## Suggested fix

JGRPP [includes a setting](https://github.com/JGRennison/OpenTTD-patches/commit/f26143f06319a56739e94ff3513c2909f49c108f) for alternate colour schemes, including green to blue, grey to red, and greyscale.

I am happy to submit a PR with whatever changes we agree on, whether that's upstreaming the setting with multiple choices from JGRPP or changing to a single different colour scheme. Personally, I like grey to red (below).

grey-red

2TallTyler:
## Motivation / Problem

Per #9168:

> [According to the NHS](https://www.nhs.uk/conditions/colour-vision-deficiency/), about 1 in 12 men and 1 in 200 women are red/green colourblind.
> 
> The linkgraph colours which show if a route is overloaded, saturated, or unused currently use a red-green colour scale. Additionally, the green "saturated" colour matches the temperate grass and is quite difficult to see.
>
> rhd

## Description

linkgraph2
_(screenshot of 2nd revision with lighter separation gray line)_

Changes to a new colour scheme using a blue-yellow-red diverging palette, as suggested by [ColorBrewer 2](https://colorbrewer2.org/#type=diverging&scheme=RdYlBu&n=11).

Additionally, the less-than-saturated scale is reversed so that Saturated is approximately in the middle of the colour gradient, instead of jumping from dark green to bright yellow as in the original. Unused is changed from white to gray so as to avoid confusion with the light blue Saturated.

This becomes the default colour scheme, ~~but a new setting is added so players can choose the original red-green scheme.~~ and no setting for the original red-green scheme is available.

The grey line which separates the linkgraph lines in each direction is lightened slightly for visibility against the darkest blue.

I closely based this off the [colorblind-friendly linkgraph colour schemes in JGRPP](https://github.com/JGRennison/OpenTTD-patches/commit/f26143f06319a56739e94ff3513c2909f49c108f), but his do not reverse the top half of the gradient.

Closes #9168.

## Limitations

I originally added a setting allowing players to choose the original green-red colour scheme, but glx22 suggested not adding a setting and simply changing the gradient, so I did that instead.

I did not incorporate any of the colour schemes from JGRPP, since this should work for all [types of colourblindness](https://www.visioncenter.org/conditions/types-of-color-blindness/) and is more visually attractive than his schemes, in my opinion.

The darkest blue of lightly-used links is hard to see on the minimap when the land colour is violet, however this is also true (and worse) when using the red-green palette on green land.

## Checklist for review

### DIFF
--- a/src/linkgraph/linkgraph_gui.cpp
+++ b/src/linkgraph/linkgraph_gui.cpp
@@ -27,9 +27,9 @@
  * "overloaded".
  */
 const uint8 LinkGraphOverlay::LINK_COLOURS[] = {
-	0x0f, 0xd1, 0xd0, 0x57,
-	0x55, 0x53, 0xbf, 0xbd,
-	0xba, 0xb9, 0xb7, 0xb5
+	0x07, 0xc7, 0xc9, 0xcb,
+	0xcd, 0x99, 0x44, 0xbe,
+	0xbb, 0xb9, 0xb7, 0xb5
 };
 
 /**
@@ -285,7 +285,7 @@ void LinkGraphOverlay::DrawContent(Point pta, Point ptb, const LinkProperties &c
 		GfxDrawLine(pta.x, pta.y + offset_y, ptb.x, ptb.y + offset_y, colour, this->scale, dash);
 	}
 
-	GfxDrawLine(pta.x, pta.y, ptb.x, ptb.y, _colour_gradient[COLOUR_GREY][1], this->scale);
+	GfxDrawLine(pta.x, pta.y, ptb.x, ptb.y, _colour_gradient[COLOUR_GREY][2], this->scale);
 }
 
 /**


glx22:
I don't think a setting is required. Blue is definitely easier to see than green, but dark blue seems to have visibility issue with the gray line used to separate the two lines.

2TallTyler:
Thanks for the feedback. I removed the setting, lightened the grey separation line, and updated the screenshot in the original PR description.

SamuXarick:
Hi. I experimented black for the middle line in my tests, and in my opinion it looks better than the grayish tone. So forgive me about my screenshots showcasing it in black, but I would like to report some issues with the choice of some blues.

I think the darkest blue is too dark. I compared it with a black middle line and it's quite hard to distinguish from it.
Also, there is one tone of blue that is so similar to water that it becomes very hard to notice. In the screenshot I only were aware of it because it crosses the ship depot.

Unnamed, 1947-12-18

Unnamed, 1948-08-25

Perhaps another tone of colour could be used, but since I'm not colourblind, I can't really suggest anything.

2TallTyler:
I disagree about the black separator line looking better. It jumps out visually, drawing attention to the wrong part of the linkgraph — the player should notice the link colors, not the separator. 

Yes, the blue does disappear into the water slightly. In my opinion, this is far less problematic than the current green disappearing into the grass. There are a couple other color gradients in the ColorBrewer link above (check the colorblind-friendly filter checkbox) but they are all quite ugly, IMO.

SamuXarick:
I just experimented with these colours, testing shades of milk-coffee brown with shades of pale green, which according to the colourbrewer site, should be colourblind friendly.

```
const uint8 LinkGraphOverlay::LINK_COLOURS[] = {
	7,                          // unused
	99, 100, 101, 102, 103,     // saturated
	59, 58, 57, 56, 55, 54      // overloaded
};
```

`GfxDrawLine(pta.x, pta.y, ptb.x, ptb.y, PC_BLACK, this->scale);`

It looks good, maybe except the darkest brown being a bit too dark. But then I went to try it on Arctic tileset and the lighter shades of green are kind of hard to notice on snow.

I think there will always be at least one colour that will end up going bad with another, no matter what. I'm going to give up on this topic.

Good luck.

jorisdrenth:
Came here via the post on Reddit and as a colourblind user I'm happy to give some feedback and background information on this topic.

First of all I'd like to say thanks for keeping colourblind players in mind when designing the game UI. A lot of game developers don't and some games are impossible for me to play because of that.

As hard as it is for you to understand colourblindness, it is for me to explain it. I see a lot of situations where non-colourblind's advise other non-colourblind's on colour usage and it makes no sense to me. Recently one of the developers at my work changed some colours of buttons on our website because some source said it would be better for colourblind users, but instead it made it worse for me. I have red-green colourblindness and so has my brother. We often have the same colour problems when we play a game together, but sometimes there's something I can't see but he can and vice versa. And I think that's the most important thing to keep in mind: no person's colourblindness is the same, even if it is the same type.

Besides that I think you should understand colour plays a different role in my life than (I think) it does in non-colourblind people's life. In the 32 years that I'm colourblind, my brain is programmed to _not_ use colour as a source to get information, as it's simply not a factor I can rely on. If family or friends come over I can almost always tell you their license plate, but it's very hard for me to remember the colour of their car.

If your definition of the problem is "a colourblind user can't distinguish the colours", changing the colour pallette would look like a viable fix. But you will never be able to fix it for all colourblind users, because the problem is different for everyone. 

Instead I would say the definition of the problem is "the only way to distinguish information is through colour". If you would fix that, it will be fixed for all colourblind users.

If you really want to make a usable interface for all colourblind people, in my opinion there are two options:

1. Give some other visual clue to distinguish information than just colour. I read a comment on Reddit by a user who suggested lines in various dotted styles, that would be great. Or you could add symbols like circles, squares and triangles to the lines.
2. Give the option to choose the colours myself. This gives the user a way to set the colours as it works for them.

That being said, I know both options would require a lot more modifications to the game than just changing the colour pallette. The blue-red one is better for me than the green-red one. But whatever colours you choose, when all different colours are next to each other in order (as in the game window in the screenshot) I can see all of them. But when I only see a line of one of them I couldn't tell which one it exactly is. Therefor I would only know if it's overloaded or not, but I wouldn't know if it's overloaded a lot or just a little bit.

Hope this gives you some insights on colourblind users. Let me know if you want more info!

2TallTyler:
[A helpful comment](https://www.reddit.com/r/openttd/comments/pttf7h/comment/hdyhbdu/?utm_source=share&utm_medium=web2x&context=3) from u/_LV426 on reddit:

> so I have Protonopia (green-reds and pinks-purple-blues tend to merge and look mucky/dull).
> 
> Generally with OTTD I've not really had that much of an issue with the ui; some of the player colour options look quite identical but I just choose other colours. From playing other games I would say the better attempts at providing colour blind support are when colour is used in conjunction with icons or text. Battefield 4 stands out in my memory as well as Planetside 2 as having good support but as they are both FPS games there is quite a lot less information to have to display.
> 
> In the included screenshot with this post for my eyes in the left hand saturation chart key the 2nd and 3rd colour bars from the bottom are quite close to each other. Along the top row of cargo list, the RM and WS are quite close in hue for me as well – I can tell there's a difference but only because the grey table gives a background, on the world screen I doubt I could tell the difference.
> 
> So (knowing nothing of what's possible here I will just spitball...) I would suggest combining some other form of differentiation – could the lines on the map change their styling to include dashes, dots, long-short, short-long-short dashes, etc. to help differentiation? Could the cargo abbreviation be included with the lines on the world screen? Could fill patterns be included? (little checker patterns, diagonal lines, etc.) A way to test would if it works in greyscale it should work for everyone.
> 
> As an aside the fullscreen filter option some games include doesn't really work for me – it's quite an obvious "desaturation" filter if that makes sense. DooM and WoW do this. Hard to explain to people who don't have colourblindness I always find :P Although I may not be able to see the full gamut I can tell when colours are "missing" or dulled down.
> 
> Edit: I should add that obviously everyone sees differently so the more customisable it could be made the better really :)

2TallTyler:
I'm closing this in favor of #9866 and #9760. I think they add the supplemental information and choice that this PR lacks.

Issue #8001, opened 2020-02-14
Ship finds path but reports it is lost


debdog:
## Version of OpenTTD
1.10.0-RC1 and master-g2196cd3cf8
Pathfinder: YAPF
OS: Devuan ASCII (conforms Debian Stretch)

## Expected result
Ship finds path and all is well.

## Actual result
Ship does find path but a message appears "Ship # is lost".
This seems to happen on longer distance routes only and where the ships do not have to alter course much. On short distances this does not happen.

## Steps to reproduce
The enclosed save game provides an example. Ships 1 to 3, on the long route, report being lost each time they leave the first waypoint in the Orders list.
Ship 4, on the short route, never issues such a message.

Let me know what more info is required and I'd be happy to provide it.

[ships_temperate.sav.txt](https://github.com/OpenTTD/OpenTTD/files/4206689/ships_temperate.sav.txt)

LordAro:
So I think this is happening because FindPath (yapf_base.hpp) is returning that it didn't find a path, but is nevertheless returning a path - possibly as far as it can find? YAPF has never been the clearest thing in the world...

Not clear what should be done here - if the pathfinder can't find a complete path, there's no way to know whether the destination is just too far away, or it genuinely can't find a route.

SamuXarick:
[Tuningville Transport, 1973-04-22.zip](https://github.com/OpenTTD/OpenTTD/files/4246361/Tuningville.Transport.1973-04-22.zip)
On this savegame, with default pf.yapf.ship_curve90_penalty = 600, ship is lost, when going from oil rig to dock. If the penalty is 100, ship won't be lost.

90 degree penalties are a recent addition as well as docking tile penalties, but it seems they're both causing the required number of search nodes to find a path to be substantially increased, going past the default limit of pf.yapf.max_search_nodes = 10000.

SamuXarick:
some test results - if rounds needed are above 10000, the ship will report it's lost
```
ship_curve45_penalty = 100 | 0     | 100   | 0
ship_curve90_penalty = 600 | 600   | 100   | 0
---------------------------+-------+-------+------

-Savegame 2: Tuningville Transport, 1973-04-22.sav
rounds needed:       13899 |  9589 |  5070 |  3641    (added docking tile cost)
rounds needed:       13341 |  7870 |  4943 |  3172  (removed docking tile cost, with PR) 

-Savegame 1: ships_temperate.sav (with 1 ship docked at the destination dock)
[...]

PR #9514, opened 2021-08-27
Change: Cap docking tile occupancy penalty to a max of one ship

SamuXarick:
## Motivation / Problem
Main problem is that there is no cap to 'docking tile occupancy penalty' - if there are 50 ships on the docking tile, the cost is ~50 times higher.

The higher the penalty/cost, the more likely the chances for the pathfinder to report ship is lost for other ships that are part of this route. If there's lots of water near the docking tile which is also the pf's destination, it will start looking for those cheaper nodes first, resulting in more search nodes being needed to find a path.

It really seems like a waste to search nodes around the destination. It's understandable that the penalty is being used to help send the ships to other docking tiles of the same multi-dock station, but... oh well.

Also refer to #8001 

## Description
This PR limits the penalty cost to a max of one ship in the docking tile. There's the cost of the tile being a docking tile which is 300 in YAPF, and then there's the cost of it being occupied by ships, which is also 300 in YAPF. So, at max, it's an added 600 cost now, instead of a cost that was dependent on the number of ships in the tile.

## Limitations
This PR may help mitigate 'ship is lost' messages which are triggered when the max number of search nodes is reached by the pf, but in all honesty, it feels insufficient still to what the game used to be before this penalty ever existed.

## Checklist for review

### DIFF
--- a/src/pathfinder/npf/npf.cpp
+++ b/src/pathfinder/npf/npf.cpp
@@ -305,11 +305,8 @@ static void NPFMarkTile(TileIndex tile)
 
 static Vehicle *CountShipProc(Vehicle *v, void *data)
 {
-	uint *count = (uint *)data;
 	/* Ignore other vehicles (aircraft) and ships inside depot. */
-	if (v->type == VEH_SHIP && (v->vehstatus & VS_HIDDEN) == 0) (*count)++;
-
-	return nullptr;
+	return v->type == VEH_SHIP && (v->vehstatus & VS_HIDDEN) == 0 ? v : nullptr;
 }
 
 static int32 NPFWaterPathCost(AyStar *as, AyStarNode *current, OpenListNode *parent)
@@ -331,7 +328,7 @@ static int32 NPFWaterPathCost(AyStar *as, AyStarNode *current, OpenListNode *par
 	if (IsDockingTile(current->tile)) {
 		/* Check docking tile for occupancy */
 		uint count = 1;
-		HasVehicleOnPos(current->tile, &count, &CountShipProc);
+		if (HasVehicleOnPos(current->tile, nullptr, &CountShipProc)) count++;
 		cost += count * 3 * _trackdir_length[trackdir];
 	}
 
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -266,11 +266,8 @@ public:
 
 	static Vehicle *CountShipProc(Vehicle *v, void *data)
 	{
-		uint *count = (uint *)data;
 		/* Ignore other vehicles (aircraft) and ships inside depot. */
-		if (v->type == VEH_SHIP && (v->vehstatus & VS_HIDDEN) == 0) (*count)++;
-
-		return nullptr;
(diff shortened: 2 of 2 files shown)


TrueBrain:
I am not sure what this PR fixes, and I think this is a bad change. It basically "soft" reverts the intention of the code, without any real argumentation why that would be better.

I am guessing this is an attempt to fix #8001, but this is not mentioned. So I am collecting context here .. I might misunderstand all of this :P

> The higher the penalty/cost, the more likely the chances for the pathfinder to report ship is lost for other ships that are part of this route. If there's lots of water near the docking tile which is also the pf's destination, it will start looking for those cheaper nodes first, resulting in more search nodes being needed to find a path.

Yes, that is exactly what Path Finders do. This is not an issue, this is their purpose in life.

> It really seems like a waste to search nodes around the destination. It's understandable that the penalty is being used to help send the ships to other docking tiles of the same multi-dock station, but... oh well.

You contradict yourself. It is not a waste, it makes multi-dock stations useful (as otherwise they are not useful, what the original change set out to fix). That you do not like it, doesn't make it "a waste". Or if there is an other argumentation here, I do not see it.

> This PR limits the penalty cost to a max of one ship in the docking tile. There's the cost of the tile being a docking tile which is 300 in YAPF, and then there's the cost of it being occupied by ships, which is also 300 in YAPF. So, at max, it's an added 600 cost now, instead of a cost that was dependent on the number of ships in the tile.

You explain what it does, which the code tells me too. But this is not a motivation. Let me put it in different words:

- "Please explain if you want to go left or right on this street"
- "I am going left, as left is the direction I am going"

Motivation explains why you are going left. Not the fact that you do.

I have no clue why you picked "1" over "2" or any other value for example. Or why you think capping it is the way to go.

As counter suggestion, I would think that some diminishing return value would be better. I can understand that `300 + N * 300` might be a bit much. But there is a difference between 1 and 2 ships at a dock. So maybe the second ship shouldn't be 300, but neither should it be 0.

TrueBrain:
Right, there has been some talk on IRC about this, lets sync up the information:

- The author of this PR didn't consider multi-docks.
- This PR seems to dismiss multi-docks all together, and soft-reverts it back to before balancing between docks was introduced. This basically means this PR is a regression.
- This PR seems to scratch an itch ("it is wasting CPU"), and not really fixes anything. It is properly called "Change", but what it actually changes is rather unclear.
- This PR does not address #8001. Plenty of other scenarios can be created where a ship is still considered lost.

Additionally, the strong suggestion made on IRC is that the "ship is lost" part is in a different part of the code. Looking back to what I wrote yesterday, it is also a bit hard to believe that adding a penalty of a single tile for each ship docked makes a difference in whether a ship is lost or not. The intention is exactly that you give the pathfinder more and more room to find other docking tiles.

In conclusion I think it is fair to say this PR is not reaching its goal, and as such I am going to close it up now.

Additional information and analysis is needed and very welcome.
To not pollute one solution with the other, please create a new PR for this; also please in such PR try to explain what cases you considered (and what you dismisses), and what you think it is actually fixing. That would greatly speed up the review process.

Lastly, I do appreciate the attempt to address these issues, and at least this PR sparked a conversation.

PR #9516, opened 2021-08-28
Codechange: remove unused strings

TrueBrain:
## Motivation / Problem

I assumed we had strings that were no longer used in our language files, as it is pretty easy to miss this. We do not validate if strings are used.

After trying to write some scripting to figure out if this is the case, I ended up figuring out how poorly we name our strings, and the weird hacks we have to reference a string.

For example, the setting-flag "zero-is-special" uses the next string after the "value" string. In other words, this special string is not referenced directly, but you have to calculate that based on another.

And it is not always in the forward direction. You can say you want to draw a "tiny" variant of a string, which is "- 1" from the original string.

After fiddling a while with a script I wrote, I finally managed to get through all the strings, and detect which are unused. Another PR will upstream that script, so we can link it to the CI, but for now, first a bit of cleanup.

## Description

Removing string is tricky, as it is hard to tell if they really aren't used. So I have been doing some digging, but please confirm my findings:

- `STR_GAME_OPTIONS_REFRESH_RATE_OTHER`: introduced in 0464a50ab8, but no reference to the string is being made.
- `STR_CONFIG_SETTING_DISABLE_UNSUITABLE_BUILDING`: removed in 188bf0fbc9.
- `STR_NETWORK_SERVER_LIST_LANGUAGE`: removed in 05612d60ae.
- `STR_NETWORK_START_SERVER_LANGUAGE_SPOKEN` and friend: removed in 05612d60ae.
- `STR_NETWORK_SERVER` and `STR_NETWORK_CLIENT`: removed in 5266359424.
- `STR_NETWORK_ERROR_NNN`: no specific commits, but none of those are referenced indirectly from what I can tell, so it is fairly safe to assume they got removed "over time".
- `STR_CONTENT_ERROR_COULD_NOT_DOWNLOAD_CONNECTION_LOST`: removed in eeb38a8e3a.
- `STR_FRAMERATE_BYTES_WARN` and friend: seems to be added to complete the list, but actually never used (see e7f6f07599e).
- `STR_NEWGRF_SETTINGS_PRESET_SAVE_QUERY`: replaced with another window, never removed (see 8755c26793).
- `STR_STATION_VIEW_EN_ROUTE_FROM`: removed in 0fc198cb005.
- `STR_VEHICLE_NAME_BUTTON`: removed in 589feba0eb.
- `STR_ERROR_INCOMPATIBLE_TRAMWAY`: introduced in c02ef3e456, but no reference to the string is being made.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1032,7 +1032,6 @@ STR_GAME_OPTIONS_GRAPHICS                                       :{BLACK}Graphics
 
 STR_GAME_OPTIONS_REFRESH_RATE                                   :{BLACK}Display refresh rate
 STR_GAME_OPTIONS_REFRESH_RATE_TOOLTIP                           :{BLACK}Select the screen refresh rate to use
-STR_GAME_OPTIONS_REFRESH_RATE_OTHER                             :other
 STR_GAME_OPTIONS_REFRESH_RATE_ITEM                              :{NUM}Hz
 STR_GAME_OPTIONS_REFRESH_RATE_WARNING                           :{WHITE}Refresh rates higher than 60Hz might impact performance.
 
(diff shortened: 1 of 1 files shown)


glx22:
Seems correct, but I prefer to wait for a third person confirmation :)

rubidium42 (approved):

PR #9524, opened 2021-08-30
Fix #9521: Don't load at just removed docks that were part of a multi-dock station

### LINKED ISSUE #9521: [Bug]: on a multidock station, if i remove one of the docks while there's a ship loading/unloading, it still continues to load/unload
### Version of OpenTTD

12.0-beta2

### Expected result

Ship should stop loading/unloading immediately after removing the dock it's unloading at.

### Actual result

Ship continues to load/unload at the multi-dock station even after the dock it's at is removed.

### Steps to reproduce

[Vorbrücken Transport, 1975-12-02.zip](https://github.com/OpenTTD/OpenTTD/files/7072372/Vorbrucken.Transport.1975-12-02.zip)
Savegame shows a ship unloading/loading at a just demolished dock.

SamuXarick:
## Motivation / Problem
Fixes #9521

Given the scenario with a station with 2 (or more) docks (with 2 docking tiles), if the ship is loading/unloading at one of the docks, and that dock is demolished, the ship still continues to load/unload, because the station facility `FACIL_DOCK` still exists.
It should instead, make the ship go to the other dock of that same station.

## Description
Problem is solved by also checking if the docking tile where the ship is positioned at is still valid and belong to the station entity.
Savegame: 
[multidock 2.zip](https://github.com/OpenTTD/OpenTTD/files/7077673/multidock.2.zip)
In this savegame, ship 1 and ship 3 are unloading at Würzbrücken Docks which consist of 2 docks with 2 docking tiles.
Würzbrucken Valley is another dock that shares the same docking tile as that where ship 3 is located.
Tests:
1 - Removing Würzbrücken Valley dock should not affect ship 3 unloading at the other dock.
2 - Removing one of the Würzbrücken Docks should make the ship that it's at the respective docking tile to leave station and head to the other dock nearby that still belongs to the same station.

## Limitations
- There's a bit of code repetition.
- I am always assuming `tile3` is a docking tile and is retrieved in that manner. Maybe there is a better way to retrieve which tiles are the docking tiles.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -2720,19 +2720,24 @@ static CommandCost RemoveDock(TileIndex tile, DoCommandFlag flags)
 		ClearDockingTilesCheckingNeighbours(tile1);
 		ClearDockingTilesCheckingNeighbours(tile2);
 
-		/* All ships that were going to our station, can't go to it anymore.
-		 * Just clear the order, then automatically the next appropriate order
-		 * will be selected and in case of no appropriate order it will just
-		 * wander around the world. */
-		if (!(st->facilities & FACIL_DOCK)) {
-			for (Ship *s : Ship::Iterate()) {
-				if (s->current_order.IsType(OT_LOADING) && s->current_order.GetDestination() == st->index) {
-					s->LeaveStation();
-				}
+		for (Ship *s : Ship::Iterate()) {
+			/* Find all ships going to our dock. */
+			if (s->current_order.GetDestination() != st->index) {
+				continue;
+			}
 
-				if (s->current_order.IsType(OT_GOTO_STATION) && s->current_order.GetDestination() == st->index) {
-					s->SetDestTile(s->GetOrderStationLocation(st->index));
-				}
+			/* Find ships that are marked as "loading" but are no longer on a
+			 * docking tile. Force them to leave the station (as they were loading
+			 * on the removed dock). */
+			if (s->current_order.IsType(OT_LOADING) && !(IsDockingTile(s->tile) && IsShipDestinationTile(s->tile, st->index))) {
+				s->LeaveStation();
+			}
+
+			/* If we no longer have a dock, mark the order as invalid and send
+			 * the ship to the next order (or, if there is none, make it
+			 * wander the world). */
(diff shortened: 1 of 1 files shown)


TrueBrain:
Although you seem to have traced the place where things go wrong, your solution is both hard to follow and possibly prune for errors. If the logic gets that complicated, it just becomes tricky.

The main issue I have with your code, is that `tile3` is in, what, 90% of the cases false. Only if two docks of the same stations have overlap, it can yield true. In other words, the `for (Ship..)` is executed in most cases. So to remove complexity, we can basically just remove that if-statement.

Next, `(!a || !b)` is the same as `!(a && b)`, which in this case is a lot more readable.

Also, a station without a dock is similar to the docking tile no longer belonging to the station. To reduce complexity, I suggest we simply remove the "does station have dock" check in favour of "does this docking-tile belong to the station". One implies the other, and it removes yet-another-check.

Lastly, the comment in front of the `if` was now heavily out-of-date. Fixing that comment would be really difficult. So instead, I suggest something like the following:

```
		for (Ship *s : Ship::Iterate()) {
			/* Find all ships going to our dock. */
			if (s->current_order.GetDestination() != st->index) {
				continue;
			}

			/* Find ships that are marked as "loading" but are no longer on a
			 * docking tile. Force them to leave the station (as they were loading
			 * on the removed dock). */
			if (s->current_order.IsType(OT_LOADING) && !(IsDockingTile(s->tile) && IsShipDestinationTile(s->tile, st->index))) {
				s->LeaveStation();
			}

			/* If we no longer have a dock, mark the order as invalid and send
			 * the ship to the next order (or, if there is none, make it
			 * wander the world). */
			if (s->current_order.IsType(OT_GOTO_STATION) && !(st->facilities & FACIL_DOCK)) {
				s->SetDestTile(s->GetOrderStationLocation(st->index));
			}
		}
```

This still fixes the problem (as far as I could tell), removes the need for `tile3`, and writes a story over multiple comments. Each statement is also easier to follow:

- Does this ship go to this station? No? Skip.
- Is it currently loading on a tile that doesn't belong to the station anymore? Leave the station.
- Are we going to the station, but there are no docks anymore? Skip to next order.

TrueBrain (approved):

PR #9526, opened 2021-08-31
Fix #9407: desync when founding a town nearby a station

### LINKED ISSUE #9407: [Bug]: CmdFoundTown may cause multiplayer desyncs due to not initialising Town::stations_near

TrueBrain:
## Motivation / Problem

When a town is founded during a game, we misuse `_generating_world` to bypass a bunch of validation that normally should happen when towns grow etc.
This is also done to skip updating `stations_near`, as it is a pretty expensive operation, and when generating a new map there is of course never a station around a town to start with. So this is safe to skip in that scenario. Not when founding a new town.

As I wrote in the commit message:

```
"stations_near" wasn't updated when founding a town near
a station. As this variable is not saved, any client joining
after the town is founded has a different value for
"stations_near", potentially causing desyncs.
```

This was reported in #9407.

## Description

JGR also made a solution for this (https://github.com/JGRennison/OpenTTD-patches/commit/b23ba0c6c0f5be9ea4d4239478b39be42fe81ee6), and although effective, it felt a bit like bringing a gun to a pillow fight.
I was also mostly wondering why `_generating_world` was misused like this, and why it would skip the `stations_near` update.

Long story short: the check on `_generating_world` is an optimization, to avoid wasting a lot of CPU on something that is never going to happen during map generation. So I looked for the next-best-thing that does the same but differently, and settled on checking if there is any station at all. During map generation of course there isn't.

In result, after this PR the code does the same for new map generation (skipping the expensive station check), but fixes the bug in #9407.

JGR's solution is also a fine solution, but it introduces several globals and more complexity that can break over time. On the other hand, his solution might fix issues we were not even aware of. I leave it to the reviewer to balance this.

## Checklist for review

### DIFF
--- a/src/station_base.h
+++ b/src/station_base.h
@@ -565,6 +565,9 @@ void RebuildStationKdtree();
 template<typename Func>
 void ForAllStationsAroundTiles(const TileArea &ta, Func func)
 {
+	/* There are no stations, so we will never find anything. */
+	if (Station::GetNumItems() == 0) return;
+
 	/* Not using, or don't have a nearby stations list, so we need to scan. */
 	std::set<StationID> seen_stations;
 
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -2310,12 +2310,10 @@ static void MakeTownHouse(TileIndex t, Town *town, byte counter, byte stage, Hou
 	if (size & BUILDING_2_TILES_X)   ClearMakeHouseTile(t + TileDiffXY(1, 0), town, counter, stage, ++type, random_bits);
 	if (size & BUILDING_HAS_4_TILES) ClearMakeHouseTile(t + TileDiffXY(1, 1), town, counter, stage, ++type, random_bits);
 
-	if (!_generating_world) {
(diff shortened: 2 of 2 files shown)


JGRennison:
The `Station::GetNumItems() != 0` check could be moved inside `ForAllStationsAroundTiles` perhaps?
The other call sites of `ForAllStationsAroundTiles` can also be called when there are no stations. (`PopulateStationsNearby` for industries also has a similar `_generating_world` check).

The reason for the slightly more complicated implementation in my branch it to avoid repeating the station tile scan for every new house in the founded town, and the catchment area scan for each found station x each house which found it. The scans only need to be done once after the town is founded. I did not make any measurements at the time but it may be worth bearing in mind.

LordAro (approved):

PR #9529, opened 2021-08-31
Fix #9527: Crash when trying to place multitile objects at map edge

### LINKED ISSUE #9527: [Crash]: NewGRF with multi-tile object causes OpenTTD 12.0-beta2 to crash on map generation
### Version of OpenTTD

OpenTTD 12.0-beta2

### Steps to reproduce

1. Create a simple 2x2 object in nml with the following switch (or use attached NewGRF)

```
switch (FEAT_OBJECTS, SELF, switch_sprite_layouts_rocks, relative_pos) {
		relative_coord(0, 0):			spritelayout_rocks;
		relative_coord(0, 1):			spritelayout_rocks;
		relative_coord(1, 0):			spritelayout_rocks;
		relative_coord(1, 1):			spritelayout_rocks;
}
```
2. Build the NewGRF with NML 0.6.0

3. Start a new game

4. OpenTTD will crash during map generation
image

The same NewGRF works properly in OpenTTD 1.10.2.

### Upload crash files

[crash.zip](https://github.com/OpenTTD/OpenTTD/files/7084901/crash.zip)
[multitile_object.zip](https://github.com/OpenTTD/OpenTTD/files/7084953/multitile_object.zip)

LordAro:
## Motivation / Problem
#9527

Building objects did not do any bounds checking. Fine for 1x1 objects, less so for others

Also spotted an issue with a previous fix in the same area

## Description

Add a check of all the tiles for validity. Not sure if there's a better way.

Also fix the check for zero-size objects, which was missing a bitshift (32 != 0). and bump the grf message up a level (to 0), as grf makers will probably want to see it
Example of issue: `0xF0 & 0x22 === 0x20 === 32 !== 2` Therefore if the lower nibble is 0, it would not be correctly detected as such.

## Checklist for review

### DIFF
--- a/src/object_cmd.cpp
+++ b/src/object_cmd.cpp
@@ -220,6 +220,9 @@ CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint3
 	int size_x = GB(spec->size, HasBit(view, 0) ? 4 : 0, 4);
 	int size_y = GB(spec->size, HasBit(view, 0) ? 0 : 4, 4);
 	TileArea ta(tile, size_x, size_y);
+	for (TileIndex t : ta) {
+		if (!IsValidTile(t)) return_cmd_error(STR_ERROR_TOO_CLOSE_TO_EDGE_OF_MAP_SUB); // Might be off the map
+	}
 
 	if (type == OBJECT_OWNED_LAND) {
 		/* Owned land is special as it can be placed on any slope. */
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -4085,8 +4085,8 @@ static ChangeInfoResult ObjectChangeInfo(uint id, int numinfo, int prop, ByteRea
 
 			case 0x0C: // Size
 				spec->size = buf->ReadByte();
-				if ((spec->size & 0xF0) == 0 || (spec->size & 0x0F) == 0) {
-					grfmsg(1, "ObjectChangeInfo: Invalid object size requested (%u) for object id %u. Ignoring.", spec->size, id + i);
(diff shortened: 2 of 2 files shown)


src/object_cmd.cpp
```diff
@@ -220,6 +220,9 @@ CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint3
 	int size_x = GB(spec->size, HasBit(view, 0) ? 4 : 0, 4);
 	int size_y = GB(spec->size, HasBit(view, 0) ? 0 : 4, 4);
 	TileArea ta(tile, size_x, size_y);
+	for (TileIndex t : ta) {
+		if (!IsValidTile(t)) return CMD_ERROR; // Might be off the map
```
James103: Could this be changed to any of the two following lines?
```c++
if (!IsValidTile(t)) return_cmd_error(STR_ERROR_OFF_EDGE_OF_MAP); // Might be off the map
```
or alternatively,
```c++
if (!IsValidTile(t)) return_cmd_error(STR_ERROR_TOO_CLOSE_TO_EDGE_OF_MAP); // Might be off the map
```

TrueBrain (approved):

JGRennison:
The existing zero-object size check is correct. Other than style issues there doesn't appear to be anything being fixed?

> Also fix the check for zero-size objects, which was missing a bitshift (32 != 0)

> Therefore if the lower nibble is 0, it would not be correctly detected as such.

I don't see how these follow from the existing code. Both nybbles are checked separately, and a bitshift has no effect.


(I've got no objections to this PR, incidentally).

LordAro:
Yes, I just became very confused about what it was doing. Commits have since been updated to reflect that it was just a codechange, but PR description had not

PR #9067, opened 2021-04-19
Feature: rework in-game Network GUI

TrueBrain:
Fixes #9035.

## Motivation / Problem

Please follow me down a rabbit hole:

For an upcoming feature we need to show server owners an "invite code" which they can hand to others to join their (private) server. This requires a place to show this "invite code".
The most sane place for this is the "Multiplayer GUI", which currently is just a client-list.

This was the only problem I wanted to resolve .. sadly .. I had a case of feature-creep.

While thinking about how to solve this, some additional things came to mind:
- #9035 rightfully mentions there is no obvious name to rename yourself in-game.
- Chatting with people is not obvious in the current GUI.
- What clients are in what company is not obvious in the current GUI.
- There is no way to "reset" (read: delete) a company from the GUI (as server owner).
- Renaming your server in-game is non-trivial.
- Changing your server from advertised to non-advertised or the other way around is non-trivial.

Basically, we made a lot of things possible via the console, but not via the GUI. So, after some internal debating, I finally made the call to do something about it.

| |    |
|  :----------: | -----------: |
|  Host Overview | image |
| Client overview (no server info) | image |
| Highlight hover | image |
| Company admin (old screenshot) | image |
| Client admin (old screenshot) | image |
| Spectator (sees "new company" line) | image |
| Confirmation dialog (old screenshot) | image |

I really did my best to scope this PR as small as possible. That is to say, I tried to add as few new features as possible, but mainly make the new GUI work better and make it more extendable towards the future. Later PRs can add new functionalities. We have a wishlist ;)

## Description

When I wrote the original Multiplayer GUI (client-list), I distinctly remember me saying: this GUI is so ugly and unfinished, sure someone will fix that up soon, as I really hate building GUIs.
Well, now 17 years later, I have the answer for you: no, nobody did. So yeah .. I spend 3 days of my life fighting the GUI system. Not because it is bad or what-ever, but because what I want it cannot deliver. So there was a constant fight going on. I think I somewhat won .. but I leave you to be the judge of that.

```
The GUI now more clearly shows some basic information about the
server you joined, your client name (and the ability to change it),
and what players are in which company.

It also contains useful buttons to press to join companies, chat
with other people, and for admins to kick/ban people.

Additionally, renamed "advertised" to "visibility"; this has to
do with future additions, but also because it is more clear in
wording.
```

[...]

PR #9532, opened 2021-09-02
Change: Return 'New company/Spectate' option to company toolbar menu

ldpl:
## Motivation / Problem

Somehow this option was removed in #9067 and since I couldn't find any reason for that change and I'm adding it back in cmclient anyway I thought I may as well PR it to vanilla. 

It's a convenient shortcut that was already there and doesn't have much to do with server lobby merging with the client list.

## Description

Returns "New company/Spectate" option to the company menu ("bald guy" button) in the main toolbar.

## Checklist for review

* *This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)*

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2107,6 +2107,8 @@ STR_NETWORK_NEED_COMPANY_PASSWORD_CAPTION                       :{WHITE}Company 
 
 # Network company list added strings
 STR_NETWORK_COMPANY_LIST_CLIENT_LIST                            :Online players
+STR_NETWORK_COMPANY_LIST_SPECTATE                               :Spectate
+STR_NETWORK_COMPANY_LIST_NEW_COMPANY                            :New company
 
 # Network client list
 STR_NETWORK_CLIENT_LIST_CAPTION                                 :{WHITE}Online Players
--- a/src/toolbar_gui.cpp
+++ b/src/toolbar_gui.cpp
@@ -204,7 +204,9 @@ static void PopupMainToolbMenu(Window *w, int widget, StringID string, int count
 
 /** Enum for the Company Toolbar's network related buttons */
 static const int CTMN_CLIENT_LIST = -1; ///< Show the client list
-static const int CTMN_SPECTATOR   = -2; ///< Show a company window as spectator
+static const int CTMN_SPECTATE    = -2; ///< Become spectator
+static const int CTMN_NEW_COMPANY = -3; ///< Create a new company
+static const int CTMN_SPECTATOR   = -4; ///< Show a company window as spectator
 
 /**
  * Pop up a generic company list menu.
@@ -222,8 +224,13 @@ static void PopupMainCompanyToolbMenu(Window *w, int widget, int grey = 0)
 
 			/* Add the client list button for the companies menu */
 			list.emplace_back(new DropDownListStringItem(STR_NETWORK_COMPANY_LIST_CLIENT_LIST, CTMN_CLIENT_LIST, false));
-			break;
 
+			if (_local_company == COMPANY_SPECTATOR) {
+				list.emplace_back(new DropDownListStringItem(STR_NETWORK_COMPANY_LIST_NEW_COMPANY, CTMN_NEW_COMPANY, NetworkMaxCompaniesReached()));
+			} else {
+				list.emplace_back(new DropDownListStringItem(STR_NETWORK_COMPANY_LIST_SPECTATE, CTMN_SPECTATE, false));
+			}
+			break;
 		case WID_TN_STORY:
 			list.emplace_back(new DropDownListStringItem(STR_STORY_BOOK_SPECTATOR, CTMN_SPECTATOR, false));
 			break;
@@ -602,6 +609,24 @@ static CallBackFunction MenuClickCompany(int index)
 		switch (index) {
 			case CTMN_CLIENT_LIST:
 				ShowClientList();
+				return CBF_NONE;
+
+			case CTMN_NEW_COMPANY:
+				if (_network_server) {
+					DoCommandP(0, CCA_NEW, _network_own_client_id, CMD_COMPANY_CTRL);
+				} else {
+					NetworkSendCommand(0, CCA_NEW, 0, CMD_COMPANY_CTRL, nullptr, {}, _local_company);
+				}
+				return CBF_NONE;
+
+
+			case CTMN_SPECTATE:
+				if (_network_server) {
+					NetworkServerDoMove(CLIENT_ID_SERVER, COMPANY_SPECTATOR);
+					MarkWholeScreenDirty();
+				} else {
+					NetworkClientRequestMove(COMPANY_SPECTATOR);
+				}
 				return CBF_NONE;
 		}
 	}


TrueBrain:
Yes, it didn't make it in the PR description. That is my bad. We talked it over outside of GitHub, and as it goes, that doesn't always make it into the PR description. It was already really lengthy, and multiple people indicated it was kinda obvious we remove those entries from the menu too.

So why are they removed? Well, several reasons honestly (these are mine, others might have more to pitch in):

First, them being there was always a hack. Back in the day, I (and I can only blame myself here) was very lazy, and couldn't find a way to add it to the network GUI. So the next best thing was used, with the idea it was "temporary". 14 years later .. yeah .. that didn't happen.
But also, it was a confusing mechanism. It either showed "new company" or "spectate", and this changed depending on who you are. That hidden in a menu, that was confusing to many players.

In addition to that, it also turns out it was really difficult to find for people. With the introduction of the new Online Players GUI it makes much more sense to put it there: all multiplayer related things close together. This is also why it was ruled "obvious" to remove it from the menu: having the same functionality twice is a bit silly, and in the GUI it makes much more sense. Finally removes the weird quirk in that menu.

As for "convenient shortcut", I think that holds true for nearly any functionality you personally use or have gotten used to. Personally, I would like to have scroll-to-HQ in the menu, but alas, we cannot endless add things to the menu. So on the convenient argument, I am not sold. As alternative, having fully customizable menus might be a fun (and goofy) idea.

I do acknowledge that for existing users it can be a bit like: huh? Where did that go? But given we did a lot to indicate there is a new GUI with this functionality, I think most players will not miss it after a bit of time. As after all, how often do players really change to spectator? In other words: I think our GUI should tailor to the 90%.

As such, I think this PR should be closed. It doesn't offer any useful functionality for most of our players, and it feels like it is only scratching an itch for a select few. I can be convinced otherwise, of course, but as it stands, I think it is obvious that #9067 removed this from the menu.

ldpl:
> First, them being there was always a hack. Back in the day, I (and I can only blame myself here) was very lazy, and couldn't find a way to add it to the network GUI. So the next best thing was used, with the idea it was "temporary". 14 years later .. yeah .. that didn't happen.
> But also, it was a confusing mechanism. It either showed "new company" or "spectate", and this changed depending on who you are. 

Feels ok to me, there is a company list and new company / spectate much like in "Online players", even better imo. If anything it's "Online players" that's out of place there since it's a list of companies linked from the list of companies. And, speaking of confusion, depending on "who you are" bald guy button opens either "online players" window or company window, much more confusing than replacing options if you ask me :p

> That hidden in a menu, that was confusing to many players.
> In addition to that, it also turns out it was really difficult to find for people. 

Not much of an issue now that it's not the only way to do it.
 
> With the introduction of the new Online Players GUI it makes much more sense to put it there: all multiplayer related things close together. This is also why it was ruled "obvious" to remove it from the menu: having the same functionality twice is a bit silly, and in the GUI it makes much more sense. Finally removes the weird quirk in that menu.

I'm not suggestting removing it from "online players" (even though it could use a bit of redesign there imo). And there are plenty of examples of other convenience "shortcuts" in the game. Including company window join button that duplicates the functions of the new GUI.

> As for "convenient shortcut", I think that holds true for nearly any functionality you personally use or have gotten used to. Personally, I would like to have scroll-to-HQ in the menu, but alas, we cannot endless add things to the menu. So on the convenient argument, I am not sold. As alternative, having fully customizable menus might be a fun (and goofy) idea.
> I do acknowledge that for existing users it can be a bit like: huh? Where did that go? But given we did a lot to indicate there is a new GUI with this functionality, I think most players will not miss it after a bit of time. As after all, how often do players really change to spectator? In other words: I think our GUI should tailor to the 90%.

How often do players really use anything in "online players"? I'd argue that new company/spectate are by far the most used functions there. Unless more players suddenly discover they can talk to companies/clients and start doing that.

EmperorJake:
"New Company" should have a confirmation window because before the change it was far too easy to accidentally create a new company.

TrueBrain:
Also interesting, turns out people use "Spectator" to mark themselves as "away". So in a use-case:

- Playing on a server that has "pause when nobody is playing"
- Going away from the keyboard for a bit, but don't want to log out
- Go to spectator instead. If nobody else is playing, the game pauses. And I don't have to login when I am back.

In short, this is not the intended use of spectators ofc, but it is a neat way of dealing with it. It is also most likely specific to certain servers, as the common player would not know about this, neither follow this rule.
Users also use this to avoid redownloading the map when they are back. So logging off is not a solution for them.

This insight helped me a bit, as I couldn't understand why people would want to go to spectators if they started to play the game. But if you created your own workflow on top of it, it makes sense of course :) 

Which circles back to this PR, as that means for "Spectate" we have more than "it is convenient" argument (on which I am still not sold). So I browsed a bit through past complaints about those buttons.

Our menu is mostly "safe". This is to say: if you click any buttons, it either asks you to confirm, shows you a GUI, changes your mouse-input, or is a toggle. This means if you click somewhere by accident, the road to recovery is always clear. There are three exceptions to this in pre-12.0:
- Spectate
- New Company
- Delete all news messages

The latter hasn't resulted in any complaints (yet), but the other two have. @EmperorJake also mentions this rightfully, that especially "New company" has been an issue. And this is what I mentioned earlier: as those two buttons swap places, the UX becomes very unclear. A user isn't shown how to get back to their original company after pressing "Spectate". Pressing the same location in the menu again DOES NOT bring you back to the earlier state. Many users reported this over time as confusing, got lost, irritated, etc. In other words: we really shouldn't revert back to this state, as it clearly wasn't a good place to be in. So this PR as-is, we really shouldn't do.

The "New Company" specifically, is also the reason we removed the lobby window: people clicking it by accident, not really knowing what it means. So in my opinion, the only place to start a new company should be in the Online Players window. As it is more likely the user will realise they wanted to join a company, and do that instead.

The "Spectate" button is the more interesting one. Especially as it is used as an "I am away" button. Having that conveniently available, becomes an argument worth talking about. Someone is at your door, you just quickly want to mark yourself as away. Opening the right GUI to find the right button, yes, that is an argument I can understand.

Which creates three questions in my head:

1) shouldn't we just create an "away" system, instead of misusing "spectator" for this? Create a "pause if everyone is away", make "I am away" a toggle in the menu, show a screen when you are away saying: click here if you are back, and don't move people to spectator. We could even make people go auto-away if they haven't touched OpenTTD for N minutes.
2) if not the "away" system, this is (claimed to be) a niche-workflow for specific types of multiplayer game. Nothing wrong with that, but is "Spectate" in this menu the best place for that? And the most clear for people not familiar with this rule on a server? Or can we do better?
3) if we do "Spectate" button in the menu, can we make it so clicking it again reverts the state you were in? Basically making it a toggle. From an UX perspective, I have a lot less issues with such a toggle.

Tnx a lot to DIscord where Azusa and Wahazar were very helpful in explaining their use-case. I really appreciate you guys, and that was most helpful :)

ldpl:
First of all, while afk is one use-case you seem to entirely forget what, I suppose, is the intended use of spectator mode - spectating. And while spectating you main want to 

1. Join another company to help it.
2. Join another company for administrative actions (if you're a server moderator).
3. Create a new company for testing things.
4. Create a new company so your messages aren't grey and can be noticed by other players.

Then there is the case when your company is doing poorly and you want a new one, that currently requires spectating first. Also, there is a twist that in this case and 3-4 of the list you may want to reset your current company which involves typing `!reset` or `!resetme` and spectating (on servers that support it).

And that's just something off the top of my head, there may be more use-cases. So while I agree that there may be better UI for these cases I think that until it's developed returning "new company/spectate" menu option is the right step as it should've never been removed in the first place. I only missed that during #9067 development because I never expected that and it was never mentioned anywhere I looked.

> The "New Company" specifically, is also the reason we removed the lobby window: people clicking it by accident, not really knowing what it means. 

As someone (and afaik the only one) who brought up that reason in the first place, let me tell you - "new company" in the menu was never the problem. One that was used as "join server" button was :p

P.S. I think "away" state should be automatic.

TrueBrain:
> 
> As someone (and afaik the only one) who brought up that reason in the first place, let me tell you - "new company" in the menu was never the problem. One that was used as "join server" button was :p
> 

I am happy you speak for everyone.

andythenorth:
I think it's important that nobody develops an ulcer over OpenTTD.  It's just cartoon trains.

PR #9534, opened 2021-09-02
Fix: use-after-free after ClientNetworkCoordinatorSocketHandler::CloseAllConnections()

TrueBrain:
## Motivation / Problem

While debugging a GC issue, I managed to cause a use-after-free. Example flow:

Start your own server, and while your server is being probed (takes ~3 seconds), kill the connection (for example by modifying the code or running your own GC that does this).

Possibly there are other ways to trigger a `CloseAllConnections`.

## Description

```
The function clears all stun-handlers. This causes all of those
objects to be destroyed.
A handler can have a pending connecter, which was only killed in
case CloseConnection() was called. This is never the case when
the object is destroyed. In result, the connecter could finish
and cause a use-after-free by calling into the (now deleted)
handler.
```

I have no clue why I initially wrote it in `CloseConnection()` with this comment, and that worries me a bit. I remember there was something special about these objects, but .. yeah, can't remember, and this seems to be the proper solution.

## Checklist for review

### DIFF
--- a/src/network/network_stun.cpp
+++ b/src/network/network_stun.cpp
@@ -100,15 +100,21 @@ NetworkRecvStatus ClientNetworkStunSocketHandler::CloseConnection(bool error)
 {
 	NetworkStunSocketHandler::CloseConnection(error);
 
-	/* If our connecter is still pending, shut it down too. Otherwise the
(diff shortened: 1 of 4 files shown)


LordAro (approved):
Sure.

PR #7235, opened 2019-02-16
Change: Non-rectangular sparse station catchment area

PeterN:
This patch changes station catchment from a single rectangle covering the largest catchment radius of the station, to individual tiles based on the catchment radius of the station tiles near them.

The non-saved calculated set of catchment tiles is stored off-map within the station class. This set is then tested against when cargo is delivered to and from industries and towns.

~~Adds a third station catchment area option:~~

Original fixed-size catchment area:
catchment0

Modified variable-size catchment area:
catchment1

New non-rectangular sparse catchment area:
catchment2

PR #9536, opened 2021-09-03
Change: Deliver cargo to the closest industry first

ldpl:
## Motivation / Problem

Among other things, #7235 changed the way stations chose the industry to deliver cargo to. Before it, industries that are closer to the station had priority, after it those with lower id got it. And that proved to be problematic as there is no way to control or even see industry id in the game. A typical example would be another industry spawning next to the existing station and stealing all the production. With distance system you can manipulate station sign to solve (or even completely prevent) it. With index system you have to move entire station which may not even be feasible at that point. Also with index system you have to be careful not to include other industries that accept same cargo in your catchment area as you have no way of telling which one will get them.

## Description

Orders the nearby industry lists by DistanceMax to the nearest industry tile from station sign (as closest equivalent of previously used CircularTileSearch). 

## Limitations

Is somewhat less performant than before but mostly affects rebuilding nearest caches that don't seem to be anywhere near time-critical. Also the algorithm can be improved if needed (I actually started on it but abandoned due to excessive complexity of the change: https://gist.github.com/ldpl/2fe7ea9bf8931cdbe632222d058474c5).

## Checklist for review

### DIFF
--- a/src/cargomonitor.cpp
+++ b/src/cargomonitor.cpp
@@ -149,9 +149,9 @@ void AddCargoDelivery(CargoID cargo_type, CompanyID company, uint32 amount, Sour
 	if (iter != _cargo_deliveries.end()) iter->second += amount;
 
 	/* Industry delivery. */
-	for (Industry *ind : st->industries_near) {
-		if (ind->index != dest) continue;
-		CargoMonitorID num = EncodeCargoIndustryMonitor(company, cargo_type, ind->index);
+	for (const auto &i : st->industries_near) {
+		if (i.industry->index != dest) continue;
+		CargoMonitorID num = EncodeCargoIndustryMonitor(company, cargo_type, i.industry->index);
 		CargoMonitorMap::iterator iter = _cargo_deliveries.find(num);
 		if (iter != _cargo_deliveries.end()) iter->second += amount;
 	}
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1040,9 +1040,10 @@ static uint DeliverGoodsToIndustry(const Station *st, CargoID cargo_type, uint n
 
 	uint accepted = 0;
 
-	for (Industry *ind : st->industries_near) {
+	for (const auto &i : st->industries_near) {
 		if (num_pieces == 0) break;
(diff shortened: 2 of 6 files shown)


PeterN:
Does this take account of the size of industries (and the station) or just their stored tile position?

J0anJosep:
Taking into account only the distance isn't the best solution in all cases.

If a new industry spawns close to a station, the station that delivered cargo to an existing industry may deliver it to the newly built industry.

Industries can also be sorted by their construction date, which could be another alternative, but sorting industries by index seems ok to me.

ldpl:
It's using the distance to the closest industry tile and the good thing about it is that it's clearly visible in the game and also relatively intuitive. Also, the main point of this PR is not to invent a perfect system but to return the one that worked fine before and to which players are already used to instead of using index that is effectively just random.

scrmjt:
Completely in support of this PR - this is very relevant when trying to serve multiple industries with specific goods, such as engineering supplies in FIRS. Without this change a large catchment area will only service one industry. You have to use some very inconvenient workarounds (such as artificially spreading out stations so they don't overlap two industries in the example below) otherwise I know of no other way to service this type of situation (which seems to be a fairly realistic scenario).

Options to split station cargo between industries or enable/disable a station serving certain industries would also be welcome if it's technically possible. Appreciate everyone's work on this - such a great game.

stations
).

2TallTyler:
I am also +1 to this change, for the reason stated by @scrmjt and because it's the original way of distributing cargo. I also think it's the most intuitive way, although of course this is subjective.

EmperorJake:
I support this change back, it was a nice trick to be able to manipulate the station sign to decide which industry you serve. Now you have to avoid overlapping the catchment area if the industries are too close, and there isn't always room for that if things are tight. This is especially pertinent with FIRS and XIS Supplies.

michicc (approved):

TrueBrain:
Not going to backport this PR, as it might change the game of people using 12.1. We tend to avoid those kind of things when we can. It of course will be in 13.0 :)

PR #9538, opened 2021-09-04
Fix #9535: Maintain a reverse dependency map of network content

### LINKED ISSUE #9535: [Bug]: Poor performance opening "Check Online Content" window due to O(N^3) behaviour
### Version of OpenTTD

master (9c74dc24), however this is not a new bug

### Expected result

Blazing fast, silky smooth, positive adjectives performance when opening the check online content window

### Actual result

Frame rate decreases dramatically when the list is being populated.
Screenshot_2021-09-03_17-13-37-trim

The performance problem seems to be created in `ClientNetworkContentSocketHandler::ReverseLookupDependency` via `ClientNetworkContentSocketHandler::CheckDependencyState`.
This seems to be because each received content item calls `CheckDependencyState` on each previously received item, which calls `ReverseLookupDependency` which again iterates all previously received items, leading overall to O(N^3) behaviour in the number of content items. See below for approximate measurements.
Screenshot_2021-09-03_17-25-28

### Steps to reproduce

1. Open the "Check Online Content" window from the main menu (not when in a game).

JGRennison:
## Motivation / Problem

This fixes the performance issues in #9535 with dependency lookup when retrieving the content list from the content server.

## Description

Instead of iterating all content to find reverse dependencies, maintain an iteratively built reverse dependency map.
When a new content item arrives, instead of calling `CheckDependencyState` on all content items, only call it on the reverse tree dependencies of the newly received item.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/network/network_content.h
+++ b/src/network/network_content.h
@@ -12,6 +12,7 @@
 
 #include "core/tcp_content.h"
 #include "core/tcp_http.h"
+#include <unordered_map>
 
 /** Vector with content info */
 typedef std::vector<ContentInfo *> ContentVector;
@@ -68,6 +69,7 @@ protected:
 	std::vector<ContentCallback *> callbacks;     ///< Callbacks to notify "the world"
 	ContentIDList requested;                      ///< ContentIDs we already requested (so we don't do it again)
 	ContentVector infos;                          ///< All content info we received
+	std::unordered_multimap<ContentID, ContentID> reverse_dependency_map; ///< Content reverse dependency map
 	std::vector<char> http_response;              ///< The HTTP response to the requests we've been doing
 	int http_response_index;                      ///< Where we are, in the response, with handling it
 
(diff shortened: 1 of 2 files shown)


TrueBrain (dismissed):

src/network/network_content.h
```diff
@@ -82,6 +84,7 @@ class ClientNetworkContentSocketHandler : public NetworkContentSocketHandler, Co
 	bool Receive_SERVER_CONTENT(Packet *p) override;
 
 	ContentInfo *GetContent(ContentID cid);
+	const ContentInfo *GetContent(ContentID cid) const { return const_cast<ClientNetworkContentSocketHandler *>(this)->GetContent(cid); }
```
LordAro: Unless I'm missing something obvious, the existing GetContent method doesn't make any changes to the object and can be declared const without making this hacky extra function

JGRennison: That's fixed now.

LordAro (approved):

PR #9542, opened 2021-09-05
Fix #9241: Grove and forest tree brushes also create rainforests

### LINKED ISSUE #9241: Some tree brushes do not mark tiles as a rainforest
## Version of OpenTTD
1.11.2, Windows 10

## Expected result
Planting rainforest trees in the scenario editor should create rainforest tiles.

## Actual result
Only the "normal" brush creates rainforest tiles, "grove" and "forest" brushes do not.

vituscze:
## Motivation / Problem

Creating rainforest in the scenario editor is exceptionally hard. The only way to make a rainforest is to plant a rainforest tree with the single-tile tree brush. See also https://github.com/OpenTTD/OpenTTD/issues/9241

## Description

This PR allows larger tree brushes to also create rainforest tiles. The behavior of random-tree brushes remains unchanged.

## Limitations

Currently, the implementation simply makes the behavior of the larger tree brushes match that of the single-tile brush. However, the zoning could be more complex (such as planting cacti also creates desert or such).

## Checklist for review

### DIFF
--- a/src/tree_gui.cpp
+++ b/src/tree_gui.cpp
@@ -29,7 +29,7 @@
 #include "safeguards.h"
 
 void PlaceTreesRandomly();
-uint PlaceTreeGroupAroundTile(TileIndex tile, TreeType treetype, uint radius, uint count);
+uint PlaceTreeGroupAroundTile(TileIndex tile, TreeType treetype, uint radius, uint count, bool set_zone);
 
 /** Tree Sprites with their palettes */
 const PalSpriteID tree_sprites[] = {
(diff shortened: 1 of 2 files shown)


TrueBrain (dismissed):

src/tree_gui.cpp
```diff
@@ -133,7 +133,7 @@ class BuildTreesWindow : public Window
 		}
 		const uint radius = this->mode == PM_FOREST_LG ? 12 : 5;
 		const uint count = this->mode == PM_FOREST_LG ? 12 : 5;
-		PlaceTreeGroupAroundTile(tile, treetype, radius, count);
+		PlaceTreeGroupAroundTile(tile, treetype, radius, count, this->tree_to_plant != TREE_INVALID);
```
LordAro, line 137: I don't think I follow why the `set_zone` parameter is necessary at all? Under what circumstances would the tree_to_plant be invalid (so picking a random type), but you *wouldn't* want to set the tropic zone?

vituscze: If you plant random trees with the single-tile brush, it doesn't create rainforest either.

LordAro: Can we get a comment explaining that?

LordAro (approved):
If I wanted to be super pedant-y, comments on their own line should use `/* */` style

Issue #9523, opened 2021-08-29
[Bug]: "ship is lost" is only reported via news


TrueBrain:
### Version of OpenTTD

12.0-beta2

### Expected result

See somewhere on the vehicle or wherever, that the ship is lost.

### Actual result

If a ship cannot reach its destination, it shows news about that once. This is true for all vehicles, but for ships even more annoying as they tend to drift towards the destination.

If you think you fixed the route, it doesn't update you with "route found!", so you are in the false sense you fixed the issue, and it never shows "ship is lost" again. This makes using ships a very frustrating experience.

### Steps to reproduce

1. Build a dock upstream
2. Don't build a lock
3. Build a depot downstream
4. Build a ship
5. Make it go to the dock
6. Observe the news message about it being lost
7. Try to find back whether it is still lost after that

What makes this extra annoying: 
8. Build a lock, but the ship still cannot reach the dock
9. You will NOT receive any notification it still cannot reach the destination

TrueBrain:
The biggest "well that is a shame" part, is that there is a flag to indicate the ship is lost in the vehicle flags (`VF_PATHFINDER_LOST`). It is just not used anywhere in the GUI.

It seems this is a pretty easy thing to address, we just need to find a clean way to present it to the user. Maybe in the status bar of the vehicle? `Couldn't find a route to ...` instead of `Heading to ...`?

James103:
There could also be an alert symbol `⚠` displayed in front of the vehicle's consist in the GUI. For example, if the `VF_PATHFINDER_LOST` flag is set, the alert symbol is shown, in order to further inform players in the vehicle listings.

PR #9543, opened 2021-09-05
Feature #9523: Display icon or text in vehicle list or in vehicle window indicating whether a vehicle is lost

SamuXarick:
## Motivation / Problem
#9523

## Description
While VF_PATHFINDER_LOST is set for a vehicle, a warning icon is displayed on vehicle window and vehicle list. On the vehicle window, status bar text will change from "heading to" to "can't reach".

Aberhattan Springs Transport, 1935-03-03

## Limitations
I am reusing a sprite! Pretty sure this is not the way to go.

## Checklist for review


glx22:
Reusing a sprite is totally fine here.

src/vehicle_gui.cpp
```diff
@@ -2732,7 +2736,7 @@ struct VehicleViewWindow : Window {
 		const Vehicle *v = Vehicle::Get(this->window_number);
 		switch (widget) {
 			case WID_VV_START_STOP:
-				size->height = std::max(size->height, std::max(GetSpriteSize(SPR_FLAG_VEH_STOPPED).height, GetSpriteSize(SPR_FLAG_VEH_RUNNING).height) + WD_IMGBTN_TOP + WD_IMGBTN_BOTTOM);
+				size->height = std::max(size->height, std::max(GetSpriteSize(SPR_WARNING_SIGN).height, std::max(GetSpriteSize(SPR_FLAG_VEH_STOPPED).height, GetSpriteSize(SPR_FLAG_VEH_RUNNING).height)) + WD_IMGBTN_TOP + WD_IMGBTN_BOTTOM);
```
LordAro: `std::max({foo, bar, baz})` using an initializer list would probably be better than chaining std::max calls

src/lang/english.txt
```diff
@@ -4142,6 +4142,11 @@ STR_VEHICLE_STATUS_HEADING_FOR_WAYPOINT_VEL                     :{LTBLUE}Heading
 STR_VEHICLE_STATUS_HEADING_FOR_DEPOT_VEL                        :{ORANGE}Heading for {DEPOT}, {VELOCITY}
 STR_VEHICLE_STATUS_HEADING_FOR_DEPOT_SERVICE_VEL                :{LTBLUE}Service at {DEPOT}, {VELOCITY}
 
+STR_VEHICLE_STATUS_CANT_REACH_STATION_VEL                       :{LTBLUE}Can't reach {STATION}, {VELOCITY}
```
LordAro: There aren't a lot of contractions in the strings at the moment, I wonder whether `Cannot reach...` would be better

LordAro (commented):
LGTM, other than a couple of minor points

TrueBrain:
Are you happy with the result? Can we un-draft this? :D

SamuXarick:
It's a bit messed up. It doesn't immediately pick up the status when the route is fixed. And in some cases, it can reach the target while still saying that it cannot be reached. I did a few testings with trains and a missing single track that gets fixed meanwhile, and it took the train some time to clear up the train is lost status.

TrueBrain:
> 
> 
> It's a bit messed up. It doesn't immediately pick up the status when the route is fixed. And in some cases, it can reach the target while still saying that it cannot be reached. I did a few testings with trains and a missing single track that gets fixed meanwhile, and it took the train some time to clear up the train is lost status.

Is this because the GUI doesn't update? Or because the PF just caches information and takes a bit of time to notice the track is restored? The first we can fix. The second .. I personally can accept that as a "known-bug". We cannot instantly update all trains if you put down a single rail :) I think the benefit of this PR out wages the possible questions that might give.

SamuXarick:
The second.

TrueBrain:
*beep beep* IRC to GitHub bot here:

```
<TrueBrain> Samu: not sure what others think, but I fully expected the icon to remain even after the route was fixed .. that is just how OpenTTD works :D
<TrueBrain> the news also remains despite the route being fixed
<TrueBrain> for me that really is not an issue .. but others might disagree :)
<TrueBrain> having some indication is already widely better than nothing, even if it can be there while the route is fixed :)
<FLHerne> TrueBrain: I think it's a bug that the news doesn't vanish too
<FLHerne> but yeah, even with this bug it's a big improvement
<_dp_> train stops being lost when it finds a path, not when tracks are "fixed", train may be on the other side of the map by that point
<_dp_> also would be very helpful if it also showed somewhere exact tile where it became lost
<_dp_> but, yeah, any indication is better than no idication
```

In other words, yes, this PR would "introduce a bug", but not one we might want to solve in this PR :)

TrueBrain (approved):

PR #9549, opened 2021-09-11
Fix #9463: [Win32] Try to work around XAudio2 crashes by catching SEH exceptions.

### LINKED ISSUE #9463: [Crash]: Game crashes on start up.
### Version of OpenTTD

1.11.2 Windows 10

### Steps to reproduce

Game crashes on start up. I've tried all the available betas (1.11.0-1.11.2 and nightly)

### Upload crash files

[crash.zip](https://github.com/OpenTTD/OpenTTD/files/6940343/crash.zip)

michicc:
## Motivation / Problem

While it looks like the XAudio2 crashes reported e.g. in #9463 are caused by third-party components, users will still blame OpenTTD at first. As such, if we can do something, user experience is improved, especially as other audio drivers besides XAudio2 are available. 

## Description

The provided crash logs point to an illegal access exception during XAudio2 creation. Windows reports such errors using the SEH mechanism which can be caught by platform-specific code. If a SEH exception is encountered, just fail driver creation and let the system do the normal fallback.

The extra function is needed as SEH handling can't be used in contexts that rely on C++ unwinding.

## Limitations

As I don't have the suspected third-party software components, I can't verify if this PR actually fixes the reported issues.

## Checklist for review

### DIFF
--- a/src/sound/xaudio2_s.cpp
+++ b/src/sound/xaudio2_s.cpp
@@ -117,6 +117,20 @@ static IXAudio2MasteringVoice* _mastering_voice = nullptr;
 static ComPtr<IXAudio2> _xaudio2;
 static StreamingVoiceContext* _voice_context = nullptr;
 
+/** Create XAudio2 context with SEH exception checking. */
+static HRESULT CreateXAudio(API_XAudio2Create xAudio2Create)
+{
+	HRESULT hr;
+	__try {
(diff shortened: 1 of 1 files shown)


glx22:
In theory it should work, but I can't test it.

TrueBrain (approved):
Doubt it makes the situation worse, so sure, why not :)

PR #9560, opened 2021-09-18
Fix: Incorrect error messages on placing water on scenario editor

SamuXarick:
## Motivation / Problem
I found some issues when placing sea/canal/river on scenario, the error messages don't always make sense and the behaviour is a little inconsistent when compared with each other. Listed below is the behaviour tested on 12.0-beta2.

On the scenario editor...
Building sea on sea: "Can't build canals here... already built"
Building sea on canal (owner none): "Can't build canals here... already built"
Building sea on canal (company owned): "Can't build canals here... already built"
Building sea on river: "Can't build canals here... already built"

Building canal on sea: builds with no error
Building canal on canal (owner none): "Can't build canals here... already built"
Building canal on canal (company owned): "Can't build canals here... already built"
Building canal on river: builds with no error

Building river on sea: builds with no error
Building river on canal (owner none): "Can't place rivers here... already built"
Building river on canal (company owned): "Can't place rivers here... already built"
Building river on river: overbuilds with no error

On a game...
Building canal on sea: builds with no error, cost £3,750
Building canal on canal (owner none): "Can't build canals here... already built"
Building canal on canal (self-owned): "Can't build canals here... already built"
Building canal on canal (other company): "Can't build canals here... already built"
Building canal on river: builds with no error, cost £3,750

## Description
Below is the behaviour after the PR, which seem a bit better.

On the scenario editor...
Building sea on sea: overbuilds with no error
Building sea on canal (owner none): builds with no error
Building sea on canal (company owned): "Can't build canals here... owned by 'company'"
Building sea on river: builds with no error

Building canal on sea: builds with no error
Building canal on canal (owner none): overbuilds with no error
Building canal on canal (company owned): "Can't build canals here... owned by 'company'"
Building canal on river: builds with no error

Building river on sea: builds with no error
Building river on canal (owner none): builds with no error
Building river on canal (company owned): "Can't place rivers here... owned by 'company'"
Building river on river: overbuilds with no error

On a game...
Building canal on sea: builds with no error, cost £3,750
Building canal on canal (owner none): "Can't build canals here... already built"
Building canal on canal (self-owned): "Can't build canals here... already built"
Building canal on canal (other company): "Can't build canals here... owned by 'company'"
Building canal on river: builds with no error, cost £3,750

## Limitations
Still incorrect error message for 'Building sea on canal (company owned)'

## Checklist for review

### DIFF
--- a/src/water_cmd.cpp
+++ b/src/water_cmd.cpp
@@ -477,8 +477,8 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 
 		bool water = IsWaterTile(current_tile);
 
-		/* can't make water of water! */
-		if (water && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
+		/* Outside the editor, prevent building canals over your own or OWNER_NONE owned canals */
+		if (water && IsCanal(current_tile) && _game_mode != GM_EDITOR && (IsTileOwner(current_tile, _current_company) || IsTileOwner(current_tile, OWNER_NONE))) continue;
 
 		ret = DoCommand(current_tile, 0, 0, flags, CMD_LANDSCAPE_CLEAR);
 		if (ret.Failed()) return ret;


SamuXarick:
It is ready to review now.

src/water_cmd.cpp
```diff
@@ -478,7 +478,7 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 		bool water = IsWaterTile(current_tile);
 
 		/* can't make water of water! */
-		if (water && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
+		if (water && IsCanal(current_tile) && Company::IsValidID(_current_company) && (IsTileOwner(current_tile, _current_company) || GetTileOwner(current_tile) == OWNER_NONE)) continue;
```
TrueBrain: As mentioned on IRC, not explaining what this does makes this a terrible line to read. So many questions:

- Why the `IsValidID`?
- Why `IsTileOwner` vs `GetTileOwner`?
- Why the check against `_current_company`?
- Why the check against `OWNER_NONE`?
- Where did the check against `OWNER_WATER` go?
- Where did the `WATER_CLASS_SEA` check go?

These might all be legit choices and be a fine solution, but nothing in this PR nor description helps me at all. You force me to read up in this code in detail to figure it out .. by that time it is often easier for me to just do it myself, instead of trying to understand what you tried to do.

Please please do better. Fix the comment above the line, and try to explain what you think this line is doing. We will take it from there. But this really is not sufficient, sorry :(

SamuXarick: I used `Company::IsValidID(_current_company)` because in SE, you're `OWNER_NONE` and in a game `_current_company` is a valid company. I needed to make the check fail in SE to allow overbuilding canals on SE-placed canals. Not sure if using `_current_company != OWNER_NONE` instead would be better - In-game I need the entire `if` to be true to skip the `current_tile` to come up with "already built" message. In the case of building a company canal over a canal of another company, I need `Company::IsValidID(_current_company)` to fail.

The check `GetTileOwner(current_tile) == OWNER_NONE` is only happening in-game, and not in SE. This is made this way to avoid overbuilding canals on SE-built/bankrupted canals, but mostly because it'd be paying the price for something that already exists. I though this behaviour was fine and didn't need change, I kept it the way it was already. This is in contrast with what happens in scenario editor, where I made it overbuild.

The check against `OWNER_WATER` and `WATER_CLASS_SEA ` are gone. Trying to explain why is actually tougher than I expected... it's mostly because of the erroneous messages that show up especially on SE. If I ignore the messages for a moment, the logic still seems valid at glance: overbuilding sea on sea doesn't have to clear the tile just to build over again. This PR makes it do the opposite, it clears the tile and builds over with the benefit that the message is gone. The drawback is that the cost is added, but since it's in SE, it doesn't really matter (or does it?).


As for what should the comment in the line above should say? I'm not really sure, honestly. It can make water of water now in some cases, with a few exceptions related to whenever the tile is a canal and who owns it. I'm not really sure what to write in the comment.

EDIT: I'm thinking of:

```
		/* Allow to make water of water, except in these specific situations.
		 * The checks done via CMD_LANDSCAPE_CLEAR can provide better error messages if any. */
```

src/water_cmd.cpp
```diff
@@ -477,8 +477,9 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
 
 		bool water = IsWaterTile(current_tile);
 
-		/* can't make water of water! */
-		if (water && (!IsTileOwner(current_tile, OWNER_WATER) || wc == WATER_CLASS_SEA)) continue;
+		/* Allow to make water of water, except in these specific situations.
+		 * The checks done via CMD_LANDSCAPE_CLEAR can provide better error messages if any. */
```
TrueBrain: This comment doesn't really help the reader. "These specific situations" is what is important. As the code is not obvious in what those situations are.
Also, the "make water of water" doesn't make sense anymore.

You have to look at it this way: if you read this in 1 year, would you still understand what this code does? And if not, will the comment help you understanding?

So what are those specific situations? Well, if I understood you correctly, your intentions are:
- Building a canal on your own canal should be skipped.
- Building a canal on a `OWNER_NONE` owned canal should be skipped.

So the comment should at least mention that.

Basically, what you made this `if` do now, is something like:
`Prevent building a canal over your own or OWNER_NONE owned canal`.

src/water_cmd.cpp
```diff
@@ -477,8 +477,9 @@ CommandCost CmdBuildCanal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32
...
+		if (water && IsCanal(current_tile) && Company::IsValidID(_current_company) && (IsTileOwner(current_tile, _current_company) || GetTileOwner(current_tile) == OWNER_NONE)) continue;
```
TrueBrain: There are two things in this line that are weird to me.
- The `IsValidID` is not obvious why it is there. So either you have to put it in the comment why, but a few lines above they use a more verbose way: `_game_mode != GM_EDITOR`.
- For one check you  use `IsTileOwner`, for the next `GetTileOwner`. Why? Is there something wrong with using `IsTileOwner` twice? (hint: there isn't, as `IsTileOwner` executes `GetTileOwner`)

TrueBrain (approved):

PR #9563, opened 2021-09-19
Add a method for remote admins to send chat messages from external sources

ldpl:
## Motivation / Problem

Many OpenTTD communities bridge ingame chat with discord/irc or other servers. Previously common solution for that would be to give the server client an empty name (so it doesn't prepend them with awkward Server: or smith) and make it send those external messages. In 12.0 it's no longer possible to give the server an empty name and while one could always name it ":" it's ridiculous and the appearance of the chat messages changed anyway so I decided to make a dedicated method for it. I also thought of making a more generic method but while it may be useful it would require an admin to somehow manage the formatting of messages in sync with possible future changes and external chat is a common enough use case anyway to deserve a dedicated method.

## Description
Add ADMIN_EXTERNAL_CHAT packet that receives message text, sender, and source as strings as well as text colour and passes it through to the clients to show.

Screenshot from 2021-09-19 17-24-59

Simple remote admin script in python for testing: [testadmin.zip](https://github.com/OpenTTD/OpenTTD/files/7192340/testadmin.zip)

## Limitations
1. I've no idea what's up with NETWORK_GAME_ADMIN_VERSION never being incremented but this is less of a protocol change than, for example, 4a1bf704ee1d8de386fe9f5b5fd87a46fb7bb615 and 4a1bf704ee1d8de386fe9f5b5fd87a46fb7bb615 and those didn't do it either.
2. External chat messages aren't sent to other connected remote admins but neither do regular chat messages from other admins.

## Checklist for review


LordAro (approved):
Seems reasonable to me. I leave final approval to @TrueBrain, who can answer the questions about protocol version numbers

TrueBrain:
I know very little about the admin protocol. I would reason that adding a packet to the server should be without bump, as it doesn't break existing clients. In this reasoning the version is similar to the major version within semver.

So, YOLO! Let's go.

PR #9567, opened 2021-09-21
Feature: Ships may ask pathfinder whether to reverse when leaving docks

### LINKED ISSUE #9566: [Bug]: Lost ship after leaving station if it docked in a perpendicular manner and doesn't reverse
### Version of OpenTTD

20210919-master-g45edd9f326

### Expected result

When the ship reaches Sentown Docks and leaves it, it should be able to reverse on the spot and go towards northwest, where it will find its way to next destination.

### Actual result

When the ship reaches Sentown Docks and leaves it, it follows the path towards southeast of it and ends up lost, unable to make a turn around until it reaches open sea.

### Steps to reproduce

[Ship can't reverse at Sentown Docks.zip](https://github.com/OpenTTD/OpenTTD/files/7204799/Ship.can.t.reverse.at.Sentown.Docks.zip)
ship becomes lost after leaving dock
t.

- Load savegame, watch ship become lost

This is probably not a bug, but more of a feature request, to ask the pathfinder whether it's better to reverse after leaving the docks.

I found this situation to be the main reason I have why some of my AI ships move so further out of place that gets them lost, even when using buoys - they enter some canal route leading to open sea and unable to return back to the right track.

SamuXarick:
## Motivation / Problem
Fix #9566 

## Description
When the ship docks perpendicular to a dock, ask the pathfinder whether it's better to reverse when leaving docks, to reach next destination.

## Limitations
ProcessOrders may be executed twice on the same tick. Not entirely sure if that is allowed.

## Checklist for review

### DIFF
--- a/src/ship_cmd.cpp
+++ b/src/ship_cmd.cpp
@@ -679,6 +679,15 @@ static void ShipController(Ship *v)
 					DiagDirection exitdir = VehicleExitDir(v->direction, v->state);
 					TileIndex tile = TileAddByDiagDir(v->tile, exitdir);
 					if (TrackStatusToTrackBits(GetTileTrackStatus(tile, TRANSPORT_WATER, 0, exitdir)) == TRACK_BIT_NONE) goto reverse_direction;
+					/* Ask pathfinder for best direction */
+					ProcessOrders(v); // updates next destination
+					bool reverse = false;
+					switch (_settings_game.pf.pathfinder_for_ships) {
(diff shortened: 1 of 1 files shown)


glx22:
I have an [alternate version](https://github.com/OpenTTD/OpenTTD/compare/master...glx22:alt_9567) for the same result, but it adds a 1 tick delay if reversing is preferred.

PR #9576, opened 2021-09-26
Revert 7ca1793: Using Trackdir keyed node is not required, Exitdir keyed node still have the correct trackdir

glx22:
## Motivation / Problem
Trackdir keyed nodes are forced when 45° and 90° penalties for ships differ (and they differ when using default values).
That implies a lot more nodes are opened when pathfinding, which implies more closed node, and when enough nodes are closed (default 10K), pathfinding just stops and ship says "can't reach destination".
But using trackdir keyed nodes for different curve penalties is not required at all, because even for exitdir keyed nodes the correct trackdir is still available (and already used to determine next tile).

## Description
Revert the commit introducing the forced switch to trackdir keyed nodes.
Hugely decreases the number of nodes opened to find a path, which increases the chance to actually find one where using trackdirs would bail out before finding it.

## Limitations
Found paths may be a little different, but in most cases they are the same.
Using trackdir keyed nodes:
trackdir
Using exitdir keyed nodes:
exitdir

## Checklist for review

    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status
)

### DIFF
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -337,16 +337,6 @@ struct CYapfShip1 : CYapfT<CYapfShip_TypesT<CYapfShip1, CFollowTrackWater    , C
 /* YAPF type 2 - uses TileIndex/DiagDirection as Node key */
 struct CYapfShip2 : CYapfT<CYapfShip_TypesT<CYapfShip2, CFollowTrackWater    , CShipNodeListExitDir > > {};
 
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):
If any problems come from this, lets document better what this was solving, so we know for next time we have this "ship PF sucks" debate :D

PeterN:
Caveat of this change is that sometimes ships will take a 90 degree turn when a 45 degree turn could've been possible. But that can be considered a performance trade-off.

PR #9589, opened 2021-10-01
Fix #9588, 140a96b: [Squirrel] Reaching memory limit during script registration could prevent further script detections

### LINKED ISSUE #9588: [Bug]: OpenTTD can think that no AI or GS are available if `Maximum memory allocation exceeded`
### Version of OpenTTD

20210929-master commit e89be8d on Windows 10 64-bit

### Expected result

When `Maximum memory allocation exceeded` occurs during parsing a script's `info.nut`, only that script is removed from the list and not any others.

### Actual result

When `Maximum memory allocation exceeded` occurs during parsing a script's `info.nut`, not only is that script removed from the list, but the following two things happen that mean you must restart the game to use AI or GS:
* Any other scripts found during execution of `rescangame` will also not be added to the list because of the same error.
* Further executions of `rescangame` will not work at all until game restart, even if the excessive allocation is prevented.

### Steps to reproduce

1. Make sure you have at least 3 GB of available RAM. The `array(1e8)` at line 1 of the script's `info.nut` will allocate approximately 1.6 GB of RAM.
2. Extract the following ZIP into your home directory's `OpenTTD\game` folder.
[_TestGS.zip](https://github.com/OpenTTD/OpenTTD/files/7255261/_TestGS.zip)
3. Start up OpenTTD.
4. Download some game scripts.
5. Notice that they don't show up.
6. Extract the following ZIP into your home directory's `OpenTTD/ai` folder.
[_TestAI.zip](https://github.com/OpenTTD/OpenTTD/files/7255279/_TestAI.zip)
7. Run `rescanai` twice.
8. Download some AIs.
9. Notice that they don't show up.
10. Exit OpenTTD.
11. Restart OpenTTD.
12. The game crashes with (parenthetical phrase added in post):
```
Crash reason:
 Exception: E06D7363 (Unhandled C++ Exception)
 Location:  00007FFAF9D74F99
 Message:   <none>
```
[crash.zip](https://github.com/OpenTTD/OpenTTD/files/7255329/crash.zip)

glx22:
## Motivation / Problem
When a script's `info.nut` triggers `Maximum memory allocation exceeded`, it can prevent the detection of other scripts of the same type.
This happens because the memory allocation triggering the limit is still accounted in the allocator for all other detections.
And that means this memory is also never freed.

Once the `allocated_size` issue was fixed, another issue appeared, a script failing to register due to memory limit would still be available ingame (and possibly other scripts that should have failed memory check). That's because scripts are scanned multiple times, and the error was not properly reset.

Finally another memory leak exists, when the memory limit is triggered in a constructor and using placement new.

Regarding the reported crash, it's because `DummyAI` registration triggers the memory limit due to a previous failing script and incomplete cleaning between registrations.

## Description
Properly count allocated memory, and free memory not referenced by squirrel before it's lost.
I added an assert to ensure all allocated memory is managed by squirrel and its garbage collector, or is taken care of otherwise.

## Limitations
As now the engine is properly reset, the failing scripts are now reported multiple times in debug console, because many rescan happen during startup.

## Checklist for review

### DIFF
--- a/src/3rdparty/squirrel/squirrel/sqobject.h
+++ b/src/3rdparty/squirrel/squirrel/sqobject.h
@@ -62,6 +62,24 @@ struct SQRefCounted
 	SQUnsignedInteger _uiRef;
 	struct SQWeakRef *_weakref;
 	virtual void Release()=0;
+
+	/* Placement new/delete to prevent memory leaks if constructor throws an exception. */
+	inline void *operator new(size_t size, SQRefCounted *place)
+	{
+		place->size = size;
+		return place;
+	}
+
+	inline void operator delete(void *ptr, SQRefCounted *place)
+	{
+		SQ_FREE(ptr, place->size);
+	}
+
+	/* Never used but required. */
+	inline void operator delete(void *ptr) { NOT_REACHED(); }
+
+private:
+	size_t size;
 };
 
 struct SQWeakRef : SQRefCounted
--- a/src/script/squirrel.cpp
+++ b/src/script/squirrel.cpp
@@ -67,7 +67,7 @@ struct ScriptAllocator {
 	 * @param requested_size The requested size that was requested to be allocated.
 	 * @param p              The pointer to the allocated object, or null if allocation failed.
 	 */
-	void CheckAllocation(size_t requested_size, const void *p)
+	void CheckAllocation(size_t requested_size, void *p)
 	{
 		if (this->allocated_size > this->allocation_limit && !this->error_thrown) {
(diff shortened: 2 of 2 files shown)


LordAro (approved):

PeterN:
Setting `place->size` inside `operator new()` is undefined behaviour as the object has not yet been constructed.

glx22:
> Setting `place->size` inside `operator new()` is undefined behaviour as the object has not yet been constructed.

And #14278 is a result of this undefined behaviour.

PR #9594, opened 2021-10-02
Fix #9591: Update station docking tiles upon placing a water object on a docking tile

### LINKED ISSUE #9591: [Bug]: Building a water object on a plain water docking tile does not correctly remove the docking tile
### Version of OpenTTD

master

### Expected result

Station::docking_station is always correct.

### Actual result

Building a water object on a plain water docking tile does not correctly remove the docking tile,
in that the station(s) associated with the docking tile do not have their docking_station tile areas updated.
This can result in multiplayer desyncs. This issue is detected by the cache checks in CheckCaches.

This was noticed because of griefers placing fake water tiles around docks, which later resulted in docking_station mismatches being logged.

### Steps to reproduce

* Set the desync log level to at least 2.
* Build a dock.
* Build water objects (e.g. from decoration NewGRF such as AuzWaterAndLandscapeObjects) on the dock's docking tiles.
* Note that a desync log message will be eventually logged.

SamuXarick:
## Motivation / Problem
Placing a water object on a docking tile doesn't run `CMD_LANDSCAPE_CLEAR` which calls `ClearTile_Water` to do the removal of docking tiles.

## Description
Borrow some part of the code from `ClearTile_Water` that deals with the removal of docking tiles and use it during object construction.

## Checklist for review

### DIFF
--- a/src/object_cmd.cpp
+++ b/src/object_cmd.cpp
@@ -32,6 +32,7 @@
 #include "date_func.h"
 #include "newgrf_debug.h"
 #include "vehicle_func.h"
+#include "station_func.h"
 
 #include "table/strings.h"
 #include "table/object_land.h"
@@ -121,7 +122,9 @@ void BuildObject(ObjectType type, TileIndex tile, CompanyID owner, Town *town, u
 			Company::Get(owner)->infrastructure.water++;
 			DirtyCompanyInfrastructureWindows(owner);
 		}
+		bool remove = IsDockingTile(t);
 		MakeObject(t, owner, o->index, wc, Random());
+		if (remove) RemoveDockingTile(t);
 		MarkTileDirtyByTile(t);
 	}
 
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -3106,8 +3106,8 @@ bool AfterLoadGame()
 		}
 	}
 
-	if (IsSavegameVersionBefore(SLV_DOCK_DOCKINGTILES)) {
-		/* All tiles around docks may be docking tiles. */
+	if (IsSavegameVersionBefore(SLV_REPAIR_OBJECT_DOCKING_TILES)) {
+		/* Placing objects on docking tiles was not updating adjacent station's docking tiles. */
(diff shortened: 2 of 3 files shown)


glx22 (commented):
Seems correct, but I think a savegame bump is needed, and an update of https://github.com/OpenTTD/OpenTTD/blob/master/src/saveload/afterload.cpp#L3109 with the new version.

src/saveload/afterload.cpp
```diff
@@ -3106,8 +3106,9 @@ bool AfterLoadGame()
 		}
 	}
 
-	if (IsSavegameVersionBefore(SLV_DOCK_DOCKINGTILES)) {
-		/* All tiles around docks may be docking tiles. */
+	if (IsSavegameVersionBefore(SLV_REPAIR_OBJECT_DOCKING_TILES)) {
+		/* Placing objects on docking tiles was not updating adjacent station's
+		 * docking tiles. */
```
LordAro: unnecessary line break

LordAro (dismissed):

src/object_cmd.cpp
```diff
@@ -121,7 +122,9 @@ void BuildObject(ObjectType type, TileIndex tile, CompanyID owner, Town *town, u
 			Company::Get(owner)->infrastructure.water++;
 			DirtyCompanyInfrastructureWindows(owner);
 		}
+		bool remove = IsDockingTile(t);
 		MakeObject(t, owner, o->index, wc, Random());
+		if (remove) RemoveDockingTile(t);
```
TrueBrain, line 127: I honestly don't know the answer, but why does this need a variable? Can't we just do `RemoveDockingTile` before `MakeObject`?

JGRennison: No. RemoveDockingTile examines the state of the tile.

TrueBrain: So `RemoveDockingTileIfIFeelLikeIt`, gotcha. Tnx :)

TrueBrain (approved):
@glx22 and @LordAro seems to think this is correct, so lets go!

PR #9600, opened 2021-10-03
Add: [Network] Keep the refresh button in lowered state while refreshing

TrueBrain:
## Motivation / Problem

I noticed on the Game Coordinator usage statistics, that we see a lot of "Aborted" connections. This happens when a new connection to the same server is being created while another one is still pending. Initially I couldn't figure out when this happens, till I was looking into an unrelated but in the Game Coordinator:

Especially when you have a TURN-connection, one that doesn't allow STUN, refreshing a server takes a few seconds. There is no visual feedback in this GUI to tell you the refresh is still pending, as the button goes up directly after pressing it. This gives you the idea you misclicked or something else went wrong, so you press again. And again. And again.

This PR sets out to resolve this.

As written in the commit message:

```
This gives user visual feedback that the refresh is still pending, and
prevents people from clicking again and again thinking nothing is
happening. This is especially true for connections that fall back to
TURN, as that takes a few seconds to kick in.

Additionally, prevent clicking on the button again while a refresh
is pending. This is only delaying a successful result.
```

## Description

Simply track that the server is being refreshed, and keep the state lowered while this holds true. Additionally, prevent clicking on the button again to give a pretty solid UX experience.

`retries` wasn't used anymore, so I removed that variable while I was at it.

## Checklist for review

### DIFF
--- a/src/network/network.cpp
+++ b/src/network/network.cpp
@@ -635,6 +635,7 @@ public:
 	{
 		NetworkGameList *item = NetworkGameListAddItem(connection_string);
 		item->status = NGLS_OFFLINE;
+		item->refreshing = false;
 
 		UpdateNetworkGameWindow();
 	}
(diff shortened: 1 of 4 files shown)


LordAro (approved):
LGTM

PR #9613, opened 2021-10-16
Codechange: "set but not used" warning when disabling assert()

TrueBrain:
## Motivation / Problem

During preparing for 12.0, the PR that disables `assert()` notice one new warning. Let's make a clean release, and address that warning.

## Description

`DebugCheckSanity()` is unused when asserts are disabled.

Owh, and we should be using `WITH_ASSERT`, not `NDEBUG` for these cases.

## Checklist for review

### DIFF
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -609,6 +609,7 @@ bool OrderList::IsCompleteTimetable() const
 	return true;
 }
 
+#ifdef WITH_ASSERT
 /**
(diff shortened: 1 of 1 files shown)


glx22:
The sanity check is actually disabled for any release build even with asserts, but maybe it should be enabled in this case.
If we go that way, `#ifndef NDEBUG` in this change and on line 1787 must be replaced with `#ifdef WITH_ASSERT` like everywhere in the code.

glx22 (approved):

PR #9616, opened 2021-10-17
Feature: Railways can be created "through" bridges, tunnels, stations…

Kuhnovic:
…. And starting from depot tiles.

## Motivation / Problem

RX3AJb1SRU

When laying track using the AutoRail tool, I often find myself accidentally starting on a bridge, tunnel or station tile. This invalidates the entire action and no rail is built. The other way around, i.e. starting at an empty tile and ending on say a station tile, that works just fine. This can be quite annoying for example when you are trying to create lots of bridges / tunnels in a crowded map.

## Description

My changes allow the user to start the AutoRail tool on train station tiles, train depot tiles, bridgehead tiles and tunnel entrance tiles. The user must drag in the right direction, or can be 45 degrees off.

In addition, the user can drag "through" a tunnel, bridge or station. This obviously doesn't apply to depots. If there are multiple tunnels or a tunnel + bridge above each other, we would follow the route that a train would take when driving along the dragged line. In other words, we simply enter the first tunnel or bridge, and exit out of the other end.

[EDIT] This now only applies when one is creating new track. When deleting track, basically we just "delete whatever we can". This approach is more in line with how we currently do things (before my changes).

Here's an impression of what all of this looks like:

1RCFk94oG3

[EDIT] And now also for roads/trams:
SJXEnptlxb

## Limitations

I like the visualization, but I need to jump through a number of hoops to get there. Worst case I can ditch the new visualization.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -574,6 +574,13 @@ CommandCost CmdBuildSingleRail(TileIndex tile, DoCommandFlag flags, uint32 p1, u
 		}
 
 		default: {
+			/* Just to get clearer error messages */
+			if (GetTileType(tile) == MP_TUNNELBRIDGE || GetTileType(tile) == MP_STATION)
+			{
+				CommandCost ret = CheckTileOwnership(tile);
+				if (ret.Failed()) return ret;
+			}
+
 			/* Will there be flat water on the lower halftile? */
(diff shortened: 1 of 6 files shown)


2TallTyler:
#11089 accomplishes some of this in a cleaner way, but I'd still be interested in seeing the "drag through" feature finished someday. 😃

PR #9620, opened 2021-10-17
Fix 3046a6ce: [Preview] building preview failed to patch LZMA

TrueBrain:
## Motivation / Problem

In 3046a6ce I assumed that the CI job and preview job would work the same way. It didn't.

## Description

As by commit message:
```
The job was started under a non-priv user, which did not have
permission to patch the emscripten files required to make LZMA
work.
```

## Limitations

There is no real way of testing this anymore, other than it now did build on my fork. But the publishing part etc can only be checked after merging. Tagging this PR with "preview" doesn't work, as this change needs to be in master before it is picked up. (security blabla).

## Checklist for review

### DIFF
--- a/.github/workflows/preview_build.yml
+++ b/.github/workflows/preview_build.yml
@@ -13,8 +13,6 @@ jobs:
     container:
       # If you change this version, change the number in the cache step too.
       image: emscripten/emsdk:2.0.31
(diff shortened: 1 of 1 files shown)


glx22 (approved):
I don't know why preview used a non-root user, but let's try with root.

PR #9628, opened 2021-10-19
Feature: Wide rivers

2TallTyler:
## Motivation / Problem

Real-life rivers and watersheds are shaped like enormous trees: thin branches gathering together into a wide trunk which flows to the sea. 

OpenTTD rivers often follow the same branching structure, but don't get any wider.

## Description

tasty

This started as a [JGRPP patch](https://github.com/JGRennison/OpenTTD-patches/pull/275) by @VacuumBreather (a.k.a. Lt. Joker).

It makes 10% of rivers longer and has them get wider as they get further from the source.

I added Doxygen, changed some code style items and added terraforming to keep rivers wide as they flow down slopes.

## Limitations

I've tested map generation pretty thoroughly for odd generation or possible terraforming under existing water tiles (which is pretty glitchy) but it's possible I could have missed a corner case.

## Checklist for review

### DIFF
--- a/src/direction_type.h
+++ b/src/direction_type.h
@@ -102,10 +102,12 @@ template <> struct EnumPropsT<DiagDirection> : MakeEnumPropsT<DiagDirection, byt
  * @see DirDiff
  */
 enum DiagDirDiff {
+	DIAGDIRDIFF_BEGIN   = 0,        ///< Used for iterations
 	DIAGDIRDIFF_SAME    = 0,        ///< Same directions
 	DIAGDIRDIFF_90RIGHT = 1,        ///< 90 degrees right
 	DIAGDIRDIFF_REVERSE = 2,        ///< Reverse directions
 	DIAGDIRDIFF_90LEFT  = 3,        ///< 90 degrees left
+	DIAGDIRDIFF_END,                ///< Used for iterations
 };
 
 /** Allow incrementing of DiagDirDiff variables */
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -32,6 +32,7 @@
 #include "saveload/saveload.h"
 #include "framerate_type.h"
 #include "landscape_cmd.h"
+#include "terraform_cmd.h"
 #include "station_func.h"
 #include <array>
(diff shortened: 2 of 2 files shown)


2TallTyler:
Fixed pointer format to follow code style, since apparently my VS2019 has been configured to auto-correct it to the wrong format for months... 😕

michicc:
I like this idea in principle, even if somebody™️ will declare it the end of the world.

One thing that sometimes looks a bit ugly is the way larger rivers meet the coast due to the fact that river tiles can't be on all slopes. I've seen it quite often that the final tile on the coast is just one tile wide due to the slopes. Ideally, some spot terraforming could create a proper wider slope.

Code wise it looks fine. One could remove the globals in favour of using the AyStar user data (like e.g. https://gist.github.com/michicc/91ac9d01052605150e6d31ba781ff01d), but this is definitely a bikeshed thing.

2TallTyler:
I'll see if I can figure out some slope terraforming.

2TallTyler:
Thanks for the gist; I've adopted it and also added slope terraforming.

Everyone: please try and break this! I've generated and inspected a lot of maps and haven't seen any graphical glitches, but earlier versions of my terraforming did undermine river tiles without deleting the water, leading to holes in the world like this:

glitch

JGRennison:
@2TallTyler In River_FoundEndNode you are interleaving calls to MakeRiver along the tiles found by the A-star path with calls to RiverMakeWider which modify the landscape. This is likely to result in your found river path no longer having tiles of the correct slope. Also, as written widening won't be done where the path is shared with an existing narrower river.
It may be worth looking into using multiple passes instead.

2TallTyler:
Thanks for the insight. It took me a while to understand what you meant, but that (and a fix for the broken line of code mentioned above) seems to have solved the issue. Additional testing appreciated!

michicc:
Needs a rebase.

michicc (approved):

Zispah:
You know, it would be better if an option is added instead of replacing.

PR #8891, opened 2021-03-24
Introduce snow/desert-coverage and custom terrain type, and move "maximum map height" to settings only

TrueBrain:
Closes #8879

Preview available here: https://preview.openttd.org/pr8891/ (play this change in your browser!)
Binaries available here: https://www.openttd.org/downloads/openttd-branches/pr8891/latest.html (if you want to try large maps)

## Motivation / Problem

Where even to start ...

If [this thread](https://www.tt-forums.net/viewtopic.php?f=29&t=88531) made one thing clear, is that many people do not understand a few of the settings in the NewGame GUI, or have a hard time using them in a useful matter. And I have been very vocal that I do not think this is a user issue, but an UX one. This Pull Requests tries to correct the most common mistakes people make.

First off, I would like to serious thank reldrid for engaging in a constructive conversation about all of this. It makes such a difference to have such conversations; seriously, tnx. Also tnx to andythenorth for some gravedigging in old threads and frosch123 for sparring about solutions and possibilities. This hasn't been easy ;)
Sadly, there was also a lot of noise on the line, people using a lot of words to say very little, lot of screenshots without context, denying facts, etc. This was absolutely not helping .. it took 3 weeks before someone managed to tell me settings where 1.11-RC1 was producing less mountains than 1.10. It was a massive drain of energy to keep engaging in those conversations.

Anyway.

To start, lets dive in: the by-now famous "maximum map height" setting on the NewGame GUI. Looking through the [MHL thread](https://www.tt-forums.net/viewtopic.php?f=33&t=40844) (tnx andythenorth for going through it), it got added "because it can be useful", under protest of the original author. It seems that it was not truly understood what it did, or what the effect would have on the player. Well, now 7 years later, we have the answer: not good :D
This setting set the height limit of the map to this value. It does not influence TGP, except that he too cannot build above this limit. So player increase the value, assuming the maps would produce higher mountains, and in some regards it is. But mostly, it isn't. People deduce all kind of weird conclusions based on it, and now we have two worlds: reality, and some vague misconception of reality. This is absolutely human, but makes the gap between perception and reality rather big.
So, let's stop annoying users, and lets split apart these settings. With this PR we now have the following settings:

- `map_height_limit` - The true limit of the map. Nothing can build over this. Defaults to "auto" and is a setting only (no longer in NewGame GUI) More on this later.
- `heightmap_height` - The brightest pixel in a heightmap will be this height. The rest is scaled in between. Replaces "maximum map height" in GUI.
[...]

PR #7340, opened 2019-03-08
Change: Replace tropic landscape map generator rules

andythenorth:
In sub-tropic climate, TGP produces very boring flat maps. These are really dull to play (apologies to whoever wrote this in TGP) :)

I've 'improved' it by deleting the sub-tropic specific code, and using the same height rules as temperate climate. This makes for much more playable maps. Desert and rainforest generation work fine. 

Gravedug from https://www.tt-forums.net/viewtopic.php?f=33&t=74647&p=1165875&hilit=tropic+landscape#p1165875

PR #9634, opened 2021-10-20
Change: Don't use special map gen for tropic

2TallTyler:
## Motivation / Problem

Subtropic maps use different map generation than other climates (so does subarctic, but that's a different story) which makes it impossible to get mountainous terrain from generated maps. This limits the types of games which can be played, even as #8891 makes it possible to have all-rainforest maps. Want a hilly rainforest? Your only option is to use a heightmap.

Eliminating the special subtropic rules has been proposed before in #7340 and an associated [forum topic](https://www.tt-forums.net/viewtopic.php?f=33&t=74647&p=1165875&hilit=tropic+landscape#p1165875), but was closed by the stale bot. Now that landscape generation has been simplified in #8891, I think it's worth revisiting.

## Description

TGP uses separate rules for subtropic and subarctic:
* Subtropic: `Half of tiles should be at lowest (0..25%) heights`
* Subarctic: `Redistribute heights to have more tiles at highest (75%..100%) range`

As shown in https://github.com/OpenTTD/OpenTTD/pull/8891#issuecomment-807341140, this appears to drag down the peaks and make the entire map flatter.

Leaving alone subarctic for now, I don't see a need for separate subtropic rules to flatten maps. Players can just choose a flatter terrain type.

Generating a map with these settings:
settings

Current subtropic generation:
tropic_current

Proposed subtropic generation:
tropic_new

## Checklist for review

### DIFF
--- a/src/tgp.cpp
+++ b/src/tgp.cpp
@@ -467,6 +467,7 @@ static void HeightMapSineTransform(height_t h_min, height_t h_max)
 		switch (_settings_game.game_creation.landscape) {
 			case LT_TOYLAND:
 			case LT_TEMPERATE:
+			case LT_TROPIC:
 				/* Move and scale 0..1 into -1..+1 */
 				fheight = 2 * fheight - 1;
 				/* Sine transform */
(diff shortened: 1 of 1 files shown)


TrueBrain:
This feels like a very cheap solution to a bigger problem, at the cost of anyone that likes the tropic mapgen :) I am not a fan of this change.

I think, and we (other developers) have been talking about this for a while now, it makes much more sense to change the approach of our "new game" window. Offer a few solid presets, and a big "advanced" button that allows you to turn every nob we can find in any way you like. One of those can be if a percentage of the map should be some type, like desert, water, snow, ... That brings in much more variation, and people can truly experiment with different forms of map generation. It even allows us to add more of these things in there, for people to find their niche.
And of course we should allow to share those presets via BaNaNaS.

But removing it outright .. that sounds like a step backwards, not forwards.

2TallTyler:
I'm totally on-board with new map generation presets and tools. And I don't want to eliminate the ability of a player to generate the map they want — I'm getting plenty of hate for my signal GUI PR right now. :)

But in the meantime, a subtropic Alpinist map looks like a temperate Hilly map to me. Changing the terrain type seems like a better way to choose the terrain type, rather than having special hidden rules which does something different than I expect. This disconnect between my choices in the New Map menu and what actually generates are what frustrates me. In any case, with or without the proposed new map generation presents/tools, I would like to see these hidden rules eliminated so that what I select is what gets generated.

TrueBrain:
It was never the idea that climates render their heightmap the same. Diversity was what the idea was with those climates. Basically, it is climates++. And I can understand that you struggle with that. But the fact still remains that your solution is the cheapest of all :D (sorry, I don't mean this in a bad way ofc).

2TallTyler:
After some discussion on Discord, let's wait for the new map generation tools to do this.

frosch123:
I don't see the point in removing the tropic-special while keeping the arctic-special.
Both arctic and tropic shift the distribution of low, medium and high altitudes.

So to allow more map options, I guess you want some "elevation gamma" setting:
* Climate-specific (default)
* Prefer low lands
* Normal
* Prefer high lands

No idea whether "gamma" is a good term, probably not, but image processing uses some similar stuff to adjust brightness/contrast/gamma.

TrueBrain:
I like your suggestion @frosch123 ; and I guess we can do with yet-another-map-generation-setting :P And it fits a lot more with any future plans. Sounds like a nice compromise between the two, not @2TallTyler ? :)

PR #9635, opened 2021-10-20
Change: Rivers may start in desert when no rainforest exists

2TallTyler:
## Motivation / Problem

Rivers cannot currently start in the desert. This is normally fine, but when the desert percentage is 100%, it is impossible for rivers to generate since the only non-desert surrounds ocean and isn't wide enough to fit the minimum length river starting at a higher elevation.

## Description

When the desert percentage is 100, allow rivers to start in the desert. Lakes are also allowed to generate, as the game already generates inland sea as oases into which rivers can flow.

Not that OpenTTD always reflects reality, but the real-world precedent for rivers starting in the desert would be water emerging from an underground spring like those which feed [qanats](https://en.wikipedia.org/wiki/Qanat).

If a player wants a 100% desert map without rivers, they can set rivers to "none".

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -1021,8 +1021,8 @@ static bool FindSpring(TileIndex tile, void *user_data)
 	int referenceHeight;
 	if (!IsTileFlat(tile, &referenceHeight) || IsWaterTile(tile)) return false;
 
-	/* In the tropics rivers start in the rainforest. */
-	if (_settings_game.game_creation.landscape == LT_TROPIC && GetTropicZone(tile) != TROPICZONE_RAINFOREST) return false;
+	/* In the tropics rivers start in the rainforest, unless there is no rainforest. */
+	if (_settings_game.game_creation.landscape == LT_TROPIC && GetTropicZone(tile) != TROPICZONE_RAINFOREST && _settings_game.game_creation.desert_coverage < 100) return false;
 
 	/* Are there enough higher tiles to warrant a 'spring'? */
 	uint num = 0;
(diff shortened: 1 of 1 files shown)


LordAro:
I like it. Though I wonder whether it might be interesting to allow rivers in deserts at a lower percentage. Or perhaps even in proportion to the amount of desert on the map
i.e.
100% desert map - 100% rivers start in desert
50% desert map - 50% rivers start in desert
etc

TrueBrain:
Similar with #9634, I really wonder if these changes are a good idea. As why 100? Why not 90? or 99? What if I just always want to have rivers start in desert? :D

Here too, I think this would work much better if we just freely allow users to pick where/when they want rivers. It would simplify our code a lot, and removes all these special cases (basically pushing them to the user). I know it is a bit annoying, as "that isn't there". But I am afraid that all these point-solutions just add up to even less maintainable code.

But this is just my opinion. @LordAro clearly disagrees :D

Eddi-z:
> 100% desert map - 100% rivers start in desert
> 50% desert map - 50% rivers start in desert etc

that makes no sense. if 50% of the map is covered in desert, and river starting points are distributed randomly, then 50% of rivers starting in desert means the same as "no restriction at all".

2TallTyler:
I chose 100% to catch this edge case where rivers are truly impossible. On a tall enough heightmap, 99% desert may still have some green peaks where rivers can start.

I'm on board with adding an option for river start locations in the new map generation toolkit. In fact, I think a lot of the options @reldred [added to JGRPP](https://github.com/JGRennison/OpenTTD-patches/pull/231) should be in there. Closing this until the new toolkit is created.

PR #9638, opened 2021-10-20
Feature: add setting to disable news about competitor's accidents

grenoult:
## Motivation / Problem

During a game of OpenTTD with many competitors, we can receive a lot of news regarding competitors vehicles involved in accidents. This is especially true when vehicles have low reliability and are rarely replaced:
- road vehicles breaking down on a rail crossing then being crashed by a train,
- airplanes failing to land on an airport and crashing.

This PR addresses the issue of receiving too many news regarding competitors vehicle crash where it doesn't bring any value or meaning to the main player. As a OpenTTD player, I want to know when my vehicles are involved in a crash but I may choose to not receive notification if competitors vehicles are destroyed.

## Description

A new setting has been added "Accidents of competitor's vehicles". It's "Full" by default. When turned off, no news will be generated when a competitor vehicle is involved in a crash. When it's "Summary", only the short version will be displayed at the bottom of the screen.
A crash, in the context of this PR, is when a road vehicle is hit by a train or when an airplane misses landing at an airport.

- Full: A news item is generated when setting is Full (by default) and a competitor vehicle is crashed (in this example, a plane crash):
image

- Summary: Short news are generated when setting is "Summary" and a competitor vehicle (red) is crashed by a train:
image

- Off: no news are generated when a competitor vehicle is crashed when settings is "Off"
image

## Limitations

- other crashes, such as UFOs and similar events are ignored as part of this PR. This has been left out because it's rare enough to cause a disruption to the player. That being said, one could argue we could update this in this PR and I could update it (probably in another PR) if requested.

## Checklist for review

- This PR adds a new translation in english.txt file. From the [documentation](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md#i-want-to-add-a-new-string), nothing else is required at this stage.

### DIFF
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -242,6 +242,7 @@ struct NewsSettings {
 	uint8 arrival_player;                                 ///< NewsDisplay of vehicles arriving at new stations of current player
 	uint8 arrival_other;                                  ///< NewsDisplay of vehicles arriving at new stations of other players
 	uint8 accident;                                       ///< NewsDisplay of accidents that occur
+	bool accident_other_company_vehicle;				  ///< if a vehicle from another company is involved in an accident, show a news item
 	uint8 company_info;                                   ///< NewsDisplay of general company information
 	uint8 open;                                           ///< NewsDisplay on new industry constructions
 	uint8 close;                                          ///< NewsDisplay about closing industries
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -242,7 +242,7 @@ struct NewsSettings {
 	uint8 arrival_player;                                 ///< NewsDisplay of vehicles arriving at new stations of current player
 	uint8 arrival_other;                                  ///< NewsDisplay of vehicles arriving at new stations of other players
 	uint8 accident;                                       ///< NewsDisplay of accidents that occur
-	bool accident_other_company_vehicle;				  ///< if a vehicle from another company is involved in an accident, show a news item
+	uint8 accident_other;                                 ///< NewsDisplay if a vehicle from another company is involved in an accident
 	uint8 company_info;                                   ///< NewsDisplay of general company information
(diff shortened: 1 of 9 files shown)


frosch123:
For "arrival of first vehicle" and "production changes of industries" there are separate settings with "full"/"summary"/"off".
Shouldn't this apply to "accidents" as well?
I guess "disasters" (ufos etc) would be a separate setting as well then.

src/settings_type.h
```diff
@@ -242,6 +242,7 @@ struct NewsSettings {
 	uint8 arrival_player;                                 ///< NewsDisplay of vehicles arriving at new stations of current player
 	uint8 arrival_other;                                  ///< NewsDisplay of vehicles arriving at new stations of other players
 	uint8 accident;                                       ///< NewsDisplay of accidents that occur
+	bool accident_other_company_vehicle;				  ///< if a vehicle from another company is involved in an accident, show a news item
```
James103:
```suggestion
	uint8 accident_other_company_vehicle;                  ///< NewsDisplay of accidents involving competitor's vehicles
```

grenoult:
> For "arrival of first vehicle" and "production changes of industries" there are separate settings with "full"/"summary"/"off". Shouldn't this apply to "accidents" as well? I guess "disasters" (ufos etc) would be a separate setting as well then.

Good point @frosch123 , I'll update my PR, thanks.

src/settings_type.h
James103: This appears to be resolved in your latest commit.

grenoult:
Closing this branch. A new PR has been created with correct commit messages: https://github.com/OpenTTD/OpenTTD/pull/9653

LordAro:
(For future reference, you can fix this with [rebasing](https://git-scm.com/docs/git-rebase) without having to open a new PR/make a new branch)

PR #9639, opened 2021-10-21
Feature: Ctrl-click to remove fully autoreplaced vehicles from list

2TallTyler:
## Motivation / Problem

When autoreplacing vehicles with another type, the old vehicle type is greyed out when none remain. Removing these empty vehicles from the list is fairly annoying, especially if you have a lot of vehicles to remove — when I play, I forget to autoreplace on a regular basis and tend to upgrade a lot of vehicle types at once.
1. Click on the vehicle
2. Move the mouse to the "Stop Replacing Vehicles" button and click
3. Repeat

## Description

When buying vehicles, we can Ctrl-click on a vehicle to hide it from the menu in a single action.

Now we can do that with autoreplace, but only for vehicles for which none remain.

This works identically to the Stop Replacing Vehicles button, so it respects groups.

## Checklist for review

### DIFF
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -578,6 +578,16 @@ public:
 				size_t engine_count = this->engines[click_side].size();
 
 				EngineID e = engine_count > i ? this->engines[click_side][i] : INVALID_ENGINE;
+
+				/* If Ctrl is pressed on the left side and we don't have any engines of the selected type, stop autoreplacing.
+				 * This is most common when we have finished autoreplacing the engine and want to remove it from the list. */
+				if (click_side == 0 && _ctrl_pressed && e != INVALID_ENGINE &&
+					(GetGroupNumEngines(_local_company, sel_group, e) == 0 || GetGroupNumEngines(_local_company, ALL_GROUP, e) == 0)) {
+						EngineID veh_from = e;
+						DoCommandP(0, this->sel_group << 16, veh_from + (INVALID_ENGINE << 16), CMD_SET_AUTOREPLACE);
+						break;
+				}
+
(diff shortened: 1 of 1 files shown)


LC-Zorg:
> When buying vehicles, we can Ctrl-click on a vehicle to hide it from the menu in a single action.

I didn't know this feature before. :) I tried to find arguments against, but I don't see them for this moment. Now that there's already a similar use of Ctrl + click, it seems quite sensible and useful.
Btw. The hiding feature doesn't work in the vehicle replacement window. It goes beyond this PR, but it would be useful and make a logical whole together.

LordAro (approved):
Sure, why not

buttercup5:
I had to go a LONG way back (2005), but it used to be that when there were no more vehicles left to replace, they would automatically be removed from the list. 

The last version where you can see this behaviour is 0.4.0.1. The release notes for 0.4.5 mention a complete rewrite of the autoreplace code so perhaps there was a technical reason for the change. If not for a technical reason, I can't think what purpose the current behaviour serves. No further actions can be performed once all the selected vehicles have been replaced, and I can't see why a player would need to be reminded of the what the old vehicle was.

I'm probably overlooking something obvious.

2TallTyler:
We discussed this on Discord, and at least one user mentioned that having the replaced vehicles hang around acted as a confirmation that the replacement was complete. I disagree and would prefer them to disappear automatically, but Ctrl-clicking is a good compromise. 😄

buttercup5:
Ah, that's fair enough. Maybe a news message would serve that purpose better though.

In any case, it's not a big issue and, like you said, ctrl+clicking is a good compromise.

PR #9642, opened 2021-10-22
Feature: Orientation of rail and road depots can be changed

Kuhnovic:
## Motivation / Problem

It's all too easy to place a rail depot in the wrong orientation. Railway station tiles can be "rotated" by simply building them again, I figured such a feature would be great for depots as well.

... I took this concept even further and created an option to automatically determine the depot orientation. More information below.

## Description

If you create a depot, and you accidentally got the orientation wrong, then just select the correct orientation and click again. The existing depot is rotated, it is not destroyed/recreated, which means orders containing the depot stay intact.

I5uB4QsmEe

**Update 1** Tried it for ship depots as well, but this was problematic. I ended up removing it again.

**Update 2**

I got the idea to have the option to automatically orient a rail or road depot when you place it, based on whether it could be connected to any nearby rail. **But I dediced to remove this again to keep the PR clean, and because it wasn't really finished.**

## Limitations

This only works on depots of the same (rail) type. This is by design.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -989,24 +989,44 @@ CommandCost CmdBuildTrainDepot(DoCommandFlag flags, TileIndex tile, RailType rai
 		cost.AddCost(_price[PR_BUILD_FOUNDATION]);
 	}
 
-	cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
-	if (cost.Failed()) return cost;
+	/* Allow the user to rotate the depot instead of having to destroy it and build it again */
+	bool rotate_existing_depot = false;
+	if (IsRailDepotTile(tile) && railtype == GetRailType(tile)) {
+		CommandCost ret = CheckTileOwnership(tile);
+		if (ret.Failed()) return ret;
 
-	if (IsBridgeAbove(tile)) return_cmd_error(STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST);
+		if (dir == GetRailDepotDirection(tile)) return_cmd_error(STR_ERROR_ALREADY_BUILT);
 
-	if (!Depot::CanAllocateItem()) return CMD_ERROR;
+		ret = EnsureNoVehicleOnGround(tile);
+		if (ret.Failed()) return ret;
+
+		rotate_existing_depot = true;
+	}
+
+	if (!rotate_existing_depot) {
+		cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
+		if (cost.Failed()) return cost;
+
+		if (IsBridgeAbove(tile)) return_cmd_error(STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST);
+
+		if (!Depot::CanAllocateItem()) return CMD_ERROR;
+	}
 
 	if (flags & DC_EXEC) {
-		Depot *d = new Depot(tile);
-		d->build_date = TimerGameCalendar::date;
+		if (rotate_existing_depot) {
+			SetRailDepotExitDirection(tile, dir);
+		} else {
+			Depot *d = new Depot(tile);
+			d->build_date = TimerGameCalendar::date;
 
-		MakeRailDepot(tile, _current_company, d->index, dir, railtype);
+			MakeRailDepot(tile, _current_company, d->index, dir, railtype);
+			MakeDefaultName(d);
+
+			Company::Get(_current_company)->infrastructure.rail[railtype]++;
+			DirtyCompanyInfrastructureWindows(_current_company);
+		}
+
 		MarkTileDirtyByTile(tile);
-		MakeDefaultName(d);
-
-		Company::Get(_current_company)->infrastructure.rail[railtype]++;
-		DirtyCompanyInfrastructureWindows(_current_company);
-
 		AddSideToSignalBuffer(tile, INVALID_DIAGDIR, _current_company);
 		YapfNotifyTrackLayoutChange(tile, DiagDirToDiagTrack(dir));
 	}
--- a/src/rail_map.h
+++ b/src/rail_map.h
@@ -530,19 +530,37 @@ static inline void MakeRailNormal(Tile t, Owner o, TrackBits b, RailType r)
 	t.m8() = r;
 }
 
-
-static inline void MakeRailDepot(Tile t, Owner o, DepotID did, DiagDirection d, RailType r)
+/**
+ * Sets the exit direction of a rail depot.
+ * @param tile Tile of the depot.
+ * @param dir  Direction of the depot exit.
+ */
+static inline void SetRailDepotExitDirection(Tile tile, DiagDirection dir)
 {
-	SetTileType(t, MP_RAILWAY);
-	SetTileOwner(t, o);
-	SetDockingTile(t, false);
-	t.m2() = did;
-	t.m3() = 0;
-	t.m4() = 0;
-	t.m5() = RAIL_TILE_DEPOT << 6 | d;
-	SB(t.m6(), 2, 4, 0);
-	t.m7() = 0;
-	t.m8() = r;
+	assert(IsRailDepotTile(tile));
+	SB(tile.m5(), 0, 2, dir);
+}
+
+/**
+ * Make a rail depot.
+ * @param tile      Tile to make a depot on.
+ * @param owner     New owner of the depot.
+ * @param depot_id  New depot ID.
+ * @param dir       Direction of the depot exit.
+ * @param rail_type Rail type of the depot.
+ */
+static inline void MakeRailDepot(Tile tile, Owner owner, DepotID depot_id, DiagDirection dir, RailType rail_type)
+{
+	SetTileType(tile, MP_RAILWAY);
+	SetTileOwner(tile, owner);
+	SetDockingTile(tile, false);
+	tile.m2() = depot_id;
+	tile.m3() = 0;
+	tile.m4() = 0;
+	tile.m5() = RAIL_TILE_DEPOT << 6 | dir;
+	SB(tile.m6(), 2, 4, 0);
+	tile.m7() = 0;
+	tile.m8() = rail_type;
 }
 
 #endif /* RAIL_MAP_H */
--- a/src/road_cmd.cpp
+++ b/src/road_cmd.cpp
@@ -1163,24 +1163,46 @@ CommandCost CmdBuildRoadDepot(DoCommandFlag flags, TileIndex tile, RoadType rt, 
 		cost.AddCost(_price[PR_BUILD_FOUNDATION]);
 	}
 
-	cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
-	if (cost.Failed()) return cost;
+	/* Allow the user to rotate the depot instead of having to destroy it and build it again */
+	bool rotate_existing_depot = false;
+	if (IsRoadDepotTile(tile) && (HasRoadTypeTram(tile) ? rt == GetRoadTypeTram(tile) : rt == GetRoadTypeRoad(tile)))
+	{
+		CommandCost ret = CheckTileOwnership(tile);
+		if (ret.Failed()) return ret;
 
-	if (IsBridgeAbove(tile)) return_cmd_error(STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST);
+		if (dir == GetRoadDepotDirection(tile)) return_cmd_error(STR_ERROR_ALREADY_BUILT);
 
-	if (!Depot::CanAllocateItem()) return CMD_ERROR;
+		ret = EnsureNoVehicleOnGround(tile);
+		if (ret.Failed()) return ret;
+
+		rotate_existing_depot = true;
+	}
+
+	if (!rotate_existing_depot) {
+		cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
+		if (cost.Failed()) return cost;
+
+		if (IsBridgeAbove(tile)) return_cmd_error(STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST);
+
+		if (!Depot::CanAllocateItem()) return CMD_ERROR;
+	}
 
 	if (flags & DC_EXEC) {
-		Depot *dep = new Depot(tile);
-		dep->build_date = TimerGameCalendar::date;
+		if (rotate_existing_depot) {
+			SetRoadDepotExitDirection(tile, dir);
+		} else {
+			Depot *dep = new Depot(tile);
+			dep->build_date = TimerGameCalendar::date;
+			MakeRoadDepot(tile, _current_company, dep->index, dir, rt);
+			MakeDefaultName(dep);
+		}
+
+		MarkTileDirtyByTile(tile);
 
 		/* A road depot has two road bits. */
 		UpdateCompanyRoadInfrastructure(rt, _current_company, ROAD_DEPOT_TRACKBIT_FACTOR);
-
-		MakeRoadDepot(tile, _current_company, dep->index, dir, rt);
-		MarkTileDirtyByTile(tile);
-		MakeDefaultName(dep);
 	}
+
 	cost.AddCost(_price[PR_BUILD_DEPOT_ROAD]);
 	return cost;
 }
--- a/regression/regression/main.nut
+++ b/regression/regression/main.nut
@@ -1107,6 +1107,7 @@ function Regression::Rail()
 	print("    IsRailTile():                  " + AIRail.IsRailTile(33411));
 	print("    BuildRailDepot():              " + AIRail.BuildRailDepot(0, 1));
 	print("    BuildRailDepot():              " + AIRail.BuildRailDepot(33411, 33411));
+	print("    BuildRailDepot():              " + AIRail.BuildRailDepot(33411, 33410));
 	print("    BuildRailDepot():              " + AIRail.BuildRailDepot(33411, 33414));
 	print("    BuildRailDepot():              " + AIRail.BuildRailDepot(33411, 33412));
 	print("    GetRailDepotFrontTile():       " + AIRail.GetRailDepotFrontTile(33411));
@@ -1203,6 +1204,7 @@ function Regression::Road()
 	print("    IsRoadTile():                  " + AIRoad.IsRoadTile(33411));
 	print("    BuildRoadDepot():              " + AIRoad.BuildRoadDepot(0, 1));
 	print("    BuildRoadDepot():              " + AIRoad.BuildRoadDepot(33411, 33411));
+	print("    BuildRoadDepot():              " + AIRoad.BuildRoadDepot(33411, 33410));
 	print("    BuildRoadDepot():              " + AIRoad.BuildRoadDepot(33411, 33414));
 	print("    BuildRoadDepot():              " + AIRoad.BuildRoadDepot(33411, 33412));
 	print("    HasRoadType(Road):             " + AIRoad.HasRoadType(33411, AIRoad.ROADTYPE_ROAD));
--- a/regression/regression/result.txt
+++ b/regression/regression/result.txt
@@ -7498,6 +7498,7 @@ ERROR: IsEnd() is invalid as Begin() is never called
     BuildRailDepot():              false
     BuildRailDepot():              false
     BuildRailDepot():              true
+    BuildRailDepot():              true
     BuildRailDepot():              false
     GetRailDepotFrontTile():       33412
     IsBuildable():                 false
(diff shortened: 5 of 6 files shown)


FLHerne:
I like this in principle, definitely would have saved me some annoyance.

For consistency the same could be applied to ship depots, if one tile of the new depot is over the old one.

Have you considered keeping the same depot when overbuilding? As I understand it, this change will cause any order lists including the original depot to gain '(Invalid Order)', and not allow overbuilding if any vehicle is in the depot.
For the most common use case ("oops I built the wrong angle") that doesn't matter, but preserving depot identity and contents might be useful when altering track layout later. Not sure it would be useful enough to be worth implementing tbh.

Code looks good to me for what it does. Pity about the duplication, but that's not this patch's fault. Regression test failure looks like it just needs the expectations updating.

Kuhnovic:
> I like this in principle, definitely would have saved me some annoyance.
> 
> For consistency the same could be applied to ship depots, if one tile of the new depot is over the old one.
> 
> Have you considered keeping the same depot when overbuilding? As I understand it, this change will cause any order lists including the original depot to gain '(Invalid Order)', and not allow overbuilding if any vehicle is in the depot. For the most common use case ("oops I built the wrong angle") that doesn't matter, but preserving depot identity and contents might be useful when altering track layout later. Not sure it would be useful enough to be worth implementing tbh.
> 
> Code looks good to me for what it does. Pity about the duplication, but that's not this patch's fault. Regression test failure looks like it just needs the expectations updating.

I tried to err on the side of caution by simply automating what the user would do: destroy the depot and create it again. That way I'm pretty certain that I don't mess anything up regarding signals/reservation/vehicles-in-depot etc. You definitely have a point, but I see the situation you describe is a bit of an edge case. I think most people will just use this feature to rectify an initial placement mistake. Then again, everybody has it's own style and habits when it comes to openttd...

I'll try to fix the regression test later this weekend :)

Kuhnovic:
eOWKq5b7p5

Implemented the functionality for ship depots as well. Not entirely sure if we should allow them to "slide" forward as seen in the GIF, but this can easily be fixed if needed.

2TallTyler:
I see nothing wrong with the sliding ship depot.

Kuhnovic:
> Have you considered keeping the same depot

@FLHerne I figured out how to turn the depot without destroying it, at least for road and train depots. No more void orders. I tried it for ship depots as well, but if the depot position changes while a ship is heading for the depot, it messes with the ship controller and the asserts start flying. Obviously it was not created with moving depots in mind :P. I couldn't find a good solution, so I stuck to the simple but crude "destroy and rebuild" method for ship depots.

LC-Zorg:
> I see nothing wrong with the sliding ship depot.

If it changes the depot ID, I see a problem with it - it doesn't concern me, but I know it will be a problem for others.
1. The player may accidentally move the depot and destroy his ships orders
2. The player may move the depot on purpose, unaware that he is destroying his ships orders

Depots may be far enough away from the ship's path that they will never be serviced.
It can be a big problem finding all ships that had this depot in their orders to fix those orders.

If the above problem cannot be avoided, it is better not to add this feature to the shipyards - PR will still be very good. :) But if it succeeds, it will be great :D

Kuhnovic:
I decided to revert my changes for the ship depot, I couldn't get it work reliably. Since ship depots only have 2 orientations to begin with, the chances of the user messing are also smaller.

2TallTyler:
I don't think I've ever placed a ship depot in the wrong orientation, because it's so obvious when aligned incorrectly.

src/rail_cmd.cpp
```diff
@@ -1007,24 +1007,42 @@ CommandCost CmdBuildTrainDepot(TileIndex tile, DoCommandFlag flags, uint32 p1, u
...
+		if (IsBridgeAbove(tile)) return_cmd_error(STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST);
+
+		if (!Depot::CanAllocateItem()) return CMD_ERROR;
+	}
 
 	if (flags & DC_EXEC) {
-		Depot *d = new Depot(tile);
-		d->build_date = _date;
+		if (rotate_existing_depot) {
+			Depot* d = Depot::GetByTile(tile);
```
LordAro: codestyle: `*` goes on the variable, not the type (also below)

LordAro (commented):
Would be nice if the regression test could be extended to actually test that you can overbuild depots - perhaps shouldn't be *too* much effort to figure out the existing code?

Kuhnovic:
> Would be nice if the regression test could be extended to actually test that you can overbuild depots - perhaps shouldn't be _too_ much effort to figure out the existing code?

I tried to add some test cases for rotating depots, but it breaks the regression tests. Somehow I can't run them properly on my own machine, they don't run all the way to the end and the error messages are not helpful. I'll have another try soon.

LordAro:
I think it's fine, it's just that the extra BuildRoadDepot call has increased the ticks slightly, so certain values like vehicle running cost(?) and positions(?) are subtly different. Just update the reference file as necessary :)

Kuhnovic:
> I think it's fine, it's just that the extra BuildRoadDepot call has increased the ticks slightly, so certain values like vehicle running cost(?) and positions(?) are subtly different. Just update the reference file as necessary :)

Are you sure? I can understand that a list order becomes slightly different, but this change in particular looks a little scary to me:

```
9268: -     Should be:          -5947
9268: +     Should be:          -5946'
```

This originates from these lines in regression/main.nut:

```
bank = AICompany.GetBankBalance(AICompany.COMPANY_SELF);

print("  SellVehicle(13):      " + AIVehicle.SellVehicle(13));
print("  IsInDepot():          " + AIVehicle.IsInDepot(12));
print("  IsStoppedInDepot():   " + AIVehicle.IsStoppedInDepot(12));
print("  StartStopVehicle():   " + AIVehicle.StartStopVehicle(12));
print("  IsInDepot():          " + AIVehicle.IsInDepot(12));
print("  IsStoppedInDepot():   " + AIVehicle.IsStoppedInDepot(12));
print("  SendVehicleToDepot(): " + AIVehicle.SendVehicleToDepot(12));
print("  IsInDepot():          " + AIVehicle.IsInDepot(12));
print("  IsStoppedInDepot():   " + AIVehicle.IsStoppedInDepot(12));

bank_after = AICompany.GetBankBalance(AICompany.COMPANY_SELF);

print("  --Accounting--");
print("    GetCosts():         " + accounting.GetCosts());
print("    Should be:          " + (bank - bank_after));
```
I interpreted this as "some money is lost somewhere", it might be rounding error, it might be something more serious.

Kuhnovic:
Scope creep FTW! I've added a feature to automatically determine the orientation of a depot when it's built. More details in the description above.

I reverted my changes for the regression tests so we can at least have green builds and playtesting for now.

src/rail_cmd.cpp
```diff
@@ -979,37 +1037,71 @@ CommandCost CmdBuildTrainDepot(DoCommandFlag flags, TileIndex tile, RailType rai
...
+	}
 
 	if (flags & DC_EXEC) {
-		Depot *d = new Depot(tile);
-		d->build_date = _date;
+		if (rotate_existing_depot) {
+			Depot *d = Depot::GetByTile(tile);
+			assert(d != nullptr);
+			_m[tile].m5 = RAIL_TILE_DEPOT << 6 | dir;
+		}
```
2TallTyler: Wrong style, should be `} else {`

src/rail_cmd.cpp
```diff
@@ -979,37 +1037,71 @@ CommandCost CmdBuildTrainDepot(DoCommandFlag flags, TileIndex tile, RailType rai
...
 		if (!_settings_game.construction.build_on_slopes || !CanBuildDepotByTileh(dir, tileh)) {
 			return_cmd_error(STR_ERROR_FLAT_LAND_REQUIRED);
 		}
 		cost.AddCost(_price[PR_BUILD_FOUNDATION]);
 	}
 
-	cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
-	if (cost.Failed()) return cost;
+	/* Allow the user to rotate the depot instead of having to destroy it and build it again */
+	const DiagDirection old_depot_direction = GetRailDepotDirection(tile);
```
2TallTyler: This might not be a rail depot. `GetRailDepotDirection()` expects a depot tile.

src/rail_cmd.cpp
```diff
@@ -979,37 +1037,71 @@ CommandCost CmdBuildTrainDepot(DoCommandFlag flags, TileIndex tile, RailType rai
...
+	if (connect_to_nearby_track) {
+		if (rotate_existing_depot) {
+			cost.AddCostIfSucceeded(RemoveAllConnectingDepotRail(TileAddByDiagDir(tile, old_depot_direction), old_depot_direction, flags));
+		}
+
+		const TileIndex adjacent_tile = TileAddByDiagDir(tile, dir);
+		if (IsTileType(adjacent_tile, MP_RAILWAY)) {
+			/* Removal of a signal would happen multiple times in test and mess up the cost. As a workaround we ignore it.*/
+			PlaceExtraDepotRail(adjacent_tile, _place_depot_extra_dir[dir], _place_depot_extra_track[dir], railtype, auto_remove_signals, flags);
+			PlaceExtraDepotRail(adjacent_tile, _place_depot_extra_dir[dir + 4], _place_depot_extra_track[dir + 4], railtype, auto_remove_signals, flags);
```
2TallTyler: Could these magic numbers be constants somehow?

src/rail_cmd.cpp
```diff
@@ -979,37 +1037,71 @@ CommandCost CmdBuildTrainDepot(DoCommandFlag flags, TileIndex tile, RailType rai
...
 	cost.AddCost(RailBuildCost(railtype));
+
+	if (connect_to_nearby_track) {
+		if (rotate_existing_depot) {
+			cost.AddCostIfSucceeded(RemoveAllConnectingDepotRail(TileAddByDiagDir(tile, old_depot_direction), old_depot_direction, flags));
+		}
+
+		const TileIndex adjacent_tile = TileAddByDiagDir(tile, dir);
+		if (IsTileType(adjacent_tile, MP_RAILWAY)) {
+			/* Removal of a signal would happen multiple times in test and mess up the cost. As a workaround we ignore it.*/
```
2TallTyler: I don't know what you mean here, but would the `CMD_NO_TEST` flag solve it?

src/rail_gui.cpp
```diff
@@ -106,48 +107,66 @@ static void GenericPlaceRail(TileIndex tile, Track track)
...
-	DIAGDIR_SW, DIAGDIR_NW, DIAGDIR_NE, DIAGDIR_SE,
-	DIAGDIR_NW, DIAGDIR_NE, DIAGDIR_NW, DIAGDIR_NE,
-};
+	const DiagDirection existing_depot_dir = IsRailDepotTile(tile) ? GetRailDepotDirection(tile) : INVALID_DIAGDIR;
+
+	DiagDirection depot_dir = _build_depot_direction;
+	if (_auto_rail_depot_direction) {
+		/* Determine in which directions it is possible to make a depot */
+		std::vector<DiagDirection> possible_dirs;
+		for (DiagDirection dir : {DIAGDIR_NE, DIAGDIR_SE, DIAGDIR_SW, DIAGDIR_NW }) {
```
2TallTyler: Truly a nitpick: inconsistent spacing between the braces and DIAGDIR constants. 🙂

src/rail_gui.cpp
```diff
@@ -106,48 +107,66 @@ static void GenericPlaceRail(TileIndex tile, Track track)
...
+		}
 
-void CcRailDepot(Commands cmd, const CommandCost &result, TileIndex tile, RailType rt, DiagDirection dir)
-{
-	if (result.Failed()) return;
+		/* Determine in which directions there is railtrack track to "snap to". Always a subset of possible_dirs. */
+		std::vector<DiagDirection> candidate_dirs;
+		for (DiagDirection dir : possible_dirs) {
+			const TileIndex next_tile = TileAddByDiagDir(tile, dir);
+			if ((IsLevelCrossingTile(next_tile) && GetCrossingRailAxis(next_tile) == DiagDirToAxis(dir))
```
2TallTyler: This is a whopper of a conditional. Is there a way to simplify it? Maybe an early-continue?

src/road_cmd.cpp
```diff
@@ -1152,25 +1161,59 @@ CommandCost CmdBuildRoadDepot(DoCommandFlag flags, TileIndex tile, RoadType rt,
...
+		if (!Depot::CanAllocateItem()) return CMD_ERROR;
+	}
 
 	if (flags & DC_EXEC) {
-		Depot *dep = new Depot(tile);
-		dep->build_date = _date;
+		if (rotate_existing_depot) {
+			Depot *d = Depot::GetByTile(tile);
+			assert(d != nullptr);
+			_m[tile].m5 = ROAD_TILE_DEPOT << 6 | dir;
```
2TallTyler: Is there not a function which can do this? Or at least some simplification to get rid of the magic number?

src/road_cmd.cpp
```diff
@@ -1152,25 +1161,59 @@ CommandCost CmdBuildRoadDepot(DoCommandFlag flags, TileIndex tile, RoadType rt,
...
+			d->build_date = _date;
```
2TallTyler: Should rotating an existing depot update the build date? Or should it keep the original? (The only consideration I 
 can think of is NewGRF sprite selection)

2TallTyler (commented):
I'm strongly +1 to overbuilding depots to rotate them, and less strongly +1 about the auto rotation. It is convenient, but the only similar thing done automatically is the automatic track and road connections from depots to existing networks, which I'm already not a fan of. To be clear, I'm not against auto orientation and would use it, it just feels somehow a bit surprising to me. 😄

src/road_cmd.cpp
michicc: There should be no direct map access outside of the `*_map.h` files (excepting saveload code), so it should get a function in `road_map.h`. This applies to the rail part, too.

Kuhnovic:
Alright, time to clean this one up. I decided to remove the auto rotation stuff again. It needs more work, and it should probably go in a PR of its own. And then there's the discussion of whether it's actually something we want to have in the game...

I've rebased the code and addressed the review feedback relevant to the code that didn't get removed.

src/road_cmd.cpp
Kuhnovic: Could there be any side effects w.r.t. time tables or some other time-based mechanism?

src/rail_map.h
```diff
@@ -530,7 +530,25 @@ static inline void MakeRailNormal(Tile t, Owner o, TrackBits b, RailType r)
 	t.m8() = r;
 }
 
+/**
+ * Sets the exit direction of a rail depot.
+ * @param t     Tile of the depot.
+ * @param dir   Direction of the depot exit.*
```
2TallTyler:
```suggestion
 * @param dir   Direction of the depot exit.
```

src/road_cmd.cpp
2TallTyler: Not that I know of, but there's no harm updating it. I don't have a preference either way.

src/road_cmd.cpp
```diff
@@ -1162,24 +1162,40 @@ CommandCost CmdBuildRoadDepot(DoCommandFlag flags, TileIndex tile, RoadType rt,
...
 
-	cost.AddCost(Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile));
-	if (cost.Failed()) return cost;
+	/* Allow the user to rotate the depot instead of having to destroy it and build it again */
+	bool rotate_existing_depot = false;
+	if (IsRoadDepotTile(tile) && (HasRoadTypeTram(tile) ? rt == RoadType::ROADTYPE_TRAM : rt == RoadType::ROADTYPE_ROAD)) {
+		return_on_cmd_failure(CheckTileOwnership(tile));
+		if (dir == GetRoadDepotDirection(tile)) return_cmd_error(STR_ERROR_ALREADY_BUILT);
+		rotate_existing_depot = true;
+		return_on_cmd_failure(EnsureNoVehicleOnGround(tile));
```
2TallTyler: To match CmdBuildTrainDepot():

```suggestion
		return_on_cmd_failure(EnsureNoVehicleOnGround(tile));
		rotate_existing_depot = true;
```

2TallTyler (commented):
Just a few nitpicks, otherwise looks good to me.

src/road_cmd.cpp
Kuhnovic: Nitpicks addressed as suggested ;)

src/rail_map.h
```diff
@@ -530,7 +530,25 @@ static inline void MakeRailNormal(Tile t, Owner o, TrackBits b, RailType r)
 	t.m8() = r;
 }
 
+/**
+ * Sets the exit direction of a rail depot.
+ * @param t     Tile of the depot.
+ * @param dir   Direction of the depot exit.
```
2TallTyler: Param name doesn't match implementation. `dir` is probably better than just `d`.

src/rail_map.h
```diff
@@ -530,7 +530,25 @@ static inline void MakeRailNormal(Tile t, Owner o, TrackBits b, RailType r)
...
+	t.m5() = RAIL_TILE_DEPOT << 6 | d;
+}
 
+/**
+ * Make a rail depot.
+ * @param t     Tile to make a depot on.
+ * @param owner New owner of the depot.
+ * @param did   New depot ID.
+ * @param dir   Direction of the depot exit.
+ * @param rt    Road type of the depot.
```
2TallTyler: `owner`, `dir`, and `rt` don't match implementation. As in my comment above, the longer versions are nicer. 🙂

2TallTyler (commented):
Sorry, found a few things I missed the first time. 🙂

src/rail_map.h
Kuhnovic: Changed all parameter names to their longer versions :)

src/command_func.h
```diff
@@ -37,6 +37,13 @@ static const CommandCost CMD_ERROR = CommandCost(INVALID_STRING_ID);
  */
 #define return_cmd_error(errcode) return CommandCost(errcode);
 
+/*
+ * Returns from a function if the provided command result is a failure.
+ *
+ * @param command_cost The result of the command to check
+ */
+#define return_on_cmd_failure(command_result) if (command_result.Failed()) return command_result;
```
LordAro: This feels at best suboptimal, at worst possibly using a temporary incorrectly. Currently it would expand to (e.g.)

```
if (CheckTileOwnership(tile).Failed()) return CheckTileOwnership(tile);
```

It's only used a few times, I'd just expand them all out (reusing the existing `cost` variable should be fine?)

PeterN: And just in favour of not adding macros generally, tbh.

Kuhnovic: I just wanted to create a convenience macro to make this a little easier without having to deal with naming a local variable, potentially running the risk over overwriting something. But I agree that my implementation was a little naive. I came up with this alternative:

`#define return_on_cmd_failure(command) { const auto return_on_cmd_failure_result = command; if (return_on_cmd_failure_result.Failed()) return return_on_cmd_failure_result; }`

But maybe I'm just overcomplicating things. You guys make the call, I'm fine with both fixing the macro or removing it altogether. I just thought it was a nice way to make things a little cleaner.

Kuhnovic: Slept on it a little and I decided to remove it. Too little gain. I've just written them out like you see elsewhere in the code.

2TallTyler (dismissed):

src/rail_map.h
```diff
@@ -530,19 +530,37 @@ static inline void MakeRailNormal(Tile t, Owner o, TrackBits b, RailType r)
...
 
+/**
+ * Sets the exit direction of a rail depot.
+ * @param tile Tile of the depot.
+ * @param dir  Direction of the depot exit.
+ */
+static inline void SetRailDepotExitDirection(Tile tile, DiagDirection dir)
+{
+	assert(IsRailDepotTile(tile));
+	tile.m5() = RAIL_TILE_DEPOT << 6 | dir;
```
rubidium42: Technically something like `SB(tile.m5(), 0, 2, dir);` should be used here.
According to the "landscape" specification the `RAIL_TILE_DEPOT` is in the highest two bits, whereas the direction is in the lowest two bits. Practically this means that someone can just start using the four middle bits for something else, and you would be resetting them here. That this is allowed for something like `MakeRailDepot` is due to the fact that you're basically building a completely new object and any new uses of those bits will undoubtedly change the calling signature of `MakeRailDepot`, or they are fine with it being set to 0.

In this case you are now resetting the "path reservation", but it is unclear whether that was intended or an oversight and it happily works. I think it is an oversight, as there is no explicit clearing of the reservation like what happens when removing a single rail tile: https://github.com/OpenTTD/OpenTTD/blob/master/src/rail_cmd.cpp#L656

Kuhnovic: Good point, definitely an oversight. Changed it to SB(...) for both rail and road.

2TallTyler:
Code looks good and my tests all work.

Should the player have to pay for rotating an existing depot? We seem to have set some sort of precedent in #9852 to not pay for these sorts of things. I don't have a strong opinion either way, but I do like being consistent. 🙂

Kuhnovic:
> Code looks good and my tests all work.
> 
> Should the player have to pay for rotating an existing depot? We seem to have set some sort of precedent in #9852 to not pay for these sorts of things. I don't have a strong opinion either way, but I do like being consistent. 🙂

I had a look at #9852 and I actually think it's pretty consistent the way it is now. For railway stations you only pay when the orientation is changed. The same is happens for rail/road depots: you get an STR_ERROR_ALREADY_BUILT error the depot is already in the current orientation, and you don't pay anything.

rubidium42 (approved):

2TallTyler:
Heh, I misread the other PR. Thanks.

PR #9645, opened 2021-10-22
Pause panning in intro game while user is interacting with the GUI

frosch123:
## Motivation / Problem

When interacting with the GUI and settings windows, the moving background of the intro game is disturbing.

## Description

The intro game animation checks the mouse position for activity.
* Panning/vehicle-following is only happening while the mouse is idle.
* When the mouse does not move for 2 seconds, panning is reactivated.
* The animation continues normally, even while panning is suppressed:
    * Non-panning actions are executed normally.
    * When the mouse becomes idle, the animation catches-up to panning-actions/vehicle-following in progress.

## Limitations

<s>This PR is in "draft" state, because I included the 12.0 titlegame for the "preview" build.</s>

## Checklist for review

### DIFF
--- a/src/intro_gui.cpp
+++ b/src/intro_gui.cpp
@@ -102,6 +102,8 @@ struct SelectGameWindow : public Window {
 	size_t cur_viewport_command_index;
 	/** Time spent (milliseconds) on current viewport command. */
 	uint cur_viewport_command_time;
+	uint mouse_idle_time;
+	Point mouse_idle_pos;
 
 	/**
 	 * Find and parse all viewport command signs.
@@ -181,6 +183,8 @@ struct SelectGameWindow : public Window {
 
 		this->cur_viewport_command_index = (size_t)-1;
 		this->cur_viewport_command_time = 0;
+		this->mouse_idle_time = 0;
+		this->mouse_idle_pos = _cursor.pos;
 	}
 
 	void OnRealtimeTick(uint delta_ms) override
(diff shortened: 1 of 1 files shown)


src/intro_gui.cpp
```diff
@@ -189,12 +193,26 @@ struct SelectGameWindow : public Window {
...
+			suppress_panning = false;
+		}
+
 		/* Determine whether to move to the next command or stay at current. */
 		bool changed_command = false;
 		if (this->cur_viewport_command_index >= intro_viewport_commands.size()) {
 			/* Reached last, rotate back to start of the list. */
 			this->cur_viewport_command_index = 0;
 			changed_command = true;
+		} else if (suppress_panning && intro_viewport_commands[this->cur_viewport_command_index].pan_to_next) {
```
nielsmh: Should probably also suppress "follow vehicle" commands.

frosch123: I changed the behavior from "continue panning in previous position" to "catch up on continue".
The former would not work with vehicle-following.

TrueBrain:
Looks good! Bit choppy when the animation continues again, but little we can do about it :D Let's ship it!

TrueBrain (approved):

LC-Zorg:
Last preview shows that you want to replace the current opntitle with the default one used in the JGR version. This is the intention? :) If not and it is supposed to work like in the previous preview, this change doesn't fix anything and it gets even worse. Much worse.

The constant changes of the background, especially in such an erratic way, only irritate and distract when the player needs to change something in the settings or worse read and understand. The necessity of constantly moving the mouse to prevent this from happening is even more irritating, especially since it doesn't help much.

I would suggest editing the current save and removing the panning signs, as well as extending the spacing between view changes.
Another, in my opinion, interesting option is the solution proposed by Kamnet, i.e. creating an intro that would use the same save with the game logo visible, and after clicking, show the menu and stop the view changes: in the current location or in the first one.

PR #9652, opened 2021-10-24
Fix #8797: Use logical rail length when placing signals

### LINKED ISSUE #8797: Wrong distance between signals with «Drag signal distance»
## Version of OpenTTD
Any with drag signal option

## Expected result
Signal every N or N-0.5 tile
Tested by drag signal distance = 1, 2 and 3, but works for any
Expected gaps marked with green and yellow rails
1 manual
2 manual
3 manual

## Actual result
Signal each N tile on straight track and each N+0.5 if gap contains diagonal tiles
Red tracks – gap over expections.
1 auto
2 auto
3 auto

## Steps to reproduce
Build some random connected tracks without juctions, get signals menu, change «Drag signal distance» in signals menu (let it be N) and build signals with ctrl+drag. Then count pieces of tracks between each signal. Straight piece of track = 1, diagonal (track on corner of tile) = 0.5. You must get N-1 on straight tracks and N-0.5 if gap contains diagonal pieces.

For common players it's not a problem, but when network contains a lot of trains with tight flow (openttdcoop style) any unexpected gap can be bottleneck of all network. For now the only solution is manually build every signal on diagonal track.
_Multicolored tracks is PURR NewGRF, it has absolutely no effect on the placement of signals_

ldpl:
## Motivation / Problem

As described in #8797 signal placer uses some visual distance instead of logical distance to place signals requiring players who need a certain logical gap to place additional signals manually.

## Description

Interprets signal distance in the UI as the logical length of that many X/Y tiles of rail. And uses the logical length of the track pieces (i.e. 192 or 128 units) when calculating the distance between signals placed.

When "keep fixed distance" setting is off (i.e. minimize_gaps = true) ensures that gap between signals never exceeds the required signal distance except for when it's on a bridge or tunnel. 
Without minimize_gaps (i.e. "pretty" mode) just tries to place signals as soon as accumulated gap >= required distance regardless of whether it succeeds or not. 
Nundinghall Transport, 15th Jan 2031

Also, this placer is mostly ready to work with segments of arbitrary logical length if that ever makes into the game.

## Limitations

* Does not account for train elongation on diagonals as there is nothing signal placer can do without knowing exact train stats.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -1209,8 +1209,11 @@ CommandCost CmdBuildSingleSignal(TileIndex tile, DoCommandFlag flags, uint32 p1,
 	return cost;
 }
 
-static bool CheckSignalAutoFill(TileIndex &tile, Trackdir &trackdir, int &signal_ctr, bool remove)
+static bool AdvanceSignalAutoFill(TileIndex &tile, Trackdir &trackdir, bool remove)
 {
+	/* We only process starting tiles of tunnels or bridges so jump to the other end before moving further. */
+	if (IsTileType(tile, MP_TUNNELBRIDGE)) tile = GetOtherTunnelBridgeEnd(tile);
+
 	tile = AddTileIndexDiffCWrap(tile, _trackdelta[trackdir]);
 	if (tile == INVALID_TILE) return false;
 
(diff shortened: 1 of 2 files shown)


Rau117:
Wow, someone remembered my old bug report, thanks!

However, will be nice to see the result on **such** a rail structure. It is quite useless, but it shows well what should be done at best (example for **«2»** distance between signals).
изображение

ldpl:
This is what it currently does with minimize_gaps and gap 2. Also there is an emscripten preview so you can try whatever you want yourself right in the browser.
Screenshot from 2021-10-25 00-37-41

LC-Zorg:
Generally it's not bad, but with the current positioning of the signaling it looks ugly. Without aligning signals, it won't be a good change for many players.

Currently (at the top), the distances between signals are 18 and 42 pixels. In addition, one of the signals is a pixel lower than the other. At the bottom, it is equal to 30 and 30.
signals on flat rails

Likewise for vertical tracks.
signals on vertical rails

I don't see why these signals should remain so unevenly placed.

Btw. I would see another method here that would help keep traffic flowing without the need of ugly signals compaction. There could be even less of this signals... In short: "Start braking when you see the red light"

ldpl:
Signal sprite offsets are a separate issue. Signal placer doesn't even build signals that close in a "pretty" mode anyway.

LC-Zorg:
Separate, but not entirely so, because this change would significantly emphasize these inequalities. For spacing settings of 3 or more, it looks fine. Setting 1 and 2 is problematic.
If these signaling positions were corrected, both modes would be pretty. ;) Besides, some players (I don't know how many, there can be a lot of them or even most, for sure me) like evenly arranged signaling, but at the same time prefer it when the game places it in front of bridges when the distance is too large. This change will remove that possibility. 

Another significant problem is the continuation of the game with the changed building rules. So maybe instead of an on/off switch, three points to choose would be better?

glx22:
If the alignment is fixed, people will complain because trains don't stop in front of signals.
Signals are placed on the tile edge for a good reason (even if it may look ugly).

LC-Zorg:
If people were to complain about it, they should be doing it now. :D

Currently
signals position - current

After change
signals position - new

In front of the station
signals position - on station

LordAro:
The real reason for the "misalignment" is so that signals do not overlap when train lines run next to each other

image

Now, that's enough off-topic discussion, please keep further discussion to the actual PR subject.

Rau117:
>  Setting 1 and 2 is problematic.

2 used by openttdcoop community. The flow of trains here is super-dense, not like in ordinary vanilla games, and it is for this density that it is very important that the signals always have the minimum gap. «1» cannot be maintained very well – you can't do simple juction on the straight track, so the most convenient one is chosen - «2».

At the current signal placement on the diagonals there are always problems, because the signal density is 1.5 times less than on straight sections, which leads to permanent traffic jams on the diagonals.

You can place signals manually on the diagonals every 2x0.5 of the tile… but this is some kind of slave labor. 

Honestly, I don’t know who in their right mind would use «1» all the time.

LordAro (approved):
#YOLO. Code looks fine. Would recommend against backporting though

PR #9653, opened 2021-10-24
Feature: add setting to hide news about competitors vehicle crash

grenoult:
## Motivation / Problem

During a game of OpenTTD with many competitors, we can receive a lot of news regarding competitors vehicles involved in accidents. This is especially true when vehicles have low reliability and are rarely replaced:
- road vehicles breaking down on a rail crossing then being crashed by a train,
- airplanes failing to land on an airport and crashing.

This PR addresses the issue of receiving too many news regarding competitors vehicle crash where it doesn't bring any value or meaning to the main player. As a OpenTTD player, I want to know when my vehicles are involved in a crash but I may choose to not receive notification if competitors vehicles are destroyed.

## Description

A new setting has been added "Accidents of competitor's vehicles". It's "Full" by default. When turned off, no news will be generated when a competitor vehicle is involved in a crash. When it's "Summary", only the short version will be displayed at the bottom of the screen.
A crash, in the context of this PR, is when a road vehicle is hit by a train or when an airplane misses landing at an airport.

- Full: A news item is generated when setting is Full (by default) and a competitor vehicle is crashed (in this example, a plane crash):
image

- Summary: Short news are generated when setting is "Summary" and a competitor vehicle (red) is crashed by a train:
image

- Off: no news are generated when a competitor vehicle is crashed when settings is "Off"
image

## Limitations

- other crashes, such as UFOs and similar events are ignored as part of this PR. This has been left out because it's rare enough to cause a disruption to the player. That being said, one could argue we could update this in this PR and I could update it (probably in another PR) if requested.

## Checklist for review

- This PR adds a new translation in english.txt file. From the [documentation](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md#i-want-to-add-a-new-string), nothing else is required at this stage.

### DIFF
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -1329,7 +1329,6 @@ static void CrashAirplane(Aircraft *v)
 	const Station *st = GetTargetAirportIfValid(v);
 	StringID newsitem;
 	TileIndex vt;
-	NewsType newstype = NT_ACCIDENT;
 	if (st == nullptr) {
 		newsitem = STR_NEWS_PLANE_CRASH_OUT_OF_FUEL;
 		vt = TileVirtXY(v->x_pos, v->y_pos);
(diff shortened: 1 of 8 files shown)


src/aircraft_cmd.cpp
```diff
@@ -1329,6 +1329,7 @@ static void CrashAirplane(Aircraft *v)
 	const Station *st = GetTargetAirportIfValid(v);
 	StringID newsitem;
 	TileIndex vt;
+	NewsType newstype = NT_ACCIDENT;
```
LordAro: Move this variable declaration slightly closer to where it is used?

grenoult: @LordAro updated, thanks

LordAro (approved):

Rau117:
"We don't want to add unnecessary settings"
Adding unnecessary settings.

TrueBrain:
> 
> 
> "We don't want to add unnecessary settings" Adding unnecessary settings.

Just to be absolutely crystal clear: do understand that if you stay on this destructive path, you have no business being part of this community. This is your last warning about your behaviour. Please learn to be constructive; once again: this is not that.

PR #9673, opened 2021-11-05
Fix #9579: Object and HQ construction is Construction cost, not Property Ma…

### LINKED ISSUE #9579: Improperly assigned objects construction and HQ relocation costs
### Version of OpenTTD

1.11.2 (probably it has always been)

### Expected result

The costs of building objects and moving the headquarters are included in the "Construction" or "Other" sections, which do not affect the company's operating profit.

### Actual result

The cost of building objects and relocating the company's seat are included in the property maintenance costs. In this way, the construction of objects and the transfer of HQ affect the company's operating profit. This impact can be very large when the objects are expensive or company value is high (HQ transfer cost is 1%). This often results in a strange drop in income for players visible on the graph, which actually did not happen.

The situation from one of the online games
Zrzut ekranu 2021-09-26 170659

### Steps to reproduce

1. Run any save, the longer the game, the better (attached one of the online game saves with very long gameplay, use company 8)
[WARP Technologies S.A., 12503-10-25.zip](https://github.com/OpenTTD/OpenTTD/files/7231740/WARP.Technologies.S.A.12503-10-25.zip)

2. Move the company HQ
3. Change your mind and move HQ again (for better effect;)
4. Look at the operating profit graph - in the case of a long game, you will only see the loss in one quarter in which you transferred your HQ
5. For the next 6 years of the game, the profit graph will be disturbed
6. The construction of objects can also greatly disturb the graph, but here the younger the game and the more expensive the objects, the clearer the effect.

2TallTyler:
…intenance

## Motivation / Problem

Per #9579:

> ### Expected result
> The costs of building objects and moving the headquarters are included in the "Construction" or "Other" sections, which do not affect the company's operating profit.
> 
> ### Actual result
> The cost of building objects and relocating the company's seat are included in the property maintenance costs.

The cost of clearing objects is already counted under Construction.

## Description

Building objects and moving the company HQ is now considered under Construction, not Property Maintenance.

Closes #9579

## Checklist for review

### DIFF
--- a/src/object_cmd.cpp
+++ b/src/object_cmd.cpp
@@ -204,7 +204,7 @@ static CommandCost ClearTile_Object(TileIndex tile, DoCommandFlag flags);
  */
 CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
 {
-	CommandCost cost(EXPENSES_PROPERTY);
+	CommandCost cost(EXPENSES_CONSTRUCTION);
 
 	ObjectType type = (ObjectType)GB(p1, 0, 16);
 	if (type >= NUM_OBJECTS) return CMD_ERROR;
@@ -526,7 +526,7 @@ static CommandCost ClearTile_Object(TileIndex tile, DoCommandFlag flags)
 			}
 
 			/* cost of relocating company is 1% of company value */
-			cost = CommandCost(EXPENSES_PROPERTY, CalculateCompanyValue(c) / 100);
+			cost = CommandCost(EXPENSES_CONSTRUCTION, CalculateCompanyValue(c) / 100);
 			break;
 		}
 
--- a/regression/regression/result.txt
+++ b/regression/regression/result.txt
@@ -639,7 +639,7 @@ ERROR: IsEnd() is invalid as Begin() is never called
     GetQuarterlyCompanyValue():      -1
   Quarter: 0
     GetQuarterlyIncome():            0
-    GetQuarterlyExpenses():          -210
+    GetQuarterlyExpenses():          0
     GetQuarterlyCargoDelivered():    0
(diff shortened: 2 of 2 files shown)


LordAro (dismissed):

2TallTyler:
Hmm, this is failing the regression test. Any idea how I'd fix that?

James103:
The failure occurs in GetQuarterlyExpenses(), which takes into account property maintenance but not construction costs, both of which this PR touches. Therefore, the regressions will need to be recorded again.

glx22:
If you run regression locally it should keep the output.

LordAro:
I know it's documented as such, but GetQuarterlyExpenses not including construction costs seems unexpected to me.

Also, I question the use of that regression test, now that it's (almost) all 0s for all 24 quarters...

2TallTyler:
I don't disagree, but I think improving regression tests would be out of scope for this PR, no?

src/object_cmd.cpp
```diff
@@ -204,7 +204,7 @@ static CommandCost ClearTile_Object(TileIndex tile, DoCommandFlag flags);
  */
 CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
 {
-	CommandCost cost(EXPENSES_PROPERTY);
+	CommandCost cost(EXPENSES_CONSTRUCTION);
```
frosch123, line 207: This also puts "purchase land" under "construction". Intended?

2TallTyler: Not intended because I forgot that's an object, but I would also consider this a bug which is fixed by this PR. Currently purchasing land falls under Property Maintenance, then selling that land counts as income under Construction. Changing it to all be Construction makes better sense to me.

LordAro (approved):

PR #9674, opened 2021-11-06
Fix: Don't show screenshot GUI in screenshots

2TallTyler:
## Motivation / Problem

Screenshots captured using the "Normal screenshot" button show the screenshot GUI. The other screenshot modes do not.

Unnamed, 1920-11-20

I often see new players posting screenshots using this button on Discord, Reddit, etc., which is typically met with suggestions to use another screen capture tool which doesn't show the screenshot GUI.

I can imagine a player wanting to show other GUI elements like vehicle windows, the status bar ("look at how much money I have!"), and even the main toolbar at top. But I don't see a reason to show the screenshot GUI itself!

## Description

Upstreams [a JGRPP commit](https://github.com/JGRennison/OpenTTD-patches/commit/a65be9b7471753603ecc37f3234125fc06c40853) to hide the screenshot GUI when taking a screenshot.

## Checklist for review

### DIFF
--- a/src/screenshot_gui.cpp
+++ b/src/screenshot_gui.cpp
@@ -13,6 +13,7 @@
 #include "screenshot.h"
 #include "widgets/screenshot_widget.h"
 #include "table/strings.h"
+#include "gfx_func.h"
 
 struct ScreenshotWindow : Window {
 	ScreenshotWindow(WindowDesc *desc) : Window(desc)
@@ -71,4 +72,25 @@ void ShowScreenshotWindow()
 {
 	CloseWindowById(WC_SCREENSHOT, 0);
 	new ScreenshotWindow(&_screenshot_window_desc);
+}
+
+/**
+ * Set the visibility of the screenshot window when taking a screenshot.
+ * @param hide Are we hiding the window or showing it again after the screenshot is taken?
+ */
+void SetScreenshotWindowVisibility(bool hide)
+{
+	ScreenshotWindow *scw = (ScreenshotWindow *)FindWindowById(WC_SCREENSHOT, 0);
+
+	if (scw == nullptr) return;
+
+	if (hide) {
+		/* Set dirty the screen area where the window is covering (not the window itself), then move window off screen. */
+		scw->SetDirty();
+		scw->left += 2 * _screen.width;
+	} else {
+		/* Return window to original position. */
+		scw->left -= 2 * _screen.width;
+		scw->SetDirty();
+	}
 }
--- a/src/screenshot_gui.h
+++ b/src/screenshot_gui.h
@@ -11,5 +11,6 @@
 #define SCREENSHOT_GUI_H
 
 void ShowScreenshotWindow();
+void SetScreenshotWindowVisibility(bool hide);
 
 #endif /* SCREENSHOT_GUI_H */
--- a/src/screenshot.cpp
+++ b/src/screenshot.cpp
@@ -12,6 +12,7 @@
 #include "viewport_func.h"
 #include "gfx_func.h"
 #include "screenshot.h"
+#include "screenshot_gui.h"
 #include "blitter/factory.hpp"
 #include "zoom_func.h"
 #include "core/endian_func.hpp"
@@ -909,8 +910,10 @@ static bool RealMakeScreenshot(ScreenshotType t, std::string name, uint32 width,
 		 * of the screenshot. This way the screenshot will always show the name
 		 * of the previous screenshot in the 'successful' message instead of the
 		 * name of the new screenshot (or an empty name). */
+		SetScreenshotWindowVisibility(true);
 		UndrawMouseCursor();
 		DrawDirtyBlocks();
+		SetScreenshotWindowVisibility(false);
 	}
 
 	_screenshot_name[0] = '\0';


src/screenshot_gui.cpp
```diff
@@ -72,3 +72,18 @@ void ShowScreenshotWindow()
 	CloseWindowById(WC_SCREENSHOT, 0);
 	new ScreenshotWindow(&_screenshot_window_desc);
 }
+
+void SetScreenshotWindowHidden(bool hidden)
+{
+	ScreenshotWindow* scw = (ScreenshotWindow*)FindWindowById(WC_SCREENSHOT, 0);
```
LordAro:
```suggestion
	ScreenshotWindow *scw = (ScreenshotWindow *)FindWindowById(WC_SCREENSHOT, 0);
```

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,18 @@ void ShowScreenshotWindow()
...
+	if (scw != nullptr) {
+		if (hidden) {
+			scw->SetDirty();
+			SetBit(scw->left, 30);
+		}
+		else {
```
LordAro:
```suggestion
		} else {
```

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,18 @@ void ShowScreenshotWindow()
...
 	new ScreenshotWindow(&_screenshot_window_desc);
 }
+
+void SetScreenshotWindowHidden(bool hidden)
+{
+	ScreenshotWindow* scw = (ScreenshotWindow*)FindWindowById(WC_SCREENSHOT, 0);
+	if (scw != nullptr) {
+		if (hidden) {
+			scw->SetDirty();
+			SetBit(scw->left, 30);
```
LordAro: What is 30 here? Do we have a define/constant for it that can be used instead?

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,18 @@ void ShowScreenshotWindow()
 	CloseWindowById(WC_SCREENSHOT, 0);
 	new ScreenshotWindow(&_screenshot_window_desc);
 }
+
+void SetScreenshotWindowHidden(bool hidden)
```
LordAro: doxygen doc comments, ideally

LordAro (dismissed):
Can't blame JGR for the codestyle issues, you did those all by yourself ;) (In theory you should be able to add JGRPP repo as a remote and just cherry-pick it)

src/screenshot_gui.cpp
glx22: 30 is to make `left` very big so the window is moved out of the main screen.
I think it doesn't need to be a define/constant, but at least a comment to explain why 30.

frosch123: Maybe it should be something like `scw->left = 2 * _screen.width`. Setting bit 30 is pretty arbitrary.

2TallTyler: I suppose then the issue becomes remembering the previous location of the window, when it's time to un-hide it. I've added a comment to explain it for now; let me know what you think.

2TallTyler:
Someday I'll learn not to stay up past my bedtime to submit PRs and miss silly things. 🙄

I prefer not to cherry-pick because going through the motions of adding the code myself helps me understand it and (when I'm paying attention) catch missing code style, Doxygen, or references to other code not present in vanilla.

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,21 @@ void ShowScreenshotWindow()
...
 }
+
+/**
+ * Hides or shows the screenshot window when taking a screenshot.
+ * @param hidden Are we hiding the window or showing it again after the screenshot is taken?
+ */
+void SetScreenshotWindowHidden(bool hidden)
+{
+	ScreenshotWindow *scw = (ScreenshotWindow *)FindWindowById(WC_SCREENSHOT, 0);
+	if (scw != nullptr) {
```
TrueBrain: `if (scw == nullptr) return` makes for much better readable code. In general, if you have an if-case that spans till the end of your function, it is more readable to invert the statement and return early :)

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,21 @@ void ShowScreenshotWindow()
 	CloseWindowById(WC_SCREENSHOT, 0);
 	new ScreenshotWindow(&_screenshot_window_desc);
 }
+
+/**
+ * Hides or shows the screenshot window when taking a screenshot.
+ * @param hidden Are we hiding the window or showing it again after the screenshot is taken?
```
TrueBrain: `hidden` confuses me. Does it mean the window was hidden, but you are unhiding it, or does it mean it will become hidden? In general, you want to use active forms for function names and parameters there in. So `hide` would be much more clear.

Similar, the function name is misleading. You are not setting it "hidden". You set the visibility of the window. So `SetScreenshotWindowVisibility` would be a lot more verbose.

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,21 @@ void ShowScreenshotWindow()
...
+ */
+void SetScreenshotWindowHidden(bool hidden)
+{
+	ScreenshotWindow *scw = (ScreenshotWindow *)FindWindowById(WC_SCREENSHOT, 0);
+	if (scw != nullptr) {
+		if (hidden) {
+			scw->SetDirty();
```
TrueBrain: The content of the if/else is a bit weird. In the first body you do `SetDirty` -> `Change left`, in the second you do `Change left` -> `SetDirty`. From a reader perspective, this suggests the order matter or something is going on. Where in reality the order really doesn't matter.

I would suggest to move the `SetDirty` outside of the `if`. Normally you do it after changing `left` , as that just reads better: change the position, mark it as dirty. The reverse: mark it as dirty, change position, feels .. not natural. But code-wise the order doesn't matter.

src/screenshot_gui.cpp
```diff
@@ -72,3 +72,21 @@ void ShowScreenshotWindow()
...
+			SetBit(scw->left, 30); // Move window out of main screen
+		} else {
+			ClrBit(scw->left, 30); // Return window to original position
```
TrueBrain: Although I get it is easy and cheap to use setting a bit, it is confusing to the reader. Why bit 30? Why not 31? Why not 29? So many questions :)

It might be better to make a const with some large number, and add/subtract that value. It still is a hack, but I understand why you do it.

There is another approach to take, but I am not sure it is better/worse. As hide/show is always called in a pair, you could use a `static` value to safely store the current `scw->left`, and set the `scw->left` to a known off-screen value when hiding. That in effect gives the same result.

Basically, now a bit in `left` is used to indicate it is hidden or not, which is just a bit nasty. And might hurt us in future refactors. Personally, I would have gone for the `static` method, but I am not sure what other developers think about it. For sure the bitset is just asking for trouble, so a const to add/subtract to `scw->left` sounds more sane to me :)

TrueBrain (commented):
This is absolute nitpicking. But as you indicate you would like to learn from these PRs, here is some feedback. Despite this PR being a few lines, there are just some issues with the logic used :)

src/screenshot_gui.cpp
JGRennison: The order does matter, it's not accidental.

TrueBrain: huh? That .. needs a comment in the code for sure ;)

src/screenshot_gui.cpp
TrueBrain: Oops, missed that @frosch123 also commented on this (but GitHub was hiding that comment from me). His solution would also work fine, as between the hide and show commands, the width of the screen cannot change. So that is also a safe thing to use in this case.

src/screenshot_gui.cpp
JGRennison: In trunk, Window::SetDirty marks the area of the screen which the window is over as dirty. If the window is off-screen, then Window::SetDirty will have no effect.

2TallTyler:
Thank you for the educational nitpicking! 😄 

Let me know what you think of these changes.

src/screenshot_gui.h
```diff
@@ -11,5 +11,6 @@
 #define SCREENSHOT_GUI_H
 
 void ShowScreenshotWindow();
+void SetScreenshotWindowVisibility(bool hidden);
```
TrueBrain: very very very minor: `hide` :)

TrueBrain (dismissed):

src/screenshot_gui.cpp
```diff
@@ -14,6 +14,8 @@
 #include "widgets/screenshot_widget.h"
 #include "table/strings.h"
 
+extern DrawPixelInfo _screen;
```
michicc: Were you searching the `gfx_func.h` header?

2TallTyler: No, I needed `_screen.width` which I believe is defined in `gfx.cpp`, however if I #include that lots of other things break. Is there a better way to do this?

michicc: Line 64 of gfx_func.h is `extern DrawPixelInfo _screen;`

2TallTyler: Would it be better to `#include gfx_func.h`?

michicc: IMHO yes, especially in a cpp file. There might be a reason to make an exception inside a header file if it would pull in too much other stuff, but that is not a concern for cpp files.

2TallTyler: Done, thanks.

michicc (approved):

PR #9093, opened 2021-04-24
Feature: [NewGRF] vehicle var 0x4E, 'date of last station departure'

andythenorth:
## Motivation / Problem

Extends NewGRF API with vehicle var 0x4E  'Last Station Departure Date'.
Using a "date" value makes this variable similar to use as "date of last service".
The use-cases are also similar (stuff that is refreshed/refilled/refurbished in stations, instead of in depots).

## Description

* vehicle property for long date (integer days since 1-1-0) of last station departure
* property will be DATE_INVALID until at least one station has been departed from
* date is set by LeaveStation()
* newgrf var 0x4E is provided to read the property
* var 0x4E provided in newgrf vehicle debug window
* exceeding max year handled in same ways as "date of last service"

## Limitations

* for trains or articulated RVs, the property is only updated for lead vehicle
  * seems like TMWFTLB to recursively set an identical value on every vehicle in the consist)
  * frosch indicates this is fine
  * newgrf can access the var with PARENT scope, no problems using it
* nml patch: https://github.com/OpenTTD/nml/pull/211
* needs newgrf documentation before PR is ready
* tested with a grf, works as expected

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #9693, opened 2021-11-13
Fix travel times stored in the linkgraph 

### LINKED ISSUE #9665: [Bug]: Travel times stored in linkgraph are incorrect for all non-direct journey types
### Version of OpenTTD

Since #9457

### Expected result

Stored link graph travel times are correct

### Actual result

Link graph stored travel times are incorrect in the case where one or more of the following order types is present between the source and destination station orders:

* Go via waypoint
* Go via station (i.e. not stopping)
* Go to station with both no loading and no unloading
* Depot/hangar order (servicing and taken, or unconditional go to depot/hangar)

`current_order_time` is the time taken by the current order, not the time taken since cargo was last loaded/unloaded and does not correspond with `last_loading_station`.

### Steps to reproduce

Create an order list using one of the above intermediary order types.

See attached for an example: [savegame](https://github.com/OpenTTD/OpenTTD/files/7448559/Linkgraph.travel.time.test.1.1990-08-16.sav.zip)

nchappe:
## Motivation / Problem

This PR fixes a few problems with travel times in the linkgraph (#9457).
- The first commit is a tiny code cleanup.
- The second commit fixes a small loss of precision that can happen (rarely) on linkgraph compression for low-capacity links.
- The third commit fixes #9665.

## Description

To fix #9665, I added a member variable `BaseConsist::time_since_last_loading` that is incremented whenever `current_order_time` is, and that is reset when the vehicle loads cargo at a station, just like `last_loading_station`.

This fully fixes the travel times in the example savegame of #9665. I also tested with a go to depot order and with a basic conditional order.

## Limitations

This PR needs a savegame upgrade.

## Checklist for review

### DIFF
--- a/src/linkgraph/refresh.cpp
+++ b/src/linkgraph/refresh.cpp
@@ -210,6 +210,8 @@ void LinkRefresher::RefreshStats(const Order *cur, const Order *next)
 			uint cargo_quantity = this->capacities[c];
 			if (cargo_quantity == 0) continue;
 
+			if (this->vehicle->GetDisplayMaxSpeed() == 0) continue;
+
 			/* If not allowed to merge link graphs, make sure the stations are
 			 * already in the same link graph. */
 			if (!this->allow_merge && st->goods[c].link_graph != st_to->goods[c].link_graph) {
--- a/src/linkgraph/refresh.cpp
+++ b/src/linkgraph/refresh.cpp
@@ -203,6 +203,7 @@ void LinkRefresher::RefreshStats(const Order *cur, const Order *next)
 	StationID next_station = next->GetDestination();
 	Station *st = Station::GetIfValid(cur->GetDestination());
 	if (st != nullptr && next_station != INVALID_STATION && next_station != st->index) {
+		Station *st_to = Station::Get(next_station);
 		for (CargoID c = 0; c < NUM_CARGO; c++) {
 			/* Refresh the link and give it a minimum capacity. */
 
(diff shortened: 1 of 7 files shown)


src/linkgraph/refresh.cpp
```diff
@@ -203,6 +203,7 @@ void LinkRefresher::RefreshStats(const Order *cur, const Order *next)
 	StationID next_station = next->GetDestination();
 	Station *st = Station::GetIfValid(cur->GetDestination());
 	if (st != nullptr && next_station != INVALID_STATION && next_station != st->index) {
+		Station *st_to = Station::Get(next_station);
```
frosch123, line 206: Can `st_to` ever be a nullptr?
Line 215 accesses it directly, whie line 225 checks the pointer first.
Possibly also remove the check in line 225?

frosch123:
The counter `time_since_last_loading` added here seems to be similar to `last_station_departure_date` from #9093.
* In #9093 an absolute time was considered better than a relative time.
* #9093 seems to care only about the date, while this PR cares about ticks.
* #9093 is pretty unclear about when to set the counter "loading", "unloading", "stopping"...

Is it the same? or is there an important difference that makes it necessary to add two similar but distinct time stamps?

nchappe:
I thought about using a date, the code would be a bit simpler but it would be a bit less precise. More importantly, it would break when using the cheat that changes the current year. While working on this PR I found out that setting the current year does break `date_of_last_service` a bit in master, I will probably open a bug or a PR about that later.

The time I store does not exactly correspond to the date of last station departure because to fix #9665 the variable introduced by my PR has to be exactly the time since `last_loading_station` was last updated, which is very specific  (see `Vehicle::LeaveStation()` for the exact conditions). So it makes sense to have three similar but separate variables `current_order_time`, `time_since_last_loading` and `last_station_departure_date`.

glx22:
Isn't it possible to just store current vehicle tick and do some math when the exact value is needed ? And remove most counters.

nchappe:
Without the time since last loading, you would probably need to store the date of last loading, which has its advantages (no need to increment the counters) and drawbacks (lower resolution, problems when using cheats and when reaching last valid year).

I grep'd for `current_order_time` to determine where I should increment `time_since_last_loading`, so assuming the former works fine, hopefully the latter will too.

nchappe:
I looked a bit more into your suggestion and I think `_date * DAY_TICKS + _date_frac` as an int64 would probably work as a tick counter. Should I replace the counter introduced in this PR with a timestamp in ticks and deduce the travel time from it?

nchappe:
As suggested, I replaced the tick counter with a date in the fix to #9665. I also added a trivial fix to #9708.

nchappe:
Could someone please have another look at this PR?

nchappe:
Ping. Cargodist is severely broken for some networks since OpenTTD 12.0 (#9665, #9708), and this PR aims to fix it.
I can always split it into several PRs if you are not convinced by some of the commits.

LordAro (dismissed):
Seems reasonable to me, but I'd like another pair of eyes on it

(Please disregard that I screwed up the rebase, we can fix that on merge... or you can fix it :) )

nchappe:
I rebased onto master. #10035 makes the third commit simpler, I now use _tick_counter instead of _date_frac to compute durations in ticks.

michicc (approved):

PR #9704, opened 2021-11-17
Fix #9702: Revert old behaviour in group GUI when not using shared orders

### LINKED ISSUE #9702: [Bug]: It is not possible to transfer vehicles sharing orders between groups (using Ctrl+drag doesn't work)
### Version of OpenTTD

since 12.0, Win10

### Expected result

You can use Ctrl+  drag to move groups of vehicles sharing orders between groups - as it was in 1.11.2 and earlier.

### Actual result

Using Ctrl+ has no effect and it is not possible to move whole groups of vehicles between created groups in the vehicle list window.
To put 200 vehicles into the newly created group, you must move each vehicle separately.

### Steps to reproduce

1. Create a simple but long road connection between the two cities
2. Buy dozens of vehicles to handle this connection
3. Create a group in the vehicle list
4. Try to put these several dozen vehicles in a new group

btzy:
## Motivation / Problem

This PR is a fix for #9702.

## Description

The group GUI now only opens the shared order list if the dropdown is set to shared orders AND the Ctrl key is pressed.  In other situations, we use the original behaviour.

When grouped by shared orders, the Ctrl key does not need to be pressed when doing a drag-and-drop operation - all vehicles that share those orders will be moved to the new group, regardless of the state of the Ctrl key when releasing the mouse.  This has already been the behaviour before the bug was introduced.

## Checklist for review

### DIFF
--- a/src/group_gui.cpp
+++ b/src/group_gui.cpp
@@ -746,17 +746,23 @@ public:
 						NOT_REACHED();
 				}
 				if (v) {
-					if (_ctrl_pressed) {
-						if (this->grouping == GB_NONE) {
-							this->SelectGroup(v->group_id);
-						} else {
-							ShowOrdersWindow(v);
-						}
+					if (_ctrl_pressed && this->grouping == GB_SHARED_ORDERS) {
+						ShowOrdersWindow(v);
 					} else {
 						this->vehicle_sel = v->index;
+
+						if (_ctrl_pressed && this->grouping == GB_NONE) {
+							/*
+							 * It only makes sense to select a group if not using shared orders
+							 * since two vehicles sharing orders can be from different groups.
+							 */
+							this->SelectGroup(v->group_id);
+						}
+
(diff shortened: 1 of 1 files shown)


src/group_gui.cpp
```diff
@@ -746,17 +746,21 @@ class VehicleGroupWindow : public BaseVehicleListWindow {
...
-							ShowOrdersWindow(v);
-						}
+					if (_ctrl_pressed && this->grouping == GB_SHARED_ORDERS) {
+						ShowOrdersWindow(v);
 					} else {
 						this->vehicle_sel = v->index;
+
+						if (_ctrl_pressed && this->grouping == GB_NONE) {
+							// It only makes sense to select a group if not using shared orders
+							// since two vehicles sharing orders can be from different groups.
```
LordAro: nitpick: Non-inline comments should use `/* */`

btzy: fixed

LordAro (approved):
Thanks for doing that so quickly :)

PR #9709, opened 2021-11-19
Feature: Build objects by area

2TallTyler:
## Motivation / Problem

Building decorative objects one tile at a time is tedious.

JGRPP includes a feature to build objects by drag-and-drop, including diagonally by holding Ctrl.

## Description

Upstreams [JGR's commit](https://github.com/JGRennison/OpenTTD-patches/commit/f44d75eaf5cad16aa24617482a3ef3623036feb7), with modifications for vanilla.

Also includes a commit to [limit the rate of object construction](https://github.com/JGRennison/OpenTTD-patches/commit/2d350d26db8efe6a4b624a813237febf2d854443), to prevent griefing.

## Limitations

I haven't touched saveload upgrades in a while and JGRPP handles it differently, so please check this carefully!

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -63,9 +63,10 @@ Company::Company(uint16 name_1, bool is_ai)
 	this->name_1 = name_1;
 	this->location_of_HQ = INVALID_TILE;
 	this->is_ai = is_ai;
-	this->terraform_limit = (uint32)_settings_game.construction.terraform_frame_burst << 16;
-	this->clear_limit     = (uint32)_settings_game.construction.clear_frame_burst << 16;
-	this->tree_limit      = (uint32)_settings_game.construction.tree_frame_burst << 16;
+	this->terraform_limit    = (uint32)_settings_game.construction.terraform_frame_burst << 16;
+	this->clear_limit        = (uint32)_settings_game.construction.clear_frame_burst << 16;
+	this->tree_limit         = (uint32)_settings_game.construction.tree_frame_burst << 16;
+	this->build_object_limit = (uint32)_settings_game.construction.build_object_frame_burst << 16;
 
 	std::fill(this->share_owners.begin(), this->share_owners.end(), INVALID_OWNER);
 	InvalidateWindowData(WC_PERFORMANCE_DETAIL, 0, INVALID_COMPANY);
@@ -267,9 +268,10 @@ void SubtractMoneyFromCompanyFract(CompanyID company, const CommandCost &cst)
 void UpdateLandscapingLimits()
 {
 	for (Company *c : Company::Iterate()) {
-		c->terraform_limit = std::min<uint64>((uint64)c->terraform_limit + _settings_game.construction.terraform_per_64k_frames, (uint64)_settings_game.construction.terraform_frame_burst << 16);
-		c->clear_limit     = std::min<uint64>((uint64)c->clear_limit     + _settings_game.construction.clear_per_64k_frames,     (uint64)_settings_game.construction.clear_frame_burst << 16);
-		c->tree_limit      = std::min<uint64>((uint64)c->tree_limit      + _settings_game.construction.tree_per_64k_frames,      (uint64)_settings_game.construction.tree_frame_burst << 16);
+		c->terraform_limit    = std::min<uint64>((uint64)c->terraform_limit    + _settings_game.construction.terraform_per_64k_frames,    (uint64)_settings_game.construction.terraform_frame_burst << 16);
+		c->clear_limit        = std::min<uint64>((uint64)c->clear_limit        + _settings_game.construction.clear_per_64k_frames,        (uint64)_settings_game.construction.clear_frame_burst << 16);
+		c->tree_limit         = std::min<uint64>((uint64)c->tree_limit         + _settings_game.construction.tree_per_64k_frames,         (uint64)_settings_game.construction.tree_frame_burst << 16);
+		c->build_object_limit = std::min<uint64>((uint64)c->build_object_limit + _settings_game.construction.build_object_per_64k_frames, (uint64)_settings_game.construction.build_object_frame_burst << 16);
 	}
 }
 
--- a/src/object_cmd.cpp
+++ b/src/object_cmd.cpp
@@ -311,6 +311,7 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 	}
 
 	int hq_score = 0;
+	uint build_object_size = 1;
 	switch (type) {
 		case OBJECT_TRANSMITTER:
 		case OBJECT_LIGHTHOUSE:
@@ -349,7 +350,14 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 			return CMD_ERROR;
 
 		default: // i.e. NewGRF provided.
+			build_object_size = size_x * size_y;
 			break;
+	}
+
+	/* Don't allow building more objects if the company has reached its limit. */
+	Company *c = Company::GetIfValid(_current_company);
+	if (c != nullptr && GB(c->build_object_limit, 16, 16) < build_object_size) {
+		return_cmd_error(STR_ERROR_BUILD_OBJECT_LIMIT_REACHED);
 	}
 
 	if (flags & DC_EXEC) {
@@ -357,12 +365,70 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 
 		/* Make sure the HQ starts at the right size. */
 		if (type == OBJECT_HQ) UpdateCompanyHQ(tile, hq_score);
+
+		/* Subtract the tile from the build limit. */
+		if (c != nullptr) c->build_object_limit -= build_object_size << 16;
 	}
 
-	cost.AddCost(ObjectSpec::Get(type)->GetBuildCost() * size_x * size_y);
+	cost.AddCost(spec->GetBuildCost() * build_object_size);
 	return cost;
 }
 
+/**
+ * Construct multiple objects in an area
+ * @param flags of operation to conduct
+ * @param tile end tile of area dragging
+ * @param start_tile start tile of area dragging
+ * @param type the object type to build
+ * @param view the view for the object
+ * @param diagonal Whether to use the Orthogonal (0) or Diagonal (1) iterator.
+ * @return the cost of this operation or an error
+ */
+CommandCost CmdBuildObjectArea(DoCommandFlag flags, TileIndex tile, TileIndex start_tile, ObjectType type, uint8 view, bool diagonal)
+{
+	if (start_tile >= MapSize()) return CMD_ERROR;
+
+	if (type >= NUM_OBJECTS) return CMD_ERROR;
+	const ObjectSpec *spec = ObjectSpec::Get(type);
+	if (view >= spec->views) return CMD_ERROR;
+
+	if (spec->size != OBJECT_SIZE_1X1) return CMD_ERROR;
+
+	Money money = GetAvailableMoneyForCommand();
+	CommandCost cost(EXPENSES_CONSTRUCTION);
+	CommandCost last_error = CMD_ERROR;
+	bool had_success = false;
+
+	const Company *c = Company::GetIfValid(_current_company);
+	int limit = (c == nullptr ? INT32_MAX : GB(c->build_object_limit, 16, 16));
+
+	TileIterator *iter = diagonal ? (TileIterator *)new DiagonalTileIterator(tile, start_tile) : new OrthogonalTileIterator(tile, start_tile);
+	for (; *iter != INVALID_TILE; ++(*iter)) {
+		TileIndex t = *iter;
+		CommandCost ret = Command<CMD_BUILD_OBJECT>::Do(flags & ~DC_EXEC, t, type, view);
+
+		/* If we've reached the limit, stop building (or testing). */
+		if (c != nullptr && --limit <= 0) break;
+
+		if (ret.Failed()) {
+			last_error = ret;
+			continue;
+		}
+
+		had_success = true;
+		if (flags & DC_EXEC) {
+			money -= ret.GetCost();
+
+			/* If we run out of money, stop building. */
+			if (ret.GetCost() > 0 && money < 0) break;
+			Command<CMD_BUILD_OBJECT>::Do(flags, t, type, view);
+		}
+		cost.AddCost(ret);
+	}
+
+	delete iter;
+	return had_success ? cost : last_error;
+}
 
 static Foundation GetFoundation_Object(TileIndex tile, Slope tileh);
 
--- a/src/object_cmd.h
+++ b/src/object_cmd.h
@@ -14,7 +14,9 @@
 #include "object_type.h"
 
 CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type, uint8 view);
+CommandCost CmdBuildObjectArea(DoCommandFlag flags, TileIndex tile, TileIndex start_tile, ObjectType type, uint8 view, bool diagonal);
 
 DEF_CMD_TRAIT(CMD_BUILD_OBJECT, CmdBuildObject, CMD_DEITY | CMD_NO_WATER | CMD_AUTO, CMDT_LANDSCAPE_CONSTRUCTION)
+DEF_CMD_TRAIT(CMD_BUILD_OBJECT_AREA, CmdBuildObjectArea, CMD_DEITY | CMD_NO_WATER | CMD_AUTO, CMDT_LANDSCAPE_CONSTRUCTION)
 
 #endif /* OBJECT_CMD_H */
--- a/src/object_gui.cpp
+++ b/src/object_gui.cpp
@@ -458,7 +458,7 @@ public:
 		}
 
 		if (_selected_object_index != -1) {
-			SetObjectToPlaceWnd(SPR_CURSOR_TRANSMITTER, PAL_NONE, HT_RECT, this);
+			SetObjectToPlaceWnd(SPR_CURSOR_TRANSMITTER, PAL_NONE, HT_RECT | HT_DIAGONAL, this);
 		}
 
 		this->UpdateButtons(_selected_object_class, _selected_object_index, _selected_object_view);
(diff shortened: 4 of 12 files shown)


LC-Zorg:
This PR is like a warm bath after all day in the cold... ;)

However, I would have a few comments that might be worth considering. I'm writing this having in mind possible improvements to this function, including PR or in the future separately.

1. There is a real risk of using objects to block other players in online games. Therefore, such a possibility should be an option or there should be a setting to limit the construction area - JGR has the following settings:
build_object_per_64k_frames = 2097152
build_object_frame_burst = 2048
If this PR is also adding them, that should be enough.
2. Is it possible to limit the stretching of the construction mesh so that the area is not larger than in the build_object_frame_burst setting? It is a bit strange when you can stretch a construction field without limit, and finally there will be, for example, 1 object on it.
disappointment
example with build_object_frame_burst = 1

3. It will be much easier to reserve an area with objects than by purchasing it. Could this change also apply to the purchase of land? It seems to me that there is no need to separate it - both are de facto objects.
4. This change will allow area-based construction of all 1x1 size objects, but many of these should not necessarily be able to do so. Perhaps the creators should be able to add markings that will prevent such construction of selected objects.

2TallTyler:
1. This PR includes such settings, as well as the global on/off switch (in Settings, chosen by the server owner for all players) for building multiple objects at once. 😄 
2. If a server wants to limit building more than one object, they should turn the feature off completely rather than limiting it with the burst limit.
3. JGRPP also includes this feature, but it is a separate command code-wise and I think would be a separate PR.
4. I don't see a problem with the player doing what they want, besides it looking silly. Forcibly limiting certain objects to a size chosen by the NewGRF author would require an addition to the NewGRF spec, and would be Too Much Work For Too Little Benefit, in my opinion.

src/saveload/saveload.h
```diff
@@ -340,6 +340,7 @@ enum SaveLoadVersion : uint16 {
 	SLV_LINKGRAPH_TRAVEL_TIME,              ///< 297  PR#9457 v12.0-RC1  Store travel time in the linkgraph.
 	SLV_DOCK_DOCKINGTILES,                  ///< 298  PR#9578 All tiles around docks may be docking tiles.
 	SLV_REPAIR_OBJECT_DOCKING_TILES,        ///< 299  PR#9594 v12.0  Fixing issue with docking tiles overlapping objects.
+	SLV_BUILD_OBJECT_BY_AREA,               ///< 300  PR#9709 Build objects by area
```
frosch123: Since SLV_TABLE_CHUNKS (#9322) savegame versions are no longer needed when adding/removing things from savegames. They are only needed if an actual conversion is required.
So I think you can remove the SLV_xxx and all savegame_version_from/to thingies.

src/settings_type.h
```diff
@@ -345,13 +345,16 @@ struct ConstructionSettings {
 	bool   freeform_edges;                   ///< allow terraforming the tiles at the map edges
 	uint8  extra_tree_placement;             ///< (dis)allow building extra trees in-game
 	uint8  command_pause_level;              ///< level/amount of commands that can't be executed while paused
+	bool   build_object_area_permitted;      ///< whether building objects by area is permitted
```
frosch123: Is there an example for when someone would want to disable this? (i.e. why is this a setting?)

src/object_cmd.cpp
```diff
@@ -347,21 +350,100 @@ CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint3
...
+		had_success = true;
+		if (flags & DC_EXEC) {
+			money -= ret.GetCost();
+			if (ret.GetCost() > 0 && money < 0) {
+				_additional_cash_required = ret.GetCost();
+				delete iter;
+				return cost;
+			}
+			DoCommand(t, type, view, flags, CMD_BUILD_OBJECT);
+		}
```
frosch123: No "\n" before "else" please.
There are some more instances of this in the PR.

src/object_cmd.cpp
```diff
@@ -347,21 +350,100 @@ CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint3
 			/* This may never be constructed using this method. */
 			return CMD_ERROR;
 
-		default: // i.e. NewGRF provided.
+		default: { // i.e. NewGRF provided.
+			const ObjectSpec *spec = ObjectSpec::Get(type);
```
frosch123: There is already a variable `spec` in this function, and it already has this value.
There are also `size_x` and `size_y` already.

src/object_cmd.cpp
```diff
@@ -312,10 +312,13 @@ CommandCost CmdBuildObject(TileIndex tile, DoCommandFlag flags, uint32 p1, uint3
 	}
 
 	int hq_score = 0;
+	int build_object_size = 0;
+	Company *c = nullptr;
 	switch (type) {
 		case OBJECT_TRANSMITTER:
 		case OBJECT_LIGHTHOUSE:
 			if (!IsTileFlat(tile)) return_cmd_error(STR_ERROR_FLAT_LAND_REQUIRED);
+			build_object_size = 1;
```
frosch123: Lighthouses and transmitters also have an objectspec, so you can do `build_object_size = size_x * size_y;` outside of the `switch`.

src/settings_type.h
2TallTyler: The usecase, as I understand it, is to make multiplayer griefing more difficult.

LC-Zorg:
> Is there an example for when someone would want to disable this? (i.e. why is this a setting?)

Point 1 in my comments above. Roads and tracks build slower - for some, it can be a new, more powerful tool for sowing chaos on servers. ;)

2TallTyler:
I tried to rebase this to use the new Command structure, but I'm getting lots of incomprehensible (to me) build errors in `network_command.cpp`. What am I doing wrong?

stormcone:
I think it is because there is a missing a `DEF_CMD_TRAIT` in `object_cmd.h`.

Similar to this:
https://github.com/OpenTTD/OpenTTD/blob/22f6502b143041b95702011c0df2c38fea9d9a88/src/object_cmd.h#L18

I'm unsure about the parameters, but something like this:
`DEF_CMD_TRAIT(CMD_BUILD_OBJECT_AREA, CmdBuildObjectArea, CMD_DEITY | CMD_NO_WATER | CMD_AUTO, CMDT_LANDSCAPE_CONSTRUCTION)`

2TallTyler:
Thanks for the tip! It builds now, but no longer works. Only one object gets built. 😕

src/object_gui.cpp
```diff
@@ -544,8 +544,37 @@ class BuildObjectWindow : public Window {
...
+			default: NOT_REACHED();
+			case DDSP_BUILD_OBJECT:
+				if (!_settings_game.construction.freeform_edges) {
+					/* When end_tile is MP_VOID, the error tile will not be visible to the
+					 * user. This happens when terraforming at the southern border. */
+					if (TileX(end_tile) == MapMaxX()) end_tile += TileDiffXY(-1, 0);
+					if (TileY(end_tile) == MapMaxY()) end_tile += TileDiffXY(0, -1);
+				}
+				ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
+				Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
```
stormcone: If you look at the original commit at JGRPP, in the `OnPlaceMouseUp` function the `DoCommandP` called with `CMD_BUILD_OBJECT_AREA`. So I think you probably also should called the `Command` here with `CMD_BUILD_OBJECT_AREA`.

```suggestion
				Command<CMD_BUILD_OBJECT_AREA>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
```

2TallTyler:
It works! One final issue: when I try to build an extremely large area of objects, which I could afford if it weren't too large an area to be allowed, I get an assertion error:
`res_test.GetCost() == res_exec.GetCost() && res_test.Failed() == res_exec.Failed()`

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
 	}
 
 	if (flags & DC_EXEC) {
 		BuildObject(type, tile, _current_company == OWNER_DEITY ? OWNER_NONE : _current_company, nullptr, view);
 
 		/* Make sure the HQ starts at the right size. */
 		if (type == OBJECT_HQ) UpdateCompanyHQ(tile, hq_score);
+
+		/* Subtract the tile from the build limit. */
+		if (c != nullptr) c->build_object_limit -= 1 << 16;
```
stormcone: Should not be the size of the object subtracted instead of 1?
```suggestion
		if (c != nullptr) c->build_object_limit -= build_object_size << 16;
```

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
 	}
 
 	cost.AddCost(ObjectSpec::Get(type)->GetBuildCost() * size_x * size_y);
```
stormcone: I know this is not from this patch, the object spec is available from line 211, so it could be simplified, once we are here :)
```suggestion
	cost.AddCost(spec->GetBuildCost() * size_x * size_y);
```

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 			/* This may never be constructed using this method. */
 			return CMD_ERROR;
 
-		default: // i.e. NewGRF provided.
+		default: { // i.e. NewGRF provided.
+			build_object_size = GB(spec->size, 0, 4) * GB(spec->size, 4, 4);
```
stormcone: In line 222-223 the width and height of the object already determined, (but a little differently), so it could be:
```suggestion
			build_object_size = size_x * size_y;
```

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
 
 		/* Make sure the HQ starts at the right size. */
 		if (type == OBJECT_HQ) UpdateCompanyHQ(tile, hq_score);
+
+		/* Subtract the tile from the build limit. */
+		if (c != nullptr) c->build_object_limit -= build_object_size << 16;
 	}
 
-	cost.AddCost(ObjectSpec::Get(type)->GetBuildCost() * size_x * size_y);
+	cost.AddCost(spec->GetBuildCost() * size_x * size_y);
```
stormcone: Now it can be more simple. :)
```suggestion
	cost.AddCost(spec->GetBuildCost() * build_object_size);
```
Sorry for the nitpicking, could not resist. :P

src/object_gui.cpp
```diff
@@ -544,8 +544,37 @@ class BuildObjectWindow : public Window {
 	void OnPlaceObject(Point pt, TileIndex tile) override
 	{
 		ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
```
stormcone: We use only the object spec, so we do not need the object class separately:
```suggestion
		const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
```

src/object_gui.cpp
```diff
@@ -544,8 +544,37 @@ class BuildObjectWindow : public Window {
...
-		Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
+		if (_settings_game.construction.build_object_area_permitted && objclass->GetSpec(_selected_object_index)->size == 0x11) {
+			VpStartPlaceSizing(tile, VPM_X_AND_Y, DDSP_BUILD_OBJECT);
+		} else {
+			Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
 				tile, objclass->GetSpec(_selected_object_index)->Index(), _selected_object_view);
```
stormcone:
```suggestion
		if (_settings_game.construction.build_object_area_permitted && spec->size == 0x11) {
			VpStartPlaceSizing(tile, VPM_X_AND_Y, DDSP_BUILD_OBJECT);
		} else {
			Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
				tile, spec->Index(), _selected_object_view);
```

src/object_gui.cpp
```diff
@@ -544,8 +544,37 @@ class BuildObjectWindow : public Window {
...
+				case DDSP_BUILD_OBJECT:
+					if (!_settings_game.construction.freeform_edges) {
+						/* When end_tile is MP_VOID, the error tile will not be visible to the
+						 * user. This happens when terraforming at the southern border. */
+						if (TileX(end_tile) == MapMaxX()) end_tile += TileDiffXY(-1, 0);
+						if (TileY(end_tile) == MapMaxY()) end_tile += TileDiffXY(0, -1);
+					}
+					ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
+					Command<CMD_BUILD_OBJECT_AREA>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
+						end_tile, start_tile, objclass->GetSpec(_selected_object_index)->Index(), _selected_object_view, (_ctrl_pressed ? true : false));
```
stormcone: Same as above:
```suggestion
					const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
					Command<CMD_BUILD_OBJECT_AREA>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
						end_tile, start_tile, spec->Index(), _selected_object_view, (_ctrl_pressed ? true : false));
```

src/company_cmd.cpp
```diff
@@ -267,9 +268,10 @@ void SubtractMoneyFromCompanyFract(CompanyID company, const CommandCost &cst)
...
-		c->clear_limit     = std::min<uint64>((uint64)c->clear_limit     + _settings_game.construction.clear_per_64k_frames,     (uint64)_settings_game.construction.clear_frame_burst << 16);
-		c->tree_limit      = std::min<uint64>((uint64)c->tree_limit      + _settings_game.construction.tree_per_64k_frames,      (uint64)_settings_game.construction.tree_frame_burst << 16);
+		c->terraform_limit    = std::min<uint64>((uint64)c->terraform_limit     + _settings_game.construction.terraform_per_64k_frames,   (uint64)_settings_game.construction.terraform_frame_burst << 16);
+		c->clear_limit        = std::min<uint64>((uint64)c->clear_limit         + _settings_game.construction.clear_per_64k_frames,       (uint64)_settings_game.construction.clear_frame_burst << 16);
+		c->tree_limit         = std::min<uint64>((uint64)c->tree_limit          + _settings_game.construction.tree_per_64k_frames,        (uint64)_settings_game.construction.tree_frame_burst << 16);
+		c->build_object_limit = std::min<uint64>((uint64)c->build_object_limit + _settings_game.construction.build_object_per_64k_frames, (uint64)_settings_game.construction.build_object_frame_burst << 16);
```
LordAro, line 274: sad alignment is sad :(

src/object_cmd.cpp
```diff
@@ -311,6 +311,8 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 	}
 
 	int hq_score = 0;
+	uint build_object_size = 1;
+	Company *c = Company::GetIfValid(_current_company);
```
LordAro: no need to get this before usage?

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
 			/* This may never be constructed using this method. */
 			return CMD_ERROR;
 
-		default: // i.e. NewGRF provided.
+		default: { // i.e. NewGRF provided.
```
LordAro: no variable declaration here anymore, so no need for braces

src/object_gui.cpp
```diff
@@ -543,9 +543,39 @@ class BuildObjectWindow : public Window {
...
-		ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
-		Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
-				tile, objclass->GetSpec(_selected_object_index)->Index(), _selected_object_view);
+		const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
+
+		if (_settings_game.construction.build_object_area_permitted && spec->size == 0x11) {
+			VpStartPlaceSizing(tile, VPM_X_AND_Y, DDSP_BUILD_OBJECT);
+		} else {
+			Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
+				tile, spec->Index(), _selected_object_view);
```
LordAro: line continuations should be double indented, but I'd say don't have the linebreak at all

src/object_gui.cpp
```diff
@@ -543,9 +543,39 @@ class BuildObjectWindow : public Window {
...
+	}
+
+	void OnPlaceDrag(ViewportPlaceMethod select_method, ViewportDragDropSelectionProcess select_proc, Point pt) override
+	{
+		VpSelectTilesWithMethod(pt.x, pt.y, select_method);
+	}
+
+	void OnPlaceMouseUp(ViewportPlaceMethod select_method, ViewportDragDropSelectionProcess select_proc, Point pt, TileIndex start_tile, TileIndex end_tile) override
+	{
+		if (pt.x != -1) {
```
LordAro: `if (pt.x == -1) return;` ? Would reduce indentation

src/settings_type.h
LordAro: Does it need a separate setting beyond the frame burst setting though? That should be able to be tuned such that you can only place one object at once, I think?

src/object_cmd.cpp
```diff
@@ -346,21 +348,90 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
+
+		had_success = true;
+		if (flags & DC_EXEC) {
+			money -= ret.GetCost();
+
+			/* If we run out of money, stop building. */
+			if (ret.GetCost() > 0 && money < 0) {
+				delete iter;
+				return cost;
+			}
```
LordAro: I wonder if it would be cleaner to break out of the loop, rather than this explicit return "hidden" in the middle

src/settings_type.h
2TallTyler: ¯\\_ (ツ)_/¯ The frame burst setting would work, but it's hidden in a config file. 

I'll remove the setting and if we get lots of reports of people wanting an easier-to-find setting, I'll add it back in a different PR.

LordAro (dismissed):

2TallTyler:
I forgot to update tooltips to indicate that you can press Ctrl to select diagonally. Ready for a re-review. 🙂

src/object_cmd.cpp
```diff
@@ -347,20 +348,85 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
+ * Construct multiple objects in an area
+ * @param flags of operation to conduct
+ * @param tile end tile of area dragging
+ * @param start_tile start tile of area dragging
+ * @param type the object type to build
+ * @param view the view for the object
+ * @param diagonal Whether to use the Orthogonal (0) or Diagonal (1) iterator.
+ * @return the cost of this operation or an error
+ */
+CommandCost CmdBuildObjectArea(DoCommandFlag flags, TileIndex tile, uint32 start_tile, ObjectType type, uint8 view, bool diagonal)
```
michicc: Use `TileIndex` for `start_tile`.

src/object_cmd.h
```diff
@@ -14,7 +14,9 @@
 #include "object_type.h"
 
 CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type, uint8 view);
-
 DEF_CMD_TRAIT(CMD_BUILD_OBJECT, CmdBuildObject, CMD_DEITY | CMD_NO_WATER | CMD_AUTO, CMDT_LANDSCAPE_CONSTRUCTION)
 
+CommandCost CmdBuildObjectArea(DoCommandFlag flags, TileIndex tile, uint32 start_tile, ObjectType type, uint8 view, bool diagonal);
```
michicc: Use `TileIndex` for `start_tile`.

Also, current _cmd.h files have all functions prototypes first and then all DEF_CMD_TRAIT.

michicc (changes requested):

src/object_cmd.cpp
```diff
@@ -347,20 +348,85 @@ CommandCost CmdBuildObject(DoCommandFlag flags, TileIndex tile, ObjectType type,
...
+{
+	if (start_tile >= MapSize()) return CMD_ERROR;
+
+	if (type >= NUM_OBJECTS) return CMD_ERROR;
+	const ObjectSpec *spec = ObjectSpec::Get(type);
+	if (view >= spec->views) return CMD_ERROR;
+
+	if (spec->size != 0x11) return CMD_ERROR;
```
michicc: Not sure if there already is a proper named constant for this, but it should either be one or at the very least get a comment explaining the magic number.

src/object_gui.cpp
```diff
@@ -543,9 +543,38 @@ class BuildObjectWindow : public Window {
 
 	void OnPlaceObject(Point pt, TileIndex tile) override
 	{
-		ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
-		Command<CMD_BUILD_OBJECT>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
-				tile, objclass->GetSpec(_selected_object_index)->Index(), _selected_object_view);
+		const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
+
+		if (spec->size == 0x11) {
```
michicc: Not sure if there already is a proper named constant for this, but it should either be one or at the very least get a comment explaining the magic number.

michicc (approved):

PR #9716, opened 2021-11-24
Fix: Don't try to rename OWNER_DEITY signs in-game

2TallTyler:
## Motivation / Problem

Many city builder Game Scripts use a sign below the town name to show the current growth rate. Players can click on this sign which brings up the GUI to edit the sign, but they are actually blocked from renaming the sign. The same goes for signs created by other players.

Why show the rename GUI if we're not allowed to do anything with it?

sign_error

## Description

If we're not allowed to rename a sign, don't even open the rename GUI.

The exception is if we're in Scenario Editor, which can open savegames by changing the file extension, and needs to be allowed to edit anything including signs.

## Checklist for review

### DIFF
--- a/src/signs_gui.cpp
+++ b/src/signs_gui.cpp
@@ -565,10 +565,14 @@ static WindowDesc _query_sign_edit_desc(
  */
 void HandleClickOnSign(const Sign *si)
 {
+	/* If we can't rename the sign, don't even open the rename GUI. */
+	if (!CompanyCanRenameSign(si)) return;
+
 	if (_ctrl_pressed && (si->owner == _local_company || (si->owner == OWNER_DEITY && _game_mode == GM_EDITOR))) {
 		RenameSign(si->index, "");
 		return;
 	}
+
 	ShowRenameSignWindow(si);
 }
 
--- a/src/signs.cpp
+++ b/src/signs.cpp
@@ -9,6 +9,7 @@
 
 #include "stdafx.h"
 #include "landscape.h"
+#include "company_func.h"
 #include "signs_base.h"
 #include "signs_func.h"
 #include "strings_func.h"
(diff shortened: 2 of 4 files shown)


LordAro (dismissed):
Comment Ian"t great - it explains "what", which the code already does. Comments should explain "why", if needed

ldpl:
It's very possible to edit signs of other companies. For good or not but currently, you just highjack the sign when you edit it. The only ownership check CmdRenameSign does is for OWNER_DEITY which is entirely different from the check in this PR: 
```
if (si->owner == OWNER_DEITY && _current_company != OWNER_DEITY && _game_mode != GM_EDITOR) return CMD_ERROR;
```

2TallTyler:
> It's very possible to edit signs of other companies.

That's...odd. I've amended the check and PR title to reflect the real itch being scratched by this PR, which is trying to edit a sign we're not allowed to change. Hopefully the comment better explains the "why" now, too.

ldpl:
Maybe move sign ownership check to a separate function? To reduce the chance of anyone making the same mistake again.

2TallTyler:
It's a pretty specific usecase, but something like `CompanyCanRenameSign()`?

2TallTyler:
`CompanyCanRenameSign()` created and now both `CmdRenameSign()` and `HandleClickOnSign()` use it instead of duplicating the same code.

src/signs_gui.cpp
```diff
@@ -569,6 +569,10 @@ void HandleClickOnSign(const Sign *si)
 		RenameSign(si->index, "");
 		return;
 	}
+
+	/* If we can't rename the sign, don't even open the rename GUI. */
+	if (!CompanyCanRenameSign(si)) return;
```
ldpl: IMO it would be more logical to check this before the ctrl-click even though currently it doesn't matter as ctrl-click condition is more strict.

LordAro (approved):

PR #9725, opened 2021-12-01
Template DoCommand and friends

michicc:
## Motivation / Problem

Command handlers are passed their data in bit-stuffed form, and with a fixed amount of bits available. This is a) prone to errors as there is no compile-time check if parameters where stuffed into the right spot, and b) limits extensibility as there's some reservation to blanket increase the parameters for one single thing.

Both problems can be solved if one allows arbitrary, typed parameters for commands using some C++ magic.

## Description and Discussion

With this PR, DoCommand and friends are spiced up using some C++17 template magic to achieve variable amounts of typed parameters. Main implementation points:
* Command parameter types are extracted by using a non-intrusive type-traits class.
* These types are reflected into templated DoCommand functions to let the command execution machinery expose typed parameters.
* Network transfer of commands is accomplished by (de-)serialising the parameters from/to an unstructured byte stream. Adding structure (type markers etc) has not benefit IMHO, as network play is revision-locked anyway.
* 
* Commands can return more than just a `CommandCost`, making most of the previously used globals obsolete.
	* Direct command execution (i.e. the old `DoCommand`) directly returns the command proc returns.
	* Indirect execution (i.e. `DoCommandP`) passes the return values on to the command callbacks.
	* Command callbacks are compile-time optimized by their type as most callbacks just want success/failure and maybe a tile reference. Some callbacks need the command parameters and the extra return values. With judicious use of `if constexpr`, the compiler can be tasked with generating matching function calls.
	* The command callback used for AI/GS is special as they use one callback for all commands and still want to capture command parameters for validation in a networking context. To allows this use, one possible form of the callback function receives the command parameters and return values in the same serialised format as used in networking.

Some additional random comments:
* Modern compilers are scarily good at optimising templates. I checked various things on godbolt along the way to make sure the implementation doesn't completely suck.
	* One of the recent test is here: [https://godbolt.org/z/ja9Y7ane6](https://godbolt.org/z/ja9Y7ane6). Looking at the resulting assembly, you'll find that everything in the `struct Do` execution layer is completely eliminated (including e.g. the various tuple gymnastics), leaving just a plain old static function call.
	* While I didn't check the complete PR with everything in godbolt, I did tests with various bits and bobs and generally found the same as in the example above. Compilers are really good with templated code these days.
* I'm not really happy with the way command callbacks for AI/GS is solved, but I didn't have any better idea.
* A bit the same is the network command unpacking that essentially uses a 2D table over CMD_* and all defined callback functions populated with template specialisations to simulate runtime template resolving. Not very elegant, but the only thing I came up with.
* As typedefs don't introduce new types, `TileIndex` is changed into a struct to be able to reliably used in template specialisation.

## Limitations

There are bugs, 100% guaranteed 😄 
This changes the on-network command packet, no idea if this necessitates any changes to the GC or the TURN relay.
I might have forgotten to update some documentation somewhere.

## Checklist for review


src/command_func.h
```diff
@@ -32,22 +32,24 @@ static const CommandCost CMD_ERROR = CommandCost(INVALID_STRING_ID);
...
 CommandCost DoCommand(const CommandContainer *container, DoCommandFlag flags);
 
-bool DoCommandP(uint32 cmd, CommandCallback *callback, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
-bool DoCommandP(uint32 cmd, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
-bool DoCommandP(const CommandContainer *container, bool my_cmd = true);
+bool DoCommandP(Commands cmd, StringID err_message, CommandCallback *callback, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
+bool DoCommandP(Commands cmd, StringID err_message, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
+bool DoCommandP(Commands cmd, CommandCallback *callback, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
+bool DoCommandP(Commands cmd, TileIndex tile, uint32 p1, uint32 p2, const std::string &text = {});
+bool DoCommandP(const CommandContainer *container, bool my_cmd = true, bool network_command = false);
```
frosch123: No idea whether this is changed later again anyway, but with so many DoCommandP overloads it's scary when there are multiple consecutive "bool" parameters. It's easy to swap them or even implicitly convert ints to bools.

Would it make sense to combine all bools (my_cmd, network_command, estimate_only) into a flag-enum again?

michicc: Yes, none of the global un-templated DoCommand* functions survive until the end. Some of the intermediary steps are just because each and every commit should compile and lead to a working build (unless I made some snafu somewhere, of course).

src/newgrf_station.cpp
```diff
@@ -223,7 +223,7 @@ static uint32 GetRailContinuationInfo(TileIndex tile)
 /* Station Resolver Functions */
 /* virtual */ uint32 StationScopeResolver::GetRandomBits() const
 {
-	return (this->st == nullptr ? 0 : this->st->random_bits) | (this->tile == INVALID_TILE ? 0 : GetStationTileRandomBits(this->tile) << 16);
+	return (this->st == nullptr ? 0 : this->st->random_bits) | (this->tile == INVALID_TILE ? 0: GetStationTileRandomBits(this->tile) << 16);
```
rubidium42: Space gone missing,

src/aircraft_cmd.cpp
```diff
@@ -1275,7 +1276,7 @@ void HandleMissingAircraftOrders(Aircraft *v)
 	const Station *st = GetTargetAirportIfValid(v);
 	if (st == nullptr) {
 		Backup<CompanyID> cur_company(_current_company, v->owner, FILE_LINE);
-		CommandCost ret = DoCommand(DC_EXEC, CMD_SEND_VEHICLE_TO_DEPOT, v->tile, v->index, 0);
+		CommandCost ret = DoCommand<CMD_SEND_VEHICLE_TO_DEPOT>::Do(DC_EXEC, v->tile, v->index, 0, {});
```
rubidium42: Not sure if anything can be done about it since there's still a `Command` struct, though `DoCommand::Do` sounds double. `Command::Do` would look and sound better. Although that change might also affect naming things like DC_EXEC.

src/command_func.h
```diff
@@ -143,6 +137,81 @@ struct DoCommandHelper<Tcmd, CommandCost(*)(DoCommandFlag, Targs...)> : protecte
...
+	 * @param callback A callback function to call after the command is finished
+	 * @param args Parameters for the command
+	 * @return \c true if the command succeeded, else \c false.
+	 */
+	static bool Post(StringID err_message, CommandCallback *callback, Targs... args)
+	{
+		return InternalPost(err_message, callback, true, false, std::forward_as_tuple(args...));
+	}
+
+	/** Toplevel docommand function for commands comming from the net.
```
rubidium42: Missing newline

src/command_func.h
```diff
@@ -78,6 +78,19 @@ struct RecursiveCommandCounter {
 	static int _counter;
 };
 
+#if defined(__GNUC__) && !defined(__clang__)
+/*
```
rubidium42, line 80: Might it be worth defining something here like `#define SILENCE_GCC_CLANG_FUNCTION_POINTER_CAST` or something, which then gets used in all other places where we push or pop the `pragma GCC diagonostics`. If at some point some other compiler needs it, you only need to change it in one location.

src/aircraft_cmd.cpp
michicc: Yeah, there can only be one `Command` struct(tm).  OTOH the current `Command` struct is used in one single place and could be easily renamed.

I wouldn't mind a different name, but I don't know what everybody else thinks about it.

michicc:
Famous last words: I might have figured MingW64 out. For whatever reason, it fails on linking when casting a function pointer to its own type, but works if I insert a detour to some other function type in between. (Fingers crossed all other compilers stay happy with it).

And, to surface one comment from Rb: Should it be `DoCommand<>::Do/Post/...` or `Command<>::Do/Post/...`, i.e. keep the old DoCommand naming or not?

src/aircraft_cmd.cpp
TrueBrain: I like the idea of `Command::Do`, feels neat.

On the subject, I do think `Command::Post` would be a confusing name, but I struggle with finding a good name. From what I remember, `Command::Do` can be executed inside a command, but `Command::Post` cannot; the latter should also always be the one to execute when starting a command.

So something like `Command::Execute`? Or `Command::Safe`? It is difficult .. but for sure "Post" confuses me, as you think it is the "post" (vs "pre") handling of the command; which it is not (I think :D) :P

michicc: `Post` was just something I could come up with that whoever came up with the initial `P` might have though of.

If we were to rename it, I like `Execute` more then `Safe`.

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+		return InternalPost(err_message, callback, my_cmd, network_command, tile, std::move(args));
+	}
+
+	template <typename Tcallback>
+	static bool InternalPost(StringID err_message, Tcallback *callback, bool my_cmd, bool network_command, TileIndex tile, std::tuple<Targs...> args)
+	{
+		auto [err, estimate_only, only_sending] = InternalPostPre(Tcmd, GetCommandFlags<Tcmd>(), tile, err_message, network_command);
+		if (err) return false;
+
+		/* Only set p2 when the command does not come from the network. */
```
TrueBrain: p2? :D

src/ai/ai_instance.cpp
```diff
@@ -92,13 +93,13 @@ ScriptInfo *AIInstance::FindLibrary(const char *library, int version)
 
 /**
  * DoCommand callback function for all commands executed by AIs.
+ * @param cmd cmd as given to DoCommandPInternal.
```
rubidium42, line 96: DoCommandPInternal does not exist anymore (also for data param).

src/game/game_instance.cpp
```diff
@@ -81,20 +82,20 @@ void GameInstance::Died()
 
 /**
  * DoCommand callback function for all commands executed by Game Scripts.
+ * @param cmd cmd as given to DoCommandPInternal.
```
rubidium42, line 85: DoCommandPInternal does not exist anymore (also for data param).

src/command.cpp
```diff
@@ -378,9 +95,7 @@ static const Command _command_proc_table[] = {
  */
 bool IsValidCommand(uint32 cmd)
```
rubidium42: Shouldn't the `uint32` be `Commands`?

src/misc/endian_buffer.hpp
```diff
@@ -0,0 +1,215 @@
+/* $Id$ */
```
rubidium42: You must have been busy with this a long time, given the Subversion comment ;)

src/misc_cmd.cpp
```diff
@@ -32,17 +33,14 @@ static_assert((LOAN_INTERVAL & 3) == 0);
...
  * Increase the loan of your company.
- * @param tile unused
  * @param flags operation to perform
- * @param p1 higher half of amount to increase the loan with, multitude of LOAN_INTERVAL. Only used when (p2 & 3) == 2.
- * @param p2 (bit 2-31) - lower half of amount (lower 2 bits assumed to be 0)
- *           (bit 0-1)  - when 0: loans LOAN_INTERVAL
- *                        when 1: loans the maximum loan permitting money (press CTRL),
- *                        when 2: loans the amount specified in p1 and p2
- * @param text unused
+ * @param cmd when LoanCommand::Interval: loans LOAN_INTERVAL
```
rubidium42: Missing comma at end of line (like the next line has)

src/misc_cmd.cpp
```diff
@@ -80,34 +78,31 @@ CommandCost CmdIncreaseLoan(TileIndex tile, DoCommandFlag flags, uint32 p1, uint
...
  * Decrease the loan of your company.
- * @param tile unused
  * @param flags operation to perform
- * @param p1 higher half of amount to decrease the loan with, multitude of LOAN_INTERVAL. Only used when (p2 & 3) == 2.
- * @param p2 (bit 2-31) - lower half of amount (lower 2 bits assumed to be 0)
- *           (bit 0-1)  - when 0: pays back LOAN_INTERVAL
- *                        when 1: pays back the maximum loan permitting money (press CTRL),
- *                        when 2: pays back the amount specified in p1 and p2
- * @param text unused
+ * @param cmd when LoanCommand::Interval: pays back LOAN_INTERVAL
```
rubidium42: Missing comma at end of line (like the next line has)

src/misc_cmd.cpp
```diff
@@ -195,34 +188,26 @@ CommandCost CmdPause(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2,
...
- * @param tile unused
  * @param flags operation to perform
- * @param p1 the amount of money to receive (if positive), or spend (if negative)
- * @param p2 unused
- * @param text unused
+ * @param amount the amount of money to receive (if positive), or spend (if negative)
  * @return the cost of this operation or an error
  */
-CommandCost CmdMoneyCheat(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdMoneyCheat(DoCommandFlag flags, int32 amount)
```
rubidium42: Shouldn't `int32` become `Money` now that's a trivial change?

src/misc_cmd.cpp
```diff
@@ -195,34 +188,26 @@ CommandCost CmdPause(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2,
...
- *           (bit 8-15) - the expenses type which should register the cost/income @see ExpensesType.
- * @param text unused
+ * @param tile tile to show text effect on (if not 0)
+ * @param delta the amount of money to receive (if positive), or spend (if negative)
+ * @param company the company ID.
+ * @param expenses_type the expenses type which should register the cost/income @see ExpensesType.
  * @return zero cost or an error
  */
-CommandCost CmdChangeBankBalance(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdChangeBankBalance(DoCommandFlag flags, TileIndex tile, int64 delta, CompanyID company, ExpensesType expenses_type)
```
rubidium42: `int64` -> `Money`?

src/misc_cmd.cpp
```diff
@@ -32,17 +33,14 @@ static_assert((LOAN_INTERVAL & 3) == 0);
...
- *                        when 2: loans the amount specified in p1 and p2
- * @param text unused
+ * @param cmd when LoanCommand::Interval: loans LOAN_INTERVAL
+ *            when LoanCommand::Max: loans the maximum loan permitting money (press CTRL),
+ *            when LoanCommand::Amount: loans the amount specified in \c amount
+ * @param amount amount to increase the loan with, multitude of LOAN_INTERVAL. Only used when cmd == LoanCommand::Amount.
  * @return the cost of this operation or an error
  */
-CommandCost CmdIncreaseLoan(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdIncreaseLoan(DoCommandFlag flags, LoanCommand cmd, int64 amount)
```
rubidium42: `int64` -> `Money`?

src/misc_cmd.cpp
```diff
@@ -80,34 +78,31 @@ CommandCost CmdIncreaseLoan(TileIndex tile, DoCommandFlag flags, uint32 p1, uint
...
- *                        when 2: pays back the amount specified in p1 and p2
- * @param text unused
+ * @param cmd when LoanCommand::Interval: pays back LOAN_INTERVAL
+ *            when LoanCommand::Max: pays back the maximum loan permitting money (press CTRL),
+ *            when LoanCommand::Amount: pays back the amount specified in \c amount
+ * @param amount amount to decrease the loan with, multitude of LOAN_INTERVAL. Only used when cmd == LoanCommand::Amount.
  * @return the cost of this operation or an error
  */
-CommandCost CmdDecreaseLoan(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdDecreaseLoan(DoCommandFlag flags, LoanCommand cmd, int64 amount)
```
rubidium42: `int64` -> `Money`?

src/misc_cmd.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef MISC_CMD_H
+#define MISC_CMD_H
+
+#include "command_type.h"
+#include "economy_type.h"
+
+enum PauseMode : byte;
+
+enum class LoanCommand : byte {
+	Interval,
```
rubidium42, line 19: Isn't coding style that all enums are all uppercase?

nielsmh: This is an `enum class` so the enum declaration forms a namespace. You can only refer to the members qualified, such as `LoanCommand::Interval`. In my opinion the convention of uppercase for constants falls away in that case, since it's no longer (as) necessary to make the enum members unique.

michicc: Current precedent in the code is that `enum class` uses normal casing for the enum members.

michicc:
Renamed to `Command<>::` from `DoCommand<>::`, but still `Command<>:Post` due to lack of consensus.

src/misc_cmd.h
rubidium42: That precedent is inconsistent though. Both "normal" case and upper case exist for `enum class`. So changing the current upper case ones to "normal" case and documenting it in the coding style could be a solution to my remark, though for a separate PR.

src/ai/ai_instance.cpp
rubidium42: You only changed the comment for one of the two parameters (likewise for game_instance.cpp)

src/road_cmd.cpp
```diff
@@ -966,59 +962,49 @@ static bool CanConnectToRoad(TileIndex tile, RoadType rt, DiagDirection dir)
...
  * @return the cost of this operation or an error
  */
-CommandCost CmdBuildLongRoad(TileIndex start_tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdBuildLongRoad(DoCommandFlag flags, TileIndex start_tile, TileIndex end_tile, RoadType rt, Axis axis, DisallowedRoadDirections drd, bool start_half, bool end_half, bool is_ai)
 {
-	DisallowedRoadDirections drd = DRD_NORTHBOUND;
-
-	if (p1 >= MapSize()) return CMD_ERROR;
-	TileIndex end_tile = p1;
+	if (end_tile >= MapSize()) return CMD_ERROR;
```
rubidium42, line 980: Isn't the tile checking in the command code (DoCommandHelper::Do) preventing a tile from being larger than the map size? Also holds for `CmdRemoveLongRoad`.

src/road_gui.cpp
```diff
@@ -124,76 +112,71 @@ void ConnectRoadToStructure(TileIndex tile, DiagDirection direction)
...
- *           bit 2: Allow stations directly adjacent to other stations.
- *           bit 3..4: Entrance direction (#DiagDirection) for normal stops.
- *           bit 3: #Axis of the road for drive-through stops.
- *           bit 5..9: The roadtype.
- *           bit 16..31: Station ID to join (NEW_STATION if build new one).
- * @param cmd Unused.
+ * @param width Width of the road stop.
+ * @param length Length of the road stop.
+ * @param is_drive_through False for normal stops, true for drive-through.
+ * @param ddir Entrance direction (#DiagDirection) for normal stops. Converted to the axis for drive-through stops.
```
rubidium42: Parameter in the function below is called `dir` not `ddir`.

src/roadveh_cmd.h
```diff
@@ -0,0 +1,23 @@
...
+/** @file roadveh_cmd.h Command definitions related to road vehicles. */
+
+#ifndef ROADVEH_CMD_H
+#define ROADVEH_CMD_H
+
+#include "command_type.h"
+#include "engine_type.h"
+#include "vehicle_type.h"
+
+CommandCost CmdBuildRoadVehicle(DoCommandFlag flags, TileIndex tile,const Engine *e, Vehicle **v);
```
rubidium42: Missing space before `const`.

src/script/api/script_object.cpp
```diff
@@ -161,78 +159,21 @@ ScriptObject::ActiveInstance::~ActiveInstance()
...
-	SetNewStoryPageElementID(_new_story_page_element_id);
 }
 
 /* static */ bool ScriptObject::GetLastCommandRes()
 {
 	return GetStorage()->last_command_res;
 }
 
-/* static */ void ScriptObject::SetNewVehicleID(VehicleID vehicle_id)
+/* static */void ScriptObject::SetLastCommandResData(CommandDataBuffer data)
```
rubidium42: Erroneously removed space?

src/script/api/script_object.hpp
```diff
@@ -263,41 +269,83 @@ friend class ScriptController;
...
+	if ((::GetCommandFlags<Tcmd>() & CMD_STR_CTRL) == 0) {
+		ScriptObjectInternal::SanitizeStringsHelper(args, std::index_sequence_for<Targs...>{});
+	}
+
+	TileIndex tile{};
+	if constexpr (std::is_same_v<TileIndex, std::tuple_element_t<0, decltype(args)>>) {
+		tile = std::get<0>(args);
+	}
+
+	/* Only set p2 when the command does not come from the network. */
```
rubidium42: `p2` doesn't exist anymore.

src/settings.cpp
```diff
@@ -1506,32 +1505,30 @@ CommandCost CmdChangeSetting(TileIndex tile, DoCommandFlag flags, uint32 p1, uin
...
- * @param p1 unused
- * @param p2 the new value for the setting
+ * @param name the name of the company setting to change
+ * @param value the new value for the setting
  * The new value is properly clamped to its minimum/maximum when setting
- * @param text the name of the company setting to change
  * @return the cost of this operation or an error
  */
-CommandCost CmdChangeCompanySetting(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdChangeCompanySetting(DoCommandFlag flags, const std::string& name, int32 value)
```
rubidium42: The `&` should be before `name`, not just after `std::string`.

src/settings_cmd.h
```diff
@@ -0,0 +1,21 @@
...
+
+/** @file settings_cmd.h Command definitions related to settings. */
+
+#ifndef SETTINGS_CMD_H
+#define SETTINGS_CMD_H
+
+#include "command_type.h"
+
+CommandCost CmdChangeSetting(DoCommandFlag flags, const std::string& name, int32 value);
+CommandCost CmdChangeCompanySetting(DoCommandFlag flags, const std::string& name, int32 value);
```
rubidium42: The `&` should be before `name`, not just after `std::string`.

src/signs_cmd.h
```diff
@@ -0,0 +1,24 @@
...
+/** @file signs_cmd.h Command definitions related to signs. */
+
+#ifndef SIGNS_CMD_H
+#define SIGNS_CMD_H
+
+#include "command_type.h"
+#include "signs_type.h"
+
+std::tuple<CommandCost, SignID> CmdPlaceSign(DoCommandFlag flags, TileIndex tile, const std::string& text);
+CommandCost CmdRenameSign(DoCommandFlag flags, SignID sign_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/company_cmd.h
```diff
@@ -0,0 +1,34 @@
...
+#include "company_type.h"
+#include "livery.h"
+
+enum ClientID : uint32;
+enum Colours : byte;
+
+CommandCost CmdCompanyCtrl(DoCommandFlag flags, CompanyCtrlAction cca, CompanyID company_id, CompanyRemoveReason reason, ClientID client_id);
+CommandCost CmdGiveMoney(DoCommandFlag flags, uint32 money, CompanyID dest_company);
+CommandCost CmdRenameCompany(DoCommandFlag flags, const std::string& text);
+CommandCost CmdRenamePresident(DoCommandFlag flags, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/depot_cmd.h
```diff
@@ -0,0 +1,22 @@
...
+
+/** @file depot_cmd.h Command definitions related to depots. */
+
+#ifndef DEPOT_CMD_H
+#define DEPOT_CMD_H
+
+#include "command_type.h"
+#include "depot_type.h"
+
+CommandCost CmdRenameDepot(DoCommandFlag flags, DepotID depot_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/goal_cmd.h
```diff
@@ -0,0 +1,33 @@
...
+#include "command_type.h"
+#include "command_type.h"
+#include "goal_type.h"
+
+std::tuple<CommandCost, GoalID> CmdCreateGoal(DoCommandFlag flags, CompanyID company, GoalType type, GoalTypeID dest, const std::string& text);
+CommandCost CmdRemoveGoal(DoCommandFlag flags, GoalID goal);
+CommandCost CmdSetGoalText(DoCommandFlag flags, GoalID goal, const std::string& text);
+CommandCost CmdSetGoalProgress(DoCommandFlag flags, GoalID goal, const std::string& text);
+CommandCost CmdSetGoalCompleted(DoCommandFlag flags, GoalID goal, bool completed);
+CommandCost CmdGoalQuestion(DoCommandFlag flags, uint16 uniqueid, uint16 target, bool is_client, uint32 button_mask, GoalQuestionType type, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/group_cmd.h
```diff
@@ -0,0 +1,47 @@
...
+enum GroupFlags : uint8;
+
+/** Action for \c CmdAlterGroup. */
+enum class AlterGroupMode : byte {
+	Rename,    ///< Change group name.
+	SetParent, ///< Change group parent.
+};
+
+std::tuple<CommandCost, GroupID> CmdCreateGroup(DoCommandFlag flags, VehicleType vt, GroupID parent_group);
+CommandCost CmdAlterGroup(DoCommandFlag flags, AlterGroupMode mode, GroupID group_id, GroupID parent_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/industry_cmd.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "command_type.h"
+#include "company_type.h"
+#include "industry_type.h"
+
+enum class IndustryAction : byte;
+enum IndustryControlFlags : byte;
+
+CommandCost CmdBuildIndustry(DoCommandFlag flags, TileIndex tile, IndustryType it, uint32 first_layout, bool fund, uint32 seed);
+CommandCost CmdIndustryCtrl(DoCommandFlag flags, IndustryID ind_id, IndustryAction action, IndustryControlFlags ctlflags, Owner company_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/news_cmd.h
```diff
@@ -0,0 +1,21 @@
...
+/** @file news_cmd.h Command definitions related to news messages. */
+
+#ifndef NEWS_CMD_H
+#define NEWS_CMD_H
+
+#include "command_type.h"
+#include "company_type.h"
+#include "news_type.h"
+
+CommandCost CmdCustomNewsItem(DoCommandFlag flags, NewsType type, NewsReferenceType reftype1, CompanyID company, uint32 reference, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/story.cpp
```diff
@@ -278,43 +269,39 @@ CommandCost CmdCreateStoryPageElement(TileIndex tile, DoCommandFlag flags, uint3
...
- *        (bit  16 -  31) - unused
- * @param p2 Id of referenced object
+ * @param tile Tile location if it is a location page element, otherwise unused.
+ * @param page_element_id The page element to update.
+ * @param reference Id of referenced object
  * @param text Text content in case it is a text or location page element
  * @return the cost of this operation or an error
  */
-CommandCost CmdUpdateStoryPageElement(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdUpdateStoryPageElement(DoCommandFlag flags, TileIndex tile, StoryPageElementID page_element_id, uint32 reference, const std::string& text)
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/story_cmd.h
```diff
@@ -0,0 +1,39 @@
...
+#include "command_type.h"
+#include "company_type.h"
+#include "date_type.h"
+#include "story_type.h"
+#include "vehicle_type.h"
+
+std::tuple<CommandCost, StoryPageID> CmdCreateStoryPage(DoCommandFlag flags, CompanyID company, const std::string& text);
+std::tuple<CommandCost, StoryPageElementID> CmdCreateStoryPageElement(DoCommandFlag flags, TileIndex tile, StoryPageID page_id, StoryPageElementType type, uint32 reference, const std::string& text);
+CommandCost CmdUpdateStoryPageElement(DoCommandFlag flags, TileIndex tile, StoryPageElementID page_element_id, uint32 reference, const std::string& text);
+CommandCost CmdSetStoryPageTitle(DoCommandFlag flags, StoryPageID page_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/town_cmd.h
```diff
@@ -0,0 +1,42 @@
...
+
+enum TownEffect : byte;
+
+std::tuple<CommandCost, Money, TownID> CmdFoundTown(DoCommandFlag flags, TileIndex tile, TownSize size, bool city, TownLayout layout, bool random_location, uint32 townnameparts, const std::string& text);
+CommandCost CmdRenameTown(DoCommandFlag flags, TownID town_id, const std::string& text);
+CommandCost CmdDoTownAction(DoCommandFlag flags, TownID town_id, uint8 action);
+CommandCost CmdTownGrowthRate(DoCommandFlag flags, TownID town_id, uint16 growth_rate);
+CommandCost CmdTownRating(DoCommandFlag flags, TownID town_id, CompanyID company_id, int16 rating);
+CommandCost CmdTownCargoGoal(DoCommandFlag flags, TownID town_id, TownEffect te, uint32 goal);
+CommandCost CmdTownSetText(DoCommandFlag flags, TownID town_id, const std::string& text);
```
rubidium42: The `&` should be before `text`, not just after `std::string`.

src/tree_cmd.cpp
```diff
@@ -378,27 +380,25 @@ void GenerateTrees()
...
  */
-CommandCost CmdPlantTree(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)
+CommandCost CmdPlantTree(DoCommandFlag flags, TileIndex tile, TileIndex start_tile, byte tree_to_plant)
 {
 	StringID msg = INVALID_STRING_ID;
 	CommandCost cost(EXPENSES_OTHER);
-	const byte tree_to_plant = GB(p1, 0, 8); // We cannot use Extract as min and max are climate specific.
 
-	if (p2 >= MapSize()) return CMD_ERROR;
+	if (start_tile >= MapSize()) return CMD_ERROR;
```
rubidium42, line 394: Isn't this already checked by `DoCommandHelper::Do`?

src/vehicle_cmd.cpp
```diff
@@ -30,83 +30,80 @@
...
 
 #include "safeguards.h"
 
-/* Tables used in vehicle.h to find the right command for a certain vehicle type */
-const uint32 _veh_build_proc_table[] = {
-	CMD_BUILD_VEHICLE | CMD_MSG(STR_ERROR_CAN_T_BUY_TRAIN),
-	CMD_BUILD_VEHICLE | CMD_MSG(STR_ERROR_CAN_T_BUY_ROAD_VEHICLE),
-	CMD_BUILD_VEHICLE | CMD_MSG(STR_ERROR_CAN_T_BUY_SHIP),
-	CMD_BUILD_VEHICLE | CMD_MSG(STR_ERROR_CAN_T_BUY_AIRCRAFT),
+/* Tables used in vehicle.h to find the right error message for a certain vehicle type */
```
rubidium42: s/vehicle.h/vehicle_func.h/

src/vehicle_func.h
```diff
@@ -115,59 +115,55 @@ const struct Livery *GetEngineLivery(EngineID engine_type, CompanyID company, En
...
 /* Functions to find the right command for certain vehicle type */
-static inline uint32 GetCmdBuildVeh(VehicleType type)
+static inline StringID GetCmdBuildVehMsg(VehicleType type)
 {
-	return _veh_build_proc_table[type];
+	return _veh_build_msg_table[type];
 }
 
-static inline uint32 GetCmdBuildVeh(const BaseVehicle *v)
+static inline StringID GetCmdBuildVehMsg(const BaseVehicle *v)
```
rubidium42, line 129: Unused function?

src/vehicle_func.h
```diff
@@ -115,59 +115,55 @@ const struct Livery *GetEngineLivery(EngineID engine_type, CompanyID company, En
...
 
-static inline uint32 GetCmdSellVeh(VehicleType type)
+static inline StringID GetCmdSellVehMsg(VehicleType type)
 {
-	return _veh_sell_proc_table[type];
+	return _veh_sell_msg_table[type];
 }
 
-static inline uint32 GetCmdSellVeh(const BaseVehicle *v)
+static inline StringID GetCmdSellVehMsg(const BaseVehicle *v)
```
rubidium42, line 139: Unused function?

src/vehicle_func.h
```diff
@@ -115,59 +115,55 @@ const struct Livery *GetEngineLivery(EngineID engine_type, CompanyID company, En
...
 
-static inline uint32 GetCmdSendToDepot(VehicleType type)
+static inline StringID GetCmdSendToDepotMsg(VehicleType type)
 {
-	return _send_to_depot_proc_table[type];
+	return _send_to_depot_msg_table[type];
 }
 
-static inline uint32 GetCmdSendToDepot(const BaseVehicle *v)
+static inline StringID GetCmdSendToDepotMsg(const BaseVehicle *v)
```
rubidium42, line 159: Unused function?

src/vehicle_func.h
```diff
@@ -115,59 +115,55 @@ const struct Livery *GetEngineLivery(EngineID engine_type, CompanyID company, En
...
 SpriteID GetVehiclePalette(const Vehicle *v);
 
-extern const uint32 _veh_build_proc_table[];
-extern const uint32 _veh_sell_proc_table[];
-extern const uint32 _veh_refit_proc_table[];
-extern const uint32 _send_to_depot_proc_table[];
+extern const StringID _veh_build_msg_table[];
+extern const StringID _veh_sell_msg_table[];
+extern const StringID _veh_refit_msg_table[];
+extern const StringID _send_to_depot_msg_table[];
```
rubidium42, line 121: Maybe not for this PR, though... these tables and all the related functions in this file are only used once or not at all. Maybe move the table to the respective users? Or even use `STR_ERROR_CAN_T_BUY_TRAIN + v->type` instead at those location?

src/command.cpp
```diff
@@ -24,187 +24,69 @@
...
-
-CommandProc CmdGiveMoney;
-CommandProc CmdMoneyCheat;
-CommandProc CmdChangeBankBalance;
-CommandProc CmdBuildCanal;
-CommandProc CmdBuildLock;
+/**
+ * Define a command with the flags which belongs to it.
+ *
+ * This struct connect a command handler function with the flags created with
```
rubidium42: s/connect/connects/?

src/command.cpp
```diff
@@ -213,174 +95,18 @@ CommandProc CmdOpenCloseAirport;
...
-	DEF_CMD(CmdAutofillTimetable,                              0, CMDT_ROUTE_MANAGEMENT      ), // CMD_AUTOFILL_TIMETABLE
-	DEF_CMD(CmdSetTimetableStart,                              0, CMDT_ROUTE_MANAGEMENT      ), // CMD_SET_TIMETABLE_START
-
-	DEF_CMD(CmdOpenCloseAirport,                               0, CMDT_ROUTE_MANAGEMENT      ), // CMD_OPEN_CLOSE_AIRPORT
-};
+static constexpr auto _command_proc_table = MakeCommandsFromTraits(std::make_integer_sequence<std::underlying_type_t<Commands>, CMD_END>{});
+
 
 /*!
  * This function range-checks a cmd, and checks if the cmd is not nullptr
```
rubidium42: It doesn't `nullptr` check anymore

src/command.cpp
```diff
@@ -603,62 +270,25 @@ bool DoCommandP(TileIndex tile, uint32 p1, uint32 p2, uint32 cmd, CommandCallbac
...
-	CommandFlags cmd_flags = GetCommandFlags(cmd);
-	/* Flags get send to the DoCommand */
-	DoCommandFlag flags = CommandFlagsToDCFlags(cmd_flags);
-
-	/* Make sure p2 is properly set to a ClientID. */
-	assert(!(cmd_flags & CMD_CLIENT_ID) || p2 != 0);
-
 	/* Do not even think about executing out-of-bounds tile-commands */
-	if (tile != 0 && (tile >= MapSize() || (!IsValidTile(tile) && (cmd_flags & CMD_ALL_TILES) == 0))) return_dcpi(CMD_ERROR);
+	if (tile != 0 && (tile >= MapSize() || (!IsValidTile(tile) && (cmd_flags & CMD_ALL_TILES) == 0))) return false;
```
rubidium42, line 291: Doesn't `DoCommandHelper::Do` already perform this check?

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+	 * @param args Parameters for the command
+	 * @see CommandProc
+	 * @return the cost
+	 */
+	static Tret Do(DoCommandFlag flags, Targs... args)
+	{
+		if constexpr (std::is_same_v<TileIndex, std::tuple_element_t<0, std::tuple<Targs...>>>) {
+			/* Do not even think about executing out-of-bounds tile-commands. */
+			TileIndex tile = std::get<0>(std::make_tuple(args...));
+			if (tile != 0 && (tile >= MapSize() || (!IsValidTile(tile) && (flags & DC_ALL_TILES) == 0))) return MakeResult(CMD_ERROR);
```
rubidium42, line 155: I `CmdLevelLand` there is a check for a tile in formerly in p1 where it returns `{ CMD_ERROR, 0, INVALID_TILE }`. This would be returning `{ CMD_ERROR, 0, 0 }`? Is that going to be a problem?

src/misc_cmd.h
michicc: `TCPConnecter::Status` and  `IndustryDirectoryWindow::SorterType` use upper case, everything else uses normal cases as far as I can see.

src/command.cpp
michicc: Yes, but this is for the `DoCommandHelper::Post/Execute` code path and is not called for `Do`.

src/road_cmd.cpp
michicc: `DoCommandHelper::Do` only checks the first parameter (i.e. what used to be `tile`), but not anything that used to be in p1/p2, as that is the current behaviour. I thought about checking all TileIndex parameters, but I'd have to verify if any command does a different TileIndex validation.

src/command_func.h
michicc: Well, it would cause a wrong red error highlight square, but I think it is safe to assume that any invalid tile would either be a straight up bug in OTTD or somebody sending faked commands over the network. And in that case I don't think the red error square is that important.

src/vehicle_func.h
michicc: Ignoring for this one (also the other GetCmd... things).

src/tree_cmd.cpp
michicc: See https://github.com/OpenTTD/OpenTTD/pull/9725#discussion_r762554773

src/road_cmd.cpp
rubidium42: Okay. Lets keep those things out of this PR then ;)
I've also had other outlandish(?) ideas. We are passing Vehicle/Town/...IDs and almost always `T obj = T::GetIfValid(id); if (obj == nullptr) return CMD_ERROR`, so I was thinking about putting that resolving to object (and validation whether `nullptr` is allowed) into some "annotation" (with C++ templating magic), That way most of the common code for resolving and checking (also with enums) can be moved out of the Cmd* functions, making it less likely that mistakes are made with the checks for data that comes from the network.
Though that is far beyond the scope of this PR. However, validating of `TileIndex` values falls somewhere in the same category.

src/command.cpp
```diff
@@ -24,187 +24,69 @@
...
-CommandProc CmdChangeTimetable;
-CommandProc CmdSetVehicleOnTime;
-CommandProc CmdAutofillTimetable;
-CommandProc CmdSetTimetableStart;
-
-CommandProc CmdOpenCloseAirport;
-
-#define DEF_CMD(proc, flags, type) {proc, #proc, (CommandFlags)flags, type}
+template<typename T, T... i>
+inline constexpr auto MakeCommandsFromTraits(std::integer_sequence<T, i...>) noexcept {
```
rubidium42, line 87: `{` on newline

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+	{
+		Tret ret{};
+		ExtractCommandCost(ret) = cost;
+		return ret;
+	}
+
+public:
+	/**
+	 * This function executes a given command with the parameters from the #CommandProc parameter list.
+	 * Depending on the flags parameter it execute or test a command.
```
rubidium42: `executes or tests`

rubidium42: I think it is useful to also mention something in this comment like the following. I am not 100% sure it's right, but it's mostly to steer people away from this function in unintended situations.
```
This function is to be called from the StateGameLoop or from within the execution of a Command.
This function must not be called from the context of a "player" (real person, AI, game script).
Use ::Post for commands directly triggered by "players".
```

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+ * but converting such pointer back to pointer to the original function type yields
+ * the pointer to the original function." */
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wcast-function-type"
+#	define SILENCE_GCC_FUNCTION_POINTER_CAST
+#endif
+
+template<Commands TCmd, typename T, bool THasTile> struct DoCommandHelper;
+
+class DoCommandHelperBase {
```
rubidium42: Should this lose its `Do` for consistency's sake? Given we gone from `DoCommand<>::Do` -> `Command<>::Do`
Similar for the `DoCommandHelper` and also check for DoCommand in the comments below as I see many of them.

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+			if (data == INVALID_CLIENT_ID) data = CLIENT_ID_SERVER;
+		}
+	}
 
-/* group_gui.cpp */
-CommandCallback CcCreateGroup;
-CommandCallback CcAddVehicleNewGroup;
+	/** Set all invalid ClientID's to the proper value. */
+	template<class Ttuple, size_t... Tindices>
+	static inline void SetClientIds(Ttuple &values, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

src/command_func.h
```diff
@@ -64,60 +65,417 @@ static inline DoCommandFlag CommandFlagsToDCFlags(CommandFlags cmd_flags)
...
+		if constexpr (std::is_same_v<ClientID, T>) {
+			return data != INVALID_CLIENT_ID;
+		} else {
+			return true;
+		}
+	}
+
+	/** Check if all ClientID arguments are set to valid values. */
+	template<class Ttuple, size_t... Tindices>
+	static inline bool AllClientIdsSet(Ttuple &values, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

src/network/network_command.cpp
```diff
@@ -14,41 +14,151 @@
...
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
+
+/* Helpers to generate the callback table from the callback list. */
+
+inline constexpr size_t _callback_tuple_size = std::tuple_size_v<decltype(_callback_tuple)>;
+
+template <size_t... i>
+inline auto MakeCallbackTable(std::index_sequence<i...>) noexcept {
```
rubidium42: newline before `{`

src/network/network_command.cpp
```diff
@@ -14,41 +14,151 @@
...
+	using Tcallback = std::tuple_element_t<Tcb, decltype(_callback_tuple)>;
+	if constexpr (std::is_same_v<Tcallback, CommandCallback * const> || std::is_same_v<Tcallback, CommandCallbackData * const> || std::is_same_v<typename CommandTraits<Tcmd>::CbArgs, typename CallbackArgsHelper<Tcallback>::Args>) {
+		return &UnpackNetworkCommand<Tcmd, Tcb>;
+	} else {
+		return nullptr;
+	}
+}
+
+template <Commands Tcmd, size_t... i>
+constexpr UnpackDispatchT MakeUnpackNetworkCommand(std::index_sequence<i...>) noexcept {
```
rubidium42: newline before `{`

src/network/network_command.cpp
```diff
@@ -14,41 +14,151 @@
...
+	return UnpackDispatchT{{ {MakeUnpackNetworkCommandCallback<Tcmd, i>()}...}};
+}
+
+template <typename T, T... i, size_t... j>
+inline constexpr auto MakeDispatchTable(std::integer_sequence<T, i...>, std::index_sequence<j...>) noexcept {
```
rubidium42: newline before `{`

src/network/network_command.cpp
```diff
@@ -319,21 +439,97 @@ const char *NetworkGameSocketHandler::ReceiveCommand(Packet *p, CommandPacket *c
...
+static inline void SetClientIdHelper(T &data, [[maybe_unused]] ClientID client_id)
+{
+	if constexpr (std::is_same_v<ClientID, T>) {
+		data = client_id;
+	}
+}
+
+/** Set all invalid ClientID's to the proper value. */
+template<class Ttuple, size_t... Tindices>
+static inline void SetClientIds(Ttuple &values, ClientID client_id, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

src/network/network_command.cpp
```diff
@@ -319,21 +439,97 @@ const char *NetworkGameSocketHandler::ReceiveCommand(Packet *p, CommandPacket *c
...
+static inline void SanitizeSingleStringHelper([[maybe_unused]] CommandFlags cmd_flags, T &data)
+{
+	if constexpr (std::is_same_v<std::string, T>) {
+		data = StrMakeValid(data.substr(0, NETWORK_COMPANY_NAME_LENGTH), (!_network_server && cmd_flags & CMD_STR_CTRL) != 0 ? SVS_ALLOW_CONTROL_CODE | SVS_REPLACE_WITH_QUESTION_MARK : SVS_REPLACE_WITH_QUESTION_MARK);
+	}
+}
+
+/** Helper function to perform validation on command data strings. */
+template<class Ttuple, size_t... Tindices>
+static inline void SanitizeStringsHelper(CommandFlags cmd_flags, Ttuple &values, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

src/script/api/script_object.hpp
```diff
@@ -263,41 +269,83 @@ friend class ScriptController;
...
+		if constexpr (std::is_same_v<std::string, T>) {
+			/* The string must be valid, i.e. not contain special codes. Since some
+			 * can be made with GSText, make sure the control codes are removed. */
+			data = ::StrMakeValid(data, SVS_NONE);
+		}
+	}
+
+	/** Helper function to perform validation on command data strings. */
+	template<class Ttuple, size_t... Tindices>
+	static inline void SanitizeStringsHelper(Ttuple &values, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

src/script/api/script_object.hpp
```diff
@@ -263,41 +269,83 @@ friend class ScriptController;
...
+	static inline void SetClientIdHelper(T &data)
+	{
+		if constexpr (std::is_same_v<ClientID, T>) {
+			if (data == INVALID_CLIENT_ID) data = (ClientID)UINT32_MAX;
+		}
+	}
+
+	/** Set all invalid ClientID's to the proper value. */
+	template<class Ttuple, size_t... Tindices>
+	static inline void SetClientIds(Ttuple &values, std::index_sequence<Tindices...>) {
```
rubidium42: newline before `{`

rubidium42 (approved):
Lets hope the guaranteed bugs are found quickly; I can't find any more.

ldpl:
I wanted to wait until I had a chance to review it properly before wringing angry posts but you blink twice and it's already merged. 

I still didn't really look much into it so maybe some of it is not that bad or at least salvageable but from what I read in the PR description this is a disaster and should be reverted asap. As it only focuses on code improvements and completely ignores the big picture. Where is any consideration for the actual use cases? Where is any analysis of potential features of the command could use? It's been a constant nuisance for many years, I've lost count of how much stuff could've benefited from a well-thought command system. Highlights, gamescripts, game recording, admin port. And all I see in this PR is just templates are cool, templates go brrrr. Even something as simple as its impact on the bandwidth usage in multiplayer is nowhere to be seen. Or some mention of that admin port shit it does. I get that breaking admin api is already commonplace but what is admin client even gonna do with this "data"? It was already bad enough previously, but at least there was some structure to it.

So yeah, changes like this make me want to just quit doing anything OpenTTD-related. You spend years trying to make the best with what you have and then you have to rewrite half of your code every time some core devs decide it's a good idea to disregard any analysis and foresight and break everything for some little gimmick.

rubidium42:
I agree it's definitely not trivial to get your head around without (prior) knowledge, and I'm sure there are things in the documentation that can be improved upon to make things clearer.

Having said that, I think that this PR is looking at the big picture of the commands within OpenTTD. The foremost two/three issues that this PR tackles is having a fixed size of 2x32 bits (+ 1 string + 1 tile index) for commands, therefor untyped parameters and a lack of flexibility with commands.

Lets take CmdPause as an example. Previously it looked like: `CommandCost CmdPause(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)` and now it looks like `CommandCost CmdPause(DoCommandFlag flags, PauseMode mode, bool pause)`. So, the tile and the text have gone and p1 and p2 have become typed. In the network packet you had 3x4 bytes + 1 byte for the text (nul-termination). Now these 13 bytes are replaced by 2 bytes.
For CmdIncreaseLoan this looked like: `CommandCost CmdIncreaseLoan(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)` and now becomes `CommandCost CmdIncreaseLoan(DoCommandFlag flags, LoanCommand cmd, Money amount)`. The old CmdIncreaseLoan's comment also mentioned: `* @param p2 (bit 2-31) - lower half of amount (lower 2 bits assumed to be 0)` so it did some bit stuffing. Why was this done this way in the past? Because of the reservation of adding another 4 bytes (i.e. p3) so we can split the 8-byte `Money` into two 4 bytes values and have space for the (sub) command. In the new way, we do not perform bitstuffing anymore. We just pass a single `Money` parameter and everything is handled under water. Now the command parameters have more space, but actually use 4 bytes fewer than before for this command.
I have to confess, that it does not always reduce the number bytes used. Take for example CmdBuildSingleSignal that goes from `CommandCost CmdBuildSingleSignal(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)` to `CommandCost CmdBuildSingleSignal(DoCommandFlag flags, TileIndex tile, Track track, SignalType sigtype, SignalVariant sigvar, bool convert_signal, bool skip_existing_signals, bool ctrl_pressed, SignalType cycle_start, SignalType cycle_stop, uint8 num_dir_cycle, byte signals_copy)`. The new variant uses 1 byte more (although with extra logic in the `EndianBufferWriter` and `EndianBufferReader` we could reduce the number of bytes by doing bitstuffing under the hood, but that was not scope of this PR).

Due to the "unlimited" amount of parameters you can now pass, there is no need to perform bitstuffing before calling a command and undoing the bitstuffing within the command. You just pass the right values into the parameters. Now, if someone changes the parameters of the function, the compilation for the callers will break instead so you can go to those locations to pass the right information, instead of hoping that you fixed all callers.
Or when we do something like increasing the number of `StationClassID`s and we require the type to become larger, we can just make the type larger and all commands will automatically do the right thing. With the old command system we had to look through all the code to see where a `StationClassID` would be bitstuffed, then we had to change the bitstuffing so the extra byte becomes available. The result: all callers of the commands that use `StationClassID` need to be identified, and if you have some other code that calls it that is not in main, we would not update it and that command call breaks. With the new code, you would not even notice the `StationClassID`'s size being made larger.

That alone gives a lot more flexibility, but there is more. You are not limited to returning "only" a `CommandCost` anymore. Take for example CmdCloneVehicle. This was `CommandCost CmdCloneVehicle(TileIndex tile, DoCommandFlag flags, uint32 p1, uint32 p2, const std::string &text)` and became `std::tuple<CommandCost, VehicleID> CmdCloneVehicle(DoCommandFlag flags, TileIndex tile, VehicleID veh_id, bool share_orders)`. With this extra flexibility of returning more than just a `CommandCost` we can get rid of using a global variable to pass around information between a Command and its callback function.

I do agree that the command logging for the admin port could use some love. We have to check what to do there, but it was already a random (meaningless) blob of data that could be changed at any time. I wonder whether we might now be able to at least give some meaning to the packet; maybe the actual name of the command, and per parameter the type and value in some more useful format. Then an user of the admin logging can at least gain some more knowledge out of the data which would be more stable than it was before.

ldpl:
Well, ok, having more data is useful in the long run, so counts as a part of "big picture" I guess. And I never doubted this change is good for refactoring the current code. But at this point, I'd expect any rewrite of the command system to at least achieve that much. What I'm more concerned about is that the way it's done here specifically hinders two of the most essential (and I'll elaborate on it more later) imo properties for the further development of the game itself and supporting code that works with it: the ability to store and analyze commands. There were already some examples for the command storage in the code previously but I guess they weren't significant enough, so here are some potential features that could've used it: 

1) Complex building tool highlights. Every building action requires some kind of preview/indication before the command can be actually executed, so that current action needs to be stored somehow. Right now previews are very simple and are handled by HighLightStyle and TileHighlightData. And while those are already a mess if you ever want a wsywg preview like cmclient does you need to store even more. If you look at cmclient there is a ridiculous construct of `HighLightStyle->citymania::ObjectHighlight->CommandContainer->citymania::ObjectTileHighlight`.
But all those user actions have a matching command so if commands can be stored there is no need for a duplicating structure and it can be directly simplified to `CommandContainer->citymania::ObjectTileHighlight`. With an added benefit of command being able to report it's detailed actions if needed (see (2)).
Screenshot from 2021-02-17 12-48-18

2) It's a must for good gameplay with server/gamespript to analyze and interfere with the command flow. A classic example would be stopping other companies from destroying your claimed town in citybuilder. And while having some kind of callback system could be a better approach to this than analyzing command packets it doesn't seem to be possible to ever get callbacks in gamescripts, let alone admin clients. So analyzing command structure or packet seems to be the only option for them to get any information on user actions, especially before their execution.

3) Command logging for either debug, game recording and/or server moderation purposes. Game recording/replay is probably not affected by this change but debug and moderation are as they require admin and/or developer to read the command output and having just a data blob there isn't helping the cause. Though it looks like this can be solved relatively easily by sending some kind of command schema along with the command names in admin api. This should also cover other cases that require some analytics on command internals. Like, while you already explained how commands are packed I still have no idea how would that impact the bandwith of a regular game and the size of game recording. If I had a schema I could easily calculate that by using the recordings of old games.

4) Ability to use arbitrary command callbacks. There was (and afaict still is) no way to reliably identify a specific command after it's made the round trip over the network. So any callbacks are identified by a pre-defined table and are validated to match it. And not only it is a big issue for any network-compatible patch like cmclient/cmserver that can't change callback table but it also prevents gamescripts from sending more than one command per tick (or maybe even 2 ticks, don't remember). And that's a huge handicap for gamescripts. For example, updating goals and story book for 15 companies can take several minutes.

Also, I realize that a single PR doesn't have to necessarily solve all the issues but it would be nice to see an acknowledgment of their existence and some plan or thoughts for solutions. Just so that it wouldn't require 4 more rewrites of the command system that only core devs could realistically implement.

nielsmh:
@ldpl Regarding your final point, the storybook and goal commands in specific would probably now benefit from a big pass to perhaps change them fundamentally. For example, allow them to send bulk updates so a single command could create an entire storybook, or update all the goals for a company. That kind of change would be out of scope for this PR, but I'm quite sure it would be worth the work to do.

ldpl:
@nielsmh Technically there is not much difference between sending a list of changes in one command and sending a list of commands with changes. Except you'll need to add complex structures to DoCommand templates, implement new GS api methods, etc. And, tbh, no matter how you wrap it up updating UI from the server side is going to suck no matter what. So this overcomplication will ultimately be for nothing. Also, it's not like UI is the only thing that can benefit from the command spam, it's just the most obvious one.

nielsmh:
You mention the update rate as a specific problem, due to the current limit of one simple update per command. If the updates could be batched into a single command, the updates that currently may take several real-time minutes could perhaps be reduced to take 32 game ticks.

ldpl:
You'll have to do batching for each command for every specific usecase instead of solvng the issue once and for all. Slow story book - batch story book, slow goals - batch goals, slow town updates - batch town text, batch industry text, batch object placement, etc, etc, etc. Every time someone comes up with something new you'll have to make a batched command.

michicc:
I can see why batching would make sense, but what does it have to do with how commands are handled in the source code?

Currently scripts are suspended after each issued command because commands are delayed at least one tick in networked games.  So the problem with any kind of command batching isn't any templated code, but that you'd have to completely redesign the script API to make command execution asynchronous from the scripts POV. This is a problem of a totally different layer than the command layer changed by this PR.

ldpl:
@michicc yes, that problem is of a wider scope than just command layer, but it's related to it nonetheless. If there is no way to identify commands on the network it will be impossible to do asynchronous execution no matter what you do to the script api. So it's something that, at the very least, should be considered when designing the command layer, otherwise you may end up in a situation where for someone to make a progress on that script api redesign he'll have to completely redo your command layer first.

PR #9732, opened 2021-12-05
Feature: Added hotkeys to tooltips and dropdown menu's

Kuhnovic:
## Motivation / Problem

OpenTTD is full of useful hotkeys, but these are never shown to the user. Having these hotkeys visible makes it much easier to learn about them.

## Description

image
Tooltips

image
Dropdown menu's

The hotkeys have been added to:
- Main toolbar
- Dropdowns of main toolbar
- Transparancy window
- Order window
- All other toolbars that have hotkeys: rail toolbar, road toolbar, etc
- Scenario editor
- Probably some other places that I forgot...

## Limitations

- The widget system and the hotkey system are completely independent, and making an association between the two was much harder than I anticipated. I eventually got things working but the solution feels a little convoluted in certain areas. If anyone has ideas for a more straightforward solution, feel free to start the discussion.
- I had to use some platform specific voodoo to fetch the character of the "toggle console" hotkey, `~ on a US keyboard. Since I only have access to a Windows machine, I could not write an implementation for other platforms (at least not one I can test).
- For some hotkeys the text becomes a little ambigious, see the image below. Which hotkey does F3 refer to? I might be able to solve this with a string control command and parameters, but I haven't gone down that rabbit hole just yet.

image

## Checklist for review

### DIFF
--- a/src/hotkeys.cpp
+++ b/src/hotkeys.cpp
@@ -7,6 +7,8 @@
 
 /** @file hotkeys.cpp Implementation of hotkey related functions */
 
+#include <sstream>
+
 #include "stdafx.h"
 #include "openttd.h"
 #include "hotkeys.h"
@@ -18,6 +20,23 @@
 
 std::string _hotkeys_file;
 
+/*
+ * Returns the actual character on the backquote button (` on US keyboard)
+ * TODO implement for other platforms.
+ */
+#ifdef _WIN32
+#include <Windows.h>
+const char* GetBackquoteButtonCharacter()
+{
+	std::stringstream ss;
+	ss << static_cast<char>(MapVirtualKey(VK_OEM_3, MAPVK_VK_TO_CHAR));
+	static std::string backquote = ss.str();
+	return backquote.c_str();
+}
+#else
+const char* GetBackquoteButtonCharacter() { return "`"; }
+#endif
+
 /**
  * List of all HotkeyLists.
  * This is a pointer to ensure initialisation order with the various static HotkeyList instances.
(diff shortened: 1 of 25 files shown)


src/hotkeys.cpp
```diff
@@ -26,27 +45,36 @@ static std::vector<HotkeyList*> *_hotkey_lists = nullptr;
...
 };
 
 /** Array of non-standard keycodes that can be used in the hotkeys config file. */
 static const KeycodeNames _keycode_to_name[] = {
-	{"SHIFT", WKC_SHIFT},
-	{"CTRL", WKC_CTRL},
-	{"ALT", WKC_ALT},
+	{"SHIFT", WKC_SHIFT, "Shift"},
+	{"CTRL", WKC_CTRL, "Ctrl"},
+	{"ALT", WKC_ALT, "Alt"},
```
nielsmh, line 64: These need to be translatable strings. For example, in Germany the Ctrl key is called *Strg*. Also keep in mind that many hotkeys on Mac systems use the *Cmd* key where other platforms use Ctrl.

src/hotkeys.cpp
```diff
@@ -215,6 +243,41 @@ static const char *KeycodeToString(uint16 keycode)
...
+	auto append = [&ss](auto text) {
+		if (ss.tellp() > 0) ss << "+";
+		ss << text;
+	};
+
+	///* WKC_GLOBAL_HOTKEY is ignored here */
+	if (keycode & WKC_SHIFT) append("Shift");
+	if (keycode & WKC_CTRL) append("Ctrl");
+	if (keycode & WKC_ALT) append("Alt");
+	if (keycode & WKC_META) append("Meta");
```
nielsmh, line 264: Needs to be translatable.

src/hotkeys.cpp
```diff
@@ -275,6 +338,28 @@ void Hotkey::AddKeycode(uint16 keycode)
...
+
+	std::stringstream ss;
+	ss << input;
+	if (!hotkey->keycodes.empty()) ss << " ";
+	for (uint16 keycode : hotkey->keycodes)
+	{
+		const bool is_global = keycode & WKC_GLOBAL_HOTKEY;
+		ss << (is_global ? "[" : "(");
+		ss << KeycodeToDisplayString(keycode);
+		ss << (is_global ? "]" : ")");
```
nielsmh, line 358: Make sure this results in meaningful results on right-to-left languages (mainly Arabic and Hebrew.)

LC-Zorg:
What would be most useful in theme of shortcuts is an option to select these shortcuts in the game interface. Then the average player wouldn't only find out what shortcuts exist, but could also change them. :) 
Regardless of the above, I think this could be a useful change, but I would have two remarks.

First of all (sorry, maybe it's obvious, but I'm almost blind when it comes to reading the code) are these shortcut tooltips permanent or depending on the chosen setting in the config file? Personally, I use completely revised shortcuts, and it would be weird to see hints that are unrelated to the facts.

The second note concerns the tooltips in the dropdown menu - I believe that they are redundant and reduce readability unnecessarily. Minimal, but they do decrease. Moreover, once someone knows about this shortcuts, this information becomes intrusive after time. In both visible cases (L / Shift + F11) these tooltips will be available after hovering over the icon. I think that is really enough. :)

Shortcut tooltips

src/hotkeys.cpp
Kuhnovic: As pointed out by dP in the development discord channel, I need to do quite a bit more work on the translating / formatting parts. I will take your comments into account as well.

Kuhnovic:
@LC-Zorg A hotkey configuration window would indeed be great, but that would be a PR on its own ;)

The hotkeys are read from hotkeys.cfg, so they will display whatever you set them to. 

The dropdown items can indeed be quite intrusive, especially if there are multiple key combination configured. I added them for completeness, but I'm not entirely sure if I want to keep them. For now I'll keep them in, I can always throw those part of the code away later.

image

Kuhnovic:
I'm giving up on this one, it's becoming a huge hassle for something that I initially thought would be relatively simple to do. I don't want to spend so much time that's going to a relatively small contribution to the overall gameplay.

PR #9739, opened 2021-12-10
Copy server invite code to the clipboard automatically

pelya:
## Motivation / Problem
When you create a server, you need to type the server invite code into email/chat letter by letter to send it to friends.

## Description
With this patch, the invite code is copied to the clipboard automatically when the server registers with the masterserver.

## Limitations
I am too lazy to add 'Copy to clipboard' button to the server info dialog, this is good enough for me.
This code works only with SDL2, other video backends will ignore the clipboard for now.

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2346,6 +2346,7 @@ STR_NETWORK_CLIENT_LIST_SERVER_INVITE_CODE                      :{BLACK}Invite c
 STR_NETWORK_CLIENT_LIST_SERVER_INVITE_CODE_TOOLTIP              :{BLACK}Invite code other players can use to join this server
 STR_NETWORK_CLIENT_LIST_SERVER_CONNECTION_TYPE                  :{BLACK}Connection type
 STR_NETWORK_CLIENT_LIST_SERVER_CONNECTION_TYPE_TOOLTIP          :{BLACK}Whether and how your server can be reached by others
+STR_NETWORK_CLIENT_LIST_SERVER_COPY_TO_CLIPBOARD                :{BLACK}Copy to clipboard
 STR_NETWORK_CLIENT_LIST_PLAYER                                  :{BLACK}Player
 STR_NETWORK_CLIENT_LIST_PLAYER_NAME                             :{BLACK}Name
 STR_NETWORK_CLIENT_LIST_PLAYER_NAME_TOOLTIP                     :{BLACK}Your player name
--- a/src/network/network_coordinator.cpp
+++ b/src/network/network_coordinator.cpp
@@ -13,7 +13,6 @@
 #include "../rev.h"
 #include "../settings_type.h"
 #include "../strings_func.h"
-#include "../video/video_driver.hpp"
 #include "../window_func.h"
 #include "../window_type.h"
 #include "network.h"
@@ -194,7 +193,6 @@ bool ClientNetworkCoordinatorSocketHandler::Receive_GC_REGISTER_ACK(Packet *p)
 	 * and _settings_client.network.server_invite_code contains the one we will
 	 * attempt to re-use when registering again. */
 	_network_server_invite_code = _settings_client.network.server_invite_code;
-	VideoDriver::GetInstance()->SetClipboardContents(_network_server_invite_code);
 
 	SetWindowDirty(WC_CLIENT_LIST, 0);
 
--- a/src/network/network_coordinator.cpp
+++ b/src/network/network_coordinator.cpp
@@ -13,6 +13,7 @@
 #include "../rev.h"
 #include "../settings_type.h"
 #include "../strings_func.h"
+#include "../video/video_driver.hpp"
 #include "../window_func.h"
 #include "../window_type.h"
 #include "network.h"
@@ -193,6 +194,7 @@ bool ClientNetworkCoordinatorSocketHandler::Receive_GC_REGISTER_ACK(Packet *p)
 	 * and _settings_client.network.server_invite_code contains the one we will
 	 * attempt to re-use when registering again. */
 	_network_server_invite_code = _settings_client.network.server_invite_code;
+	VideoDriver::GetInstance()->SetClipboardContents(_network_server_invite_code);
 
 	SetWindowDirty(WC_CLIENT_LIST, 0);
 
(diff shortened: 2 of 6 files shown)


2TallTyler:
I'm +1 to this feature, but not how it's hidden from anyone who hasn't seen this PR. 😃 

Personally I would prefer a `Copy to clipboard` button, although it would need to work with all video drivers.

Eilon:
Setting clipboard data is "destructive" in the sense that it clears out whatever the user may have already had in their clipboard, which the user may have deemed valuable. Also, as @2TallTyler mentions, how would anyone even know that this happened so that they could actually make use of it?

I think adding an explicit button would be a better design, even if it's more work.

nielsmh (dismissed):
You need to make a button for copying the code, that makes it clear that clicking this button will replace the user's clipboard contents.
Don't silently overwrite the user's clipboard.

pelya:
I've added a 'copy to clipboard' button to GUI

LC-Zorg:
Please, don't care too much about this. Perhaps my opinion will be parochial and overly conservative for you, but...
I don't understand what this invitation code play is for at all. For me, but I also think that for many other average player, it makes no sense at all. Just like showing MD5sum or GRF ID in the NewGRF information - this information is completely useless for the player and only creates questions that the player doesn't need to know the answers to. You want to simplify the interface, but adding a button to such an enigmatic function doesn't simplify it at all.
I see this as the same function as the server password now. The difference is that it is more difficult to remember, so much that you need to add another function, which is in this PR. In addition, this code is shown even where a password is not needed at all. I'm sorry, but I really don't understand this.

2TallTyler:
The invite code is for players to join their friends' games without having to list the game publicly. See #9434. But the value of this approach has nothing to do with this PR, which simply aims to make it easier to share the invite code.

pelya:
I thought invite code was a way to join servers behind NAT? It looks pretty
important to me.

nielsmh:
@LC-Zorg 
The invite code is the "address" of the game. When you want a friend to join your game, instead of having to figure out your IP address and set up port forwarding, you send the invite code to your friend. Your friend then enters the invite code, and gets connected to your game. (There is a blog post with more technical details if you're interested.)

That's why the invite code is important: It's the easy way to let others join your private game.

LC-Zorg:
Ok, I don't use it and I don't think it will be any different in the future for me, but I can understand there is a need for some to hide the server. However, I think there are at least one maybe two significant issues with this feature that make its functionality significantly limited, and in its current form, at least from my point of view, it is more problematic than useful. It is independent of this PR, so I will write about it elsewhere. (#9746) (#9747)

If I understand the code correctly, this copy button is in the client list window in the server settings field. If so, I think there is no problem and it is a good solution. :) Previously, I thought this might be for the server list window, and that wouldn't be good.

LordAro (commented):
Like others I am generally +1 to this feature, but I don't understand the current implementation

Why is `set_clipboard_text` stored at all? Why not just call `SDL_SetClipboardText` directly in `VideoDriver::SetClipboardContents` ?

pelya:
Because it must be called from the SDL main thread, not from the game
thread.

rubidium42 (approved):
When actually squashing/merging, maybe note that it is SDL2 for now?

One tricky thing from the user's point of view is that is a button that does nothing (unless they use SDL2). That might be considered  buggy behaviour, so does that make this patch bugged? 
And what would be the best way to remedy this? Implementing it for the other platforms as well in this PR? Or ship it buggy and hope the maintainers fix those bugs? Or introduce some manner where the UI can ask the video driver whether it supports writing to the clipboard and then disable the button when the video driver doesn't support it?

pelya:
Isn't SDL2 used for all release binaries that are published to openttd.org?
I guess we can surround this code with
`#if SDL_VERSION_ATLEAST(2,0,0)`

glx22:
SDL2 is used only on linux builds.

pelya:
Eh okay. SDL2 is designed to abstract away Linux/Windows/MacOS differences,
it seems counter-intuitive that OpenTTD is not using this feature.

Well, we can copypaste SDL2 implementation at least.

https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowsclipboard.c#L51

https://github.com/libsdl-org/SDL/blob/main/src/video/cocoa/SDL_cocoaclipboard.m#L29

2TallTyler:
I don't see a point shipping a build with a button that doesn't work for the majority of users. (I imagine Linux will be more prevalent among OpenTTD users than the average, but W3 Schools says [it's used by](https://www.w3schools.com/browsers/browsers_os.asp) about 4.3% of users.)

pelya:
Yes, releasing the feature only for Linux makes little sense.
I'm not using Windows or MacOS, I can only point to SDL2 clipboard
implementation. Someone else will need to fill in the missing code in this
patch.

nielsmh:
I was about to say that the Win32 and Cocoa video drivers also support setting the clipboard, but I was  mixing that up with the _getting_ clipboard function, which is implemented in the OS code and not video driver.

https://github.com/OpenTTD/OpenTTD/blob/7f0efbe00e3c100319a1633deb1d6bc9eeb8e86b/src/os/windows/win32.cpp#L511-L531
https://github.com/OpenTTD/OpenTTD/blob/7f0efbe00e3c100319a1633deb1d6bc9eeb8e86b/src/os/macosx/macos.mm#L191-L206
https://github.com/OpenTTD/OpenTTD/blob/7f0efbe00e3c100319a1633deb1d6bc9eeb8e86b/src/os/unix/unix.cpp#L271-L289

I think it would make the most sense to move this code to the video driver (because the concept of a clipboard is usually quite tightly bound to the "exterior" windowing system used), and then of course implement setting the clipboard for all platforms as well.

LordAro (changes requested):
No issues with this at all, but it needs to support more than just SDL2 before being merged

2TallTyler:
As @pelya [posted on TT-Forums](https://www.tt-forums.net/viewtopic.php?p=1252352#p1252352) a month ago, he will likely not be working on this any time in the near future.

2TallTyler:
This is a great feature but the PR can't be merged without supporting all OSs. Since @pelya has indicated that he's not willing/able to add that support, I'm closing the PR. Feel free to reopen if you disagree or have code that makes it work for other OSs. 🙂

PR #9744, opened 2021-12-12
Fix: if vehicles only refit to cargo-slots >= 32, the default cargo was wrong.

frosch123:
## Motivation / Problem

If vehicles are only refittable to cargo-slots >= 32, their default cargo would say "passengers".

## Description

In this case the default cargo uses the first refittable-cargo.
The refit-mask is a uint64, but FindFirstBit truncated that to uint32.
Apprently if "no bit is set", FindFirstBit returns 0 (= passengers), instead of asserting.

## Checklist for review

### DIFF
--- a/src/core/bitmath_func.cpp
+++ b/src/core/bitmath_func.cpp
@@ -24,7 +24,7 @@ const uint8 _ffb_64[64] = {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


glx22 (approved):
LGTM

Issue #9167, opened 2021-05-01
Linkgraphs don't show direction of each link


2TallTyler:
## Version of OpenTTD
1.11.1 and earlier

## Expected result
Linkgraphs like the Cargo Flow Legend overlay or the smallmap linkgraph, clearly indicate the direction of each link, perhaps with an arrow of some type.

## Actual result
![rhd](https://user-images.githubusercontent.com/55058389/116788591-7e7f4700-aa78-11eb-94fc-259bc03475e6.png)

Which colour is passengers from Treninghall to Martston, and which colour is the return trip?

(Answer: it matches the side road vehicles drive on, but this isn't indicated anywhere.)

This problem exists for both the overlay and smallmap linkgraphs.

## Steps to reproduce

1. Start a game with cargo distribution enabled for passengers (symmetric is fine). 
2. Connect two towns with a bus route.
3. Open the Cargo Flow Legend from the map icon in the main toolbar, and enable Passengers for your company.

glx22:
I tried something, but I'm not sure it's a good idea.
![Ninfingley Transport, 24 Jun 1960](https://user-images.githubusercontent.com/2952192/116790198-5bbe5400-aab3-11eb-87ac-f8e90c1c9a96.png)

2TallTyler:
I like where you're going, but it's still somewhat ambiguous — could be pointed either direction as an arrowhead or a comet tail.

glx22:
Another try.
![Ninfingley Transport, 16 Avr 1960](https://user-images.githubusercontent.com/2952192/116790695-02a3ef80-aab6-11eb-8a08-38fe1c89c93e.png)

2TallTyler:
Better, but it's very subtle. Perhaps the arrows could be shorter, maybe even a fixed length instead of half the link distance?

glx22:
Fixed length is complicated, in this one the arrow length is 1/6th of link distance with an empty head.
Anyway not being able to zoom in the minimap doesn't really help to see the arrow, same goes for the green on green.
![Ninfingley Transport, 1er Déc 1960](https://user-images.githubusercontent.com/2952192/116797606-b2de1c00-aae7-11eb-9946-ef0f2ecb6081.png)

James103:
> Anyway not being able to zoom in the minimap doesn't really help to see the arrow, same goes for the green on green.

Ignore the reference above, the actual issue that this is referring to is #8998.

2TallTyler:
[...]

PR #9760, opened 2021-12-22
Feature: [Linkgraph] Show a tooltip with statistics when hovering a link

nchappe:
## Motivation / Problem

Cargodist considers many parameters to determine how cargo should be routed.
These parameters are not all exposed to the player:
the linkgraph overlay shows the saturation of links, but there is nothing
for the capacities, or for the travel times.

## Description

This PR shows a tooltip when a link of the overlay is hovered.
The tooltip contains the source and destination stations, the link capacity, its usage and its average travel time.
For bidirectional links, data are shown for both directions of travel.

I believe this could be useful for various reasons:
- The station names help differentiate close but distinct links, and clarify which color corresponds to each direction, mitigating #9167.
- The link usage is expressed as a percentage, which is more precise than the color code. It mitigates #9168 too.
- When e.g. adding a vehicle to a saturated route, it is interesting to see how the capacity of the link evolves.
- Travel times can help players update their timetables after replacing vehicles.
- It can help debugging linkgraph code (that was actually my motivation to write the first version of this patch a few months ago)

### Corner cases

- The tooltip is only shown if only one cargo type is selected. Showing information for all cargo types would need a huge tooltip.
- For the same reason, in case several links cross, the tooltip only shows information for one of them.
- As all cargo types are always shown in the minimap, it does not benefit from this feature.

### Limitations

I had to declare a C-like string as static because TooltipsWindow segfaults if its given string has a temporary RAW_STRING argument that does not outlive the TooltipsWindow.

A possible alternative would be to add a GetString call in the constructor of TooltipsWindow and store the resulting string once and for all in a buffer. This is how JGR's patch pack solves the same problem.

### Alternatives

A window would be more flexible than a tooltip, but it would raise many questions about its layout, etc.
I preferred to keep this PR relatively simple.

## Checklist for review

### DIFF
--- a/src/linkgraph/linkgraph_gui.cpp
+++ b/src/linkgraph/linkgraph_gui.cpp
@@ -218,8 +218,10 @@ void LinkGraphOverlay::AddLinks(const Station *from, const Station *to)
 		const LinkGraph &lg = *LinkGraph::Get(ge.link_graph);
 		ConstEdge edge = lg[ge.node][to->goods[c].node];
 		if (edge.Capacity() > 0) {
-			this->AddStats(lg.Monthly(edge.Capacity()), lg.Monthly(edge.Usage()),
-					ge.flows.GetFlowVia(to->index), from->owner == OWNER_NONE || to->owner == OWNER_NONE,
+			this->AddStats(c, lg.Monthly(edge.Capacity()), lg.Monthly(edge.Usage()),
+					ge.flows.GetFlowVia(to->index),
+					edge.TravelTime() / DAY_TICKS,
+					from->owner == OWNER_NONE || to->owner == OWNER_NONE,
 					this->cached_links[from->index][to->index]);
 		}
 	}
@@ -235,14 +237,16 @@ void LinkGraphOverlay::AddLinks(const Station *from, const Station *to)
  * @param new_shared If the new link is shared.
  * @param cargo LinkProperties to write the information to.
  */
-/* static */ void LinkGraphOverlay::AddStats(uint new_cap, uint new_usg, uint new_plan, bool new_shared, LinkProperties &cargo)
+/* static */ void LinkGraphOverlay::AddStats(CargoID new_cargo, uint new_cap, uint new_usg, uint new_plan, uint32 time, bool new_shared, LinkProperties &cargo)
 {
 	/* multiply the numbers by 32 in order to avoid comparing to 0 too often. */
 	if (cargo.capacity == 0 ||
-			std::max(cargo.usage, cargo.planned) * 32 / (cargo.capacity + 1) < std::max(new_usg, new_plan) * 32 / (new_cap + 1)) {
+			cargo.Usage() * 32 / (cargo.capacity + 1) < std::max(new_usg, new_plan) * 32 / (new_cap + 1)) {
+		cargo.cargo = new_cargo;
 		cargo.capacity = new_cap;
 		cargo.usage = new_usg;
 		cargo.planned = new_plan;
+		cargo.time = time;
 	}
 	if (new_shared) cargo.shared = true;
 }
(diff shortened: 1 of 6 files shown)


src/linkgraph/linkgraph_gui.cpp
```diff
@@ -333,6 +336,54 @@ void LinkGraphOverlay::DrawStationDots(const DrawPixelInfo *dpi) const
 	GfxDrawLine(x + w2, y - w1, x + w2, y + w2, border_colour);
 }
 
+bool LinkGraphOverlay::ShowTooltip(Point pt)
+{
+	if (!HasExactlyOneBit(this->cargo_mask))
+		return false;
```
nielsmh:
```suggestion
	if (!HasExactlyOneBit(this->cargo_mask)) return false;
```

src/linkgraph/linkgraph_gui.h
```diff
@@ -24,9 +24,13 @@
 struct LinkProperties {
 	LinkProperties() : capacity(0), usage(0), planned(0), shared(false) {}
 
+	/** Return the usage of the link to display. */
+	uint Usage() const { return std::max(usage, planned); }
```
nielsmh:
```suggestion
	uint Usage() const { return std::max(this->usage, this->planned); }
```

nielsmh (commented):
Two things I'm worried about (having not tested the code):

Is the LinkGraphOverlay::ShowTooltip sufficiently fast on big networks?

Does this even work when the tooltip delay is zero so you need to hold RMB to show them?

nchappe:
Thanks for the feedback!
I fixed the coding style and RMB tooltips. For this one I had to change a bit the mouse handling logic of window.cpp, I'm not entirely sure what I did is valid so I put it in a separate commit.

On the performance of `LinkGraphOverlay::ShowTooltip`, its complexity is linear in the number of links for the selected cargo type.
I agree this is a bit much for large networks, so now hover events are only generated every `_settings_client.gui.hover_delay_ms` when the cursor is static (they were generated at each frame before). With the default setting of 250ms, `ShowTooltip` will be called at most 4 times per second.

2TallTyler:
Might we get a `Preview` label to make testing this easier?

nielsmh:
Examples of the tooltip:
image
image

It appears to only work when you have exactly one cargo type selected in the cargo flow legend, if I select more than one cargo type the tooltips won't show up.

nchappe:
> It appears to only work when you have exactly one cargo type selected in the cargo flow legend, if I select more than one cargo type the tooltips won't show up.

I just pushed a new version that displays statistics for *the most saturated one* when a link has several cargo types. This is already how the color to use for a link is determined, so the two will always match.

2TallTyler:
In my opinion, this and #9866 close #9167 and #9168.

One point of minor confusion for me regarding links which are over capacity — e.g. `219 passengers transported back (547% of capacity) in 10 days`.

To me, this reads like 219 passengers is 547% of capacity. Obviously it's not, but perhaps there's a better way to word this. What about `219 passengers transported back in 10 days (547% of capacity)`?

nchappe:
219 passengers/month *is* 547% of capacity. Even without this PR, all dark red links in the overlay correspond to links with a 200%+ usage/capacity ratio.

The "capacity" of a link is only a rough estimate, based on the capacity of the vehicles that took the link in the past few months.
As for the usage of a link, it is also approximative, and it's a bit more complex as the value used is actually max(planned, usage), with 'planned' an (even rougher) estimate of the future demand based on the number of passengers waiting for this link.

By construction, the estimated usage of a link never exceeds 100% of its capacity, but the demand can, thus the occasional >100% usage ratios. This is suboptimal, but reworking the formulae to be more accurate is definitely out of scope for this PR. I may give it a try at some point in the future though :)

In any case, I realize the tooltip text is a bit confusing as is.
I read your proposed wording as '219 passengers crossed this link the last 10 days' when the intended meaning of the tooltip is actually 'about 219 passengers cross this link each month, and the journey takes them 10 days on average'.
Maybe `219 passengers transported back per month (547% of capacity) in an average of 10 days` would be less ambiguous? I'm open to alternative wording suggestions!

Moth-Tolias:
it looks like there are two different stats trying to be condensed into a single line here.
how does this sound?
`219 passengers (547% of current capacity) to be transported back per month`
`average return journey time: 10 days`

that stops it from sounding like more than 100% capacity is being transported, and also helps prevent capacity and time from being confused.

nchappe:
Thanks for the suggestions. I slightly reworded the tooltip:

```
X passengers to be transported per month from A to B (x% of capacity)
Y passengers to be transported back (y% of capacity)
Average travel time: n days
```

The travel time is now on a separate line, and is an average for the two directions of travel.

2TallTyler:
The `to be transported` helps a lot, indicating demand versus actual stats.

michicc (approved):

PR #9761, opened 2021-12-22
Change: add a timestamp in name of crash files

glx22:
## Motivation / Problem
When a crash happens, it overwrites previous crash files.
But different crash may happen before a player decides to open an issue, and they can have very different causes.

## Description
Add a timestamp to the filenames, so newer crashes don't overwrite maybe useful data from previous crashes.

## Checklist for review

### DIFF
--- a/src/crashlog.cpp
+++ b/src/crashlog.cpp
@@ -325,6 +325,25 @@ char *CrashLog::LogRecentNews(char *buffer, const char *last) const
 }
 
 /**
+ * Create a timestamped filename.
+ * @param filename      The begin where to write at.
+ * @param filename_last The last position in the buffer to write to.
+ * @param ext           The extension for the filename.
+ * @param with_dir      Whether to prepend the filename with the personal directory.
+ * @return the number of added characters.
+ */
+int CrashLog::CreateFileName(char *filename, const char *filename_last, const char *ext, bool with_dir) const
+{
+	static std::string crashname;
+
+	if (crashname.empty()) {
+		UTCTime::Format(filename, filename_last, "crash%Y%m%d%H%M%S");
+		crashname = filename;
+	}
+	return seprintf(filename, filename_last, "%s%s%s", with_dir ? _personal_dir.c_str() : "", crashname.c_str(), ext);
+}
+
+/**
  * Fill the crash log buffer with all data of a crash log.
  * @param buffer The begin where to write at.
  * @param last   The last position in the buffer to write to.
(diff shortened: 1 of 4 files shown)


LordAro:
I think there's a windows error message referencing crash.log somewhere as well

Also, the issue template :)

glx22:
Windows crash dialog says something about `crash.dmp` and lists generated files with their real fullpath.
I could add `<timestamp>` in all `crash.xxx` in issue template, but I don't think it's necessary.

src/crashlog.cpp
```diff
@@ -324,6 +324,25 @@ char *CrashLog::LogRecentNews(char *buffer, const char *last) const
 	return buffer;
 }
 
+/**
+ * Create a timestamped filename.
+ * @param filename      The begin where to write at.
+ * @param filename_last The last position in the buffer to write to.
+ * @param ext           The extension for the filename.
+ * @param with_dir      Wether to append directory in front of filename.
```
rubidium42: s/wether/whether/
Maybe something like "Whether to suffix the file name with the directory"? Not sure whether "append ... in front" or "suffix" would be the prefered way to write it in English though.

src/crashlog.cpp
```diff
@@ -324,6 +324,25 @@ char *CrashLog::LogRecentNews(char *buffer, const char *last) const
...
+ * @return the number of added characters.
+ */
+int CrashLog::CreateFileName(char *filename, const char *filename_last, const char *ext, bool with_dir) const
+{
+	static std::string crashname;
+
+	if (crashname.empty()) {
+		UTCTime::Format(filename, filename_last, "crash%Y%m%d%H%M%S");
```
rubidium42, line 340: I wonder how likely it is for this to trigger an allocation that OOMs the crashlog when its being generated. I seem to remember we were fairly wary about memory allocation before at least the crash log and dump were written. Not sure how big a problem it is, because it's a very small amount of memory that gets allocated though.

JGRennison: On POSIX platforms at least, strictly speaking you should not call malloc or similar functions from a signal handler (including the crash log handler) because these are not re-entrant or signal-safe. OOM is not really an issue because if it happens it's more likely that the OOM killer will kill the whole process outright anyway. In the vast majority of cases you can get away with using malloc, and indeed the backtrace call implementations in the Unix crash log handler call malloc.
On Windows both writing the crash log and showing the dialog with the crash log contents in will allocate vastly more memory than this, so if an OOM event occurs any crash log will never make it to the user.

This string content here is short enough that it'll probably fit into the inline buffer anyway.

rubidium42 (approved):

PR #9768, opened 2021-12-24
Fix/incorrect company values

benda:
## Motivation / Problem

Company value calculation is missing the value of the company's shares in other companies

Discussed as issue 1 below:

https://github.com/OpenTTD/OpenTTD/discussions/9765

I view this as a defect.

## Description

The problem is solved by first calculating all of the company values without including value of shares in other companies, then adding in the value of the shares which now is known. I do not think this causes issues with bankruptcy or buying/selling shares; it does affect those of course since they now use the new company values.

Note that when buying shares, the company value does not change, since instead of holding the cash it now holds the shares of an equivalent value. 

Note that when selling shares, the company value does not change, since instead of holding the shares it now holds cash of an equivalent value. 

## Limitations

I think this PR includes a commit that was already merged (the first commit), I'm not sure how to remove that from this PR. 

I haven't tested multiplayer, this works fine on saved games in single player.

I am not a C++ programmer so my change should be thoroughly reviewed. 

## Checklist for review

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -69,7 +69,7 @@ that comes with vcpkg. After that, you can run something similar to this:
 ```powershell
 mkdir build
 cd build
-cmake.exe .. -G'Visual Studio 16 2019' -DCMAKE_TOOLCHAIN_FILE="<location of vcpkg>\vcpkg\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="x64-windows-static"
+cmake.exe .. -G"Visual Studio 16 2019" -DCMAKE_TOOLCHAIN_FILE="<location of vcpkg>\vcpkg\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="x64-windows-static"
 ```
 
 Change `<location of vcpkg>` to where you have installed vcpkg. After this
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -112,23 +112,47 @@ static PriceMultipliers _price_base_multiplier;
  */
 Money CalculateCompanyValue(const Company *c, bool including_loan)
 {
+	Money value = 0;
+
+	Money companyValues[8];
+
+	for (const Company* co : Company::Iterate()) {
+	  companyValues[co->index]	= CalculateCompanyValueExcludingShares(co);
+	}
+
+	for (const Company* co : Company::Iterate()) {
+
+		for (int i = 0; i < 4; i++)
+		{
+			if (co->share_owners[i] == c->index)
+			{
+				value += (companyValues[co->index] / 4);
+			}
+		}
+	}
+
+	return std::max<Money>(value + companyValues[c->index], 1);
+}
+
+Money CalculateCompanyValueExcludingShares(const Company* c, bool including_loan)
+{
 	Owner owner = c->index;
 
 	uint num = 0;
 
-	for (const Station *st : Station::Iterate()) {
+	for (const Station* st : Station::Iterate()) {
 		if (st->owner == owner) num += CountBits((byte)st->facilities);
 	}
 
 	Money value = num * _price[PR_STATION_VALUE] * 25;
 
-	for (const Vehicle *v : Vehicle::Iterate()) {
+	for (const Vehicle* v : Vehicle::Iterate()) {
 		if (v->owner != owner) continue;
 
 		if (v->type == VEH_TRAIN ||
-				v->type == VEH_ROAD ||
-				(v->type == VEH_AIRCRAFT && Aircraft::From(v)->IsNormalAircraft()) ||
-				v->type == VEH_SHIP) {
+			v->type == VEH_ROAD ||
+			(v->type == VEH_AIRCRAFT && Aircraft::From(v)->IsNormalAircraft()) ||
+			v->type == VEH_SHIP) {
 			value += v->value * 3 >> 1;
 		}
 	}
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -103,35 +103,32 @@ static PriceMultipliers _price_base_multiplier;
 
 /**
  * Calculate the value of the company. That is the value of all
- * assets (vehicles, stations, etc) and money minus the loan,
+ * assets (vehicles, stations, shares) and money minus the loan,
  * except when including_loan is \c false which is useful when
  * we want to calculate the value for bankruptcy.
- * @param c              the company to get the value of.
+ * @param c the company to get the value of.
  * @param including_loan include the loan in the company value.
  * @return the value of the company.
  */
 Money CalculateCompanyValue(const Company *c, bool including_loan)
 {
-	Money value = 0;
+	Money companyValue = 0;
 
-	Money companyValues[8];
+	Money companyValuesExcludingShares[MAX_COMPANIES];
 
 	for (const Company* co : Company::Iterate()) {
-	  companyValues[co->index]	= CalculateCompanyValueExcludingShares(co);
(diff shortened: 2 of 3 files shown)


src/economy.cpp
```diff
@@ -103,38 +103,61 @@ static PriceMultipliers _price_base_multiplier;
...
+	Money companyValuesExcludingShares[MAX_COMPANIES];
+
+	for (const Company* co : Company::Iterate()) {
+	  companyValuesExcludingShares[co->index]	= CalculateCompanyValueExcludingShares(co);
+	}
+
+	for (const Company* co : Company::Iterate()) {
+		for (int i = 0; i < 4; i++){
+			if (co->share_owners[i] == c->index) {
+				companyValue += (companyValuesExcludingShares[co->index] / 4);
```
ldpl: Why calculate and store company values that aren't even going to be used? CalculateCompanyValueExcludingShares can just be called here once for each relevant company.

benda: I wasn't sure how to get which shares the relevant company owns without iterating everything first. On this particular line though (126), CalculateCompanyValueExcludingShares will be called up to 3 times for each 25% share, so it would be  called multiple times for the same company, no ?

ldpl: they're all the same, you can just count and multiply

benda: OK, I committed with the implementation I think you are recommending.

src/economy.cpp
```diff
@@ -103,38 +103,65 @@ static PriceMultipliers _price_base_multiplier;
...
+}
+
+Money CalculateCompanyValueExcludingShares(const Company* c, bool including_loan)
 {
 	Owner owner = c->index;
 
 	uint num = 0;
 
-	for (const Station *st : Station::Iterate()) {
+	for (const Station* st : Station::Iterate()) {
```
2TallTyler: Pointers are written with the asterisk after the space, as in `Station *st`. This applies to several places in your PR. The way this alone is changed here, though, makes me wonder if your IDE is autocorrecting it to the wrong style (a problem I've experienced previously).

src/economy.cpp
```diff
@@ -103,38 +103,65 @@ static PriceMultipliers _price_base_multiplier;
...
+		sharesOwned = 0;
+
+		for (int i = 0; i < 4; i++){
+			if (co->share_owners[i] == c->index) {
+				sharesOwned++;
+			}
+		}
+
+		if (sharesOwned > 0)
+		{
```
2TallTyler: The opening curly brace for if statements stays on the same line.

```suggestion
		if (sharesOwned > 0) {
```

src/economy.cpp
```diff
@@ -103,38 +103,65 @@ static PriceMultipliers _price_base_multiplier;
...
  * except when including_loan is \c false which is useful when
  * we want to calculate the value for bankruptcy.
- * @param c              the company to get the value of.
+ * @param c the company to get the value of.
  * @param including_loan include the loan in the company value.
  * @return the value of the company.
  */
 Money CalculateCompanyValue(const Company *c, bool including_loan)
+{
+	Money companySharesValue = 0;
```
2TallTyler: Variables in OpenTTD are all lowercase with underscores. Also, a better name might be:
```suggestion
	Money owned_shares_value = 0;
```

src/economy.cpp
```diff
@@ -103,38 +103,65 @@ static PriceMultipliers _price_base_multiplier;
...
+
+	int sharesOwned = 0;
```
2TallTyler: The OpenTTD code base is fairly careful about right-sizing variables (no matter the "best practice" internet arguments about unsigned integers). This, and your `i` iterators, would fit in `uint8`, which has a range of 0-255.
```suggestion
	uint8 shares_owned = 0;
```

COMPILING.md
```diff
@@ -69,7 +69,7 @@ that comes with vcpkg. After that, you can run something similar to this:
 ```powershell
 mkdir build
 cd build
-cmake.exe .. -G'Visual Studio 16 2019' -DCMAKE_TOOLCHAIN_FILE="<location of vcpkg>\vcpkg\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="x64-windows-static"
+cmake.exe .. -G"Visual Studio 16 2019" -DCMAKE_TOOLCHAIN_FILE="<location of vcpkg>\vcpkg\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="x64-windows-static"
```
2TallTyler: Assuming you've set the upstream properly as described in [CONTRIBUTING.md](https://github.com/OpenTTD/OpenTTD/blob/master/CONTRIBUTING.md#pull-requests) and are using the command line (as opposed to GitHub Desktop, which doesn't really have full functionality), you should be able to rebase this fork on OpenTTD:master with the commands:
`git fetch upstream`
`git rebase upstream/master`

Then you can do `git rebase -i HEAD~4` to start an [interactive rebase](https://docs.github.com/en/get-started/using-git/about-git-rebase), which allows you to drop unwanted commits, squash commits together, and reword them to match [OpenTTD commit message format.](https://wiki.openttd.org/en/Development/Coding%20style#commit-message)

To push your changed commits to this PR, you'll need to force-push your commits with `git push -f origin HEAD`.

Also, assuming you're new to GitHub, I quite like the built-in Git GUI tool for building commits, particularly when you're adding things in multiple commits. It's also good for checking trailing whitespace, which is failing the commit checker.. You can start it with `git gui`. 😃

2TallTyler (commented):
I'm +1 to this change, and my comments are just [code style](https://wiki.openttd.org/en/Development/Coding%20style) things.

Personally, I would consider simply counting owned shares in company value a fix, while the other items in #9765 would be changes which warrant a separate PR.

LC-Zorg:
This is not a good change. Overall, incorporating the value of other companies shares into goodwill is fine, but I don't like the fact that buying 100% of the stock takes over the money of the acquired company. In practice, this means that taking over any company costs nothing, because in fact you pay for the infrastructure itself and the vehicles which are no longer worth anything. This change further strengthens this highly flawed mechanism.

In my opinion, when introducing any changes to the function of shares, it would be worth first to set a goal, how it should work, what is the purpose of buying shares and what it brings to the game. Currently, the purchase of shares in practice doesn't serve any other purpose than the possibility of unilateral takeover of companies using the robbery method (without the consent of the company owner). Such a mechanism might be ok in the past when only offline games existed. Nowadays, with a lot of games being played in multiplayer, such an arbitrary way of buying shares and taking over companies is definitely flawed.

src/economy.cpp
benda: Thanks and sorry - yes I think this was my IDE (VS2019) - I made the requested changes.

src/economy.cpp
benda: I made the requested change.

src/economy.cpp
benda: I made the requested change.

src/economy.cpp
benda: I made the requested change. Thanks for all your help!

COMPILING.md
benda: OK this is a bit complex for me, I'll try to get to this when I have more time. Thanks for the detailed instructions.

benda: I ran into a lot of conflicts - I think the simplest thing is for me to delete this PR, delete my branch, make a new branch, redo my changes in one commit manually, and make a new PR. Is that OK?

benda: Sorry for all the hassle; new PR is here: https://github.com/OpenTTD/OpenTTD/pull/9770

PR #9770, opened 2021-12-25
Fix: Company values do not properly account for shares

benda:
## Motivation / Problem

This is a redo of

https://github.com/OpenTTD/OpenTTD/pull/9768

to have everything into 1 commit. Had some rebase conflicts with prior PR. 

Company value calculation is missing the value of the company's shares in other companies

Discussed as issue 1 below:

https://github.com/OpenTTD/OpenTTD/discussions/9765

## Description

The problem is solved by first calculating all of the company values without including value of shares in other companies, then adding in the value of the shares which now is known. I do not think this causes issues with bankruptcy or buying/selling shares; it does affect those of course since they now use the new company values.

Note that when buying shares, the company value does not change, since instead of holding the cash it now holds the shares of an equivalent value.

Note that when selling shares, the company value does not change, since instead of holding the shares it now holds cash of an equivalent value.

## Limitations

I haven't tested multiplayer, this works fine on saved games in single player.

I am not a C++ programmer so my change should be thoroughly reviewed.

## Checklist for review

### DIFF
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -167,6 +167,7 @@ struct Company : CompanyProperties, CompanyPool::PoolItem<&_company_pool> {
 };
 
 Money CalculateCompanyValue(const Company *c, bool including_loan = true);
+Money CalculateCompanyValueExcludingShares(const Company *c, bool including_loan = true);
 
 extern uint _next_competitor_start;
 extern uint _cur_company_tick_index;
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -103,14 +103,33 @@ static PriceMultipliers _price_base_multiplier;
 
 /**
  * Calculate the value of the company. That is the value of all
- * assets (vehicles, stations, etc) and money minus the loan,
+ * assets (vehicles, stations, shares) and money minus the loan,
  * except when including_loan is \c false which is useful when
  * we want to calculate the value for bankruptcy.
- * @param c              the company to get the value of.
+ * @param c the company to get the value of.
  * @param including_loan include the loan in the company value.
  * @return the value of the company.
  */
 Money CalculateCompanyValue(const Company *c, bool including_loan)
+{
+	Money owned_shares_value = 0;
+
+	for (const Company *co : Company::Iterate()) {
+		uint8 shares_owned = 0;
+
+		for (uint8 i = 0; i < 4; i++) {
+			if (co->share_owners[i] == c->index) {
+				shares_owned++;
+			}
+		}
+
+		owned_shares_value += (CalculateCompanyValueExcludingShares(co) / 4) * shares_owned;
+	}
+
+	return std::max<Money>(owned_shares_value + CalculateCompanyValueExcludingShares(c), 1);
+}
+
+Money CalculateCompanyValueExcludingShares(const Company *c, bool including_loan)
 {
 	Owner owner = c->index;
 


src/economy.cpp
```diff
@@ -103,14 +103,38 @@ static PriceMultipliers _price_base_multiplier;
...
+ * @param c the company to get the value of.
  * @param including_loan include the loan in the company value.
  * @return the value of the company.
  */
-Money CalculateCompanyValue(const Company *c, bool including_loan)
+Money CalculateCompanyValue(const Company* c, bool including_loan)
+{
+	Money owned_shares_value = 0;
+
+	uint8 shares_owned = 0;
```
LordAro: variable is only used inside the loop, no need to declare it before it's used (very old-style C ! )

src/economy.cpp
```diff
@@ -103,14 +103,38 @@ static PriceMultipliers _price_base_multiplier;
...
+ * assets (vehicles, stations, shares) and money minus the loan,
  * except when including_loan is \c false which is useful when
  * we want to calculate the value for bankruptcy.
- * @param c              the company to get the value of.
+ * @param c the company to get the value of.
  * @param including_loan include the loan in the company value.
  * @return the value of the company.
  */
-Money CalculateCompanyValue(const Company *c, bool including_loan)
+Money CalculateCompanyValue(const Company* c, bool including_loan)
```
LordAro: codestyle: `*` goes on the variable, not the type

src/economy.cpp
```diff
@@ -103,14 +103,38 @@ static PriceMultipliers _price_base_multiplier;
...
+
+		for (uint8 i = 0; i < 4; i++) {
+			if (co->share_owners[i] == c->index) {
+				shares_owned++;
+			}
+		}
+
+		if (shares_owned > 0) {
+			owned_shares_value += (CalculateCompanyValueExcludingShares(co) / 4) * shares_owned;
+		}
```
LordAro: probably not worth bothering with the if statement here - CalculateCompanyValueExcludingShares isn't a particularly costly function to call needlessly (if `shares_owned == 0`)

src/economy.cpp
```diff
@@ -133,8 +157,9 @@ Money CalculateCompanyValue(const Company *c, bool including_loan)
 		}
 	}
 
-	/* Add real money value */
 	if (including_loan) value -= c->current_loan;
+
+	/* Add real money value */
```
LordAro: why this comment move?

src/company_base.h
```diff
@@ -166,6 +166,7 @@ struct Company : CompanyProperties, CompanyPool::PoolItem<&_company_pool> {
 };
 
 Money CalculateCompanyValue(const Company *c, bool including_loan = true);
+Money CalculateCompanyValueExcludingShares(const Company* c, bool including_loan = true);
```
LordAro: same issue with the `*`

LordAro (changes requested):
Purely looking at the code, haven't thought about the change itself yet

src/company_base.h
benda: Fixed

src/economy.cpp
benda: Comment applies to line below the comment. There was an unnecessary space.

src/economy.cpp
benda: Fixed

src/economy.cpp
benda: Fixed

src/economy.cpp
ldpl: CalculateCompanyValueExcludingShares iterates all vehicles and stations so it is somewhat costly. Enough to add a check to skip it in the most common case.

LordAro (dismissed):
Sure. (Needs some squashing/rebasing)

LordAro (approved):

PR #9772, opened 2021-12-28
Fix missing icon

matthijskooijman:
These are two fixes for small problems I noticed while preparing the Debian package
 - The openttd.32.bmp icon was no longer installed since the cmake conversion, but it is still referenced by the SDL driver.
 - ~~There was a super-long line that lintian complains about. Apparently super-long lines are discouraged because they might hinder other lintian checks and are generally a sign of bad code. There were some more complaints (about lang files, the changelog, a data uri in shell.html and the 3rdparty glext.h), but it did not seem appropriate or possible to change those, so I ignored those warnings instead.~~ Commit dropped, turned out the line was parsed and needs to stay one line.

## Checklist for review

* The bug fix is important enough to be backported? -> No
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md) -> No
* This PR affects the save game format? (label 'savegame upgrade') -> No
* This PR affects the GS/AI API? (label 'needs review: Script API') -> No (just touches a comment in the API)
* This PR affects the NewGRF API? (label 'needs review: NewGRF') -> No

### DIFF
--- a/media/baseset/CMakeLists.txt
+++ b/media/baseset/CMakeLists.txt
@@ -17,6 +17,7 @@ set(BASESET_OTHER_SOURCE_FILES
         ${CMAKE_CURRENT_SOURCE_DIR}/openttd.grf
         ${CMAKE_CURRENT_SOURCE_DIR}/opntitle.dat
         ${CMAKE_CURRENT_SOURCE_DIR}/orig_extra.grf
+        ${CMAKE_CURRENT_SOURCE_DIR}/../openttd.32.bmp
 )
 
 # Done by the subdirectories, if nforenum / grfcodec is installed


src/script/api/script_error.hpp
```diff
@@ -114,7 +114,13 @@ class ScriptError : public ScriptObject {
...
+		ERR_AREA_NOT_CLEAR,                           // [STR_ERROR_BUILDING_MUST_BE_DEMOLISHED, STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST, STR_ERROR_MUST_DEMOLISH_RAILROAD,
+		                                              //  STR_ERROR_MUST_DEMOLISH_AIRPORT_FIRST, STR_ERROR_MUST_DEMOLISH_CARGO_TRAM_STATION_FIRST,
+							      //  STR_ERROR_MUST_DEMOLISH_TRUCK_STATION_FIRST, STR_ERROR_MUST_DEMOLISH_PASSENGER_TRAM_STATION_FIRST,
+							      //  STR_ERROR_MUST_DEMOLISH_BUS_STATION_FIRST, STR_ERROR_BUOY_IN_THE_WAY, STR_ERROR_MUST_DEMOLISH_DOCK_FIRST,
+							      //  STR_ERROR_GENERIC_OBJECT_IN_THE_WAY, STR_ERROR_COMPANY_HEADQUARTERS_IN, STR_ERROR_OBJECT_IN_THE_WAY,
+							      //  STR_ERROR_MUST_REMOVE_ROAD_FIRST, STR_ERROR_MUST_REMOVE_RAILROAD_TRACK, STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST,
+							      //  STR_ERROR_MUST_DEMOLISH_TUNNEL_FIRST, STR_ERROR_EXCAVATION_WOULD_DAMAGE]
```
nielsmh: Indentation seems to be inconsistent here.

src/script/api/script_error.hpp
```diff
@@ -114,7 +114,13 @@ class ScriptError : public ScriptObject {
 		ERR_ALREADY_BUILT,                            // [STR_ERROR_ALREADY_BUILT, STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST, STR_ERROR_TREE_ALREADY_HERE]
 
 		/** Area isn't clear, try to demolish the building on it */
-		ERR_AREA_NOT_CLEAR,                           // [STR_ERROR_BUILDING_MUST_BE_DEMOLISHED, STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST, STR_ERROR_MUST_DEMOLISH_RAILROAD, STR_ERROR_MUST_DEMOLISH_AIRPORT_FIRST, STR_ERROR_MUST_DEMOLISH_CARGO_TRAM_STATION_FIRST, STR_ERROR_MUST_DEMOLISH_TRUCK_STATION_FIRST, STR_ERROR_MUST_DEMOLISH_PASSENGER_TRAM_STATION_FIRST, STR_ERROR_MUST_DEMOLISH_BUS_STATION_FIRST, STR_ERROR_BUOY_IN_THE_WAY, STR_ERROR_MUST_DEMOLISH_DOCK_FIRST, STR_ERROR_GENERIC_OBJECT_IN_THE_WAY, STR_ERROR_COMPANY_HEADQUARTERS_IN, STR_ERROR_OBJECT_IN_THE_WAY, STR_ERROR_MUST_REMOVE_ROAD_FIRST, STR_ERROR_MUST_REMOVE_RAILROAD_TRACK, STR_ERROR_MUST_DEMOLISH_BRIDGE_FIRST, STR_ERROR_MUST_DEMOLISH_TUNNEL_FIRST, STR_ERROR_EXCAVATION_WOULD_DAMAGE]
```
frosch123: The comment is not just a comment. It is used by the squirrel-bindings generator:
https://github.com/OpenTTD/OpenTTD/blob/adb134e49a66f8f54d9014f977604b7d32177aa6/cmake/scripts/SquirrelExport.cmake#L540

I don't think you can wrap it over multiple lines.
That's why the PR-checks fail.

src/script/api/script_error.hpp
matthijskooijman: Ah, good point. Seems my editor switched to all spaces on the second line, making the indentation dependent on my tabsize settings. I've now fixed this by leading with tabs until the ident level of the `ERR_AREA_NOT_CLEAR` on the first line, and then filling the rest with spaces, which should make it look as expected at any tab size.

I've done a force push with this change.

src/script/api/script_error.hpp
matthijskooijman: Ah, right. It looked a bit like that, but I couldn't find any reference to this file being parsed, but I guess this script is just applied to all files then maybe.

Changing that script to support wrapping is probably not worth the effort, so I'll just drop this commit and add a motivated exception in the Debian package :-)

matthijskooijman:
Hm, there is a link failure in the CI, but I don't think my change could have caused that. Any known problems in that area?

glx22:
Yeah, not your fault.
MSYS2 ld for x86_64 targets seems broken, but only for debug builds (which are used as CI check).

rubidium42:
Technically the bmp is only needed when it's being built with SDL or SDL2, so it should probably also only be copied in those cases. I have no idea how to accomplish that with CMake though.
Conceptually the copying of the bmp probably belongs more in the cmake file of the parent folder, as it has little to do with basesets. On the other hand, opntitle.dat doesn't belong there either so I'm not that bothered with that.

LordAro (approved):
#shipit

Issue #287, opened 2006-08-18
Bugfix for error in NewGRF callback 31 code for aircraft


DorpsGek:
**_mart3p_ opened the ticket and wrote:**

> The code added at r5822 to implement NewGRF callback 31 has been placed in the wrong function in aircraft_cmd.c. The code has been added to CmdSellAircraft, it should be in CmdStartStopAircraft (it is correct for all other vehicle types).
> 
> This patch corrects the problem.
> 

**Attachments**
- [GRF_Callback31_Bugfix_r5934.patch](https://bugs.openttd.org/task/287/getfile/297/GRF_Callback31_Bugfix_r5934.patch) (1.67 KiB)

**Reported version**: _trunk_
**Operating system**: _All_

---

DorpsGek:
**_Bjarni_ wrote:**

> Peter1138 better look at this one because he committed r5822

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/287#comment430_

DorpsGek:
**_Bjarni_ closed the ticket.**

**Reason for closing: _Implemented_**

> 

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/287_

PR #9781, opened 2022-01-06
Change: Automatic screenshot numbering

larryfenn:
Invoking the `screenshot` console command with a filename ending in %d will cause it to be numbered automatically.

## Motivation / Problem

As part of running my OpenTTD server I use console scripts to capture a screenshot of the game periodically. I use these screenshots to make time-lapse videos of the game.

However, the default behavior of the console command `screenshot` with a user-provided filename will always save just to that filename, causing it to be overwritten if it exists. The only way to get a 'reel' of screenshots automatically is to omit the filename argument, in which case the screenshots will be saved as e.g. `screenshot.png`, `screenshot#1.png`, `screenshot#2.png`, etc.

## Description

The proposed change causes the automatic numbering to happen to any user provided filename that ends in `%d`.

## Limitations

* The `%d` token _must_ be at the end of the user provided filename.
* The automatically numbered files start with no number, and then count up from 1. This matches the existing behavior for generated filenames, where the first file generated is `screenshot.png`, followed by `screenshot#1.png`.

## Checklist for review

### DIFF
--- a/src/screenshot.cpp
+++ b/src/screenshot.cpp
@@ -680,8 +680,18 @@ static const char *MakeScreenshotName(const char *default_fn, const char *ext, b
 		}
 	}
 
-	/* Add extension to screenshot file */
 	size_t len = strlen(_screenshot_name);
+
+	/* Handle user-specified filenames ending in %d with automatic numbering */
+	if (len >= 2) {
+		if (_screenshot_name[len - 2] == '%' && _screenshot_name[len - 1] == 'd') {
+			generate = true;
+			len -= 2;
+			_screenshot_name[len] = '\0';
+		}
+	}
+
+	/* Add extension to screenshot file */
 	seprintf(&_screenshot_name[len], lastof(_screenshot_name), ".%s", ext);
 
 	const char *screenshot_dir = crashlog ? _personal_dir.c_str() : FiosGetScreenshotDir();
--- a/src/screenshot.cpp
+++ b/src/screenshot.cpp
@@ -683,12 +683,10 @@ static const char *MakeScreenshotName(const char *default_fn, const char *ext, b
 	size_t len = strlen(_screenshot_name);
 
 	/* Handle user-specified filenames ending in %d with automatic numbering */
-	if (len >= 2) {
-		if (_screenshot_name[len - 2] == '%' && _screenshot_name[len - 1] == 'd') {
-			generate = true;
-			len -= 2;
-			_screenshot_name[len] = '\0';
-		}
+	if (StrEndsWith(_screenshot_name, "%d")) {
+		generate = true;
+		len -= 2;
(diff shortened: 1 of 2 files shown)


src/screenshot.cpp
```diff
@@ -680,8 +680,18 @@ static const char *MakeScreenshotName(const char *default_fn, const char *ext, b
 		}
 	}
 
-	/* Add extension to screenshot file */
 	size_t len = strlen(_screenshot_name);
+
+	/* Handle user-specified filenames ending in %d with automatic numbering */
+	if (len >= 2) {
```
rubidium42: There is a `StrEndsWith` function, which would probably make the code a bit clearer. So instead of the len and character compares, just use `if (StrEndsWith(_screenshot_name, "%d")) {`.

larryfenn: Thanks for the tip, I've switched the logic to just use `StrEndsWith`

TrueBrain:
Hmm. I am a bit torn. On one hand, I fully understand why you did the "has to end with" part. On the other hand, for `%d` that feels really odd.

Maybe we should do something like: when it ends with `#`, or something. Purely because anyone who understands `%d` would not expect it to only work at the end.

That all said, for sure this PR misses a change in documentation for the console command, mentioning this. Otherwise only those who have seen this PR know that this is now possible :) Mind updating the docs for this console command?

larryfenn:
I've added a documentation string to the console command - are there other places in the code where the `screenshot` command is documented?

LC-Zorg:
What do you think about using an actual date in front of each "screenshot" or "Donnington Transport, 2037-08-20"? I mean something like this:
`real date + hour` + `screenshot` / `company name, game date` that is for example: `2022-05-09 23.28.37 screenshot` or `2022-05-09 23.28.37 Donnington Transport, 1950-03-28`. In my opinion it would be useful also in case of all other screenshots made by players. Thanks to this, it would be easier to keep the chronology in the album and thus find the photos you are looking for. The history of the game would also be created in this way. For server and automatic screenshots, this can also make finding events easier.

With screenshot numbers like `#1`, `#2`, `#3` ... `#287` there may be a problem that if you remove the middle positions, the newest screenshots will most likely have the first free name, which will destroy the chronology. This problem existed on Sony phones and, until recently, on Windows as well. This is obviously a different use, but the experience was really terrible. However, if you want to keep this form, it is worth making sure that each subsequent screenshot never uses a number lower than the last one.

2TallTyler (approved):
@LC-Zorg's suggestions could be another enhancement PR but shouldn't preclude this one from being merged. Anyone writing a script to automate taking screenshots should be more than capable of sorting their screenshots.

Commits need squashing but that can happen when merging. 🙂

PR #9831, opened 2022-03-10
Fix: removing long roads doesn't prioritize refusal of local authority over other errors

### LINKED ISSUE #9820: [Bug]: Destroying road when local authority doesn't allow it gives the wrong error message
### Version of OpenTTD

12.1, Windows 11

### Expected result

When I try to delete a road but the local authority doesn't allow it, I should get an error message saying "Local authority doesn't allow this", no matter what button I use.

### Actual result

When I try to delete a road using the   button but the local authority doesn't allow it, I get an error message saying "there is no road". If I use the  button instead, it gives the correct error message.

### Steps to reproduce

1. Make the local authority of a town mad at you by destroying buildings, trees, etc.
2. Try to destroy a road owned by that town by using the  button.
3. You will get an error message saying "there is no road", see video below.

https://user-images.githubusercontent.com/40611243/154799894-1b63124a-1fbb-4a2f-99c5-b745d9988032.mp4

TrueBrain:
Fixes #9820

## Motivation / Problem

Some errors are more equal than others. Ownership errors already were. Now local authority errors are too. See #9820 for example.

## Description

See patch. Can't explain it in more words.

## Checklist for review

### DIFF
--- a/src/road_cmd.cpp
+++ b/src/road_cmd.cpp
@@ -1113,8 +1113,14 @@ std::tuple<CommandCost, Money> CmdRemoveLongRoad(DoCommandFlag flags, TileIndex 
 				cost.AddCost(ret);
 				had_success = true;
 			} else {
-				/* Ownership errors are more important. */
-				if (last_error.GetErrorMessage() != STR_ERROR_OWNED_BY) last_error = ret;
+				/* Some errors are more equal than others. */
+				switch (last_error.GetErrorMessage()) {
+					case STR_ERROR_OWNED_BY:
+					case STR_ERROR_LOCAL_AUTHORITY_REFUSES_TO_ALLOW_THIS:
+						break;
+					default:
+						last_error = ret;
+				}
 			}
 		}
 


src/road_cmd.cpp
```diff
@@ -1113,8 +1113,14 @@ std::tuple<CommandCost, Money> CmdRemoveLongRoad(DoCommandFlag flags, TileIndex
 				cost.AddCost(ret);
 				had_success = true;
 			} else {
-				/* Ownership errors are more important. */
-				if (last_error.GetErrorMessage() != STR_ERROR_OWNED_BY) last_error = ret;
+				/* Some errors are more equal than others. */
```
James103, line 1116:
```suggestion
				/* Some errors are more important than others. */
```
The word "important" describes it better than "equal".

TrueBrain: No. Humor is important!

TrueBrain: PS: I can be bought off with cookies (or beer) to give me another opinion. Just saying ;)

michicc (approved):

PR #9832, opened 2022-03-11
Change: Vanilla industry tiles have 8/8 acceptance by default

### LINKED ISSUE #9791: [Bug]: Sawmill acceptance limitations
### Version of OpenTTD

12.1

### Expected result

Any tile of an industry should have full acceptance/generation, for consistency. Currently, mines, forests, factories, farms, and steel mills follow this behaviour.

### Actual result

Sawmills can only accept wood if the middle south-western tile is covered by a freight-type station (ie not bus or passenger tram).

I believe this is a simplification to account for the variants of the industries, as there are both 3x3 sawmills and 'corner cut' sawmill types, while other industries exhibiting the desired behaviour are simple rotations, or variants have the same footprint.

As an aside, power stations only accept from the central 4 tiles, and this would further confirm the above theory, with the 2x4 'long' variant, and the 3x3 -1 'cut corner' variant having only 4 accepting tiles.

While simply to avoid for a veteran player looking at acceptance/supply while placing, this inconsistent behaviour [is illogical for newer players 
](https://old.reddit.com/r/openttd/comments/s6gm6i/why_isnt_it_accepting_timber_please/)

### Steps to reproduce

https://imgur.com/a/2UH1mVu

Initially detected in JGRPP, but confirmed in 12.1. No GRFs. 

Can be confirmed from new game (no older saves compatible to test against) - not a 'delayed demand from new industry' problem.

Searches `is:issue sawmill coverage`, `is:issue sawmill acceptance` `is:issue sawmill wood` returned no relevant results

2TallTyler:
## Motivation / Problem

In the vanilla industry set, the Sawmill, Oil Refinery, and Power Station don't have 8/8 cargo acceptance on each tile. All other industries do.

This is behavior copied from TTD (or so the story goes), but as discussed in #9791, it doesn't really add anything to the game and often confuses players.

## Description

Give 8/8 cargo acceptance to vanilla Sawmill, Oil Refinery, and Power Station tiles.

Add a setting for traditional tile acceptance behavior, used by existing savegames automatically.

Do not change tiles provided by a NewGRF, even if they just inherit vanilla tiles (these will keep their original acceptance, unchanged by this PR).

Closes #9791.

## Limitations

* Oil Rigs have strange cargo acceptance, with one tile accepting 8/8 Passengers and another accepting 8/8 Mail, both in the second cargo slot. But the built-in station seems to circumvent the problem of players finding the right place to build a station.

## Checklist for review

### DIFF
--- a/src/industry_cmd.cpp
+++ b/src/industry_cmd.cpp
@@ -411,14 +411,16 @@ static Foundation GetFoundation_Industry(TileIndex tile, Slope tileh)
 static void AddAcceptedCargo_Industry(TileIndex tile, CargoArray &acceptance, CargoTypes &always_accepted)
 {
 	IndustryGfx gfx = GetIndustryGfx(tile);
-	const IndustryTileSpec *itspec = GetIndustryTileSpec(gfx);
 	const Industry *ind = Industry::GetByTile(tile);
+	const IndustryTileSpec *itspec = GetIndustryTileSpec(gfx);
+	const IndustrySpec *indspec = GetIndustrySpec(ind->type);
 
 	/* Starting point for acceptance */
 	auto accepts_cargo = itspec->accepts_cargo;
 	auto cargo_acceptance = itspec->acceptance;
 
-	if (itspec->special_flags.Test(IndustryTileSpecialFlag::AcceptsAllCargo)) {
+	/* Default (non-NewGRF) industry tiles might accept all cargoes. */
+	if (itspec->special_flags.Test(IndustryTileSpecialFlag::AcceptsAllCargo) || (!_settings_game.economy.traditional_industry_acceptance && !indspec->grf_prop.HasGrfFile() && !itspec->grf_prop.HasGrfFile())) {
 		/* Copy all accepted cargoes from industry itself */
 		for (const auto &a : ind->accepted) {
 			auto pos = std::ranges::find(accepts_cargo, a.cargo);
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1431,6 +1431,9 @@ STR_CONFIG_SETTING_INDUSTRY_PLATFORM_HELPTEXT                   :Amount of flat 
 STR_CONFIG_SETTING_MULTIPINDTOWN                                :Allow multiple similar industries per town: {STRING2}
 STR_CONFIG_SETTING_MULTIPINDTOWN_HELPTEXT                       :Normally, a town does not want more than one industry of each type. With this setting, it will allow several industries of the same type in the same town
 
+STR_CONFIG_SETTING_TRADITIONAL_INDUSTRY_ACCEPTANCE              :Traditional industry tile acceptance
+STR_CONFIG_SETTING_TRADITIONAL_INDUSTRY_ACCEPTANCE_HELPTEXT     :When enabled, only some tiles of the default Oil Refinery, Power Station, and Sawmill accept cargo
+
 STR_CONFIG_SETTING_SIGNALSIDE                                   :Show signals: {STRING2}
 STR_CONFIG_SETTING_SIGNALSIDE_HELPTEXT                          :Select on which side of the track to place signals
 ###length 3
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -798,6 +798,11 @@ bool AfterLoadGame()
 		_settings_game.linkgraph.recalc_time     *= CalendarTime::SECONDS_PER_DAY;
 	}
 
+	/* Keep existing industry tile cargo acceptance. */
+	if (IsSavegameVersionBefore(SLV_TRADITIONAL_INDTILE_ACCEPTANCE)) {
+		_settings_game.economy.traditional_industry_acceptance = true;
+	}
+
 	/* Load the sprites */
 	GfxLoadSprites();
 	LoadStringWidthTable();
--- a/src/saveload/saveload.h
+++ b/src/saveload/saveload.h
@@ -415,6 +415,8 @@ enum SaveLoadVersion : uint16_t {
 	SLV_SIGN_TEXT_COLOURS,                  ///< 363  PR#14743 Configurable sign text colors in scenario editor.
 	SLV_BUOYS_AT_0_0,                       ///< 364  PR#14983 Allow to build buoys at (0x0).
 
+	SLV_TRADITIONAL_INDTILE_ACCEPTANCE,     ///< 365  PR#9832  Setting for vanilla industries to accept cargo on all tiles.
+
 	SL_MAX_VERSION,                         ///< Highest possible saveload version
 };
 
--- a/src/settingentry_gui.cpp
+++ b/src/settingentry_gui.cpp
@@ -843,6 +843,7 @@ SettingsContainer &GetSettingsTree()
 				industries->Add(new SettingEntry("construction.raw_industry_construction"));
 				industries->Add(new SettingEntry("construction.industry_platform"));
 				industries->Add(new SettingEntry("economy.multiple_industry_per_town"));
+				industries->Add(new SettingEntry("economy.traditional_industry_acceptance"));
 				industries->Add(new SettingEntry("game_creation.oil_refinery_limit"));
 				industries->Add(new SettingEntry("economy.type"));
 				industries->Add(new SettingEntry("station.serve_neutral_industries"));
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -578,6 +578,7 @@ struct EconomySettings {
 	bool give_money; ///< allow giving other companies money
 	bool mod_road_rebuild; ///< roadworks remove unnecessary RoadBits
 	bool multiple_industry_per_town; ///< allow many industries of the same type per town
+	bool traditional_industry_acceptance; ///< default industries don't accept cargo on all tiles
 	uint8_t town_growth_rate; ///< town growth rate
 	uint8_t larger_towns; ///< the number of cities to build. These start off larger and grow twice as fast
 	uint8_t initial_city_size; ///< multiplier for the initial size of the cities compared to towns
--- a/src/table/settings/economy_settings.ini
+++ b/src/table/settings/economy_settings.ini
@@ -142,6 +142,13 @@ str      = STR_CONFIG_SETTING_MULTIPINDTOWN
 strhelp  = STR_CONFIG_SETTING_MULTIPINDTOWN_HELPTEXT
 
 [SDT_BOOL]
+var      = economy.traditional_industry_acceptance
+def      = false
+str      = STR_CONFIG_SETTING_TRADITIONAL_INDUSTRY_ACCEPTANCE
+strhelp  = STR_CONFIG_SETTING_TRADITIONAL_INDUSTRY_ACCEPTANCE_HELPTEXT
+cat      = SC_ADVANCED
+
+[SDT_BOOL]
 var      = economy.bribe
 def      = true
 str      = STR_CONFIG_SETTING_BRIBE


TrueBrain (dismissed):
Personally I find this an excellent change! I find it a very annoying game mechanic with no added benefit other than to annoy users.

But I am curious if there are devs disagreeing :) So I leave merging for a second opinion!

frosch123 (dismissed):
Independent on whether this feature is liked or not (I personally like if multiple tiles have to be covered), the implementation is problematic:
* The tile properties defined in src/table/*h act as default values for NewGRF-defined items.
* If you were to change the acceptance of tile 0 (coal mine) you would probably break > 50% of all industry NewGRF.
* Changing the other tiles probably breaks older non-FIRS-based NewGRF, like ECS. (did not check)

The correct way to implement this, would be:
* Do not change src/table/*h ever.
* Grep for INDTILE_SPECIAL_ACCEPTS_ALL_CARGO, and add an additional condition that makes it implicitly set for all non-NewGRF defined industries. (You cannot set INDTILE_SPECIAL_ACCEPTS_ALL_CARGO explicitly, because it is set per tile, which may be used by both NewGRF-defined and default industries.)

2TallTyler:
Aha, I forgot that this would affect NewGRFs. 🤔 

[This looks like the place](https://github.com/OpenTTD/OpenTTD/blob/master/src/industry_cmd.cpp#L423) to insert a `|| "is a vanilla industry tile"`, but what does that look like? What does `GetIndustryGfx()` return if the tile is not provided by a NewGRF?

frosch123:
Both IndustrySpec and IndustryTileSpec have a "grf_prop" member. If "grffile" is "nullptr", it is not defined by a NewGRF.
I guess you want to check whether both industry and tile have no NewGRF set.

src/industry_cmd.cpp
```diff
@@ -412,6 +412,7 @@ static void AddAcceptedCargo_Industry(TileIndex tile, CargoArray &acceptance, Ca
 {
 	IndustryGfx gfx = GetIndustryGfx(tile);
 	const IndustryTileSpec *itspec = GetIndustryTileSpec(gfx);
+	const IndustrySpec *indspec = GetIndustrySpec(GetIndustryType(tile));
```
frosch123: Bikeshed: "ind" is already queried, so "ind->type" gives the same as "GetIndustryType(tile)".

src/industry_cmd.cpp
```diff
@@ -420,7 +421,8 @@ static void AddAcceptedCargo_Industry(TileIndex tile, CargoArray &acceptance, Ca
 	MemCpyT(accepts_cargo, itspec->accepts_cargo, lengthof(accepts_cargo));
 	MemCpyT(cargo_acceptance, itspec->acceptance, lengthof(cargo_acceptance));
 
-	if (itspec->special_flags & INDTILE_SPECIAL_ACCEPTS_ALL_CARGO) {
+	/* Non-NewGRF (vanilla) industry tiles accept all cargos. */
+	if (itspec->special_flags & INDTILE_SPECIAL_ACCEPTS_ALL_CARGO || (itspec->grf_prop.grffile == nullptr && indspec->grf_prop.grffile == nullptr)) {
```
frosch123: Bikeshed: I would have put "( )" around the "(a & b)" term.

LordAro: I'd consider pulling the grffile condition into its own variable. Self documenting code!
Relatedly, "accept all cargos" could be clearer - "accept all cargoes of that industry". Also, cargoes.

src/industry_cmd.cpp
```diff
@@ -411,16 +411,21 @@ static Foundation GetFoundation_Industry(TileIndex tile, Slope tileh)
...
 
 	/* Starting point for acceptance */
 	CargoID accepts_cargo[lengthof(itspec->accepts_cargo)];
 	int8 cargo_acceptance[lengthof(itspec->acceptance)];
 	MemCpyT(accepts_cargo, itspec->accepts_cargo, lengthof(accepts_cargo));
 	MemCpyT(cargo_acceptance, itspec->acceptance, lengthof(cargo_acceptance));
 
-	if (itspec->special_flags & INDTILE_SPECIAL_ACCEPTS_ALL_CARGO) {
+	/* Non-NewGRF (vanilla) industry tiles accept all cargoes of that industry. */
+	if ((itspec->special_flags & INDTILE_SPECIAL_ACCEPTS_ALL_CARGO) || (is_vanilla_industry && is_vanilla_industrytile)) {
```
LordAro: a single variable would do, surely? :p No need for it to be at the top of the function either, put it near usage. Not writing C89...

PeterN:
This could "break" save games that rely on this mechanic?

This behaviour is surely intentional as there's no other reason for the x/8 system to exist.

2TallTyler:
Yes, it could, and I believe it is intentional for those three industries — the rest have 8/8 acceptance on every tile.

Is it possible to not apply this effect to older savegames? Or do savegames get updated as soon as they're loaded and saved in a newer version of OpenTTD?

nielsmh:
A possible way would be to introduce new industry tiles for all industries that are identical to the original industry tiles, except they have 8/8 acceptance everywhere. New industries built use the new tiles, existing industries use the old tiles.
I have no idea what that would do for NewGRF compatibility, it would probably be difficult.

PeterN:
You could, of course, just do that with a NewGRF and not change the game. Or change the game and make a NewGRF to provide the old behaviour, if the game itself must be changed.

Leo512bit:
Maybe have an option in the settings menu to enable/disable for backwards compatibly?

2TallTyler:
This may be Too Much Work For Too Little Benefit.

2TallTyler:
It's "catchment areas should work as advertised," especially since most industries don't do that.

Regarding answering questions about this quirk, I don't think leaving new players out in the cold unless they ask for help is good game design.

nielsmh:
I'm airing an idea of instead making a default NewGRF that ships "in-box" with the game, and is enabled by default for new games, possibly by a special checkbox instead of adding it to the NewGRF configuration UI.
This in-box default NewGRF could basically fix things with the default game rules that are commonly seen as undesirable, such as this industry tile accepts, and various vehicles expiring and becoming unavailable.

PeterN:
That seems better to me, especially considering all the other things we won't "fix" because NewGRF allows changing it.

PeterN:
Vanilla industries tiles *don't* have 8/8 acceptance, it's not just a story ;-)

image

audigex:
After some discussion on Discord, I propose re-opening this PR

The current behaviour of vanilla industries, while arguably not truly a "bug" in the literal sense (the mechanic is intended), is confusing and unintutive - and particularly so to new players who it can easily catch out. I've seen this come up several times on TT-Forums, Discord, and Reddit, so it's clearly a recurring issue and the feedback from players who encounter it is that they think its a bug until the mechanic is explained to them, at which point they think it's a bad mechanic.

When a feature has a behaviour that is, to the average user, indistinguishable from a bug... it should probably be changed

The most obvious "fixes" seem to be

1. This PR (a fairly simple change that keeps current behaviour more or less the same except for this "fix")
2. Making the INDUSTRY display the catchment area, rather than the station (a major change that changes how users expect the game to work)
3. Keeping the catchment display on the station, but colouring the squares based on whether they accept cargo. It's something... but it's still unintuitive and would likely need to be explained
4. Create a "community bugfixes" newGRF as proposed above. A workable solution but only if the newGRF is included by default in new games, otherwise the new users this issue most affects simply won't find it

Of those options, this PR seems like the cleanest and most intuitive change

Rito13:
> The most obvious "fixes" seem to be

[#14760](https://github.com/OpenTTD/OpenTTD/pull/14760) is also an option.

src/industry_cmd.cpp
```diff
@@ -412,14 +412,17 @@ static Foundation GetFoundation_Industry(TileIndex tile, Slope tileh)
...
+	const IndustrySpec *indspec = GetIndustrySpec(ind->type);
 
 	/* Starting point for acceptance */
 	auto accepts_cargo = itspec->accepts_cargo;
 	auto cargo_acceptance = itspec->acceptance;
 
-	if (itspec->special_flags.Test(IndustryTileSpecialFlag::AcceptsAllCargo)) {
+	/* Non-NewGRF (vanilla) industry tiles accept all cargoes of that industry. */
+	bool is_vanilla_industry = indspec->grf_prop.grffile == nullptr && itspec->grf_prop.grffile == nullptr;
+	if (itspec->special_flags.Test(IndustryTileSpecialFlag::AcceptsAllCargo) || is_vanilla_industry) {
```
PeterN: This behaviour change needs to be configurable, not always on.

audigex:
> > The most obvious "fixes" seem to be
> 
> [#14760](https://github.com/OpenTTD/OpenTTD/pull/14760) is also an option.

Certainly, although I'd call it a sticking plaster over the actual problem - the mechanic would still be unintuitive, it would just be a little more visible. I'd venture most players aren't looking at that window if the UI is showing that the station "covers" the industry

I'd like to see 14760 merged too, I think it would be a nice UI improvement - but I'd still say it should be combined with a more robust improvement to remove how unintuitive this is

PeterN:
Enjoy 662 bytes of NewGRF:

https://bananas.openttd.org/package/newgrf/504e4901

2TallTyler:
Per @PeterN's request I've added a setting for legacy cargo acceptance. It is off by default so new players get the new simplified behavior.

I suppose I could add a savegame bump and force the setting to off on old savegames, if anyone thinks that is needed.

audigex:
> Enjoy 662 bytes of NewGRF:
> 
> 
> 
> https://bananas.openttd.org/package/newgrf/504e4901
> 
> 
> 
> 

The newGRF is a reasonable solution... if it's included by default with new save games that use vanilla industries

Otherwise it'd just 662 bytes of wasted space on BaNaNaS that nobody will ever use, because the people impacted by this "not a bug" feature are confused new players who won't be aware the "fix" newGRF exists

PeterN:
IMO, default should be off for existing savegames. Toggling this mid-game might have implications, and may need to trigger a refresh of station cargo acceptance.

And no, of course we won't be making an addon NewGRF part of a players default config. Addons are for players (or server admins) to configure. But the NewGRF is now there to avoid shit like "it's a waste of effort"... as I've already put in the effort.

ldpl:
If you know about this misfeature you just place station closer, not download some mods.

Even generally speaking fixing poor game design with mods is never a good idea.  Vanilla features should be adjustable in vanilla way, not with some obscure mods. Does this even work with other vanilla industry "fixes" like opengfx+ or manual industries?

src/industry_cmd.cpp
```diff
@@ -412,14 +412,17 @@ static Foundation GetFoundation_Industry(TileIndex tile, Slope tileh)
...
+	const IndustryTileSpec *itspec = GetIndustryTileSpec(gfx);
+	const IndustrySpec *indspec = GetIndustrySpec(ind->type);
 
 	/* Starting point for acceptance */
 	auto accepts_cargo = itspec->accepts_cargo;
 	auto cargo_acceptance = itspec->acceptance;
 
-	if (itspec->special_flags.Test(IndustryTileSpecialFlag::AcceptsAllCargo)) {
+	/* Default (non-NewGRF) industry tiles might accept all cargoes. */
+	bool is_default_industry = indspec->grf_prop.grffile == nullptr && itspec->grf_prop.grffile == nullptr;
```
PeterN: Also, you could do:
`bool is_default_industry = !indspec->grf_prop.HasGrfFile() && !itspec->grf_prop.HasGrfFile();`

PeterN: Although actually, these tests are unnecessary if AcceptsAllCargo is set and the new config option is not set, so maybe don't break it out into a local variable.

src/lang/english.txt
```diff
@@ -1429,6 +1429,9 @@ STR_CONFIG_SETTING_INDUSTRY_PLATFORM_HELPTEXT                   :Amount of flat
 STR_CONFIG_SETTING_MULTIPINDTOWN                                :Allow multiple similar industries per town: {STRING2}
 STR_CONFIG_SETTING_MULTIPINDTOWN_HELPTEXT                       :Normally, a town does not want more than one industry of each type. With this setting, it will allow several industries of the same type in the same town
 
+STR_CONFIG_SETTING_LEGACY_INDUSTRY_ACCEPTANCE                   :Legacy industry tile acceptance
+STR_CONFIG_SETTING_LEGACY_INDUSTRY_ACCEPTANCE_HELPTEXT          :When enabled, only some tiles of the default Oil Refinery, Power Station, and Sawmill accept cargo
```
PeterN: So yeah, I also object to this being called "Legacy".

audigex:
> IMO, default should be off for existing savegames. Toggling this mid-game might have implications, and may need to trigger a refresh of station cargo acceptance.
> 
> 
> 
> And no, of course we won't be making an addon NewGRF part of a players default config. Addons are for players (or server admins) to configure. But the NewGRF is now there to avoid shit like "it's a waste of effort"... as I've already put in the effort.

The point of it being a waste of effort was "it won't be used" rather than "I don't want to do it"

As the intermediate comment points out, people who know about this "technically not a bug" feature will just work around, they don't need a newGRF and won't bother with it

The change we need is for new players to not be confused by an unintuitive and misleading feature. If including that newGRF by default isn't an option, we're back to this PR (or some other change with a similar goal) being needed

Just letting new players be confused in perpetuity and then telling them to download a newGRF to fix it for their next game, is clearly not the way to go

Defaulting to off for existing saves but on for new ones, seems like the cleanest path forward

2TallTyler:
I've renamed "Legacy" to "Traditional" to match the main toolbar setting, and added savegame handling so existing saves get the setting enabled automatically to keep their existing behavior.

PeterN:
I'm still against this being the default.

2TallTyler:
Like all settings intended for new players, it's only helpful if it's enabled by default. (For new games, anyway, it automatically keeps existing behaviour for existing saves).

If it's not enabled by default, then we shouldn't have it at all.

I am not strongly attached to this PR, as it's not something I'd use, I'm just tired of seeing new players confused.

2TallTyler:
¯\\_(ツ)_/¯

PR #9833, opened 2022-03-11
Feature: Allow disabling local authority control of company actions

2TallTyler:
## Motivation / Problem

Local authorities are a feature that are unpopular with some players, mostly due to their sensitivity to tree removal. Several Game Scripts effectively disable this feature by forcibly setting each company's rating to "Outstanding" on a regular basis. It would be better to have a way to fix this frustration within the game.

## Description

~~I added a new `Unrestricted` option in the existing `Town council's attitude towards area restructuring` setting.~~

Per @nielsmh's suggestion, I changed "Permissive" to "Lenient" and added a new "Permissive" option.

When chosen, the company's rating still changes with their actions, but the checks for a minimum rating to do certain actions are bypassed, either by requiring the minimum possible rating or ignoring the check entirely. This includes:
- Bulldozing roads
- Bulldozing houses
- Converting a NRT road to a different type
- Building a station
- Building more than 2 airports, if noise control is off*

*Note: If noise levels are on, they work as normal with new values half as restrictive as the "Lenient" attitude.

## Limitations

The values in the new "Permissive" airport noise setting are invented and not balanced for anything. I suspect few would play with both disabled local authority control and noise limits active, but I didn't want to silently bypass the separate noise level feature.

## Checklist for review

### DIFF
--- a/src/table/settings/economy_settings.ini
+++ b/src/table/settings/economy_settings.ini
@@ -284,6 +284,14 @@ min      = 800
 max      = 65535
 cat      = SC_EXPERT
 
+[SDT_VAR]
+var      = economy.town_noise_population[3]
+type     = SLE_UINT16
+def      = 400
+min      = 100
+max      = 65535
+cat      = SC_EXPERT
+
 [SDT_BOOL]
 var      = economy.infrastructure_maintenance
 from     = SLV_166
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -675,7 +675,8 @@ static CommandCost ClearTile_Town(TileIndex tile, DoCommandFlag flags)
 	Town *t = Town::GetByTile(tile);
 
 	if (Company::IsValidID(_current_company)) {
-		if (rating > t->ratings[_current_company] && !(flags & DC_NO_TEST_TOWN_RATING) && !_cheats.magic_bulldozer.value) {
+		if (rating > t->ratings[_current_company] && !(flags & DC_NO_TEST_TOWN_RATING) &&
+				!_cheats.magic_bulldozer.value && _settings_game.difficulty.town_council_tolerance != TOWN_COUNCIL_PERMISSIVE) {
 			SetDParam(0, t->index);
 			return_cmd_error(STR_ERROR_LOCAL_AUTHORITY_REFUSES_TO_ALLOW_THIS);
 		}
@@ -3527,6 +3528,9 @@ static void UpdateTownUnwanted(Town *t)
  */
 CommandCost CheckIfAuthorityAllowsNewStation(TileIndex tile, DoCommandFlag flags)
 {
+	/* The required rating is hardcoded to RATING_VERYPOOR (see below), not the authority attitude setting, so we can bail out like this. */
+	if (_settings_game.difficulty.town_council_tolerance == TOWN_COUNCIL_PERMISSIVE) return CommandCost();
+
 	if (!Company::IsValidID(_current_company) || (flags & DC_NO_TEST_TOWN_RATING)) return CommandCost();
 
 	Town *t = ClosestTownFromTile(tile, _settings_game.economy.dist_local_authority);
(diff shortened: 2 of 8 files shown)


src/table/settings/economy_settings.ini
```diff
@@ -284,6 +284,14 @@ min      = 800
...
 cat      = SC_EXPERT
 
+[SDT_VAR]
+var      = economy.town_noise_population[3]
+type     = SLE_UINT16
+def      = 400
+min      = 100
+max      = 65535
+cat      = SC_EXPERT
+
```
stormcone, line 294: Is this setting required? I think this is a leftover from somewhere else.

2TallTyler: Yes, it's used here: https://github.com/OpenTTD/OpenTTD/blob/master/src/town.h#L122

src/town_cmd.cpp
```diff
@@ -675,9 +675,10 @@ static CommandCost ClearTile_Town(TileIndex tile, DoCommandFlag flags)
 	Town *t = Town::GetByTile(tile);
 
 	if (Company::IsValidID(_current_company)) {
-		if (rating > t->ratings[_current_company] && !(flags & DC_NO_TEST_TOWN_RATING) && !_cheats.magic_bulldozer.value) {
-			SetDParam(0, t->index);
-			return_cmd_error(STR_ERROR_LOCAL_AUTHORITY_REFUSES_TO_ALLOW_THIS);
+		if (rating > t->ratings[_current_company] && !(flags & DC_NO_TEST_TOWN_RATING) &&
+			!_cheats.magic_bulldozer.value && _settings_game.difficulty.town_council_tolerance != TOWN_COUNCIL_UNRESTRICTED) {
+				SetDParam(0, t->index);
+				return_cmd_error(STR_ERROR_LOCAL_AUTHORITY_REFUSES_TO_ALLOW_THIS);
```
michicc: Existing code seems to mostly invert the indentation here, i.e. indent lines of multi-line ifs twice while keeping the statements inside the if at one indent.

Moth-Tolias:
could the root issue not be resolved by reducing sensitivity to tree removal?

2TallTyler:
That's a legacy and game balance feature that can't really be changed without adding another setting, since this existing one doesn't touch penalties.

nielsmh:
I don't like the term "unrestricted" a whole lot in this use. The setting can end up reading as "the attitude level is unrestricted" which can mean that there is no upper and no lower bound to how happy/angry they can get.

I'd rather rename the current "Permissive" setting to "Lenient", and then make "Permissive" the name of the new value.

2TallTyler:
Thanks, I've done that. The new setting is unfortunately at the bottom of the option list to avoid breaking savegames by changing the value of each option. Is there a way to fix this?

Also, I've added a commit which improves the setting name and helptext. In-game, the local authority is always called that, not "town council," so searching for "local authority" needs to find the setting.

michicc (approved):

2TallTyler:
@nielsmh Thoughts on "Indifferent" instead of repurposing "Permissive"?

nielsmh:
Yeah that might actually be a better term all in all, at least for explaining the purpose. "They don't care what you do."
Although the company rating still changes, they have an opinion of you, they just don't act on it.

How about "Disabled"? The town local authority's powers over you are.

2TallTyler:
I thought about "Disabled" when I first started this PR but it almost sounds like a slur against the local authority, like we're calling them "disabled". I think I prefer "Indifferent" despite how the town's opinion still changes. "Disabled" has the same problem.

LC-Zorg:
So this setting will practically act like a magic bulldozer? Ok, but this means that nothing will change in online gaming and local authorities will be just the same annoying. The truth is, what is by far the most annoying thing about local authorities is their oversensitivity to tree removal, as Moth-Tolias said. I wouldn't call it a legacy or a balance worth it to be preserved. If you look closely, removing just 21 tree tiles, even accidentally, prevents the player from operating near the city. This means that later in the game, any track or road construction prevents the player from building a station. In order to be able to do so, he has to perform the sick ritual of cutting and planting trees. Ugh ... For me such a balance is not any balance, it's just a bug.

Sorry, but I don't like this change. It is not just about skipping online games, but most of all about the way it changes the player's abilities. Such a change will only worsen understanding of the game. Players were used to what each rating meant. If a player saw that he had a bad rating, he knew what to expect. Now, for each setting, responses and constraints will be different. The words for each setting will have a different meaning. In my opinion, this is not the right way. The changes should concern the impact they have on the assessment of individual activities. Whereas, a ... bad assessment should remain bad.

2TallTyler:
No, it's not like Magic Bulldozer. It completely removes any power from the local authority to control your actions, including building stations and upgrading roads. 

Changing the balance and behavior of trees is an invitation for endless arguments here and angry threads on TT-Forums. I think disabling the local authorities mechanic entirely is a much easier solution. If you have a concrete proposal for how to rebalance tree removal, I think it would be worth a separate discussion thread.

PR #9834, opened 2022-03-13
Add: [NewGRF] Map seed as global variable.

michicc:
## Motivation / Problem

NewGRF lacks access to a stable random value that is both global and fixed during runtime of a game.

```
This is useful to provide a feature-agnostic, stable random value that differs between games.
One of the possible uses is to e.g. use it to create pseudo-random regions for towns or industries.
```

## Description

The map game seed is (usually) a random value and stable, so we can just expose it as an Action 2 global variable.
Game seed is stored as a setting and doesn't change, so it should be desync-safe.

## Limitations

Global variables in OTTD are handled identical for both Action 2 and Action 6/7/9, but the game seed is not yet set when Action 6/7/9 are executed. A similar problem is present with var 0x20, which is documented to be only usable in Act 2 (reads game settings as well), but can be happily read (possibly with a different value) in Act 6/7/9 as OTTD does not check this.

As such, a documentation only limit seems to have been acceptable in the past.

## Checklist for review

### DIFF
--- a/src/genworld.cpp
+++ b/src/genworld.cpp
@@ -93,7 +93,6 @@ static void _GenerateWorld()
 		_generating_world = true;
 		if (_network_dedicated) Debug(net, 3, "Generating map, please wait...");
 		/* Set the Random() seed to generation_seed so we produce the same map with the same seed */
-		if (_settings_game.game_creation.generation_seed == GENERATE_NEW_SEED) _settings_game.game_creation.generation_seed = _settings_newgame.game_creation.generation_seed = InteractiveRandom();
 		_random.SetSeed(_settings_game.game_creation.generation_seed);
 		SetGeneratingWorldProgress(GWP_MAP_INIT, 2);
 		SetObjectToPlace(SPR_CURSOR_ZZZ, PAL_NONE, HT_NONE, WC_MAIN_WINDOW, 0);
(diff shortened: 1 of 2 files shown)


frosch123:
* Why is the mapgen seed not known during NewGRF activation? The mapgen seed is a game setting like any other, so already present when NewGRF are loaded, just like start_year or other settings. (I think the reason why the snow-line is not listed in Action6/7/9/D is because a variable-snowline depends on the date)
* Values that are constant during a game usually go into ActionD / GetPatchVariable(), to save the very limited global-variable slots.
* Scenarios use the same mapgen seed for every game, so last time this was discussed, a separate NewGRF seed was preferred.
* (possibly de-railing, feel free to ignore: AIs can request random data in script parameters, so it was suggested to do the same for NewGRF via Action14; but that is certainly more work)

michicc:
> * Why is the mapgen seed not known during NewGRF activation? The mapgen seed is a game setting like any other, so already present when NewGRF are loaded, just like start_year or other settings. (I think the reason why the snow-line is not listed in Action6/7/9/D is because a variable-snowline depends on the date)

Map seed in `_settings_game` is set in `_GenerateWorld` unless using the full GUI, which is sequenced after NewGRF loading.

> * Values that are constant during a game usually go into ActionD / GetPatchVariable(), to save the very limited global-variable slots.

See, one learns new things each time 😄 Would still need resolving the above point I guess.

> * Scenarios use the same mapgen seed for every game, so last time this was discussed, a separate NewGRF seed was preferred.

This sounds like both values would actually have a use. Map gen seed for something like regions, which I imagine is something that should be constant for a scenario, and a general random for other dynamic uses like events.
I am open to add a different NewGRF seed, I just have to find a good spot to save it and when to generate it.

michicc:
Moved the map seed to a patch variable and slightly re-ordered `GenerateWorld` to make the stored seed valid early enough for it.
It's still the map seed though, not some extra random seed.

frosch123 (approved):
I don't think scenarios are popular enough to justify having two seeds :)

PR #9842, opened 2022-04-01
Fix: No longer open a duplicate multiplayer window upon canceling password entry

### LINKED ISSUE #9736: [Bug]: Multiplayer window doubles when exiting password entering is cancelled
### Version of OpenTTD

Version: 12.1 OS: Windows 10

### Expected result

expectation: 
image
clicking the cancel button in the password entering phase closes the password entering window and brings you back to the multiplayer server list that you were previously were on.

### Actual result

image
Clicking Cancel yields another multiplayer server list window, now two of the same window is generated.

### Steps to reproduce

1. Open Multiplayer Menu
2. Join any passworded server
3. When requested for password, click cancel

AaronKatzin:
## Motivation / Problem

Fixes: #9736

## Description

When OnQueryTextFinished() is called and no password has been filled, we call ShowNetworkGameWindow() causing a new network game window to appear.
But this is unnecessary as the original window is still there too.

## Limitations

~~Since there are no game servers running on master, I was only able to test this change on release 12.1 17dfc6e28aac5ad2290e849b03893a416be97b03 (to avoid version mismatch).
If there's a way I can work around a version mismatch to test on master I'd be glad to do so.~~
Never mind, I figured out how I can test this feature based on master: I hosted my own server locally and ran a second instance of openttd in parallel.
The fix worked as expected there too.

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -2218,7 +2218,6 @@ struct NetworkJoinStatusWindow : Window {
 	{
 		if (StrEmpty(str)) {
 			NetworkDisconnect();
-			ShowNetworkGameWindow();
 			return;
 		}
 


LordAro (approved):
Thanks!

PR #9848, opened 2022-04-07
Feature: [Actions] Package builds for the Windows Store automatically

orudge:
## Motivation / Problem

The Windows Store builds currently need to be put together by hand by me. While this is not a particularly challenging task, it takes time and relies on me being free to do it.

## Description

This PR will add a 'Windows Store' job to the release action, which will generate the .appx bundle that is uploaded to the Microsoft Store (comprising x86, x64 and ARM64 binaries, plus the common package with data files, OpenGFX, OpenSFX and OpenMSX).

The .appx is only generated for tagged builds, as it has little value for end-users, and won't be uploaded to AWS along the other builds.

## Limitations

Microsoft requires all .appx bundles to be signed. This however isn't with a commercial code signing certificate (as we use for the standard Windows builds), but with a special certificate using an ID that is linked to our Windows Store entry (Microsoft then re-signs it using their own certificate once verified). For this PR, we generate a self-signed certificate for each build and include it in the artifacts - if a user wishes to test the build out on their own machine, they can then add that certificate to their trusted store and run it.

This PR doesn't implement support for automatically submitting new builds to Microsoft. This is, I believe, possible, but will require some set up (an Azure Active Directory environment is required, for instance), which I will continue to look into. There are also some limitations in terms of versioning, so we'd only want to upload final release versions, and we'd need to automate release notes generation for the Store (or simplify those notes so they just point to our web site).

The versions of OpenGFX, OpenSFX and OpenMSX to download are hard-coded; this was the case for the pre-CPack NSIS installer too.

The list of languages embedded into the app manifests is currently static; it might be nice to be able to generate these automatically, though we'd need to map some of them from the ISO codes we use to the codes that Microsoft uses.

## Checklist for review

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -850,6 +850,189 @@ jobs:
         path: build/bundles
         retention-days: 5
 
+  windows-appx:
+    name: Windows Store
+    needs:
+    - source
+    - windows
+
+    if: needs.source.outputs.is_tag == 'true'
+    runs-on: windows-latest
+
+    steps:
+    - name: Download source
+      uses: actions/download-artifact@v2
+      with:
+        name: internal-source
+
+    - name: Unpack source
+      shell: bash
+      run: |
+        tar -xf source.tar.gz --strip-components=1
+
+    - name: Download x86 build
+      uses: actions/download-artifact@v2
+      with:
+        name: openttd-windows-x86
+
+    - name: Download x64 build
+      uses: actions/download-artifact@v2
+      with:
+        name: openttd-windows-x64
+
+    - name: Download arm64 build
+      uses: actions/download-artifact@v2
+      with:
+        name: openttd-windows-arm64
+
+    - name: Unpack builds
+      shell: bash
+      run: |
+        mkdir builds
+        cd builds
+
+        function extract {
+            mkdir $1
+
+            # Extract the zip version of the release
+            unzip ../openttd-*-windows-$2.zip -d $1
+
+            # Remove the extraneous directory
+            mv $1/openttd-*-windows-$2/* $1/
+            rmdir $1/openttd-*-windows-$2
+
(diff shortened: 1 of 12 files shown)


LordAro (approved):
¯\\\_(ツ)_/¯

.github/workflows/release.yml
```diff
@@ -850,6 +850,189 @@ jobs:
...
+
+        rm -f opensfx-all.zip opensfx-*.tar
+
+    - name: Generate signing certificate
+      shell: cmd
+      run: |
+        cd builds
+
+        REM We need to provide a signed .appx to the Windows Store, so generate a certificate with password 'password'
+        call ..\os\windows\winstore\generate-key.bat "CN=78024DA8-4BE4-4C77-B12E-547BBF7359D2" password cert.pfx
```
TrueBrain, line 974: So we generate a new certificate every time, to sign a file to send to Microsoft .. that sounds ... useful? Lol .. I am sure Microsoft has its reasons ...

TrueBrain (approved):

.github/workflows/release.yml
TrueBrain: Ugh, hit enter too quickly: should the CN number be private? (read: in a GitHub Secret?)

orudge: We generate a new certificate each time, though we could alternatively manually generate a certificate and use that every time (which might be easier if anybody wanted to test the .appx outside the Store, in that they'd just need to trust that certificate once). Our standard code signing certificate however doesn't work here because the app has to have the CN that the Store gave us. Microsoft then discards our certificate and signs it with their own certificate upon upload to the store.

In terms of the CN being secret, no, it doesn't need to be; anybody can see it by installing the app from the store and poking around in the manifest files that get installed to your machine.

PR #9861, opened 2022-04-18
Change: recognize fedora/rhel distros in cmake

nikolas:
This allows for rpm packages to be created with cpack.


glx22 (approved):
Can't really test it, but the code seems to do the right thing.

PR #9872, opened 2022-04-30
Add: Show video driver name in Game Options window

nielsmh:
## Motivation / Problem

It's difficult for users to be sure exactly what video driver is being used currently.

## Description

Let video drivers describe themselves and print the info in the Game Options window.

image

## Limitations

Some OpenGL drivers have really long names, which can make the window excessively wide

## Checklist for review

### DIFF
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -310,6 +310,7 @@ struct GameOptionsWindow : Window {
 			case WID_GO_BASE_SFX_DROPDOWN:     SetDParamStr(0, BaseSounds::GetUsedSet()->name); break;
 			case WID_GO_BASE_MUSIC_DROPDOWN:   SetDParamStr(0, BaseMusic::GetUsedSet()->name); break;
 			case WID_GO_BASE_MUSIC_STATUS:     SetDParam(0, BaseMusic::GetUsedSet()->GetNumInvalid()); break;
+			case WID_GO_VIDEO_DRIVER_INFO:     SetDParamStr(0, VideoDriver::GetInstance()->GetInfoString()); break;
 			case WID_GO_REFRESH_RATE_DROPDOWN: SetDParam(0, _settings_client.gui.refresh_rate); break;
 			case WID_GO_RESOLUTION_DROPDOWN: {
 				auto current_resolution = GetCurrentResolutionIndex();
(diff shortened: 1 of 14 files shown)


LordAro:
Looks good. Seems like `glGetString` doesn't exist on Linux? Existing functions seem to have a `_` prefix

src/settings_gui.cpp
```diff
@@ -323,6 +323,7 @@ struct GameOptionsWindow : Window {
 				}
 				break;
 			}
+			case WID_GO_VIDEO_DRIVER_INFO:     SetDParamStr(0, VideoDriver::GetInstance()->GetInfoString()); break;
```
LordAro: Should probably be multiple lines

LordAro (commented):
Wonder if it would be worth splitting driver name and ogl driver name onto separate lines. Would avoid the line-too-long issue

src/video/video_driver.hpp
```diff
@@ -178,6 +178,11 @@ class VideoDriver : public Driver {
 		return ZOOM_LVL_OUT_4X;
 	}
 
+	virtual const char * GetInfoString() const
```
TrueBrain: Nitpicking: space after * seems random :)

PS: nice idea, this PR :D

src/settings_gui.cpp
nielsmh: All of the above ones (before the one multiline case) are formatted the same way. If I moved it before the multiline case it would fit in?

src/video/video_driver.hpp
nielsmh: I honestly don't remember what the convention is, I just feel like for functions returning pointers or references, attaching the sigil to the function name feels wrong since the function name itself isn't a pointer/reference.

TrueBrain: I just noticed you did it without the space the other 3 times in this PR :) my OCD triggered on that :p

nielsmh:
Tried it on my Linux machine.

I'm not sure putting the "base" driver and GPU driver on separate lines is a good idea, if only because vertical space is already at a premium in the Options window. Also, hiding the GPU driver line when using a non-accelerated base driver is annoying.

image

TrueBrain:
Do you mind adding this information in the crash log too? The driver is already in there, but not the OpenGL extended information, as far as I can see. (doesn't have to be this PR btw .. but it just sparked to mind)

LordAro (approved):
shipit

PR #9879, opened 2022-05-06
Add: Commandline option to skip NewGRF scanning

nielsmh:
## Motivation / Problem

Sometimes when testing stuff, waiting for the NewGRF scanning just takes annoyingly long time, and you don't need NewGRFs anyway for what you're testing. But you also don't want to switch to a different config/file structure.

## Description

Add a commandline option `-Q` (for Quick) that skips all scanning for NewGRF files.

## Limitations

Haven't tested how this behaves with the in-game content downloading UI or anything like that.

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -201,6 +201,8 @@ static void ShowHelp()
 		"  -x                  = Never save configuration changes to disk\n"
 		"  -X                  = Don't use global folders to search for files\n"
 		"  -q savegame         = Write some information about the savegame and exit\n"
+		"  -Q                  = Don't scan for/load NewGRF files on startup\n"
+		"  -QQ                 = Disable NewGRF scanning/loading entirely\n"
 		"\n",
 		lastof(buf)
 	);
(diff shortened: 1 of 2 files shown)


glx22:
If I follow the code properly it disables all scans, even manual ones.
I think it should be possible to disable only the automatic first scan.

nielsmh:
Hmm, I could make it so `-Q` disables the first scan, and `-QQ` disables all scans.

src/openttd.cpp
```diff
@@ -201,6 +201,7 @@ static void ShowHelp()
 		"  -x                  = Never save configuration changes to disk\n"
 		"  -X                  = Don't use global folders to search for files\n"
 		"  -q savegame         = Write some information about the savegame and exit\n"
+		"  -Q                  = Don't scan for NewGRF files (quick startup)\n"
```
TrueBrain: I think this will give people the impression using this gives them a better experience (for most mere mortals it will not). Maybe make more clear what the impact is? (Maybe by removing quick startup remark? Dunno .. describing things is hard :D)×

Alternatively, for network stuff we use the env to all developers to modify things normal users shouldn't,  to kinda avoid this problem. Not sure if that is a road to take here, but just mentioning. 

Either way, please also document QQ :)

TrueBrain (approved):

PR #9880, opened 2022-05-06
Codechange: Document the _ship_subcoord table

nielsmh:
## Motivation / Problem

The code as-is looks like black magic.

## Description

While working on my multithreading project, documenting this just happened.

## Limitations

None?

## Checklist for review

### DIFF
--- a/src/ship_cmd.cpp
+++ b/src/ship_cmd.cpp
@@ -516,38 +516,53 @@ static inline TrackBits GetAvailShipTracks(TileIndex tile, DiagDirection dir)
 	return tracks;
 }
 
-static const byte _ship_subcoord[4][6][3] = {
+/** Structure for ship sub-coordinate data for moving into a new tile via a Diagdir onto a Track. */
+struct ShipSubcoordData {
+	byte x_subcoord; ///< New X sub-coordinate on the new tile
+	byte y_subcoord; ///< New Y sub-coordinate on the new tile
+	Direction dir;   ///< New Direction to move in on the new track
+};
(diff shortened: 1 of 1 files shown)


glx22 (approved):

src/ship_cmd.cpp
```diff
@@ -516,38 +516,53 @@ static inline TrackBits GetAvailShipTracks(TileIndex tile, DiagDirection dir)
...
+	// DIAGDIR_NE
 	{
-		{15, 8, 1},
-		{ 0, 0, 0},
-		{ 0, 0, 0},
-		{15, 8, 2},
-		{15, 7, 0},
-		{ 0, 0, 0},
+		{15,  8, DIR_NE},      // TRACK_X
+		{ 0,  0, INVALID_DIR}, // TRACK_Y
```
frosch123, line 534: I don't undestand this. INVALID_DIR is 0xFF, not 0. DirDifference() asserts on that.

frosch123 (changes requested):

src/ship_cmd.cpp
nielsmh: As described in the comment, the table maps from a pair of Diagdir and Track to a set of coordinates. Only three of the tracks are reachable from one tile to another via a specific Diagdir, so only three of the six entries for each Diagdir are valid. The entries with INVALID_DIR are for the Tracks not reachable via the Diagdir.
Should I be more explicit about this in the comment?

frosch123: Thanks, I did not notice that.

frosch123 (approved):

PR #9885, opened 2022-05-10
CVE-2022-30292 fix

PavelBlinnikov:
## Motivation / Problem

This is the fix for CVE-2022-30292 which is the critical Squirrel vulnerability:
- [Security advisory](https://github.com/sprushed/CVE-2022-30292)
- [MITRE reference](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2022-30292)

### DIFF
--- a/src/3rdparty/squirrel/squirrel/sqbaselib.cpp
+++ b/src/3rdparty/squirrel/squirrel/sqbaselib.cpp
@@ -816,6 +816,7 @@ static SQInteger thread_call(HSQUIRRELVM v)
 	SQObjectPtr o = stack_get(v,1);
 	if(type(o) == OT_THREAD) {
 		SQInteger nparams = sq_gettop(v);
+		sq_reservestack(_thread(o), nparams + 3);
 		_thread(o)->Push(_thread(o)->_roottable);
 		for(SQInteger i = 2; i<(nparams+1); i++)
 			sq_move(_thread(o),v,i);


nielsmh:
I believe this is not actually an issue that will be present in OpenTTD. As the below code shows, the SQVM::Push() function is patched in our Squirrel version to either (in debug builds) assert on stack overrun, or (in release builds) automatically extend the stack.

https://github.com/OpenTTD/OpenTTD/blob/8d5115615002508694c80e9defcc6e768cc66a11/src/3rdparty/squirrel/squirrel/sqvm.cpp#L1543-L1555

The advisory on this is quite thin on information, but it looks like a heap buffer overrun. My instinct says this would probably be difficult to exploit, but I'm not at all an expert.

PavelBlinnikov:
But OpenTTD has asserts disabled by default in release builds. This is possible to exploit.

glx22:
All squirrel allocations (stack resizing included) are accounted, and the script is killed if it allocated too much.

PavelBlinnikov:
> All squirrel allocations (stack resizing included) are accounted, and the script is killed if it allocated too much.

That does nothing with the vulnerability. Because this is not a simple OOM but a vulnerability of memory corruption.

TrueBrain:
Tnx for this PR @PavelBlinnikov , much appreciated :D

As OpenTTD runs a rather modified Squirrel version, I am trying to understand the impact this has for us. Would you be able to share with us some more details on what is actually going on and how it can be exploited?

I can understand you are not willing to share that publicly yet (and in fact, I would appreciate if you didn't yet :p), but an email to either info@ or truebrain@ on openttd.org would be much appreciated :D

Alternatively we need to do the analysis ourselves ofc ;) Basically I need to know if we need to CVE OpenTTD and push a release or not :D

Tnx a bunch!

PavelBlinnikov:
> Tnx for this PR @PavelBlinnikov , much appreciated :D
> 
> As OpenTTD runs a rather modified Squirrel version, I am trying to understand the impact this has for us. Would you be able to share with us some more details on what is actually going on and how it can be exploited?
> 
> I can understand you are not willing to share that publicly yet (and in fact, I would appreciate if you didn't yet :p), but an email to either info@ or truebrain@ on openttd.org would be much appreciated :D
> 
> Alternatively we need to do the analysis ourselves ofc ;) Basically I need to know if we need to CVE OpenTTD and push a release or not :D
> 
> Tnx a bunch!

After completed analysis I can say that modified version of Squirrel in OpenTTD is not vulnerable because of patched SQVM::Push function. So there is really no problem here. Thank you for the quick response.

TrueBrain:
Tnx a lot for looking further into this, it is much appreciated :D And happy we are not affected.

Tnx again, and awesome you are putting in the effort here!

PR #9827, opened 2022-02-23
Feature: Improved finance window

2TallTyler:
## Motivation / Problem

fianances_old

`Group expenses in company finance window` has been on by default since #8463. Even so, the window is cluttered with redundant text ("Income" and "Running Costs") and a whole bunch of black numbers.

Additionally, the "Income" is incorrectly used to describe "Revenue" — income is revenue minus expenses.

## Description

financeGUI

I've reworked the window inspired by a design by @LC-Zorg [on TT-Forums](https://www.tt-forums.net/viewtopic.php?t=86948). Each category gained a title with an indented breakdown below.

Per discussion here and on Discord, the total was kept below the category instead of being in-line with the title, as I originally proposed.

The expense/revenue names have been significantly simplified, both for vehicle types as well as "Property Maintenance" becoming "Infrastructure".

Income is renamed to Revenue in several places, both in strings and source code. The script API has been left unchanged.

The balance sheet at bottom left has been reorganized to emphasize the bank balance available for construction, not the company's own funds, which would be negative until the bank balance is higher than the loan.

I have also added the loan interest above the maximum loan. Both of these are purely informational, as the loan interest currently cannot be changed during a game.

## Limitations

The un-grouped finance window no longer exists, a casualty of how I've rewritten the window. It was awful and needed to disappear anyway.

The new names for Revenue, Operating Expenses, and Capital Expenses are American English. Feel free to tell me I'm a silly yank if the terms are different in the UK. 😄 

LC-Zorg's proposal includes a slider to change how much loan is borrowed or repaid per click. I don't consider that in scope for this PR.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #9887, opened 2022-05-11
Change: Use an indent, not a dash, to list train capacity

### LINKED ISSUE #9882: [Bug]: Negative train total cargo
### Version of OpenTTD

Built from latest master (0dce7bf85d56b879e6e7e42b2bef4ef827275cf9) on Ubuntu 22.04

### Expected result

Total cargo that the train is carrying should be a positive number

### Actual result

Total cargo that the train is carrying is presented as a negative number:
Screenshot from 2022-05-08 16-14-33

### Steps to reproduce

Unsure what scenario caused the issue.
I only see it on trains, not on trucks.
I don't have any boats or planes in this scenario so I'm not sure if the bug is present there too.

2TallTyler:
## Motivation / Problem

In #9882 a player thought the dash used as a list entry prefix for in the train Total Cargo GUI was a minus sign.

## Description

We discussed other characters and reached a consensus that no character would be better.

I also added an indent (the same size used in the finance window in #9827) for the list items.

indent

Closes #9882.

## Checklist for review


src/train_gui.cpp
```diff
@@ -363,6 +363,11 @@ void DrawTrainDetails(const Train *v, int left, int right, int y, int vscroll_po
 	/* get rid of awkward offset */
 	y -= WD_MATRIX_TOP;
 
+	/* Indent the total cargo capacity details */
+	int EXP_INDENT = 10;
```
glx22: Should at least be `const`, and the `EXP_` part comes from expenses window ;)

glx22 (approved):

PR #9889, opened 2022-05-12
Fix #9883: Show cost/income float over end tile of rail or road construction

### LINKED ISSUE #9883: [Bug]: The cost of building roads and tracks is often shown out of sight of the player
### Version of OpenTTD

Since at least 1.4.0, this is an old bug

### Expected result

The cost of building roads and tracks is shown at the end of the constructed section, which is always visible to the player - in the same way as for all other construction options.

### Actual result

The costs of building roads and tracks are the only ones shown at the place where construction starts, which means that the player is often not able to see how much he has paid for their construction, especially when the section is really long. It can be said that the longer the section built, the bigger the problem is, because although the construction cost becomes more and more important with the length, it is also harder and harder to notice it, and at some point it becomes impossible.

In all of the examples below, I was building from top to down.
10 Planting trees
11 Demolishing
12 Terraforming
13 Canal construction
14 Bridge construction
15 Track or road reconstruction

Everywhere, the cost is shown at the drop point, which is logical and functional. The exception is the construction of roads and tracks.
16 Road construction
17 Track construction

### Steps to reproduce

Just build a long or very long section of road or track and try to see how much it cost to build.
Note that this behavior is different from the rest of the build actions.

2TallTyler:
## Motivation / Problem

As documented in #9883, when building or removing long stretches of rail or road segments, the cost/income float appears over the first tile where it can be out of sight of the player. This is unlike other things like landscaping, bridges, etc.

## Description

[CommandHelperBase::InternalPostResult](https://github.com/OpenTTD/OpenTTD/blob/646d160c3803385ce805ddb4eee817dbafcad674/src/command.cpp#L253) only takes one tile parameter to pass to `ShowCostOrIncomeAnimation()`. Other build/remove commands which work properly have the end_tile as the first parameter and start_tile as the second parameter. I'm no expert on C++ templates but it appears that it's taking only the first TileIndex parameter, so I swapped the parameter order such that InternalPostResult receives the end tile rather than the start.

Closes #9883.

## Limitations

I didn't find where the template looks for the first TileIndex only, so I may have broken things I'm not aware of.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -922,8 +922,8 @@ static CommandCost CmdRailTrackHelper(DoCommandFlag flags, TileIndex tile, TileI
  * Build rail on a stretch of track.
  * Stub for the unified rail builder/remover
  * @param flags operation to perform
- * @param tile start tile of drag
  * @param end_tile end tile of drag
+ * @param start_tile start tile of drag
  * @param railtype railroad type normal/maglev (0 = normal, 1 = mono, 2 = maglev), only used for building
  * @param track track-orientation
  * @param auto_remove_signals false = build up to an obstacle, true = fail if an obstacle is found (used for AIs).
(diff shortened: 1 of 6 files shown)


LordAro (approved):
I'm not a huge fan of end_tile & start_tile being "backwards", but I don't think there's anything that can be done about it due to the "special" nature of the first tile in the command args. Unless @michicc can think of a better solution?

ldpl:
> I'm not a huge fan of end_tile & start_tile being "backwards", but I don't think there's anything that can be done about it due to the "special" nature of the first tile in the command args. Unless @michicc can think of a better solution?

IMO it would be better to make `tile` into an explicit attribute of the `CommandCost`.

michicc:
There is precedence for the "backwards" variant already, so let's just go with it instead of rejigging all of `CommandCost`.

ldpl:
This swap didn't take into account special handling that first tile parameter has. Afaict it now has double checks for the end tile and none for the start one possibly introducing a security vulnerability.

PR #9891, opened 2022-05-12
Feature: Towns don't build parallel, redundant bridges

2TallTyler:
## Motivation / Problem

As discussed in #9299, cities often build parallel road bridges over ocean, which look silly.

## Description

Before building a bridge over ocean, towns search the nearby area for parallel road bridges. This area is one bridge-length perpendicular and half a bridge-length parallel to the starting bridge head, to find bridges on the same shoreline but not bridges "behind" the proposed bridge, such as on a small island.

The easiest way to test this change is in Scenario Editor with the Expand button on towns. I've prepared a simple before-and-after example to prove that this works ([bridges_test.zip](https://github.com/OpenTTD/OpenTTD/files/8681201/bridges_test.zip)) but I encourage people to experiment and find any corner cases that break this and might need more tuning.

Closes #9299 (might need converting to an Issue first).

## Limitations

This is intended to work alongside #9890, which caps bridge length at the original 11 tiles. Extremely long 20+ tile bridges would have an enormous search area and would find all sorts of parallel bridges which are not actually redundant.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -1161,6 +1161,27 @@ static bool CanRoadContinueIntoNextTile(const Town *t, const TileIndex tile, con
 }
 
 /**
+ * CircularTileSearch proc which checks for a nearby parallel bridge to avoid building redundant bridges.
+ * @param tile The tile to search.
+ * @param user_data Reference to the valid direction of the proposed bridge.
+ * @return true if another bridge exists, else false.
+ */
+static bool RedundantBridgeExistsNearby(TileIndex tile, void *user_data)
+{
+	/* Don't look into the void. */
+	if (!IsValidTile(tile)) return false;
+
+	/* Only consider bridge head tiles. */
+	if (!IsBridgeTile(tile)) return false;
+
+	/* Only consider road bridges. */
+	if (GetTunnelBridgeTransportType(tile) != TRANSPORT_ROAD) return false;
+
+	/* If the bridge is facing the same direction as the proposed bridge, we've found a redundant bridge. */
+	return (GetTileSlope(tile) & InclinedSlope(ReverseDiagDir(*(DiagDirection *)user_data)));
+}
+
+/**
  * Grows the town with a bridge.
  *  At first we check if a bridge is reasonable.
  *  If so we check if we are able to build it.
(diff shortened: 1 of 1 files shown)


nielsmh:
I didn't watch carefully what order things happened in, but I think this is outside the intent of this patch?

image

2TallTyler:
I'm not sure what's going on there, but I rewrote the function to use a CircularTileSearch() instead of my own proprietary search function and remove possibilities for error. Hope that fixes it.

LC-Zorg:
Even now, if it doesn't work perfectly, the effect is much better - bridges are created during the game by more than half as much as before.

After change
City bridges - after change

Before change (red bridges are the ones cities haven't built thanks to restrictions)
City bridges - before change x


The situation presented by Nielsmh most likely comes from the map editor or from the beginning of the game, right after generating the cities. It seems that in this case the function has no chance to work because the game builds many bridges at the same time and cannot relate to those that exist, because they simply do not exist.

In most cases, when the generated cities are not large, this is not a significant problem. The problem arises when the initial size of cities is large - then there may be more bridges and they may be longer.

I don't know if this preview has implemented the #9890 length limitation for city bridges. If so, it doesn't work.
City bridges - after map generation with big cities

I think it might be worthwhile to specify a minimum search radius to be at least 5, for example, so that also small bridges would not be built side by side.

Search radius: 1:1 (actual)
City bridges - search radius 1-1

Search radius: minimum 5
City bridges - search radius min 5





I guess there shouldn't be a problem, but perhaps, just in case, it would also be worth checking the performance of this solution and whether it doesn't slow down the game in case when cities can build many bridges at the same time.
I wanted to create dedicated test save, but I don't know how to read it in this version of the game.

It would be nice if You could change the generation of bridges also during the initial generation of cities. Especially if You wanted to go further and put a restriction as well so that bridges are not built along the quays. But even in its present form, it's a useful improvement. :) Thanks, thanks, thanks :)

LC-Zorg:
Regarding the limitations of building bridges along the quays, perhaps such a simplified test could be done? 
City bridges - bridges along the quay, test
Instead of searching a large area, only a few or a dozen tiles to check.

It would not rule out all silly cases, but for the vast majority it should be effective.

2TallTyler:
The code you tested is different than the screenshot by nielsmh, so that bug may not exist anymore (I certainly hope not!). I just pushed an update which implements #9890 (it previously did not, as you noticed).

As an aside, map generation doesn't handle road or bridge construction any differently than when a town grows during gameplay. It still grows one tile at a time using the same code. (It particularly doesn't build many bridges at the same time, because computers don't really do multiple things at the same time. 😉)

I don't think a minimum search radius is appropriate. Short bridges should be able to be built near each other, as we often see that in real life, as in your screenshot of the many bridges over the canal. Look at Chicago, for example.

I'll do some thinking about not building bridges parallel to the coast.

michicc (approved):

PR #8616, opened 2021-01-30
Add: Currency:  Indonesian Rupiah (IDR)

WilloIzCitron:
Adds a Currency of indonesia
With conversion rate of 19 to GBP.

PR #8386, opened 2020-12-15
Change: extend the allowed range for max loan setting

ldpl:
From 100000 - 500000 to 0 - 2000000000.
0 to allow gamescripts do game modes without loan. 2000000000 since it's enough for all forseable reasons and int32 is the max settings gui currently supports (well, it was a bit less but I fixed that). Also fixed gui overflow with iranian rials while at it.

Issue #8253, opened 2020-07-02
Very large company values are not graphed correctly


James103:
## Version of OpenTTD
master (2020-07-01)

## Expected result
* Company value of 2^63-1 is recorded on the company value graph.
* Company value of e.g. 2^63-10000 is recorded on the company value graph, with data points being visible.
* The scale in these cases runs from 0 to 2^63-1.

## Actual result
* Company value of 2^63-1 is not recorded on the company value graph at all (scale still runs from -5 to 4).
![image](https://user-images.githubusercontent.com/37945304/86307343-279cad00-bbdc-11ea-9c8c-8e703d909335.png)

* Company value of e.g. 2^63-10000 is still recorded on the company value graph, but the points do not appear.
![image](https://user-images.githubusercontent.com/37945304/86307414-561a8800-bbdc-11ea-9562-889cb039d5d0.png)

* The scale runs from 0 to (2^63-1)/10 (~922 billion million) instead of from 0 to 2^63-1 (~9.22 trillion million).
(see above images)

## Steps to reproduce
1. Load the attached savegame.
2. Fast forward a few years. The money (and company value) will max out.
3. Open the company value graph. It appears blank.
4. Open your company finances window, and hit the "Borrow £10,000" button a couple of times.
5. Fast forward a few months.
6. Take a look at the company info window and company value graph.
7. Hit the "Repay £10,000" button until the loan has been repaid.
8. Take a look at the company info window and company value graph.

## Savegame
[Gronnway Transport, 363576-09-15.zip](https://github.com/OpenTTD/OpenTTD/files/4861520/Gronnway.Transport.363576-09-15.zip)
(The savegame contains a single company with maxed-out (2^63-1) company value and cash, and has been optimized to run at up to several in-game years per second on high-end modern hardware.)

TrueBrain:
You are an absolute madman to make a savegame for this and clear steps to reproduce :D Awesome, tnx :)

LordAro:
The first part's quite fun - the graphing code uses INT64_MAX as "INVALID_DATAPOINT" . Of course, if it actually gets that value, it just ignores all the data.

https://github.com/OpenTTD/OpenTTD/blob/1e529e16329b905e625bb69cc501edbc478186bc/src/graph_gui.cpp#L39

A hacky fix might be to use INT64_MIN as a sentinel value, but I'd say that in real world conditions one is no less likely to appear than the other, so probably wouldn't actually fix anything

[...]

PR #7616, opened 2019-06-04
Codechange: Change KRW currency rate

telk5093:
Since 2017, I think because of the Brexit, KRW against GBP has been about 1GBP≒1,500KRW.
OpenTTD is using 1GBP=1,850KRW, which makes very difficult to calculate.

This PR makes 1GBP=1,500KRW which is a close rate in nowadays, and also easier to calculate
But I have no idea if I change openttd's currency rate, please let me aware.

PR #9897, opened 2022-05-24
Update IDR currency to current exchange rate

JulianGro:
## Motivation / Problem
This PR updates the exchange rate of IDR, since the current exhange rate does not fit at all.

## Checklist for review

### DIFF
--- a/src/currency.cpp
+++ b/src/currency.cpp
@@ -66,7 +66,7 @@ static const CurrencySpec origin_currency_specs[CURRENCY_END] = {
 	{    8, "", CF_NOEURO, u8"\u00a5",     "",               0, STR_GAME_OPTIONS_CURRENCY_CNY    }, ///< chinese renminbi
 	{   10, "", CF_NOEURO, "HKD" NBSP,     "",               0, STR_GAME_OPTIONS_CURRENCY_HKD    }, ///< hong kong dollar
 	{   90, "", CF_NOEURO, u8"\u20b9",     "",               0, STR_GAME_OPTIONS_CURRENCY_INR    }, ///< Indian Rupee
-	{   19, "", CF_NOEURO, "Rp",           "",               0, STR_GAME_OPTIONS_CURRENCY_IDR    }, ///< Indonesian Rupiah
+	{18357, "", CF_NOEURO, "Rp",           "",               0, STR_GAME_OPTIONS_CURRENCY_IDR    }, ///< Indonesian Rupiah
 	{    5, "", CF_NOEURO, "RM",           "",               0, STR_GAME_OPTIONS_CURRENCY_MYR    }, ///< Malaysian Ringgit
 };
 


orudge:
It looks like IDR was added in https://github.com/OpenTTD/OpenTTD/pull/8616 as 19 instead of 19,000 due to theoretical overflow issues. I don't know if anyone is especially likely to have more than £485 trillion in a game though...

James103:
As a workaround, you could make a custom currency with the following parameters:
* Exchange rate: Rp&nbsp;19&NoBreak;**,000** = £ 1 (the bolded text is part of the Suffix and not the number itself)
* Separator: `,` (depending on locale)
* Prefix: `"Rp" NBSP`
* Suffix: `,000` (depending on locale)

JulianGro:
> 

That is a nice idea. I have no idea how to switch between , and . depending on locale though.

telk5093:
Note that there was a similar issue of KRW (#7616) but it is rejected.

JulianGro:
So I have given this some thought and exploration.
From my point of view the best thing to do is actually to go forward with the current changes.

Adding a "thousand" to the currency seems very alien to at least the Indonesian I have talked to.
Having the rate be rounded to something like 18000 makes the last three digits never change in-game, which looks weird to me and isn't how the rest of the currencies are. (Even on GBP you have to pay £4,921 for a bus, instead of a more realistic price like £4,900 or £5,000)
Also trying to add ".000" or ",000" as a postfix is too much for me and my knowledge of C++ and the codebase; Hardcoding it to ".000" would look weird in locales that would normally use ",", and haven't been able to use something like `digitsepcur` instead.

For the currency overflow consideration, I don't think this should be too much of a problem. There is already a currency that has a 4901 multiplicator. And realistically, if there is overflow problems you are going to run into them eventually. This would just make it 4 times faster than on the currency with the current highest multiplicator.

nielsmh:
All money calculations in the game are still done in whole GBP, and currency conversion is only done to display the figures. I'm not sure if currency display also uses the OverflowSafeInt type, but if it does then any overflow should instead saturate and it would rather look like you hit the ceiling for amount of money in the bank long before you actually do.
I'm not sure how common it is to ever get that much money in regular games, but I suppose if enough complaints come in it could be possible to add some sort of "you have more money than we can show" indicator, or possibly even an ultra-long multiplication of sorts.

James103:
> I'm not sure if currency display also uses the OverflowSafeInt type

I can confirm that it does by loading the savegame from #8253, spending some money while the currency is GBP, then switching to a currency such as EUR, USD, or JPY (where more than one of that currency = £1), then spending some more money and seeing that the available balance remains at INT64_MAX in the displayed currency.

ldpl:
Overflow isn't just about the total amount of money in company, there are some other places that do calculations with currency rates, they all need checking. For example, I had to fix one of such issues in settings display for #8386 as even 5k rates were buggy prior to that change. So it's unclear what's the safe max for the multiplier is, someone needs to go over all the code that uses rates and check for the limits. I know for sure 50k multiplier is buggy so 19k is getting pretty close.

LordAro:
Yeah, intentional design decision as stated in the original PR. Changing the suffix might be more appropriate, but as you say, there's no real good alternative

PR #9901, opened 2022-05-29
Fix: Improve performance when calculating company value

SamuXarick:
## Motivation / Problem
Commit 1:
Some scripts make heavy use of CalculateCompanyValue and AI companies can't own shares of other AIs anyway, so there is no point calculating company values just to multiply it by zero in the end.

Commit 2:
`GetAmountOwnedBy` can do the same functionality as that for loop it replaces.

## Description
Commit 1:
Don't compute the value of the other company when the shares a company has on the other company is zero.

Commit 2:
N/A

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -101,6 +101,8 @@ Economy _economy;
 Prices _price;
 static PriceMultipliers _price_base_multiplier;
 
+extern int GetAmountOwnedBy(const Company *c, Owner owner);
+
 /**
  * Calculate the value of the company. That is the value of all
  * assets (vehicles, stations, shares) and money minus the loan,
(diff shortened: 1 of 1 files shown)


src/economy.cpp
```diff
@@ -115,18 +117,12 @@ Money CalculateCompanyValue(const Company *c, bool including_loan)
...
-				shares_owned++;
-			}
-		}
-
-		owned_shares_value += (CalculateCompanyValueExcludingShares(co) / 4) * shares_owned;
+		if (shares_owned > 0) owned_shares_value += (CalculateCompanyValueExcludingShares(co) / 4) * shares_owned;
 	}
 
-	return std::max<Money>(owned_shares_value + CalculateCompanyValueExcludingShares(c), 1);
+	return owned_shares_value + CalculateCompanyValueExcludingShares(c);
```
2TallTyler, line 125: Why was the max() dropped here?

SamuXarick: `CalculateCompanyValueExcludingShares` already ensures the minimum value is 1.

2TallTyler (approved):
Sure, looks like a straightforward improvement. Should probably be squashed to `Codechange: Improve performance when calculating company value` as nothing is really being fixed.

PR #9910, opened 2022-06-01
Fix: Refresh company window when switching companies

SamuXarick:
This is to refresh buy and sell share buttons states

## Motivation / Problem
Switching between companies does not properly update the buy and sell share button states

## Description
The problem is solved by invalidating all Company information windows when switching companies.

## Limitations
I'm uncertain if there is a better way to achieve the same in other means.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -114,8 +114,11 @@ void SetLocalCompany(CompanyID new_company)
 
 	_current_company = _local_company = new_company;
 
-	/* Delete any construction windows... */
-	if (switching_company) CloseConstructionWindows();
+	if (switching_company) {
+		InvalidateWindowClassesData(WC_COMPANY);
+		/* Delete any construction windows... */
+		CloseConstructionWindows();
+	}
 
 	/* ... and redraw the whole screen. */
 	MarkWholeScreenDirty();


src/company_cmd.cpp
```diff
@@ -121,6 +121,7 @@ void SetLocalCompany(CompanyID new_company)
 	MarkWholeScreenDirty();
 	InvalidateWindowClassesData(WC_SIGN_LIST, -1);
 	InvalidateWindowClassesData(WC_GOALS_LIST);
+	if (switching_company) InvalidateWindowClassesData(WC_COMPANY);
```
LordAro: any particular reason this needs to be done at the end of the function, or can it be bundled into one of the existing `if (switching_company)` blocks?

SamuXarick: It can be moved to the second block

2TallTyler (approved):
To reproduce the bug this fixes:

1. Start a network game
2. Create three companies and wait for them to be old enough to purchase shares
3. Use Company A to buy shares in Company B, making the Sell Shares button available. Leave the Company B overview window open
4. Join Company C
5. Notice that the Company B window still has the option to Sell Shares, despite Company C not owning any

I've tested and this does indeed fix that bug. 🙂

PR #9915, opened 2022-06-04
Fix #9914, 86818e5: prevent useless pathfinder run for blocked vehicles

### LINKED ISSUE #9914: [Bug]: Blocked vehicles constantly running pathfinder
### Version of OpenTTD

20220602-master-g0a3d5f5ff8, windows 10

### Expected result

A vehicle trying to enter a new tile, calls the pathfinder to get a direction, but that direction is blocked. Cache that direction to avoid calling the pathfinder again when retrying to enter the new tile.

### Actual result

A vehicle trying to enter a new tile, calls the pathfinder to get a direction, but that direction is blocked. Does not cache that direction, and calls the pathfinder again when retrying to enter the new tile.

### Steps to reproduce

This issue is similar to #7670.
In short, when a road vehicle finds itself blocked by another as it is about to enter a new tile, it calls the pathfinder for directions.

In this report I found 2 new cases where the issue was not handled by the fix provided for #7670.

That fix does work on cases where the pathfinder does provide a path cache, but on these 2 new situations, it does not cache anything, but still returns a trackdir. This isn't handled properly. Since the vehicle is blocked, it doesn't advance to the new tile, and next vehicletick it repeats the same logic. The sluggish behaviour becomes very noticeable especially when the road network the vehicles are in is huge.

[save.zip](https://github.com/OpenTTD/OpenTTD/files/8838536/save.zip)

- Load the savegames and unpause.

SamuXarick:
## Motivation / Problem
Fixes #9914 
The way the current pathfinder works (yapf), it sometimes returns only one tile/td pair which is not path cached. 86818e5 expected a path cache to always exist to prevent re-running the pathfinder but that's not always the case.

## Description
~~Unconditionally~~ push tile/td to the path cache when the vehicle is blocked by another if the pathfinder was run or its path cache was consumed. This way, the next time it asks the pathfinder for directions, it will consume from the cache, instead of doing a costly full search.

## Checklist for review

### DIFF
--- a/src/roadveh_cmd.cpp
+++ b/src/roadveh_cmd.cpp
@@ -872,18 +872,20 @@ static int PickRandomBit(uint bits)
 /**
  * Returns direction to for a road vehicle to take or
  * INVALID_TRACKDIR if the direction is currently blocked
- * @param v        the Vehicle to do the pathfinding for
- * @param tile     the where to start the pathfinding
- * @param enterdir the direction the vehicle enters the tile from
+ * @param v                  the Vehicle to do the pathfinding for
+ * @param tile               the where to start the pathfinding
+ * @param enterdir           the direction the vehicle enters the tile from
+ * @param[out] ran_pf_cache  whether the pathfinder ran or consumed from cache
  * @return the Trackdir to take
  */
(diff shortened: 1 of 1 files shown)


LordAro:
Superseded by #9917

PR #9921, opened 2022-06-12
Fix: On company closure, only sell shares back after all company value calculations are done

SamuXarick:
## Motivation / Problem
When a company closes, shares are sold. To get the value of these shares, company value calculations are issued. But the way it's currently implemented, it sells the share, increasing the value of that company. This is done in a loop.

If another share value is required involving the same company which recently got its value increased, the value is now different. In essence, what I'm trying to say is that the same company can have different values while the closure command is still running, skewing subsequent share sells.

## Description
```
	/* We're going to check if there are shares to sell, and if there are
	 * we're not selling them immediately, as that would create different
	 * company values for the same company while the command is still running.
	 * Instead, we will first gather the costs from selling the shares and
	 * their owners, and only after that we actually sell them. */
```

## Limitations
This may not be an issue at all, depending on who you ask. After all, the calculations that are in place can be seen as multiple 'Sell 25% shares' happening in a sequence, meaning that having different company values is fine. However, since the execution order of which shares are sold first is not controlled by the player, I still went with this solution.

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -325,20 +325,30 @@ void ChangeOwnershipOfCompanyItems(Owner old_owner, Owner new_owner)
 
 	assert(old_owner != new_owner);
 
+	/* We're going to check if there are shares to sell, and if there are
+	 * we're not selling them immediately, as that would create different
+	 * company values for the same company while the command is still running.
+	 * Instead, we will first gather the costs from selling the shares and
+	 * their owners, and only after that we actually sell them. */
+	Money money[MAX_COMPANIES] = {};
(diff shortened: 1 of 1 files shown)


2TallTyler:
I don't quite understand the need for this change. When a company is closed, its value disappears into the ether. And unless I'm somehow mistaken, a company's value doesn't change based on who owns shares of it, so remaining companies wouldn't care either. Who benefits from a more accurate accounting of how much a closed company is worth? (Honest question, I think I'm missing something.)

2TallTyler:
Closing this because the intention has not been explained after two months. Feel free to reopen if/when you want to explain. 😃

PR #9927, opened 2022-06-18
Change: GetQuarterlyExpenses now returns the additive inverse of the value

SamuXarick:
This ensures that the value '-1' always refer to an invalid result, while values '>= 0' as valid values.

## Motivation / Problem
GetQuarterlyExpenses could return '-1' for invalid and valid results, because expenses are retrieved as negative values.

## Description
This ensures that the value '-1' always refer to an invalid result, while values '>= 0' as valid values.

## Limitations
Old behaviour is mainained for old scripts, before 13, always a negative value for valid expenses. New behaviour is always a positive value for valid expenses.

## Checklist for review

### DIFF
--- a/bin/ai/compat_1.10.nut
+++ b/bin/ai/compat_1.10.nut
@@ -6,3 +6,10 @@
  */
 
 AILog.Info("1.10 API compatibility in effect.");
+
+/* 13 returns the additive inverse of the value. */
+AICompany._GetQuarterlyExpenses <- AICompany.GetQuarterlyExpenses
+AICompany.GetQuarterlyExpenses <- function(company, quarter)
+{
+	return -AICompany._GetQuarterlyExpenses(company, quarter);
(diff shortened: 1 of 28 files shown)


bin/ai/compat_1.10.nut
```diff
@@ -6,3 +6,10 @@
  */
 
 AILog.Info("1.10 API compatibility in effect.");
+
+/* 13 returns positive. */
```
James103:
```suggestion
/* 13 returns the additive inverse of the value. */
```

SamuXarick: thx, missed that one

glx22:
For me this is not needed. If caller makes sure the arguments are valid then it can consider `-1` as a valid value.

LordAro:
Agreed, not needed

PR #9928, opened 2022-06-19
Change: Remove scrollbar from town authority actions panel

Joel-Milligan:
## Motivation / Problem

When a company has more than 4 actions available against a town authority, you must scroll down in a short window to access the later actions.

This can be annoying if you wish to do the same action to a number of different towns, or if you repeatedly close and open the local authority window to perform the same action periodically, such as 'Fund New Building' every few days when attempting to grow a town.

This will also enable players to quickly see which actions are available to them, instead of needing to scroll through the list to see if a certain action is an option.

I believe players that commonly deal with local authorities would enjoy this feature, as the interface is simpler (removed scrollbar) and easier to use (all actions available in one click).

## Description

Remove scrollbar from the actions panel of the local authority window, and instead resize the height of the panel to show the available options.

Only 2 actions available:
image

All 8 actions available:
image

## Limitations

Redrawing the local authority window can be slow if number of available actions rapidly changes (multiple times per second).

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3482,14 +3482,14 @@ STR_LOCAL_AUTHORITY_ACTION_EXCLUSIVE_TRANSPORT                  :Buy exclusive t
 STR_LOCAL_AUTHORITY_ACTION_BRIBE                                :Bribe the local authority
 
 ###length 8
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_SMALL_ADVERTISING            :{YELLOW}Initiate a small local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a small radius around the town centre.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_MEDIUM_ADVERTISING           :{YELLOW}Initiate a medium local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a medium radius around the town centre.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_LARGE_ADVERTISING            :{YELLOW}Initiate a large local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a large radius around the town centre.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_ROAD_RECONSTRUCTION          :{YELLOW}Fund the reconstruction of the urban road network.{}Causes considerable disruption to road traffic for up to 6 months.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_STATUE_OF_COMPANY            :{YELLOW}Build a statue in honour of your company.{}Provides a permanent boost to station rating in this town.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_NEW_BUILDINGS                :{YELLOW}Fund the construction of new buildings in the town.{}Provides a temporary boost to town growth in this town.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{YELLOW}Buy 1 year's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_BRIBE                        :{YELLOW}Bribe the local authority to increase your rating, at the risk of a severe penalty if caught.{}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_SMALL_ADVERTISING            :{PUSH_COLOUR}{YELLOW}Initiate a small local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a small radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_MEDIUM_ADVERTISING           :{PUSH_COLOUR}{YELLOW}Initiate a medium local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a medium radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_LARGE_ADVERTISING            :{PUSH_COLOUR}{YELLOW}Initiate a large local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a large radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_ROAD_RECONSTRUCTION          :{PUSH_COLOUR}{YELLOW}Fund the reconstruction of the urban road network.{}Causes considerable disruption to road traffic for up to 6 months.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_STATUE_OF_COMPANY            :{PUSH_COLOUR}{YELLOW}Build a statue in honour of your company.{}Provides a permanent boost to station rating in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_NEW_BUILDINGS                :{PUSH_COLOUR}{YELLOW}Fund the construction of new buildings in the town.{}Provides a temporary boost to town growth in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 1 year's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_BRIBE                        :{PUSH_COLOUR}{YELLOW}Bribe the local authority to increase your rating, at the risk of a severe penalty if caught.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
 
 # Goal window
 STR_GOALS_CAPTION                                               :{WHITE}{COMPANY} Goals
--- a/src/town.h
+++ b/src/town.h
@@ -200,18 +200,6 @@ Town *CalcClosestTownFromTile(TileIndex tile, uint threshold = UINT_MAX);
 
 void ResetHouses();
 
-void ClearTownHouse(Town *t, TileIndex tile);
-void UpdateTownMaxPass(Town *t);
-void UpdateTownRadius(Town *t);
-CommandCost CheckIfAuthorityAllowsNewStation(TileIndex tile, DoCommandFlag flags);
-Town *ClosestTownFromTile(TileIndex tile, uint threshold);
-void ChangeTownRating(Town *t, int add, int max, DoCommandFlag flags);
-HouseZonesBits GetTownRadiusGroup(const Town *t, TileIndex tile);
-void SetTownRatingTestMode(bool mode);
-uint GetMaskOfTownActions(int *nump, CompanyID cid, const Town *t);
-bool GenerateTowns(TownLayout layout);
-const CargoSpec *FindFirstCargoWithTownEffect(TownEffect effect);
-
 /** Town actions of a company. */
 enum TownActions {
 	TACT_NONE             = 0x00, ///< Empty action set.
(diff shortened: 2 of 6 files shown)


nielsmh:
What if the player is running low screen resolution and there are many companies and all actions available? Make sure the window won't overflow the screen.
One of the original design criteria for this window was that it had to fit on a 640x480 screen when using the default sprite font. I don't know if 640x480 is still a design target for the GUI.

Joel-Milligan:
Here is the window at 640x480, with 15 companies with a reputation against the town and all actions available.
Looks to fit well within it the height of the smallest resolution.

image

Unrelated to my change, I did notice that the options menu for example does seem to overflow, so I'm not sure if 640x480 is still a design target?

image

LC-Zorg:
In my opinion, a very good initiative. :) 
I don't think it's worth considering the resolution that is only available on old monitors that just spoil your eyesight, but following Nielmsh remark, I think it could have been done differently and retained the ability to downscale if necessary window height (be aware of the enlarged GUI and font).

Currently, by increasing the size of the window, the part with the list of companies that have interacted with local authorities is enlarged. In my opinion, there is no point in enlarging the empty space. With this topic, I remembered trying to extend this window a few years ago, just to see these hidden options. Instead of the expected expansion of the list, the top, empty part has increased. Completely pointless.

Icon of pointless



So, in my opinion, the top part should only expand automatically (with a minimum height for 6-8 companies (currently 8), so that in most cases the same size of the window will be kept), while changing the size of the window by the player should increase or decrease the part of the list available actions, but not more than 8 and not less than 3 lines. The scroll bar should remain.
It is also reduce the height of the list of available actions first, and then the empty space in the part with the list of companies, but not more than to 2 lines (empty or not). When increasing the size of the window, first you would enlarge the part with the list of companies to the default 8 lines, and then the part with the list of available actions.

Anyway, even in the proposed form, it is definitely more functional than it is currently. :)

Joel-Milligan:
Following discussions in IRC, I've updated the actions menu to now show all actions, with actions that are unable to be performed greyed out and unable to be executed. Any actions disabled via Settings should not show up in the list at all, greyed out or otherwise.

image

The only limitation of this change is the lack of reasoning behind why the greyed out actions are unable to be executed.
The most common case where the company can't afford it, is not so much an issue, as the cost of the action is in the description, but other situations I'm not so sure how to go about implementing.

LC-Zorg:
This seems to make sense, but as it stands it looks very bad with this font.
Is it possible to use a gray font without black shading?
Gray without shading + Explanation

Such actions marked in gray could be inactive. Though then less experienced players would not know why and it could be irritating. Some message would be needed. Or cost in the explanation can be marked in red.

Joel-Milligan:
Removed shadow from grey text and display cost in red if you can't afford the action in new update.

image

src/lang/english.txt
```diff
@@ -3460,7 +3460,7 @@ STR_LOCAL_AUTHORITY_ACTION_EXCLUSIVE_TRANSPORT                  :Buy exclusive t
 STR_LOCAL_AUTHORITY_ACTION_BRIBE                                :Bribe the local authority
 
 ###length 8
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_SMALL_ADVERTISING            :{YELLOW}Initiate a small local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a small radius around the town centre.{}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_SMALL_ADVERTISING            : passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a small radius around the town centre.{}Cost: {CURRENCY_LONG}
```
glx22: Unintended change

Joel-Milligan: Fixed on latest push

glx22:
I think it should be possible to do something similar to `STR_ORDER_GO_TO_STATION_CAN_T_USE_STATION` to prevent string duplication. The strings would be like `{PUSH_COLOUR}{YELLOW}Description{POP_COLOUR}Cost` and you can change text colour for cost part when calling DrawString.

Joel-Milligan:
Fixed the issue with selecting actions disabled in settings.
Attempted to remove the string duplication following `STR_ORDER_GO_TO_STATION_CAN_T_USE_STATION` but doesn't work despite looking ok?

src/town_gui.cpp
```diff
@@ -110,17 +104,15 @@ struct TownAuthorityWindow : Window {
 		if (buttons != displayed_actions_on_previous_painting) this->SetDirty();
 		displayed_actions_on_previous_painting = buttons;
 
-		this->vscroll->SetCount(numact + 1);
-
-		if (this->sel_index != -1 && !HasBit(buttons, this->sel_index)) {
```
glx22, line 116: This block is important. For instance open town action window and select "transport rights", now keep the window open and disable "transport rights" in the setting, the line is removed (as expected) but still selected.
The issue is there are now 2 bitmasks, one for available actions and one for enabled actions, and this block was using the available mask and of course can't use the enabled mask because the meaning is different.

src/town_cmd.cpp
```diff
@@ -3311,7 +3332,7 @@ uint GetMaskOfTownActions(int *nump, CompanyID cid, const Town *t)
 	}
 
 	if (nump != nullptr) *nump = num;
-	return buttons;
+	return buttons & GetMaskOfTownActionsSettings(&num, cid, t); // Only return actions that aren't disabled in settings
```
glx22: `nump` might be incorrect after this bitand, but it seems `nump` is now useless anyway as it was only used to set scrollbar count.

src/town_cmd.cpp
```diff
@@ -3265,6 +3265,36 @@ static TownActionProc * const _town_action_proc[] = {
...
 };
 
+/**
+ * Get a list of all actions allowable by settings.
+ * @param nump if not nullptr add put the number of available actions in it
+ * @param cid the company that is querying the town
+ * @param t the town that is queried
+ * @return bitmasked value of enabled actions
+ */
+uint GetMaskOfTownActionsSettings(int* nump, CompanyID cid, const Town* t)
```
glx22: I think all this can be done via a private window variable and `OnInvalidateData()` handler. That way it's a cached value easily usable anywhere in the window code.

glx22 (commented):
Some things I noticed while reading the diff.
And with the cache idea, `GetNthSetBit()` could make use of it internally (removing the need for mask arg)

src/town_cmd.cpp
Joel-Milligan: Implemented as a private window var. Not sure about the `OnInvalidateData()` handler, didn't seem to be triggered by settings changes. However cached var does prevent recalculations within each paint.

src/town_cmd.cpp
Joel-Milligan: Removed nump, no longer needed.

src/town_gui.cpp
Joel-Milligan: Now deselecting the option if disabled in settings while window is open.

glx22:
You can use `InvalidateWindowClassesData(WC_TOWN_AUTHORITY[, optional data])` to replace `SetWindowClassesDirty(WC_TOWN_AUTHORITY)` in https://github.com/OpenTTD/OpenTTD/blob/master/src/table/settings/economy_settings.ini#L126 and friends.
(I noticed a `InvalidateWindowData(WC_TOWN_AUTHORITY, town_id)` call in `CmdTownRating()`, but the window has no handler for it yet).
And please use `this->` when referencing own members.

src/town_cmd.cpp
Joel-Milligan: Following this point in the right direction, I was able to figure out how to set the callback on setting change. I looked everywhere but the INI file :P.

src/town_gui.cpp
```diff
@@ -71,56 +68,69 @@ struct TownAuthorityWindow : Window {
...
+	 * @return Bitmask of actions enabled in the settings.
+	 */
+	static TownActions GetEnabledActions()
+	{
+		TownActions enabled = TACT_ALL;
+
+		if (!_settings_game.economy.fund_roads) ClrBit(enabled, 3);
+		if (!_settings_game.economy.fund_buildings) ClrBit(enabled, 5);
+		if (!_settings_game.economy.exclusive_rights) ClrBit(enabled, 6);
+		if (!_settings_game.economy.bribe) ClrBit(enabled, 7);
```
glx22:
```suggestion
		if (!_settings_game.economy.fund_roads) CLRBITS(enabled, TACT_ROAD_REBUILD);
		if (!_settings_game.economy.fund_buildings) CLRBITS(enabled, TACT_FUND_BUILDINGS);
		if (!_settings_game.economy.exclusive_rights) CLRBITS(enabled, TACT_BUY_RIGHTS);
		if (!_settings_game.economy.bribe) CLRBITS(enabled, TACT_BRIBE);
```
To avoid magic values.

src/town_gui.cpp
```diff
@@ -698,7 +727,6 @@ struct TownDirectoryWindow : public Window {
 
 			this->towns.shrink_to_fit();
 			this->towns.RebuildDone();
-			this->vscroll->SetCount((uint)this->towns.size()); // Update scrollbar as well.
```
LordAro: What's this change doing here?

src/town_gui.cpp
```diff
@@ -297,6 +316,16 @@ struct TownAuthorityWindow : Window {
 	{
 		this->SetDirty();
 	}
+
+	void OnInvalidateData(int data = 0, bool is_gui = true) override
```
LordAro: all other OnInvalidateData functions call the second parameter `gui_scope`. Should probably match...

src/town_gui.cpp
Joel-Milligan: Removed that line, because we no longer use the vscroll in this window, so the line does nothing.

LordAro: I don't follow. This window is the town directory window, which definitely still uses the scrollbar

nielsmh: To be clear, _this_ is the scrollbar you made non-functional by deleting that line:
image

And to be extra clear, this is the same generated world, just with the window made larger to fit more towns. It still has many more, but you can't view the list because the scroll bar doesn't work.
image

Joel-Milligan: Yep very true, got lost in the sauce and didn't realise what object I was working in. Fixed.

src/town_gui.cpp
```diff
@@ -194,33 +231,15 @@ struct TownAuthorityWindow : Window {
...
-
-				if (--pos < 0) {
-					DrawString(r.left + WD_FRAMERECT_LEFT, r.right - WD_FRAMERECT_RIGHT, y, STR_LOCAL_AUTHORITY_ACTIONS_TITLE);
-					y += FONT_HEIGHT_NORMAL;
-				}
-
-				for (int i = 0; buttons; i++, buttons >>= 1) {
-					if (pos <= -5) break; ///< Draw only the 5 fitting lines
+					Money action_cost = _price[PR_TOWN_ACTION] * _town_action_costs[this->sel_index] >> 8;
+					bool affordable = action_cost < Company::Get(_local_company)->money;
```
PeterN: This doesn't check that `_local_company` is a valid company.

2TallTyler: I'd imagine there's a `Company::GetIfValid()` to use here.

2TallTyler (commented):
@Joel-Milligan Are you interested in finishing this PR? I'm going through the backlog for low-hanging fruit and this is on my list.

Joel-Milligan:
> @Joel-Milligan Are you interested in finishing this PR? I'm going through the backlog for low-hanging fruit and this is on my list.

Yep definitely interested in finishing this one off, just fell off my radar. I'll fix up the feedback tonight.

src/town_gui.cpp
Joel-Milligan: Using `Company::GetIfValid()` instead of `Company::Get()` in latest commit.
Under what circumstances would `_local_company` not be a valid company? I want to see if I can produce any issues under that situation.

LordAro:
(commit checker is currently failing as the PR is too old, and it can't find the correct merge-base as a result - rebase to fix)

src/town_gui.cpp
```diff
@@ -185,6 +195,33 @@ struct TownAuthorityWindow : Window {
 		}
 	}
 
+	/** Draws the contents of the actions panel. May re-initialise window to resize panel, if the list does not fit. */
+	void DrawActions()
+	{
+		NWidgetBase* nwid = this->GetWidget<NWidgetBase>(WID_TA_COMMAND_LIST);
```
LordAro: codestyle: `*` goes on the variable

src/town_gui.cpp
```diff
@@ -185,6 +195,33 @@ struct TownAuthorityWindow : Window {
...
+			/* Don't show actions if disabled in settings. */
+			if (!HasBit(this->enabled_actions, i)) continue;
+
+			/* Set colour of action based on ability to execute and if selected. */
+			TextColour action_colour = TC_GREY | TC_NO_SHADE;
+			if (HasBit(this->available_actions, i)) action_colour = TC_ORANGE;
+			if (this->sel_index == i) action_colour = TC_WHITE;
+
+			DrawString(left + WD_FRAMERECT_LEFT, right - WD_FRAMERECT_RIGHT, y,
+				STR_LOCAL_AUTHORITY_ACTION_SMALL_ADVERTISING_CAMPAIGN + i, action_colour);
```
LordAro: codestyle: no need for a linebreak here (or, if there is, we use 2 tabs for continuation)

src/town_gui.cpp
```diff
@@ -298,6 +317,16 @@ struct TownAuthorityWindow : Window {
 	{
 		this->SetDirty();
 	}
+
+	void OnInvalidateData(int data = 0, bool gui_scope = true) override
+	{
+		if (!gui_scope) return;
+
+		this->enabled_actions = GetEnabledActions();
```
LordAro: missing `this->` on method

LordAro (commented):
Minor nitpicks

2TallTyler (dismissed):

src/town.h
```diff
@@ -200,6 +200,28 @@ Town *CalcClosestTownFromTile(TileIndex tile, uint threshold = UINT_MAX);
...
+	TACT_FUND_BUILDINGS = 0x20, ///< Fund new buildings.
+	TACT_BUY_RIGHTS = 0x40, ///< Buy exclusive transport rights.
+	TACT_BRIBE = 0x80, ///< Try to bribe the council.
+
+	TACT_COUNT = 8,    ///< Number of available town actions.
+
+	TACT_ADVERTISE = TACT_ADVERTISE_SMALL | TACT_ADVERTISE_MEDIUM | TACT_ADVERTISE_LARGE, ///< All possible advertising actions.
+	TACT_CONSTRUCTION = TACT_ROAD_REBUILD | TACT_BUILD_STATUE | TACT_FUND_BUILDINGS, ///< All possible construction actions.
+	TACT_FUNDS = TACT_BUY_RIGHTS | TACT_BRIBE, ///< All possible funding actions.
+	TACT_ALL = TACT_ADVERTISE | TACT_CONSTRUCTION | TACT_FUNDS, ///< All possible actions.
```
michicc: Not a bug in any way, but the enum had some nice aligned formatting before the move.

2TallTyler (approved):
Indentation fixed

PR #8444, opened 2020-12-27
Feature: Close adjacent level crossings as if they were one large crossing

Eddi-z:
## Motivation / Problem

Level crossings on parallel tracks can be dangerous, especially for articulated road vehicles, or if multiple road vehicles follow each other. Often, a vehicle is waiting for a level crossing to open, when it suddenly gets struck in its "tail" by another train.

## Description

Updated patch from https://www.tt-forums.net/viewtopic.php?f=33&t=46091

This patch tries to avoid the situation that a vehicle gets stuck inbetween two tracks by closing all level crossings simultaneously, and instead allowing road vehicles to leave a crossing they already entered. Best combined with a path signal some tiles ahead of a level crossing, to give vehicles time to leave if a train approaches.

## Limitations

on loading an old savegame, road vehicles that were previously waiting for a crossing to open, may now immediately start up, causing them to run straight into a train. some savegame conversion update is needed.

~~also, a call to `SndPlayTileFx(SND_0E_LEVEL_CROSSING, tile);` fell off and i don't remember where it went, or why i removed it.~~
**Update:** level crossing sound is now changed to only appear when a train approaches the crossing, instead of sometimes when it is being closed.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #9931, opened 2022-06-23
Feature: Multi-track level crossings

2TallTyler:
## Motivation / Problem

Multiple-track level crossings in vanilla OpenTTD are unsafe since road vehicle protection is per-track and not for the entire crossing. Vehicles will happily stop on the near track to wait for a train to pass on the far track, and then get smushed by a train.

## Description

single_sided
_(NewGRFs: U&RaTT roads and U&ReRMM rails, both by Ufiby)_

JGRPP includes several improvements to level crossings. This PR only includes some features, starting with Eddi's original [patch](https://www.tt-forums.net/viewtopic.php?p=836749#p836749) and various improvements by JGR including code improvements and a feature to only draw crossing signals (when sprites are provided by NewGRF) on the edges of a multi-track crossing.

What this PR includes:

- When an approaching train causes a level crossing to be barred, adjacent tracks are also blocked.
- Road vehicles are always allowed to exit blocked crossings even if that means crossing multiple tracks.
- When using a NewGRF railtype which includes custom crossing signal sprites, only the crossing signals at the edges of the crossing are drawn.

What this PR does NOT include from JGRPP:
- Safer level crossings (trains cannot reserve a path over a level crossing unless it is clear of road vehicles) ([commit](https://github.com/JGRennison/OpenTTD-patches/commit/0073c4e71eefa6ff499082ff1fc05d40c983f232))
- Road vehicles cannot be manually stopped on level crossings ([commit](https://github.com/JGRennison/OpenTTD-patches/commit/0acb4fdd2f94fa4a241600500f4b08552188e10e))
- Settings to enable or disable multi-track crossings

There is a saveload conversion for savegames where road vehicles may be stopped on a level crossing waiting for a train. Any road vehicle stopped on a level crossing is teleported to the nearest depot, if one exists. (If no depot is reachable, it'll still get smashed.) The conversion then iterates through every tile on the map to update level crossings and block adjacent tiles if necessary. I tried looking only at tiles reserved by trains, but couldn't find an easy way to do this. Iterating through every tile is a bit more expensive but as a one-time saveload conversion I hope it's okay.

Closes #8444 (draft PR of the same patch)

## Limitations

- ~~In savegames where a road vehicle is sitting on an open level crossing, stopped at a barred level crossing on the next track (behavior which this PR bans), if another train reserves the crossing where this vehicle sets, it will start driving and crash into the side of the train on the next track.~~ Now fixed. Savegame for demonstration: [rv_smash.zip](https://github.com/OpenTTD/OpenTTD/files/8970738/rv_smash.zip)
- During saveload conversion, if a vehicle stopped on a level crossing cannot find a depot, it will not get teleported and will likely get smashed. This should be rare, but is still possible.
- Vanilla level crossings are drawn using a single sprite, so a NewGRF railtype which includes custom crossing signal sprites is required to only show crossing signals on the crossing edges
- Level crossings are still unsafe if you're using block signals, or if your road vehicles take a long time to clear the crossing. Path signals should be placed far in advance of the crossing to cause it to activate and allow road vehicles to clear the crossing before the train arrives. This is existing behavior, with the only change being that the more tracks you have, the longer the time it takes to clear.

## Checklist for review

### DIFF
--- a/src/road_cmd.cpp
+++ b/src/road_cmd.cpp
@@ -380,6 +380,8 @@ static CommandCost RemoveRoad(TileIndex tile, DoCommandFlag flags, RoadBits piec
 			uint len = GetTunnelBridgeLength(other_end, tile) + 2;
 			cost.AddCost(len * 2 * RoadClearCost(existing_rt));
 			if (flags & DC_EXEC) {
+				MarkDirtyAdjacentLevelCrossingTiles(tile, GetCrossingRoadAxis(tile));
+
 				/* A full diagonal road tile has two road bits. */
 				UpdateCompanyRoadInfrastructure(existing_rt, GetRoadOwner(tile, rtt), -(int)(len * 2 * TUNNELBRIDGE_TRACKBIT_FACTOR));
 
@@ -780,6 +782,7 @@ CommandCost CmdBuildRoad(DoCommandFlag flags, TileIndex tile, RoadBits pieces, R
 				MakeRoadCrossing(tile, company, company, GetTileOwner(tile), roaddir, GetRailType(tile), rtt == RTT_ROAD ? rt : INVALID_ROADTYPE, (rtt == RTT_TRAM) ? rt : INVALID_ROADTYPE, town_id);
 				SetCrossingReservation(tile, reserved);
 				UpdateLevelCrossing(tile, false);
+				MarkDirtyAdjacentLevelCrossingTiles(tile, GetCrossingRoadAxis(tile));
 				MarkTileDirtyByTile(tile);
 			}
 			return CommandCost(EXPENSES_CONSTRUCTION, 2 * RoadBuildCost(rt));
@@ -1703,7 +1706,42 @@ static void DrawTile_Road(TileInfo *ti)
 				SpriteID rail = GetCustomRailSprite(rti, ti->tile, RTSG_CROSSING) + axis;
 				DrawGroundSprite(rail, pal);
 
-				DrawRailTileSeq(ti, &_crossing_layout, TO_CATENARY, rail, 0, PAL_NONE);
+				const Axis road_axis = GetCrossingRoadAxis(ti->tile);
+				const DiagDirection dir1 = AxisToDiagDir(road_axis);
+				const DiagDirection dir2 = ReverseDiagDir(dir1);
+				uint adjacent_diagdirs = 0;
+				for (DiagDirection dir : { dir1, dir2 }) {
+					const TileIndex t = TileAddByDiagDir(ti->tile, dir);
+					if (t < MapSize() && IsLevelCrossingTile(t) && GetCrossingRoadAxis(t) == road_axis) {
+						SetBit(adjacent_diagdirs, dir);
+					}
+				}
+
+				switch (adjacent_diagdirs) {
+					case 0:
+						DrawRailTileSeq(ti, &_crossing_layout, TO_CATENARY, rail, 0, PAL_NONE);
+						break;
+
+					case (1 << DIAGDIR_NE):
+						DrawRailTileSeq(ti, &_crossing_layout_SW, TO_CATENARY, rail, 0, PAL_NONE);
+						break;
+
+					case (1 << DIAGDIR_SE):
+						DrawRailTileSeq(ti, &_crossing_layout_NW, TO_CATENARY, rail, 0, PAL_NONE);
+						break;
+
+					case (1 << DIAGDIR_SW):
+						DrawRailTileSeq(ti, &_crossing_layout_NE, TO_CATENARY, rail, 0, PAL_NONE);
+						break;
+
+					case (1 << DIAGDIR_NW):
+						DrawRailTileSeq(ti, &_crossing_layout_SE, TO_CATENARY, rail, 0, PAL_NONE);
+						break;
+
+					default:
+						/* Show no sprites */
+						break;
+				}
 			} else if (draw_pbs || tram_rti != nullptr || road_rti->UsesOverlay()) {
 				/* Add another rail overlay, unless there is only the base road sprite. */
 				PaletteID pal = draw_pbs ? PALETTE_CRASH : PAL_NONE;
@@ -2033,7 +2071,16 @@ static TrackStatus GetTileTrackStatus_Road(TileIndex tile, TransportType mode, u
 					if (side != INVALID_DIAGDIR && axis != DiagDirToAxis(side)) break;
 
 					trackdirbits = TrackBitsToTrackdirBits(AxisToTrackBits(axis));
-					if (IsCrossingBarred(tile)) red_signals = trackdirbits;
+					if (IsCrossingBarred(tile)) {
+						red_signals = trackdirbits;
+						auto mask_red_signal_bits_if_crossing_barred = [&](TileIndex t, TrackdirBits mask) {
+							if (IsLevelCrossingTile(t) && IsCrossingBarred(t)) red_signals &= mask;
+						};
+						/* Check for blocked adjacent crossing to south, keep only southbound red signal trackdirs, allow northbound traffic */
+						mask_red_signal_bits_if_crossing_barred(TileAddByDiagDir(tile, AxisToDiagDir(axis)), TRACKDIR_BIT_X_SW | TRACKDIR_BIT_Y_SE);
+						/* Check for blocked adjacent crossing to north, keep only northbound red signal trackdirs, allow southbound traffic */
+						mask_red_signal_bits_if_crossing_barred(TileAddByDiagDir(tile, ReverseDiagDir(AxisToDiagDir(axis))), TRACKDIR_BIT_X_NE | TRACKDIR_BIT_Y_NW);
+					}
 					break;
 				}
 
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -24,6 +24,7 @@
 #include "../string_func.h"
 #include "../date_func.h"
 #include "../roadveh.h"
+#include "../roadveh_cmd.h"
 #include "../train.h"
 #include "../station_base.h"
 #include "../waypoint_base.h"
@@ -56,7 +57,6 @@
 #include "../disaster_vehicle.h"
 #include "../ship.h"
 #include "../water.h"
-
 
 #include "saveload_internal.h"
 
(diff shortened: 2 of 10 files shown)


Eddi-z:
> In savegames where a road vehicle is sitting on an open level crossing, stopped at a barred level crossing on the next track

i always felt uneasy about this, and this is why i never really felt confident submitting it as a PR

2TallTyler:
Realistically, any network unsafe enough for this to occur would almost certainly be experiencing road/rail collisions on a regular basis. It's not ideal but I can't think of a way to implement this without such a side effect or some other undesirable feature like a default-off setting. I think it's a small price to pay.

LC-Zorg:
For me, this is one of the best features present in JGRPP that is missing from the official version. If I had made a wish list, this feature would definitely be ranked very high. :)

Regarding "Safety level crossings", good thing you didn't combine it. This is an annoying feature that additionally makes it possible to block train traffic, what was impossible ever before.
Sejfti lewel krosing mini
A good alternative to such a function could be to optionally prevent the construction of track crossings with roads of different companies. Then accidents can always be prevented.

Regarding the addition of options, I think it would be redundant, if not to say useless. Most likely, after turning it on, no one will ever want to turn it off. :) As for the collisions when first starting the games after adding the features, I think they will be something rare or very rare and as you said they will not surprise the player because with this style of building they are something common. However, if the player absolutely wants to avoid them, he will also be able to do this by stopping selected vehicles before run the game, so I don't see the problem here.

Eddi-z:
> As for the collisions when first starting the games after adding the features, I think they will be something rare or very rare and as you said they will not surprise the player because with this style of building they are something common. 

I strongly disagree with this reasoning. It is something rare and acceptable in a world where people actively seek out experimental patches, but in a general release, with potentially millions of people loading up their old savegames, they will be totally not expecting this kind of mass failure accross their whole map. this will almost certainly cause a maintenance nightmare. we should not run into this blindly.

JGRennison:
Teleporting vehicles would be excessively fiddly when it comes to articulated vehicles, and road layouts including things like junctions, corners, tunnel/bridges, etc.

If a road vehicle is already part way across a level crossing when the tile that they're on also becomes barred, very likely the vehicle will be destroyed imminently with or without this change.
That said, you could modify GetTileTrackStatus_Road to check for the presence of train vehicles on the tile, at the risk of some performance cost.

By far the easiest and most robust solution to the problem of new features and old savegames is a setting.

Eddi-z:
one other alternative would be to pause the game and display a warning to the player that he should manually stop the road vehicles before unpausing.

2TallTyler:
I decided to handle this in the saveload conversion, which does two things:

1. Any road vehicle stopped on a level crossing is teleported to the nearest depot, if one exists. (If no depot is reachable, it'll still get smashed.)
2. We iterate through every tile on the map to update level crossings and block adjacent tiles if necessary. I tried looking only at tiles reserved by trains, but couldn't find an easy way to do this. Iterating through every tile is a bit more expensive but as a one-time saveload conversion I hope it's okay.

LC-Zorg:
> 1. Any road vehicle stopped on a level crossing is teleported to the nearest depot, if one exists. (If no depot is reachable, it'll still get smashed.)

This may not be a good solution. Does teleportation only take place to a depot that the vehicle can reach on wheels? Is there a distance limit? If the answer to any of these questions is no, it may cause more confusion than a collision with train. This can be really annoying when the vehicle suddenly ends up on the other side of the map or in a place where it cannot find its way. In my opinion, the solution suggested by True Brain was better. At least in terms of gameplay, it would not confuse players.

By the way, I'm still not convinced that these safeguards are needed at all. It doesn't matter to me, but I just don't see much sense in it. Really, if these accidents were to be especially noticeable after launching an updated game, there would have to be a lot of such dangerous railroad crossings and tons of vehicles on the map. But even if the player has such a save, accidents on such a map are something completely common and whether there is 6 accidents per week instead of 3 will not make much difference. And it certainly won't be a reason to complain if, after increasing the number of collisions in first few days, they will be significantly reduced later in the game. It is certainly not possible for dozens of collisions to suddenly occur after updating the game if they were rare before. 

Out of curiosity, to check it and prove my thesis, I made a test save.
**[Level Crossing Crash Spam STD, 2010-01-01.zip](https://github.com/OpenTTD/OpenTTD/files/9035721/Level.Crossing.Crash.Spam.STD.2010-01-01.zip)**
- Map 1024x256
- More than 4,000 vehicles
- 68 double crossings, 8 quad crossings, some with very heavy traffic
- All passages are safe in the JGRPP version with the Close adjacent level crossing function enabled

**Results:**
I run the first few months of save in regular version 12.2. Not much is happening, because the trains are just about to go hunting.
**1. In the case of continuing the game in version 12.2**, there were 127 accidents in the next year, which means that the message about the collision appeared on average every 3 days, i.e. practically without interruption.
**2. If the game is continued in the JGRPP version with the function disabled**, the number of collisions is huge, much greater than in the regular version of the game - 628 vehicles are destroyed during the one year! Massive collisions of many vehicles occur very often.
**3. In the case of starting the save from the beginning in the JGRPP version with the function enabled**, there were only 15 accidents in 10 years. With a few minor tweaks, there would have been any collision.

### Game update simulation:
After the same few months, in version 12.2 I saved the game and restarted it.
**4. In the case of continuing the game in version 12.2** (details for point 1)
- there are 3 collisions in the first week
- there are 17 collisions occur in the first month
- there are 127 collisions in the year

**5. In the case of continuing the game in the JGRPP version with the function of closing adjacent level crossing enabled**
- there are 6 collisions in the first week
- there are 10 collisions in the first month
- there are 20 collisions during the year (the next 10 collisions take place on a 4-track crossing, and the reason is that vehicles cannot avoid wrecks at crossings)
- there are no collisions in the following year

Obviously, these are results for a game where distances between signals are kept safe. In the case of the average, less familiar player, the effect of enabling the function will not be that great, but there should be a more or less noticeable improvement. It is also worth taking into account that practically no one builds in such a way that in his games there is another collision every now and then. Usually, there are less than 10 of them per year. Scaling the above results, it can be assumed that in the case of an average player's save, these collisions will also be less in the first year. I don't think that's a reason to complain. :)

Moth-Tolias:
is it not possible to just teleport the vehicles off of the crossing, rather than all the way to a depo?

Eddi-z:
The main problem with teleporting vehicles, as hinted earlier, is making sure whether the place where you put them is a valid location, which is tricky when you have articulated vehicles. do you attempt to put them along a road? can you put them in a wormhole state (like, leaving a tunnel)? putting them in a depot avoids these issues.

2TallTyler:
Depot teleportation finds a depot using the same function as when you press the "Go to depot" button on a road vehicle, so yes, it is always reachable (although it could potentially be stuck on the wrong side of a one-way road post-teleportation if the player uses these in odd ways). I've added a distance limit of 512 tiles (Manhattan distance) for teleportation.

My previous implementation didn't teleport articulated vehicles properly, thus the latest push to fix this. It works in my test savegames using Road Hog vehicles, but the save provided by @LC-Zorg produces a crash: [crash20220708164528.zip](https://github.com/OpenTTD/OpenTTD/files/9073812/crash20220708164528.zip)

I'm not sure why, since it's not an assertion failure, and even on vanilla 12.2 my 2CC in NML sprites are corrupted and replaced by ? symbols. Not sure if it's that or something else. I'm marking this as a draft until this is fixed and it's truly ready for review.

2TallTyler:
Further testing revealed that the crash in @LC-Zorg's savegame is (I think) unrelated to this PR, and just triggered a crash because I'm running a debug build. `RoadStop.GetOccupied()` threw a read access violation because `this was nullptr.`

I did fix one bug where I didn't initialize `reverse` with a value and rebased. Ready for code review and for people to try and break teleportation! Here's a more comprehensive savegame: [RVSmash3.zip](https://github.com/OpenTTD/OpenTTD/files/9557437/RVSmash3.zip)

2TallTyler:
No changes, just rebased to fix a merge conflict in `saveload.h`.

src/road_cmd.cpp
```diff
@@ -1703,7 +1706,42 @@ static void DrawTile_Road(TileInfo *ti)
...
+				uint adjacent_diagdirs = 0;
+				for (DiagDirection dir : { dir1, dir2 }) {
+					const TileIndex t = TileAddByDiagDir(ti->tile, dir);
+					if (t < MapSize() && IsLevelCrossingTile(t) && GetCrossingRoadAxis(t) == road_axis) {
+						SetBit(adjacent_diagdirs, dir);
+					}
+				}
+
+				switch (adjacent_diagdirs) {
+				case 0:
```
michicc: Wrong indentation.

src/saveload/afterload.cpp
```diff
@@ -57,12 +57,12 @@
 #include "../ship.h"
 #include "../water.h"
 
-
 #include "saveload_internal.h"
 
 #include <signal.h>
 
 #include "../safeguards.h"
+#include <roadveh_cmd.h>
```
michicc: Wrong position for the #include and also wrong style (<> vs. "").

src/table/road_land.h
```diff
@@ -53,6 +53,54 @@ static const DrawTileSprites _crossing_layout = {
 	{0, PAL_NONE}, _crossing_layout_ALL
 };
 
+static const DrawTileSeqStruct _crossing_layout_SW_ALL[] = {
+	//TILE_SEQ_LINE(2, PAL_NONE,  0,  0, 3, 3)
+	//TILE_SEQ_LINE(4, PAL_NONE,  0, 13, 3, 3)
```
michicc: Is there a deeper meaning in these commented out lines? If yes, add a text comment, if not, they should just vanish.

src/train_cmd.cpp
```diff
@@ -1674,29 +1674,92 @@ static bool TrainApproachingCrossing(TileIndex tile)
...
 	assert(IsLevelCrossingTile(tile));
+	bool set_barred;
 
-	/* reserved || train on crossing || train approaching crossing */
-	bool new_state = HasCrossingReservation(tile) || HasVehicleOnPos(tile, nullptr, &TrainOnTileEnum) || TrainApproachingCrossing(tile);
+	/* We force the crossing to be barred when an adjacent crossing is barred, otherwise let it decide for itself. */
+	if (force_barred) {
+		set_barred = force_barred;
+	} else {
+		set_barred = CheckLevelCrossing(tile);
```
michicc: `set_barred = forced_barred || CheckLevelCrossing(tile);`? (|| does short-circuit evaluation)

michicc (dismissed):

2TallTyler:
I don't think it's necessary to allow vehicles to play Frogger on unsafe crossings. The signals just need to be placed further away to give vehicles time to cross. If the crossing is too wide and the signals can't be placed far enough away due to signal spacing required by heavy traffic, it's probably time for a bridge.

Bouke:
> This would not be easy to add. I don't think it's necessary to allow vehicles to play Frogger on unsafe crossings. (…)

I agree that we shouldn’t have to avoid all sorts of crashes. However some smarts wouldn’t hurt. Maybe road vehicles should turn around if they’re on a multi-track crossing and turning around would be safer:
- there are (more) (reserved) crossings ahead
- turning around is the shortest number of tiles to leave the multi-track crossing

2TallTyler:
"Some smarts" is more difficult than it sounds, given that the vehicles currently have no logic for checking blocked crossing tiles by themselves. The road tiles themselves set the valid directions across themselves based on the crossing's status. All this PR does is propagate that status to the entire crossing (well, a bit more nuanced: the first tile gets blocked and vehicles in the crossing can continue).

Turning around definitely isn't possible without significant changes to vehicle code. Stopping vehicles short of blocked crossings would be possible but I'm not sure it's worth the effort. Just put the signals further away or don't cross so many tracks at once.

michicc (approved):
😇

PR #9953, opened 2022-07-17
Fix: Trains only considered railtype-poweredness, and mostly ignored railtype-compatibility.

frosch123:
## Motivation / Problem

Consider Andy's train:

| Vehicle | Compatible | Powered |
| --- | --- | --- |
| Engine 1 | RAIL, ELRL, HIGH | ELRL |
| Engine 2 | RAIL, HIGH | HIGH |

Expected behavior:
* Train can only run on HIGH tracks.
    * All engines are compatible with RAIL and HIGH; but only some with ELRL.
    * Some engines are powered on ELRL, some on HIGH; but none on just RAIL.

Actual behavior:
* Compatibilty is ignored.
* The train runs on ELRL and HIGH tracks.

Secondary issue:
* If different engines of a train are powered on different railtypes, a train may get stranded when transitioning between railtypes.
    * Engines in the front may lose power when entering the new railtype.
    * Engines in the back may not yet have entered the new railtype, so are not powered either.
* If trains have enough momentum they can safely transition these power gaps.
* If a train stops due to signals, breakdowns, ... it is stuck with zero power.

## Description

Primary issue:
* Trains are allowed to enter tracks, if
    * All vehicles of the consist are compatible.
    * At least one engine is powered.

Secondary issue:
* Passengers can now be drafted to push the train, providing a minimum of 1 hp in all cases.

## Limitations

PBS path reservation/freeing sometimes uses `v->compatible_railtypes` instead of `GetRailTypeInfo(v->railtype)->compatible_railtypes`. This needs checking.

## Checklist for review

### DIFF
--- a/src/ground_vehicle.cpp
+++ b/src/ground_vehicle.cpp
@@ -56,12 +56,14 @@ void GroundVehicle<T, Type>::PowerChanged()
 
 	this->gcache.cached_air_drag = air_drag + 3 * air_drag * number_of_parts / 20;
 
+	if (total_power == 0) {
+		/* Ensure that vehicles are never completely stuck */
+		total_power = 1;
+	}
+
 	max_te *= GROUND_ACCELERATION; // Tractive effort in (tonnes * 1000 * 9.8 =) N.
 	max_te /= 256;  // Tractive effort is a [0-255] coefficient.
 	if (this->gcache.cached_power != total_power || this->gcache.cached_max_te != max_te) {
-		/* Stop the vehicle if it has no power. */
-		if (total_power == 0) this->vehstatus |= VS_STOPPED;
-
 		this->gcache.cached_power = total_power;
 		this->gcache.cached_max_te = max_te;
 		SetWindowDirty(WC_VEHICLE_DETAILS, this->index);
--- a/src/rail_type.h
+++ b/src/rail_type.h
@@ -49,6 +49,7 @@ enum RailTypes : uint64 {
 	RAILTYPES_ELECTRIC = 1 << RAILTYPE_ELECTRIC, ///< Electrified rails
 	RAILTYPES_MONO     = 1 << RAILTYPE_MONO,     ///< Monorail!
 	RAILTYPES_MAGLEV   = 1 << RAILTYPE_MAGLEV,   ///< Ever fast maglev
+	RAILTYPES_ALL      = UINT64_MAX,             ///< All railtypes
 	INVALID_RAILTYPES  = UINT64_MAX,             ///< Invalid railtypes
 };
 DECLARE_ENUM_AS_BIT_SET(RailTypes)
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -114,7 +114,8 @@ void Train::ConsistChanged(ConsistChangeFlags allowed_changes)
 	const RailVehicleInfo *rvi_v = RailVehInfo(this->engine_type);
 	EngineID first_engine = this->IsFrontEngine() ? this->engine_type : INVALID_ENGINE;
 	this->gcache.cached_total_length = 0;
-	this->compatible_railtypes = RAILTYPES_NONE;
+	RailTypes all_compatible = RAILTYPES_ALL;
+	RailTypes any_powered = RAILTYPES_NONE;
 
 	bool train_can_tilt = true;
 	int min_curve_speed_mod = INT_MAX;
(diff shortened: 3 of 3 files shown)


frosch123:
@michicc Can you please have a look at the implications for PBS. Why do TryReserveSafeTrack/FollowTrainReservation/FreeTrainTrackReservation use `GetRailTypeInfo(v->railtype)->compatible_railtypes`?

JGRennison:
> @michicc Can you please have a look at the implications for PBS. Why do TryReserveSafeTrack/FollowTrainReservation/FreeTrainTrackReservation use `GetRailTypeInfo(v->railtype)->compatible_railtypes`?

I'm not really michicc, but I've looked into this previously...
The PBS reservation needs to end at a safe waiting point. A change of track type where the current train is not powered on the far side is not a valid place to end the reservation as another train of a different compatible rail type could be powered on both of the rail types and come from the other direction. In particular if the tile on the far side is already reserved the reservations would become joined from the point of view of one side but not the other, which is bad news.
Unfortunately this still sometimes doesn't work as some rail type GRFs define incorrect compatibility tables (i.e. where A is compatible with B but B is not compatible with A), so this can still cause train crashes and other signalling problems in vanilla if you're unlucky.

andythenorth:
> Unfortunately this still sometimes doesn't work as some rail type GRFs define incorrect compatibility tables (i.e. where A is compatible with B but B is not compatible with A), so this can still cause train crashes and other signalling problems in vanilla if you're unlucky.

Oops.  I wasn't aware that compatibility A -> B also required compatibility B -> A :o

Some of the test cases I'm using in grf might be invalid on this basis?

E.g. I have the literal case described: railtype A compatible with B; railtype B is not compatible with A.

I believe the standardised scheme is also created on this basis (the axle load stuff is hierarchical).

planetmaker:
> > Unfortunately this still sometimes doesn't work as some rail type GRFs define incorrect compatibility tables (i.e. where A is compatible with B but B is not compatible with A), so this can still cause train crashes and other signalling problems in vanilla if you're unlucky.
> 
> Oops. I wasn't aware that compatibility A -> B also required compatibility B -> A fearful
> 
> Some of the test cases I'm using in grf might be invalid on this basis?
> 
> E.g. I have the literal case described: railtype A compatible with B; railtype B is not compatible with A.
> 
> I believe the standardised scheme is also created on this basis (the axle load stuff is hierarchical).

Is this British sarcasm served dry but the dead pan voice gone missing in writing? :)

The you're sure right that A->B does not imply B->A with the classical example of standard RAIL and electrified rail ELRL

frosch123:
Examples for compatibility:
* Gauges:
    * NARROW and RAIL are not compatible to each other.
    * NARROW and RAIL are compatible to MIXED.
    * If there were MIXED vehicles, they probably were also compatible to NARROW and RAIL.
    * So, here "compatibility" is symmetric, but not transitive.
* Clearance:
    * RAIL is compatible to HIGH.
    * HIGH is not compatible to RAIL.
    * So, here "compatible" is intended to be asymmetric.

In all cases, "enforcing symmetry" is not job of the NewGRF, since it does not know about other NewGRF.
OpenTTD needs to either support asymmetric compatibility, or enforce symmetry on load.

I guess I need to thing more, whether asymmetric compatibilty can be done for PBS, or whether HIGH cannot be implemented.

andythenorth:
> Is this British sarcasm served dry but the dead pan voice gone missing in writing? :)

Nah, it's straight questions 😉   Either the spec is wrong, or my test cases are wrong, or railtypes are being used wrong, or any combination of those.

> The you're sure right that A->B does not imply B->A with the classical example of standard RAIL (A) and electrified rail ELRL (B)

As per Frosch RAIL and ELRL are mutually compatible, as required.  But ELRL is not powered on RAIL etc etc 😄

michicc:
The intended logic for path signals is that track/reservation following is matching what a train can actually drive.

Let's say we have a vehicle of type A and A is only compatible with itself. In this case an A -> B transition should be a possible reservation end point. In case the reservation has to be removed, no reservation following from A to B should occur.

If we now also have a vehicle of type B that is compatible with A, then this vehicle can reserve over an B -> A boundary. If the A tile is already reserved by the A vehicle, the reservation should just fail. But it may not stop on the transition because then the reservation could clash with the A vehicle.

Whether `v->compatible_railtypes` or `GetRailTypeInfo(v->railtype)->compatible_railtypes` is the right thing to use is probably to be determined.

JGRennison:
Following on from the above, and the discord discussion, see attached for an example savegame illustrating the problem with asymmetric rail type compatibility.
[PR 9953 Example, 1990-01-03.sav.zip](https://github.com/user-attachments/files/20712483/PR.9953.Example.1990-01-03.sav.zip)

NAAN is defined as compatible with NABE, but NABE is not compatible with NAAN.

PR #9958, opened 2022-07-26
Add: [YAPF] Rail depot penalty based on the number of trains waiting in it

nchappe:
## Motivation / Problem

Non-stopped trains waiting in depot are not taken into account in YAPF cost calculations, because they may not have any track reserved. This can for instance be a problem for 'go to nearest depot' orders, as a train may choose to go to a depot full of waiting trains when it could go to a slightly more distant empty depot instead.

## Description

This PR adds a penalty for trains in depot to YAPF.
`rail_pbs_cross_penalty` seemed to be the most relevant penalty as it is used for reserved tracks.

## Checklist for review

### DIFF
--- a/src/depot.cpp
+++ b/src/depot.cpp
@@ -11,6 +11,7 @@
 #include "depot_base.h"
 #include "order_backup.h"
 #include "order_func.h"
+#include "train.h"
 #include "window_func.h"
 #include "core/pool_func.hpp"
 #include "vehicle_gui.h"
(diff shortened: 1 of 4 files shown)


2TallTyler:
I'm in favor of this, but I recognize that it could be confusing to players when a vehicle bypasses a nearby depot to go further away. I'd be curious to see what happens when a large number of vehicles is in a depot (say, after cloning 20+ trains and starting them all) — maybe a maximum penalty would be a good idea, if there's any precedent for that in other penalty calculations.

nchappe:
To answer your question, the default value of `rail_pbs_cross_penalty` is `3 * YAPF_TILE_LENGTH` so after cloning 20 trains, a train going to the nearest depot would prefer an empty depot 60 (free) tiles away.

I agree this can be confusing, so I added a cap of 8 trains. Beyond 8 trains stuck in the same depot, I guess the player should probably rethink their network anyways.
The max "trains in depot" penalty is now equivalent to a distance of 24 free tiles, or (I think) 6 occupied tiles.

PeterN:
Building lists of vehicles in a depot within the pathfinder seems like it might be bad idea for pathfinder performance.

2TallTyler:
Maybe having depots cache the current number of non-stopped trains inside them would be a better solution.

2TallTyler:
@nchappe Any plans to resume work on this based on the review comments?

nchappe:
I agree with the performance concerns! I reworked the PR so that the number of trains waiting in each depot is cached.

src/depot.cpp
```diff
@@ -47,3 +48,39 @@ Depot::~Depot()
...
+				if (!v->IsPrimaryVehicle()) continue;
+				if (!v->IsInDepot()) continue;
+				if (v->vehstatus.Test(VehState::Stopped)) continue;
+				break;
+		}
+
+		Depot *d = Depot::GetByTile(v->tile);
+		if (!d) continue;
+		d->running_vehicles++;
+	}
```
PeterN, line 85: This entire function should just use the Vehicle Tile Hash, so iterating the vehicle pool is not necessary.

Doing so might even be fast enough to not require any caching.

src/openttd.cpp
```diff
@@ -1250,6 +1251,7 @@ void StateGameLoop()
 			SaveOrLoad(name, SLO_SAVE, DFT_GAME_FILE, AUTOSAVE_DIR, false);
 		}
 
+		RebuildDepotOccupancyCache(VEH_TRAIN);
```
PeterN, line 1254: Updating this information every game tick seems excessive. There are some places in the code which are called when vehicles enter and exit depots, which could simply increment/decrement as needed.

src/pathfinder/yapf/yapf_costrail.hpp
```diff
@@ -400,6 +401,13 @@ class CYapfCostRailT : public CYapfCostBase {
 			} else if (IsRailDepotTile(cur.tile)) {
 				/* We will end in this pass (depot is possible target) */
 				end_segment_reason.Set(EndSegmentReason::Depot);
+				/* Add a penalty for each non-stopped train in the depot.
+				 * The penalty is capped at 8 trains. */
+				const Depot *d = Depot::GetByTile(cur.tile);
+				if (d) {
```
PeterN, line 407: `if (d != nullptr) {`

src/openttd.cpp
2TallTyler: For easy reference, [this commit](https://github.com/OpenTTD/OpenTTD/pull/11945/commits/c09f8963a832a78d1774fcc25311821e0b88d1cf) in my depot unbunching commit shows everywhere a vehicle leaves or enters a depot. 🙂

PR #9959, opened 2022-07-27
Feature: Constantly update destination of 'any depot' orders

nchappe:
## Motivation / Problem

Consider this line with a depot on each side, and a train T with a 'go to nearest depot' order (or an auto servicing order).

```
     D1
T====+===>
     D2
```

Right now the pathfinder will choose once and for all to go to D1 or to D2 when it first reaches the nearest depot order (see YapfTrainFindNearestDepot).
Say both depots are free at this time and it chooses D1. While on its way, another train could block the way to D1. In this case, T will wait for the path to D1 to be free, it won't be re-routed to D2.

This PR allows the destinations of 'any depot' orders to be updated if YAPF finds a less costly path.

## Description

When a train reaches the end of a segment, YAPF is called to determine which way the train should go next. It looks for the best path from the train to any valid destination tile, and reserves the relevant part of this path.
This PR marks *all* train depots as valid destination tiles for 'nearest depot' destinations (commit 2), instead of just the one from the call to YapfTrainFindNearestDepot.
So if for some reason the cost to reach the initial "nearest" depot increases YAPF may find a less costly depot. In the example above, the train would be re-routed to D2 when it reaches the PBS just before the D1/D2 intersection if D1 is blocked at this point.
This comes at no performance cost (YAPF may even return faster if it finds a better path).

The destination depot from YapfTrainFindNearestDepot is still useful as it serves as an initial hint to speed up subsequent calls to YAPF.

The destination shown in the vehicle window is updated in real-time thanks to commits 3 to 5.

## Limitations

- NPF is not supported.
- This is only implemented for trains, this feature would not benefit other vehicles as much.

## Checklist for review

### DIFF
--- a/src/depot_type.h
+++ b/src/depot_type.h
@@ -10,8 +10,12 @@
 #ifndef DEPOT_TYPE_H
 #define DEPOT_TYPE_H
 
-typedef uint16 DepotID; ///< Type for the unique identifier of depots.
+#include "station_type.h"
+
+typedef StationID DepotID; ///< Type for the unique identifier of depots.
 struct Depot;
+
+static const DepotID INVALID_DEPOT = INVALID_STATION;
 
 static const uint MAX_LENGTH_DEPOT_NAME_CHARS = 32; ///< The maximum length of a depot name in characters including '\0'
 
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -2318,13 +2318,16 @@ static const byte _initial_tile_subcoord[6][4][3] = {
  * @param[out] path_found Whether a path has been found or not.
  * @param do_track_reservation Path reservation is requested
  * @param[out] dest State and destination of the requested path
+ * @param[out] final_dest Final tile of the best path found
  * @return The best track the train should follow
  */
-static Track DoTrainPathfind(const Train *v, TileIndex tile, DiagDirection enterdir, TrackBits tracks, bool &path_found, bool do_track_reservation, PBSTileInfo *dest)
+static Track DoTrainPathfind(const Train *v, TileIndex tile, DiagDirection enterdir, TrackBits tracks, bool &path_found, bool do_track_reservation, PBSTileInfo *dest, TileIndex *final_dest)
 {
+	if (final_dest != nullptr) *final_dest = INVALID_TILE;
+
 	switch (_settings_game.pf.pathfinder_for_trains) {
 		case VPF_NPF: return NPFTrainChooseTrack(v, path_found, do_track_reservation, dest);
-		case VPF_YAPF: return YapfTrainChooseTrack(v, tile, enterdir, tracks, path_found, do_track_reservation, dest);
+		case VPF_YAPF: return YapfTrainChooseTrack(v, tile, enterdir, tracks, path_found, do_track_reservation, dest, final_dest);
 
 		default: NOT_REACHED();
 	}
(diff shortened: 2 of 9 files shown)


src/order_base.h
```diff
@@ -20,6 +20,8 @@
 #include "date_type.h"
 #include "saveload/saveload.h"
 
+static_assert (INVALID_STATION == INVALID_DEPOT);
```
LordAro: Why is this here?

src/depot_type.h
```diff
@@ -13,6 +13,8 @@
 typedef uint16 DepotID; ///< Type for the unique identifier of depots.
 struct Depot;
 
+static const DepotID INVALID_DEPOT = 0xFFFF;
```
LordAro: Thinking about the static_assert, why is this not just defined as `= INVALID_STATION` ?

src/train_cmd.cpp
```diff
@@ -2454,16 +2458,23 @@ class VehicleOrderSaver {
...
 		old_last_station_visited(_v->last_station_visited),
 		index(_v->cur_real_order_index),
-		suppress_implicit_orders(HasBit(_v->gv_flags, GVF_SUPPRESS_IMPLICIT_ORDERS))
+		suppress_implicit_orders(HasBit(_v->gv_flags, GVF_SUPPRESS_IMPLICIT_ORDERS)),
+		restored(false)
 	{
 	}
 
-	~VehicleOrderSaver()
+	void restore()
```
LordAro: function should start with a capital letter (we're not the STL!), and missing documentation.

Though I'm not quite clear why it can't just be part of the destructor (documentation!)

src/train_cmd.cpp
```diff
@@ -2454,16 +2458,23 @@ class VehicleOrderSaver {
...
 		this->v->current_order = this->old_order;
 		this->v->dest_tile = this->old_dest_tile;
 		this->v->last_station_visited = this->old_last_station_visited;
 		SB(this->v->gv_flags, GVF_SUPPRESS_IMPLICIT_ORDERS, 1, suppress_implicit_orders ? 1: 0);
+		this->restored = true;
+	}
+
+	~VehicleOrderSaver()
+	{
+		if (!this->restored) restore();
```
LordAro: missing `this->`

src/train_cmd.cpp
```diff
@@ -2318,13 +2318,16 @@ static const byte _initial_tile_subcoord[6][4][3] = {
  * @param[out] path_found Whether a path has been found or not.
  * @param do_track_reservation Path reservation is requested
  * @param[out] dest State and destination of the requested path
+ * @param[out] final_dest Final tile of the best path found
```
LordAro, line 2321: maybe a note to say this is only used for any depot orders?

LordAro (commented):
Some minor quibbles

src/train_cmd.cpp
nchappe: I fixed the case and documented the function.

I restore the order a bit before returning from ChooseTrainTrack. Doing it RAII-style with the destructor of VehicleOrderSaver would introduce a new 100-line scope in an already complex function.
I added Restore() to make this more readable.

src/train_cmd.cpp
nchappe: I would prefer to keep it as is because I can see other uses for it, for instance to show which platform of a station or which tile of a waypoint a train is heading to.

src/depot_type.h
nchappe: Thanks for the feedback. DepotID and INVALID_DEPOT are now defined in terms of their station counterparts.

michicc (approved):

src/depot_type.h
```diff
@@ -10,9 +10,13 @@
 #ifndef DEPOT_TYPE_H
 #define DEPOT_TYPE_H
 
-typedef uint16 DepotID; ///< Type for the unique identifier of depots.
+#include "station_type.h"
+
+typedef StationID DepotID; ///< Type for the unique identifier of depots.
```
JGRennison, line 15: This is a bit late I know, but this seems misleading from a type point of view.

A DepotID is not actually a StationID even if they happen to have the same field width.
i.e. values of type DepotID and StationID are not interchangeable, and they do not point into the same pools.

Issue #7889, opened 2020-01-01
[Request] Company rating via Server Admin Port


Hezkore:
I think it would be really useful if the server admin port was able to tell what company is leading according to the Company League Table found in-game.

![openttd_2020-01-01_04-44-13](https://user-images.githubusercontent.com/1727541/71637923-5f3c9380-2c51-11ea-8581-31fe7a9065fb.png)

The information could be stored inside the company class itself as the value `rating` or similar, along with the already existing values like `inaugurated` and `president`.

nielsmh:
While theoretically any enhancements to the admin port ought to be made through Game Script custom messages, I agree that having the admin port more useful by default is not bad.

Anyone wanting to look at this can start here:
https://github.com/OpenTTD/OpenTTD/blob/90f4abe03f7c1e1abeae3b2a2b4a37dd6086ded6/src/network/network_admin.cpp#L312-L345

marimeireles:
hey! I wanna work on this one :)

nielsmh:
@marimeireles Nice, welcome :)

One thing to keep in mind if you extend the network protocol, there are existing software that works with the existing protocol, so any changes should aim to stay compatible. I think just adding new fields to the end of the existing message should be safe, but I'm not entirely sure. The alternative would be to define a new type of packet with the additional data.

Hezkore:
> 
> 
> hey! I wanna work on this one :)

Any news?

pon-droid:
If @marimeireles isn't working on it already, would it be alright if I could have a try?

Hezkore:
> If @marimeireles isn't working on it already, would it be alright if I could have a try?

I think it's fair to say that @marimeireles isn't working on it, as is been two years heh.

marimeireles:
Oops, sorry folks. Go for it!

TrueBrain:
This information is already available in `COMPANY_ECONOMY`, as performance rating. You can just sort this list, and the one with the highest score is #1 in rating!

If you have any questions about this, do let us know!

PR #9973, opened 2022-08-21
Add #7889: Company rating via Server Admin Port

pon-droid:
## Motivation / Problem
Enhancing the admin port by sending company 'ranking' information as well. 

## Description

The ranking is according to the performance history of the company, the same as the company league table. The company rating is modified by calling _SendCompanyInfo_.

## Limitations
- Modifies company objects
- Duplicates sorting functionality with the company league table

I originally attempted to reuse the sorting done from the company league table GUI class, but then the admin port would have to wait for the player to open it up before getting rating information.

This is also my first PR for any project so there are definitely more things to be called out here, please do that. 

## Checklist for review

### DIFF
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -79,6 +79,8 @@ struct CompanyProperties {
 
 	Year inaugurated_year;           ///< Year of starting the company.
 
+	byte rating;                     ///< Ranking nth of the company with 1st being best performing according to performance history.
+
 	byte months_of_bankruptcy;       ///< Number of months that the company is unable to pay its debts
 	CompanyMask bankrupt_asked;      ///< which companies were asked about buying it?
 	int16 bankrupt_timeout;          ///< If bigger than \c 0, amount of time to wait for an answer on an offer to buy this company.
@@ -108,7 +110,7 @@ struct CompanyProperties {
 	CompanyProperties()
 		: name_2(0), name_1(0), president_name_1(0), president_name_2(0),
 		  face(0), money(0), money_fraction(0), current_loan(0), colour(0), block_preview(0),
-		  location_of_HQ(0), last_build_coordinate(0), share_owners(), inaugurated_year(0),
+		  location_of_HQ(0), last_build_coordinate(0), share_owners(), inaugurated_year(0), rating(0),
 		  months_of_bankruptcy(0), bankrupt_asked(0), bankrupt_timeout(0), bankrupt_value(0),
 		  terraform_limit(0), clear_limit(0), tree_limit(0), is_ai(false), engine_renew_list(nullptr) {}
 };
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -79,8 +79,6 @@ struct CompanyProperties {
 
 	Year inaugurated_year;           ///< Year of starting the company.
 
-	byte rating;                     ///< Ranking nth of the company with 1st being best performing according to performance history.
-
 	byte months_of_bankruptcy;       ///< Number of months that the company is unable to pay its debts
 	CompanyMask bankrupt_asked;      ///< which companies were asked about buying it?
 	int16 bankrupt_timeout;          ///< If bigger than \c 0, amount of time to wait for an answer on an offer to buy this company.
@@ -110,7 +108,7 @@ struct CompanyProperties {
 	CompanyProperties()
 		: name_2(0), name_1(0), president_name_1(0), president_name_2(0),
 		  face(0), money(0), money_fraction(0), current_loan(0), colour(0), block_preview(0),
-		  location_of_HQ(0), last_build_coordinate(0), share_owners(), inaugurated_year(0), rating(0),
+		  location_of_HQ(0), last_build_coordinate(0), share_owners(), inaugurated_year(0),
 		  months_of_bankruptcy(0), bankrupt_asked(0), bankrupt_timeout(0), bankrupt_value(0),
 		  terraform_limit(0), clear_limit(0), tree_limit(0), is_ai(false), engine_renew_list(nullptr) {}
 };
--- a/src/network/network_admin.cpp
+++ b/src/network/network_admin.cpp
@@ -332,7 +332,7 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
 	p->Send_uint8 (CeilDiv(c->months_of_bankruptcy, 3)); // send as quarters_of_bankruptcy
 
 	for (auto owner : c->share_owners) {
-	  p->Send_uint8(owner);
+		p->Send_uint8(owner);
 	}
 
 
--- a/src/network/network_admin.cpp
+++ b/src/network/network_admin.cpp
@@ -137,7 +137,6 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendError(NetworkErrorCode er
 
 	p->Send_uint8(error);
 	this->SendPacket(p);
-
 	std::string error_message = GetString(GetNetworkErrorMsg(error));
 
 	Debug(net, 1, "[admin] The admin '{}' ({}) made an error and has been disconnected: '{}'", this->admin_name, this->admin_version, error_message);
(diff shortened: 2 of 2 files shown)


src/company_base.h
```diff
@@ -79,6 +79,8 @@ struct CompanyProperties {
 
 	Year inaugurated_year;           ///< Year of starting the company.
 
+	byte rating;                     ///< Ranking nth of the company with 1st being best performing according to performance history.
```
TrueBrain: You create this variable, but only write and read to it in a single function. That means you don't actually need to allocate the variable here. Saves a bit of memory that is always allocated while it is only used in a very specific case :) But we will see in a bit that I think we don't even need to do this :)

src/network/network_admin.cpp
```diff
@@ -333,11 +332,27 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
 	p->Send_uint8 (CeilDiv(c->months_of_bankruptcy, 3)); // send as quarters_of_bankruptcy
 
 	for (auto owner : c->share_owners) {
-		p->Send_uint8(owner);
+	  p->Send_uint8(owner);
```
TrueBrain: OpenTTD uses tabs to to align statement. I think you replace them with spaces here.

src/network/network_admin.cpp
```diff
@@ -333,11 +332,27 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
...
+	for(Company* company: Company::Iterate()){
+	  companies.emplace_back(company);
+	}
+
+	std::sort(companies.begin(), companies.end(), [](Company * const &c1, Company * const &c2){
+	  return c2->old_economy[0].performance_history < c1->old_economy[0].performance_history;
+	});
+
+	for(size_t t = 0; t < companies.size(); t++){
+	  companies[t]->rating = t+1;
```
TrueBrain: I am wondering about this block of logic, if that is the right thing to do. We manually sort the performance history here, and deduce a rating out of that. But why not let the admin-port clients do that? And just send over the `performance_history` to them? Saves us a lot of complex calculations, gives them more information than they receive now, and avoids nasty problems like: if you now request two companies with a slight delay, they might have the exact same rating as they so happened to have changed `performance_history`.

If we would go for this `rating` solution, the logic here could use some cleverness. You now calculate the rating for all companies, while you are only interested in a single one. So why not just calculate the single one? For example, it could be something like this (this replaces line 338 till 349 in your code):

```
int rating = 1;
for (auto company: Company::Iterate()) {
  if (c->old_economy[0].performance_history < company->old_economy[0].performance_history) rating++;
}
```

Basically, we find all companies that have a higher rating than us, meaning we drop one rating. The result is the exact same, but without the two loops and the sorting :) At least .. I think it is the same :P My simple testing shows it is :D But you do need to test it first :)

src/network/network_admin.cpp
```diff
@@ -333,11 +332,27 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
...
 	}
 
+	p->Send_uint8 (c->rating);
```
TrueBrain: I have no clue what the impact is of just adding data to an admin-port packet. Do clients break? I think we have a version-number somewhere; it might be needed to bump that number. Maybe someone like @nielsmh knows?

TrueBrain (commented):
Excellent start for a Pull Request to OpenTTD! Thank you for that!

src/network/network_admin.cpp
nielsmh: Yeah that algorithm is simpler, the current way of getting everything, sorting it, and then counting, is excessive.
It could also be written with a `std::count_if()` call.

src/network/network_admin.cpp
nielsmh: Each packet contains a 16 bit size field at the start, so clients that just ignore data at the end of the packet that they don't know about should keep working. I'm pretty sure clients are supposed to do that.
I think the protocol version or packet type id should only be changed if fields are removed/reordered/redefined.

ldpl:
I don't think this is a good idea. OpenTTD already sends performance ratings over the admin port so clients can figure out ranking themselves if they want. And adding it to the game itself does nothing but bloat the codebase needlessly IMO. Neither does rating really belong to company info logically as it's an attribute of a company list, not a single company on its own.

Also, the company league table is pointless and needs some way to customize it with proper ratings, not spreading its stupidity into more places.

src/network/network_admin.cpp
```diff
@@ -336,8 +335,15 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
 		p->Send_uint8(owner);
 	}
 
+
+	uint8 rating = std::count_if(Company::Iterate().begin(), Company::Iterate().end(), [&](auto company){
+	return c->old_economy[0].performance_history < company->old_economy[0].performance_history;});
+
+	p->Send_uint8 (rating);
```
TrueBrain, line 342: I posted this in an earlier comment, I am still not sure about the answer here. But do we actually want to send the rating itself to the client? As they can fetch the performance themselves, and deduce the rating based on it.

One use-case I can imagine is that you only want to fetch the company-info every N minutes and display that on some website. In that case I think it is a valid case, as you don't want to bother with fetching the economy and doing the calculation yourself.

But that said, the one thing that annoys me slightly about this, is that if you fetch multiple companies over time, you can have two companies with the same rating just because the game-state updated between your requests. The `COMPANY_ECONOMY` packet solves this by always sending all companies at once (and also includes the performance rating).

An alternative could be to add the rating in that packet explicit. Albeit a bit silly, it wouldn't hurt either. But changing that packet in a backwards compatible manner is nearly impossible.

Owh, the choices ... :D

Either way, curious what other developers think. @nielsmh , given you replied in the original issue too, how do you view this?

src/network/network_admin.cpp
```diff
@@ -137,7 +137,6 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendError(NetworkErrorCode er
 
 	p->Send_uint8(error);
 	this->SendPacket(p);
-
```
TrueBrain, line 140: Please in PRs always try to minimize the change. Removing these whiteline (as for the 2 others in this PR) carries no weight, and only distracts from your change. So it is better to remove them.

Otherwise we have constant changes in newlines in PRs, depending on the mood of the author ;) :D

src/network/network_admin.cpp
```diff
@@ -336,8 +335,15 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
 		p->Send_uint8(owner);
 	}
 
+
+	uint8 rating = std::count_if(Company::Iterate().begin(), Company::Iterate().end(), [&](auto company){
```
TrueBrain, line 339: a bit lower is a function `SendCompanyUpdate`, which sends the same info as this packet, minus some immutable fields (no clue how we ever were like: yeah, that is a great idea! Let's have two nearly identical packets, but .. here we are :D). As rating can change, and if we go for this change, I think this code should also be added there. Just to avoid confusion with any admin-port user :)

src/network/network_admin.cpp
```diff
@@ -336,8 +335,15 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCompanyInfo(const Company
...
+	return c->old_economy[0].performance_history < company->old_economy[0].performance_history;});
```
TrueBrain, line 340: Coding style is a bit off here. I suggest something like this:

```
uint8 rating = std::count_if(Company::Iterate().begin(), Company::Iterate().end(), [&](auto company) {
    return c->old_economy[0].performance_history < company->old_economy[0].performance_history;
});
```

That way it is a bit more clear where the body of the lambda starts and ends :)

src/network/network_admin.cpp
nielsmh: When I wrote the reply in the original issue, I didn't realize that it was about sending the ranking position, and that the information for a client to deduce that was already available to the client.

Seeing this now, I'm very inclined to say close this and the original issue.

TrueBrain:
I am really sorry @pon-droid , but it seems we let you astray. As per https://github.com/OpenTTD/OpenTTD/pull/9973#discussion_r957633496, the original issue should have been closed with: already implemented, and I will do as such in a moment.

But don't despair! There is nothing wrong with your PR itself. And tons of other small issues that require attention. So if you are up for it, maybe pick up one of the other small issues? :D

Tnx, and sorry :)

PR #9978, opened 2022-08-26
Fix #9977: Use a deque rather than linked list for console buffer

### LINKED ISSUE #9977: [Crash]: Console command `clear` can stack overflow and crash
### Version of OpenTTD

e9507e7ffa on Windows 10, 64-bit

### Steps to reproduce

1. `debuglevel 9`
2. Wait a few minutes or until a few ten-thousand lines are logged to the in-game console.
3. `clear`
4. Crash

Alternatively,
1. Create a file with the name `a.txt` and the contents as follows:
```
exec a.txt
exec a.txt
exec a.txt
```
2. Run `exec a.txt`
3. Run `clear` (or just wait)

Doing this in master bypasses the crash handler, while doing the same in JGRPP 0.48.1 crashes properly. The crash files from the latter are attached.

### Upload crash files

The following crash report was done on JGRPP 0.48.1: [crash-20220825T204753Z.zip](https://github.com/OpenTTD/OpenTTD/files/9428214/crash-20220825T204753Z.zip)

Doing the same in vanilla generates only `crash.dmp`, whose file size is 0 bytes.

nielsmh:
## Motivation / Problem

The internal console uses a linked list for its backlog buffer. Deleting items from the backlog is implemented via recursion. This means that deleting a too long backlog will overflow the stack and crash the game.

A linked list also makes scrolling back inefficient, as the only way to look up old lines in the history are by walking the entire backlog.

## Description

Replace the linked list with a `std::deque`. Keep inserting the newest line in front (position 0), and truncate by deleting the end. The deque data structure makes this reasonably efficient.

## Limitations

Mostly smoke-tested, but it does pass the second reproduction method given in #9977 (the recursive script execution filling the backlog).

## Checklist for review

### DIFF
--- a/src/console_gui.cpp
+++ b/src/console_gui.cpp
@@ -21,6 +21,8 @@
 #include "console_func.h"
 #include "rev.h"
 #include "video/video_driver.hpp"
+#include <deque>
+#include <string>
 
 #include "widgets/console_widget.h"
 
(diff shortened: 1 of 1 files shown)


src/console_gui.cpp
```diff
@@ -198,8 +142,13 @@ struct IConsoleWindow : Window
 	 */
 	void Scroll(int amount)
 	{
-		int max_scroll = std::max(0, IConsoleLine::size + 1 - this->height / this->line_height);
-		IConsoleWindow::scroll = Clamp<int>(IConsoleWindow::scroll + amount, 0, max_scroll);
+		size_t max_scroll = std::max<size_t>(0, _iconsole_buffer.size() + 1 - this->height / this->line_height);
```
frosch123: `size_t` is unsigned. This clamping to zero won't work, but create huge numbers.

src/console_gui.cpp
```diff
@@ -235,9 +187,12 @@ struct IConsoleWindow : Window
...
 
-		if (IConsoleLine::Truncate() &&
-				(IConsoleWindow::scroll > IConsoleLine::size)) {
-			IConsoleWindow::scroll = std::max(0, IConsoleLine::size - (this->height / this->line_height) + 1);
+		assert(this->height >= 0 && this->line_height > 0);
+		size_t visible_lines = (size_t)(this->height / this->line_height);
+
+		if (TruncateBuffer() && IConsoleWindow::scroll + visible_lines > _iconsole_buffer.size()) {
+			size_t max_scroll = (visible_lines > _iconsole_buffer.size()) ? 0 : _iconsole_buffer.size() + 1 - visible_lines;
+			IConsoleWindow::scroll = std::min<size_t>(IConsoleWindow::scroll, max_scroll);
```
nielsmh, line 195: I'm not entirely sure if this is correct, and I can't think of a good way to exercise the truncation code right now.

frosch123 (approved):

src/console_gui.cpp
frosch123: lgtm

PR #9979, opened 2022-08-26
Fix #9940: Print debuglevel parse errors to console when changed from console

### LINKED ISSUE #9940: [Bug]: `debuglevel` with invalid argument prints error to external dialog and may execute partially
### Version of OpenTTD

19af139 on Windows 10

### Expected result

Running `debuglevel` with invalid/unknown arguments, such as `debuglevel a` should print the error message to the in-game console instead of an external dialog box.

The command's argument should be fully parsed and checked for errors *before* the current debug levels are written to, with abort beforehand (leaving debug levels unchanged) if parsing fails.

### Actual result

Running `debuglevel` with invalid/unknown arguments, such as `debuglevel a` will print the error message to an external dialog box (treated as a separate window) and not the in-game console.

In addition, running `debuglevel s3npf9` will show "Unknown debug level 's3'" and leave debug levels unchanged, while `debuglevel npf9s3` will also show "Unknown debug level 's3'", but change the debug levels, setting `npf` to 9.

### Steps to reproduce

1. Run `debuglevel npf9s3yapf9` while in the main menu.
2. An external dialog box shows: "Unknown debug level 's3'"
3. Close the external dialog box.
4. NPF pathfinder debug info shows up in the console, as if `debuglevel npf9` was set.
5. No YAPF pathfinder debug info shows up in the console. If `debuglevel yapf9` was set, such info would have showed up.

nielsmh:
## Motivation / Problem

The `SetDebugString` function doesn't care if it gets called from a commandline parameter or from the internal console. It always uses error output as if it was called during startup where the GUI might not be available.

## Description

Make it take a callback for reporting parse errors.

## Limitations

Slightly more annoying to call, but it's only called from 3 places.

## Checklist for review

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -167,21 +167,23 @@ void SetDebugString(const char *s, void (*error_func)(const char *))
 	char *end;
 	const char *t;
 
-	/* global debugging level? */
+	/* Store planned changes into map during parse */
+	std::map<const char *, int> new_levels;
+
+	/* Global debugging level? */
 	if (*s >= '0' && *s <= '9') {
 		const DebugLevel *i;
 
 		v = strtoul(s, &end, 0);
 		s = end;
 
-		for (i = debug_level; i != endof(debug_level); ++i) *i->level = v;
+		for (i = debug_level; i != endof(debug_level); ++i) {
+			new_levels[i->name] = v;
+		}
 	}
 
-	/* individual levels */
+	/* Individual levels */
 	for (;;) {
-		const DebugLevel *i;
-		int *p;
-
 		/* skip delimiters */
 		while (*s == ' ' || *s == ',' || *s == '\t') s++;
 		if (*s == '\0') break;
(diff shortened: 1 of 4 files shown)


src/debug.cpp
```diff
@@ -203,7 +204,8 @@ void SetDebugString(const char *s)
 		if (p != nullptr) {
 			*p = v;
 		} else {
-			ShowInfoF("Unknown debug level '%.*s'", (int)(s - t), t);
+			std::string error_string = std::string("Unknown debug level '") + std::string(t, s - t) + std::string("'");
```
LordAro: Perhaps fmt?

LordAro:
Doesn't fix the other part of #9940 (partial parses still affect debuglevel), as far as I can tell?

nielsmh:
Oh, I totally missed that the bug report was supposed to be two bug reports.

LordAro (dismissed):

frosch123 (approved):

PR #9983, opened 2022-08-27
Fix #9971: Don't assert that time moves forward

### LINKED ISSUE #9971: [Crash]: Assertion failed at line 968 of framerate_gui.cpp when pausing from brief hyper-fast-forward (1M+ TPS)
### Version of OpenTTD

88a046a7 on Windows 10, 64-bit (via Steam)

### Steps to reproduce

1. Load the attached saved game.
2. Open the frame rate window.
3. Unpause the game.
4. Open the game loop window from the frame rate window.
5. Briefly press TAB to fast-forward, then hit pause.
6. The game asserts here:

https://github.com/OpenTTD/OpenTTD/blob/88a046a76e97ca15c6fc55e808d33141fc4f9bff/src/framerate_gui.cpp#L968

---

My setup for getting 1.25 million ticks per second is as follows:
1. Use a system with an Intel Core i5-9400F CPU (base clock: 2.90 GHz, boost: 3.90 GHz on a single core), or another CPU with better single-threaded performance.
2. Run OpenTTD on that system with stock configs.
3. Disable the fast-forward speed limit (set the value in Interface -> Fast forward speed limit to `0`, which means "no limit")
7. Load the attached savegame (`a.sav` in the crash.zip), which has been pre-prepared for maximum simulation speed.
8. Open the frame rate window.
9. Unpause, fast-forward, open the "Game loop" time graph, and watch.

### Upload crash files

[crash20220819225606-a.zip](https://github.com/OpenTTD/OpenTTD/files/9385797/crash20220819225606-a.zip)

nielsmh:
## Motivation / Problem

It seems that if simulation goes too fast, the tick time measurements can get inaccurate and produce measurements that appear to go backwards in time. Maybe it has to do with threads being migrated between CPU cores? No idea.

## Description

Just ignore measurements that appear to do time travel. Don't try to draw them.

## Limitations

I tried the reproduction given in #9971 and I can't make it crash with or without this patch. I did test with builds with asserts enabled, and reached about 1.3 million ticks/s.
It probably has to do with specific CPU models that have performance counters behaving in specific ways.

## Checklist for review

### DIFF
--- a/src/framerate_gui.cpp
+++ b/src/framerate_gui.cpp
@@ -965,7 +965,7 @@ struct FrametimeGraphWindow : Window {
 					(int)Scinterlate<int64>(x_zero, x_max, 0, (int64)draw_horz_scale, (int64)draw_horz_scale - (int64)time_sum),
 					(int)Scinterlate<int64>(y_zero, y_max, 0, (int64)draw_vert_scale, (int64)value)
 				};
-				assert(newpoint.x <= lastpoint.x);
+				if (newpoint.x > lastpoint.x) continue; // don't draw backwards
 				GfxDrawLine(lastpoint.x, lastpoint.y, newpoint.x, newpoint.y, c_lines);
 				lastpoint = newpoint;
 


glx22 (approved):
Seems safe to me.

ldpl:
May depend on the compiler as high_resolution_clock is not guaranteed to be steady and can alias to system_clock with all the ensuing adjustments.

PR #9994, opened 2022-09-04
Fix #9993: Handle DPI changes on macOS and Windows

### LINKED ISSUE #9993: [Bug]: Moving window between retina and non-retina causes incorrect scaling
### Version of OpenTTD

12.2, macOS Monterey 12.5

### Expected result

OpenTTD should render correctly regardless of what screen it launched and whether the window was moved between displays.

### Actual result

OpenTTD determines the GUI zoom based on the scaling factor of the display it launched on. After moving the window to another display with a different scaling factor, the GUI zoom is not updated. As a result the GUI is rendered at a with an incorrect size (at 200% or 50%). In order to actually play on a different display, OpenTTD needs to be (re)started while having the pointer on the desired display.

bug0

### Steps to reproduce

Either:
1. Launch OpenTTD on a retina display
2. Move the window to a non-retina display
3. The GUI is rendered at 200%.

Or:
1. Launch OpenTTD on a non-retina display
2. Move the window to a retina display
3. The GUI is rendered at 50%.

Bouke:
## Motivation / Problem

Fixes #9993. When the application window is moved between retina and non-retina displays (macOS) or different DPI settings (Windows), the application should update the GUI zoom. Any opened windows should remain at their original position, but corrected for the new GUI zoom level.

## Description

When we determine that the window has moved to another display with a different scaling factor, we update `_gui_zoom` and apply the zoom scale factor to all internal `Window`s and their `viewport`. This is done by calling `UpdateGUIZoom()` and comparing the before and after value of `_gui_zoom`. A new zoom helper method `AdjustByZoom` scales any current value according to the difference in zoom (up/down). Then we allow the video driver to update the window/buffer and when that is done, we ReInit all windows to rerender them at the new zoom. This requires some specific ordering as the game tries to keep `Window`s visible: within the main window, below the toolbar and above the statusbar. My initial attempt didn't account for this and caused `Window`s to move around as a result.

On Windows we need to resize the window to match the new DPI configuration to keep the same size on-screen. Furthermore we opt-in to the [V2 DPI Awareness so that the _non-client area_ (title bar) is resized by the OS](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enablenonclientdpiscaling), on supported versions ([Windows 10, version 1607 and later](https://docs.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process#setting-default-awareness-with-the-application-manifest)).

Demo on macOS:
fix0

## Limitations

<s>- The fix is only implemented for the macOS video backend. The bug could be present on other systems, however I have no experience nor any setup to verify this. When this is the case, we might want to introduce a more generic hook to change `_gui_zoom`.

- When switching displays while on the main screen, the title "OpenTTD" is repositioned. This is not a new bug. Fixed in #9997.

- Sometimes the new scaling is not taking effect. I have not discovered how to reproduce.

- Station names get cut-off when the original display is non-retina. I _think_ this is caused by the max length calculation only being performed once.

- I need to make myself familiar with `_use_hidpi` / `_allow_hidpi_window` and see how that affects this PR.

- The initial GUI rescale doesn't apply when going from Retina -> non-Retina. I need to look into this. </s>

- The changes for Windows could use a critical pair of eyes. It appears to work fine in my VM when moving between displays with different DPI settings though.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.<s>

### DIFF
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -17,10 +17,12 @@
 #include "settings_type.h"
 #include "network/network.h"
 #include "network/network_func.h"
+#include "window_gui.h"
 #include "window_func.h"
 #include "newgrf_debug.h"
 #include "thread.h"
 #include "core/backup_type.hpp"
+#include "viewport_func.h"
 
 #include "table/palettes.h"
 #include "table/string_colours.h"
@@ -2040,6 +2042,38 @@ void UpdateGUIZoom()
 	} else {
 		_font_zoom = static_cast<ZoomLevel>(_font_zoom_cfg);
 	}
+}
+
+/**
+ * Resolve GUI zoom level and adjust GUI to new zoom, if auto-suggestion is requested.
+ * @returns true when the zoom level has changed, caller must call ReInitAllWindows(true)
+ * after resizing the application's window/buffer.
+ */
+bool AdjustGUIZoom()
+{
+	auto old_zoom = _gui_zoom;
+	UpdateGUIZoom();
+	if (old_zoom == _gui_zoom) return false;
+	GfxClearSpriteCache();
+	VideoDriver::GetInstance()->ClearSystemSprites();
+	ClearFontCache();
+	GfxClearSpriteCache();
+	UpdateAllVirtCoords();
+
+	/* Adjust all window sizes to match the new zoom level, so that they don't appear
+	   to move around when the application is moved to a screen with different DPI. */
+	auto zoom_shift = old_zoom - _gui_zoom;
+	for (Window *w : Window::Iterate()) {
+		w->left = AdjustByZoom(w->left, zoom_shift);
+		w->top = AdjustByZoom(w->top, zoom_shift);
+		w->width = AdjustByZoom(w->width, zoom_shift);
+		w->height = AdjustByZoom(w->height, zoom_shift);
+		if (w->viewport != nullptr) {
+			w->viewport->zoom = Clamp(ZoomLevel(w->viewport->zoom - zoom_shift), _settings_client.gui.zoom_min, _settings_client.gui.zoom_max);
+		}
+	}
+
+	return true;
 }
 
 void ChangeGameSpeed(bool enable_fast_forward)
--- a/src/video/cocoa/cocoa_wnd.mm
+++ b/src/video/cocoa/cocoa_wnd.mm
@@ -1269,8 +1269,12 @@ void CocoaDialog(const char *title, const char *message, const char *buttonLabel
 /** Screen the window is on changed. */
 - (void)windowDidChangeBackingProperties:(NSNotification *)notification
 {
+	bool did_adjust = AdjustGUIZoom();
+
 	/* Reallocate screen buffer if necessary. */
 	driver->AllocateBackingStore();
+
+	if (did_adjust) ReInitAllWindows(true);
 }
 
 /** Presentation options to use for full screen mode. */
--- a/src/video/win32_v.cpp
+++ b/src/video/win32_v.cpp
@@ -38,6 +38,10 @@
 #define PM_QS_INPUT 0x20000
 #endif
 
+#ifndef WM_DPICHANGED
+#define WM_DPICHANGED 0x02E0
+#endif
+
 bool _window_maximize;
 static Dimension _bck_resolution;
 DWORD _imm_props;
@@ -668,6 +672,24 @@ LRESULT CALLBACK WndProcGdi(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
 					break;
 			}
 			return TRUE;
+		}
+
+		case WM_DPICHANGED: {
+			auto did_adjust = AdjustGUIZoom();
+
+			/* Resize the window to match the new DPI setting. */
+			RECT *prcNewWindow = (RECT *)lParam;
+			SetWindowPos(hwnd,
+				NULL,
+				prcNewWindow->left,
+				prcNewWindow->top,
+				prcNewWindow->right - prcNewWindow->left,
+				prcNewWindow->bottom - prcNewWindow->top,
+				SWP_NOZORDER | SWP_NOACTIVATE);
+
+			if (did_adjust) ReInitAllWindows(true);
+
+			return 0;
 		}
 
 /* needed for wheel */
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -3343,6 +3343,13 @@ void HideVitalWindows()
 	CloseWindowById(WC_STATUS_BAR, 0);
 }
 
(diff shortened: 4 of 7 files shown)


James103:
Commit checker is failing for 2 reasons:
1. All instances of four consecutive space characters used for indentation must be replaced with a tab character.
2. The commit message must be amended. Put `Fix #9993: ` before the existing title.
Full title: `Fix #9993: Support switching (non-)retina displays on macOS`

src/video/cocoa/cocoa_wnd.mm
```diff
@@ -1269,8 +1272,36 @@ - (void)windowDidEnterFullScreen:(NSNotification *)aNotification
...
+	if (suggestedZoomLevel != newSuggestedZoomLevel) {
+		GfxClearSpriteCache();
+		UpdateGUIZoom();
+		driver->ClearSystemSprites();
+		ClearFontCache();
+		GfxClearSpriteCache();
+
+		// Apply scaling factor to all windows so when GameSizeChanged is called
+		// from AllocateBackingStore the windows still fit within the main window.
+		float scaling = float(suggestedZoomLevel) / float(newSuggestedZoomLevel);
```
Bouke: This is incorrect as the zoom levels have log scale. I could use `UnScaleByZoom` here to get the scaling factor.

.editorconfig
```diff
@@ -4,6 +4,6 @@ root = true
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{c,cpp,h,hpp}]
+[*.{c,cpp,h,hpp,mm}]
```
nielsmh: I'd really like this to be a separate commit, tbh. Fixing the editor config is logically separate from the UI logic change.

src/video/cocoa/cocoa_wnd.mm
nielsmh: Use `-` to find the difference between scaling levels. You probably don't need to use floats for this at all, nothing else does.

Bouke: Indeed, I didn't factor in the possibility of a bit shift. That could work: shifting by the difference.

nielsmh: That's exactly what the Zoom and UnZoom functions (`zoom_func.h`) already do.

Bouke: @nielsmh Using `-` would produce [-2, 2] for the current implementation. Is there a method available to apply this zoom difference, or do I need a combination of `ScaleByZoom` and `UnScaleByZoom` to achieve what I want?

```
if (diff < 0) { _value_ = ScaleByZoom(_value_, -diff); }
else { _value_ = ScaleByZoom(_value_, diff); }
```

.editorconfig
Bouke: Separate commit in this PR, or also separate PR?

src/video/cocoa/cocoa_wnd.mm
nielsmh: Yeah you should take the difference between old and new level, if it's negative then use UnScaleByZoom by the absolute value, if it's positive then use ScaleByZoom. You should really read `zoom_func.h`, it's very short and all the functions are documented.

.editorconfig
michicc: Same PR is fine, unless you want to aim for claiming #10000 😄

Bouke: I’d need to sit out two more issues/PRs 😇

src/video/cocoa/cocoa_wnd.mm
Bouke: Thanks! I’ve updated the code. To reduce duplication I’ve introduced a helper with the condition shown above.

src/gfx.cpp
```diff
@@ -2042,6 +2044,37 @@ void UpdateGUIZoom()
 	}
 }
 
+/**
+ * Resolve GUI zoom level and adjust GUI to new zoom, if auto-suggestion is requested.
+ * @returns true when the zoom level has changed, caller must call ReInitAllWindows(true)
+ * after resizing the application's window/buffer.
+ */
+bool AdjustGUIZoom() {
```
LordAro: codestyle: opening brace of functions should be on their own line

src/window.cpp
```diff
@@ -3352,9 +3358,13 @@ void ReInitAllWindows(bool zoom_changed)
 	extern void InitDepotWindowBlockSizes();
 	InitDepotWindowBlockSizes();
 
+	// When _gui_zoom has changed, we need to resize toolbar and statusbar first,
+	// so EnsureVisibleCaption uses the updated size information.
```
LordAro: codestyle: multiline comments should use `/* */`

LordAro (dismissed):
LGTM, but I'd quite like a review from someone who actually understands this stuff

@michicc ?

src/gfx.cpp
```diff
@@ -2042,6 +2044,38 @@ void UpdateGUIZoom()
...
+	GfxClearSpriteCache();
+	VideoDriver::GetInstance()->ClearSystemSprites();
+	ClearFontCache();
+	GfxClearSpriteCache();
+	UpdateAllVirtCoords();
+
+	/* Adjust all window sizes to match the new zoom level, so that they don't appear
+	   to move around when the application is moved to a screen with different DPI. */
+	auto zoom_shift = old_zoom - _gui_zoom;
+	for (Window* w : Window::Iterate()) {
```
michicc: Formatting:
```suggestion
	for (Window *w : Window::Iterate()) {
```

src/video/win32_v.cpp
```diff
@@ -670,6 +674,24 @@ LRESULT CALLBACK WndProcGdi(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
 			return TRUE;
 		}
 
+		case WM_DPICHANGED: {
+			auto did_adjust = AdjustGUIZoom();
+
+			/* Resize the window to match the new DPI setting. */
+			RECT* const prcNewWindow = (RECT*)lParam;
```
michicc: Not sure actually want a const pointer to a modifiable RECT, and it is not reused anyway, so just (and also formatting):
```suggestion
			RECT *prcNewWindow = (RECT *)lParam;
```

michicc (commented):
Can't actually properly test this (especially of OSX), but the code looks sound to me.

src/video/win32_v.cpp
Bouke: I took the sample code from here: https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged, but I'll change it regardless.

michicc (approved):

PR #9996, opened 2022-09-04
Fix #9995: Adjust viewport zoom level for HiDPI displays

### LINKED ISSUE #9995: [Bug]: Vehicle viewer uses unexpected zoom level on retina display
### Version of OpenTTD

12.2, macOS Monterey 12.5

### Expected result

Clicking a train (or other vehicle) should render their window with the "normal" zoom level, regardless of OpenTTD being run on a retina or non-retina display.

### Actual result

The vehicle viewport renders at 50% size when run on a retina display.

bug1

### Steps to reproduce

1. Run game on a retina display
2. Open vehicle window
3. Notice the zoom level is zoomed out one step too many

Bouke:
## Motivation / Problem

Fixes #9995.

## Description

On HiDPI screens the zoom level is increased for detailed rendering. This causes hard-coded zoom levels to be off by this adjustment. To fix these default zoom levels, we adjust based on `_gui_zoom`.

Expected viewport zoom:
OpenTTD 9996 002

Actual viewport zoom:
OpenTTD 9996 003

Fixed viewport zoom:
OpenTTD 9996 fix 003

I've tested with all viewports: trains, buses, airplanes, ships, towns, news, industries, waypoint, extra viewport.

The change is also applied to savegames, when saving them we save current zoom, but un-adjust it for the current `_gui_zoom`. This means that regardless of the user's display scaling, loading the game will restore the viewport at the same experienced viewport.

## Limitations

- Savegames saved from by a HiDPI prior to this PR will restore with a zoomed-out viewport, as we now regard the saved zoom level as display-independent. This allows for a consistent viewport restoration regardless of display dpi settings.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.<s>

### DIFF
--- a/src/news_gui.cpp
+++ b/src/news_gui.cpp
@@ -310,7 +310,7 @@ struct NewsWindow : Window {
 		/* Initialize viewport if it exists. */
 		NWidgetViewport *nvp = this->GetWidget<NWidgetViewport>(WID_N_VIEWPORT);
 		if (nvp != nullptr) {
-			nvp->InitializeViewport(this, ni->reftype1 == NR_VEHICLE ? 0x80000000 | ni->ref1 : (uint32)GetReferenceTile(ni->reftype1, ni->ref1), ZOOM_LVL_NEWS);
+			nvp->InitializeViewport(this, ni->reftype1 == NR_VEHICLE ? 0x80000000 | ni->ref1 : (uint32)GetReferenceTile(ni->reftype1, ni->ref1),ScaleZoomGUI(ZOOM_LVL_NEWS));
 			if (this->ni->flags & NF_NO_TRANSPARENT) nvp->disp_flags |= ND_NO_TRANSPARENCY;
 			if ((this->ni->flags & NF_INCOLOUR) == 0) {
 				nvp->disp_flags |= ND_SHADE_GREY;
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -2736,7 +2736,7 @@ public:
 		}
 		this->FinishInitNested(window_number);
 		this->owner = v->owner;
-		this->GetWidget<NWidgetViewport>(WID_VV_VIEWPORT)->InitializeViewport(this, this->window_number | (1 << 31), _vehicle_view_zoom_levels[v->type]);
+		this->GetWidget<NWidgetViewport>(WID_VV_VIEWPORT)->InitializeViewport(this, this->window_number | (1 << 31), ScaleZoomGUI(_vehicle_view_zoom_levels[v->type]));
 
 		this->GetWidget<NWidgetCore>(WID_VV_START_STOP)->tool_tip       = STR_VEHICLE_VIEW_TRAIN_STATUS_START_STOP_TOOLTIP + v->type;
 		this->GetWidget<NWidgetCore>(WID_VV_RENAME)->tool_tip           = STR_VEHICLE_DETAILS_TRAIN_RENAME + v->type;
--- a/src/zoom_func.h
+++ b/src/zoom_func.h
@@ -80,6 +80,26 @@ static inline int UnScaleGUI(int value)
 }
 
 /**
+ * Scale zoom level relative to GUI zoom.
+ * @param value zoom level to scale
+ * @return scaled zoom level
+ */
+static inline ZoomLevel ScaleZoomGUI(ZoomLevel value)
+{
+	return std::clamp(ZoomLevel(value + (ZOOM_LVL_GUI - ZOOM_LVL_OUT_4X)), ZOOM_LVL_MIN, ZOOM_LVL_MAX);
+}
+
+/**
+ * UnScale zoom level relative to GUI zoom.
+ * @param value zoom level to scale
+ * @return un-scaled zoom level
+ */
+static inline ZoomLevel UnScaleZoomGUI(ZoomLevel value)
+{
+	return std::clamp(ZoomLevel(value - (ZOOM_LVL_GUI - ZOOM_LVL_OUT_4X)), ZOOM_LVL_MIN, ZOOM_LVL_MAX);
+}
+
+/**
  * Scale traditional pixel dimensions to GUI zoom level.
  * @param value Pixel amount at #ZOOM_LVL_BASE (traditional "normal" interface size).
  * @return Pixel amount at #ZOOM_LVL_GUI (current interface size).
--- a/src/zoom_type.h
+++ b/src/zoom_type.h
@@ -35,7 +35,7 @@ enum ZoomLevel : byte {
 	ZOOM_LVL_VIEWPORT = ZOOM_LVL_OUT_4X, ///< Default zoom level for viewports.
 	ZOOM_LVL_NEWS     = ZOOM_LVL_OUT_4X, ///< Default zoom level for the news messages.
 	ZOOM_LVL_INDUSTRY = ZOOM_LVL_OUT_8X, ///< Default zoom level for the industry view.
-	ZOOM_LVL_TOWN     = ZOOM_LVL_OUT_8X, ///< Default zoom level for the town view.
+	ZOOM_LVL_TOWN     = ZOOM_LVL_OUT_4X, ///< Default zoom level for the town view.
 	ZOOM_LVL_AIRCRAFT = ZOOM_LVL_OUT_4X, ///< Default zoom level for the aircraft view.
 	ZOOM_LVL_SHIP     = ZOOM_LVL_OUT_4X, ///< Default zoom level for the ship view.
 	ZOOM_LVL_TRAIN    = ZOOM_LVL_OUT_4X, ///< Default zoom level for the train view.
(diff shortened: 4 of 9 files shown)


src/vehicle_gui.cpp
```diff
@@ -2736,7 +2736,9 @@ struct VehicleViewWindow : Window {
 		}
 		this->FinishInitNested(window_number);
 		this->owner = v->owner;
-		this->GetWidget<NWidgetViewport>(WID_VV_VIEWPORT)->InitializeViewport(this, this->window_number | (1 << 31), _vehicle_view_zoom_levels[v->type]);
+
+        auto zoom_level = ZoomLevel(_vehicle_view_zoom_levels[v->type] + _gui_zoom - ZOOM_LVL_OUT_4X); // adjust zoom level for GUI scaling (HiDPI)
+        this->GetWidget<NWidgetViewport>(WID_VV_VIEWPORT)->InitializeViewport(this, this->window_number | (1 << 31), zoom_level);
```
James103: On each of the two lines you changed, you need to indent with 2 tab characters instead of 8 spaces.

Bouke: Thanks! I did see your comment on the other PR, will adjust my editor and update the PR. Strange that the `.editorconfig` isn't taking effect in CLion?

src/zoom_type.h
```diff
@@ -35,7 +35,7 @@ enum ZoomLevel : byte {
 	ZOOM_LVL_VIEWPORT = ZOOM_LVL_OUT_4X, ///< Default zoom level for viewports.
 	ZOOM_LVL_NEWS     = ZOOM_LVL_OUT_4X, ///< Default zoom level for the news messages.
 	ZOOM_LVL_INDUSTRY = ZOOM_LVL_OUT_8X, ///< Default zoom level for the industry view.
-	ZOOM_LVL_TOWN     = ZOOM_LVL_OUT_8X, ///< Default zoom level for the town view.
+	ZOOM_LVL_TOWN     = ZOOM_LVL_OUT_4X, ///< Default zoom level for the town view.
```
Bouke, line 38: Note: in `town_gui.cpp:334` the wrong constant was used (`ZOOM_LVL_NEWS`). This has been corrected and the constant has been updated to reflect the current zoom level for towns.

src/news_gui.cpp
```diff
@@ -311,7 +311,8 @@ struct NewsWindow : Window {
 		/* Initialize viewport if it exists. */
 		NWidgetViewport *nvp = this->GetWidget<NWidgetViewport>(WID_N_VIEWPORT);
 		if (nvp != nullptr) {
-			nvp->InitializeViewport(this, ni->reftype1 == NR_VEHICLE ? 0x80000000 | ni->ref1 : (uint32)GetReferenceTile(ni->reftype1, ni->ref1), ZOOM_LVL_NEWS);
+			auto zoom_level = ZoomLevel(ZOOM_LVL_NEWS + _gui_zoom - ZOOM_LVL_OUT_4X); // adjust zoom level for GUI scaling (HiDPI)
```
Bouke: I probably should introduce a helper method for adjusting the zoom level, especially as it's been repeated a few times for every viewport flavour.

src/zoom_func.h
```diff
@@ -68,6 +68,26 @@ static inline int UnScaleGUI(int value)
...
 }
 
+/**
+ * Scale zoom level relative to GUI zoom.
+ * @param value zoom level to scale
+ * @return scaled zoom level
+ */
+static inline ZoomLevel ScaleZoomGUI(ZoomLevel value)
+{
+	return ZoomLevel(value + (ZOOM_LVL_GUI - ZOOM_LVL_OUT_4X));
```
frosch123: I guess this method is similar to ScaleGUITrad in that it scales constants defined for the original/traditional zoom level.
Semantically ZOOM_LVL_OUT_4X is an arbitrary value, which changes meaning, if we ever add more zoom levels. The traditional zoom level is ZOOM_LVL_BASE.
```suggestion
	return ZoomLevel(value + (ZOOM_LVL_GUI - ZOOM_LVL_BASE));
```
Maybe the `ScaleZoomGUI` could be named in a ways, so it groups/associated better to the existing ScaleGUITrad, since both are used in similar context.

Bouke: I originally picked `ZOOM_LVL_OUT_4X` because that's the constant that's being defaulted to by the video driver:

https://github.com/OpenTTD/OpenTTD/blob/2d2a5dd9661f842b0399d367883e6bbf1f19cdca/src/video/video_driver.hpp#L176-L178

However I'm fine with the change; maybe we should update the video driver's default as well then?

Bouke: > Maybe the ScaleZoomGUI could be named in a ways, so it groups/associated better to the existing ScaleGUITrad, since both are used in similar context.

How about `ScaleZoomGUITrad` and `UnScaleZoomGUITrad`?

frosch123:
When changing the interface-zoom setting in-game, while windows with viewports are opened, the windows resize (ReInit), but the zoom levels remain the same. Is there an easy fix for this?

src/zoom_func.h
```diff
@@ -68,6 +68,26 @@ static inline int UnScaleGUI(int value)
...
+}
+
+/**
+ * UnScale zoom level relative to GUI zoom.
+ * @param value zoom level to scale
+ * @return un-scaled zoom level
+ */
+static inline ZoomLevel UnScaleZoomGUI(ZoomLevel value)
+{
+	return ZoomLevel(value - (ZOOM_LVL_GUI - ZOOM_LVL_OUT_4X));
```
frosch123: Hmm, wouldn't these functions need some clamping to the range [ZOOM_LVL_BEGIN, ZOOM_LVL_END)?
I am also unsure what happens, if the max-zoom-in level is limited in the settings, maybe that is a limiting factor as well?

src/zoom_func.h
Bouke: > I guess this method is similar to ScaleGUITrad in that it scales constants defined for the original/traditional zoom level.

Indeed, however the default viewport's `ZoomLevel`s are defined assuming a GUI zoom of `ZOOM_LVL_OUT_4X`. 

> Semantically ZOOM_LVL_OUT_4X is an arbitrary value, which changes meaning, if we ever add more zoom levels. The traditional zoom level is ZOOM_LVL_BASE.

`ZOOM_LVL_BASE` is currently defined to be `2` while I need `ZOOM_LVL_OUT_4X` (currently defined as `4`), as that's what's being used here:

https://github.com/OpenTTD/OpenTTD/blob/2d2a5dd9661f842b0399d367883e6bbf1f19cdca/src/video/video_driver.hpp#L178

src/saveload/misc_sl.cpp
```diff
@@ -39,7 +39,7 @@ void SaveViewportBeforeSaveGame()
 	if (w != nullptr) {
 		_saved_scrollpos_x = w->viewport->scrollpos_x;
 		_saved_scrollpos_y = w->viewport->scrollpos_y;
-		_saved_scrollpos_zoom = w->viewport->zoom;
+		_saved_scrollpos_zoom = UnScaleZoomGUI(w->viewport->zoom);
```
frosch123: I am not sure whether this scale/unscale in the saveload code makes sense.
If you are on ZOOM_LVL_OUT_32X before saving, this would try to save non-existing 64X.
If it is clamped to 32X in the savegame, then loading the game will scale it to 16X, so different that before.

I think this change of zoom-levels in save/load cycles is weird. Maybe it's better to just store/restore the zoom-level as-it, and not try to adjust the zoom level in case you switch the display between save/load?

Bouke:
> When changing the interface-zoom setting in-game, while windows with viewports are opened, the windows resize (ReInit), but the zoom levels remain the same. Is there an easy fix for this?

This is not in relation to my PR? As the behaviour you're describing is also in 12.2 on my machine.

What behaviour do you want? Should the viewports change zoom, or don't you want them to resize? -- The resizing is caused by the windows resizing to their new minimal size. So when you go from 1x to 2x, the windows double in size. 

The zoom is interesting though, what would be the expected zoom level of viewports when the GUI is doubled in size? Should the viewport's zoom be scaled according to the GUI, or the default zoomlevel? I _think_ the latter, while this PR implements the former.

src/zoom_func.h
frosch123: Yeah, video_driver.hpp would also break if there would be more zoom levels added.

src/saveload/misc_sl.cpp
Bouke: Saving non-existing 64X would be bad and the value would need to be clamped on load (if that's not done already). I think it would be good to store a GUI-zoom independent zoom level though, so save games load consistently. Indeed between different screens on the same machine, but also when save games are shared between players. I can split it off into a separate PR so it can be reviewed/discussed separately?

Bouke: I've split this off into #10007.

src/zoom_func.h
Bouke: Resolving this as I'm currently using the same constant as the method I'm matching.

src/zoom_func.h
Bouke: I've updated the code to clamp to the range `[ZOOM_LVL_MIN, ZOOM_LVL_MAX]`. Applying the settings is not something that I feel these methods should do: that's up to call-site to apply the settings (if needed).

Bouke:
I've rebased the PR onto master. I think it's good to go as-is.

michicc (approved):

PR #9997, opened 2022-09-04
Fix: title positioning on HiDPI displays

Bouke:
## Motivation / Problem

The title positioning on the main screen is incorrect on HiDPI displays; the offsets/padding isn't adjusted for the zoom. This causes the title to display at a different position when the window is shown on a retina vs non-retina display (macOS).

## Description

fix2

## Limitations

None afaik.

## Checklist for review

### DIFF
--- a/src/main_gui.cpp
+++ b/src/main_gui.cpp
@@ -248,8 +248,8 @@ struct MainWindow : Window
 		this->DrawWidgets();
 		if (_game_mode == GM_MENU) {
 			static const SpriteID title_sprites[] = {SPR_OTTD_O, SPR_OTTD_P, SPR_OTTD_E, SPR_OTTD_N, SPR_OTTD_T, SPR_OTTD_T, SPR_OTTD_D};
-			static const uint LETTER_SPACING = 10;
-			int name_width = (lengthof(title_sprites) - 1) * LETTER_SPACING;
+			uint letter_spacing = ScaleGUITrad(10);
+			int name_width = (lengthof(title_sprites) - 1) * letter_spacing;
 
 			for (uint i = 0; i < lengthof(title_sprites); i++) {
 				name_width += GetSpriteSize(title_sprites[i]).width;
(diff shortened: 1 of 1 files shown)


src/main_gui.cpp
```diff
@@ -247,18 +247,19 @@ struct MainWindow : Window
 	{
 		this->DrawWidgets();
 		if (_game_mode == GM_MENU) {
+			int scaling_factor = ZOOM_LVL_OUT_4X / _gui_zoom;
```
nielsmh: This is wrong, the zoom level is a logarithmic scale not linear, and the most zoomed in (largest scaling factor) has a value of 0 so this would cause division by zero at 4x GUI scale. Use the `UnScaleByZoom` function to scale GUI elements.

Bouke: Thanks for the pointer. I've updated the PR accordingly.

src/main_gui.cpp
```diff
@@ -248,17 +248,17 @@ struct MainWindow : Window
 		this->DrawWidgets();
 		if (_game_mode == GM_MENU) {
 			static const SpriteID title_sprites[] = {SPR_OTTD_O, SPR_OTTD_P, SPR_OTTD_E, SPR_OTTD_N, SPR_OTTD_T, SPR_OTTD_T, SPR_OTTD_D};
-			static const uint LETTER_SPACING = 10;
-			int name_width = (lengthof(title_sprites) - 1) * LETTER_SPACING;
+			uint letter_spacing = UnScaleByZoom(40, _gui_zoom);
```
frosch123: There is `ScaleGUITrad` to scale traditional pixel values. It will also work in the future, when the base-zoom-level is changed again for 8k screens.
```suggestion
			uint letter_spacing = ScaleGUITrad(10);
```

Bouke: This is great, thank you for the pointer.

frosch123 (approved):

PeterN:
Is it worth assigning ScaleGUITrad(50) to a variable outside the loop? I suppose compilers are good at optimising.

PR #10001, opened 2022-09-04
Feature: [GS] Scriptable league tables

ldpl:
Default company league table came from the original game and has long lost its purpose since. As the base game has no goal it's kind of pointless to rank companies by some arbitrary rating that doesn't even facilitate good gameplay. So while adding goals to the game is the job of gamescripts, right now they have no way to change the league table mechanics having to misuse global goals for that purpose leading to player confusion.
So I designed a GS API for league tables based on the "league tables" of CityMania servers. As they're the most complex ones I've seen and cover all other use cases I know of, nicely showcasing:
- There may be multiple league tables, not just one.
- Score can be of some custom type like time.
- Ordering can also be quite complex (companies that reached the goal are ordered by time, who didn't by population). 
- It can include companies that are not in the game anymore
- May need to include custom information (e.g. list of players).

Screenshot from 2022-08-28 11-32-32

So I kind of got it all working but before I go finish all the boring technical details I'd like to see what opinions there are on this PR in general and some details in particular:

- Naming of things. Currently, I called the table itself LeagueTable and its elements (rows) LeagueTableElement like in story pages. But it seems kinda long and generic, should I just call them rows? Also, I added `Link`, `LinkType` and `LinkTargetID` types that currently are kind of out of the scope of LeagueTables but are only used by them. The idea is to eventually use them in other scriptable UI like goals or story book but I don't want to do that in this PR. So should I name them like LeagueTableLink for now? I'm also looking for better names for all the other stuff.

- Default league table and scoring. As I said, I don't see much point in keeping it in the game anymore. But in case someone feels strongly about it I kinda wanted to avoid removing it in this PR. So far I made it so that it shows default league menu options (league table, detailed rating, highscores) only if there are no custom league tables in the game. If there are it only shows the list of tables as GS author can implement highscores as a custom league table if they wish so and making custom rating details is just something I don't see possible without a full framework for custom UI.

- Attributes. LeagueTable is quite simple and only has a title, header and footer. 
    LeagueTableElement though has a lot:
    - `table` - id of the table it belongs to
    - `rating` - hidden value used only for ordering elements in the table.
    - `score` - string with a visible score to show.
    - `text` - main text of the element (i.e. company name).
    - `link` - `Link(target_type, target_id)` for the object to open when this element is clicked.

    So I separated the logical "score" used for sorting and the visual one because, as I already said, the score can be quite complex and I don't think the game should limit it to certain types. And yet it needs some value to use for sorting the elements.
 
- GS API. Currently, I've added the `GSLeagueTable` class with the following methods:
    - `New(title, header, footer) -> LeagueTableID` - Add a new league table.
    - `NewElement(table, rating, company, text, score, link_type, link_target) -> LeagueTableElementID` - Add an element (row) to the table.
    - `UpdateElementData(element, company, text, link_type, link_target)`.
    - `UpdateElementScore(element, rating, score)`.
    - `RemoveElement(element)`.
    I decided to split updating into two methods as IMO score is updated much more often and in different places than data is.

TODO:
- [x] Saveload
- [x] RTL
- [x] Comments

GS to test this (implements default league table in GS): 
[testleaguetable.zip](https://github.com/OpenTTD/OpenTTD/files/9877171/testleaguetable.zip)
Screenshot from 2022-09-05 01-14-45


src/os/macosx/macos.mm
```diff
@@ -1,6 +1,6 @@
 /*
  * This file is part of OpenTTD.
- * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is free software; you can redis@ibute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
```
James103: Please revert that change.

2TallTyler:
I haven't reviewed the code or tested this, but I think this is an excellent path forward from the current scoring system. As broken as it is, I agree with your decision to keep it as a fallback when no custom scoring is present.

npabisz:
As owner of GS server I agree that whole system of goals should be reworked, I mean its 2022. This game should put more attention into multiplayer.

ldpl:
Since no one commented on the api or naming I went on with what I had and finished the PR.

James103:
Reimplementing the CityMania goal tables in the new league table system as it currently stands (with no header or footer) would mean losing the ability to have a detailed description of what the goal is or scores mean.

For example, without a header or footer, the league table would only have a title of "Goal: 10,000 people", the elements, and nothing else. With a header, it can be used to show the full goal and what the scores mean. With a footer, it can be used to show something along the lines of "... and X more" if there are too many companies/entries to show at once.

Therefore, league tables should (in this implementation) allow for a header and/or footer.

ldpl:
I added header a footer. Doesn't look particularly nice but so is the rest of scriptable UI. 
I've updated the test GS to showcase them.

P.S. Stupid github doesn't want to attach new screenshot for some reason :(

src/league_cmd.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Create a new league table.
+ * @param flags type of operation
+ * @param title Title of the league table
+ * @param header Text to show above the table
+ * @param footer Text to show below the table
+ * @return the cost of this operation or an error
+ */
+std::tuple<CommandCost, LeagueTableID>  CmdCreateLeagueTable(DoCommandFlag flags, const std::string &title, const std::string &header, const std::string &footer)
```
LordAro:
```suggestion
std::tuple<CommandCost, LeagueTableID> CmdCreateLeagueTable(DoCommandFlag flags, const std::string &title, const std::string &header, const std::string &footer)
```

src/league_cmd.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+#include "safeguards.h"
+
+LeagueTableElementPool _league_table_element_pool("LeagueTableElement");
+INSTANTIATE_POOL_METHODS(LeagueTableElement)
+
+LeagueTablePool _league_table_pool("LeagueTable");
+INSTANTIATE_POOL_METHODS(LeagueTable)
+
+bool IsValidLink(Link link) {
```
LordAro: opening function brace should be on a newline

\+ several others

src/league_gui.cpp
```diff
@@ -0,0 +1,453 @@
...
+};
+
+static inline StringID GetPerformanceTitleFromValue(uint value)
+{
+	return _performance_titles[std::min(value, 1000u) >> 6];
+}
+
+class PerformanceLeagueWindow : public Window {
+private:
+	GUIList<const Company*> companies;
```
LordAro:
```suggestion
	GUIList<const Company *> companies;
```

LordAro: I know this was in the original code, but why does there need to be a copy of the companies list stored within the window?

src/league_gui.cpp
```diff
@@ -0,0 +1,453 @@
...
+
+	if (_ctrl_pressed) {
+		ShowExtraViewportWindow(xy);
+	} else {
+		ScrollMainWindowToTile(xy);
+	}
+}
+
+
+class ScriptLeagueWindow : public Window {
```
LordAro, line 254: Could do with some doxy comments throughout

src/league_type.h
```diff
@@ -0,0 +1,40 @@
...
+
+struct Link {
+	LinkType type;
+	LinkTargetID target;
+	Link(LinkType type, LinkTargetID target): type{type}, target{target} {}
+	Link(): Link(LT_NONE, 0) {}
+};
+
+typedef uint8 LeagueTableID; ///< ID of a league table
+struct LeagueTable;
```
LordAro, line 33: Feels odd to have these forward declared in a header. Is that normal compared to other types?

src/script/api/script_league.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+
+/* static */ bool ScriptLeagueTable::IsValidLeagueTable(LeagueTableID table_id)
+{
+	return ::LeagueTable::IsValidID(table_id);
+}
+
+/* static */ ScriptLeagueTable::LeagueTableID ScriptLeagueTable::New(Text *title, Text *header, Text *footer)
+{
+	CCountedPtr<Text> title_counter(title);
```
LordAro, line 29: this feels very std::shared_ptr... but that's something for the future.

src/toolbar_gui.cpp
```diff
@@ -669,60 +670,96 @@ static CallBackFunction MenuClickGoal(int index)
...
-}
+/* --- Graphs and League Table button menu --- */
 
 /**
- * Handle click on the entry in the Graphs menu.
- *
- * @param index Graph to show.
- * @return #CBF_NONE
+ * Enum for the League Toolbar's and Graph Toolbar's related buttons.
+ * Use continuos numbering as League Toolbar can be combined into the Graph Toolbar.
```
LordAro:
```suggestion
 * Use continuous numbering as League Toolbar can be combined into the Graph Toolbar.
```

src/toolbar_gui.cpp
```diff
@@ -669,60 +670,96 @@ static CallBackFunction MenuClickGoal(int index)
...
-	return CBF_NONE;
-}
+/* --- Graphs and League Table button menu --- */
 
 /**
- * Handle click on the entry in the Graphs menu.
- *
- * @param index Graph to show.
- * @return #CBF_NONE
+ * Enum for the League Toolbar's and Graph Toolbar's related buttons.
```
LordAro, line 679: Not actually an enum. Could it be?

LordAro (commented):
No issues in principle. Mostly nitpicks

src/league_gui.cpp
ldpl: As I understand it, because it's inconvenient to keep list constantly updated, so to avoid glitches due to resorting it stores a copy and updates it episodically with full redraw of the window.

src/league_type.h
ldpl: I was copying `story_type.hpp`, seems pretty common.

src/toolbar_gui.cpp
ldpl: Possibly, but `CTMN_*` in the same file isn't technically a enum either but is called so in the comment.

src/league_gui.cpp
```diff
@@ -0,0 +1,451 @@
...
+			SetDParamStr(0, lt->header);
+			ir.top = DrawStringMultiLine(ir.left, ir.right, ir.top, UINT16_MAX, STR_JUST_RAW_STRING, TC_BLACK);
+		}
+
+		int icon_y_offset = (this->line_height - this->icon_size.height) / 2;
+		int text_y_offset = (this->line_height - FONT_HEIGHT_NORMAL) / 2;
+
+		uint rank_left, rank_right, icon_left, text_left, text_right, score_left, score_right;
+		int spacer = WidgetDimensions::scaled.framerect.left + WidgetDimensions::scaled.framerect.right;
+		if (_current_text_dir == TD_RTL) {
```
PeterN: Without changing too much how it's all calculated, something like this should work (and handle both ltr & rtl):

```
Rect rank = ir.WithWidth(this->rank_width, rtl);
Rect icon = ir.Indent(this->rank_width + spacer, rtl).WithWidth(this->icon_size.width, rtl);
Rect text = ir.Indent(this->rank_width + spacer + this->icon_size.width, rtl).WithWidth(this->text_width, rtl);
Rect score = ir.Indent(this->rank_width + spacer + this->icon_size.width + this->text_width, rtl).WithWidth(this->score_width, rtl);
```

`spacer` should be either `hsep_normal` or `hsep_wide`, depending how you want it to look.

This basically ignores `ir.right` / Width, but you've precalculated your widths so that should be okay.

src/league_gui.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+	void UpdateWidgetSize(int widget, Dimension *size, const Dimension &padding, Dimension *fill, Dimension *resize) override
+	{
+		if (widget != WID_SLT_BACKGROUND) return;
+
+		auto lt = LeagueTable::GetIfValid(this->table);
+		if (lt == nullptr) return;
+
+		this->icon_size = GetSpriteSize(SPR_COMPANY_ICON);
+		this->line_height = std::max<int>(this->icon_size.height + ScaleGUITrad(2), FONT_HEIGHT_NORMAL);
```
PeterN: Could be `WidgetDimensions::scaled.fullbevel.Vertical()` if you want to avoid a magic 2.

(`fullbevel` is `bevel` but always scaled, regardless of the scaled bevels option...)

src/league_gui.cpp
ldpl: Added some

src/toolbar_gui.cpp
```diff
@@ -2020,7 +2057,7 @@ struct MainToolbarWindow : Window {
 			case MTHK_STORY: ShowStoryBook(_local_company); break;
 			case MTHK_GOAL: ShowGoalsList(_local_company); break;
 			case MTHK_GRAPHS: ShowOperatingProfitGraph(); break;
-			case MTHK_LEAGUE: ShowCompanyLeagueTable(); break;
+			case MTHK_LEAGUE: ShowPerformanceLeagueTable(); break;  // FIXME
```
LordAro: ?

LordAro (dismissed):
LGTM. Would appreciate another set of eyes on it

src/league_cmd.h
```diff
@@ -0,0 +1,29 @@
...
+/** @file league_cmd.h Command definitions related to league tables. */
+
+#ifndef LEAGUE_CMD_H
+#define LEAGUE_CMD_H
+
+#include "league_type.h"
+#include "command_type.h"
+#include "company_type.h"
+
+std::tuple<CommandCost, LeagueTableID>  CmdCreateLeagueTable(DoCommandFlag flags, const std::string &title, const std::string &header, const std::string &footer);
```
2TallTyler:
```suggestion
std::tuple<CommandCost, LeagueTableID> CmdCreateLeagueTable(DoCommandFlag flags, const std::string &title, const std::string &header, const std::string &footer);
```

src/league_cmd.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+#include "safeguards.h"
+
+LeagueTableElementPool _league_table_element_pool("LeagueTableElement");
+INSTANTIATE_POOL_METHODS(LeagueTableElement)
+
+LeagueTablePool _league_table_pool("LeagueTable");
+INSTANTIATE_POOL_METHODS(LeagueTable)
+
+bool IsValidLink(Link link)
```
2TallTyler, line 34: Doxygen?

src/league_cmd.cpp
```diff
@@ -0,0 +1,170 @@
...
+ */
+CommandCost CmdUpdateLeagueTableElementData(DoCommandFlag flags, LeagueTableElementID element, CompanyID company, const std::string &text, LinkType link_type, LinkTargetID link_target)
+{
+	if (_current_company != OWNER_DEITY) return CMD_ERROR;
+	auto lte = LeagueTableElement::GetIfValid(element);
+	if (lte == nullptr) return CMD_ERROR;
+
+	Link link{link_type, link_target};
+	if (!IsValidLink(link)) return CMD_ERROR;
+	if (company != INVALID_COMPANY && !Company::IsValidID(company)) return CMD_ERROR;
```
2TallTyler: This seems grouped in an odd place with `link` checks. Maybe it belongs in the section above, near the check that enforces `_current_company == OWNER_DEITY`?

src/league_gui.cpp
```diff
@@ -0,0 +1,454 @@
...
+	return _performance_titles[std::min(value, 1000u) >> 6];
+}
+
+class PerformanceLeagueWindow : public Window {
+private:
+	GUIList<const Company *> companies;
+	uint ordinal_width; ///< The width of the ordinal number
+	uint text_width;    ///< The width of the actual text
+	int line_height;    ///< Height of the text lines
+	Dimension icon;     ///< Dimenion of the company icon.
```
2TallTyler:
```suggestion
	Dimension icon;     ///< Dimension of the company icon.
```

src/league_gui.cpp
```diff
@@ -0,0 +1,454 @@
...
+		this->line_height = std::max<int>(this->icon.height + WidgetDimensions::scaled.vsep_normal, FONT_HEIGHT_NORMAL);
+
+		for (const Company *c : Company::Iterate()) {
+			SetDParam(0, c->index);
+			SetDParam(1, c->index);
+			SetDParam(2, _performance_titles[widest_title]);
+			widest_width = std::max(widest_width, GetStringBoundingBox(STR_COMPANY_LEAGUE_COMPANY_NAME).width);
+		}
+
+		this->text_width = widest_width + WidgetDimensions::scaled.hsep_indent * 3; // Keep some extra spacing
```
2TallTyler, line 160: Multiplying `hsep_indent` by a magic number to create some new non-standard amount of space isn't particularly pretty. Is there another constant that could be used, or created?

src/league_gui.cpp
```diff
@@ -0,0 +1,454 @@
...
+	0,
+	_nested_performance_league_widgets, lengthof(_nested_performance_league_widgets)
+);
+
+void ShowPerformanceLeagueTable()
+{
+	AllocateWindowDescFront<PerformanceLeagueWindow>(&_performance_league_desc, 0);
+}
+
+
```
2TallTyler: Extra newline?

src/league_gui.cpp
```diff
@@ -0,0 +1,454 @@
...
+static void HandleLinkClick(Link link)
+{
+	/* Determine dst coordinate for goal and try to scroll to it. */
```
2TallTyler: `dst`?

src/league_gui.cpp
ldpl: It does happen in other places: https://github.com/OpenTTD/OpenTTD/blob/master/src/cheat_gui.cpp#L244
Idk any better way to do it.

src/league_gui.cpp
PeterN: Hehe, copy & paste of my typo :-)

src/league_gui.cpp
PeterN: Copy & paste of my changes too, it's basically "make some arbitrary space", which may or may not be equal in different places.

LordAro (approved):

PR #10003, opened 2022-09-05
Fix #9903: Don't build houses around half-tile roads after a bridge or tunnel

### LINKED ISSUE #9903: [Bug]: Towns incorrectly build houses at half-tile roads connected to bridge or tunnel, when towns are allowed to build roads is off
### Version of OpenTTD

12.1, macOS 11.5.1

### Expected result

Towns should not be allowed to build houses at the half-tile dead end roads

### Actual result

Sunningstone Transport, 1962-06-15
From left to right, the second and fourth roads have the unexpected behaviour of building houses at dead ends

### Steps to reproduce

1. Turn off town are allowed to build roads
2. Build half-tile dead end roads at the end of bridges or tunnels, similar to the examples in the screenshot above
3. Grow the town

ldpl:
## Motivation / Problem

Closes #9903

## Description

Adds a check for the half-tile road before building a house. 
AFACT only matters if it's the end of the tunnel/bridge or starting tile. 
Didn't check for `_settings_game.economy.allow_town_roads` as that seems irrelevant since towns don't build half-tile roads anyway.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -1622,6 +1622,9 @@ static TownGrowthResult GrowTownInTile(TileIndex *tile_ptr, RoadBits cur_rb, Dia
 			return TownGrowthResult::Continue;
 		}
 
+		/* Don't build houses around half-tile roads in the rare case we ended up on one. */
+		if (HasExactlyOneBit(cur_rb)) allow_house = false;
+
 		/* Possibly extend the road in a direction.
 		 * Randomize a direction and if it has a road, bail out. */
 		target_dir = RandomDiagDir();


LordAro (dismissed):

PeterN:
Why not? I'm not clear on the conditions here. Plenty of towns in my games have dead-end half-tile roads with houses around them where it wouldn't make sense not to have a house..

2TallTyler:
Building houses at half-tile dead-ends tends to stunt town growth, both by blocking potential pathways for growth and by creating stub ends which cause the growth walk to end without growing the town.

I'm +1 to this change.

2TallTyler:
end

Hmm, managed to break it with a company-owned road.

ldpl:
> Hmm, managed to break it with a company-owned road.
I can't reproduce this, can you explain in a detail what you did here? And, first of all, are town allowed to build their own roads? If they are this is normal and towns are allowed to continue into half-tile roads as those can be extended.

ldpl:
Looking at this PR now everyone seems confused to what it actually does. It does not forbid towns to build around half-tile roads. That was already done long ago. But only when towns are not allowed to build their own roads. What this does is removes a single exception to that rule found in #9903.

2TallTyler:
Re-reading this months later, I see now that this only applies to town-owned roads immediately after a bridge or tunnel. In this case, I am happy to approve this, although I believe it'll need a rebase and force-push for CI to be okay with merging (the commit is too far back). 🙂

2TallTyler:
I think you forgot to fetch and rebase, commit checker is failing for exactly what I predicted. 🙂 
`No common parent found for this merge commit (max-depth of 256 reached)`

ldpl:
I remembered to fetch but not to rebase xD

zephyris:
Rebased because I like the fix to preventing houses around the dead end roads after bridges/tunnels, but reviewing the code I can't see why it wouldn't affect all dead end roads. Am I missing something?

src/town_cmd.cpp
```diff
@@ -1622,6 +1622,9 @@ static TownGrowthResult GrowTownInTile(TileIndex *tile_ptr, RoadBits cur_rb, Dia
 			return TownGrowthResult::Continue;
 		}
 
+		/* Don't build houses around half-tile roads in the rare case we ended up on one. */
+		if (HasExactlyOneBit(cur_rb)) allow_house = false;
```
zephyris, line 1626: I don't see how this only affects half tile roads after bridges, surely would need to be part of the bridge teleport check above?

PR #10004, opened 2022-09-05
Fix: Don't stop growth attempt if house is not allowed

ldpl:
## Motivation / Problem

When OpenTTD tries to build a house but is not allowed it stops the growth attempt completely even though it could've succeeded further down that road. AFACT there are two distinct cases that are affected (I'm using cmclient for screenshots since it shows stopped growth with the orange square):
- Blocking grid intersections with stations or one-way roads when towns are allowed to build roads;
Screenshot from 2022-09-06 02-13-51
- Using road types that don't allow houses
Screenshot from 2022-09-06 02-15-01

In both cases, it results in a drastic reduction of the effective growth speed as it needs to chance on a road tile when picking the house spot for it to fail early and the search to continue.

## Description

Simply don't set the growth stopping flag if the house spot was not allowed.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -1593,8 +1593,6 @@ static void GrowTownInTile(TileIndex *tile_ptr, RoadBits cur_rb, DiagDirection t
 		}
 
 		if (!TownCanGrowRoad(tile)) return;
-
-		_grow_town_result = GROWTH_SEARCH_STOPPED;
 	}
 
 	/* Return if a water tile */


LordAro (dismissed):

frosch123:
This change does not make a lot of sense to me.
With the same argument you could remove all usages of `GROWTH_SEARCH_STOPPED`.
It's always set before attempting to build a road.

ldpl:
@frosch123 I haven't looked into the road building logic, maybe it shouldn't stop growth in some cases as well. But for houses I explained two cases where it clearly shouldn't stop as it kills the growth with for no apparent reason. Why should town stop growing when player just put some bus stops on roads?

2TallTyler:
I wonder if it would be possible (and preferable) to explicitly handle these cases, rather than implicitly grouping all the growth failures together.

2TallTyler:
Do you have a fancy cmclient screenshot of the failed growth tiles with this change, for comparison? I guess my concern would be running and running the growth search unnecessarily in situations where it should bail out, which is why I suggested explicitly handling these "house failed but keep looking" cases.

ldpl:
I can make client screenshot tomorrow but there would be nothing special to see there. Normally growth fails in dead ends and road loops, not on house failures.

ldpl:
IMG_20221121_185024_575
IMG_20221121_184903_991

PeterN:
I think the reasoning is something like:

This is a fix because when using a fixed layout town, some growth locations may be blocked by e.g. stations or one-way roads that affect town growth in a way that does not happen with a random town layout. This causes a reduction in the size of a town (when other variables are the same) over time for a fixed layout town vs a random layout town.

2TallTyler:
@ldpl, you said on Discord last December that you were reconsidering your approach to this, so I dismissed the PR approval. I haven't heard anything on the subject, but assume you aren't going this route. I'm going to close this PR to keep the review queue from getting too long, but if you change your mind feel free to re-open and let us know. 😃

ldpl:
I weren't reconsidering it. I entertained the frosch idea and while it possible to simplify the code by removing _grow_town_result it's not really related to this pr and depends on it being merged first anyway: https://github.com/OpenTTD/OpenTTD/compare/master...ldpl:OpenTTD:fix-growth-result

P.S. I don't think I can reopen bugs.

2TallTyler:
So you want this reopened for consideration?

ldpl:
> This change does not make a lot of sense to me. With the same argument you could remove all usages of `GROWTH_SEARCH_STOPPED`. It's always set before attempting to build a road.
> 
> If road/tunnel/bridge construction suceeds it is replaced with `GROWTH_SUCCEED`, otherwise it stops.

Though reading this comment again it doesn't seem to mean that at all xD It's just plain wrong as building road is non-recoverable so it should always stop but failing house placement can and should just continue down the road.

Also, that refactoring branch is probably broken after #10343. But this PR should be fine.

PR #10006, opened 2022-09-06
Make modifier keys configurable and split Ctrl into remove and function.

ldpl:
## Motivation / Problem

Ctrl magic is a hot mess. There is no particular logic behind it leading to some conflicting situations like with rail where ctrl is used for removal and stations that use it for joining.

## Description

This PR tries to sort it out a little bit by separating it into two modifiers, one for remove (bulldozer) mode and one for miscellaneous functions. This allows using the remove modifier key with all the tools that support bulldozer mode including stations, waypoints, and signals. Unfortunately, there are only so many modifiers on the keyboard so with ctrl and shift already taken alt was the only one left and it's not very reliable as it can be used for os functions. So I added the settings to freely map modifier keys to their function allowing, for example, to map the remove mode to shift instead of estimation. I've already tried this change in CityMania patch pack and it's super handy.

**UPD** Also added `{FN_KEY}, {REMOVE_KEY}, and {ESTIMATE_KEY}` string codes to use in the tooltips. If modifier is not set it shows as "unassigned", in red color to indicate it's a feature, not a bug :p At least, that's the best solution I could come up with.

## Checklist for review

### DIFF
--- a/src/gfx_func.h
+++ b/src/gfx_func.h
@@ -52,8 +52,10 @@ extern byte _dirkeys;        ///< 1 = left, 2 = up, 4 = right, 8 = down
 extern bool _fullscreen;
 extern byte _support8bpp;
 extern CursorVars _cursor;
-extern bool _ctrl_pressed;   ///< Is Ctrl pressed?
-extern bool _shift_pressed;  ///< Is Shift pressed?
+extern bool _shift_pressed;    ///< Is Shift pressed?
+extern bool _fn_pressed;       ///< Is function modifier pressed?
+extern bool _remove_pressed;   ///< Is remove modifier pressed?
+extern bool _estimate_pressed; ///< Is esitame modifier pressed?
 extern uint16_t _game_speed;
 
 extern bool _left_button_down;
@@ -71,7 +73,7 @@ extern Palette _cur_palette; ///< Current palette
 void HandleToolbarHotkey(int hotkey);
 void HandleKeypress(uint keycode, char32_t key);
 void HandleTextInput(const char *str, bool marked = false, const char *caret = nullptr, const char *insert_location = nullptr, const char *replacement_end = nullptr);
-void HandleCtrlChanged();
+void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed);
 void HandleMouseEvents();
 void UpdateWindows();
 void ChangeGameSpeed(bool enable_fast_forward);
--- a/src/gfx_func.h
+++ b/src/gfx_func.h
@@ -73,7 +73,7 @@ extern Palette _cur_palette; ///< Current palette
 void HandleToolbarHotkey(int hotkey);
 void HandleKeypress(uint keycode, char32_t key);
 void HandleTextInput(const char *str, bool marked = false, const char *caret = nullptr, const char *insert_location = nullptr, const char *replacement_end = nullptr);
-void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed);
+void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed, bool alt_pressed);
 void HandleMouseEvents();
 void UpdateWindows();
 void ChangeGameSpeed(bool enable_fast_forward);
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -284,13 +284,13 @@ STR_TOOLTIP_CLOSE_WINDOW                                        :{BLACK}Close wi
 STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS                              :{BLACK}Window title - drag this to move window
 STR_TOOLTIP_SHADE                                               :{BLACK}Shade window - only show the title bar
 STR_TOOLTIP_DEBUG                                               :{BLACK}Show NewGRF debug information
-STR_TOOLTIP_DEFSIZE                                             :{BLACK}Resize window to default size. Ctrl+Click to store current size as default
-STR_TOOLTIP_STICKY                                              :{BLACK}Mark this window as uncloseable by the 'Close All Windows' key. Ctrl+Click to also save state as default
+STR_TOOLTIP_DEFSIZE                                             :{BLACK}Resize window to default size. {FN_KEY}+Click to store current size as default
+STR_TOOLTIP_STICKY                                              :{BLACK}Mark this window as uncloseable by the 'Close All Windows' key. {FN_KEY}+Click to also save state as default
 STR_TOOLTIP_RESIZE                                              :{BLACK}Click and drag to resize this window
 STR_TOOLTIP_TOGGLE_LARGE_SMALL_WINDOW                           :{BLACK}Toggle large/small window size
 STR_TOOLTIP_VSCROLL_BAR_SCROLLS_LIST                            :{BLACK}Scroll bar - scrolls list up/down
 STR_TOOLTIP_HSCROLL_BAR_SCROLLS_LIST                            :{BLACK}Scroll bar - scrolls list left/right
-STR_TOOLTIP_DEMOLISH_BUILDINGS_ETC                              :{BLACK}Demolish buildings etc. on a square of land. Ctrl+Click+Drag to select the area diagonally. Also press Shift to show cost estimate only
+STR_TOOLTIP_DEMOLISH_BUILDINGS_ETC                              :{BLACK}Demolish buildings etc. on a square of land. {FN_KEY}+Click+Drag to select the area diagonally. Also press {ESTIMATE_KEY} to show cost estimate only
 
 # Show engines button
 ###length VEHICLE_TYPES
@@ -314,6 +314,13 @@ STR_WARNING_PASSWORD_SECURITY                                   :{YELLOW}Warning
 # On screen keyboard window
 STR_OSK_KEYBOARD_LAYOUT                                         :`1234567890-=\qwertyuiop[]asdfghjkl;'  zxcvbnm,./ .
 STR_OSK_KEYBOARD_LAYOUT_CAPS                                    :~!@#$%^&*()_+|QWERTYUIOP{{}}ASDFGHJKL:"  ZXCVBNM<>? .
+
+# Modifier key names, must follow order of ModifierKey enum
+###length 4
+STR_MODIFIER_KEY_UNASSIGNED                                     :{PUSH_COLOUR}{RED}Unassigned{POP_COLOUR}
+STR_MODIFIER_KEY_SHIFT                                          :Shift
+STR_MODIFIER_KEY_CTRL                                           :Ctrl
+STR_MODIFIER_KEY_ALT                                            :Alt
 
 # Measurement tooltip
 STR_MEASURE_LENGTH                                              :{BLACK}Length: {NUM}
@@ -395,10 +402,10 @@ STR_TOOLBAR_TOOLTIP_DISPLAY_GOALS_LIST                          :{BLACK}Open goa
 STR_TOOLBAR_TOOLTIP_DISPLAY_GRAPHS                              :{BLACK}Open company graphs and cargo payment rates
 STR_TOOLBAR_TOOLTIP_DISPLAY_COMPANY_LEAGUE                      :{BLACK}Open company league table
 STR_TOOLBAR_TOOLTIP_FUND_CONSTRUCTION_OF_NEW                    :{BLACK}Open industry directory, industry chain, or fund construction of a new industry
-STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_TRAINS              :{BLACK}Open list of company's trains. Ctrl+Click to show or hide vehicle groups, opposite of the chosen setting
-STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_ROAD_VEHICLES       :{BLACK}Open list of company's road vehicles. Ctrl+Click to show or hide vehicle groups, opposite of the chosen setting
-STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_SHIPS               :{BLACK}Open list of company's ships. Ctrl+Click to show or hide vehicle groups, opposite of the chosen setting
-STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_AIRCRAFT            :{BLACK}Open list of company's aircraft. Ctrl+Click to show or hide vehicle groups, opposite of the chosen setting
+STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_TRAINS              :{BLACK}Open list of company's trains. {FN_KEY}+Click to show or hide vehicle groups, opposite of the chosen setting
+STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_ROAD_VEHICLES       :{BLACK}Open list of company's road vehicles. {FN_KEY}+Click to show or hide vehicle groups, opposite of the chosen setting
+STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_SHIPS               :{BLACK}Open list of company's ships. {FN_KEY}+Click to show or hide vehicle groups, opposite of the chosen setting
+STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_AIRCRAFT            :{BLACK}Open list of company's aircraft. {FN_KEY}+Click to show or hide vehicle groups, opposite of the chosen setting
 STR_TOOLBAR_TOOLTIP_ZOOM_THE_VIEW_IN                            :{BLACK}Zoom in
 STR_TOOLBAR_TOOLTIP_ZOOM_THE_VIEW_OUT                           :{BLACK}Zoom out
 STR_TOOLBAR_TOOLTIP_BUILD_RAILROAD_TRACK                        :{BLACK}Build railway infrastructure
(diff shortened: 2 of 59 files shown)


src/lang/english.txt
```diff
@@ -3889,16 +3908,16 @@ STR_BUY_VEHICLE_SHIP_BUY_REFIT_VEHICLE_BUTTON                   :{BLACK}Buy and
 STR_BUY_VEHICLE_AIRCRAFT_BUY_REFIT_VEHICLE_BUTTON               :{BLACK}Buy and Refit Aircraft
 
 ###length VEHICLE_TYPES
-STR_BUY_VEHICLE_TRAIN_BUY_VEHICLE_TOOLTIP                       :{BLACK}Buy the highlighted train vehicle. Shift+Click shows estimated cost without purchase
-STR_BUY_VEHICLE_ROAD_VEHICLE_BUY_VEHICLE_TOOLTIP                :{BLACK}Buy the highlighted road vehicle. Shift+Click shows estimated cost without purchase
-STR_BUY_VEHICLE_SHIP_BUY_VEHICLE_TOOLTIP                        :{BLACK}Buy the highlighted ship. Shift+Click shows estimated cost without purchase
-STR_BUY_VEHICLE_AIRCRAFT_BUY_VEHICLE_TOOLTIP                    :{BLACK}Buy the highlighted aircraft. Shift+Click shows estimated cost without purchase
+STR_BUY_VEHICLE_TRAIN_BUY_VEHICLE_TOOLTIP                       :{BLACK}Buy the highlighted train vehicle. {ESTIMATE_KEY}{ESTIMATE_KEY}+Click shows estimated cost without purchase
```
James103: `{ESTIMATE_KEY}` is duplicated here.
```suggestion
STR_BUY_VEHICLE_TRAIN_BUY_VEHICLE_TOOLTIP                       :{BLACK}Buy the highlighted train vehicle. {ESTIMATE_KEY}+Click shows estimated cost without purchase
```

EmperorJake:
This would be great. On my Mac I've remapped the command key to the Fn key (which is where the Ctrl key should be) so that the CTRL-C, etc. key commands are the same as I'm used to on other computers. While this works great at the OS level, OpenTTD doesn't recognise it and still makes me use the Command key for things like direction flipping and signal dragging, while combination keystrokes like Ctrl-X were able to be remapped in the .cfg. This results in a very inconsistent experience where I have to remember to hit Command instead of Ctrl for certain things.

src/lang/english.txt
```diff
@@ -2797,14 +2816,14 @@ STR_STATION_BUILD_NOISE                                         :{BLACK}Noise ge
...
-STR_LANDSCAPING_TOOLTIP_PURCHASE_LAND                           :{BLACK}Purchase land for future use. Ctrl selects the area diagonally. Shift toggles building/showing cost estimate
+STR_LANDSCAPING_TOOLTIP_LOWER_A_CORNER_OF_LAND                  :{BLACK}Lower a corner of land. Dragging lowers the first selected corner and levels the selected area to the new corner height. {FN_KEY} selects the area diagonally. {ESTIMATE_KEY} toggles building/showing cost estimate
+STR_LANDSCAPING_TOOLTIP_RAISE_A_CORNER_OF_LAND                  :{BLACK}Raise a corner of land. Dragging raises the first selected corner and levels the selected area to the new corner height. {FN_KEY} selects the area diagonally. {ESTIMATE_KEY} toggles building/showing cost estimate
+STR_LANDSCAPING_LEVEL_LAND_TOOLTIP                              :{BLACK}Level an area of land to the height of the first selected corner. {FN_KEY} selects the area diagonally. {ESTIMATE_KEY} toggles building/showing cost estimate
+STR_LANDSCAPING_TOOLTIP_PURCHASE_LAND                           :{BLACK}Purchase land for future use. {ESTIMATE_KEY} toggles building/showing cost estimate
```
2TallTyler: I may have screwed this up while rebasing, sorry. 🙁 

```suggestion
STR_LANDSCAPING_TOOLTIP_PURCHASE_LAND                           :{BLACK}Purchase land for future use. {FN_KEY} selects the area diagonally. {ESTIMATE_KEY} toggles building/showing cost estimate
```

src/lang/english.txt
```diff
@@ -2630,16 +2649,16 @@ STR_RAIL_TOOLBAR_ELRAIL_CONSTRUCTION_CAPTION                    :Electrified Rai
...
-STR_RAIL_TOOLBAR_TOOLTIP_BUILD_RAILROAD_SIGNALS                 :{BLACK}Build railway signals. Ctrl toggles semaphore/light signals{}Dragging builds signals along a straight stretch of rail. Ctrl builds signals up to the next junction or signal{}Ctrl+Click toggles opening the signal selection window. Shift toggles building/showing cost estimate
-STR_RAIL_TOOLBAR_TOOLTIP_BUILD_RAILROAD_BRIDGE                  :{BLACK}Build railway bridge. Shift toggles building/showing cost estimate
-STR_RAIL_TOOLBAR_TOOLTIP_BUILD_RAILROAD_TUNNEL                  :{BLACK}Build railway tunnel. Shift toggles building/showing cost estimate
-STR_RAIL_TOOLBAR_TOOLTIP_TOGGLE_BUILD_REMOVE_FOR                :{BLACK}Toggle build/remove for railway track, signals, waypoints and stations. Hold Ctrl to also remove the rail of waypoints and stations
-STR_RAIL_TOOLBAR_TOOLTIP_CONVERT_RAIL                           :{BLACK}Convert/Upgrade the type of rail. Shift toggles building/showing cost estimate
+STR_RAIL_TOOLBAR_TOOLTIP_BUILD_RAILROAD_TRACK                   :{BLACK}Build railway track. {FN_KEY} toggles build/remove for railway construction. {ESTIMATE_KEY} toggles building/showing cost estimate
```
2TallTyler: Should this (and the next line) be {REMOVE_KEY}?

ldpl:
@2TallTyler I've rebased and fixed REMOVE_KEY

2TallTyler:
Code looks fine, but now that I've actually tried this change — not being able to bind both FN_KEY and REMOVE_KEY to Ctrl is a major and unwelcome change. I'm all for adding customizability (for accessibility, Mac OS keyboards, etc) but am not a fan of forcing players to re-learn a new hotkey for remove. Alt in particular is uncomfortable for me.

I would prefer to allow FN and REMOVE to use the same key, with optional separation into different keys.

ldpl:
@2TallTyler I've tried implementing it in compatible way in cmclient and didn't like it in the slightest. Having it support the old scheme messes up all the logic tremendously and kind of defeats the purpose of separation. Just consider the tooltips like "Does A, with Fn B, with remove C", they'll have to have a special version for compatibility case without C. Also, IMO, estimate is just a waste of a modifier key, I have bound Ctrl to Fn and Shift to remove, it's way more convenient than Alt, I just didn't want to mess up the default shift-click behavior in this PR.

2TallTyler:
What's the "purpose of separation," though? It sounds like you're trying to scratch a personal itch, which I have no problem with as long as it doesn't result in everyone else being forced out of their habits and opening angry bug reports. The backlash from hiding block signals was bad enough, and that was both possible and easy to revert to the old behavior.

I'd be happy to approve customizable modifier keys that default to the current behavior (I'd probably change my own settings to use Shift for Remove, as you suggest), but the current implementation is a hard no from me.

ldpl:
@2TallTyler the purpose of separation is to somewhat sort out the hotkey mess. You may think it's fine to have a hotkey for bulldozer mode that only works for half of the tools but I find it ridiculous. My personal itch is already scratched by cmclient, I just think it's the job of vanilla to make breaking changes and leave compatible crap for patchpacks.
Besides, I doubt that many people are using ctrl-removal and even if they do extending it to all tools is a nice trade-off for key change.

2TallTyler:
I don't think the current hotkey system is fine, I just don't think it's broken enough to put the target for breaking it on my back. If another maintainer wants to approve this, I won't complain.

nielsmh:
When building railroad track, holding Ctrl temporarily switches into removal mode. When building railroad signals, holding Ctrl builds a different kind of signals. That makes no sense. It will take a bunch of getting used to, but I think it will be better in the end.
A further extension (maybe down the line) would be to expand the Remove tool to work with anything. Currently, there are some items you can use Remove on and some you can't, and having a universal toggle-able Remove function would be great IMO.

However one issue I see with being able to take away the Estimate key is that you may lose the function for aborting a drag-build of something. Yes you can press Esc while dragging to cancel, but that also de-selects the tool, so you have to select the tool again.

ldpl:
Aborting with shift is kinda in the spacebar heating territory imo. I think eventual solution to that would be to make esc abort dragging and switch to tools that are less reliant on dragging like polyrail.

nielsmh:
Sorry for bikeshedding this, but looking at the code, I don't really like the usage of "Fn" for the generic modifier, because it corresponds to the label of a real key on many keyboards, but definitely does not _map_ to that key. Suggested alternatives: "Special" or "Super".

Additionally, I think it would be beneficial (also for the purpose of better accessibility) to name the variables `_build_mode.special`, `_build_mode.remove`, `_build_mode.estimate` or similar. Call it a "mode" instead of "pressed", and (perhaps later on) also give the player an option to configure toggles instead of holds for the functions.

And yes I agree that abusing Estimate for aborting building is not the right thing, it's just that it's the best available right now.

ldpl:
It can be said that Super is also a key ;) Anyway I don't really care for the naming if more people agree on smth like special I'll change it. It's not a player-facing name anyway, I think the only place where player sees it is in the settings.

Now, `_build_mode.special` is just weird imo. Smth like `_build_mode == BuildMode::Special` would be more appropriate syntax then or `_buildl_mode.is_special()`. 

And this PR doesn't remove Estimate mode so it's more of a concern for the future.

src/window.cpp
```diff
@@ -2687,14 +2687,35 @@ void HandleKeypress(uint keycode, WChar key)
...
 /**
- * State of CONTROL key has changed
+ * Handle the modifier key state changes.
  */
-void HandleCtrlChanged()
-{
-	/* Call the event, start with the uppermost window. */
-	for (Window *w : Window::IterateFromFront()) {
-		if (w->OnCTRLStateChange() == ES_HANDLED) return;
+void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed) {
```
rubidium42: Newline before `{`.

src/lang/english.txt
```diff
@@ -1009,7 +1009,7 @@ STR_GAME_OPTIONS_VIDEO_VSYNC_TOOLTIP                            :{BLACK}Check th
 STR_GAME_OPTIONS_VIDEO_DRIVER_INFO                              :{BLACK}Current driver: {RAW_STRING}
 
 STR_GAME_OPTIONS_GUI_SCALE_FRAME                                :{BLACK}Interface size
-STR_GAME_OPTIONS_GUI_SCALE_TOOLTIP                              :{BLACK}Drag slider to set interface size. Hold Ctrl for continuous adjustment
+STR_GAME_OPTIONS_GUI_SCALE_TOOLTIP                              :{BLACK}Drag slider to set interface size. Hold {FN_KEY} for continuous adjustment
```
rubidium42: All the changes of english.txt in the first commit should probably be moved to the last commit, as `{FN_KEY}` has not been added at this point yet.

src/window.cpp
```diff
@@ -2687,14 +2687,35 @@ void HandleKeypress(uint keycode, WChar key)
...
+	bool old_fn_pressed = _fn_pressed;
+	bool old_remove_pressed = _remove_pressed;
+
+	_shift_pressed = shift_pressed;
+
+	/* Hardwire modifiers leaving remove inaccessible for now. */
```
rubidium42: At this point you could still wire `_remove_pressed` to `ctrl_pressed`. Then this commit itself doesn't functionally change anything.

src/window.cpp
```diff
@@ -2689,15 +2689,15 @@ void HandleKeypress(uint keycode, WChar key)
 /**
  * Handle the modifier key state changes.
  */
-void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed) {
+void HandleModifierKeys(bool shift_pressed, bool ctrl_pressed, bool alt_pressed) {
```
rubidium42: Newline before `{`.

src/settings_table.cpp
```diff
@@ -443,6 +443,33 @@ static bool CheckMaxHeightLevel(int32 &new_value)
 	return true;
 }
 
+static bool CheckModifier(int32 &new_value, ModifierKey other1, ModifierKey other2)
```
rubidium42: Technically and practically you do not need the `&` here anymore, since you're not going to change the value.
I'd maybe even use `ModifierKey key` here as parameter, and do the cast in the functions below.

src/window.cpp
```diff
@@ -2689,15 +2689,15 @@ void HandleKeypress(uint keycode, WChar key)
...
 	bool old_fn_pressed = _fn_pressed;
 	bool old_remove_pressed = _remove_pressed;
 
 	_shift_pressed = shift_pressed;
 
-	/* Hardwire modifiers leaving remove inaccessible for now. */
+	/* Hardwire modifier keys. */
 	_fn_pressed = ctrl_pressed;
-	_remove_pressed = false;
+	_remove_pressed = alt_pressed;
```
rubidium42: One massive problem, for me at least, is that Alt + dragging moves the whole window of OpenTTD for me. This is on Linux with Xorg and LXDE, but I'm seeing queries on how to disable that for other window managers.

Essentially this makes it impossible for me to drag-remove anything.

src/gfx_func.h
```diff
@@ -52,8 +52,10 @@ extern byte _dirkeys;        ///< 1 = left, 2 = up, 4 = right, 8 = down
 extern bool _fullscreen;
 extern byte _support8bpp;
 extern CursorVars _cursor;
-extern bool _ctrl_pressed;   ///< Is Ctrl pressed?
-extern bool _shift_pressed;  ///< Is Shift pressed?
+extern bool _shift_pressed;    ///< Is Shift pressed?
+extern bool _fn_pressed;       ///< Is function modifier pressed?
```
rubidium42, line 56: I find the naming of this confusing. It has nothing to do with the 'Fn' button that exists on my keyboard.
Although I do not immediately have a better name for it. I'd almost say '_alternate_behaviour_pressed' or something similar, as it's an indication of alternate behaviour. However, that would somewhat conflict with the actual 'Alt' button on the keyboard.

src/window_gui.h
```diff
@@ -537,12 +537,18 @@ struct Window : ZeroedMemoryAllocator {
 	virtual EventState OnHotkey(int hotkey);
 
 	/**
-	 * The state of the control key has changed
+	 * The state of the functional modifier key has changed
```
rubidium42: It's 'function modifier' in other places.

src/strings.cpp
```diff
@@ -515,6 +515,12 @@ static char *FormatGenericCurrency(char *buff, const CurrencySpec *spec, Money n
 	return buff;
 }
 
+static char *FormatModifierKey(char *buff, ModifierKey key, const char *last) {
```
rubidium42: Newline before `{`.

src/strings.cpp
```diff
@@ -515,6 +515,12 @@ static char *FormatGenericCurrency(char *buff, const CurrencySpec *spec, Money n
...
+	int64 args[] = { 1 };
+	StringParameters tmp_params(args);
```
rubidium42: `StringParameters tmp_params(nullptr, 0, nullptr);`

rubidium42 (commented):
The Alt+drag combination being taken over by the window manager on (many?) Linux distributions by default is a major concern for me with this PR. So, if you ever wondered why there were no keyboard shortcuts with Alt in OpenTTD. This is the underlying issue.

It essentially means I have to make a choice between two of function, estimate and removal. As the third can't be used with dragging.

src/window.cpp
ldpl: Maybe but why? The whole point of this commit is to separate them.

src/window.cpp
ldpl: On whatever linuxes i used alt-dragging haven't been moving windows for at least 10 years and even back then there was an option to disable. And even if you for some reason don't want to disable it and don't want to reassign useless estimate on alt you can remove with bulldozer or bulldozer hotkey, that's still the better way to remove anything without this pr as ctrl only removes what in vanilla, just rails?

src/gfx_func.h
ldpl: Well, this is not very helpful. I asked chatgpt to give me some options and here is what it got, just pick any:
mod, action, func, ctrl, control, cmd, op, opt, command, system, task, meta, flag, adjust, toggle, override, enhance, change, var, switch, transform, shift, update, altered, modifier.
Also, I can add: extra, magic, alter

src/strings.cpp
ldpl: `error: no matching function for call to ‘StringParameters::StringParameters(std::nullptr_t, int, std::nullptr_t)`
I guess something changed since so I just use `MakeParameters()`, that seems to work.

ldpl:
Ok, I think I addressed every actionable comment here. But I'm not sure I fancy spending multiple hours on aimlessly rebasing this ever again. So can you please finally come up to some decision? At the very least agree on a new name for ctrl, I could then just get rid of the controversial fn/remove split and just allow reassigning shift and ctrl as they are. While that's totally not the initial intention of this PR it would still help to reduce the code footprint of an actual split massively.

TrueBrain:
Is it just me, or does this PR also change behaviour? I might misread it, but it seems like this does something with alt and removing of things. That is kinda unexpected to me to find in this PR, and I would expect that to be a separate PR? But I might be misreading this, as I am just looking at code :)

ldpl:
It does change behaviour, it was a remove-fn splitting PR first and other things just got dragged in as a consequence. But I can change it into doing configuration first and split in a separate PR. As soon as everyone agrees on how the magic ctrl should be named ;)

PR #10018, opened 2022-09-10
Codechange: Don't store tree counter in the map array

ldpl:
## Motivation / Problem

Trees add a lot of entropy to the map array drastically increasing the size of the save file.
According to my estimations, this change decreases the size by 10-15%.

## Description

The tree counter is just used to determine each 8th or 16th call to the `TileLoop_Trees` and that can easily be done without storing it in the map array. I plan to do some follow-up tree patches but this one is so free I decided to PR it separately.

## Limitations

It does change the order in which tiles are processed in the tile loop so it's not a perfect replica but it's ok as the rate is still the same. Regression tests had to be updated though as the random state blew up.

## Checklist for review

### DIFF
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -2334,8 +2334,7 @@ bool AfterLoadGame()
 			if (IsTileType(t, MP_TREES)) {
 				uint density = GB(_m[t].m2, 6, 2);
 				uint ground = GB(_m[t].m2, 4, 2);
-				uint counter = GB(_m[t].m2, 0, 4);
-				_m[t].m2 = ground << 6 | density << 4 | counter;
+				_m[t].m2 = ground << 6 | density << 4;
 			}
 		}
 	}
@@ -3187,6 +3186,15 @@ bool AfterLoadGame()
 					u->UpdatePosition();
 				}
 				RoadVehLeaveDepot(rv, false);
+			}
+		}
+
+		if (IsSavegameVersionBeforeOrAt(SLV_MULTITRACK_LEVEL_CROSSINGS)) {
+			/* Reset unused tree counters to reduce the savegame size. */
+			for (TileIndex t = 0; t < map_size; t++) {
+				if (IsTileType(t, MP_TREES)) {
+					SB(_m[t].m2, 0, 4, 0);
+				}
 			}
 		}
 
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -174,9 +174,6 @@ static void PlaceTree(TileIndex tile, uint32 r)
 		if (ground != TREE_GROUND_SNOW_DESERT && ground != TREE_GROUND_ROUGH_SNOW && ground != TREE_GROUND_SHORE) {
 			SetTreeGroundDensity(tile, (TreeGround)GB(r, 28, 1), 3);
 		}
-
-		/* Set the counter to a random start value */
-		SetTreeCounter(tile, (TreeGround)GB(r, 24, 4));
 	}
 }
 
@@ -710,10 +707,14 @@ static void TileLoop_Trees(TileIndex tile)
 
 	AmbientSoundEffect(tile);
 
-	uint treeCounter = GetTreeCounter(tile);
+	/* _tick_counter is incremented by 256 between each call, so ignore lower 8 bits.
+	 * Also, we use a simple hash to spread the updates evenly over the map.
+	 * 11 and 9 are just some co-prime numbers for better spread.
+	 */
+	uint32 cycle = 11 * TileX(tile) + 9 * TileY(tile) + (_tick_counter >> 8);
 
 	/* Handle growth of grass (under trees/on MP_TREES tiles) at every 8th processings, like it's done for grass on MP_CLEAR tiles. */
-	if ((treeCounter & 7) == 7 && GetTreeGround(tile) == TREE_GROUND_GRASS) {
+	if ((cycle & 7) == 7 && GetTreeGround(tile) == TREE_GROUND_GRASS) {
 		uint density = GetTreeDensity(tile);
 		if (density < 3) {
 			SetTreeGroundDensity(tile, TREE_GROUND_GRASS, density + 1);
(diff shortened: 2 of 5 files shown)


src/tree_cmd.cpp
```diff
@@ -707,10 +704,13 @@ static void TileLoop_Trees(TileIndex tile)
 
 	AmbientSoundEffect(tile);
 
-	uint treeCounter = GetTreeCounter(tile);
+	/* _tick_couter is incremented by 256 between each call, so ignore lower 8 bits.
```
```suggestion
	/* _tick_counter is incremented by 256 between each call, so ignore lower 8 bits.
```

2TallTyler (commented):
If I understand this correctly: Instead of keeping track of when each tree should grow based on a random initial counter, we randomly pick trees to grow and assume that the result is approximately the same? Trees also have a growth/density state, which is not affected. 

Is that correct?

2TallTyler:
From a discussion on Discord:

@2TallTyler:
> JGR It looks like you cherry-picked #10018 for your patchpack — does it do what the tin says? Any oddities to report? 🙂

@JGRennison:
> It does what it says on the tin, there's nothing wrong with it that i can see
> I haven't made any actual measurements of savegame sizes, however
> There isn't any observable change in tree behaviour

...

@ldpl:
> tbh randomization in 10018 is kinda temporary and can be improved but I doubt anyone would notice that anyway

@PeterN:
> I would :p

@ldpl:
> only if you're looking for it on fast forward :p 
> just don't say it changed in the changelog xDD
> or, whoever has fast pc can can try to find some decent looking primes for lcg :p 

@PeterN: 
> I find it odd to consider changing behaviour purely to decrease savegame size. 🤷

@ldpl: 
> I find it odd to store random values in the savegame :p 

...

Based on my understanding of how trees currently work and how this changes that, I think this PR is a worthwhile change. If we can get the same results without bloating the savegame with unnecessary noise, we should. I imagine it would most benefit people joining network games with large maps, which is a known issue. Most network games don't have trees for this very reason. 

That said, if the current randomization is "kinda temporary and can be improved" let's wait for that before we merge. 😄

ldpl:
It was temporary in a sense that I planned more entropy reduction patches for trees that would need a better random/hash anyway. Unfortunately, circumstances changed so I probably won't return to that work for a while. 

In any case, this is a fine (and simple) change on its own. Also, I improved hash a little bit so now I can't see any pattern whatsoever even on fast forward. Btw, keep in mind that tile loop adds its own randomization on top of this.

2TallTyler:
Sure, fine by me. Except it doesn't build now. 😉

PeterN:
It isn't (and cannot be) the same result, but it might be close enough.

As the usage of the map array has changed, it might be desirable to bump the savegame version and clear the existing data on existing maps, and ensure those bits are cleared when creating tree tiles, otherwise the benefits only appear for new maps.

src/tree_cmd.cpp
```diff
@@ -720,11 +721,7 @@ static void TileLoop_Trees(TileIndex tile)
 
 	if (_settings_game.construction.extra_tree_placement == ETP_NO_GROWTH_NO_SPREAD) return;
```
rubidium42, line 725: There's an interesting behavioural question to answer here. Should the grass grow or not?
With this setting on 'no growth and no spread', it previously disabled it for about 87.5% of the tree tiles, and it introduced an 8x speed up for the remaining 12.5% of the tiles. Since either the `treeCounter & 7` is actually 7 and it triggers every time because the tree counter does not get updated, or it never triggers as (again) the tree counter does not get updated.

As it stands now the grass is now allowed to grow for 100% of the tiles. If grass should not grow like on 87.5% of the tiles previously, then this check needs to be moved up to just after `AmbientSoundEffect(tile);`.

src/tree_cmd.cpp
```diff
@@ -720,11 +721,7 @@ static void TileLoop_Trees(TileIndex tile)
...
 
-	if (GetTreeCounter(tile) < 15) {
-		AddTreeCounter(tile, 1);
-		return;
-	}
-	SetTreeCounter(tile, 0);
+	if ((cycle & 15) < 15) return;
```
rubidium42: I would use `!= 15` instead of `< 15`. Both because that would be consistent with the previous if-statement, and because the `<` kind of implies there might be more than one case when there is onlyy one case.

src/tree_map.h
```diff
@@ -215,50 +215,6 @@ static inline void SetTreeGrowth(TileIndex t, uint g)
 	SB(_m[t].m5, 0, 3, g);
 }
 
-/**
```
rubidium42, line 218: If you remove it, also set it to 0 in the map structure in `AfterLoadGame` just to prevent nasty surprises in the future.

You can guard it with `IsSavegameVersionBeforeOrAt(SLV_<whatever_it_currently_is>)`, which would mean that it will set the field to 0 for every savegame until the next savegame bump is made. This way you will immediately see the result in (network) savegame size, at a tiny cost during loading.

While you're at it, in afterload.cpp around line 2352 (within a `IsSavegameVersionBefore(SLV_135)`) the counter gets read and written. That read can go and the counter can be removed from the bit-or. I wouldn't complicate the clearing logic of the counter by checking whether it is after this savegame version though.

src/tree_cmd.cpp
ldpl: I'd say it was introduced in #8415 accidentally so this fixes it. Even though non-growing grass is an interesting feature to have doing that only under the trees is a bug. Also, idk what 87.5% you're talking about, I just tested and no grass grows under trees in 13.0.

rubidium42 (approved):

2TallTyler (approved):

src/saveload/afterload.cpp
```diff
@@ -3190,6 +3189,15 @@ bool AfterLoadGame()
 			}
 		}
 
+		if (IsSavegameVersionBeforeOrAt(SLV_MULTITRACK_LEVEL_CROSSINGS)) {
```
michicc, line 3192: Sure you want this savegame version? Last is (since not long) now `SLV_LINKGRAPH_EDGES` (with a wrong comment).

rubidium42: Technically it should indeed have been `SLV_LINKGRAPH_EDGES`, but given the speed the two savegame versions after `SLV_MULTITRACK_LEVEL_CROSSINGS` got added to master, I doubt there are many (if any) savegames that are made between NewGRF road stops got added and this was merged. So, in the grand scheme of things it does not really matter.
I fixed the comment with #10471 by just squashing that tiny change into the same savegame version. Similarly, technically not the right thing to do, but essentially harmless.

PR #10034, opened 2022-09-18
Feature: Button to honk a vehicle's horn

2TallTyler:
## Motivation / Problem

Some transport games (Transport Fever 2, Mashinky) allow the player to honk the horn of a vehicle when riding along. Recently there was some (probably joking) discussion about adding this to OpenTTD, which apparently I took seriously. 😄 

## Description

honk

Adds a horn button to train and ship windows.

## Limitations

- Airplanes don't have horns, so they don't have the button
- In the original TTD sound pack, road vehicles never honk their horns. In OpenSFX they sometimes do. I chose to not include the button for road vehicles.
- Sprite suggestions or contributions welcome. 🙂 

## Checklist for review


glx22:
Don't forget to update https://github.com/OpenTTD/OpenTTD/blob/d0bdc138575b9be68233ab610f1a50477c02b131/src/table/sprites.h#L57

2TallTyler:
Accidentally drew the sprite too big. _Now_ it's ready for review. 😄

frosch123:
`_settings_client.sound.vehicle` disables vehicle sounds. A button which does nothing is weird, so either that button should somehow ignore the setting, or the setting should disable the button.

src/vehicle_gui.cpp
```diff
@@ -2977,6 +2989,17 @@ struct VehicleViewWindow : Window {
 					ShowOrdersWindow(v);
 				}
 				break;
+			case WID_VV_HONK_HORN: // honk horn
+				switch (v->type) {
+					case VEH_TRAIN:
+						v->PlayLeaveStationSound();
+						break;
+					case VEH_SHIP:
+						PlayShipSound(v);
```
frosch123: Isn't this the same as `v->PlayLeaveStationSound`?
I.e. no need to make `PlayShipSound` external, and no need to switch on vehicle type.

2TallTyler:
It now plays regardless of `_settings_client.sound.vehicle`. That said, if the player turns the Effects volume slider all the way down as a way to disable sound effects (instead of disabling them in Settings), they won't hear the horn.

Kuhnovic:
Haha I love the idea. It would be even better if you could send the command over the network so that you can honk to other players during multiplayer. Maybe a bit too much work though for a gimmicky feature like this, but it would be fun!

2TallTyler:
Rebased atop #10052, which changed `media/baseset/openttd.grf`.

Sending horn sounds over the network sounds like an easy way to annoy other players, so I don't think I'll add that. 😃

2TallTyler:
This is too silly a feature to justify adding a button to an already-busy UI. It would get used a few times and then just be in the way. It might be a fun hotkey-only hidden feature if anyone wants to run with it, but that person isn't me right now. 🙂

Arastais:
That's unfortunate. Was actually looking forward to this feature 😢

2TallTyler:
Feel free to steal my code and do what you want with it. I think a hotkey would be nice but I wouldn't use it enough to justify the time spent making a PR.

Rau771:
Huh, 3 years later this was mentioned, so… how about re-adding this button, but to the top panel, next to "centre main view"? It won't make window any larger so wouldn't hurt, unlike adding it into right side along with "actually useful" buttons.

изображение

_The debug button might need to be moved to the left_

2TallTyler:
It was added as a hotkey in https://github.com/OpenTTD/OpenTTD/pull/10110 .

Rau771:
> It was added as a hotkey in #10110 .

Yes, yes… but it's a non-standard hotkey (at least for OpenTTD) so to find out about it, you have to "press all the buttons and see what happens" (or know about hotkeys.cfg). I don't think a lot of players do this, and even if they do, it's likely to be an unnecessary feature for their purposes. But a tiny button somewhere in an inconspicuous corner could be way better to find and be used for a new players.

PR #10035, opened 2022-09-18
Change: Make _tick_counter 64bit to avoid wrapping

ldpl:
## Motivation / Problem

I mostly need this for the procedural tree PR I'm currently working on but this isn't the first time I run into it being inconvenient and error-inducing that _tick_counter wraps every 2 in-game years. So I think this PR is a good change on its own as even right now `OnTick_Station` is doing slightly inconsistent ticks because of the wrapping, which isn't a big issue but still. Also, #9693 can probably benefit from this change as well.

## Description

Use uint64 for _tick_counter. I decided to skip uint32 as ~159000 in-game years are clearly not enough~ as computers are mostly 64bit now anyway and it's large enough to not overflow even if the tick rate is changed in the future or in some patch pack.

## Limitations

- NewGRFs still get a 16bit tick counter because spec.
- MIngw printf format specifiers are stupid.

## Checklist for review


src/newgrf_profiling.cpp
```diff
@@ -109,7 +109,7 @@ uint32 NewGRFProfiler::Finish()
 
 	fputs("Tick,Sprite,Feature,Item,CallbackID,Microseconds,Depth,Result\n", f);
 	for (const Call &c : this->calls) {
-		fprintf(f, "%u,%u,0x%X,%u,0x%X,%u,%u,%u\n", c.tick, c.root_sprite, c.feat, c.item, (uint)c.cb, c.time, c.subs, c.result);
+		fprintf(f, "%I64u,%u,0x%X,%u,0x%X,%u,%u,%u\n", c.tick, c.root_sprite, c.feat, c.item, (uint)c.cb, c.time, c.subs, c.result);
```
glx22: Shouldn't that use `OTTD_PRINTF64U`?

ldpl: Yeah, I just tried to put it directly to figure out wth is wrong with mingw.

LordAro (approved):
Seems reasonable

PR #10042, opened 2022-09-22
Fix 9aef7b8c: Don't dispatch hover event if already hovering.

PeterN:
## Motivation / Problem

When using timed tooltips, the tooltip window is destroyed and recreated every frame. This is invisible but unnecessary.

## Description

`MC_HOVER` is dispatched every mouse loop, with triggers GuiShowTooltips() which heartlessly destroys its window and creates another one.

This is fixed by not dispatching the `MC_HOVER` event if already hovering.

## Limitations

This also affects the `OnHover()` window event, however this is not ever used so there's no impact.

## Checklist for review

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -2966,7 +2966,7 @@ void HandleMouseEvents()
 			hover_pos = _cursor.pos;
 			hover_time = std::chrono::steady_clock::now();
 			_mouse_hovering = false;
-		} else {
+		} else if (!_mouse_hovering) {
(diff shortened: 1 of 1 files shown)


James103:
Commit checker is failing because the first part of the title is incorrect. The "r20118" needs to be replaced with the hash of the corresponding commit, which is 9aef7b8c3dab0fe2d6e8855b7effcb89908c699e.

LordAro (approved):

PR #10045, opened 2022-09-23
Cleanup: Split fontcache implementations.

PeterN:
## Motivation / Problem

fontcache.cpp contains 4 font cache related classes, 

1. Base `FontCache`.
2. The sprite-based `SpriteFontCache`.
3. `TrueTypeFontCache` base font cache for platform specific implementations.
4. `FreeTypeFontCache` font cache.

This is left over from the original system which was then expanded. There is therefore a bit of a mishmash of similar looking classes all smashed together.

The platform specific `Win32FontCache` and `CoreTextFontCache` implementations are already in separate files.

Additionally some parts of the fontcache system reference freetype even though they have nothing to do with it.

And the GetUnicodeGlyph method should be internal to the `SpriteFontCache` but it is exposed throughout the font cache system.

## Description

The font cache implementations have been split into separate files into a fontcache subdirectory.

`FreeTypeSettings` has been renamed `FontCacheSettings`.

GetUnicodeGlyph has been removed from FontCache and made private in SpriteFontCache.

## Checklist for review


src/debug.cpp
```diff
@@ -72,7 +72,7 @@ struct DebugLevel {
 	DEBUG_LEVEL(oldloader),
 	DEBUG_LEVEL(npf),
 	DEBUG_LEVEL(yapf),
-	DEBUG_LEVEL(freetype),
+	DEBUG_LEVEL(fontcache),
```
LordAro, line 75: I suppose technically this makes it a 'change' commit

src/fontcache/freetypefontcache.cpp
```diff
@@ -0,0 +1,334 @@
...
+#endif /* WITH_FREETYPE */
+
+
+#if !defined(_WIN32) && !defined(__APPLE__) && !defined(WITH_FONTCONFIG) && !defined(WITH_COCOA)
+
+#ifdef WITH_FREETYPE
+FT_Error GetFontByFaceName(const char *font_name, FT_Face *face) { return FT_Err_Cannot_Open_Resource; }
+#endif /* WITH_FREETYPE */
+
+#endif /* !defined(_WIN32) && !defined(__APPLE__) && !defined(WITH_FONTCONFIG) && !defined(WITH_COCOA) */
```
LordAro: Why separate this from the main WITH_FREETYPE block?

src/fontcache_internal.h
```diff
@@ -19,59 +19,4 @@ static const int MAX_FONT_SIZE = 72; ///< Maximum font size.
 static const byte FACE_COLOUR = 1;
 static const byte SHADOW_COLOUR = 2;
```
LordAro: fontcache_internal is now left with just 3 constants. Can they go somewhere else?

LordAro (commented):
<3

src/fontcache/freetypefontcache.cpp
PeterN: I remembered cleaning this bit up, but that must have been in a slightly different iteration!

src/fontcache_internal.h
PeterN: I think so, They are all only used by TrueTypeFontCache implementations, but the two colour definitions could be considered generic enough to just live in fontcache.h.

I think for now I should simply move them along with the other TrueType bits.

LordAro (approved):

Issue #9697, opened 2021-11-14
[Bug]: Long server name sometimes forces Online Players window to be very wide


LC-Zorg:
### Version of OpenTTD

12.0 / 12.1 Win10

### Expected result

You can significantly reduce the width of the window. Saving window size works.

### Actual result

The width of the window varies almost each time the game is started and gives the impression of being random. Very often this width is senselessly large and cannot be reduced. Sometimes changing the language unlocks the possibility of reducing the width. Saving the window size doesn't work.

The minimum width of the window after entering the server
![Senselessly big window](https://user-images.githubusercontent.com/67662046/141700150-26570fd1-71df-4b66-8e18-4b9b25a8cb3e.png)

The minimum width of the window after closing the game and re-entering the server
![Narrower window after changing the language](https://user-images.githubusercontent.com/67662046/141700157-3b01e6df-7017-4245-a719-0ef5081bd689.png)

This method doesn't always work.

### Steps to reproduce

The situation is repetitive, but I don't see the key. The problem seems to arise after trying to save the window size.

James103:
This may be caused by the length of the server name being a limiting factor in the width of the online players window. Some servers put extra info in the server name, which inflates the length of said server name.

To confirm this, will you please try the following?
1. Start and/or join a server with a very long name.
2. Attempt to reproduce the bug to show that the long name increases the minimum width of the window.
3. Start and/or join a server with a very short name.
4. Attempt to reproduce the bug to show that the short name is not as much of a factor in determining the minimum width of the window.

LC-Zorg:
As you can see above, these two screenshots are from the same server, which actually has a long name. And as you can see, this doesn't the reason to prevent the window from reducing the width.

PR #10054, opened 2022-09-26
Added widget showing company slots remaining

### LINKED ISSUE #9703: [Bug]: No indication that server is full within the game
### Version of OpenTTD

12.1

### Expected result

Online players window shows some clear indication that server is full and new company can't be created.

### Actual result

(New company) line just disappears and players, especially newbies, get confused trying to find it.
Screenshot from 2021-11-17 03-21-21

Screenshot from 2021-11-17 03-14-53

### Steps to reproduce

1. Join a full server
2. ???

zachtyson:
## Motivation / Problem
I've seen a few new players join and try to create a company, not realizing that the game has reached the maximum amount of companies, e.g #9703 , which is easier to see at a glance. 

## Description
Adds a single line above company/client count to display the amount of remaining companies that can be created.

## Limitations
The addition simply shows the amount of existing companies minus the max amount of companies. Can be changed from '0 company slots' to 'FULL' if desired

Closes #9703 - edit 2TT

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -1341,6 +1341,7 @@ static const NWidgetPart _nested_client_list_widgets[] = {
 		NWidget(NWID_VERTICAL),
 			NWidget(WWT_MATRIX, COLOUR_GREY, WID_CL_MATRIX), SetMinimalSize(180, 0), SetResize(1, 1), SetFill(1, 1), SetMatrixDataTip(1, 0, STR_NULL), SetScrollbar(WID_CL_SCROLLBAR),
 			NWidget(WWT_PANEL, COLOUR_GREY),
+				NWidget(WWT_TEXT, COLOUR_GREY, WID_CL_CLIENT_COMPANY_REMAINING), SetFill(1, 0), SetMinimalTextLines(1, 0), SetResize(1, 0), SetPadding(2, 1, 2, 1), SetAlignment(SA_CENTER), SetDataTip(STR_NETWORK_CLIENT_LIST_CLIENT_COMPANY_REMAINING, STR_NULL),
 				NWidget(WWT_TEXT, COLOUR_GREY, WID_CL_CLIENT_COMPANY_COUNT), SetFill(1, 0), SetMinimalTextLines(1, 0), SetResize(1, 0), SetPadding(2, 1, 2, 1), SetAlignment(SA_CENTER), SetDataTip(STR_NETWORK_CLIENT_LIST_CLIENT_COMPANY_COUNT, STR_NULL),
 			EndContainer(),
 		EndContainer(),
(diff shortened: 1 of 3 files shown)


James103:
In the lines you added, the code must be indented with tabs instead of spaces.
Commit checker will fail until you resolve this.

LordAro:
Seems reasonable. Just a few things:
* Can you add a screenshot of what it looks like?
* "Company slots" feels like odd English to me... not that I can think of anything better at the moment. I guess "Companies remaining" is too long? "Open slots" "Unused companies" ? IDK
* Don't modify any language files other than english.txt - everything else is exclusively handled via our translator teams (even if it's identical!)
* We like rebasing rather than merge commits :) (As admins we can forcibly squash PRs into a single commit and fix it on merge, but it's much easier to deal with)

zachtyson:
Untitled

Gotcha, I changed it from to open slots also

Bouke:
The server list is using x/y notation (x used, y available):

image
image

Maybe we should use the same here for consistency, instead of introducing a new concept "open slot(s)".

James103:
If it is desired to keep it as one line only at the bottom, it could say something like "3 / 100 clients - 5 / 15 companies"

LordAro:
Mm, does seem reasonable to keep it consistent. Total possible number of clients seems redundant when you're already connected to the server, but again, consistent.

zachtyson:
Yea I get the need for consistency, I can show a preview of those proposed changes if y'all want once I get off work

LC-Zorg:
In my opinion and for me the problem is a bit that the bar and the create a new company button disappears. For example, it will disappear if the list is long and you scroll down. The same is true when it is not possible to start a company because all available slots have been taken. Honestly, I still haven't got used to it until today.

I like the idea of ​​Bouke and James103. It takes up less space, is equally pithy and consistent with what is in the server list window.

I would also have some idea what else can be changed in this or other PR.

1. If the number of companies reaches the maximum number, the number is red - this is more noticeable and I think it is understandable.
2. The button for establishing a new company can be moved to the bottom of the window (I have already proposed something like that before - [here](https://github.com/OpenTTD/OpenTTD/discussions/9504#discussioncomment-1398870)).
- When there is free space, the button is active
- When there is no free space, the button is still visible but inactive and shaded
- When the player is in his own company, the bar with the create company button is collapsed
Numbers of clients and companies v1 0


By the way, I would like to remind about the problem of the inability to reduce the width of this window, which is almost always much wider than it should (#9697). I mention it, if only so that this PR doesn't somehow reinforce this bug.

2TallTyler:
@zachtyson Are you still interested in working on this? Would be nice to have in the upcoming OpenTTD 13 release. 🙂

src/network/network_gui.cpp
```diff
@@ -1785,6 +1786,10 @@ struct NetworkClientListWindow : Window {
 				break;
 			}
 
+			case WID_CL_CLIENT_COMPANY_REMAINING:
+				SetDParam(0, _settings_client.network.max_companies - Company::GetNumItems());
```
rubidium42, line 1790: This setting does not get updated from the server. So, it will not be right when joining the server and whenever the server changes it won't be right. Unless "accidentally" the configuration value at the server happens to be the same as for your client, which is quite likely when testing it locally.

I would suggest you look at `NetworkMaxCompaniesReached` and add a similar function at those locations that returns the number of companies that are remaining. I would then also suggest changing `NetworkMaxCompaniesReached` so it uses that new function and compares that count with 0. That way the logic that chooses between the variables containing the maximum number of companies does not get duplicated and if the logic requires changes it is less likely that the other function is missed.

Issue #7247, opened 2019-02-19
With lots of vehicles, PerformanceAccumulator has a large performance impact itself


PeterN:
## Version of OpenTTD

master-gef7e47a53a

## Expected result

Performance meter monitors performance of vehicle ticks with minimal adverse affect.

## Actual result

Due to large amount of individual timing, Performance meter consumes significant CPU itself when timing for lots of vehicles.

## Steps to reproduce

* Load the 'wentbourne' save.
* Open the Frame rate and observe the Simulation rate.
* Edit source to disable the PerformanceAccumulator for vehicles. Comment out the PerformanceAccumulator lines in src/roadveh_cmd.cpp, src/aircraft_cmd.cpp, src/ship_cmd.cpp, and src/train_cmd.cpp
* Open the Frame rate window and observe the Simulation rate.

On my particular system, the simulation rate in master is around 8 fps.
After disabling the PerformanceAccumulator, and no other changes, the simulation rate increases by 50% to around 12 fps.

PeterN:
Link to savegame: https://fuzzle.org/~petern/ottd/wentbourne.sav

nielsmh:
Unfortunately fixing this will probably require rather large restructuring of the vehicle ticks code, to separate vehicle types into each their own arrays, so all road vehicles, trains, etc. can be processed as single groups.

It might be possible to somehow compile two versions of the vehicle tick functions, one with and one without measurements, and change the dynamic dispatch depending on a setting, but that's likely also tricky to get right.

James103:
Calculating benchmarks from the initial comment in this issue... 

(Note: tps = ticks per second (sim rate), different from fps = frames per second (graphics rate). This is despite OpenTTD listing sim rate and graphics rate in the same units)
* Simulation rate (control): 8 tps (125 mspt from above, 700-800 mspt on a Core 2 Duo T5800 2 GHz)
* Simulation rate (PerformanceAccumulator disabled): 12 tps (83.33 mspt)
* Simulation rate improvement: 41.67 mspt 
* Number of vehicles in saved game: 13,899 (NOTE: Road vehicles > 5000 (= 5499) in single company, possibly hacked max vehicles?)

Therefore, the Vehicle PerformanceAccumulator takes ~3 microseconds (on a mid-range CPU) per vehicle per tick. Not a lot, but it adds up over thousands of vehicles to ~42 mspt at 14K vehicles.

PeterN:
[...]

PR #10055, opened 2022-09-27
Fix: reduce framerate overhead in Train::Tick

Bouke:
## Motivation / Problem

After profiling a save game with lots of trains (few hundred), I noticed 20% of `Train::Tick()` was spent on measuring the game loop. By moving the measurement to the only relevant case, I reduced the game loop duration by 25% (15ms to 12ms).

## Description

`Train::Tick()` is a noop for all but front-engine / crashed vehicles. Starting a `framerate` is rather cheap, but not free, and introduces a lot of overhead for such close loops. This PR changes `Train::Tick` to only start a `framerate` for the front engine. As `Train::Tick()` is a noop in other cases, in practice there should not be a difference to the actual generated measurements.

Before:
Screenshot 2022-09-27 at 10 32 13

After:
Screenshot 2022-09-27 at 10 35 57

## Alternatives considered

Currently the frame rate is determined _per-vehicle_, but we could instead do this _per-type_. So for 100 trains we would do a single measurement combining 100 `Tick()`s instead of 100 measurements of a single `Tick()`. This would mean we have to enumerate the vehicles more, once per type (4 types + 1 'other'). A small experiment didn't yield improved noticeably.

## Checklist for review

<s>* The bug fix is important enough to be backported? (label: 'backport requested')

### DIFF
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -3910,11 +3910,11 @@ Money Train::GetRunningCost() const
  */
 bool Train::Tick()
 {
-	PerformanceAccumulator framerate(PFE_GL_TRAINS);
-
 	this->tick_counter++;
 
 	if (this->IsFrontEngine()) {
+		PerformanceAccumulator framerate(PFE_GL_TRAINS);
+
 		if (!(this->vehstatus & VS_STOPPED) || this->cur_speed > 0) this->running_ticks++;
 
 		this->current_order_time++;


PeterN:
Re Alternatives considered, I tried that a few years ago and it improved performance for me but for others it was detrimental, so I didn't progress much further.

https://github.com/OpenTTD/OpenTTD/compare/master...PeterN:OpenTTD:vehicle-tick-split

JGRennison:
It's probably worth linking this to https://github.com/OpenTTD/OpenTTD/issues/7247

On the alternative of doing ticks in vehicle type groups, with a single performance measurement around each type group, I am doing this in my branch. To avoid the extra iteration overhead I'm maintaining caches of per-vehicle type lists (and the cached lists have exclusions like non including non-head trains, etc).

PeterN:
That's the PR I couldn't find :-) Thanks!

Bouke:
> It's probably worth linking this to #7247
> 
> On the alternative of doing ticks in vehicle type groups, with a single performance measurement around each type group, I am doing this in my branch. To avoid the extra iteration overhead I'm maintaining caches of per-vehicle type lists (and the cached lists have exclusions like non including non-head trains, etc).

I can imagine this gives small improvements throughout the game where currently the whole collection is iterated. Is there any objection to upstreaming these changes?

JGRennison:
> I can imagine this gives small improvements throughout the game where currently the whole collection is iterated. Is there any objection to upstreaming these changes?

As written it's only usable for the tick cache, and general iteration in other places is as before.
Most of it is in these commits if you want to take a look: https://github.com/JGRennison/OpenTTD-patches/commit/e5b61e0b6cf, https://github.com/JGRennison/OpenTTD-patches/commit/e28e2f638e8, https://github.com/JGRennison/OpenTTD-patches/commit/99b79c26c6c, https://github.com/JGRennison/OpenTTD-patches/commit/455e694e79f

Bouke:
> > I can imagine this gives small improvements throughout the game where currently the whole collection is iterated. Is there any objection to upstreaming these changes?
> 
> As written it's only usable for the tick cache, and general iteration in other places is as before. Most of it is in these commits if you want to take a look: [JGRennison/OpenTTD-patches@e5b61e0b6cf](https://github.com/JGRennison/OpenTTD-patches/commit/e5b61e0b6cf), [JGRennison/OpenTTD-patches@e28e2f638e8](https://github.com/JGRennison/OpenTTD-patches/commit/e28e2f638e8), [JGRennison/OpenTTD-patches@99b79c26c6c](https://github.com/JGRennison/OpenTTD-patches/commit/99b79c26c6c), [JGRennison/OpenTTD-patches@455e694e79f](https://github.com/JGRennison/OpenTTD-patches/commit/455e694e79f)

Without further diving into the specifics of your changes, would it make sense to upstream them in some way? Is it worthwhile of our time to spend time on preparing a patch?

PeterN:
As it stands, for me this improves train ticks of the Wentbourne save from 50ms to 40ms. Admittedly this change does skip a bit of the train tick handle, but I don't think that's just from not counting the IsFrontEngine() test...

Bouke:
> As it stands, for me this improves train ticks of the Wentbourne save from 50ms to 40ms. Admittedly this change does skip a bit of the train tick handle, but I don't think that's just from not counting the IsFrontEngine() test...

Do you also see a 10ms decrease of the total game loop? Any differences due to only measuring `IsFrontEngine()` would show up there.

michicc (approved):

michicc:
This does not block any kind of further improvements like in JGRPP, so why not.

PR #10058, opened 2022-09-28
Feature: [UI] Split AI/Game Script configuration windows and add them to world gen window

Arastais:
## Motivation / Problem
A while ago, @frosch123 mentioned in the discord that adding Game Script, AI, and NewGRF settings buttons to the world gen window is an old request/idea but always gets swamped by other requests. As I personally also want this feature, I decided to do it myself. This will allow users to edit Game Script/AI settings, NewGRF settings, and Game Script configuration from both world generation windows (new game and heightmap). 

What's most convenient is that users can now edit Game Script paramters directly from the world generation window, without having to go back to main menu > AI/Game Script Setttings > select Game Script > Configure, and can instead just click "Configure Game Script". This also aligns with OpenTTD's goal of improving the user interface; I made the changes in the way that they are not too intrusive (the new buttons do not over-complicate the world generation window) and have the same styling as the other buttons. This will be most enjoyed by any players who regularly use and change either their game script or their game script configuration/parameters.

## Description
This commit simply adds 3 buttons to the bottom of the world generation window, each of which just show windows which are already in the game:
eb61390770a79342ee5799bee2c57d4a

The first two buttons do exactly the same as their main menu counterpart; the "Configure Game Script" button directly shows the Game Script Parameters window for the currently selected game script and allows players to directly configure the Game Script. As mentioned earlier, this allows players to easily and fully configure their NewGRFs and scripts right from the world generation screen.

The buttons are double the height of the other buttons so they are easy to see but are also at the bottom of the window so they do not intrude any part of the world gen interface (The idea to put them at the bottom was thanks to @ldpl). They are also colored the same as the other buttons (excluding the "Generate" button) to match OpenTTD's theme. 

## Limitations
This feature solves the problem of inconvenient menu navigation for all players. The feature is pretty much complete, and nothing else needs to be added to it. It is possible to change the look and feel of the buttons in the future, or add more buttons with similar functionality, but my aim was to make functionality that was already in the game more convenient and user-friendly. As this is a simple addition, there should be no bugs or problems that occur because of it.

The only notable source code change is that `void ShowAISettingsWindow(CompanyID slot)` in `ai_gui.cpp` was added via decleration to its corresponding header file (i.e. the header file for the AI gui) and is no longer static, so that it could be visible to the world gen GUI file (`genworld_gui.cpp`). This should not have a real effect on any part of the game however.

It is important to note however that **I did add strings to `english.txt`**, some of which were duplicates. I itentionally did not re-use the strings from the main menu (e.g. reusing "NewGRF Settings") because I did not want them to share the same strings (causing edits to them to relfect in both places) and possibly confuse people if they were looking for a seperate set of strings. It would also make the strings inconsistent: the first two buttons would use shared strings, while the "Configure Game Script" button would have its own string in a different section in `english.txt`. Not to mention, I don't see string codes being re-used anywhere in OpenTTD. Regardless, this could be easily changed if need be, but I ended up doing what I believed is best.
## Checklist for review

### DIFF
--- a/src/ai/ai_gui.cpp
+++ b/src/ai/ai_gui.cpp
@@ -182,7 +182,7 @@ struct AIListWindow : public Window {
 			for (int i = 0; i < this->selected; i++) it++;
 			GetConfig(slot)->Change((*it).second->GetName(), (*it).second->GetVersion());
 		}
-		InvalidateWindowData(WC_GAME_OPTIONS, WN_GAME_OPTIONS_AI);
+		InvalidateWindowData(WC_GAME_OPTIONS, slot == OWNER_DEITY ? WN_GAME_OPTIONS_GS : WN_GAME_OPTIONS_AI);
 		InvalidateWindowClassesData(WC_AI_SETTINGS);
 		CloseWindowByClass(WC_QUERY_STRING);
 		InvalidateWindowClassesData(WC_TEXTFILE);
@@ -276,7 +276,7 @@ static WindowDesc _ai_list_desc(
  * Open the AI list window to chose an AI for the given company slot.
  * @param slot The slot to change the AI of.
  */
-static void ShowAIListWindow(CompanyID slot)
+void ShowAIListWindow(CompanyID slot)
 {
 	CloseWindowByClass(WC_AI_LIST);
 	new AIListWindow(&_ai_list_desc, slot);
@@ -296,7 +296,7 @@ struct AISettingsWindow : public Window {
 	int clicked_row;                      ///< The clicked row of settings.
 	int line_height;                      ///< Height of a row in the matrix widget.
 	Scrollbar *vscroll;                   ///< Cache of the vertical scrollbar.
-	typedef std::vector<const ScriptConfigItem *> VisibleSettingsList;
+	typedef std::vector<const ScriptConfigItem *> VisibleSettingsList; ///< typdef for a vector of script settings
 	VisibleSettingsList visible_settings; ///< List of visible AI settings
 
 	/**
@@ -318,15 +318,6 @@ struct AISettingsWindow : public Window {
 		this->FinishInitNested(slot);  // Initializes 'this->line_height' as side effect.
 
 		this->RebuildVisibleSettings();
-	}
-
-	void SetStringParameters(int widget) const override
-	{
-		switch (widget) {
-			case WID_AIS_CAPTION:
-				SetDParam(0, (this->slot == OWNER_DEITY) ? STR_AI_SETTINGS_CAPTION_GAMESCRIPT : STR_AI_SETTINGS_CAPTION_AI);
-				break;
-		}
 	}
 
 	/**
@@ -476,7 +467,7 @@ struct AISettingsWindow : public Window {
 						wi_rect.top = pt.y - rel_y + (this->line_height - SETTING_BUTTON_HEIGHT) / 2;
 						wi_rect.bottom = wi_rect.top + SETTING_BUTTON_HEIGHT - 1;
 
-						/* For dropdowns we also have to check the y position thoroughly, the mouse may not above the just opening dropdown */
+						/* If the mouse is still held but dragged outside of the dropdown list, keep the dropdown open */
 						if (pt.y >= wi_rect.top && pt.y <= wi_rect.bottom) {
 							this->clicked_dropdown = true;
 							this->closing_dropdown = false;
@@ -533,24 +524,15 @@ struct AISettingsWindow : public Window {
 	void OnQueryTextFinished(char *str) override
 	{
 		if (StrEmpty(str)) return;
-		VisibleSettingsList::const_iterator it = this->visible_settings.begin();
-		for (int i = 0; i < this->clicked_row; i++) it++;
-		const ScriptConfigItem config_item = **it;
-		if (_game_mode == GM_NORMAL && ((this->slot == OWNER_DEITY) || Company::IsValidID(this->slot)) && (config_item.flags & SCRIPTCONFIG_INGAME) == 0) return;
 		int32 value = atoi(str);
-		this->ai_config->SetSetting(config_item.name, value);
-		this->SetDirty();
+
+		SetValue(value);
 	}
 
 	void OnDropdownSelect(int widget, int index) override
 	{
 		assert(this->clicked_dropdown);
-		VisibleSettingsList::const_iterator it = this->visible_settings.begin();
-		for (int i = 0; i < this->clicked_row; i++) it++;
-		const ScriptConfigItem config_item = **it;
-		if (_game_mode == GM_NORMAL && ((this->slot == OWNER_DEITY) || Company::IsValidID(this->slot)) && (config_item.flags & SCRIPTCONFIG_INGAME) == 0) return;
-		this->ai_config->SetSetting(config_item.name, index);
-		this->SetDirty();
+		SetValue(index);
 	}
 
 	void OnDropdownClose(Point pt, int widget, int index, bool instant_close) override
@@ -597,13 +579,23 @@ private:
 			|| (config_item.flags & SCRIPTCONFIG_INGAME) != 0
 			|| _settings_client.gui.ai_developer_tools;
 	}
+
+	void SetValue(int value)
+	{
+		VisibleSettingsList::const_iterator it = this->visible_settings.begin();
+		for (int i = 0; i < this->clicked_row; i++) it++;
+		const ScriptConfigItem config_item = **it;
+		if (_game_mode == GM_NORMAL && ((this->slot == OWNER_DEITY) || Company::IsValidID(this->slot)) && (config_item.flags & SCRIPTCONFIG_INGAME) == 0) return;
+		this->ai_config->SetSetting(config_item.name, value);
+		this->SetDirty();
+	}
 };
 
 /** Widgets for the AI settings window. */
 static const NWidgetPart _nested_ai_settings_widgets[] = {
 	NWidget(NWID_HORIZONTAL),
 		NWidget(WWT_CLOSEBOX, COLOUR_MAUVE),
-		NWidget(WWT_CAPTION, COLOUR_MAUVE, WID_AIS_CAPTION), SetDataTip(STR_AI_SETTINGS_CAPTION, STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS),
+		NWidget(WWT_CAPTION, COLOUR_MAUVE, WID_AIS_CAPTION), SetDataTip(STR_AI_SETTINGS_CAPTION_AI, STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS),
 		NWidget(WWT_DEFSIZEBOX, COLOUR_MAUVE),
 	EndContainer(),
 	NWidget(NWID_HORIZONTAL),
@@ -683,7 +675,7 @@ void ShowScriptTextfileWindow(TextfileType file_type, CompanyID slot)
 static const NWidgetPart _nested_ai_config_widgets[] = {
 	NWidget(NWID_HORIZONTAL),
 		NWidget(WWT_CLOSEBOX, COLOUR_MAUVE),
-		NWidget(WWT_CAPTION, COLOUR_MAUVE), SetDataTip(STR_AI_CONFIG_CAPTION, STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS),
+		NWidget(WWT_CAPTION, COLOUR_MAUVE), SetDataTip(STR_AI_CONFIG_CAPTION_AI, STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS),
 	EndContainer(),
 	NWidget(WWT_PANEL, COLOUR_MAUVE, WID_AIC_BACKGROUND),
 		NWidget(NWID_VERTICAL), SetPIP(4, 4, 4),
@@ -705,16 +697,13 @@ static const NWidgetPart _nested_ai_config_widgets[] = {
 			EndContainer(),
 		EndContainer(),
 		NWidget(NWID_SPACER), SetMinimalSize(0, 9),
-		NWidget(WWT_FRAME, COLOUR_MAUVE), SetDataTip(STR_AI_CONFIG_GAMESCRIPT, STR_NULL), SetPadding(0, 5, 4, 5),
-			NWidget(WWT_MATRIX, COLOUR_MAUVE, WID_AIC_GAMELIST), SetMinimalSize(288, 14), SetFill(1, 0), SetMatrixDataTip(1, 1, STR_AI_CONFIG_GAMELIST_TOOLTIP),
-		EndContainer(),
 		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE), SetPIP(7, 0, 7),
-			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_CHANGE), SetFill(1, 0), SetMinimalSize(93, 0), SetDataTip(STR_AI_CONFIG_CHANGE, STR_AI_CONFIG_CHANGE_TOOLTIP),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_CHANGE), SetFill(1, 0), SetMinimalSize(93, 0), SetDataTip(STR_AI_CONFIG_CHANGE_AI, STR_AI_CONFIG_CHANGE_TOOLTIP),
 			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_CONFIGURE), SetFill(1, 0), SetMinimalSize(93, 0), SetDataTip(STR_AI_CONFIG_CONFIGURE, STR_AI_CONFIG_CONFIGURE_TOOLTIP),
-			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_CLOSE), SetFill(1, 0), SetMinimalSize(93, 0), SetDataTip(STR_AI_SETTINGS_CLOSE, STR_NULL),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_TEXTFILE + TFT_README), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_TEXTFILE_VIEW_README, STR_NULL),
 		EndContainer(),
 		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE), SetPIP(7, 0, 7),
-			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_TEXTFILE + TFT_README), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_TEXTFILE_VIEW_README, STR_NULL),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_CLOSE), SetFill(1, 0), SetMinimalSize(93, 0), SetDataTip(STR_AI_SETTINGS_CLOSE, STR_NULL),
 			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_TEXTFILE + TFT_CHANGELOG), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_TEXTFILE_VIEW_CHANGELOG, STR_NULL),
 			NWidget(WWT_PUSHTXTBTN, COLOUR_YELLOW, WID_AIC_TEXTFILE + TFT_LICENSE), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_TEXTFILE_VIEW_LICENCE, STR_NULL),
 		EndContainer(),
@@ -762,21 +751,6 @@ struct AIConfigWindow : public Window {
 			case WID_AIC_NUMBER:
 				SetDParam(0, GetGameSettings().difficulty.max_no_competitors);
 				break;
-			case WID_AIC_CHANGE:
-				switch (selected_slot) {
-					case OWNER_DEITY:
-						SetDParam(0, STR_AI_CONFIG_CHANGE_GAMESCRIPT);
-						break;
-
-					case INVALID_COMPANY:
-						SetDParam(0, STR_AI_CONFIG_CHANGE_NONE);
-						break;
-
-					default:
-						SetDParam(0, STR_AI_CONFIG_CHANGE_AI);
-						break;
-				}
-				break;
 		}
 	}
 
@@ -788,32 +762,11 @@ struct AIConfigWindow : public Window {
 				*size = maxdim(*size, NWidgetScrollbar::GetHorizontalDimension());
 				break;
 
-			case WID_AIC_GAMELIST:
-				this->line_height = FONT_HEIGHT_NORMAL + WD_MATRIX_TOP + WD_MATRIX_BOTTOM;
-				size->height = 1 * this->line_height;
-				break;
-
 			case WID_AIC_LIST:
 				this->line_height = FONT_HEIGHT_NORMAL + WD_MATRIX_TOP + WD_MATRIX_BOTTOM;
 				resize->height = this->line_height;
 				size->height = 8 * this->line_height;
 				break;
-
-			case WID_AIC_CHANGE: {
-				SetDParam(0, STR_AI_CONFIG_CHANGE_GAMESCRIPT);
-				Dimension dim = GetStringBoundingBox(STR_AI_CONFIG_CHANGE);
-
-				SetDParam(0, STR_AI_CONFIG_CHANGE_NONE);
-				dim = maxdim(dim, GetStringBoundingBox(STR_AI_CONFIG_CHANGE));
-
-				SetDParam(0, STR_AI_CONFIG_CHANGE_AI);
-				dim = maxdim(dim, GetStringBoundingBox(STR_AI_CONFIG_CHANGE));
-
-				dim.width += padding.width;
-				dim.height += padding.height;
-				*size = maxdim(*size, dim);
-				break;
-			}
 		}
 	}
 
@@ -824,8 +777,6 @@ struct AIConfigWindow : public Window {
 	 */
 	static bool IsEditable(CompanyID slot)
 	{
-		if (slot == OWNER_DEITY) return _game_mode != GM_NORMAL || Game::GetInstance() != nullptr;
-
 		if (_game_mode != GM_NORMAL) {
 			return slot > 0 && slot <= GetGameSettings().difficulty.max_no_competitors;
 		}
@@ -841,20 +792,6 @@ struct AIConfigWindow : public Window {
 	void DrawWidget(const Rect &r, int widget) const override
 	{
 		switch (widget) {
-			case WID_AIC_GAMELIST: {
-				StringID text = STR_AI_CONFIG_NONE;
-
-				if (GameConfig::GetConfig()->GetInfo() != nullptr) {
-					SetDParamStr(0, GameConfig::GetConfig()->GetInfo()->GetName());
-					text = STR_JUST_RAW_STRING;
-				}
-
-				DrawString(r.left + 10, r.right - 10, r.top + WD_MATRIX_TOP, text,
-						(this->selected_slot == OWNER_DEITY) ? TC_WHITE : (IsEditable(OWNER_DEITY) ? TC_ORANGE : TC_SILVER));
-
-				break;
-			}
-
 			case WID_AIC_LIST: {
 				int y = r.top;
 				for (int i = this->vscroll->GetPosition(); this->vscroll->IsVisible(i) && i < MAX_COMPANIES; i++) {
@@ -899,13 +836,6 @@ struct AIConfigWindow : public Window {
 				break;
 			}
 
-			case WID_AIC_GAMELIST: {
-				this->selected_slot = OWNER_DEITY;
-				this->InvalidateData();
-				if (click_count > 1 && this->selected_slot != INVALID_COMPANY && _game_mode != GM_NORMAL) ShowAIListWindow((CompanyID)this->selected_slot);
-				break;
-			}
-
 			case WID_AIC_LIST: { // Select a slot
 				this->selected_slot = (CompanyID)this->vscroll->GetScrolledRowFromWidget(pt.y, this, widget);
 				this->InvalidateData();
@@ -947,7 +877,7 @@ struct AIConfigWindow : public Window {
 				if (!_network_available) {
 					ShowErrorMessage(STR_NETWORK_ERROR_NOTAVAILABLE, INVALID_STRING_ID, WL_ERROR);
 				} else {
-					ShowNetworkContentListWindow(nullptr, CONTENT_TYPE_AI, CONTENT_TYPE_GAME);
+					ShowNetworkContentListWindow(nullptr, CONTENT_TYPE_AI);
 				}
 				break;
 		}
@@ -968,10 +898,10 @@ struct AIConfigWindow : public Window {
 
 		this->SetWidgetDisabledState(WID_AIC_DECREASE, GetGameSettings().difficulty.max_no_competitors == 0);
 		this->SetWidgetDisabledState(WID_AIC_INCREASE, GetGameSettings().difficulty.max_no_competitors == MAX_COMPANIES - 1);
-		this->SetWidgetDisabledState(WID_AIC_CHANGE, (this->selected_slot == OWNER_DEITY && _game_mode == GM_NORMAL) || this->selected_slot == INVALID_COMPANY);
+		this->SetWidgetDisabledState(WID_AIC_CHANGE, this->selected_slot == INVALID_COMPANY);
 		this->SetWidgetDisabledState(WID_AIC_CONFIGURE, this->selected_slot == INVALID_COMPANY || GetConfig(this->selected_slot)->GetConfigList()->size() == 0);
-		this->SetWidgetDisabledState(WID_AIC_MOVE_UP, this->selected_slot == OWNER_DEITY || this->selected_slot == INVALID_COMPANY || !IsEditable((CompanyID)(this->selected_slot - 1)));
-		this->SetWidgetDisabledState(WID_AIC_MOVE_DOWN, this->selected_slot == OWNER_DEITY || this->selected_slot == INVALID_COMPANY || !IsEditable((CompanyID)(this->selected_slot + 1)));
+		this->SetWidgetDisabledState(WID_AIC_MOVE_UP, this->selected_slot == INVALID_COMPANY || !IsEditable((CompanyID)(this->selected_slot - 1)));
+		this->SetWidgetDisabledState(WID_AIC_MOVE_DOWN, this->selected_slot == INVALID_COMPANY || !IsEditable((CompanyID)(this->selected_slot + 1)));
 
 		for (TextfileType tft = TFT_BEGIN; tft < TFT_END; tft++) {
 			this->SetWidgetDisabledState(WID_AIC_TEXTFILE + tft, this->selected_slot == INVALID_COMPANY || (GetConfig(this->selected_slot)->GetTextfile(tft, this->selected_slot) == nullptr));
--- a/src/ai/ai_gui.hpp
+++ b/src/ai/ai_gui.hpp
@@ -12,8 +12,10 @@
 
 #include "../company_type.h"
 
+void ShowAIListWindow(CompanyID slot);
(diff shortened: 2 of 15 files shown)


2TallTyler:
I'm very happy to see this being worked on, and it's always nice to see new contributors getting involved. 

A couple thoughts:
- I'm not sure about the button for configuring the current GS. NewGRF parameters are changed from inside the NewGRF settings window so these don't match. I like the convenience but the inconsistency bothers me. I could be convinced one way or the other. 🙂 
- I think the Generate button should be at the bottom of the window, underneath the new buttons. You might need to rearrange the contents of the columns to match the lengths. I do like how your new buttons are double-height for visibility.

Arastais:
> I'm very happy to see this being worked on, and it's always nice to see new contributors getting involved.
> 
> A couple thoughts:
> 
>* I'm not sure about the button for configuring the current GS. NewGRF parameters are changed from inside the NewGRF settings window so these don't match. I like the convenience but the inconsistency bothers me. I could be convinced one way or the other. 🙂
> 
>* I think the Generate button should be at the bottom of the window, underneath the new buttons. You might need to rearrange the contents of the columns to match the lengths. I do like how your new buttons are double-height for visibility.

@2TallTyler, thanks for your interest. Here's what I think:
1. Like you say, the NewGRF parameters are configured from inside the settings window, while the Game Script has its own dedicated window. Additionally, I wanted to put a dedicated button for the Game Script as there can only be one selected at a time, so having a button to open the dedicated window for it makes sense. For NewGRF on the other hand, you can have multiple active at the same time, so it doesn't make sense to have a dedicated button for configuring NewGRFs; it would basically just be the settings window anyway since you would have to chose which NewGRF to configure. While I do see the presented inconsistency, I think that whatever convenience the button brings is worth, and makes the interface feel more natural as a whole; this dedicated configure button only works for the Game Script like I said, and I believe that it is the most convenient part of this feature.

2. Turns out, having the generate button underneath was my original idea, but I ruled it out as, either on the landscape generation or height map generation (as the 'Generate' button is of a different height between these two), it would cause the columns to be uneven in terms of length, even after rearrangement. Here is what I mean (I moved sea level to the right side):
image
This is for the "New Game" version of the window. The "Play Heightmap" version of this window is actually even, since the generate button is only double height. As you can see, there's a gap after the town names on this window; This also creates inconsistency as this gap is not present on the heightmap version of this window. The best solution I can see to the gap is centering the "Map edges" text and button, but again there's still the consistency of this not being present.
Regardless, if you prefer this version (how it is in the screenshot above and below), let me know and I will commit this version instead. For reference, here's how it would look when playing a heightmap (I moved rivers to the right side):
image

As a side note, would anyone happen to know why the commit check is saying that the commit message is in the incorrect format? It seems to me like it is, unless it's somehow too long. (I do know why the tabs check is failing, and I'll fix that at the end).

glx22:
> As a side note, would anyone happen to know why the commit check is saying that the commit message is in the incorrect format? It seems to me like it is, unless it's somehow too long. (I do know why the tabs check is failing, and I'll fix that at the end).

It's because of the merge commit, your commit message is fine.

2TallTyler:
Good argument for the configure GS button. I'm sold. 

I definitely prefer the generate button below. The gap below town names doesn't bother me — I'm pretty sure a previous version of the game had a similar gap somewhere in this menu which was only eliminated.

Arastais (commented):
The lines that Actions is complaining about has already been fixed in [33da4df](https://github.com/OpenTTD/OpenTTD/pull/10058/commits/33da4df367f48dd4653de9ec1a3fb8d6c6f7b522); It seems to be complaining about an earlier commit. The current changes to `genworld_widget.h` only adds 4 lines with correct tab spacing. Also, the commit message complaint is referring to the commit that merges master into my branch.

glx22:
Indeed, commit checker checks each commit individually.

2TallTyler:
You'll want to do an [interactive rebase](https://www.sitepoint.com/git-interactive-rebase-guide/) to combine all six commits into one, then force-push.

src/genworld_gui.cpp
```diff
@@ -183,6 +183,16 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 				NWidget(NWID_SPACER), SetFill(1, 1),
 			EndContainer(),
 		EndContainer(),
+		NWidget(NWID_SPACER), SetMinimalSize(0, 6), SetFill(1, 1),
+		/* NewGRF and AI/GS Settings */
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE), SetPIP(10, 0, 10),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_NEWGRF_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_NEWGRF_SETTINGS, STR_NULL), SetFill(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_SCRIPT_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_SCRIPT_SETTINGS, STR_NULL), SetFill(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_CONFIG_GS_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_CONFIG_GS, STR_MAPGEN_CONFIG_GS_TOOLTIP), SetFill(1, 0),
```
TrueBrain: Was it a deliberate choice to only give one of the three buttons a tooltip? And if so, why?

src/genworld_gui.cpp
```diff
@@ -183,6 +183,16 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 				NWidget(NWID_SPACER), SetFill(1, 1),
 			EndContainer(),
 		EndContainer(),
+		NWidget(NWID_SPACER), SetMinimalSize(0, 6), SetFill(1, 1),
+		/* NewGRF and AI/GS Settings */
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE), SetPIP(10, 0, 10),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_NEWGRF_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_NEWGRF_SETTINGS, STR_NULL), SetFill(1, 0),
```
TrueBrain: I don't really like the double height. I get your reasoning, but if we start doing this in UIs, people increase the height of buttons they think are important, creating a bit of a wild-west effect. I think we can better stick to an uniform look, and just keep their normal height. One opinion, would love to hear others :)

src/genworld_gui.cpp
```diff
@@ -271,14 +280,24 @@ static const NWidgetPart _nested_heightmap_load_widgets[] = {
...
 					EndContainer(),
 				EndContainer(),
 			EndContainer(),
 		EndContainer(),
+		NWidget(NWID_SPACER), SetMinimalSize(0, 6), SetFill(1, 1),
+		/* NewGRF and AI/GS Settings */
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE), SetPIP(10, 0, 10),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_NEWGRF_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_NEWGRF_SETTINGS, STR_NULL), SetFill(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_SCRIPT_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_SCRIPT_SETTINGS, STR_NULL), SetFill(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_ORANGE, WID_GL_CONFIG_GS_BUTTON), SetMinimalSize(0, 24), SetDataTip(STR_MAPGEN_CONFIG_GS, STR_MAPGEN_CONFIG_GS_TOOLTIP),  SetFill(1, 0),
```
TrueBrain: Even if no gamescript is selected, this button is clickable. What does that do in that case? (hopefully not crash :D). Too lazy to test, sorry :P

TrueBrain (commented):
Although maybe not for this PR, but the 3rd button kinda bothers me. And I understand your reasoning for making that button, but the UI now feels weird.

NewGRF -> AI/GS -> GS

It is a bit confusing why GS is there twice. In an ideal world, I guess, the AI/GS should be come just AI, and the GS should allow changing the GS. That would give a much more solid flow, I would think.

But I can fully understand if that is a bit too much for a change at once, but I would love to read your view on it.

src/genworld_gui.cpp
2TallTyler: I could go either way on this. Note that the Generate button is already bigger because it's important. 😉

src/genworld_gui.cpp
Arastais: No, actually good catch. What happened was that I was originally reusing strings (I had set it to the same tooltip as the one on the title screen), but then realized it was better not to reuse strings, so I made it null because I intended to make a tooltip for it but got caught up and forgot. I'll correct that in the next FP

src/genworld_gui.cpp
Arastais: I personally did this because I felt that if I put a single height row below the water edge selections it would disturb the flow of the UI too much and be harder to see, it wasn't just because it was important, though @2TallTyler's point is valid. Also, it discerns these 3 from the rest of the settings, which are land generation based; These 3 are outside of that scope.

TrueBrain: Maybe a panel might also work out here. GUI design is hard :(

PeterN:
1. I think the GS settings button is not needed as that comes under AI/GS configuration, much like configuring individual AIs.
2. I also don't like the choice of double height buttons.
3. Also I don't really see why you have moved the Sea Level dropdown and the Generate button, and made that really really huge.

2TallTyler:
Moving the Generate button was my suggestion above. 😛 It looked strange to have a lot of buttons below the Generate button. Maybe keeping it at the bottom, but making it normal or double height instead of triple, would look better.

Arastais:
> 1. I think the GS settings button is not needed as that comes under AI/GS configuration, much like configuring individual AIs.
> 
>     2. I also don't like the choice of double height buttons.
> 
>     3. Also I don't really see why you have moved the Sea Level dropdown and the Generate button, and made that really really huge.

1. The whole point of this is so that you can skip a bunch of steps. Instead of having to open up the menu, select the game script, then select the configure GS button, which is way smaller too, you can do all that in one click.
2. It seems me and @2TallTyler agreed on double height buttons, but it's not unanimous. Maybe it's worthy of a discussion in the discord.
3. This was @2TallTyler's suggestion and I honestly like it more than the original design, even with single height buttons. Do note that I started with the original design, which is seen in the pictures of the first post of this thread.

src/genworld_gui.cpp
Arastais: It just shows nothing. I think this is better than the alternative, which is to have the button disabled in the window
image

TrueBrain: Good, but also confirms that this would be rather confusing to a lot of of players :)

PeterN:
I think splitting AI/GS windows should be a separate commit before adding the buttons to the worldgen window.

And please rebase to remove the merge.

src/ai/ai_gui.cpp
```diff
@@ -594,13 +575,21 @@ struct AISettingsWindow : public Window {
 	{
 		return _game_mode == GM_MENU || ((this->slot != OWNER_DEITY) && !Company::IsValidID(this->slot)) || (config_item.flags & SCRIPTCONFIG_INGAME) != 0;
 	}
+	void SetValue(int value) {
```
glx22: missing newline

src/ai/ai_gui.hpp
```diff
@@ -11,9 +11,10 @@
 #define AI_GUI_HPP
 
 #include "../company_type.h"
-
```
glx22, line 14: don't remove the newline

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,477 @@
...
+			}
+		}
+
+		this->vscroll->SetCount((int)this->visible_settings.size());
+	}
+
+	void UpdateWidgetSize(int widget, Dimension* size, const Dimension& padding, Dimension* fill, Dimension* resize) override
+	{
+		switch (widget) {
+		case WID_GSC_SETTINGS:
```
glx22: wrong identation

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,477 @@
...
+	 */
+	static bool IsEditable()
+	{
+		return _game_mode != GM_NORMAL || Game::GetInstance() != nullptr;
+	}
+
+	void DrawWidget(const Rect& r, int widget) const override
+	{
+		switch (widget) {
+		case WID_GSC_GSLIST: {
```
glx22: indentation again

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,477 @@
...
+				StringID str;
+				TextColour colour;
+				uint idx = 0;
+				if (StrEmpty(config_item.description)) {
+					if (!strcmp(config_item.name, "start_date")) {
+						/* Build-in translation */
+						str = STR_AI_SETTINGS_START_DELAY;
+						colour = TC_LIGHT_BLUE;
+					}
+					else {
```
glx22:
```suggestion
					} else {
```

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,477 @@
...
+						str = STR_JUST_STRING;
+						colour = TC_ORANGE;
+					}
+				}
+				else {
```
glx22:
```suggestion
				} else {
```

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,477 @@
...
+		if (widget >= WID_GSC_TEXTFILE && widget < WID_GSC_TEXTFILE + TFT_END) {
+			if (GameConfig::GetConfig() == nullptr) return;
+
+			ShowScriptTextfileWindow((TextfileType)(widget - WID_GSC_TEXTFILE), (CompanyID)OWNER_DEITY);
+			return;
+		}
+
+		switch (widget) {
+
+		case WID_GSC_GSLIST: {
```
glx22: indentation

glx22 (commented):
I won't comment on all locations with coding style issues ;)
Most are caused by VS.

Arastais:
Sorry about all the force pushes, had to fix merge conflicts and add your suggestions

This is what the AI window looks like now:
wg-settings-ai

And this is the Game Script window:
wg-settings-gs

I have tested both of them as much as I could and they both work as intended. The reset button in the Game Script window works the same way it did before, so it doesn't reset values it's not supposed to in-game. **Note however** that because I changed an existing StringID in english.txt (instead of making a new one) the "Select Game Script"/"Select AI" texts will show up incorrectly (they'll be blue and won't say select), so translations should be updated properly. 

To coincide with the split, the title menu now looks like this:
wg-settings-menu

And the in-game toolbar looks like this:
wg-settings-toolbar

The world gen windows look the same as before these recent force-pushes.

The title screen may look off now, but I did what felt the simplest in terms of changing the layout of the buttons. The layout of the title screen can obviously be discussed.

> I think splitting AI/GS windows should be a separate commit before adding the buttons to the worldgen window.
If this is agreed upon, I can remove it being added to worldgen window and change this PR to just splitting the config windows.

src/ai/ai_gui.cpp
```diff
@@ -594,13 +575,22 @@ struct AISettingsWindow : public Window {
 	{
 		return _game_mode == GM_MENU || ((this->slot != OWNER_DEITY) && !Company::IsValidID(this->slot)) || (config_item.flags & SCRIPTCONFIG_INGAME) != 0;
 	}
+
+	void SetValue(int value) {
```
glx22:
```suggestion
	void SetValue(int value)
	{
```

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+			case WID_GSC_CHANGE:  // choose other Game Script
+				ShowAIListWindow((CompanyID)OWNER_DEITY);
+				break;
+
+			case WID_GSC_CONTENT_DOWNLOAD:
+				if (!_network_available) {
+					ShowErrorMessage(STR_NETWORK_ERROR_NOTAVAILABLE, INVALID_STRING_ID, WL_ERROR);
+				}
+				else {
```
glx22:
```suggestion
				} else {
```

src/ai/gs_gui.cpp
```diff
@@ -0,0 +1,475 @@
...
+		HideDropDownMenu(this);
+		this->CloseChildWindows(WC_QUERY_STRING);
+	}
+private:
+	bool IsEditableItem(const ScriptConfigItem &config_item) const
+	{
+		return _game_mode == GM_MENU || (config_item.flags & SCRIPTCONFIG_INGAME) != 0;
+	}
+
+	void SetValue(int value) {
```
glx22:
```suggestion
	void SetValue(int value)
	{
```

glx22 (commented):
Seems VS doesn't want to cooperate ;)

Arastais:
As a side note for the main menu/title screen UI, a good option would be to remove the highscore table (similar to what #7786 does) and thus remove the button from the main menu and replace it with the "NewGRF Settings" button. This way "AI Settings" and "Game Script Settings" will be on their own row and the UI will be properly lined up again. Keep in mind that this of course would interfere with #7786 since the "NewGRF Settings" button goes where the "Help and Manuals" button would go in #7786.

LC-Zorg:
I like the idea of ​​shortcuts in the new game window. I also like the idea of ​​creating separate windows for AI and GS selection and settings. It could use other changes there as well, but that's a separate topic. However, I don't know if the proposed changes are the best way. What I definitely miss in this window, apart from these three shortcuts, is also a shortcut to gameplay settings - not all settings, including those related to the interface, but those that define the gameplay. In general, there are some interesting concepts that would be worth considering.

Some time ago, Andrew350 proposed, probably not the first, to combine the NewGRF and AI/GS setup windows. This solution solves the problem of the three buttons in the intro.
https://user-images.githubusercontent.com/50259667/102848133-e639cd00-43c8-11eb-8697-23125bd8c6e0.png

Over a decade ago andythenorth created a very interesting and inspiring mockup for me, showing the concept of a new game settings window (I think).
https://www.tt-forums.net/download/file.php?id=128801

Also TrueBrain introduced the concept of redesigning the new game window as well as the intro, where the main change would be to add a preset selection option, which would be the player's primary choice and which I think is a really enjoyable and useful solution.
https://user-images.githubusercontent.com/1663690/108421875-14575400-7236-11eb-8468-78dbebe6e2c7.png
https://user-images.githubusercontent.com/1663690/108423928-c859de80-7238-11eb-99e0-20e5ffe00474.png

I also have my own concept largely inspired by the ideas above. It is not complete and not everything is consistent.
In brief:
1. By pressing "Singleplayer" you open a new game creation window
2. The basic choice are presets, that is, saved various configurations of map, game, AI, GS and NewGRF settings
3. Random maps are under the "Heightmap" tab
4. At the bottom of the "Heightmap" tab you have shortcuts to AI, GS and NewGRF settings, as well as in a separate line to gameplay settings
5. The shortcuts AI, GS and NewGRF open the same window with a different tab
6. A shortcut to gameplay settings opens the window of all merged settings with the "Gameplay" tab open.
Presets and setting shortcuts

Sorry, I'm not presenting you with a ready-made concept that you can simply approve or reject. However, I mention it because I think it is worth considering what it should look like. :)

Arastais:
@LC-Zorg The main goal of this PR was to add the GS/AI settings to the world gen window. I split GS and AI after the fact because there was a disagreement about the GS parameter config button, and I agreed with the suggestion to split them. However, I wanted to split them as minimally as possible and keeping to the same layout/design as much as I could. Thus, while your suggestions are valid, I feel they are beyond the scope of this pr, especially when trying to drastically alter the title menu or NewGRF settings as well. Your suggestions would be a good PR on its own.

2TallTyler:
Needs a rebase 🙂

src/ai/CMakeLists.txt
```diff
@@ -11,4 +11,6 @@ add_files(
     ai_instance.hpp
     ai_scanner.cpp
     ai_scanner.hpp
+    gs_gui.cpp
```
frosch123: Shouldn't these go into src/game/ ?

frosch123:
I like this a lot :)

Is there a reason to keep the 3 buttons in the title menu?
They only affect new games, and now they are in the mapgen gui. So I would remove them from the title menu.

I am not sure how the scenario editor should work (I never use it myself). Maybe the easiest option would be to also add the three buttons to `_nested_create_scenario_widgets` ?

src/ai/CMakeLists.txt
Arastais: forgot that the `game` folder is for Game Scripts lol. Fixed this in the most recent force-push at time of writing.

Arastais:
> Needs a rebase 🙂

Fixed merge conflicts then rebased👏 Should be good now since, despite the failed commit check, the commit message is in the correct format.
<br/>
> Is there a reason to keep the 3 buttons in the title menu? They only affect new games, and now they are in the mapgen gui. So I would remove them from the title menu.

I would think to keep them in the title menu; a couple reasons come off the top of my head. They're convenient for if you just want quick access to ai, game script, or newgrf straight from when you load the game (for example, you load the game just to check what newgrf or ais you have). Not to mention, new players probably wouldn't notice the existence of ai, game script, or newgrf as much if they're underneath having to create a new game.
<br/>

> I am not sure how the scenario editor should work (I never use it myself). Maybe the easiest option would be to also add the three buttons to `_nested_create_scenario_widgets` ?

The scenario editor has the in-game toolbar which has the settings button with the "Game Script settings" and "AI Settings" options already. Adding to `_nested_create_scenario_widgets` is thus not necessary because it would be redundant.

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file game_gui.cpp %Window for configuring the Games */
```
2TallTyler:
```suggestion
/** @file game_gui.cpp %Window for configuring the Game Script */
```

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+#include "../table/strings.h"
+#include "../game/game.hpp"
+#include "../game/game_config.hpp"
+#include "../game/game_info.hpp"
+#include "../game/game_instance.hpp"
+
+#include "../safeguards.h"
+
+
+/** Widgets for the configure AI window. */
```
2TallTyler: Quite a few of the Doxygen headers in this file say AI window. Shouldn't it be GS window?

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+	0,
+	_nested_gs_config_widgets, lengthof(_nested_gs_config_widgets)
+);
+
+/**
+ * Window to configure which AIs will start.
+ */
+struct GSConfigWindow : public Window {
+	ScriptConfig* gs_config; ///< The configuration we're modifying.
+	int line_height;         ///< Height of a single AI-name line.
```
2TallTyler: Also says AI instead of GS

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+				this->line_height = FONT_HEIGHT_NORMAL + WD_MATRIX_TOP + WD_MATRIX_BOTTOM;
+				size->height = 1 * this->line_height;
+				break;
+		}
+	}
+
+	/**
+	 * Can the GS config be edited?
+	 * @param slot The slot to query.
+	 * @return True if and only if the given AI Config slot can e edited.
```
2TallTyler:
```suggestion
	 * @return True if the given GS Config slot can be edited, otherwise false.
```

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+					this->CloseChildWindows(WC_QUERY_STRING);
+					HideDropDownMenu(this);
+					this->clicked_row = num;
+					this->clicked_dropdown = false;
+				}
+
+				bool bool_item = (config_item.flags & SCRIPTCONFIG_BOOLEAN) != 0;
+
+				int x = pt.x - wid->pos_x;
+				if (_current_text_dir == TD_RTL) x = wid->current_x - 1 - x;
```
2TallTyler, line 312: Is this `1` an indent of some type? Probably should be a constant. (See #9827 for some constant indentation examples)

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+				x -= 4;
```
2TallTyler, line 313: Also probably a constant?

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+						const NWidgetBase* wid = this->GetWidget<NWidgetBase>(WID_GSC_SETTINGS);
+						int rel_y = (pt.y - (int)wid->pos_y) % this->line_height;
+
+						Rect wi_rect;
+						wi_rect.left = pt.x - (_current_text_dir == TD_RTL ? SETTING_BUTTON_WIDTH - 1 - x : x);
+						wi_rect.right = wi_rect.left + SETTING_BUTTON_WIDTH - 1;
+						wi_rect.top = pt.y - rel_y + (this->line_height - SETTING_BUTTON_HEIGHT) / 2;
+						wi_rect.bottom = wi_rect.top + SETTING_BUTTON_HEIGHT - 1;
+
+						/* For dropdowns we also have to check the y position thoroughly, the mouse may not above the just opening dropdown */
```
2TallTyler: This comment doesn't make sense. Words out of order?

src/window_type.h
```diff
@@ -13,6 +13,7 @@
 /** %Window numbers. */
 enum WindowNumberEnum {
 	WN_GAME_OPTIONS_AI = 0,          ///< AI settings.
+	WN_GAME_OPTIONS_GS,
```
2TallTyler: Missing Doxygen

src/intro_gui.cpp
```diff
@@ -449,9 +451,11 @@ static const NWidgetPart _nested_select_game_widgets[] = {
 
 	/* 'script settings' and 'newgrf settings' buttons */
```
2TallTyler: Needs updating now that GS setting button is there too

src/genworld_gui.cpp
```diff
@@ -815,6 +835,18 @@ struct GenerateLandscapeWindow : public Window {
 				_settings_newgame.game_creation.water_borders = (_settings_newgame.game_creation.water_borders == BORDERS_RANDOM) ? 0 : BORDERS_RANDOM;
 				this->InvalidateData();
 				break;
+
+			case WID_GL_AI_BUTTON: //AI Settings
```
2TallTyler: Comments always have a space after the `//` (also the next two cases)

src/genworld_gui.cpp
```diff
@@ -276,14 +286,24 @@ static const NWidgetPart _nested_heightmap_load_widgets[] = {
...
 								NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_TOWNNAME_DROPDOWN), SetDataTip(STR_BLACK_STRING, STR_MAPGEN_TOWN_NAME_DROPDOWN_TOOLTIP), SetFill(1, 0),
 							EndContainer(),
 						EndContainer(),
-						NWidget(WWT_PUSHTXTBTN, COLOUR_GREEN, WID_GL_GENERATE_BUTTON), SetMinimalSize(84, 0), SetDataTip(STR_MAPGEN_GENERATE, STR_NULL), SetFill(1, 1),
 					EndContainer(),
 				EndContainer(),
 			EndContainer(),
 		EndContainer(),
+		NWidget(NWID_SPACER), SetMinimalSize(0, 6), SetFill(1, 1),
+		/* NewGRF and AI/GS Settings */
```
2TallTyler:
```suggestion
		/* AI, GS, and NewGRF settings */
```

src/genworld_gui.cpp
```diff
@@ -186,6 +187,16 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 				NWidget(NWID_SPACER), SetFill(1, 1),
 			EndContainer(),
 		EndContainer(),
+		NWidget(NWID_SPACER), SetMinimalSize(0, 6), SetFill(1, 1),
+		/* NewGRF and AI/GS Settings */
```
2TallTyler:
```suggestion
		/* AI, GS, and NewGRF settings */
```

2TallTyler (commented):
I haven't delved too deep into how the code works, but here's a bunch of code style and documentation items.

The failed commit check is a format thing. You can't have more than one type. Personally I'd call this a Feature.

src/game/game_gui.cpp
Arastais: Interestingly enough, other files in the `src/game` folder say `Game` instead of `Game Script`, but I'll commit your suggestion anyway

src/game/game_gui.cpp
Arastais: This section is copied over from the `AISettingsWindow`, since that window is now embedded into the GS config window. Thus, this `1` is just ported over from that window (See [ai_gui.cpp](https://github.com/OpenTTD/OpenTTD/blob/8c4a70e05ea6ad1387961f96e377a00fac66fdd7/src/ai/ai_gui.cpp#L458)) and I didn't end up changing it.

I was considering looking through the code and also making all the re-used code between both `ai_gui.cpp` and `game_gui.cpp` actually shared instead of duplicated, but decided not to because I wanted to make the changes as minimal as possible for this PR. If I should make more significant changes and make the code shared then let me know.

src/game/game_gui.cpp
Arastais: Same reasoning as [above](https://github.com/OpenTTD/OpenTTD/pull/10058/#discussion_r1015415356).

src/game/game_gui.cpp
Arastais: Same reasoning as https://github.com/OpenTTD/OpenTTD/pull/10058/#discussion_r1015415356. Was copied over from `ai_gui.cpp`

src/game/game_gui.cpp
2TallTyler: After some discussion on Discord, it seems that #10114 might make this a non-issue.

src/game/game_gui.cpp
2TallTyler: It should either be changed to make sense or removed, then. No sense keeping it the way it is. 🙂

2TallTyler (changes requested):

src/genworld_gui.cpp
```diff
@@ -126,7 +126,7 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 						EndContainer(),
 						NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_DATE, STR_NULL), SetFill(1, 1),
 						NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_SMOOTHNESS, STR_NULL), SetFill(1, 1),
-						NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_QUANTITY_OF_RIVERS, STR_NULL), SetFill(1, 1),
```
2TallTyler, line 129: This went missing and didn't get re-added elsewhere. 🙁

2TallTyler (commented):
If you're moving items around in the generate world GUI because of the moved Generate button, I'd suggest the following:

Left side (land generation):
- Map size
- Land generator
- Terrain type
- Variety distribution
- Smoothness
- Map edges

Right side (everything besides land):
- Date
- Sea level
- Rivers
- Town names
- No. of towns
- No. of industries

That gives a nice even 6 items on each side. 🙂

Arastais:
> This went missing and didn't get re-added elsewhere. 🙁

Good catch; The merge conflict was an absolute mess so I forgot to put that back in. It's there now.
<br/>
> If you're moving items around in the generate world GUI because of the moved Generate button, I'd suggest the following:
> 
> Left side (land generation):
> 
>     * Map size
> 
>     * Land generator
> 
>     * Terrain type
> 
>     * Variety distribution
> 
>     * Smoothness
> 
>     * Map edges
> 
> 
> Right side (everything besides land):
> 
>     * Date
> 
>     * Sea level
> 
>     * Rivers
> 
>     * Town names
> 
>     * No. of towns
> 
>     * No. of industries
> 
> 
> That gives a nice even 6 items on each side. 🙂

You forgot about snow/desert coverage, so I can't do this exactly 😭. I tried to make the new game menu as consistent as possible with the heightmap generation menu, while also trying to keep these suggestions. Thus, I've moved rivers to the left half on both windows, but kept sea level on the right since that's only on the new game menu. I've also added some comments in the source code to help clarify the structure of the GUI tree and cleaned up the structure a bit.

Here's what the windows look like now:
new_game
heightmap

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+			case WID_GSC_GSLIST:
+				this->line_height = FONT_HEIGHT_NORMAL + WD_MATRIX_TOP + WD_MATRIX_BOTTOM;
+				size->height = 1 * this->line_height;
+				break;
+		}
+	}
+
+	/**
+	 * Can the GS config be edited?
+	 * @param slot The slot to query.
```
2TallTyler: Parameter doesn't exist

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+	NWidget(NWID_HORIZONTAL),
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_MAUVE, WID_GSC_ACCEPT), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_AI_SETTINGS_CLOSE, STR_NULL),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_MAUVE, WID_GSC_RESET), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_AI_SETTINGS_RESET, STR_NULL),
+		EndContainer(),
+		NWidget(WWT_RESIZEBOX, COLOUR_MAUVE),
+	EndContainer(),
+};
+
+
```
2TallTyler: Extra newline

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+#include "../network/network_content.h"
+#include "../textfile_gui.h"
+#include "../widgets/dropdown_type.h"
+#include "../widgets/dropdown_func.h"
+#include "../hotkeys.h"
+#include "../core/geometry_func.hpp"
+#include "../guitimer_func.h"
+#include "../company_cmd.h"
+#include "../misc_cmd.h"
+
```
2TallTyler, line 34: Extra newline (maybe two)

src/game/game_gui.cpp
```diff
@@ -0,0 +1,467 @@
...
+	ScriptConfig* gs_config; ///< The configuration we're modifying.
+	int line_height;         ///< Height of a single GS-name line.
+	int clicked_button;      ///< The button we clicked.
+	bool clicked_increase;   ///< Whether we clicked the increase or decrease button.
+	bool clicked_dropdown;   ///< Whether the dropdown is open.
+	bool closing_dropdown;   ///< True, if the dropdown list is currently closing.
+	GUITimer timeout;        ///< Timeout for unclicking the button.
+	int clicked_row;         ///< The clicked row of settings.
+	Scrollbar* vscroll;      ///< Cache of the vertical scrollbar.
+	typedef std::vector<const ScriptConfigItem*> VisibleSettingsList;
```
2TallTyler: No doxygen

2TallTyler (commented):
The new layout looks great. And gives me space to do #10093. 😉 

Sorry, I don't know if these are new changes to review or if I missed them before.

src/game/game_gui.cpp
Arastais: `ai_gui.cpp` has a single matching newline, so I'll keep one newline here for consistency.

2TallTyler (approved):
Thanks for being patient with all my nitpicks. 🙂

PR #10065, opened 2022-09-30
Fix #8252: Remove duplicate functionality in `screenshot.cpp`

### LINKED ISSUE #8252: Possible duplication of functionality in `screenshot.cpp`
## Version of OpenTTD
Since e04ca904a9455afc63aa87db775fe4463ab899b7

The introduction of the minimap screenshot feature appears to be duplicating functionality that will be used elsewhere for drawing the widgets within the minimap window; specifically, the `GetMinimapOwner` function is very likely to be replicating functionality used by the _owner information_ overlay of the minimap window.

## Expected result
We shouldn't have multiple codepaths that perform substantially similar functions.

Arastais:
## Motivation / Problem
Fixes #8252 by removing all duplicate (small map tile related) functionality in [`screenshot.cpp`](https://github.com/OpenTTD/OpenTTD/blob/73e5c57e6ba3d49f2ecd4c781d8e4ff128764c2d/src/screenshot.cpp#L1004).

## Description
This fix completely removes the [`GetMinimapOwner(...)`](https://github.com/OpenTTD/OpenTTD/blob/73e5c57e6ba3d49f2ecd4c781d8e4ff128764c2d/src/screenshot.cpp#L1004) function and the first half of the [`MinimapScreenCallback(...)`](https://github.com/OpenTTD/OpenTTD/blob/73e5c57e6ba3d49f2ecd4c781d8e4ff128764c2d/src/screenshot.cpp#L1025) function, as both duplicated functionality already present within [`GetSmallMapOwnerPixels(...)`](https://github.com/OpenTTD/OpenTTD/blob/73e5c57e6ba3d49f2ecd4c781d8e4ff128764c2d/src/smallmap_gui.cpp#L563) in `smallmap_gui.cpp`. `MinimapScreenCallback(...)` now just reuses `GetSmallMapOwnerPixels(...)` with a byte mask to get the palette colour.

`GetSmallMapOwnerPixels(...)` had to be changed for this however (since screenshots don't include the heightmap while the minimap does) and thus was given a new enum parameter: `include_heightmap`. It determines if and how the minimap pixels should include the heightmap/contour. In `screenshot.cpp` (specifically  `MinimapScreenCallback(...)`), this parameter is set to `IH_NEVER` so that the heightmap will never be shown regardless and thus the pixels will always be the default land color. In `smallmap_gui.cpp`, this parameter is set to `IH_IF_ENABLED` so that the heightmap will only be shown if the user presses the button to enable it in the minimap GUI. Critically, I made this parameter an enum and not a boolean because:
1. I wanted the parameter to be more clear to developers than just a boolean. In my opinion, an enum value of `IH_NEVER` seems more clear than just a true/false, and the source code of the function (not to mention the function itself) is generally easier to read.
2. I wanted developers to have more control over how the function handles heightmap colours/pixels without having to change the function. If in the future, a developer ever wants to re-use the function again, they can easily specify how the heightmap colours should be handled so that it better matches the intended functionality. 
3. I wanted developers to be able to add any values in the future to the enum, so that one could easily add ways to handle the small map heightmap, without having to change the function prototype.

Importantly, there is also a third, currently unused option: `IH_ALWAYS`, which always shows the heightmap colours/pixels regardless of any player input. While this option may seem unnecessary (since it's unused right now), I put it in anyway for future-proofing. I wanted to add this functionality/value because it could be used for future features; A good example of such a feature would be taking a minimap screenshot with the contour/heightmap showing as well. This idea is also a good example of my reasoning for making this parameter an enum in the first place.

## Limitations

While **all parts of the commit that actually fix the issue don't have any limitations or problems**, the new parameter for `GetSmallMapOwnerPixels(...)` does have a small caveat with the `IH_ALWAYS` option (which again, is not a problem right now since it's currently unused). 

If `IH_ALWAYS` is passed, the function may return a 32-bit palette colour or a heightmap colour. Thus, any function that wants to always include the heightmap, but also use the palette colours when a heightmap colour isn't returned, has to manually determine if the return value is a palette colour or a heightmap colour. An example of such function would be if `MinimapScreenCallback(...)` wanted to always show the heightmap, i.e. if you wanted to take a minimap screenshot with the heightmap shown. While I could've corrected this by significantly changing `GetSmallMapOwnerPixels(...)`, I intentionally did not. I do not know how any future developers will/would use `IH_ALWAYS` (or more specifically, how they will use `GetSmallMapOwnerPixels(...)`); Thus, I did not want to remove any possibly desired functionality (maybe developers would want both palette colours and heightmap colours returned), since I really don't know what exact behavior future developers would want. Also, I think that significantly changing the function for an option that has yet to be used is not ideal. I will leave it up to developers using the function to handle and check the return value as they wish. Of course, if this is disagreed upon, I can change the function to be more clear about which type of value/colour it returns (e.g. changing the function to also indicate if the returned value is a heightmap or palette colour). Hopefully this all makes sense; I know that it's a very niche caveat and may never even come up, but I thought I might as well talk about it for reference.

## Checklist for review

### DIFF
--- a/src/smallmap_gui.cpp
+++ b/src/smallmap_gui.cpp
@@ -556,15 +556,18 @@ static inline uint32 GetSmallMapVegetationPixels(TileIndex tile, TileType t)
 /**
  * Return the colour a tile would be displayed with in the small map in mode "Owner".
  *
- * @param tile The tile of which we would like to get the colour.
- * @param t    Effective tile type of the tile (see #SmallMapWindow::GetTileColours).
+ * @note If include_heightmap is IH_NEVER, the return value can safely be used as a palette colour (by masking it to a uint8)
+ * @param tile              The tile of which we would like to get the colour.
+ * @param t                 Effective tile type of the tile (see #SmallMapWindow::GetTileColours).
+ * @param include_heightmap Whether to return the heightmap/contour colour of this tile (instead of the default land tile colour)
  * @return The colour of tile in the small map in mode "Owner"
  */
-static inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t)
+inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t, IncludeHeightmap include_heightmap)
 {
 	Owner o;
 
 	switch (t) {
+		case MP_VOID:     return MKCOLOUR_XXXX(PC_BLACK);
 		case MP_INDUSTRY: return MKCOLOUR_XXXX(PC_DARK_GREY);
 		case MP_HOUSE:    return MKCOLOUR_XXXX(PC_DARK_RED);
 		default:          o = GetTileOwner(tile); break;
@@ -577,7 +580,8 @@ static inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t)
 	if ((o < MAX_COMPANIES && !_legend_land_owners[_company_to_list_pos[o]].show_on_map) || o == OWNER_NONE || o == OWNER_WATER) {
 		if (t == MP_WATER) return MKCOLOUR_XXXX(PC_WATER);
 		const SmallMapColourScheme *cs = &_heightmap_schemes[_settings_client.gui.smallmap_land_colour];
-		return _smallmap_show_heightmap ? cs->height_colours[TileHeight(tile)] : cs->default_colour;
+		return ((include_heightmap == IncludeHeightmap::IfEnabled && _smallmap_show_heightmap) || include_heightmap == IncludeHeightmap::Always)
+			? cs->height_colours[TileHeight(tile)] : cs->default_colour;
 	} else if (o == OWNER_TOWN) {
 		return MKCOLOUR_XXXX(PC_DARK_RED);
 	}
@@ -814,7 +818,7 @@ inline uint32 SmallMapWindow::GetTileColours(const TileArea &ta) const
 			return GetSmallMapVegetationPixels(tile, et);
 
 		case SMT_OWNER:
-			return GetSmallMapOwnerPixels(tile, et);
+			return GetSmallMapOwnerPixels(tile, et, IncludeHeightmap::IfEnabled);
 
 		default: NOT_REACHED();
 	}
--- a/src/smallmap_gui.h
+++ b/src/smallmap_gui.h
@@ -27,6 +27,15 @@ void ShowSmallMap();
 void BuildLandLegend();
 void BuildOwnerLegend();
 
+/** Enum for how to include the heightmap pixels/colours in small map related functions */
+enum class IncludeHeightmap {
+	Never,      ///< Never include the heightmap
+	IfEnabled,  ///< Only include the heightmap if its enabled in the gui by the player
+	Always      ///< Always include the heightmap
+};
+
+uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t, IncludeHeightmap include_heightmap);
+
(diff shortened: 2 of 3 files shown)


src/smallmap_gui.cpp
```diff
@@ -556,28 +556,32 @@ static inline uint32 GetSmallMapVegetationPixels(TileIndex tile, TileType t)
...
-static inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t)
+inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t, IncludeHeightmap include_heightmap)
 {
 	Owner o;
 
 	switch (t) {
+		case MP_VOID:     return MKCOLOUR_XXXX(PC_BLACK);
 		case MP_INDUSTRY: return MKCOLOUR_XXXX(PC_DARK_GREY);
 		case MP_HOUSE:    return MKCOLOUR_XXXX(PC_DARK_RED);
+		case MP_WATER:    return MKCOLOUR_XXXX(PC_WATER);
```
michicc: Sorry, but this is a regression. Canals, ship depots and locks can be company-owned and are `MP_WATER` tiles. With this change, they are not visible on the smallmap anymore.

src/smallmap_gui.h
```diff
@@ -27,6 +27,15 @@ void ShowSmallMap();
 void BuildLandLegend();
 void BuildOwnerLegend();
 
+/** Enum for how to include the heightmap pixels/colours in small map related functions */
+enum IncludeHeightmap {
+	IH_NEVER,                  ///< Never include the heightmap
+	IH_IF_ENABLED,             ///< Only include the heightmap if its enabled in the gui by the player
+	IH_ALWAYS                  ///< Always include the heightmap
+};
```
michicc: Not a sticking point but just a suggestion: I think we tend to prefer class enums now for new things to limit scope polution. Declaration could look like e.g.:
```suggestion
enum class IncludeHeightmap {
	Never,                  ///< Never include the heightmap
	IfEnabled,             ///< Only include the heightmap if its enabled in the gui by the player
	Always                  ///< Always include the heightmap
};
```

Call sites would then use it as `IncludeHeightmap::Never` and so on.

michicc (dismissed):

src/smallmap_gui.h
Arastais: Will change it to scoped enum. I assumed that because all OpenTTD code i've seen so far does not do so, that I shouldn't as generally I personally prefer scoped enums anyway. I will remember this for the future.

src/smallmap_gui.cpp
Arastais: Ok, I will revert this part of the owner check.

src/smallmap_gui.cpp
```diff
@@ -575,9 +578,10 @@ static inline uint32 GetSmallMapOwnerPixels(TileIndex tile, TileType t)
 	}
 
 	if ((o < MAX_COMPANIES && !_legend_land_owners[_company_to_list_pos[o]].show_on_map) || o == OWNER_NONE || o == OWNER_WATER) {
-		if (t == MP_WATER) return MKCOLOUR_XXXX(PC_WATER);
+		if(t == MP_WATER) return MKCOLOUR_XXXX(PC_WATER);
```
stormcone: Sorry for the nitpick, but where did the space go? :)

michicc:
I think you need to rebase onto current master to the annotations fail to go away.

michicc (approved):

PR #10086, opened 2022-10-17
Doc: Compress PNG images.

Fdawgs:
## Motivation / Problem

Repo size can always be smaller!

## Description

Losslessly compress documentation PNG images using [Pngcrush](https://pmt.sourceforge.io/pngcrush/) to reduce the repo size by a few bytes.

## Limitations

N/A

## Checklist for review


LordAro:
This needs to be done extremely carefully, as PNG compressors tend to fiddle around with image palettes, and OTTD/grfcodec/etc is very specific about the image palette of the image - changing that will cause issues

Also, there's a conflict - which raises the other issue - as soon as this images get edited, the compression will get undone, and so will only serve to bloat the repo size.

Compressing the images in docs is fine though

FLHerne:
*Does* this decrease the repo size? The old versions of the files are still in the history.

`git clone` by default copies the entire repository including all previous commits, so now most people are downloading both the old *and* new versions of the files (or one and a binary diff) resulting in an increase. Similarly, everyone with an existing clone will have to download the new versions when they `git pull`. The only benefit is to users making a shallow clone with `git clone --depth` which is relatively unusual.

Fdawgs:
Good point @FLHerne, will drop this PR. Thanks!

PR #10087, opened 2022-10-17
Change: Double speed for speed rating at stations for road vehicle.

condac:
## Motivation / Problem

Road vehicle never get any speed rating bonus on stations with current formula. Playing with only road vehicles it is impossible to reach the 80% needed to trigger the extra boost on industrial growth. 
With statue and the fastest road vehicle in late game only 79% is reached for cargo after the inital bonus for new vehicles have disapeared. But with trains you can breach the 80% even without statues. 
Beeing a sandbox focused game the players should have the oportunity to trigger all game mechanics using any type of transport. 

## Description
By doubling the road vehicle speed the same way that ships have a double bonus, road vehicles will follow the evolution that trains have very close if compared to the fastest type avaible each year:
Station rating based on fastest avaible type_year
Red line is current bonus, Yellow is after this patch. 

## Note

The speed for road vehicles are multiplied by 4 in the code and then divided by 2 here, and on the wiki it say that road vehicles speed is divided by 2 when used in the station rating. Making it so that people thinks that road vehicles have a bonus already of a factor of 2 for the rating formula. Or one might think that road vehicles have a penalty. 

The reality when looking at the numbers used here is that `t = front->vcache.cached_max_speed` is not the speed &times; 4 but the speed &times; 2 when it reaces this formula. So road vehicles have the same formula as trains. One might think that the intention from the beginning was to have road vehicles at &times; 2 because it looks like max_speed is set to &times; 4 but for some reason it is only &times; 2 when this math is done. 

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1755,7 +1755,7 @@ static void LoadUnloadVehicle(Vehicle *front)
 			case VEH_SHIP:
 			case VEH_ROAD:
 				/* cached_max_speed from ships and road vehicles are double in size when we fetch the values here
-				 * in order for road veichles to get speed bonus on station ratings we keep the values as is */
+				 * in order for road vehicles and ships to get speed bonus on station ratings we keep the values as is */
 				t = front->vcache.cached_max_speed;
 				break;
 
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1753,11 +1753,10 @@ static void LoadUnloadVehicle(Vehicle *front)
 		switch (front->type) {
 			case VEH_TRAIN:
 			case VEH_SHIP:
-				t = front->vcache.cached_max_speed;
-				break;
-
 			case VEH_ROAD:
-				t = front->vcache.cached_max_speed / 2;
+				/* cached_max_speed from ships and road vehicles are double in size when we fetch the values here
+				 * in order for road veichles to get speed bonus on station ratings we keep the values as is */
+				t = front->vcache.cached_max_speed;
 				break;
 
 			case VEH_AIRCRAFT:


src/economy.cpp
```diff
@@ -1757,7 +1757,7 @@ static void LoadUnloadVehicle(Vehicle *front)
 				break;
 
 			case VEH_ROAD:
-				t = front->vcache.cached_max_speed / 2;
+				t = front->vcache.cached_max_speed;
 				break;
```
James103: This switch case can be merged with the one immediately above it now that both share the same code, creating one block as follows:
```c++
case VEH_TRAIN:
case VEH_ROAD:
case VEH_SHIP:
	t = front->vcache.cached_max_speed;
	break;
```

PeterN: I would also appreciate a comment (for ships as well) which explains why the conversion of half-units to units is deliberately not done, rather than possibly just forgotten about.

condac: I find it hard to explain what is hapening here and I agree that the code does not really reflect what is hapening when the max_speed is alreaddy double in size before this function is called. But i have inserted a comment and made the optimizations as suggested.

src/economy.cpp
```diff
@@ -1753,11 +1753,10 @@ static void LoadUnloadVehicle(Vehicle *front)
 		switch (front->type) {
 			case VEH_TRAIN:
 			case VEH_SHIP:
-				t = front->vcache.cached_max_speed;
-				break;
-
 			case VEH_ROAD:
-				t = front->vcache.cached_max_speed / 2;
+				/* cached_max_speed from ships and road vehicles are double in size when we fetch the values here
+				 * in order for road veichles to get speed bonus on station ratings we keep the values as is */
```
2TallTyler:
```suggestion
				 * in order for road vehicles and ships to get speed bonus on station ratings we keep the values as is */
```

2TallTyler (commented):
The switch case isn't truly needed anymore unless you think it adds readability. You could just do:

```
if (front->type == VEH_AIRCRAFT) {
	t = Aircraft::From(front)->GetSpeedOldUnits(); // Convert to old units.
} else {
	/* cached_max_speed from ships and road vehicles are double in size when we fetch the values here
	 * in order for road veichles to get speed bonus on station ratings we keep the values as is */
	t = front->vcache.cached_max_speed;
}
```

In the comment it may also be worth noting why trains don't get a conversion. (Are their speed units accurate already?)

src/economy.cpp
condac: I think the rewrite with if instead of the switch-case are not neccesary, it just adds clutter to the pull request and do not improve readability in my mind, and if the max_speed values change in the future it might be needed again.

frosch123:
The original TTD behavior is that the speed is halved for road vehicles here, so road vehicles are the only vehicles which cannot maximize the speed bonus.

Changing the game mechanics, like modifying ratings or entirely removing the rating mechanic, is subject to add-ons, like NewGRF. There are already NewGRF which change the rating mechanics. The mechanic suggested in this PR can also be implemented in a NewGRF.

So, I vote for rejecting this PR, it does not meet the [goals of the project](https://github.com/OpenTTD/OpenTTD/blob/master/CONTRIBUTING.md#project-goals).

michicc:
I concur with frosch here.

condac:
Please update wiki to not say the speed is dubbled or halfed, because it is not true. If the wiki is not true then this will always be confused as a bug.

PR #10092, opened 2022-10-21
Change: Don't make wide rivers using original landscape generator

2TallTyler:
## Motivation / Problem

In [a comment on the Wide Rivers commit](https://github.com/OpenTTD/OpenTTD/commit/664771d085737f52838c6aa86e0e37bddf9dc47a#commitcomment-87469798), @Limyx826 suggested that maybe wide rivers shouldn't be built when using the original terrain generator.

## Description

Don't generate wide rivers when using original landscape generator.

(Yes, I'm aware that rivers didn't exist at all in TTD, but this change took me all of 5 minutes.)

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -1273,8 +1273,10 @@ static void River_FoundEndNode(AyStar *aystar, OpenListNode *current)
 		}
 	}
 
-	/* If the river is a main river, go back along the path to widen it. */
-	if (data->main_river) {
+	/* If the river is a main river, go back along the path to widen it.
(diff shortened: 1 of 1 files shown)


Limyx826:
My comment about the wide rivers in the Original generation is due to the fact if river is enabled in the generation setting the wide river will cause some weird shapes on the terrain due how Original generation work. Then again this is up to debate whether wide river should be enabled in Original generation.

michicc (approved):
I don't think it's that bad, but potato potato.

PR #10101, opened 2022-10-23
Fix a9a21e78: Depots aren't really stations.

michicc:
## Motivation / Problem

Depots aren't actually stations, see https://github.com/OpenTTD/OpenTTD/pull/9959#pullrequestreview-1152104926

## Description

Revert the typedef to the original, we don't need that and yet can still use the same numerical value for invalid.

## Checklist for review

### DIFF
--- a/src/depot_type.h
+++ b/src/depot_type.h
@@ -12,10 +12,10 @@
 
 #include "station_type.h"
 
-typedef StationID DepotID; ///< Type for the unique identifier of depots.
+typedef uint16 DepotID; ///< Type for the unique identifier of depots.
 struct Depot;
 
-static const DepotID INVALID_DEPOT = INVALID_STATION;
+static const DepotID INVALID_DEPOT = (DepotID)INVALID_STATION;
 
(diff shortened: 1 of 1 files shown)


src/depot_type.h
```diff
@@ -12,10 +12,10 @@
 
 #include "station_type.h"
 
-typedef StationID DepotID; ///< Type for the unique identifier of depots.
+typedef uint16 DepotID; ///< Type for the unique identifier of depots.
 struct Depot;
 
-static const DepotID INVALID_DEPOT = INVALID_STATION;
+static const DepotID INVALID_DEPOT = (DepotID)INVALID_STATION;
```
LordAro, line 18: If depots aren't stations, why is INVALID_DEPOT defined using INVALID_STATION ?

michicc: Because `Order` uses the same field for different things, and in context of #9959 it makes stuff easier if invalid stations and invalid depots are guaranteed to have the same sentinel value.

LordAro (approved):

PR #10104, opened 2022-10-24
Update: [Dev] Improve .editorconfig with C++ formatting rules based on OpenTTD code style

Arastais:
## Motivation / Problem
The current .editorconfig file is very limited and doesn't provide any formatting rules for C++ files. This can cause IDEs to auto-format based on a developer's setting, causing an extra hassle for either the developer and even maintainers if it goes unnoticed.

## Description
This PR only adds almost all possible [C++ specific formatting rules](https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019) to the repo's .editorconfig file and sets their values to best match OpenTTD. **It only touches the .editorconfig file** and is simply a quality-of-life addition for developers/maintainers - it will not be seen by players at all.

Each rule has an end-of-line comment describing what the key-value combination does. The file is sectioned and ordered by the same ordering on the respective listed web pages for the rules/keys.

Most IDEs should support these formatting rules, and the mainstream ones (VS, VS Code, IntelliJ/ReSharper, etc.) for sure do. Worst case (if a developer's IDE doesn't support these rules), then they're the same as before. 

This will prevent developers from having to manually fix incorrect formatting (including wrong auto-formatting done by the IDE) or forget to write in the [OpenTTD code style](https://wiki.openttd.org/en/Development/Coding%20style). This will also mean that maintainers will have to review/suggest code style changes or mistakes as often. Overall, I think this an important quality-of-life improvement that will save both developers and maintainers a lot of time and hassle.

## Limitations
Obviously, not all rules in [OpenTTD's coding style](https://wiki.openttd.org/en/Development/Coding%20style) can be automatically followed by the IDE with just the .editorconfig, but this will prevent any wrong auto-formatting for sure.

Several rules were "skipped" (i.e. not given an value and thus will not override developers' settings for those rules) for a given reason. These reasons are listed in the comments of the rule.

`indent_size`, `tab_width`, and `end_of_line` were not given rules, as they are global and not clearly specified in [OpenTTD's code style](https://wiki.openttd.org/en/Development/Coding%20style)

A good way to review this is to make a test C++ source file and make code style mistakes on purpose, especially common ones, and see if the auto-formatter will correct them as expected.
## Checklist for review

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
 [*.{c,cpp,h,hpp}]
 indent_style = tab
 charset = utf-8
+#indent_size = <integer>
+#tab_width = <integer>
+#end_of_line = <lf, cf, crlf>
+
+### C++ specific editor config rules ###
+## Visual Studio ## https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019 ##
+# Indentation #
+cpp_indent_braces = false                                                       ; Do not indent braces
+cpp_indent_multi_line_relative_to = innermost_parenthesis                       ; When a new line is typed, it's indented relatively to the innermost open parenthesis
+cpp_indent_within_parentheses = indent                                          ; Indent new lines within parenthesis
+cpp_indent_preserve_within_parentheses = true                                   ; In existing code, do not use the setting for alignment of new lines within parentheses
+cpp_indent_case_contents = true                                                 ; Indent case contents
+cpp_indent_case_labels = true                                                   ; Indent case labels
+cpp_indent_case_contents_when_block = false                                     ; Do not indent braces following a case statement
+cpp_indent_lambda_braces_when_parameter = false                                 ; Do not indent braces of lambdas used as parameters
+#cpp_indent_goto_labels                                                         ; No override for "Position of goto labels" (Unspecified)
+cpp_indent_preprocessor = leftmost_column                                       ; Move preprocessor directives to the leftmost column
+cpp_indent_access_specifiers = false                                            ; Do not indent access specifiers
+cpp_indent_namespace_contents = true                                            ; Indent namespace contents
+cpp_indent_preserve_comments = true                                             ; Preserve indentation of comments
+
+# Newline #
+cpp_new_line_before_open_brace_namespace = same_line                            ; Keep open braces for namespaces on the same line, but add a space before the brace
+cpp_new_line_before_open_brace_type = same_line                                 ; Keep open braces for types/classes on the same line, but add a space before the brace
+cpp_new_line_before_open_brace_function = new_line                              ; Move open braces for functions to a new line
+cpp_new_line_before_open_brace_block = same_line                                ; Keep open braces for control blocks on the same line, but add a space before the brace
+cpp_new_line_before_open_brace_lambda = same_line                               ; Keep open braces for lambdas on the same line, but add a space before the brace
+cpp_new_line_scope_braces_on_separate_lines = true                              ; Place scope braces on separate lines
+cpp_new_line_close_brace_same_line_empty_type = false                           ; Do not move closing braces to the same line as opening braces for empty types
+cpp_new_line_close_brace_same_line_empty_function = false                       ; Do not move closing braces to the same line as opening braces for empty function bodies
+cpp_new_line_before_catch = false                                               ; Do not place 'catch' and similar keywords on a new line
+cpp_new_line_before_else = false                                                ; Do not place 'else' on a new line
+cpp_new_line_before_while_in_do_while = false                                   ; Do not place 'while' in a do-while loop on a new line
+
+# Spacing #
+cpp_space_before_function_open_parenthesis = remove                             ; Remove spaces between function names and opening parentheses of argument lists
+cpp_space_within_parameter_list_parentheses = false                             ; Do not insert a space within parentheses of an argument list
+cpp_space_between_empty_parameter_list_parentheses = false                      ; Do not insert a space between parentheses when argument list is empty
+cpp_space_after_keywords_in_control_flow_statements = true                      ; Insert space between keyword and opening parenthesis in control flow statements
+cpp_space_within_control_flow_statement_parentheses = false                     ; Do not insert a space within parentheses of a control statement
+cpp_space_before_lambda_open_parenthesis = false                                ; Do not insert a space before opening parenthesis of lambda argument lists
+cpp_space_within_cast_parentheses = false                                       ; Do not insert a space within parentheses of a C-style cast
+cpp_space_after_cast_close_parenthesis = false                                  ; Do not insert a space after closing parenthesis of C-style cast
+cpp_space_within_expression_parentheses = false                                 ; Do not insert a space within parentheses of a parenthesized expression
+cpp_space_before_block_open_brace = true                                        ; Insert space before opening brace of blocks
+cpp_space_between_empty_braces = false                                          ; Do not insert a space between empty braces
+cpp_space_before_initializer_list_open_brace = false                            ; Do not insert a space before opening brace of uniform initialization and initializer lists
+#cpp_space_within_initializer_list_braces                                       ; No override for "Insert space within braces of uniform initialization and initializer lists" (Spacing for this varies)
+cpp_space_preserve_in_initializer_list = true                                   ; Preserve spaces inside uniform initialization and initializer lists
+cpp_space_before_open_square_bracket = false                                    ; Do not insert space before opening square brackets
+cpp_space_within_square_brackets = false                                        ; Do not insert space within square bracket
+cpp_space_before_empty_square_brackets = false                                  ; Do not insert space before empty square brackets
+cpp_space_between_empty_square_brackets = false                                 ; Do not insert space between empty square brackets
+cpp_space_group_square_brackets = true                                          ; Group square brackets for multi-dimensional arrays together
+cpp_space_within_lambda_brackets = false                                        ; Do not insert space within square brackets for lambdas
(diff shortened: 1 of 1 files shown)


PeterN:
The EditorConfig specification says comments should go on individual lines, can you please move them up.

PeterN:
I tested auto formatting one source file and the only bits it got wrong are where we randomly decide to indent the right hand side of assignments, which I wouldn't expect any editor to magically guess what we want. Nice.

.editorconfig
```diff
@@ -7,3 +7,71 @@ trim_trailing_whitespace = true
 [*.{c,cpp,h,hpp}]
 indent_style = tab
 charset = utf-8
+
+## C++ specific editor config rules (see https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019) ##
+# Indentation #
+cpp_indent_braces = false                                   #Do not indent braces
+cpp_indent_multi_line_relative_to = innermost_parenthesis   #When a new line is typed, it's indented relatively to the innermost open parenthesis
+cpp_new_line_before_open_brace_function = new_line          #Open braces for functions are moved to new line
```
glx22: Duplicated later in the file

Arastais:
Turns out the JetBrains IDEs (namely IntelliJ IDEA, ReSharper, and Rider) have a lot more C++ specific rules than visual studio, so I've added all of them.

Feel free to test it again (even in visual studio to check for unexpected behavior), but it should be good to go.

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_space_around_ternary_operator = insert                                      ; Insert spaces around conditional operators
+
+# Wrapping #
+cpp_wrap_preserve_blocks = never                                                ; Always apply New Lines settings for blocks
+
+
+
+## Rider & ReSharper ## https://www.jetbrains.com/help/resharper/EditorConfig_CPP_CppBlankLinesPageScheme.html ##
+# Blank Lines #
+#cpp_keep_blank_lines_in_declarations                                           ; No override for "Max blank lines in declerations" (Unspecified; Not consistent across source code)
```
rubidium42: decleration -> declaration

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+#cpp_blank_lines_around_function_definition                                     ; No override for "Number of blank lines around function definitions﻿" (Unspecified; Not consistent across source code)
+#cpp_blank_lines_around_single_line_function_definition                         ; No override for "Number of blank lines around single line function definitions﻿" (Unspecified; Not consistent across source code)
+#cpp_blank_lines_around_namespace                                               ; No override for "Number of blank lines around namespaces﻿" (Unspecified; Not consistent across source code)
+#cpp_blank_lines_around_other_declaration                                       ; No override for "Number of blank lines around other definitions and declarations﻿" (Unspecified; Not consistent across source code)
+
+# Braces Layout #
+cpp_namespace_declaration_braces = end_of_line                                  ; Keep open braces for namespace declerations on the same line, but add a space before the brace
```
rubidium42: decleration -> declaration

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_linkage_specification_braces = end_of_line                                  ; Keep open braces for linkage declerations on the same line, but add a space before the brace
```
rubidium42: decleration -> declaration

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+#cpp_align_multiline_calls_chain                                                ; No override for "Align multiline chained method calls" (Unspecified)
+#cpp_outdent_dots                                                               ; No override for "Outdent '.' and '->' in chained method calls﻿ on new lines" (Unspecified)
+cpp_align_multiline_binary_expressions_chain = false                            ; Do not align multiline chained binary expressions
+cpp_int_align_fix_in_adjacent = true                                            ; Fix column alignment in adjacent lines﻿
+cpp_int_align_eq = true                                                         ; Align assignments with adjacent assignments
+cpp_int_align_declaration_names = false                                         ; Do not align declaration names with adjacent declaration names
+cpp_int_align_comments = true                                                   ; Align end-of-line comments with adjacent end-of-line comments
+
+# Spaces #
+cpp_space_before_comma = false                                                  ; Do not put space after ANY commas
```
rubidium42: after -> before

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_space_after_comma = true                                                    ; Put space before ALL commas
```
rubidium42: before -> after

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_space_before_ptr_in_data_member = true                                      ; Put space before ptr in declaration of variable﻿
+cpp_space_after_ptr_in_data_member = false                                      ; Do not put space after ptr in declaration of variable﻿
+cpp_space_before_ptr_in_data_members = true                                     ; Put space before ptr in declaration of multiple variables﻿
+cpp_space_after_ptr_in_data_members = false                                     ; Do not put space after ptr in declaration of multiple variables﻿
+cpp_space_before_ptr_in_method = false                                          ; Do not put space before ptr in return type of function﻿
+cpp_space_after_ptr_in_method = true                                            ; Put space after ptr in return type of function﻿
```
rubidium42: I can't quickly find a single function with `const char* Function()` compared to about 100 cases of `const char *Function()`

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_space_before_ptr_in_abstract_decl = false                                   ; Do not put space before ptr in abstract declaration﻿
+cpp_space_before_ref_in_data_member = true                                      ; Put space before ref in declaration of variable﻿
+cpp_space_after_ref_in_data_member = false                                      ; Do not put space after ref in declaration of variable﻿
+cpp_space_before_ref_in_data_members = true                                     ; Put space before ref in declaration of multiple variables﻿
+cpp_space_after_ref_in_data_members = false                                     ; Do not put space after ref in declaration of multiple variables﻿
+cpp_space_before_ref_in_method = false                                          ; Do not put space before ref in return type of function﻿
+cpp_space_after_ref_in_method = true                                            ; Put space after ref in return type of function﻿
```
rubidium42: I cannot find a single `string& Function()`, but I can find a few `string &Function()` cases. I think this is the wrong way around just like it was for pointers.

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+cpp_space_after_semicolon_in_for_statement = true                               ; Put space after semicolon﻿ in 'for' statements
+cpp_space_before_for_colon = true                                               ; Put space before ':' in range-based for loop﻿
+cpp_space_after_for_colon = true                                                ; Put space after ':' in range-based for loop﻿
+cpp_space_before_colon_in_case = true                                           ; Put space before colon in switch case or label statement﻿
+cpp_space_after_colon_in_case = true                                            ; Put space after colon in switch case or label statement﻿
+cpp_space_around_binary_operator = true                                         ; Put space around binary operator﻿
+cpp_space_around_assignment_operator = true                                     ; Put space around assignment operator﻿
+cpp_space_around_member_access_operator = false                                 ; Do not put space around dot, '->', '.*' and '->.'﻿
+cpp_space_within_parentheses = true                                             ; Put space within any parentheses﻿
```
rubidium42: I don't think this is often the case, and if it's the case then the space has been placed there usually for alignment of parameter purposes as AFAIK the default would be not having spaces within parentheses.

.editorconfig
```diff
@@ -7,3 +7,251 @@ trim_trailing_whitespace = true
...
+#cpp_space_after_type_parameter_constraint_colon                                ; No override for "After C++/CLI generic constraint colon﻿" (Unspecified)
+cpp_space_after_keywords_in_control_flow_statements = true                      ; Put space before parentheses﻿ of control statements﻿
+cpp_space_between_parentheses_of_control_flow_statements = false                ; Do not put space within parentheses﻿﻿ of control statements﻿
+cpp_space_before_semicolon_in_for_statement = false                             ; Do not put space before semicolon﻿ in 'for' statements
+cpp_space_after_semicolon_in_for_statement = true                               ; Put space after semicolon﻿ in 'for' statements
+cpp_space_before_for_colon = true                                               ; Put space before ':' in range-based for loop﻿
+cpp_space_after_for_colon = true                                                ; Put space after ':' in range-based for loop﻿
+cpp_space_before_colon_in_case = true                                           ; Put space before colon in switch case or label statement﻿
```
rubidium42: I don't think this is true.

rubidium42 (commented):
I haven't tested it, so there might be more cases where things are wrong. Though I like the idea of codifying the coding style in this way, as long as the tooling helps.

Arastais:
@rubidium42 I fixed the incorrect rules you pointed out; Thanks for checking those. I don't use resharper so didn't have an incentive to check and also didn't double-check those because I was essentially copying the rules from the visual studio section.

Too bad I didn't spell check 😵

rubidium42:
I just pushed a simple change to move the inline comments, which are forbidden by the specifications as mentioned before by @PeterN, to the line before it. 
I'm not sure what @LordAro meant exactly with the comment about the space after the `#`. For now I made proper comments have the space after the `#` and cases where the configuration is commented out without a space after the `#`.

2TallTyler (approved):
I don't have an encyclopedic knowledge of our coding style but I trust that @rubidium42 does. 😛 Comments look fine to me, I can read it just fine.

PR #10110, opened 2022-10-25
Feature: Hotkey to honk a vehicle's horn

Arastais:
## Motivation / Problem
@2TallTyler created #10034 so that players could press a button in the UI to honk a vehicles horn. It was closed as to not clutter the vehicle view UI for an unnecessary feature, and as such the feature was never added. It was suggested to add the feature to a "hidden" hotkey instead.

## Description
This PR is co-authored by @2TallTyler. It takes his vehicle honking code and ties it to a hotkey instead of a UI button. This means that when you have a vehicle view window open, pressing 'H' (the hotkey set for it) will honk a vehicle's horn. This allows the honking feature to be retained while not cluttering/changing the vehicle view UI at all.

## Limitations
- As before, does not do anything with road vehicles (as there's no susceptible sound for honking in `sound_type.h`) and airplanes (because they don't have horns).
- Does not function in multiplayer; Since the hotkey/horn is spammable, it would probably be too annoying in multiplayer.
- There is no setting to turn this functionality off. Pressing the hotkey will always play the sound (unless effects are turned to 0%; it will still play if effects sounds are disabled instead of turned down).

## Checklist for review

### DIFF
--- a/src/newgrf_sound.cpp
+++ b/src/newgrf_sound.cpp
@@ -181,11 +181,12 @@ SoundID GetNewGRFSoundID(const GRFFile *file, SoundID sound_id)
  * Checks whether a NewGRF wants to play a different vehicle sound effect.
  * @param v Vehicle to play sound effect for.
  * @param event Trigger for the sound effect.
+ * @param force Should we play the sound effect even if vehicle sound effects are muted?
  * @return false if the default sound effect shall be played instead.
  */
-bool PlayVehicleSound(const Vehicle *v, VehicleSoundEvent event)
+bool PlayVehicleSound(const Vehicle *v, VehicleSoundEvent event, bool force)
 {
-	if (!_settings_client.sound.vehicle) return true;
+	if (!_settings_client.sound.vehicle && !force) return true;
 
 	const GRFFile *file = v->GetGRF();
 	uint16 callback;
--- a/src/ship_cmd.cpp
+++ b/src/ship_cmd.cpp
@@ -272,16 +272,10 @@ void Ship::MarkDirty()
 	this->UpdateCache();
 }
 
-static void PlayShipSound(const Vehicle *v)
+void Ship::PlayLeaveStationSound(bool force) const
 {
-	if (!PlayVehicleSound(v, VSE_START)) {
-		SndPlayVehicleFx(ShipVehInfo(v->engine_type)->sfx, v);
-	}
-}
-
-void Ship::PlayLeaveStationSound() const
-{
-	PlayShipSound(this);
+	if (PlayVehicleSound(this, VSE_START, force)) return;
+	SndPlayVehicleFx(ShipVehInfo(this->engine_type)->sfx, this);
 }
 
 TileIndex Ship::GetOrderStationLocation(StationID station)
@@ -398,7 +392,7 @@ static bool CheckShipLeaveDepot(Ship *v)
 	v->UpdateViewport(true, true);
 	SetWindowDirty(WC_VEHICLE_DEPOT, v->tile);
 
-	PlayShipSound(v);
+	v->PlayLeaveStationSound();
 	VehicleServiceInDepot(v);
 	InvalidateWindowData(WC_VEHICLE_DEPOT, v->tile);
 	SetWindowClassesDirty(WC_SHIPS_LIST);
--- a/src/vehicle_base.h
+++ b/src/vehicle_base.h
@@ -450,8 +450,9 @@ public:
 
 	/**
 	 * Play the sound associated with leaving the station
+	 * @param force Should we play the sound even if sound effects are muted? (horn hotkey)
 	 */
-	virtual void PlayLeaveStationSound() const {}
+	virtual void PlayLeaveStationSound(bool force = false) const {}
 
 	/**
 	 * Whether this is the primary vehicle in the chain.
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -41,6 +41,7 @@
 #include "order_cmd.h"
 #include "roadveh_cmd.h"
 #include "train_cmd.h"
+#include "hotkeys.h"
 
 #include "safeguards.h"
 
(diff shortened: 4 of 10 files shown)


src/newgrf_sound.cpp
```diff
@@ -183,9 +183,9 @@ SoundID GetNewGRFSoundID(const GRFFile *file, SoundID sound_id)
  * @param event Trigger for the sound effect.
  * @return false if the default sound effect shall be played instead.
```
2TallTyler, line 185: Doxygen needs updating for the `bool force` param

src/vehicle_base.h
```diff
@@ -440,8 +440,9 @@ struct Vehicle : VehiclePool::PoolItem<&_vehicle_pool>, BaseVehicle, BaseConsist
 
 	/**
 	 * Play the sound associated with leaving the station
+	 * @param force Should we play the sound even if sound effects are muted? (meant for horn)
```
2TallTyler:
```suggestion
	 * @param force Should we play the sound even if sound effects are muted? (horn hotkey)
```

src/vehicle_gui.cpp
```diff
@@ -3085,15 +3076,44 @@ struct VehicleViewWindow : Window {
...
+	&VehicleViewWindow::hotkeys
+);
 
 /** Shows the vehicle view window of the given vehicle. */
 void ShowVehicleViewWindow(const Vehicle *v)
 {
 	AllocateWindowDescFront<VehicleViewWindow>((v->type == VEH_TRAIN) ? &_train_view_desc : &_vehicle_view_desc, v->index);
 }
 
+
```
2TallTyler: Accidental newline 🙂

2TallTyler (commented):
Just a few nitpicks.

The feature works as advertised with the hotkey trigger. The hotkey is correctly added to the hotkeys config file.

One quirk: When multiple vehicle windows are open, the hotkey makes all of them honk. I don't see a problem with this (and think it's preferable to randomly selecting one to make the sound) but welcome other opinions.

2TallTyler (dismissed):

LordAro:
Given `force` is false in almost all cases, I wonder if a default parameter might make the code a bit cleaner and reduce the size of the change

michicc:
Indeed, a default param on `PlayVehicleSound` would make a lot of sense.

LC-Zorg:
> One quirk: When multiple vehicle windows are open, the hotkey makes all of them honk. I don't see a problem with this (and think it's preferable to randomly selecting one to make the sound) but welcome other opinions.

The honking of all vehicles seems strange and a bit... pointless. Would it add anything of value to the game and would players really like to use it as such? I think it would be worth changing. I imagine four possible solutions here:
1. This function only works for a vehicle window that was opened or used last
2. This function only works for the vehicle closest to the center of the screen
3. This feature only works for the vehicle that is being tracked
4. This function only works when one vehicle window is open and only for that vehicle.

By the way, that button wouldn't be so bad if it were an option available in the settings. :)
Moreover, it could be one of the "extra button" options. Earlier, I thought about adding another button - a shortcut to the list of vehicles sharing orders. Some time ago on the forum there was also a proposal to add a button that would show the route of the vehicle on the game map (I really liked this idea). Upon the idea of ​​adding a signal button, I figured it could be one of the options to choose from.
additional button

But you could also think of adding all three buttons, divided into "info buttons" on one side of the window and "way buttons" on the other side.
more buttons

I mention this because it might be worth considering going back to the button idea. :)
Then there would be no problem with which vehicle should make the sound or where the sound is coming from.

2TallTyler:
Adding additional button space is a bit overkill/out of scope for a silly easter egg like this.

What about if the hotkey only worked when the camera is following a vehicle? That's how this feature works in Transport Fever 2 and Mashinky.

Arastais:
> Given `force` is false in almost all cases, I wonder if a default parameter might make the code a bit cleaner and reduce the size of the change

> Indeed, a default param on `PlayVehicleSound` would make a lot of sense.

Default param is now `false` for both `PlayVehicleSound` and `PlayLeaveStationSound`
<br/>
> Adding additional button space is a bit overkill/out of scope for a silly easter egg like this.
> 
> What about if the hotkey only worked when the camera is following a vehicle? That's how this feature works in Transport Fever 2 and Mashinky.

Hotkey now only honks a vehicle's horn if you are following that vehicle (by pressing double click on the viewport icon)


I've also gone ahead and removed `PlayShipSound` by making `Ship::PlayLeaveStationSound` closer to `Train::PlayLeaveStationSound` to make the source code cleaner and more consistent.

src/ship_cmd.cpp
```diff
@@ -272,16 +272,12 @@ void Ship::MarkDirty()
...
-	if (!PlayVehicleSound(v, VSE_START)) {
-		SndPlayVehicleFx(ShipVehInfo(v->engine_type)->sfx, v);
-	}
-}
+	if (PlayVehicleSound(this, VSE_START, force)) return;
 
-void Ship::PlayLeaveStationSound() const
-{
-	PlayShipSound(this);
+	EngineID engtype = this->engine_type;
```
2TallTyler: This variable is only used once and isn't that complicated. Does it need to exist?

src/vehicle.cpp
```diff
@@ -2731,7 +2731,7 @@ void Vehicle::ShowVisualEffect() const
 		}
 	} while ((v = v->Next()) != nullptr);
 
-	if (sound) PlayVehicleSound(this, VSE_VISUAL_EFFECT);
+	if (sound) PlayVehicleSound(this, VSE_VISUAL_EFFECT, false);
```
2TallTyler: Unneeded `false` now that it's a default parameter value?

src/vehicle_gui.cpp
```diff
@@ -3151,6 +3132,17 @@ struct VehicleViewWindow : Window {
 		}
 	}
 
+	EventState OnHotkey(int hotkey) override
+	{
+		const Window* mainwindow = FindWindowById(WC_MAIN_WINDOW, 0);
+		const Vehicle* v = Vehicle::Get(window_number);
+		/* If the hotkey is not for any widget in the UI (i.e. for honking) and we're following this vehicle */
+		if (hotkey == WID_VV_HONK_HORN && mainwindow->viewport->follow_vehicle == v->index) {
```
2TallTyler: I wonder if we could early-return from this function if `hotkey != WID_VV_HONK_HORN` before creating these variables. I don't think they're particularly expensive but if this fires on every open window every time the user presses a hotkey, it might add up.

2TallTyler (commented):
Works perfectly. I do have a couple nitpicks with the new code though. 🙂

src/vehicle_gui.cpp
Arastais: Didn't think about this, actually. I assume you're referring to `FindWindowById`, which iterates through all open windows and matches the Id and class. On paper this specific iteration might seem costly but like you inferred It's not expensive at all in practice (it's only checking two integer members and there can only be so many windows at once) so you won't notice much of a difference. 

It doesn't hurt though, so I've added what you suggested in the simplest way possible.

src/ship_cmd.cpp
Arastais: I copied this over from `train_cmd.cpp` like I said to simplify the source code (mainly by removing `PlayShipSound()`) and keeping `train_cmd.cpp` and `ship_cmd.cpp` consistent with one another; `train_cmd.cpp` also has the local `engtype` variable. I didn't think much of it and just aimed for consistency, so I didn't bother changing it. Now that you're pointing it out, I've gone ahead and fixed this for both `train_cmd.cpp` and `ship_cmd.cpp`

src/vehicle.cpp
Arastais: I must've missed this when checking the changes. I've removed it.

Arastais (commented):
@2TallTyler I've implemented all your suggestions.

2TallTyler (approved):

PR #10112, opened 2022-10-30
Fix #10023: Allow negative input in text fields when needed

### LINKED ISSUE #10023: [Bug]: "Autorenew when vehicle is X months after max age" text input does not accept negative numbers
### Version of OpenTTD

f279fc57721d558c92a476f80ade51f6d63cbd74

### Expected result

The text input for the setting "Autorenew when vehicle is X months after maximum age" should accept the negative sign `-`, allowing input of negative numbers. This is because the setting's range ( $[-12, 12]$ ) includes negative numbers.

The text inputs for other settings whose range includes negative numbers (if any) should also accept negative numbers.

### Actual result

The text input for the setting "Autorenew when vehicle is X months after maximum age" does not accept `-` for a negative sign, restricting the range of numbers that the setting can take when setting via text input to $[0, 12]$ .

### Steps to reproduce

1. Invoke the text input for the setting "Autorenew when vehicle is X months after maximum age" (category "Company").
2. Try typing a negative number.
3. The negative sign is ignored and only the absolute value of the number entered is used.

hallonsoda79:
## Motivation / Problem
Some settings like "Autorenew when vehicle is X months after maximum age" have negative values in the valid range and should therefore allow negative sign ('-') in the text field.

## Description
Only in a few rare cases settings span between positive and negative values in the valid range. 
For example "Autorenew when vehicle is X months after max age" have [-12,12] as a valid range. 
In these cases also negative sign ('-')  will be allowed in the text field with this fix.

## Checklist for review

### DIFF
--- a/src/string.cpp
+++ b/src/string.cpp
@@ -494,12 +494,13 @@ bool strtolower(std::string &str, std::string::size_type offs)
 bool IsValidChar(WChar key, CharSetFilter afilter)
 {
 	switch (afilter) {
-		case CS_ALPHANUMERAL:  return IsPrintable(key);
-		case CS_NUMERAL:       return (key >= '0' && key <= '9');
-		case CS_NUMERAL_SPACE: return (key >= '0' && key <= '9') || key == ' ';
-		case CS_ALPHA:         return IsPrintable(key) && !(key >= '0' && key <= '9');
-		case CS_HEXADECIMAL:   return (key >= '0' && key <= '9') || (key >= 'a' && key <= 'f') || (key >= 'A' && key <= 'F');
-		default: NOT_REACHED();
+	case CS_ALPHANUMERAL:   return IsPrintable(key);
+	case CS_NUMERAL:        return (key >= '0' && key <= '9');
+	case CS_NUMERAL_SPACE:  return (key >= '0' && key <= '9') || key == ' ';
+	case CS_NUMERAL_SIGNED: return (key >= '0' && key <= '9') || key == '-';
+	case CS_ALPHA:          return IsPrintable(key) && !(key >= '0' && key <= '9');
+	case CS_HEXADECIMAL:    return (key >= '0' && key <= '9') || (key >= 'a' && key <= 'f') || (key >= 'A' && key <= 'F');
+	default: NOT_REACHED();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


src/string.cpp
```diff
@@ -494,12 +494,13 @@ bool strtolower(std::string &str, std::string::size_type offs)
...
-		case CS_ALPHA:         return IsPrintable(key) && !(key >= '0' && key <= '9');
-		case CS_HEXADECIMAL:   return (key >= '0' && key <= '9') || (key >= 'a' && key <= 'f') || (key >= 'A' && key <= 'F');
-		default: NOT_REACHED();
+	case CS_ALPHANUMERAL:   return IsPrintable(key);
+	case CS_NUMERAL:        return (key >= '0' && key <= '9');
+	case CS_NUMERAL_SPACE:  return (key >= '0' && key <= '9') || key == ' ';
+	case CS_NUMERAL_SIGNED: return (key >= '0' && key <= '9') || key == '-';
+	case CS_ALPHA:          return IsPrintable(key) && !(key >= '0' && key <= '9');
+	case CS_HEXADECIMAL:    return (key >= '0' && key <= '9') || (key >= 'a' && key <= 'f') || (key >= 'A' && key <= 'F');
+	default: NOT_REACHED();
```
nielsmh: Indentation seems to have gotten wrong here. (Maybe check your editor settings.)

hallonsoda79: Ok thanks. Yes, vscode wanted to format the code in a completely different way so I had to turn off auto format here.

glx22:
What happens if you type `-` in the middle of the number ?

hallonsoda79:
> What happens if you type `-` in the middle of the number ?

I have done some testing of non valid input and it looks it comes down to the behavior of the "atoll(str)" function inside "OnQueryTextFinished":
* Input "----" will result in autorenew 0 months after (since atoll function will return 0)
* Input "-11-33" will result in autorenew 11 months before maximum age (since atoll function will return -11)
* Input "54-22" will result in autorenew 12 month after maximum age (since atoll will return 54 and then it will later choose maximum value of this setting)

I think above test input and outcome are acceptable, please let me know what you think.

glx22 (approved):

Issue #8998, opened 2021-04-10
Minimap too small on Retina displays


Crazor:
## Version of OpenTTD
1.11.0

## Expected result
I expect to be able to zoom into the minimap at least one, better two more steps.

## Actual result
The minimap is too small on retina screens even when fully zoomed in

## Steps to reproduce
Run OpenTTD on a Mac with Retina display, open the minimap, zoom in all the way.

## System and Settings
* MacBook Air 2020 13" (2560x1600px)
* System display settings: Select "scaled" and then the rightmost icon for maximum area
* UI element size: quad
* Font size: double
* openttd.cfg:
```
small_font = /Users/dhe/Library/Fonts/DejaVuSans.ttf
medium_font = /Users/dhe/Library/Fonts/DejaVuSans-Bold.ttf
large_font = /Users/dhe/Library/Fonts/DejaVuSerif-Bold.ttf
mono_font = /Users/dhe/Library/Fonts/DejaVuSansMono-Bold.ttf
small_size = 14
medium_size = 15
large_size = 20
mono_size = 14
```

The above settings give me a nice and playable OpenTTD experience. Viewport zoom levels and UI size are perfect:
<img width="1792" alt="Bildschirmfoto 2021-04-10 um 16 12 53" src="https://user-images.githubusercontent.com/128295/114272752-b108c780-9a17-11eb-84ff-6db97d9a1bd5.png">

The minimap is not, however. This is a 128x256 map:
<img width="1792" alt="Bildschirmfoto 2021-04-10 um 16 13 03" src="https://user-images.githubusercontent.com/128295/114272794-d7c6fe00-9a17-11eb-929c-2a1cac670f62.png">

I think this configuration could use at least one, if not two more levels to zoom in. I briefly looked into the source code and noticed that as zoom levels are "increase to zoom out", there is no easy way for me to set a zoom level below 1. Fractional zoom levels seem to need a bunch of work. Maybe the minimap scaling should be always blown up to 2x or 4x size depending on the UI element size setting? The following is a 4096x4096 map zoomed out all the way, so maybe there should be one or two additional zoom-out levels added as well then:
<img width="1792" alt="Bildschirmfoto 2021-04-10 um 16 17 56" src="https://user-images.githubusercontent.com/128295/114272928-6e93ba80-9a18-11eb-946f-4097f48f04cd.png">

JGRennison:
[...]

Issue #9009, opened 2021-04-11
Feature request: mouse cursor scale


Hezkore:
I run the game at double interface scale with a TTF font.
Most of the game looks great, but the mouse cursor looks like a giant hideous monster.
I'd really like an option _(just a checkbox)_ for the mouse cursor to NOT be scaled with the rest of the interface.

I think everyone gets the idea here.
But I've been fiddling around with the code myself, and this is before and after.

Before/Scaled Cursor:
![big](https://user-images.githubusercontent.com/1727541/114287924-7e88ba00-9a6b-11eb-8e41-de25dc216b9e.png)

After/Unscaled Cursor:
![small](https://user-images.githubusercontent.com/1727541/114287929-821c4100-9a6b-11eb-8b6a-029e2a3fee8f.png)

LordAro:
Not a bad idea - I think a setting with "GUI size", "1x", "2x", "4x" options would work quite well (defaulting to GUI size)

PeterN:
1) You should see the cursor from the original graphics!
2) There are special cursors that are built up from engine parts (drag & drop) which would need attention.
3) Less is more.

Hezkore:
I'd be happy with just a "Scale mouse pointer" checkbox, that scales the mouse pointer with the interface scale.

nielsmh:
There is also a (separate) concern about having a high-contrast mouse cursor, possibly as a forced override of the baseset.

PR #10114, opened 2022-10-31
Feature: Variable interface scaling

PeterN:
## Motivation / Problem

Powers-of-2 interface scaling, allowing 1x, 2x, and 4x zoom, is too limited, and does not provide a comfortable middle ground.

Additionally window scaling is currently broken. Although padding is scaled at the nested widget level, individually it is left up to each window to scale appropriately.

## Description

This massive set of changes:

1) Changes many rendering routines to work with `Rect` rather than individual left, top, right and bottom, as supplied to `DrawWidget`.
2) Adds `RectPadding`, and new functions to `Rect` apply padding and Indent/Width operations that understand RTL layout decisions.
3) Replaces the `WidgetDrawDistances` enum with pre-scaled `WidgetDimensions`. As padding is now pre-scaled, window scaling is now far more consistent.

**4) Replaces separate GUI and font zoom choices with a single interface scale slider that ranges from 100% to 500%. This can be set in 1% increments, but the UI is limited to 25% increments for ease of use.**

5) Bevels are now scaled. Chunky. The option to toggle this is not currently saved, and may be removed.

image

## Limitations

It's big. This is actually split into a series of branches in my tree, but PRs don't support that AFAIK.

## Checklist for review


src/core/geometry_type.hpp
```diff
@@ -43,12 +43,178 @@ struct Dimension {
...
+	 * @return height of Rect.
+	 */
+	inline int Height() const { return this->bottom - this->top + 1; }
+
+	/**
+	 * Copy and shrink Rect by s pixels.
+	 * @param s number of pixels to remove from each side of Rect.
+	 * @return the new smaller Rect.
+	 */
+	[[nodiscard]] inline Rect Shrink(int s) const {
```
LordAro: Opening function braces should be on a newline

2TallTyler:
I'd appreciate some clarification (in the code) on the various name variations of `Rect tr`, `...ir`, `...rr`, `...br`, etc. I can't figure out the pattern. 🙂

PeterN:
> I'd appreciate some clarification (in the code) on the various name variations of `Rect tr`, `...ir`, `...rr`, `...br`, etc. I can't figure out the pattern. 🙂

No hard pattern to be honest:
`ir` is usually inner rect
`tr` is usually text rect
`br` is bevel rect

PeterN:
Update to interface size auto-detection which was missing. OS DPI scale is now taken literally, so 150% in the OS means 150% in game. This may need tweaking as previous 150% meant 2x.

LordAro:
wasm... still needs work

image

(the game options window)

Also it crashed with `(index):1 Error: Assertion failed at line 1718 of /__w/OpenTTD/OpenTTD/src/widget.cpp: given_width >= this->smallest_x && given_height >= this->smallest_y` when I clicked on "Load game"

src/settings_gui.cpp
```diff
@@ -474,10 +466,44 @@ struct GameOptionsWindow : Window {
...
+				if (ClickSliderWidget(this->GetWidget<NWidgetBase>(widget)->GetCurrentRect(), pt, MIN_INTERFACE_SCALE, MAX_INTERFACE_SCALE, this->gui_scale)) {
+					if (!_ctrl_pressed) this->gui_scale = ((this->gui_scale + 12) / 25) * 25;
+					this->SetWidgetDirty(widget);
+					this->SetWidgetLoweredState(WID_GO_GUI_SCALE_AUTO, false);
+				}
+
+				if (click_count > 0) this->mouse_capture_widget = widget;
+				break;
+
+			case WID_GO_GUI_SCALE_AUTO:
```
LordAro, line 488: Needs a FALLTHROUGH here, I think?

src/gfx_func.h
```diff
@@ -139,7 +139,7 @@ bool FillDrawPixelInfo(DrawPixelInfo *n, int left, int top, int width, int heigh
  */
 static inline int CenterBounds(int min, int max, int size)
 {
-	return min + (max - min - size + 1) / 2;
+	return (min + max - size + 1) / 2;
```
rubidium42, line 178: Technically this is less overflow safe; imagine `min` to be 2^30, max 2^30 + 100 and size 50. The old code would return 2^30 + 25, the new code would return -2^30 + 26. However, the old code is not guaranteed to always return the right value, see e.g. https://www.youtube.com/watch?v=sBtAGxBh-XI for the mess you can get with getting the midpoint of two values.
An even simpler version would be `std::midpoint(min, max - size + 1)` (with C++20).
Having said that, I guess it's outrageous to have such large coordinates to draw with, so I guess in the end it's not a big issue.

src/core/geometry_type.hpp
```diff
@@ -49,6 +49,121 @@ struct Rect {
...
+			? Rect {this->left,          this->top, this->right - indent, this->bottom}
+			: Rect {this->left + indent, this->top, this->right,          this->bottom};
+	}
+
+	/**
+	 * Copy Rect and set its height.
+	 * @param width height in pixels for new Rect.
+	 * @return the new resized Rect.
+	 */
+	inline Rect SetHeight(int height) const
```
rubidium42: Can't you just do `SetHeight(int height, bool end = false)` in the next function and omit this?

src/core/geometry_type.hpp
```diff
@@ -49,6 +49,121 @@ struct Rect {
...
+		return {this->left + x, this->top + y, this->right + x, this->bottom + y};
+	}
+
+	/**
+	 * Copy Rect and set its width.
+	 * @param width width in pixels for new Rect.
+	 * @param end   if set, set width at end of Rect, i.e. on right.
+	 * @return the new resized Rect.
+	 */
+	inline Rect SetWidth(int width, bool end) const
```
rubidium42: These setter names are quite misleading. Maybe use something like `WithWidth`, so you would get `rect.WithWidth(10, false)`?

src/core/geometry_type.hpp
```diff
@@ -49,6 +49,121 @@ struct Rect {
...
+		return {this->left, this->top, this->right, this->top + height - 1};
+	}
+
+	/**
+	 * Copy Rect and set its height.
+	 * @param width height in pixels for new Rect.
+	 * @param end   if set, set height at end of Rect, i.e. at bottom.
+	 * @return the new resized Rect.
+	 */
+	inline Rect SetHeight(int height, bool end) const
```
rubidium42: `WithHeight` to not mislead people?

src/widgets/slider.cpp
```diff
@@ -23,12 +23,16 @@ static const int SLIDER_WIDTH = 3;
  * @param r     Rectangle to draw the widget in
  * @param value Value to put the slider at
  */
-void DrawSliderWidget(Rect r, int min_value, int max_value, int value)
+void DrawSliderWidget(Rect r, int min_value, int max_value, int value, std::map<int, StringID> labels)
```
rubidium42: `const std::map<int, StringID> &labels` to prevent copying the map for each draw?

src/widgets/slider.cpp
```diff
@@ -42,10 +46,23 @@ void DrawSliderWidget(Rect r, int min_value, int max_value, int value)
 	GfxDrawLine(wedge[1].x, wedge[1].y, wedge[2].x, wedge[2].y, _current_text_dir == TD_RTL ? shadow : light, t);
 	GfxDrawLine(wedge[0].x, wedge[0].y, wedge[1].x, wedge[1].y, shadow, t);
 
+	int x;
+	for (auto label : labels)
+	{
```
rubidium42: The `{` should be on the line before.

src/core/geometry_type.hpp
PeterN: WithWidth/Height works for me, also added [[nodiscard]] hint just in case.

stormcone:
Is it intentional that the labels in the scaling slider don't scale?

scaling

PeterN:
> Is it intentional that the labels in the scaling slider don't scale?

They do scale, but currently the sprite font (along with all sprites) can only be 1x, 2x or 4x. Truetype fonts will be scaled appropriately. 3x and 5x sprite font scaling can be possible, but non-integer scaling of it looks pretty bad.

LC-Zorg:
I like it. Especially for the possibility of x1,25 and x1,5 zoom. :) These values seem most appropriate for FHD and QHD resolutions.

The problem with high magnifications is that after changing the setting, if it turns out to be too high, returning to the previous setting can be troublesome or even impossible. This problem also occurs today. Therefore, maybe on the occasion of this change, it would also be worth introducing an appropriate limitation? I think even a simple equation could solve the problem effectively:
horizontal resolution / 1000 (maybe 800?) = maximum available multiplier. For FHD 1920x1080, the limit would be x1.92 that is x1.75 if /1000 or x2.4 that is x2.25 if /800. Selecting a higher value would automatically jump to the available maximum value.

reldred:
Just had a quick play with it, working very nicely now. Didn't know just how much I was missing chunky bevels in my life. 2x scaling now feels like I'm literally just using integer scaling on my graphics card except for the viewports. Very well done.

I'll have to try 3x zoom on my macbook (I tested on a 3440x1440 desktop panel), but it looked very clean on the desktop at least.

LordAro:
Personally, I vote we merge it asap and fix any remaining issues before the final release - the more eyes the better with something like this

PeterN:
Hmm, should I remove the "Chunky Bevels" option, or finish it off (save it to settings)?

(Perhaps reword it if keeping it...)

LordAro:
I like having the option - I can definitely see people preferring the "cleaner" look at higher zoom levels. How much code is it to keep it?

PeterN:
Kinda barely any. It's just the GUI code and a single if-condition (and settings when it's added.) There's no extra conditions anywhere else.

Behind the scenes there is `bevel` and `fullbevel` dimensions, the first is only used for drawing the bevels, the latter is used when working out padding to keep everything consistent. That could be removed if the option is removed, but it's probably more work to do that than not...

PeterN:
I'll add it to saved settings, can always be removed later. And I'm going to rename/reword some of the strings. Coming soon :-)

PeterN:
Updated to save the "chunky bevels" setting, along with changing some of the strings around (to avoid translation conflicts)

src/gfx_func.h
PeterN: Yeah, I think at the point we are hitting those possibilities it's well outside the range of being displayable. Though this commit itself is not particularly relevant to the rest.

src/company_gui.cpp
```diff
@@ -159,15 +155,14 @@ static uint GetMaxCategoriesWidth()
...
-	int offs_right = _current_text_dir == TD_LTR ? 0 : EXP_INDENT;
+	const Rect tr = r.Indent(WidgetDimensions::scaled.hsep_indent, _current_text_dir == TD_RTL);
 
 	int y = start_y;
 	ExpensesType et;
 
 	for (uint i = 0; i < list.length; i++) {
 		et = list.et[i];
-		DrawString(r.left + offs_left, r.right - offs_right, y, STR_FINANCES_SECTION_CONSTRUCTION + et);
+		DrawString(tr.left, tr.right, y, STR_FINANCES_SECTION_CONSTRUCTION + et);
```
LordAro: Possible future work to change DrawString to take a Rect as well, rather than the left/right/top param?

src/company_gui.cpp
```diff
@@ -1269,81 +1262,81 @@ static const NWidgetPart _nested_select_company_manager_face_widgets[] = {
...
-							NWidget(WWT_TEXT, INVALID_COLOUR, WID_SCMF_LIPS_MOUSTACHE_TEXT), SetFill(1, 0), SetPadding(0, WD_FRAMERECT_RIGHT, 0, WD_FRAMERECT_LEFT),
+							NWidget(WWT_TEXT, INVALID_COLOUR, WID_SCMF_LIPS_MOUSTACHE_TEXT), SetFill(1, 0), SetPadding(WidgetDimensions::unscaled.framerect),
 								SetDataTip(STR_FACE_MOUSTACHE, STR_NULL), SetTextColour(TC_GOLD), SetAlignment(SA_VERT_CENTER | SA_RIGHT),
 							NWidget(WWT_PUSHARROWBTN, COLOUR_GREY, WID_SCMF_LIPS_MOUSTACHE_L), SetDataTip(AWV_DECREASE, STR_FACE_LIPS_MOUSTACHE_TOOLTIP),
 							NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SCMF_LIPS_MOUSTACHE), SetDataTip(STR_WHITE_STRING, STR_FACE_LIPS_MOUSTACHE_TOOLTIP),
 							NWidget(WWT_PUSHARROWBTN, COLOUR_GREY, WID_SCMF_LIPS_MOUSTACHE_R), SetDataTip(AWV_INCREASE, STR_FACE_LIPS_MOUSTACHE_TOOLTIP),
 						EndContainer(),
 						NWidget(NWID_HORIZONTAL),
-							NWidget(WWT_TEXT, INVALID_COLOUR, WID_SCMF_CHIN_TEXT), SetFill(1, 0), SetPadding(0, WD_FRAMERECT_RIGHT, 0, WD_FRAMERECT_LEFT),
+							NWidget(WWT_TEXT, INVALID_COLOUR, WID_SCMF_CHIN_TEXT), SetFill(1, 0), SetPadding(WidgetDimensions::unscaled.framerect),
```
LordAro, line 1318: How bad/hard would it be to introduce some defaults, like `SetPadding(WidgetDimensions::unscaled.framerect),` ?

src/company_gui.cpp
PeterN: It does, but in this case y/start_y is provided separately. Although `tr.top = start_y;` would work, and then just increment tr.top instead of y.

src/company_gui.cpp
PeterN: Most widgets already have built in padding, but WWT_TEXT doesn't. But in most cases, WWT_TEXT is used without extra padding anyway.

src/dock_gui.cpp
```diff
@@ -531,25 +531,34 @@ struct BuildDocksDepotWindow : public PickerWindowBase {
...
+		DrawPixelInfo tmp_dpi;
 
-		DrawShipDepotSprite(this->GetWidget<NWidgetBase>(WID_BDD_X)->pos_x + x1, this->GetWidget<NWidgetBase>(WID_BDD_X)->pos_y + y1, AXIS_X, DEPOT_PART_NORTH);
-		DrawShipDepotSprite(this->GetWidget<NWidgetBase>(WID_BDD_X)->pos_x + x2, this->GetWidget<NWidgetBase>(WID_BDD_X)->pos_y + y2, AXIS_X, DEPOT_PART_SOUTH);
-		DrawShipDepotSprite(this->GetWidget<NWidgetBase>(WID_BDD_Y)->pos_x + x2, this->GetWidget<NWidgetBase>(WID_BDD_Y)->pos_y + y1, AXIS_Y, DEPOT_PART_NORTH);
-		DrawShipDepotSprite(this->GetWidget<NWidgetBase>(WID_BDD_Y)->pos_x + x1, this->GetWidget<NWidgetBase>(WID_BDD_Y)->pos_y + y2, AXIS_Y, DEPOT_PART_SOUTH);
+		switch (widget) {
+			case WID_BDD_X:
+			case WID_BDD_Y:
+				Axis axis = widget == WID_BDD_X ? AXIS_X : AXIS_Y;
```
LordAro: Variable inside case statement - should be surrounded by braces (it doesn't *strictly* matter because it's the last case statement, but is mildly more correct)

src/industry_gui.cpp
```diff
@@ -979,11 +962,11 @@ class IndustryViewWindow : public Window
...
 		if (!i->text.empty()) {
 			SetDParamStr(0, i->text);
-			y += WD_PAR_VSEP_WIDE;
-			y = DrawStringMultiLine(left + WD_FRAMERECT_LEFT, right - WD_FRAMERECT_RIGHT, y, UINT16_MAX, STR_JUST_RAW_STRING, TC_BLACK);
+			ir.top += WidgetDimensions::scaled.vsep_wide;
+			ir.top = DrawStringMultiLine(ir.left, ir.right, ir.top, UINT16_MAX, STR_JUST_RAW_STRING, TC_BLACK);
 		}
 
-		return y + WD_FRAMERECT_BOTTOM;
+		return ir.top - 1 + WidgetDimensions::scaled.framerect.bottom;
```
LordAro, line 970: why -1 ?

src/lang/english.txt
```diff
@@ -1008,21 +1008,19 @@ STR_GAME_OPTIONS_VIDEO_VSYNC_TOOLTIP                            :{BLACK}Check th
...
+STR_GAME_OPTIONS_GUI_SCALE_AUTO_TOOLTIP                         :{BLACK}Check this box to detect interface size automatically
+
+STR_GAME_OPTIONS_GUI_SCALE_BEVELS                               :{BLACK}Scale bevels
+STR_GAME_OPTIONS_GUI_SCALE_BEVELS_TOOLTIP                       :{BLACK}Check this box to scale bevels by interface size
+
+STR_GAME_OPTIONS_GUI_SCALE_1X                                   :{TINY_FONT}1x
+STR_GAME_OPTIONS_GUI_SCALE_2X                                   :{TINY_FONT}2x
+STR_GAME_OPTIONS_GUI_SCALE_3X                                   :{TINY_FONT}3x
+STR_GAME_OPTIONS_GUI_SCALE_4X                                   :{TINY_FONT}4x
+STR_GAME_OPTIONS_GUI_SCALE_5X                                   :{TINY_FONT}5x
```
LordAro: Why put TINY_FONT in the language file? We're not expecting translations to use something else, surely?

src/misc_gui.cpp
```diff
@@ -798,39 +796,36 @@ void QueryString::DrawEditBox(const Window *w, int wid) const
...
+	if (!FillDrawPixelInfo(&dpi, fr.left, fr.top, fr.Width(), fr.Height())) return;
 
 	DrawPixelInfo *old_dpi = _cur_dpi;
 	_cur_dpi = &dpi;
 
 	/* We will take the current widget length as maximum width, with a small
 	 * space reserved at the end for the caret to show */
 	const Textbuf *tb = &this->text;
-	int delta = std::min(0, (right - left) - tb->pixels - 10);
+	int delta = std::min(0, (fr.right - fr.left) - tb->pixels - 10);
```
LordAro, line 823: fr.Width() ?

src/misc_gui.cpp
```diff
@@ -856,17 +851,16 @@ Point QueryString::GetCaretPosition(const Window *w, int wid) const
...
+	int clearbtn_width = sprite_size.width + WidgetDimensions::scaled.imgbtn.Horizontal();
 
-	int left   = wi->pos_x + (rtl ? clearbtn_width : 0);
-	int right  = wi->pos_x + (rtl ? wi->current_x : wi->current_x - clearbtn_width) - 1;
+	Rect r = wi->GetCurrentRect().Indent(clearbtn_width, !rtl).Shrink(WidgetDimensions::scaled.framerect);
 
 	/* Clamp caret position to be inside out current width. */
 	const Textbuf *tb = &this->text;
-	int delta = std::min(0, (right - left) - tb->pixels - 10);
+	int delta = std::min(0, (r.right - r.left) - tb->pixels - 10);
```
LordAro, line 860: r.Width() ?

src/network/network_chat_gui.cpp
```diff
@@ -106,7 +107,7 @@ void CDECL NetworkAddChatMessage(TextColour colour, uint duration, const std::st
 void NetworkReInitChatBoxSize()
 {
 	_chatmsg_box.y       = 3 * FONT_HEIGHT_NORMAL;
-	_chatmsg_box.height  = MAX_CHAT_MESSAGES * (FONT_HEIGHT_NORMAL + NETWORK_CHAT_LINE_SPACING) + 4;
+	_chatmsg_box.height  = MAX_CHAT_MESSAGES * (FONT_HEIGHT_NORMAL + ScaleGUITrad(NETWORK_CHAT_LINE_SPACING)) + ScaleGUITrad(4);
```
LordAro, line 110: 4, 10 & 3 in this file could be replaced with constants?

src/newgrf_gui.cpp
```diff
@@ -833,21 +834,22 @@ struct NewGRFWindow : public Window, NewGRFScanCallback {
...
 
 				bool rtl = _current_text_dir == TD_RTL;
-				uint text_left    = rtl ? r.left + WD_FRAMERECT_LEFT : r.left + square.width + 15;
-				uint text_right   = rtl ? r.right - square.width - 15 : r.right - WD_FRAMERECT_RIGHT;
-				uint square_left  = rtl ? r.right - square.width - 5 : r.left + 5;
-				uint warning_left = rtl ? r.right - square.width - warning.width - 10 : r.left + square.width + 10;
+				uint text_left    = rtl ? tr.left : tr.left + square.width + 13;
+				uint text_right   = rtl ? tr.right - square.width - 13 : tr.right;
+				uint square_left  = rtl ? tr.right - square.width - 3 : tr.left + 3;
+				uint warning_left = rtl ? tr.right - square.width - warning.width - 8 : tr.left + square.width + 8;
```
LordAro, line 852: scary arbitrary numbers

src/rail_gui.cpp
```diff
@@ -1254,11 +1254,11 @@ struct BuildRailStationWindow : public PickerWindowBase {
...
 			case WID_BRAS_PLATFORM_DIR_X:
 				/* Set up a clipping area for the '/' station preview */
-				if (FillDrawPixelInfo(&tmp_dpi, r.left, r.top, r.right - r.left + 1, r.bottom - r.top + 1)) {
+				if (FillDrawPixelInfo(&tmp_dpi, r.left, r.top, r.Width(), r.Height())) {
 					DrawPixelInfo *old_dpi = _cur_dpi;
 					_cur_dpi = &tmp_dpi;
-					int x = ScaleGUITrad(31) + 1;
-					int y = r.bottom - r.top - ScaleGUITrad(31);
+					int x = (r.Width()  - ScaleSpriteTrad(64)) / 2 + ScaleSpriteTrad(31);
+					int y = (r.Height() + ScaleSpriteTrad(58)) / 2 - ScaleSpriteTrad(31);
```
LordAro, line 1261: 64/58 scalings used in a few places. Pull it out into a variable?

src/rail_gui.cpp
```diff
@@ -1673,15 +1673,14 @@ struct BuildSignalWindow : public PickerWindowBase {
...
 		Dimension sprite_size = GetSpriteSize(image, &offset);
-		const NWidgetBase *widget = this->GetWidget<NWidgetBase>(widget_index);
-		int x = widget->pos_x - offset.x +
-				(widget->current_x - sprite_size.width + offset.x) / 2;  // centered
-		int y = widget->pos_y - sig_sprite_bottom_offset + WD_IMGBTN_TOP +
-				(widget->current_y - WD_IMGBTN_TOP - WD_IMGBTN_BOTTOM + sig_sprite_size.height) / 2; // aligned to bottom
+		Rect ir = r.Shrink(WidgetDimensions::scaled.imgbtn);
+		int x = CenterBounds(ir.left, ir.right, sprite_size.width - offset.x); // centered
+		int y = ir.top - sig_sprite_bottom_offset +
+				(ir.Height() + sig_sprite_size.height) / 2; // aligned to bottom
```
LordAro, line 1683: arbitrary line break

src/road_gui.cpp
```diff
@@ -974,15 +974,23 @@ struct BuildRoadDepotWindow : public PickerWindowBase {
...
 	{
 		if (!IsInsideMM(widget, WID_BROD_DEPOT_NE, WID_BROD_DEPOT_NW + 1)) return;
 
-		DrawRoadDepotSprite(r.left + 1 + ScaleGUITrad(31), r.bottom - ScaleGUITrad(31), (DiagDirection)(widget - WID_BROD_DEPOT_NE + DIAGDIR_NE), _cur_roadtype);
+		DrawPixelInfo tmp_dpi;
+		if (FillDrawPixelInfo(&tmp_dpi, r.left, r.top, r.Width(), r.Height())) {
+			DrawPixelInfo *old_dpi = _cur_dpi;
+			_cur_dpi = &tmp_dpi;
+			int x = (r.Width()  - ScaleSpriteTrad(64)) / 2 + ScaleSpriteTrad(31);
+			int y = (r.Height() + ScaleSpriteTrad(48)) / 2 - ScaleSpriteTrad(31);
```
LordAro, line 990: 48? everything else uses 58

src/roadveh_cmd.cpp
```diff
@@ -94,10 +94,10 @@ int RoadVehicle::GetDisplayImageWidth(Point *offset) const
 	int reference_width = ROADVEHINFO_DEFAULT_VEHICLE_WIDTH;
 
 	if (offset != nullptr) {
-		offset->x = ScaleGUITrad(reference_width) / 2;
+		offset->x = ScaleSpriteTrad(reference_width) / 2;
 		offset->y = 0;
 	}
-	return ScaleGUITrad(this->gcache.cached_veh_length * reference_width / VEHICLE_LENGTH);
+	return ScaleSpriteTrad(this->gcache.cached_veh_length * reference_width / VEHICLE_LENGTH);
 }
```
LordAro, line 101: these functions seem gui-related, perhaps move to roadveh_gui.cpp ?

src/settings_gui.cpp
```diff
@@ -475,10 +465,45 @@ struct GameOptionsWindow : Window {
...
+				if (ClickSliderWidget(this->GetWidget<NWidgetBase>(widget)->GetCurrentRect(), pt, MIN_INTERFACE_SCALE, MAX_INTERFACE_SCALE, this->gui_scale)) {
+					if (!_ctrl_pressed) this->gui_scale = ((this->gui_scale + 12) / 25) * 25;
+					this->SetWidgetDirty(widget);
+				}
+
+				if (click_count > 0) this->mouse_capture_widget = widget;
+				break;
+
+			case WID_GO_GUI_SCALE_AUTO:
+			{
```
LordAro, line 489: unnecessary brace? (and on the wrong line) (also unnecessary braces for WID_GO_GUI_SCALE_BEVEL_BUTTON)

src/settings_gui.cpp
```diff
@@ -57,22 +57,6 @@ static const StringID _autosave_dropdown[] = {
...
-
-static const StringID _font_zoom_dropdown[] = {
-	STR_GAME_OPTIONS_FONT_ZOOM_DROPDOWN_AUTO,
-	STR_GAME_OPTIONS_FONT_ZOOM_DROPDOWN_NORMAL,
-	STR_GAME_OPTIONS_FONT_ZOOM_DROPDOWN_2X_ZOOM,
-	STR_GAME_OPTIONS_FONT_ZOOM_DROPDOWN_4X_ZOOM,
-	INVALID_STRING_ID,
-};
-
 static Dimension _circle_size; ///< Dimension of the circle +/- icon. This is here as not all users are within the class of the settings window.
```
LordAro, line 60: does this still need to be global?

src/station_gui.cpp
```diff
@@ -809,7 +811,7 @@ static const NWidgetPart _nested_station_view_widgets[] = {
  */
 static void DrawCargoIcons(CargoID i, uint waiting, int left, int right, int y)
 {
-	int width = ScaleGUITrad(10);
+	int width = ScaleSpriteTrad(10);
```
LordAro, line 814: 10 !

src/textfile_gui.cpp
```diff
@@ -148,17 +148,15 @@ void TextfileWindow::SetupScrollbars(bool force_reflow)
...
 
-	const int x = r.left + WD_FRAMETEXT_LEFT;
-	const int y = r.top + WD_FRAMETEXT_TOP;
-	const int right = r.right - WD_FRAMETEXT_RIGHT;
-	const int bottom = r.bottom - WD_FRAMETEXT_BOTTOM;
+	Rect fr = r.Shrink(WidgetDimensions::scaled.frametext);
 
 	DrawPixelInfo new_dpi;
-	if (!FillDrawPixelInfo(&new_dpi, x, y, right - x + 1, bottom - y + 1)) return;
+	if (!FillDrawPixelInfo(&new_dpi, fr.left, fr.top, fr.Width(), fr.Height())) return;
```
LordAro, line 154: FillDrawPixelInfo should take a Rect param?

src/town_gui.cpp
```diff
@@ -196,28 +205,33 @@ struct TownAuthorityWindow : Window {
...
 				}
 				break;
-			case WID_TA_COMMAND_LIST: {
+
+			case WID_TA_RATING_INFO:
+				this->DrawRatings(r);
+				break;
+
+			case WID_TA_COMMAND_LIST:
+			{
```
LordAro: bad linebreak

src/town_gui.cpp
```diff
@@ -420,26 +434,30 @@ struct TownViewWindow : Window {
 				SetDParam(2, cargo->Index());
 				SetDParam(3, this->town->goal[i]);
 			}
-			DrawString(cargo_text_left, cargo_text_right, y += FONT_HEIGHT_NORMAL, string);
+			DrawString(tr.Indent(20, rtl), string);
```
LordAro, line 443: 20?

src/tree_gui.cpp
```diff
@@ -58,8 +58,8 @@ static Dimension GetMaxTreeSpriteSize()
 	Dimension size, this_size;
 	Point offset;
 	/* Avoid to use it uninitialized */
-	size.width = 32; // default width - WD_FRAMERECT_LEFT
-	size.height = 39; // default height - BUTTON_BOTTOM_OFFSET
+	size.width = ScaleGUITrad(32); // default width - WD_FRAMERECT_LEFT
+	size.height = ScaleGUITrad(39); // default height - BUTTON_BOTTOM_OFFSET
```
LordAro, line 62: these constants don't exist anymore

src/vehicle_gui.cpp
```diff
@@ -3355,15 +3344,15 @@ void SetMouseCursorVehicle(const Vehicle *v, EngineImageType image_type)
...
+		if (total_width >= ScaleSpriteTrad(2 * (int)VEHICLEINFO_FULL_VEHICLE_WIDTH)) break;
 
 		PaletteID pal = (v->vehstatus & VS_CRASHED) ? PALETTE_CRASH : GetVehiclePalette(v);
 		VehicleSpriteSeq seq;
 
 		if (rotor_seq) {
 			GetCustomRotorSprite(Aircraft::From(v), image_type, &seq);
 			if (!seq.IsValid()) seq.Set(SPR_ROTOR_STOPPED);
-			y_offset = - ScaleGUITrad(5);
+			y_offset = - ScaleSpriteTrad(5);
```
LordAro: space between `-` and number (function call) is a bit odd (i know, it was already there)

src/widget.cpp
```diff
@@ -24,6 +24,208 @@
...
+Dimension GetScaledSpriteSize(SpriteID sprid)
+{
+	Point offset;
+	Dimension d = GetSpriteSize(sprid, &offset, ZOOM_LVL_OUT_4X);
+	d.width  -= offset.x;
+	d.height -= offset.y;
+	return ScaleGUITrad(d);
+}
+
+void SetupWidgetDimensions()
```
LordAro, line 199: doccomments? (+ several others)

src/widget.cpp
```diff
@@ -393,11 +605,17 @@ static inline void DrawVerticalScrollbar(const Rect &r, Colours colour, bool up_
 	GfxFillRect(r.left, r.top + height, r.right, r.bottom - height, c2);
 	GfxFillRect(r.left, r.top + height, r.right, r.bottom - height, c1, FILLRECT_CHECKER);
 
+	/* track positions */
+	int left  = r.left + r.Width() * 3 / 11;
+	int right = r.left + r.Width() * 8 / 11;
```
LordAro: arbitrary numbers? (that are repeated in DrawHorizontalScrollbar...)

src/widget_type.h
```diff
@@ -920,8 +923,7 @@ struct NWidgetPartWidget {
  * Widget part for storing padding.
  * @ingroup NestedWidgetParts
  */
-struct NWidgetPartPaddings {
-	uint8 top, right, bottom, left; ///< Paddings for all directions.
+struct NWidgetPartPaddings : RectPadding {
 };
```
LordAro, line 927: Any particular reason for leaving it as a distinct struct, rather than making it a typedef?

src/zoom_func.h
```diff
@@ -99,34 +99,24 @@ static inline ZoomLevel UnScaleZoomGUI(ZoomLevel value)
 	return std::clamp(ZoomLevel(value - (ZOOM_LVL_GUI - ZOOM_LVL_OUT_4X)), ZOOM_LVL_MIN, ZOOM_LVL_MAX);
 }
 
-/**
- * Scale traditional pixel dimensions to GUI zoom level.
+/*
```
LordAro: doxyment removal?

LordAro (commented):
Just idly scrolled through the whole diff. These are the thoughts I had along the way. Feel free to ignore any/all of it

src/lang/english.txt
PeterN: No but we want the string to be tiny. I suppose I was thinking a bit to generically with the list of strings, but given the use-case I guess I could just specify FS_SMALL when drawing.

src/misc_gui.cpp
PeterN: Not sure, Width() adds 1, it's not the same as `right - left`. Maybe the existing code is obiwan though.

src/industry_gui.cpp
PeterN: Because otherwise it would be 1 pixel too large ;-)

(It returns the required 'bottom' position, rather than height)

src/lang/english.txt
PeterN: Well, it turns out `GetStringBoundingBox` with a StringID doesn't accept a font-size parameter, only with a raw string.

Arastais:
Out of curiosity, would this PR fix #8998 and/or #9009

Arastais:
Out of curiosity, would this PR fix #8998 and/or #9009 l

PeterN:
> Out of curiosity, would this PR fix issues like #8998 and/or #9009 in any way?

No, they would need to be separately addressed still.

LordAro (approved):
#Shipit

PR #10119, opened 2022-11-01
Change: Cycle through visible signals by default

### LINKED ISSUE #10118: [Bug]: «Сycle trough signal types» set to «Path signals only» by default
### Version of OpenTTD

OpenTTD 12.2 and newer

### Expected result

### Actual result

### Steps to reproduce

2. Open Railway signals menu

2TallTyler:
## Motivation / Problem

#10118 

The signal GUI by default shows only path signals. There's a separate setting to choose whether Ctrl-click cycles through path signals or all visible signals. When we only see path signals, this second setting does nothing. However, if a player expands the signal GUI to use more signals, they likely want to cycle through all visible signals (as was the default before #8688). Right now the default is path signals only, so they'd need to go into Settings and change that.

## Description

Changing the default only affects people who use block/presignals and don't already have an OpenTTD config file, and we can close #10118. 🙂 

## Limitations

The issue also requests an option to cycle through block signals only. That isn't covered in this PR.

## Checklist for review


frosch123:
I don't understand the purpose of this setting. Is there some history with it?

I would understand a setting with values:
* one cycle through all (visible) signals
* two cycles: (oneway-pbs, twoway-pbs) and (normal block, pre, exit, combo)

But a setting that converts a non-path signal into a path signal and not back?
I think we should rather remove this setting :)

2TallTyler:
I agree. The function of this setting should be changed to two options:
1. Cycle through all visible signals
2. Cycle through this type of signal (block/pre or path)

frosch123:
I wondered how often you actually use this:
* You only cycle signals when building them. One click for a block signal, 2 for a presignal, 3 for exit, ...
* The wrap-around is only relevant if you misclick and click too often.

I think a setting for this rare "misclick" thingie is a bit too much. So I think just removing this setting is quite valid.

2TallTyler:
I don't disagree for how I play the game, but I've learned not to assume that my way is best.

PR #10123, opened 2022-11-02
Feature: Display power-to-weight ratio in ground vehicle details GUI

M3Henry:
## Motivation / Problem

When building a train, deciding how long the train or how many engines to have requires considering the power to weight ratio when the train is loaded, however the ground vehicle details GUI only displays the vehicles current weight and not the fully loaded weight, and the default wagons vary their loaded weights fairly widely.

## Description

Display the loaded power-to-weight ratio of the ground vehicle in the details tab. The units are selected using the user's power and weights setting.

Screenshot from 2022-11-02 08-12-47

## Limitations

Alternative: Display maximum weight in brackets next to the current weight in the vehicle details GUI.

> Weight 370 / 550 t

or

> Weight 370 (550) t

## Checklist for review


src/vehicle.cpp
```diff
@@ -2987,3 +2987,27 @@ void GetVehicleSet(VehicleSet &set, Vehicle *v, uint8 num_vehicles)
...
+	{
+		const Engine *e = u->GetEngine();
+		max_weight += e->GetDisplayWeight();
+		if (!e->CanCarryCargo()) continue;
+
+		auto cargo = u->cargo_type;
+		uint32 cargo_weight = CargoSpec::Get(cargo)->weight;
+		cargo_weight *= FreightWagonMult(cargo);
+		cargo_weight *= e->DetermineCapacity(u);
+		max_weight += cargo_weight / 16;
```
2TallTyler: Some clarification about the magic number 16 would be good here.

src/strings.cpp
```diff
@@ -1252,6 +1265,19 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 				break;
 			}
 
+			case SCC_POWER_TO_WEIGHT: { // {POWER_TO_WEIGHT}
+				auto setting = _settings_game.locale.units_power * 3u + _settings_game.locale.units_weight;
+				assert(setting < lengthof(_units_power_to_weight));
+
+				auto const& x = _units_power_to_weight[setting];
```
2TallTyler:
```suggestion
				auto const &x = _units_power_to_weight[setting];
```

src/vehicle.cpp
```diff
@@ -2987,3 +2987,27 @@ void GetVehicleSet(VehicleSet &set, Vehicle *v, uint8 num_vehicles)
...
+	}
+
+	if (max_weight == 0) return 0;
+	return GetGroundVehicleCache()->cached_power * 10u / max_weight;
```
2TallTyler, line 3014: Doxygen header says `@return power-to-weight ratio in 10ths of hp(I) per tonne` — needs updating? (This appears to return in hp per tonne)

2TallTyler (commented):
The vehicle buy menu lists maximum weight in parenthesis with the units included twice: `Weight: 13 t (17 t)`. I think this should be included in the vehicle details window, where the first number is the current weight (not the empty weight as in the buy menu).

For trains it makes sense to only have the power-to-weight ratio in the details GUI, but for road vehicles I wonder if it makes sense to include it in the buy menu, since RVs cannot be composed of multiple units. The use case I have in mind is for NewGRFs like Mop's Expanded Road Vehicles which include many choices, each with a different power-to-weight ratio for a specific purpose. You wouldn't want to send a heavy and weak road train on a mountainous route and might want to see that an electrified truck has a higher power-to-weight ratio.

Other opinions welcome, since that does add clutter to the buy menu.

src/vehicle.cpp
M3Henry: This is correct, the game uses hp(I) rather than hp(M) internally.

src/vehicle_base.h
```diff
@@ -234,7 +234,7 @@ struct Vehicle : VehiclePool::PoolItem<&_vehicle_pool>, BaseVehicle, BaseConsist
 	Vehicle *previous_shared;           ///< NOSAVE: pointer to the previous vehicle in the shared order chain
 
 public:
-	friend void FixOldVehicles();
```
2TallTyler: I'm sure you didn't mean to reformat this 😃

src/vehicle.cpp
```diff
@@ -2987,3 +2987,29 @@ void GetVehicleSet(VehicleSet &set, Vehicle *v, uint8 num_vehicles)
...
+
+	for (const Vehicle* u = this; u != nullptr; u = u->Next()) {
+		max_weight += u->GetMaxWeight();
+	}
+
+	return max_weight;
+}
+
+/**
+ * Calculates the minimum power-to-weight ratio usinging the maximum weight of the ground vehicle
```
2TallTyler:
```suggestion
 * Calculates the minimum power-to-weight ratio using the maximum weight of the ground vehicle
```

2TallTyler (dismissed):

2TallTyler (dismissed):
Code looks good and I would use this feature.

2TallTyler (approved):

M3Henry:
I swear there was a button to merge last time. :confused:

PR #10124, opened 2022-11-02
Change: Don't show vehicle reliability when breakdowns disabled

2TallTyler:
## Motivation / Problem

The build vehicle GUI and vehicle details GUI are full of useful information, but they're also...full. 

It would be nice to hide useless information, like reliability and breakdown information when breakdowns are disabled.

## Description

reliability

Hide these lines of text and make the GUIs smaller when breakdowns are disabled. If the setting is changed while a GUI is open, it will grow to fit the content.

## Limitations

If you repeatedly enable and disable breakdowns while a train details window is open, the details window will lose lines until it's only three lines high. This does not break anything and I don't expect many people to do this, but I'm also not sure why it happens. 🙂 

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -928,10 +928,12 @@ int DrawVehiclePurchaseInfo(int left, int right, int y, EngineID engine_number, 
 		DrawString(left, right, y, STR_PURCHASE_INFO_DESIGNED_LIFE);
 		y += FONT_HEIGHT_NORMAL;
 
-		/* Reliability */
-		SetDParam(0, ToPercent16(e->reliability));
-		DrawString(left, right, y, STR_PURCHASE_INFO_RELIABILITY);
-		y += FONT_HEIGHT_NORMAL;
+		/* Reliability, if breakdowns are enabled */
+		if (_settings_game.difficulty.vehicle_breakdowns != 0) {
+			SetDParam(0, ToPercent16(e->reliability));
+			DrawString(left, right, y, STR_PURCHASE_INFO_RELIABILITY);
+			y += FONT_HEIGHT_NORMAL;
+		}
 	}
 
 	if (refittable) y = ShowRefitOptionsList(left, right, y, engine_number);
@@ -1115,7 +1117,8 @@ struct BuildVehicleWindow : Window {
 		widget->tool_tip    = STR_SHOW_HIDDEN_ENGINES_VEHICLE_TRAIN_TOOLTIP + type;
 		widget->SetLowered(this->show_hidden_engines);
 
-		this->details_height = ((this->vehicle_type == VEH_TRAIN) ? 10 : 9);
+		this->details_height = ((this->vehicle_type == VEH_TRAIN) ? 9 : 8);
+		if (_settings_game.difficulty.vehicle_breakdowns != 0) this->details_height++; // Only show a line for max reliability if breakdowns are enabled
 
 		this->FinishInitNested(tile == INVALID_TILE ? (int)type : (int)tile);
 
--- a/src/settings_table.cpp
+++ b/src/settings_table.cpp
@@ -247,6 +247,15 @@ static void RoadVehSlopeSteepnessChanged(int32 new_value)
 	}
 }
 
+/**
+ *Vehicle breakdown setting changed, redraw vehicle and buy menu GUIs to show / hide reliability info
+ * @param new_value Unused
+ */
+static void VehicleBreakdownsChanged(int32 new_value)
+{
+	ReInitAllWindows(false);
+}
+
 static void TownFoundingChanged(int32 new_value)
 {
 	if (_game_mode != GM_EDITOR && _settings_game.economy.found_town == TF_FORBIDDEN) {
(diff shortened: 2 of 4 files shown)


nielsmh:
One of the things the reliability in the Purchase window also indicates, is that a vehicle model is on the way to be phased out. You can tell that by seeing the max reliability dropping from the peak it has had (if you monitor it). That's still relevant information even when playing without breakdowns.

I wonder if it'd be worth adding a separate indication about that, somehow. Perhaps as a separate PR.

Arastais:
@nielsmh I personally agree with your second point; I think this PR should be merged but also a follow-up PR should be made (or maybe even add to this PR since it's relevant) to add a dedicated line showing key information on when a vehicle is phased out. I think this would be much clearer to players, as there would be no reliability info when breakdowns are turned off (so people don't get confused and maybe think breakdowns are not actually off), and the phase out date/info would show regardless.

Also in my opinion the limitation of losing lines is quite significant, and should be sorted out before merge. If a player just turns the setting on and off a couple of times throughout the game (which is not uncommon), it would remove one line, and I think removing even one line is problematic enough, and I can easily see it being reported as a bug in the future.

2TallTyler:
> If a player just turns the setting on and off a couple of times throughout the game (which is not uncommon), it would remove one line, and I think removing even one line is problematic enough, and I can easily see it being reported as a bug in the future.

Sorry, I wasn't clear about this. It only removes the lines for windows which remain open during the setting changes. New windows are always initialized with the correct number of lines. It's just something odd about reinitializing the same window repeatedly.

nielsmh:
Thought about showing "being phased out": Change the line with Designed and Year into being `Available: <startyear>-<endyear>` when there is less than 2 or 5 years until the vehicle stops being available to purchase. Perhaps make the end year red to highlight it. Keep it as showing just Designed year until near the EOL date.

Arastais:
>Sorry, I wasn't clear about this. It only removes the lines for windows which remain open during the setting changes. New windows are always initialized with the correct number of lines. It's just something odd about reinitializing the same window repeatedly.

@2TallTyler It's not that significant of a problem in that case 👏. My guess is that `ReInitAllWindows(bool)` cannot handle reinitializing the window all by itself specifically after the setting is changed. You could try invalidating the window data (e.g. `InvalidateWindowClassesData(...)`) or maybe even force closing the window (e.g. `CloseWindowByClass(...)`).
<br/>

> Thought about showing "being phased out": Change the line with Designed and Year into being `Available: <startyear>-<endyear>` when there is less than 2 or 5 years until the vehicle stops being available to purchase. Perhaps make the end year red to highlight it. Keep it as showing just Designed year until near the EOL date.

@nielsmh Honestly a great idea. Even better, you probably don't even need to change the designed/year when it's near the EOL date, it could _always_ be `Available: <startyear>-<endyear>` and just have that as the permanent text, since `<startyear>` is when it was designed.

James103:
> Even better, you probably don't even need to change the designed/year when it's near the EOL date - it could _always_ be `Available: <startyear>-<endyear>` and just have that as the permanent text, since `<startyear>` is when it was designed.

Would that change also apply (a) if breakdowns are on and (b) if "Vehicles never expire" is on?

Arastais:
> > Even better, you probably don't even need to change the designed/year when it's near the EOL date - it could _always_ be `Available: <startyear>-<endyear>` and just have that as the permanent text, since `<startyear>` is when it was designed.
> 
> Would that change also apply (a) if breakdowns are on and (b) if "Vehicles never expire" is on?

a. Yes, I would assume breakdowns only control reliability and breakdown information that isn't related to lifespan. If breakdowns are turned off, then the respective lines would not be shown, but the change would still apply.
b. Probably not. The easiest option I think would be to just show `Available: <startyear>`

nielsmh:
Window layouts becoming weird when changing options that cause re-layout is pretty much a staple at this point. It's difficult to make work reliably in my experience, and not worth spending much time on.

One thing regarding the displaying end-year for availability, is that it should only be shown when close to the end year. This is by the idea that a manufacturer usually doesn't know far ahead in time when they'll stop supporting a thing, but they might have a plan 5 years into the future on phasing out a model. The string needs a variation without ending year regardless, for games where "Vehicles never expire" is turned on.

LC-Zorg:
I don't think the vehicle window is the right place to show the phasing out information. Usually, if the breakdowns are turned off, it doesn't matter to the player until he wants to buy new vehicles. Then the right place would be the vehicle purchase window, but here also I don't see big sense in this, since he can see this fact just in this that the vehicle is unavailable.

But what about when breakdowns are turned off but servicing is still on? In this case, reliability drops. The question is, does it matter? Could the level of reliability have an impact on, for example, maintenance costs?

I admit that I have an idea here. It goes even a bit further than this PR provides.

In brief:
- For better readability, individual pieces of information are divided thematically
- All breakdowns related items are in one part
- If the breakdowns are disabled, the entire breakdown part, including the setting of service intervals, is hidden.
- Off topic: Easier access to information about cargo and vehicle by splitting information into two tabs instead of four
Improved vehicle info window

This suggestion also relates to #10123

2TallTyler:
The eventual solution to this problem is much bigger/different and I don't have plans to complete it anytime soon.

PR #10134, opened 2022-11-04
New industry creation/closure behavior

micler:
## Motivation / Problem

This branch brings two new options in the game: "Industry creation behavior" and "Industry Closure Behavior", to better manage industry creations (default behavior is always increasing the number of industries until the map is full), and industry closures (prevent unserviced industries from closing)

## Description

Subject was discussed here: https://www.tt-forums.net/viewtopic.php?t=90314
The branch adds a new "Industry creation" parameter, with four options: Original/Keep stable count/ Adapt to population / Funding Only. (the latter disappearing from industry density, which becomes a map generation option only)
The other parameter, "industry closure", allows the player to prevent an industry from closing if Unserviced or Linked to a town without player ratings.
Also, industries with an active subsidies won't close.

## Limitations

I have tested with several tests cases (new game, load old save, load scenario), no pitfall.
Untested with newgrf industries.

## Checklist for review

Save is impacted. New save version has been added, with compatibility in afterload.
Translation files are altered. I have modified english.txt and also done french for the new string I added. Since I have changed the FUNDING_ONLY choice, moving from industry_density to industry_creation, I have modified all language files to reflect the change.

### DIFF
--- a/src/genworld_gui.cpp
+++ b/src/genworld_gui.cpp
@@ -97,6 +97,7 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 					NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_QUANTITY_OF_SEA_LAKES, STR_NULL), SetFill(1, 1),
 					NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_NUMBER_OF_TOWNS, STR_NULL), SetFill(1, 1),
 					NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_NUMBER_OF_INDUSTRIES, STR_NULL), SetFill(1, 1),
+					NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_BEHAVIOR_OF_INDUSTRIES, STR_NULL), SetFill(1, 1),
 					NWidget(WWT_TEXT, COLOUR_ORANGE), SetDataTip(STR_MAPGEN_BORDER_TYPE, STR_NULL), SetFill(1, 1),
 				EndContainer(),
 				/* Widgets at the right of the labels. */
@@ -112,7 +113,8 @@ static const NWidgetPart _nested_generate_landscape_widgets[] = {
 					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_VARIETY_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
 					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_WATER_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
 					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_TOWN_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
-					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_INDUSTRY_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
+					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_INDUSTRY_DENSITY_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
+					NWidget(WWT_DROPDOWN, COLOUR_ORANGE, WID_GL_INDUSTRY_CREATION_PULLDOWN), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
 					NWidget(WWT_TEXTBTN, COLOUR_ORANGE, WID_GL_BORDERS_RANDOM), SetDataTip(STR_JUST_STRING, STR_NULL), SetFill(1, 0),
 				EndContainer(),
 			EndContainer(),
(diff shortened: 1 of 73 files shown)


frosch123:
There are multiple existing features in this area:
* Environment > Industires > Economy type: Frozen
* GameScript > NewGRF communication on production increase/decrease and closure. (GSIndustry::SetControlFlags, industry variable 47 'gs_disallows_*')
* Add-ons like OpenGFX+Industries, Manual Industries, ...

Please explain:
* How do these new settings interact and extend the existing methods?
* What is missing in the existing add-on framework, which would be covered by these settings?
* Why did you choose to add hard-coded behavior inside the core OpenTTD instead of extending the add-ons?

micler:
Hi,
I've answered below:


Le 05/11/2022 à 14:56, frosch a écrit :
>
> There are multiple existing features in this area:
>
>   * Environment > Industires > Economy type: Frozen
>
Doesn't answer the need at all.
>
>   * GameScript > NewGRF communication on production increase/decrease
>     and closure. (GSIndustry::SetControlFlags, industry variable 47
>     'gs_disallows_*')
>
Can't find this one ? what does it do ?
>
>   * Add-ons like OpenGFX+Industries, Manual Industries, ...
>
I wanted this feature to works whatever industry set we use.
>
> Please explain:
>
>   * How do these new settings interact and extend the existing methods?
>   * What is missing in the existing add-on framework, which would be
>     covered by these settings?
>   * Why did you choose to add hard-coded behavior inside the core
>     OpenTTD instead of extending the add-ons?
>
I'll start by answering the last question: seeing how industries are 
created , I didn't seem like a good idea to try to go around the 
behavior of the "core" of the game.
I don't believe that piling up extensions is a good answer when the 
implementation of the features is easily done in the code, especially 
when the modifications are "deep". (with multiple addons it's much 
harder to handle conflicts).
Honestly, I didn't have the time to test every script to see if it could 
answer the need. There a re far too many of them. I checked the two 
scripts relation to industry creation, and neither seem to answer the need.
Another point, more personnal: I used to run several addons before, I 
gave up and now I use an almost vanilla game. I dont think many people 
are using addons (especially game scripts), it can be used for niche 
features/ expert players, but I guess most users don't even bother with 
them. (that's one thing that bothering me with scenarios: having to 
install X addons to run it).

The proposed features should work with all extensions, as it changes the 
target count for each industry type (creation), and adds extra checks 
before allowing an industry to close.
The original way the game works is to steadily increase the amount of 
wanted industries on the map, meaning that until every single town has 
an industry (except oil), it will spawn new industries. It did work on 
the origin 256 square maps, but doesn't work anymore with large map, 
that's another reason why I modified the code: I believe the original 
way to go is obsolete. New options either freezes the number of 
industries (with 12% variations), or allows it to grow linked to the 
world population (also with 12% variation). It doesn't mean that 
industries are frozen, they can still shut down. Simply we won't spawn 
new industries if the amount already on the map is enough.

Note that the original behavior is still available for both options, the 
only thing becoming standard is not closing an industry with a subsidy. 
(which was to me a bug)

Zifox

Message ID: ***@***.***>

Thedfkhjzeofzeofhoze

LC-Zorg:
I think this is a change in the right direction, but I don't really like this path.

Referring to the original problem that you described on the forum, i.e. the desire to maintain a specific industry density (this is not ensured by any NewGRF, nor the freeze mode, the scripts in their current form are a very poor solution due to the possibility of choosing only one and many other problems), I think another method could be used, similar to the one used in the rest of the create new game window settings.

What I mean. The player could also choose one of the currently available options for determining the **initial industry density** (non, minimal, very low, low, normal, high), but moreover he would have the option of setting his own, more precise value (0-100?) As it is possible with the number of cities, terrain height or sea level. From my point of view, it would be good because very often the "low" setting often turns out to be very high, and the "minimum" setting is, in turn, too small. I see two variants here. In the first, the value would be in the range 0-100, where 0 is nothing and 100 is the maximum possible density. In the second variant, the number would be unlimited and would represent the literal total number of enterprises on the map. Both variants have their advantages and disadvantages, but I think the latter would be better as it would be used by players who rather know what they want and what to expect.

Additionally, a second setting would also be available here that would define the **maximal industry density during the game**. Thanks to this, the player could determine whether their number should increase, decrease or remain at the same level. Ideally, this setting could remain configurable during the game. Then, if the player changes them, the number of enterprises on the map would also change accordingly - sometimes it turns out that there are too few or too many enterprises, so it's good to be able to do so. This setting, which is currently available, although it can be changed, has only little effect on the current game and sometimes none. By default, this setting would be the same as the initial number (if A = Low, B = Low), although the actual value for the latter setting could be slightly larger. This setting could be hidden in the settings window in the industries tab. By the way, I think this current setting might also fit better in this tab.

Industry density - few changes

micler:
Hi,

I don't think that's much different from what I did, it fits in the two 
parameters I have in my proposal, whatever name we give them. Your 
*initial industry density *is the actual *industry density* with a an 
extra custom entry, and *maximal industry density during the game* is my 
new *industry creation*, only with more choices (1). (none = funding 
only, keep initial count, adapt to population(2), very low, low, normal 
high, custom) .

We could add a new custom value for initial industry creation, but we 
have to keep in mind that by default, the number of industries is 
limited by the number of towns (unless multiple industries becomes the 
standard). So the custom value could be a percentage of the number of 
towns, or a value that we would have to cap (without knowledge from the 
player that his request would not be met when creating a new map). Maybe 
a warning could be issued when clicking generate as a pre-build check.

I'd keep the options separated in "World generation" and 
"Environnement", as one is used in generation/editor, the other in game.

Contrary to your statement, the industry density option has no impact 
during the game, it's only used during game creation. (used in 
GetNumberOfIndustries, called from GenerateIndustries). As I discovered 
during my search, the amound of industries is increased regularly, 
scaled using map size. (see void IndustryBuildData::MonthlyLoop())

(1) see screenshot attached
(2) I'm very fond of adapting to the world population, it gives a boost 
in the late stages of the game when town population explodes. I've 
dropped the  "mininal" choice which does nothing, it's mapped to the 
same value as funding only in the code: 0.)

A shame we didn't have this discussion beforehand on the forum. :)

Zifox


Le 05/11/2022 à 17:42, LC-Zorg a écrit :

TrueBrain:
Although I really like the creativity of these ideas, from an OpenTTD perspective they are always a bit problematic when they are hardcoded. In the "old days" we used to do that, and what we ended up with that every other month or so someone would find another nice way to do town-growth, industries, cargo, etc. But maintaining that in a codebase like OpenTTD's became very impractical.

So, what we set out years ago, that we stop adding "yet another algorithm" to OpenTTD, and instead see if we can enable APIs to do the same thing. This is basically what @frosch123 was asking about: why can't we do this?

I really think we can have so much better industry behaviour etc, but having it hardcoded in OpenTTD is just not going to fly. We need to be able to enable other people to have their own ideas about it, and implement those. NewGRFs and GameScript (GS) are the tools to make that happen.
And yes, I am aware that there might be limitations to either of those to make your idea happen, but to keep OpenTTD maintainable and still a success in the next 10 years, we are better off spending our time in overcoming these.

In result, I am sorry (and I actually am, as I kinda like the concept of your PR), but I have to reject this Pull Request. If you are open and willing to help us extend the GameScript API to do this better, please do let us know. We are happy to help you out how to get started and how to get that done. But another method of doing it hardcoded is just not going to happen anymore :(

I do hope you understand our decision here :)

PR #10141, opened 2022-11-05
Fix #10140: [SDL2] Lock mouse while moving viewport

### LINKED ISSUE #10140: [Bug]: SDL2 / Wayland: Move viewport with mouse position locked does not lock mouse
### Version of OpenTTD

1.12 Fedora

### Expected result

When moving map/viewport, mouse should stay on fixed position.

### Actual result

During move, mouse is rendered on fixed position, but after move finish, mouse "jumps" to new position.
It is issue with SDL2 library - pull request is created.

### Steps to reproduce

Set viewport scroll behaviour to "...mouse position locked"
Move the map (with right button).
Mouse should stay on fixed position.

tomec-martin:
Setting viewport scroll behaviour does not affect mouse locking - mouse position is not locked. Mouse does not warp, when SDL RelativeMouseMode is false. This is workaround for SDL 2.24.0 (and hopefully next versions).

## Motivation / Problem
Fix of map scrolling with mouse position locked on Fedora 36
https://github.com/OpenTTD/OpenTTD/issues/10140

## Description
Setting of RelativeMouseMode in SDL solved this issue for me (Fedora 36). I am not sure what this change will do on other platforms.

## Limitations
Fix is used only with SDL library of version >= 2.24.0. 

## Checklist for review

### DIFF
--- a/src/video/sdl2_v.cpp
+++ b/src/video/sdl2_v.cpp
@@ -367,13 +367,18 @@ static uint ConvertSdlKeycodeIntoMy(SDL_Keycode kc)
 bool VideoDriver_SDL_Base::PollEvent()
 {
 	SDL_Event ev;
+	// workaround for old SDL libraries
+	bool is_SDL_2_0 = this->sdl_lib_major == 2 && this->sdl_lib_minor < 24;
 
 	if (!SDL_PollEvent(&ev)) return false;
 
 	switch (ev.type) {
 		case SDL_MOUSEMOTION:
-			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, true)) {
+			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, is_SDL_2_0)) {
+				SDL_SetRelativeMouseMode(is_SDL_2_0 ? SDL_FALSE : SDL_TRUE);
 				SDL_WarpMouseInWindow(this->sdl_window, _cursor.pos.x, _cursor.pos.y);
+			} else {
+				SDL_SetRelativeMouseMode(SDL_FALSE);
 			}
 			HandleMouseEvents();
 			break;
@@ -500,9 +505,6 @@ static const char *InitializeSDL()
 	 * UpdateWindowSurface() to update the window's texture instead of
 	 * its surface. */
 	SDL_SetHint(SDL_HINT_FRAMEBUFFER_ACCELERATION, "0");
-#ifndef __EMSCRIPTEN__
-	SDL_SetHint(SDL_HINT_MOUSE_RELATIVE_MODE_WARP, "1");
-#endif
 
 	/* Check if the video-driver is already initialized. */
 	if (SDL_WasInit(SDL_INIT_VIDEO) != 0) return nullptr;
@@ -517,6 +519,10 @@ const char *VideoDriver_SDL_Base::Initialize()
 
 	const char *error = InitializeSDL();
 	if (error != nullptr) return error;
+	SDL_version linked;
+	SDL_GetVersion(&linked);
+	this->sdl_lib_major = linked.major;
+	this->sdl_lib_minor = linked.minor;
 
 	FindResolutions();
 	Debug(driver, 2, "Resolution for display: {}x{}", _cur_resolution.width, _cur_resolution.height);
--- a/src/video/sdl2_v.h
+++ b/src/video/sdl2_v.h
@@ -88,6 +88,9 @@ private:
 	bool edit_box_focused;
 
 	int startup_display;
+
+	/** linked SDL library version - for backward compatibility workarounds */
+	int sdl_lib_major, sdl_lib_minor;
 };
 
 #endif /* VIDEO_SDL_H */


tomec-martin:
Problem exists with all SDL2 versions. I was able to make fix only with SDL 2.24.0. For older SDL2 versions (SDL 2.0.*) it remains unchanged (I dont know how to fix it with older versions)

LordAro:
Still don't understand. Why was this only fixed with SDL 2.24 ? Why can't it be fixed with older versions?

src/video/sdl2_v.h
```diff
@@ -88,6 +88,9 @@ class VideoDriver_SDL_Base : public VideoDriver {
 	bool edit_box_focused;
 
 	int startup_display;
+
+	/** linked SDL library version - for backward compatibility workarounds */
+	int sdl_lib_major, sdl_lib_minor;
```
LordAro, line 93: Why not just store the SDL_version object?

src/video/sdl2_v.cpp
```diff
@@ -500,9 +505,6 @@ static const char *InitializeSDL()
 	 * UpdateWindowSurface() to update the window's texture instead of
 	 * its surface. */
 	SDL_SetHint(SDL_HINT_FRAMEBUFFER_ACCELERATION, "0");
-#ifndef __EMSCRIPTEN__
-	SDL_SetHint(SDL_HINT_MOUSE_RELATIVE_MODE_WARP, "1");
-#endif
```
LordAro, line 505: Why this removal?

src/video/sdl2_v.cpp
```diff
@@ -367,13 +367,18 @@ static uint ConvertSdlKeycodeIntoMy(SDL_Keycode kc)
 bool VideoDriver_SDL_Base::PollEvent()
 {
 	SDL_Event ev;
+	// workaround for old SDL libraries
```
LordAro, line 370: codestyle: comments on their own line should be `/* ... */`

src/video/sdl2_v.cpp
```diff
@@ -517,6 +519,10 @@ const char *VideoDriver_SDL_Base::Initialize()
 
 	const char *error = InitializeSDL();
 	if (error != nullptr) return error;
+	SDL_version linked;
+	SDL_GetVersion(&linked);
+	this->sdl_lib_major = linked.major;
+	this->sdl_lib_minor = linked.minor;
```
LordAro, line 525: Why not just store the `SDL_version` struct, rather than splitting it out like this?

LordAro:
Absolutely, and thank you for the video evidence actually showing what the issue is!

I'm only looking for an explanation about why this has only become an issue now (and only with SDL 2.24 ?) No one thus far has given any explanation why the fix needs to be specific to this version of SDL...

(and the comments above)

stormcone:
I don't experience this problem under Fedora 37 with the settings below. Maybe it depends on other things, not just on the sdl.

$ sdl2-config --version
2.26.0


mouse

kulkom:
I think that it actually does considering that we both are running the same distro and only I am having this issue.
For the record, my enviroment is:
- A Samsung NP300E5C laptop running fully updated Fedora 37
- Using graphics, music and sounds from the original TTD (copied to /usr/share/openttd/data and /usr/share/openttd/gm)
- Usually I am playing with a touchpad, but connecting a USB mouse and disabling the touchpad in the GNOME settings and with the appropriate Fn-key does not help
- This issue occurs even when all settings are reset to default in-game or deleted manually (`rm -rf .local/share/openttd; rm -rf .config/openttd`). Deleting settings manually like that actually resets the default graphics back to OpenGFX (probably because OpenGFX is a dependency of OpenTTD on Fedora) so using OpenGFX doesn't solve this issue.

LordAro:
I think you're onto a winner with Wayland being the ultimate cause here.

I found https://bugzilla.redhat.com/show_bug.cgi?id=1242322 which would suggest the same thing. At the time, SDL2 on Wayland definitely didn't support relative mouse stuff, but that was quite a while ago and https://github.com/libsdl-org/SDL/commit/f11a4409992ca41c50055c27bb7ba42164f30f39 has been in place since 2.0.5...

The SDL changelog mentions a couple of Wayland/mouse related things in 2.24
https://github.com/libsdl-org/SDL/blob/d1b4810542183a415e42b11b516bf8d5b0a6d58c/WhatsNew.txt#L61 (which was amended in 2.26 ? Maybe ?)
https://github.com/libsdl-org/SDL/blob/d1b4810542183a415e42b11b516bf8d5b0a6d58c/WhatsNew.txt#L101
https://github.com/libsdl-org/SDL/blob/d1b4810542183a415e42b11b516bf8d5b0a6d58c/WhatsNew.txt#L38 (2.26)
Could be related?

kulkom:
Yup, for me it appears that this issue is caused by using Wayland. Switching to X11 makes this issue disappear completely.

PatrickJosh:
I experience the same behaviour with OpenTTD on Fedora with SDL 2.26. I'd like to mention that using Wayland is the default for SDL on Fedora. Hence, moving the viewport by mouse is broken out of the box, currently. 

If someone faces this problem as well but does not want to switch to X11 completely: You can start OpenTTD by forcing it to use X11 as follows: `SDL_VIDEODRIVER=x11 openttd`. However, this makes the game look blurry if fractional scaling is enabled.

TrueBrain:
https://gitmemories.com/libsdl-org/SDL/issues/5546 more suggests this problem should be fixed by SDL 2.26. Could you give that a go and let us know if that indeed does solve the problem?

Given now we know it is caused by the SDL backend, and not by the SDL version, I assume the approach of this PR is no longer valid? If something like this would be to land, it would have to check if the backend is Wayland, and compensate for that, not?

But I just hope SDL 2.26 fixes this issue, and we can close up this PR :) Let us know :)

ghisvail:
The Flatpak version uses SDL 2.26.2 and still requires this patch to work properly on Wayland.

src/video/sdl2_v.cpp
ghisvail: Because leaving it may give a false information that mouse relative mode wrap is defined solely by the target platform (anything not WebAssembly if I am correct?), whereas this patch overrides it at runtime based on the SDL version check.

src/video/sdl2_v.cpp
```diff
@@ -367,13 +367,18 @@ static uint ConvertSdlKeycodeIntoMy(SDL_Keycode kc)
 bool VideoDriver_SDL_Base::PollEvent()
 {
 	SDL_Event ev;
+	// workaround for old SDL libraries
+	bool is_SDL_2_0 = this->sdl_lib_major == 2 && this->sdl_lib_minor < 24;
```
ghisvail, line 371: Could we use [SDL_VERSION_ATLEAST](https://wiki.libsdl.org/SDL2/SDL_VERSION_ATLEAST) instead?

src/video/sdl2_v.cpp
ghisvail: Maybe using `SDL_VERSION_ATLEAST` would drop the necessity to store version information altogether.

src/video/sdl2_v.cpp
```diff
@@ -367,13 +367,18 @@ static uint ConvertSdlKeycodeIntoMy(SDL_Keycode kc)
...
 	if (!SDL_PollEvent(&ev)) return false;
 
 	switch (ev.type) {
 		case SDL_MOUSEMOTION:
-			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, true)) {
+			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, is_SDL_2_0)) {
+				SDL_SetRelativeMouseMode(is_SDL_2_0 ? SDL_FALSE : SDL_TRUE);
 				SDL_WarpMouseInWindow(this->sdl_window, _cursor.pos.x, _cursor.pos.y);
+			} else {
+				SDL_SetRelativeMouseMode(SDL_FALSE);
```
ghisvail, line 381: Since the SDL version is unlikely to change from one cursor update to the next, I wonder whether a one-time config at build time would be possible instead of a runtime check?

ghisvail:
The following patch seems to be working on the Flatpak package:

```diff
diff --git a/src/video/sdl2_v.cpp b/src/video/sdl2_v.cpp
index 51c490d23..c7870a8f2 100644
--- a/src/video/sdl2_v.cpp
+++ b/src/video/sdl2_v.cpp
@@ -370,10 +370,17 @@ bool VideoDriver_SDL_Base::PollEvent()

 	if (!SDL_PollEvent(&ev)) return false;

+	const char *sdl_video_driver = SDL_GetCurrentVideoDriver();
+	const bool sdl_on_wayland =
+		(sdl_video_driver != nullptr) ? std::string(sdl_video_driver) == "wayland" : false;
+
 	switch (ev.type) {
 		case SDL_MOUSEMOTION:
-			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, true)) {
+			if (_cursor.UpdateCursorPosition(ev.motion.x, ev.motion.y, !sdl_on_wayland)) {
+				SDL_SetRelativeMouseMode(sdl_on_wayland ? SDL_FALSE : SDL_TRUE);
 				SDL_WarpMouseInWindow(this->sdl_window, _cursor.pos.x, _cursor.pos.y);
+			} else {
+				SDL_SetRelativeMouseMode(SDL_FALSE);
 			}
 			HandleMouseEvents();
 			break;
-- 
2.39.1
```

SuborbitalPigeon:
With the latest version from flatpak, with this patch and forcing X11 through flatseal, the mouse scrolling problem also manifests.

ghisvail:
> With the latest version from flatpak, with this patch and forcing X11 through flatseal, the mouse scrolling problem also manifests.

Why would you force X11 out-of-curiosity?

SuborbitalPigeon:
If I've set a custom font, under Wayland OpenTTD crashes (https://github.com/OpenTTD/OpenTTD/issues/10232), but with X11 it's fine (but then this bug occurs).

TrueBrain:
Although the conversation in this PR has been useful, the content of the PR is, as mentioned above, not the way to go. This is not related to the SDL version, but completely related to X11 vs Wayland.

As such, I am going to close this PR here, and we can continue the conversation in #10140. I will make a comment there with the current state, but short: a solution that works for both X11 and Wayland is harder than we would like.

TrueBrain:
And to ping a bunch of people who have been helpful so far, please see #10140 for two PRs that might address this issue. I could really use some input if they actually do, or make things worse. Tnx a lot!

PR #10142, opened 2022-11-06
Codechange: [MinGW] use pe-bigobj-x86-64 format for x64 debug builds

glx22:
## Motivation / Problem
When trying to understand why #10001 compilation was failing for MinGW x64 debug builds, I compared `objdump -t network_command.cpp.obj` from master and #10001.
```
objdump -t snippet for #10001
[85159](sec -32108)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x0000000000000000 .refptr._Z19CmdClearOrderBackup13DoCommandFlag9TileIndex8ClientID
objdump -t snippet for master
[80970](sec 31671)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x0000000000000000 .refptr._Z19CmdClearOrderBackup13DoCommandFlag9TileIndex8ClientID
```
I noticed a lot of negative section values in #10001, so I though maybe we have too many sections (even if the compiler doesn't complain about that). I then tried to enable pe-bigobj-x86-64 and that solved the issue for #10001.

Then I decided to test something and commented out the ASLR disabling, and #10001 still build fine.
So I tried the same on master, and it builds too.

So in the end it seems OpenTTD again managed to hit some compiler limits (remember the settings hell on macos).

## Description
Add `-Wa,-mbig-obj` compile flag for MinGW x64 debug builds.
As a side effect, ASLR disabling in link options is not needed anymore, and removed.

## Checklist for review

### DIFF
--- a/cmake/CompileFlags.cmake
+++ b/cmake/CompileFlags.cmake
@@ -44,8 +44,8 @@ macro(compile_flags)
             "$<$<NOT:$<CONFIG:Debug>>:-fstack-protector>" # Prevent undefined references when _FORTIFY_SOURCE > 0
         )
         if(CMAKE_SIZEOF_VOID_P EQUAL 8)
-            add_link_options(
-                "$<$<CONFIG:Debug>:-Wl,--disable-dynamicbase,--disable-high-entropy-va,--default-image-base-low>" # ASLR somehow breaks linking for x64 Debug builds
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):
I don't even ... how do you figure out these things?

PR #8895, opened 2021-03-26
Feature: let AI developers edit non-editable AI/Game Script Parameters

perezdidac:
## Motivation / Problem

There is no way to modify non-editable AI/Game Script params during the game or during the scenario editor.

## Description

This code change makes non-editable AI/Game Script params editable during the game or during the scenario editor if _ai_developer_tools_ is set to true in the config file. This will help AI developers and other developers being able to change the params they created in those cases.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10152, opened 2022-11-11
Feature: Allow AI/GS to be fully modified in scenario editor

Arastais:
## Motivation / Problem
There is no way to edit non-editable GS/AI Script parameters in the scenario editor, even though they should since they are not an existing save. You can currently change the script itself in the scenario editor, but not any of its settings/parameters. This problem is also partially outlined in #8895. Currently the workaround is to set `CONFIG_INGAME`, but that is not a good solution for obvious reasons. 

## Description
This change simply allows all GS and AI script parameters to be modified in the scenario editor. This PR is also another partial solution to the problems outlined in #8895.

As a side note, I've also changed out the tabs for spaces in the checks for if a setting is editable to better align them.

## Limitations
The `scenario_developer` setting is intentionally ignored, since I feel being able to modify the AI/GS parameters should be the default behavior for the scenario editor, and I don't think `scenario_developer` applies to this. This can easily be changed it need be however.

## Checklist for review

### DIFF
--- a/src/ai/ai_gui.cpp
+++ b/src/ai/ai_gui.cpp
@@ -575,9 +575,10 @@ private:
 	bool IsEditableItem(const ScriptConfigItem &config_item) const
 	{
 		return _game_mode == GM_MENU
-			|| ((this->slot != OWNER_DEITY) && !Company::IsValidID(this->slot))
-			|| (config_item.flags & SCRIPTCONFIG_INGAME) != 0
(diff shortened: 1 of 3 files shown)


2TallTyler (approved):
Thanks for this. I agree that the `scenario_developer` setting should not influence this (and is probably not properly named, obviously that's out of scope here).

PR #10168, opened 2022-11-13
Change: Requested changes to the Finance GUI

### LINKED ISSUE #10160: [Bug]: Finance window - The title "Net profit" is most probably wrong (2/6)
### Version of OpenTTD

13.0-beta1

### Expected result

Titles in the finance window accurately refer to reality, but at the same time take into account the reality of this game and are not misleading.

### Actual result

I'm not fully convinced that "net profit" is the right title to summarize the company's annual financials. Depending on the country and tax system, these terms may have slightly different meanings. To me, this title seems confusing and not entirely appropriate in this game. First of all, the amount is not always positive, so the phrase "profit" will not always be accurate. Secondly, "net profit" is not a final summary of the annual finances. Dividends are deducted from the net profit. In the case of this game, all dividends go to the owner of the company, i.e. the player, or simply to the company's account. Dividends can also be extracted by script and in the future theoretically they may become an integral part of the game, but in the current form this cannot be done because you cannot subtract anything from the final amount. It seems to me that a more accurate term here  
would be "Financial result", which can be both positive and negative, can include dividends paid and is a de facto summary of all income and expenses.
76B
From wikipedia:
Financial result wiki

### Steps to reproduce

An economist would be helpful here...

### LINKED ISSUE #10163: [Detail]: Finance window - The location of total of revenues or expense seems not entirely legible and confusing (4/6)
### Version of OpenTTD

13.0-beta1

### Expected result

The layout of the information is legible, not overloaded, easy to get used to.

### Actual result

The location of the revenues or expense summary can be confusing, especially when not all lines are filled in.
In this arrangement, it looks a bit like the value from the above part is assigned to the one below. In my proposal, that's why I put the summary at the top of the lists. This is not a rare solution, but well... If you keep the same layout as before, perhaps it would be better to use the existing solutions as well. Anyway, I see several solutions. Will post them in separate posts. Please rate what you think is better.

### Steps to reproduce

n/a

### LINKED ISSUE #10165: [Detail]: Finance Window - The financial summary can be confusing (5/6)
### Version of OpenTTD

13.0-beta1

### Expected result

The design of the window takes into account the players habits of the layout where the number at the top was the most important.

### Actual result

It may be hypersensitivity, but it may be worth paying attention to. In my proposal, apart from naming the missing category and changing the order, it was important to use different colors of the text to mark what is less and what is more important. This is missing in its current form, and since the order has been changed, it's possible that this will at least initially confuse players. For me it is confusing now.

There are a bit of possibilities, some of which I will present below in separate posts.

### Steps to reproduce

Play a bit 13.0
When you want to buy a new vehicle to see if you can afford it, what amount are you looking at: own funds at the top or bank balance at the bottom?

2TallTyler:
## Motivation / Problem

#10160 
#10163
#10165

## Description

changed

Closes #10160 by rewording `Net Profit` to `Profit`. Dividends exist in the Villages Is Villages GS so Net Profit is no good.

Closes #10163 by adding a `Total:` summary to each category. This existed before #9827 but I removed it as unneeded clutter. I'm fine with it now because I was able to remove so much other unneeded text.

Closes #10165 by changing the Bank Balance amount to white, as supported by @michicc. Other suggestions included using gray text, which has no precedence in other GUIs, or putting totals on top, which was previously rejected in discussions of #9827.

## Checklist for review


michicc (dismissed):

PeterN:
I think just "Profit" is better than "Financial Result", to be honest.

LordAro:
Indeed. I didn't think that "Financial Result" was a good string

2TallTyler:
Oops, misread your original comment, sorry @LordAro.

It now says `Profit`.

PeterN:
I don't think Total should have a colon suffix. None of the other lines do :)

michicc (approved):

PR #10027, opened 2022-09-12
Feature: Purchase land multiple tiles at a time

2TallTyler:
## Motivation / Problem

Purchasing land one tile at a time is tedious.

## Description

Purchased land is an actually an object, so I based this atop #9709 to use the new multi-object tool.

If server admins want to stop players from griefing with purchased land, they can change the `frame burst` and `per_64k_frames` settings for objects.

Depends on #9709.

## Limitations

I decided against adding new settings to rate limit purchasing land in network games (`frame burst` and `per_64k_frames`) and am using the object settings added in #9709. I don't see a use case to distinguish the two, but it would be relatively easy to separate in CmdBuildObjectArea() if anyone things it's necessary.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10173, opened 2022-11-14
Change: Allow building canal by area outside editor

2TallTyler:
## Motivation / Problem

In Scenario Editor, we can build canals by area, but in-game we are restricted to building a single-tile-wide canal in one action.

This restriction feels arbitrary and makes building things like this slower:

canal

## Description

Remove the arbitrary restriction on building an area of canal tiles outside Scenario Editor.

I believe this follows a precedent set by #9709 and #10027 where we make it easier for players to build more objects or purchased tiles at once.

I can't think of an exploit that limiting canal construction could be trying to prevent, that wouldn't be an issue with other terraforming.

## Limitations

It is not possible to build canals diagonally with Ctrl like rivers in Scenario Editor, because the Ctrl key is already used to modify the canal tool to flood dry tiles into sea, when in Scenario Editor on dry sea-level tiles.

## Checklist for review

### DIFF
--- a/src/dock_gui.cpp
+++ b/src/dock_gui.cpp
@@ -192,7 +192,7 @@ struct BuildDocksToolbarWindow : Window {
 	{
 		switch (this->last_clicked_widget) {
 			case WID_DT_CANAL: // Build canal button
-				VpStartPlaceSizing(tile, (_game_mode == GM_EDITOR) ? VPM_X_AND_Y : VPM_X_OR_Y, DDSP_CREATE_WATER);
+				VpStartPlaceSizing(tile, VPM_X_AND_Y, DDSP_CREATE_WATER);
 				break;
 
 			case WID_DT_LOCK: // Build lock button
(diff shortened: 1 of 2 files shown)


Arastais:
> It is not possible to build canals diagonally with Ctrl like rivers in Scenario Editor, because the Ctrl key is already used to modify the canal tool to flood dry tiles into sea, when in Scenario Editor on dry sea-level tiles.

Since building diagonally is quite important, I suggest you move the current functionality to the alt key or to a new button in the gui all together (would require making a new icon, but it can just be the river icon but wider). Adding a button to the gui would also have the benefit of being more clear to the player; you can show an error message for when you're not allowed to flood, and you can't accidentally build a canal instead of flooding because you ctrl-clicked on somewhere that wasn't sea-level, but didn't notice

PeterN (approved):
Simple enough, though might make it slightly harder to build long single-width canals as you have to be more precise.

PR #10201, opened 2022-11-29
Fix: Missing extra padding when drawing tooltip text.

PeterN:
## Motivation / Problem

This padding is included when calculating the size of the tooltips, the difference causes a mismatch in height for some tooltips.

image

## Description

Apply padding when drawing.

image

Note that this tooltip wraps onto 4 lines because tooltips have a maximum width. While putting one word on the last line is arguable, the intention of this change is to make the padding consistent between sizing calculation and drawing.

The second commit changes the layout so that a grey panel isn't needlessly drawn underneath the tooltip.

## Checklist for review

### DIFF
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -656,7 +656,7 @@ void HideFillingPercent(TextEffectID *te_id)
 }
 
 static const NWidgetPart _nested_tooltips_widgets[] = {
-	NWidget(WWT_PANEL, COLOUR_GREY, WID_TT_BACKGROUND), SetMinimalSize(200, 32), EndContainer(),
+	NWidget(WWT_EMPTY, INVALID_COLOUR, WID_TT_BACKGROUND), EndContainer(),
 };
 
 static WindowDesc _tool_tips_desc(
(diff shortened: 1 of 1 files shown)


src/misc_gui.cpp
```diff
@@ -656,7 +656,7 @@ void HideFillingPercent(TextEffectID *te_id)
 }
 
 static const NWidgetPart _nested_tooltips_widgets[] = {
-	NWidget(WWT_PANEL, COLOUR_GREY, WID_TT_BACKGROUND), SetMinimalSize(200, 32), EndContainer(),
+	NWidget(WWT_EMPTY, COLOUR_GREY, WID_TT_BACKGROUND), SetMinimalSize(200, 32), EndContainer(),
```
2TallTyler: It's it's empty, does it need to be grey? Isn't there an invalid colour to use? (Also, why does there need to be an invisible widget under the tooltip?) Lots of questions, sorry. 😛

PeterN: The tooltip widget is drawn completely by the Window's DrawWidget handler, so WWT_EMPTY is appropriate. WWT_PANEL draws a filled area with shaded bevel, which was then drawn over. But INVALID_COLOUR would work though. That SetMinimalSize is unneeded too, I guess.

2TallTyler (approved):

PR #10204, opened 2022-11-30
Feature: Add alternative setting for right-click close window option to exclude pinned windows

ShreksHellraiser:
## Motivation / Problem

The right-click to close windows option is a great gameplay improvement, however it makes it very easy and common to close windows you don't intend to. 

## Description

My solution to this issue is to change the right-click to close windows setting to be a dropdown rather than a simple toggle, with a third option to exclude pinned windows. When set to this third mode, simply pin a window and right-click will no longer close the window.

## Limitations

Since I changed the type of the config option from a `bool` to a `uint8` the game will give you a warning upon launching if you last saved settings from a version that did not make this change. This window can be easily dismissed and as far as I can tell there are no negative effects from this.

I've been looking into solutions for this, but modifying the setting loading code to smoothly change this format to the new type is not a satisfactory solution.

I'm also not very satisfied with my if condition in `window.cpp`, it's quite long, but I'm not sure of a better way to split it.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1618,6 +1618,10 @@ STR_CONFIG_SETTING_RIGHT_MOUSE_BTN_EMU_OFF                      :Off
 
 STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE                        :Close window on right-click: {STRING2}
 STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_HELPTEXT               :Closes a window by right-clicking inside it. Disables the tooltip on right-click!
+###length 3
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_NO                     :No
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_YES                    :Yes
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_YES_EXCEPT_STICKY      :Yes, except sticky
 
 STR_CONFIG_SETTING_AUTOSAVE                                     :Autosave: {STRING2}
 STR_CONFIG_SETTING_AUTOSAVE_HELPTEXT                            :Select interval between automatic game saves
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -160,8 +160,8 @@ enum IniFileVersion : uint32_t {
 	IFV_GAME_TYPE,                                         ///< 2  PR#9515  Convert server_advertise to server_game_type.
 	IFV_LINKGRAPH_SECONDS,                                 ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	IFV_NETWORK_PRIVATE_SETTINGS,                          ///< 4  PR#10762 Move no_http_content_downloads / use_relay_service to private settings.
-
 	IFV_AUTOSAVE_RENAME,                                   ///< 5  PR#11143 Renamed values of autosave to be in minutes.
+	IFV_RIGHT_CLICK_CLOSE,                                 ///< 6  PR#10204 Add alternative right click to close windows setting.
 
 	IFV_MAX_VERSION,       ///< Highest possible ini-file version.
 };
@@ -1332,6 +1332,12 @@ void LoadFromConfig(bool startup)
 				case 4: _settings_client.gui.autosave_interval = std::chrono::minutes(120); break;
 				default: break;
 			}
+		}
+
+		/* Persist the right click close option from older versions. */
+		if (generic_version < IFV_RIGHT_CLICK_CLOSE && IsConversionNeeded(generic_ini, "gui", "right_mouse_wnd_close", "right_click_wnd_close", &old_item)) {
+			auto old_value = BoolSettingDesc::ParseSingleValue(old_item->value->c_str());
+			_settings_client.gui.right_click_wnd_close = old_value.value_or(false) ? RCC_YES : RCC_NO;
 		}
 
 		_grfconfig_newgame = GRFLoadConfig(generic_ini, "newgrf", false);
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -77,6 +77,13 @@ enum ParticipateSurvey {
 	PS_YES,
 };
 
+/** Right-click to close window actions. */
+enum RightClickClose : uint8_t {
+	RCC_NO = 0,
+	RCC_YES,
+	RCC_YES_EXCEPT_STICKY,
+};
+
 /** Settings related to the difficulty of the game */
 struct DifficultySettings {
 	byte   competitor_start_time;            ///< Unused value, used to load old savegames.
@@ -154,7 +161,7 @@ struct GUISettings {
 	uint8_t  scrollwheel_scrolling;            ///< scrolling using the scroll wheel?
 	uint8_t  scrollwheel_multiplier;           ///< how much 'wheel' per incoming event from the OS?
 	bool   timetable_arrival_departure;      ///< show arrivals and departures in vehicle timetables
-	bool   right_mouse_wnd_close;            ///< close window with right click
+	RightClickClose  right_click_wnd_close;  ///< close window with right click
 	bool   pause_on_newgame;                 ///< whether to start new games paused or not
 	SignalGUISettings signal_gui_mode;       ///< select which signal types are shown in the signal GUI
 	SignalCycleSettings cycle_signal_types;  ///< Which signal types to cycle with the build signal tool.
--- a/src/table/settings/gui_settings.ini
+++ b/src/table/settings/gui_settings.ini
@@ -18,6 +18,7 @@ static void SpriteZoomMinChanged(int32_t new_value);
 
 static constexpr std::initializer_list<const char*> _osk_activation{"disabled", "double", "single", "immediately"};
 static constexpr std::initializer_list<const char*> _savegame_date{"long", "short", "iso"};
+static constexpr std::initializer_list<const char*> _right_click_close{"no", "yes", "except sticky"};
 
 static const SettingVariant _gui_settings_table[] = {
 [post-amble]
@@ -128,12 +129,17 @@ def      = false
 str      = STR_CONFIG_SETTING_SMOOTH_SCROLLING
 strhelp  = STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT
 
-[SDTC_BOOL]
-var      = gui.right_mouse_wnd_close
-flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
-def      = false
+[SDTC_OMANY]
+var      = gui.right_click_wnd_close
+type     = SLE_UINT8
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_DROPDOWN
+def      = RCC_NO
+min      = RCC_NO
+max      = RCC_YES_EXCEPT_STICKY
+full     = _right_click_close
 str      = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE
 strhelp  = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_HELPTEXT
+strval   = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_NO
 cat      = SC_BASIC
 
 ; We might need to emulate a right mouse button on mac
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -819,8 +819,11 @@ static void DispatchRightClickEvent(Window *w, int x, int y)
 		if (w->OnRightClick(pt, wid->index)) return;
 	}
 
-	/* Right-click close is enabled and there is a closebox */
-	if (_settings_client.gui.right_mouse_wnd_close && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
+	/* Right-click close is enabled and there is a closebox. */
+	if (_settings_client.gui.right_click_wnd_close == RCC_YES && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
+		w->Close();
+	} else if (_settings_client.gui.right_click_wnd_close == RCC_YES_EXCEPT_STICKY && (w->flags & WF_STICKY) == 0 && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
+		/* Right-click close is enabled, but excluding sticky windows. */
 		w->Close();
 	} else if (_settings_client.gui.hover_delay_ms == 0 && !w->OnTooltip(pt, wid->index, TCC_RIGHT_CLICK) && wid->tool_tip != 0) {
 		GuiShowTooltips(w, wid->tool_tip, TCC_RIGHT_CLICK);
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -1780,7 +1780,7 @@ static SettingsContainer &GetSettingsTree()
 				general->Add(new SettingEntry("gui.errmsg_duration"));
 				general->Add(new SettingEntry("gui.window_snap_radius"));
 				general->Add(new SettingEntry("gui.window_soft_limit"));
-				general->Add(new SettingEntry("gui.right_mouse_wnd_close"));
+				general->Add(new SettingEntry("gui.right_click_wnd_close"));
 			}
 
 			SettingsPage *viewports = interface->Add(new SettingsPage(STR_CONFIG_SETTING_INTERFACE_VIEWPORTS));


LC-Zorg:
Player #623071 review: For my way of navigating the game interface, this would be a redundant option, although I do occasionally close a window I wanted to keep open. As the new behavior would be an added option, not a replacement, I have nothing against. Ok, just one thing: the red/green button is more visible, but I'm not convinced that's very important. However, I would have two suggestions.

1. Please make right-click closing the default setting. I fully agree that this is a great gameplay improvement. It makes the game so much easier, I can't see any compelling arguments against it, so why shouldn't new players be able to take advantage of it? :)

2. In the case of setting where clicking on a pinned window doesn't close it, maybe a good solution would be to allow it to be closed using Ctrl + click?

MasonGulu:
I don't really have any authority to change default behavior, I'm just adding another option to help with accessibility.

I could add a modifier key to close the window, but I feel like at that point it's just as easy to unpin the window and close it. Adding a modifier key also increases the complexity of documenting this, would I hardcore the modifier, or would I use an existing keybind?

I'd like to receive some more input before trying to learn how user keyboard and keybind input works. But thank you for the feedback!

src/right_click_type.h
```diff
@@ -0,0 +1,20 @@
...
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file right_click_type.h Types related to right click options. */
+#ifndef RIGHT_CLICK_TYPE_H
+#define RIGHT_CLICK_TYPE_H
+
+#include "core/enum_type.hpp"
+
+enum RightClickClose : uint8 {
```
PeterN: This enum should either include a prefix `RCC_` on each value or be declared as `enum class` and use Pascal case values.

src/settings_type.h
```diff
@@ -144,7 +145,7 @@ struct GUISettings {
 	uint8  scrollwheel_scrolling;            ///< scrolling using the scroll wheel?
 	uint8  scrollwheel_multiplier;           ///< how much 'wheel' per incoming event from the OS?
 	bool   timetable_arrival_departure;      ///< show arrivals and departures in vehicle timetables
-	bool   right_mouse_wnd_close;            ///< close window with right click
+	uint8   right_mouse_wnd_close;            ///< close window with right click
```
PeterN: Use the declared enum type?

src/table/settings/gui_settings.ini
```diff
@@ -130,12 +130,16 @@ def      = false
...
-[SDTC_BOOL]
+[SDTC_VAR]
 var      = gui.right_mouse_wnd_close
-flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
-def      = false
+type     = SLE_UINT8
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_DROPDOWN
+def      = 0
+min      = 0
+max      = 2
```
PeterN: You might be able to use the enum values here instead of integers, not sure though.

src/window.cpp
```diff
@@ -805,7 +805,10 @@ static void DispatchRightClickEvent(Window *w, int x, int y)
 	}
 
 	/* Right-click close is enabled and there is a closebox */
-	if (_settings_client.gui.right_mouse_wnd_close && w->nested_root->GetWidgetOfType(WWT_CLOSEBOX)) {
+	if (_settings_client.gui.right_mouse_wnd_close == CLOSE && w->nested_root->GetWidgetOfType(WWT_CLOSEBOX)) {
+		w->Close();
+	/* Right-click close is enabled, but excluding pinned windows */
```
PeterN: Indentation

src/window.cpp
```diff
@@ -805,7 +805,10 @@ static void DispatchRightClickEvent(Window *w, int x, int y)
...
+	} else if (_settings_client.gui.right_mouse_wnd_close == CLOSE_EXCEPT_STICKY  && w->nested_root->GetWidgetOfType(WWT_CLOSEBOX) && (w->flags & WF_STICKY) == 0) {
```
PeterN: Spurious double-space before first &&

PeterN (dismissed):

src/settings_type.h
glx22:
```suggestion
	uint8  right_mouse_wnd_close;            ///< close window with right click
```

glx22: And @PeterN has a point too.

src/settings_type.h
```diff
@@ -68,6 +68,13 @@ enum UseRelayService {
 	URS_ALLOW,
 };
 
+/** Right-click to close window actions. */
+enum RightClickClose {
```
PeterN: I think it still needs `: uint8` hereto force the enum type as the settings type is SLE_UINT8.

MasonGulu:
I've changed the enum type. 

Is the setting type warning something that needs fixed/mitigated?

2TallTyler:
> Is the setting type warning something that needs fixed/mitigated?

I think so, yes. Most players don't know that the warning is harmless, and being greeted by a warning immediately upon opening a newly-updated game release is bound to result in bug reports, complaints, and a bad taste in the mouth of some players.

How to actually fix that, I don't know. 🤷

MasonGulu:
I am looking into changing the setting type to `SDT_OMANY` so I can define a `load` function to handle conversion.

How do I fix the last commit name? The coding style guide doesn't mention a commit name for catching up to master.

2TallTyler:
Merge commits shouldn't be included. You'll need to rebase and force push.

2TallTyler:
@MasonGulu: Still interested in working on this? I would suggest asking in the `#openttd-development` channel in the official Discord server if you aren't sure how to update the setting to avoid the error message.

MasonGulu:
I have resolved the warning on first launch issue by renaming the option to `right_click_wnd_close` from `right_mouse_wnd_close`. This has the added benefit of not interfering with older game versions, in cases when the config is shared between different instances.

src/window.cpp
```diff
@@ -820,7 +820,10 @@ static void DispatchRightClickEvent(Window *w, int x, int y)
 	}
 
 	/* Right-click close is enabled and there is a closebox */
-	if (_settings_client.gui.right_mouse_wnd_close && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
+	if (_settings_client.gui.right_click_wnd_close == RCC_YES && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
+		w->Close();
+	/* Right-click close is enabled, but excluding pinned windows */
+	} else if (_settings_client.gui.right_click_wnd_close == RCC_YES_EXCEPT_STICKY && (w->flags & WF_STICKY) == 0&& (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
```
2TallTyler:
```suggestion
	} else if (_settings_client.gui.right_click_wnd_close == RCC_YES_EXCEPT_STICKY && (w->flags & WF_STICKY) == 0 && (w->window_desc->flags & WDF_NO_CLOSE) == 0) {
```

src/window.cpp
2TallTyler: Comment matches the `else if` below, unless you mean something else?

src/table/settings/gui_settings.ini
2TallTyler: Was this not possible?

src/settings_type.h
```diff
@@ -154,7 +161,7 @@ struct GUISettings {
 	uint8  scrollwheel_scrolling;            ///< scrolling using the scroll wheel?
 	uint8  scrollwheel_multiplier;           ///< how much 'wheel' per incoming event from the OS?
 	bool   timetable_arrival_departure;      ///< show arrivals and departures in vehicle timetables
-	bool   right_mouse_wnd_close;            ///< close window with right click
+	uint8   right_click_wnd_close;            ///< close window with right click
```
2TallTyler: Indentation:
```suggestion
	uint8  right_click_wnd_close;            ///< close window with right click
```

src/table/settings/gui_settings.ini
ShreksHellraiser: This slipped my mind when I picked this back up, this is possible and has been done in the latest commit

2TallTyler:
Try an interactive rebase (`git rebase -i head~2` to edit the last two commits, then `fixup` the latest commit) to squash that commit and make commit checker happy 😉.

2TallTyler (dismissed):
Looks good to me. Thanks for the contribution!

(Since you haven't done this before: from here it waits for another developer to look it over and either request changes, or agree that it's suitable for inclusion in OpenTTD and hit the Merge button 🙂.)

src/settings_type.h
```diff
@@ -77,6 +77,13 @@ enum ParticipateSurvey {
 	PS_YES,
 };
 
+/** Right-click to close window actions. */
+enum RightClickClose : uint8 {
+	RCC_NO_ACTION,
+	RCC_YES,
+	RCC_YES_EXCEPT_STICKY,
```
TrueBrain, line 84: These names are a bit unbalanced. `no-action` sounds good, but `yes` is then a bit weird. I would either expect no/yes, or some other wording for yes.

To make code more explicit, I don't mind the `RCC_NO_ACTION`. A suggestion could be `RCC_CLOSE_ALL` as the second option (and `RCC_CLOSE_EXCEPT_STICKY` for the third). But otherwise please make it `RCC_NO`. Either choice is fine by me.

The reason for this is that it might otherwise confuse future-us in a year or so, thinking that there was some special meaning here.

src/table/settings/gui_settings.ini
```diff
@@ -130,12 +130,17 @@ def      = false
 str      = STR_CONFIG_SETTING_SMOOTH_SCROLLING
 strhelp  = STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT
 
-[SDTC_BOOL]
-var      = gui.right_mouse_wnd_close
-flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
-def      = false
+; This was originally right_mouse_wnd_close, renamed to change type to uint8
```
TrueBrain: This comment doesn't actually add value, funny enough. It is useful for this PR, sure, but once it is merged, it is no longer useful. Imagine if we leave everything we once removed in comments behind, the codebase would be littered with those comments :)

For this we use version-control. If someone actually wants to know these things, they can dive into the old history themselves :)

src/window.cpp
TrueBrain: The comment is in the wrong place. It is unclear it is a comment for the `else if` statement. Instead, if you move it inside of the body of that statement, and indent it, it becomes much more clear what is going on. So something like:

```
} else if (...) {
    /* Right-click close is enabled, but excluding pinned windows. */
   w->Close();
```

PS: missing dot at end of sentence in that comment. Please add that. Years ago we changed that sentences should end in dots; it is only older code that doesn't do that yet. You don't have to change older code; only when you touch it :)

src/table/settings/gui_settings.ini
```diff
@@ -130,12 +130,17 @@ def      = false
 str      = STR_CONFIG_SETTING_SMOOTH_SCROLLING
 strhelp  = STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT
 
-[SDTC_BOOL]
-var      = gui.right_mouse_wnd_close
```
TrueBrain, line 132: I do not like that we reset the setting of everything. As a player, that would be rather frustrating: because someone wanted to add an option, everyone has to suffer, basically ;)

But, it is not the first time we encounter this situation .. and the solution takes a tiny bit of work.

~The best example I know of, is `_old_units`. You can grep the code for that, and see how it is done. It does require a savegame bump to be handled properly, which might be a bit complicated for your first PR, so if you can't figure it out, do ask for help. But basically:~

~Keep the old setting around, and use `var` to store it in an `_old_NNN` variable. Add a new `SLV_NNN` to bump the savegame version. In `afterload.cpp`, check all savegames before the new version, and use the `_old_NNN` to write to the new `right_click_wnd_close` to correct value.~

~This way, nobody that was using "yes" will notice any difference.~

~PS: `NNN` ofcourse is some name you have to come up with. Not literally `NNN` ;)~

TrueBrain (changes requested):
Sorry, I think @2TallTyler jumped the gun here. A few more things to pick up here :)

src/settings_type.h
TrueBrain: This wasn't done? Any reason why it was resolved and not done? :)

src/table/settings/gui_settings.ini
```diff
@@ -130,12 +130,17 @@ def      = false
...
+var      = gui.right_click_wnd_close
+type     = SLE_UINT8
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_DROPDOWN
+def      = RCC_NO_ACTION
+min      = RCC_NO_ACTION
+max      = RCC_YES_EXCEPT_STICKY
 str      = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE
 strhelp  = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_HELPTEXT
+strval   = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_NO_ACTION
 cat      = SC_BASIC
```
TrueBrain, line 143: I thought I wrote a comment about this, but it seems either GitHub ate it, or I never pressed the little green button. Anyway, on the off chance of repeating myself:

Please change the type to `SDTC_OMANY` and add a `full` entry. See `gui.autosave` for an example. Now you can attach a list to it, that translates the integer to a name and back. That way, we don't see in the configuration:

`right_click_wnd_close = 2`, but instead `right_click_wnd_close = close-except-sticky`. This makes it much more understandable.

The list has to be in the same order as the enum entries are.

We have been doing this for all new enums; there are only a ton of old ones we haven't done yet.

src/table/settings/gui_settings.ini
TrueBrain: Okay, forget that, there is an easier way to do this. ~When you changed the type to `OMANY`, we can simply define a `load` to do conversion. Besides, my earlier story assumed we saved this setting in the savegame .. we don't. Even easier :D~

~Basically, with `OMANY`, it first tries to use the supplied list to convert the string to a number. If that fails, it will check if a `load` function is defined. Here you can convert `true` / `false` to the correct value. An example of this is `ConvertLandscape`, and a very recent PR I made to confirm this actually works, adding `ConvertAutosave`:~

https://github.com/OpenTTD/OpenTTD/pull/11143

~As you can see in that PR, it didn't work out-of-the-box, as the `load` was never connected in the `SDTC_OMANY` yet. But that is fixed wit that PR :)~

~If you follow that, the code becomes much easier. You do need to rename the variable back to `right_mouse_wnd_close` for it to work. But you do need to wait for that PR to be merged, ofc :)~

TrueBrain: And a follow-up: you can still rename the setting if we so like, and #11143 shows how you can convert the old value to the new value. It is a bit of code, and requires a configuration-version-bump, but with having the example from #11143 I am sure you can figure it out :D And if not, let us know!

PS: this would not require a `load` in the setting.

src/settings.cpp
```diff
@@ -1262,6 +1263,20 @@ void LoadFromConfig(bool startup)
...
+		/* Persist the right click close option from older versions. */
+		if (generic_version < IFV_RIGHT_CLICK_CLOSE) {
+			IniGroup *gui = generic_ini.GetGroup("gui", false);
+			if (gui != nullptr) {
+				IniItem *right_mouse_wnd_close = gui->GetItem("right_mouse_wnd_close");
+				if (right_mouse_wnd_close != nullptr) {
+					if (right_mouse_wnd_close->value) {
+						std::string str = right_mouse_wnd_close->value.value();
+						_settings_client.gui.right_click_wnd_close = strcmp(str.c_str(), "true") == 0 ? 1 : 0;
+					}
```
TrueBrain: We should reusing setting code here. Will write some code for that tomorrow. This are reminder to myself :)

src/lang/english.txt
```diff
@@ -1618,6 +1618,10 @@ STR_CONFIG_SETTING_RIGHT_MOUSE_BTN_EMU_OFF                      :Off
 
 STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE                        :Close window on right-click: {STRING2}
 STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_HELPTEXT               :Closes a window by right-clicking inside it. Disables the tooltip on right-click!
+###length 3
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_NO                     :No
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_YES                    :Yes
+STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_YES_EXCEPT_STICKY      :Yes, except pinned
```
TrueBrain: In two places it is called `pinned`, while in all other places it is called `sticky`. I think we should go for sticky, as that seems to be the term in use. What do you think?

src/settings.cpp
TrueBrain: I am figure out with other developers what we exactly want here, and it seems we are narrowing down on a solution. Hang tight, I will help out with this soon :)

TrueBrain: After a rebase, you can do something like:

```
		if (generic_version < IFV_RIGHT_CLICK_CLOSE && IsConversionNeeded(generic_ini, "gui", "right_mouse_wnd_close", "right_click_wnd_close", &old_item)) {
			auto old_value = BoolSettingDesc::ParseSingleValue(old_item->value->c_str());
			_settings_client.gui.right_click_wnd_close = old_value.value_or(false) ? RCC_YES : RCC_NO;
		}
```

Ignoring any typos I might have made :D

src/lang/english.txt
ShreksHellraiser: Code seems to call it sticky, while all strings refer to it as pinned. My code was a little inconsistent on that naming scheme so I can fix that.

TrueBrain: I couldn't find any reference to "pinned" in the strings. Did I miss any? (Honest question, to be clear)

ShreksHellraiser: Looking again I seem to have misremembered, the only reference to pinned I can find is in a comment https://github.com/OpenTTD/OpenTTD/blob/master/src/window.cpp#L3300

MasonGulu:
That should be all my references to "pinned" swapped out for "sticky", rebased on the latest master branch, and with the config loading code provided

TrueBrain (approved):

TrueBrain:
Tnx again for your work on this, and your patience while we figured out what we actually wanted :)

PR #10215, opened 2022-12-04
Feature: Expand all towns in the scenario editor

perezdidac:
## Motivation / Problem

* As a user, I want to make all towns grow in the scenario editor in case I want to play a game with towns and cities that have more population.
* I asked around on Discord and people seemed excited about adding this functionality.

## Description

This code change adds a button in the town window in the scenario editor that will expand all towns on the map.

## Limitations

This code change does not add a button to decrease the population of all downs.

## Checklist for review

N/A

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2851,6 +2851,8 @@ STR_FOUND_TOWN_RANDOM_TOWN_BUTTON                               :{BLACK}Random T
 STR_FOUND_TOWN_RANDOM_TOWN_TOOLTIP                              :{BLACK}Found town in random location
 STR_FOUND_TOWN_MANY_RANDOM_TOWNS                                :{BLACK}Many random towns
 STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP                             :{BLACK}Cover the map with randomly placed towns
+STR_FOUND_TOWN_EXPAND_ALL_TOWNS                                 :{BLACK}Expand all towns
+STR_FOUND_TOWN_EXPAND_ALL_TOWNS_TOOLTIP                         :{BLACK}Make all towns grow slightly
 
 STR_FOUND_TOWN_NAME_TITLE                                       :{YELLOW}Town name:
 STR_FOUND_TOWN_NAME_EDITOR_TITLE                                :{BLACK}Enter town name
--- a/src/town_gui.cpp
+++ b/src/town_gui.cpp
@@ -1061,7 +1061,9 @@ static const NWidgetPart _nested_found_town_widgets[] = {
 		NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_TF_RANDOM_TOWN), SetMinimalSize(156, 12), SetFill(1, 0),
 										SetDataTip(STR_FOUND_TOWN_RANDOM_TOWN_BUTTON, STR_FOUND_TOWN_RANDOM_TOWN_TOOLTIP), SetPadding(0, 2, 1, 2),
 		NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_TF_MANY_RANDOM_TOWNS), SetMinimalSize(156, 12), SetFill(1, 0),
-										SetDataTip(STR_FOUND_TOWN_MANY_RANDOM_TOWNS, STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP), SetPadding(0, 2, 0, 2),
+										SetDataTip(STR_FOUND_TOWN_MANY_RANDOM_TOWNS, STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP), SetPadding(0, 2, 1, 2),
+		NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_TF_EXPAND_ALL_TOWNS), SetMinimalSize(156, 12), SetFill(1, 0),
+										SetDataTip(STR_FOUND_TOWN_EXPAND_ALL_TOWNS, STR_FOUND_TOWN_EXPAND_ALL_TOWNS_TOOLTIP), SetPadding(0, 2, 0, 2),
 		/* Town name selection. */
 		NWidget(WWT_LABEL, COLOUR_DARK_GREEN), SetMinimalSize(156, 14), SetPadding(0, 2, 0, 2), SetDataTip(STR_FOUND_TOWN_NAME_TITLE, STR_NULL),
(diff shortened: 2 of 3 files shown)


perezdidac:
I just tested the deployment https://preview.openttd.org/pr10215/ and it does work. I might add a 10x multiplier since they don't grow much in every click.

src/lang/english.txt
```diff
@@ -2851,6 +2851,8 @@ STR_FOUND_TOWN_RANDOM_TOWN_BUTTON                               :{BLACK}Random T
 STR_FOUND_TOWN_RANDOM_TOWN_TOOLTIP                              :{BLACK}Found town in random location
 STR_FOUND_TOWN_MANY_RANDOM_TOWNS                                :{BLACK}Many random towns
 STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP                             :{BLACK}Cover the map with randomly placed towns
+STR_FOUND_TOWN_EXPAND_ALL_TOWNS                                 :{BLACK}Expand all towns
+STR_FOUND_TOWN_EXPAND_ALL_TOWNS_TOOLTIP                         :{BLACK}Make all towns slightly grow
```
James103: A better wording for this string would be: <code>Make all towns <b>grow slightly</b></code> (emphasis mine)

perezdidac: Thanks! Fixed.

LC-Zorg:
It's funny, but not so long ago, that's exactly what I missed. :) I was working on a scenario where many cities were very small, literally 2-4 buildings, which often made it difficult for them to start growing at all. So I think this could be a useful feature. :)

If I had to make a comment, it would be that small cities grow too much, and big ones too little, often hardly at all. In the case of the expansion function that is in the city window, each press will enlarge the city by about 20%. Is it possible (easy to do) to make the multi-city extension work similarly?

I also think it might be worth adding a confirmation request for at least "Many random towns" - this feature is beside and accidentally pressing it when expanding towns can cause unpleasant results.

src/town_cmd.cpp
```diff
@@ -2272,6 +2272,19 @@ bool GenerateTowns(TownLayout layout)
 	return false;  // we are still without a town? we failed, simply
 }
 
+void ExpandTowns()
```
2TallTyler: Instead of redefining town growth, this should call `CMD_EXPAND_TOWN` inside the loop. You probably want to adapt the OnClick handler for `WID_TV_EXPAND` to get its warning message.

src/town_gui.cpp
```diff
@@ -1062,6 +1062,8 @@ static const NWidgetPart _nested_found_town_widgets[] = {
 										SetDataTip(STR_FOUND_TOWN_RANDOM_TOWN_BUTTON, STR_FOUND_TOWN_RANDOM_TOWN_TOOLTIP), SetPadding(0, 2, 1, 2),
 		NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_TF_MANY_RANDOM_TOWNS), SetMinimalSize(156, 12), SetFill(1, 0),
 										SetDataTip(STR_FOUND_TOWN_MANY_RANDOM_TOWNS, STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP), SetPadding(0, 2, 0, 2),
```
2TallTyler: Need some padding between these two buttons. Bottom padding on the button above would match the other buttons here.

```suggestion
										SetDataTip(STR_FOUND_TOWN_MANY_RANDOM_TOWNS, STR_FOUND_TOWN_RANDOM_TOWNS_TOOLTIP), SetPadding(0, 2, 1, 2),
```

src/widgets/town_widget.h
```diff
@@ -49,6 +49,7 @@ enum TownFoundingWidgets {
 	WID_TF_NEW_TOWN,          ///< Create a new town.
 	WID_TF_RANDOM_TOWN,       ///< Randomly place a town.
 	WID_TF_MANY_RANDOM_TOWNS, ///< Randomly place many towns.
+	WID_TF_EXPAND_ALL_TOWNS,  ///< Make all towns slightly grow.
```
2TallTyler: Might as well match the tooltip. 🙂 

```suggestion
	WID_TF_EXPAND_ALL_TOWNS,  ///< Make all towns grow slightly.
```

2TallTyler (changes requested):
If you can get these changes done promptly, this would be good to have in OpenTTD 13 (Release Candidate build and feature-freeze coming soon ™️). 😃

src/town_gui.cpp
perezdidac: Fixed!! thank you :D

perezdidac:
> It's funny, but not so long ago, that's exactly what I missed. :) I was working on a scenario where many cities were very small, literally 2-4 buildings, which often made it difficult for them to start growing at all. So I think this could be a useful feature. :)
> 
> If I had to make a comment, it would be that small cities grow too much, and big ones too little, often hardly at all. In the case of the expansion function that is in the city window, each press will enlarge the city by about 20%. Is it possible (easy to do) to make the multi-city extension work similarly?
> 
> I also think it might be worth adding a confirmation request for at least "Many random towns" - this feature is beside and accidentally pressing it when expanding towns can cause unpleasant results.

I made the changes that @2TallTyler suggested and now it effectively does what you suggest :D

src/town_cmd.cpp
```diff
@@ -2272,6 +2272,14 @@ bool GenerateTowns(TownLayout layout)
...
+{
+	for (Town *t : Town::Iterate()) {
+		Command<CMD_EXPAND_TOWN>::Do(DC_EXEC, t->index, 0);
+	}
+
+	UpdateAirportsNoise();
```
LordAro: Why is this here? (and only this?) I'd expect noise levels to be handled by each expand-town cmd

src/town_cmd.cpp
```diff
@@ -2272,6 +2272,14 @@ bool GenerateTowns(TownLayout layout)
 	return false;  // we are still without a town? we failed, simply
 }
 
+void ExpandTowns()
```
LordAro: this function is small enough that it might be worth considering not having it at all, just sticking it in the loop in town_gui.cpp ?

but if not, requires some doxygen comments :)

2TallTyler: I agree, the whole loop could be moved to the OnClick handler.

src/town_cmd.cpp
perezdidac: I looked at similar code paths and thought it was necessary, and only done after all towns have been expanded. Fixed it!

src/town_cmd.cpp
perezdidac: Fixed this too, thanks for the suggestions!

2TallTyler (approved):

perezdidac:
Thank you guys!

PR #10220, opened 2022-12-05
Feature: Engine variant groups

PeterN:
## Motivation / Problem

See #8446.

Currently NewGRFs that want to provide a variety of different paint-jobs for the same vehicle can use cargo subtypes. However, these are very hidden from players; an engine must be purchased, and then refitted while in the depot, before the player even knows what liveries exist. This process is also tricky to get right with autoreplace.

Since a long time, OpenTTD has enough engine IDs to not really worry about wasting them, so it's possible to provide different liveries by creating separate engines. However this results in a very long purchase list.

## Description

Enter engine variant groups. Grouped variants are collapsed by default, and can be expanded:

image
image

Variants groups are collapsed by default, to avoid cluttering the purchase list. They can be expanded to reveal all the variants. Multiple levels of grouping is also supported. When purchasing a variant, the selected variant is remembered (but not saved) and that variant will be the purchased variant if the group is collapsed. This allows the player to open the group, pick a style, and close it again.

Because the variants are separate engine IDs, they can have independent stats, dates, names, etc, and can be independently autoreplaced.

This implements a new NewGRF property for all vehicle types to specify the primary engine ID for a variant. This must exist as a defined engine else it will be ignored.

| Feature | Property | Size | |
| ------- | -------- | ---- | - |
| Trains | 0x2F | WORD | Set variant engine ID |
| RoadVehs | 0x26 | WORD | Set variant engine ID |
| Ships | 0x20 | WORD | Set variant engine ID |
| Aircraft | 0x20 | WORD | Set variant engine ID |
| Trains | 0x30 | DWORD | Set extra flags |
| RoadVehs | 0x27 | DWORD | Set extra flags |
| Ships | 0x21 | DWORD | Set extra flags |
| Aircraft | 0x21 | DWORD | Set extra flags |

### Extra Flags ###

Extra flags is a set of bits, much like misc flags but that is full up.

| Value | Bit | Feature |
| ----- | --- | ------- |
| 0x01 | 0 | Disable "New Vehicle" news message for this engine |
| 0x02 | 1 | Disable "Exclusive Preview" for this engine |
| 0x04 | 2 | Variants - Include this variant when primary engine has "Exclusive Preview" |
| 0x08 | 3 | Variants - (Attempt to) Synchronize reliability the  primary engine. |

This is implemented by https://github.com/OpenTTD/nml/pull/272

## Limitations

Probably some, let's find out?

## Checklist for review

### DIFF
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -125,6 +125,10 @@ class ReplaceVehicleWindow : public Window {
 
 			/* Add variants if not folded */
 			if ((item.flags & (EngineDisplayFlags::HasVariants | EngineDisplayFlags::IsFolded)) == EngineDisplayFlags::HasVariants) {
+				/* Add this engine again as a child */
+				if ((item.flags & EngineDisplayFlags::Shaded) == EngineDisplayFlags::None) {
+					target.emplace_back(item.engine_id, item.engine_id, EngineDisplayFlags::None, indent + 1);
+				}
 				AddChildren(source, target, item.engine_id, indent + 1, side);
 			}
 		}
@@ -617,7 +621,22 @@ public:
 				uint i = this->vscroll[click_side]->GetScrolledRowFromWidget(pt.y, this, widget);
 				size_t engine_count = this->engines[click_side].size();
 
-				EngineID e = engine_count > i ? this->engines[click_side][i].engine_id : INVALID_ENGINE;
+				EngineID e = INVALID_ENGINE;
+				if (i < engine_count) {
+					const auto &item = this->engines[click_side][i];
+					const Rect r = this->GetWidget<NWidgetBase>(widget)->GetCurrentRect().Shrink(WidgetDimensions::scaled.matrix).WithWidth(WidgetDimensions::scaled.hsep_indent * (item.indent + 1), _current_text_dir == TD_RTL);
+					if ((item.flags & EngineDisplayFlags::HasVariants) != EngineDisplayFlags::None && IsInsideMM(r.left, r.right, pt.x)) {
+						/* toggle folded flag on engine */
+						assert(item.variant_id != INVALID_ENGINE);
+						Engine *engine = Engine::Get(item.variant_id);
+						engine->display_flags ^= EngineDisplayFlags::IsFolded;
+
+						InvalidateWindowData(WC_REPLACE_VEHICLE, (VehicleType)this->window_number, 0); // Update the autoreplace window
+						InvalidateWindowClassesData(WC_BUILD_VEHICLE); // The build windows needs updating as well
+						return;
+					}
+					if ((item.flags & EngineDisplayFlags::Shaded) == EngineDisplayFlags::None) e = item.engine_id;
+				}
 
 				/* If Ctrl is pressed on the left side and we don't have any engines of the selected type, stop autoreplacing.
 				 * This is most common when we have finished autoreplacing the engine and want to remove it from the list. */
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -33,11 +33,11 @@
 
 #include "safeguards.h"
 
-void DrawEngineList(VehicleType type, const Rect &r, const GUIEngineList *eng_list, uint16 min, uint16 max, EngineID selected_id, bool show_count, GroupID selected_group);
+void DrawEngineList(VehicleType type, const Rect &r, const GUIEngineList &eng_list, uint16 min, uint16 max, EngineID selected_id, bool show_count, GroupID selected_group);
 
-static bool EngineNumberSorter(const EngineID &a, const EngineID &b)
+static bool EngineNumberSorter(const GUIEngineListItem &a, const GUIEngineListItem &b)
 {
-	return Engine::Get(a)->list_position < Engine::Get(b)->list_position;
+	return Engine::Get(a.engine_id)->list_position < Engine::Get(b.engine_id)->list_position;
 }
 
 /**
@@ -113,6 +113,22 @@ class ReplaceVehicleWindow : public Window {
 		return true;
 	}
 
+	void AddChildren(const GUIEngineList &source, GUIEngineList &target, EngineID parent, int indent, int side)
+	{
+		for (const auto &item : source) {
+			if (item.variant_id != parent || item.engine_id == parent) continue;
+
+			const Engine *e = Engine::Get(item.engine_id);
+			EngineDisplayFlags flags = item.flags;
+			if (e->display_last_variant != INVALID_ENGINE) flags &= ~EngineDisplayFlags::Shaded;
+			target.emplace_back(e->display_last_variant == INVALID_ENGINE ? item.engine_id : e->display_last_variant, item.engine_id, flags, indent);
+
+			/* Add variants if not folded */
+			if ((item.flags & (EngineDisplayFlags::HasVariants | EngineDisplayFlags::IsFolded)) == EngineDisplayFlags::HasVariants) {
+				AddChildren(source, target, item.engine_id, indent + 1, side);
+			}
+		}
+	}
 
 	/**
 	 * Generate an engines list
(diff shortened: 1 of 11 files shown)


LC-Zorg:
> 1. Avoid spamming news message for new-engine types.
> 2. Disallow exclusive preview for variants?
> 3. Reliability syncing?

I think that it should all depend on the author of the set. Variants can only be a different livery, but they can also be very different machines (an example of many versions of the Boeing 737 and 747).
It could also be that the new variant introduces a new livery that will be available for many vehicles, so there should be a message, but at the same time, the reliability should remain the same as for the rest of the variants because technically it is still the same vehicle.
This may not be formally correct, but maybe a vehicle tag system for variants could be added. Something like: "no news", "no test" or "news", "test" etc.

Regarding the reliability itself, I still strongly believe that it should be determined by the sets authors and not randomly. In fact, it is very common for one vehicle or even a version of it to have significantly lower or higher reliability than the others. This characteristic is currently not transferable to the game world. I don't know if that's related. I'm writing just in case, because it's something worth fixing in this game.

PeterN:
Rebased and added extra flags as documented in the first post.

andythenorth:
Test grf. Doesn't use any flags yet.  Variants available from 1960s onwards.

PeterN:
I've noticed reliability is very samey at the moment, regardless of the sync setting, so trying to improve the seed.

andythenorth:
Horse with flags - bits 0, 1 and 3 are now used by variants.
[iron-horse-2.63.1-19-gac530925.zip](https://github.com/OpenTTD/OpenTTD/files/10202229/iron-horse-2.63.1-19-gac530925.zip)

arikover:
I tested this new feature, and it is quite nice.
I noticed 2 problems, which I think are bugs:

1. Cargo filter now only works for trains.
Steps to reproduce:
- Open any non-train vehicle list (RVs for example)
- Filter by any cargo
What should happen: list shoud only contain vahicle that can be refitted to the chosen cargo
What happens: list stays unchanged
When: for road vehicles, boats and planes. Trains are filtered correctly

2. Engine Variant Group GUI: when choosing the first variant of a group, the container is also highlighted, see animated GIF
GUI_Problem
(nothing important really)

PeterN:
The cargo filter issue is now fixed.

andythenorth:
Added a newer Horse with more variants: 
[iron-horse-2.63.1-40-g8e590ea5.zip](https://github.com/OpenTTD/OpenTTD/files/10273139/iron-horse-2.63.1-40-g8e590ea5.zip)

src/autoreplace_gui.cpp
```diff
@@ -113,19 +113,39 @@ class ReplaceVehicleWindow : public Window {
 		return true;
 	}
 
+	void AddChildren(const GUIEngineList &source, GUIEngineList &target, EngineID parent, int indent, int side)
+	{
+		for (auto &item : source) {
```
LordAro: const ?

src/build_vehicle_gui.cpp
```diff
@@ -1078,6 +1097,27 @@ struct BuildVehicleWindow : Window {
 		}
 	}
 
+	void AddChildren(const GUIEngineList &source, EngineID parent, int indent)
+	{
+		for (auto &item : source) {
```
LordAro: const?

src/engine.cpp
```diff
@@ -655,7 +679,20 @@ void StartupOneEngine(Engine *e, Date aging_date)
...
+	}
+
+	SetRandomSeed(_settings_game.game_creation.generation_seed ^ seed ^
+	              (re->index << 16) ^
+				  (re->info.base_intro << 12) ^
+				  (re->info.decay_speed << 8) ^
+				  (re->info.lifelength << 4) ^
+				  re->info.retire_early ^
+	              e->type ^
+	              e->GetGRFID());
```
LordAro: some very strange line breaks and indenting here

glx22: maybe put the operators at begin of line

glx22:
A test build is available on https://www.openttd.org/downloads/openttd-branches/pr10220/latest

2TallTyler (approved):
As we've discussed on Discord, this PR may be suited for a "merge now for widespread player testing, and the author has promised to fix any bugs" approach. 

There have been some player and NewGRF author disagreements on Discord about alternate approaches, including continuing to use cargo subtypes to allow free depot repainting. In my opinion, variants does not preclude future changes including the hypothetical livery menu that has previously eluded us, because even in that case it has legitimate use cases for other vehicle similarities like different size trailers, single or multiple engine consists (e.g. double engines are used in Iron Horse to autoreplace one big steam loco with two smaller electrics), ships with different deck fittings, etc. 

Plus, the UI can always be iterated upon in the future without changing GRF spec — for example I'd like to enable buying a random variant (Ctrl+click on the buy button?) and other might want a button to manually change the variant inside a depot. 

Merry Christmas, @andythenorth 😄

Brickblock1:
Would it be possible to add some form of shortcut to close all currently open groups?

andythenorth:
I will do the GRF spec status / grf wiki updates.

andythenorth:
I did grf wiki and status tracker updates.  They might bear some review, but they're basically done.

NyanGoat:
Could/should this be expanded to include categories?
Same implementation as this, just it would be a non-purchaseable item, mostly there to provide some graphics and text for said category.

Example could be grouping trains under nationality, or under purpose.
So you could have f.ex the parent group be "American", "British", "German" and then place said trains under there.
Or, in the example of Horse, you could have "General Purpose" and then said trains under there etc.

NyanGoat:
Just as an example to what I mean above:
image

Just a quick and dirty hackjob, and even though you can fake the result like this, a proper implementation where the default entry would just be a category and not an actual purchasable vehicle would be to prefer.

I also notice that, as I set the "category" to be VEHICLE_NEVER_EXPIRES, even though the next two vehicles have an expiration date, they don't seem to expire, not sure if it's related to variant_group inheritance or something else I might have broken though.

PeterN:
You could try setting 'group' engine as unbuildable (no climates perhaps), and it might still appear (but unselectable) in the list as the other buildable engines force it to appear.

Engine life should only be affected if the sync reliability with parent flag is set, and even then it should still take account of the vehicle's own life length. Hmm.

NyanGoat:
Thought I had tried NO_CLIMATES, but seemingly not, as I had not removed the bitmask() part of the string.
Did try it now though, and when I click the item in the menu, the game crashes.
image

The same happens if you also try to make the vehicle otherwise unavailable, by f.ex setting the introduction date in the future.

For the expiration, I tested it again now by changing the lifetime of the vehicles to a shorter time and just fast forwarding, and they do expire as normal, so that was a non-issue, or at least I am unable to reproduce it at the moment.

PeterN:
Well... thanks, I will look into the assertion!

PeterN:
@NyanGoat Could you provide the NewGRF used for this so I can replicate?

NyanGoat:
This is with version 20221229-master-gf90156f74c
Add attached GRF to game.
Start in year 1940.
Open Vehicles, Available Vehicles.
Click One of the Mail Truck lines (for some reason there are two after i changed to NO_CLIMATE, I think it's related to it adding one entry per item in the group, as there should be two available at 1940. Starting a game at 1910 only shows one entry.)

[moprv_variant_test.zip](https://github.com/OpenTTD/OpenTTD/files/10326402/moprv_variant_test.zip)

PeterN:
Thanks, found it :-)

NyanGoat:
Thanks for the quick fix :)

Tested your change locally and I see that setting the group/category vehicle as unavailable does work as mentioned, where it shows the group/category and the two intended vehicles only.

But, once you buy one of the two vehicles, it replaces the default chosen vehicle with the recently bought vehicle (which from what I remember was intended in the spec).
Would it be possible to perhaps add one more flag to the extra_flags spec, f.ex VEHICLE_FLAG_FORCE_DEFAULT_VARIANT or such to make a chosen variant the default variant in the purchase list for that group, if desired?

George-VB:
If the vehicle, that is a parent vehicle for the current one, is defined below in the grf, the list is not displayed correctly. If you reapply the GRF, the list becomes fixed
image
image
image
branch 98 
https://github.com/George-VB/xussrset
[xussr.zip](https://github.com/OpenTTD/OpenTTD/files/10413909/xussr.zip)

George-VB:
> You could try setting 'group' engine as unbuildable (no climates perhaps), and it might still appear (but unselectable) in the list as the other buildable engines force it to appear.

In case group engine is unavailable via no climates, it's additional_text is not displayed, even when provided. Could it be changed?

PR #10225, opened 2022-12-07
Fix #10223: Crash when vehicle cloning fails on order cloning.

### LINKED ISSUE #10223: [Crash]: Error: Assertion failed
### Version of OpenTTD

13.0-beta2, nightly(20221205)

### Steps to reproduce

1. Click on Clone train button and then on train which to clone
2. Crashed with screenshot below
2022-12-07 15_39_43-Error!

### Bug location

_/OpenTTD/blob/master/src/core/pool_type.hpp_
```
inline Titem *Get(size_t index)
{
  assert(index < this->first_unused);
  return this->data[index];
}
```

### How to solve it

1. After little bit testing I found out it is crashing when you are trying to clone a train, which has recently deleted station on the route list

### Upload crash files

1. Actually its showing the dialog box and when clicking Ok, its still showing again and it is not possible to go back to the game and no crash logs are created.

michicc:
## Motivation / Problem

See #10223. Cloning a vehicle can crash when cloning the order list fails.

## Description / Limitations

This was caused by a misconversion in the command rewrite. #10223 mentions another line below, but that is correct as `CheckCompanyHasMoney` actually modifies the `CommandCost` parameter.

## Checklist for review

### DIFF
--- a/src/vehicle_cmd.cpp
+++ b/src/vehicle_cmd.cpp
@@ -958,7 +958,7 @@ std::tuple<CommandCost, VehicleID> CmdCloneVehicle(DoCommandFlag flags, TileInde
 		if (result.Failed()) {
 			/* The vehicle has already been bought, so now it must be sold again. */
 			Command<CMD_SELL_VEHICLE>::Do(flags, w_front->index, true, false, INVALID_CLIENT_ID);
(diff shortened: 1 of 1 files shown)


glx22 (approved):
We should maybe recheck all #9725 conversions, but it's a tedious task.

PR #10227, opened 2022-12-08
Fix 3c047b1: AIGroup.GetProfitLastYear could get values different than those displayed in GUI

SamuXarick:
## Motivation / Problem
`AIGroup.GetProfitLastYear` was only accounting profits of vehicles which had an age higher than `VEHICLE_PROFIT_MIN_AGE`, which is different than what is displayed on group_gui widget `WID_GL_INFO`

`GroupStatistics::profit_last_year` is only used for displaying the colour of the profit icons in the vehicle list window, but for the AI API method, it should be expected to account for all vehicles in the group, and it doesn't.

## Description
The proposed solution is to create another `GroupStatistics` profit that includes all vehicles and which would match the value  from `WID_GL_INFO`:
The original `GroupStatistics::profit_last_year` is renamed to `GroupStatistics::profit_last_year_min_age`, and a second `GroupStatistics::profit_last_year` replaces the original.
The resulting `GroupStatistics::profit_last_year_min_age` is to be used for the icon purposes in vehicle group lists and the new `GroupStatistics::profit_last_year` is to be used for the `AIGroup.GetProfitLastYear` method. No actual changes to the AI code is necessary then.
For clarity, `GroupStatistics::num_profit_vehicle` is renamed to `GroupStatistics::num_vehicle_min_age` as well as a few other items related to this, to try make it easier to understand what they refer to.

With the addition of the new profit, autoreplace posed an issue when trying to copy group and vehicle statistics from the old vehicle to the new vehicle. The group was getting the last year profit of the new vehicle which was still 0, before the profit was updated over from the old vehicle. The end result after autoreplace was completed, was that the group statistics didn't account for the missing profit. It was only copied between vehicles, but not when updating the group. This is solved by cirurgically updating the last year profits of the affected groups in `CopyVehicleConfigAndStatistics`

## Checklist for review

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -24,13 +24,14 @@ extern GroupPool _group_pool; ///< Pool of groups.
 /** Statistics and caches on the vehicles in a group. */
 struct GroupStatistics {
 	uint16 num_vehicle;                     ///< Number of vehicles.
+	Money profit_last_year;                 ///< Sum of profits for all vehicles.
 	uint16 *num_engines;                    ///< Caches the number of engines of each type the company owns.
 
 	bool autoreplace_defined;               ///< Are any autoreplace rules set?
 	bool autoreplace_finished;              ///< Have all autoreplacement finished?
 
 	uint16 num_profit_vehicle;              ///< Number of vehicles considered for profit statistics;
-	Money profit_last_year;                 ///< Sum of profits for all vehicles.
+	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
 
 	GroupStatistics();
 	~GroupStatistics();
@@ -41,6 +42,7 @@ struct GroupStatistics {
 	{
 		this->num_profit_vehicle = 0;
 		this->profit_last_year = 0;
+		this->profit_last_year_min_age = 0;
 	}
 
 	void ClearAutoreplace()
@@ -105,7 +107,7 @@ static inline bool IsAllGroupID(GroupID id_g)
 uint GetGroupNumEngines(CompanyID company, GroupID id_g, EngineID id_e);
 uint GetGroupNumVehicle(CompanyID company, GroupID id_g, VehicleType type);
 uint GetGroupNumProfitVehicle(CompanyID company, GroupID id_g, VehicleType type);
-Money GetGroupProfitLastYear(CompanyID company, GroupID id_g, VehicleType type);
+Money GetGroupProfitLastYearMinAge(CompanyID company, GroupID id_g, VehicleType type);
 
 void SetTrainGroupID(Train *v, GroupID grp);
 void UpdateTrainGroupID(Train *v);
--- a/src/group.h
+++ b/src/group.h
@@ -23,15 +23,13 @@ extern GroupPool _group_pool; ///< Pool of groups.
 
 /** Statistics and caches on the vehicles in a group. */
 struct GroupStatistics {
-	uint16 num_vehicle;                     ///< Number of vehicles.
 	Money profit_last_year;                 ///< Sum of profits for all vehicles.
+	Money profit_last_year_min_age;         ///< Sum of profits for vehicles considered for profit statistics.
 	uint16 *num_engines;                    ///< Caches the number of engines of each type the company owns.
-
+	uint16 num_vehicle;                     ///< Number of vehicles.
+	uint16 num_vehicle_min_age;             ///< Number of vehicles considered for profit statistics;
 	bool autoreplace_defined;               ///< Are any autoreplace rules set?
 	bool autoreplace_finished;              ///< Have all autoreplacement finished?
-
-	uint16 num_vehicle_min_age;             ///< Number of vehicles considered for profit statistics;
-	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
 
 	GroupStatistics();
 	~GroupStatistics();
--- a/src/group.h
+++ b/src/group.h
@@ -57,6 +57,7 @@ struct GroupStatistics {
 
 	static void CountVehicle(const Vehicle *v, int delta);
 	static void CountEngine(const Vehicle *v, int delta);
+	static void AddProfitLastYear(const Vehicle *v);
 	static void VehicleReachedProfitAge(const Vehicle *v);
 
 	static void UpdateProfits();
--- a/src/group.h
+++ b/src/group.h
@@ -30,7 +30,7 @@ struct GroupStatistics {
 	bool autoreplace_defined;               ///< Are any autoreplace rules set?
 	bool autoreplace_finished;              ///< Have all autoreplacement finished?
 
-	uint16 num_profit_vehicle;              ///< Number of vehicles considered for profit statistics;
+	uint16 num_vehicle_min_age;             ///< Number of vehicles considered for profit statistics;
 	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
 
 	GroupStatistics();
@@ -40,8 +40,9 @@ struct GroupStatistics {
 
 	void ClearProfits()
 	{
-		this->num_profit_vehicle = 0;
 		this->profit_last_year = 0;
+
+		this->num_vehicle_min_age = 0;
 		this->profit_last_year_min_age = 0;
 	}
 
@@ -58,7 +59,7 @@ struct GroupStatistics {
 	static void CountVehicle(const Vehicle *v, int delta);
 	static void CountEngine(const Vehicle *v, int delta);
 	static void AddProfitLastYear(const Vehicle *v);
-	static void VehicleReachedProfitAge(const Vehicle *v);
+	static void VehicleReachedMinAge(const Vehicle *v);
 
 	static void UpdateProfits();
 	static void UpdateAfterLoad();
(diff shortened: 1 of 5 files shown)


2TallTyler (commented):
This could potentially break AIs which use group profits to determine whether a route is successful, using the minimum age for its intended purpose to let vehicles settle into a route before determining their profitability.

(Full disclaimer: I've never written an AI nor looked at how they work.)

What would you say to these concerns?

SamuXarick:
> This could potentially break AIs which use group profits to determine whether a route is successful, using the minimum age for its intended purpose to let vehicles settle into a route before determining their profitability.
> 
> (Full disclaimer: I've never written an AI nor looked at how they work.)
> 
> What would you say to these concerns?

Documentation doesn't say which vehicles in the group it refers to, only "the profit of a group".
https://docs.openttd.org/ai-api/classAIGroup.html#a9514b20e3fe704bd548352cb434527ea

It's cousin `GetProfitThisYear` is documented in the same fashion.

https://github.com/OpenTTD/OpenTTD/blob/3af2c7fff6e4e6799392ce754d0f541b435fe8f0/src/script/api/script_group.cpp#L163-L184
Code for both functions are different. `GetProfitLastYear` looks up from the group statistics cache, while `GetProfitThisYear` manually iterates every vehicle that belongs to the group, without any criteria regarding age.

I've also checked from Bananas which AIs use `GetProfitLastYear`, and it's only RailwAI.
```
		local list = AIVehicleList_Group(group)
		local vehicle = list.Begin()
		//Do not remove group if vehicle age is less than ~2 years
		if (AIVehicle.GetAge(vehicle) < 800)
			continue
		local profit = (AIGroup.GetProfitThisYear(group) + AIGroup.GetProfitLastYear(group))/list.Count()
		if (profit < lowestProfitPerVehicle) {
			lowestProfitPerVehicle = profit
			leastProfitableGroup = group
		}
```

It is being used together with `GetProfitThisYear` as a sum of both in order to get the average profit per vehicle of a group. `AIVehicleList_Group` is a list of all vehicles in the group. Nothing suggests that profits of vehicles below Mininmum age should be unnacounted for.
The code seems to skip groups in which the first vehicle found in a group is younger than 800 days, though.

2TallTyler (dismissed):
If RailwayAI is the only user of this API and it has a separate check for vehicles younger than 800 days (VEHICLE_PROFIT_MIN_AGE is 730 days), this sounds safe to me.

src/group.h
```diff
@@ -24,13 +24,14 @@ extern GroupPool _group_pool; ///< Pool of groups.
...
 	uint16 num_vehicle;                     ///< Number of vehicles.
+	Money profit_last_year;                 ///< Sum of profits for all vehicles.
 	uint16 *num_engines;                    ///< Caches the number of engines of each type the company owns.
 
 	bool autoreplace_defined;               ///< Are any autoreplace rules set?
 	bool autoreplace_finished;              ///< Have all autoreplacement finished?
 
 	uint16 num_profit_vehicle;              ///< Number of vehicles considered for profit statistics;
-	Money profit_last_year;                 ///< Sum of profits for all vehicles.
+	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
```
rubidium42: Something to consider about the workings of C++. The actual data in this structures and classes are placed in the order they are written in the code. However, the access of (some) memory by the CPU is often much faster at multiples of their register size. For this the compiler just adds a bit of unused memory, so things get aligned again.

In this case, on x86-64 in the original code, there will be 6 unused bytes between `num_vehicle` and `num_engines` and another 4 bytes between `num_profit_vehicle` and `profit_last_year`. So, preferably the fields are from big to small, so there is no need to create those gaps.
```
struct GroupStatistics {
	Money profit_last_year;                 ///< Sum of profits for all vehicles.
	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
	uint16_t *num_engines;                  ///< Caches the number of engines of each type the company owns.
	uint16_t num_vehicle;                   ///< Number of vehicles.
	uint16_t num_profit_vehicle;            ///< Number of vehicles considered for profit statistics;
	bool autoreplace_defined;               ///< Are any autoreplace rules set?
	bool autoreplace_finished;              ///< Have all autoreplacement finished?
};
```
This uses the same amount of memory (on x86-64) but with an extra field for the money you want to store.

rubidium42: Naming of these variables does need to be improved/unified in some manner.
There is `num_vehicle` and `num_profit_vehicle` vs `profit_last_year` and `profit_last_year_min_age`. This brings a whole load of questions to me, as `num_profit_vehicles` is not linked to `profit_last_year` though the comment somewhat implies that.
So, some sort of unified naming should be devised so it becomes easier to link the two sets of variables together. I'm not quite sure what to use though. However, I do not really like `min_age` to be used as that might not be all (now or in the future). In other words, it should be about the concept of considering it for performance ratings and not the way the consideration has been implemented.

src/vehicle_base.h
```diff
@@ -749,6 +749,8 @@ struct Vehicle : VehiclePool::PoolItem<&_vehicle_pool>, BaseVehicle, BaseConsist
 
 		this->profit_this_year = src->profit_this_year;
 		this->profit_last_year = src->profit_last_year;
+
+		GroupStatistics::AddProfitLastYear(this);
```
rubidium42: I don't see anything removing profit, so why would autoreplace/autorenew add profit to the statistics? It's not removed by deleting the old vehicle as far as I can see. So this looks like you're going to double count the profits of last year.
At best you will create a difference in the statistics between a game that is loaded from savegame after the autrenew and a game that has been running during the autorenew. At worst this will cause a desync if something uses these `GroupStatistics` in the main game logic.

rubidium42 (commented):
The underlying `GroupStatistics` code looks way too complicated for what it needs to be doing as far as I can see. To me there should be one function to add/remove a vehicle from statistics and one function to clear and repopulate the statistics making use of the previous function.

src/vehicle_base.h
SamuXarick: I created a debug branch to test the autoreplace issue: https://github.com/SamuXarick/OpenTTD/tree/group-get-profit-last-year-debug

Savegame: [autoreplace test.zip](https://github.com/OpenTTD/OpenTTD/files/5821631/autoreplace.test.zip)
AI: [autoreplace test.zip](https://github.com/OpenTTD/OpenTTD/files/5821636/autoreplace.test.zip)

How to debug test:
1 - Savegame contains the test AI that is constantly calling AIGroup.GetProfitLastYear.
2 - AIGroup.GetProfitLastYear has two methods for retrieving the last year profit. One computes manually by iterating all vehicles in the group, the other retrieves directly from the group statistics last year profit. It compares the result of both if they match with an assert.
3 - Load the savegame and fastforward, wait for the vehicle to become old and trigger the autorenew.

SamuXarick: https://github.com/OpenTTD/OpenTTD/blob/921c6591f95d3c46432e7c3c7cc48ac8a629d9c3/src/train_cmd.cpp#L1419-L1427

Okay, I see what you mean. I'm investigating if there's another way to update the profit_last_year.

Train code also use this function, and it is now broken.

SamuXarick: And it is now fixed again. I moved it away from `CopyVehicleConfigAndStatistics` in `vehicle_base.h` to 
`CopyHeadSpecificThings` in `autoreplace_cmd.cpp`

https://github.com/OpenTTD/OpenTTD/blob/dd6bf061cdb0305a993bcdfe764a966ffb8e1029/src/autoreplace_cmd.cpp#L425

src/group.h
SamuXarick: I tried to come up with a better naming. `num_vehicle`/`profit_last_year` are related to each other, and `num_vehicle_min_age`/`profit_last_year_min_age` is the other duo.
```
/** Statistics and caches on the vehicles in a group. */
struct GroupStatistics {
	uint16 num_vehicle;                     ///< Number of vehicles.
	Money profit_last_year;                 ///< Sum of profits for all vehicles.
	uint16 *num_engines;                    ///< Caches the number of engines of each type the company owns.

	bool autoreplace_defined;               ///< Are any autoreplace rules set?
	bool autoreplace_finished;              ///< Have all autoreplacement finished?

	uint16 num_vehicle_min_age;             ///< Number of vehicles considered for profit statistics;
	Money profit_last_year_min_age;         ///< Sum of profits for considered vehicles.
```

rubidium42 (changes requested):
The minimum age logic bothers me immensely. If I build a vehicle on 01-12-2020, then at 01-01-2022 the profit of 2021 is not considered but on 01-12-2022 it is considered and the profit of the group last year rises? In my opinion it should be relatively simple, when the vehicle is at least 1 year old on 01-01 of that year the profit of the previous year should be counted for the whole year. The profit of the previous year of the vehicle is not going to change, so it would be weird to not include it until a later date that year.
But that is beyond the scope of this story. 

I did do some testing and wrote some (ugly) cache validation and everything seemed to work just fine.

Only thing that remains is reordering the fields in GroupStatistics: profit_last_year, profile_last_year_min_age, num_engines, num_vehicle, num_vehicle_min_age, autoreplace_defined, autoreplace_finished. That way less memory gets wasted.

rubidium42 (approved):
I think we should squash during merge, as some changes are happening in a weird order.
For example the second commit duplicating some code, and then the third commit introducing a helper function to remove the duplication. Instead of creating the helper function first, and then then using it.

RailwAI:
> If RailwayAI is the only user of this API and it has a separate check for vehicles younger than 800 days (VEHICLE_PROFIT_MIN_AGE is 730 days), this sounds safe to me.

I expect that this bugfix can only be beneficial for my AI, I never noticed that the function return value was different from the GUI, but assumed it was the same. I only excluded groups as a whole, but assumed that every vehicle in the group would be taken into account when I don't exclude the group.

PR #10229, opened 2022-12-10
Fix #10224: Don't fast forward after autosave if tab key is no longer pressed

### LINKED ISSUE #10224: [Bug]: Fast forward is toggling on after autosave even if Tab isn't pressed
### Version of OpenTTD

12.2, Steam, Windows 10

### Expected result

Fast forward shouldn't toggle on after autosave if `Tab` isn't pressed.

### Actual result

Fast forward is toggling on after autosave even if `Tab` button isn't pressed.

### Steps to reproduce

- Load into any game
- Press `Tab` for fast forward
- When autosave starts, release `Tab` button
- Once autosave will be completed, fast forward will be toggled on despite the fact that `Tab` wasn't pressed.

https://user-images.githubusercontent.com/22667749/206229839-4565d55f-eb77-4d3e-9c47-4982407f43de.mp4

AaronKatzin:
If fast forward is via key press

## Motivation / Problem

Fix #10224

## Description

Upon autosave, we save game speed in order to restore fast forward upon completion of autosave.
This caused a bug in the case where fast forward was due to key press which was released during the autosave.
So if fast forward is due to key press, save normal game speed. The graphics driver will take care of game speed after the autosave completes just like it does during game-play based on the key being pressed or not. 

## Limitations

n/a

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -2920,9 +2920,11 @@ static inline void ClearSaveLoadState()
  */
 static void SaveFileStart()
 {
-	// If fast forward is via key, save normal game speed.
-	// If the key continues to be pressed, the graphics driver will continue to fast forward the game.
-	// If the key is released during autosave, the normal speed will be restored.
+	/**
+	 * If fast forward is via key, save normal game speed.
+	 * If the key continues to be pressed, the graphics driver will continue to fast forward the game.
+	 * If the key is released during autosave, the normal speed will be restored.
+	 */
 	_sl.game_speed = VideoDriver::GetInstance()->isFastForwardViaKey() ? 100 : _game_speed;
 	_game_speed = 100;
 	SetMouseCursorBusy(true);
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -2925,7 +2925,7 @@ static void SaveFileStart()
 	 * If the key continues to be pressed, the graphics driver will continue to fast forward the game.
 	 * If the key is released during autosave, the normal speed will be restored.
 	 */
-	_sl.game_speed = VideoDriver::GetInstance()->isFastForwardViaKey() ? 100 : _game_speed;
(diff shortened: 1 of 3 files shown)


src/saveload/saveload.cpp
```diff
@@ -2920,7 +2920,10 @@ static inline void ClearSaveLoadState()
  */
 static void SaveFileStart()
 {
-	_sl.game_speed = _game_speed;
+	// If fast forward is via key, save normal game speed.
+	// If the key continues to be pressed, the graphics driver will continue to fast forward the game.
+	// If the key is released during autosave, the normal speed will be restored.
```
glx22: Multiline comment uses `/* */`

src/video/video_driver.hpp
```diff
@@ -222,6 +222,9 @@ class VideoDriver : public Driver {
 		bool unlock; ///< Stores if the lock did anything that has to be undone.
 	};
 
+	bool isFastForwardViaKey(){
+		return fast_forward_via_key;
+	}
```
glx22:
```suggestion
	bool IsFastForwardViaKey()
	{
		return this->fast_forward_via_key;
	}
	
```
Applying https://wiki.openttd.org/en/Development/Coding%20style

src/saveload/saveload.cpp
```diff
@@ -2920,7 +2920,12 @@ static inline void ClearSaveLoadState()
  */
 static void SaveFileStart()
 {
-	_sl.game_speed = _game_speed;
+	/**
+	 * If fast forward is via key, save normal game speed.
+	 * If the key continues to be pressed, the graphics driver will continue to fast forward the game.
+	 * If the key is released during autosave, the normal speed will be restored.
+	 */
+	_sl.game_speed = VideoDriver::GetInstance()->isFastForwardViaKey() ? 100 : _game_speed;
```
glx22:
```suggestion
	_sl.game_speed = VideoDriver::GetInstance()->IsFastForwardViaKey() ? 100 : _game_speed;
```
Follow-up of the other suggested change.

2TallTyler:
Superseded by #10230 - thanks for the PR! 🙂

PR #10230, opened 2022-12-10
Fix #10224: Don't fiddle with fast-forward when saving.

### LINKED ISSUE #10224: [Bug]: Fast forward is toggling on after autosave even if Tab isn't pressed

PeterN:
## Motivation / Problem

See #10224. Fast-forward can get stuck on if tab is released while saving.

## Description

The original comment about saving with fast-forward on was written 18 years ago, and predates lots of changes to how saveload work.

As an alternative to #10229, I propose that it's probably no longer relevant, and instead of adding workarounds, we can simply remove it.

## Limitations

Not actually benchmarked (saving!?) to see if it is relevant...

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -210,7 +210,6 @@ struct SaveLoadParams {
 	StringID error_str;                  ///< the translatable error message to show
 	char *extra_msg;                     ///< the error message
 
-	uint16 game_speed;                   ///< The game speed when saving started.
 	bool saveinprogress;                 ///< Whether there is currently a save in progress.
 };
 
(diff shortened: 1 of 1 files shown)


LordAro (commented):
What happens if ffwd is such that another autosave happens while the previous one is still being saved?

rubidium42:
> What happens if ffwd is such that another autosave happens while the previous one is still being saved?

Basically nothing; https://github.com/OpenTTD/OpenTTD/blob/8063fcb6e03f4a2d20d0b28f6b93ca9a57fd923f/src/saveload/saveload.cpp#L3244

Though the question is whether this slowing down behaviour and (attempting) to not skip a save is wanted or not. For doing things as fast as possible I would reckon the slow downs are annoying, except when you want to get to the last savegame before some crash or something fairly expediously. The former case can currently be handled by not enabling autosave, if you do not want the slow downs.

One the other hand, arguably CPUs of most machines running OpenTTD nowadays will be multi-core, so there is likely less conflict for resources for compressing the savegame than there was back when parallel compression got added.

So, less is more?

michicc (approved):

andythenorth:
I have done quite a bit of FFWD speed-testing for performance (comparing hardware acceleration on / off, different blitters etc).

For the 'go as fast as possible' case, autosave would usually be turned off.

PR #10233, opened 2022-12-13
Fix #10208: allow to use specific underlay for road/tram tunnels

### LINKED ISSUE #10208: [Bug]: NRT road tunnels lack custom overlays
### Version of OpenTTD

13.0 b2

### Expected result

Creating a good-looking NRT road tunnel is almost impossible, because the game always overlays a standard road tile on top of the tunnel underlay.

Untitled

Image shows a rail tunnel, a road tunnel as built (with the roadtile overlayed), and the road tunnel as drawn (mocked in). There is currently no way to change or hide the overlayed road tile.

### Actual result

Newgrf authors should be able to specify a custom overlay (as with Railtype action 3 cargo 03). This would not break compatibility with existing newgrfs (which will continue to work as before if the action 3 is not defined), nor would it cause conflicts between road + tram each providing graphics for the tile; a custom tram overlay will simply appear above a custom road overlay, as with the defaults.

see https://github.com/OpenTTD/nml/issues/79 for some further discussion.

### Steps to reproduce

Build a road tunnel with any NRT road grf.

glx22:
## Motivation / Problem
As shown in #10208, drawing road/tram underlay sprites on top of tunnel entrance ground can look very ugly.
And if the underlay is not drawn we're stuck with default road sprite for all tunnels.  

## Description
Disable drawing of underlays for tunnels, as they are intended for plain tiles and can be larger than the available space in tunnel.
Add cargo 03 for roadtype/tramtype Action3 (same as railtype Action3 cargo 03) with 4 sprites in SW, NW, NE, SE order. This allows to "repaint" default road from baseset tunnel entrance sprite to look like the actual road.
Of course road takes precedence over tram as always.

## Checklist for review

### DIFF
--- a/src/tunnelbridge_cmd.cpp
+++ b/src/tunnelbridge_cmd.cpp
@@ -1307,8 +1307,28 @@ static void DrawTile_TunnelBridge(TileInfo *ti)
 			const RoadTypeInfo *road_rti = road_rt == INVALID_ROADTYPE ? nullptr : GetRoadTypeInfo(road_rt);
 			const RoadTypeInfo *tram_rti = tram_rt == INVALID_ROADTYPE ? nullptr : GetRoadTypeInfo(tram_rt);
 			uint sprite_offset = DiagDirToAxis(tunnelbridge_direction) == AXIS_X ? 1 : 0;
+			bool draw_underlay = true;
 
-			DrawRoadOverlays(ti, PAL_NONE, road_rti, tram_rti, sprite_offset, sprite_offset);
+			/* Road underlay takes precedence over tram */
+			if (road_rti != nullptr) {
+				if (road_rti->UsesOverlay()) {
+					SpriteID ground = GetCustomRoadSprite(road_rti, ti->tile, ROTSG_TUNNEL);
+					if (ground != 0) {
+						DrawGroundSprite(ground + tunnelbridge_direction, PAL_NONE);
+						draw_underlay = false;
+					}
+				}
+			} else {
+				if (tram_rti->UsesOverlay()) {
+					SpriteID ground = GetCustomRoadSprite(tram_rti, ti->tile, ROTSG_TUNNEL);
+					if (ground != 0) {
+						DrawGroundSprite(ground + tunnelbridge_direction, PAL_NONE);
+						draw_underlay = false;
+					}
+				}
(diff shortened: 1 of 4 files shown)


src/tunnelbridge_cmd.cpp
```diff
@@ -1308,7 +1308,22 @@ static void DrawTile_TunnelBridge(TileInfo *ti)
...
 			uint sprite_offset = DiagDirToAxis(tunnelbridge_direction) == AXIS_X ? 1 : 0;
 
-			DrawRoadOverlays(ti, PAL_NONE, road_rti, tram_rti, sprite_offset, sprite_offset);
+			/* Road underlay takes precedence over tram */
+			if (road_rti != nullptr) {
+				if (road_rti->UsesOverlay()) {
+					SpriteID ground = GetCustomRoadSprite(road_rti, ti->tile, ROTSG_TUNNEL);
+					if (ground != 0) DrawGroundSprite(ground + tunnelbridge_direction, PAL_NONE);
+				}
+			} else {
```
PeterN, line 1321: What happens if the road type use overlays but doesn't provide tunnel sprites?

glx22: Original tunnel ground (with entrance and road) is still drawn before tunnel underlay, so default road is visible if no tunnel specific sprites are provided.

PikkaBird: As I mentioned at #10208, continuing to display the standard road overlay if the tunnel overlay action 3 is not defined would retain the old behaviour which may be expected by existing newgrfs. OTOH, displaying no overlay if the action 3 is not used would better mirror the rail tunnel spec. Which is better?

glx22: It's easy to restore old behaviour when tunnel underlay sprites are not provided (it's controlled by a single boolean, fixed to `false` for now).
And I agree it might be better for backward compatibility with existing newgrf.

glx22:
Test with many NRT grf and disabled underlay
image
Same save but keeping underlay as fallback if no tunnel specific sprites
image

It looks like NRT authors like to redesign tunnel entrances :)

PeterN (approved):
LGTM now that it is "backwards compatible"

PR #10235, opened 2022-12-13
Fix #10218: Sloped river tiles need water both up and downstream

### LINKED ISSUE #10218: [Bug]: Banks of wide rivers on sloped terrain
### Version of OpenTTD

OpenTTD 13.0-beta2, Windows 8

### Expected result

Normal banks of a river delta:

river00

### Actual result

Banks that appear when there is a slope next to a river delta:

river01

river02

### Steps to reproduce

1. Random world generation with Terrain type: Mountainous, Variety distribution: Medium, Smoothness: Smooth or Rough

2TallTyler:
## Motivation / Problem

When widening rivers, we use a CircularTileSearch to attempt to terraform and make river tiles within the river's radius. This determines the desired slope automatically based on whether the target tile is part of a continuous row of sloped river tiles. This was missing a check to ensure the river is actually sloping down to the sea, so the row of sloped tiles could end up propagating inland, as seen in #10218.

I don't know what's going on with coast tiles being drawn in #10218, but this also avoids that happening. 🙂 

## Description

Ensure that sloped tiles always have water both upstream and downstream from themselves.

New ocean tiles created by river terraforming will flood after river generation is complete, but on non-ocean tiles we can take a shortcut to create river tiles up and/or downstream, if they don't already exist.

Closes #10218.

A second commit moves a `cur_slope` update so it doesn't get run if no terraforming is needed, and gets rid of a `RandomRange(1)` (which always returns 0) that must be vestigial code from the patch's development which somehow snuck past review.

## Limitations

It's still possible for a long river paralleling a coast to have a wide row of sloped tiles down to the sea. But this looks much less wrong than #10218 and is more of a design choice than a bug. 😉 I'm hesitant to write a lot of code to solve this and potentially introduce other edge cases where multiple rivers flow into the ocean near each other. 

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -1176,9 +1176,9 @@ static bool RiverMakeWider(TileIndex tile, void *data)
 				Command<CMD_TERRAFORM_LAND>::Do(DC_EXEC | DC_AUTO, tile, to_change, true);
 			}
 		}
+		/* Update cur_slope after possibly terraforming. */
+		cur_slope = GetTileSlope(tile);
 	}
-	/* Update cur_slope after possibly terraforming. */
-	cur_slope = GetTileSlope(tile);
 
 	/* Sloped rivers need water both upstream and downstream. */
 	if (IsInclinedSlope(cur_slope)) {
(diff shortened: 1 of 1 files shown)


src/landscape.cpp
```diff
@@ -1176,9 +1176,41 @@ static bool RiverMakeWider(TileIndex tile, void *data)
...
+		/* Downstream might be new ocean created by our terraforming, and it hasn't flooded yet. */
+		bool downstream_is_ocean = GetTileZ(downstream_tile) == 0 && (GetTileSlope(downstream_tile) == SLOPE_FLAT || IsSlopeWithOneCornerRaised(GetTileSlope(downstream_tile)));
+
+		/* If downstream is dry, flat, and not ocean, try making it a river tile. */
+		if (!IsWaterTile(downstream_tile) && !downstream_is_ocean) {
+			if (GetTileSlope(downstream_tile) == SLOPE_FLAT) {
+				MakeRiver(downstream_tile, Random());
+			} else {
+				return false;
+			}
```
PeterN: Would be better as `... != SLOPE_FLAT) return false;`, to avoid excess nesting. Same below.

PeterN (approved):

PR #9745, opened 2021-12-13
Fix #9720: Delay start of GS/AI to after loading of savegame

glx22:
Fixes #9720.

## Motivation / Problem
See https://github.com/OpenTTD/OpenTTD/issues/9720#issuecomment-980808344

## Description
During loading, AI/GS saved data is stored in `ScriptConfig`.
AI/GS are started after loading is fully completed.
On AI/GS start, the saved data is put on stack right after initialisation.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

Issue #9720, opened 2021-11-27
[Crash]: Can't start game with CityGrowthLimiter GS


Thundo54:
### Version of OpenTTD

12.1 (compiled on RaspberryPi), aarch64 (RP3B+)

### Steps to reproduce

1. Open Scenario Editor
2. Configure GS CGL 
3.  Save scenario
4.  Start game with new scenario
5.  Save game to run in on Raspberry Pi
6.  Start game on RPi3B+ with command `/opt/openttd-12.1/build/openttd -D -g /root/.openttd/save/autosave/mygame.sav`
7. Crash while starting game (as dedicated server)

### Upload crash files

[crash.zip](https://github.com/OpenTTD/OpenTTD/files/7612391/crash.zip)

glx22:
Please attach your savegame.

frosch123:
I reproduced this, here is a simple scenario that uses 'CityGrowthLimiter v6'
[cgl.zip](https://github.com/OpenTTD/OpenTTD/files/7612549/cgl.zip)

The bug is triggered by the script in town.nut during static initialisation:
```
class Town {
        ...
        paxRequired = GSController.GetSetting("min_transport_pax");
        ...
}
```
* The script wants to read some script configuration setting during loading of the scenario.
* OpenTTD is not yet finished loading the scenario and actually thinks it still is in the main menu (GM_MENU).
* OpenTTD then tries to use the new-game script configuration, instead of the one from the scenario. (this is already wrong)
* However, this new-game script configuration is not available on dedicated servers, so OpenTTD crashes.

The underlying issue in OpenTTD is actually pretty severe:
* When loading savegames/scenarios, AI and GS are started inside the saveload code:
    * AIPLChunkHandler::Load
    * GSDTChunkHandler::Load
* At this point of time, the savegame is not yet completely loaded, AfterLoad did not run yet, ... so the gamestate is still incomplete or even invalid.
* AI/GS code running at this time get in contact with all kind of invalid data. (static initialisation, script's Load() method)

I think to fix this, loading AI/GS needs to be changed significantly:
* AIPLChunkHandler::Load and GSDTChunkHandler::Load have to store the raw script data in some temporary location.
* Loading and starting AI/GS must wait until the end of AfterLoad, when the gamestate is ready.

James103:
I can reproduce this with a custom AI specifically made for this, so it is not just limited to the provided Game Script and save game.
[...]

PR #10241, opened 2022-12-14
Fix #10206: Disable scripts in intro game

### LINKED ISSUE #10206: Assert when loading titlegame with AI
### Version of OpenTTD

13.0-beta2

### Expected result

According to our longtime Title Game Competition rules, AI, GS, and NewGRFs used in a title game are automatically disabled.

I would expect if an AI is present, the AI company is deleted or the game loads the all-water fallback instead, like when a NewGRF is present.

### Actual result

If an AI is used, we get an assertion failure:
`Assertion failed at line 94 of src\ai\ai_config.cpp: strcmp("start_date", name) == 0`

Logs and emergency save:
[crash20221201165815.zip](https://github.com/OpenTTD/OpenTTD/files/10134454/crash20221201165815.zip)

### Steps to reproduce

1. Download the attached savegame and `opntitle.dat` of the same savegame
2. Place `opntitle.dat` in your baseset folder, replacing the file that's already there
3. Start OpenTTD

[title.zip](https://github.com/OpenTTD/OpenTTD/files/10134482/title.zip)

glx22:
Fixes #10206

## Motivation / Problem
Using scripts in intro game can cause crashes.
We always said intro should not use scripts or newgrfs, but we only enforced the newgrf part (with broken save error because disabling newgrf in a running game is never a good idea).
Dedicated servers may also crash because of scripts during the load test as it's done in `GM_MENU` (unrelated but it's funny to see the broken save error when starting a dedicated server with `-g` and a save containing newgrf)

## Description
Discard script data on load and skip starting of scripts (like it's done for network games).
As OpenTTD requires a script to be started for `is_ai` companies, prevent using any random ones (that may cause crash by accessing settings) by forcing Dummy AI for intro.
And don't show AIDebugWindow when an AI dies in intro (as Dummy AI will instant die).

## Limitations
This also solves the crash reported in #9720 but starting scripts before the gamestate is ready is still an issue, and #9745 should handle that when done.

## Checklist for review

### DIFF
--- a/src/ai/ai_scanner.cpp
+++ b/src/ai/ai_scanner.cpp
@@ -10,6 +10,7 @@
 #include "../stdafx.h"
 #include "../debug.h"
 #include "../network/network.h"
+#include "../openttd.h"
 #include "../core/random_func.hpp"
 
 #include "../script/squirrel_class.hpp"
@@ -59,6 +60,11 @@ void AIScannerInfo::RegisterAPI(class Squirrel *engine)
 
 AIInfo *AIScannerInfo::SelectRandomAI() const
 {
+	if (_game_mode == GM_MENU) {
+		Debug(script, 0, "The intro game should not use AI, loading 'dummy' AI.");
+		return this->info_dummy;
+	}
+
 	uint num_random_ais = 0;
 	for (const auto &item : info_single_list) {
 		AIInfo *i = static_cast<AIInfo *>(item.second);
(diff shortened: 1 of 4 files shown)


src/ai/ai_scanner.cpp
```diff
@@ -65,7 +66,7 @@ AIInfo *AIScannerInfo::SelectRandomAI() const
 		if (i->UseAsRandomAI()) num_random_ais++;
 	}
 
-	if (num_random_ais == 0) {
+	if (_game_mode == GM_MENU || num_random_ais == 0) {
 		Debug(script, 0, "No suitable AI found, loading 'dummy' AI.");
```
LordAro: This debug message doesn't really fit now

2TallTyler (approved):

PR #10249, opened 2022-12-17
Minor sprite aligner enhancements

PeterN:
## Motivation / Problem

Using the debug sprite aligner tool does not give very useful feedback on what is happening when aligning a sprite.

## Description

This PR adds two buttons:

- **Centre sprite** toggles whether the sprite should be either centred within the panel, or drawn with the offset centred. This has the effect of reversing the visual operation of the nudge buttons, although their action doesn't change.
- **Crosshair** toggles drawing a dotted crosshair at the sprite offset. This makes the offset clearly visible and so requires less fiddling with relative adjustments.

image

The wording and position of the buttons isn't all that great to be honest, but putting them there did mean that the window didn't need to be rearranged.

## Checklist for review


PeterN:
Changed wording to make it not look like and action, and also improved the layout of the list.

image

PeterN:
Millions of force-pushes to satisfy the CI string linter :)

2TallTyler (commented):
I'm not sure the Offset/Sprite centred button should be be lowered depending on its setting. That implies an on/off toggle, not an A/B toggle. I could be easily convinced if there's a reason it's this way. 🙂

PeterN:
True, it was originally a toggle, then I decided to make it clearer with the text changing instead. I'll change it.

PeterN:
Hmm, actually that's how our buttons work, without it being lowered, the text wouldn't change. See the "Manual" / "Freeform" edges buttons on the terrain generator.

2TallTyler (approved):
In that case, looks fine to me!

PR #9689, opened 2021-11-11
Add: Specify minimum version for macOS

orudge:
## Motivation / Problem

OpenTTD will open on macOS versions earlier than 10.14, but will crash.

Fixes #9688.

## Description

This adds a minimum version to the Info.plist so the Finder will not allow the application to launch. It will display a minimum version warning instead.

## Limitations

Tested on macOS 10.11 on my ancient MacBook Pro; with this patch it was greyed out and displayed a warning, without the patch it opened and crashed.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10253, opened 2022-12-17
Change: Set minimum macOS version to 10.13

newbthenewbd:
## Motivation / Problem

Following the addition of C++17 variants to the OpenTTD codebase, the deployment target and minimum system version for the game were set to 10.14. However, as testing seems to clearly show[^1][^2], `std::variant`s, along with the specific source of the problem - the `std::bad_variant_access` exceptions - do actually work on macOS 10.13 as well.

MacOS 10.13 is a milestone version, the last one to support Macs from 2009 (dropping, as a cursory search seems to show, no devices that 10.12 supported), and 10.14 only officially works with ones from 2012 onwards, with the exception of some Metal-capable desktop variants.[^3] This means that for little reason, three years' worth of perfectly good computers are being locked out of the game.

## Description

This PR sets the macOS deployment targets and version limits to 10.13.

## Limitations

It turns out that this PR is not even needed in the first place for the `openttd` executable generated by `make` to be launched from within the terminal.[^2] It is, however, essential to allow launching the game from within the .app.

Although these changes were enough for me in order to get the build working fine with LLVM Clang 15 from MacPorts, it might be the case that, say, Apple Clang in Xcode complains. If this were the case, adding `add_definitions(-D_LIBCPP_DISABLE_AVAILABILITY)` somewhere in the CMake setup would probably silence the error, but then it's not up to me to decide whether losing an oversensitive availability guard is too much to pay for potentially gaining a few players. If it's not obvious enough that I am quite biased... :)

## Checklist for review

[^1]: https://stackoverflow.com/a/53868971
  In particular, I am referring to this part, and not the hackier workarounds:
  2a) ... and just use it
[^2]: Apparently, a simple workaround for the version limit exists, even though it's not modifying the Info.plist: I unpacked the contents of the `MacOS` and `Resources` directories from the «incompatible» 12.2 OpenTTD.app into a single external directory, and ran `./openttd`. This is the result:
  Zrzut ekranu 2022-12-17 o 16 36 22
  Appears quite working to me, in contrast to the 10.12 crashes mentioned in #9689.
[^3]: Compare the compatibility charts: [10.12](https://support.apple.com/kb/sp742), [10.13](https://support.apple.com/kb/sp765), [10.14](https://support.apple.com/kb/sp777)

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -162,7 +162,7 @@ jobs:
 
     runs-on: macos-latest
     env:
-      MACOSX_DEPLOYMENT_TARGET: 10.14
+      MACOSX_DEPLOYMENT_TARGET: 10.13
 
     steps:
     - name: Checkout
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -469,7 +469,7 @@ jobs:
 
     runs-on: macos-11
     env:
-      MACOSX_DEPLOYMENT_TARGET: 10.14
+      MACOSX_DEPLOYMENT_TARGET: 10.13
 
     steps:
     - name: Download source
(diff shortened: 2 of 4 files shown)


glx22 (approved):
Seems to build fine (minus some link time warnings due to cached vcpkg libs, but we can kill the cache before merging).
The only way to really test is to merge and get a nightly build.

PR #10256, opened 2022-12-18
Fix #10248: Ground sprites depend on foundation sprite offsets.

### LINKED ISSUE #10248: [Bug]: groundtiles on foundation should not be childsprites of foundation sprites
### Version of OpenTTD

13.0 b2

### Expected result

A groundsprite on a foundation should be positioned mathematically, rather than being parented to the foundation sprites.

### Actual result

With extra zoom grfs, some error in a calculation somewhere appears to attach the groundsprite to the foundation sprite two pixels to the left of where it ought to be. This error can be seen in my WIP grf, as well as the zbase baseset - both of which have used the workaround of clipping the foundation short to position the groundtile correctly.

Untitled
zbase

This calculation could be fixed, but that would cause errors for grfs (including base sets) which have previously worked around the error. Instead, the groundtile should be simply placed in the correct position (as building sprites on foundations are) instead of being parented to the foundation.

### Steps to reproduce

Look closely at the zbase graphics

PeterN:
## Motivation / Problem

As per #10248, ground sprites drawn on top of foundations are drawn as child sprites with the foundation as the parent sprite.

Child sprites are currently drawn relative to the parent sprite, **including** the parent sprite's offsets. This means that when adjusting offsets of foundation sprites, the ground sprite on top also moves.

This can be clearly seen when adjusting sprite offsets:

image

In this case I have only adjust the sprite offset of the foundation sprite, and this has also shifted the ground sprite of the building above.

## Description

This PR decouples this relationship for foundation child sprites by adding a mode that draws the child sprite relative to the parent sprite's bounding box instead of bounding box + sprite offset.

This requires a little bit of extra fiddling when drawing railway half-tile foundation sprites as in that case the bounding box is shifted. This worked before purely due to the the parent sprites internal offsets happening to be correct.

With this change, adjust the sprite offset of the foundation sprite no longer move the ground sprite above:

image

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -500,7 +500,7 @@ void DrawFoundation(TileInfo *ti, Foundation f)
 		if (!IsNonContinuousFoundation(f)) {
 			/* Lower part of foundation */
 			AddSortableSpriteToDraw(
-				leveled_base + (ti->tileh & ~SLOPE_STEEP), PAL_NONE, ti->x, ti->y, 16, 16, 7, ti->z
+				leveled_base + (ti->tileh & ~SLOPE_STEEP), PAL_NONE, ti->x, ti->y, TILE_SIZE, TILE_SIZE, TILE_HEIGHT - 1, ti->z
 			);
 		}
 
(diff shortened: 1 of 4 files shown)


PikkaBird:
If it's of any use to anyone testing, I've attached a test grf with un-nerfed foundation sprites, and so exhibits the bug this PR is intended to fix. Pre-PR, dropping the western or northern corner of a tile and creating a foundation causes the ground tile above to shift to the left.

Lumpyton Transport, 1920-02-07

[foundationtest.zip](https://github.com/OpenTTD/OpenTTD/files/10268690/foundationtest.zip)

PeterN:
This isn't really anything to do with NewGRF to be honest, so there's nothing to review on that side.

LordAro (dismissed):
Yay magic numbers

PeterN:
Less magic numbers :-)

2TallTyler (approved):

PR #10259, opened 2022-12-18
Draft: plus / minus icon redraw

andythenorth:
Relates to #10220


LordAro:
I think -1 to this change without any additional explanation. black on grey works much less well than orange on grey

andythenorth:
I think the correct thing to do is 
-  decide what colour we want to use, the orange doesn't match nearby yellow text in the cases where it's used, seems like it should
- improve the current icon: the shading is incomplete and not at the normal angle for font shading

Peter has an svg example that looks much better, but depends on an svg icon replacement branch.

glx22:
Maybe it should be a character, or a recolourable sprite.

LC-Zorg:
I'm not a big fan of the graphic of this button, but it's not something that bothers me either.
Anyway, I had some fun with it... :)

**Orange:** round with improved shading, squares, semisquares
Iron Horse - orange

**Grey:** round, squares, semisquares
Iron Horse - grey
I've been checking out other colors, but I don't really see any others that would fit well here.

**Yellow:** as like buttons
Iron Horse v1 1 yellow as like buttons
Here it looks good, but in the settings window it can be problematic because it suggests that you have to press this little button, although it is not.

**Setting window**
Here it would be most useful to change the color of the inscriptions in the type and category fields of settings (they are barely visible). Different colors of the icons may be even better in this case.
Setting window icons

PeterN:
I can look at this (already done some fiddling as it's also involved in my SVG branch) but I think it can wait til after 13.0-rc1

TrueBrain:
Few months later, I look at the PR, and I really have no clue what this PR is about, sorry :) Would you mind adding the template information to the description, so the next reader has an idea what this is actually about? Thank you :)

PR #10262, opened 2022-12-20
Change: Support flipping shorter engines without NewGRF support.

PeterN:
## Motivation / Problem

"It's impossible to flip shorter rail vehicles without NewGRF support"

(Flip activated by ctrl-click in the depot, which relies on EF_RAIL_FLIPS)

Below is an example of what happens if flipping was not prohibited, and the NewGRF provides no* support. It's not entirely legit because you can't do this in master...

image

*For UKRS3 it's mixed, the engines support it but the wagons do not.

## Description

This PR allows any* rail vehicle to be flipped without extra NewGRF by:
1. Flipping the direction before setting up the bounding box.
2. Minor adjustment to offset for odd-length engines to get the sprite into the correct position.

Correct drawing in the depot and mouse cursor is also handled.

This PR changes the meaning of EF_RAIL_FLIPS:
1. If not set, any rail vehicle (except multi-head or articulated) can be flipped, and OpenTTD will handle the offsets.
2. If set, then the flip is still allowed, but OpenTTD no longer handles the offsets and leaves it to the NewGRF.

This should effectively mean this is backwards compatible without add any new flags. But I might have missed something...

Below is an example using the same NewGRFs (and save file...) as above.

image

## Limitations

:shrug:

Could potentially be compatibility problems if a NewGRF has implemented support for flipping vehicles, but NOT set the EF_RAIL_FLIPS flag. That is unlikely, so is bound to exist somewhere.

## Checklist for review

### DIFF
--- a/src/engine_type.h
+++ b/src/engine_type.h
@@ -155,7 +155,7 @@ enum EngineMiscFlags {
 	EF_ROAD_TRAM  = 0, ///< Road vehicle is a tram/light rail vehicle
 	EF_USES_2CC   = 1, ///< Vehicle uses two company colours
 	EF_RAIL_IS_MU = 2, ///< Rail vehicle is a multiple-unit (DMU/EMU)
-	EF_RAIL_FLIPS = 3, ///< Rail vehicle can be flipped in the depot
+	EF_RAIL_FLIPS = 3, ///< Rail vehicle has old depot-flip handling
 	EF_AUTO_REFIT = 4, ///< Automatic refitting is allowed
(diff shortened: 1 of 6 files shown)


andythenorth:
Works as expected for me when tested.

LordAro (dismissed):
Sure, why not

2TallTyler (approved):

andythenorth:
I've updated grf spec status and relevant wiki pages.
https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status

PR #10278, opened 2022-12-22
Add: 'font' console command to configure fonts

### LINKED ISSUE #8596: Provide GUI for changing of fonts ingame
## Version of OpenTTD
1.10.3
## Expected result
Possibility to freely choose the size and font type in the game options window or the settings window, without having to close the game, open the configuration file and also knowledge what and where to change.
## Actual result
The player has very limited options to change the font by step-size multiplier of the font size: x1, x2, x4, which rarely gives a satisfactory result. This choice often doesn't allow to adjust the fonts in the game to your monitor and the player's needs.
## Comment
The default setting is especially disadvantageous for the smallest font. Its enlargement by the available setting makes the average font unnecessarily large.
On the other hand, the default medium font, although quite good in the game, works very badly in a readme file, which is tiring to read after the first line of text.
Here all the work that nielsmh is doing with the manual window (#7786) seems meaningless. :( Even if the entire encyclopedia was posted there, reading it would still be very tedious. Yes, you can change this font, but even many players involved do not know about this possibility, and since the manual will be mainly for players starting their adventure with the game, it is this they are more unlikely to know how to change it.
The font can be freely selected in the configuration file. It is not complicated at all. However, you need to know where and what to do. You also have to exit the game to make a change. Another downside of this form is that you can't see what the effect will be. Not all fonts look good in-game. To find the right one, sometimes even several dozen attempts are needed, which are associated with closing and opening the game each time.
For me, the ability to change the font during the game would not be particularly important, but it would certainly be useful for most players and for the game itself, the interface of which may overwhelm new players with its complexity and not the best, and sometimes poor readability - the default small font is too small for most modern monitors.

glx22:
Closes #8596.

## Motivation / Problem
Configuring fonts requires multiple iterations of 
- editing `openttd.cfg`
- start OpenTTD to see the result
- close OpenTTD

So a console command to do the same from a running OpenTTD will be an improvement.

## Description
I implemented `font` console command.
The command allows to list loaded fonts, with indication of the config as loaded font may differ (fallback mechanism ...).

Usually fonts are initialised before the command is accessible, but `FS_MONO` has to be different as it's not loaded unless a TextFile is displayed. So I included a mechanism to make sure all non sprite fonts are loaded when listing them.

For "normal" fonts I used `CheckForMissingGlyphs()` to initialise and check the modified font, but for "monospace" I directly `InitFontCache(true)` because the validation is only done when loading a file in a TextFile window, but that should at least validate the user selected font could work.

Fallback mechanism is also a huge annoyance as it reinitialises the fonts and may invalidate a perfectly fine font for a given size.
I tried to work around this by temporary storing in the config the fonts currently loaded for other `FontSize` when changing config for a font. That way if other sizes use fallback font they will still use it while validating the new font.
Of course if the new font is not valid and a fallback is searched, other sizes, which might have been valid while not being the fallback font, will end up using the fallback font.
We really need a per character fallback mechanism instead of the global one we currently have, but that will be another journey.

## Limitations
I tested only with non freetype windows, but I think it should work for other platforms as long as `FontCache::GetFontName()` overrides return a sensible value.

## Checklist for review

### DIFF
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -13,6 +13,11 @@
 #include "blitter/factory.hpp"
 #include "gfx_layout.h"
 #include "fontcache/spritefontcache.h"
+#include "openttd.h"
+#include "settings_func.h"
+#include "strings_func.h"
+#include "viewport_func.h"
+#include "window_func.h"
 
 #include "safeguards.h"
 
@@ -71,6 +76,49 @@ bool GetFontAAState(FontSize size, bool check_blitter)
 	if (check_blitter && BlitterFactory::GetCurrentBlitter()->GetScreenDepth() != 32) return false;
 
 	return GetFontCacheSubSetting(size)->aa;
+}
+
+void SetFont(FontSize fontsize, const std::string& font, uint size, bool aa)
+{
+	FontCacheSubSetting *setting = GetFontCacheSubSetting(fontsize);
+	bool changed = false;
+
+	if (setting->font != font) {
+		setting->font = font;
+		changed = true;
+	}
+
+	if (setting->size != size) {
+		setting->size = size;
+		changed = true;
+	}
+
+	if (setting->aa != aa) {
+		setting->aa = aa;
+		changed = true;
+	}
+
+	if (!changed) return;
+
+	if (fontsize != FS_MONO) {
+		/* Try to reload only the modified font. */
+		FontCacheSettings backup = _fcsettings;
+		for (FontSize fs = FS_BEGIN; fs < FS_END; fs++) {
+			if (fs == fontsize) continue;
+			FontCache *fc = FontCache::Get(fs);
+			GetFontCacheSubSetting(fs)->font = fc->HasParent() ? fc->GetFontName() : "";
+		}
+		CheckForMissingGlyphs();
+		_fcsettings = backup;
+	} else {
+		InitFontCache(true);
+	}
+
+	LoadStringWidthTable();
+	UpdateAllVirtCoords();
+	ReInitAllWindows(true);
+
+	if (_save_config) SaveToConfig();
 }
 
 /**
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -70,15 +70,8 @@ bool GetFontAAState(FontSize size, bool check_blitter)
 	/* AA is only supported for 32 bpp */
 	if (check_blitter && BlitterFactory::GetCurrentBlitter()->GetScreenDepth() != 32) return false;
 
-	switch (size) {
-		default: NOT_REACHED();
-		case FS_NORMAL: return _fcsettings.medium.aa;
-		case FS_SMALL:  return _fcsettings.small.aa;
-		case FS_LARGE:  return _fcsettings.large.aa;
-		case FS_MONO:   return _fcsettings.mono.aa;
-	}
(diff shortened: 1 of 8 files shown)


LordAro:
Confirmed works well on Linux + SDL2.

Command syntax may need some work though - explicit "name" and "size" arguments seem a bit odd, any particular reason for them? why not just `font medium "Liberation Sans" [15] [noaa|aa]` ?

Using "" to reset to sprite doesn't seem to work. Helpfully, just using any other string ("sprite", "flibble", etc) works fine as it just falls back to the sprite font for unrecognised fonts

the aa setting for fonts appears to be stateful. noaa seems to be the default (should it?), but whichever it was previously set to appears to be retained, even if you go via the sprite font.

glx22:
The idea behind explicit "name" and "size" is to allow to modify only part of the config, like changing font size without touching the font name.

Using "" just works fine for me, and it should also work with freetype as the first thing in `LoadFreeTypeFont()` is `if (settings->font.empty()) return;` which skips loading and falls back to sprite. It's the same in `LoadWin32Font()` and `LoadCoreTextFont()`.

If you don't specify aa or noaa, the current value is kept, same for name and size.

LordAro:
ah yeah, looks like i was just forgetting the `name` prefix previously. Told you it was confusing :p

glx22:
I understand the confusion, the syntax is intended to work like editing `openttd.cfg` manually.
I guess I could add a 'font [medium|small|large|mono] <font_name>' shortcut.
Hmm even 'font [medium|small|large|mono] [<font_name>] [<size>] [aa|noaa]' should be doable as it's possible to "check" value types (basic integer/non integer).

LordAro (dismissed):
LGTM.

PeterN:
ReInitAllWindows() should be called with true (the parameter naming is a bit wrong, sorry) so that dimensions are recalculated.

Viewport signs need to invalidated if the fonts are changed as the sign sizes change. UpdateAllVirtCoords().

Might need to call LoadStringWidthTable(), though I'm not sure when that data is used anymore.

src/fontcache.cpp
```diff
@@ -70,15 +74,49 @@ bool GetFontAAState(FontSize size, bool check_blitter)
...
+				if (fs == fontsize) continue;
+				FontCache *fc = FontCache::Get(fs);
+				GetFontCacheSubSetting(fs)->font = fc->HasParent() ? fc->GetFontName() : "";
+			}
+			CheckForMissingGlyphs();
+			_fcsettings = backup;
+		} else {
+			InitFontCache(true);
+		}
+		ReInitAllWindows(false);
```
PeterN: Call with true to ensure windows get resized fully for the font change.

May need to call `LoadStringWidthTable()` here.
Do need to call `UpdateAllVirtCoords()` here.

src/fontcache.cpp
```diff
@@ -70,15 +74,49 @@ bool GetFontAAState(FontSize size, bool check_blitter)
...
+		setting->size = size;
+		changed = true;
+	}
+
+	if (setting->aa != aa) {
+		setting->aa = aa;
+		changed = true;
+	}
+
+	if (changed) {
```
PeterN: if (!changed) return;

PeterN (changes requested):

PeterN (approved):

Issue #10238, opened 2022-12-14
One-way road arrows don't look good on slopes


PikkaBird:
### Version of OpenTTD

13.0 b2

### Expected result

Both the base graphics and newgrfs treat the one-way arrows as diegetic, ie as if they were in-game road markings rather a UI indicator. Action 5 type 09 should be extended with an additional 12 sprites to allow providing sloped versions of the one-way arrows.

### Actual result

The one-way arrows do not have sloped versions and thus look naff on slopes.
![Jollyburg Transport, 1976-09-30](https://user-images.githubusercontent.com/39507755/207490924-c45b56ec-0ae4-437c-85cb-bb1f88433bd5.png)

### Steps to reproduce

Build a one-way road on a slope.

LC-Zorg:
When introducing changes, it would also be worth considering the possibility of assigning different direction signs to different roads.
![Zrzut ekranu 2021-03-13 162456](https://user-images.githubusercontent.com/67662046/207740899-6d4113b7-9399-420f-bf1a-b7afeeac24a1.png)

michicc:
Support added in #10282

PR #10282, opened 2022-12-25
Add: Slope-aware and roadtype-specific one-way sprites

FLHerne:
## Motivation / Problem
As in #10238:
 * The baseset only includes one-way road indicators for flat ground. These look bad when placed on slopes.
 * NewGRF road types can't provide their own type-specific indicators. The baseset ones may be inappropriate for some road types (e.g. painted markings on dirt roads are silly) @LC-Zorg.

## Description

Action5 type 0x09 is extended to accept a set of 18 sprites. 
The first 6 are unchanged from before this patch, followed by similar groups for north-facing and then south-facing slopes.

For compatibility with older basesets and NewGRFs, if an Action5 provides only sprites for flat tiles these will be used on slopes as in previous versions.

Action3 type 0x0B, for roadtypes only, may optionally provide a set of 18 sprites as above for use with that roadtype.

OpenTTD/OpenGFX#82
OpenTTD/nml#274

## Limitations

~~This doesn't yet allow the baseset itself to provide indicators for sloped tiles, because I haven't yet understood how to add sprites to the baseset without breaking compatibility. I intend to do that before merging.~~
This is done now.
Using flat tiles for slopes, if no sloped tiles are provided, for compatibility might be confusing if for some reason a GRF wanted to override only the flat tiles. I can't think of a reason to do that.

## Checklist for review


glx22:
Basesets use action5, so handling more sprites here should be doable, if only first 6 are provided use openttd included ones.

FLHerne:
I know it's possible, I just haven't yet figured out the details of how it's done. Will look again when it's not 3am on Christmas Day...

glx22:
https://github.com/OpenTTD/OpenTTD/blob/master/src/newgrf.cpp#L6160
Looks like changing `ONEWAY_SPRITE_COUNT` should be enough.

It's already handled like OpenTTD GUI block (and we increased it many times).

FLHerne:
Action5 is supported now; this is ready for review.

src/newgrf.cpp
```diff
@@ -6250,9 +6250,17 @@ static void GraphicsNew(ByteReader *buf)
 		if (offset <= depot_no_track_offset && offset + num > depot_no_track_offset) _loaded_newgrf_features.tram = TRAMWAY_REPLACE_DEPOT_NO_TRACK;
 	}
 
+	/* If the baseset or grf only provides sprites for flat tiles (pre #10282), duplicate those for use on slopes. */
+	bool dup_oneway_sprites = ((type == 0x09) && (offset + num) <= SPR_ONEWAY_SLOPE_N_OFFSET);
```
2TallTyler: Parenthesis don't make sense here: the `)` after `num` should be after the offset constant.

src/table/sprites.h
```diff
@@ -290,8 +290,10 @@ static const SpriteID SPR_TRAMWAY_DEPOT_NO_TRACK       = SPR_TRAMWAY_BASE + 113;
 static const uint16 TRAMWAY_SPRITE_COUNT = 119;
 
 /** One way road sprites */
-static const SpriteID SPR_ONEWAY_BASE = SPR_TRAMWAY_BASE + TRAMWAY_SPRITE_COUNT;
-static const uint16 ONEWAY_SPRITE_COUNT = 6;
+static const SpriteID SPR_ONEWAY_BASE           = SPR_TRAMWAY_BASE + TRAMWAY_SPRITE_COUNT;
+static const SpriteID SPR_ONEWAY_SLOPE_N_OFFSET = 6;
+static const SpriteID SPR_ONEWAY_SLOPE_S_OFFSET = 12;
```
2TallTyler, line 295: Are these names for North and South? Should they be for axis instead?

2TallTyler:
I think the failed commit check is an actions failure and not a problem with your PR. Hope it'll run properly next time you push it.

src/table/sprites.h
FLHerne: Yes and no respectively. The sprites are grouped by (flat) (NE and NW slopes) (SE and SW slopes), with each of those corresponding closely to the pre-existing set of sprites.

Grouping them by axis would make it more complicated to retain compatibility with existing basesets and grfs.

See oneway.png in the OpenGFX or NML PRs; the first six sprites are unchanged from previously and the ugly red arrows point uphill. (Pikka agreed to draw some actual slope sprites at some point.)

michicc:
> ~This doesn't yet allow the baseset itself to provide indicators for sloped tiles, because I haven't yet understood how to add sprites to the baseset without breaking compatibility. I intend to do that before merging.~ This is done now. 

Ideally, the one-way road graphics from openttd.grf would be updated, too. One could argue if the graphics should be in openttd.grf or in orig_extra.grf.
Right now they are in openttd.grf, which means any baseset that does not provide one-way arrows will get the ones from there. Adding them there would preserve this, while needing code to do the legacy behaviour when a baseset only sets the old amount. Adding to orig_extra.grf would avoid this, but results in a different look if a baseset uses the game-supplied arrow sprites.

PeterN:
Something is not right with it. I reran the CI but it didn't resolve it.

src/newgrf.cpp
FLHerne: Fixed now, thanks.

FLHerne:
Added sloped graphics to `openttd.grf`. I got a bit lazy with the 'impassable' sprites because they're useless so no-one will ever see them; gave them better offsets but didn't redraw the sprite.

These will only be visible if no one-way sprites at all are provided by a baseset or GRF; i.e. when using the TTD baseset.
If a baseset or GRF provides flat one-way sprites they'll be used on slopes per Action5 compat; this looks less bad IMO than using a completely mismatched style especially for zBase.

michicc (approved):

michicc (approved):

PR #10283, opened 2022-12-25
Fix #10150: Force FS_SMALL for small viewport signs.

### LINKED ISSUE #10150: [Bug]: Names of cities on the map are shortened pointlessly after changing the font
### Version of OpenTTD

12.2 (not new bug)

### Expected result

City names are only shortened if they are really long or they are never shortened.

### Actual result

After changing the font, the names of cities on the game map (zoom out) are shortened against any visible logic. Very often, short names are abbreviated while long names remain unchanged. Moreover, the abbreviated name of the city + ... also very often have the same or very similar length as not abbreviated. This is rare for original English names and quite common for Fictional Town Names. Not all fonts are affected.

small Arial

small Verdana

Better town names

Better town names 2

### Steps to reproduce

1. Change small font to Arial, Calibri or Verdana.
2. Generate a map with multiple cities, preferably with some add-on that change the names of the cities.
3. More or less of the cities will have a shortened name in the zoom out view.

PeterN:
## Motivation / Problem

As per #10150, small town sign labels may be cropped with Truetype fonts.

image

This is a workaround for string widths being different with mixed font-sizes.

For some reason (with Uniscribe at least), the string bounding box of "{FS_SMALL}Text" can be different when called starting with FS_SMALL and FS_NORMAL.

## Description

image

This is resolved by explicitly drawing small signs with FS_SMALL, although the underlying problem is not touched.

## Checklist for review

### DIFF
--- a/src/viewport.cpp
+++ b/src/viewport.cpp
@@ -1722,7 +1722,7 @@ static void ViewportDrawStrings(ZoomLevel zoom, const StringSpriteToDrawVector *
 			}
 		}
 
-		DrawString(x + WidgetDimensions::scaled.fullbevel.left, x + w - 1 - WidgetDimensions::scaled.fullbevel.right, y + WidgetDimensions::scaled.fullbevel.top, ss.string, colour, SA_HOR_CENTER);
+		DrawString(x + WidgetDimensions::scaled.fullbevel.left, x + w - 1 - WidgetDimensions::scaled.fullbevel.right, y + WidgetDimensions::scaled.fullbevel.top, ss.string, colour, SA_HOR_CENTER, false, small ? FS_SMALL : FS_NORMAL);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/viewport.cpp
```diff
@@ -1323,7 +1323,7 @@ void ViewportAddString(const DrawPixelInfo *dpi, ZoomLevel small_from, const Vie
 		int shadow_offset = 0;
 		if (string_small_shadow != STR_NULL) {
 			shadow_offset = 4;
-			AddStringToDraw(sign->center - sign_half_width + shadow_offset, sign->top, string_small_shadow, params_1, params_2, INVALID_COLOUR, sign->width_small);
+			AddStringToDraw(sign->center - sign_half_width + shadow_offset, sign->top, string_small_shadow, params_1, params_2, INVALID_COLOUR, sign->width_small | 0x8000);
```
2TallTyler, line 1326: Is there not a flag enum for this?

PeterN: No, it is only used here, and only tested with `HasBit(ss.width, 15)`. I have in mind a proper rewrite so don't really want to start adding flags.

2TallTyler (approved):

PR #10290, opened 2022-12-29
Change: Big UFO disaster targets current location of a random train

LordAro:
## Motivation / Problem
See #10076
Closes #10076

## Description
As suggested, Big UFO now targets the current location of a random train.
Code largely copied from the standard UFO - differs in that it uses the current tile of the train, rather than the index, because we're not actively targeting a train

## Limitations
Limited testing, because disasters are a pain to spawn. Potential feature: Cheat to spawn a disaster.

## Checklist for review

### DIFF
--- a/src/disaster_vehicle.cpp
+++ b/src/disaster_vehicle.cpp
@@ -40,6 +40,7 @@
 #include "sound_func.h"
 #include "effectvehicle_func.h"
 #include "roadveh.h"
+#include "train.h"
 #include "ai/ai.hpp"
 #include "game/game.hpp"
 #include "company_base.h"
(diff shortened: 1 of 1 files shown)


src/disaster_vehicle.cpp
```diff
@@ -578,17 +579,27 @@ static bool DisasterTick_Big_Ufo(DisasterVehicle *v)
...
+
+		if (n == 0) {
+			/* If there are no targetable trains, destroy the UFO. */
+			delete v;
+			return false;
+		}
+
+		n = RandomRange(n); // Choose one of them.
+		for (const Train *t : Train::Iterate()) {
+			/* Find (n+1)-th road vehicle. */
```
stormcone:
```suggestion
			/* Find (n+1)-th train. */
```

LordAro: Can you tell the code was copied?

stormcone: I actually read it in the description. :)

src/disaster_vehicle.cpp
```diff
@@ -578,17 +579,27 @@ static bool DisasterTick_Big_Ufo(DisasterVehicle *v)
 		}
 		v->current_order.SetDestination(1);
 
-		TileIndex tile_org = RandomTile();
-		TileIndex tile = tile_org;
-		do {
-			if (IsPlainRailTile(tile) &&
-					Company::IsHumanID(GetTileOwner(tile))) {
```
glx22, line 585: AIs are not really able to manage Big UFO effects, that's why only human owned tiles were targeted.
I think the new method should also filter out AI owned vehicles.

JGRennison:
On the testing side, you could probably locally cherry pick this: https://github.com/JGRennison/OpenTTD-patches/commit/4cdc253c38fe7378e2e9cf0ff58bf2e4bc0baa68

rubidium42 (dismissed):

nielsmh:
Shouldn't this also test if the train is actually running and not in a depot?
I suppose it's not a real problem, and targeting depots is valid too. (Do trains in depot have their correct tile stored?)

glx22:
Yeah skipping crashed and hidden vehicles (in depot, tunnel) might be a good idea.

rubidium42:
> Yeah skipping crashed and hidden vehicles (in depot, tunnel) might be a good idea.

Though then you also have to do the same for road vehicles, so they are consistent. And then the scope of the change grows significantly.

glx22:
For road vehicle the target is the vehicle itself, it can move while the UFO is coming. Here we just want a tile, and the IsPlainRailTile() check has been removed.

JGRennison:
> > Yeah skipping crashed and hidden vehicles (in depot, tunnel) might be a good idea.
> 
> Though then you also have to do the same for road vehicles, so they are consistent. And then the scope of the change grows significantly.

The various disasters are extremely arbitrary. There isn't some underlying consistency to be exposed and any possible changes to small UFO road vehicles don't have to be done in the same PR.

LordAro:
¯\\\_(ツ)_/¯

glx22 (approved):

PR #10299, opened 2022-12-31
Enable CodeQL code scanning

rubidium42:
## Motivation / Problem

There used to be a website called lgtm.com, but that has been deprecated in favour of this solution.
The goal is to resolve potential code issues before an actual release, and this way it should even be able to do it with pull requests.

## Description

Just followed the manual at https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository#setting-up-code-scanning-manually though disabled Javascript testing.

## Limitations

Compiling will be slower, as CodeQL building takes about 45 minutes (MinGW takes about 30 minutes). There are three "configurations" that are supported out of the box:
- default: 50 high security alerts -> https://github.com/rubidium42/OpenTTD/pull/2/checks?check_run_id=10378139264
- security-extended: 83 high security alerts -> https://github.com/rubidium42/OpenTTD/pull/3/checks?check_run_id=10378116897
- security-and-quality: 83 high security alerts, 38 errors, 106 warnings and 3151 notes -> https://github.com/rubidium42/OpenTTD/pull/4/checks?check_run_id=10378214008

Someone needs to go through the errors and warnings first, to either dismiss them or fix them in a separate PR. However, I'm not sure whether people without merge rights will be able to dismiss warnings in their PRs and how that will look.

Furthermore there's a question whether some of the warnings/notes might be too noisy for OpenTTD's coding style. So, the first question would be... which configuration do we want to use?

## Checklist for review

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,78 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches:
+    - master
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install dependencies
+      run: |
+        echo "::group::Update apt"
+        sudo apt-get update
+        echo "::endgroup::"
+
+        echo "::group::Install dependencies"
+        sudo apt-get install -y --no-install-recommends \
+          liballegro4-dev \
+          libfontconfig-dev \
+          libicu-dev \
+          liblzma-dev \
+          liblzo2-dev \
+          libsdl2-dev \
+          zlib1g-dev \
+          # EOF
+        echo "::endgroup::"
+      env:
+        DEBIAN_FRONTEND: noninteractive
+
+    - name: Set number of make jobs
+      run: |
+        echo "MAKEFLAGS=-j$(nproc)" >> $GITHUB_ENV
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: cpp
+        config-file: ./.github/codeql/codeql-config.yml
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v2
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+      with:
+        category: /language:cpp
+        upload: False
+        output: sarif-results
+
+    - name: Filter out table & generated code
+      uses: advanced-security/filter-sarif@v1
+      with:
+        patterns: |
+          +**/*.*
+          -**/table/*.*
+          -**/generated/**/*.*
+        input: sarif-results/cpp.sarif
+        output: sarif-results/cpp.sarif
+
+    - name: Upload results
+      uses: github/codeql-action/upload-sarif@v2
+      with:
+        sarif_file: sarif-results/cpp.sarif
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,10 @@
+name: openttd
+queries:
+- uses: security-and-quality
+query-filters:
+- exclude:
+    id:
+    # Only feasible way is to move away from fopen; fopen_s is optional C11 and not implemented on most platforms.
+    - cpp/world-writable-file-creation
+    # Basically OpenTTD's coding style for adding things like ..._INVALID to enumerations
+    - cpp/irregular-enum-init


glx22:
I quickly checked the default report. In theory the issues are valid, but I think they are highly improbable in our usage.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
+# For most projects, this workflow file will not need changing; you simply need
```
TrueBrain: Please take a look at other workflows to match the style a bit more. I made some comments here to give a direction to what I mean :)

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
```
TrueBrain: Please remove all these comments; it doesn't add value after the initial run through it

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+name: "CodeQL"
```
TrueBrain: No need for the quotes :)

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+
+on:
+  push:
+    branches: [ "master" ]
```
TrueBrain:
```
on:
  push:
    branches:
    - master
```

The silly inline list syntax of YAML is only annoying to read :)

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '19 2 * * 4'
```
TrueBrain: please drop two spaces from this line:

```
  schedule:
  - cron: ..
```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Use only 'java' to analyze code written in Java, Kotlin or both
+        # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
```
TrueBrain: There is no need for a strategy here, as we only run a single entry. You can just replace `${{ matrix.language }}` with `cpp`, and we are good :)

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+
+    steps:
+    - name: Checkout repository
```
TrueBrain: We call this `- name: Checkout` in all other workflows. Just to keep things more the same.

TrueBrain (commented):
I like the very verbose one. Especially as only errors and above are reported in the PR itself, but if you want, you can see what else CodeQL discovered. There are some pretty decent finds in there. And it doesn't actually cost anything extra to run them.

I also think we should enable this for all PRs. Although it takes even longer to finish up a PR, it might prevent silly mistakes and if people take care of the notes, improve the quality of our codebase.

We of course have done this mostly on the background over the years, either with coverity or with lgtm, but I think it is time to drag it to the foreground, especially with how well it integrated with GitHub.

For example, if we dismiss an alart because we consider it not relevant for us, it will remain in that state till we tell it otherwise. From the same interface as the one we use to make PRs. So that is nice.

TrueBrain:
The main thing that is a bit annoying, that it also finds "problems" in the generated code. Mostly "unused variable" and "unused function". These make up most of the notes CodeQL generates. Ideally, CodeQL wouldn't report those.

But looking at the manual, what they say: if you don't want reports on files, don't compile them (with a custom build step). The problem for us: they are all header files, and we cannot not compile them. So I am not actually sure how we could ignore them.

`path-ignore` wouldn't work, as that doesn't work for a compiled language. So that is a bit of a mystery to me. But we could also just mass-close them, as they are clearly marked as "generated", and forgot about them. So it is not a real biggy .. just annoying :P

---

The other thing I noticed, that CodeQL creates a release build. And this is most likely why it is also taking so much more time. If we would use a custom build step, maybe we could reduce the build-time by using a debug build. Not sure if it actually matters, but just something I spotted.

LordAro:
a release build also uses LTO, which may (or may not) allow CodeQL to find more issues?

TrueBrain:
> a release build also uses LTO, which may (or may not) allow CodeQL to find more issues?

That seems not to be the case. It intercepts the compiler-call to know what files to analyze .. and that is about it.

Some other ideas to speed up the build:

- https://github.com/github/codeql/discussions/11260#discussioncomment-4140432
  (read: builds run on a single core, instead of using both cores available on the GitHub Actions runners)
- As it intercepts calls to gcc, but doesn't actually care about the result, I wonder if we could symlink `c++` to a dummy script, just to CodeQL sees the call. That might speed things up too. Just a bit hackish.

If the first suggestion already makes CodeQL faster than msys2, nobody cares anymore how long it actually takes. So maybe that is the best option here :)

rubidium42:
> Please take a look at other workflows to match the style a bit more. I made some comments here to give a direction to what I mean :)

I hope I made the right changes. I left some comments alone as they might be important (later?).


>     * [Is there any way to tell a github workflow's autobuilder to use parallel make? github/codeql#11260 (comment)](https://github.com/github/codeql/discussions/11260#discussioncomment-4140432)
>       (read: builds run on a single core, instead of using both cores available on the GitHub Actions runners)

This already chops 10-15 minutes from the run time. In this particular case 33 minutes for the code scanning and 29 for the slowest msys2, however that finished after the code scanning. Where about 50% of the time is spent in the build and 50% of the time is spent in the analyse phases.


>     * As it intercepts calls to gcc, but doesn't actually care about the result, I wonder if we could symlink `c++` to a dummy script, just to CodeQL sees the call. That might speed things up too. Just a bit hackish.

That might be extremely hackish, as strgen and friends probably need to run to get the valid automatically generated files needed for a normal compile. That seems way too hackish to me.

rubidium42:
> But looking at the manual, what they say: if you don't want reports on files, don't compile them (with a custom build step). The problem for us: they are all header files, and we cannot not compile them. So I am not actually sure how we could ignore them.

Adding `autofilter:true` to the filter when you go through a listing already seems to remove most of the generated code, leaving about 1000 errors/notes, instead of 3000. So that at least helps when going manually through the list.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches:
+    - master
+  schedule:
+  - cron: '19 2 * * 4'
```
glx22: push and pull_request triggers should be enough, I don't think it needs a schedule trigger

rubidium42:
I wonder how we can disable certain checks. I tried the following, but that failed and I have no idea how to figure out why.
```
    - name: Initialize CodeQL
      uses: github/codeql-action/init@v2
      with:
        languages: cpp
        queries: security-and-quality
        query-filters:
        - exclude:
          id: cpp/world-writable-file-creation

```

TrueBrain:
> github/codeql-action/ini

The `with` statement is rather limited, and I think this is the reason CodeQL only supports `exclude` by using their configuration file.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#excluding-specific-queries-from-analysis

Basically, create `./.github/codeql/codeql-config.yml` and put in it:
```
query-filters:
- exclude:
  id: cpp/world-writable-file-creation
```

Now call the `init` step with:

```
    - name: Initialize CodeQL
      uses: github/codeql-action/init@v2
      with:
        languages: cpp
        queries: security-and-quality
        config-file: ./.github/codeql/codeql-config.yml
```

Didn't actually test it, but knowing their excellent manuals, that should do the trick :)

rubidium42:
> Didn't actually test it, but knowing their excellent manuals, that should do the trick :)

There's one little tidbit that made it really hard to figure out what was going wrong when building with this. The `id` after `exclude:` must be indented more, otherwise you'll get an error message which Google doesn't know yet. So, `id` may not be at the same level as `exclude`.
The error message was "Query filter must have exactly one key".

I also added some extra filtering that should remove all warnings from tables and generated. However, writing this I wonder whether that might have been a bit too blunt and that we should only filter out specific warnings/notes from tables and generated instead.

TrueBrain:
Sorry, that was my bad ... poor copy/paste from their manual :(

I think skipping the generated code in general is fine honestly, especially for now :)

TrueBrain (approved):

PR #10306, opened 2023-01-02
Codechange: address CodeQL issue "Multiplication result converted to larger type"

TrueBrain:
## Motivation / Problem

Mostly I was curious how hard it is to address the issues found by CodeQL in #10299.

## Description

Address the multiplication issues found by CodeQL. Mostly this is easily avoided by casting the first value to the bigger type. Sometimes this allowed for some code cleanup.

The `BufferSize` is worth mentioning: it gives the suggestion it should be used for .. well .. buffer size, so a `size_t` felt more appropriate than the `int`. Having it signed is very weird anyway, so changing it to unsigned doesn't hurt anyone.

Most of the hits are unlikely to cause any problems for us, with two exceptions:
- The linkgraph could, potentially, wrap before we expect it
- The money cheat could, potentially, wrap before we expect it

Both will most likely go unnoticed by the end-user anyway.

I cherry-picked the CodeQL change, to check if I actually addressed all of them. It is very tricky to run the CodeQL locally, to see if you actually fixed the issues mentioned. I will remove that commit when I convinced myself it actually does solve those issues.

## Checklist for review

### DIFF
--- a/src/blitter/base.hpp
+++ b/src/blitter/base.hpp
@@ -170,7 +170,7 @@ public:
 	 * @param height The height of the buffer-to-be.
 	 * @return The size needed for the buffer.
 	 */
-	virtual int BufferSize(int width, int height) = 0;
+	virtual size_t BufferSize(uint width, uint height) = 0;
 
 	/**
 	 * Called when the 8bpp palette is changed; you should redraw all pixels on the screen that
--- a/src/cheat_gui.cpp
+++ b/src/cheat_gui.cpp
@@ -56,7 +56,7 @@ static int32 _money_cheat_amount = 10000000;
  */
 static int32 ClickMoneyCheat(int32 p1, int32 p2)
 {
-	Command<CMD_MONEY_CHEAT>::Post(p2 * _money_cheat_amount);
+	Command<CMD_MONEY_CHEAT>::Post(Money(_money_cheat_amount) * p2);
 	return _money_cheat_amount;
 }
 
--- a/src/ground_vehicle.cpp
+++ b/src/ground_vehicle.cpp
@@ -141,7 +141,7 @@ int GroundVehicle<T, Type>::GetAcceleration() const
 	}
 	/* Air drag; the air drag coefficient is in an arbitrary NewGRF-unit,
 	 * so we need some magic conversion factor. */
-	resistance += (area * this->gcache.cached_air_drag * speed * speed) / 1000;
+	resistance += static_cast<int64>(area) * this->gcache.cached_air_drag * speed * speed / 1000;
 
 	resistance += this->GetSlopeResistance();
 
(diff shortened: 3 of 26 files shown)


src/blitter/base.hpp
```diff
@@ -170,7 +170,7 @@ class Blitter : public SpriteEncoder {
 	 * @param height The height of the buffer-to-be.
 	 * @return The size needed for the buffer.
 	 */
-	virtual int BufferSize(int width, int height) = 0;
+	virtual size_t BufferSize(int width, int height) = 0;
```
nielsmh: Arguably the arguments should also be unsigned, a negative buffer dimension doesn't make sense.

src/cheat_gui.cpp
```diff
@@ -56,7 +56,7 @@ static int32 _money_cheat_amount = 10000000;
  */
 static int32 ClickMoneyCheat(int32 p1, int32 p2)
 {
-	Command<CMD_MONEY_CHEAT>::Post(p2 * _money_cheat_amount);
+	Command<CMD_MONEY_CHEAT>::Post(Money(p2) * _money_cheat_amount);
```
rubidium42: Conceptually I would do `Money(_money_cheat_amount) * p2` as the amount is the actual money, whereas p2 is just a multiplier.

src/ground_vehicle.cpp
```diff
@@ -141,7 +141,7 @@ int GroundVehicle<T, Type>::GetAcceleration() const
 	}
 	/* Air drag; the air drag coefficient is in an arbitrary NewGRF-unit,
 	 * so we need some magic conversion factor. */
-	resistance += (area * this->gcache.cached_air_drag * speed * speed) / 1000;
+	resistance += static_cast<size_t>(area) * this->gcache.cached_air_drag * speed * speed / 1000;
```
rubidium42: `resistance` is not a `size_t` but `int64`, so it is still broken for 32 bit builds.

src/blitter/base.hpp
TrueBrain: Totally agree, but I considered it out-of-scope of this PR :) Otherwise it becomes such a rabbit-hole :P

In good news, most of the time width/height are unsigned .. just sometimes they are not ;)

PS: bah, I was curious, so I fixed this one :P

src/ground_vehicle.cpp
TrueBrain: awh, and here I thought I checked them all :( Nicely spotted :D

src/cheat_gui.cpp
```diff
@@ -43,7 +43,7 @@
  * This variable is semantically a constant value, but because the cheat
  * code requires to be able to write to the variable it is not constified.
  */
-static int32 _money_cheat_amount = 10000000;
+static Money _money_cheat_amount = 10000000;
```
rubidium42: This change would require line 189 to change as well to SLE_INT64 (if that even works).

TrueBrain: No, it silently casts it to the right thing. But let's not make this PR more complicated than we have to ;) Fixed it by leaving it an `int32`

rubidium42 (approved):
LGTM ;)

Issue #6041, opened 2014-06-11
Huge linkgraph savegame section


DorpsGek:
**_Rubidium_ opened the ticket and wrote:**

> The savegame of https://www.tt-forums.net/viewtopic.php?f=31&t=70715 has a linkgraph section of about 275 MiB. This is more than the allowed 256 MiB, and shows some weakness of the linkgraph with respect to memory usage, especially in the savegame.
> 
> We need to figure out whether the savegame size can be reduced significantly in this case.
> 
> To load the savegame, use the attached diff.

**Attachments**
- [to_load.diff](https://bugs.openttd.org/task/6041/getfile/9857/to_load.diff) (0.84 KiB)

**Reported version**: _1.4.1_
**Operating system**: _All_

---

DorpsGek:
**_fonsinchen_ wrote:**

> r26646 reduces the size of the saveload chunk but is not straightforward to backport to 1.4 as it also bumps the saveload version.
> r26650 allows larger saveload chunks, of up to 4GiB
> 
> Some patches to backport r26646, reactivating the minor saveload version have been proposed. However, as r26650 can easily be backported I think that's unnecessary. With such large link graphs you already need a lot of RAM to keep the in-memory representation of the link graph around and the RAM requirements are still moderate compared to the CPU requirements of actually running such a game. On my computer the given savegame runs unplayably slow, but uses only about 500MiB to 1GiB of the 4GiB of RAM available. The save games are somewhat larger without r26646 but as the example shows they can be compressed very efficiently.
> 
> We could use a sparse vector or matrix implementation to optimize the in-memory representation of link graphs for size, but as most of the time link graphs are much smaller than the example given here I don't think it's worth it.

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/6041#comment13384_

DorpsGek:
**_Rubidium_ closed the ticket.**

**Reason for closing: _Fixed_**

> 

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/6041_

PR #10314, opened 2023-01-03
Codechange: Don't use a dense matrix to store LinkGraph edges.

michicc:
## Motivation / Problem

The LinkGraph uses a dense matrix to store information about links (edges) between stations (nodes).
This matrix can be potentially big and contains mostly unused entries since it is not used to store distances anymore (see #6041).
The savegame is already only storing the used entries of the matrix.

Additionally, the source and destination of each edge are purely determined positionally, which necessitates wrapper classes around nodes and edges and custom iterator implementations to walk the edges, just so the callers aren't exposed to this.
The `SmallMatrix` container used to store the edges is also not C++-aware, which is worked around by simulating constructors.

## Description

This PR moves the edges to be stored into a simple sorted `std::vector` in each node.
Each edge also stores its destination, so there's no need for any custom wrappers and iterators anymore.
The annotations of the LinkGraphJob are given reference to the source elements, removing the position dependence there, too.

As the linkgraph was the only user of `SmallMatrix`, it can go, too.

## Limitations

Not tested for performance.
Sparsely used dense matrices are usually not very cache-friendly, so not having them is probably good. Also, resizing the matrix is expensive.

Minimally checked for correctness as I am not an avid cdist player and don't have a large sample set of reference savegames at hand.

## Checklist for review


LordAro:
>Not tested for performance.

Would be very useful to get some numbers on this

src/saveload/saveload.h
```diff
@@ -344,6 +344,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_LAST_LOADING_TICK,                  ///< 301  PR#9693 Store tick of last loading for vehicles.
 	SLV_MULTITRACK_LEVEL_CROSSINGS,         ///< 302  PR#9931 v13.0  Multi-track level crossings.
 
+	SLV_LINKGRAPH_EDGES,                    ///< 303  PR#xxxxx Explicitly store link graph edges destination.
```
LordAro: you can do this now :)

michicc:
> Would be very useful to get some numbers on this

Anybody got some good test savegames for this? I don't play much with cargodist (well, more like at all), so my collection is severely lacking for something with noticeable linkgraph CPU load.

JGRennison:
I've pinched some ideas from here into my branch, and in doing so have added some extra things which might be worth considering here.

The number of EdgeAnnotation instances per NodeAllocation is relatively small, and does not change after initialisation.
To reduce the allocation overheads of std::vector instances per NodeAllocation, you could allocate all the EdgeAnnotation instances in a single std::vector or other contiguous block, and use std::span instances within NodeAllocation. This would probably get you better memory locality.
You can do the same thing for DemandAnnotation instances once the demand allocation phase is done, but this is a bit more fiddly.

In EdgeAnnotation, the `const LinkGraph::BaseEdge &base` reference could be removed outright.
It is (in most cases) 8 bytes, and only things which are actually needed through it are the capacity and the distance annotation (which can be precomputed), which also total to 8 bytes. This also would probably get you better memory locality.

LordAro (approved):
@JGRennison PRing your additional changes would be appreciated too

PR #10321, opened 2023-01-05
Fix: prevent corrupted GRF files to allocate stupid amounts of memory

rubidium42:
## Motivation / Problem

CodeQL complaining, and actually having a point that there is absolutely no check on the amount of memory that is going to get allocated.

## Description

Limit the amount of memory to allocate for a single sprite to 64 MiB, which is about 5 times more than the largest original sprite accounting for 4x zoom and 5 bpp. Though I'm not sure whether the end-game sprites get scaled by 4x at all, so even that might be a significant overestimate of the maximum sprite size.

## Limitations

Super huge sprites can't be loaded anymore, but where would we even draw such a sprite?

## Checklist for review

### DIFF
--- a/src/spriteloader/grf.cpp
+++ b/src/spriteloader/grf.cpp
@@ -59,6 +59,15 @@ static bool WarnCorruptSprite(const SpriteFile &file, size_t file_pos, int line)
  */
 bool DecodeSingleSprite(SpriteLoader::Sprite *sprite, SpriteFile &file, size_t file_pos, SpriteType sprite_type, int64 num, byte type, ZoomLevel zoom_lvl, byte colour_fmt, byte container_format)
 {
+	/*
+	 * Original sprite height was max 255 pixels, with 4x extra zoom => 1020 pixels.
+	 * Original maximum width for sprites was 640 pixels, with 4x extra zoom => 2560 pixels.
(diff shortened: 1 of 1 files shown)


src/spriteloader/grf.cpp
```diff
@@ -59,6 +59,15 @@ static bool WarnCorruptSprite(const SpriteFile &file, size_t file_pos, int line)
  */
 bool DecodeSingleSprite(SpriteLoader::Sprite *sprite, SpriteFile &file, size_t file_pos, SpriteType sprite_type, int64 num, byte type, ZoomLevel zoom_lvl, byte colour_fmt, byte container_format)
 {
+	/*
+	 * Original sprite height was max 255 pixels, with 4x extra zoom => 1020 pixels.
+	 * Original maximum width for sprites was 640 pixels, with 4x extra zoom => 2560 pixels.
+	 * Now up to 5 bits per pixel => 1020 * 2560 * 5 => ~ 12.5 MiB.
```
glx22:
```suggestion
	 * Now up to 5 bytes per pixel => 1020 * 2560 * 5 => ~ 12.5 MiB.
```

PeterN (approved):

PR #8749, opened 2021-02-26
WIP Feature: Setting to loop the same year forever

2TallTyler:
## Motivation / Problem

The goal of this PR is to add an option to freeze technological progress in a chosen year, by failing to advance the year after December 31st. While this can be faked using the cheat menu or “Vehicles never expire,” both are tedious and the latter doesn’t affect the introduction of new houses, towns building newer roadtypes, and other NewGRFs which progress through time.

The ability to remain in a certain year (really a period of time, but commonly a year) is often requested by players who want to explore a certain period in history without the frustration or immersion-breaking of the aforementioned hacks. Transport Fever 2 includes a setting to repeat a certain year forever, to support this style of play. I see several new opportunities in offering this game mode:
- Early-year games such as the American Wild West can actually reach a late-game network size and bank account without cheating time and money — without this, by the time a large network can be built, the era is long past.
- Industry sets can be designed for this mode without regard to supporting realistic technological progression, and can offer industry chains which were phased out long ago. Imagine sending ships to hunt whales, delivering coal to gasworks for town gas production, and harvesting ice from frozen lakes to ship around the world. Currently, the NewGRF author can either choose to force these industries to close, frustrating the player, or keep them around forever, which feels silly.
- Players who want to build in one time period and not have to worry about vehicle and tech upgrades can do so, without having to start their game in the future or use the aforementioned hacks. Many transport simulator games, like Cities in Motion, Cities: Skylines, Mini Metro, and Airport CEO; train simulators like Trainz, Derail Valley, Train Sim World, and MSTS; and logistics games like Factorio, Satisfactory, and Rise of Industry do not have history progression and let the player focus on other things.

## Description

#7938 proposed a “groundhog year” but did not consider the many potential side effects of looping a gameplay year. This PR is my attempt to address these and either make the necessary changes or confirm that no change is required. Thanks to those who participated in #8384 (especially @andythenorth) to suggest potential issues.

This also adds a "game year" which functions like the repeating years in Transport Fever 2, displaying a counter of how many years have passed since the game started. This means graphs, the company finance window, etc., still provide useful information.

There is plenty yet to be done, but I figure it's good to start the conversation before I get too far down rabbitholes like changing how time is measured. 🙂 This change to `_game_year` is the area where I'm least confident in my work. Please see the "Help needed" section in Limitations.

For the sake of reviews, I’ve tried to organize my changes into logical commits. We can squash later before merging.

### Modified to work with looped years:
[...]

PR #10322, opened 2023-01-05
Feature: Change speed of calendar progress

### LINKED ISSUE #10205: [Bug]: MILLISECONDS_PER_TICK should be 27, not 30
### Version of OpenTTD

All of them

### Expected result

Running TTD(Patch) and OpenTTD side-by-side should result in the game time progressing at roughly the same rate.

TTD and TTDPatch both appear to run at approximately 27 ms per tick, which results in 37 ticks per second. This puts a game day of 74 ticks at exactly 2 seconds. A game year is 12m10s.

### Actual result

Running TTD(Patch) and OpenTTD side-by-side, the TTD(Patch) game runs faster. After one month, OpenTTD is behind by around 3 days, 2 months it is behind by approximately 6 days.

OpenTTD actually uses 30 ms per tick, resulting in 33.333 ticks per second. A game day of 74 ticks is 2.466 seconds. A game year is 15 minutes.

### Steps to reproduce

1. Start TTD.
2. Start OpenTTD.
3. Compare the dates (easier with TTDPatch).

🤷

2TallTyler:
## Motivation / Problem

In the [daylength](https://github.com/OpenTTD/OpenTTD/discussions/8397#discussioncomment-224382) discussion, @andythenorth gives motivations for players to configure how fast technology progresses in OpenTTD:

> Play 1960 trains forever!
> 
> For many players, the game seems to progress quickly. A new generation of vehicles appears, so we upgrade and tune our networks. Then just as we finish, boom! A new generation of vehicles appears.
> 
> On busy or large maps it can be hard to keep up.
> 
> Slower progression of vehicles would make a more relaxing games for some players.

Additional usecases might be:

* A GS freezes technology progress and advances it when the player completes transport goals ("research" in many games).
* Early-year games such as the American Wild West can actually reach a late-game network size and bank account without cheating time and money — without this, by the time a large network can be built, the era is long past.
* Industry sets can be designed for frozen progression without regard to supporting realistic technological progression, and can offer industry chains which were phased out long ago. Imagine sending ships to hunt whales, delivering coal to gasworks for town gas production, and harvesting ice from frozen lakes to ship around the world. Currently, the NewGRF author can either choose to force these industries to close, frustrating the player, or keep them around forever, which feels silly.

## Description

This implements @nielsmh's [approach described](https://github.com/OpenTTD/OpenTTD/pull/9789#issuecomment-1013852571) in my previous attempt to implement this (#9789).

Date/time handling is separated into two systems:
#### Calendar time for technology and world appearance
- [x] Vehicle introduction and obsolescence
- [x] Available airports, stations, objects, etc.
- [x] NewGRF variables which can be used to determine visual styles or behavior based on the time of year
- [x] Snow line movement

#### Economy time
- [x] Industry and house production/consumption, etc.
- [x] Town growth
- [x] Industry production changes, closure, and spawning
- [x] NewGRF variables for things like FIRS production boosts
- [x] AI/GS API additions for things like town growth scripts
- [x] Running and maintenance costs, loan interest, etc.
- [x] Vehicle and group profit statistics
- [x] Finance graphs
- [x] Cargo payment rates graph
- [x] Local authority reputation, station ratings, etc.
- [x] Vehicle service interval and time since last service
- [x] Linkgraph tooltips and update interval setting
- [x] Autosave frequency and other settings
- [ ] Network game duration in server list (replacing "Years") - @TrueBrain will help with this before merging.

Calendar time is kept in a common Gregorian calendar, with 365 days in a year, leap years, variable length months, and so on.

Economy time is kept in (approximate) real time. There are logical days with a duration of 1.998 seconds, logical months with a duration of 1 minute, and logical years with a duration of slightly more than 12 minutes.

Internally a calendar with 12 months of 30 days is observed. There are no leap periods. 

In the user interface, all elements that run on economy time are presented in real time units:
#### Seconds (multiples of 2), for things which were originally daily
- [x] Timetables
#### Minutes, for things which were originally monthly:
- [x] Industry production per minute
- [x] Passengers last minute
- [x] Vehicle service intervals and time since last service
- [x] Subsidies
 #### Economic periods (sometimes described as simply 12 minutes) for things which were originally yearly:
- [x] Running costs
- [x] Company finances
- [x] Vehicle and group profit statistics
- [x] Company performance rating (league table) 

Internally most events will still be referenced as daily/monthly/yearly.

## Limitations

This changes MILLISECONDS_PER_TICK from 30ms to 27ms to make an in-game day last 1.99 seconds instead of 2.2 seconds, making real-world time feasible. This actually matches TTD and closes #10205. 😉 

## Checklist for review


PeterN:
Your checklist doesn't mention town growth, is that covered under house generation?

src/date_type.h
```diff
@@ -22,14 +22,18 @@ typedef uint8  Day;   ///< Type for the day of the month, note: 1 based, first d
...
+ * 1 tick is approximately 27 ms.
+ * 1 day is thus about 2 seconds (74 * 27 = 1998) on a machine that can run OpenTTD normally
  */
 static const int DAY_TICKS         =  74; ///< ticks per day
 static const int DAYS_IN_YEAR      = 365; ///< days per year
 static const int DAYS_IN_LEAP_YEAR = 366; ///< sometimes, you need one day more...
 static const int MONTHS_IN_YEAR    =  12; ///< months per year
 
+static const int DAYS_IN_ECONOMY_MONTH  = 30;  ///< economy days per econony month -- at 2 seconds a day this equals 1 minute per month
+static const int MONTHS_IN_ECONOMY_YEAR = 12; ///< months in economy economy year
```
James103:
```suggestion
static const int DAYS_IN_ECONOMY_MONTH  = 30; ///< economy days per econony month -- at 2 seconds a day, this equals 1 minute per month
static const int MONTHS_IN_ECONOMY_YEAR = 12; ///< months in economy economy year
```

src/gfx_type.h
```diff
@@ -313,8 +313,12 @@ enum SpriteType : byte {
 	ST_INVALID  = 4,      ///< Pseudosprite or other unusable sprite, used only internally
 };
 
-/** The number of milliseconds per game tick. */
-static const uint MILLISECONDS_PER_TICK = 30;
+/**
+ * The number of milliseconds per game tick.
+ * The value 27 together with a day length of 74 ticks makes one day 1998 milliseconds, almost exactly 2 seconds.
+ * With a 2 second day, one standard month is 1 minute, and one standard year slightly over 12 minutes.
```
James103:
```suggestion
 * With a 2 second day, one standard month is 1 minute, and one standard year is slightly over 12 minutes.
```

src/lang/english.txt
```diff
@@ -3494,10 +3506,10 @@ STR_LOCAL_AUTHORITY_ACTION_BRIBE                                :Bribe the local
...
 STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_STATUE_OF_COMPANY            :{PUSH_COLOUR}{YELLOW}Build a statue in honour of your company.{}Provides a permanent boost to station rating in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
 STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_NEW_BUILDINGS                :{PUSH_COLOUR}{YELLOW}Fund the construction of new buildings in the town.{}Provides a temporary boost to town growth in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 1 year's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 12 minute's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
```
James103:
```suggestion
STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 12 minutes of exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
```

Alternatively, if the apostrophe is to be kept, that part could instead say "Buy 12 minutes' exclusive transport rights ..."

src/saveload/saveload.h
```diff
@@ -344,6 +344,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_LAST_LOADING_TICK,                  ///< 301  PR#9693 Store tick of last loading for vehicles.
 	SLV_MULTITRACK_LEVEL_CROSSINGS,         ///< 302  PR#9931 Multi-track level crossings.
 
+	SLV_NOCALENDAR,                         ///< 303  PR#xxxx Separate calendar time and economy time.
```
```suggestion
	SLV_NOCALENDAR,                         ///< 303  PR#10322 Separate calendar time and economy time.
```

James103 (commented):
I found some minor items whose grammar could be improved.

LC-Zorg:
The preview game crashes very quickly (once when I trying to build a station, the second time when the vehicle arrived at the station, the third time when I was not doing anything), so it's hard to judge how it works. Overall it might be good maybe very good but I have a few doubts.

1 - Release dates for new vehicles
Will a vehicle that should appear in 1990 always appear in (ca.)1990, or is it dependent on the calendar setting or the game start date?

2 - Running costs, production, everything else based on real time
Are you planning this as an option or the only choice?

3 - Percentage setting of the calendar speed
Whenever I set costs in any add-on, I think why there's just no percentage setting, it would make it so much easier to customize different add-ons. Here, however, I'm not really sure if it's better than the form used by JGR or kaomoneus. Not a very important note, just thinking out loud ;)

4 - Broader game pace setting instead of just calendar pace
What do you think about making it a group of settings rather than a single setting? I mean, the pace of production, including the amount of passengers, could be regulated in a similar way. Thanks to this, you could create a game where not only time changes slower (or faster), but also where everything except the pace of vehicles is slower, as in the JGR version. Such a game, when you don't have to use thousands of vehicles on every connect and can focus on the development of a wider world, can be equally interesting. Something like that:

Settings > Generating world > Game pace:
- calendar progress speed
- industries / economy speed - it determines how fast, how much enterprises and cities produce and thus how fast you earn

Yes, this idea probably excludes the third point. :)

nielsmh:
> 1 - Release dates for new vehicles Will a vehicle that should appear in 1990 always appear in (ca.)1990, or is it dependent on the calendar setting or the game start date?

Why would it depend on the game start date? When the calendar shows 1990 things that appear in 1990 will appear.

> 2 - Running costs, production, everything else based on real time Are you planning this as an option or the only choice?

It is real-time, and real-time only. Real-time based on the wall clock in your real room, nothing production or earning or scoring related is based on the game calendar.
Production happens on the same game tick increments as it has always done, vehicles move the same distance per game tick as they have always done, running costs might be slightly different because they are calculated per 30*74 ticks (one "economy month) instead of per 28, 29, 30 or 31 days of 74 ticks.

> 3 - Percentage setting of the calendar speed Whenever I set costs in any add-on, I think why there's just no percentage setting, it would make it so much easier to customize different add-ons. Here, however, I'm not really sure if it's better than the form used by JGR or kaomoneus. Not a very important note, just thinking out loud ;)

You need to re-state this question. The game setting is already configured in percent.
Personally I think it would be better to define the technology calendar day length in game ticks (default = 74) than a percentage setting.

> 4 - Broader game pace setting instead of just calendar pace What do you think about making it a group of settings rather than a single setting? I mean, the pace of production, including the amount of passengers, could be regulated in a similar way. 

In my understanding/opinion, this is out of scope.

LC-Zorg:
1
Because economic time is different from calendar time. This fact alone raises some doubts.

2
I understand that everything works the same (with a slight acceleration). It's about the way it's presented and the point of reference used. I believe that presenting the level of production, the number of passengers, and especially costs and revenues in relation to real time (minutes) as the ONLY form is not a good way to implement this function.
minutes

3
It's not really important. Any other form, including ticks, can be added if it will be needed.

4
Yes, it's not something that needs to be done now (or ever). Just an idea that might be related to this change.

2TallTyler:
It's possible, although it would require changing timetables from using dates to using ticks. I don't think many people use timetables for anything beyond separating vehicles with shared orders and I plan to upstream the JGRPP autoseparation feature soon™️ (my next big project after this), so I'm not convinced it's worth the effort. Rounding to even seconds is only a visual change, after all -- the timetable resolution is unchanged.

2TallTyler:
Notes on the company finance window: Before this PR, it showed data separated by year. This is economic data and shouldn't be scaled with calendar time, so I've moved it to economy time where it will continue to be broken up into 12-minute periods of gameplay.

How do I label each period? A few ideas I've had and rejected:
- Relative time in minutes: `36 minutes ago, 24 minutes ago, 12 minutes ago` (many ways to write it). Problem: relative time seems like it should be a rolling total, but the data is in discrete periods which start from scratch every New Year's Day (which no longer matches up with the calendar, even at 100% speed since economy years only have 360 days).
- Use calendar dates, scaled to the calendar progress speed, `Jan 1-Jan 30, Jan 31-Mar 2, Mar 3-Apr 3 ..` Problem: this seems incredibly arbitrary and doesn't work at all if calendar progression is frozen entirely.

The eventual solution:
- Abstract "production periods" or "pentas" of 12 minutes, `0, 1, 2 ..`.  Problem: the player has no way to know how far through each period they are, since economy time is (intentionally) not displayed anywhere. This is currently the "least bad" option, although I welcome suggestions.

Note: As with many of these changes, continuing to allow the use of calendar years, either automatically if the calendar progress speed is 100%, or by an optional setting, or not possible because economic years don't match calendar years: they only have 360 days per "year".

James103:
Regarding the timetable, everything timetable-related doesn't have to be in seconds; some players may be better off having timetable real times displayed in (hours), minutes and seconds (instead of just seconds).

For example, instead of "This timetable will take 1,234 seconds to complete", that same string can instead say "This timetable will take 20 minutes and 34 seconds to complete". Instead of "A: in 123 sec", there could be "A: in 1m 23s" or "A: in 01:23".

2TallTyler:
I tried splitting times into minutes/seconds, and although I didn't quite get the string parameters right, I decided it looked too cluttered with twice the numbers. Plus, if you're trying to do the math you'll have to convert minutes/seconds to just seconds. Either way is a tradeoff for one group, and I think all seconds is much easier to read.

All seconds:
seconds

Mixed minutes/seconds:
dparams

PeterN:
> Mixed minutes/seconds:
Stay for 5s
Travel for 1:04

FLHerne:
Not sure who that would benefit.

People using timetables for anything besides separation need to do calculations on the timetable periods; displaying times in minutes:seconds will make that more complicated.

The other people probably won't have such long waiting times at all.

PeterN:
But that IS the timetable period, it's economy ticks (not calendar ticks) converted into seconds. Displaying as calendar days doesn't make sense if time is slowed or stopped.

FLHerne:
> But that IS the timetable period, it's economy ticks (not calendar ticks) converted into seconds. Displaying as calendar days doesn't make sense if time is slowed or stopped.

I meant that IMO they should be in seconds, even for periods >60s, rather than minutes:seconds.

(tbh I still find the real-world time thing a bit jarring, but have no better ideas for that purpose)

James103:
By changing vehicle daily loops to run off economy time, that means vehicles age based off economy time, meaning various age displays will need changing. Would you prefer age to be in minutes or keep it displayed in 12-minute economic periods?

Option A:
> Age: 125 minutes (180)
> Life: 180 minutes

Option B:
> Age: 10 periods (15)
> Life: 15 periods

Current:
> Age: 10 years (15)
> Life: 15 years

2TallTyler:
Good catch. Vehicle age is a gameplay element which impacts replacement (autorenew), station ratings, etc., so it should probably be moved to economy time.

I'll keep it in periods, because that's the unit used for running costs and profit stats.

James103:
Regarding updating subsidies to economy time, these strings in the subsidy list window may need updating:

Subsidy on offer
Current:
> Passengers from Parnford to Wuham (by January 1951)

With economy time:
> Passengers from Parnford to Wuham (12 minutes left)

Subsidy awarded:
Current:
> Passengers from Parnford to Wuham (Parnford Transport, until January 1952)

With economy time:
> Passengers from Parnford to Wuham (Parnford Transport, 12 minutes left)

LC-Zorg:
@2TallTyler I have one important question: will it still be possible to play with the division of time into days, months and years, or will this be the only way to present time?

2TallTyler:
Thanks, @James103. I'd implemented that earlier but the commit got lost in a rebase. It's now found. 🙂 

@LC-Zorg: Unfortunately, with this approach it would not be possible to display economy time in calendar units. While the information is still stored correlated to an economy date, these do not align with calendar dates because "economy months" (which last 1 minute) only have 30 "days", thus an "economy year" has 360 days. Trying to sync the two would be an exercise in futility and even if possible, would create lots of new bugs. I would prefer to support both dates and real-time units, but I don't see a viable way to do so.

LC-Zorg:
I didn't want to believe it, it seemed too absurd to be real, but you are actually doing it... 
I consider the very idea of the length of the calendar year to be different from the economic year to be strongly flawed and bad even if it were an option, but as an option it would not be something I could not accept. I believe it's just messing up the game's code even more, which will make further development of the game even more difficult, but ultimately as a player I might just never use it. But if it's a change that doesn't give a choice, I think it's radically wrong. For me, it just ruins the game, and I'm sure a lot of other players will feel the same way. Time, especially in an economic game, is not an insignificant detail, but one of its foundations. All financial statements are on an annual, quarterly or monthly basis. It is never presented in terms of 90 or 360 days. To anyone, or at least to many of those who know anything about economics, this will look like a caricature. Once a financial summary in March, another time in July, another time in October. Various costs that are charged at the beginning of the month, with this change, this term will change all the time. This is grotesque! Another big caveat of mine is that without asking, you completely change the player's point of view from the position of a participant in the events to the highly technical position of an outside observer. I don't know if you see it, but that's how you change the nature of the game. This is not a minor change or improvement. What you are doing is a revolution that has nothing to do with the development rules of this game. In doing so, you are destroying the entire history of this game, where time has always been shown in the form of days, months and years in the game. By doing so, you ignore the habits and preferences of many players, as well as what they may have liked about the game.

The function of changing the pace of the time passage can be implemented in a way that does not affect the core game in any way.

2TallTyler:
Thank you for sharing your concerns. Please know that I don't propose such a major change carelessly. (And that it is just that: a proposal, not guaranteed to be approved and merged into OpenTTD.)

> The function of changing the pace of the time passage can be implemented in a way that does not affect the core game in any way.

I would welcome a technical proposal to achieve this. I've made other attempts before (#8749, #9789) and the problem that always results is that separating technology and economy time results in two time systems. Either the player is presented with both (confusing) or one is hidden (even more confusing), which is why in this PR I've converted economy time to real-time with relative measurements instead of fixed points in time — neatly solving this issue. Everybody knows how long seconds, minutes, etc., are.

The only alternative I have not tried is JGRPP daylength with tileloop updates (ocean flooding, grass growing, etc.) properly disconnected from calendar dates and with the economy scaled automatically to retain the vanilla economy even at slow calendar progression. This would require automatically choosing a suitable date unit to display things like timetables, service intervals, etc., to have enough precision even at slow day lengths. For example, if an in-game day takes 5 real hours, choosing service intervals in days (default: 150 days, or 5 real minutes) is not acceptable, and it would need to be changed to in-game hours or minutes. This would be...problematic to implement.

LC-Zorg:
> Please know that I don't propose such a major change carelessly.

But the problem is, no matter how much you refine this change, it's still going to be bad. Maybe not for everyone, but definitely for many.

> I would welcome a technical proposal to achieve this.

This probably won't to be a technical proposition, but my observation is that even today it is possible to use real time in game. It is not necessary to create artificial and unsuitable units of economic time. Today, it is possible to save and display the production level on a monthly, weekly, daily and actual minute basis (2222 ticks). If you slow down time, it's still possible. The month will still count from 1 to 1, the week will still have 7 days (the in-game date should then show what day of the week it is), and the actual minutes will have more ticks. The same can apply to all other aspects of gameplay. The timetable can also be shown today in real seconds, and with a greater slowdown of the game also in hours and minutes from the game world. Each player could use the form that would suit him best, but the "automatic" setting could also be used, which could be the default and would automatically select the most optimal form depending on the level of slowing down the time passage. Perhaps it would also be better to use the fractional tempo setting, i.e. 1/2, 1/3, 1/4, ... 1/649 (1 real second = 1 in-game minute), so it might be easier to calculate how much ticks should contain one real second and one minute. As for stopping the flow of time completely, if it were possible to slow down to 1/649, I don't know if that would still make sense, but if so, I think adding a "Groundhog year" option wouldn't be a bad thing. It could be in the same tab as the time rate setting. Both options could complement each other. Compared to completely stopping time, this would be better because it would be possible to change the date and seasons (snow or other things), and it would also be possible to summarize the year normally, but all the time the same. It can be summarized quarterly or month instead of year. Yes, the game would probably have to create and save a lot more data. In order to limit their excess, it would be possible to limit the possibility of choosing too short intervals in relation to the selected pace of the passage of time.

LordAro:
Alright that's enough. No more unfounded speculation from you. Any more and there will be consequences.

You don't like it, you don't use it, even under the assumption it goes in as-is.

planetmaker:
I'd like to encourage you to go forward with this approach. Daylength (or however one calls it) has been controversial for ages - and there probably is no ideal solution, and definitely none to satisfy everybody. This looks very reasonable and workable.

michicc:
I would like an OpenTTD with some kind of daylength, and I don't have any specific implementation concept that I would consider superior over another.

But I still do want to point out two things that should at least be taken into consideration:
* Quite some things in NewGRF-land are defined by days or months, like the industry production callback. Currently, this means that the callback is called 7 or 8 times a month, and I think that ECS does actually care about this variance. Whether this is something that should be preserved is certainly debatable, but a non-calendar economy time would make it more difficult.
* Seconds are only actually seconds if the game speed is exactly 1x, which it won't when fast-forwarding or not having enough computing power. Thus, strictly speaking, you'd still need to show a second "real-time" clock, which might or might match your PC clock.

In my personal opinion (that is in no way more important than any other opinion), I quite like having a single consistent time base. Traditionally, this has been the calendar date. I am quite aware that if you change the literal day length, some things in OpenTTD that are shown in days don't really make sense (like the service interval). It would still be possible to scale them on display only, like e.g. timetable durations that are internally always stored in ticks. For high daylength factors, some sub-unit like hours would probably be needed, just to have a more intuitive units to ticks that is better resolution than days.

A decoupled time base is probably advantageous if you want fine-grained daylength factors over a large range, but would make any kind of consistent game balancing even more impossible than it already is. We might have crossed that game-balance bridge long ago though.
In contrast, some games like e.g. Transport Fever 2 offer only a very limited daylength subset which has only 1x, 2x and 4x. It basically just applies a rename in a few places so that 1980, 1981, 1982, 1983 become either 1980 H1, 1980 H2, 1981 H1, 1981 H2 or 1980 Q1, 1980 Q2, 1980 Q3, 1980 Q4. Makes for an easy implementation and at least a chance of game balance, but is very limiting in possiblities.

TL;DR: There's no solution that will make everybody happy.

PeterN:
Is it possible to have economy time related to calendar time, so the length of the economy month is variable instead of moving to fixed 30-economy days?

At 1x, there'd be 1:1 relationship with economy and calendar, and the game is effectively the same as without this.

Changing day length mid-game would make it go out of sync though, but that may not be a huge problem -- it would just be as the PR already is.

2TallTyler:
Superseded by #10605, in which real-world time display is optional.

PR #10327, opened 2023-01-07
Resolve CodeQL's inconsistent nullness checks

rubidium42:
## Motivation / Problem

CodeQL complaining about inconsistent nullness checks. For example `v->Next()` being checked for `nullptr` in some cases and not in other cases.
Reviewing the alerts did not bring any shocking bugs to light, except the potential to crash OpenTTD with a corrupt savegame.

With this #10299 will not show the red cross denoting a build failure anymore, but get a nice green checkmark because all the security alerts and errors have been resolved.

## Description

When loading data from a savegame, and it could conceivably be `nullptr` in the savegame but should not be, add a check and mark the savegame as corrupt.
For getting the main window, originally `FindWindowByID(WC_MAIN_WINDOW, 0)`, introduce `GetMainWindow()` that asserts that the `Window` actually exists, and use that throughout the code. Also check whether the current checks for `nullptr` on it were actually necessary; they do not seem to be.
The rest is mostly adding `assert(x != nullptr);` to "document" the check has been explicitly left out (for release builds).

## Limitations

None that I'm aware of.

## Checklist for review

### DIFF
--- a/src/newgrf_gui.cpp
+++ b/src/newgrf_gui.cpp
@@ -1464,6 +1464,9 @@ private:
 				this->avails.push_back(c);
 			} else {
 				const GRFConfig *best = FindGRFConfig(c->ident.grfid, HasBit(c->flags, GCF_INVALID) ? FGCM_NEWEST : FGCM_NEWEST_VALID);
+				/* Never triggers; FindGRFConfig returns either c, or a newer version of c. */
+				assert(best != nullptr);
+
 				/*
 				 * If the best version is 0, then all NewGRF with this GRF ID
 				 * have version 0, so for backward compatibility reasons we
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -1179,6 +1179,18 @@ Window *FindWindowByClass(WindowClass cls)
 }
 
 /**
+ * Get the main window, i.e. FindWindowById(WC_MAIN_WINDOW, 0).
+ * If the main window is not available, this function will trigger an assert.
+ * @return Pointer to the main window.
+ */
+Window *GetMainWindow()
+{
+	Window *w = FindWindowById(WC_MAIN_WINDOW, 0);
+	assert(w != nullptr);
+	return w;
+}
+
+/**
  * Close a window by its class and window number (if it is open).
  * @param cls Window class
  * @param number Number of the window within the window class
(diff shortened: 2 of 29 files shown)


src/window.cpp
```diff
@@ -1178,6 +1178,17 @@ Window *FindWindowByClass(WindowClass cls)
 	return nullptr;
 }
 
+/**
+ * Get the main window, i.e. FindWindowById(WC_MAIN_WINDOW, 0).
+ * This function asserts the main window is available.
+ * @return Pointer to the main window.
+ */
+Window *GetMainWindow() {
```
LordAro:
```suggestion
Window *GetMainWindow()
{
```

LordAro (commented):
Feels like there are a lot of "redundant" asserts being added. Hope it doesn't affect debug builds too much...

src/window.cpp
```diff
@@ -1178,6 +1178,18 @@ Window *FindWindowByClass(WindowClass cls)
 	return nullptr;
 }
 
+/**
+ * Get the main window, i.e. FindWindowById(WC_MAIN_WINDOW, 0).
+ * This function asserts the main window is available.
```
TrueBrain: missing `if` I think?

Reading again, I get what you mean, but it reads difficult. How about:

`@note  This function asserts if the main window is not available.`

Or maybe:

`This function asserts that the main window is available`

I just don't know, something is odd, and I can't put my finger on it :P

src/newgrf_gui.cpp
```diff
@@ -1464,6 +1464,7 @@ struct NewGRFWindow : public Window, NewGRFScanCallback {
 				this->avails.push_back(c);
 			} else {
 				const GRFConfig *best = FindGRFConfig(c->ident.grfid, HasBit(c->flags, GCF_INVALID) ? FGCM_NEWEST : FGCM_NEWEST_VALID);
+				assert(best != nullptr); // At worst case the GRF of this loop iteration is found
```
TrueBrain: I fail to decipher what the comment means?

rubidium42: Is `Best will never be nullptr, as it will either find a newer NewGRF or it will find itself, i.e. best == c.` better?
Yeah, in hind sight it's not a really good comment. No idea why exactly anymore, although "itself" of the proposal is not really clear on its own as it's not super obvious whether that refers to `best` or `c`. That might be why it was mentioning the loop iteration, so it points to `c` but that didn't make it much clearer.

src/window.cpp
rubidium42: What about `@note If the main window is not available, this function will trigger an assert.`? Maybe without the `@note`?

I agree it's difficult to convey that it behaves slightly differently than just `FindWindowById(WC_MAIN_WINDOW, 0)`, but for a good reason: now you know it ain't `nullptr` and otherwise you don't.

TrueBrain: One could argue it should return `Window &`, and not have this comment at all :P I am fine with your suggestion, with or without `@note`.

src/newgrf_gui.cpp
TrueBrain: Ah, so, you mean to say: will never trigger. Maybe that is the valuable piece of information here. Suggestion:

`Never triggers; FindGRFConfig returns either c, or a newer version of c`

Or, as extension to your suggestion:

`Cannot trigger: best will never be nullptr, as it will either find a newer NewGRF or it will find itself, i.e. best == c`

But yeah, even without the "cannot trigger" this comment is a lot more clear to me :) Go for it!

TrueBrain (approved):

PR #8051, opened 2020-03-30
Fix: Don't rerandomise a script's settings when loading from a save

LordAro:
ScriptConfig stuff is all pretty hairy, but I'm pretty sure this is what's wanted. Besides, Samu's AfterFix AI is literally the only one on BaNaNaS that makes use of the random settings anyway.

Closes #7486

PR #7515, opened 2019-04-15
Fix: Really check for start_date once an AI company slot becomes available.

SamuXarick:
<Samu> basically, if there's no slots, it checks daily, using the interval 3601 - 3600
<Samu> and never reaches 0
<Samu> if it gets a start_date lower than that 3601-3600 interval, then the real countdown starts, using that start_date value. When it reaches 0, it tries to make ai company

PR #7486, opened 2019-04-08
Fix: AI/GS settings with the flag SCRIPTCONFIG_RANDOM could be altered after loading from a savegame.

SamuXarick:
Non-Random AI/GS configured in the main menu with their settings left at their defaults, could have their SCRIPTCONFIG_RANDOM flagged settings randomized when loading from a saved game.

This patch anchors them as unchangeable to prevent game load from randomizing them.

Issue #6460, opened 2016-05-06
Can't access Random AI Parameters in a game


DorpsGek:
**_Samu_ opened the ticket and wrote:**

> Steps:
> \- launch OpenTTD
> \- click AI/Game Script Settings
> \- Configure at least 1 Random AI...
> (screenshot 1) - when a Random AI is selected in the main menu, the Configure button is not disabled.
> \- start new game
> \- open Options > AI/Game script settings
> \- select the Random AI...
> bug: (screenshot 2) - when a Random AI is selected in a game, the Configure button is disabled.
> \- Click Select AI, chose Random AI again and click Accept.
> (screenshot 3) - when a AI slot changes its script to Random AI, the Configure button becomes enabled.
> 
> In my opinion, the indended behaviour would be to have the Configure button not disabled in all these cases.

**Attachments**
- [1 - in menu.png](https://bugs.openttd.org/task/6460/getfile/10514/1+-+in+menu.png) (95.88 KiB)
- [2 - in game, before changing Random AI.png](https://bugs.openttd.org/task/6460/getfile/10515/2+-+in+game%2C+before+changing+Random+AI.png) (100.84 KiB)
- [3 - in game, after changing Random AI to Random AI.png](https://bugs.openttd.org/task/6460/getfile/10516/3+-+in+game%2C+after+changing+Random+AI+to+Random+AI.png) (100.84 KiB)

**Reported version**: _1.6.0_
**Operating system**: _Windows_

---

DorpsGek:
**_Samu_ wrote:**

> I discovered something more.
> 
> \- follow the same steps up to screenshot 2
> \- save the game, load it back
> \- open Options > AI/Game script settings
> \- select the Random AI...
> ... it becomes exactly as in screenshot 3
> 
> 

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/6460#comment14179_

DorpsGek:
**_Samu_ wrote:**

> https://paste.openttdcoop.org/phsw73z9z
> 
> The problem is the missing start_date when starting a new game. This fixes it.

**Attachments**
- [add start_date parameter for Random AIs on new game v1 r27603.patch](https://bugs.openttd.org/task/6460/getfile/10562/add+start_date+parameter+for+Random+AIs+on+new+game+v1+r27603.patch) (0.53 KiB)

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/6460#comment14226_

PR #10330, opened 2023-01-07
Codechange: Rework 'start_date' parameter for AIs as a game setting

SamuXarick:
## Motivation / Problem
The point of the PR is to eliminate the need to control when "start_date" parameter is pushed into an AI config, then control when it's to be randomized or deviated, also to make it more predictable when loading from a savegame, what it initializes to, make the savegame thus more reproducible. By making it a game setting, it is removed from the AI config, no more need to babysit it.

See #7515, #7486, #8051, #6460

## Description
StartDelay

`start_date` is removed from the list of AI parameters, and is added as `ai_start_next` on the list of game settings and on AI settings where you setup which AIs to start.

## Limitations
Moves 'start_date' AI parameter away from scripts losing its "per-AI" customization, into a game setting. It is now a "one for all" approach for the feature.
Default value was set to that of Easy Preset.

## Checklist for review

### DIFF
--- a/src/ai/ai_core.cpp
+++ b/src/ai/ai_core.cpp
@@ -15,6 +15,7 @@
 #include "../network/network.h"
 #include "../window_func.h"
 #include "../framerate_type.h"
+#include "../script/api/script_object.hpp"
 #include "ai_scanner.hpp"
 #include "ai_instance.hpp"
 #include "ai_config.hpp"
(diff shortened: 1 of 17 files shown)


src/ai/ai_core.cpp
```diff
@@ -293,13 +294,17 @@
 
 /* static */ int AI::GetStartNextTime()
 {
-	/* Find the first company which doesn't exist yet */
-	for (CompanyID c = COMPANY_FIRST; c < MAX_COMPANIES; c++) {
-		if (!Company::IsValidID(c)) return AIConfig::GetConfig(c, AIConfig::SSS_FORCE_GAME)->GetSetting("start_date");
+	int start_next = _settings_game.ai.ai_start_next;
+	if (start_next > 0 && AI::START_NEXT_DEVIATION != 0) {
```
glx22: Useless !=0 check as `AI::START_NEXT_DEVIATION` is always 60

src/ai/ai_core.cpp
```diff
@@ -293,13 +294,17 @@
...
+		/* Add random deviation */
+		start_next += InteractiveRandomRange(AI::START_NEXT_DEVIATION * 2 + 1) - AI::START_NEXT_DEVIATION;
+
+		/* ai_start_delay = 0 is a special case, where random deviation does not occur.
+		 * If ai_start_delay was not already 0, then a minimum value of 1 must apply. */
+		start_next = Clamp(start_next, std::max((int)AI::START_NEXT_MIN, 1), AI::START_NEXT_MAX);
```
glx22: `AI::START_NEXT_MIN` is 0, so std::max will always return 1

src/openttd.cpp
```diff
@@ -386,9 +386,6 @@ void MakeNewgameSettingsLive()
 		_settings_game.ai_config[c] = nullptr;
 		if (_settings_newgame.ai_config[c] != nullptr) {
 			_settings_game.ai_config[c] = new AIConfig(_settings_newgame.ai_config[c]);
-			if (!AIConfig::GetConfig(c, AIConfig::SSS_FORCE_GAME)->HasScript()) {
-				AIConfig::GetConfig(c, AIConfig::SSS_FORCE_GAME)->Change(nullptr);
-			}
```
glx22, line 394: This change feels wrong

glx22 (commented):
I didn't check GUI stuff in details, but it feels more complicated than it needs to be (I can be wrong).

src/openttd.cpp
SamuXarick: This was the fix for #6460. The "start_date" parameter for random ais was at times not part of their configuration. This PR undoes that fix because "start_date" is no longer part of an AI configuration.

src/ai/ai_core.cpp
SamuXarick: I see, I was trying to be thorough. The equivalent code had those checks in place. As long as no one changes the defaults, I can remove the useless check.

src/ai/ai_core.cpp
SamuXarick: Same as above.

PR #10340, opened 2023-01-12
Feature: Set a custom number of industries in map generation window

### LINKED ISSUE #10339: [Bug]: Industry density settings are not granular enough
### Version of OpenTTD

12.0

### Expected result

2000*2000 map, alpine, terragenesis. No industry GRF's

If I set "no of industries" to "Minimal", I get precisely 1 of each. Ok, so that's definitely minimal, I'd prefer some more than that.
I set it to "Very Low", and the map is absolutely littered with industries!

image

That isn't what I'd call a "very low" number, even for a 2k map: There's almost 100 coal mines there!

Set it to "Low" and I get over a 1200 industries on the map!

Set it to "Normal" and now I'm looking at almost 900 combined coal mines and farms. Probably 2400 industries total! The "Normal" map looks like this:
image
image

"High" tried to generate 5120 industries. How do I know? Because it took so long to generate them all the dialog was up on the screen for several seconds! But the map itself only has about 3500 - doesn't look like it could fit the rest on!

====

"Very Low" on 256 * 256 gets me 1 of each industry only though.
"Very Low" on 1k * 1k gets quite a lot of industries (though proportionally far less than 2k).
"Very Low" on 4k * 4k gets you 2500 industries!

=====

To me at least, "Very Low" on the larger maps is what I expect "High" to look like.

### Actual result

As above

### Steps to reproduce

As above

2TallTyler:
## Motivation / Problem

As described in #10339, the existing industry density settings are not very granular, especially on larger maps.

## Description

Upstreams a JGRPP commit (b69521ea87800dc44d7e3a4a2632ea34a439e04e) which adds a custom industry number, like the existing custom town number.

Like with custom town number, the custom industry number is stored in `openttd.cfg` but is not added to the Settings menu.

Closes #10339.

## Limitations

- If the custom industry number is set below the number of required industries (those without the optional flag, where if they cannot be generated the player will see a warning message that an industry failed to generate), the game will still attempt to generate one of each of those industries. The resulting total will be higher than the player's choice. I see this as a feature, not a bug, but welcome other opinions. 🙂 
- The default setting for industry density is now Normal, where it had previously been `ID_END - 1` or High. I think this was a mistake so I've fixed it. (Default town density is Normal.)

## Checklist for review


src/table/settings/difficulty_settings.ini
```diff
@@ -92,6 +92,13 @@ strhelp  = STR_CONFIG_SETTING_INDUSTRY_DENSITY_HELPTEXT
 strval   = STR_FUNDING_ONLY
 cat      = SC_BASIC
 
+[SDT_VAR]
```
rubidium42: This is meant for what is stored in the PATS chunk according to the document, so this does not quite fit here. I would suggest world_settings.ini.
And I wonder whether it needs to be saved just like `custom_town_number` is, as there is a path from an industry GUI to the access of `customer_industry_number` (via `GenerateIndustries`, which is not called only from initial game generation)

2TallTyler: My reason for saving in the config file is that this allows a player to store the ideal industry number for their favorite map size, instead of having to remember it. And it's consistent with `custom_town_number`. I could be convinced otherwise. 🙂 

(I did move it to `world_settings.ini` though, good call on that.)

rubidium42: The whole reason I asked was because I noticed a savegame version in `world_settings.ini` for `customer_town_number`: https://github.com/OpenTTD/OpenTTD/blob/master/src/table/settings/world_settings.ini#L268
Having it saved in the savegame does not exclude it from being saved in the configuration file.
When trying the preview I noticed that the intro game just worked, so a minimum version is not needed anymore since the savegame chunk rewrite. Testing locally also showed that it is actually saved in the savegame already with this change and slightly older versions of OpenTTD are able to load those savegames. So a savegame bump is not needed (anymore).

rubidium42 (approved):

PR #10342, opened 2023-01-12
Feature: Press Ctrl to build a diagonal area of trees

2TallTyler:
## Motivation / Problem

When planting trees in-game, you can drag an area of trees. (Scenario Editor has brushes and works completely differently.)

This area can only be orthogonal, not diagonal...until now.

## Description

Hold Ctrl to select a diagonal area, as with terraforming, demolishing, buying land, or building rivers or objects.

## Checklist for review

### DIFF
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -384,9 +384,10 @@ void GenerateTrees()
  * @param tile end tile of area-drag
  * @param start_tile start tile of area-drag of tree plantation
  * @param tree_to_plant tree type, TREE_INVALID means random.
+ * @param diagonal Whether to use the Orthogonal (false) or Diagonal (true) iterator.
  * @return the cost of this operation or an error
  */
-CommandCost CmdPlantTree(DoCommandFlag flags, TileIndex tile, TileIndex start_tile, byte tree_to_plant)
+CommandCost CmdPlantTree(DoCommandFlag flags, TileIndex tile, TileIndex start_tile, byte tree_to_plant, bool diagonal)
 {
 	StringID msg = INVALID_STRING_ID;
 	CommandCost cost(EXPENSES_OTHER);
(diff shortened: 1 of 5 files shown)


src/tree_cmd.cpp
```diff
@@ -398,8 +399,16 @@ CommandCost CmdPlantTree(DoCommandFlag flags, TileIndex tile, TileIndex start_ti
 	Company *c = (_game_mode != GM_EDITOR) ? Company::GetIfValid(_current_company) : nullptr;
 	int limit = (c == nullptr ? INT32_MAX : GB(c->tree_limit, 16, 16));
 
-	TileArea ta(tile, start_tile);
-	for (TileIndex current_tile : ta) {
+	std::unique_ptr<TileIterator> iter;
+	if (diagonal) {
+		iter = std::make_unique<DiagonalTileIterator>(tile, start_tile);
+	}
+	else {
```
glx22:
```suggestion
	} else {
```

src/tree_cmd.cpp
```diff
@@ -398,8 +399,15 @@ CommandCost CmdPlantTree(DoCommandFlag flags, TileIndex tile, TileIndex start_ti
 	Company *c = (_game_mode != GM_EDITOR) ? Company::GetIfValid(_current_company) : nullptr;
 	int limit = (c == nullptr ? INT32_MAX : GB(c->tree_limit, 16, 16));
 
-	TileArea ta(tile, start_tile);
-	for (TileIndex current_tile : ta) {
+	std::unique_ptr<TileIterator> iter;
```
rubidium42: Seeing this (seemingly) copied bit of code made me wonder whether it's better to not copy the construction of the iterator, but change it into a function that gets reused. Also has the bonus of removing a few delete statements by move all those cases to smart pointers:
https://github.com/OpenTTD/OpenTTD/compare/master...rubidium42:OpenTTD:tileiterator_create 

I made #10345 to let the iterators use smart pointers and unify the creation of diagonal and orthogonal tile iterators.

rubidium42: #10345 has been merged, so you should use `std::unique_ptr<TileIterator> iter = TileIterator::Create(tile, start_tile, diagonal);` now instead of these six lines.

2TallTyler: Thanks for that. 🙂 Updated.

rubidium42 (approved):

PR #10346, opened 2023-01-13
Fix: Link variants to parents when finalising engines.

PeterN:
## Motivation / Problem

It is currently not possible to define a variant whose parent is defined later in the NewGRF.

In this example TE10 is the parent of several variants, but is defined later:

image

## Description

This is resolved by deferring mapping variant IDs to the engine finalisation stage. This ensures that definition-order of engines within the NewGRF does not matter.

image

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1329,7 +1329,7 @@ static ChangeInfoResult RailVehicleChangeInfo(uint engine, int numinfo, int prop
 				break;
 
 			case 0x2F: // Engine variant
-				ei->variant_id = GetNewEngineID(_cur.grffile, VEH_TRAIN, buf->ReadWord());
+				ei->variant_id = buf->ReadWord();
 				break;
 
 			case 0x30: // Extra miscellaneous flags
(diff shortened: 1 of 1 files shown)


src/newgrf.cpp
```diff
@@ -9001,13 +9001,16 @@ static void FinaliseEngineArray()
 			}
 		}
 
-		if (!HasBit(e->info.climates, _settings_game.game_creation.landscape)) continue;
-
-		/* Set appropriate flags on variant engine */
+		/* Do final mapping on variant engine ID and set appropriate flags on variant engine */
 		if (e->info.variant_id != INVALID_ENGINE) {
-			Engine::Get(e->info.variant_id)->display_flags |= EngineDisplayFlags::HasVariants | EngineDisplayFlags::IsFolded;
+			e->info.variant_id = GetNewEngineID(e->grf_prop.grffile, VEH_TRAIN, e->info.variant_id);
```
rubidium42: Shouldn't `VEH_TRAIN` be something like `e->type`?

PeterN: Oops :-)

PeterN:
Fix using the wrong engine type.

As a note, this change also moves the check to be above the condition that checks for climate. This is so that variant hierarchy is still populated ignoring that check.

rubidium42 (approved):

PR #8535, opened 2021-01-08
Fix #8506: Towns shouldn't add junctions to NewGRF roads they cannot build

2TallTyler:
Fixes #8506

## Motivation / Problem

Per #8506, towns add junctions to NewGRF roads which have the `TOWN_BUILD` flag disabled.

If a town cannot build a road type, I would assume that it should not be able to add junctions to roads of that type.

## Description

I have created a savegame to test. You will need the RattRoads NewGRF, available from Bananas.
[HighwayRoadTest.zip](https://github.com/OpenTTD/OpenTTD/files/5789609/HighwayRoadTest.zip)

While the town is not able to build junctions on the highway, the town growth algorithm can still walk down the highway to build houses past the highway.

Note that while the RattRoads Highway also has the `NO_HOUSES` flag set, this is not checked as I think it is irrelevant to towns building said road.

## Limitations

There could be side effects I'm not aware of. Please help test this!

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

Issue #8506, opened 2021-01-05
Town builds unnecessary junctions on highway


kiwitreekor:
## Version of OpenTTD
1.10.3

## Expected result
Towns should not build unnecessary junctions on highway (disallowed construction of houses along roadtype and disabled to be built by towns)

## Actual result
![image](https://user-images.githubusercontent.com/24273113/103709980-2d45d600-4ff7-11eb-9896-dc231aa0d2fa.png)
Towns build unnecessary junctions

## Steps to reproduce
1. Build highway connected with town
2. Wait until town grows

Eddi-z:
"Highway" is not a default road in OpenTTD, so you might want to update your "steps to reproduce" by which NewGRFs need to be loaded.

kiwitreekor:
I used Rattroad 1.0.2 to reproduce it, but it is reproducible with any newgrf which includes a roadtype with flag bit 2 set and bit 4 cleared.

glx22:
Does it happen for all town layouts ?

kiwitreekor:
![image](https://user-images.githubusercontent.com/24273113/103731487-9b56c100-5028-11eb-9eb1-623894188e93.png)
2x2 layout
![image](https://user-images.githubusercontent.com/24273113/103731807-5bdca480-5029-11eb-84d9-e8a951ad72b1.png)
3x3 layout
![image](https://user-images.githubusercontent.com/24273113/103732001-d7d6ec80-5029-11eb-9252-8daa875ee40b.png)
Default layout

Yes, it happens.

kiwitreekor:
I found that town can build roads after making junctions, I'm not sure whether this is intended or not.
In my opinion, It doesn't seem natural that a city can make a junction on a highway.

glx22:
@kiwitreekor can you check if #8535 solves the issue ? (A preview is available for this PR): https://preview.openttd.org/pr8535/

kiwitreekor:
#8535 fixes the issue, thanks!

JGRennison:
For reference I added a road/tram type flag in my branch to prevent towns from modifying tiles of a particular road/tram type. https://github.com/JGRennison/OpenTTD-patches/commit/946442ab6be8897b2efa6157866feded7a27770f
This is a bit more restrictive than the change proposed in #8535 in that it also covers some other town road operations such as removing free ends.
This sort of mechanism could be an alternative way of solving this.

[...]

PR #10347, opened 2023-01-13
Fix #10343: Towns attempt and fail to extend disallowed roadtypes

### LINKED ISSUE #10343: [Bug]: Towns attempt and fail to extend disallowed roadtypes
### Version of OpenTTD

nightly 20230112

### Expected result

The NewRoadType spec includes a property that disallows towns from building particular types of roads - so that a town does not inappropriately consist entirely of motorways or other non-urban roadtype. The town growth function does not appear to account for this, and will attempt to build off non-town roadtypes when expanding. This results in the town creating a lot of ugly T junctions on non-town roadtypes.

Instead, the town should not attempt to extend road from a non-town roadtype. It shouldn't *ignore* these roadtypes altogether, as if they were not roads, because that would mean it wouldn't find areas of town on the other side of the road...

### Actual result

New Godfield Transport, 1950-01-28
Setup, with non-town roadtype around the edge of the town.

New Godfield Transport, 1960-01-31
After 10 years, with spiky non-town road.

New Godfield Transport, 1950-02-15
Without roadtypes.

New Godfield Transport, 1960-01-27
After 10 years.

### Steps to reproduce

See above screenshots.

PeterN:
## Motivation / Problem

As per #10343, and also way back in #8506/#8535, towns are able to build road pieces for road types even if the the roadtype disallows it.

image

## Description

This is resolved by checking if the tile currently has a non-buildable roadtype, and disallow building a road. This is slightly different from the fix in #8535 as that also inadvertently prevented building houses along non-buildable roadtypes, though that may be permitted.

image

This also resolves and additional concern if a NewGRF disallows towns from building all roadtypes, by forcing the default ROADTYPE_ROAD slot to be town-buildable if no other roadtypes are town-buildable.

Closes #10343

## Limitations

Some NewGRFs maybe rely on the previously broken behaviour. Those NewGRFs can be fixed by the author.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -1362,6 +1362,20 @@ static inline bool RoadTypesAllowHouseHere(TileIndex t)
 	return !allow;
 }
 
+/** Test if town can grow road onto a specific tile.
+ * @param town Town that is building.
+ * @param tile Tile to build upon.
+ * @return true iff the tile's road type don't prevent extending the road.
+ */
+static bool TownCanGrowRoad(const Town *town, TileIndex tile)
+{
+	if (!IsTileType(tile, MP_ROAD)) return true;
+
+	/* Allow extending on roadtypes which can be built by town, or if the road type matches the type the town will build. */
+	RoadType rt = GetRoadTypeRoad(tile);
+	return HasBit(GetRoadTypeInfo(rt)->flags, ROTF_TOWN_BUILD) || GetTownRoadType(town) == rt;
+}
+
 /**
  * Grows the given town.
  * There are at the moment 3 possible way's for
@@ -1438,6 +1452,8 @@ static void GrowTownInTile(TileIndex *tile_ptr, RoadBits cur_rb, DiagDirection t
 		}
 
 	} else if (target_dir < DIAGDIR_END && !(cur_rb & DiagDirToRoadBits(ReverseDiagDir(target_dir)))) {
+		if (!TownCanGrowRoad(t1, tile)) return;
+
 		/* Continue building on a partial road.
 		 * Should be always OK, so we only generate
 		 * the fitting RoadBits */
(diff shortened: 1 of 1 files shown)


PeterN:
And yes, this does affect town growth...

src/town_cmd.cpp
```diff
@@ -1445,6 +1445,9 @@ static void GrowTownInTile(TileIndex *tile_ptr, RoadBits cur_rb, DiagDirection t
 
 		if (!_settings_game.economy.allow_town_roads && !_generating_world) return;
 
+		/* Don't allow junctions on roadtypes which can't be built by towns. */
```
rubidium42: Should both of these maybe be before the `_grow_town_result = GROWTH_SEARCH_STOPPED;`? After all, based on the comment it looks like it's definitely going to create some road, so it does not need to search further and then you still prevent it from building the road.
I've got no idea what the effect would be nor whether that would be a welcome effect.

PeterN: Not sure, there is a distinction between GROWTH_SUCCEED and GROWTH_SEARCH_STOPPED, but I don't know what.

PeterN:
Rewrite of the change to cope with the issue of the town being unable to extend even the roadtype it is going to use.

src/town_cmd.cpp
```diff
@@ -1362,6 +1362,20 @@ static inline bool RoadTypesAllowHouseHere(TileIndex t)
 	return !allow;
 }
 
+/** Test if town can grow road onto a specific tile.
+ * @param town Town that is building.
+ * @param tile Tile to build upon.
+ * @return true iif the tile's road type don't prevent extending the road.
```
rubidium42: Tiny typo: s/iif/iff/

rubidium42 (approved):

PR #8492, opened 2021-01-03
Fix: Don't allow towns to terraform certain floodable tiles

SamuXarick:
## Motivation / Problem
road and house flooded
During world generation, town growth algorithm is speed up, allowing the town to perform multiple sequential growth actions, like building houses, roads, and terraforming. If it happens to lower terrain adjacent to sea and then in one of those sequences it builds a road or a house there, it will end up flooded during the "run tile loop" phase of world generation.
The result could be similar to what the screenshot shows: disconnected roads and/or houses.

This happens very rarely. I tried 12k towns 4096x4096 map, and detected no more than 3 flooded road or house tiles.

## Description
Prevent rare occurrences of roads and houses being flooded, resulting in disconnected towns.

If the terraform command finds a certain floodable tile, the terraform fails. This works even for a sequence of commands during town generation. So far, the tests have not yet produced any flooded houses or roads.

## Limitations
Void tiles were a problem. Terraform command ignores them. If the town terraformed one of the tiles at the edges of the map, there was no checking happening for the void tiles, and it could result in a flood coming in the direction of the void tile. I added checks coming from the edge tiles. Terraform command shall fail if the resulting slope on an edge tile is flat and has a height of zero.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10348, opened 2023-01-13
Fix: Various Wide River issues

2TallTyler:
## Motivation / Problem

@SamuXarick identified several issues with Wide Rivers and suggested improvements.

### Cleanup
1. Fixed an accidental double semicolon.
2. One usage of `MarkTileDirtyByTile()` used the iterator for a `CircularTileSearch()` which worked since it was initialized to the current tile, but should just reference the current tile to be less...weird.

### Fixes
1. Looking for adjacent river tiles to determine the desired slope accidentally called `tile` instead of `other_tile` -- see the check directly above it for a sloped tile to see how it was intended. As it was, that `if` path was never taken. I've tested the fix briefly and it seems to not really affect anything, but it might be more efficient.
2. Adding upstream or downstream water tiles did not convert desert tiles or mark tiles dirty.

### Changes
1. After rivers are widened, some ocean tiles might not have flooded yet due to terraforming. Force a flood check to allow #8492 to work properly.

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -1093,7 +1093,7 @@ static bool RiverMakeWider(TileIndex tile, void *data)
 	/* If the tile is at height 0 after terraforming but the ocean hasn't flooded yet, don't build river. */
 	if (GetTileMaxZ(tile) == 0) return false;
 
-	TileIndex origin_tile = *(TileIndex *)data;;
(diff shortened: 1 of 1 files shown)


Ufiby:
My suggestions

michicc (approved):

PR #10344, opened 2023-01-13
Change: Try making Scripts usage of Random more predictable

SamuXarick:
## Motivation / Problem
The use of InteractiveRandom in Script based functions means that when attempting to reproduce the same game play they could not repeat the same events.

## Description
Make AIs or GSs do the same thing when the games are restarted in single player wherever possible.

## Limitations
Note that there are certain limitations though where reproduceability isn't garanteed, such as GS's ability to modify game settings, but those are not dealt with in here.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10349, opened 2023-01-14
Change: make scripts reproducible

rubidium42:
## Motivation / Problem

For debugging scripts it can be useful to have them behave in a reproducible manner.
Currently there are a few functions that are not reproducible due to the use of `InteractiveRandom` such as building/prospecting industries, creating towns, and setting the presidents gender.

In multiplayer this is even worse, because the game will not use `Random` but `InteractiveRandom` all the time! This is not addressed by #10344.

## Description

Refactor some of the command code that internally makes a choice between `Random` and `InteractiveRandom` to accept the `Randomizer` that is to be used. Also refactor the script code to have one location for its source of random.

This is followed by the introduction of a per-script `Randomizer` state that is saved in the savegame, making scripts reproducible between multiplayer runs. This obviously requires no other user commands to be issued for the script to be truely reproducible, but that was already the case for single player as well.
However, technically it should be possible to desync-debug-dump the commands and then rerun the game while only injecting the commands that did not originate from scripts and it should basically be running in sync.

- [x] Refactor command code 
- [x] Refactor script code
- [x] Figure out what to do with the `InteractiveRandomRange` in script_config.cpp and ai_scanner.cpp; #10330 will help get rid of one case in script_config.cpp.
- [x] Introduce per-script `Randomizer`

As little bonus, the regression should not be as susceptible to changes in code as the random it uses internally is now not linked to the main random. As such, there should be fewer changes in the part where it tests its own random functions.

## Limitations

Random debugging is slightly compromised, as the two places where now a `Randomizer` is passed and used in the command code will not trigger the desync-debug output if making a -DRANDOM_DEBUG build. However, one of those is in town name generation at the start of the game, and the other is for the face generation during the creation of a new company.
This could be resolved by using `std::source_location` instead of the macro mayhem that is currently used, however that requires C++20 and Clang 15 has a bug (fixed in trunk) making that unusable at the moment.

For simplicity the randomizer state is stored in an array in `ScriptObject` with one entry per `Owner` until `OWNER_END`. This means that a few entries will never be used, but the complexity of adding to an existing per-script object in code will outweigh this choice and memory waste. Well, at least with our current 15 companies.
However, due to this the `InteractiveRandomRange` calls in script_config.cpp and ai_scanner.cpp that are used when setting up the AIs and Game Script can be made reproducible by using the randomizer state for `OWNER_NONE`.

Saving the randomizer states of scripts in the savegame makes little sense as it is infeasible for a script to save/know its complete state. For example, let's fund a new industry. For saving I need to store that I'm going to fund the industry, and then I need to run the command to fund it. What if the script gets interrupted because of op codes just before the call to fund the industry. After reloading I can start that command again and it might work. However, if we went into the API, which itself does a random call!, and then the game gets saved before the command is actually executed; now after reloading the random inside the API call has already been done but calling the API again will mean an extra random has been done and you are not in-sync anymore.
Something similar is conceivable when saving happens between the command being executed and the script getting the result. The loaded script is not getting the result and has the state that the command still needs to be executed. So, the only way this is going to work is to abolish the custom save/load scripts, and just to save the entire stack and heap of the AI including all constructed C++ objects. That's definitely way too much effort for such a relatively small gain.

So, loading the same savegame multiple times should yield a reproducible result to eachother and starting new games with the same settings/seed should yield a reproducible result. Continueing to run a game after saving will not yield a reproducible result compared to loading that just saved savegame.

Final note: I have not actually fully tested the reproducibility yet! However, restarting a game seems to be doing the right thing.

## Checklist for review


glx22:
I did my test with 21b16ea.

First step was to generate a game in master (no AI config, no scripts, no newgrf) and save it.

Then I loaded it in master and typed `startai` in the console (with varying delay between commands)
image

Then I reloaded it in master and again typed `startai` in the console (with varying delay between commands)
image

Then I loaded it in 10349(21b16ea) and typed `startai` in the console (with varying delay between commands)
image

Then I reloaded it in 10349(21b16ea) and tagain yped `startai` in the console (with varying delay between commands)
image

As we can see, the AIs are different in the 2 master tests, but they are the same in 10349.

Company colour, faces and president are still different, but IIRC they use Random() during company creation command.

glx22 (approved):

PR #10350, opened 2023-01-14
Codechange: refactor removal of desert around river tiles

SamuXarick:
## Motivation / Problem
Code repetition to do the same thing in multiple places

## Description
Factor removal of desert directly around river tiles

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -1064,11 +1064,7 @@ static bool MakeLake(TileIndex tile, void *user_data)
 	for (DiagDirection d = DIAGDIR_BEGIN; d < DIAGDIR_END; d++) {
 		TileIndex t2 = tile + TileOffsByDiagDir(d);
 		if (IsWaterTile(t2)) {
-			MakeRiver(tile, Random());
-			MarkTileDirtyByTile(tile);
-			/* Remove desert directly around the river tile. */
-			TileIndex t = tile;
-			CircularTileSearch(&t, RIVER_OFFSET_DESERT_DISTANCE, RiverModifyDesertZone, nullptr);
+			MakeRiverAndModifyDesertZoneAround(tile);
 			return false;
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/water_cmd.cpp
```diff
@@ -427,6 +427,15 @@ bool RiverModifyDesertZone(TileIndex tile, void *)
 	return false;
 }
```
2TallTyler, line 429: Needs Doxygen

src/water_cmd.cpp
```diff
@@ -427,6 +427,15 @@ bool RiverModifyDesertZone(TileIndex tile, void *)
...
 
+void DoRiverModifyDesertZone(TileIndex tile, uint8 random_bits) {
+	MakeRiver(tile, random_bits);
```
2TallTyler: Instead of passing Random() as a parameter into random_bits, why not just call Random() here? (Honest question, I haven't done much with randomness and may be missing something.)

2TallTyler (commented):
This looks helpful, but why create a separate function instead of moving the desert zone modifications into the existing `MakeRiver()`?

src/water.h
```diff
@@ -39,6 +39,7 @@ void MakeWaterKeepingClass(TileIndex tile, Owner o);
 void CheckForDockingTile(TileIndex t);
 
 bool RiverModifyDesertZone(TileIndex tile, void *data);
+void DoRiverModifyDesertZone(TileIndex tile, uint8 random_bits);
```
rubidium42: The function does not only do that, but much more. What about: `MakeRiverAndMakeNearbyDesertTropical`? It's a longer name, but you immediately know what it's doing. Okay, this name is not perfect as it's not really "tropical" but the normal in the tropical map.
I'd maybe even rename `RiverModifyDesertZone` to `MakeDesertTropical`.

src/water_cmd.cpp
glx22: I agree, remove the `random_bits` param and just use `Random()` in MakeRiver call.

src/landscape.cpp
```diff
@@ -1453,11 +1431,7 @@ static std::tuple<bool, bool> FlowRiver(TileIndex spring, TileIndex begin, uint
 				/* We only want a lake if the river is long enough. */
 				DistanceManhattan(spring, lakeCenter) > min_river_length) {
 			end = lakeCenter;
```
glx22, line 1433: Unrelated to the PR, but that's a "nice" abuse of a local var to backup another local var.

SamuXarick: ah, so it can be removed?

glx22: no it's still needed, it's used to restore `lakecenter` after `CircularTileSearch()` calls

SamuXarick: nevermind, it can't, it's used below

2TallTyler (approved):

PR #10351, opened 2023-01-14
Codechange: Minor cleanups to Windows font handling

PeterN:
## Motivation / Problem

* Fallback font list uses custom C-style list allocation.
* Glyph rendering attempts to guess memory size required for a glyph to prevent two calls to `GetGlyphOutline`.
* Trying to render a glyph that is too large can fail at the `AllocaM` stage before width & height are checked.

## Description

* Simplify by using `std::vector` and `std::wstring` for fallback font list instead.
* The `GetGlyphOutline` API basically is intended to be called twice. Removing the guesswork simplifies the flow.
* Due to the above, we know the glyph width & height before trying to allocate, so do this check first. Gives a better error.

## Checklist for review

### DIFF
--- a/src/os/windows/font_win32.cpp
+++ b/src/os/windows/font_win32.cpp
@@ -436,22 +436,13 @@ void Win32FontCache::ClearFontCache()
 	GLYPHMETRICS gm;
 	MAT2 mat = { {0, 1}, {0, 0}, {0, 0}, {0, 1} };
 
-	/* Make a guess for the needed memory size. */
-	DWORD size = this->glyph_size.cy * Align(aa ? this->glyph_size.cx : std::max(this->glyph_size.cx / 8l, 1l), 4); // Bitmap data is DWORD-aligned rows.
-	byte *bmp = AllocaM(byte, size);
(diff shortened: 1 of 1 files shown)


src/os/windows/font_win32.cpp
```diff
@@ -233,54 +233,29 @@ static std::string GetEnglishFontName(const ENUMLOGFONTEX *logfont)
...
-
-		if (this->items == this->capacity) {
-			this->capacity += 10;
-			this->fonts = ReallocT(this->fonts, this->capacity);
+		for (const auto &entry : this->fonts) {
+			if (wcscmp(entry.get(), font) == 0) return false;
 		}
 
-		this->fonts[this->items++] = wcsdup(font);
+		this->fonts.emplace_back(wcsdup(font));
```
rubidium42: Essentially you're mixing `malloc` with `delete` here due to `unique_ptr`. What about `std::vector<std::wstring>` and getting rid of `wcsdup`?
Maybe even `std::wstring_view &font` and the C++ compare between `entry.get()` and `font`?

PeterN: Oops, stupid. Yes, `wstring`/`wstring_view` work here.

rubidium42 (approved):

PR #10353, opened 2023-01-14
Refactor some CodeQL warnings away

rubidium42:
## Motivation / Problem

CodeQL warnings about not following "Rule of Two" and "Comparison result is always the same".

## Description

The "Rule of Two" is about that if you declare a copy-constructor you should also declare a ssignment and vice versa, so you do not get conflicts between the default and custom behaviour. In case of sqclass.h the assignment is explicitly deleted. In case of fmt/core.h I did a check of more recent versions and the copy-constructor got removed. Since it's declared to be the default implementation, this should not be a problem for us either.

The "Comparison result is always the same" are more funky. In console_cmds.cpp's zoomto function the GetArgumentInteger always gets an unsigned number and since `ZOOM_LVL_MIN` is 0, checking for the parameter to be less than makes no sense. I wanted to opt to remove it, but then users can try negative numbers and they will get an error that the value is too large. So I just check for the minus-character instead and added a `static_assert` in case `ZOOM_LVL_MIN` ever becomes a positive number.
Also in console_cmds.cpp there is a check in the font function for the argument index being less than the number of arguments, but it already checked for argument index + 1 before. On the other hand the code looked more consistent before.
In network/core/game_info.cpp there is a check for the number of NewGRFs to not exceed `NETWORK_MAX_GRF_COUNT`, however the value being read it an uint8 and the value it is checked against is > 255. That can obviously never happen, and as such the check is pointless. However, to prevent handy people only changing `NETWORK_MAX_GRF_COUNT` to get more NewGRFs a `static_assert` is added to at least be triggered to check the network protocol.

And a small typo as bonus.

## Limitations

None as far as I know.

## Checklist for review

### DIFF
--- a/src/3rdparty/fmt/core.h
+++ b/src/3rdparty/fmt/core.h
@@ -529,7 +529,6 @@ void check_format_string(S);
 
 struct error_handler {
   constexpr error_handler() = default;
-  constexpr error_handler(const error_handler&) = default;
 
   // This function is intentionally not constexpr to give a compile-time error.
   FMT_NORETURN FMT_API void on_error(const char* message);
--- a/src/3rdparty/squirrel/squirrel/sqclass.h
+++ b/src/3rdparty/squirrel/squirrel/sqclass.h
@@ -10,6 +10,7 @@ struct SQClassMember {
 		val = o.val;
 		attrs = o.attrs;
 	}
+	SQClassMember& operator=(SQClassMember &o) = delete;
 	SQObjectPtr val;
 	SQObjectPtr attrs;
 };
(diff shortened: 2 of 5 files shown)


src/3rdparty/fmt/core.h
```diff
@@ -529,7 +529,6 @@ void check_format_string(S);
 
 struct error_handler {
   constexpr error_handler() = default;
-  constexpr error_handler(const error_handler&) = default;
```
glx22, line 532: Probably better to fully update the library files.
See https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h#L649-L650

rubidium42: Oh, yeah. Forgot to write down that I had the exact same idea, and actually went ahead with it for a while but for some reasons I don't remember anymore it didn't really want to co-operate, so I stopped that attempt.
Then I'd rather go to `std::format`, but that requires C++20 and the compilers aren't quite up to speed yet.

glx22: Ok, I updated the files, MSVC builds and regression pass, but gcc in wsl fails to build.

glx22 (approved):

PR #10355, opened 2023-01-15
Change #10077: make maximum loan a positive multiple of the loan interval

### LINKED ISSUE #10077: [Bug]: Incorrect loan limit amount - other than what the player selected
### Version of OpenTTD

since 1.11.0-beta1, related to #8386, the bug is also present in the JGRPP version

### Expected result

Full title:
The loan limit is always a multiple of £ 50,000, despite the possibility to enter a different value. Setting a loan limit below £ 50,000 gives you a limit of 0 which makes it impossible to start the game.

The report de facto concerns two related errors:
1. The loan limit value entered by the player is not used in the game - I don't know since when the problem exists, the loan setting in the window is available from 0.3.4
2. The entered value of the loan limit below £ 50,000 in the game is reduced to £ 0 - the problem appeared with #8386
Solving one of the problems will most likely affect the other, so I attach both of them in one report.

**Three main possibilities:**
1. Enabling the player to actually choose any loan limit, which in the game will be as much as the player entered in the setting. If a player chooses £ 10,000, there will be £ 10,000 in play. If £ 5,492, there will be £ 5,492 in game.
2. The visible loan limit, after entering any amount in the settings window, will be automatically rounded up to a multiple of £ 50,000 or another amount, eg £ 10,000; The description should contain appropriate information regarding the rounding of the amount.
3. Removal of the possibility to enter any loan amount - the selected amount will always be a multiple, although selecting larger amounts will be cumbersome. Then adding the Ctrl+click or (and) Alt+click function, which will give a click x10 (x100) effect (also useful in other cases)

### Actual result

The loan limit is always a multiple of £ 50,000, which the player cannot see in the settings window. The value entered by the player, although visible in the settings window, is not valid in the game. In fact, a multiple of 50,000 will be used, always less than or equal to the one entered. As of 1.11.0 a problem with this is that setting a loan limit below £ 50,000 gives you a limit of 0 which prevents you from starting the game without cheating.

In my opinion, setting a low or very low limit can be an interesting challenge for players ("from rags to riches"), which is currently impossible.

### Steps to reproduce

1. If you set a loan limit of £ 49,999 or less in the settings window, it will be £ 0 in game.
2. If you set a loan limit of £ 50,000 or a multiple of this in the settings window, the game will use that limit exactly.
3. If you set a loan limit above £ 50,000 in the settings window, but other than a multiple of that amount, it will be rounded down to that multiple in the game. So, for example, instead of £ 90,000 it will be £ 50,000.

rubidium42:
## Motivation / Problem

Fixes #10077 as far as I am aware. Lots of small issues related to the maximum loan setting.

## Description

First and foremost problem is that you can set the maximum loan below the first multiple of the interval for maximum loan. In the original code this would be 50 000, so setting it to 20 000 means you would have to wait quite a while in 1920 before you can actually take out a loan. 
Furthermore setting it to 0 means you can never get a loan. However, a loan of 0 can be useful if there is a Game Script that will provide you with funds. So, allow 0 but round everything 1 GBP up to loan interval GBP and show a warning in the setting that a Game Script is required.
screenshot

The second issue is that you can enter any random amount, but the maximum loan shown in the game will still be a multiple of the interval for maximum loan. I added a pre check callback, so when the user enters it in the user interface the setting will get (silently) rounding towards the nearest multiple of the interval for maximum loan.

A third issue, that was not noted in #10077, is that the initial money/loan at company start is applying inflation twice to the maximum loan; it got minimum of the default initial loan and inflation corrected maximum loan, and then inflation corrected that minimum. When the inflation corrected maximum loan was less than the initial loan, then the initial money/loan of the company would be more than allowed by the maximum loan shown in the company finances window.

Finally I changed the maximum loan interval to the interval of taking out a loan. This means more granularity for the person setting the maximum loan, and allows slightly more loan when inflation sets in due to rising earlier to say 310 000 instead of waiting until 350 000 before the first increase in maximum loanx.

## Limitations

Loading the configuration from savegame/settings.ini will not clamp the maximum loan setting to a multiple of the interval for maximum loan.

## Checklist for review

### DIFF
--- a/src/table/settings/difficulty_settings.ini
+++ b/src/table/settings/difficulty_settings.ini
@@ -98,9 +98,10 @@ type     = SLE_UINT32
 from     = SLV_97
 flags    = SF_NEWGAME_ONLY | SF_SCENEDIT_TOO | SF_GUI_CURRENCY
 def      = 300000
-min      = 0
+min      = LOAN_INTERVAL
 max      = 2000000000
-interval = 50000
+pre_cb   = [](auto &new_value) { new_value = (new_value + LOAN_INTERVAL / 2) / LOAN_INTERVAL * LOAN_INTERVAL; return true; }
+interval = LOAN_INTERVAL
 str      = STR_CONFIG_SETTING_MAXIMUM_INITIAL_LOAN
 strhelp  = STR_CONFIG_SETTING_MAXIMUM_INITIAL_LOAN_HELPTEXT
 strval   = STR_JUST_CURRENCY_LONG
--- a/src/table/settings/difficulty_settings.ini
+++ b/src/table/settings/difficulty_settings.ini
@@ -96,7 +96,7 @@ cat      = SC_BASIC
 var      = difficulty.max_loan
 type     = SLE_UINT32
 from     = SLV_97
-flags    = SF_NEWGAME_ONLY | SF_SCENEDIT_TOO | SF_GUI_CURRENCY
+flags    = SF_NEWGAME_ONLY | SF_SCENEDIT_TOO | SF_GUI_CURRENCY | SF_GUI_0_IS_SPECIAL
 def      = 300000
 min      = LOAN_INTERVAL
 max      = 2000000000
@@ -104,7 +104,7 @@ pre_cb   = [](auto &new_value) { new_value = (new_value + LOAN_INTERVAL / 2) / L
 interval = LOAN_INTERVAL
 str      = STR_CONFIG_SETTING_MAXIMUM_INITIAL_LOAN
 strhelp  = STR_CONFIG_SETTING_MAXIMUM_INITIAL_LOAN_HELPTEXT
-strval   = STR_JUST_CURRENCY_LONG
+strval   = STR_CONFIG_SETTING_MAXIMUM_INITIAL_LOAN_VALUE
 cat      = SC_BASIC
 
 [SDT_VAR]
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -560,7 +560,8 @@ Company *DoStartupNewCompany(bool is_ai, CompanyID company = INVALID_COMPANY)
 	ResetCompanyLivery(c);
 	_company_colours[c->index] = (Colours)c->colour;
 
-	c->money = c->current_loan = (std::min<int64>(INITIAL_LOAN, _economy.max_loan) * _economy.inflation_prices >> 16) / 50000 * 50000;
+	/* Scale the initial loan based on the inflation rounded down to the loan interval. The maximum loan has already been inflation adjusted. */
+	c->money = c->current_loan = std::min<int64>((INITIAL_LOAN * _economy.inflation_prices >> 16) / LOAN_INTERVAL * LOAN_INTERVAL, _economy.max_loan);
 
 	std::fill(c->share_owners.begin(), c->share_owners.end(), INVALID_OWNER);
 
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -757,8 +757,8 @@ bool AddInflation(bool check_year)
  */
 void RecomputePrices()
 {
-	/* Setup maximum loan */
-	_economy.max_loan = ((uint64)_settings_game.difficulty.max_loan * _economy.inflation_prices >> 16) / 50000 * 50000;
+	/* Setup maximum loan as a rounded down multiple of LOAN_INTERVAL. */
+	_economy.max_loan = ((uint64)_settings_game.difficulty.max_loan * _economy.inflation_prices >> 16) / LOAN_INTERVAL * LOAN_INTERVAL;
 
 	/* Setup price bases */
 	for (Price i = PR_BEGIN; i < PR_END; i++) {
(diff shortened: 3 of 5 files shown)


src/table/settings/difficulty_settings.ini
```diff
@@ -98,9 +98,10 @@ type     = SLE_UINT32
 from     = SLV_97
 flags    = SF_NEWGAME_ONLY | SF_SCENEDIT_TOO | SF_GUI_CURRENCY
 def      = 300000
-min      = 0
+min      = LOAN_INTERVAL
 max      = 2000000000
-interval = 50000
+pre_cb   = [](auto new_value) { new_value = (new_value + LOAN_INTERVAL / 2) / LOAN_INTERVAL * LOAN_INTERVAL; return true; }
```
LordAro: `auto &` to make it more explicit? Kinda looks like it's modifying a local variable at the moment otherwise

LordAro (commented):
I'd be in favour of backporting this. It's definitely an issue that should be fixed

src/table/settings/difficulty_settings.ini
rubidium42: Sure, but then you'd get a few other instances as bonus ;)

ldpl:
Loan 0 is a useful case for game modes that don't need loan. GameScripts can give player as much money as they see fit for the intended gameplay.

Also, what's the point of limiting it to only be multiples of LOAN_INTERVAL at all? Having it loan/repay different amounts on some button clicks doesn't even count as an ui issue imo. If anything it would be better for buttons to do smth like 10% of the loan at one click. And I can see extremely low loans be somewhat useful for some special cases like starting with only one horse carriage or something. Though not nearly as useful as 0 loan ofc.

LC-Zorg:
> Also, what's the point of limiting it to only be multiples of LOAN_INTERVAL at all?

I think rounded loan amounts just look better. Another problem is that when you use a different currency, if you omit the rounding, the results will often not look the way the player wants. For example, if you enter 700,000zł (£1 = 6zł), the limit in the game will be 699,996zł, if you enter 650,000zł, it will be 649,996zł and so on. I haven't taken any loans for a long time, maybe I'm wrong, but it seems to me that in the real world, if the bank sets an upper limit, it's round amounts, e.g. $100,000 or $550,000 instead of $98,748 or $543,992. Loan amounts during inflation are also rounded.

Rounding to the nearest amount is probably the best way. For me, it works very well, although the possibility of starting the game with an even lower loan is a bit missing.

So, also taking into account ldpl comment, what do you think about adding the ability to choose the amount of 5000, 1000 and 0? So the interval would be: 0 <> 1,000 <> 5,000 <> 10,000 <> 20,000 <> 30,000 <> ...

rubidium42:
> Also, what's the point of limiting it to only be multiples of LOAN_INTERVAL at all?

The taking and repaying loans can only be done with a multiple of LOAN_INTERVAL. Changing that is a lot more work, as it affects the UI but also the scripts. For the scripts it's relatively easy, but for the UI you need some way to account for maybe wanted smaller or bigger steps. That's not a hurdle I want to take.
If someone wants to take that hurdle and make the interval 1000 or 500 or 100 or 1.Fine, then that also immediately updates the behaviour of what maximum loans can be set.

I can imagine you want no loans, so what about:
screenshot

> I think rounded loan amounts just look better. Another problem is that when you use a different currency, if you omit the rounding, the results will often not look the way the player wants. For example, if you enter 700,000zł (£1 = 6zł), the limit in the game will be 699,996zł, if you enter 650,000zł, it will be 649,996zł and so on.

Money in code is always in pounds, so the rounding to 10.000 also happens on that level. Entering 700,000zl will give you 720.000zl (or 120.000 pounds).
With Belgian Francs it will be even "weirder": 810.000 Francs. But that's always bound to happen when you use weird exchange rates.

Only caveat of the rounding, currency conversions and the 0 is special, is that entering 1-5zl will result in "No loan" and 6zl yields 60.000zl. But alas, there is nothing I can do about that without a huge change in the UI.

ldpl:
> > Also, what's the point of limiting it to only be multiples of LOAN_INTERVAL at all?
> 
> The taking and repaying loans can only be done with a multiple of LOAN_INTERVAL. Changing that is a lot more work, as it affects the UI but also the scripts. For the scripts it's relatively easy, but for the UI you need some way to account for maybe wanted smaller or bigger steps. That's not a hurdle I want to take. If someone wants to take that hurdle and make the interval 1000 or 500 or 100 or 1.Fine, then that also immediately updates the behaviour of what maximum loans can be set.

I guess that can be considered out of scope of this PR since it didn't work anyway.

> I can imagine you want no loans, so what about: 

Yeah, that seems fine. It's also a bit of a workaround for extremely low loans since GS can just give a tiny bit of money. Won't exactly be a loan but loans are kinda pointless anyway.

> Only caveat of the rounding, currency conversions and the 0 is special, is that entering 1-5zl will result in "No loan" and 6zl yields 60.000zl. But alas, there is nothing I can do about that without a huge change in the UI.

You can round up ;)

rubidium42:
>> But alas, there is nothing I can do about that without a huge change in the UI.
>
> You can round up ;)

If only ;) 5zl first becomes 0 GBP and rounding 0 up to the nearest multiple of 10 000 is still 0. Let me take a guess that you would then say, you can round up when converting currency, but then 89 995zl => 14 999.1|6| => 15 000 GBP and that gets rounded to 20 000 GBP which is also wrong.

To solve this, the rounding needs to be done before the conversion to GBP. Which means adding a per-setting configurable value for the rounding value with all the related logic in setting UI and amending a lot of tables, or an earlier pre-validation callback but then the callback needs to know whether it's called from the UI or the (remote) console command, and it needs to support 64 bit values. Which is, in the context of this minor nitpick, a huge change in the UI.

LordAro (approved):

PR #10365, opened 2023-01-16
Fix #10362: NewGRF bridges without speed limits.

### LINKED ISSUE #10362: [Bug]: Bridges cannot not have a speed limit
### Version of OpenTTD

nightly 20230115

### Expected result

All infrastructure in the game with speed limits (road and rail types) can have that speed limit disabled by setting it to 0, except bridges.

### Actual result

Setting bridge speed limit to 0 in NewGRF results in a speed limit of 0 (which means vehicles cross at stall/crawl speed). Setting it to 0xFFFF results in a speed limit of 65,918 km/h.

Setting a bridge speed limit to either 0 or 0xFFFF in NewGRF should disable the speed limit for that bridge (even if that just means setting the speed limit to 0xFFFF and hiding the speed info text in the "select bridge" menu).

### Steps to reproduce

Build a bridge and get over it.

FLHerne:
## Motivation / Problem
See #10362.

## Description
For bridges, a max speed of 0xFFFF (i.e. no effective limit) is no longer displayed as a limit in the UI.

A max speed of 0 is also considered unlimited, for similarity to the roadtype and railtype interface.

## Limitations
This changes behaviour for existing NewGRF-defined bridges with 0 speed, but that makes no sense.

I considered special-casing 0 max speed in the code (rather than just the GRF interface) as with rail/roadtypes, but it would have required intrusive changes in many more locations for no benefit I can think of.

## Checklist for review

### DIFF
--- a/src/bridge_gui.cpp
+++ b/src/bridge_gui.cpp
@@ -130,6 +130,23 @@ private:
 		this->SetWidgetDirty(WID_BBS_BRIDGE_LIST);
 	}
 
+	/**
+	 * Get the StringID to draw in the selection list and set the appropriate DParams.
+	 * @param bridge_data the bridge to get the StringID of.
+	 * @return the StringID.
+	 */
+	StringID GetBridgeSelectString(const BuildBridgeData &bridge_data) const
+	{
+		SetDParam(0, bridge_data.spec->material);
+		SetDParam(1, bridge_data.spec->speed);
+		SetDParam(2, bridge_data.cost);
+		/* If the bridge has no meaningful speed limit, don't display it. */
+		if (bridge_data.spec->speed == UINT16_MAX) {
+			return _game_mode == GM_EDITOR ? STR_SELECT_BRIDGE_INFO_NAME : STR_SELECT_BRIDGE_INFO_NAME_COST;
+		}
+		return _game_mode == GM_EDITOR ? STR_SELECT_BRIDGE_INFO_NAME_MAX_SPEED : STR_SELECT_BRIDGE_INFO_NAME_MAX_SPEED_COST;
+	}
+
 public:
 	BuildBridgeWindow(WindowDesc *desc, TileIndex start, TileIndex end, TransportType transport_type, byte road_rail_type, GUIBridgeList *bl) : Window(desc),
 		start_tile(start),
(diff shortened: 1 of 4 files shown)


src/lang/english.txt
```diff
@@ -2698,8 +2698,10 @@ STR_BUILD_SIGNAL_DRAG_SIGNALS_DENSITY_INCREASE_TOOLTIP          :{BLACK}Increase
 STR_SELECT_RAIL_BRIDGE_CAPTION                                  :{WHITE}Select Rail Bridge
 STR_SELECT_ROAD_BRIDGE_CAPTION                                  :{WHITE}Select Road Bridge
 STR_SELECT_BRIDGE_SELECTION_TOOLTIP                             :{BLACK}Bridge selection - click on your preferred bridge to build it
-STR_SELECT_BRIDGE_INFO                                          :{GOLD}{STRING},{} {VELOCITY} {WHITE}{CURRENCY_LONG}
-STR_SELECT_BRIDGE_SCENEDIT_INFO                                 :{GOLD}{STRING},{} {VELOCITY}
+STR_SELECT_BRIDGE_INFO                                          :{GOLD}{STRING}
+STR_SELECT_BRIDGE_INFO_MAX_SPEED                                :{GOLD}{STRING},{} {VELOCITY}
+STR_SELECT_BRIDGE_INFO_COST                                     :{GOLD}{STRING},{} {WHITE}{2:CURRENCY_LONG}
```
```suggestion
STR_SELECT_BRIDGE_INFO_COST                                     :{GOLD}{0:STRING},{} {WHITE}{2:CURRENCY_LONG}
```
(applying #10111)

src/bridge_gui.cpp
```diff
@@ -130,6 +130,22 @@ class BuildBridgeWindow : public Window {
...
 
+	StringID GetBridgeSelectString(const BuildBridgeData &bridge_data) const {
+		/**
+		* Get the StringID to draw in the selection list and set the appropriate DParams.
+		* @param bridge_data the bridge to get the StringID of.
+		* @return the StringID.
+		*/
+		SetDParam(2, bridge_data.cost);
+		SetDParam(1, bridge_data.spec->speed);
+		SetDParam(0, bridge_data.spec->material);
```
2TallTyler: Super nitpicky, but do these need to be out of order? (I see that they were originally, but since you're redoing it anyway... 🙂)

src/bridge_gui.cpp
```diff
@@ -130,6 +130,22 @@ class BuildBridgeWindow : public Window {
 		this->SetWidgetDirty(WID_BBS_BRIDGE_LIST);
 	}
```
2TallTyler, line 132: Doxygen should be here, not L134-138.

src/bridge_gui.cpp
```diff
@@ -130,6 +130,22 @@ class BuildBridgeWindow : public Window {
...
 
+	/**
+	* Get the StringID to draw in the selection list and set the appropriate DParams.
+	* @param bridge_data the bridge to get the StringID of.
+	* @return the StringID.
+	*/
+	StringID GetBridgeSelectString(const BuildBridgeData &bridge_data) const {
```
rubidium42:
```suggestion
	/**
	 * Get the StringID to draw in the selection list and set the appropriate DParams.
	 * @param bridge_data the bridge to get the StringID of.
	 * @return the StringID.
	 */
	StringID GetBridgeSelectString(const BuildBridgeData &bridge_data) const {
```
A bit pedantic, but the first asterisk of each of the lines should be under eachother.

rubidium42 (dismissed):

2TallTyler (dismissed):

src/lang/english.txt
```diff
@@ -2698,8 +2698,10 @@ STR_BUILD_SIGNAL_DRAG_SIGNALS_DENSITY_INCREASE_TOOLTIP          :{BLACK}Increase
 STR_SELECT_RAIL_BRIDGE_CAPTION                                  :{WHITE}Select Rail Bridge
 STR_SELECT_ROAD_BRIDGE_CAPTION                                  :{WHITE}Select Road Bridge
 STR_SELECT_BRIDGE_SELECTION_TOOLTIP                             :{BLACK}Bridge selection - click on your preferred bridge to build it
-STR_SELECT_BRIDGE_INFO                                          :{GOLD}{STRING},{} {VELOCITY} {WHITE}{CURRENCY_LONG}
-STR_SELECT_BRIDGE_SCENEDIT_INFO                                 :{GOLD}{STRING},{} {VELOCITY}
+STR_SELECT_BRIDGE_INFO                                          :{GOLD}{STRING}
```
PeterN: This string is incompatible (and used in different circumstances) so should have a different name IMHO.

2TallTyler (approved):

PR #10375, opened 2023-01-18
Change: Log AI/GS Squirrel crashes in white text for readability

andythenorth:
## Motivation / Problem

* improve legibility of errors in AI / GS debug log

## Description

* white on grey is more legible than red on grey

## Limitations

* probably someone will disagree

## Checklist for review

No
* No
* No
 Not really
No

### DIFF
--- a/src/ai/ai_gui.cpp
+++ b/src/ai/ai_gui.cpp
@@ -1166,7 +1166,7 @@ struct AIDebugWindow : public Window {
 					TextColour colour;
 					switch (log->type[pos]) {
 						case ScriptLog::LOG_SQ_INFO:  colour = TC_BLACK;  break;
-						case ScriptLog::LOG_SQ_ERROR: colour = TC_RED;    break;
+						case ScriptLog::LOG_SQ_ERROR: colour = TC_WHITE;    break;
 						case ScriptLog::LOG_INFO:     colour = TC_BLACK;  break;
 						case ScriptLog::LOG_WARNING:  colour = TC_YELLOW; break;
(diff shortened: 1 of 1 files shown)


SamuXarick:
I disagree 👎 . There's also yellow text for Warnings.

LC-Zorg:
I think you're both right. The red error color is used not only here. I think it should stay red as something important. But yeah, a red wall of text isn't fun to read and pick up important points. So what do you think about the orange/gold color of the bug description with the red color of the main information preserved?

Something like this...
Debug recolor 2 1

versus this...
Debug recolor 0

LC-Zorg:
In principle, the description could also be white. However, it seems to me that orange fits better in this place and is also quite readable.
Debug recolor 2 2

andythenorth:
White is the most legible.  

I picked white because it's the most legible.

Everything is irrelevant. Unix shell doesn't go to red for errors unless you install some fancy text formatter for it.

2TallTyler:
With the disclaimer that I've only dabbled with AI/GS development and haven't really used the log much...

I support making this more legible, although I think yellow warning text and white error text isn't quite right. Maybe they can both be white? I also like the orange error text as it's suitably severe looking while being significantly more readable than red.

andythenorth:
Amended PR, GSLog.Error now remains red, I don't use it so I don't care.  If the semantics of the colour are more important than legibility so be it.

Squirrel crashes are white. This is the change I wish to see made.  At this point the script is dead, the meaning of the colour is irrelevant.

image

2TallTyler (approved):
Suggested squashed commit title: `Change: Log AI/GS Squirrel crashes in white text for readability`

PR #10376, opened 2023-01-18
Fix: [Script] Ensure building/removing rail lines can't be backwards and allow building/removing rail lines coming from or heading to the void

SamuXarick:
## Motivation / Problem
It is possible to build and remove a rail line backwards, using `ScriptRail.BuildRail` and `ScriptRail.RemoveRail`.
It is not possible to build or remove a rail line that is facing the void.

## Description
- Add the missing precondition that forbids `from` tile to be the same as `to` tile.
- For lines longer than 1 piece which makes the value of `diag_offset` to be 2 or more, the test was incomplete. It was not checking the direction going from a tile to another. The code below would solve it, but it would be difficult to describe in the documentation, so I opted for `DiagdirBetweenTiles`.
```
EnforcePrecondition(false, diag_offset <= 1 ||
			(::TileX(from) == ::TileX(tile) && ::TileX(tile) == ::TileX(to) && (int)::TileY(from) - (int)::TileY(tile) == ((int)::TileY(tile) - (int)::TileY(to)) / abs((int)::TileY(tile) - (int)::TileY(to))) ||
			(::TileY(from) == ::TileY(tile) && ::TileY(tile) == ::TileY(to) && (int)::TileX(from) - (int)::TileX(tile) == ((int)::TileX(tile) - (int)::TileX(to)) / abs((int)::TileX(tile) - (int)::TileX(to))));
```
- Replace the checks that test `IsValidTile` for `from` and `to` with a `MapSize` check. This allows building or removing rail facing the void.

## Limitations
If the void tiles on top left and top right are disabled, then the edge tiles are always ocean that can't be cleared. So there should be no worries about placing rail tiles there.

## Checklist for review

### DIFF
--- a/bin/ai/compat_0.7.nut
+++ b/bin/ai/compat_0.7.nut
@@ -392,3 +392,88 @@ AIRoad.HasRoadType <- function(tile, road_type)
 	}
 	return false;
 }
+
+/* 14 ensures rail can't be built or removed backwards and allows
+ * 'from' and 'to' tiles to be invalid, as long as they're void tiles. */
+__BuildRemoveRailHelper <- function(from, tile, to)
+{
+	local map_log_x = function()
+	{
+		local x = AIMap.GetMapSizeX();
+		local pos = 0;
+
+		if ((x & 0xFFFFFFFF) == 0) { x = x >> 32; pos += 32; }
+		if ((x & 0x0000FFFF) == 0) { x = x >> 16; pos += 16; }
+		if ((x & 0x000000FF) == 0) { x = x >> 8;  pos += 8;  }
+		if ((x & 0x0000000F) == 0) { x = x >> 4;  pos += 4;  }
+		if ((x & 0x00000003) == 0) { x = x >> 2;  pos += 2;  }
+		if ((x & 0x00000001) == 0) { pos += 1; }
+
+		return pos;
+	};
+
+	/* Workaround to circumvent IsValidTile. We want to give a pass to void tiles */
+	local is_tile_inside_map = function(t)
+	{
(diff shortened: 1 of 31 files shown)


LordAro:
Missing justification: What happens currently if from == to ?

glx22:
I think it's to match the documentation stating `@pre from != to.`, but maybe the documentation is wrong (didn't check anything).

SamuXarick:
> Missing justification: What happens currently if from == to ?

building rail

glx22:
So if `from == to` it somehow looks like `BuildRailTrack`

glx22:
> building rail
> 
Documentation says
```
Parameters
    from	The tile just before the tile to build on.
    tile	The first tile to build on.
    to	The tile just after the last tile to build on. 
```
I agree not everything is enforced, but it seems to build what it's supposed to build in the screenshot.

I'm not sure all the compat stuff is needed, as the function signature didn't change.

SamuXarick:
Here's some examples what's happening right now with BuildRail. Running the same script...
image
Top right: this PR, AI API version requirements set to 14 - fails building backwards, and can build facing the void
Bottom left: 13.0-RC1, AI API version requirements set to 13 - builds backwards, can't build facing the void
Bottom right: this PR, AI API version requirements set to 13 - mimics 13.0-RC1

2TallTyler:
This PR is trying to solve two issues. 
- The missing precondition should be a separate PR. One thing per PR.
- Building to/from the void is a solution in search of a problem. And the function in the compat files is overly complex and impossible to review. 

I'm closing this PR, but feel free to open a new one to add the missing precondition.

PR #10380, opened 2023-01-21
Hide Tile and TileExtended implementation details inside Map structure

rubidium42:
## Motivation / Problem

Initial trigger was CodeQL complaining about short global variable names `_m` and `_me`. Then realizing that the whole idea of splitting `_m` and `_me` is basically arbitrary for performance reasons, so why would the user need to be aware of that.
Similarly CodeQL complains about functions defined in code blocks, so when I saw the same globals for map size related purposes being used in a similar way, I though about just creating a `Map` structure with all the size related information as well as the tiles, that provides simple access to the different fields of tiles.

With this reordering/refactoring where the different fields of the tiles are going to be stored becomes quite trivial.

## Description

Move map size related accessors to Map, as well abstracting away the Tile and TileExtended structs into Map accessors for the different fields.
Also some minor improvements (first few commits) to the code, so other more approprate functions are used instead of directly accessing the map.

The use/migrate changes have been mostly done by use of bash scripts. In case of conflicts undoing the changes and running the appropriate one(s) of the scripts below could help with conflict solution.
```
for i in `grep '_m\[' -R .|sed s/:.*//|uniq|grep -v map_func.h`; do sed "s/_me\[\([^]]*\)\]\.m\([0-9]\)/Map::m\2(\1)/g" -i $i; sed "s/_m\[\([^]]*\)\]\.m\([0-9]\)/Map::m\2(\1)/g" -i $i; sed "s/_m\[\([^]]*\)\]\.type/Map::type(\1)/g" -i $i; sed "s/_m\[\([^]]*\)\]\.height/Map::height(\1)/g" -i $i; done

for i in `grep ScaleByMapSize -R .|sed s/:.*//`; do sed s/ScaleByMapSize/Map::ScaleBySize/g -i $i; done
for i in `grep MapLog -R .|sed s/:.*//`; do sed s/MapLog/Map::Log/g -i $i; done
for i in `grep MapMax -R .|sed s/:.*//`; do sed s/MapMax/Map::Max/g -i $i; done
for i in `grep MapSize -R .|sed s/:.*//`; do sed 's/ MapSize/ Map::Size/g' -i $i; done
```

## Limitations

Breaks a lot of other patches.
Still no memory safety for the tile arrays; was thinking about using std::vector but decided against it for this PR, but that could be added relatively easily.

## Checklist for review

### DIFF
--- a/src/map_func.h
+++ b/src/map_func.h
@@ -16,22 +16,6 @@
 #include "direction_func.h"
 
 /**
- * Pointer to the tile-array.
- *
- * This variable points to the tile-array which contains the tiles of
- * the map.
- */
-extern TileBase *_m;
-
-/**
- * Pointer to the extended tile-array.
- *
- * This variable points to the extended tile-array which contains the tiles
- * of the map.
- */
-extern TileExtended *_me;
-
-/**
  * Wrapper class to abstract away the way the tiles are stored. It is
  * intended to be used to access the "map" data of a single tile.
  *
@@ -40,6 +24,36 @@ extern TileExtended *_me;
  */
 class Tile {
 private:
+	friend struct Map;
+	/**
+	 * Data that is stored per tile. Also used TileExtended for this.
+	 * Look at docs/landscape.html for the exact meaning of the members.
+	 */
+	struct TileBase {
+		byte   type;   ///< The type (bits 4..7), bridges (2..3), rainforest/desert (0..1)
+		byte   height; ///< The height of the northern corner.
+		uint16 m2;     ///< Primarily used for indices to towns, industries and stations
+		byte   m1;     ///< Primarily used for ownership information
+		byte   m3;     ///< General purpose
+		byte   m4;     ///< General purpose
+		byte   m5;     ///< General purpose
+	};
+
+	static_assert(sizeof(TileBase) == 8);
+
+	/**
+	 * Data that is stored per tile. Also used TileBase for this.
+	 * Look at docs/landscape.html for the exact meaning of the members.
+	 */
+	struct TileExtended {
+		byte m6;   ///< General purpose
+		byte m7;   ///< Primarily used for newgrf support
+		uint16 m8; ///< General purpose
+	};
+
+	static TileBase *base_tiles;         ///< Pointer to the tile-array.
+	static TileExtended *extended_tiles; ///< Pointer to the extended tile-array.
+
 	TileIndex tile; ///< The tile to access the map data for.
 
 public:
@@ -74,7 +88,7 @@ public:
 	 */
 	debug_inline byte &type()
 	{
-		return _m[tile].type;
+		return base_tiles[tile].type;
 	}
 
 	/**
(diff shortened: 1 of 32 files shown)


PeterN:
Not much to say but I like it.

LordAro:
Would be good to see some profiling to make sure all the extra function calls aren't adding any significant overhead (both debug and release)

rubidium42:
> Would be good to see some profiling to make sure all the extra function calls aren't adding any significant overhead (both debug and release)

Some statistics with `-v null -g wentbourne.sav`, GCC10 on Debian with an i5-3337U; release without asserts average of 9 runs:
Original: 64.9 seconds 
This: 65.3 seconds (+0.7%)
This with merged `TileBase + TileExtended`: 66.5 (+2.4%)
This with merged `TileBase + TileExtended` and pointer in `Tile`: 65.7 (+1.2%)
The last one can probably be improved more when usage of `Tile` gets pushed further into the call stack, so it needs to do fewer lookups.

With debug and average of 3 runs:
Original: 489 seconds
This: 574 seconds (+17%)
This with `FORCEINLINE` for `Tile`: 485 seconds (-0.6%)

So, with optimisation there is barely a noticable difference, although the current split in two arrays is slightly faster. By resolving the pointer and reusing that you can claw back some of the performance loss. 
Note: release with asserts is somewhere around 50% slower!

With debug it's definitely another story. This current version is significantly slower, and "only" enabling inlining of small functions (`-finline-small-functions`) does not seem to achieve something useful. However, "reintroducing" `FORCEINLINE` (https://meghprkh.github.io/blog/posts/c++-force-inline/) for use in such special and trivial wrapper cases brings it back to the original speed.
Note that I added `FORCEINLINE` after finding out that this version is significantly slower than the original, however I doubt that `FORCEINLINE` will have an appreciable effect on release builds, so I have not created statistics with that change.

I can imagine `FORCEINLINE` also being used in other cases of trivial wrapper functions where we just get hit by a huge penalty in performance due to not inlining at all. Though we should set up some rules for its use, like only for trivial functions that only return some value and that are called really often.

glx22:
Release builds probably already inline trivial wrappers.

rubidium42:
> I can imagine `FORCEINLINE` also being used in other cases of trivial wrapper functions where we just get hit by a huge penalty in performance due to not inlining at all. Though we should set up some rules for its use, like only for trivial functions that only return some value and that are called really often.

This got added with #10406, so now there's also no statistically significant regression on the speed of debug builds anymore.

src/map_func.h
```diff
@@ -16,26 +16,223 @@
 #include "direction_func.h"
 
 /**
- * Pointer to the tile-array.
+ * Wrapper class to abstract away the way the tiles are stored. It is
+ * intended to be used to access the "map" data of a single tile.
  *
- * This variable points to the tile-array which contains the tiles of
- * the map.
+ * The wrapper is expected to be fully optimized awa by the compiler, even
```
LordAro:
```suggestion
 * The wrapper is expected to be fully optimized away by the compiler, even
```

LordAro (dismissed):
#yolo

glx22 (approved):

PR #10382, opened 2023-01-21
Hide map size related information in Map class

rubidium42:
## Motivation / Problem

Quite a few `_map...` globals that could easily be hidden and made inaccessible inside a class. Also removes the need for using `extern _map...` in functions, and makes it harder to access them in an unwanted manner.
I had included this in #10380, but after some talking to others another more future proof idea has emerged that has to be implemented. However, these changes are separate from that so can also be merged separately.
 

## Description

Move the `Map*` functions into a new class `Map` that also contains their cached values, and only set them upon a call to `Map::Allocate` (previously `AllocateMap`). It does not mess with `Tile` and `TileExtended`; that is part of #10380.

## Limitations

Potentially some conflict in patches, but in almost each conflict changing `Map` to `Map::` will solve the issue.

## Checklist for review

Maybe enable "ignore whitespace changes" for reviewing map_func.h, as most of that is just changing indentation.

### DIFF
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -529,12 +529,12 @@ void SetAircraftPosition(Aircraft *v, int x, int y, int z)
 
 	Aircraft *u = v->Next();
 
-	int safe_x = Clamp(x, 0, MapMaxX() * TILE_SIZE);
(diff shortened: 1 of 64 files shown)


michicc (approved):
Might make backporting to 13.x harder, but maybe we should just get on with releasing stuff instead.

PR #10384, opened 2023-01-21
Change #10370: Increased max length of vehicle type names

### LINKED ISSUE #10370: Player-input vehicle name limit shorter than NewGRF name limit
### Version of OpenTTD

13.0-beta2

### Expected result

When NewGRFs can set a long name for vehicles, the player should be able to set a name of the similar length.

### Actual result

The length limit when manually renaming a vehicle is much lower than the length of default names in many NewGRFs.
Screenshot 2023-01-18 152431

### Steps to reproduce

1. Start a game with a NewGRF (for example, "2cc Trains in NML 3.1").
2. Open the vehicle purchase window/Available Trains window
3. Try to rename a vehicle with a longer name
4. Notice that you cannot edit the entire length of the existing name.

merni-ns:
From 32 to 128.

## Motivation / Problem

Closes #10370. NewGRFs are able to set names for vehicle types with >32 characters, but players cannot edit such names without deleting much of them, so that it fits within 31 characters.

## Description

Increased the limit for player-set vehicle type names from 32 to 128 characters. Per subjective observation, this is enough for most practical cases (as I have not seen a NewGRF yet where vehicle names are >128 chars.) Further, such a long name would probably exceed the width of the screen anyway. 

## Limitations

It is a quick and dirty fix that just increases a numerical hard-limit. It does not solve the underlying problem -- technically, a NewGRF can create names longer than 128 characters, which a player wouldn't be able to edit while preserving its entirety.

Probably increases the memory and space usage. The extent of such increase is unknown, but *hopefully* not significant.

The change has not been well tested.

## Checklist for review

### DIFF
--- a/src/engine_type.h
+++ b/src/engine_type.h
@@ -182,7 +182,7 @@ enum EngineFlags {
 	ENGINE_EXCLUSIVE_PREVIEW = 2, ///< This vehicle is in the exclusive preview stage, either being used or being offered to a company.
 };
 
-static const uint MAX_LENGTH_ENGINE_NAME_CHARS = 32; ///< The maximum length of an engine name in characters including '\0'
+static const uint MAX_LENGTH_ENGINE_NAME_CHARS = 128; ///< The maximum length of an engine name in characters including '\0'
 
 static const EngineID INVALID_ENGINE = 0xFFFF; ///< Constant denoting an invalid engine.
 


LC-Zorg:
The limit of 32 characters indeed does seem too low, but does 128 really make sense? Here I would even ask another question: does it make sense to allow NewGRF creators to create such long names? For me, this is the fundamental error here. Because later on, the player is forced to stretch the vehicle selection window ridiculously in order to be able to see the entire vehicle names, which just interferes with the game. I believe that if a creator had a problem with not being able to use such long names, then that problem should remain their problem, not the players problem. It's really always possible to use shorter names. The game should not allow to create problems.

Two examples of why allowing long names is a mistake.
1. V4 Trains - here is the longest vehicle name I've come across
Long vehicle names - V4 Trains 0 0
It uses 98 characters though there are two errors and to be consistent with other names it should be 108.
I like this set, apart from numerous rather minor bugs, it is refined and has much better balanced cars than the 2cc Trains it is based on. But these long names are a misunderstanding. Since I use a much narrower window (as most players probably do), I never see them in their entirety. Thus, in this case, I'm unable to find out that this locomotive is a hybrid and also has a diesel engine. Such detailed information as power or axle layout (Bo'Bo') should be in the description and not in the name. 
This is how the shortened name and the description adapted to them could look like.
Long vehicle names - V4 Trains 1 1

2. Auz TrainDepot - add-on with the longest station names, which significantly affects the size of the station selection window
Station construction window 9 - auz train depot

Yes, these are not vehicle names. However, I mention this because of the very principle of allowing very long names, which I consider a bug. In the case of stations, it is even greater because the length of names and descriptions determines the width of the station selection window. If someone creates their add-ons on a big screen, they may not even notice this problem. But later, when another player uses this add-on on a smaller screen (laptop), such a ridiculously large station selection window that covers half of the screen will only be irritating. What's more, the player won't even know what's causing it. Only after a long investigation I managed to find the culprit, because it was one of the many graphics of one of the many add-ons, which had a senselessly long name.

merni-ns:
> Here I would even ask another question: does it make sense to allow NewGRF creators to create such long names? For me, this is the fundamental error here.

> I believe that if a creator had a problem with not being able to use such long names, then that problem should remain their problem, not the players problem. 

I agree with you. Indeed, since I play on a laptop, I have had the exact 2 problems you mentioned, among others. But to fix it for the GRFs requires discussion about what the exact limit should be and whether/how to avoid breaking existing GRFs. That shoukd be done, but as a stopgap I think something like this is needed. If not 128, some smaller limit.

Btw, I didn't increase it with the intention that players should set 128 char long names (though 32 is too small). But as currently implemented they cannot even see more than 32 chars on the edit window to decide which parts of the name to keep.



------- Original Message -------

rubidium42:
Just changing the constant is likely not working just right when not using only ASCII characters. The length you give is in characters, that can be ASCII characters but also CJK ideographic characters. ASCII gets encoded as one by, most CJK ideographic as two or three bytes, just like accented Latin characters or Greek characters require at least two bytes to be encoded. The longest currently supported encoding of unicode characters takes 4 bytes.

This means that, in the worst case, your 128 characters could be 512 bytes. Or rather 509 as the 128 includes the termination which is always 1 byte, so technically one would never use the last 3 bytes of the allocation.

The way I read the string validation for commands coming from the network, it takes the first `NETWORK_COMPANY_NAME_LENGTH` (i.e. 128) bytes and makes a valid string out of that. So, when you are at your maximum vehicle name length and use some non-ASCII characters, some of the last characters go missing. Also, I think that it removes the control codes for things like string colors (after it trimmed the string to 128 bytes), but that might already be removed by the text editor UI.

So, then you think: lets change that validation limit to 512 bytes. What are the repercussions of that? Well, that means that you need to be able to store the 512 bytes of string in a command packet. That definitely fits, as there are 1460 bytes in the current packet available. But then you notice a command with three strings, which means it should have at least space for 3x512 bytes and that's way more than 1460.

Thus you must also change the maximum size the command packets may be.

So all in all, is this really about needing bigger vehicle type/engine names? Or is this a work around for NewGRF authors writing complete descriptions of a vehicle as their name? Would it help if the text editor UI allowed working on longer strings but refuse accepting too long strings, while it shows a <number of used>/<number of allowed> counter? Maybe that solves the issues with editing the strings, or at least makes it more bearable?

PR #10386, opened 2023-01-21
Fix: Some Script::IsValidVehicle checks need to be complemented with IsPrimaryVehicle

SamuXarick:
## Motivation / Problem
Some Script::IsValidVehicle checks need to be complemented with IsPrimaryVehicle, where Free Wagon vehicles are excluded, otherwise wrong values can be returned for some functions, or vehicle properties are accessed when they're not supposed to.

## Description
I tried my best judgement to add IsPrimaryVehicle checks where I think are necessary. I could be wrong in some, needs to be evaluated case by case.

## Limitations
May break some AIs, if they tried to use it on free wagon vehicles, unlikely, but...

## Checklist for review

### DIFF
--- a/src/script/api/script_vehicle.hpp
+++ b/src/script/api/script_vehicle.hpp
@@ -97,8 +97,19 @@ public:
 	 * Checks whether the given vehicle is valid and owned by you.
 	 * @param vehicle_id The vehicle to check.
 	 * @return True if and only if the vehicle is valid.
+	 * @note Also returns true when the leading part of the vehicle is a wagon.
+	 *       Use IsPrimaryVehicle() to check for a valid vehicle with a leading engine.
 	 */
 	static bool IsValidVehicle(VehicleID vehicle_id);
+
+	/**
+	 * Checks whether this is a primary vehicle.
+	 * @param vehicle_id The vehicle to check.
+	 * @pre IsValidVehicle(vehicle_id).
+	 * @return True if the vehicle is a primary vehicle.
+	 * @note Returns false when the leading part of the vehicle is a wagon.
+	 */
+	static bool IsPrimaryVehicle(VehicleID vehicle_id);
 
 	/**
 	 * Get the number of wagons a vehicle has.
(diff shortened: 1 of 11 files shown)


glx22:
I'd had a `Script::IsStartableVehicle()` to simplify the code and the documentation.

rubidium42 (dismissed):
Choice wise, i.e. what preconditions to change, it looks okay to me. There are some places where it's debatable whether to not change it to `IsPrimaryVehicle`, such as `GetState` or `IsStoppedInDepot` though it does not necessarily hurt to have them free wagon chains.

I'm not sure whether `IsPrimaryVehicle` is the right name though. I thought about it keeping it's original name, and adding a new function that also allows free wagon chains, however that might break existing scripts so that's not something I want to do.
I am wondering whether `IsPrimaryVehicle` is the best name for the new function, even thought it's the name on an internal function. Primarily as having a name that fully explains what it means by itself is favourable in an API, and what is a "primary" vehicle? One of the main four, i.e. not a disaster or effect vehicle? `IsStartableVehicle` might be a better name, as it really only returns true on something you can start, except maybe crashed vehicles... Yes, this is a difficult issue, but maybe a more proficient English speaker/writer has a better idea?

src/script/api/script_vehicle.hpp
```diff
@@ -97,9 +97,19 @@ class ScriptVehicle : public ScriptObject {
 	 * Checks whether the given vehicle is valid and owned by you.
 	 * @param vehicle_id The vehicle to check.
 	 * @return True if and only if the vehicle is valid.
+	 * @note Also returns true when the leading part of the vehicle is a wagon.
```
2TallTyler, line 100: It might be useful to direct people to use IsPrimaryVehicle() instead.
```suggestion
	 * @note Also returns true when the leading part of the vehicle is a wagon. Use IsPrimaryVehicle() to check for a valid vehicle with a leading engine.
```

2TallTyler (approved):
Looks good to me.

PR #10399, opened 2023-01-22
Feature: [NewGRF] Engine name callback.

michicc:
## Motivation / Problem
 
With the engine variants feature, the engine purchase list can potentially contain a lot of redundant information in the shown engine name due to having to contain both the engine and the variant.

## Description

To "prettify" the engine list, the engine name needs to be dynamic depending on the context it is drawn, i.e. is it the top-level in the engine list or an expanded variant, the vehicle info panel of a built vehicle, or an engine preview offer.

As dynamic data in NewGRF is usually handled by callbacks, this PR introduces and engine name callback that is passed the context where the name is used, allowing a NewGRF to decide what to show. Because most places that show the engine name don't have a real vehicle, the callback is always run in 'purchase list' scope, even for the vehicle info panel for consistency.

To avoid any performance regressions, the callback has to be enabled with a flag. As the current callback mask property for vehicles is already full, a second property (like for e.g. industries or houses) is introduced.

As callback string results are limited to the D0xx string range which is shared with all other callbacks and has only a max of 1024 strings, this PR also extends the D8/DCxx range up the the max of FFFF. The re-used string IDs are special TTDPatch-only strings that OTTD has never supported anyway and thus where completely unused. To enable usage of these strings, the callback supports the text stack so you can use `{STRING}` to reference them.

### API details

Callback number: 0x161
Flag bit 0 in property 31/28/22/22 (train/RV/ship/aircraft).
Callback is executed in 'purchase list' scope, i.e. cargotype FF and only the purchase list vars are available.

Variable 10 contains the name context. The supported values are analogue to the ones passed in [var 10 during sprite drawing](https://newgrf-specs.tt-wiki.net/wiki/Action2/Vehicles#Drawing_different_sprites_on_map_in_various_GUIs).
<table>
<tr>
	<td>Bits 0 - 7 of variable 10
	<td>Meaning
	<td>Extra data in Bits 8 - 31
<tr>
	<td>00
	<td>No specific context available (like e.g. AI/GS name query)
	<td>
<tr>
	<td>11
	<td>Vehicle details GUI
	<td>Indentation level in engine list (i.e. for variants)
<tr>
	<td>20
	<td>Purchase list/autoreplace window
	<td>
<tr>
	<td>21
	<td>Engine preview/news message
	<td>
</table>

A return value of 0x400 or a failed callback cause the game to use the use the normal Action 4 name string.
The returned string can use up to 6 parameters from the text stack.

## Limitations

Callbacks that are coded inefficiently could be a performance problem. As the name of an engine type is often only visible once on screen (i.e. it is rare to have vehicle details, purchase list and a news window open at the same) and thus only drawn with one context, caching the resulting string and only re-querying if the context is different would be possible. This would interfere with including dynamic data via the text stack, but I am not sure we should care about that.

Marked as draft for now as I am not a NewGRF author and might have missed something obvious with the API design.

## Checklist for review


michicc:
Modified NML train example NewGRF attached that demonstrates dynamic names for engine variants depending on where they are shown.

[example_train.zip](https://github.com/OpenTTD/OpenTTD/files/10475578/example_train.zip)


I have a rough NML patch for the callback and the corresponding flag. As far as I could see, there is no way to get NML to allocate a string into the D8xx range when used inside/from a switch (even if put into the text stack), so I'll gladly leave this part to someone else 😄

michicc:
Minimal basic NML patch: [cb_name.patch](https://github.com/OpenTTD/OpenTTD/files/10475640/cb_name.patch)

And because the uploaded example doesn't contain it: Yes, I did test using `{STRING}` and the text stack.

andythenorth:
Confirmed the nml patch works for simple cases.  Didn't test exceeding the 1024 string limit, nor did I test text stack.

`
switch (FEAT_TRAINS, SELF, foo_switch_name_purchase, getbits(extra_callback_info1, 8, 8)) {
    1: return string(STR_CABBAGE);
    return string(STR_PONY);
}
switch (FEAT_TRAINS, SELF, foo_switch_name, getbits(extra_callback_info1, 0, 8)) {
    0x20: foo_switch_name_purchase;
    return string(STR_CAT);
`

with a corresponding call (as example) 

`
graphics {
    name: foo_switch_name;
}
`

andythenorth:
> NB you can save some text IDs by returning CB_RESULT_NO_TEXT to make OTTD use the name property string (which does not count towards the string limit) for "uninteresting" contexts.

michicc:
Nobody screamed bloody murder yet... so I guess it's ready for review.

michicc:
Updated the example with some gratuitous text stack, search for `sw_icm_name_text_p` in the nml.

[example_train.zip](https://github.com/OpenTTD/OpenTTD/files/10493444/example_train.zip)

andythenorth:
> Updated the example with some gratuitous text stack, search for `sw_icm_name_text_p` in the nml.

Confirmed working 
image

src/aircraft_gui.cpp
```diff
@@ -34,7 +34,7 @@ void DrawAircraftDetails(const Aircraft *v, const Rect &r)
 	int y = r.top;
 	for (const Aircraft *u = v; u != nullptr; u = u->Next()) {
 		if (u->IsNormalAircraft()) {
-			SetDParam(0, u->engine_type);
+			SetDParam(0, u->engine_type | (static_cast<uint64>(EngineNameContext::VehicleDetails) << 32));
```
rubidium42: Might it be useful to hide the exact details of the bitstuffing?
```
uint64 EngineNameContextPack(EngineNameContext context, EngineID engine_id) {
    return (static_cast<uint64>(context) << 32) | engine_id;
}

std::tuple<EngineNameContext, EngineID> EngineNameContextUnpack(uint64 packed) {
    return { static_cast<EngineNameContext>(packed >> 32), static_cast<EngineID>(packed) };
}
```

Even though I whine a lot about Java missing really useful C++ features at work, here C++ kind of misses a feature of Java. The ability to add functions to enumerations. It would have been really nice to be able to write `EngineNameContext::VehicleDetails.Pack(u->engine_type)` or something similar. With a custom one for the purchase list's indent that allows two parameters.

src/strings.cpp
```diff
@@ -1371,16 +1372,36 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 			}
 
 			case SCC_ENGINE_NAME: { // {ENGINE}
-				const Engine *e = Engine::GetIfValid(args->GetInt32(SCC_ENGINE_NAME));
+				int64 arg = args->GetInt64(SCC_ENGINE_NAME);
+				const Engine *e = Engine::GetIfValid(arg & 0xFFFFFFFF);
```
rubidium42: `GB(arg, 0, 32)`? Or the unpack function?

src/strings.cpp
```diff
@@ -1371,16 +1372,36 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
+						if (callback != CALLBACK_FAILED && callback < 0x400) {
+							const GRFFile *grffile = e->GetGRF();
+							assert(grffile != nullptr);
+
+							StartTextRefStackUsage(grffile, 6);
+							uint64 tmp_dparam[6] = { 0 };
+							WChar tmp_type[6] = { 0 };
+							StringParameters tmp_params(tmp_dparam, 6, tmp_type);
+							buff = GetStringWithArgs(buff, GetGRFStringID(grffile->grfid, 0xD000 + callback), &tmp_params, last);
+							StopTextRefStackUsage();
```
rubidium42: A `break;` here, and then remove the first else block and the second else?
Similarly this massive else branch could be removed if adding a `break` after 1382 (new code). That would remove some indentation and makes it clearer that in any case something is written (except when `e == nullptr`), as there would be a final fallback with 
```
StringParameters tmp_params(nullptr, 0, nullptr);
buff = GetStringWithArgs(buff, e->info.string_id, &tmp_params, last);
```

src/aircraft_gui.cpp
michicc: Went for a pack function, but no explicit unpack as it is done only in once place.

rubidium42 (approved):
Looks okay to me, though I can't say much useful about the NewGRF stuff, especially the extending of the extended range.

andythenorth:
Left panel should show "Koploper Type 1"? 

image

PeterN:
Look at build_vehicle_gui.cpp:1042 if you want to tweak this, "show_count" is true only for the left side of the autoreplace window.

michicc:
I took the different contexts from what is distinguished for vehicle sprites (see [specs wiki](https://newgrf-specs.tt-wiki.net/wiki/Action2/Vehicles#Drawing_different_sprites_on_map_in_various_GUIs)) which explicitly combines purchase list and autoreplace. Luckily there are still 16 unused bits in var 10, so we could add more context bits 16-24 without breaking 13.1 (which a new value for bits 0..7 would do).

In general, it would help to do the extensive testing before a release, not after.

ldpl:
In general, it may help for extensive testing to actually state the problem PR is supposed to solve and not just ignore that part of the template ;)

andythenorth:
Yeah I can't be fucked.  I'm exiting GitHub as well thanks.

andythenorth:
GRF Spec status: https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status

andythenorth:
Proposal: add bit 0x12 for context "AutoreplaceListVehicleInUse". (or a shorter name for same).

This would cover the left hand "Vehicles in use" panel of autoreplace.  Right hand side would continue to use context bit 0x21.

image

This consumes a bit from the "reserved for other future GUIs" range, per nfo spec linked in preceding comment.

We'd need to

- update nfo spec
- update nmlc
- add this to the still-to-be-written nml docs

There _could_ be a case where existing grfs now rely on autoreplace context returning bit 0x20 for both sides of the panel; those grfs will unfortunately need to update their behaviour IMO.

Brickblock1:
I don't see any problem with this and I think it might even have been asked for before variants.

andythenorth:
#10666 done and merged.  NewGRF Spec Status updated.

PR #10403, opened 2023-01-23
Codechange: replace bridge cost loop by an O(1) calculation

jengelh:
The bridge costing function has poor scalability.
Looped addition is trivial and there is usually a
constant time algorithm that can be used in its place.
    
Performing 2*10^9 iterations on a i5-1135G7 CPU:
    
    len  runtime_old  runtime_new
    8      0.967s       0.517s
    16     1.706s       0.517s
    32     3.419s       0.517s
    64     6.554s       0.517s
    128    9.557s       0.517s

### DIFF
--- a/src/tunnelbridge_cmd.cpp
+++ b/src/tunnelbridge_cmd.cpp
@@ -12,6 +12,7 @@
  */
 
 #include "stdafx.h"
+#include <cmath>
 #include "newgrf_object.h"
 #include "viewport_func.h"
 #include "command_func.h"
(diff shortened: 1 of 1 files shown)


LordAro:
Well that's something. How did you even find this? :D

However, we don't like floating point numbers anywhere near the game logic as it can cause issues across the various platforms we support. Can you do an integer-based solution instead?

src/tunnelbridge_cmd.cpp
```diff
@@ -100,23 +101,18 @@ void ResetBridges()
...
-	length -= 2;
-	int sum = 2;
-	for (int delta = 1;; delta++) {
-		for (int count = 0; count < delta; count++) {
-			if (length == 0) return sum;
-			sum += delta;
-			length--;
-		}
-	}
+	length -= 2; /* ignore bridge heads */
```
2TallTyler, line 113: [OpenTTD code style](https://wiki.openttd.org/en/Development/Coding%20style) dicates `// comment` for comments on the same line as code, and `/* comment */` for comments on their own line. 🙂

FLHerne:
> What are those issues? There are already sqrt calls in e.g. tgp.cpp.

OpenTTD uses a lockstep multiplayer model - every client runs the game simulation independently, and only commands are sent over the network. For all clients to maintain the same state, everything within the game loop must be totally deterministic.

Floating-point calculations aren't reliably deterministic across all platforms - while almost all modern stuff uses IEEE 754, older hardware that OpenTTD can be compiled on does odd things (e.g. 80-bit floats on x87 or 128-bit on PowerPC), and something like sqrt may be subject to different approximations depending on the compiler or libc.

`tgp.cpp` is exempt from the above because map generation is never possible in multiplayer.

TrueBrain:
I am going to close this PR for now. If you can come up with a solution that doesn't involve floats, please do reopen it, as it would be most welcome. But as pointed out, having floats in this part of the code could possibly result in some nasty side-effects.

Still, well found place to optimize OpenTTD :D Cheers!

PR #10408, opened 2023-01-24
Codechange: Store ObjectSpecs in std::vector

PeterN:
## Motivation / Problem

Space for 64000 object types is statically allocated, on the off-chance that 64000 object types will be loaded.

As ObjectSpec is around 64 bytes (down from 72 recently), this is 4MB even if no extra object types are loaded.

Some loops also test all 64000 entries in case an object type is defined.

## Description

This is resolved by storing object types in a vector instead of a fixed-size C-style array.

This requires a minor change to how objectspecs are bound to classes as the vector will be reallocated during load, so they need to be bound after all objectspecs are loaded.

This also fixes an issue where the default objects objects are set in default classes before those default objects are actually initialised.

## Checklist for review

### DIFF
--- a/src/newgrf_object.cpp
+++ b/src/newgrf_object.cpp
@@ -90,6 +90,18 @@ uint ObjectSpec::Index() const
 	return this - _object_specs;
 }
 
+/**
+ * Tie all ObjectSpecs to their class.
+ */
+/* static */ void ObjectSpec::BindToClasses()
+{
+	for (auto &spec : _object_specs) {
+		if (spec.IsEnabled() && spec.cls_id != INVALID_OBJECT_CLASS) {
+			ObjectClass::Assign(&spec);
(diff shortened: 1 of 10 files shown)


src/newgrf_object.cpp
```diff
@@ -87,37 +101,45 @@ bool ObjectSpec::IsAvailable() const
  */
 uint ObjectSpec::Index() const
 {
-	return this - _object_specs;
+	return this - &_object_specs[0];
```
glx22:
```suggestion
	return this - _object_specs.data();
```
I think it's better to use `data()`

rubidium42 (dismissed):
I hope there's nothing wrong with the code that was not caught because some of the builds failed.

rubidium42 (approved):

Issue #10287, opened 2022-12-27
[Bug]: Useless buttons when founding town in game


hpiirai:
### Version of OpenTTD

20221226-master-g6caed5f15e, Debian GNU/Linux bookworm/sid

### Expected result

The game shows only user interface elements that might actually be used during a game.

### Actual result

When founding towns in the game is allowed, the in-game **Found town** dialogue displays UI elements that don’t make sense outside the scenario editor; specifically **Random Town**, **Many random towns**, **Expand all towns**, and the **Large** town size:
![Unnamed, 1950-01-11](https://user-images.githubusercontent.com/45715592/209690143-04d9475c-5b93-4cf9-9156-9576ba8b5fe7.png)

Moreover, when the setting is  **Allowed** rather than **Allowed, custom town layout**, the entire **Town road layout** section is useless:
![Unnamed, 1950-01-18](https://user-images.githubusercontent.com/45715592/209690147-68f5ba14-a8eb-4496-86dd-f7f9d33ed1fe.png)

### Steps to reproduce

1. On the main screen, open **Settings**.
2. Find the setting **Founding towns in game** and set it to **Allowed** or **Allowed, custom town layout**.
3. Start a new game.
4. Open the **Found town** dialogue.

redone9211:
i would like to contribute to this issue , can you let me know how i can solve it.

2TallTyler:
@LC-Zorg - Do you have any ideas for redesigning this window? It's currently a mess of unrelated buttons, some toggles and some actions...and I'm about to add another button to load a list of towns from an external file. I thought maybe you have mockups floating around your hard drive already. 😃

LC-Zorg:
I don't have. I think once I was wondering if something could be changed here, but I didn't have any good idea and to be honest I'm not even sure if it really makes sense to change anything here. Ok, this window during the game could look a bit simpler, i.e. without options that the player can't use anyway. Maybe I'll glue something together, but not today and probably not tomorrow.

In addition, there is one significant problem when creating scenarios that is worth solving: you can accidentally press Many random towns, which is very burdensome to undo. I can't see it, but I suppose you added/will add a location import button below that button. You press a few pixels higher and... you have a disaster. ;D It's worth adding a confirmation window there. Possibly, I also thought that instead of the red confirmation window, a window would appear to indicate the number of cities that would be generated, but I don't know if this is really needed by anyone. I think a simple confirmation would suffice.

[...]

PR #10409, opened 2023-01-25
Feature: Import town data from JSON file

2TallTyler:
## Motivation / Problem

smallmap

Creating real-world scenarios in Scenario Editor is slow and time-consuming.
1. Aligning towns to the correct location without a real map underlay is tricky and involves a lot of guessing.
2. NewGRFs are hard-coded to the scenario file. While many can be loaded after building the scenario, house NewGRFs cannot be changed without deleting all towns, which is most of the work involved in creating a typical scenario.
3. Changing map size requires starting from scratch.

## Description

import_gui

[Read the documentation](https://github.com/OpenTTD/OpenTTD/blob/master/docs/importing_town_data.md)!

The Town Generation GUI in Scenario Editor now lets you load towns from a JSON file, each containing the town's name, location, target OpenTTD population, and whether it is marked as a city in the game.

The intended way of generating this information is in an image editor, with the accompanying heightmap and a labeled map aligned with each other in different layers. The pixel coordinates of the town become (with some easy conversion) the tile coordinates of the town in OpenTTD.

The benefits of this approach are:

1. Placing towns accurately is much easier using a map underlay such as OpenStreetMap to match town locations with the corresponding heightmap.
2. Storing town data in a JSON file instead of as an OpenTTD Scenario (.scn) doesn't require storing NewGRFs, particularly the NewGRF house set. You can load the data with whatever house set you choose.
3. Town coordinates are scaled by the map size, so you can choose the size that works for you.

This standardized format is an API which allows future tools to generate this data and feed it to OpenTTD.

You can download a sample heightmap and town data set from the .zip below, and copy both to `OpenTTD\scenario\heightmap`.

[example_data.zip](https://github.com/OpenTTD/OpenTTD/files/10495233/example_data.zip)

If you'd like to try out this PR, you can [download a pre-compiled build](https://cdn.openttd.org/openttd-branches/pr10409/20240626-pr10409-geba7366775/).

## Out of scope for this PR

- Feature: Save town data to JSON file using Scenario Editor
- Feature: Package heightmap and town data files in .tar files, and allow upload/download from BaNaNaS
- Change: Hide disabled buttons on Town Generation GUI (#10287)
- Feature: Import trees/desert/rivers/objects/etc. from file

## Checklist for review

### DIFF
--- a/docs/importing_town_data.md
+++ b/docs/importing_town_data.md
@@ -1,0 +1,97 @@
+# Importing Town Data into OpenTTD
+
+To aid players in scenario creation, OpenTTD's Scenario Editor can import town data from external JSON files. This enables players to use an image editing program to align town coordinates with a real-world heightmap using a map underlay, instead of guessing at the correct locations in Scenario Editor itself.
+
+This town data consists of a JSON file storing an array of town data objects, each containing a name, location, target OpenTTD population, and whether it is marked as a city in the game.
+
+This document describes the standard format for this JSON file and outlines a workflow for creating this data effectively.
+
+## Table of contents
+
+- Why load external data?
+- How to use this feature
+  - Creating geodata
+    - Town data format standards
+    - Town data values
+  - Loading geodata into OpenTTD
+- Tutorial: Creating town data
+
+## Why load external data?
+
+There are three benefits to using an image editing program to create towns instead of the OpenTTD Scenario Editor.
+
+1. Placing towns accurately is much easier using a map underlay such as OpenStreetMap to match town locations with the corresponding heightmap.
+2. Storing town data in a JSON file instead of as an OpenTTD Scenario (.scn) doesn't require choosing your NewGRF house set before placing towns.
+3. Town coordinates are scaled by the map size, so you can load the data onto whatever size map you like.
+
+## How to use this feature
+
+### Creating geodata
+
+Town data is a text file in the JSON format, with a list of towns, each containing a coordinate location and properties: name, population, and whether or not it should be a city in OpenTTD.
+
+The format of this file is standardized for importing into OpenTTD and must be followed for OpenTTD to properly parse the data.
+
+For use in OpenTTD, you will also need a matching heightmap of the terrain features, as a PNG.
+
+#### Town data format standards
+
+The following code sample is complete and can be used in OpenTTD.
+- The list of towns is enclosed in an array marked with square brackets `[]`
+- Each town is enclosed in curly braces `{}`, with a comma after each town except for the last in the list.
+- The properties separated by commas except for the last.
+- Property names are enclosed in double quotes `""` with a colon `:` separating it from the property value.
+- The name property value is enclosed in double quotes `"London"`, while all other property values `44910`, `true`, etc., are not.
+
+```
+[
+    {
+        "name": "London",
+        "population": 44910,
+        "city": true,
+        "x": 0.7998046875,
+        "y": 0.708984375
+    },
+    {
+        "name": "Canterbury",
+        "population": 217.16,
+        "city": false,
+        "x": 0.83251953125,
+        "y": 0.828125
+    }
+]
+```
+
+#### Town data values
+
+- Population is scaled down for use in OpenTTD. It is possible to generate huge cities by using a large number, but there is a practical limit to town size. The larger the town, the longer it will take to import town data, since towns are placed at a relatively small size and then expanded until the population is greater than the player-defined target.
+- X and Y coordinates are a proportion of the total map dimension, between 0 and 1. Just take the pixel coordinates of the town's location in the corresponding heightmap (more detail in the tutorial below) and divide each by the maximum value.
+  - For example, London is at `726, 1638` in a 1024 px by 2048 px heightmap, so `726 / 1024 = 0.7998046875` and `1638 / 2048 = 0.708984375` gives the correct coordinates for OpenTTD.
+  - The reason for these proportional coordinates is so the data can be used for any map size.
+- 0,0 is (approximately) the very top tile in OpenTTD. You can see tile coordinates in-game with the Land Info Tool.
+- In most image editing programs, 0,0 is the top-left corner of the image. You can rotate the image however you want relative to compass north to orient the map to your liking. Make sure you crop and resize the image before recording town locations.
+- In OpenTTD, X and Y axis are swapped compared to most image editing programs and the standard Cartesian coordinate system. From the 0,0 origin at top left, X is the axis along the left side and Y is the axis along the right side. You can still measure X and Y coordinates in your image editing program, just swap them before importing into OpenTTD or towns won't line up with your heightmap.
+
+### Loading geodata into OpenTTD
+Using geodata to create a real-world location in OpenTTD is done in the Scenario Editor.
+
+1. Choose the NewGRFs you want to use in the game.
+2. Load the heightmap which you created in the geodata workflow. Either rotation will work, but the clockwise rotation is considered "correct" and the coordinates in the Land Info Tool will match your data; counter clockwise maps will align properly but the coordinates won't match your data.
+3. In the Town Generation window, click `Load from file` and choose the .json file containing town data. The default directory to search for town data is `OpenTTD\scenario\heightmap`.
+4. (Optional) Manually add industries, rivers, trees, and objects.
+5. Save the game as a Scenario and exit to the main menu.
+6. Load the game with Play Scenario and enjoy.
+
+Sometimes it's not possible to place a town, such as when the heightmap is very rough and a flat tile can't be found with a 16-tile radius of the target tile. In such cases, a sign will be placed on the target tile with the name of the town. The player can then place the town manually or change the heightmap settings and try again. This fallback also helps debug errors with data creation, such as if towns end up in the ocean.
+
+## Tutorial: Creating town data
+
+1. Load both your heightmap and a labeled map like OpenStreetMap as layers in an image editing program. You can use a free/open-source program like QGIS to acquire, align, and export these map images, if you like.
+2. Crop the image to your desired bounds, ensuring the aspect ratio is supported in OpenTTD (1:1, 1:2, 1:4, etc.).
+3. Resize the image to one of OpenTTD's supported map sizes, such as 512 px by 1024 px. Some image editors let you do this part of step 2. You can always load heightmaps and town data at a reduced size, so you may want to make this larger than your intended use in case you want it later.
+4. Use the labeled map layer to find the pixel coordinates of each town you'd like to include in your map. In GIMP this is displayed in the bottom left corner of the image window, and in Photoshop you need to enable the Info panel (F8) and switch to pixel units of measurement if not already.
+5. Some spreadsheets including Google Sheets can export data as JSON, so you may want to record it there, to export after step 8. Or you can build the JSON file manually.
+6. Adjust population numbers for OpenTTD.
+7. Change coordinates from pixels to proportion (0-1) of the total dimension: `x / maximum_x` and `y / maximum_y`, as described in "Town data values" above.
+8. Swap X and Y coordinates before importing to OpenTTD, since OpenTTD uses a reverse X and Y system than most image editors.
+9. Save the heightmap and town data files in your `OpenTTD\scenario\heightmap` folder.
--- a/src/fios.cpp
+++ b/src/fios.cpp
@@ -84,6 +84,10 @@ void FileList::BuildFileList(AbstractFileType abstract_filetype, SaveLoadOperati
 			FiosGetHeightmapList(fop, show_dirs, *this);
 			break;
 
+		case FT_TOWN_DATA:
+			FiosGetTownDataList(fop, show_dirs, *this);
+			break;
+
 		default:
 			NOT_REACHED();
 	}
@@ -180,6 +184,7 @@ bool FiosBrowseTo(const FiosItem *item)
 		case FIOS_TYPE_OLD_SCENARIO:
 		case FIOS_TYPE_PNG:
 		case FIOS_TYPE_BMP:
+		case FIOS_TYPE_JSON:
 			return false;
 	}
 
@@ -552,6 +557,42 @@ void FiosGetHeightmapList(SaveLoadOperation fop, bool show_dirs, FileList &file_
 	std::string base_path = FioFindDirectory(HEIGHTMAP_DIR);
 	Subdirectory subdir = base_path == *_fios_path ? HEIGHTMAP_DIR : NO_DIRECTORY;
 	FiosGetFileList(fop, show_dirs, &FiosGetHeightmapListCallback, subdir, file_list);
+}
+
+/**
+ * Callback for FiosGetTownDataList.
+ * @param fop Purpose of collecting the list.
+ * @param file Name of the file to check.
+ * @return a FIOS_TYPE_JSON type of the found file, FIOS_TYPE_INVALID if not a valid JSON file, and the title of the file (if any).
+ */
+static std::tuple<FiosType, std::string> FiosGetTownDataListCallback(SaveLoadOperation fop, const std::string &file, const std::string_view ext)
+{
+	if (fop == SLO_LOAD) {
+		if (StrEqualsIgnoreCase(ext, ".json")) {
+			return { FIOS_TYPE_JSON, GetFileTitle(file, SAVE_DIR) };
+		}
+	}
+
+	return { FIOS_TYPE_INVALID, {} };
+}
+
+/**
+ * Get a list of town data files.
+ * @param fop Purpose of collecting the list.
+ * @param show_dirs Whether to show directories.
+ * @param file_list Destination of the found files.
+ */
+void FiosGetTownDataList(SaveLoadOperation fop, bool show_dirs, FileList &file_list)
+{
+	static std::optional<std::string> fios_town_data_path;
+
+	if (!fios_town_data_path) fios_town_data_path = FioFindDirectory(HEIGHTMAP_DIR);
+
+	_fios_path = &(*fios_town_data_path);
+
+	std::string base_path = FioFindDirectory(HEIGHTMAP_DIR);
+	Subdirectory subdir = base_path == *_fios_path ? HEIGHTMAP_DIR : NO_DIRECTORY;
+	FiosGetFileList(fop, show_dirs, &FiosGetTownDataListCallback, subdir, file_list);
 }
 
 /**
--- a/src/fios.h
+++ b/src/fios.h
@@ -107,6 +107,7 @@ void ShowSaveLoadDialog(AbstractFileType abstract_filetype, SaveLoadOperation fo
 void FiosGetSavegameList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
 void FiosGetScenarioList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
 void FiosGetHeightmapList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
+void FiosGetTownDataList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
 
 bool FiosBrowseTo(const FiosItem *item);
 
--- a/src/fios_gui.cpp
+++ b/src/fios_gui.cpp
@@ -23,6 +23,7 @@
 #include "querystring_gui.h"
 #include "engine_func.h"
 #include "landscape_type.h"
+#include "genworld.h"
 #include "timer/timer_game_calendar.h"
 #include "core/geometry_func.hpp"
 #include "gamelog.h"
@@ -172,6 +173,48 @@ static constexpr NWidgetPart _nested_load_heightmap_dialog_widgets[] = {
 	EndContainer(),
 };
 
+/** Load town data */
+static constexpr NWidgetPart _nested_load_town_data_dialog_widgets[] = {
+	NWidget(NWID_HORIZONTAL),
+		NWidget(WWT_CLOSEBOX, COLOUR_GREY),
+		NWidget(WWT_CAPTION, COLOUR_GREY, WID_SL_CAPTION),
+		NWidget(WWT_DEFSIZEBOX, COLOUR_GREY),
+	EndContainer(),
+	/* Current directory and free space */
+	NWidget(WWT_PANEL, COLOUR_GREY, WID_SL_BACKGROUND), SetFill(1, 0), SetResize(1, 0), EndContainer(),
+
+	/* Filter box with label */
+	NWidget(WWT_PANEL, COLOUR_GREY), SetFill(1, 1), SetResize(1, 1),
+		NWidget(NWID_HORIZONTAL), SetPadding(WidgetDimensions::unscaled.framerect.top, 0, WidgetDimensions::unscaled.framerect.bottom, 0),
+				SetPIP(WidgetDimensions::unscaled.frametext.left, WidgetDimensions::unscaled.frametext.right, 0),
+			NWidget(WWT_TEXT, COLOUR_GREY), SetFill(0, 1), SetDataTip(STR_SAVELOAD_FILTER_TITLE , STR_NULL),
+			NWidget(WWT_EDITBOX, COLOUR_GREY, WID_SL_FILTER), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_LIST_FILTER_OSKTITLE, STR_LIST_FILTER_TOOLTIP),
+		EndContainer(),
+	EndContainer(),
+	/* Sort Buttons */
+	NWidget(NWID_HORIZONTAL),
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_SORT_BYNAME), SetDataTip(STR_SORT_BY_CAPTION_NAME, STR_TOOLTIP_SORT_ORDER), SetFill(1, 0), SetResize(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_SORT_BYDATE), SetDataTip(STR_SORT_BY_CAPTION_DATE, STR_TOOLTIP_SORT_ORDER), SetFill(1, 0), SetResize(1, 0),
+		EndContainer(),
+		NWidget(WWT_PUSHIMGBTN, COLOUR_GREY, WID_SL_HOME_BUTTON), SetAspect(1), SetDataTip(SPR_HOUSE_ICON, STR_SAVELOAD_HOME_BUTTON),
+	EndContainer(),
+	/* Files */
+	NWidget(NWID_HORIZONTAL),
+		NWidget(WWT_PANEL, COLOUR_GREY, WID_SL_FILE_BACKGROUND),
+			NWidget(WWT_INSET, COLOUR_GREY, WID_SL_DRIVES_DIRECTORIES_LIST), SetFill(1, 1), SetPadding(2, 2, 2, 2),
+					SetDataTip(0x0, STR_SAVELOAD_LIST_TOOLTIP), SetResize(1, 10), SetScrollbar(WID_SL_SCROLLBAR), EndContainer(),
+		EndContainer(),
+		NWidget(NWID_VSCROLLBAR, COLOUR_GREY, WID_SL_SCROLLBAR),
+	EndContainer(),
+	/* Load button */
+	NWidget(NWID_HORIZONTAL, NC_EQUALSIZE),
+		NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_LOAD_BUTTON), SetResize(1, 0), SetFill(1, 0),
+				SetDataTip(STR_SAVELOAD_LOAD_BUTTON, STR_SAVELOAD_LOAD_TOWN_DATA_TOOLTIP),
+		NWidget(WWT_RESIZEBOX, COLOUR_GREY),
+	EndContainer(),
+};
+
 /** Save game/scenario */
 static constexpr NWidgetPart _nested_save_dialog_widgets[] = {
 	NWidget(NWID_HORIZONTAL),
@@ -242,6 +285,7 @@ static const TextColour _fios_colours[] = {
 	TC_ORANGE,       // DFT_GAME_FILE
 	TC_YELLOW,       // DFT_HEIGHTMAP_BMP
 	TC_ORANGE,       // DFT_HEIGHTMAP_PNG
+	TC_LIGHT_BROWN,  // DFT_TOWN_DATA_JSON
 	TC_LIGHT_BLUE,   // DFT_FIOS_DRIVE
 	TC_DARK_GREEN,   // DFT_FIOS_PARENT
 	TC_DARK_GREEN,   // DFT_FIOS_DIR
@@ -330,6 +374,7 @@ public:
 					break;
 
 				default:
+					/* It's not currently possible to save town data. */
 					NOT_REACHED();
 			}
 		}
@@ -354,6 +399,10 @@ public:
 
 			case FT_HEIGHTMAP:
 				caption_string = (this->fop == SLO_SAVE) ? STR_SAVELOAD_SAVE_HEIGHTMAP : STR_SAVELOAD_LOAD_HEIGHTMAP;
+				break;
+
+			case FT_TOWN_DATA:
+				caption_string = STR_SAVELOAD_LOAD_TOWN_DATA; // It's not currently possible to save town data.
 				break;
 
 			default:
@@ -391,6 +440,7 @@ public:
 				break;
 
 			case FT_HEIGHTMAP:
+			case FT_TOWN_DATA:
 				o_dir.name = FioFindDirectory(HEIGHTMAP_DIR);
 				break;
 
@@ -634,6 +684,9 @@ public:
 				if (this->abstract_filetype == FT_HEIGHTMAP) {
 					this->Close();
 					ShowHeightmapLoad();
+				} else if (this->abstract_filetype == FT_TOWN_DATA) {
+					this->Close();
+					LoadTownData();
 				} else if (!_load_check_data.HasNewGrfs() || _load_check_data.grf_compatibility != GLC_NOT_FOUND || _settings_client.gui.UserIsAllowedToChangeNewGRFs()) {
 					_switch_mode = (_game_mode == GM_EDITOR) ? SM_LOAD_SCENARIO : SM_LOAD_GAME;
 					ClearErrorMessages();
@@ -689,7 +742,7 @@ public:
 				} else if (!_load_check_data.HasErrors()) {
 					this->selected = file;
 					if (this->fop == SLO_LOAD) {
-						if (this->abstract_filetype == FT_SAVEGAME || this->abstract_filetype == FT_SCENARIO) {
+						if (this->abstract_filetype == FT_SAVEGAME || this->abstract_filetype == FT_SCENARIO || this->abstract_filetype == FT_TOWN_DATA) {
 							this->OnClick(pt, WID_SL_LOAD_BUTTON, 1);
 						} else {
 							assert(this->abstract_filetype == FT_HEIGHTMAP);
@@ -856,6 +909,7 @@ public:
 
 				switch (this->abstract_filetype) {
 					case FT_HEIGHTMAP:
+					case FT_TOWN_DATA:
 						this->SetWidgetDisabledState(WID_SL_LOAD_BUTTON, this->selected == nullptr || _load_check_data.HasErrors());
 						break;
 
@@ -908,6 +962,14 @@ static WindowDesc _load_heightmap_dialog_desc(
 	_nested_load_heightmap_dialog_widgets
 );
 
+/** Load town data */
+static WindowDesc _load_town_data_dialog_desc(
+	WDP_CENTER, "load_town_data", 257, 320,
+	WC_SAVELOAD, WC_NONE,
+	0,
+	_nested_load_town_data_dialog_widgets
+);
+
 /** Save game/scenario */
 static WindowDesc _save_dialog_desc(
 	WDP_CENTER, "save_game", 500, 294,
@@ -929,6 +991,17 @@ void ShowSaveLoadDialog(AbstractFileType abstract_filetype, SaveLoadOperation fo
 		new SaveLoadWindow(_save_dialog_desc, abstract_filetype, fop);
 	} else {
 		/* Dialogue for loading a file. */
-		new SaveLoadWindow((abstract_filetype == FT_HEIGHTMAP) ? _load_heightmap_dialog_desc : _load_dialog_desc, abstract_filetype, fop);
+		switch (abstract_filetype) {
+			case FT_HEIGHTMAP:
+				new SaveLoadWindow(_load_heightmap_dialog_desc, abstract_filetype, fop);
+				break;
+
+			case FT_TOWN_DATA:
+				new SaveLoadWindow(_load_town_data_dialog_desc, abstract_filetype, fop);
+				break;
+
+			default:
+				new SaveLoadWindow(_load_dialog_desc, abstract_filetype, fop);
+		}
 	}
 }
--- a/src/genworld.cpp
+++ b/src/genworld.cpp
@@ -10,6 +10,10 @@
 #include "stdafx.h"
 #include "landscape.h"
 #include "company_func.h"
+#include "town_cmd.h"
+#include "signs_cmd.h"
+#include "3rdparty/nlohmann/json.hpp"
+#include "strings_func.h"
 #include "genworld.h"
 #include "gfxinit.h"
 #include "window_func.h"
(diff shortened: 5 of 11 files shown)


src/misc/json.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+	if (read != filesize) return false;
+
+	/* Add a trailing \0 to mark the end of the string */
+	text = ReallocT(text, filesize + 1);
+	text[filesize] = '\0';
+
+	/* The parser doesn't like tabs or various newline characters, so we'll replace them with a space. */
+	for (char *p = text; *p != '\0'; p++) {
+		if (*p == '\t' || *p == '\r' || *p == '\n' || *p == '\r\n') *p = ' ';
```
glx22: As reported by CI '\r\n' is indeed not a single char but 2 chars, and they are already handled anyway :)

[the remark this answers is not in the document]
PeterN: Strings are not chars :-)

PeterN:
json11 is archived so I'm not sure it's a good choice to start implementing with.

src/misc/json.hpp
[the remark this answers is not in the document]
nielsmh: This is probably triggering on the `*p == '\r\n'` which almost certainly does not do what the author intended it to.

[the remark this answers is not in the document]
2TallTyler: I am the author of this file. 😛 I fixed it already, though.

LC-Zorg:
For scenario creators this is a really great feature. :) Thanks to this, the scenario of the map of Poland, the Visegrad group (Poland, Czech Republic, Slovakia, Hungary) and Europe, which used to be also on the server, was created. The first scenario already had a lot of updates, including changes to industries sets and scripts, which required the map to be regenerated from scratch. Without this feature, it would hardly make sense. Here I would like to point out that a version of the game was used to generate the map (added the aforementioned McZapkie's repeatable map generation tools) which also allowed importing rivers, objects (stone and transmitters as borders) and enterprises - features that would be equally useful. I admit that although I helped a lot in refining these maps, I've never used this version of the game (needs to be compiled) and I don't know exactly how it works. I know that rivers and objects can be imported from an image file and all functions are only available from the console level so it's not very convenient for the average user.

> Should BaNaNaS support uploading and downloading this data (maybe bundled with the heightmap as a .tar)? I want to consider compatibility with future data layers like trees and desert (not objects though, that's a manual-only tool). I would need help with this.

I think definitely yes, but maybe it would be worth adding a new format?
- **Heightmap** - the height map itself
- **Map** - height map with other data or layers added (city locations, rivers, water depth, deserts, snow, trees, roads and more), the player still has the option to choose his own industrial set, roads, buildings, vehicles and script; allowing for the collection of, for example, the water layer itself or cities would not make sense without making available the base it refers to
- **Scenario** - fully ready map

In the case of the Map and the Scenario, it would be good if it was possible to export individual layers to graphic files (water, trees, etc.) or text files (cities, enterprises)

The ability to use more layers could be added gradually. Although I write this as a theoretician, it seems to me that the most important thing is to define the framework of the basis on which all this could be based.

For the future:
As for deserts and snow, I think it would be worthwhile to go even further and make maps with many different substrates. For example, a map of Europe from snowy Scandinavia to the deserts of northern Africa (look at tt-distas server with map of Europe). The same with the map of the United States itself, where desert or rocky areas can connect directly to snow-covered areas. Snow could be a separate layer, where different colors on png file would correspond to seasonal changes.

When it comes to rivers and lakes, it took me over 30 hours to refine them on the Visegrad map. I think being able to save this layer, outside of cities, would be most helpful.


> Redesign the Town Generation GUI. It was bad enough before I added another button.

I'm not sure if it's worth changing it any more significantly - I wrote a bit more [here #10287](https://github.com/OpenTTD/OpenTTD/issues/10287#issuecomment-1405841468). :)

2TallTyler:
> json11 is archived so I'm not sure it's a good choice to start implementing with.

There has been some discussion about this on Discord, but without any resolution. I'd appreciate any thoughts on an alternative library.

From what I've found, the main differences seem to be whether they're one huge header file or several smaller header files. I quite like the look of this one, but its single include header file is 24,640 lines...
https://github.com/nlohmann/json
https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp

RapidJSON also looks nice and is apparently smaller, but is a bunch more files to include 
https://rapidjson.org/index.html
https://github.com/Tencent/rapidjson/tree/master/include/rapidjson

MiguelHorta:
> > json11 is archived so I'm not sure it's a good choice to start implementing with.
> 
> There has been some discussion about this on Discord, but without any resolution. I'd appreciate any thoughts on an alternative library.

Hello,
I think the first thing that is intriguing me is why did you chose to include it in-tree? I don't see why it is necessary(, or even advisable). 

* It is not used by code that deals with game logic. So you gain nothing from having a stable curated snapshot;
* It is not trivial code. You can't easily check for its correctness. Sometime in the future there will be fixes upstream.
* It is not a critical dependency. There might be ports/packaging that benefit from having it disabled. 

Personalty, I would chose to bring it as a dependency, that can be disabled during configuration phase.

> 
> From what I've found, the main differences seem to be whether they're one huge header file or several smaller header files. I quite like the look of this one, **but its single include header file is 24,640 lines...** https://github.com/nlohmann/json https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp

Does that matter? I don't think that's a worthy metric. I think the correct questions are:

* Is is actively maintained? Is it used/backed by a large organization/product?
* Is it well documented?
* Does it support every target in which openttd is currently compiled?
* If used as dependency, is there support for CMAKE, VCPKG, brew, and is present in major Linux distributions repositories?
* Performance wise, how does it compares? (https://github.com/miloyip/nativejson-benchmark)
* Does it integrate well in openttd's code, or does it feels alien? i.e. does it make use of modern C++ features?
* Does it follows/implements fully the json spec? Does it adds anything on top of json spec that is noteworthy or deal-breaking?

I don't have a suggestion at the moment, but I think that this list of questions should be useful to reach a consensus.

src/genworld.cpp
```diff
@@ -326,3 +329,103 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+
+		/* Find the target tile for the town. */
+		TileIndex tile = TileXY(town.x_percent * Map::MaxX(), town.y_percent * Map::MaxY());
+
+		/* Try founding on the target tile, and if that doesn't work, find the nearest suitable tile up to 16 tiles away.
+		 * The target might be on water, blocked somehow, or on a steep slope that can't be terraformed by the founding command. */
+		TileIndex search_tile = tile;
+		bool success = CircularTileSearch(&search_tile, 16, 0, 0, TryFoundTownNearby, &town);
+
+		/* If we still fail to found the town, we'll create a sign at the intended location and tell the player how many towns we failed to create in an error message. */
```
TrueBrain: This is kinda nice :) A bit weird, but also nice :)

src/genworld.cpp
```diff
@@ -326,3 +329,103 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		}
+
+		/* Now that we've created the town, let's grow it to the target population. */
+		Town* t = ClosestTownFromTile(search_tile, UINT_MAX);
+		uint tries = 500;
+
+		do {
+			Command<CMD_EXPAND_TOWN>::Post(t->index, 0);
+			if (tries-- == 0) break;
+		} while (t->cache.population < town.population);
```
TrueBrain: This is a bit weird .. only 500 tries? Why not 500 failures? Feels a bit random, 500.

src/genworld.h
```diff
@@ -82,13 +82,23 @@ enum GenWorldProgress {
 	GWP_CLASS_COUNT
 };
 
+/** Town data imported from JSON files and used to place towns. */
+struct ExternalTownData {
+	std::string name;            ///< The name of the town.
+	uint population;             ///< The target population of the town when created in OpenTTD. If input is blank, defaults to 0.
+	bool is_city;                ///< Should it be created as a city in OpenTTD? If input is blank, defaults to false.
+	double x_percent;            ///< The X coordinate of the town, as a percentage 0..1 of the maximum X coordinate.
+	double y_percent;            ///< The Y coordinate of the town, as a percentage 0..1 of the maximum Y coordinate.
```
TrueBrain: Something to think about, but do we really need a double here?

src/fileio.cpp
```diff
@@ -42,6 +42,7 @@ static const char * const _subdirs[] = {
 	"save" PATHSEP "autosave" PATHSEP,
 	"scenario" PATHSEP,
 	"scenario" PATHSEP "heightmap" PATHSEP,
+	"scenario" PATHSEP "towndata" PATHSEP,
```
TrueBrain: Just as a FYI, if you want to be able to load heightmaps + towndata, they can't be in different folders. How the content system works, is that 1 content piece needs everything in a single folder.

But I also like your idea of having this for SE-only, that people need to load towndata on top of their scenario / heightmap. But just the headsup.

src/genworld.cpp
2TallTyler: Why not 500 tries? This is to expand the town to the target population specified by the player in the JSON file, so we need some way to break out of the do/while loop in case their target population is not reachable. (Which is extremely likely, since some players will inevitably try to match real-world city populations of 1M+, which doesn't work in OpenTTD...)

Alternate suggestions welcome. 😃

src/fileio.cpp
2TallTyler: Ah, thanks for noting this. Something to consider as we move forward with this and consider bundling town data with heightmaps or other info.

docs/importing_town_data.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Table of contents
+
+1. Why load external data?
+2. How to use this feature
+   - Creating geodata
+     - Town data format standards
+	 - Town data values
+   - Loading geodata into OpenTTD
+3. Tutorial: Creating town data
```
PeterN: TOC indenting looks weird. L14-15 should be at the same level based on the header type, but there a tab on L15. Why use numbers for the top-level which aren't present on the sections themselves.

docs/importing_town_data.md
```diff
@@ -0,0 +1,98 @@
...
+    }
+]
+```
+
+#### Town data values
+
+- Population is scaled down for use in OpenTTD. It is possible to generate huge cities by using a large number, but there is a practical limit to town size. The larger the town, the longer it will take to import town data, since towns are placed at a relatively small size and then expanded until the population is greater than the player-defined target.
+- X and Y coordinates are a percent of the total map dimension, between 0 and 1. Just take the pixel coordinates of the town's location in the corresponding heightmap (more detail in the tutorial below) and divide each by the maximum value.
+  - For example, London is at `726, 1638` in a 1024 px by 2048 px heightmap, so `726 / 1024 = 0.7998046875` and `1638 / 2048 = 0.708984375` gives the correct coordinates for OpenTTD.
+  - The reason for these percentage coordinates is so the data can be used for any map size.
```
PeterN: By definition, percent is 0-100, not 0-1. This is a proportion, or ratio, or some other wording. Normalised to range [0,1].

src/fileio.cpp
PeterN: Not who this affects considerations, but I would think that town placement data is only useful for a specific heightmap. A scenario can be considered a heightmap and town data pre-combined.

src/fios_gui.cpp
```diff
@@ -320,6 +364,7 @@ struct SaveLoadWindow : public Window {
 					break;
 
 				default:
+					// It's not currently possible to save town data.
 					NOT_REACHED();
```
PeterN: In general we only use "//" style comments at the end of line.

src/fios_gui.cpp
```diff
@@ -168,6 +169,48 @@ static const NWidgetPart _nested_load_heightmap_dialog_widgets[] = {
...
+			NWidget(NWID_VSCROLLBAR, COLOUR_GREY, WID_SL_SCROLLBAR),
+		EndContainer(),
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_LOAD_BUTTON), SetResize(1, 0), SetFill(1, 0),
+					SetDataTip(STR_SAVELOAD_LOAD_BUTTON, STR_SAVELOAD_LOAD_TOWN_DATA_TOOLTIP),
+			NWidget(WWT_RESIZEBOX, COLOUR_GREY),
+		EndContainer(),
+	EndContainer(),
+EndContainer(),
+};
```
PeterN: This is out of date with some other layout changes—the panel at the bottom is wrong, and there's an extra EndContainer() in the wrong place.

Ideally this wouldn't be duplicated at all.

src/genworld.cpp
```diff
@@ -326,3 +330,137 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	return (Command<CMD_FOUND_TOWN>::Post(tile, TSZ_SMALL, town.is_city, _settings_game.economy.town_layout, false, 0, town.name));
+}
+
+/**
+ * Load town data from _file_to_saveload and place its contents into the world.
+ * @return True if the JSON file is valid and has north and south extents, otherwise false.
+ */
+void LoadTownData()
+{
+	using json = nlohmann::json;
```
PeterN: I don't think this is a good idea.

src/genworld.cpp
```diff
@@ -326,3 +330,137 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	}
+
+	json town_data;
+
+	try {
+		town_data = json::parse(text);
+	}
+	catch (json::exception&) {
+		ShowErrorMessage(STR_TOWN_DATA_ERROR_LOAD_FAILED, STR_TOWN_DATA_ERROR_JSON_FORMATTED_INCORRECTLY, WL_ERROR);
+	}
```
PeterN: `catch` on the same like as the `}`

You probably want to `return;` when inside the catch too.

src/genworld.cpp
```diff
@@ -326,3 +330,137 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+
+	text = ReallocT(text, filesize);
+	size_t read = fread(text, 1, filesize, handle);
+	fclose(handle);
+
+	if (read != filesize) return;
+
+	/* Add a trailing \0 to mark the end of the string */
+	text = ReallocT(text, filesize + 1);
+	text[filesize] = '\0';
```
PeterN: Pretty sure you should be able to read directly into a std::string (resized to filesize), instead of affing with C-style memory allocation and null-termination.

src/genworld.h
```diff
@@ -82,13 +82,23 @@ enum GenWorldProgress {
 	GWP_CLASS_COUNT
 };
 
+/** Town data imported from JSON files and used to place towns. */
+struct ExternalTownData {
+	std::string name; ///< The name of the town.
+	uint population;  ///< The target population of the town when created in OpenTTD. If input is blank, defaults to 0.
+	bool is_city;     ///< Should it be created as a city in OpenTTD? If input is blank, defaults to false.
+	float x_percent;  ///< The X coordinate of the town, as a percentage 0..1 of the maximum X coordinate.
+	float y_percent;  ///< The Y coordinate of the town, as a percentage 0..1 of the maximum Y coordinate.
```
PeterN: 0..1 is not a percentage.

src/genworld.h
```diff
@@ -82,13 +82,23 @@ enum GenWorldProgress {
...
+};
```
PeterN: Does this struct need to be exposed outside of genworld.cpp?

src/genworld.cpp
2TallTyler: Yes, I just can't figure out how... 😕

src/fios_gui.cpp
2TallTyler: This GUI alone doesn't have an online content button, so I'm not sure how to de-duplicate it.

I believe I have fixed the layout.

src/fileio.cpp
2TallTyler: My rationale for doing towns and heightmaps separately instead of bundling them in a .tar or similar, is that the heightmap import often doesn't go perfectly the first try. The max height is wrong, or some coasts need cleanup, etc. Loading towns takes a lot of time, so it would be frustrating to force them to happen simultaniously.

For now, this format is for scenario creators, not a full "reproducible scenario" format for use by the average player.

That still doesn't answer the question of where to put the files. I guess we could always move them later.

2TallTyler:
Updated and rebased. There are still outstanding review comments which I need help with.

src/genworld.cpp
2TallTyler: Figured it out. 😃

src/fileio.cpp
2TallTyler: I've moved town data into the heightmap directory. I don't plan to bundle anything in this PR, but this will make it easier later.

ahyangyi:
> I think definitely yes, but maybe it would be worth adding a new format?
> 
> * **Heightmap** - the height map itself
> * **Map** - height map with other data or layers added (city locations, rivers, water depth, deserts, snow, trees, roads and more), the player still has the option to choose his own industrial set, roads, buildings, vehicles and script; allowing for the collection of, for example, the water layer itself or cities would not make sense without making available the base it refers to
> * **Scenario** - fully ready map
> 
> In the case of the Map and the Scenario, it would be good if it was possible to export individual layers to graphic files (water, trees, etc.) or text files (cities, enterprises)
> 
> The ability to use more layers could be added gradually. Although I write this as a theoretician, it seems to me that the most important thing is to define the framework of the basis on which all this could be based.

It seems to me that a tarball with a well-documented naming scheme is good enough as a container. If we have future layers, just add them as new files in the tarball. And they can always be optional, and individually-exportable, because of that tarballness.

ahyangyi:
> `Feature: Import real-world town data via JSON file`. Not sure 'real-world' is necessary or true here. It's perfectly fine to have non-real-world town data too.
> 
> As heightmaps can be used on a variety of map sizes, maybe the number of towns created can be scalable via `_settings_newgame.difficulty.number_towns` & `_settings_newgame.game_creation.custom_town_number` (the usual town scaling settings)? Might be something to think about later although such a choice might have a bearing on the json format.

Good point, some kind of "detail level" control seems useful. Allows us to utilize the same json file in multiple ways. The most naive solution would be:
* The game decides how many towns to place with some mysterious black box algorithm
* It sorts towns by population, and places the most populated towns from the json

One problem is that some towns might be more important than their population suggests.  "Washington DC and Las Vegas have small population, let's drop them" might just be wrong. A simple patch would be allow the json provider to specify an importance level, and default to using population when that's not specified.

PeterN:
I'm not sure:

* download town data from online content.
* navigate file browser to download location.
* un-tar the file.
* open town data json in text data (do spreadsheets support json?)
* remove lines

is simpler than:

* download town data from online content.
* obey existing in-game limit for towns when loading town data.

but... :)

Anyway, I was not suggesting to implement it now, just enquiring about considerations to be made. But as the data is json I suppose it's simple enough to just add whatever is needed later.

2TallTyler:
Fair enough. Excel can import JSON natively, Google Sheets needs a plugin. But as you say, let's leave that for later. 🙂

src/fios.cpp
```diff
@@ -554,6 +559,42 @@ void FiosGetHeightmapList(SaveLoadOperation fop, bool show_dirs, FileList &file_
...
+/**
+ * Callback for FiosGetTownDataList.
+ * @param fop Purpose of collecting the list.
+ * @param file Name of the file to check.
+ * @return a FIOS_TYPE_JSON type of the found file, FIOS_TYPE_INVALID if not a valid JSON file, and the title of the file (if any).
+ */
+static std::tuple<FiosType, std::string> FiosGetTownDataListCallback(SaveLoadOperation fop, const std::string &file, const std::string_view ext)
+{
+	if (fop == SLO_LOAD) {
+		if (StrEqualsIgnoreCase(ext, ".json")) {
```
PeterN, line 571: This seems very wide-open. What would happen if we added  some different metadata later that is also stored in a file ending .json?

src/fios_gui.cpp
```diff
@@ -168,6 +169,49 @@ static constexpr NWidgetPart _nested_load_heightmap_dialog_widgets[] = {
...
+	EndContainer(),
+	/* Current directory and free space */
+	NWidget(WWT_PANEL, COLOUR_GREY, WID_SL_BACKGROUND), SetFill(1, 0), SetResize(1, 0), EndContainer(),
+
+	/* Filter box with label */
+	NWidget(WWT_PANEL, COLOUR_GREY), SetFill(1, 1), SetResize(1, 1),
+		NWidget(NWID_HORIZONTAL), SetPadding(WidgetDimensions::unscaled.framerect.top, 0, WidgetDimensions::unscaled.framerect.bottom, 0),
+			SetPIP(WidgetDimensions::unscaled.frametext.left, WidgetDimensions::unscaled.frametext.right, 0),
+				NWidget(WWT_TEXT, COLOUR_GREY), SetFill(0, 1), SetDataTip(STR_SAVELOAD_FILTER_TITLE , STR_NULL),
+				NWidget(WWT_EDITBOX, COLOUR_GREY, WID_SL_FILTER), SetFill(1, 0), SetMinimalSize(50, 12), SetResize(1, 0),
```
PeterN: SetMinimalSize is not needed here :)

src/fios_gui.cpp
```diff
@@ -168,6 +169,49 @@ static constexpr NWidgetPart _nested_load_heightmap_dialog_widgets[] = {
...
+					SetDataTip(STR_LIST_FILTER_OSKTITLE, STR_LIST_FILTER_TOOLTIP),
+		EndContainer(),
+	EndContainer(),
+	/* Sort Buttons */
+	NWidget(NWID_HORIZONTAL),
+		NWidget(NWID_HORIZONTAL, NC_EQUALSIZE),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_SORT_BYNAME), SetDataTip(STR_SORT_BY_CAPTION_NAME, STR_TOOLTIP_SORT_ORDER), SetFill(1, 0), SetResize(1, 0),
+			NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_SL_SORT_BYDATE), SetDataTip(STR_SORT_BY_CAPTION_DATE, STR_TOOLTIP_SORT_ORDER), SetFill(1, 0), SetResize(1, 0),
+		EndContainer(),
+		NWidget(WWT_PUSHIMGBTN, COLOUR_GREY, WID_SL_HOME_BUTTON), SetMinimalSize(12, 12), SetDataTip(SPR_HOUSE_ICON, STR_SAVELOAD_HOME_BUTTON),
```
PeterN: Should be using `SetAspect()` now, see the other SPR_HOUSE_ICON widgets.

src/fios_gui.cpp
```diff
@@ -920,7 +983,16 @@ void ShowSaveLoadDialog(AbstractFileType abstract_filetype, SaveLoadOperation fo
...
+		switch (abstract_filetype) {
+			case FT_HEIGHTMAP:
+				sld = &_load_heightmap_dialog_desc;
+				break;
+			case FT_TOWN_DATA:
+				sld = &_load_town_data_dialog_desc;
+				break;
+			default:
+				sld = &_load_dialog_desc;
+			}
```
PeterN: Too much indentation on the closing }

src/genworld.cpp
```diff
@@ -337,3 +341,147 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	uint population;    ///< The target population of the town when created in OpenTTD. If input is blank, defaults to 0.
+	bool is_city;       ///< Should it be created as a city in OpenTTD? If input is blank, defaults to false.
+	float x_proportion; ///< The X coordinate of the town, as a proportion 0..1 of the maximum X coordinate.
+	float y_proportion; ///< The Y coordinate of the town, as a proportion 0..1 of the maximum Y coordinate.
+};
+
+/**
+ * Helper for CircularTileSearch to found a town on or near a given tile.
+ * @param tile The tile to try founding the town upon.
+ * @param user_data The ExternalTownData to attempt to found.
```
PeterN, line 358: Missing `@return`

src/genworld.cpp
```diff
@@ -337,3 +341,147 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		return;
+	}
+
+	/* Check for JSON formatting errors with the array of towns. */
+	if (!town_data.is_array()) {
+		ShowErrorMessage(STR_TOWN_DATA_ERROR_LOAD_FAILED, STR_TOWN_DATA_ERROR_JSON_FORMATTED_INCORRECTLY, WL_ERROR);
+		return;
+	}
+
+	std::list<std::pair<Town *, uint> > towns;
```
PeterN: `std::vector<std::pair<Town *, uint>> towns`?

src/genworld.cpp
```diff
@@ -337,3 +341,147 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		/* If we still fail to found the town, we'll create a sign at the intended location and tell the player how many towns we failed to create in an error message.
+		 * This allows the player to diagnose a heightmap misalignment, if towns end up in the sea, or place towns manually, if in rough terrain. */
+		if (!success) {
+			Command<CMD_PLACE_SIGN>::Post(tile, town.name);
+			failed_towns++;
+			continue;
+		}
+
+		Town *t = ClosestTownFromTile(search_tile, UINT_MAX);
+		towns.push_back(std::make_pair(t, town.population));
```
PeterN: `towns.emplace_back(t, town.population);`

PeterN: But also, CmdFoundTown does return the TownID, so it shouldn't be necessary to search for it.

src/genworld.cpp
```diff
@@ -337,3 +341,147 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+};
+
+/**
+ * Helper for CircularTileSearch to found a town on or near a given tile.
+ * @param tile The tile to try founding the town upon.
+ * @param user_data The ExternalTownData to attempt to found.
+ */
+static bool TryFoundTownNearby(TileIndex tile, void *user_data)
+{
+	ExternalTownData town = *(ExternalTownData *)user_data;
```
PeterN: `*static_cast<ExternalTownData *>(user_data);`

JGRennison: Presumably this should also be `const ExternalTownData &town =`?

PeterN: Good point, it's making a copy of it right now, whoops!

src/fios.cpp
2TallTyler: We could invent a file extension, at the cost of scenario creators needing to change file extension after exporting JSON from the spreadsheet used to generate the data. I did experiment with `.geojson` but that's actually a standardized format.

ahyangyi:
> Trying to override the data specified by the scenario creator within OpenTTD feels like a violation of KISS (Keep It Simple, Silly). In my opinion, we should let people filter the data themselves using a spreadsheet. If someone wishes to add some automatic filtering in the future, it is easy to add fields to the JSON spec.
> 
> As far as including additional layers -- loading a heightmap often requires some experimentation with max height. I would prefer to keep the load process using multiple steps instead of doing the entire (time-consuming) process with one click, so packing everything into a .tar and loading each layer manually would be my choice of workflow.

On the one hand, overriding the data specified by the json file feels like a violation of KISS. On the other hand, ignoring town density settings at setup also feels like a violation of KISS... So I guess it's a perspective issue.

As long as there is a way to use all towns from the json and nothing more, I'll be happy.

2TallTyler:
import_gui

The button to import town data is in the Towns menu, while the town density controls are in the landscape generation menu. Other nearby buttons, like random towns, don't respect the density/custom number limit, so I would not expect importing town data to have any connection to such a limit.

2TallTyler:
Rebased to fix conflicts.

src/fios_gui.cpp
```diff
@@ -168,6 +169,49 @@ static constexpr NWidgetPart _nested_load_heightmap_dialog_widgets[] = {
...
+
+	/* Filter box with label */
+	NWidget(WWT_PANEL, COLOUR_GREY), SetFill(1, 1), SetResize(1, 1),
+		NWidget(NWID_HORIZONTAL), SetPadding(WidgetDimensions::unscaled.framerect.top, 0, WidgetDimensions::unscaled.framerect.bottom, 0),
+			SetPIP(WidgetDimensions::unscaled.frametext.left, WidgetDimensions::unscaled.frametext.right, 0),
+				NWidget(WWT_TEXT, COLOUR_GREY), SetFill(0, 1), SetDataTip(STR_SAVELOAD_FILTER_TITLE , STR_NULL),
+				NWidget(WWT_EDITBOX, COLOUR_GREY, WID_SL_FILTER), SetFill(1, 0), SetResize(1, 0),
+					SetDataTip(STR_LIST_FILTER_OSKTITLE, STR_LIST_FILTER_TOOLTIP),
+		EndContainer(),
+	EndContainer(),
```
PeterN: Incorrect NWidgetPart indenting here

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	}
+
+	/* If we couldn't found a town (or multiple), display a message to the player with the number of failed towns. */
+	if (failed_towns > 0) {
+		SetDParam(0, failed_towns);
+		ShowErrorMessage(STR_TOWN_DATA_ERROR_FAILED_TO_FOUND_TOWN, INVALID_STRING_ID, WL_WARNING);
+	}
+
+	/* Now that we've created the towns, let's grow them to their target populations. */
+	for (auto &item : towns) {
```
LordAro: const?

src/fios.h
```diff
@@ -107,6 +107,7 @@ void ShowSaveLoadDialog(AbstractFileType abstract_filetype, SaveLoadOperation fo
 void FiosGetSavegameList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
 void FiosGetScenarioList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
 void FiosGetHeightmapList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
+void FiosGetTownDataList(SaveLoadOperation fop, bool show_dirs, FileList& file_list);
```
LordAro:
```suggestion
void FiosGetTownDataList(SaveLoadOperation fop, bool show_dirs, FileList &file_list);
```

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
 
 	_GenerateWorld();
 }
+
+/** Town data imported from JSON files and used to place towns. */
+struct ExternalTownData {
+	TownID town_id;     ///< The TownID of the town in OpenTTD. Not imported, but set during the founding proceess and stored here for convenience.
```
LordAro: new code style says don't align comments :p

src/fios.cpp
2TallTyler: Actually, I think you might be mistaken. This callback is only triggered when the user clicks "Load [town data] from file" -- so it does NOT reserve the file extension `.json` for town data. Unless I'm very mistaken, which is always possible. 😄

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		feature.at("name").get_to(town.name);
+		feature.at("population").get_to(town.population);
+		feature.at("city").get_to(town.is_city);
+
+		/* Set town coordinates. */
+		feature.at("x").get_to(town.x_proportion);
+		feature.at("y").get_to(town.y_proportion);
+
+		/* Check for improper coordinates and warn the player. */
+		if (town.x_proportion <= 0 || town.y_proportion <= 0 || town.x_proportion >= 1 || town.y_proportion >= 1) {
```
TrueBrain:
```suggestion
		if (town.x_proportion <= 0.0 || town.y_proportion <= 0.0 || town.x_proportion >= 1.0 || town.y_proportion >= 1.0) {
```

Don't confuse your reviewers :P

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+			SetDParamStr(0, town.name);
+			ShowErrorMessage(STR_TOWN_DATA_ERROR_LOAD_FAILED, STR_TOWN_DATA_ERROR_BAD_COORDINATE, WL_ERROR);
+			return;
```
TrueBrain, line 457: `return`? Not `continue`? Punish all towns because 1 is wrong? It can be a choice, ofc. Just checking if it is the right one :) Same with the `return` statements above.

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		SetDParam(0, failed_towns);
+		ShowErrorMessage(STR_TOWN_DATA_ERROR_FAILED_TO_FOUND_TOWN, INVALID_STRING_ID, WL_WARNING);
+	}
+
+	/* Now that we've created the towns, let's grow them to their target populations. */
+	for (const auto &item : towns) {
+		Town *t = item.first;
+		uint population = item.second;
+
+		uint tries = 500;
```
TrueBrain: I still don't understand this `500`. Population can now never grow over 500 growth-ticks. That seems weird / random to me :)

Maybe you want to check if there was no population growth for 500 ticks, and abort? Or at least comment what your thought it for this value.

Either way, the 500 for sure should be a constant, as magic numbers mid-way is just weird :)

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		/* Find the target tile for the town. */
+		TileIndex tile = TileXY(town.x_proportion * Map::MaxX(), town.y_proportion * Map::MaxY());
+
+		/* Try founding on the target tile, and if that doesn't work, find the nearest suitable tile up to 16 tiles away.
+		 * The target might be on water, blocked somehow, or on a steep slope that can't be terraformed by the founding command. */
+		TileIndex search_tile = tile;
+		bool success = CircularTileSearch(&search_tile, 16, 0, 0, TryFoundTownNearby, &town);
+
+		/* If we still fail to found the town, we'll create a sign at the intended location and tell the player how many towns we failed to create in an error message.
+		 * This allows the player to diagnose a heightmap misalignment, if towns end up in the sea, or place towns manually, if in rough terrain. */
```
TrueBrain, line 480: Shouldn't this only happen in "development mode" or something? Dunno, just asking :) I am fine with this, to be clear.

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	/* The command succeeded, send the ID back through user_data. */
+	town.town_id = id;
+	return true;
+}
+
+/**
+ * Load town data from _file_to_saveload and place its contents into the world.
+ * @return True if the JSON file is valid and has north and south extents, otherwise false.
+ */
+void LoadTownData()
```
TrueBrain, line 379: This function does more than it suggests. It doesn't only load, but it also validates, tries to build them, and grows them.

Normally I would suggest to split up the function a bit more, as it is really long (and a bit hard to follow). But as you have `returns` in your for-loop, that is tricky. Might still be worth looking into it, see if you can split it up a bit.

If not, no harm done. Just checking.

src/genworld.cpp
```diff
@@ -337,3 +341,157 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+	if (id == INVALID_TOWN) return false;
+
+	/* The command succeeded, send the ID back through user_data. */
+	town.town_id = id;
+	return true;
+}
+
+/**
+ * Load town data from _file_to_saveload and place its contents into the world.
+ * @return True if the JSON file is valid and has north and south extents, otherwise false.
```
TrueBrain: Nitpick: if you write `True if ..., otherwise false`, just write: `True iff ...`. Saves the reader some time not having to read the `otherwise false` non-sense :P :P

TrueBrain: What does `north and south extents` mean? I can't find any clue into what that references :)

2TallTyler: Oops! Somehow this comment survived from several iterations ago. It is totally irrelevant now.

src/genworld.cpp
2TallTyler: It used to be split up, but I think it makes more sense combined. I will change the function doxygen to mention everything it does.

src/genworld.cpp
2TallTyler: No, this is for the scenario author to diagnose problems. A town can fail to found if it can't find a flat tile (heightmap too hilly), or if the author gets the rotation wrong and a town ends up in the ocean.

src/genworld.cpp
2TallTyler: Yes, founding and growing towns takes a significant amount of time, so I am intentionally bailing out early. I expect the scenario author would not want to continue if their data was wrong, so this saves them time.

src/genworld.cpp
```diff
@@ -337,3 +341,167 @@ void GenerateWorld(GenWorldMode mode, uint size_x, uint size_y, bool reset_setti
...
+		/* Grid towns can grow almost forever, but the town growth algorithm gets less and less efficient as it wanders roads randomly,
+		 * so we set an arbitrary limit. With a flat map and a 3x3 grid layout this results in about 4900 houses, or 2800 houses with "Better roads." */
+		int try_limit = 1000;
+
+		/* If a town repeatedly fails to grow, continuing to try only wastes time. */
+		int fail_limit = 10;
+
+		/* Grow by a constant number of houses each time, instead of growth based on current town size.
+		 * We want our try limit to apply in a predictable way, no matter the road layout and other geography. */
+		const int HOUSES_TO_GROW = 10;
```
TrueBrain, line 510: One final nitpick: should we move these consts up like other files, and make `1000` and `10` also consts (next to these variables, ofc), so we get `HOUSE_GROW_TRY_LIMIT` and `HOUSE_GROW_FAIL_LIMIT`?

Btw, excellent comment around `try_limit`, exactly addresses my concern/issue with it :)

TrueBrain:
@2TallTyler btw, did anyone besides you tested this, if it works as they would expect and that they can do what they want with it? Not a blocker, more wondering out loud.

src/genworld.cpp
2TallTyler: I tried that, and it was messy. These only apply to this function and aren't shared by anything else in the file, so I personally like them as specific as possible. 

Since you phrased your remark as a question, my answer is "In my opinion, no." If your remark was actually a request, I will happily move them for you. 😄

TrueBrain: Nope. It was an actual question :) Maybe another dev has another opinion, but as long as you had a reason for it, I am pretty much okay with this :)

(and I guess they have till we merge this to speak up :P )

2TallTyler:
If anyone wants to try this out, they can get a [pre-compiled build here](https://cdn.openttd.org/openttd-branches/pr10409/20240626-pr10409-geba7366775/) (link also in PR description). I've posted an announcement on Discord, so I hope we get some good feedback. 😄

PimvdOetelaar:
> If anyone wants to try this out, they can get a [pre-compiled build here](https://cdn.openttd.org/openttd-branches/pr10409/20240626-pr10409-geba7366775/) (link also in PR description). I've posted an announcement on Discord, so I hope we get some good feedback. 😄

Tested:
- on my system, folder towndata is not automatically created. I have the regular folder <root>\OpenTTD (ai, baseset, etc), then the actual game is in <root>\OpenTTD_PR10409. In case that has to do with anything.
- Creating heightmap and adding towns goes smoothly and quickly in the scenario. 

I'm now going to give putting together an own heightmap and town list and see if this can be done relatively quickly. The method from the tutorial is tedious on large maps and I want to see if getting a town list can be at least partly automated.

2TallTyler:
Thanks for the feedback. I've actually moved away from a separate `towndata` folder to instead keep everything in `heightmap`, but missed updating one mention of this in the PR description. 😄 

Yes, the pixel coordinate method in the tutorial is tedious — although it is faster than manually placing towns in Scenario Editor. I am hoping that this PR is an incremental improvement which can be improved further by someone developing an external tool to create the data in a more automated way. 🙂

MasterOktagon:
When rotating the heightmap (in the example to the left) the town data is not, resulting in wrongly placed cities and cities placed in the water (=only signs).

2TallTyler:
Thanks for the feedback. The clockwise direction is "correct" with respect to the coordinates, but I've made it possible to use counter-clockwise heightmaps by translating the town data if the heightmap was loaded in a counter-clockwise orientation. I have updated the documentation accordingly.

Note that the pre-compiled build linked above has not been updated with this change.

pvillaverde:
Hey there! I saw this a while ago but I had already worked on a base scenario doing the rivers and adding each town, village or settlement one by one. Would it be possible (some day) to export the cities? And ideally the rivers too, because heightmaps do not create them (Real rivers I mean)

2TallTyler:
Exporting cities to a file in this format would be trivially easy. I am not including that feature in this PR to keep it a manageable size -- although I have no plans to add the feature myself.

Rivers are significantly harder, because if the map isn't perfectly set up with all river elevation changes on a valid river slope, you either get disconnected rivers, or have to terraform when placing rivers. When I added wide rivers, I saw just how tricky it can be to terraform rivers properly, without undermining other rivers, etc. It's doable, but not trivial.

TrueBrain (approved):

PR #10381, opened 2023-01-21
Change: allow GS to mutate vehicle orders (previously AI-only)

andythenorth:
## Motivation / Problem

* For features or gameplay changes:
    * What was the motivation to develop this feature?
        * GS can't manipulate vehicle orders. 
    * Does this address any problem with the gameplay or interface?
        * It addresses that GS can't manipulate vehicle orders.
    * Which group of players do you think would enjoy this feature?
        * This feature is about extending the GS API to be more comprehensive, and not trying to anticipate in advance if GS authors can be trusted to use it wisely :)

## Description

* For features or gameplay changes:
    * What does this feature do?
        *  Makes existing AI control over vehicle orders available to GS 

## Limitations

Describe here
* Is the problem solved in all scenarios?
   * Not sure, I just removed the -game flag from doxygen comments
* Is this feature complete? Are there things that could be added in the future?
   * Not sure, I just removed the -game flag from doxygen comments
* Are there things that are intentionally left out?
   * Not sure, I just removed the -game flag from doxygen comments
* Do you know of a bug or corner case that does not work?
   * Not sure, I just removed the -game flag from doxygen comments
   * Works in my simplistic test script
   * Might need pre-condition enforced for company etc, not sure?
## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
    * Nope
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
    * Nope 
* This PR affects the save game format? (label 'savegame upgrade')
    * Not AFAIK
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
        * This will need doing, I will push again when I've figured out how to do that
    * The compatibility wrappers (compat_*.nut) need updating.
        * Not sure, hoping a reviewer might tell me :)
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * Nope

PR #10411, opened 2023-01-25
Expose more functions to game scripts

SamuXarick:
## Motivation / Problem
1. A few functions could work without needing the GS to be in CompanyMode. These have been changed to allow it:
- GSGroup::IsValidGroup
- GSGroup::GetNumEngines
- GSGroup::GetNumVehicles
- GSGroup::GetEngineReplacement
- GSGroupList
- GSVehicleList_Group
- GSVehicleList_DefaultGroup

2. One special method is added and only exposed for Game Scripts to complement for the changes above:
- GSGroup::GetOwner

3. Expose more events to Game Scripts. They and all other events have been templated as requested:
- GSEventEnginePreview
- GSEventCompanyAskMerger
- GSEventVehicleLost
- GSEventVehicleWaitingInDepot
- GSEventVehicleUnprofitable
- GSEventEngineAvailable
- GSEventDisasterZeppelinerCrashed
- GSEventDisasterZeppelinerCleared
- GSEventAircraftDestTooFar
- GSEventVehicleAutoReplaced

## Description
- `ScriptGroup::IsValidGroup`, `ScriptGroupList`, `ScriptVehicleList_Group` and `ScriptVehicleList_DefaultGroup` lets GS to get groups of any company without being in the company.

- `ScriptGroup::GetOwner` didn't exist before. It was added and is only accessible for the GS.

## Limitations
- `ScriptGroup::GetNumEngines`, `ScriptGroup::GetNumVehicles` and `ScriptGroup::GetEngineReplacement` have extra conditions to make it less restrictive to use when in deity mode. If the group exists and is not `GROUP_DEFAULT` and not `GROUP_ALL`, then the company of that group could be retrieved from the group->owner itself, without the need to be in company scope, but since those functions can also be supplied with `GROUP_DEFAULT` and `GROUP_ALL` as parameters, you need to be in a valid company scope mode for those.

- `ScriptEventEngineAvailable` may happen to not be broadcasted for AIs, only to GS, because a GS can also deal with non-AI companies.

## Checklist for review

### DIFF
--- a/src/autoreplace_cmd.cpp
+++ b/src/autoreplace_cmd.cpp
@@ -20,7 +20,7 @@
 #include "core/random_func.hpp"
 #include "vehiclelist.h"
 #include "road.h"
-#include "ai/ai.hpp"
+#include "script/script_trigger.hpp"
 #include "news_func.h"
 #include "strings_func.h"
 #include "autoreplace_cmd.h"
@@ -481,7 +481,7 @@ static CommandCost ReplaceFreeUnit(Vehicle **single_unit, DoCommandFlags flags, 
 
 			*single_unit = new_v;
 
-			AI::NewEvent(old_v->owner, new ScriptEventVehicleAutoReplaced(old_v->index, new_v->index));
+			ScriptTrigger::NewEvent<ScriptEventVehicleAutoReplaced>(old_v->owner, old_v->index, new_v->index);
 		}
 
 		/* Sell the old vehicle */
@@ -640,7 +640,7 @@ static CommandCost ReplaceChain(Vehicle **chain, DoCommandFlags flags, bool wago
 				/* Success ! */
 				if (flags.Test(DoCommandFlag::Execute) && new_head != old_head) {
 					*chain = new_head;
-					AI::NewEvent(old_head->owner, new ScriptEventVehicleAutoReplaced(old_head->index, new_head->index));
+					ScriptTrigger::NewEvent<ScriptEventVehicleAutoReplaced>(old_head->owner, old_head->index, new_head->index);
 				}
 
 				/* Transfer cargo of old vehicles and sell them */
@@ -709,7 +709,7 @@ static CommandCost ReplaceChain(Vehicle **chain, DoCommandFlags flags, bool wago
 					TransferCargo(old_head, new_head, true);
 					*chain = new_head;
 
-					AI::NewEvent(old_head->owner, new ScriptEventVehicleAutoReplaced(old_head->index, new_head->index));
+					ScriptTrigger::NewEvent<ScriptEventVehicleAutoReplaced>(old_head->owner, old_head->index, new_head->index);
 				}
 
 				/* Sell the old vehicle */
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -21,6 +21,7 @@
 #include "autoreplace_gui.h"
 #include "string_func.h"
 #include "ai/ai.hpp"
+#include "game/game.hpp"
 #include "core/pool_func.hpp"
 #include "engine_gui.h"
 #include "engine_func.h"
@@ -1126,6 +1127,9 @@ static void NewVehicleAvailable(Engine *e)
 
 	/* Only broadcast event if AIs are able to build this vehicle type. */
 	if (!IsVehicleTypeDisabled(e->type, true)) AI::BroadcastNewEvent(new ScriptEventEngineAvailable(index));
+
+	/* Only send the event to Game Script if engine can be built. */
+	if (!IsVehicleTypeDisabled(e->type, false)) Game::NewEvent(new ScriptEventEngineAvailable(index));
 
 	/* Only provide the "New Vehicle available" news paper entry, if engine can be built. */
 	if (!IsVehicleTypeDisabled(e->type, false) && !e->info.extra_flags.Test(ExtraEngineFlag::NoNews)) {
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -20,8 +20,7 @@
 #include "window_func.h"
 #include "autoreplace_gui.h"
 #include "string_func.h"
-#include "ai/ai.hpp"
-#include "game/game.hpp"
+#include "script/script_trigger.hpp"
 #include "core/pool_func.hpp"
 #include "engine_gui.h"
 #include "engine_func.h"
@@ -992,7 +991,7 @@ static IntervalTimer<TimerGameCalendar> _calendar_engines_daily({TimerGameCalend
 				 * boost that they wouldn't have gotten against other human companies. The check on
 				 * the line below is just to make AIs not notice that they have a preview if they
 				 * cannot build the vehicle. */
-				if (!IsVehicleTypeDisabled(e->type, true)) AI::NewEvent(e->preview_company, new ScriptEventEnginePreview(i));
+				if (!IsVehicleTypeDisabled(e->type, true)) ScriptTrigger::NewEvent<ScriptEventEnginePreview>(e->preview_company, e->preview_company, i);
 				if (IsInteractiveCompany(e->preview_company)) ShowEnginePreviewWindow(i);
 			}
 		}
--- a/src/script/api/script_event_types.hpp
+++ b/src/script/api/script_event_types.hpp
@@ -439,17 +439,19 @@ private:
 
 /**
  * Event Company Ask Merger, indicating a company can be bought (cheaply) by you.
- * @api ai
+ * @api ai game
  */
 class ScriptEventCompanyAskMerger : public ScriptEvent {
 public:
 #ifndef DOXYGEN_API
 	/**
+	 * @param buyer The company that can buy.
 	 * @param owner The company that can be bought.
 	 * @param value The value/costs of buying the company.
 	 */
-	ScriptEventCompanyAskMerger(Owner owner, Money value) :
+	ScriptEventCompanyAskMerger(Owner buyer, Owner owner, Money value) :
 		ScriptEvent(ET_COMPANY_ASK_MERGER),
+		buyer(ScriptCompany::ToScriptCompanyID(buyer)),
 		owner(ScriptCompany::ToScriptCompanyID(owner)),
 		value(value)
 	{}
@@ -461,6 +463,14 @@ public:
 	 * @return The converted instance.
 	 */
 	static ScriptEventCompanyAskMerger *Convert(ScriptEvent *instance) { return (ScriptEventCompanyAskMerger *)instance; }
+
+	/**
+	 * Get the CompanyID of the company that can purchase the other.
+	 * @return The CompanyID of the company that can purchase.
+	 * @note If the company is bought this will become invalid.
+	 * @api -ai
+	 */
+	ScriptCompany::CompanyID GetBuyerID() { return this->buyer; }
 
 	/**
 	 * Get the CompanyID of the company that can be bought.
@@ -483,8 +493,9 @@ public:
 	bool AcceptMerger();
 
 private:
+	ScriptCompany::CompanyID buyer; ///< The company that is buying.
 	ScriptCompany::CompanyID owner; ///< The company that is in trouble.
-	Money value;                ///< The value of the company, i.e. the amount you would pay.
+	Money value;                ///< The value of the company in trouble, i.e. the amount the buyer would pay.
 };
 
 /**
--- a/src/script/api/script_event_types.hpp
+++ b/src/script/api/script_event_types.hpp
@@ -568,7 +568,7 @@ private:
 
 /**
  * Event Vehicle Lost, indicating a vehicle can't find its way to its destination.
- * @api ai
+ * @api ai game
  */
 class ScriptEventVehicleLost : public ScriptEvent {
 public:
@@ -601,7 +601,7 @@ private:
 
 /**
  * Event VehicleWaitingInDepot, indicating a vehicle has arrived a depot and is now waiting there.
- * @api ai
+ * @api ai game
  */
 class ScriptEventVehicleWaitingInDepot : public ScriptEvent {
 public:
@@ -634,7 +634,7 @@ private:
 
 /**
  * Event Vehicle Unprofitable, indicating a vehicle lost money last year.
- * @api ai
+ * @api ai game
  */
 class ScriptEventVehicleUnprofitable : public ScriptEvent {
 public:
@@ -733,7 +733,7 @@ private:
 
 /**
  * Event Engine Available, indicating a new engine is available.
- * @api ai
+ * @api ai game
  */
 class ScriptEventEngineAvailable : public ScriptEvent {
 public:
@@ -808,7 +808,7 @@ private:
 
 /**
  * Event Disaster Zeppeliner Crashed, indicating a zeppeliner has crashed on an airport and is blocking the runway.
- * @api ai
+ * @api ai game
  */
 class ScriptEventDisasterZeppelinerCrashed : public ScriptEvent {
 public:
@@ -841,7 +841,7 @@ private:
 
 /**
  * Event Disaster Zeppeliner Cleared, indicating a previously crashed zeppeliner has been removed, and the airport is operating again.
- * @api ai
+ * @api ai game
  */
 class ScriptEventDisasterZeppelinerCleared : public ScriptEvent {
 public:
@@ -909,7 +909,7 @@ private:
  * Event AircraftDestTooFar, indicating the next destination of an aircraft is too far away.
  * This event can be triggered when the current order of an aircraft changes, usually either when
  * loading is done or when switched manually.
- * @api ai
+ * @api ai game
  */
 class ScriptEventAircraftDestTooFar : public ScriptEvent {
 public:
@@ -1177,7 +1177,7 @@ public:
 
 /**
  * Event VehicleAutoReplaced, indicating a vehicle has been auto replaced.
- * @api ai
+ * @api ai game
  */
 class ScriptEventVehicleAutoReplaced : public ScriptEvent {
 public:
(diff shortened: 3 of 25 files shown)


SamuXarick:
Notice these logged issues:
https://github.com/OpenTTD/OpenTTD/actions/runs/4007525928/jobs/6880479353#step:9:213
https://github.com/OpenTTD/OpenTTD/actions/runs/4007525928/jobs/6880479353#step:9:381
https://github.com/OpenTTD/OpenTTD/actions/runs/4007525928/jobs/6880479353#step:9:464

I don't know CMake, but on IRC, @glx22 mentioned something about it:
> <glx[d]> Might work, but needs deep look in how the squirrel binding is done, and probably update of the cmake script
> <glx[d]> Overloading constructors was never done before for the API

My goal was to show the right constructor to the respective doc page. AI gets shown one, GS gets shown the other.

2TallTyler:
Is this ready for review? Andy asked for #10381 to be merged, but it's missing the changelog and company mode preconditions that you added here, so it's better to merge this and close Andy's.

SamuXarick:
> Is this ready for review? Andy asked for #10381 to be merged, but it's missing the changelog and company mode preconditions that you added here, so it's better to merge this and close Andy's.

It kind of is, I'm marking it ready for review. What I'd still want to do was to make some group functions precondition less restrictive when in deity mode, those that only query some information detail from a group.
If the group exists and is not GROUP_DEFAULT and not GROUP_ALL, then the company of that group could be retrieved from the group->owner itself, without the need to be in company scope, but since those functions can also be supplied with GROUP_DEFAULT and GROUP_ALL as parameters, they need to be in company scope for those.

src/script/api/script_company.cpp
```diff
@@ -40,11 +40,14 @@
 
 /* static */ bool ScriptCompany::IsMine(ScriptCompany::CompanyID company)
 {
+	if (ScriptObject::GetCompany() == OWNER_DEITY) return false;
```
rubidium42: I would use the `EnforcePrecondition` and add the comment that the GS needs to be in a `ScriptCompanyMode`. That way all the functions are consistent, after all it also makes no sense to call this when not in a `ScriptCompanyMode`.

src/script/api/script_group.cpp
```diff
@@ -26,11 +26,18 @@
 /* static */ bool ScriptGroup::IsValidGroup(GroupID group_id)
 {
 	const Group *g = ::Group::GetIfValid(group_id);
-	return g != nullptr && g->owner == ScriptObject::GetCompany();
+	return g != nullptr && (g->owner == ScriptObject::GetCompany() || ScriptObject::GetCompany() == OWNER_DEITY);
+}
+
+/* static */ ScriptCompany::CompanyID ScriptGroup::GetOwner(GroupID group_id) {
```
rubidium42: `{` on new line

src/script/api/script_group.cpp
```diff
@@ -26,11 +26,18 @@
...
+	if (!IsValidGroup(group_id)) return ScriptCompany::COMPANY_INVALID;
+
+	return static_cast<ScriptCompany::CompanyID>((int)::Group::Get(group_id)->owner);
```
rubidium42: The cast via `(int)` seems weird. Is the `(int)` really needed? Because then why not use just `(ScriptCompany::CompanyID)`?

src/script/api/script_group.cpp
```diff
@@ -148,13 +166,15 @@
 
 /* static */ EngineID ScriptGroup::GetEngineReplacement(GroupID group_id, EngineID engine_id)
 {
+	if (ScriptObject::GetCompany() == OWNER_DEITY) return ::INVALID_ENGINE;
```
rubidium42: Doesn't `EnforcePrecondition(::INVALID_ENGINE, ScriptObject::GetCompany() == OWNER_DEITY)` work?

src/script/api/script_group.cpp
```diff
@@ -128,18 +143,21 @@
...
 {
+	EnforcePrecondition(false, ScriptObject::GetCompany() != OWNER_DEITY);
 	if (HasWagonRemoval() == enable_removal) return true;
 
 	return ScriptObject::Command<CMD_CHANGE_COMPANY_SETTING>::Do("company.renew_keep_length", enable_removal ? 1 : 0);
 }
 
 /* static */ bool ScriptGroup::HasWagonRemoval()
 {
+	if (ScriptObject::GetCompany() == OWNER_DEITY) return false;
```
rubidium42: Why no `EnforcePrecondition`?

src/script/api/script_group.cpp
```diff
@@ -104,13 +116,15 @@
 
 /* static */ int32 ScriptGroup::GetNumEngines(GroupID group_id, EngineID engine_id)
 {
+	if (ScriptObject::GetCompany() == OWNER_DEITY) return -1;
```
rubidium42: Doesn't `EnforcePrecondition(-1, ScriptObject::GetCompany() == OWNER_DEITY)` work?

src/script/api/script_group.cpp
```diff
@@ -104,13 +116,15 @@
...
 	if (!IsValidGroup(group_id) && group_id != GROUP_DEFAULT && group_id != GROUP_ALL) return -1;
 
 	return GetGroupNumEngines(ScriptObject::GetCompany(), group_id, engine_id);
 }
 
 /* static */ int32 ScriptGroup::GetNumVehicles(GroupID group_id, ScriptVehicle::VehicleType vehicle_type)
 {
+	if (ScriptObject::GetCompany() == OWNER_DEITY) return -1;
```
rubidium42: Doesn't `EnforcePrecondition(-1, ScriptObject::GetCompany() == OWNER_DEITY)` work?

src/script/api/script_event_types.hpp
```diff
@@ -381,8 +392,9 @@ class ScriptEventCompanyAskMerger : public ScriptEvent {
 	 * @param owner The company that can be bought.
 	 * @param value The value/costs of buying the company.
 	 */
-	ScriptEventCompanyAskMerger(Owner owner, int32 value) :
+	ScriptEventCompanyAskMerger(Owner buyer, Owner owner, int32 value) :
```
rubidium42: No necessarily this PR, but it seems silly to have an `int32` value instead of a `Money` value. Otherwise the AI will potentially get a lower value presented and accepts it, without knowing how expensive it actually is.

src/autoreplace_cmd.cpp
```diff
@@ -470,6 +471,7 @@ static CommandCost ReplaceFreeUnit(Vehicle **single_unit, DoCommandFlag flags, b
 			*single_unit = new_v;
 
 			AI::NewEvent(old_v->owner, new ScriptEventVehicleAutoReplaced(old_v->index, new_v->index));
+			Game::NewEvent(new ScriptEventVehicleAutoReplaced(old_v->index, new_v->index));
```
rubidium42: I'd almost template creating a new event in a manner that creates two events to keep the practical code duplication relatively low.

That would be something like:
```
class Script {
public:
    template <class EventClass, typename ... Args> 
    static void NewEvent(int owner, Args... args) {
        AI::NewEvent(owner, new EventClass(args...));
        Game::NewEvent(new EventClass(args...));
    }
};
```
and then `Script::NewEvent<ScriptEventVehicleAutoReplaced>(old_v->owner, old_v->index, new_v->index);`. Yes, when a company needs to be passed to the constructor you pass the owner twice in a row, or you make a second templated function `Script::NewEventWithOwner` or something that under water passes the owner twice to `Script::NewEvent`.

I'm not sure whether either way with the templating is the best solution though.

rubidium42, line 474: I'd almost template creating a new event in a manner that creates two events to keep the practical code duplication relatively low.

That would be something like:
```
class Script {
public:
    template <class EventClass, typename ... Args> 
    static void NewEvent(int owner, Args... args) {
        AI::NewEvent(owner, new EventClass(args...));
        Game::NewEvent(new EventClass(args...));
    }
};
```
and then `Script::NewEvent<ScriptEventVehicleAutoReplaced>(old_v->owner, old_v->index, new_v->index);`. Yes, when a company needs to be passed to the constructor you pass the owner twice in a row, or you make a second templated function `Script::NewEventWithOwner` or something that under water passes the owner twice to `Script::NewEvent`.

I'm not sure whether either way with the templating is the best solution though.

src/script/api/script_company.cpp
```diff
@@ -308,11 +317,13 @@
 
 /* static */ bool ScriptCompany::SetPrimaryLiveryColour(LiveryScheme scheme, Colours colour)
 {
+	EnforcePrecondition(false, ScriptObject::GetCompany() != OWNER_DEITY);
 	return ScriptObject::Command<CMD_SET_COMPANY_COLOUR>::Do((::LiveryScheme)scheme, true, (::Colours)colour);
 }
 
 /* static */ bool ScriptCompany::SetSecondaryLiveryColour(LiveryScheme scheme, Colours colour)
 {
+	EnforcePrecondition(false, ScriptObject::GetCompany() != OWNER_DEITY);
```
rubidium42: Maybe make a `ScriptObject::IsScriptCompanyScopeActive()` function? That would make the `EnforcePrecondition`s map better to the `@pre` comments. Though, since it already exists in other places in the code, it might not be wise to add that to this PR.

src/script/api/script_group.cpp
SamuXarick: I merely copied how it was done on ScriptVehicle::GetOwner. It looks like that, too. Git blame didn't show anything suspicious about it. The first time it appeared was already like this.

src/script/api/script_group.cpp
SamuXarick: I tried to understand why some places use EnforcePrecondition and others don't. I thought it was probably to do with DoCommands being run. In the majority of cases, if it runs a DoCommand, then it uses EnforcePreconditon, and if it doesn't then it's done without EnforcePrecondition.

src/script/api/script_group.cpp
SamuXarick: done

src/script/api/script_company.cpp
SamuXarick: I added the comment that the GS needs to be in a valid Company. However, I'm unsure about using the EnforcePrecondition.

src/script/api/script_event_types.hpp
SamuXarick: A new PR was open for this.

src/autoreplace_cmd.cpp
SamuXarick: I created a new commit at the end, adding 2 templates. I didn't know where to put this, so I created a new file just for that. Not sure if the file name is fine, or if this could be put in some existing file. Please let me know.

I have however, been unable to do it properly for the case of AI::BroadcastNewEvent, which has a default parameter.
https://github.com/OpenTTD/OpenTTD/blob/59251d3c6b9bf2a0ae85aefc2c9e9f91a46f3031/src/ai/ai.hpp#L117-L120

```
	/**
	 * Broadcast a new event to all active AIs, and to the Game Script.
	 * @param skip_company The company to skip broadcasting for, or MAX_COMPANIES to skip none.
	 */
	template <class ScriptEventType, typename ... Args>
	static void BroadcastNewEvent(CompanyID skip_company, Args ... args) {
		AI::BroadcastNewEvent(new ScriptEventType(args...), skip_company);
		Game::NewEvent(new ScriptEventType(args...));
	}
```

When there is a `skip_company`, such in the case of a new company, I pass `Script::BroadcastNewEvent<ScriptEventCompanyNew>(c->index, c->index);`, but in the case I want to broadcast to all companies, I need to pass `MAX_COMPANIES`, which is the majority of the situations. One such example is `Script::BroadcastNewEvent<ScriptEventCompanyInTrouble>(MAX_COMPANIES, c->index);`
Ideally I wouldn't have to use `MAX_COMPANIES` everywhere, but I don't know how to do it in the template.

src/script/api/script_company.hpp
```diff
@@ -129,8 +129,8 @@ class ScriptCompany : public ScriptObject {
 	/**
 	 * Check if a CompanyID is your CompanyID, to ease up checks.
 	 * @param company The company index to check.
+	 * @game @pre ScriptCompanyMode::IsValid().
```
rubidium42: If it's a precondition, then use `EnforceCompanyModeValid(false);` in the actual implementation.

src/script/api/script_group.cpp
rubidium42: That it exists somewhere else does not necessarily mean that it is correct. Just to prove a point, yesterday I got some review comments on code that I copied and I amended my code and the code where I copied it from.

src/script/api/script_group.cpp
```diff
@@ -110,20 +117,27 @@
 
 /* static */ SQInteger ScriptGroup::GetNumEngines(GroupID group_id, EngineID engine_id)
 {
-	EnforceCompanyModeValid(-1);
-	if (!IsValidGroup(group_id) && group_id != GROUP_DEFAULT && group_id != GROUP_ALL) return -1;
+	EnforceDeityOrCompanyModeValid(-1);
```
rubidium42, line 121: What is the rationale for all this complicated logic, over just requiring company mode for all ScriptGroup calls?
Why is it needed that a GS can iterate over almost, but not all, groups?

Especially because to reach all functionality of this function, I still need to switch to company mode. I really do not see a reason for a GS to iterate (almost) all the groups, I actually do not see much reason for a GS to access groups but I can agree that if adding it takes only five characters there is little harm. However, making the API complicated for something that will barely get used seems overkill.

src/script/script.hpp
```diff
@@ -0,0 +1,42 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file script.hpp Base functions for all Scripts. */
```
rubidium42: This comment is misleading, it's not some base fo all scripts but rather very narrow for event triggering of scripts.
I'd almost call it script_event.hpp, but that's also already in the api folder, and even though technically not a problem it doesn't seem like something worth in the long run.

Maybe 'script_trigger.hpp', and then 'Functionality to trigger events in AI and game scripts' as comment? Then I'd also rename Script to ScriptTrigger. It's not as concise, but it definitely is much less misleading.

src/script/script.hpp
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Broadcast a new event to all active AIs, and to the Game Script.
+	 * @param skip_company The company to skip broadcasting for, or MAX_COMPANIES to skip none.
+	 */
+	template <class ScriptEventType, typename ... Args>
+	static void BroadcastNewEvent(CompanyID skip_company, Args ... args) {
+		AI::BroadcastNewEvent(new ScriptEventType(args...), skip_company);
+		Game::NewEvent(new ScriptEventType(args...));
+	}
+};
```
rubidium42: I really like these new helpers. It would be nice to introduce them first (including changing the existing ones), and then as you add the different events change those calls from AI::NewEvent to the new helper. Now the order is a bit weird, since you're adding Game::NewEvent calls to replace them a few commits later.

src/autoreplace_cmd.cpp
rubidium42: Maybe introduce a `BroadcastNewEventExceptForCompany` which is exactly like `BroadcastNewEvent` right now, and remove `skip_company` from the current `BroadcastNewEvent`? It is an extra function, but much clearer in the code.

src/script/api/script_group.cpp
SamuXarick: The rationale, if it is querying information without modifying anything in the company, then it should be done without being in the company. Too bad that for this and 2 other functions to retrieve information for GROUP_DEFAULT and GROUP_ALL, a specific company must be provided. The only way for that is to EnforceCompanyModeValid. The other groups are company owned, which makes it possible to query this information without being in the company.

rubidium42: You say that if you don't modify the company you should not need to be in the company, but then still require the GS to be in the company to get the complete picture.

And to be advocate's devil: why shouldn't this function, in deity mode, not return the total number of engines of all companies in the default or all group? Similarly, shouldn't ScriptVehicleList_DefaultGroup contain all vehicles that are not in a group if you're deity? After all, in deity mode basically all the other lists are the combined result of all companies. 

That way you do not need to construct complicated preconditions, or add new precondition check functions. And the GS does not need to iterate through all the companies if it wants the combined information.

src/script/script.hpp
SamuXarick: I renamed some things, the file, the Class name, and comment.

src/script/api/script_company.hpp
SamuXarick: done

src/script/api/script_group.cpp
SamuXarick: Done for this case, not for the other GetOwner.

src/engine.cpp
```diff
@@ -911,7 +910,7 @@ void EnginesDailyLoop()
 				 * boost that they wouldn't have gotten against other human companies. The check on
 				 * the line below is just to make AIs not notice that they have a preview if they
 				 * cannot build the vehicle. */
-				if (!IsVehicleTypeDisabled(e->type, true)) AI::NewEvent(e->preview_company, new ScriptEventEnginePreview(i));
+				if (!IsVehicleTypeDisabled(e->type, true)) if (!IsVehicleTypeDisabled(e->type, true)) ScriptTrigger::NewEvent<ScriptEventEnginePreview>(e->preview_company, e->preview_company, i);
```
rubidium42: Double ifs

SamuXarick: fixed

PR #10414, opened 2023-01-26
Codechange: use RAII to automatically restore _cur_dpi after use

rubidium42:
## Motivation / Problem

The code is "full" of 
```
DrawPixelInfo *old_pi = _cur_dpi; 
_cur_dpi = &tmp_dpi; 
<code>; 
_cur_dpi = old_dpi;
```
constructs on which CodeQL complains that an address of a stack variable is saved in a global variable. Or rather a local variable's address is stored in non-local memory.

## Description

Introduce a simpler variant of `Backup` that does nothing more than store the old value and restore it when the object goes out of scope, in other words use the RAII paradigm also used for things like `lock_guard`.
Then introduce this in all places where swaps between `old_dpi` and `_cur_dpi` are made.

## Limitations

This introduces one new warning, though solves 23. Technically I would classify that new warning as a non-issue (maybe even a false positive), as the `AutoRestoreBackup` will automatically reset `_cur_dpi` in this context before the reference that has been passed in as new value goes out of scope.

## Checklist for review

### DIFF
--- a/src/core/backup_type.hpp
+++ b/src/core/backup_type.hpp
@@ -144,4 +144,46 @@ private:
 	const int line;
 };
 
+/**
+ * Class to backup a specific variable and restore it upon destruction of this object to prevent
+ * stack values going out of scope before resetting the global to its original value. Contrary to
+ * #Backup this restores the variable automatically and there is no manual option to restore.
+ */
+template <typename T>
+struct AutoRestoreBackup {
+	/*
+	 * There is explicitly no only original constructor version, as that would make it possible
+	 * for the new value to go out of scope before this object goes out of scope, thus defeating
+	 * the whole goal and reason for existing of this object.
+	 */
+
+	/**
+	 * Backup variable and switch to new value.
+	 * @param original Variable to backup.
+	 * @param new_value New value for variable.
+	 */
+	AutoRestoreBackup(T &original, T new_value) : original(original), original_value(original)
+	{
+		original = new_value;
+	}
+
+	/**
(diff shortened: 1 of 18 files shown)


src/core/backup_type.hpp
```diff
@@ -144,4 +144,48 @@ struct Backup {
...
+
+	/**
+	 * Backup variable and switch to new value.
+	 * @param original Variable to backup.
+	 * @param new_value New value for variable.
+	 */
+	template <typename U>
+	AutoRestoreBackup(T &original, const U &new_value) : original(original), original_value(original)
+	{
+		/* Note: We use a separate typename U, so type conversions are handled by assignment operator. */
```
LordAro: Why? When would the types not match?

rubidium42: Here never. I got left after copying Backup for apparantly no reason.

src/core/backup_type.hpp
```diff
@@ -144,4 +144,46 @@
...
+	 */
+
+	/**
+	 * Backup variable and switch to new value.
+	 * @param original Variable to backup.
+	 * @param new_value New value for variable.
+	 */
+	AutoRestoreBackup(T &original, T new_value) : original(original), original_value(original)
+	{
+		original = new_value;
```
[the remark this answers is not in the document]
rubidium42: This alert is suppressed as this class' whole purpose is to go out of scope and thus restore the non-local memory before the local variable address goes out of scope.

michicc (approved):

PR #10422, opened 2023-01-28
Change: [Actions] several improvements for the release workflow

### LINKED ISSUE #10301: [Bug]: GitHub Actions Release workflow publishes too many files to the CDN
The Windows Store part of the release workflow publishes two files to the CDN it really shouldn't:

- `output/OpenTTD.appxbundle`
- `cert.pfx`

This causes the CDN indexer to fail, as it doesn't know how to make those files available.

Please make sure those files are no longer published to the CDN, or, if we do want to publish them, please generate md5sum / sha1sum and sha256sum of those files before making an artifact out of them, like the other jobs do.

TrueBrain:
## Motivation / Problem

This has been debated a lot of times. But it boils down to this:

- We don't build Ubuntu/Debian for nightlies
- We only build them for actual releases
- They fail often, as either the compiler got too old, or it throws a warning, or something else is fishy.
- We only find this out during a release, causing a lot of hassle to fix it again

Basically, the maintenance burden is really high.

On the other side, we have a Generic Linux build, which should (famous last words) work for everyone running Linux. How do we know? It is the version we publish on Steam. So far, not a single report has been made that the Linux version on Steam is not working. So that somewhat suggests it works for a larger audience.

Additionally, we only publish binaries for Ubuntu/Debian. Not for RedHat/Fedora/CentOS/Arch/... So this was already a bit unfair, in that light.

To reduce the maintenance burden on us, I suggest we push back these binaries to the distros, and let them publish these files. They are much better equipped to do so. And it is likely most users can use our Generic Linux builds just fine, if they want to use the latest version. Or just use Steam, ofc ;)

Additionally, while at it, also tackled #10301 and automated GOG upload.

Fixes #10301

## Description

- Remove the "we only build this for releases" binaries of our pipeline
- Split up release pipeline into several smaller ones, by using reusing workflows
- Upload windows-store binaries into "internal" folder on the CDN
- Automate uploading to GOG

## Limitations

Looking at statistics, the Ubuntu/Debian builds are downloaded as often as the MacOS is from our website. There can be a few implications:
- People download Ubuntu/Debian builds, because that is what they are running. But if they are not offered to them, they would be fine downloading the Generic one.
- The amount of downloads via our websites are a fraction of our actual users, and most use GOG / Steam to get the Linux builds (which are the Generic ones)

## Checklist for review


glx22:
IIRC there's a duplicate dep for generic build, but probably can go in another PR

michicc (dismissed):
No better test than a release.

michicc (approved):

PR #10432, opened 2023-01-29
Fix #10419: Water infrastructure accounting when building ship depots and docks

### LINKED ISSUE #10419: [Bug]: Water infrastructure total mismatch when building ship depot over water objects on owned water
### Version of OpenTTD

master

### Expected result

Water infrastructure total is correct

### Actual result

Water infrastructure total is incorrect in the case below.

This may eventually lead to a multiplayer desync.

### Steps to reproduce

1. Build two canal pieces
2. Build object(s) on one or both of these pieces
3. Build a ship depot on the two canal pieces
4. The water infrastructure total is now incorrect
5. Removing the ship depot and canal pieces now leaves a water infrastructure of 1 or 2 (depending on whether one or both tiles had objects).

JGRennison:
## Motivation / Problem

See #10419

Building docks has the same failure cases as ship depots, above.

## Description

Fix water infrastructure accounting when building a ship depot or dock over an object on a canal.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -2535,14 +2535,17 @@ CommandCost CmdBuildDock(DoCommandFlag flags, TileIndex tile, StationID station_
 
 	if (flags & DC_EXEC) {
 		st->ship_station.Add(tile);
-		st->ship_station.Add(tile + TileOffsByDiagDir(direction));
+		TileIndex flat_tile = tile + TileOffsByDiagDir(direction);
+		st->ship_station.Add(flat_tile);
 		st->AddFacility(FACIL_DOCK, tile);
 
 		st->rect.BeforeAddRect(dock_area.tile, dock_area.w, dock_area.h, StationRect::ADD_TRY);
 
 		/* If the water part of the dock is on a canal, update infrastructure counts.
-		 * This is needed as we've unconditionally cleared that tile before. */
-		if (wc == WATER_CLASS_CANAL) {
+		 * This is needed as we've cleared that tile before.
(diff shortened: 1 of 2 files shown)


glx22 (dismissed):

src/station_cmd.cpp
```diff
@@ -2535,14 +2535,15 @@ CommandCost CmdBuildDock(DoCommandFlag flags, TileIndex tile, StationID station_
...
+		st->ship_station.Add(flat_tile);
 		st->AddFacility(FACIL_DOCK, tile);
 
 		st->rect.BeforeAddRect(dock_area.tile, dock_area.w, dock_area.h, StationRect::ADD_TRY);
 
 		/* If the water part of the dock is on a canal, update infrastructure counts.
-		 * This is needed as we've unconditionally cleared that tile before. */
-		if (wc == WATER_CLASS_CANAL) {
+		 * This is needed as we've cleared that tile before. */
+		if (wc == WATER_CLASS_CANAL && !(HasTileWaterClass(flat_tile) && GetWaterClass(flat_tile) == WATER_CLASS_CANAL && IsTileOwner(flat_tile, _current_company))) {
```
rubidium42, line 2548: Oh, how fun that `ClearTile` for objects does not actually clear the tile but removes the object. That might be messing more things up; maybe also locks?
Maybe at add a note in the comment in these places that for `MP_OBJECT` clearing could yield a `MP_WATER` tile instead of `MP_CLEAR`? I got very confused by this, as it did clear so the tile should have been `MP_CLEAR` and as such the `HasTileWaterClass` would already/still be returning false.

rubidium42 (approved):
Thanks for the comment update.

Issue #7534, opened 2019-04-21
GSGoal.QuestionClient sends message to wrong client


ldpl:
CmdGoalQuestion uses client index instead of client id but indexes are not synced on network so whoever receives it is somewhat random. 
Problem with sending client_id though is that it needs 32 bits and CmdGoalQuestion only have 27 available. 
Technically it can be squeezed there as only 10 bits is needed to send 988 possible button combinations but enumerating them isn't fun. 
So... any other ideas?

ldpl:
After a bit of though imo best solution atm would be to somehow do a 8-bit client id. After all there can only be 255 clients and 32-bit client id is only useful during connection phase. As how to do that I think of moving client creation into command queue as it's done for companies so that indexes will automatically become synced. Or mb just do some kind of sync index / set id command.

PR #10435, opened 2023-01-31
Client id mistype

rubidium42:
## Motivation / Problem

Samu and dP on chat.
Samu (paraphrased): `ResolveClientID` does not return same value for range `2**31 - 2**32-1` as entered. Such client IDs would be returned as negative numbers.
dP (paraphrased): `QuestionClient` does not work with `ClientID`s in range `2**16 - 2**32-1`.

## Description

For first issue: remove cast to `int32`, just let automatic promotion to `int64` take place so the `uint32` value remains positive

For second issue: remove `EnforcePrecondition` and change `target` from `uint16` to `uint32` in `CmdGoalQuestion`.

## Limitations

None AFAIK.

## Checklist for review

### DIFF
--- a/cmake/scripts/SquirrelExport.cmake
+++ b/cmake/scripts/SquirrelExport.cmake
@@ -298,7 +298,7 @@ foreach(LINE IN LISTS SOURCE_LINES)
                 string(APPEND SQUIRREL_EXPORT "\n	/* Allow enums to be used as Squirrel parameters */")
                 foreach(ENUM IN LISTS ENUMS)
                     string(APPEND SQUIRREL_EXPORT "\n	template <> inline ${ENUM} GetParam(ForceType<${ENUM}>, HSQUIRRELVM vm, int index, SQAutoFreePointers *ptr) { SQInteger tmp; sq_getinteger(vm, index, &tmp); return (${ENUM})tmp; }")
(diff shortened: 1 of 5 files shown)


SamuXarick:
Unnamed, 1935-01-01#5
ClientID is now good

michicc (approved):

LordAro:
Just to link them together, the Client ID issue is #7534

SamuXarick:
Found something that may require attention
https://github.com/OpenTTD/OpenTTD/blob/1897e98c3dec300a0b97cc1222906bff9fadea09/src/script/api/script_window.hpp.in#L66-L71

NUMBER_ALL value, what will happen to it?

rubidium42:
> NUMBER_ALL value, what will happen to it?

Technically I guess it's undefined, as the specification does not define whether an enum should be signed or unsigned. As long as all the values fit, then it's "fine". Practically whatever it is, it will end up in Squirrel land and then be passed back to the API where it gets cast into an unsigned integer and then the comparison itself will (implicitly) cast either the enum value or the parameter to the type of the other and do the comparison. So everything should be fine.
Having said that, maybe just force the type for all enumerations in the script API to not confuse people about the different behaviours.

PR #10443, opened 2023-02-02
Fix: [Script] ScriptBase::Rand() return value was between -MIN(int32) and MAX(int32)

glx22:
Also ensure the parameters for ScriptBase::RandRange() and ScriptBase::Chance() are in [0-MAX(uint32)] range

## Motivation / Problem
Documentation says ScriptBase::Rand() return a random value between 0 and MAX(uint32), but the actually returned value is between -MIN(int32) and MAX(int32).
ScriptBase::RandRange() has the same issue with values above MAX(int32) becoming negative.

## Description
Use SQInteger as return type, so uint32 values are casted to int64 instead of int32.
Also use SQInteger for parameters, allowing to check them to be in the uint32 range.

## Limitations
No compatibility scripts as the documentation was always correct.

## Checklist for review


src/script/api/script_base.cpp
```diff
@@ -10,36 +10,46 @@
 #include "../../stdafx.h"
 #include "script_base.hpp"
 #include "script_error.hpp"
+#include "script_log.hpp"
 
 #include "../../safeguards.h"
 
-/* static */ uint32 ScriptBase::Rand()
+/* static */ SQInteger ScriptBase::Rand()
```
SamuXarick, line 16: Isn't it better to use int64 than SQInteger? I always thought SQInteger was specifically used outside the API.

src/script/api/script_base.cpp
```diff
@@ -10,36 +10,46 @@
...
 {
 	return ScriptObject::GetRandomizer().Next();
 }
 
-/* static */ uint32 ScriptBase::RandItem(int unused_param)
+/* static */ SQInteger ScriptBase::RandItem(SQInteger unused_param)
```
SamuXarick, line 21: unused_param doesn't really need to change.

src/script/api/script_base.cpp
```diff
@@ -10,36 +10,46 @@
...
 {
 	return ScriptBase::Rand();
 }
 
-/* static */ uint ScriptBase::RandRange(uint max)
+/* static */ SQInteger ScriptBase::RandRange(SQInteger max)
 {
+	if (max < 0 || max > UINT32_MAX) {
+		ScriptLog::Warning("Clamping RandRange() parameter to 0..MAX(uint32)");
```
SamuXarick: ScriptLog::Warning is a bit too intrusive, better just tell a clamping may happen in the description, rather than outputing it into the log.

SamuXarick (commented):
Mind if I submit general feedback?

src/script/api/script_base.cpp
glx22: `SQInteger` is the type used in every squirrel internal functions (especially `sq_pushinteger()` to pass a number to the squirrel VM and `sq_getinteger()` to get a number from the VM). It is indeed an int64, but I think using SQInteger is better as it hides implementation details from the API.

src/script/api/script_base.cpp
glx22: True, but my idea would be to use `SQInteger` everywhere in the API for generic numbers.

src/script/api/script_base.cpp
glx22: Agreed, will update.

src/script/api/script_base.cpp
rubidium42: Then, as a reminder for a later (cleanup) story: remove the casts from squirrel_helper_type.hpp:49-51 as well as 57.

rubidium42 (approved):
Should we add a note to the changelogs about the behavior being incorrect based on the specification in everything before OpenTTD 14? And that there is no compatibility layer that reintroduces this buggy behavior to scripts requesting an older API version?
I'm not sure though, as this is just fixing the behavior to be like it's specified. Though it might also break some expectations.

src/script/api/script_base.cpp
glx22: Ah you mean squirrel_helper.hpp

PR #10444, opened 2023-02-03
Fix #10059: Clamp config item values to int, disallow negative random deviation, and allow values up to 12 digits including '-'

### LINKED ISSUE #10059: [Crash]: CTD and incorrect parameter value when adding GSInfo setting with certain int max values
### Version of OpenTTD

OpenTTD 12.2, Windows 10 64-bit

### Steps to reproduce

1. Create an info.nut file for a game script
2. Under the GSInfo derived class create a setting
3. Set the max value of this setting really high (as an example I did 4 billion, i dont know the exact range this happens at)
    - Note that setting this value too high (such as 4 trillion) will cause it to work normally. In this case, if you try to enter 4 billion as a value, it will stop typing at 400 million.
5. Set the other parameters of the setting to something random (It didn't matter in my testing)
6. Add a label for this setting for the value 0
7. Load the game into the "Configure Game Script Parameters" window
8. Click the dropdown (which is shown instead of a number picker)
9. Crash the game to desktop

See below for the screenshot at crash time. **I have attached a minimal game script setup to test this functionality:** [MinimalCTDTest.zip](https://github.com/OpenTTD/OpenTTD/files/9662175/MinimalCTDTest.zip)

**NOTE:** if you skip step 5 and do not put a label, the game will not crash but you will instead have incorrect parameters:
image
In this case, it does not show a drop down but the correct number picker, although it is grayed out. Inputting any number into the parameter will cause it to change from 0 to -29496729

### Upload crash files

The exception is "0xC0000094: The thread tried to divide an integer value by an integer divisor of zero."
[crash20220928063447.log](https://github.com/OpenTTD/OpenTTD/files/9662170/crash20220928063447.log)
crash20220928063447

SamuXarick:
## Motivation / Problem
We can't input custom negative values via query string on a config item and it can't be more than 10 digits.
Crash report is caused by implicit 64 bit to 32 bit conversions during construction of a script's settings.
We can't create labels for settings with a negative value.

## Description
- Clamp config item values to int32. They're already correctly read as int64 (or SQInteger).
- Disallow negative random_deviation. It would cause weirdness during AddDeviation, because it expects uint. If it is negative, the abs of it is set instead.
- Allow custom values via query string to be negative and up to 12 digits, to allow for a large enough number, like -2 147 483 648.
- Also allow saving to the savegame and openttd.cfg numbers these large.
- AddLabels now checks the second character of key_string identifiers. If it finds a '_', then the integer that comes after becomes negative, allowing adding labels to negative values of a setting.

## Limitations
I suppose none.

## Checklist for review

### DIFF
--- a/src/script/api/script_info_docs.hpp
+++ b/src/script/api/script_info_docs.hpp
@@ -254,13 +254,16 @@ public:
 	 *  user will see the corresponding name.
 	 * @param setting_name The name of the setting.
 	 * @param value_names A table that maps values to names. The first
-	 *   character of every identifier is ignored and the rest should
+	 *   character of every identifier is ignored, the second character
+	 *   could be '_' to indicate the value is negative, and the rest should
 	 *   be an integer of the value you define a name for. The value
 	 *   is a short description of that value.
 	 * To define labels for a setting named "competition_level" you could
 	 * for example call it like this:
 	 * AddLabels("competition_level", {_0 = "no competition", _1 = "some competition",
 	 * _2 = "a lot of competition"});
+	 * Another example, for a setting with a negative value:
+	 * AddLabels("amount", {__1 = "less than one", _0 = "none", _1 = "more than one"});
 	 *
 	 * @note This is a function provided by OpenTTD, you don't have to
 	 * include it in your Script but should just call it from GetSettings.
--- a/src/script/api/script_info_docs.hpp
+++ b/src/script/api/script_info_docs.hpp
@@ -218,20 +218,27 @@ public:
 	 *    store the current configuration of Scripts. Required.
 	 *  - description A single line describing the setting. Required.
 	 *  - min_value The minimum value of this setting. Required for integer
-	 *    settings and not allowed for boolean settings.
+	 *    settings and not allowed for boolean settings. The value will be
+	 *    clamped in the range [MIN(int32), MAX(int32)] (inclusive).
 	 *  - max_value The maximum value of this setting. Required for integer
-	 *    settings and not allowed for boolean settings.
+	 *    settings and not allowed for boolean settings. The value will be
+	 *    clamped in the range [MIN(int32), MAX(int32)] (inclusive).
 	 *  - easy_value The default value if the easy difficulty level
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [MIN(int32), MAX(int32)] (inclusive).
 	 *  - medium_value The default value if the medium difficulty level
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [MIN(int32), MAX(int32)] (inclusive).
 	 *  - hard_value The default value if the hard difficulty level
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [MIN(int32), MAX(int32)] (inclusive).
 	 *  - custom_value The default value if the custom difficulty level
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [MIN(int32), MAX(int32)] (inclusive).
 	 *  - random_deviation If this property has a nonzero value, then the
 	 *    actual value of the setting in game will be randomized in the range
 	 *    [user_configured_value - random_deviation, user_configured_value + random_deviation] (inclusive).
+	 *    random_deviation sign is ignored and the value is clamped in the range [0, MAX(int32)] (inclusive).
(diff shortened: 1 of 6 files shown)


glx22 (commented):
Documentation should mention the clamping and the allowed range.

src/script/script_info.cpp
```diff
@@ -252,7 +252,14 @@ SQInteger ScriptInfo::AddLabels(HSQUIRRELVM vm)
 		if (SQ_FAILED(sq_getstring(vm, -1, &label))) return SQ_ERROR;
 		/* Because squirrel doesn't support identifiers starting with a digit,
 		 * we skip the first character. */
-		int key = atoi(key_string + 1);
+		key_string++;
+		int sign = 1;
+		if (strchr(key_string, '_')) {
```
glx22:
```suggestion
		if (*key_string == '_') {
```

src/script/script_info.cpp
```diff
@@ -252,7 +252,14 @@ SQInteger ScriptInfo::AddLabels(HSQUIRRELVM vm)
...
+			/* When the second character is '_', it indicates the value is negative. */
+			sign = -1;
+			key_string++;
+		}
+		int key = atoi(key_string);
```
glx22: You don't apply the sign.

src/script/api/script_info_docs.hpp
```diff
@@ -218,17 +218,23 @@ class ScriptInfo {
 	 *    store the current configuration of Scripts. Required.
 	 *  - description A single line describing the setting. Required.
 	 *  - min_value The minimum value of this setting. Required for integer
-	 *    settings and not allowed for boolean settings.
+	 *    settings and not allowed for boolean settings. The value will be
+	 *    clamped in the range [-2147483648, 2147483647] (inclusive).
```
glx22: Didn't check all documentation, but some places use `MIN(int32) .. MAX(int32)` style.
Not sure which one is better, but consistency is good.

src/script/api/script_info_docs.hpp
```diff
@@ -218,17 +218,23 @@ class ScriptInfo {
...
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [-2147483648, 2147483647] (inclusive).
 	 *  - custom_value The default value if the custom difficulty level
-	 *    is selected. Required.
+	 *    is selected. Required. The value will be clamped in the range
+	 *    [-2147483648, 2147483647] (inclusive).
 	 *  - random_deviation If this property has a nonzero value, then the
 	 *    actual value of the setting in game will be randomized in the range
 	 *    [user_configured_value - random_deviation, user_configured_value + random_deviation] (inclusive).
```
glx22, line 240: This one is also clamped but to `0 .. MAX(int32)`

SamuXarick: It's not exactly "clamped". I just don't know how to describe it. It's Clamp(abs(res)) in code. -30 becomes 30, not 0.

glx22: Sign is ignored, and it's clamped.

src/script/script_config.cpp
```diff
@@ -216,7 +216,7 @@ std::string ScriptConfig::SettingsToString() const
 	char *s = string;
 	*s = '\0';
 	for (const auto &item : this->settings) {
-		char no[10];
+		char no[12];
```
rubidium42: First of all the commit message is not correct compared to this code. The 12 in here contains a trailing `\0`, so at most 11 digits could be placed.
Secondly, the whole function is trying hard to stay within the artificial 1024 byte limit, but then returns a theoreticall unbounded `string`, so why not use proper C++ constructs to create this settings string?
Thirdly, the ini file reading seems to use a 1024 byte buffer as well, so how can the written value ever be correctly read, because I think there is some key in front of the configuration.

So all-in-all it's more broken than this commit fixes, though I can agree that the second and third issue are not for this PR. The commit message definitely is broken, and the current number is magic-y without any comments, so what about something like:
`char no[10 + 1 + 1]; // Maximum of 10 digits for MIN/MAX_INT32, 1 for the sign and 1 for '\0'.`
And if the "12" gets used in many more places, a constant would be better as it can be reused, maybe something like `INT32_DIGITS_WITH_SIGN_AND_TERMINATION`.

SamuXarick: I placed INT32_DIGITS_WITH_SIGN_AND_TERMINATION in the script_config.hpp with comment, added includes in script_gui.cpp and game_gui.cpp to read this value. Also edited 2 commit messages.

rubidium42 (approved):

PR #10448, opened 2023-02-05
Add: use https:// for content-service connections

TrueBrain:
## Motivation / Problem

Our content-service still runs over `http://`. This is more and more problematic:

1) slowly companies start to disallow the usage of `http://`, and only allow `https://`, connections to be made from client machines.
2) running an HTTP-server on the Internet limits where we can host this. Services like Cloudflare and GitHub doesn't really fancy non-HTTPS connections anymore. Ideally we would like to migrate our CDN to Cloudflare R2, which doesn't allow `http://` on public buckets.

As I was also looking into automated reporting of crashes to Sentry, which also requires libcurl / winhttp, I was like: why not add libcurl to start with, and see how that goes.

This also paves a way towards cloud-saves, as for this we really do need a secure connection to the cloud. As last "added bonus", it means we finally get ride of our custom "HTTP client", which was `HTTP/1.0` only.

Basically, time for OpenTTD to join 2023 ;)

## Description

This requires the use of WinHTTP (for Windows) or libcurl (for all others).

On Linux it requires ca-certificates to be installed, so the HTTPS certificate can be validated. It is really likely this is installed on any modern machine, as most connections these days are HTTPS.

(On MacOS and Windows the certificate store is filled by default)

Reminder: in case the http(s):// connection cannot be established, OpenTTD falls back to a custom TCP-based connection to fetch the content from the content-service.

### Why winhttp for Windows, and not libcurl?

Well, Windows has an excellent HTTP client on-board, and this means we have one less dependency for Windows. Which means we don't have to look into shipping a DLL next to our binaries. So it is just a win/win.

## Limitations

- If an OS doesn't have the certificate store filled, the HTTPS connection cannot be validated. We could ship our own CAcerts, and feed that to libcurl / winhttp, but this overly complicates the setup. As I think it is rather unlikely a modern computer doesn't have a certificate store filled, I don't foresee any real issues with this.
- If the above assumption is wrong, we would see an increase in TCP traffic towards our infrastructure (as the `https://` connections are failing). This would mean we would have to rush another release after this is released. Not a biggy, in my opinion.
- Emscripten's method to download URLs (Emscripten Fetch) is a mess; streaming only works on Firefox, and without it, it makes a shitty user experience. So for now, that will still use the fallback method.

## Things to fix outside this PR

- Several improvements can be done for both libcurl as winhttp implementation. Additionally, the way HTTP requests are made can surely need a polish. Most noticeable, it would be better to run a single thread that handles all HTTP traffic, and make a simple queue to run them down. But to limit the scope of this PR, this is not done. Let's first see if HTTPS is working for most people :)

## Checklist for review


andythenorth:
Builds and runs on macOS.  libcurl appears to be included with macOS (might need XCode is installed).

Downloaded a grf, worked fine.

stormcone:
It works on Fedora 37.
Commit: [15ddf94793cea17e5581938fa063126b2add39c5](https://github.com/TrueBrain/OpenTTD/commit/15ddf94793cea17e5581938fa063126b2add39c5)

Log:
<details>

$ ./openttd -d net=4
dbg: [net] Starting network
dbg: [net] Initializing UDP listeners
dbg: [net] Network online, multiplayer available
dbg: [net] Connected to content.openttd.org:3978
dbg: [net] HTTP request to https://binaries.openttd.org/bananas
dbg: [net] Network revision name: norev0000-uunknown
dbg: [net] HTTP callback: 423 bytes
dbg: [net] HTTP request succeeded
dbg: [net] HTTP request to http://bananas-2.cdn.openttd.org/newgrf/415a1101/0b86ead276c998da0c4265e83505040b/415a1101-C.O.N.V.E.Y.O.R._B.E.L.T.S-1.0.0.tar.gz
dbg: [net] HTTP callback: 14220 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 12576 bytes
dbg: [net] HTTP callback: 2016 bytes
dbg: [net] HTTP request succeeded
dbg: [net] HTTP request to http://bananas-2.cdn.openttd.org/newgrf/54530102/0aca956a9a97f341d27173b69c3f203a/54530102-CFL_Trains_Luxembourgish_Train-1.04.tar.gz
dbg: [net] HTTP callback: 14233 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 12576 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 2076 bytes
dbg: [net] HTTP request succeeded
dbg: [net] HTTP request to http://bananas-2.cdn.openttd.org/newgrf/4d430207/c3aaab4fe3d3956af4d82ca62091cd5d/4d430207-PolRoad-5.5.tar.gz
dbg: [net] HTTP callback: 14230 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 16384 bytes
dbg: [net] HTTP callback: 806 bytes
dbg: [net] HTTP request succeeded
</details>

glx22:
Windows 10, IPv6
```
dbg: [net] Connected to content.openttd.org:3978
dbg: [net] HTTP request to https://binaries.openttd.org/bananas
dbg: [net] HTTP request status code: 200
dbg: [net] HTTP request to http://<redacted url to real content>.tar.gz
dbg: [net] HTTP request succeeded
dbg: [net] HTTP request status code: 200
dbg: [net] HTTP request succeeded
```
Works fine.

src/network/core/http_curl.cpp
```diff
@@ -0,0 +1,216 @@
...
+NetworkHTTPRequest::~NetworkHTTPRequest()
+{
+	if (this->curl) {
+		curl_multi_remove_handle(this->multi_handle, this->curl);
+		curl_multi_cleanup(this->multi_handle);
+
+		curl_easy_cleanup(this->curl);
+	}
+
+	free(this->data);
```
rubidium42, line 178: Why not make `data` a `std::unique_ptr<const char*>`? Then you don't need to do the manual memory management.
Technically might want to use `std::unique_ptr` with a customer deleter for `curl` and `multi_handle` too, so you can leave out the manual close completely, but that seems more finicky here than that it's worth.

src/network/core/http_winhttp.cpp
```diff
@@ -0,0 +1,301 @@
...
+			/* Next step: read the data in a temporary buffer. */
+			char *buffer = size == 0 ? nullptr : MallocT<char>(size);
+			WinHttpReadData(this->request, buffer, size, 0);
+		} break;
+
+		case WINHTTP_CALLBACK_STATUS_READ_COMPLETE:
+			Debug(net, 4, "HTTP callback: {} bytes", length);
+
+			this->callback->OnReceiveData(static_cast<char *>(info), length);
+			free(info);
```
rubidium42, line 158: I'd at cross-reference that this is freeing the allocation that is made in the `WINHTTP_CALLBACK_STATUS_DATA_AVAILABLE` case in a comment.

src/network/core/http_winhttp.cpp
```diff
@@ -0,0 +1,301 @@
...
+			WinHttpQueryDataAvailable(this->request, nullptr);
+		} break;
+
+		case WINHTTP_CALLBACK_STATUS_DATA_AVAILABLE:
+		{
+			/* Retrieve the amount of data available to process. */
+			DWORD size = *(DWORD *)info;
+
+			/* Next step: read the data in a temporary buffer. */
+			char *buffer = size == 0 ? nullptr : MallocT<char>(size);
```
rubidium42, line 149: I'd at cross-reference that this is allocating the memory that is freed in the `WINHTTP_CALLBACK_STATUS_READ_COMPLETE` case in a comment.

src/network/core/http_winhttp.cpp
```diff
@@ -0,0 +1,301 @@
...
+			}
+
+			/* Next step: query for any data. */
+			WinHttpQueryDataAvailable(this->request, nullptr);
+		} break;
+
+		case WINHTTP_CALLBACK_STATUS_DATA_AVAILABLE:
+		{
+			/* Retrieve the amount of data available to process. */
+			DWORD size = *(DWORD *)info;
```
rubidium42, line 145: Should we do some sort of sanity check on this size?

src/network/core/http_winhttp.cpp
rubidium42, line 149: If the data is available, then it should be in some sort of buffer. Why is this copy of the data needed, especially if we're copying it later to disk as well?

src/network/core/http_winhttp.cpp
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * Destructor of the HTTP request.
+ *
+ * Makes sure all handlers are closed, and all memory is free'd.
+ */
+NetworkHTTPRequest::~NetworkHTTPRequest()
+{
+	if (this->request) {
+		WinHttpCloseHandle(this->request);
+		WinHttpCloseHandle(this->connection);
```
rubidium42, line 264: Technically leaks resources (in release builds without asserts) in case the `connection` is opened by the `request` isn't.

src/network/core/http_curl.cpp
TrueBrain: There are more of those improvements we can do, but that would touch more than just the http-part. I think I prefer to leave that for another PR, if you don't mind?

src/network/core/http_winhttp.cpp
TrueBrain: What sanity check do you suggest? I can't really think of one ... that would mean the Windows API did something weird, so I find it difficult to think about what to check for.

src/network/core/http_winhttp.cpp
TrueBrain: That is a question for the WinHttp API gods, and not one I can answer :( This is how the documentation says how to implement this. Not sure what more I can add, sorry :)

https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpreaddata for reference. To quote:

`Pointer to a buffer that receives the data read. Make sure that this buffer remains valid until WinHttpReadData has completed.`

src/network/core/http_winhttp.cpp
TrueBrain: I was lazy, and you are right :P Let me fix it properly :)

src/network/core/http_curl.cpp
rubidium42: Fine by me.

src/network/core/http_winhttp.cpp
rubidium42: I don't really know either, though that could mean allocating a few GB of memory.
I also have no clue what happens when reading/allocating a buffer smaller that this size. Would that just call this callback multiple times? Then limiting to something like 1 MB might be useful to not use a lot of memory for a 32 bpp base set.

src/network/core/http_winhttp.cpp
rubidium42: To quote, on the same page: `Although calling WinHttpReadData immediately after completion of WinHttpReceiveResponse avoids the expense of a buffer copy, doing so requires that the application use a fixed-length buffer for reading.`
So it is possible to do it without the extra copy, but I have no clue what the impact of it would be. It's mostly a check question to prevent the extra copying if that's reasonably easy to attain.

TrueBrain: How I read that, is that it internally doesn't allocate an additional buffer. So out of our hands :)

So yeah, let's do this for now; hopefully someone with more knowledge about WinHttp comes along and goes: NOOOO, and fixes it :D

src/network/core/http_winhttp.cpp
TrueBrain: I did some checking, also with other projects. Both the Windows manual suggests to do it like this, but software like `wine` and `duckstation` also do it like this. So I am guessing, but I cannot confirm that, that Windows take care of this. Otherwise a lot of these applications would constantly run out of memory :)

WinHttp is just a bit weird to understand, so I rather stick with what the manual says, as that is most likely to work. If we ever get reports about out-of-memory issues, we can pick it up from there :)

CMakeLists.txt
```diff
@@ -124,6 +124,13 @@ find_package(LibLZMA)
 find_package(LZO)
 find_package(PNG)
 
+if(WIN32)
+    # Windows uses WinHttp for HTTP requests.
+else()
+    # All other targets use libcurl.
```
rubidium42, line 131: What about Emscripten? I know that if it doesn't find it, it will not fail but if `CURL_FOUND` then the CMakeLists.txt for src/network/core will add both http_curl.cpp and http_emscripten.cpp, which as far as I can see both add the same symbols causing build failures.

COMPILING.md
```diff
@@ -4,6 +4,7 @@
 
 OpenTTD makes use of the following external libraries:
 
+- (required) libcurl: content downloads
```
rubidium42: If libcurl is required, then zlib must at least be required as well. Being able to download the data, but not decompressing it means the functionality is still pointless.
Technically a build without libcurl would still work, if it would fall back to the Emscripten http fallback. Also, it's only required for `!Windows && !Emscripten`, so maybe it does not belong in the block of the document that describes the libraries required/encouraged along all platforms.

rubidium42: Line 22-24: `OpenTTD does not require any of the libraries to be present, ...` is not valid anymore if going with the idea that curl is required. Do we want to lose the ability to build without library dependencies? Or should we make a set of them just required?

COMPILING.md
```diff
@@ -45,6 +46,7 @@ by following the `Quick Start` instructions of their
 After this, you can install the dependencies OpenTTD needs. We advise to use
 the `static` versions, and OpenTTD currently needs the following dependencies:
 
+- curl
```
rubidium42: Including lines 58 and 59: why require/install curl when using winhttp?

rubidium42 (commented):
Maybe the crashlog code should also log the version of libcurl?

CMakeLists.txt
TrueBrain: It is impossible to build libcurl for emscripten, but I can add some extra letters here to make that more obvious.

COMPILING.md
TrueBrain: Wauw, that is a lot to unpack :P

Currently, having libcurl without zlib is not useful, but that can (and will) change over time. So having that in this list feels like something that people will forget when they use a http connection for something else. So I think we shouldn't do that :)

I will make more clear it is linux only :)

rubidium42 (approved):
The pedant in me would say, what about MacOS in COMPILING.md, but then there's a whole cesspit with other relatively small oversights like that, such as not mentioning BSDs which do require the SDL/freetype stuff, and the choice between sdl/allegro, and...

PR #10458, opened 2023-02-08
Codechange: Move AI settings back to ai_gui

SamuXarick:
This also opens GS's own config window from Debug window, and reverts caption fixes.

## Motivation / Problem
I think the Script Settings window move to script_gui was precipitated. Yes, it can do both AI and GS, but GS now has it's own window which also does settings. The only reason this was moved away from ai_gui was due to the Settings button on the Script Debug window.

## Description
My proposal is to make the Settings button from Debug window to open the GS Config window instead. And with that done, move Script Settings window back to ai_gui and rename it and its properties appropriately to AI Settings.
A revert is also done for some strings, as this window is now solely for AI Settings. I added an assert just to make sure no OWNER_DEITY can access it.

## Limitations
The GS Config window could perhaps be improved so that it only shows settings and not the config part when accessed via Debug window, but it's currently okay for now. The uneccessary buttons are shown as disabled.

## Checklist for review

### DIFF
--- a/src/ai/ai_gui.cpp
+++ b/src/ai/ai_gui.cpp
@@ -9,11 +9,14 @@
 
 #include "../stdafx.h"
 #include "../error.h"
+#include "../settings_gui.h"
+#include "../querystring_gui.h"
 #include "../company_base.h"
 #include "../window_func.h"
 #include "../network/network.h"
 #include "../settings_func.h"
 #include "../network/network_content.h"
+#include "../widgets/dropdown_func.h"
 #include "../core/geometry_func.hpp"
 
 #include "ai.hpp"
@@ -68,7 +71,7 @@ static const NWidgetPart _nested_ai_config_widgets[] = {
 
 /** Window definition for the configure AI window. */
 static WindowDesc _ai_config_desc(
-	WDP_CENTER, "settings_script_config", 0, 0,
+	WDP_CENTER, "settings_ai_config", 0, 0,
 	WC_GAME_OPTIONS, WC_NONE,
 	0,
 	_nested_ai_config_widgets, lengthof(_nested_ai_config_widgets)
@@ -96,7 +99,7 @@ struct AIConfigWindow : public Window {
 	void Close() override
 	{
 		CloseWindowByClass(WC_SCRIPT_LIST);
-		CloseWindowByClass(WC_SCRIPT_SETTINGS);
+		CloseWindowByClass(WC_AI_SETTINGS);
 		this->Window::Close();
 	}
 
@@ -221,7 +224,7 @@ struct AIConfigWindow : public Window {
 				break;
 
 			case WID_AIC_CONFIGURE: // change the settings for an AI
-				ShowScriptSettingsWindow((CompanyID)this->selected_slot);
+				ShowAISettingsWindow((CompanyID)this->selected_slot);
 				break;
 
 			case WID_AIC_CLOSE:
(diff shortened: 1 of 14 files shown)


src/table/settings/gui_settings.ini
```diff
@@ -779,7 +779,7 @@ cat      = SC_EXPERT
 var      = gui.ai_developer_tools
 flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
 def      = false
-post_cb  = [](auto) { InvalidateWindowClassesData(WC_SCRIPT_SETTINGS); }
+post_cb  = [](auto) { InvalidateWindowClassesData(WC_AI_SETTINGS); InvalidateWindowData(WC_GAME_OPTIONS, WN_GAME_OPTIONS_GS); }
```
glx22: While at it, maybe also invalidate debug window (to show/hide break string)

src/ai/ai_gui.cpp
```diff
@@ -264,6 +267,354 @@ struct AIConfigWindow : public Window {
 	}
 };
 
+
+/**
+ * Window for settings the parameters of an AI.
+ */
+struct AISettingsWindow : public Window {
+	CompanyID slot;                       ///< The currently show company's setting.
+	ScriptConfig *script_config;          ///< The configuration we're modifying.
```
glx22: Could be renamed back to ai_config.

src/lang/english.txt
```diff
@@ -4623,10 +4623,8 @@ STR_SCREENSHOT_WORLD_SCREENSHOT                                 :{BLACK}Whole ma
 STR_SCREENSHOT_HEIGHTMAP_SCREENSHOT                             :{BLACK}Heightmap screenshot
 STR_SCREENSHOT_MINIMAP_SCREENSHOT                               :{BLACK}Minimap screenshot
 
-# Script Parameters
-STR_AI_SETTINGS_CAPTION                                         :{WHITE}{STRING} Parameters
-STR_AI_SETTINGS_CAPTION_AI                                      :AI
-STR_AI_SETTINGS_CAPTION_GAMESCRIPT                              :Game Script
+# AI Parameters
+STR_AI_SETTINGS_CAPTION_AI                                      :{WHITE}AI Parameters
```
glx22: Translators will be mad ;)
Maybe removing STR_AI_SETTINGS_CAPTION_AI and STR_AI_SETTINGS_CAPTION_GAMESCRIPT, and changing STR_AI_SETTINGS_CAPTION to `{WHITE}AI Parameters` would be better (will generate many warnings on compile but should be fine)

glx22 (commented):
I think GSConfigWindow must also be looked at, it seems to mix usage of `this->gs_config` and `GameConfig::GetConfig()`.

And ScriptListWindow could also cache ScriptConfig for the slot in constructor instead calling `GetConfig()` everytime. Then the static function can be removed.

src/ai/ai_gui.cpp
```diff
@@ -264,6 +267,354 @@ struct AIConfigWindow : public Window {
...
+ * @param slot The CompanyID of the Script to change the settings.
+ */
+void ShowAISettingsWindow(CompanyID slot)
+{
+	CloseWindowByClass(WC_SCRIPT_LIST);
+	CloseWindowByClass(WC_AI_SETTINGS);
+	new AISettingsWindow(&_ai_settings_desc, slot);
+}
+
 /** Open the AI config window. */
```
glx22: I think we usually try to keep the `ShowXXXWindow()` near the window definition.

Edit: looks like the line selection somehow failed, I wanted to point 618-623

glx22 (commented):
I forgot something in the previous review.

Oh and after a quick test following an intuition, there's an existing "bug" (ie not introduced by this PR) with ScriptListWindow, once it's opened, selecting another AI 'line' in AIConfigWindow is not propagated, and applying change in ScriptListWindow will affect the previously selected 'line'.
Maybe ScriptListWindow should be closed when AIConfigWindow selection is changed, or ScriptListWindow could somehow get the new slot.
Same applies to AISettingsWindow.

src/ai/ai_gui.cpp
```diff
@@ -264,6 +267,354 @@ struct AIConfigWindow : public Window {
...
+						/* Increase button clicked */
+						new_val += config_item.step_size;
+						if (new_val > config_item.max_value) new_val = config_item.max_value;
+						this->clicked_increase = true;
+					} else {
+						/* Decrease button clicked */
+						new_val -= config_item.step_size;
+						if (new_val < config_item.min_value) new_val = config_item.min_value;
+						this->clicked_increase = false;
+					}
```
Arastais: Possible programmatic simplification (although it could be argued that it's less human-readable now):
```suggestion
					} else {
						/* Increase/Decrease button clicked */
						this->clicked_increase = (x >= SETTING_BUTTON_WIDTH / 2);
						new_val = this->clicked_increase ?
							std::min(config_item.max_value, new_val + config_item.step_size) :
							std::max(config_item.min_value, new_val - config_item.step_size);
					}
```

glx22: Same could apply to GSConfigWindow

src/table/settings/gui_settings.ini
SamuXarick: I just tried it, but the invalidate code of that window doesn't really do anything in that regard.

SamuXarick:
<Samu> I've been having 2nd thoughts about the ai settings fiasco, and came to realise that if I'm actually going to have the gsconfig window look more like the ai settings window without the config part, why not just use the ai settings window instead, and thus making my whole PR point invalid
<Samu> but then I wake up and see ppl embracing my PR 
<Samu> I made a bad judgement
<Samu> but okay
<Samu> let's see where it goes

<Samu> To do: debug window invalidate data showing / hiding break
<Samu> To do 2: hide config buttons from gs config if accessed via debug window
<Samu> To do 3: the propagated thing glx mentioned
<Samu> though i don't really consider it to be a bug
<Samu> more like a feature
<Samu> I don't know how to show/hide widgets

2TallTyler:
Reviewing your PR is not the same as embracing it. If you changed your mind and don't want it merged, you can just close this PR.

SamuXarick:
I am, or was about to undo the Script Settings move I've done, and go back to what it was before. The remaining changes would be these:
https://github.com/OpenTTD/OpenTTD/compare/master...SamuXarick:OpenTTD:simplify-script-gui-related-functions

glx22:
Personally I'm fine with both options.

PR #10459, opened 2023-02-09
Fix: [Script] Use Money instead of int32 for presenting the value of a company to AIs and show buy company dialog window even when playing in the AI company

SamuXarick:
## Motivation / Problem
No necessarily this PR, but it seems silly to have an `int32` value instead of a `Money` value. Otherwise the AI will potentially get a lower value presented and accepts it, without knowing how expensive it actually is.

_Originally posted by @rubidium42 in https://github.com/OpenTTD/OpenTTD/pull/10411#discussion_r1101492146_

I have something against this
https://github.com/OpenTTD/OpenTTD/blob/59251d3c6b9bf2a0ae85aefc2c9e9f91a46f3031/src/company_cmd.cpp#L705-L707
The is_ai check is unnecessary. If the company is not AI, AI::NewEvent will deal with it accordingly and won't send the event anyway. And if I'm the local company in the AI company, I'd still like to receive the buyout popup message.

## Description
No longer clamp bankrupt value to int32, and use Money instead.
Show the buy company dialog even when playing in the AI company.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -702,9 +702,9 @@ static void HandleBankruptcyTakeover(Company *c)
 	SetBit(c->bankrupt_asked, best->index);
 
 	c->bankrupt_timeout = TAKE_OVER_TIMEOUT;
-	if (best->is_ai) {
-		AI::NewEvent(best->index, new ScriptEventCompanyAskMerger(c->index, c->bankrupt_value));
-	} else if (IsInteractiveCompany(best->index)) {
+
+	AI::NewEvent(best->index, new ScriptEventCompanyAskMerger(c->index, c->bankrupt_value));
+	if (IsInteractiveCompany(best->index)) {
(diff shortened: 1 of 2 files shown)


src/company_cmd.cpp
```diff
@@ -702,9 +702,9 @@ static void HandleBankruptcyTakeover(Company *c)
 	SetBit(c->bankrupt_asked, best->index);
 
 	c->bankrupt_timeout = TAKE_OVER_TIMEOUT;
-	if (best->is_ai) {
```
glx22, line 705: Why remove the filter ?

glx22: Oh I missed the edition in first post

2TallTyler (approved):
In the future these two unrelated commits should be in separate PRs, but this looks fine to me.

PR #10468, opened 2023-02-12
Change: [Actions] also attempt a build without any (encouraged) dependencies

TrueBrain:
## Motivation / Problem

We weren't actually testing if you can build OpenTTD without zlib, libpng, ...
Trust but validate.

## Description

Use the "dedicated server" build to also build without all the additional dependencies.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -94,6 +94,9 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


TrueBrain:
Sadly, GitHub Runners already install all the encouraged dependencies we use, so this is not actually doing what I was hoping it would.

An alternative is to add another Windows build which does this, but .. like .. we have enough builds as it is. The extra time required doesn't outweighs the benefits.

PR #10471, opened 2023-02-12
Fix #10469, 5e14a20: [Script] League Table rating element is a int64 everywhere else

### LINKED ISSUE #10469: [Bug]: League Table: zero is lower than any negative number
### Version of OpenTTD

13.0, Windows 11 Pro

### Expected result

Zero is higher than any negative number.

### Actual result

Zero is lower than any negative number.

Unnamed, 1953-11-28

### Steps to reproduce

While working on a GS using LeagueTable functions, I noticed an issue with sorting values.

I think https://github.com/OpenTTD/OpenTTD/blob/64523709bffa32b2a61ff8aeb522182636b2efea/src/league_base.h#L33
is influencing the sorting

Here's the Game Script
[Leage Table.zip](https://github.com/OpenTTD/OpenTTD/files/10716042/Leage.Table.zip)

SamuXarick:
## Motivation / Problem
Closes #10469
Rating is a int64 in the Cmd functions, in the script API functions, but not on the league table element struct. Sorting through negative values would cause 0 to be below them.

## Description
Turn rating in struct into int64, and add savegame bump and conversion.

## Limitations
Probably will cause issues if the values saved on an old savegame were already above INT64_MAX to begin with, they would turn negative after load, most likely.

## Checklist for review

### DIFF
--- a/src/league_base.h
+++ b/src/league_base.h
@@ -30,7 +30,7 @@ extern LeagueTablePool _league_table_pool;
  **/
 struct LeagueTableElement : LeagueTableElementPool::PoolItem<&_league_table_element_pool> {
 	LeagueTableID table;  ///< Id of the table which this element belongs to
-	uint64 rating;        ///< Value that determines ordering of elements in the table (higher=better)
+	int64 rating;         ///< Value that determines ordering of elements in the table (higher=better)
 	CompanyID company;    ///< Company Id to show the color blob for or INVALID_COMPANY
 	std::string text;     ///< Text of the element
 	std::string score;    ///< String representation of the score associated with the element
--- a/src/saveload/league_sl.cpp
+++ b/src/saveload/league_sl.cpp
@@ -16,13 +16,14 @@
 #include "../safeguards.h"
 
 static const SaveLoad _league_table_elements_desc[] = {
-	 SLE_VAR(LeagueTableElement, table,       SLE_UINT8),
-	 SLE_VAR(LeagueTableElement, rating,      SLE_UINT64),
-	 SLE_VAR(LeagueTableElement, company,     SLE_UINT8),
-	SLE_SSTR(LeagueTableElement, text,        SLE_STR | SLF_ALLOW_CONTROL),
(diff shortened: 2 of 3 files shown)


ldpl (dismissed):
I don't remember why it ended up with different types. Probably I changed it at some point and missed one place. At the end of the day it's just an arbitrary field only used for sorting so it doesn't matter much what type it is. But as GS types are signed having it stored as signed as well makes more sense. Probably GS methods can even accept SQInteger or whatever is GS type called. But as SQInteger isn't used in other functions yet it's probably fine to leave it as int64 is for this PR.

Also, savegame conversion stuff and version bump is logically redundant as it is just a int64 stored in that uint64 anyway. But I don't know of a better way to express that in the new save format so I guess it's fine as it is.

glx22:
I don't think any changes in the savegame stuff is needed, using SLE_UINT64 for an int64 should work fine.

ldpl:
> I don't think any changes in the savegame stuff is needed, using SLE_UINT64 for an int64 should work fine.

It feels kinda dumb to have to use a wrong type to bypass a type check.

rubidium42:
> > I don't think any changes in the savegame stuff is needed, using SLE_UINT64 for an int64 should work fine.
> 
> It feels kinda dumb to have to use a wrong type to bypass a type check.

There is a middle ground solution `SLE_VAR(LeagueTableElement, rating, SLE_FILE_U64 | SLE_VAR_I64),`, which could be accompanied by a comment that the type in the file is technically wrong, but that a savegame bump is an excessive change. But that, when a savegame bump is required for the LeageTableElement, this should be corrected by a proper change as well.

glx22 (approved):

glx22:
Backport can only be partial due to savegame bump, but uint64 -> int64 itself can go to release branch I think.

rubidium42:
> Backport can only be partial due to savegame bump, but uint64 -> int64 itself can go to release branch I think.

Hmm, true. How to annotate that for the poor soul that's going to try to backport?

PR #10480, opened 2023-02-14
Codechange: move curl into a thread so simplify code

TrueBrain:
## Motivation / Problem

The async variant of curl gave too much trouble / noise. Switching to a threaded/sync solution brings it on-par with WinHttp, and makes the code a lot easier.

## Description

With a thread, we can just run curl_easy_perform() and let CURL and threads handle the blocking part.

With async solution there are too many things to keep track of, and it makes "when to update the GUI" tricky. By using a thread that all gets a lot simpler, as the game-thread and download-thread run side-by-side.

This is similar to how the WinHttp backend already works.

## Checklist for review

### DIFF
--- a/src/network/core/http_curl.cpp
+++ b/src/network/core/http_curl.cpp
@@ -12,216 +12,169 @@
 #include "../../stdafx.h"
 #include "../../debug.h"
 #include "../../rev.h"
+#include "../../thread.h"
 #include "../network_internal.h"
 
 #include "http.h"
 
+#include <atomic>
+#include <condition_variable>
 #include <curl/curl.h>
+#include <memory>
+#include <mutex>
+#include <queue>
 
 #include "../../safeguards.h"
 
 /** Single HTTP request. */
 class NetworkHTTPRequest {
-private:
+public:
+	/**
+	 * Create a new HTTP request.
+	 *
+	 * @param uri      the URI to connect to (https://.../..).
+	 * @param callback the callback to send data back on.
+	 * @param data     optionally, the data we want to send. When set, this will be a POST request, otherwise a GET request.
+	 */
+	NetworkHTTPRequest(const std::string &uri, HTTPCallback *callback, const char *data = nullptr) :
+		uri(uri),
+		callback(callback),
+		data(data)
+	{
+	}
+
+	/**
+	 * Destructor of the HTTP request.
+	 */
+	~NetworkHTTPRequest()
+	{
+		free(this->data);
+	}
+
 	std::string uri;        ///< URI to connect to.
 	HTTPCallback *callback; ///< Callback to send data back on.
 	const char *data;       ///< Data to send, if any.
-
-	CURL *curl = nullptr;          ///< CURL handle.
-	CURLM *multi_handle = nullptr; ///< CURL multi-handle.
-
-public:
-	NetworkHTTPRequest(const std::string &uri, HTTPCallback *callback, const char *data = nullptr);
-
-	~NetworkHTTPRequest();
-
-	void Connect();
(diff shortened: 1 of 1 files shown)


src/network/core/http_curl.cpp
```diff
@@ -12,216 +12,156 @@
...
+	{
+	}
 
-	~NetworkHTTPRequest();
+	/**
+	 * Destructor of the HTTP request.
+	 */
+	~NetworkHTTPRequest()
+	{
+		free(this->data);
```
LordAro, line 51: I know it was in the old curl too, but where is this memory coming from such that it needs to be freed with `free` ? Can it use new/delete instead? Or better, something automatic?

TrueBrain: Rubidium asked the same :D

The reason I hold off on that, as that is somewhere inside the network_content code. I have no problems addressing that, but I tried to keep these PRs as low-impact as possible, read: only address the http part.

Next PR will address these things, together with some other bugs and troubles we have, if you don't mind. If you do, let me know :)

frosch123 (dismissed):

src/network/core/http_curl.cpp
```diff
@@ -12,216 +12,165 @@
...
- * Destructor of the HTTP request.
- *
- * Makes sure all handlers are closed, and all memory is free'd.
- */
-NetworkHTTPRequest::~NetworkHTTPRequest()
-{
-	if (this->curl) {
-		curl_multi_remove_handle(this->multi_handle, this->curl);
-		curl_multi_cleanup(this->multi_handle);
+		std::shared_ptr<NetworkHTTPRequest> request = _http_requests.front();
```
rubidium42: I think we do not need `shared_ptr`, but can use `unique_ptr` if `move`-ing from `front()`:
```suggestion
		std::unique_ptr<NetworkHTTPRequest> request = std::move(_http_requests.front());
```
The other `shared`s in this PR also need to become `unique` for this to work.

src/network/core/http_curl.cpp
```diff
@@ -12,216 +12,165 @@
...
+	_http_thread_exit = false;
+	StartNewThread(&_http_thread, "ottd:http", &HttpThread);
 }
 
 void NetworkHTTPUninitialize()
 {
-	curl_share_cleanup(_curl_share);
 	curl_global_cleanup();
+
+	_http_thread_exit = true;
```
frosch123, line 170: "_http_thread_exit = true" + "notify_one" must be inside the mutex.
Otherwise they could be executed just before the http-thread goes to sleep.

On the plus side, "_http_thread_exit" is then protected by the mutex, and no longer needs to be atomic.

TrueBrain: Hmm, not sure I completely agree.

That is, for the `notify_one`, you are absolutely right. I should get the lock before I notify :) Nice catch there :D
But for `_http_thread_exit`, the lock is not meant to guard that; at least, I wrote the lock to guard `_http_requests`. Then I think I get where you are coming for, but I also think you missed the read in `CURLOPT_XFERINFOFUNCTION`. As this is also outside the lock.

Now I could add a lock in that callback, but that seems wasteful. It is just easier to make `_http_thread_exit` atomic. Although I doubt it matters for a `bool`, but that is yet-another-story :P

Now we could also add a signal on the queue to notify: job's done, but to prepare for a functional "cancel", I needed some logic in `CURLOPT_XFERINFOFUNCTION` anyway, and this felt cheaper.

At least, that is my reasoning .. now you will point to where my reasoning failed :D

frosch123 (approved):

PR #10479, opened 2023-02-14
Fix 5e14a20: [GS] Save league table title into string

SamuXarick:
## Motivation / Problem
Unnamed, 1955-06-17

GetEncodedText is changing the title to the last non-null text param. In the example screenshot, it's the footer.

## Description
Move title handling two lines lower as header and footer are stored in strings anyway

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10483, opened 2023-02-14
Codechange: [Script] Don't expose static buffers outside of ScriptText

glx22:
Supersede #10479.

## Motivation / Problem
As shown in #10479, in `ScriptLeagueTable::New()` encoded_title directly store the static buffer without copying its value first, so in the end encoded_title value will be the result of the last `GetEncodedText()` call (footer in the example given).

## Description
Convert the static buffer content into an `std::string` on return.
~~The return type can't be `std::string` directly due to the special handling of SCRIPT_TEXT_MAX_PARAMETERS (which in theory can't never happen because it's already checked when adding the parameters), so I kept the current mechanism using `std::optional<std::string>`.
I also extracted the "too many parameters" check from `EnforcePreconditionEncodedText()`, to allow checking the validity when an empty string is allowed (again not really needed but safer).~~ Removed `ScriptError::ERR_PRECONDITION_TOO_MANY_PARAMETERS` and replaced its handling with a fatal error, as it's really not supposed to happen.
Also used `std::string` to store RawString value, removing the manual stredup/free.
Updated the docs to reflect empty string being allowed when nullptr is.

## Checklist for review

### DIFF
--- a/src/script/api/script_text.cpp
+++ b/src/script/api/script_text.cpp
@@ -11,6 +11,7 @@
 #include "../../string_func.h"
 #include "../../strings_func.h"
 #include "script_text.hpp"
+#include "../script_fatalerror.hpp"
 #include "../../table/control_codes.h"
 
 #include "table/strings.h"
@@ -181,7 +182,8 @@ const char *ScriptText::GetEncodedText()
 	static char buf[1024];
 	int param_count = 0;
 	this->_GetEncodedText(buf, lastof(buf), param_count);
-	return (param_count > SCRIPT_TEXT_MAX_PARAMETERS) ? nullptr : buf;
+	if (param_count > SCRIPT_TEXT_MAX_PARAMETERS) throw Script_FatalError("A string had too many parameters");
+	return buf;
 }
 
 char *ScriptText::_GetEncodedText(char *p, char *lastofp, int &param_count)
@@ -208,8 +210,7 @@ char *ScriptText::_GetEncodedText(char *p, char *lastofp, int &param_count)
 
 const char *Text::GetDecodedText()
 {
-	const char *encoded_text = this->GetEncodedText();
-	if (encoded_text == nullptr) return nullptr;
+	const std::string &encoded_text = this->GetEncodedText();
 
 	static char buf[1024];
 	::SetDParamStr(0, encoded_text);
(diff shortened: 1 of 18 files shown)


src/script/api/script_text.cpp
```diff
@@ -17,15 +17,10 @@
 
 #include "../../safeguards.h"
 
-RawText::RawText(const char *text) : text(stredup(text))
+RawText::RawText(const char *text) : text(text)
```
frosch123, line 21: Can `text` be `nullptr`?

src/script/api/script_text.hpp
```diff
@@ -21,17 +23,17 @@ class Text : public ScriptObject {
 public:
 	/**
 	 * Convert a ScriptText to a normal string.
-	 * @return A string (in a static buffer), or nullptr.
+	 * @return A string, or <empty>.
 	 * @api -all
 	 */
-	virtual const char *GetEncodedText() = 0;
+	virtual std::optional<const std::string> GetEncodedText() = 0;
```
frosch123: The semantics of `optional<const T>` confuses me. The result is a value, no reference, so `const` is weird.

src/script/api/script_text.cpp
glx22: Based of my understanding of https://github.com/OpenTTD/OpenTTD/blob/master/cmake/scripts/SquirrelExport.cmake#L38-L47 it cannot be `nullptr`

src/script/api/script_text.hpp
glx22: If we replace the special handling of SCRIPT_TEXT_MAX_PARAMETERS by some kind of assertion check, it's not even needed to use std::optional. The number of parameters is already checked in constructor (throws "invalid parameter"), and AddParam() or SetParam() can't add more than the limit (no throw, the call just fails and says "unknown error").

Maybe I should just throw in AddParam() and SetParam() to give useful info, and also find a way to kill the script in case it managed to add more than SCRIPT_TEXT_MAX_PARAMETERS (the special handling requiring std::optional), or just assume it can't happen, but I don't like removing a safety check.

Edit: I replaced handling of SCRIPT_TEXT_MAX_PARAMETERS as a precondition error by a fatal error, as it makes more sense that way. This removes the need to use std::optional.

rubidium42 (approved):

PR #10485, opened 2023-02-15
Fix #10131: actually cancel downloads when pressing cancel 

### LINKED ISSUE #10131: [Bug]: Cancelling in-game content download does not actually cancel it.
### Version of OpenTTD

git master, Windows 11

### Expected result

Clicking on the cancel button when downloading content in game should cancel the download.

### Actual result

The cancel button closes the download window, but the download still continues behind the scenes.

While it is still download, the content window allows the content to be reselected, but attempting to download again results in an error "Could not decompress the download file".

This does not affect the background download, which will complete and eventually show that the content has been downloaded.

### Steps to reproduce

1. Select a large file to download (large is easier to intercept).
2. Cancel the download.
3. (Optional, try to select and download again...)
4. Wait a bit.
5. Download has completed despite being cancelled.

TrueBrain:
## Motivation / Problem

Pressing cancel didn't actually cancel, but kept downloading in the background. Which gave hilarious results if you try to download another file, as then the filepointers got mixed up, and the whole world got disappointed in you.

## Description

Introduced a "IsCancelled" state, to notify the backends they need to terminate as soon as they can. This to avoid sending pointers back and forth .. the backend now polls the callback object for its state. Not the best solution, but the current state of code is .. well, there are weird things in there.

While at it, found (and solved) various of other bugs. If requested, I have no problem making them into separate PRs. But in short:

- Finally did the much requested change to use `std::string` instead of some custom malloc/free for the content-body
- Download progress was not always starting at zero. While at it, also initialize it as zero. Not really important, but looks much cleaner.

I refrained from making any other changes / fixes .. this code kinda needs a redo anyway, but also on a fundamental level: the download list is fetched in full, every time. This is a long list by now, and is very slow. Ideally, we approach this smarter, either by categories, backend-searching, etc. So I didn't want to spend too much time on this code, and just fixed the problems I spotted.

## Checklist for review

### DIFF
--- a/src/network/core/http.h
+++ b/src/network/core/http.h
@@ -30,6 +30,15 @@ struct HTTPCallback {
 	 */
 	virtual void OnReceiveData(const char *data, size_t length) = 0;
 
+	/**
+	 * Check if there is a request to cancel the transfer.
+	 *
+	 * @return true iff the connection is cancelled.
+	 * @note Cancellations are never instant, and can take a bit of time to be processed.
+	 * The object needs to remain valid until the OnFailure() callback is called.
+	 */
+	virtual bool IsCancelled() const = 0;
+
 	/** Silentium */
 	virtual ~HTTPCallback() {}
 };
--- a/src/network/core/http.h
+++ b/src/network/core/http.h
@@ -42,9 +42,9 @@ public:
 	 *
 	 * @param uri      the URI to connect to (https://.../..).
 	 * @param callback the callback to send data back on.
-	 * @param data     optionally, the data we want to send. When set, this will be a POST request, otherwise a GET request.
+	 * @param data     the data we want to send. When non-empty, this will be a POST request, otherwise a GET request.
 	 */
-	static void Connect(const std::string &uri, HTTPCallback *callback, const char *data = nullptr);
+	static void Connect(const std::string &uri, HTTPCallback *callback, const std::string data = "");
 
 	/**
 	 * Do the receiving for all HTTP connections.
--- a/src/network/core/http_curl.cpp
+++ b/src/network/core/http_curl.cpp
@@ -167,8 +167,10 @@ void HttpThread()
 		 * do about this. */
 		curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 0L);
 		curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, +[](void *userdata, curl_off_t dltotal, curl_off_t dlnow, curl_off_t ultotal, curl_off_t ulnow) -> int {
-			return _http_thread_exit ? 1 : 0;
+			const HTTPCallback *callback = static_cast<HTTPCallback *>(userdata);
+			return (callback->IsCancelled() || _http_thread_exit) ? 1 : 0;
 		});
+		curl_easy_setopt(curl, CURLOPT_XFERINFODATA, request->callback);
 
 		/* Perform the request. */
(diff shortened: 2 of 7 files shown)


src/network/core/http_curl.cpp
```diff
@@ -36,24 +36,16 @@ class NetworkHTTPRequest {
 	 * @param callback the callback to send data back on.
 	 * @param data     optionally, the data we want to send. When set, this will be a POST request, otherwise a GET request.
 	 */
-	NetworkHTTPRequest(const std::string &uri, HTTPCallback *callback, const char *data = nullptr) :
+	NetworkHTTPRequest(const std::string &uri, HTTPCallback *callback, std::string &data) :
```
frosch123: Why a mutable string reference?

src/network/core/http_curl.cpp
```diff
@@ -36,24 +36,16 @@ class NetworkHTTPRequest {
 	 * @param callback the callback to send data back on.
 	 * @param data     optionally, the data we want to send. When set, this will be a POST request, otherwise a GET request.
```
frosch123: "optionally" now refers to an empty string?

src/network/core/http.h
```diff
@@ -30,6 +30,15 @@ struct HTTPCallback {
...
 	virtual void OnReceiveData(const char *data, size_t length) = 0;
 
+	/**
+	 * Check if there is a request to cancel the transfer.
+	 *
+	 * @return true iff the connection is cancelled.
+	 * @note Cancellations are never instant, and can take a bit of time to be processed.
+	 * The object needs to remain valid until the OnFailure() callback is called.
+	 */
+	virtual bool IsCancelled() = 0;
```
frosch123: Could be "const" for the weak :)

src/network/core/http_winhttp.cpp
```diff
@@ -171,20 +173,22 @@ void NetworkHTTPRequest::WinHttpCallback(DWORD code, void *info, DWORD length)
...
-			this->callback->OnFailure();
 			this->finished = true;
+			this->callback->OnFailure();
 			return;
 	}
 }
 
 static void CALLBACK StaticWinHttpCallback(HINTERNET handle, DWORD_PTR context, DWORD code, void *info, DWORD length)
 {
+	if (context == NULL) return;
```
frosch123: Maybe `nullptr` makes mingw32 happy?

src/network/network_content.cpp
```diff
@@ -591,6 +594,13 @@ void ClientNetworkContentSocketHandler::OnFailure()
 		fclose(this->curFile);
```
frosch123, line 594: Should the file be deleted on failure?

src/network/core/http_winhttp.cpp
TrueBrain: a `nullptr` is not valid, as it is a `DWORD_PTR`. So .. I am afraid I will have to use `0` .........

src/network/network_content.cpp
TrueBrain: This wasn't done before, so I didn't dare to touch it. I would think that makes sense, on the other hand, leaving it behind allows you to diagnose errors .. I was a bit meh about it, either way. But not for this PR at least, I would think :)

src/network/core/http.h
TrueBrain: You get a const, and you get a const, and you get a const!

frosch123 (approved):

PR #9038, opened 2021-04-14
Add test framework and some basic unit tests

WimLeflere:
## Motivation

Tests for new and existing code will make development & refactoring easier.
And will make it easier to detect regressions and unintended changes.

## Description

Added Catch2 header only test framework.
Math tests from #7796

## Limitations

Prototype implementation to get feedback

Catch2 doesn't seem to be available in the Ubuntu 20.04 package repository
CMake FetchContent could be used instead: https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md

My CMake knowledge is to limited and/or the OpenTTD CMake setup is to advanced for me to do a correct test target integration. The current solution with CMake arguments is not ideal.
Other options, in 'quality' order
1. create a openttd_lib which is used by openttd and openttd_test
has the advantage of a single compile for most of the sources
2. re-use the sources list for openttd in openttd_test
two compiles, but no manual action needed to keep in sync
3. specify the required source files for openttd_test separately
lots of dependencies make it hard to include all the required files, manual updates needed for new files

PR #10491, opened 2023-02-18
Fix #10222: Adjust line drawing algorithm

### LINKED ISSUE #10222: [Bug]: Even-width horizontal/vertical lines have intial 'head' in wrong place.
### Version of OpenTTD

git master, Windows 11

### Expected result

Even-width horizontal/vertical lines should have no 'head'.

### Actual result

The first pixel of an even-width horizontal or vertical line is off-by-one relative to the rest of the line:

image

### Steps to reproduce

1. Write code to draw an even-width line. :-)
2. Observe line.

nikolas:
This adjusts the line drawing algorithm by setting up a few variables in the loop before the pixel is set, rather than after. This change appears to fix this off-by-one error.

I was testing with a few lines in the console_gui, shown below.

Before change:
Screenshot_2023-02-17_23-36-38

After change:
Screenshot_2023-02-17_23-40-32

Closes #10222

### DIFF
--- a/src/blitter/common.hpp
+++ b/src/blitter/common.hpp
@@ -109,11 +109,6 @@ void Blitter::DrawLineGeneric(int x1, int y1, int x2, int y2, int screen_width, 
 		}
 
 		while (x1 != x2) {
-			if (dash_count < dash) {
-				for (int y = y_low; y != y_high; y += stepy) {
-					if (y >= 0 && y < screen_height) set_pixel(x1, y);
-				}
-			}
 			if (frac_low >= 0) {
 				y_low += stepy;
 				frac_low -= dx;
@@ -122,6 +117,12 @@ void Blitter::DrawLineGeneric(int x1, int y1, int x2, int y2, int screen_width, 
 				y_high += stepy;
 				frac_high -= dx;
 			}
+			if (dash_count < dash) {
+				for (int y = y_low; y != y_high; y += stepy) {
+					if (y >= 0 && y < screen_height) set_pixel(x1, y);
+				}
+			}
+
 			x1++;
 			frac_low += dy;
 			frac_high += dy;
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):

ldpl:
Does it preserve the starting/ending point though?

2TallTyler:
I don't see why it wouldn't. The moved code only changes the iterator scoped within each for loop.

ldpl:
Because that code moves the drawing boundaries when changing x to the next column (or y to row for second loop).
So placing the drawing loop in between essentially means using half-computed new boundaries instead. 
As I see it that can only be correct if those boundaries were initially incorrect in that exact half-step way, which seems unlikely unless explained.
In other words it's unlikely that just randomly moving code around could accidentally make it correct and I see neither explanation nor sufficient testing to prove it right.

nikolas:
@ldpl from my testing this addresses the bug noted here: #10222 

I didn't move the code around randomly - it was an intentional adjustment to the line drawing algorithm, which is admittedly pretty complex.

The start and and end point looked fine from my testing, but I'll do some more thorough manual testing. It is, if nothing else, an interesting area of the code.

ldpl:
Well, at the very least I can confirm it's still broken:
Screenshot from 2023-02-21 01-47-17

```
	GfxFillRect(100, 100, 120, 120, PC_GREY);
	GfxDrawLine(100, 100, 120, 120, PC_YELLOW, 1, 0);
```

nikolas:
Ah okay - interesting. So the bug these changes fixed happens only in the case where the line has an even-number width. I'm seeing the same rendering of this diagonal yellow line both before and after these changes.

But I will look into this as it would be good to fix this too.

nikolas:
This sort of elaborate algorithm that handles all sorts of inputs, demanding exacting results, really calls for some unit tests. I'd love to set something like that up here.

2TallTyler:
Relevant to unit tests: #9038.

ldpl:
And this shows an actual regression:
Screenshot from 2023-02-21 02-20-07
```
	GfxFillRect(100, 100, 120, 120, PC_GREY);
	GfxDrawLine(120, 120, 100, 100, PC_RED, 1, 0);
	GfxDrawLine(110, 120, 110, 100, PC_RED, 1, 0);
	GfxDrawLine(120, 110, 100, 110, PC_RED, 1, 0);
	GfxDrawLine(120, 100, 100, 120, PC_RED, 1, 0);
```

nikolas:
Thank you for pointing this out - I didn't do thorough enough testing with single-pixel diagonal lines.

PR #10494, opened 2023-02-19
Add: Use specific error message when vehicle cannot go to station/waypoint

### LINKED ISSUE #9807: [Bug]: Error message when trying to order articulated vehicles to non-drive-thru stop is unclear
### Version of OpenTTD

12.1

### Expected result

I should be able to send a bus using GO TO to a station, instead...

### Actual result

When sending a bus to a station that is attached to another station (like train station) I get the message: Can't insert new order ... vehicle can't go to that station. 
It happend few times already. 

image

Any tips on how to get around the error? 

### Steps to reproduce

[SAVE](https://drive.google.com/file/d/1vRI8pLWzYLLOQHdfElc5TLVc7Sy384nz/view?usp=sharing)

1. Try sending bus number 58 to a station Strzebielin Wiekszy Woods. 
2. You'll gen an error message.

2TallTyler:
## Motivation / Problem

Players are often confused when they try and send an articulated road vehicle to a bay station, because the error message does not explain why the vehicle cannot use the station. See #9807.

The same goes for sending buses to truck stations and vice versa. 

## Description

Upstreams a JGRPP commit to add a specific error message, and builds off a commit to add a two-string error message.

This requires explicitly casting a few CommandCost arguments to determine which overload to use. I also added a ShowErrorMessage overload which simply takes a CommandCost instead of individual attributes of that object, per LordAro's request.

Closes #9807.

## Limitations

The "no buoy" and "no waypoint" should almost never be reached, but we already have error handling for this, so why not add a couple strings to be descriptive?

## Checklist for review

### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -259,7 +259,7 @@ void CommandHelperBase::InternalPostResult(const CommandCost &res, TileIndex til
 	if (res.Failed()) {
 		/* Only show the error when it's for us. */
 		if (estimate_only || (IsLocalCompany() && err_message != 0 && my_cmd)) {
-			ShowErrorMessage(err_message, res.GetErrorMessage(), WL_INFO, x, y, res.GetTextRefStackGRF(), res.GetTextRefStackSize(), res.GetTextRefStack());
+			ShowErrorMessage(err_message, x, y, res);
 		}
 	} else if (estimate_only) {
 		ShowEstimatedCostOrIncome(res.GetCost(), x, y);
--- a/src/command_type.h
+++ b/src/command_type.h
@@ -22,12 +22,13 @@ struct GRFFile;
  * a possible error message/state together.
  */
 class CommandCost {
-	ExpensesType expense_type; ///< the type of expence as shown on the finances view
-	Money cost;       ///< The cost of this action
-	StringID message; ///< Warning message for when success is unset
-	bool success;     ///< Whether the command went fine up to this moment
-	const GRFFile *textref_stack_grffile; ///< NewGRF providing the #TextRefStack content.
-	uint textref_stack_size;   ///< Number of uint32 values to put on the #TextRefStack for the error message.
+	ExpensesType expense_type;                  ///< the type of expence as shown on the finances view
+	Money cost;                                 ///< The cost of this action
+	StringID message;                           ///< Warning message for when success is unset
+	bool success;                               ///< Whether the command went fine up to this moment
+	const GRFFile *textref_stack_grffile;       ///< NewGRF providing the #TextRefStack content.
+	uint textref_stack_size;                    ///< Number of uint32 values to put on the #TextRefStack for the error message.
+	StringID extra_message = INVALID_STRING_ID; ///< Additional warning message for when success is unset
 
 	static uint32 textref_stack[16];
 
@@ -38,9 +39,9 @@ public:
 	CommandCost() : expense_type(INVALID_EXPENSES), cost(0), message(INVALID_STRING_ID), success(true), textref_stack_grffile(nullptr), textref_stack_size(0) {}
 
 	/**
-	 * Creates a command return value the is failed with the given message
+	 * Creates a command return value with one, or optionally two, error message strings.
 	 */
-	explicit CommandCost(StringID msg) : expense_type(INVALID_EXPENSES), cost(0), message(msg), success(false), textref_stack_grffile(nullptr), textref_stack_size(0) {}
+	explicit CommandCost(StringID msg, StringID extra_msg = INVALID_STRING_ID) : expense_type(INVALID_EXPENSES), cost(0), message(msg), success(false), textref_stack_grffile(nullptr), textref_stack_size(0), extra_message(extra_msg) {}
 
 	/**
 	 * Creates a command cost with given expense type and start cost of 0
@@ -98,11 +99,12 @@ public:
 	 * Makes this #CommandCost behave like an error command.
 	 * @param message The error message.
 	 */
-	void MakeError(StringID message)
+	void MakeError(StringID message, StringID extra_message = INVALID_STRING_ID)
 	{
 		assert(message != INVALID_STRING_ID);
 		this->success = false;
 		this->message = message;
+		this->extra_message = extra_message;
 	}
 
 	void UseTextRefStack(const GRFFile *grffile, uint num_registers);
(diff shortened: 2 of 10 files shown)


src/lang/english.txt
```diff
@@ -5084,6 +5084,20 @@ STR_ERROR_CAN_T_COPY_ORDER_LIST                                 :{WHITE}Can't co
 STR_ERROR_TOO_FAR_FROM_PREVIOUS_DESTINATION                     :{WHITE}... too far from previous destination
 STR_ERROR_AIRCRAFT_NOT_ENOUGH_RANGE                             :{WHITE}... aircraft has not enough range
 
+# Extra messages which go on the third line of errors,
+# explaining why orders failed
```
LordAro: no need for line break

src/lang/english.txt
```diff
@@ -5084,6 +5084,20 @@ STR_ERROR_CAN_T_COPY_ORDER_LIST                                 :{WHITE}Can't co
...
+STR_ERROR_NO_RAIL_STATION                                       :{WHITE}There is no railway station
+STR_ERROR_NO_BUS_STATION                                        :{WHITE}There is no bus station
+STR_ERROR_NO_TRUCK_STATION                                      :{WHITE}There is no lorry station
+STR_ERROR_NO_DOCK                                               :{WHITE}There is no dock
+STR_ERROR_NO_AIRPORT                                            :{WHITE}There is no airport/heliport
+STR_ERROR_NO_STOP_COMPATIBLE_ROAD_TYPE                          :{WHITE}There are no stops with a compatible road/tram type
```
LordAro: Feels like this one should be split into explicit road & tram strings?

src/lang/english.txt
```diff
@@ -5084,6 +5084,20 @@ STR_ERROR_CAN_T_COPY_ORDER_LIST                                 :{WHITE}Can't co
...
+STR_ERROR_NO_STOP_ARTIC_VEH                                     :{WHITE}There are no stops which are suitable for articulated road vehicles.{}Articulated road vehicles require a drive-through stop, not a bay stop
```
LordAro: Is "bay stop" the usual term? I'm not aware of it used anywhere else

LordAro: Seems no real need for abbreviate articulated. similar for helicopters below

src/command_type.h
```diff
@@ -42,6 +43,16 @@ class CommandCost {
 	 */
 	explicit CommandCost(StringID msg) : expense_type(INVALID_EXPENSES), cost(0), message(msg), success(false), textref_stack_grffile(nullptr), textref_stack_size(0) {}
 
+	/**
+	 * Creates a command return value with two lines of detailed error messages.
+	 */
+	static CommandCost DualErrorMessage(StringID msg, StringID extra_msg)
```
LordAro: I wonder if instead of this static function + initialising extra_message in the class declaration, it'd be better to just use a constructor.

you could even just use the existing constructor with a default INVALID_STRING_ID parameter

src/command.cpp
```diff
@@ -259,7 +259,7 @@ void CommandHelperBase::InternalPostResult(const CommandCost &res, TileIndex til
 	if (res.Failed()) {
 		/* Only show the error when it's for us. */
 		if (estimate_only || (IsLocalCompany() && err_message != 0 && my_cmd)) {
-			ShowErrorMessage(err_message, res.GetErrorMessage(), WL_INFO, x, y, res.GetTextRefStackGRF(), res.GetTextRefStackSize(), res.GetTextRefStack());
+			ShowErrorMessage(err_message, res.GetErrorMessage(), WL_INFO, x, y, res.GetTextRefStackGRF(), res.GetTextRefStackSize(), res.GetTextRefStack(), res.GetExtraErrorMessage());
```
LordAro: possible extra work: Add a version of ShowErrorMessage that just takes in a commandcost as a parameter. 5 separate attributes of the same object seems excessive

src/lang/english.txt
2TallTyler: Hmm, the tooltip for freight road stops says "Lorry Loading Bay," as does the tooltip for choosing the orientation. They're also called that on the smallmap. Bus stations are just called "bus station."

If you have a better suggestion I'd welcome it. A "bay stop" makes sense to me and I'm not sure if we call it anything else elsewhere.

src/order_cmd.cpp
```diff
@@ -1576,10 +1576,11 @@ CommandCost CmdCloneOrder(DoCommandFlag flags, CloneOptions action, VehicleID ve
 
 			/* Trucks can't copy all the orders from busses (and visa versa),
 			 * and neither can helicopters and aircraft. */
+			Station *st;
```
rubidium42: Why is this not combined at line 1581? It doesn't seem to be used after the loop.

src/vehicle.cpp
```diff
@@ -2903,6 +2903,57 @@ bool CanVehicleUseStation(const Vehicle *v, const Station *st)
...
+
+			for (; rs != nullptr; rs = rs->next) {
+				/* The vehicle is articulated and can therefore not go to a standard road stop. */
+				if (IsStandardRoadStopTile(rs->xy) && rv->HasArticulatedPart()) {
+					err = STR_ERROR_NO_STOP_ARTICULATED_VEHICLE;
+					continue;
+				}
+				/* The vehicle cannot go to this roadstop (different roadtype) */
+				if (!HasTileAnyRoadType(rs->xy, rv->compatible_roadtypes)) {
+					return RoadTypeIsRoad(rv->roadtype) ? STR_ERROR_NO_STOP_COMPATIBLE_ROAD_TYPE : STR_ERROR_NO_STOP_COMPATIBLE_TRAM_TYPE;
```
rubidium42: This is broken; I do not think there should be a return here, but rather a continue like the case before. Or maybe the logic of this part of the function needs to be changed.

Start a game with trams and articulated road vehicles (e.g. generic trams + HEQS). Build a drive through tram and then add a bay station road. Build an articulated road vehicle, and create an order for that station. Result: "no compatible road type", but there is... it's just a bay.

Also, when changing this code, think about what should take precedence. If you change this to work the same as for the articulated vehicle, then a situation with an articulated bus and a bay bus and drive through tram can return either message depending on the order of the underlying road stops.

rubidium42 (changes requested):

rubidium42 (approved):

PR #10507, opened 2023-02-24
Fix: Don't send unused tile field over the network

ldpl:
## Motivation / Problem

CommandPacket contains a TileIndex field that is rarely used because it's either 0 or duplicates the first tile argument that is sent separately. Also tile can be passed to any command, even if it doesn't have a tile argument and has no use for it or, on contrary, it can be ommited when it's actually used. This makes it harder to understand code and simpler to make mistakes.

## Description

This PR removes the tile field from CommandPacket and deals with the command handling according to command type:

- Has first TileIndex argument. cp->tile just duplicates the argument so no additional handling is needed.
- Has implicit TileIndex argument. Mark these commands with the new CMD_LOCATION flag to make it more explicit. That makes sure tile is always sent as the first argument in the packet and Post calls so it can't be missed (like in CcBuildWagon). Tile is mainly used for the error location and is handled internally, so it's not included in command handler or Do function arguments.
- Has no tile argument. Just remove unused cp->tile and make sure tile can't be passed by accident (like in OnDragDrop_Vehicle).

## Limitations

This PR focuses on network side of things and doesn't go further in solving TileIndex confusion like 
- Using both 0 and INVALID_TILE
- Implicit validation of first (and only first) TileIndex argument.
- Implicit error location handling. 
- It's mostly unnecessary to pass location over the network just to show the error.

## Checklist for review

* **The bug fix is important enough to be backported? (label: 'backport requested')**

### DIFF
--- a/src/command_func.h
+++ b/src/command_func.h
@@ -37,7 +37,7 @@ static const CommandCost CMD_ERROR = CommandCost(INVALID_STRING_ID);
  */
 #define return_cmd_error(errcode) return CommandCost(errcode);
 
-void NetworkSendCommand(Commands cmd, StringID err_message, CommandCallback *callback, CompanyID company, TileIndex location, const CommandDataBuffer &cmd_data);
+void NetworkSendCommand(Commands cmd, StringID err_message, CommandCallback *callback, CompanyID company, const CommandDataBuffer &cmd_data);
 
 bool IsValidCommand(Commands cmd);
 CommandFlags GetCommandFlags(Commands cmd);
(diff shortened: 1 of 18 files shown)


src/command_func.h
```diff
@@ -446,8 +434,13 @@ struct CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), true> : protected C
 template <Commands Tcmd, typename Tret, typename... Targs>
 struct CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), false> : CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), true>
 {
-	/* Import Post overloads from our base class. */
-	using CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), true>::Post;
+	/* Do not allow Post without explicit location. */
+	static inline bool Post(StringID err_message, Targs... args)=delete;
```
PeterN: `) = delete` to match existing uses.

src/command_func.h
```diff
@@ -492,6 +484,7 @@ struct CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), false> : CommandHel
 #endif
 
 template <Commands Tcmd>
-using Command = CommandHelper<Tcmd, typename CommandTraits<Tcmd>::ProcType, std::is_same_v<TileIndex, std::tuple_element_t<0, typename CommandTraits<Tcmd>::Args>>>;
+
```
michicc: Sneaky newline.

michicc (approved):

PR #10512, opened 2023-02-26
Doc: Add Coding Style into the repo for easier access

Limyx826:
## Motivation / Problem
Originally, I am working on adding SGD into the game with VS Code. However, when I am doing some changes to the code, I found that doing commit message and hooks and whatnot are somewhat tedious that you need to refer to the wiki online rather than a document that already in the repo. 

## Description
~I add .vscode into .gitignore.~
~Adding SGD currency into the game.~
Transcribe Coding Style from the wiki to .md file. 

Rendered: https://github.com/Limyx826/OpenTTD/blob/newbr/CODINGSTYLE.md

## Limitations
No

## Checklist for review

* The bug fix is important enough to be backported? (label: 'backport requested') ❎
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md) ❎
* This PR affects the save game format? (label 'savegame upgrade') ❎
* This PR affects the GS/AI API? (label 'needs review: Script API') ❎
* This PR affects the NewGRF API? (label 'needs review: NewGRF') ❎

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -1,0 +1,490 @@
+**Why a set coding style is important** <br>
+This project is an open source project. To get open source working as intended, many people should be able to comprehend the code. This implies that there is a well documented and uniform flow of code.
+That does not necessarily mean that a contributor should not write optimised yet cryptic code - one should always care about performance.  However, other developers need to understand the code which means complicated parts of code **must** have good documentation.
+
+**Why we require that EVERYTHING is documented**<br>
+What is simple to some might appear very complicated to others. Documentation helps these others. Anyone should be able to improve the code. But the main reason is, that when patch contributors want their patches added to the common codebase, the code needs to be reviewed by one or more developers. Documentation makes reviewing much easier.
+
+## Coding style for OpenTTD
+### Functions
+* Function names use [CamelCase](http://www.wikipedia.org/wiki/Camelcase) without underscores.
+* Opening curly bracket **{** for a function starts on the next line.
+* Use Foo() instead of Foo(void).
+```c++
+void ThisIsAFunction()
+{
+}
+```
+
+### Variables
+* Variable names are all lowercase, and use "_" as separator.
+* Global variables are preceded by an underscore. ("_") Use descriptive names because leading underscores are often used for system / compiler variables.
+* Own members of classes should always be referenced using "this->"
+* Pointers and references should have their reference symbol next to the name (compatibility with current code).
+* Variables that are declared below one another should have their type, name or reference operator, and assignment operator aligned by spaces.
+* There are set names for many variables. Those are (but not limited to): Vehicle *u, *v, *w; Station *st; Town *t; Window *w; Engine *e.
+* For multiple instances, use numbers "*t1, *t2" or postfixes "*st_from, *st_to".
+* Declare variables upon first usage.
+* Declare iterators in their loop.
+* There is a space between '*' and 'const' in "const pointers"
+```c++
+int     number         = 10;
+Vehicle *u_first_wagon = v->next;
+int     value          = v->value;
+
+uint32 _global_variable = 3750;
+
+static const char * const _global_array[] = {
+	"first string",
+	"second string",
+	"another string",
+	"last string followed by comma",
+};
+
+protected:
+	char protected_array[10];
+
+for (int i = 0;;);
+```
+
+* Give the variables expedient names, this increases the readability of the code
+```c++
+bool is_maglev_train = true;
+if (!is_maglev_train) DoSomething();
+```
+
+* Some people like to introduce copies of variables to increase readability, which can waste memory. However, in some cases, especially in code pieces which are often called, it makes sense to cache some calculated variables.
+```c++
+/* Unoptimized code:
+ * foo is not touched inside the loop!
+ */
+for (uint8 i = 0; i < 100000; i++) {
+	/* Do something */
+	if (value_to_check == (foo * 4) % 5 + 6) DoSomething();
+	/* Do something else */
+}
+
+/* Better:
+ * The used value of foo is calculated outside the loop.
+ */
+const uint32 bar = (foo * 4) % 5 + 6;
+for (uint8 i = 0; i < 100000; i++) {
+	/* Do something */
+	if (value_to_check == bar) DoSomething();
+	/* Do something else */
+}
+```
+
+### Enumerations / static consts
+* Enumerations store integers that belong logically together (railtypes, string IDs, etc.).
+* Enumeration names also use CamelCase.
+* Unscoped enumerators are all caps with "_" between the words.
+* Scoped enumerators are use CamelCase.
+* Enums are not used to store single numbers.
+* Enums have consecutive numbers OR
+* Enums have consecutive powers of two. Powers of two (bits) are written in hex or with the shift operator.
+* Enums may have special enumerators: "_BEGIN", "_END", and "INVALID_").  See example.
+* The invalid always has 0xFF, 0xFFFF, 0xFFFFFFFF as a value.
+* Other special values are consecutively less than the invalid.
+* Variables that hold enumerators should have the type of the enumeration.
+```c++
+enum DiagDirection {
+	DIAGDIR_BEGIN = 0,
+	DIAGDIR_NE  = 0,
+	DIAGDIR_SE  = 1,
+	DIAGDIR_SW  = 2,
+	DIAGDIR_NW  = 3,
+	DIAGDIR_END,
+	INVALID_DIAGDIR = 0xFF,
+	BROKEN_DIAGDIR = 0xFE,
+};
+
+enum {
+	DEPOT_SERVICE       = (1 << 0),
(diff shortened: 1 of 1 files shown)


TrueBrain:
This seems to be two separate changes; we very much prefer that those are also two separate PRs.

This simply as we might agree with one, and not with the other. Or one is easier to review than the other. Etc etc. Having them combined makes our life a lot more complex :)

Additionally, please only change `english.txt` translations. Other translations are handled by our webtranslator, which is a bit sensitive if you take away his job :(

Cheers!

FLHerne:
I agree with @TrueBrain's comment, the SGD commit should be a separate PR.

In your latest push to this PR, you now have that commit and a separate commit reverting it, which isn't quite right either. ;-)

The coding style doc should go with the other more-detailed development documentation as `docs/coding_style.md`, the few files in the root directory are more of an index.

Please put "Singapore dollar" or "currency" in the SGD commit message because not everyone knows what "SGD" means out of context.

Note the commit checker failure, looks like some trivial whitespace issue.

Limyx826:
I see whether I can cherry pick out the commits and rebase into this branch otherwise I will create a new branch and pull request.

FLHerne:
Also typo: 'gitinnore' in one commit message.

FLHerne:
The usual way to drop commits from a branch would be `git rebase -i master` and delete the lines corresponding to the unwanted commits. Make sure you preserve them on another branch to avoid losing work.

LordAro:
In general, we don't like "user specific" things to be added to the project gitignore - they should by preference be added to the user's [global gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) instead. I've seen another trick where you can just add a .gitignore file inside the .vscode folder that ignores everything, including itself. But the global gitignore seems the "proper" solution.

(`.vs` is different because it's used by VS as part of the Windows build ...I think)

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+	SE_END,              ///< End of the enumeration, used for iterations
+};
+```
+
+### Control flow
+* Put a space before the parentheses in **if**, **switch**, **for** and **while** statements.
+* Use curly braces and put the contained statements on their own lines (e.g., don't put them directly after the **if**).
+* Opening curly bracket **{** stays on the first line, closing curly bracket **}** gets a line to itself (except for the **}** preceding an **else**, which should be on the same line as the **else**).
+* When only a single statement is contained, the brackets can be omitted. In this case, put the single statement on the same line as the preceding keyword (**if**, **while**, etc.). Note that this is only allowed for if statements without an **else** clause.
+* All fall throughs must be documented, using a FALL THROUGH comment.
```
rubidium42: This is now done with the `FALLTHROUGH;` define/macro, not a comment. This adds the compiler appropriate attribute that a fall through is expected.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+
+* If you are writing one or more template class in the dedicated header file, use file.hpp for its name instead of file.h. This will let others know that it is template library (includes also implementation), not just header with declarations.
+
+### Other important rules
+* Put a space before and after binary operators: "a + b", "a == b", "a & b", "a <<= b", etc.. Exceptions are ".", "->" and "[]" (no spaces) and "," (just space after it).
+* Put parenthesis where it improves readability: "*(b++)" instead of "*b++", and "if ((a & b) && c == 2)" instead of "if (a & b && c == 2)".
+* Do not put external declarations in implementation (i.e. cpp) files.
+* Use const where possible.
+* Do not typedef enums and structs.
+* Don't treat non-flags as flags: use "if (char_pointer != NULL && *char_pointer != '\0')", not "if (char_pointer && *char_pointer)".
```
rubidium42: `NULL` is now `nullptr`. Also twice on the next line.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+* Use "free(p)" instead of "if (p != NULL) free(p)". "free(NULL)" doesn't hurt anyone.
+* No trailing whitespace. The svn server will not allow tabs or space on the end of lines.
```
rubidium42: We've not used svn for a long time, though one of the github workflows checks for trailing whitespace and some situations of wrong tabs.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+```c++
+void ThisIsAFunction()
+{
+}
+```
+
+### Variables
+* Variable names are all lowercase, and use "_" as separator.
+* Global variables are preceded by an underscore. ("_") Use descriptive names because leading underscores are often used for system / compiler variables.
+* own members of classes should always be referenced using "this->"
```
rubidium42: Capitalisation of `own`.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+| **@see** | Starts a paragraph where one or more cross-references to classes, functions, methods, variables, files or URL may be specified. Two names joined by either :: or # are understood as referring to a class and one of its members. One of several overloaded methods or constructors may be selected by including a parenthesized list of argument types after the method name. [Here](http://www.stack.nl/~dimitri/doxygen/autolink.html) you can find detailed information about this feature. | @see OtherFunc() |
+| **@b** | Displays the following word using a bold font. Equivalent to &lt;b&gt;word&lt;/b&gt;. To put multiple words in bold use &lt;b&gt;multiple words&lt;/b&gt;.| ...@b this and @b that... |
```
rubidium42, line 360: Are all these HTML encoding really needed in markdown? As far as I can find markdown automatically converts these to their HTML encoding when preparing for visualisation.
Also on the next line and maybe in other places?

rubidium42 (changes requested):
There are some things wrong in the coding style, compared to the actual coding style. Could you also fix them in ther source of this text?
Please remove the .gitignore change from this PR.

And given this is a copy of some other source, how are we going to maintain that they are in sync? Should the other one maybe link to github if this gets merged?

CODINGSTYLE.md
Limyx826: I ain't sure the counterpart of those encoding in markdown. So if you have any idea of what these are, please let me know and I will make the change.

Limyx826:
Must have been something wrong with the commit when I remove the gitignore one.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+
+### Other important rules
+* Put a space before and after binary operators: "a + b", "a == b", "a & b", "a <<= b", etc.. Exceptions are ".", "->" and "[]" (no spaces) and "," (just space after it).
+* Put parenthesis where it improves readability: "*(b++)" instead of "*b++", and "if ((a & b) && c == 2)" instead of "if (a & b && c == 2)".
+* Do not put external declarations in implementation (i.e. cpp) files.
+* Use const where possible.
+* Do not typedef enums and structs.
+* Don't treat non-flags as flags: use "if (char_pointer != nullptr && *char_pointer != '\0')", not "if (char_pointer && *char_pointer)".
+* Use "free(p)" instead of "if (p != nullptr) free(p)". "free(nullptr)" doesn't hurt anyone.
+* No trailing whitespace. The github workflows will not allow tabs or space on the end of lines.
```
2TallTyler:
```suggestion
* No trailing whitespace. The GitHub workflows will not allow tabs or space on the end of lines.
```

Limyx826:
> And given this is a copy of some other source, how are we going to maintain that they are in sync? Should the other one maybe link to github if this gets merged?

I think so as much, the wiki should link back to GitHub version. As far as I know OpenTTD wiki has not been updated since the change in server<sup>I think?</sup>, also having a local copy for the contributors is better than having an internet connection to access the coding style.

2TallTyler (commented):
You'll want to [squash your commits](https://stackoverflow.com/questions/44173558/how-to-squash-with-git-rebase-i) using an interactive rebase.

I agree that the Wiki should link to the official docs in this GitHub repo.

CODINGSTYLE.md
```diff
@@ -0,0 +1,490 @@
...
+
+### Enumerations / static consts
+* Enumerations store integers that belong logically together (railtypes, string IDs, etc.).
+* Enumeration names also use CamelCase.
+* Unscoped enumerators are all caps with "_" between the words.
+* Scoped enumerators are use CamelCase.
+* Enums are not used to store single numbers.
+* Enums have consecutive numbers OR
+* Enums have consecutive powers of two. Powers of two (bits) are written in hex or with the shift operator.
+* Enums may have special enumerators: "_BEGIN", "_END", and "INVALID_").  See example.
```
rubidium42: When rendering the current variant in GitHub it makes 'END", and "INVALID' italic, and more importantly it removes the underscores. Maybe reordering them solves the issue?
```suggestion
* Enums may have special enumerators: "INVALID_", "_BEGIN", and "_END").  See example.
```

rubidium42 (approved):
Please squash on merge.

PR #10513, opened 2023-02-26
Fix: build on platforms without native atomic

hack3ric:
## Motivation / Problem

Platforms/compilers like riscv64 GCC does not have proper atomic support, so linking against libatomic is needed in order to build.

## Description

This PR introduces `CheckAtomic.cmake` from LLVM to automatically detect the level of support of atomic types. If it doesn't support then `-latomic` is automatically appended to build flags.

## Limitations

No limitation as far as I know.

## Checklist for review

### DIFF
--- a/cmake/CheckAtomic.cmake
+++ b/cmake/CheckAtomic.cmake
@@ -1,0 +1,87 @@
+# atomic builtins are required for threading support.
+
+INCLUDE(CheckCXXSourceCompiles)
+INCLUDE(CheckLibraryExists)
+
+# Sometimes linking against libatomic is required for atomic ops, if
+# the platform doesn't support lock-free atomics.
+
+function(check_working_cxx_atomics varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++17")
+  check_cxx_source_compiles("
+#include <atomic>
+std::atomic<int> x;
+std::atomic<short> y;
+std::atomic<char> z;
+int main() {
+  ++z;
+  ++y;
+  return ++x;
+}
+" ${varname})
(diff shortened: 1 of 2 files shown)


cmake/CheckAtomic.cmake
```diff
@@ -0,0 +1,87 @@
...
+
+INCLUDE(CheckCXXSourceCompiles)
+INCLUDE(CheckLibraryExists)
+
+# Sometimes linking against libatomic is required for atomic ops, if
+# the platform doesn't support lock-free atomics.
+
+function(check_working_cxx_atomics varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++11")
```
LordAro: Is the fact that we use c++17 relevant here? I guess probably not, but maybe it should be consistent?

cmake/CheckAtomic.cmake
```diff
@@ -0,0 +1,87 @@
...
+  CHECK_CXX_SOURCE_COMPILES("
```
LordAro: function calls should be lowercase

cmake/CheckAtomic.cmake
```diff
@@ -0,0 +1,87 @@
...
+  return ++x;
+}
+" ${varname})
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endfunction(check_working_cxx_atomics)
+
+function(check_working_cxx_atomics64 varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "-std=c++11 ${CMAKE_REQUIRED_FLAGS}")
+  CHECK_CXX_SOURCE_COMPILES("
```
LordAro: lowercase (also consistent with the rest of the file)

LordAro (approved):
LGTM

PR #8984, opened 2021-04-10
Feature: build vehicle name filter

perezdidac:
## Motivation / Problem

As discussed in https://github.com/OpenTTD/OpenTTD/discussions/8723, users will benefit from having the option to filter the Build Vehicle list by typing a string. This would be especially useful when playing with NewGRF that include tons of vehicles.

## Description

This change adds a textbox to the Build Vehicle window so users can enter a keyword to find vehicles that match that keyboard. This works for all the types of vehicles and it's compatible with the existing filters.

This is what it looks like:

image

The text boxes do not immediately get focus to not break user's hotkeys.

## Limitations

Nope.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10519, opened 2023-02-28
Feature: Filter engine build menu by name and NewGRF extra text

2TallTyler:
## Motivation / Problem

In #8984, @perezdidac started the work of adding a text filter to the build vehicle menu, but hasn't been touched in 10 months. Let's finish it.

## Description

Implements #8984 with a change to put the editbox below the other lines:

filter

I added the ability to search NewGRF extra text, which is often used to describe a vehicle's role:

extra_text

It also works with the vehicle name callback added in #10399 (there's an example trainset in that PR):

name_callback

If variants are used (unreleased Iron Horse 2 in this screenshot), a subvariant that meets the filter, nested inside a parent does not, will show up underneath its greyed-out parent variant, collapsed by default.

variants

Closes #8984.

## Limitations

I tried swapping the cargo filter dropdown and the editbox in the GUI, and the cargo filter all by itself looked much more wrong than a full-width textbox (which we have elsewhere in the GUI, including the object menu).

swapped_lines

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -34,6 +34,9 @@
 #include "train_cmd.h"
 #include "vehicle_cmd.h"
 #include "zoom_func.h"
+#include "querystring_gui.h"
+#include "stringfilter_type.h"
+#include "hotkeys.h"
 
 #include "widgets/build_vehicle_widget.h"
 
@@ -69,6 +72,7 @@ static const NWidgetPart _nested_build_vehicle_widgets[] = {
 				NWidget(WWT_TEXTBTN, COLOUR_GREY, WID_BV_SHOW_HIDDEN_ENGINES),
 				NWidget(WWT_DROPDOWN, COLOUR_GREY, WID_BV_CARGO_FILTER_DROPDOWN), SetResize(1, 0), SetFill(1, 0), SetDataTip(STR_JUST_STRING, STR_TOOLTIP_FILTER_CRITERIA),
 			EndContainer(),
+			NWidget(WWT_EDITBOX, COLOUR_GREY, WID_BV_FILTER), SetMinimalSize(128, 0), SetResize(1, 0), SetFill(1, 0), SetDataTip(STR_LIST_FILTER_OSKTITLE, STR_LIST_FILTER_TOOLTIP),
 		EndContainer(),
 	EndContainer(),
 	/* Vehicle list. */
@@ -854,6 +858,29 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
 	return y;
 }
 
+
+/**
+ * Try to get the NewGRF engine additional text callback as an optional std::string.
+ * @param engine The engine whose additional text to get.
+ * @return The std::string if present, otherwise std::nullopt.
+ */
+static std::optional<std::string> GetNewGRFAdditionalText(EngineID engine)
+{
+	uint16 callback = GetVehicleCallback(CBID_VEHICLE_ADDITIONAL_TEXT, 0, 0, engine, nullptr);
+	if (callback == CALLBACK_FAILED || callback == 0x400) return std::nullopt;
+	const GRFFile *grffile = Engine::Get(engine)->GetGRF();
+	assert(grffile != nullptr);
+	if (callback > 0x400) {
+		ErrorUnknownCallbackResult(grffile->grfid, CBID_VEHICLE_ADDITIONAL_TEXT, callback);
+		return std::nullopt;
+	}
+
+	StartTextRefStackUsage(grffile, 6);
+	std::string result = GetString(GetGRFStringID(grffile->grfid, 0xD000 + callback));
+	StopTextRefStackUsage();
+	return result;
+}
+
 /**
  * Display additional text from NewGRF in the purchase information window
  * @param left   Left border of text bounding box
@@ -864,19 +891,9 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
  */
 static uint ShowAdditionalText(int left, int right, int y, EngineID engine)
 {
-	uint16 callback = GetVehicleCallback(CBID_VEHICLE_ADDITIONAL_TEXT, 0, 0, engine, nullptr);
-	if (callback == CALLBACK_FAILED || callback == 0x400) return y;
-	const GRFFile *grffile = Engine::Get(engine)->GetGRF();
-	assert(grffile != nullptr);
-	if (callback > 0x400) {
-		ErrorUnknownCallbackResult(grffile->grfid, CBID_VEHICLE_ADDITIONAL_TEXT, callback);
-		return y;
-	}
-
-	StartTextRefStackUsage(grffile, 6);
-	uint result = DrawStringMultiLine(left, right, y, INT32_MAX, GetGRFStringID(grffile->grfid, 0xD000 + callback), TC_BLACK);
-	StopTextRefStackUsage();
-	return result;
+	auto text = GetNewGRFAdditionalText(engine);
+	if (!text) return y;
+	return DrawStringMultiLine(left, right, y, INT32_MAX, *text, TC_BLACK);
 }
 
 void TestedEngineDetails::FillDefaultCapacities(const Engine *e)
@@ -1091,6 +1108,11 @@ void DisplayVehicleSortDropDown(Window *w, VehicleType vehicle_type, int selecte
 	ShowDropDownMenu(w, _engine_sort_listing[vehicle_type], selected, button, 0, hidden_mask);
 }
 
+/** Enum referring to the Hotkeys in the build vehicle window */
+enum BuildVehicleHotkeys {
+	BVHK_FOCUS_FILTER_BOX, ///< Focus the edit box for editing the filter string
+};
+
 /** GUI for building vehicles. */
 struct BuildVehicleWindow : Window {
 	VehicleType vehicle_type;                   ///< Type of vehicles shown in the window.
@@ -1111,6 +1133,9 @@ struct BuildVehicleWindow : Window {
 	int details_height;                         ///< Minimal needed height of the details panels, in text lines (found so far).
 	Scrollbar *vscroll;
 	TestedEngineDetails te;                     ///< Tested cost and capacity after refit.
+
+	StringFilter string_filter;                 ///< Filter for vehicle name
+	QueryString vehicle_editbox;                ///< Filter editbox
 
 	void SetBuyVehicleText()
 	{
@@ -1149,7 +1174,7 @@ struct BuildVehicleWindow : Window {
 		}
 	}
 
-	BuildVehicleWindow(WindowDesc *desc, TileIndex tile, VehicleType type) : Window(desc)
+	BuildVehicleWindow(WindowDesc *desc, TileIndex tile, VehicleType type) : Window(desc), vehicle_editbox(MAX_LENGTH_VEHICLE_NAME_CHARS * MAX_CHAR_LENGTH, MAX_LENGTH_VEHICLE_NAME_CHARS)
 	{
 		this->vehicle_type = type;
 		this->listview_mode = tile == INVALID_TILE;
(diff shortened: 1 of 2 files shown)


audigex:
2tall

Additional Text length testing - appears to be 2046 characters

audigex:
It would be nice if the nested results were expanded down the tree to the search result, I think? Not a huge issue, but that would seem more intuitive to me

src/build_vehicle_gui.cpp
```diff
@@ -1330,6 +1357,41 @@ struct BuildVehicleWindow : Window {
...
+	bool FilterByExtraText(const Engine* e)
+	{
+		/* Do not filter if the filter text box is empty */
+		if (this->string_filter.IsEmpty()) {
+			return true;
+		}
+
+		StringID extra_text = GetNewGRFAdditionalText(e->index);
+		if (extra_text != INVALID_STRING_ID) {
+			char buffer[MAX_LENGTH_VEHICLE_NAME_CHARS * MAX_CHAR_LENGTH];
```
James103: According to https://github.com/OpenTTD/OpenTTD/pull/10519#issuecomment-1447426417, the max buffer length here appears to be 2046 ASCII characters.

However, the string is likely null-terminated, so the buffer should at least be set to 2048 bytes just in case.

2TallTyler:
I expanded the buffer to `DRAW_STRING_BUFFER = 2048`, but am still getting odd results with Iron Horse. I suspect the NewGRF is using the text stack to generate these results, which may be the root of the problem, but I have no idea how to solve it.

horse

13.0 shows the correct result:

2TallTyler:
I still need to fix usage of the NewGRF text stack, but a new to-do item is filtering the NewGRF vehicle name callback for variants. Does anybody know of a NewGRF that uses this feature?

src/build_vehicle_gui.cpp
```diff
@@ -1354,6 +1416,9 @@ struct BuildVehicleWindow : Window {
 			/* Filter now! So num_engines and num_wagons is valid */
 			if (!FilterSingleEngine(eid)) continue;
 
+			/* Filter by name or NewGRF extra text */
+			if (!(FilterByName(e) || FilterByExtraText(e))) continue;
```
frosch123: This break the AND behavior when filtering with multiple words.
AND/OR and multiple words should be handled by `StringFilter`, so it is important to pass all data at once, and call `ResetState` and `GetState` only once.

src/build_vehicle_gui.cpp
```diff
@@ -854,31 +858,43 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
...
 	assert(grffile != nullptr);
 	if (callback > 0x400) {
 		ErrorUnknownCallbackResult(grffile->grfid, CBID_VEHICLE_ADDITIONAL_TEXT, callback);
-		return y;
+		return INVALID_STRING_ID;
 	}
 
 	StartTextRefStackUsage(grffile, 6);
-	uint result = DrawStringMultiLine(left, right, y, INT32_MAX, GetGRFStringID(grffile->grfid, 0xD000 + callback), TC_BLACK);
+	StringID result = GetGRFStringID(grffile->grfid, 0xD000 + callback);
```
frosch123: This StringID is only valid until StopTextRefStackUsage. You have to call `GetString` already here.

src/build_vehicle_gui.cpp
```diff
@@ -1785,13 +1861,45 @@
...
+	{
+		switch (hotkey) {
+			case BVHK_FOCUS_FILTER_BOX:
+				this->SetFocusedWidget(WID_BV_FILTER);
+				SetFocusedWindow(this); // The user has asked to give focus to the text box, so make sure this window is focused.
+				break;
+
+			default:
+				return ES_NOT_HANDLED;
+		}
```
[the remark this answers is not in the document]
2TallTyler: Should this be fixed or dismissed? Hotkey handling, like GUI stuff, tends to use switches so it's a question of consistency.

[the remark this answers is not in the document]
rubidium42: I think dismissing them is counter productive, as then you would probably dismiss all of them and any new cases get this alert again.
So, there is a choice to be made: 
* do we want to keep this coding style, then explicitly disable this specific check from the CodeQL configuration (.github/codeql/codeql-config.yml, where two are already disabled).
* do we want to comply with this alert, then make a (separate) PR that changes all these cases from a `switch` to an `if`.

I would have a slight favour for the first solution just because this makes extending these kinds of callbacks easier in the future. But I can also live with changing them to a simple `if`

On the other hand, the `OnEditboxChanged` above it uses the `if` paradigm.
Conceptually, I would replace the break with `return ES_HANDLED` as you're done handling the event at that point.

src/build_vehicle_gui.cpp
```diff
@@ -854,28 +858,41 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
...
- * Display additional text from NewGRF in the purchase information window
- * @param left   Left border of text bounding box
- * @param right  Right border of text bounding box
- * @param y      Top border of text bounding box
- * @param engine Engine to query the additional purchase information for
- * @return       Bottom border of text bounding box
+ *
  */
-static uint ShowAdditionalText(int left, int right, int y, EngineID engine)
+static std::tuple<bool, std::string> GetNewGRFAdditionalText(EngineID engine)
```
rubidium42: Clippy-mode: "it looks like you're trying to implement `std::optional`" ;)
But seriously, I think `std:optional` is what you conceptually want here. Also, is the comment of this function missing?

Sadly we're still at C++17 and not C++23, as then we could use the monadic functionality and write `ShowAdditionalText` as:
```
return GetNewGRFAdditionalText(engine)
        .and_then([=] (auto &text) { return DrawStringMultiLine(left, right, y, INT32_MAX, text, TC_BLACK); }
        .value_or(y);
```

src/build_vehicle_gui.cpp
```diff
@@ -854,28 +858,41 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
...
+ * Display additional text from NewGRF in the purchase information window
+ * @param left   Left border of text bounding box
+ * @param right  Right border of text bounding box
+ * @param y      Top border of text bounding box
+ * @param engine Engine to query the additional purchase information for
+ * @return       Bottom border of text bounding box
+ */
+static uint ShowAdditionalText(int left, int right, int y, EngineID engine)
+{
+	std::tuple<bool, std::string> packed = GetNewGRFAdditionalText(engine);
```
rubidium42: In general, with tuples and the likes you more likely want to use a structured binding declaration like:
`auto& [succeeded, text] = GetNewGRFAdditionalText(engine);` as figuring out the indices later on when unpacking is not that nice.

src/stringfilter_type.h
```diff
@@ -59,6 +59,7 @@ struct StringFilter {
 	bool IsEmpty() const { return this->word_index.size() == 0; }
 
 	void ResetState();
+	void AddLine(std::string string);
```
rubidium42: `const std::string &string` as `std::string string` makes a copy, which doesn't seem needed by the looks of it. And maybe rename the parameter to `str` to match the implementation.

src/stringfilter.cpp
```diff
@@ -116,6 +116,19 @@ void StringFilter::AddLine(const char *str)
 	}
 }
 
+/**
+ * Pass another text line from the current item to the filter.
+ *
+ * You can call this multiple times for a single item, if the filter shall apply to multiple things.
+ * Before processing the next item you have to call ResetState().
+ *
+ * @param string Another line from the item.
```
rubidium42: Does not match the parameter; I'd suggest going for `str` as then there is no clash with the class name.

Maybe we should get a GitHub work flow that checks the doxygen logs for missing/extra comments, so it can warn about these inconsistencies automatically.

2TallTyler: Yep, no excuse on this or the missing doxygen elsewhere besides sloppy self-review. Sorry about that. 🙂

src/build_vehicle_gui.cpp
2TallTyler: Thanks for the advice. I've learned a lot from this review. 🙂 I've never used `std::optional` but it's exactly what I want.

src/build_vehicle_gui.cpp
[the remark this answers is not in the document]
2TallTyler: I'll make a separate PR proposing to change the CodeQL configuration. That seems like a better place to have this discussion.

`OnEditboxChanged` seems to use `if` globally when there's only one case, except for a few uses where it doesn't even check the widget. The only use of `switch` is the network game list, which has two editboxes.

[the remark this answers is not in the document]
2TallTyler: Resolved by #10537.

src/build_vehicle_gui.cpp
```diff
@@ -1359,6 +1409,9 @@ struct BuildVehicleWindow : Window {
 			/* Filter now! So num_engines and num_wagons is valid */
 			if (!FilterSingleEngine(eid)) continue;
 
+			/* Filter by name or NewGRF extra text */
+			if (!(FilterByText(e))) continue;
```
rubidium42: Unneeded parentheses.

src/build_vehicle_gui.cpp
```diff
@@ -1405,6 +1458,9 @@ struct BuildVehicleWindow : Window {
 			if (!IsEngineBuildable(eid, VEH_ROAD, _local_company)) continue;
 			if (this->filter.roadtype != INVALID_ROADTYPE && !HasPowerOnRoad(e->u.road.roadtype, this->filter.roadtype)) continue;
 
+			/* Filter by name or NewGRF extra text */
+			if (!(FilterByText(e))) continue;
```
rubidium42: Unneeded parentheses.

src/build_vehicle_gui.cpp
```diff
@@ -1422,6 +1478,10 @@ struct BuildVehicleWindow : Window {
 			if (!this->show_hidden_engines && e->IsVariantHidden(_local_company)) continue;
 			EngineID eid = e->index;
 			if (!IsEngineBuildable(eid, VEH_SHIP, _local_company)) continue;
+
+			/* Filter by name or NewGRF extra text */
+			if (!(FilterByText(e))) continue;
```
rubidium42: Unneeded parentheses.

src/build_vehicle_gui.cpp
```diff
@@ -1449,7 +1509,11 @@ struct BuildVehicleWindow : Window {
 			/* First VEH_END window_numbers are fake to allow a window open for all different types at once */
 			if (!this->listview_mode && !CanVehicleUseStation(eid, st)) continue;
 
+			/* Filter by name or NewGRF extra text */
+			if (!(FilterByText(e))) continue;
```
rubidium42: Unneeded parentheses.

src/build_vehicle_gui.cpp
```diff
@@ -1335,6 +1364,27 @@ struct BuildVehicleWindow : Window {
...
+	bool FilterByText(const Engine *e)
+	{
+		/* Do not filter if the filter text box is empty */
+		if (this->string_filter.IsEmpty()) {
+			return true;
+		}
+
+		/* Filter engine name */
+		char buffer[DRAW_STRING_BUFFER];
+		GetString(buffer, e->info.string_id, lastof(buffer));
```
rubidium42: Since you added a `std::string` `AddLine` version, you can use the `std::string` returning `GetString` here.

src/build_vehicle_gui.cpp
```diff
@@ -854,31 +858,45 @@ static int DrawAircraftPurchaseInfo(int left, int right, int y, EngineID engine_
...
+ * @param y      Top border of text bounding box
+ * @param engine Engine to query the additional purchase information for
+ * @return       Bottom border of text bounding box
+ */
+static uint ShowAdditionalText(int left, int right, int y, EngineID engine)
+{
+	auto text = GetNewGRFAdditionalText(engine);
+	if (!text) return y;
+	uint result = DrawStringMultiLine(left, right, y, INT32_MAX, *text, TC_BLACK);
+	return result;
```
rubidium42: Could return the result directly here, without the extra variable.

src/stringfilter.cpp
```diff
@@ -117,6 +117,19 @@ void StringFilter::AddLine(const char *str)
...
 
+/**
+ * Pass another text line from the current item to the filter.
+ *
+ * You can call this multiple times for a single item, if the filter shall apply to multiple things.
+ * Before processing the next item you have to call ResetState().
+ *
+ * @param str Another line from the item.
+ */
+void StringFilter::AddLine(const std::string &str)
```
rubidium42: Since you added this, maybe remove the `.c_str()` on the 5 cases that could now use this function?

2TallTyler: Yes, although it would make more sense to me as a separate PR?

2TallTyler: Now #10743. 🙂

rubidium42 (approved):
Looks okay to me.

PR #10523, opened 2023-02-28
Add: maximum number of companies allowed to the client list

rubidium42:
## Motivation / Problem

See #10054: "I've seen a few new players join and try to create a company, not realizing that the game has reached the maximum amount of companies, e.g #9703, which is easier to see at a glance."

I could not push to that PR, so a new one is needed. Closes #10054.

## Description

This PR adds the maximum number of companies to the current label with "{NUM} clients / {NUM} companies" at the bottom of the client list, changing it to: "{NUM} clients - {NUM}/{NUM} companies" (while accounting for the right plurals).

Since the number of clients allowed in the server is not available at the client, I have not made it consistent "{NUM}/{NUM} clients - {NUM}/{NUM} companies". This also keeps the line slightly narrower.

I became aware that #10054 did not update the UIs properly when changing the maximum, so the appropriate dirtying of the client list window was added. This also solves the issue that the server name does not get properly updated at both the server and client.

example

## Limitations

It does not make it 100% explicit that the server's maximum number of companies is reached, as in: that is not explicitly stated. However, there not being a button to start a new company and the number of companies being the same as the maximum should be enough.
However, I added a tooltip to the label so people can be aware of the meaning of everything.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2358,7 +2358,8 @@ STR_NETWORK_CLIENT_LIST_NEW_COMPANY                             :(New company)
 STR_NETWORK_CLIENT_LIST_NEW_COMPANY_TOOLTIP                     :{BLACK}Create a new company and join it
 STR_NETWORK_CLIENT_LIST_PLAYER_ICON_SELF_TOOLTIP                :{BLACK}This is you
 STR_NETWORK_CLIENT_LIST_PLAYER_ICON_HOST_TOOLTIP                :{BLACK}This is the host of the game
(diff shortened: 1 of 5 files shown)


PeterN (approved):

2TallTyler (commented):
Do we use "num/max" anywhere already? Most of what I've seen (vehicle ages mostly) use "num (max)". 

I could absolutely be wrong, but I'm not near a computer to check and wanted to weigh in before this gets merged. 🙂

PeterN:
It's in the server list at least...
image

2TallTyler:
That's as good a place as any! 😃

PR #10529, opened 2023-03-01
Fix: crash due to incomplete deity/company mode precondition checks

rubidium42:
## Motivation / Problem

Inconsistent/missing checks for being in deity/company mode. For example, for setting the industry text there is no explicit check for it being called from deity mode. If called from company mode the call will eventually fail in the command processing, instead of earlier.

Furthermore there is a potential crash to desktop when the company mode becomes invalid.

## Description

The crashes to desktop are due to the company of the CompanyMode becoming invalid after a merge or bankruptcy, and then the removed company is still being used in getters and commands. This causes dereferences of nullptr.
This is solved by checking in `ScriptCompanyMode::IsValid` for the company of the company mode still being a valid company.

All cases where `ScriptObject::GetCompany` is used, including all commands, are suspect for this potential issue of using an invalid company. So, add a precondition check for either the right company mode or deity mode being active to all those functions.

Furthermore a script has been made that finds such cases, and this is integrated in a workflow that checks for this during builds on github.

Finally the last error for all these precondition checks is changed to `ERR_PRECONDITION_INVALID_COMPANY` instead of the less specific `ERR_PRECONDITION_INVALID`.

## Limitations

None as far as I know.

## Checklist for review

### DIFF
--- a/.github/script-missing-mode-enforcement.py
+++ b/.github/script-missing-mode-enforcement.py
@@ -1,0 +1,71 @@
+"""
+Script to scan the OpenTTD's script API for functions that miss checks for the
+function being called from the right mode (deity or company mode).
+
+When a function calls either ScriptObject::Command or ScriptObject::GetCompany
+then the function is considered dangerous. When one of the mode enforcement
+macros from script_error.hpp, i.e. EnforceDeityMode, EnforceCompanyModeValid or
+EnforceDeityOrCompanyModeValid, are called in the function, then we consider
+that the function has mode enforcement.
+
+Any dangerous function for which no enforcement is found are emitted as errors.
+"""
+
+import glob
+import re
+import sys
+
+
+def check_mode_enforcement(path):
+    errors = []
+    with open(path, "r") as reader:
+        mode_enforcement_found = False
+        dangerous_function = False
+        for line in reader:
+            # Line does not start with a tab and have <word>::<word>. That looks like the begin of a function, so reset the state.
+            if re.match(r"^[^\t].*\w::\w", line):
+                mode_enforcement_found = False
+                dangerous_function = False
+                currentFunction = line
+                continue
+
+            if re.match(
+                r"\t(EnforceDeityMode|EnforceCompanyModeValid|EnforceDeityOrCompanyModeValid|EnforceDeityOrCompanyModeValid_Void)\(",
+                line,
+            ):
+                # Mode enforcement macro found
+                mode_enforcement_found = True
+                continue
+
+            if re.match(r".*(ScriptObject::Command|ScriptObject::GetCompany).*", line):
+                # Dangerous function found
+                dangerous_function = True
+                continue
+
+            # Line with only a closing bracket. That looks like the end of a function, so check for the dangerous function without mode enforcement
+            if re.match(r"^}$", line) and dangerous_function and not mode_enforcement_found:
+                function_name = currentFunction.rstrip("\n").replace("/* static */ ", "")
+                errors.append(f"{path}: {function_name}")
+
+    return errors
+
+
+def main():
+    errors = []
+    for path in sorted(glob.glob("src/script/api/*.cpp")):
+        # Skip a number of files that yield only false positives
+        if path.endswith(("script_object.cpp", "script_companymode.cpp", "script_controller.cpp", "script_game.cpp")):
+            continue
+
+        errors.extend(check_mode_enforcement(path))
+
+    if errors:
+        print("Mode enforcement was expected in the following files/functions:")
+        print("\n".join(errors))
+        sys.exit(1)
+
+    print("OK")
+
+
+if __name__ == "__main__":
+    main()
--- a/src/script/api/script_airport.cpp
+++ b/src/script/api/script_airport.cpp
@@ -91,6 +91,7 @@
 
 /* static */ SQInteger ScriptAirport::GetNumHangars(TileIndex tile)
 {
+	EnforceDeityOrCompanyModeValid(-1);
 	if (!::IsValidTile(tile)) return -1;
 	if (!::IsTileType(tile, MP_STATION)) return -1;
 
@@ -103,6 +104,7 @@
 
 /* static */ TileIndex ScriptAirport::GetHangarOfAirport(TileIndex tile)
 {
+	EnforceDeityOrCompanyModeValid(INVALID_TILE);
 	if (!::IsValidTile(tile)) return INVALID_TILE;
 	if (!::IsTileType(tile, MP_STATION)) return INVALID_TILE;
 	if (GetNumHangars(tile) < 1) return INVALID_TILE;
--- a/src/script/api/script_depotlist.cpp
+++ b/src/script/api/script_depotlist.cpp
@@ -16,6 +16,7 @@
 
 ScriptDepotList::ScriptDepotList(ScriptTile::TransportType transport_type)
 {
+	EnforceDeityOrCompanyModeValid_Void();
 	::TileType tile_type;
 	switch (transport_type) {
 		default: return;
--- a/src/script/api/script_sign.cpp
+++ b/src/script/api/script_sign.cpp
@@ -21,6 +21,7 @@
 
 /* static */ bool ScriptSign::IsValidSign(SignID sign_id)
 {
+	EnforceDeityOrCompanyModeValid(false);
 	const Sign *si = ::Sign::GetIfValid(sign_id);
 	return si != nullptr && (si->owner == ScriptObject::GetCompany() || si->owner == OWNER_DEITY);
 }
@@ -36,6 +37,7 @@
 {
 	CCountedPtr<Text> counter(name);
 
+	EnforceDeityOrCompanyModeValid(false);
 	EnforcePrecondition(false, IsValidSign(sign_id));
 	EnforcePrecondition(false, name != nullptr);
 	const std::string &text = name->GetDecodedText();
@@ -63,6 +65,7 @@
 
 /* static */ bool ScriptSign::RemoveSign(SignID sign_id)
 {
+	EnforceDeityOrCompanyModeValid(false);
 	EnforcePrecondition(false, IsValidSign(sign_id));
 	return ScriptObject::Command<CMD_RENAME_SIGN>::Do(sign_id, "");
 }
@@ -71,6 +74,7 @@
 {
 	CCountedPtr<Text> counter(name);
 
+	EnforceDeityOrCompanyModeValid(INVALID_SIGN);
 	EnforcePrecondition(INVALID_SIGN, ::IsValidTile(location));
 	EnforcePrecondition(INVALID_SIGN, name != nullptr);
 	const std::string &text = name->GetDecodedText();
--- a/src/script/api/script_town.cpp
+++ b/src/script/api/script_town.cpp
@@ -281,6 +281,7 @@
 {
 	CCountedPtr<Text> counter(name);
 
+	EnforceDeityOrCompanyModeValid(false);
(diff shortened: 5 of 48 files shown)


src/script/api/script_depotlist.cpp
```diff
@@ -16,6 +16,7 @@
 
 ScriptDepotList::ScriptDepotList(ScriptTile::TransportType transport_type)
 {
+	EnforceDeityOrCompanyModeValid();
```
glx22: This is wrong for every builds, but only MSVC warns.

rubidium42: I know, I've fixed it locally.

I'm working on splitting this PR in smaller, hopefully easier to review PRs as it has become a bit of a mess in this branch. Like first a PR for the trivial refactor, and then update this PR to add all the missing ones with logic to prevent them going missing in the future.

src/script/api/script_airport.cpp
```diff
@@ -91,24 +91,26 @@
 
 /* static */ int32 ScriptAirport::GetNumHangars(TileIndex tile)
 {
+	EnforceDeityOrCompanyModeValid(false);
```
SamuXarick:
```suggestion
	EnforceDeityOrCompanyModeValid(-1);
```

src/script/api/script_airport.cpp
```diff
@@ -91,24 +91,26 @@
...
-	if (st->owner != ScriptObject::GetCompany() && ScriptObject::GetCompany() != OWNER_DEITY) return -1;
+	if (st->owner != ScriptObject::GetCompany() && ScriptCompanyMode::IsValid()) return -1;
 	if ((st->facilities & FACIL_AIRPORT) == 0) return -1;
 
 	return st->airport.GetNumHangars();
 }
 
 /* static */ TileIndex ScriptAirport::GetHangarOfAirport(TileIndex tile)
 {
+	EnforceDeityOrCompanyModeValid(false);
```
SamuXarick:
```suggestion
	EnforceDeityOrCompanyModeValid(INVALID_TILE);
```

src/script/api/script_sign.cpp
```diff
@@ -71,6 +74,7 @@
 {
 	CCountedPtr<Text> counter(name);
 
+	EnforceDeityOrCompanyModeValid(false);
```
SamuXarick:
```suggestion
	EnforceDeityOrCompanyModeValid(INVALID_SIGN);
```

src/script/api/script_town.cpp
```diff
@@ -277,11 +281,12 @@
...
 
-	EnforcePrecondition(false, ScriptObject::GetCompany() == OWNER_DEITY || _settings_game.economy.found_town != TF_FORBIDDEN);
+	EnforceDeityOrCompanyModeValid(false);
+	EnforcePrecondition(false, ScriptCompanyMode::IsDeity() || _settings_game.economy.found_town != TF_FORBIDDEN);
 	EnforcePrecondition(false, ::IsValidTile(tile));
 	EnforcePrecondition(false, size == TOWN_SIZE_SMALL || size == TOWN_SIZE_MEDIUM || size == TOWN_SIZE_LARGE)
-	EnforcePrecondition(false, size != TOWN_SIZE_LARGE || ScriptObject::GetCompany() == OWNER_DEITY);
-	if (ScriptObject::GetCompany() == OWNER_DEITY || _settings_game.economy.found_town == TF_CUSTOM_LAYOUT) {
+	EnforcePrecondition(false, ScriptCompanyMode::IsDeity() || size != TOWN_SIZE_LARGE);
+	if (ScriptCompanyMode::IsDeity() || _settings_game.economy.found_town == TF_CUSTOM_LAYOUT) {
```
SamuXarick: You switched the evaluation order. Does it matter?

SamuXarick (changes requested):

src/script/api/script_town.cpp
rubidium42: No, it was done for consistency's sake with the check three lines before and to be consistent with the `@pre` documentation.

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,66 @@
...
+import sys
+
+def checkModeEnforcement(path):
+	errors = []
+	with open(path, 'r') as reader:
+		modeEnforcementFound = False
+		dangerousFunction = False
+		line = reader.readline()
+
+		while line != '':  # The EOF char is an empty string
```
LordAro: `for line in reader:`

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+
+Any dangerous function for which no enforcement is found are emitted as errors.
+"""
+
+import glob
+import re
+import sys
+
+def checkModeEnforcement(path):
+	errors = []
```
TrueBrain: It is really uncommon for Python scripts to use tabs; although fully allowed, it is uncommon.

The other scripts for OpenTTD we have been writing with the more default "4 spaces". Maybe good to follow that trend here too?

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+	with open(path, 'r') as reader:
+		modeEnforcementFound = False
+		dangerousFunction = False
+		for line in reader:
+			# Line does not start with a tab and have <word>::<word>. That looks like the begin of a function, so reset the state.
+			if re.match("^[^\t].*\w::\w", line):
+				modeEnforcementFound = False
+				dangerousFunction = False
+				currentFunction = line
```
TrueBrain: For ease of reading, mind doing a `continue` here, and dropping the `else`?

The endless deeper indentation makes for hard-to-read code :)

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+			else:
+				if re.match("\t(EnforceDeityMode|EnforceCompanyModeValid|EnforceDeityOrCompanyModeValid|EnforceDeityOrCompanyModeValid_Void)\\(", line):
+					# Mode enforcement macro found
+					modeEnforcementFound = True
+
+				if re.match(".*(ScriptObject::Command|ScriptObject::GetCompany).*", line):
+					# Dangerous function found
+					dangerousFunction = True
```
TrueBrain: Should the above cases also contain a `continue`? I assume they are not expected to happen on the same line?

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+
+				# Line with only a closing bracket. That looks like the end of a function, so check for the dangerous function without mode enforcement
+				if re.match("^}$", line) and dangerousFunction and not modeEnforcementFound:
+					errors.append("%s: %s" % (path, currentFunction.rstrip('\n').replace("/* static */ ", "")))
```
TrueBrain: old-style strings are not used that commonly anymore. Suggestion:
```python
function_name = currentFunction.rstrip('\n').replace("/* static */ ", "")
errors.append(f"{path}: {function_name}")
```

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+that the function has mode enforcement.
+
+Any dangerous function for which no enforcement is found are emitted as errors.
+"""
+
+import glob
+import re
+import sys
+
+def checkModeEnforcement(path):
```
TrueBrain: Very nitpicking, feel free to ignore: Python normally uses this_way_of_making_function and similar for this_way_of_variables. Camelcasing really isn't used commonly.

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+		modeEnforcementFound = False
+		dangerousFunction = False
+		for line in reader:
+			# Line does not start with a tab and have <word>::<word>. That looks like the begin of a function, so reset the state.
+			if re.match("^[^\t].*\w::\w", line):
+				modeEnforcementFound = False
+				dangerousFunction = False
+				currentFunction = line
+			else:
+				if re.match("\t(EnforceDeityMode|EnforceCompanyModeValid|EnforceDeityOrCompanyModeValid|EnforceDeityOrCompanyModeValid_Void)\\(", line):
```
TrueBrain: It is wise to use `r""` for regular expressions. This is a "raw" string, and avoids weird things like escaping etc. Not required, just so you are aware that exists.

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+
+	return errors
+
+
+def main():
+	errors = []
+	for path in sorted(glob.glob("src/script/api/*.cpp")):
+		# Skip a number of files that yield only false positives
+		if not path.endswith(("script_object.cpp", "script_companymode.cpp", "script_controller.cpp", "script_game.cpp")):
+			errors = errors + checkModeEnforcement(path)
```
TrueBrain: The python way is: `errors.extend(checkModeEnforcement(path))` :)

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+				if re.match("^}$", line) and dangerousFunction and not modeEnforcementFound:
+					errors.append("%s: %s" % (path, currentFunction.rstrip('\n').replace("/* static */ ", "")))
+
+	return errors
+
+
+def main():
+	errors = []
+	for path in sorted(glob.glob("src/script/api/*.cpp")):
+		# Skip a number of files that yield only false positives
```
TrueBrain: Comment is a bit deceiving. You don't actually skip ;) What might be easier to read is to do what the comment says:

```python
if path.endswith(("script_object.cpp", "script_companymode.cpp", "script_controller.cpp", "script_game.cpp")):
    continue

errors.extend(checkModeEnforcement(path))
```

100% absolutely nitpicking, so feel free to say: nah, I'm good.

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+		if not path.endswith(("script_object.cpp", "script_companymode.cpp", "script_controller.cpp", "script_game.cpp")):
+			errors = errors + checkModeEnforcement(path)
+
+	if errors:
+		print("Mode enforcement was expected in the following files/functions:")
+		for error in errors:
+			print(error)
+		sys.exit(1)
+	else:
```
TrueBrain: no need for an `else` after a `sys.exit` :)

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+	errors = []
+	for path in sorted(glob.glob("src/script/api/*.cpp")):
+		# Skip a number of files that yield only false positives
+		if not path.endswith(("script_object.cpp", "script_companymode.cpp", "script_controller.cpp", "script_game.cpp")):
+			errors = errors + checkModeEnforcement(path)
+
+	if errors:
+		print("Mode enforcement was expected in the following files/functions:")
+		for error in errors:
+			print(error)
```
TrueBrain: `print("\n".join(errors))`, if you like. Again, matter of opinion, just this join way is more pythonic :D

.github/script-missing-mode-enforcement.py
```diff
@@ -0,0 +1,62 @@
...
+		sys.exit(1)
+	else:
+		print("OK")
+
+
+if __name__ == "__main__":
+    main()
```
TrueBrain, line 71: These days we use a tool called `black` to reformat our Python code. It basically applies the default coding style of Python to a file (https://peps.python.org/pep-0008/ and others). This saves you a lot of time and annoyance.

`pip install black && black -l 120 <filename>` is what we run in all other Python projects we build for OpenTTD. It fixes a few of my comments.

TrueBrain (approved):

PR #10532, opened 2023-03-02
Fix: deity should see all signs, AIs should not see signs of deity

rubidium42:
## Motivation / Problem

The GS can only see the signs from itself. The AI can see signs from itself and GS.

For depots, station, vehicles it is that the AI can see them of its own, whereas the GS can see them of everybody.

Why this inconsistency? I reckon because a mistake was made when amending the code.

## Description

Let GS see the signs from everybody, and AIs see only their own signs.

## Limitations

Breaks AIs that are written to read the sign of the GS, though I wonder whether that has been done.

## Checklist for review

### DIFF
--- a/src/script/api/script_sign.cpp
+++ b/src/script/api/script_sign.cpp
@@ -22,7 +22,7 @@
 /* static */ bool ScriptSign::IsValidSign(SignID sign_id)
 {
 	const Sign *si = ::Sign::GetIfValid(sign_id);
-	return si != nullptr && (si->owner == ScriptObject::GetCompany() || si->owner == OWNER_DEITY);
+	return si != nullptr && (si->owner == ScriptObject::GetCompany() || ScriptObject::GetCompany() == OWNER_DEITY);
 }
 
 /* static */ ScriptCompany::CompanyID ScriptSign::GetOwner(SignID sign_id)


PeterN:
Can players see GS signs? If so, AI should be able to as well.

glx22:
`ScriptSign::IsValidSign()` is used to determine if the sign can be modified, so for now a sign is valid if AI can touch it.

PeterN:
Ah okay, editing is another matter indeed :-)

rubidium42:
> `ScriptSign::IsValidSign()` is used to determine if the sign can be modified, so for now a sign is valid if AI can touch it.

`ScriptSign::IsValidSign` is also used to create the list of signs.


> Can players see GS signs? If so, AI should be able to as well.

Yes, they can. Players can also see depots, stations and vehicles from other players, but AIs cannot.

Currently AIs cannot see the signs from other companies, except those from the GS. The GS cannot see signs from other companies, whereas GS can see depots, stations and vehicles of other companies. That's quite inconsistent with the others.

frosch123:
Does this affect the [Script Communication Library](https://wiki.openttd.org/en/Development/Script/Script%20communication%20protocol)? GS and AI communicate via signs.

rubidium42:
Nobody seems to really second this change, so...

PeterN:
Oh. I do think GS should be able to see all signs, I'm not sure AIs shouldn't be able to see other signs, given players can see other player's signs. Unless something changed.

2TallTyler:
I agree, GS should be able to see all signs. No opinion on whether AIs should be able to see signs from other companies or those created by GS. (Are signs created in Scenario Editor owned by GS too?)

frosch123:
This discussion took a weird turn. GS generally use GSCompanyMode to access (read, act) companies.

rubidium42:
The whole point of this PR was that there is an inconsistency between the behaviour of signs compared to stations, depots and vehicles. 
If you say that AIs should get a list of all signs because human players can see all signs, then the AI should also get a list of all vehicles, stations and depots because human players can see those as well.

PeterN:
Okay, I'll happily "log out" of this because I have no idea about any of it. Sorry for the noise.

frosch123:
More explicit, so that I understand it. This PR changes:

| Object | Visible to AI or GS in GSCompanyMode | Visible to GS outside GSCompanyMode |
| ---- | ---- |---- |
| Company owned station | yes | yes |
| Buoy (station with OWNER_NONE) | yes | yes |
| Vehicle | yes | yes |
| Company built sign | yes | no -> yes |
| Scenario editor built sign (sign with OWNER_DEITY) | yes -> no | yes |

I disagree with both changes:
* Making company-owned signs visible to GS.
    * Yes, it would be consistent with stations and vehicles, but:
    * Signs can have OWNER_DEITY and vehicles/stations cannot. GS can build/remove signs. The GS needs a way to distinguish own/scenario signs from company signs. GSSign::GetOwner does not work for this, because GSCompany does not know about OWNER_DEITY.
* Making scenario-editor-built signs invisible to AI.
    * They are visible to players as well, while players can show/hide signs of other companies.

PR #9034, opened 2021-04-12
Don't complain that you cannot remove rail/road if it doesn't exist.

Xaroth:
## Motivation / Problem

Currently, when you place a 10-piece length of rail on an existing 9-piece length of rail, you don't get an error, even though you told it to re-build 9 pieces of rail. If you do the opposite (9 pieces of rail over an existing 10-piece length of rail), you get an error. This is somewhat inconsistent, not to mention quite a nuisance while building.

## Description

For both the creation and removal of rail and road pieces, we no longer complain when the requested change is already done. This makes building a bit more idempotent, and hopefully a little bit less frustrating.

## Limitations

I'm not 100% how this affects AI scripts. Some AIscript guru might know more about this.

I also purposefully did not touch any of the water functions, as that is a can of worms I'm not confident in opening.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10534, opened 2023-03-03
Change: Don't show error when rail/road construction/removal is already done

2TallTyler:
## Motivation / Problem

I thought I'd finish #9034.

> Currently, when you place a 10-piece length of rail on an existing 9-piece length of rail, you don't get an error, even though you told it to re-build 9 pieces of rail. If you do the opposite (9 pieces of rail over an existing 10-piece length of rail), you get an error. This is somewhat inconsistent, not to mention quite a nuisance while building.

## Description

Instead of giving a descriptive error when we try to modify a tile where the change is already done (overbuilding an existing tile or removing from a tile that has nothing to remove), throw a command return with no error message.

Closes #9034.

## Limitations

- This is visible to AIs, since the command now succeeds instead of fails. I don't know if this breaks anything.
- This does not affect water infrastructure, per the original PR.

## Checklist for review


src/command_func.h
```diff
@@ -27,6 +27,11 @@
  */
 static const CommandCost CMD_ERROR = CommandCost(INVALID_STRING_ID);
 
+/**
+ * Define a return value for a failed command with no cost and no error.
+ */
+static const CommandCost CMD_NOTHING = CommandCost();
```
rubidium42: `CommandCost()` is currently "command succeeded, but cost nothing". See for example `CmdPause`, `CmdForceTrainProceed`, or `CmdStartStopVehicle`.
I'm not sure documenting it as a a failure that is not an error and costs nothing is really warranted, or clear.

2TallTyler: Ah, thanks. I guess this does touch AIs then (not that it necessarily breaks anything, but needs considering).

ldpl: Also, there are plenty of commands returning `CommandCost()` already, adding another way to do the same thing seems like a bad idea.

2TallTyler: Good point, thanks.

rubidium42:
If you use autoroad and double click, you will still get an error message. This is due to passing `INVALID_AXIS` via `_place_road_dir` to the command in `BuildRoadToolbarWindow::OnPlaceMouseUp`. It is not caused by this PR, but since it claims to not show an error message when it is already done, I'm simply pointing out a situation when an error can still happen.

I'm not sure about the solution, but silencing the error itself in the command does not look like a good solution as I would rather consider passing an invalid value as a bug in the caller, not the callee. So this should be solved somehow in the UI.

2TallTyler:
@rubidium42 I'm still not really sure what you mean in your review comment. Are you asking me to fix the GUI caller in this PR?

rubidium42:
> @rubidium42 I'm still not really sure what you mean in your review comment. Are you asking me to fix the GUI caller in this PR?

Well, that's a good question. When testing this PR I found this behavior when building the same bit of road twice. Practically it is overbuilding the same tile, so according to the description it should not give an error.
However, since it's caused somewhere completely different it could also be considered a separate, but related, bug. The error message can be "hidden" in the same way as this PR does, but conceptually that does not sit well with me because the passed data is essentially incorrect.

So long story shorter: I noticed I could create an error message building the same thing twice, whereas the description suggests that should not give an error. So it might have been overlooked and in scope, or it might be something unrelated and out scope. Depending on how you see that, it means fixing the issue in this PR or leaving it for another PR.

2TallTyler:
Let's be honest with myself, I'm never going to finish this. 🙂

PR #10536, opened 2023-03-04
Change: simplify/unify company mode scope checks

rubidium42:
## Motivation / Problem

`GSCompanyMode` is a complicated mechanism, and depending on the scope it could or could not be active. For a script it is essentially unknowable whether you are in a company mode scope or not.

Next to this there are a lot of comparisons of `GSObject::GetCompany()` against `OWNER_DEITY` to determine whether a call is in the company mode or not. That in itself is not a problem, but... a company can go bankrupt and then it is still in a "valid" company scope but the calls will fail in various was, including crashes to desktops. However, `GSObject::GetCompany() != OWNER_DEITY` will not solve the issue on its own, but that needs to be moved to its own function.

## Description

This PR does not prevent or fix the above problem with companies going out of scope, but it lays the foundation for fixing that in #10529. Primarily by moving the trivial changes from that PR into this PR.

The first commit adds `GSCompanyMode::IsValid()` and `GSCompanyMode::IsDeity()`. This `IsValid()` will be changed in #10529 to also perform a check on the validity of the company. In #10529 there is also a check script that gets added to check whether a function checks for the precondition. To make validation with a script easier, separate enforce helper macros `EnforceCompanyModeValid(<retval>)` and `EnforceDeityMode(<retval>)` are added, which also saves on the amount of boiler plate code that needs to be written in many functions.

The second commit replaces:
* `GSObject::GetCompany() != OWNER_DEITY` -> `GSCompanyMode::IsValid()`
* `GSObject::GetCompany() == OWNER_DEITY` -> `GSCompanyMode::IsDeity()`
* `EnforcePrecondition(<retval>, GSObject::GetCompany() != OWNER_DEITY)` -> `EnforceCompanyModeValid(<retval>)`
* `EnforcePrecondition(<retval>, GSObject::GetCompany() == OWNER_DEITY)` -> `EnforceDeityMode(<retval>)`
* Comments mentioning company mode being active or not, now mention `GSCompanyMode::IsValid()` or `GSCompanyMode::IsDeity()` instead.

## Limitations

The crash-to-desktop bug is not fixed yet, that is for #10529.

## Checklist for review

### DIFF
--- a/src/script/api/script_companymode.hpp
+++ b/src/script/api/script_companymode.hpp
@@ -47,6 +47,22 @@ public:
 	 *   in when the instance was created.
 	 */
 	~ScriptCompanyMode();
+
+	/**
+	 * Check whether a company mode is valid. In other words, are commands
+	 * being executed under some company.
+	 * @return true When a company mode is valid.
+	 * @post !ScriptCompanyMode::IsDeity().
+	 */
+	static bool IsValid();
+
+	/**
+	 * Check whether the company mode is not active, i.e. whether we are a deity.
+	 * In other words, are commands are not being executed under some company.
+	 * @return true When we are a deity, i.e. company mode is not active.
+	 * @post !ScriptCompanyMode::IsValid().
+	 */
+	static bool IsDeity();
 };
 
 #endif /* SCRIPT_COMPANYMODE_HPP */
--- a/src/script/api/game_changelog.hpp
+++ b/src/script/api/game_changelog.hpp
@@ -18,6 +18,8 @@
  * This version is not yet released. The following changes are not set in stone yet.
  *
  * API additions:
+ * \li GSCompanyMode::IsValid
+ * \li GSCompanyMode::IsDeity
  * \li GSTown::ROAD_LAYOUT_RANDOM
  * \li GSVehicle::IsPrimaryVehicle
  *
--- a/src/script/api/script_airport.cpp
+++ b/src/script/api/script_airport.cpp
@@ -72,7 +72,7 @@
 
 /* static */ bool ScriptAirport::BuildAirport(TileIndex tile, AirportType type, StationID station_id)
 {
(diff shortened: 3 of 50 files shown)


src/script/api/script_companymode.hpp
```diff
@@ -47,6 +47,24 @@ class ScriptCompanyMode : public ScriptObject {
 	 *   in when the instance was created.
 	 */
 	~ScriptCompanyMode();
+
+	/**
+	 * Check whether a company mode is valid. In other words, are commands
+	 * being executed under some company.
+	 * @return true When a company mode is valid.
+	 * @post !ScriptCompanyMode::IsDeity().
+	 * @api gs
```
glx22: it's `@api game` IIRC

src/script/api/script_companymode.hpp
```diff
@@ -47,6 +47,24 @@ class ScriptCompanyMode : public ScriptObject {
...
+	 */
+	static bool IsValid();
+
+	/**
+	 * Check whether the company mode is not active, i.e. whether we are a deity.
+	 * In other words, are commands are not being executed under some company.
+	 * @return true When we are a deity, i.e. company mode is not active.
+	 * @post !ScriptCompanyMode::IsValid().
+	 * @api gs
```
glx22: it's `@api game` IIRC

src/script/api/script_companymode.hpp
rubidium42: Well... doxygen_filter.awk:23 changes gs to game, so it should function. But the whole `ScriptCompanyMode` is already only for the game script, so it's not even needed at all.

src/script/api/script_companymode.hpp
rubidium42: Well... doxygen_filter.awk:23 changes gs to game, so it should function. But the whole `ScriptCompanyMode` is already only for the game script, so it's not even needed at all.

glx22 (approved):

PR #10537, opened 2023-03-04
Codechange: [CodeQL] Disable CodeQL check on trivial switch statements

2TallTyler:
## Motivation / Problem

In #10519 I got a `No trivial switch statements` [complaint from CodeQL](https://github.com/OpenTTD/OpenTTD/security/code-scanning/3467) and discussed it with @rubidium42, who wrote:
> I think dismissing them is counter productive, as then you would probably dismiss all of them and any new cases get this alert again.
So, there is a choice to be made:

> do we want to keep this coding style, then explicitly disable this specific check from the CodeQL configuration (.github/codeql/codeql-config.yml, where two are already disabled).
do we want to comply with this alert, then make a (separate) PR that changes all these cases from a `switch` to an `if`.
I would have a slight favour for the first solution just because this makes extending these kinds of callbacks easier in the future. But I can also live with changing them to a simple `if`

## Description

This PR seems like a better place to discuss such a decision.

## Limitations

Our GUI code also triggers `Long switch case` complaints. This is not changed here (although if those above my pay grade decide we want to ignore that too, I can add it to this PR).

## Checklist for review

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -8,3 +8,5 @@ query-filters:
     - cpp/world-writable-file-creation
     # Basically OpenTTD's coding style for adding things like ..._INVALID to enumerations
     - cpp/irregular-enum-init
+    # Our GUI code tends to use switches for OnClick handlers, DrawWidget, and UpdateWidgetSize. Sometimes GUIs just don't have many elements, but we want to keep consistency.
+    - cpp/trivial-switch


rubidium42:
For those who have access to it, there are currently [76 alerts for "cpp/trivial-switch"](https://github.com/OpenTTD/OpenTTD/security/code-scanning?query=is%3Aopen+branch%3Amaster+rule%3Acpp%2Ftrivial-switch)

glx22:
I'm personally fine with replacing trivial switches with simple `if` as it reduces indentation and returns early.
In the reported CodeQL "issue"
```
	if (hotkey != BVHK_FOCUS_FILTER_BOX) return ES_NOT_HANDLED:

	this->SetFocusedWidget(WID_BV_FILTER);
	SetFocusedWindow(this); // The user has asked to give focus to the text box, so make sure this window is focused.
	return ES_HANDLED;
```
doesn't look like a bad change.

rubidium42 (approved):

PR #10538, opened 2023-03-04
Show the number of hidden vehicles on the button

ldpl:
## Motivation / Problem

It's unclear just by looking at buy vehicle or autoreplace window whether there are any vehicles hidden. Only way to know is clicking the button and checking what changed. And considering they can be hidden by accident or other players in the company it's not particularly convenient.

## Description

Simply add a number to the button showing how many vehicles are currently not shown because they're hidden. 

Screenshot from 2023-03-04 19-42-47

## Checklist for review

### DIFF
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -88,6 +88,7 @@ class ReplaceVehicleWindow : public Window {
 	byte sort_criteria;           ///< Criteria of sorting vehicles.
 	bool descending_sort_order;   ///< Order of sorting vehicles.
 	bool show_hidden_engines;     ///< Whether to show the hidden engines.
+	uint num_hidden_engines;      ///< Number of currently hidden engines.
 	RailType sel_railtype;        ///< Type of rail tracks selected. #INVALID_RAILTYPE to show all.
 	RoadType sel_roadtype;        ///< Type of road selected. #INVALID_ROADTYPE to show all.
 	Scrollbar *vscroll[2];
@@ -146,9 +147,9 @@ class ReplaceVehicleWindow : public Window {
 		byte side = draw_left ? 0 : 1;
 
 		GUIEngineList list;
+		this->num_hidden_engines = 0;
 
 		for (const Engine *e : Engine::IterateType(type)) {
-			if (!draw_left && !this->show_hidden_engines && e->IsVariantHidden(_local_company)) continue;
 			EngineID eid = e->index;
 			switch (type) {
 				case VEH_TRAIN:
@@ -173,6 +174,10 @@ class ReplaceVehicleWindow : public Window {
 				if (num_engines == 0 && EngineReplacementForCompany(Company::Get(_local_company), eid, this->sel_group) == INVALID_ENGINE) continue;
 			} else {
 				if (!CheckAutoreplaceValidity(this->sel_engine[0], eid, _local_company)) continue;
+				if (e->IsVariantHidden(_local_company)) {
+					this->num_hidden_engines++;
+					if (!this->show_hidden_engines) continue;
+				}
 			}
 
 			list.emplace_back(eid, e->info.variant_id, (side == 0) ? EngineDisplayFlags::None : e->display_flags, 0);
@@ -454,6 +459,10 @@ public:
 			case WID_RV_ROAD_TYPE_DROPDOWN:
 				SetDParam(0, this->sel_roadtype == INVALID_ROADTYPE ? STR_REPLACE_ALL_ROADTYPE : GetRoadTypeInfo(this->sel_roadtype)->strings.replace_text);
 				break;
+
+			case WID_RV_SHOW_HIDDEN_ENGINES:
+				SetDParam(0, this->num_hidden_engines);
+				break;
 		}
 	}
 
@@ -499,7 +508,14 @@ public:
 
 	void OnPaint() override
 	{
-		if (this->engines[0].NeedRebuild() || this->engines[1].NeedRebuild()) this->GenerateLists();
+		if (this->engines[0].NeedRebuild() || this->engines[1].NeedRebuild()) {
+			auto old_num_hidden_engines = this->num_hidden_engines;
+			this->GenerateLists();
+			if (old_num_hidden_engines != this->num_hidden_engines) {
+				this->ReInit();
+				return;
+			}
+		}
 
 		Company *c = Company::Get(_local_company);
 
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1126,6 +1126,7 @@ struct BuildVehicleWindow : Window {
 	bool descending_sort_order;                 ///< Sort direction, @see _engine_sort_direction
 	byte sort_criteria;                         ///< Current sort criterium.
 	bool show_hidden_engines;                   ///< State of the 'show hidden engines' button.
+	uint num_hidden_engines;                    ///< Number of currently hidden engines.
 	bool listview_mode;                         ///< If set, only display the available vehicles and do not show a 'build' button.
 	EngineID sel_engine;                        ///< Currently selected engine, or #INVALID_ENGINE
 	EngineID rename_engine;                     ///< Engine being renamed.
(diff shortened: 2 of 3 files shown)


src/autoreplace_gui.cpp
```diff
@@ -175,6 +176,12 @@ class ReplaceVehicleWindow : public Window {
 				if (!CheckAutoreplaceValidity(this->sel_engine[0], eid, _local_company)) continue;
 			}
 
+
+			if (!draw_left && e->IsHidden(_local_company)) {
```
glx22: Can't go in the else above ?

ldpl: I generally prefer having unrelated checks separate. Though in this case they're already mixed up.

2TallTyler (commented):
The button doesn't consider the maximum DParam value when choosing size, so if you hide more than 9 vehicles it doesn't fit.

digits

man-of-eel:
> The button doesn't consider the maximum DParam value when choosing size, so if you hide more than 9 vehicles it doesn't fit.
> 
> digits

can't you fix that by resizing the window?

2TallTyler:
"You" meaning the player? No, it doesn't resize the button.

ldpl:
Dunno if this is the best way to handle resizing but that's the best one I could think of.

2TallTyler (commented):
Couldn't you just invalidate/resize the window whenever the player hides or unhides a vehicle?

ldpl:
> Couldn't you just invalidate/resize the window whenever the player hides or unhides a vehicle?

well, I kinda did exactly that

2TallTyler:
I mean have it triggered by the OnClick() action, directly after the command that actually shows or hides the vehicle: https://github.com/ldpl/OpenTTD/blob/num-hidden/src/build_vehicle_gui.cpp#L1568. That seems much cleaner to me.

ldpl:
OnClick just sends a network command, nothing has changed yet so there is no point to resize.
Actual engine list is rebuilt OnPaint, so that's where I added resize check.

src/autoreplace_gui.cpp
```diff
@@ -173,8 +174,13 @@ class ReplaceVehicleWindow : public Window {
 				if (num_engines == 0 && EngineReplacementForCompany(Company::Get(_local_company), eid, this->sel_group) == INVALID_ENGINE) continue;
 			} else {
 				if (!CheckAutoreplaceValidity(this->sel_engine[0], eid, _local_company)) continue;
+				if (e->IsHidden(_local_company)) {
+					this->num_hidden_engines++;
+					if (!this->show_hidden_engines) continue;
+				}
 			}
 
+
```
2TallTyler: Random newline snuck in

src/build_vehicle_gui.cpp
```diff
@@ -1722,7 +1728,14 @@ struct BuildVehicleWindow : Window {
 
 	void OnPaint() override
 	{
-		this->GenerateBuildList();
+		auto old_num_hidden_engines = this->num_hidden_engines;
+		auto sel_id = this->GenerateBuildList();
```
2TallTyler: Why not `uint` and `EngineID` to match? I know it's a matter of personal opinion whether auto (and var in C#) is more or less readable than explicit types, but in code reviews without Intellisense it's definitely less readable. 🙂 Same goes for the counterpart in the autoreplace GUI.

2TallTyler (commented):
I don't understand the flow of drawing code well enough to have an opinion on where that reinit should go, but here's a couple nitpicks I'm qualified for. 😛

rubidium42:
> Dunno if this is the best way to handle resizing but that's the best one I could think of.

Normally you would do something like https://github.com/OpenTTD/OpenTTD/blob/master/src/town_gui.cpp#L904-L910 where you determine the width with the maximum number of digits you expect. Here I guess that would probably be 3 digits, as I doubt someone is going to hide a thousand engines or more.

ldpl:
Meanwhile me using vehicle hiding for the first time in actual game xD
Screenshot from 2023-04-13 22-30-04

But, yeah, I don't like that approach that you just oversize it for the most common case and then crop when it gets high. So I think resizing it dynamically as I did here is a better UX. I was mostly just wondering if there is some existing way for a button to resize to its content automatically.

andythenorth:
This adds visual noise to a cramped window, but it is quite easy to accidentally hide vehicles and not be aware. 

I have foolishly wasted time hunting for newgrf 'bugs' cause by me accidentally hiding the vehicle.

2TallTyler:
@ldpl Any chance you feel like finishing this for 14.0?

src/autoreplace_gui.cpp
```diff
@@ -173,6 +174,10 @@ class ReplaceVehicleWindow : public Window {
 				if (num_engines == 0 && EngineReplacementForCompany(Company::Get(_local_company), eid, this->sel_group) == INVALID_ENGINE) continue;
 			} else {
 				if (!CheckAutoreplaceValidity(this->sel_engine[0], eid, _local_company)) continue;
+				if (e->IsVariantHidden(_local_company)) {
```
2TallTyler, line 177: Why move this here instead of keeping on L151? Assuming we want to `continue` as soon as possible for performance.

src/build_vehicle_gui.cpp
```diff
@@ -1192,7 +1193,10 @@ struct BuildVehicleWindow : Window {
 		this->descending_sort_order = _engine_sort_last_order[type];
 		this->show_hidden_engines   = _engine_sort_show_hidden_engines[type];
 
+		this->SetCargoFilterArray();
 		this->UpdateFilterByTile();
+		this->eng_list.ForceRebuild();
+		this->GenerateBuildList(); // generate the list, since we need it in the next line
```
2TallTyler: Is this comment still correct, since the `next line` has changed? 😛

src/build_vehicle_gui.cpp
```diff
@@ -1321,20 +1322,16 @@ struct BuildVehicleWindow : Window {
 		this->te.FillDefaultCapacities(e);
 	}
 
-	void OnInit() override
-	{
-		this->SetCargoFilterArray();
-	}
-
 	/** Filter the engine list against the currently selected cargo filter */
```
2TallTyler: Doxygen needs updating to show the return type.

src/build_vehicle_gui.cpp
```diff
@@ -1427,58 +1426,58 @@ struct BuildVehicleWindow : Window {
...
+			}
 
 			/* Filter by name or NewGRF extra text */
 			if (!FilterByText(e)) continue;
 
 			this->eng_list.emplace_back(eid, e->info.variant_id, e->display_flags, 0);
-
-			if (eid == this->sel_engine) sel_id = eid;
 		}
-		this->SelectEngine(sel_id);
```
2TallTyler, line 1447: This function, and its counterparts below for ships and aircrafts, don't follow the `GenerateBuildTrainList()` example above, which removes this call but adds a return. Why? Ideally the doxygen would better explain what these functions actually do, because (no fault of yours) they are somewhat hard to read and understand.

src/autoreplace_gui.cpp
ldpl: IsVariantHidden needs to be the last check for num_hidden_engines to only count vehicles that were hidden and not removed from the list for other reasons.

src/build_vehicle_gui.cpp
ldpl: None of the functions around have a full doxygen comment so idk what the proper way of commenting here. But I added some mention of return value to the comment.

src/build_vehicle_gui.cpp
ldpl: I think it does... IIRC I moved it here because CreateNestedTree somehow requires engine list to count hidden and calculate the width of the button.

src/build_vehicle_gui.cpp
ldpl: Well, the answer is in GenerateBuildList, trains are a special case that do their own cargo filtering while other vehicle types use FilterEngineList that also adds this return.

src/build_vehicle_gui.cpp
ldpl: Ok, it's actually FinishInitNested that needed the list so I updated the comment

PR #7575, opened 2019-05-07
Feature: Industry production graph

kiwitreekor:
* This patch adds a window showing industry production history in last 24 months. (which already exists in Chris Sawyer's Locomotion)
production_history

PR #10541, opened 2023-03-05
Feature: Industry production graph

anatolyeltsov:
## Motivation / Problem

Finish the work started by @kiwitreekor in [this pull request](https://github.com/OpenTTD/OpenTTD/pull/7575).

## Description

This patch adds a window showing industry production history in last 24 months.

Most work was done by @kiwitreekor.
What I've done:

- rebased on current master
- made [requested changes](https://github.com/OpenTTD/OpenTTD/pull/7575#pullrequestreview-1224383919)
- fixed a [bug that I found](https://github.com/OpenTTD/OpenTTD/pull/7575#issuecomment-1454938018) during making changes
- renamed `_legend_excluded_cargo` to `_legend_excluded_cargo_payment_rates` in `graph_gui.cpp` for more consistency

Screenshot from 2023-03-05 06-59-55

Closes #7575. (edit 2TT)

## Limitations

When turning on/off cargos graph box can move a little as y axis labels are changing (like from 5 to 100) and therefore they take different amount of space.

## Checklist for review

### DIFF
--- a/src/graph_gui.cpp
+++ b/src/graph_gui.cpp
@@ -22,9 +22,9 @@
 #include "timer/timer.h"
 #include "timer/timer_window.h"
 #include "timer/timer_game_tick.h"
-#include "timer/timer_game_calendar.h"
 #include "timer/timer_game_economy.h"
 #include "zoom_func.h"
+#include "industry.h"
 
 #include "widgets/graph_widget.h"
 
@@ -35,7 +35,8 @@
 
 /* Bitmasks of company and cargo indices that shouldn't be drawn. */
 static CompanyMask _legend_excluded_companies;
-static CargoTypes _legend_excluded_cargo;
+static CargoTypes _legend_excluded_cargo_payment_rates;
+static CargoTypes _legend_excluded_cargo_production_history;
 
 /* Apparently these don't play well with enums. */
 static const OverflowSafeInt64 INVALID_DATAPOINT(INT64_MAX); // Value used for a datapoint that shouldn't be drawn.
@@ -176,6 +177,7 @@ protected:
 	static const int PAYMENT_GRAPH_X_STEP_DAYS    = 10; ///< X-axis step label for cargo payment rates "Days in transit".
 	static const int PAYMENT_GRAPH_X_STEP_SECONDS = 20; ///< X-axis step label for cargo payment rates "Seconds in transit".
 	static const int ECONOMY_QUARTER_MINUTES = 3;  ///< Minutes per economic quarter.
+	static const int ECONOMY_MONTH_MINUTES = 1;  ///< Minutes per economic month.
 
 	static const TextColour GRAPH_AXIS_LABEL_COLOUR = TC_BLACK; ///< colour of the graph axis label.
 
@@ -190,13 +192,14 @@ protected:
 	/* The starting month and year that values are plotted against. */
 	TimerGameEconomy::Month month;
 	TimerGameEconomy::Year year;
+	uint8_t month_increment; ///< month increment between vertical lines. must be divisor of 12.
 
 	bool draw_dates = true; ///< Should we draw months and years on the time axis?
 
 	/* These values are used if the graph is being plotted against values
 	 * rather than the dates specified by month and year. */
 	uint16_t x_values_start;
-	uint16_t x_values_increment;
+	int16_t x_values_increment;
 
 	StringID format_str_y_axis;
 	uint8_t colours[GRAPH_MAX_DATASETS];
@@ -409,7 +412,7 @@ protected:
 				SetDParam(1, year);
 				DrawStringMultiLine(x, x + x_sep, y, this->height, month == 0 ? STR_GRAPH_X_LABEL_MONTH_YEAR : STR_GRAPH_X_LABEL_MONTH, GRAPH_AXIS_LABEL_COLOUR, SA_LEFT);
 
-				month += 3;
+				month += this->month_increment;
 				if (month >= 12) {
 					month = 0;
 					year++;
@@ -498,7 +501,8 @@ protected:
 			format_str_y_axis(format_str_y_axis)
 	{
 		SetWindowDirty(WC_GRAPH_LEGEND, 0);
-		this->num_vert_lines = 24;
+		this->num_vert_lines = GRAPH_NUM_MONTHS;
+		this->month_increment = 3;
 	}
 
 	void InitializeWindow(WindowNumber number)
@@ -532,7 +536,7 @@ public:
 				SetDParam(1, year);
 				x_label_width = std::max(x_label_width, GetStringBoundingBox(month == 0 ? STR_GRAPH_X_LABEL_MONTH_YEAR : STR_GRAPH_X_LABEL_MONTH).width);
 
-				month += 3;
+				month += this->month_increment;
 				if (month >= 12) {
 					month = 0;
 					year++;
@@ -591,7 +595,7 @@ public:
 	 * Update the statistics.
 	 * @param initialize Initialize the data structure.
 	 */
-	void UpdateStatistics(bool initialize)
+	virtual void UpdateStatistics(bool initialize)
 	{
 		CompanyMask excluded_companies = _legend_excluded_companies;
 
@@ -605,7 +609,7 @@ public:
 			nums = std::min(this->num_vert_lines, std::max(nums, c->num_valid_stat_ent));
 		}
 
-		int mo = (TimerGameEconomy::month / 3 - nums) * 3;
+		int mo = (TimerGameEconomy::month / this->month_increment - nums) * this->month_increment;
 		auto yr = TimerGameEconomy::year;
 		while (mo < 0) {
 			yr--;
@@ -961,7 +965,7 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 		this->x_values_increment = (TimerGameEconomy::UsingWallclockUnits() ? PAYMENT_GRAPH_X_STEP_SECONDS : PAYMENT_GRAPH_X_STEP_DAYS);
 
 		this->CreateNestedTree();
-		this->vscroll = this->GetScrollbar(WID_CPR_MATRIX_SCROLLBAR);
+		this->vscroll = this->GetScrollbar(WID_GRAPH_MATRIX_SCROLLBAR);
 		this->vscroll->SetCount(_sorted_standard_cargo_specs.size());
 
 		auto *wid = this->GetWidget<NWidgetCore>(WID_GRAPH_FOOTER);
@@ -985,14 +989,14 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 
 		int i = 0;
 		for (const CargoSpec *cs : _sorted_standard_cargo_specs) {
-			if (HasBit(_legend_excluded_cargo, cs->Index())) SetBit(this->excluded_data, i);
+			if (HasBit(_legend_excluded_cargo_payment_rates, cs->Index())) SetBit(this->excluded_data, i);
 			i++;
 		}
 	}
 
 	void UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize) override
 	{
-		if (widget != WID_CPR_MATRIX) {
+		if (widget != WID_GRAPH_MATRIX) {
 			BaseGraphWindow::UpdateWidgetSize(widget, size, padding, fill, resize);
 			return;
 		}
@@ -1016,7 +1020,7 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 
 	void DrawWidget(const Rect &r, WidgetID widget) const override
 	{
-		if (widget != WID_CPR_MATRIX) {
+		if (widget != WID_GRAPH_MATRIX) {
 			BaseGraphWindow::DrawWidget(r, widget);
 			return;
 		}
@@ -1029,7 +1033,7 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 		for (auto it = first; it != last; ++it) {
 			const CargoSpec *cs = *it;
 
-			bool lowered = !HasBit(_legend_excluded_cargo, cs->Index());
+			bool lowered = !HasBit(_legend_excluded_cargo_payment_rates, cs->Index());
 
 			/* Redraw frame if lowered */
 			if (lowered) DrawFrameRect(line, COLOUR_BROWN, FR_LOWERED);
@@ -1052,18 +1056,18 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 	void OnClick([[maybe_unused]] Point pt, WidgetID widget, [[maybe_unused]] int click_count) override
 	{
 		switch (widget) {
-			case WID_CPR_ENABLE_CARGOES:
+			case WID_GRAPH_ENABLE_CARGOES:
 				/* Remove all cargoes from the excluded lists. */
-				_legend_excluded_cargo = 0;
+				_legend_excluded_cargo_payment_rates = 0;
 				this->excluded_data = 0;
 				this->SetDirty();
 				break;
 
-			case WID_CPR_DISABLE_CARGOES: {
+			case WID_GRAPH_DISABLE_CARGOES: {
 				/* Add all cargoes to the excluded lists. */
 				int i = 0;
 				for (const CargoSpec *cs : _sorted_standard_cargo_specs) {
-					SetBit(_legend_excluded_cargo, cs->Index());
+					SetBit(_legend_excluded_cargo_payment_rates, cs->Index());
 					SetBit(this->excluded_data, i);
 					i++;
 				}
@@ -1071,10 +1075,10 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 				break;
 			}
 
-			case WID_CPR_MATRIX: {
-				auto it = this->vscroll->GetScrolledItemFromWidget(_sorted_standard_cargo_specs, pt.y, this, WID_CPR_MATRIX);
+			case WID_GRAPH_MATRIX: {
+				auto it = this->vscroll->GetScrolledItemFromWidget(_sorted_standard_cargo_specs, pt.y, this, WID_GRAPH_MATRIX);
 				if (it != _sorted_standard_cargo_specs.end()) {
-					ToggleBit(_legend_excluded_cargo, (*it)->Index());
+					ToggleBit(_legend_excluded_cargo_payment_rates, (*it)->Index());
 					this->UpdateExcludedData();
 					this->SetDirty();
 				}
@@ -1085,7 +1089,7 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 
 	void OnResize() override
 	{
-		this->vscroll->SetCapacityFromWidget(this, WID_CPR_MATRIX);
+		this->vscroll->SetCapacityFromWidget(this, WID_GRAPH_MATRIX);
 	}
 
 	void OnGameTick() override
@@ -1146,12 +1150,12 @@ static constexpr NWidgetPart _nested_cargo_payment_rates_widgets[] = {
 			NWidget(WWT_EMPTY, COLOUR_BROWN, WID_GRAPH_GRAPH), SetMinimalSize(495, 0), SetFill(1, 1), SetResize(1, 1),
 			NWidget(NWID_VERTICAL),
 				NWidget(NWID_SPACER), SetMinimalSize(0, 24), SetFill(0, 1),
-				NWidget(WWT_PUSHTXTBTN, COLOUR_BROWN, WID_CPR_ENABLE_CARGOES), SetDataTip(STR_GRAPH_CARGO_ENABLE_ALL, STR_GRAPH_CARGO_TOOLTIP_ENABLE_ALL), SetFill(1, 0),
-				NWidget(WWT_PUSHTXTBTN, COLOUR_BROWN, WID_CPR_DISABLE_CARGOES), SetDataTip(STR_GRAPH_CARGO_DISABLE_ALL, STR_GRAPH_CARGO_TOOLTIP_DISABLE_ALL), SetFill(1, 0),
+				NWidget(WWT_PUSHTXTBTN, COLOUR_BROWN, WID_GRAPH_ENABLE_CARGOES), SetDataTip(STR_GRAPH_CARGO_ENABLE_ALL, STR_GRAPH_CARGO_TOOLTIP_ENABLE_ALL), SetFill(1, 0),
+				NWidget(WWT_PUSHTXTBTN, COLOUR_BROWN, WID_GRAPH_DISABLE_CARGOES), SetDataTip(STR_GRAPH_CARGO_DISABLE_ALL, STR_GRAPH_CARGO_TOOLTIP_DISABLE_ALL), SetFill(1, 0),
 				NWidget(NWID_SPACER), SetMinimalSize(0, 4),
 				NWidget(NWID_HORIZONTAL),
-					NWidget(WWT_MATRIX, COLOUR_BROWN, WID_CPR_MATRIX), SetFill(1, 0), SetResize(0, 2), SetMatrixDataTip(1, 0, STR_GRAPH_CARGO_PAYMENT_TOGGLE_CARGO), SetScrollbar(WID_CPR_MATRIX_SCROLLBAR),
-					NWidget(NWID_VSCROLLBAR, COLOUR_BROWN, WID_CPR_MATRIX_SCROLLBAR),
+					NWidget(WWT_MATRIX, COLOUR_BROWN, WID_GRAPH_MATRIX), SetFill(1, 0), SetResize(0, 2), SetMatrixDataTip(1, 0, STR_GRAPH_CARGO_PAYMENT_TOGGLE_CARGO), SetScrollbar(WID_GRAPH_MATRIX_SCROLLBAR),
+					NWidget(NWID_VSCROLLBAR, COLOUR_BROWN, WID_GRAPH_MATRIX_SCROLLBAR),
 				EndContainer(),
 				NWidget(NWID_SPACER), SetMinimalSize(0, 24), SetFill(0, 1),
 			EndContainer(),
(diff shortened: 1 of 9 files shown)


2TallTyler:
I like this feature, and from my testing it seems to work.

In the previous PR for this feature (#7575), there was discussion about how saving industry production data might bloat savegame file size. Can you provide some data on this, testing a variety of map sizes and industry density, so we can get a better idea of how this will change?

There was also discussion about using a different interval to graph production. I think that's out of scope and shouldn't block this PR from being merged, assuming it won't have a tremendous effect on file size.

anatolyeltsov:
Sorry, didn't see your comment earlier.
Rebased on current master and did some tests with default industries and temperate climate:

**Master:**
256 x 256 map, 50 industries, initial: 98,4 Kb
256 x 256 map, 50 industries, after 2 years: 102,5 Kb
1024 x 1024 map, 500 industries, initial: 1548 Kb
1024 x 1024 map, 500 industries, after 2 years: 1600 Kb 
512 x 512 map, 1000 industries, initial: 411,7 Kb
512 x 512 map, 1000 industries, after 2 years: 424.9 Kb
4096 x 4096 map, 2500 industries, initial: 27,3 MB
4096 x 4096 map, 2500 industries, after 2 months*: 26,5 MB

**This PR:**
256 x 256 map, 50 industries, initial: 96,6 Kb
256 x 256 map, 50 industries, after 2 years: 101,4 Kb
1024 x 1024 map, 500 industries, initial: 1561 Kb
1024 x 1024 map, 500 industries, after 2 years: 1612 Kb
512 x 512 map, 1000 industries, initial: 429,5 Kb
512 x 512 map, 1000 industries, after 2 years: 445,7 Kb
4096 x 4096 map, 2500 industries, initial: 27,7 MB
4096 x 4096 map, 2500 industries, after 2 months*: 26,9 MB

\* couldn't FF to 2 years as it was veeery slow process for me (game speed with FF was like 1.2x only) and also for whatever reason the later file is smaller

_So I would say from these examples that there's nothing really bad in terms of savegame sizes._

ldpl:
Did you test the same seed? 256x256 initial size is smaller with production which makes no sense. Also, there is not much point comparing initial sizes as production history is all zeros that compress extremely well. So would be nice to see some worst case comparison with high industry density and so 5k+ industries on the map and full year history. Ideally with pre-compression size and compresion time comparison.

Though having it being lost within the random noise in this test looks kind of promising.

anatolyeltsov:
No, seeds were random so tests are no 100% accurate but anyway.

Tried 512 x 512 map with 5000 industries:

**Master:**
Initial: 480,5 Kb
After 2 years: 484,7 Kb

**This PR:**
Initial: 446,2 Kb
After 2 years: 466,2 Kb

Saving game took less than a second for me in this case.

src/saveload/compat/industry_sl_compat.h
```diff
@@ -29,9 +29,9 @@ const SaveLoadCompat _industry_sl_compat[] = {
 	SLC_VAR("prod_level"),
 	SLC_VAR("this_month_production"),
 	SLC_VAR("this_month_transported"),
-	SLC_VAR("last_month_pct_transported"),
-	SLC_VAR("last_month_production"),
-	SLC_VAR("last_month_transported"),
+	SLC_VAR("past_pct_transported[0]"),
+	SLC_VAR("past_production[0]"),
+	SLC_VAR("past_transported[0]"),
```
PeterN: I don't think this is right, as this is the label used within the savegame.

PeterN:
The unit tests show that  last month production isn't being reported properly, possibly due to incorrect savegame changes, but maybe something else.

anatolyeltsov:
I've made a save before and after this PR and it seems that if I try to open older savegame in game based on this PR "last month production" for each industry became zero (therefore tests failed I think).

As I understand we need to get data from different fields/params depending on save load version. I'm not a big expert in all these compatibility things so not sure what to do in this regard at the moment but I'll try to figure it out.

PeterN:
That's what the SLV_CONDARR() stuff should do, but because the name of the variable has slightly changed it's upsetting it.

Might be we need to add a macro to allow overriding the used name (like for the SLEG versions.)

anatolyeltsov:
Seems like I've fixed it by adding SLE_CONDARRNAME (as you suggested if I understood you correctly + we already have SLE_CONDVAR and SLE_CONDVARNAME pair) and according changes in code. Now everything is OK including regression tests.

Should this addition go as another PR first?

2TallTyler:
Probably not another PR since it would be unused at that point, but a separate commit in this one.

PeterN:
Indeed, just rebase so that commit is first.

anatolyeltsov:
Done! But I'm not sure about spacing in `src/saveload/industry_sl.cpp` - it's become a little bit even more mess

src/saveload/industry_sl.cpp
```diff
@@ -26,27 +26,30 @@ static const SaveLoad _industry_desc[] = {
...
-	SLE_CONDARR(Industry, last_month_pct_transported, SLE_UINT8,  16,             SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SL_MAX_VERSION),
-	SLE_CONDARR(Industry, last_month_production,      SLE_UINT16,  2,               SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
-	SLE_CONDARR(Industry, last_month_production,      SLE_UINT16, 16,             SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SL_MAX_VERSION),
-	SLE_CONDARR(Industry, last_month_transported,     SLE_UINT16,  2,               SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
-	SLE_CONDARR(Industry, last_month_transported,     SLE_UINT16, 16,             SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SL_MAX_VERSION),
+	SLE_CONDARR(Industry, this_month_production,      SLE_UINT16,   2,               SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
+	SLE_CONDARR(Industry, this_month_production,      SLE_UINT16,  16,             SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SL_MAX_VERSION),
+	SLE_CONDARR(Industry, this_month_transported,     SLE_UINT16,   2,               SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
+	SLE_CONDARR(Industry, this_month_transported,     SLE_UINT16,  16,             SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SL_MAX_VERSION),
```
PeterN: I don't think you need to adjust any of the lines before this point just to add this space.

src/widgets/graph_widget.h
```diff
@@ -40,15 +40,15 @@ enum PerformanceHistoryGraphWidgets {
...
-	WID_CPR_MATRIX_SCROLLBAR,///< Cargo list scrollbar.
+	WID_CPR_BACKGROUND,       ///< Background of the window.
+	WID_CPR_HEADER,           ///< Header.
+	WID_CPR_GRAPH,            ///< Graph itself.
+	WID_CPR_RESIZE,           ///< Resize button.
+	WID_CPR_FOOTER,           ///< Footer.
+	WID_CPR_ENABLE_CARGOES,   ///< Enable cargoes button.
+	WID_CPR_DISABLE_CARGOES,  ///< Disable cargoes button.
+	WID_CPR_MATRIX,           ///< Cargo list.
+	WID_CPR_MATRIX_SCROLLBAR, ///< Cargo list scrollbar.
```
PeterN: Don't think this is necessary either

src/saveload/industry_sl.cpp
```diff
@@ -26,27 +26,30 @@ static const SaveLoad _industry_desc[] = {
...
+	SLE_CONDARRNAME(Industry, past_pct_transported[0], "last_month_pct_transported", SLE_UINT8, 16, SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SLV_PRODUCTION_HISTORY),
+	SLE_CONDARR(Industry, past_pct_transported,       SLE_UINT8,  384,             SLV_PRODUCTION_HISTORY, SL_MAX_VERSION),
+	SLE_CONDARRNAME(Industry, past_production[0], "last_month_production", SLE_UINT16,  2, SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
+	SLE_CONDARRNAME(Industry, past_production[0], "last_month_production", SLE_UINT16, 16, SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SLV_PRODUCTION_HISTORY),
+	SLE_CONDARR(Industry, past_production,            SLE_UINT16, 384,             SLV_PRODUCTION_HISTORY, SL_MAX_VERSION),
+	SLE_CONDARRNAME(Industry, past_transported[0], "last_month_transported", SLE_UINT16, 2,  SL_MIN_VERSION, SLV_EXTEND_INDUSTRY_CARGO_SLOTS),
+	SLE_CONDARRNAME(Industry, past_transported[0], "last_month_transported", SLE_UINT16, 16, SLV_EXTEND_INDUSTRY_CARGO_SLOTS, SLV_PRODUCTION_HISTORY),
+	SLE_CONDARR(Industry, past_transported,           SLE_UINT16, 384,             SLV_PRODUCTION_HISTORY, SL_MAX_VERSION),
```
PeterN: You could put the 3 new lines together at the buttom -- this would give you 6 similar lines together, then the 3 new ones.

src/industry_gui.cpp
```diff
@@ -831,9 +832,24 @@ class IndustryViewWindow : public Window
...
 
+		int count = 0;
+		const Industry *i = Industry::Get(window_number);
+		for (size_t j = 0; j < INDUSTRY_NUM_OUTPUTS; j++) {
+			if (!IsCargoIDValid(i->produced_cargo[j])) continue;
+			count++;
+		}
+		if (count < 1) {
+			this->DisableWidget(WID_IV_GRAPH);
+		}
```
PeterN: The actual count isn't cared about here, and similar code is used in `ScriptTileList_IndustryProducing::ScriptTileList_IndustryProducing()`, may be worth adding this as a member of Industry in industry.h instead.

```suggestion
		const Industry *i = Industry::Get(window_number);
		if (!i->ProducesCargo()) this->DisableWidget(WID_IV_GRAPH);
```

And I think this should be good for the actual test there:

```c++
	/** Test if this industry produces any cargo.
	 * @return true iff industry produces at least one valid cargo.
	 */
	inline bool ProducesCargo() const
	{
		const auto &found = std::find_if(this->produced_cargo, endof(this->produced_cargo), [](CargoID cid) { return IsCargoIDValid(cid); });
		return found != endof(this->produced_cargo);
	}
```

src/industry_cmd.cpp
```diff
@@ -2425,12 +2426,21 @@ static void UpdateIndustryStatistics(Industry *i)
...
+				i->past_production[k][j] = i->past_production[k-1][j];
+			}
+			i->past_production[0][j] = i->this_month_production[j];
 			i->this_month_production[j] = 0;
 
-			i->last_month_transported[j] = i->this_month_transported[j];
+			for (size_t k = lengthof(i->past_transported) - 1; k >= 1; k--){
+				i->past_transported[k][j] = i->past_transported[k-1][j];
+			}
+			i->past_transported[0][j] = i->this_month_transported[j];
```
PeterN: Missing spaces in `[k - 1]` each time.

Switching the array around from `[month][slot]` to `[slot][month]` would give simpler code here I think -- although on second thoughts that would make the saveload changes a nightmare!

PeterN (dismissed):

src/widgets/graph_widget.h
anatolyeltsov: Does OpenTTD have any strict rules about formatting spaces? Seems like we have same spacing in other places so it's a little bit confusing

src/saveload/industry_sl.cpp
anatolyeltsov: Wouldn't it be less clear for those who'll read this code?
UPD: oh, I get the idea wrong. Yep, you're right, this will look better.

src/widgets/graph_widget.h
PeterN: In general yes, but you are fixing stuff that you otherwise don't touch -- that's why it's not necessary here.

It could be a separate commit that does just this if you like, though.

anatolyeltsov: Good point, thanks!

src/industry_gui.cpp
anatolyeltsov: Should I update the code for `ScriptTileList_IndustryProducing::ScriptTileList_IndustryProducing()` in separate commit of this PR too?

PeterN: Yes can do, just explain in the commit title that it is being done to allow re-use of the code.

anatolyeltsov:
Hmm.. whats wrong with tests here?

2TallTyler:
CI checks were broken for a while, try force-pushing to run them again.

PeterN:
Draft PR#10853 restructures the accepts/produced data to make it easier to iterate through.

One effect of that is production/transported data is shuffled around in such a way that it can be extended from 2 months (this and last) to 24 months quite easily (and even monthly rotation of figures doesn't need to change), and without any (further) savegame changes.

I'm not sure if that PR is acceptable right now, but it's something to bear in mind.

anatolyeltsov:
Made it work with your latest changes, just need a little clean up!
Will update this pull request in the next few days.

PeterN:
Sorry for smashing over it your PR, but the existing layout did need a proper restructuring before adding more deep arrays.

anatolyeltsov:
No problems. Now this PR has much less changes!

src/industry_gui.cpp
```diff
@@ -1197,6 +1210,7 @@ static const NWidgetPart _nested_industry_view_widgets[] = {
 	EndContainer(),
 	NWidget(NWID_HORIZONTAL),
 		NWidget(WWT_PUSHTXTBTN, COLOUR_CREAM, WID_IV_DISPLAY), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_INDUSTRY_DISPLAY_CHAIN, STR_INDUSTRY_DISPLAY_CHAIN_TOOLTIP),
+	NWidget(WWT_PUSHTXTBTN, COLOUR_CREAM, WID_IV_GRAPH), SetFill(1, 0), SetResize(1, 0), SetDataTip(STR_INDUSTRY_VIEW_PRODUCTION_GRAPH, STR_INDUSTRY_VIEW_PRODUCTION_GRAPH_TOOLTIP),
```
PeterN: Needs to be aligned with the widgets above & below this line.

src/industry.h
```diff
@@ -78,7 +78,7 @@ struct Industry : IndustryPool::PoolItem<&_industry_pool> {
 		CargoID cargo; ///< Cargo type
 		uint16_t waiting; ///< Amount of cargo produced
 		uint8_t rate; ///< Production rate
-		std::array<ProducedHistory, 2> history; ///< History of cargo produced and transported
+		std::array<ProducedHistory, 25> history; ///< History of cargo produced and transported for this month and 24 previous months
```
PeterN, line 82: Although this is backwards compatible, it is not forwards compatible, so it may be wise to bump the savegame version anyway, even if no other saveloiad changes are made.

On the other hand there's no release with the current format so it may not be a problem.

ldpl:
Finally got to do some proper apples-to-apples comparison on this one. And results are pretty interesting, while file size increase is pretty mild, time spent in SlSaveChunks is up almost 50%. That's the non-threaded part of save process so it means higher lag for everyone in multiplayer game when client connects and on autosave. I didn't measure compression time directly but judging by SaveFileToDisk and uncompressed size that's also up about 5-10%, meaning longer connection time. IMO that's a pretty heavy price to pay for just a few charts. I'd very much prefer the game to only save data I may actually need than waste resources on random noise.

Also, that's on default industries, it will be even worse on some newgrf with rapid production changes.

[Screenshot from 2023-05-31 03-48-40](https://github.com/OpenTTD/OpenTTD/assets/413570/997bb260-5b31-43af-9dfb-55d65a399c45)
Relevant test settings: 2048x2048 map, high industry density, smooth economy

PeterN:
Massive map with slightly more data is slightly slower to save *shock*

The same amount of data is stored regardless of how many cargo types are involved and how many months has been played.

And due to updated the layout (in master), it isn't just saving an array of zeros when not filled in, either.

ldpl:
50% is not insignificant and that's not even the worst case. Servers I run have up to 10k industries and that's not a "massive" map, I sometimes get complaints from players that they run out of space. So for me this PR effectively doubles the already high lag on each save for basically nothing. As very little of what saved here is actual useful data. No one will ever look at these thousands of random fluctuation charts. It would be much better to save history for industries that matter instead. Can have more history with nearly zero performance impact. Because as it is currently I'd very much like a setting to disable this feature as it's just not worth it for me.

ldpl:
High industry density works really well for multiplayer goal games. As you can get 2-3 industries with one station.
- It makes cargo start more viable and comparable to passenger+mail.
- It's easier to spot places with good production, especially without modified client.
- Allows for more players to play on a relatively small map without running out of resources.

2TallTyler:
> @ldpl: It would be much better to save history for industries that matter instead.

So, industries with an amount transported over 0%? That's an idea.

I'm not aware of any production changes, vanilla, or NewGRF, that are based on previous production levels, so players do not need that information for non-served industries (whether they'll ask for it is another story 😛).

The Production Graph button could be greyed out if the industry's history is not enabled, possibly with a hover tooltip to say "Transport cargo from this industry to track production history" or similar.

Does the new storage format allow us to store either just last month's production, or the full history, depending on the industry?

ldpl:
IMO, the most interesting are industries that player takes cargo to (secondaries). For those I'd say even 10 years of history won't be too much. 
For primary industries it can be somewhat educational to just see how it grows but ultimately it's just some random, so single production number is enough for the gameplay. If anything, having a station rating history chart would be more useful in this case.

ldpl:
IMO, the most interesting are industries that player takes cargo to (secondaries). For those I'd say even 10 years of history won't be too much. In custom clients this function is covered by delivery cargo graph having cargo selection and it's good enough for most cases, at least in relatively short games with few secondary industries connected.

For primary industries it can be somewhat educational to just see how it grows but ultimately it's just some random, so single production number is good enough for the gameplay. If anything, having a station rating history chart would be more useful in this case.

anatolyeltsov:
> So, industries with an amount transported over 0%? That's an idea.

What if player had transported something but then stopped? Should we check if he had transported anything in any of 24 months?

> Does the new storage format allow us to store either just last month's production, or the full history, depending on the industry?

I think we can check if anything was transported [here](https://github.com/OpenTTD/OpenTTD/blob/master/src/industry_cmd.cpp#L2403) 🤔

ldpl:
> What if player had transported something but then stopped? Should we check if he had transported anything in any of 24 months?

I'd say that's such a rare case it's irrelevant.

anatolyeltsov:
But what if we check only current and/or previous months and it takes more than month (or two) for train or truck to comeback? 🤔 

Maybe it makes sense to add simple boolean for industry - something like used_at_least_once (couldn't come up with a better name at the moment) - and set it when something was transported for the first time? 🤔 🤔

anatolyeltsov:
Any suggestions on further development?

2TallTyler (commented):
I imagine the next step would be to not save production history if the industry has never been served. I think your idea of adding a boolean is fine, something like `has_been_serviced` to keep with how we word things elsewhere in the game (we use "service" in newspapers). The goal is to make this as performance-friendly as possible for the cases described by @ldpl.

PeterN:
The linked patch goes one step further than #11133 (actually makes that one redundant) and replaces the accepted and produced arrays with vectors. Although memory is used by the vector, overall for vanilla (and most other) industries (2 or 3 cargo types) this will use less memory, and again use slightly less disk space.

This one is more complex and there may be missing bounds checking elsewhere -- although most code touching this already uses iterators instead of array index access. In fact this may slightly improve performance, as code will no longer be checking a full 16 entries every time, but then again it might affect cache usage. But not benchmarked and probably not particularly noticeable.

To be honest with vanilla this doesn't gain all that much, but when the history length is extended from 2 to 25 entries, I would expect the saving to be more worthwhile -- without this each industry will use 1600 bytes just for (mostly empty) production history.

https://github.com/OpenTTD/OpenTTD/compare/master...PeterN:OpenTTD:industry-vector

2TallTyler:
So this PR should wait for the industry vector patch to be finished and merged?

andythenorth:
If this is merged, could we expose the data to Script (GS/AI) as another PR?

Currently I'm caching industry accepted/produced/transported history in GS as a Squirrel table, and writing it to the savegame.

anatolyeltsov:
@PeterN do you still consider applying the patch mentioned above?

anatolyeltsov:
Rebased on current master and now there's one problem.

Everything is still ok with calendar time mode:
graph_ip_calendar_right
but in wallclock mode it doesn't seem right now:
graph_ip_wallclock_right

Was thinking about making X-axis with minuses like this:
graph_ip_wallclock_right_minus
but then `x_values_start`, `x_values_increment` and `label` variables in `src/graph_gui.cpp` should be `int16_t` instead of `uint16_t` and I'm not sure if it's ok or not.

Any thoughts on this?

anatolyeltsov:
Rebased on current master and made according changes.
But not sure if I should change anything as in [this PR](https://github.com/OpenTTD/OpenTTD/pull/12190/files#diff-462d87cebbf7d7ed2240aac6d8a404d7ac11a50af17ff42b306b53ddafa41cd1).

2TallTyler:
Hi @anatolyeltsov, #12159 has finally been merged so this PR can be rebased and properly reviewed. 😄

Thanks for your patience!

anatolyeltsov:
@2TallTyler Hi! I'm on vacation right now, will be able to do it in the end of month.

anatolyeltsov:
Finally had time and rebased on latest master.

Calendar mode
Screenshot from 2024-08-14 14-04-04

Wallclock mode
Screenshot from 2024-08-14 14-15-54

I've changed `x_values_increment` from _uint_ to _int_ and made it negative for this type of graph - not sure if it's ok but seems better than it [was before](https://github.com/OpenTTD/OpenTTD/pull/10541#issuecomment-1937862201).

**Not yet implemented:** [has_been_serviced](https://github.com/OpenTTD/OpenTTD/pull/10541#pullrequestreview-1529132082) variable

src/graph_gui.cpp
```diff
@@ -1410,6 +1414,284 @@ struct PerformanceRatingDetailWindow : Window {
...
+
+		this->CreateNestedTree();
+		this->vscroll = this->GetScrollbar(WID_GRAPH_MATRIX_SCROLLBAR);
+
+		int count = 0;
+		const Industry *i = Industry::Get(window_number);
+		for (size_t j = 0; j < i->produced.size(); j++) {
+			if (!IsValidCargoID(i->GetProduced(j).cargo)) continue;
+			count++;
+		}
```
PeterN:
```c++
for (const auto &p : i->produced) {
    if (!IsValidCargoID(p.cargo)) continue;
    count++;
}
```

src/graph_gui.cpp
```diff
@@ -1410,6 +1414,284 @@ struct PerformanceRatingDetailWindow : Window {
...
+		int index = 0;
+		const Industry *i = Industry::Get(this->window_number);
+		const CargoSpec *cs;
+		for (size_t j = 0; j < i->produced.size(); j++) {
+			if (!IsValidCargoID(i->GetProduced(j).cargo)) continue;
+
+			cs = CargoSpec::Get(i->GetProduced(j).cargo);
+			if (HasBit(_legend_excluded_cargo_production_history, cs->Index())) SetBit(this->excluded_data, index);
+			index++;
+		}
```
PeterN:
```c++
for (const auto &p : i->produced) {
    if (!IsValidCargoID(p.cargo)) continue;
    if (HasBit(_legend_excluded_cargo_production_history, p.cargo)) SetBit(this->excluded_data, index);
    index++;
}
```

src/graph_gui.cpp
```diff
@@ -1410,6 +1414,284 @@ struct PerformanceRatingDetailWindow : Window {
...
+	void UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize) override
+	{
+		if (widget != WID_GRAPH_MATRIX) {
+			BaseGraphWindow::UpdateWidgetSize(widget, size, padding, fill, resize);
+			return;
+		}
+
+		const Industry *i = Industry::Get(this->window_number);
+		const CargoSpec *cs;
+		for (size_t j = 0; j < i->produced.size(); j++) {
```
PeterN: Same :)

src/graph_gui.cpp
```diff
@@ -1410,6 +1414,284 @@ struct PerformanceRatingDetailWindow : Window {
...
+		bool rtl = _current_text_dir == TD_RTL;
+
+		int pos = this->vscroll->GetPosition();
+		int max = pos + this->vscroll->GetCapacity();
+
+		Rect line = r.WithHeight(this->line_height);
+		const Industry *i = Industry::Get(this->window_number);
+		const CargoSpec *cs;
+
+		for (size_t j = 0; j < i->produced.size(); j++) {
```
PeterN: Same :) There are more but remaining as well :)

PeterN (changes requested):
Basically most of the for-loops indexing `i->produced` can use range-for instead.

And `cs->Index()` is the same as the CargoID, so often no need for CargoSpec::Get in places too.

anatolyeltsov:
Thanks for pointing out! 👍🏻 Fixed 😃

And finally implemented [has_been_serviced](https://github.com/OpenTTD/OpenTTD/pull/10541#pullrequestreview-1529132082) variable but I'm pretty sure that my code is not perfect and/or elegant 🥲

Also have some questions:

1. Do we need to change button tooltip text (adding smth like _You need to service this industry at least once to see it's production history_)?
If yes, do we need to have two different texts - for active and inactive buttons?
2. Is it ok that industry window needs to be reopened to see button availability change?

PeterN:
I don't really see the point in having that flag. The history array will still be allocated, just not populated, so it's not going to save any memory.

It's also incorrectly implemented as the flag would need to be saved in the savegame.

anatolyeltsov:
Oh, my bad. Absolutely forgot about savegame. Reverted it.
But I'm still worried about performance/memory usage issues that were talked above.

PeterN (approved):

PeterN:
Could you please squash it back down to one commit? Saves us any mistakes if/when merging.

2TallTyler:
> But I'm still worried about performance/memory usage issues that were talked above.

Perhaps you could provide some benchmarking. Create some common map sizes in vanilla OpenTTD, then load them in this PR and let them run (no companies or vehicles), then see how the file sizes differ. (There may a better way to do this, suggestions welcome!) Just to see approximately how much bigger savegames would be with this feature.

PeterN:
> > But I'm still worried about performance/memory usage issues that were talked above.
> 
> Perhaps you could provide some benchmarking. Create some common map sizes in vanilla OpenTTD, then load them in this PR and let them run (no companies or vehicles), then see how the file sizes differ. (There may a better way to do this, suggestions welcome!) Just to see approximately how much bigger savegames would be with this feature.

Remember we switched the fixed 16-slot array with a vector to reduce some of the overhead already. It's an additional 92 bytes per produced cargo per industry.

PeterN:
> > But I'm still worried about performance/memory usage issues that were talked above.
> 
> Perhaps you could provide some benchmarking. Create some common map sizes in vanilla OpenTTD, then load them in this PR and let them run (no companies or vehicles), then see how the file sizes differ. (There may a better way to do this, suggestions welcome!) Just to see approximately how much bigger savegames would be with this feature.

On a random savegame taken from a multiplayer 14.1 game. Map size 1024x2048, 9 players, 1698 industries:

| Version | Date | Save size | 
| ----- | ---- | ------ |
| 14.1 | 1992-09-14 | 3734884 |
| master | 1992-09-14 | 3737976 |
| master | 1994-11-05 | 3742680 |
| PR10541 | 1992-09-14 | 3740056 |
| PR10541 | 1994-11-05 | 3754416 |

Slightly larger but nothing concerning, I reckon.

PR #10543, opened 2023-03-05
Feature: Region-based pathfinder for ships (buoys no longer required!)

Kuhnovic:
## Remarks for reviewers:
 - The code contains lots of debug drawing code which I need in order to fix things. Please ignore this code for now.
 - If the destination is unreachable an random path is generated. This makes the ship move around aimlessly. This makes it visually clear that the ship is lost, but perhaps it looks a little too silly.
 - The node limiting of the tile-level ship pathfinder is based on the assumption that https://github.com/OpenTTD/OpenTTD/pull/10623 will be merged soon, which will only use the ExitDir-based variant of the ship node.
 - Although there is no need for buoys anymore, users can still use them as waypoints if they want to. I don't think we should remove them.
 
## Motivation / Problem

Anybody who has tried using ships has undoubtedly noticed that ships get lost and need buoys to guide them to the destination. On water the YAPF ship pathfinder quickly reaches its 10000 (default) node limit and throws in the towel. This means we either have to increase the node limit which can lead to large performance hits, or use buoys to create an intermediate destination that is still within this 10000 node limit. It is also very unclear how far buoys have to be spaced apart since you can't see how far the pathfinder is able to search, which depends on the complexity of the map.

## Description
image

The map is divided up into so called WaterRegions of 16x16 tiles (screenshot is from when it was still 8x8). Within such a region connected-component-labeling (CCL) is used to identify different areas of water that are connected. This is done once when a map is created / loaded, and regions are invalidated when changes are made to the map. Invalidated regions get updated automatically when they are needed.

The YAPF ship pathfinder first does a call to a "WaterRegion pathfinder" which only looks at the high level interconnectivity between regions (either via region edges or via aqueducts). This creates a high level path to the destination. The regular YAPF ship pathfinder then looks for a path several regions ahead of its current position. Effectively this gives the tile-level pathfinder an "intermediate destination". When close to the final destination the lookahead is ignored and the pathfinding is simply done using the tile-level YAPF pathfinder.

The high level search is MUCH faster than the tile-level search. I had 5000 ships (the game's total vehicle limit) traveling across a 1024x1024 map with very complicated water geometry, this ran great on my Intel i7 3610QM laptop from 2012. It still runs at about 1,5x the normal game speed when I hold down TAB. 

Another big benefit is quick rejection of cases where no path is available. Such cases are horrible for any A* pathfinder because it requires searching the entire map until there are no new nodes to be added (or until a predetermined node limit is reached, like YAPF does). Since the high level pathfinder effectively searches 256 tiles per region it is able to very quickly determine if there is any path at all, and if there is none then the tile-level pathfinder won't even run.

## Limitations

The main downside is that it no longer guarantees optimal paths. Through extensive tweaking I was able to get very good paths that are often optimal or only slightly suboptimal. IMO a slightly suboptimal path is still a lot better than having to fiddle with buoys to even reach the destination _at all_.

## Credit where credit is due

Shoutout to JazzyJaffa from the tt forums. I stumbled across his thread and that inspired me to go this route
https://www.tt-forums.net/viewtopic.php?t=58905 . I ended up writing my own implementation, simply because I'm stubborn, and because I wanted to make some different design decisions, e.g. go for square regions instead of arbitrary shapes.

## "Why didn't you use Jump Point Search?"

I actually had a sort-of working JPS implementation, but I did not continue with it because:
 - If there is no path, JPS will still search the whole map (assuming a worst case). It will do it more efficiently than regular A*, but it still doesn't scale well to large maps.
 - JPS does not take turning penalties into account, which can cause ships to do unnecessary 90 degree turns relatively often. This can be worked around but is pretty tedious.
 - JPS assumes we can travel to any direction from the start of search, but this is not the case since ships might have to turn or reverse. This makes the "expansion pattern" around the start very complicated.
 - I ran into the patch from JazzyJaffa and got the idea for regions :)

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -1,0 +1,379 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file water_regions.cpp Handles dividing the water in the map into square regions to assist pathfinding. */
+
+#include "stdafx.h"
+#include "map_func.h"
+#include "water_regions.h"
+#include "map_func.h"
+#include "tilearea_type.h"
+#include "track_func.h"
+#include "transport_type.h"
+#include "landscape.h"
+#include "tunnelbridge_map.h"
+#include "follow_track.hpp"
+#include "ship.h"
+
+using TWaterRegionTraversabilityBits = uint16_t;
+constexpr TWaterRegionPatchLabel FIRST_REGION_LABEL = 1;
+constexpr TWaterRegionPatchLabel INVALID_WATER_REGION_PATCH = 0;
+
+static_assert(sizeof(TWaterRegionTraversabilityBits) * 8 == WATER_REGION_EDGE_LENGTH);
+
+static inline TrackBits GetWaterTracks(TileIndex tile) { return TrackStatusToTrackBits(GetTileTrackStatus(tile, TRANSPORT_WATER, 0)); }
+static inline bool IsAqueductTile(TileIndex tile) { return IsBridgeTile(tile) && GetTunnelBridgeTransportType(tile) == TRANSPORT_WATER; }
+
+static inline int GetWaterRegionX(TileIndex tile) { return TileX(tile) / WATER_REGION_EDGE_LENGTH; }
+static inline int GetWaterRegionY(TileIndex tile) { return TileY(tile) / WATER_REGION_EDGE_LENGTH; }
+
+static inline int GetWaterRegionMapSizeX() { return Map::SizeX() / WATER_REGION_EDGE_LENGTH; }
+static inline int GetWaterRegionMapSizeY() { return Map::SizeY() / WATER_REGION_EDGE_LENGTH; }
+
+static inline TWaterRegionIndex GetWaterRegionIndex(int region_x, int region_y) { return GetWaterRegionMapSizeX() * region_y + region_x; }
+static inline TWaterRegionIndex GetWaterRegionIndex(TileIndex tile) { return GetWaterRegionIndex(GetWaterRegionX(tile), GetWaterRegionY(tile)); }
+
+/**
+ * Represents a square section of the map of a fixed size. Within this square individual unconnected patches of water are
+ * identified using a Connected Component Labeling (CCL) algorithm. Note that all information stored in this class applies
+ * only to tiles within the square section, there is no knowledge about the rest of the map. This makes it easy to invalidate
+ * and update a water region if any changes are made to it, such as construction or terraforming.
+ */
+class WaterRegion
+{
+private:
+	std::array<TWaterRegionTraversabilityBits, DIAGDIR_END> edge_traversability_bits{};
+	bool has_cross_region_aqueducts = false;
+	TWaterRegionPatchLabel number_of_patches = 0; // 0 = no water, 1 = one single patch of water, etc...
+	const OrthogonalTileArea tile_area;
+	std::array<TWaterRegionPatchLabel, WATER_REGION_NUMBER_OF_TILES> tile_patch_labels{};
+	bool initialized = false;
+
+	/**
+	 * Returns the local index of the tile within the region. The N corner represents 0,
+	 * the x direction is positive in the SW direction, and Y is positive in the SE direction.
+	 * @param tile Tile within the water region.
+	 * @returns The local index.
+	 */
+	inline int GetLocalIndex(TileIndex tile) const
+	{
+		assert(this->tile_area.Contains(tile));
+		return (TileX(tile) - TileX(this->tile_area.tile)) + WATER_REGION_EDGE_LENGTH * (TileY(tile) - TileY(this->tile_area.tile));
+	}
+
+public:
+	WaterRegion(int region_x, int region_y)
+		: tile_area(TileXY(region_x * WATER_REGION_EDGE_LENGTH, region_y * WATER_REGION_EDGE_LENGTH), WATER_REGION_EDGE_LENGTH, WATER_REGION_EDGE_LENGTH)
+	{}
+
+	OrthogonalTileIterator begin() const { return this->tile_area.begin(); }
+	OrthogonalTileIterator end() const { return this->tile_area.end(); }
+
+	bool IsInitialized() const { return this->initialized; }
+
+	void Invalidate() { this->initialized = false; }
+
+	/**
+	 * Returns a set of bits indicating whether an edge tile on a particular side is traversable or not. These
+	 * values can be used to determine whether a ship can enter/leave the region through a particular edge tile.
+	 * @see GetLocalIndex() for a description of the coordinate system used.
+	 * @param side Which side of the region we want to know the edge traversability of.
+	 * @returns A value holding the edge traversability bits.
+	 */
+	TWaterRegionTraversabilityBits GetEdgeTraversabilityBits(DiagDirection side) const { return edge_traversability_bits[side]; }
+
+	/**
+	 * @returns The amount of individual water patches present within the water region. A value of
+	 * 0 means there is no water present in the water region at all.
+	 */
+	int NumberOfPatches() const { return this->number_of_patches; }
+
+	/**
+	 * @returns Whether the water region contains aqueducts that cross the region boundaries.
+	 */
+	bool HasCrossRegionAqueducts() const { return this->has_cross_region_aqueducts; }
+
+	/**
+	 * Returns the patch label that was assigned to the tile.
+	 * @param tile The tile of which we want to retrieve the label.
+	 * @returns The label assigned to the tile.
+	 */
+	TWaterRegionPatchLabel GetLabel(TileIndex tile) const
+	{
+		assert(this->tile_area.Contains(tile));
+		return this->tile_patch_labels[GetLocalIndex(tile)];
+	}
+
+	/**
+	 * Performs the connected component labeling and other data gathering.
+	 * @see WaterRegion
+	 */
+	void ForceUpdate()
+	{
+		this->has_cross_region_aqueducts = false;
+
+		this->tile_patch_labels.fill(INVALID_WATER_REGION_PATCH);
+
+		for (const TileIndex tile : this->tile_area) {
+			if (IsAqueductTile(tile)) {
+				const TileIndex other_aqueduct_end = GetOtherBridgeEnd(tile);
+				if (!tile_area.Contains(other_aqueduct_end)) {
+					this->has_cross_region_aqueducts = true;
+					break;
+				}
+			}
+		}
+
+		TWaterRegionPatchLabel current_label = 1;
+		TWaterRegionPatchLabel highest_assigned_label = 0;
+
+		/* Perform connected component labeling. This uses a flooding algorithm that expands until no
+		 * additional tiles can be added. Only tiles inside the water region are considered. */
+		for (const TileIndex start_tile : tile_area) {
+			static std::vector<TileIndex> tiles_to_check;
+			tiles_to_check.clear();
+			tiles_to_check.push_back(start_tile);
+
+			bool increase_label = false;
+			while (!tiles_to_check.empty()) {
+				const TileIndex tile = tiles_to_check.back();
+				tiles_to_check.pop_back();
+
+				const TrackdirBits valid_dirs = TrackBitsToTrackdirBits(GetWaterTracks(tile));
+				if (valid_dirs == TRACKDIR_BIT_NONE) continue;
+
+				if (this->tile_patch_labels[GetLocalIndex(tile)] != INVALID_WATER_REGION_PATCH) continue;
+
+				this->tile_patch_labels[GetLocalIndex(tile)] = current_label;
+				highest_assigned_label = current_label;
+				increase_label = true;
+
+				for (const Trackdir dir : SetTrackdirBitIterator(valid_dirs)) {
+					/* By using a TrackFollower we "play by the same rules" as the actual ship pathfinder */
+					CFollowTrackWater ft;
+					if (ft.Follow(tile, dir) && this->tile_area.Contains(ft.m_new_tile)) tiles_to_check.push_back(ft.m_new_tile);
+				}
+			}
+
+			if (increase_label) current_label++;
+		}
+
+		this->number_of_patches = highest_assigned_label;
+		this->initialized = true;
+
+		/* Calculate the traversability (whether the tile can be entered / exited) for all edges. Note that
+		 * we always follow the same X and Y scanning direction, this is important for comparisons later on! */
+		this->edge_traversability_bits.fill(0);
+		const int top_x = TileX(tile_area.tile);
+		const int top_y = TileY(tile_area.tile);
+		for (int i = 0; i < WATER_REGION_EDGE_LENGTH; ++i) {
+			if (GetWaterTracks(TileXY(top_x + i, top_y)) & TRACK_BIT_3WAY_NW) SetBit(this->edge_traversability_bits[DIAGDIR_NW], i); // NW edge
+			if (GetWaterTracks(TileXY(top_x + i, top_y + WATER_REGION_EDGE_LENGTH - 1)) & TRACK_BIT_3WAY_SE) SetBit(this->edge_traversability_bits[DIAGDIR_SE], i); // SE edge
+			if (GetWaterTracks(TileXY(top_x, top_y + i)) & TRACK_BIT_3WAY_NE) SetBit(this->edge_traversability_bits[DIAGDIR_NE], i); // NE edge
+			if (GetWaterTracks(TileXY(top_x + WATER_REGION_EDGE_LENGTH - 1, top_y + i)) & TRACK_BIT_3WAY_SW) SetBit(this->edge_traversability_bits[DIAGDIR_SW], i); // SW edge
+		}
+	}
+
+	/**
+	 * Updates the patch labels and other data, but only if the region is not yet initialized.
+	 */
+	inline void UpdateIfNotInitialized()
+	{
+		if (!this->initialized) ForceUpdate();
+	}
+};
+
+std::vector<WaterRegion> _water_regions;
+
+TileIndex GetTileIndexFromLocalCoordinate(int region_x, int region_y, int local_x, int local_y)
+{
+	assert(local_x >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	assert(local_y >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	return TileXY(WATER_REGION_EDGE_LENGTH * region_x + local_x, WATER_REGION_EDGE_LENGTH * region_y + local_y);
+}
+
+TileIndex GetEdgeTileCoordinate(int region_x, int region_y, DiagDirection side, int x_or_y)
+{
+	assert(x_or_y >= 0 && x_or_y < WATER_REGION_EDGE_LENGTH);
+	switch (side) {
+		case DIAGDIR_NE: return GetTileIndexFromLocalCoordinate(region_x, region_y, 0, x_or_y);
+		case DIAGDIR_SW: return GetTileIndexFromLocalCoordinate(region_x, region_y, WATER_REGION_EDGE_LENGTH - 1, x_or_y);
+		case DIAGDIR_NW: return GetTileIndexFromLocalCoordinate(region_x, region_y, x_or_y, 0);
+		case DIAGDIR_SE: return GetTileIndexFromLocalCoordinate(region_x, region_y, x_or_y, WATER_REGION_EDGE_LENGTH - 1);
+		default: NOT_REACHED();
+	}
+}
+
+WaterRegion &GetUpdatedWaterRegion(uint16_t region_x, uint16_t region_y)
+{
+	WaterRegion &result = _water_regions[GetWaterRegionIndex(region_x, region_y)];
+	result.UpdateIfNotInitialized();
+	return result;
+}
+
+WaterRegion &GetUpdatedWaterRegion(TileIndex tile)
+{
+	WaterRegion &result = _water_regions[GetWaterRegionIndex(tile)];
+	result.UpdateIfNotInitialized();
+	return result;
+}
+
+/**
+ * Returns the index of the water region
+ * @param water_region The Water region to return the index for
+ */
+TWaterRegionIndex GetWaterRegionIndex(const WaterRegionDesc &water_region)
+{
+	return GetWaterRegionIndex(water_region.x, water_region.y);
+}
+
+/**
+ * Returns the center tile of a particular water region.
+ * @param water_region The water region to find the center tile for.
+ * @returns The center tile of the water region.
+ */
+TileIndex GetWaterRegionCenterTile(const WaterRegionDesc &water_region)
+{
+	return TileXY(water_region.x * WATER_REGION_EDGE_LENGTH + (WATER_REGION_EDGE_LENGTH / 2), water_region.y * WATER_REGION_EDGE_LENGTH + (WATER_REGION_EDGE_LENGTH / 2));
+}
+
+/**
+ * Returns basic water region information for the provided tile.
+ * @param tile The tile for which the information will be calculated.
+ */
+WaterRegionDesc GetWaterRegionInfo(TileIndex tile)
+{
+	return WaterRegionDesc{ GetWaterRegionX(tile), GetWaterRegionY(tile) };
+}
+
+/**
+ * Returns basic water region patch information for the provided tile.
+ * @param tile The tile for which the information will be calculated.
+ */
+WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile)
+{
+	WaterRegion &region = GetUpdatedWaterRegion(tile);
+	return WaterRegionPatchDesc{ GetWaterRegionX(tile), GetWaterRegionY(tile), region.GetLabel(tile)};
+}
+
+/**
+ * Marks the water region that tile is part of as invalid.
+ * @param tile Tile within the water region that we wish to invalidate.
+ */
+void InvalidateWaterRegion(TileIndex tile)
+{
+	const int index = GetWaterRegionIndex(tile);
+	if (index > static_cast<int>(_water_regions.size())) return;
+	_water_regions[index].Invalidate();
+}
+
+/**
+ * Calls the provided callback function for all water region patches
+ * accessible from one particular side of the starting patch.
+ * @param water_region_patch Water patch within the water region to start searching from
+ * @param side Side of the water region to look for neigboring patches of water
+ * @param callback The function that will be called for each neighbor that is found
+ */
+static inline void VisitAdjacentWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, DiagDirection side, TVisitWaterRegionPatchCallBack &func)
+{
+	const WaterRegion &current_region = GetUpdatedWaterRegion(water_region_patch.x, water_region_patch.y);
+
+	const TileIndexDiffC offset = TileIndexDiffCByDiagDir(side);
+	const int nx = water_region_patch.x + offset.x;
+	const int ny = water_region_patch.y + offset.y;
+
+	if (nx < 0 || ny < 0 || nx >= GetWaterRegionMapSizeX() || ny >= GetWaterRegionMapSizeY()) return;
+
+	const WaterRegion &neighboring_region = GetUpdatedWaterRegion(nx, ny);
+	const DiagDirection opposite_side = ReverseDiagDir(side);
+
+	/* Indicates via which local x or y coordinates (depends on the "side" parameter) we can cross over into the adjacent region. */
+	const TWaterRegionTraversabilityBits traversability_bits = current_region.GetEdgeTraversabilityBits(side)
+		& neighboring_region.GetEdgeTraversabilityBits(opposite_side);
+	if (traversability_bits == 0) return;
+
+	if (current_region.NumberOfPatches() == 1 && neighboring_region.NumberOfPatches() == 1) {
+		func(WaterRegionPatchDesc{ nx, ny, FIRST_REGION_LABEL }); // No further checks needed because we know there is just one patch for both adjacent regions
+		return;
+	}
+
+	/* Multiple water patches can be reached from the current patch. Check each edge tile individually. */
+	static std::vector<TWaterRegionPatchLabel> unique_labels; // static and vector-instead-of-map for performance reasons
+	unique_labels.clear();
+	for (int x_or_y = 0; x_or_y < WATER_REGION_EDGE_LENGTH; ++x_or_y) {
+		if (!HasBit(traversability_bits, x_or_y)) continue;
+
+		const TileIndex current_edge_tile = GetEdgeTileCoordinate(water_region_patch.x, water_region_patch.y, side, x_or_y);
+		const TWaterRegionPatchLabel current_label = current_region.GetLabel(current_edge_tile);
+		if (current_label != water_region_patch.label) continue;
+
+		const TileIndex neighbor_edge_tile = GetEdgeTileCoordinate(nx, ny, opposite_side, x_or_y);
+		const TWaterRegionPatchLabel neighbor_label = neighboring_region.GetLabel(neighbor_edge_tile);
+		if (std::find(unique_labels.begin(), unique_labels.end(), neighbor_label) == unique_labels.end()) unique_labels.push_back(neighbor_label);
+	}
+	for (TWaterRegionPatchLabel unique_label : unique_labels) func(WaterRegionPatchDesc{ nx, ny, unique_label });
+}
+
+/**
+ * Calls the provided callback function on all accessible water region patches in
+ * each cardinal direction, plus any others that are reachable via aqueducts.
+ * @param water_region_patch Water patch within the water region to start searching from
+ * @param callback The function that will be called for each accessible water patch that is found
+ */
+void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, TVisitWaterRegionPatchCallBack &callback)
+{
+	const WaterRegion &current_region = GetUpdatedWaterRegion(water_region_patch.x, water_region_patch.y);
+
+	/* Visit adjacent water region patches in each cardinal direction */
+	for (DiagDirection side = DIAGDIR_BEGIN; side < DIAGDIR_END; side++) VisitAdjacentWaterRegionPatchNeighbors(water_region_patch, side, callback);
+
+	/* Visit neigboring water patches accessible via cross-region aqueducts */
+	if (current_region.HasCrossRegionAqueducts()) {
+		for (const TileIndex tile : current_region) {
+			if (GetWaterRegionPatchInfo(tile) == water_region_patch && IsAqueductTile(tile)) {
+				const TileIndex other_end_tile = GetOtherBridgeEnd(tile);
+				if (GetWaterRegionIndex(tile) != GetWaterRegionIndex(other_end_tile)) callback(GetWaterRegionPatchInfo(other_end_tile));
+			}
+		}
+	}
+}
+
+std::vector<WaterRegionSaveLoadInfo> GetWaterRegionSaveLoadInfo()
+{
+	std::vector<WaterRegionSaveLoadInfo> result;
+	for (WaterRegion &region : _water_regions) result.push_back({ region.IsInitialized() });
+	return result;
+}
+
+void LoadWaterRegions(const std::vector<WaterRegionSaveLoadInfo> &save_load_info)
+{
+	_water_regions.clear();
+	_water_regions.reserve(save_load_info.size());
+	TWaterRegionIndex index = 0;
+	for (const auto &loaded_region_info : save_load_info) {
+		const int region_x = index % GetWaterRegionMapSizeX();
+		const int region_y = index / GetWaterRegionMapSizeX();
+		WaterRegion &region = _water_regions.emplace_back(region_x, region_y);
+		if (loaded_region_info.initialized) region.ForceUpdate();
+		index++;
+	}
+}
+
+/**
+ * Initializes all water regions. All water tiles will be scanned and interconnected water patches within regions will be identified.
+ */
+void InitializeWaterRegions()
+{
+	_water_regions.clear();
+	_water_regions.reserve(static_cast<size_t>(GetWaterRegionMapSizeX()) * GetWaterRegionMapSizeY());
+
+	for (int region_y = 0; region_y < GetWaterRegionMapSizeY(); region_y++) {
+		for (int region_x = 0; region_x < GetWaterRegionMapSizeX(); region_x++) {
+			_water_regions.emplace_back(region_x, region_y).ForceUpdate();
+		}
+	}
+}
--- a/src/pathfinder/water_regions.h
+++ b/src/pathfinder/water_regions.h
@@ -1,0 +1,73 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file water_regions.h Handles dividing the water in the map into regions to assist pathfinding. */
+
+#ifndef WATER_REGIONS_H
+#define WATER_REGIONS_H
+
+#include "tile_type.h"
+#include "map_func.h"
+
+using TWaterRegionPatchLabel = uint8_t;
+using TWaterRegionIndex = uint;
+
+constexpr int WATER_REGION_EDGE_LENGTH = 16;
+constexpr int WATER_REGION_NUMBER_OF_TILES = WATER_REGION_EDGE_LENGTH * WATER_REGION_EDGE_LENGTH;
+
+/**
+ * Describes a single interconnected patch of water within a particular water region.
+ */
+struct WaterRegionPatchDesc
+{
+	int x; ///< The X coordinate of the water region, i.e. X=2 is the 3rd water region along the X-axis
+	int y; ///< The Y coordinate of the water region, i.e. Y=2 is the 3rd water region along the Y-axis
+	TWaterRegionPatchLabel label; ///< Unique label identifying the patch within the region
+
+	bool operator==(const WaterRegionPatchDesc &other) const { return x == other.x && y == other.y && label == other.label; }
+	bool operator!=(const WaterRegionPatchDesc &other) const { return !(*this == other); }
+};
+
+
+/**
+ * Describes a single square water region.
+ */
+struct WaterRegionDesc
+{
+	int x; ///< The X coordinate of the water region, i.e. X=2 is the 3rd water region along the X-axis
+	int y; ///< The Y coordinate of the water region, i.e. Y=2 is the 3rd water region along the Y-axis
+
+	WaterRegionDesc(const int x, const int y) : x(x), y(y) {}
+	WaterRegionDesc(const WaterRegionPatchDesc &water_region_patch) : x(water_region_patch.x), y(water_region_patch.y) {}
+
+	bool operator==(const WaterRegionDesc &other) const { return x == other.x && y == other.y; }
+	bool operator!=(const WaterRegionDesc &other) const { return !(*this == other); }
+};
+
+TWaterRegionIndex GetWaterRegionIndex(const WaterRegionDesc &water_region);
+
+TileIndex GetWaterRegionCenterTile(const WaterRegionDesc &water_region);
+
+WaterRegionDesc GetWaterRegionInfo(TileIndex tile);
+WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile);
+
+void InvalidateWaterRegion(TileIndex tile);
+
+using TVisitWaterRegionPatchCallBack = std::function<void(const WaterRegionPatchDesc &)>;
+void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, TVisitWaterRegionPatchCallBack &callback);
+
+void InitializeWaterRegions();
+
+struct WaterRegionSaveLoadInfo
+{
+	bool initialized;
+};
+
+std::vector<WaterRegionSaveLoadInfo> GetWaterRegionSaveLoadInfo();
+void LoadWaterRegions(const std::vector<WaterRegionSaveLoadInfo> &save_load_info);
+
+#endif /* WATER_REGIONS_H */
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -14,46 +14,64 @@
 
 #include "yapf.hpp"
 #include "yapf_node_ship.hpp"
+#include "yapf_ship_regions.h"
+#include "../water_regions.h"
 
 #include "../../safeguards.h"
+
+constexpr int NUMBER_OR_WATER_REGIONS_LOOKAHEAD = 4;
+constexpr int MAX_SHIP_PF_NODES = (NUMBER_OR_WATER_REGIONS_LOOKAHEAD + 1) * WATER_REGION_NUMBER_OF_TILES * 4; // 4 possible exit dirs per tile.
+
+constexpr int SHIP_LOST_PATH_LENGTH = 8; // The length of the (aimless) path assigned when a ship is lost.
 
 template <class Types>
 class CYapfDestinationTileWaterT
 {
 public:
-	typedef typename Types::Tpf Tpf;                     ///< the pathfinder class (derived from THIS class)
+	typedef typename Types::Tpf Tpf;                     ///< the pathfinder class (derived from THIS class).
 	typedef typename Types::TrackFollower TrackFollower;
-	typedef typename Types::NodeList::Titem Node;        ///< this will be our node type
-	typedef typename Node::Key Key;                      ///< key to hash tables
+	typedef typename Types::NodeList::Titem Node;        ///< this will be our node type.
+	typedef typename Node::Key Key;                      ///< key to hash tables.
 
 protected:
 	TileIndex    m_destTile;
 	TrackdirBits m_destTrackdirs;
 	StationID    m_destStation;
 
+	bool                 m_has_intermediate_dest = false;
+	TileIndex            m_intermediate_dest_tile;
+	WaterRegionPatchDesc m_intermediate_dest_region_patch;
+
 public:
 	void SetDestination(const Ship *v)
 	{
 		if (v->current_order.IsType(OT_GOTO_STATION)) {
-			m_destStation   = v->current_order.GetDestination();
-			m_destTile      = CalcClosestStationTile(m_destStation, v->tile, STATION_DOCK);
+			m_destStation = v->current_order.GetDestination();
+			m_destTile = CalcClosestStationTile(m_destStation, v->tile, STATION_DOCK);
 			m_destTrackdirs = INVALID_TRACKDIR_BIT;
 		} else {
-			m_destStation   = INVALID_STATION;
-			m_destTile      = v->dest_tile;
+			m_destStation = INVALID_STATION;
+			m_destTile = v->dest_tile;
 			m_destTrackdirs = TrackStatusToTrackdirBits(GetTileTrackStatus(v->dest_tile, TRANSPORT_WATER, 0));
 		}
 	}
 
+	void SetIntermediateDestination(const WaterRegionPatchDesc &water_region_patch)
+	{
+		m_has_intermediate_dest = true;
+		m_intermediate_dest_tile = GetWaterRegionCenterTile(water_region_patch);
+		m_intermediate_dest_region_patch = water_region_patch;
+	}
+
 protected:
-	/** to access inherited path finder */
-	inline Tpf &Yapf()
+	/** To access inherited path finder. */
+	inline Tpf& Yapf()
 	{
 		return *static_cast<Tpf*>(this);
 	}
 
 public:
-	/** Called by YAPF to detect if node ends in the desired destination */
+	/** Called by YAPF to detect if node ends in the desired destination. */
 	inline bool PfDetectDestination(Node &n)
 	{
 		return PfDetectDestinationTile(n.m_segment_last_tile, n.m_segment_last_td);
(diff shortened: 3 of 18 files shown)


rubidium42:
There is one thing to consider with the connectivity of a region; a region can be connected to multiple other regions, but you might not be able to go through the region. See the attached example where a ship gets lost, while it works just fine in master.
[poc.zip](https://github.com/OpenTTD/OpenTTD/files/10911879/poc.zip)

Kuhnovic:
> There is one thing to consider with the connectivity of a region; a region can be connected to multiple other regions, but you might not be able to go through the region. See the attached example where a ship gets lost, while it works just fine in master. [poc.zip](https://github.com/OpenTTD/OpenTTD/files/10911879/poc.zip)

The algorithm takes this into account. I opened your savegame and it runs fine right out of the gate, so I'm afraid it's a bug in the region invalidation / updating, which has been very difficult to get right. All regions are updated once when a game is loaded, so that would explain why it worked for me. Can you load your savegame again and (hopefully) confirm that it runs just fine for you as well?

rubidium42:
> > There is one thing to consider with the connectivity of a region; a region can be connected to multiple other regions, but you might not be able to go through the region. See the attached example where a ship gets lost, while it works just fine in master. [poc.zip](https://github.com/OpenTTD/OpenTTD/files/10911879/poc.zip)
> 
> @rubidium42 The algorithm takes this into account. I opened your savegame and it runs fine right out of the gate, so I'm afraid it's a bug in the region invalidation / updating, which has been very difficult to get right. All regions are updated once when a game is loaded, so that would explain why it worked for me. Can you load your savegame again and (hopefully) confirm that it runs just fine for you as well?

Yes, I can confirm that after reloading it works. That, sadly, makes the problem actually even worse, as we now know the current implementation causes desyncs. See docs/desync.md for more information about them, but essentially this is a cache mismatch.

Kuhnovic:
> Yes, I can confirm that after reloading it works. That, sadly, makes the problem actually even worse, as we now know the current implementation causes desyncs. See docs/desync.md for more information about them, but essentially this is a cache mismatch.

I was able to reproduce the issue, should be fixed now. After reading more about desyncs I realize that the way I store my region data needs to be made more robust. I think I should try to store everything in the m1,m2 etc tiledata an not in a separate data structure. I'll see if I can come up with a different solution for this.

Kuhnovic:
Alright, big update. I've gone over the code and cleaned things up. I've also tried to decouple the pathfinding logic and the water region logic further. They are both quite complicated, so I want to separate those two domains as best as possible.

I have still left the debug drawing there, it is absolutely essential to fix bugs. I will remove it later.

src/pathfinder/yapf/yapf_base.hpp
```diff
@@ -43,6 +44,9 @@
  *  declaration. There are some examples. For another example look at
  *  test_yapf.h (part or unittest project).
  */
+
+#define ENABLE_YAPF_DEBUG_DRAWING 1 // TODO KB Remove
```
rubidium42: Maybe move all the debug code that you want to remove later on to a separate commit. That would also make it easier to retain in some local branch if you want to use it later on, or maybe to fix-up/improve to be able to be used more generically. Although filling the code with lots of `#ifdef`s for debugging isn't that pretty.

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+	Node       *m_hash_next;
+	Node       *m_parent;
+	int         m_cost;
+	int         m_estimate;
+
+	/* TODO KB Remove these functions, they exist only to allow the debug drawing code in yapf_base.hpp to compile */
+	Trackdir GetTrackdir() { return TRACKDIR_BEGIN; }
+	Tile GetTile() { return 0;}
+
+	inline void Set(Node* parent, const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
	inline void Set(Node *parent, const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+	{
+		m_key.Set(water_region_patch);
+		m_hash_next = nullptr;
+		m_parent = parent;
+		m_cost = 0;
+		m_estimate = 0;
+	}
+
+	inline void Set(Node* parent, const Key& key)
```
rubidium42:
```suggestion
	inline void Set(Node *parent, const Key &key)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+		const int dx = m_key.m_water_region_patch.x - m_parent->m_key.m_water_region_patch.x;
+		const int dy = m_key.m_water_region_patch.y - m_parent->m_key.m_water_region_patch.y;
+		if (dx > 0 && dy == 0) return DIAGDIR_SW;
+		if (dx < 0 && dy == 0) return DIAGDIR_NE;
+		if (dx == 0 && dy > 0) return DIAGDIR_SE;
+		if (dx == 0 && dy < 0) return DIAGDIR_NW;
+		return INVALID_DIAGDIR;
+	}
+
+	inline Node* GetHashNext() { return m_hash_next; }
```
rubidium42:
```suggestion
	inline Node *GetHashNext() { return m_hash_next; }
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+	bool HasOrigin(const WaterRegionPatchDesc &water_region_patch)
+	{
+		return std::find(m_origin_keys.begin(), m_origin_keys.end(), CYapfRegionPatchNodeKey{ water_region_patch }) != m_origin_keys.end();
+	}
+
+	void PfSetStartupNodes()
+	{
+		for (const CYapfRegionPatchNodeKey &origin_key : m_origin_keys)
+		{
```
rubidium42:
```suggestion
		for (const CYapfRegionPatchNodeKey &origin_key : m_origin_keys) {
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+template <class Types>
+class CYapfOriginRegionT
+{
+public:
+	typedef typename Types::Tpf Tpf;              ///< the pathfinder class (derived from THIS class)
+	typedef typename Types::NodeList::Titem Node; ///< this will be our node type
+	typedef typename Node::Key Key;               ///< key to hash tables
+
+protected:
+	inline Tpf& Yapf() { return *static_cast<Tpf*>(this); }
```
rubidium42:
```suggestion
	inline Tpf &Yapf() { return *static_cast<Tpf*>(this); }
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+private:
+	std::vector<CYapfRegionPatchNodeKey> m_origin_keys;
+
+public:
+	void AddOrigin(const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
	void AddOrigin(const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+	Key m_dest;
+
+public:
+	void SetDestination(const WaterRegionPatchDesc& water_region_patch)
+	{
+		m_dest.Set(water_region_patch);
+	}
+
+protected:
+	Tpf& Yapf() { return *static_cast<Tpf*>(this); }
```
rubidium42:
```suggestion
	Tpf &Yapf() { return *static_cast<Tpf*>(this); }
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+		const WaterRegionPatchDesc start_water_region_patch = GetWaterRegionPatchInfo(start_tile);
+
+		/* We reserve one pathfinder node per water region. Some regions will have separate patches of water which results in multiple
+		 * nodes, where others regions don't contain any water or are simply unreachable. The chance of hitting this node limit is small. */
+		Tpf pf(static_cast<int>(Map::Size()) / WATER_REGION_NUMBER_OF_TILES);
+		pf.SetDestination(start_water_region_patch);
+
+		if (v->current_order.IsType(OT_GOTO_STATION)) {
+			DestinationID station_id = v->current_order.GetDestination();
+			const BaseStation* station = BaseStation::Get(station_id);
```
rubidium42:
```suggestion
			const BaseStation *station = BaseStation::Get(station_id);
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+			TileArea tile_area;
+			station->GetTileArea(&tile_area, STATION_DOCK);
+			for (const auto& tile : tile_area) {
```
rubidium42:
```suggestion
			for (const auto &tile : tile_area) {
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+		std::vector<WaterRegionPatchDesc> path = { start_water_region_patch };
+		path.reserve(max_returned_path_length);
+		if (pf.HasOrigin(start_water_region_patch)) return path;
+
+		/* Find best path */
+		if (!pf.FindPath(v)) return {}; // Path not found
+
+		Node *node = pf.GetBestNode();
+		for (int i = 0; i < max_returned_path_length - 1; ++i) {
+			if (node) {
```
rubidium42:
```suggestion
			if (node != nullptr) {
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+				node = node->m_parent;
+				if (node) path.push_back(node->m_key.m_water_region_patch);
```
rubidium42:
```suggestion
				if (node != nullptr) path.push_back(node->m_key.m_water_region_patch);
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+			}
+		}
+
+		/* Draw the rest of the path black */
+		while (node)
+		{
+			node = node->m_parent;
+			if (node) {
```
rubidium42:
```suggestion
			if (node != nullptr) {
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+		Node *node = pf.GetBestNode();
+		for (int i = 0; i < max_returned_path_length - 1; ++i) {
+			if (node) {
+				node = node->m_parent;
+				if (node) path.push_back(node->m_key.m_water_region_patch);
+			}
+		}
+
+		/* Draw the rest of the path black */
+		while (node)
```
rubidium42:
```suggestion
		while (node != nullptr)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+	/** Called by YAPF to calculate the cost from the origin to the given node.
+	 *  Calculates only the cost of given node, adds it to the parent node cost
+	 *  and stores the result into Node::m_cost member */
+	inline bool PfCalcCost(Node &n, const TrackFollower *tf)
+	{
+		n.m_cost = n.m_parent->m_cost + ManhattanDistance(n.m_key, n.m_parent->m_key);
+
+		/* Incentivise zigzagging by adding a slight penalty when the search continues in the same direction. */
+		Node *grandparent = n.m_parent->m_parent;
+		if (grandparent) {
```
rubidium42:
```suggestion
		if (grandparent != nullptr) {
```

src/saveload/CMakeLists.txt
```diff
@@ -44,4 +44,5 @@ add_files(
     town_sl.cpp
     vehicle_sl.cpp
     waypoint_sl.cpp
+	water_regions_sl.cpp
```
rubidium42:
```suggestion
    water_regions_sl.cpp
```

src/viewport.cpp
```diff
@@ -1266,10 +1288,11 @@ static void ViewportAddLandscape()
...
 				if (min_visible_height < potential_bridge_height + MAX_TILE_EXTENT_TOP) last_row = false;
-			} else {
-				/* Outside of map. If we are on the north border of the map, there may still be a bridge visible,
-				 * so we need to loop over more rows to possibly find one. */
-				if ((tilecoord.x <= 0 || tilecoord.y <= 0) && min_visible_height < potential_bridge_height + MAX_TILE_EXTENT_TOP) last_row = false;
+			}
+ else {
+	 /* Outside of map. If we are on the north border of the map, there may still be a bridge visible,
+		* so we need to loop over more rows to possibly find one. */
+		if ((tilecoord.x <= 0 || tilecoord.y <= 0) && min_visible_height < potential_bridge_height + MAX_TILE_EXTENT_TOP) last_row = false;
```
rubidium42: What happened here?

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,83 @@
...
+ /** @file water_regions.h Handles dividing the water in the map into regions to assist pathfinding. */
+
+#ifndef WATER_REGIONS_H
+#define WATER_REGIONS_H
+
+#include "tile_type.h"
+#include "map_func.h"
+
+#include <vector>
+#include <functional>
```
rubidium42: These are included via `stdafx.h`, so not needed here anymore.

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,83 @@
...
+
+using TWaterRegionPatchLabel = uint8;
+using TWaterRegionIndex = uint;
```
rubidium42:
```suggestion
using TWaterRegionPatchLabel = uint8_t;
using TWaterRegionIndex = uint;
```
We'll be going to the std types, so `uint8_t`.

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,83 @@
...
+
+TWaterRegionIndex GetWaterRegionIndex(const WaterRegionDesc &water_region);
+
+WaterRegionDesc GetWaterRegionInfo(TileIndex tile);
+WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile);
+
+void InvalidateWaterRegion(TileIndex tile);
+
+using TVisitWaterRegionPatchCallBack = std::function<void(const WaterRegionPatchDesc &)>;
+void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc& water_region_patch, TVisitWaterRegionPatchCallBack& callback);
```
rubidium42:
```suggestion
void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, TVisitWaterRegionPatchCallBack &callback);
```

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Returns the center tile of a particular water region.
+ * @param water_region The water region to find the center tile for.
+ * @returns The center tile of the water region.
+ */
+inline TileIndex GetWaterRegionCenterTile(const WaterRegionDesc& water_region)
```
rubidium42:
```suggestion
inline TileIndex GetWaterRegionCenterTile(const WaterRegionDesc &water_region)
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+#include "landscape.h"
+#include "tunnelbridge_map.h"
+#include "follow_track.hpp"
+#include "ship.h"
+#include "viewport_func.h" // For debug drawing
+
+#include <vector>
+#include <array>
+#include <unordered_set>
+#include <unordered_map>
```
rubidium42: Are these still needed? Since some of them are included in `stdafx.h`.

rubidium42: `unorderd_set`/`unordered_map` seem a bit tricky in the pathfinder, as they can yield different order of elements on different platforms. So, that's something to take into account when testing/reviewing.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+				for (const Trackdir dir : SetTrackdirBitIterator(valid_dirs)) {
+					/* By using a TrackFollower we "play by the same rules" as the actual ship pathfinder */
+					CFollowTrackWater ft;
+					if (ft.Follow(tile, dir) && this->tile_area.Contains(ft.m_new_tile)) tiles_to_check.push_back(ft.m_new_tile);
+				}
+			}
+
+			if (increase_label) current_label++;
+		}
+		for (const auto [tile, label] : temp_labels) tile_patch_labels[GetLocalIndex(tile)] = label;
```
rubidium42: When multiple tiles within the same local index have a different labels, then this is nondeterministic and a desync threat.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+	assert(local_x >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	assert(local_y >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	return TileXY(WATER_REGION_EDGE_LENGTH * region_x + local_x, WATER_REGION_EDGE_LENGTH * region_y + local_y);
+}
+
+TileIndex GetEdgeTileCoordinate(int region_x, int region_y, DiagDirection side, int x_or_y) // TODO better name
+{
+	assert(x_or_y >= 0 && x_or_y < WATER_REGION_EDGE_LENGTH);
+	switch (side)
+	{
```
rubidium42:
```suggestion
	switch (side) {
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+	const WaterRegion &neighboring_region = GetUpdatedWaterRegion(nx, ny);
+	const DiagDirection opposite_side = ReverseDiagDir(side);
+
+	/* Indicates via which local x or y coordinates (depends on the "side" parameter) we can cross over into the adjacent region. */
+	const TWaterRegionTraversabilityBits traversability_bits = current_region.GetEdgeTraversabilityBits(side)
+		& neighboring_region.GetEdgeTraversabilityBits(opposite_side);
+	if (traversability_bits == 0) return;
+
+	if (current_region.NumberOfPatches() == 1 && neighboring_region.NumberOfPatches() == 1) {
+		func(WaterRegionPatchDesc{ nx, ny, FIRST_REGION_LABEL }); // No further checks needed because we know there is just one patch for both adjacent regions
```
rubidium42, line 300: Maybe a `return` here, and un-indenting the `else`.  Now I'm wondering whether there is something that needs to happen after the `if-else` when reading this line.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+	}
+}
+
+/**
+ * Calls the provided callback function on all accessible water region patches in
+ * each cardinal direction, plus any others that are reachable via aqueducts.
+ * @param water_region_patch Water patch within the water region to start searching from
+ * @param callback The function that will be called for each accessible water patch that is found
+ */
+void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc& water_region_patch, TVisitWaterRegionPatchCallBack& callback)
```
rubidium42:
```suggestion
void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, TVisitWaterRegionPatchCallBack &callback)
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+void DEBUG_MarkRegion(TWaterRegionIndex region_index, int color, bool override)
+{
+	DEBUG_MarkRegion(GetWaterRegionX(region_index), GetWaterRegionY(region_index), -1, color, override);
+}
+
+std::vector<WaterRegionSaveLoadInfo> GetWaterRegionSaveLoadInfo()
+{
+	std::vector<WaterRegionSaveLoadInfo> result;
+	for (WaterRegion &region : _water_regions) result.push_back(WaterRegionSaveLoadInfo{ region.IsInitialized()});
```
rubidium42: Inconsistent space in the `WaterRegionSaveLoadInfo`, but why not use `emplace_back`?
```suggestion
	for (WaterRegion &region : _water_regions) result.emplace_back(region.IsInitialized());
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+void InitializeWaterRegions()
+{
+	DEBUG_clearDrawInstructions();
+
+	_water_regions.clear();
+	_water_regions.reserve(static_cast<size_t>(GetWaterRegionMapSizeX()) * GetWaterRegionMapSizeY());
+
+	for (int region_y = 0; region_y < GetWaterRegionMapSizeY(); region_y++) {
+		for (int region_x = 0; region_x < GetWaterRegionMapSizeX(); region_x++) {
+			_water_regions.push_back(WaterRegion{ region_x, region_y });
```
rubidium42:
```suggestion
			_water_regions.emplace_back(region_x, region_y);
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,405 @@
...
+	_water_regions.clear();
+	_water_regions.reserve(save_load_info.size());
+	TWaterRegionIndex index = 0;
+	for (const auto &loaded_region_info : save_load_info) {
+		const int region_x = index % GetWaterRegionMapSizeX();
+		const int region_y = index / GetWaterRegionMapSizeX();
+		WaterRegion water_region{ region_x, region_y};
+		if (loaded_region_info.initialized) water_region.ForceUpdate();
+		if (!loaded_region_info.initialized) DEBUG_MarkRegion(GetWaterRegionX(index), GetWaterRegionY(index), -1, 1); // 0 = normal, 1 = red, 2 = blue, 3 = black, 4 = transparent
+		_water_regions.push_back(std::move(water_region));
```
rubidium42: Maybe use the "method" of `InitializeWaterRegions`, i.e. just put the element in the `_water_regions` and then `_water_regions.back().ForceUpdate()` in the `if`-statement?

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+	static_assert(sizeof(TWaterRegionPatchLabel) == sizeof(byte)); // Important for the hash calculation
+
+	inline void Set(const WaterRegionPatchDesc& water_region_patch)
+	{
+		m_water_region_patch = water_region_patch;
+	}
+
+	inline int CalcHash() const { return m_water_region_patch.label | GetWaterRegionIndex(m_water_region_patch) << 8; }
+	inline bool operator == (const CYapfRegionPatchNodeKey &other) const { return CalcHash() == other.CalcHash(); }
```
rubidium42:
```suggestion
	inline bool operator==(const CYapfRegionPatchNodeKey &other) const { return CalcHash() == other.CalcHash(); }
```
To be consistent with earlier cases.

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -31,20 +38,33 @@ class CYapfDestinationTileWaterT
...
-			m_destStation   = v->current_order.GetDestination();
-			m_destTile      = CalcClosestStationTile(m_destStation, v->tile, STATION_DOCK);
+			m_destStation = v->current_order.GetDestination();
+			m_destTile = CalcClosestStationTile(m_destStation, v->tile, STATION_DOCK);
 			m_destTrackdirs = INVALID_TRACKDIR_BIT;
-		} else {
-			m_destStation   = INVALID_STATION;
-			m_destTile      = v->dest_tile;
+		}
+		else {
```
rubidium42:
```suggestion
		} else {
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -31,20 +38,33 @@ class CYapfDestinationTileWaterT
...
+			m_destStation = INVALID_STATION;
+			m_destTile = v->dest_tile;
 			m_destTrackdirs = TrackStatusToTrackdirBits(GetTileTrackStatus(v->dest_tile, TRANSPORT_WATER, 0));
 		}
 	}
 
+	void SetIntermediateDestination(const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
	void SetIntermediateDestination(const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -158,43 +219,73 @@ class CYapfFollowShipT
...
+			 * to the final destination tile. The low-level pathfinder might actually prefer a different docking tile in a nearby region. Without
+			 * caching the full path the ship can get stuck in a loop. */
+			const WaterRegionPatchDesc end_water_patch = GetWaterRegionPatchInfo(node->GetTile());
+			const WaterRegionPatchDesc start_water_patch = GetWaterRegionPatchInfo(tile);
+			std::deque<std::pair<TileIndex, Trackdir>> draw_stuff; // TODO KB remove
+			while (node->m_parent) {
+				const WaterRegionPatchDesc node_water_patch = GetWaterRegionPatchInfo(node->GetTile());
+				if (node_water_patch == start_water_patch || (!is_intermediate_destination && node_water_patch != end_water_patch)) {
+					path_cache.push_front(node->GetTrackdir());
+					draw_stuff.push_front({ node->GetTile(), node->GetTrackdir() });
```
rubidium42:
```suggestion
					draw_stuff.emplace_front(node->GetTile(), node->GetTrackdir());
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -209,33 +300,44 @@ class CYapfFollowShipT
...
+		}
+
+		/* Create pathfinder instance */
+		Tpf pf(MAX_SHIP_PF_NODES);
+		/* Set origin and destination nodes */
 		if (trackdir == nullptr) {
 			pf.SetOrigin(tile, TrackdirToTrackdirBits(td1) | TrackdirToTrackdirBits(td2));
-		} else {
+		}
+		else {
```
rubidium42:
```suggestion
		} else {
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -209,33 +300,44 @@ class CYapfFollowShipT
...
 		pf.SetDestination(v);
-		/* find best path */
+		if (high_level_path.size() > 1) pf.SetIntermediateDestination(high_level_path.back());
+		pf.RestrictSearch(high_level_path);
+
+		/* Find best path */
 		if (!pf.FindPath(v)) return false;
 
-		Node *pNode = pf.GetBestNode();
+		Node* pNode = pf.GetBestNode();
```
rubidium42: Moving the `*` in the wrong direction.

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -209,33 +300,44 @@ class CYapfFollowShipT
...
 		while (pNode->m_parent != nullptr) {
 			pNode = pNode->m_parent;
 		}
 
 		Trackdir best_trackdir = pNode->GetTrackdir();
 		if (trackdir != nullptr) {
 			*trackdir = best_trackdir;
-		} else {
+		}
+		else {
```
rubidium42:
```suggestion
		} else {
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -268,7 +370,8 @@ class CYapfCostShipT
 		if (HasTrackdir(TrackdirCrossesTrackdirs(td1), td2)) {
 			/* 90-deg curve penalty */
 			return Yapf().PfGetSettings().ship_curve90_penalty;
-		} else if (td2 != NextTrackdir(td1)) {
+		}
+		else if (td2 != NextTrackdir(td1)) {
```
rubidium42:
```suggestion
		} else if (td2 != NextTrackdir(td1)) {
```

rubidium42 (commented):
I'm seeing potential desync risks, which need to be investigated further.

Kuhnovic:
@rubidium42 Thanks for the review, I'll try to address your points soon.

There certainly is desync potential, in fact it wouldn't be the first time. Let me give an high-level overview of what I did to prevent desyncs:
- Every water region has an "initialized" member. This boolean is stored in the savegame.
- Terraforming and any water-related construction sets the initialized member of the corresponding water region to false
- When the high-level ship pathfinder runs and bumps into a non-initialized water region, it tells it to update. This means individual patches of water are identified within the region, and some other stuff. After this the initialized boolean becomes true, and the pathfinder can use the information stored for the water region. (Bonus: this lazy updating also helps to prevent constant updating to water regions when lots of sea tiles are getting flooded)
- When loading a savegame, all water regions that are marked as initialized in the savegame get updated. The others are left uninitialized. This keeps things bug-compatible: there have been bugs where certain construction actions did not properly invalidate the water region. Initializing all water regions would "fix" that water region when loading a savegame, causing a desync. This exact thing was the cause of the desync mentioned in one of the earlier comments.

Assuming the tile data and pathfinder calls are all properly in sync, the pathfinder on all clients should all be updating their water regions with the same game state at the same time (when the pathfinder bumps into them). As long as the initialized state of the water regions is properly synced then this should work. Unless I'm missing something that is. I'm certainly no expect on this topic, so please take any misplaced confidence with a grain of salt :)

_P.S. Reading this makes me realize that "initialized" is probably not the right term for this, but you get the idea._

src/viewport.cpp
Kuhnovic: I apparently completely murdered this function. All of the changes in viewport.cpp are debug drawing related and will be reverted. I'll make sure this gets a TODO.

src/pathfinder/water_regions.cpp
Kuhnovic: Looks like I don't really need the temp_labels anymore, I can directly write to the std::array instead with some modifications. It's cleaner too.

src/pathfinder/water_regions.cpp
Kuhnovic: I did something slightly different, but along those lines

Kuhnovic:
@rubidium42 Valid point about the std::unordered_map, that can indeed cause a different labeling order on different platforms. Should be fixed now, see my updated code.

@glx22 The case you describe does not happen, because in such a case all clients will have initialized-regions-that-should-have-been invalidated. As long as the "initialized-ness" of all water regions across clients (and any savegame when joining) matches we should be in sync.

TrueBrain:
We would like to test this out in a playtest, but for our build-system to build this PR, it kinda needs a rebase. Are you able to make that happen one of these days? Tnx :)

Kuhnovic:
> We would like to test this out in a playtest, but for our build-system to build this PR, it kinda needs a rebase. Are you able to make that happen one of these days? Tnx :)

She should be good to go! Fingers crossed that I didn't break anything

glx22:
Test build available on https://cdn.openttd.org/openttd-branches/pr10543/20231211-pr10543-g6f73cc8dcf/

2TallTyler:
Binaries to test this PR can be found at: https://cdn.openttd.org/openttd-branches/pr10543/20231211-pr10543-g6f73cc8dcf/

2TallTyler:
The test game this past weekend was a success, and I think proved to those of us present that this is a fantastic improvement to OpenTTD. Responding to your "notes for reviewers":
- Wandering lost ships makes sense to me. I intentionally built a ship with impossible orders during the test game and it was fine.
- I don't think #10623 will be merged soon. There's been some discussion about waiting for more data from the survey (which will be released in 14.0), or at the very least, it needs somebody to step up and make the new pathfinder settings mentioned in the comments.
- I agree that we should keep buoys, of course.

So, what's next? Clean it up, fix the warnings, move the debug code to its own commit (if you still want to keep it around for now), maybe split the actual change into logical commits to help us review it?

Kuhnovic:
I will remove the debug drawing for sure. Not sure how much sense it will make to split the code into separate commits, but I'll have a look.

I can make it work without all the changes from #10623. The main thing was the TileIndex/Trackdir vs TileIndex/DiagDir as nodekey. This has certain implications on the node limit since a Trackdir can have 12 different values per tile, Diagdir can only have 4. Not a showstopper by any means, just something I need to take into account.

The xmas holidays are coming up and I have quite a busy schedule lately, so I can't make any promises on when I'll work on this. But it's definitely on the top of my development-hobby-list ;)

LordAro:
I'm not quite sure what form it should take, but I think it'd be good to have the debug drawing in the game, hidden behind debug mode and/or suitable hotkeys

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,72 @@
...
+#include "tile_type.h"
+#include "map_func.h"
+
+using TWaterRegionPatchLabel = uint8_t;
+using TWaterRegionIndex = uint;
+
+constexpr int WATER_REGION_EDGE_LENGTH = 16;
+constexpr int WATER_REGION_NUMBER_OF_TILES = WATER_REGION_EDGE_LENGTH * WATER_REGION_EDGE_LENGTH;
+
+/* Describes a single interconnected patch of water within a particular water region */
```
TrueBrain: Please make these doxygen comments. And sentences end with a dot. I am mentioning this once, but goes for all further statements.

```
/**
 * ...
 */
```

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,72 @@
...
+struct WaterRegionPatchDesc
+{
+	int x; // Water region X coordinate
```
TrueBrain: Please make these doxygen comments (`///<`). And sentences end with a dot. Goes for all further statements.

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -54,28 +73,34 @@ class CYapfDestinationTileWaterT
...
 	{
 		return PfDetectDestinationTile(n.m_segment_last_tile, n.m_segment_last_td);
 	}
 
 	inline bool PfDetectDestinationTile(TileIndex tile, Trackdir trackdir)
 	{
-		if (m_destStation != INVALID_STATION) {
-			return IsDockingTile(tile) && IsShipDestinationTile(tile, m_destStation);
+		if (m_has_intermediate_dest) {
+			/* GetWaterRegionInfo is much faster than GetWaterRegionPatchInfo so we try that first */
```
TrueBrain: Sentences end with a dot (and here too, please apply this to everywhere :D )

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -158,43 +218,73 @@ class CYapfFollowShipT
...
-				/* Skip tiles at end of path near destination. */
-				if (skip > 0) skip--;
-				if (skip == 0 && steps > 0 && steps < YAPF_SHIP_PATH_CACHE_LENGTH) {
-					path_cache.push_front(pNode->GetTrackdir());
+		/* Try one time without restricting the search area, which generally results in better and more natural looking paths.
+		 * However the pathfinder can hit the node limit in certain situations such as long aqueducts or maze-like terrain.
+		 * If that happens we run the pathfinder again, but restricted only to the regions provided by the region pathfinder. */
+		for (int attempt = 0; attempt < 2; ++attempt) {
+			DEBUG_clearDrawInstructions();
+			for (const auto& region_patch : high_level_path) DEBUG_MarkRegion(region_patch.x, region_patch.y, region_patch.label, attempt == 0 ? 4 : 1, true);
```
TrueBrain: more nitpicking: `&` should be on the right side :P

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,330 @@
...
+/* We don't need a follower but the YAPF requires one */
+struct DummyFollower : public CFollowTrackWater {};
+
+/** Config struct of YAPF for route planning.
+ *  Defines all 6 base YAPF modules as classes providing services for CYapfBaseT.
+ */
+template <class Tpf_, class Tnode_list>
+struct CYapfRegion_TypesT
+{
+	/** Types - shortcut for this struct type */
```
TrueBrain: Bit odd comments. Repeating of the variable name isn't needed, but seems they can use a freshening up in general :)

TrueBrain (commented):
Bunch of nitpicking :)

src/pathfinder/yapf/yapf_ship_regions.cpp
Kuhnovic: I've improved the style of this struct for the yapf_ship and yapf_ship_regions classes.

Kuhnovic:
All debug drawing code should now be removed. I fully agree with LordAro's comment that some form of debug drawing would be convenient, but I won't make that part of the scope of this PR. It's big enough already :p

I've also been thinking about simplifying the water region invalidation. There are InvalidateWaterRegions calls in many places in the code, and even though I spend quite some time on it there could be edge cases that I've missed. This could lead to water regions not getting updated on an important event, say for example when a dock is built. This in turn could leads to ships not finding a path or taking huge detours.

As an alternative I can simply invalidate a water region when DoClearSquare (in landscape.cpp) is called. This would essentially invalidate the water region on any change. That includes changes to non-water tiles. It would be quite robust, but also a bit of a primitive approach. If you have an opinion on this please let me know.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+inline bool IsAqueductTile(TileIndex tile) { return IsBridgeTile(tile) && GetTunnelBridgeTransportType(tile) == TRANSPORT_WATER; }
+
+inline int GetWaterRegionX(TileIndex tile) { return TileX(tile) / WATER_REGION_EDGE_LENGTH; }
+inline int GetWaterRegionY(TileIndex tile) { return TileY(tile) / WATER_REGION_EDGE_LENGTH; }
+
+inline int GetWaterRegionMapSizeX() { return Map::SizeX() / WATER_REGION_EDGE_LENGTH; }
+inline int GetWaterRegionMapSizeY() { return Map::SizeY() / WATER_REGION_EDGE_LENGTH; }
+
+inline TWaterRegionIndex GetWaterRegionIndex(int region_x, int region_y) { return GetWaterRegionMapSizeX() * region_y + region_x; }
+inline TWaterRegionIndex GetWaterRegionIndex(TileIndex tile) { return GetWaterRegionIndex(GetWaterRegionX(tile), GetWaterRegionY(tile)); }
```
rubidium42: Maybe use `static inline` to denote that these are only used within this file?

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file water_regions.h Handles dividing the water in the map into square regions to assist pathfinding. */
```
rubidium42:
```suggestion
 /** @file water_regions.cpp Handles dividing the water in the map into square regions to assist pathfinding. */
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+/**
+ * Represents a square section of the map of a fixed size. Within this square individual unconnected patches of water are
+ * identified using a Connected Component Labeling (CCL) algorithm. Note that all information stored in this class applies
+ * only to tiles within the square section, there is no knowledge about the rest of the map. This makes it easy to invalidate
+ * and update a water region if any changes are made to it, such as construction or terraforming.
+ */
+class WaterRegion
+{
+private:
+	std::array<TWaterRegionTraversabilityBits, 4> edge_traversability_bits{};
```
rubidium42:
```suggestion
	std::array<TWaterRegionTraversabilityBits, DIAGDIR_END> edge_traversability_bits{};
```
We do not really like magic numbers, and since this array seems to be indexed by a `DiagDirection`, this seems a better fit.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+	void ForceUpdate()
+	{
+		this->has_cross_region_aqueducts = false;
+
+		this->tile_patch_labels.fill(INVALID_WATER_REGION_PATCH);
+
+		for (const TileIndex tile : this->tile_area) {
+			if (IsAqueductTile(tile)) {
+				const TileIndex other_aqueduct_end = GetOtherBridgeEnd(tile);
+				if (!tile_area.Contains(other_aqueduct_end)) this->has_cross_region_aqueducts = true;
```
rubidium42: You can also break out of the loop in the `if`s case.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+			tiles_to_check.clear();
+			tiles_to_check.push_back(start_tile);
+
+			bool increase_label = false;
+			while (!tiles_to_check.empty()) {
+				const TileIndex tile = tiles_to_check.back();
+				tiles_to_check.pop_back();
+
+				const TrackdirBits valid_dirs = TrackBitsToTrackdirBits(GetWaterTracks(tile));
+				if (!valid_dirs) continue;
```
rubidium42:
```suggestion
				if (valid_dirs == TRACKDIR_BIT_NONE) continue;
```
The `!` is generally only allowed for booleans.

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+std::vector<WaterRegion> _water_regions;
+
+TileIndex GetTileIndexFromLocalCoordinate(int region_x, int region_y, int local_x, int local_y)
+{
+	assert(local_x >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	assert(local_y >= 0 && local_y < WATER_REGION_EDGE_LENGTH);
+	return TileXY(WATER_REGION_EDGE_LENGTH * region_x + local_x, WATER_REGION_EDGE_LENGTH * region_y + local_y);
+}
+
+TileIndex GetEdgeTileCoordinate(int region_x, int region_y, DiagDirection side, int x_or_y) // TODO better name
```
rubidium42: Maybe something like `edge_coordinate`? Or `side_coordinate`?

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+		index++;
+	}
+}
+
+/**
+ * Initializes all water regions. All water tiles will be scanned and interconnected water patches within regions will be identified.
+ */
+void InitializeWaterRegions()
+{
+
```
rubidium42: Unneeded newline.

src/pathfinder/water_regions.h
```diff
@@ -0,0 +1,73 @@
...
+constexpr int WATER_REGION_EDGE_LENGTH = 16;
+constexpr int WATER_REGION_NUMBER_OF_TILES = WATER_REGION_EDGE_LENGTH * WATER_REGION_EDGE_LENGTH;
+
+/**
+ * Describes a single interconnected patch of water within a particular water region.
+ */
+struct WaterRegionPatchDesc
+{
+	int x; ///< Water region X coordinate
+	int y; ///< Water region Y coordinate
```
rubidium42: This and `WaterRegionDesc` both call these `Water region ? coordinate`. However, as far as I could figure out they are not the same. I think for `WaterRegionPatchDesc` they are essentially `Tile X/Y % WATER_REGION_EDGE_LENGTH`, and for `WaterRegionDesc` they are essentially `Tile X/Y / WATER_REGION_EDGE_LENGTH`. 
To avoid confusion, I would give them a different name, and maybe write in the comment of `WaterRegionPatchDesc` something along the line of `Coordinates are within region, i.e. x=2 is the third tile within the region along the X-axis`.
For `WaterRegionDesc` maybe something along the line of `Coordinates are of region, i.e. x=2 is the third region along the X-axis.`

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -126,92 +151,154 @@ class CYapfFollowShipT
...
+			}
 		}
 	}
 
-	/** return debug report character to identify the transportation type */
+	/** Restricts the search by creating corridor or water regions through which the ship is allowed to travel. */
+	inline void RestrictSearch(const std::vector<WaterRegionPatchDesc> &path)
+	{
+		m_water_region_corridor.clear();
+		for (const WaterRegionPatchDesc& path_entry : path) m_water_region_corridor.push_back(path_entry);
```
rubidium42:
```suggestion
		for (const WaterRegionPatchDesc &path_entry : path) m_water_region_corridor.push_back(path_entry);
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -126,92 +151,154 @@ class CYapfFollowShipT
...
+			Node *node = pf.GetBestNode();
+			if (attempt == 0 && !path_found) continue; // Try again with restricted search area.
+			if (!path_found || !node) return INVALID_TRACKDIR;
+
+			/* Return only the path within the current water region if an intermediate destination was returned. If not, cache the entire path
+			 * to the final destination tile. The low-level pathfinder might actually prefer a different docking tile in a nearby region. Without
+			 * caching the full path the ship can get stuck in a loop. */
+			const WaterRegionPatchDesc end_water_patch = GetWaterRegionPatchInfo(node->GetTile());
+			const WaterRegionPatchDesc start_water_patch = GetWaterRegionPatchInfo(tile);
+			std::deque<std::pair<TileIndex, Trackdir>> draw_stuff; // TODO KB remove
```
rubidium42: Remaining debug code?

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file yapf_ship.cpp Implementation of YAPF for water regions, which are used for finding intermediate ship destinations. */
```
rubidium42:
```suggestion
 /** @file yapf_ship_regions.cpp Implementation of YAPF for water regions, which are used for finding intermediate ship destinations. */
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+constexpr int NODES_PER_REGION = 4;
+constexpr int MAX_NUMBER_OF_NODES = 65536;
+
+/** Yapf Node Key that represents a single patch of interconnected water within a water region. */
+struct CYapfRegionPatchNodeKey {
+	WaterRegionPatchDesc m_water_region_patch;
+
+	static_assert(sizeof(TWaterRegionPatchLabel) == sizeof(byte)); // Important for the hash calculation.
+
+	inline void Set(const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
	inline void Set(const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+	typedef Tkey_ Key;
+	typedef CYapfRegionNodeT<Tkey_> Node;
+
+	Tkey_       m_key;
+	Node       *m_hash_next;
+	Node       *m_parent;
+	int         m_cost;
+	int         m_estimate;
+
+	/* TODO KB Remove these functions, they exist only to allow the debug drawing code in yapf_base.hpp to compile. */
```
rubidium42: More debug remnants?

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+public:
+	typedef typename Types::Tpf Tpf;              ///< The pathfinder class (derived from THIS class).
+	typedef typename Types::NodeList::Titem Node; ///< This will be our node type.
+	typedef typename Node::Key Key;               ///< Key to hash tables.
+
+protected:
+	Key m_dest;
+
+public:
+	void SetDestination(const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
	void SetDestination(const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+	typedef typename Types::NodeList::Titem Node;        ///< This will be our node type.
+	typedef typename Node::Key Key;                      ///< Key to hash tables.
+
+protected:
+	inline Tpf& Yapf() { return *static_cast<Tpf*>(this); }
+
+public:
+	inline void PfFollowNode(Node &old_node)
+	{
+		TVisitWaterRegionPatchCallBack visitFunc = [&](const WaterRegionPatchDesc& water_region_patch)
```
rubidium42:
```suggestion
		TVisitWaterRegionPatchCallBack visitFunc = [&](const WaterRegionPatchDesc &water_region_patch)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+{
+public:
+	typedef typename Types::Tpf Tpf;              ///< the pathfinder class (derived from THIS class).
+	typedef typename Types::TrackFollower TrackFollower;
+	typedef typename Types::NodeList::Titem Node; ///< this will be our node type.
+	typedef typename Node::Key Key;               ///< key to hash tables.
+
+protected:
+	/** To access inherited path finder. */
+	Tpf& Yapf() { return *static_cast<Tpf*>(this); }
```
rubidium42:
```suggestion
	Tpf &Yapf() { return *static_cast<Tpf*>(this); }
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+/** Cost Provider of YAPF for water regions. */
+template <class Types>
+class CYapfCostRegionT
+{
+public:
+	typedef typename Types::Tpf Tpf;              ///< the pathfinder class (derived from THIS class).
+	typedef typename Types::TrackFollower TrackFollower;
+	typedef typename Types::NodeList::Titem Node; ///< this will be our node type.
+	typedef typename Node::Key Key;               ///< key to hash tables.
```
rubidium42: Forgot some capitalisation here in the comments.

src/pathfinder/yapf/yapf_ship_regions.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ /** @file yapf_ship.cpp Implementation of YAPF for water regions, which are used for finding intermediate ship destinations. */
```
rubidium42:
```suggestion
 /** @file yapf_ship_regions.h Implementation of YAPF for water regions, which are used for finding intermediate ship destinations. */
```

src/pathfinder/water_regions.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+/**
+ * Calls the provided callback function for all water region patches
+ * accessible from one particular side of the starting patch.
+ * @param water_region_patch Water patch within the water region to start searching from
+ * @param side Side of the water region to look for neigboring patches of water
+ * @param callback The function that will be called for each neighbor that is found
+ */
+inline void VisitAdjacentWaterRegionPatchNeighbors(const WaterRegionPatchDesc& water_region_patch, DiagDirection side, TVisitWaterRegionPatchCallBack& func)
```
rubidium42:
```suggestion
inline void VisitAdjacentWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_patch, DiagDirection side, TVisitWaterRegionPatchCallBack &func)
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+class CYapfFollowRegionT
+{
+public:
+	typedef typename Types::Tpf Tpf;                     ///< The pathfinder class (derived from THIS class).
+	typedef typename Types::TrackFollower TrackFollower;
+	typedef typename Types::NodeList::Titem Node;        ///< This will be our node type.
+	typedef typename Node::Key Key;                      ///< Key to hash tables.
+
+protected:
+	inline Tpf& Yapf() { return *static_cast<Tpf*>(this); }
```
rubidium42:
```suggestion
	inline Tpf &Yapf() { return *static_cast<Tpf*>(this); }
```

src/saveload/water_regions_sl.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+struct WRGNChunkHandler : ChunkHandler {
+	WRGNChunkHandler() : ChunkHandler('WRGN', CH_TABLE) {}
+
+	void Save() const override
+	{
+		SlTableHeader(_water_region_desc);
+
+		int index = 0;
+		for (WaterRegionSaveLoadInfo& region : GetWaterRegionSaveLoadInfo()) {
```
rubidium42:
```suggestion
		for (WaterRegionSaveLoadInfo &region : GetWaterRegionSaveLoadInfo()) {
```

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -14,68 +14,92 @@
...
 
+		if (m_destStation != INVALID_STATION) return IsDockingTile(tile) && IsShipDestinationTile(tile, m_destStation);
+
 		return tile == m_destTile && ((m_destTrackdirs & TrackdirToTrackdirBits(trackdir)) != TRACKDIR_BIT_NONE);
 	}
 
 	/**
 	 * Called by YAPF to calculate cost estimate. Calculates distance to the destination
-	 *  adds it to the actual cost from origin and stores the sum to the Node::m_estimate
+	 *  adds it to the actual cost from origin and stores the sum to the Node::m_estimate.
```
rubidium42:
```suggestion
	 * adds it to the actual cost from origin and stores the sum to the Node::m_estimate.
```

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,316 @@
...
+	typedef typename Node::Key Key;               ///< key to hash tables.
+
+protected:
+	/** To access inherited path finder. */
+	Tpf& Yapf() { return *static_cast<Tpf*>(this); }
+
+public:
+	/** Called by YAPF to calculate the cost from the origin to the given node.
+	 *  Calculates only the cost of given node, adds it to the parent node cost
+	 *  and stores the result into Node::m_cost member. */
```
rubidium42:
```suggestion
	/**
	 * Called by YAPF to calculate the cost from the origin to the given node.
	 * Calculates only the cost of given node, adds it to the parent node cost
	 * and stores the result into Node::m_cost member. 
	 */
```

rubidium42 (commented):
While you're fixing these, also please do a rebase as there's a (merge) conflict.

src/pathfinder/water_regions.cpp
Kuhnovic: IMO the fact that it's in a cpp file kind of implies that already. But I'd rather go for consistency if that's something we already do for other functions / files.

src/pathfinder/water_regions.cpp
Kuhnovic: I'm not sure if I referred to the function name or to x_or_y. But I think they are both pretty clear actually. Although a bit odd, "x_or_y" at least somewhat captures the fact that it's coordinate changes depending on the provided DiagDirection.

src/pathfinder/water_regions.h
Kuhnovic: They are in fact the same, both are `TileX/Y / WATER_REGION_EDGE_LENGTH`... but I understand the confusion. I've changed the comments to better reflect their meaning based on your suggestion.

src/pathfinder/water_regions.cpp
rubidium42: I referred to the `x_or_y` as I thought that was what the TODO is about. If both are pretty clear, then I'd say: remove the TODO.

src/pathfinder/water_regions.cpp
rubidium42: `static inline` is more prevalent, though an argument could be made that only `inline` is better. It's definitely better when in header files, so I made a PR to remove `static` from there. I couldn't quite think of a strong enough reason to remove it from the headers, where we use `static` to imply the function isn't used from the outside.

rubidium42 (commented):
Did you notice the trailing white space error already?

rubidium42 (dismissed):

src/pathfinder/yapf/yapf_ship_regions.cpp
```diff
@@ -0,0 +1,314 @@
...
+	/** To access inherited path finder. */
+	Tpf &Yapf() { return *static_cast<Tpf*>(this); }
+
+public:
+	/**
+	 * Called by YAPF to calculate the cost from the origin to the given node.
+	 * Calculates only the cost of given node, adds it to the parent node cost
+	 * and stores the result into Node::m_cost member.
+	 */
+	inline bool PfCalcCost(Node &n, const TrackFollower */*tf*/)
```
TrueBrain: What happened here? GitHub shows it as red .. :D

Ah, pointer followed by a comment. Mind removing the comment? Tnx :)

Kuhnovic: Done!

TrueBrain (approved):

TrueBrain:
Tnx again for all your work on this. It looks awesome :D

Issue #8022, opened 2020-02-25
Ship automatic service causes them to be stuck in a loop


SamuXarick:
## Version of OpenTTD
~~OpenTTD 20200223-master-gd1b7eb2de1~~ - Resolved by #11768
OpenTTD 20240128-master-g90084dc74d
## Expected result
Ships either do service or not.
## Actual result
Ships are stuck in a loop due to automatic servicing distance limit triggering it. They can't decide whether to head to ship depot or the dock, which are opposite directions.
## Steps to reproduce
~~[Charningstone Transport, 1960-11-25.zip](https://github.com/OpenTTD/OpenTTD/files/4250827/Charningstone.Transport.1960-11-25.zip)~~ - Resolved by #11768

[Charningstone Transport, 1964-05-30.zip](https://github.com/OpenTTD/OpenTTD/files/14077702/Charningstone.Transport.1964-05-30.zip) - Not resolved

SamuXarick:
I have a fix for this which envolves using the pathfinder, but the idea has been rejected in the past, which was before ship path cache. Is it perhaps worth approaching it again?
https://github.com/SamuXarick/OpenTTD/tree/find-closest-reachable-ship-depot

SamuXarick:
For this specific savegame, it is fixed by #11768, but I'm not sure this is solved in other situations.

hpiirai:
Just a reminder that this bug is still present in **20251213-master-g1091fa6071**.

Try [Gribble & Co., 1877-06-21.sav.gz](https://github.com/user-attachments/files/24140752/Gribble.Co.1877-06-21.sav.gz) if you dare.

PR #10544, opened 2023-03-05
Fix #5713: Use pathfinder to find closest ship depot

### LINKED ISSUE #5713: Boats sent to unreachable depots
**_Leftie_ opened the ticket and wrote:**

> Hiya.
> 
> Boats sometimes attempt to go to places they are simply unable to reach. I first saw this on the TT-F forum, here http://www.tt-forums.net/viewtopic.php?f=29&t=68364&p=1093057#p1093055
> 
> I then tried to replicate it to see if it was still possible in the current version (1.3.2). Using YAPF, i managed to recreate this. Boats attempted to go to a depot that was nearer, but was completely inaccessible through any valid means, and totally ignored a boat yard further away, but in clear open waters.
> 
> My attempt at replication is in the below screenshot, using OpenTTD 1.3.2 and YAPF as the pathfinder in advanced settings.

**Reported version**: _1.3.2_
**Operating system**: _All_

---

SamuXarick:
## Motivation / Problem
https://github.com/OpenTTD/OpenTTD/issues/8022 - Ships are stuck in a loop due to automatic servicing distance limit triggering it. They can't decide whether to head to ship depot or the dock, which are opposite directions.

## Description
This patch provides the closest reachable ship depot, by utilizing the pathfinder, and removes current masters' method.

I created 3 commits. Commit 1 is the base core of the changes. Commit 2 and 3 are "optional", in the sense that they improve the experience and results, but I recognize they make the code larger or harder to review.

1. ### _Use pathfinder to find closest ship depot_
Yapf searches for a depot by using a breadth-first search approach on both the high-level and low-level pathfinders. The approach I took was to base it as much as I could on the existing `ChooseShipTrack`.

Some changes were made to this function, to adapt for the necessary means to deal with a `max_penalty` path cost and to return the tile of the depot found.

On the high-level part of the pathfinder however, a new method exclusively for searching depots was introduced, `FindShipDepotRegionPath`.

Unlike the original `FindWaterRegionPath`, the depot method has no known destination, so the current ship position is added as the origin, the path is built upside down instead, and the whole of it is reversed then returned to maintain the same behaviour expected for `ChooseShipTrack`, which has been adapted to use spans of the path returned where needed.

Intermediate destinations and automatic servicing required special care of my part. After a lot of experimentation, I decided it was best to simply return no depot found when in this situation even when sometimes there would be no issue with returning the depot. My reasoning for this decision is to prevent exactly what I'm supposed to be solving after all, avoid ships going back and forth undecided whether to service or not. Ships could find themselves outside of an intermediate destination on subsequent automatic service requests and have no ship depot found within the allowed `max_penalty` cost.

2. ### _Find nearest depot to entry edge in intermediate region_
This add-on improves the search of a ship depot when an intermediate region is set. It's not perfect by any means, but it's an attempt to guess which depot is closest to the edge in which the region is entered from. Should there be more than one ship depot in that region, this theoretically helps return the depot which is closest.

3. ### _Ships may reverse on find closest depot order_
This add-on improves the search of a ship depot by letting the pathfinder know whether the ship is allowed to reverse. Ships may be allowed to reverse when manually ordered to find the closest ship depot or executing an updated order to find the closest ship depot after departing from a station.

## Limitations
This was rejected in the past, when ship cache didn't exist and OPF and NPF were still options. But as of today, I think of it as complete and clean enough to warrant a PR again. I've been rebasing it over the years to keep it current.

## Checklist for review

### DIFF
--- a/src/aircraft.h
+++ b/src/aircraft.h
@@ -110,7 +110,7 @@ struct Aircraft final : public SpecializedVehicle<Aircraft, VEH_AIRCRAFT> {
 	uint Crash(bool flooded = false) override;
 	TileIndex GetOrderStationLocation(StationID station) override;
 	TileIndex GetCargoTile() const override { return this->First()->tile; }
-	ClosestDepot FindClosestDepot() override;
+	ClosestDepot FindClosestDepot(bool may_reverse = false) override;
 
 	/**
 	 * Check if the aircraft type is a normal flying device; eg
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -394,7 +394,7 @@ CommandCost CmdBuildAircraft(DoCommandFlags flags, TileIndex tile, const Engine 
 }
 
 
-ClosestDepot Aircraft::FindClosestDepot()
+ClosestDepot Aircraft::FindClosestDepot([[maybe_unused]] bool may_reverse)
 {
 	const Station *st = GetTargetAirportIfValid(this);
 	/* If the station is not a valid airport or if it has no hangars */
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -1894,10 +1894,11 @@ VehicleOrderID ProcessConditionalOrder(const Order *order, const Vehicle *v)
  * Update the vehicle's destination tile from an order.
  * @param order the order the vehicle currently has
  * @param v the vehicle to update
+ * @param may_reverse Whether the vehicle is allowed to reverse when executing the updated order.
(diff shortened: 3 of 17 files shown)


Kuhnovic:
I think it's a good idea. The shortest manhattan distance doesn't necessarily have to be the actual shortest distance.

I just want to point out that I'm working on this PR which will conflict with this: https://github.com/OpenTTD/OpenTTD/pull/10543 . With my region-based approach searching for a depot can become a lot more efficient.

PR #10548, opened 2023-03-06
Change: Groups cache vehicle lists

SamuXarick:
## Motivation / Problem
AIs creating lists of vehicles will slow down OpenTTD especially with high amount of vehicles in the game.

## Description
Create a `vehicle_list` member for `GroupStatistics`. Every group, including the `group_default` and the `group_all`, will keep track of `const Vehicle *`'s as they're added or removed from them.

## Limitations
This is a work in progress.

1. I'm looking into the rest of OpenTTD code for places where the use of this cache could be beneficial, but...
2. ... `AfterLoad` is the biggest hindrance to this approach. Some of its routines can at times call long established functions that are now being switched away from `VehicleListPool` iteration to the cached group vehicle list version. I am experimenting populating `GroupStatistics`, which is where these lists reside, as early as possible in the `AfterLoadGame` function.

The following have been changed so far:
- Iterate group vehicle lists for `GenerateVehicleSortList`.
- Iterate group vehicle lists for `ScriptVehicleList_Station`, `ScriptVehicleList_Depot`, `ScriptVehicleList_Group` and `ScriptVehicleList_DefaultGroup`.
- Iterate aircraft group list for `UpdateAirplanesOnNewStation`.
- Iterate group vehicle lists for `DisasterTick_Ufo` and `DisasterTick_Big_Ufo`. As a compromise, the result of this change differs from the original. The iteration order changed from lowest to highest vehicle index to lowest to highest company index plus the order at which each vehicle was added to each list regardless of their indexes.
- Iterate group vehicle lists for `UpdateCompanyRatingAndValue`.
- Iterate group vehicle lists for `SettingsDisableElrail`.
- Iterate group vehicle lists for `GetPreviewCompany` and `NewVehicleAvailable`. As a compromise, the result of this change differs from the original. Since the list only contains primary vehicles, free wagons aren't taken into consideration.
- Iterate group vehicle lists for `GroupStatistics::UpdateProfits`.
- Iterate group vehicle list for `PropagateChildLivery`.
- Iterate group vehicle list for `CmdAddSharedVehicleGroup`.
- Iterate group vehicle list for `CmdRemoveAllVehiclesGroup`.
- Iterate group vehicle lists for `WhoCanServiceIndustry`.
- Iterate group vehicle lists for `NetworkPopulateCompanyStats` and `NetworkAutoCleanCompanies`.
- Iterate group vehicle lists for `RemoveOrderFromAllVehicles`. Additionally make the function receive one parameter more with the owner of the destination, to help it perform faster.

## Checklist for review

### DIFF
--- a/src/elrail.cpp
+++ b/src/elrail.cpp
@@ -625,10 +625,13 @@ void SettingsDisableElrail(int32_t new_value)
 		}
 	}
 
-	/* Fix the total power and acceleration for trains */
-	for (Train *t : Train::Iterate()) {
-		/* power and acceleration is cached only for front engines */
-		if (t->IsFrontEngine()) {
+	for (const Company *c : Company::Iterate()) {
+		/* Fix the total power and acceleration for trains.
+		 * These values are only cached for front engines which
+		 * is what the collection enumerates. */
+		const VehicleList &vehicle_list = c->group_all[VEH_TRAIN].vehicle_list;
+		for (const Vehicle *v : vehicle_list) {
+			Train *t = Train::From(Vehicle::Get(v->index));
 			t->ConsistChanged(CCF_TRACK);
 		}
 	}
--- a/src/vehiclelist.cpp
+++ b/src/vehiclelist.cpp
@@ -116,8 +116,11 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 
 	switch (vli.type) {
 		case VL_STATION_LIST:
-			for (const Vehicle *v : Vehicle::Iterate()) {
-				if (v->type == vli.vtype && v->IsPrimaryVehicle()) {
+			for (const Company *c : Company::Iterate()) {
+				/* Only iterate over all companies in the case of neutral Oil Rig stations */
+				if (vli.company != c->index && vli.company != OWNER_NONE) continue;
+				const VehicleList &vehicle_list = c->group_all[vli.vtype].vehicle_list;
+				for (const Vehicle *v : vehicle_list) {
 					for (const Order *order : v->Orders()) {
 						if ((order->IsType(OT_GOTO_STATION) || order->IsType(OT_GOTO_WAYPOINT) || order->IsType(OT_IMPLICIT))
 								&& order->GetDestination() == vli.index) {
@@ -141,37 +144,48 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 		}
 
 		case VL_GROUP_LIST:
+			if (vli.index == DEFAULT_GROUP) {
+				const VehicleList &vehicle_list = Company::Get(vli.company)->group_default[vli.vtype].vehicle_list;
+				for (const Vehicle *v : vehicle_list) {
+					list->push_back(v);
+				}
+				break;
+			}
+
 			if (vli.index != ALL_GROUP) {
-				for (const Vehicle *v : Vehicle::Iterate()) {
-					if (v->type == vli.vtype && v->IsPrimaryVehicle() &&
-							v->owner == vli.company && GroupIsInGroup(v->group_id, vli.index)) {
-						list->push_back(v);
+				for (const Group *g : Group::Iterate()) {
+					if (g->owner != vli.company || g->vehicle_type != vli.vtype) continue;
+					if (GroupIsInGroup(g->index, vli.index)) {
+						const VehicleList &vehicle_list = g->statistics.vehicle_list;
+						for (const Vehicle *v : vehicle_list) {
+							list->push_back(v);
+						}
 					}
 				}
 				break;
 			}
 			FALLTHROUGH;
 
-		case VL_STANDARD:
-			for (const Vehicle *v : Vehicle::Iterate()) {
-				if (v->type == vli.vtype && v->owner == vli.company && v->IsPrimaryVehicle()) {
-					list->push_back(v);
-				}
+		case VL_STANDARD: {
+			const VehicleList &vehicle_list = Company::Get(vli.company)->group_all[vli.vtype].vehicle_list;
+			for (const Vehicle *v : vehicle_list) {
+				list->push_back(v);
 			}
 			break;
+		}
 
-		case VL_DEPOT_LIST:
-			for (const Vehicle *v : Vehicle::Iterate()) {
-				if (v->type == vli.vtype && v->IsPrimaryVehicle()) {
-					for (const Order *order : v->Orders()) {
-						if (order->IsType(OT_GOTO_DEPOT) && !(order->GetDepotActionType() & ODATFB_NEAREST_DEPOT) && order->GetDestination() == vli.index) {
-							list->push_back(v);
-							break;
-						}
+		case VL_DEPOT_LIST: {
+			const VehicleList &vehicle_list = Company::Get(vli.company)->group_all[vli.vtype].vehicle_list;
+			for (const Vehicle *v : vehicle_list) {
+				for (const Order *order : v->Orders()) {
+					if (order->IsType(OT_GOTO_DEPOT) && !(order->GetDepotActionType() & ODATFB_NEAREST_DEPOT) && order->GetDestination() == vli.index) {
+						list->push_back(v);
+						break;
 					}
 				}
 			}
 			break;
+		}
 
 		default: return false;
 	}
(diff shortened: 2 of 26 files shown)


PeterN:
You've stated a performance problem but not backed up your solution. Can you provide some metrics as to what benefit this achieves?

What about if there are no running AIs that care about vehicle groups? Or a client on a multiplayer server?

src/vehiclelist.cpp
```diff
@@ -115,9 +115,13 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 	list->clear();
 
 	switch (vli.type) {
-		case VL_STATION_LIST:
-			for (const Vehicle *v : Vehicle::Iterate()) {
-				if (v->type == vli.vtype && v->IsPrimaryVehicle()) {
+		case VL_STATION_LIST: {
```
glx22: `{` not needed

src/vehiclelist.cpp
```diff
@@ -128,6 +132,7 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 				}
 			}
 			break;
+		}
```
glx22: `}` not needed

src/vehiclelist.cpp
```diff
@@ -140,38 +145,49 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 			break;
 		}
 
-		case VL_GROUP_LIST:
-			if (vli.index != ALL_GROUP) {
-				for (const Vehicle *v : Vehicle::Iterate()) {
-					if (v->type == vli.vtype && v->IsPrimaryVehicle() &&
-							v->owner == vli.company && GroupIsInGroup(v->group_id, vli.index)) {
-						list->push_back(v);
+		case VL_GROUP_LIST: {
```
glx22: `{` not needed

src/vehiclelist.cpp
```diff
@@ -140,38 +145,49 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
...
+			if (vli.index == DEFAULT_GROUP) {
+				const GroupStatistics &stats = Company::Get(vli.company)->group_default[vli.vtype];
+				for (const auto vehicle_id : stats.vehicle_list) {
+					list->push_back(Vehicle::Get(vehicle_id));
+				}
+				break;
+
+			} else if (vli.index != ALL_GROUP) {
```
glx22: You can drop the `else` as there's a `break` in previous `if`

src/vehiclelist.cpp
```diff
@@ -140,38 +145,49 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
...
+					if (GroupIsInGroup(g->index, vli.index)) {
+						for (const auto vehicle_id : Group::Get(g->index)->statistics.vehicle_list) {
+							list->push_back(Vehicle::Get(vehicle_id));
+						}
 					}
 				}
 				break;
 			}
 			FALLTHROUGH;
+		}
```
glx22: `}` not needed

SamuXarick:
> You've stated a performance problem but not backed up your solution. Can you provide some metrics as to what benefit this achieves?
> 
> What about if there are no running AIs that care about vehicle groups? Or a client on a multiplayer server? Have you measured any negative effect of maintaining this info when it won't be required?

I measured 13.0 vs this PR using 14 AIs and GS on a savegame starting with:
Name | Value | Type
-- | -- | --
_vehicle_pool.items | 36615 | unsigned __int64
_depot_pool.items | 31854 | unsigned __int64
_station_pool.items | 64000 | unsigned __int64
_group_pool.items | 0 | unsigned __int64

Over the course of the game, each AI will be creating groups and moving at least a vehicle into them. They will still leave some vehicles in the default group at the end.
The AI Scripts will be constantly looping `AIVehicleList_DefaultGroup`, `AIVehicleList_Station`, `AIVehicleList_Depot`, and once the groups are created, `AIVehicleList_Group`.

(PS: just noticed VehicleList_Depot are always empty, vehicles don't have orders going to depot, my bad.)

The Game Script will be constantly looping `GSVehicleList_Station` and `GSVehicleList_Depot`.

Test AI used in savegame: [CoronaAI PR #10548.zip](https://github.com/OpenTTD/OpenTTD/files/10930959/CoronaAI.PR.10548.zip)
Test GS used in savegame: [CoronaGS PR #10548.zip](https://github.com/OpenTTD/OpenTTD/files/10930961/CoronaGS.PR.10548.zip)
savegame: [corona ai gs limits lzma9.zip](https://github.com/OpenTTD/OpenTTD/files/10930970/corona.ai.gs.limits.lzma9.zip)

SamuXarick:
Worst case scenario test: 64k groups, 0 vehicles. It's faster by iterating vehicles in this case 👎 :
image
[64k groups.zip](https://github.com/OpenTTD/OpenTTD/files/10934385/64k.groups.zip)

glx22:
Having many more groups than vehicles is probably not that frequent.

SamuXarick:
Another test, using my League Tables Game Script. Any of the "infrastructure efficiency" tests will run `GSVehicleList_Station`. The savegame is from an issue report, it's recommended to uninstall all AIs, and only run the League Tables GS for testing this. It is slow already without AIs. The league tables will only start computing if it finds that the window is open:
image
Game Script: [League Tables.zip](https://github.com/OpenTTD/OpenTTD/files/10934817/League.Tables.zip)
Savegame: https://github.com/OpenTTD/OpenTTD/files/10832374/go.to.nearest.road.vehicle.depot.zip

src/elrail.cpp
```diff
@@ -625,10 +625,12 @@ void SettingsDisableElrail(int32_t new_value)
...
-	/* Fix the total power and acceleration for trains */
-	for (Train *t : Train::Iterate()) {
-		/* power and acceleration is cached only for front engines */
-		if (t->IsFrontEngine()) {
+	for (const Company *c : Company::Iterate()) {
+		const GroupStatistics &stats = c->group_all[VEH_TRAIN];
+		/* Fix the total power and acceleration for trains */
+		for (const Vehicle *v : stats.vehicle_list) {
+			Train *t = Train::From(Vehicle::Get(v->index));
+			/* power and acceleration is cached only for front engines */
```
nielsmh: This comment is no longer relevant, since it related to the original following line, that tests for `IsFrontEngine()`. But since your collection now only contains front engines, it no longer relates to this. The more appropriate location might be to move it outside the loop and note that the power/acceleration values are only cached for front engines and that is what this collection enumerates.

src/vehiclelist.cpp
```diff
@@ -116,8 +116,11 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
 
 	switch (vli.type) {
 		case VL_STATION_LIST:
-			for (const Vehicle *v : Vehicle::Iterate()) {
-				if (v->type == vli.vtype && v->IsPrimaryVehicle()) {
+			for (const Company *c : Company::Iterate()) {
+				/* Only iterate over all companies in the case of neutral Oil Rig stations */
+				if (vli.vtype != VEH_AIRCRAFT && vli.vtype != VEH_SHIP && vli.company != c->index) continue;
```
nielsmh: I'm not sure if this is correct. Shouldn't it rather test for the station owner being the neutral company, and not care about the vehicle type in question at all?

I also somewhat doubt this is going to be a performance improvement in this specific case. On the other hand, it's "only" a GUI use and not performance critical for the simulation itself.

nielsmh (commented):
I believe the biggest gain for actual game performance would be splitting out the vehicle types for the core game loop. There already is (or has been) a PR for that but it has not been merged.

src/vehiclelist.cpp
SamuXarick: Fixed. Thanks.

There's still a positive gain. In the worst case, it iterates only one vehicle type per company.

PeterN:
I'm kind of tempted to close this because it's clear you are just making this up as you go, with no real goal or defined problem.

2TallTyler:
I agree. Your fork can be your own personal sandbox to try things, but please don't open a PR until you have a clearly-defined problem, a solution that you know works, and the evidence to show us why we should adopt your solution.

If you're looking for some clearly-defined problems to solve, you could try the "good first issue" tag in Issues.

PR #10558, opened 2023-03-08
Fix: Improve grfmessage for ShipVehicleChangeInfo.

J0anJosep:
Improve the grfmsg when trying to assign an invalid cargo type to a ship.


2TallTyler (approved):
Good catch, thanks.

PR #10570, opened 2023-03-13
Fix: inconsistent partial Z calculations for different slopes

### LINKED ISSUE #10514: [Bug]: Inconsistent (partial) Z
### Version of OpenTTD

13.0, though probably any ;)

### Expected result

When different slopes connect to each other, there should not be weird/inconsistent jumps in the height or apparant slope (of the slope).

See the attach spreadsheet with partial Zs of different slopes that do fit together.
proposed

### Actual result

Inconsistent mess, especially with steep slopes and slopes with two adjacent corners raised.

See the attached screenshot of a spreadsheet with the partial Zs of different slopes that should fit together. For example CD18 the height is 1, while the adjacent tiles at CC19 and CE17 are 0. This means that a vehicle going along CC19->CD18->CE17 jumps up by 1 for a short time. This can be seen anywhere around the perimeter of the steep slope tiles against the single corner raised tiles.
Going more to the right, with the two adjacent raised corner tiles there as discontinuities with the one raised corner tiles as well. Similarly going from DY19->DZ18->EA17 gives such a bump, and something similar happens at row 48 around column EO.
actual

### Steps to reproduce

No savegame needed.

Attached is the spreadsheet so you can take a deeper look at it. The original tab is the current behaviour, generated by dumping all the x/y sub locations within a tile for each of the slopes and putting that into spreadsheets. The proposed tab is one variant that seems consistent, but there still needs to be a discussion about what it needs to be exactly. At the moment I'm not quite sure yet, but wanted to make the issue to at least get a place to put images and spreadsheets to make communication about the issue easier.
[partial z comparison.ods](https://github.com/OpenTTD/OpenTTD/files/10833615/partial.z.comparison.ods)

rubidium42:
## Motivation / Problem

Fixes #10514; the partial Z calculations result in inconsistent Z locations when looking at adjacent slopes. Similarly a 3-corner slope + the appropriate 1 corner slope did not result in the steep slope.
It was initially triggered by a CodeQL alert about the function having too little documentation, and trying to figure out what the result of all the slopes were.
Furthermore a plane going across tunnel entrances at their top would jump up and down in height due to the partial Z for some locations being in the "ground vehicle" realm, and in the other locations in the normal "ground slope" realm.

## Description

Redo basically all the partial Z calculations, and update ground vehicle code where needed.
Next to that add a load of `static_assert`s tests to check whether the composition of slopes yield the expected results, and that the partial Z values of the slopes are as expected.
Furthermore a distinction is made between ground vehicles and other calls to `GetSlopePixelZ` to account for the different behaviour for things as shadows for tunnel/bridge entrances compared to the road vehicles.

[frosch's train](https://github.com/frosch123/CaterpillarSnakeSlug/releases/tag/v1) seems to work [correctly](https://github.com/OpenTTD/OpenTTD/files/10995358/pr10570.zip), just like all weird slope related cases I could think of with road vehicles turning around on slopes.

## Limitations

The flickering of aircraft shadows on tunnel/bridge entrances is not completely solved.

## Checklist for review


michicc:
I'm wondering if the test thingy should be its own compile target (like the regression tests). The tests itself look fine, but making it a real target might allow adding more tests that are not pure `constexpr` later on.

Everything else looks fine, even if I'm assuming the maths are good instead of re-tracing them myself.

rubidium42:
> I'm wondering if the test thingy should be its own compile target (like the regression tests). The tests itself look fine, but making it a real target might allow adding more tests that are not pure `constexpr` later on.
> 
> Everything else looks fine, even if I'm assuming the maths are good instead of re-tracing them myself.

I have considered adding a separate compile target, but did not add it because there seemed to be some issues with adding a proper system, e.g. #9038. I'm also wondering the usefulness with cross-compiles for targets you cannot run, but then the regression tests can't run there either. Also such a system would essentially double the build time, as it's a separate build step with a different set of defines... although just to exclude the main functions. It probably can be done more efficient, but that's definitely out of scope for this PR.

michicc (approved):

PR #10572, opened 2023-03-15
Fix #10568: "can savegame be loaded check" failed in dedicated server

### LINKED ISSUE #10568: [Bug]: Dedicated server: Loading a scenario, or a save with NewGRF
### Version of OpenTTD

13.0, Debian 11.6

### Expected result

The game shouldn't error out.

### Actual result

```openttd@openttd:~/openttd-13.0-linux-generic-amd64$ ./openttd -D -g ~/.local/share/openttd/save/gamestart.sav 
(...)
dbg: [sl] Nie można wczytać gry [Translation: can't load the game]
Zepsuta zapisana gra - The intro game must not use NewGRF [Translation: Saved game broken - (...)]
```
The game still loads with the content in question, it just print an error to the console and then ignores it. Haven't had any problems after running the DS for 6 hours with that "error".

### Steps to reproduce

1. Run the dedicated server, trying to load a save or a scenario with NewGRF content with `-g` argument (`./openttd -D -g ~/.local/share/openttd/save/gamestart.sav`)

rubidium42:
## Motivation / Problem

Fixes #10568.

If savegame has NewGRFs, it complains NewGRFs are not allowed in intro game and returns SL_REINIT.
If loading fails, it usually returns SL_REINIT which doesn't trigger check.
When game mode is changed, so the intro game check is not done... the game will just crash slightly later as the NewGRFs are not loaded at this time. In other words, with the current way NewGRFs are loaded the check always fails but it does not trigger the actual error path that stops the server.

## Description

Move the logic for handling these kinds errors to `SafeLoad`, where the NewGRFs have already been loaded.
Then in case a load error (anything other than `SL_OK`) occurs, and we are coming from the menu state, i.e. it is the first savegame that is loaded, and we are a dedicated server, then we show an error message and close the server.

In the other cases nothing happens, but the indentation has changed. Also, I see no reason why the intro savegame is inherently not multiplayer safe, so I changed the comments to more reflect reality. After all, after the error and without the special case for the dedicated server, the server just disconnects and would load the intro game/menu where you cannot join.

## Limitations

People might complain that the dedicated server doesn't keep running with a broken/corrupt savegame, instead of just creating a new game. Though it is "just" reintroducing functionality that got broken in the past.

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -969,37 +969,51 @@ bool SafeLoad(const std::string &filename, SaveLoadOperation fop, DetailedFileTy
 
 	_game_mode = newgm;
 
-	switch (lf == nullptr ? SaveOrLoad(filename, fop, dft, subdir) : LoadWithFilter(lf)) {
-		case SL_OK: return true;
+	SaveOrLoadResult result = (lf == nullptr) ? SaveOrLoad(filename, fop, dft, subdir) : LoadWithFilter(lf);
+	if (result == SL_OK) return true;
 
-		case SL_REINIT:
-			if (_network_dedicated) {
-				/*
-				 * We need to reinit a network map...
-				 * We can't simply load the intro game here as that game has many
-				 * special cases which make clients desync immediately. So we fall
-				 * back to just generating a new game with the current settings.
-				 */
-				Debug(net, 0, "Loading game failed, so a new (random) game will be started");
-				MakeNewGame(false, true);
-				return false;
-			}
-			if (_network_server) {
-				/* We can't load the intro game as server, so disconnect first. */
-				NetworkDisconnect();
-			}
-
-			switch (ogm) {
-				default:
-				case GM_MENU:   LoadIntroGame();      break;
-				case GM_EDITOR: MakeNewEditorWorld(); break;
-			}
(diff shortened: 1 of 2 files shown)


src/openttd.cpp
```diff
@@ -969,37 +969,51 @@ bool SafeLoad(const std::string &filename, SaveLoadOperation fop, DetailedFileTy
...
-				 */
-				Debug(net, 0, "Loading game failed, so a new (random) game will be started");
-				MakeNewGame(false, true);
-				return false;
-			}
-			if (_network_server) {
-				/* We can't load the intro game as server, so disconnect first. */
-				NetworkDisconnect();
-			}
+	SaveOrLoadResult result = lf == nullptr ? SaveOrLoad(filename, fop, dft, subdir) : LoadWithFilter(lf);
```
glx22:
```suggestion
	SaveOrLoadResult result = (lf == nullptr) ? SaveOrLoad(filename, fop, dft, subdir) : LoadWithFilter(lf);
```
I know it's not really needed, but it helps the reading (many `=` symbols here).

glx22 (approved):

PR #10573, opened 2023-03-16
Fix: [Script] Access to enum/consts defined outside of main.nut

glx22:
## Motivation / Problem
OpenTTD extended squirrel with `require()` which allows to use AI/GS libraries and also splitting AI/GS into multiple files.

`require()` is a runtime function, and this detail is important to note.

When squirrel compiles a file, it gathers all enum/const into a table global to the VM, then during bytecode generation if an identifier corresponds to an enum/consts it is directly replaced by the value. Else a variable is assumed and an opcode to resolve the identifier at runtime is emitted.
At runtime when an identifier needs to be resolved, it is checked for in local variables, then in global scope (the root table), and only there.

Now the issue with `require()` and enum/consts is the compiler can't know at compile time which enum/consts will be defined by the "required" file, so when it encounters them it assumes they are variables to be resolved at runtime, but enum/consts are not considered at all at runtime. The result is "the index 'blah' does not exist", and script crashes.

## Description
My solution to this issue is to set consts table as delegate of root table. That way if an enum/const from a "required" file is accessed by "requiring" file, it still won't be found in root table, but the search will fallback to the consts table.

## Checklist for review

### DIFF
--- a/src/script/squirrel.cpp
+++ b/src/script/squirrel.cpp
@@ -566,6 +566,10 @@ void Squirrel::Initialize()
 
 	sq_pushroottable(this->vm);
 	squirrel_register_global_std(this);
+
+	/* Set consts table as delegate of root table, so consts/enums defined via require() are accessible */
+	sq_pushconsttable(this->vm);
+	sq_setdelegate(this->vm, -2);
 }
 
 class SQFile {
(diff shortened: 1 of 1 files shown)


src/script/squirrel.cpp
```diff
@@ -764,6 +768,12 @@ void Squirrel::Uninitialize()
...
 	ScriptAllocatorScope alloc_scope(this);
 
+	/* Remove the delegation */
+	sq_pushroottable(this->vm);
+	sq_pushnull(this->vm);
+	sq_setdelegate(this->vm, -2);
+	sq_pop(this->vm, 1);
+
 	/* Clean up the stuff */
 	sq_pop(this->vm, 1);
```
glx22, line 778: I'm not sure this works as expected when uninitialisation is done while the script is running (like AI reset).
It will pop stack top, but the intention was to clear the stack I think.

Anyway doesn't really matter as sq_close() will clear the stack anyway.

rubidium42 (approved):

PR #10575, opened 2023-03-18
RFC: Change: Vehicle service uses reliability threshold, not time interval

2TallTyler:
## Motivation

My ongoing rework of NotDayLength (#10322) to offer both in-game dates and real-world time units means I have to rework service intervals.

## Problem

Currently, we offer two ways of setting service intervals in the vehicle information window and the settings menu (which sets default values):

- Service the vehicle every N days
- Service the vehicle if its reliability drops below N percent

This is in addition to vehicle orders, which offer a range of fine-grained ways to control vehicle service:
- Go to nearest depot order
- Go to a specific depot order
- Go to a specific depot order, but only if needed
- Conditional order: Reliability
- Conditional order: Maximum reliability
- Conditional order: Remaining lifetime
- Conditional order: Requires service

I cannot see a valid usecase for setting vehicle intervals in days. Either you've matched dates with the timetable and want the vehicle to service at a specific segment of the route (and should use orders instead) or you want the vehicle to service often enough to maintain good reliability (and should use the reliability percent mode instead). Or you want to roleplay maintenance schedules, maybe? (but that should be done with orders). I welcome explanation of other usecases I may have overlooked. 🙂 

You may ask: What's wrong with keeping it the way it is?

The dual-unit system currently employed is, in my opinion, a dumpster fire in both user interface and the code. Specifying the units as "N days/%" is confusing to the player. You don't know which is being used and they scale in opposite directions — a higher percentage has similar results to fewer days. If you change the "Service intervals are in percents" setting, all your chosen defaults get reset to hard-coded defaults which vary by type of vehicle.

Behind the scenes, in addition to these hardcoded defaults we have several custom clamping functions to ensure that these settings stay within proper bounds for whatever mode they're operating in. Fine now, but a pain to extend.

And guess who needs to extend them? NotDayLength would require a third unit option, real-world seconds. This needs new defaults, new clamping bounds, and a third unit in the user interface (I'd have to rewrite it to show only one unit at a time, because "N seconds/days/%" is absolutely unacceptable). I can make these changes, but would really prefer not to make a bad design even worse.

## Description

This PR removes the concept of setting time-based service intervals, leaving only the percent-based reliability threshold.

I realize that there are some corner cases where this might change how players' savegames operate. I think the safest way to handle this is to disable breakdowns if any vehicles have custom service intervals.

The PR is a draft because if we move forward with this route, I'd want to rename the remaining "service intervals" references (including possibly the vehicle property itself) to "reliability threshold" to better describe how it's used.

(I'm fully aware that I'm probably not removing the vehicle properties or the game setting properly, but that's a bridge to cross after "can we remove a feature." 😛 )

## Limitations

Workflow

## Checklist for review

### DIFF
--- a/src/saveload/saveload.h
+++ b/src/saveload/saveload.h
@@ -345,6 +345,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_MULTITRACK_LEVEL_CROSSINGS,         ///< 302  PR#9931 v13.0  Multi-track level crossings.
 	SLV_NEWGRF_ROAD_STOPS,                  ///< 303  PR#10144 NewGRF road stops.
 	SLV_LINKGRAPH_EDGES,                    ///< 304  PR#10314 Explicitly store link graph edges destination, PR#10471 int64 instead of uint64 league rating
+	SLV_SERVICE_INTERVAL_PERCENT_ONLY,      ///< 305  PR#10575 Don't use days for service intervals, only percents.
+
 
 	SL_MAX_VERSION,                         ///< Highest possible saveload version
 };
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -348,7 +348,6 @@ CommandCost CmdBuildAircraft(DoCommandFlag flags, TileIndex tile, const Engine *
 
 		v->vehicle_flags = 0;
 		if (e->flags & ENGINE_EXCLUSIVE_PREVIEW) SetBit(v->vehicle_flags, VF_BUILT_AS_PROTOTYPE);
-		v->SetServiceIntervalIsPercent(Company::Get(_current_company)->settings.vehicle.servint_ispercent);
 
 		v->InvalidateNewGRFCacheOfChain();
 
--- a/src/base_consist.cpp
+++ b/src/base_consist.cpp
@@ -37,8 +37,5 @@ void BaseConsist::CopyConsistPropertiesFrom(const BaseConsist *src)
 	if (HasBit(src->vehicle_flags, VF_TIMETABLE_STARTED)) SetBit(this->vehicle_flags, VF_TIMETABLE_STARTED);
 	if (HasBit(src->vehicle_flags, VF_AUTOFILL_TIMETABLE)) SetBit(this->vehicle_flags, VF_AUTOFILL_TIMETABLE);
 	if (HasBit(src->vehicle_flags, VF_AUTOFILL_PRES_WAIT_TIME)) SetBit(this->vehicle_flags, VF_AUTOFILL_PRES_WAIT_TIME);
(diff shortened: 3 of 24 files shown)


JGRennison:
> Service the vehicle if its reliability drops below N percent

This is not actually what the reliability in percent mode does. What it does and the resultant threshold is rather non-obvious, whereas time interval based servicing is simple to understand and mostly works as expected by normal users.

src/saveload/saveload.h
```diff
@@ -345,6 +345,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_MULTITRACK_LEVEL_CROSSINGS,         ///< 302  PR#9931 v13.0  Multi-track level crossings.
 	SLV_NEWGRF_ROAD_STOPS,                  ///< 303  PR#10144 NewGRF road stops.
 	SLV_LINKGRAPH_EDGES,                    ///< 304  PR#10314 Explicitly store link graph edges destination, PR#10471 int64 instead of uint64 league rating
+	SLV_SERVICE_INTERVAL_PERCENT_ONLY,      ///< 305  PR#10575 Don't use days for service intervals, only percents.
```
James103, line 348: I just noticed while reviewing this that savegame versions after 205 aren't grouped by 5's; there should be a second newline after line 342 / SLV_299 and another after line 347 / SLV_304.

2TallTyler: Thanks, I've opened a separate PR to fix that.

LC-Zorg:
Before you start changing anything further, could you first present your intentions (https://github.com/OpenTTD/OpenTTD/pull/10322) to the players with all the disadvantages of this solution and ask them if they want it at all? For me, what you're trying to implement is the total destruction of this game, and for me, it spells the end of it. A lot of players will feel the same way. If you think that what I write is _unfounded speculation_, do a fair presentation and see for yourself. What you are doing is a change too radical to be done behind the backs of the players.

> Specifying the units as "N days/%" is confusing to the player
>"N seconds/days/%" is absolutely unacceptable

This argument is completely flawed. Just because days/% is confusing (I agree with that) doesn't mean it can't be represented better, and therefore doesn't mean it should be removed.
Service intervals setting
(By changing the currency you change the start or end of many settings text, so I think you can do the same with percentages, days and possibly seconds or minutes)

You should also take into account that the rate of reliability decline can vary greatly (nonsensical function), which can cause unexpected behavior, i.e. too frequent servicing.
I'm skipping the aspect of whether the current form suits your planed intentions or not. As of today, your project is slightly but still dominated by negative ratings and I'm convinced that when it will be presented to a wider public, it will simply be crushed, especially by players who like the competitive element of this game, so it is not worth considering the legitimacy of additional changes until the major change with its defects was not accepted by the players.

2TallTyler:
In case it was lost in translation, the "RFC" title is "Request For Comments". So, thanks for your comment. 🙂

Given JGR's comment and my investigations into how the reliability percentage feature does indeed not work as I'd expect, it looks like I'll just have to bite the bullet and add real-time units.

PR #10579, opened 2023-03-25
Codechange: Simplify some CodeQL-flagged trivial switches

2TallTyler:
## Motivation / Problem

Today on Discord we discussed CodeQL warnings triggered in #10519 which I proposed to dismiss in #10537.

@frosch123 wrote:
> turning that into an "if" does not make the code better
> and whoever adds a new widget next, will have to readd the switch
> warnings are a suggestion, but depending on context you do not have to follow them
...
> codeql does the checks based on the knowledge it has. all/most widget methods get a "widget" id passed, because we do not do proper oo-gui stuff
> as a result all widget-handlers have a switch-case, because the cases are usually not related (and would be separate methods in any modern gui framework)
> code should reflect intention. it does not make sense to work around warnings, just because the code-checker does not know the intention of the code

@LordAro wrote:
> i agree, disable the trivial switch check rather than anything else
> possibly with a cursory review of them to fix any "genuine" issues

I took that cursory review. This PR aims to fix what I'd consider potentially "genuine" issues.

## Description

Simplify some switches which don't seem like they would need expanding in the future:

- Two switches which use a `NOT_REACHED()` instead of an `assert()`
- One switch in MenuClickSubsidies() which is documented as unused
- Two switches in OpenGL which just need an if/else
- One switch in an OnInvalidateData() which only has one case. Using an `if` for this is done elsewhere, such as in `SpriteAlignerWindow`.

I would hope that with these resolved, we could merge #10537, clearing the way for #10519 to be merged without warnings. 🙂 

## Limitations

Trivial switches for widgets and hotkeys are left alone so they can be easily added to later.

Trivial switches for vehicle types are left alone because it's a multi-nested can of worms that I don't want to open right now. 😛 

## Checklist for review

### DIFF
--- a/src/object_gui.cpp
+++ b/src/object_gui.cpp
@@ -565,20 +565,17 @@ public:
 	{
 		if (pt.x == -1) return;
 
-		switch (select_proc) {
-			default: NOT_REACHED();
-			case DDSP_BUILD_OBJECT:
-				if (!_settings_game.construction.freeform_edges) {
-					/* When end_tile is MP_VOID, the error tile will not be visible to the
-						* user. This happens when terraforming at the southern border. */
-					if (TileX(end_tile) == Map::MaxX()) end_tile += TileDiffXY(-1, 0);
-					if (TileY(end_tile) == Map::MaxY()) end_tile += TileDiffXY(0, -1);
-				}
-				const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
-				Command<CMD_BUILD_OBJECT_AREA>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
-					end_tile, start_tile, spec->Index(), _selected_object_view, (_ctrl_pressed ? true : false));
-				break;
+		assert(select_proc == DDSP_BUILD_OBJECT);
+
+		if (!_settings_game.construction.freeform_edges) {
+			/* When end_tile is MP_VOID, the error tile will not be visible to the
+			 * user. This happens when terraforming at the southern border. */
+			if (TileX(end_tile) == Map::MaxX()) end_tile += TileDiffXY(-1, 0);
+			if (TileY(end_tile) == Map::MaxY()) end_tile += TileDiffXY(0, -1);
 		}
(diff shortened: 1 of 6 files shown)


src/toolbar_gui.cpp
```diff
@@ -552,9 +552,7 @@ static CallBackFunction ToolbarSubsidiesClick(Window *w)
  */
 static CallBackFunction MenuClickSubsidies(int index)
 {
-	switch (index) {
-		case 0: ShowSubsidiesList(); break;
```
glx22, line 556: Looks like this one was not really working outside of single player mode (hardcoded to company 0 only)

2TallTyler: I believe this is dropdown index 0? (Subsidy button has only one dropdown)

src/object_gui.cpp
```diff
@@ -565,20 +565,17 @@ class BuildObjectWindow : public Window {
...
-				}
-				const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
-				Command<CMD_BUILD_OBJECT_AREA>::Post(STR_ERROR_CAN_T_BUILD_OBJECT, CcPlaySound_CONSTRUCTION_OTHER,
-					end_tile, start_tile, spec->Index(), _selected_object_view, (_ctrl_pressed ? true : false));
-				break;
+		assert(select_proc == DDSP_BUILD_OBJECT);
+
+		if (!_settings_game.construction.freeform_edges) {
+			/* When end_tile is MP_VOID, the error tile will not be visible to the
+				* user. This happens when terraforming at the southern border. */
```
rubidium42: The second line of the comment aligns weirdly. Though it was already broken in the old situation.

src/toolbar_gui.cpp
rubidium42: Yes, there is only one button in the dropdown, as it's a list of subsidies for everyone. It doesn't care about the compan when opening it.

rubidium42 (dismissed):

src/object_gui.cpp
2TallTyler: Might as well fix it.

rubidium42 (approved):

PR #10582, opened 2023-03-28
Fix #10574: Force town-buildable roads to be not hidden.

### LINKED ISSUE #10574: [Bug]: Map generation fails when no town buildable roads are available
### Version of OpenTTD

20230316-master-g0bf6d80cd7 (Win10 64)

### Expected result

Without failures of creating a new map

### Actual result

Message

### Steps to reproduce

1. Download NewGRF: or U&RaTT, or Rattroads
2. Starting game in 1900 year

PeterN:
## Motivation / Problem

Some NewGRFs set all town-buildable roads as hidden, which is not the intention of the hidden flag.

Since the hidden flag was enforced, these NewGRFs can no longer be used in the game. As per #10574, map generation fails, and if loading a savegame, towns will not be able to grow.

## Description

This is resolved by forcing town-buildable roads to be not hidden.

## Limitations

This is just a suggestion of how to work around this issue without reverting the changes which enforce the check.

An alternative is to test if there are no town-buildable roads and then force a non-hidden road to be town-buildable, but that would probably cause different problems.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -4516,6 +4516,7 @@ static ChangeInfoResult RoadTypeChangeInfo(uint id, int numinfo, int prop, ByteR
 
 			case 0x10: // Road Type flags
 				rti->flags = (RoadTypeFlags)buf->ReadByte();
+				if (HasBit(rti->flags, ROTF_TOWN_BUILD)) ClrBit(rti->flags, ROTF_HIDDEN);
 				break;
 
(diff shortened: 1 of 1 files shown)


PeterN:
Hmm, actually this doesn't wholly fix the issue with buggy NewGRFs. Just because all town-buildable roads are allowed again, doesn't mean a town is able to build a road.

2TallTyler:
What are we doing with this one?

PR #10585, opened 2023-03-30
Fix #10574 -- revert the change and don't list unavailable road types to game script.

### LINKED ISSUE #10574: [Bug]: Map generation fails when no town buildable roads are available

PeterN:
## Motivation / Problem

See #10574. 

## Description

Re-allow town-buildable to override hidden, but do check for availability for game script listing.

## Checklist for review

### DIFF
--- a/src/road.cpp
+++ b/src/road.cpp
@@ -112,7 +112,14 @@ bool HasRoadTypeAvail(const CompanyID company, RoadType roadtype)
 {
 	if (company == OWNER_DEITY || company == OWNER_TOWN || _game_mode == GM_EDITOR || _generating_world) {
 		const RoadTypeInfo *rti = GetRoadTypeInfo(roadtype);
-		return rti->label != 0 && (rti->flags & ROTFB_HIDDEN) == 0;
+		if (rti->label == 0) return false;
+
+		/*
+		 * Do not allow building hidden road types, except when a town may build it.
+		 * The GS under deity mode, as well as anybody in the editor builds roads that are
+		 * owned by towns. So if a town may build it, it should be buildable by them too.
+		 */
+		return (rti->flags & ROTFB_HIDDEN) == 0 || (rti->flags & ROTFB_TOWN_BUILD) != 0;
 	} else {
 		const Company *c = Company::GetIfValid(company);
 		if (c == nullptr) return false;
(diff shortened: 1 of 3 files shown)


src/script/api/script_railtypelist.cpp
```diff
@@ -18,6 +18,6 @@ ScriptRailTypeList::ScriptRailTypeList()
 {
 	EnforceDeityOrCompanyModeValid_Void();
 	for (RailType rt = RAILTYPE_BEGIN; rt != RAILTYPE_END; rt++) {
-		if (ScriptCompanyMode::IsDeity() || ::HasRailtypeAvail(ScriptObject::GetCompany(), rt)) this->AddItem(rt);
+		if (::HasRailtypeAvail(ScriptObject::GetCompany(), rt)) this->AddItem(rt);
```
rubidium42: This function has no special case for town/deity, so this for a GS (not in company scope/mode) will dereference a null/invalid pointer.

PeterN: Oops, forgot to check that.

PeterN:
Updated to remove anything to do with rail types.

rubidium42 (dismissed):

Andrew350:
FWIW, as a NewGRF developer affected by the initial change this would be my preferred solution compared to the other options so far, although I'm obviously a bit biased here since this means no changes necessary on our side.

My only question is if this solution has any negative effect on the GS side of things? From what @rubidium42 explained, the reason for the change in the first place was that a GS could have issues determining which roadtypes were actually viable to build based on the way we define some of them with both the HIDDEN and TOWN_BUILD flags.

If that was the main issue and this solution solves it for both sides in a good way (which it does from my perspective, at least) then this seems like the ideal fix to me.

Brickblock1:
I would also prefer this option, while I haven't made a roadset I still think that this option is more flexible and therefore better.

michicc:
Needs a rebase after the recent CI changes.

rubidium42 (approved):

PR #10593, opened 2023-04-02
Fix #10493, Fix #8424: Collision sensitivity lowered to correct incorrect crash circumstances.

### LINKED ISSUE #8424: Trains can crash through depots
## Version of OpenTTD
1.10.3

## Expected result
The train entering the depot should continue going into the depot.
## Actual result
The train crashes into train on the other side of the depot.
## Steps to reproduce
Make a straight line of track with depots at the end and two one-way signals facing either other in the middle.  Create two trains going to the other depot.  One train will wait at the signal while the other will go back and forth in the middle segment.  Stop the train that's waiting for a path and delete the track in front of it.  Place a depot on that square for the other train to go into.  The second train will crash into the first train through the back wall of the depot.
openttd_bug

### LINKED ISSUE #10493: [Bug]: Trains on different track types can collide in end-to-end stations
### Version of OpenTTD

13.0, Windows 10

### Expected result

Trains of different types (railway, monorail, maglev) arrive in the end-to-end stations without colliding (as neither will pass onto the 'other' track type), or the signalling system will prevent them from using the track simultaneously.

### Actual result

The trains are allowed to use the station simultaneously and collide.

### Steps to reproduce

To recreate from scratch:
1. Place a Railway, Monorail, or Maglev station
2. Using a different track type, place another station on the end of the first, as if you were extending it
3. Timetable trains to use both halves of the station, and ensure they will be stopped in the platform simultaneously
4. When the second of the two trains arrives, it will collide with the one already in the platform

Attached save has this scenario set up, with a Monorail - Maglev station. After unpausing, the monorail train will arrive second, colliding with the maglev.

[MonoMaglev.zip](https://github.com/OpenTTD/OpenTTD/files/10776589/MonoMaglev.zip)

Station layout in question - a 1x3 Monorail station end-on to a 1x3 Maglev station.
image

(Based on a report on Reddit: https://www.reddit.com/r/openttd/comments/115mc5d/in_rare_cases_using_a_path_signal_can_cause/)

TrevorShelton:
Collision sensitivity was lowered in a way that both problematic circumstances appear to be rectified while true collisions still seem to function fully. (An idea of making the trains "park in" a little less also comes to mind, but this patch seems sufficient to resolve the overall issue.)

### DIFF
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -3138,8 +3138,8 @@ static Vehicle *FindTrainCollideEnum(Vehicle *v, void *data)
 	if (hash & ~15) return nullptr;
 
 	/* Slower check using multiplication */
-	int min_diff = (Train::From(v)->gcache.cached_veh_length + 1) / 2 + (tcc->v->gcache.cached_veh_length + 1) / 2 - 1;
-	if (x_diff * x_diff + y_diff * y_diff > min_diff * min_diff) return nullptr;
+	int min_diff = (Train::From(v)->gcache.cached_veh_length + tcc->v->gcache.cached_veh_length) / 2 - 2;
+	if (x_diff * x_diff + y_diff * y_diff >= min_diff * min_diff) return nullptr;
 
 	/* Happens when there is a train under bridge next to bridge head */
 	if (abs(v->z_pos - tcc->v->z_pos) > 5) return nullptr;


PeterN:
This change allows shorter vehicles to overlap without crashing considerably, and certain combinations of lengths (but not the smallest) will never crash even if exactly on the same position, as the minimum distance goes negative (past zero).

TrevorShelton:
I have put the viable alternative in a new pull request.

PR #9002, opened 2021-04-10
Change cargo age to have more effect on cargo payment.

michicc:
Or: Make transfer networks suck.
Or: How to break cargo ageing and cargodist in one go.

## Motivation / Problem

There's lot of talk about cargo ageing, profitability and strange cargo curves. We should try our hardest to make sure routes fail to be profitable.

Cargo profit depends on distance and time taken until delivery. There's a hard limit for the time taken, which means each cargo has a point from which on time (and thus vehicle speed) don't matter anymore. Further away means more profit then, not matter if horse or Concorde. Cargo also doesn't care about being stored for hundreds of years in transfer stations before delivery.

## Description

Increase age range of cargo packets. Age in-transit cargo at stations. Add a fourth time range to profit calculations that asymptotically approaches zero.

Cargo that is in-transit (i.e. has an age > 0) will be aged in stations. Default ticks before ageing is 740, except pax/mail which are aged every 370 ticks. A NewGRF cargo prop is provided to modify the rate.

On profit calculation, if the time factor was previously clamped to 31, it will now continue with a scaled `1/(x+1)` function asymptotically to zero with 4 bits of fixed point precision.

## Limitations

YOLO

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10596, opened 2023-04-04
Change: Increase max cargo age and let min cargo payment approach zero.

ldpl:
There was one good idea in #9002 so let's merge it! ;)

## Motivation / Problem

There is a minimum time factor for payment rates. That may have been fine on small maps in the original game but nowadays it causes cargo payments go to infinity with longer route and encourages players to build corner-to corner on 4k maps instead of staying withing the optimal payment range. Also makes cargo aging newgrf property much less useful as it effectively just scales the payment chart.

image

## Description

This PR smoothly scales time factor to 1/16 after it reaches the previous min time factor of 31. Payments with time factor above 31 stay exactly the same (for vanilla cargoes that's about a year en route). Compared to 5715227a5e8dd14e71fa964882307756a1cf2c72 this handles the case when time factor hits 31 before the end of days2 range and avoids rounding issue with odd days2 values.
Screenshot from 2023-04-04 17-34-56

CityMania profit calculator (https://citymania.org/tools/profit) is also updated with on option to use this PR.

## Limitations

- Min of 1/16 is not zero so after a certain point it still goes to infinity linearly but it's small enough to not be noticeable even on 4k maps.
- NewGRF api wasn't changed so profit callback is still limited to 255 days.

## Checklist for review

### DIFF
--- a/src/cargopacket.cpp
+++ b/src/cargopacket.cpp
@@ -383,7 +383,7 @@ void VehicleCargoList::AgeCargo()
 	for (ConstIterator it(this->packets.begin()); it != this->packets.end(); it++) {
 		CargoPacket *cp = *it;
 		/* If we're at the maximum, then we can't increase no more. */
-		if (cp->days_in_transit == 0xFF) continue;
+		if (cp->days_in_transit == UINT16_MAX) continue;
 
 		cp->days_in_transit++;
 		this->cargo_days_in_transit += cp->count;
--- a/src/cargopacket.h
+++ b/src/cargopacket.h
@@ -44,7 +44,7 @@ struct CargoPacket : CargoPacketPool::PoolItem<&_cargopacket_pool> {
 private:
 	Money feeder_share;     ///< Value of feeder pickup to be paid for on delivery of cargo.
 	uint16 count;           ///< The amount of cargo in this packet.
-	byte days_in_transit;   ///< Amount of days this packet has been in transit.
+	uint16 days_in_transit; ///< Amount of days this packet has been in transit.
 	SourceType source_type; ///< Type of \c source_id.
 	SourceID source_id;     ///< Index of source, INVALID_SOURCE if unknown/invalid.
 	StationID source;       ///< The station where the cargo came from first.
@@ -126,10 +126,10 @@ public:
 	/**
 	 * Gets the number of days this cargo has been in transit.
 	 * This number isn't really in days, but in 2.5 days (CARGO_AGING_TICKS = 185 ticks) and
-	 * it is capped at 255.
+	 * it is capped at UINT16_MAX.
 	 * @return Length this cargo has been in transit.
 	 */
-	inline byte DaysInTransit() const
+	inline uint16 DaysInTransit() const
 	{
 		return this->days_in_transit;
 	}
@@ -221,8 +221,8 @@ public:
 	};
 
 protected:
-	uint count;                 ///< Cache for the number of cargo entities.
-	uint cargo_days_in_transit; ///< Cache for the sum of number of days in transit of each entity; comparable to man-hours.
+	uint count;                   ///< Cache for the number of cargo entities.
+	uint64 cargo_days_in_transit; ///< Cache for the sum of number of days in transit of each entity; comparable to man-hours.
 
 	Tcont packets;              ///< The cargo packets in this list.
 
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -975,7 +975,7 @@ Money GetPrice(Price index, uint cost_factor, const GRFFile *grf_file, int shift
 	return cost;
 }
 
-Money GetTransportedGoodsIncome(uint num_pieces, uint dist, byte transit_days, CargoID cargo_type)
+Money GetTransportedGoodsIncome(uint num_pieces, uint dist, uint16 transit_days, CargoID cargo_type)
 {
 	const CargoSpec *cs = CargoSpec::Get(cargo_type);
 	if (!cs->IsValid()) {
@@ -985,7 +985,7 @@ Money GetTransportedGoodsIncome(uint num_pieces, uint dist, byte transit_days, C
 
 	/* Use callback to calculate cargo profit, if available */
 	if (HasBit(cs->callback_mask, CBM_CARGO_PROFIT_CALC)) {
-		uint32 var18 = std::min(dist, 0xFFFFu) | (std::min(num_pieces, 0xFFu) << 16) | (transit_days << 24);
+		uint32 var18 = std::min(dist, 0xFFFFu) | (std::min(num_pieces, 0xFFu) << 16) | (std::min<uint16>(transit_days, 0xFFu) << 24);
 		uint16 callback = GetCargoCallback(CBID_CARGO_PROFIT_CALC, 0, var18, cs);
 		if (callback != CALLBACK_FAILED) {
 			int result = GB(callback, 0, 14);
@@ -1002,25 +1002,35 @@ Money GetTransportedGoodsIncome(uint num_pieces, uint dist, byte transit_days, C
 
 	static const int MIN_TIME_FACTOR = 31;
 	static const int MAX_TIME_FACTOR = 255;
+	static const int TIME_FACTOR_FRAC_BITS = 4;
+	static const int TIME_FACTOR_FRAC = 1 << TIME_FACTOR_FRAC_BITS;
 
 	const int days1 = cs->transit_days[0];
 	const int days2 = cs->transit_days[1];
 	const int days_over_days1 = std::max(   transit_days - days1, 0);
 	const int days_over_days2 = std::max(days_over_days1 - days2, 0);
+	int days_over_max = MIN_TIME_FACTOR - MAX_TIME_FACTOR;
+	if (days2 > -days_over_max) days_over_max += transit_days - days1;
+	else days_over_max += 2 * (transit_days - days1) - days2;
 
 	/*
 	 * The time factor is calculated based on the time it took
 	 * (transit_days) compared two cargo-depending values. The
-	 * range is divided into three parts:
+	 * range is divided into four parts:
 	 *
 	 *  - constant for fast transits
 	 *  - linear decreasing with time with a slope of -1 for medium transports
 	 *  - linear decreasing with time with a slope of -2 for slow transports
+	 *  - after hitting MIN_TIME_FACTOR, the time factor will be asymptotically decreased to a limit of 1 with a scaled 1/(x+1) function.
 	 *
 	 */
-	const int time_factor = std::max(MAX_TIME_FACTOR - days_over_days1 - days_over_days2, MIN_TIME_FACTOR);
-
-	return BigMulS(dist * time_factor * num_pieces, cs->current_payment, 21);
+	if (days_over_max > 0) {
+		const int time_factor = std::max(2 * MIN_TIME_FACTOR * TIME_FACTOR_FRAC * TIME_FACTOR_FRAC / (days_over_max + 2 * TIME_FACTOR_FRAC), 1); // MIN_TIME_FACTOR / (x/(2 * TIME_FACTOR_FRAC) + 1) + 1, expressed as fixed point with TIME_FACTOR_FRAC_BITS.
+		return BigMulS(dist * time_factor * num_pieces, cs->current_payment, 21 + TIME_FACTOR_FRAC_BITS);
+	} else {
+		const int time_factor = std::max(MAX_TIME_FACTOR - days_over_days1 - days_over_days2, MIN_TIME_FACTOR);
+		return BigMulS(dist * time_factor * num_pieces, cs->current_payment, 21);
+	}
 }
 
 /** The industries we've currently brought cargo to. */
@@ -1094,7 +1104,7 @@ static uint DeliverGoodsToIndustry(const Station *st, CargoID cargo_type, uint n
  * @return Revenue for delivering cargo
  * @note The cargo is just added to the stockpile of the industry. It is due to the caller to trigger the industry's production machinery
  */
-static Money DeliverGoods(int num_pieces, CargoID cargo_type, StationID dest, TileIndex source_tile, byte days_in_transit, Company *company, SourceType src_type, SourceID src)
+static Money DeliverGoods(int num_pieces, CargoID cargo_type, StationID dest, TileIndex source_tile, uint16 days_in_transit, Company *company, SourceType src_type, SourceID src)
 {
 	assert(num_pieces > 0);
 
--- a/src/economy_func.h
+++ b/src/economy_func.h
@@ -28,7 +28,7 @@ extern Prices _price;
 int UpdateCompanyRatingAndValue(Company *c, bool update);
 void StartupIndustryDailyChanges(bool init_counter);
 
-Money GetTransportedGoodsIncome(uint num_pieces, uint dist, byte transit_days, CargoID cargo_type);
+Money GetTransportedGoodsIncome(uint num_pieces, uint dist, uint16 transit_days, CargoID cargo_type);
 uint MoveGoodsToStation(CargoID type, uint amount, SourceType source_type, SourceID source_id, const StationList *all_stations, Owner exclusivity = INVALID_OWNER);
 
 void PrepareUnload(Vehicle *front_v);
--- a/src/saveload/cargopacket_sl.cpp
+++ b/src/saveload/cargopacket_sl.cpp
@@ -88,14 +88,15 @@
 SaveLoadTable GetCargoPacketDesc()
 {
 	static const SaveLoad _cargopacket_desc[] = {
-		     SLE_VAR(CargoPacket, source,          SLE_UINT16),
-		     SLE_VAR(CargoPacket, source_xy,       SLE_UINT32),
-		     SLE_VAR(CargoPacket, loaded_at_xy,    SLE_UINT32),
-		     SLE_VAR(CargoPacket, count,           SLE_UINT16),
-		     SLE_VAR(CargoPacket, days_in_transit, SLE_UINT8),
-		     SLE_VAR(CargoPacket, feeder_share,    SLE_INT64),
-		 SLE_CONDVAR(CargoPacket, source_type,     SLE_UINT8,  SLV_125, SL_MAX_VERSION),
-		 SLE_CONDVAR(CargoPacket, source_id,       SLE_UINT16, SLV_125, SL_MAX_VERSION),
+		SLE_VAR(CargoPacket, source,          SLE_UINT16),
+		SLE_VAR(CargoPacket, source_xy,       SLE_UINT32),
+		SLE_VAR(CargoPacket, loaded_at_xy,    SLE_UINT32),
+		SLE_VAR(CargoPacket, count,           SLE_UINT16),
+		SLE_CONDVAR(CargoPacket, days_in_transit, SLE_FILE_U8 | SLE_VAR_U16, SL_MIN_VERSION, SLV_MORE_CARGO_AGE),
+		SLE_CONDVAR(CargoPacket, days_in_transit, SLE_UINT16, SLV_MORE_CARGO_AGE, SL_MAX_VERSION),
+		SLE_VAR(CargoPacket, feeder_share,    SLE_INT64),
+		SLE_CONDVAR(CargoPacket, source_type,     SLE_UINT8,  SLV_125, SL_MAX_VERSION),
+		SLE_CONDVAR(CargoPacket, source_id,       SLE_UINT16, SLV_125, SL_MAX_VERSION),
 	};
 	return _cargopacket_desc;
 }
--- a/src/saveload/saveload.h
+++ b/src/saveload/saveload.h
@@ -349,6 +349,7 @@ enum SaveLoadVersion : uint16 {
 
 	SLV_VELOCITY_NAUTICAL,                  ///< 305  PR#10594 Separation of land and nautical velocity (knots!)
 	SLV_CONSISTENT_PARTIAL_Z,               ///< 306  PR#10570 Conversion from an inconsistent partial Z calculation for slopes, to one that is (more) consistent.
+	SLV_MORE_CARGO_AGE,                     ///< 307  PR#10596 Track cargo age for a longer period.
 
 	SL_MAX_VERSION,                         ///< Highest possible saveload version
 };
--- a/src/script/api/script_cargo.cpp
+++ b/src/script/api/script_cargo.cpp
@@ -76,7 +76,7 @@
 
 	distance = Clamp<SQInteger>(distance, 0, UINT32_MAX);
 
-	return ::GetTransportedGoodsIncome(1, distance, Clamp(days_in_transit * 2 / 5, 0, 255), cargo_type);
+	return ::GetTransportedGoodsIncome(1, distance, Clamp(days_in_transit * 2 / 5, 0, UINT16_MAX), cargo_type);
 }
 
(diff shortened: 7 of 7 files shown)


LC-Zorg:
Some basic questions:
1. Why do you want to make it pointless to build extensive transport networks on large or long maps?
2. Such a change will make many existing player saves with large networks on large maps, that they have spent tens of hours creating, unplayable as companies will only generate losses. Have you considered it?
3. Is building small, compact companies occupying only a patch of large maps have to be the only style allowed by the economy in this game?

LordAro:
Less of the accusatory tone please

ldpl:
@LC-Zorg Do you realize just how long of a distance we're talking here? It takes thousands of tiles to even begin noticing the difference. For some train consists like end-game lev4 (TL12) it's nearly impossible see the change at all even if you send it corner to corner on 4k map. Also, as it would be traveling for over a year so other elements of the game will start to fall apart as well. E.g. you won't get accurate yearly profit and so accurate score.
 Besides, it's not like all profit suddenly goes to 0, even long route will still be earning enough to cover running costs. And any network that spans that long should be earning more than enough money to keep everything afloat anyway. So I don't think any real games would be affected much unless it's something built specifically to exploit the payment formula, like sending grossly overweight Kirbys across 4k map.

2TallTyler:
I support this change. The CityMania profit calculator is a really helpful tool to show just how large a distance we're talking about before this has an effect, so thanks for that.

PR needs a rebase before I can add my approval. 😃

2TallTyler (approved):

LC-Zorg:
But here you don't need thousands of tiles to notice the difference. :/ In the presented rate decrease graph, everyone can see a very large difference already at a distance of only 70 days. If the transport takes 300 instead of 230 days, the drop in the rate will be about 70% compared to what it is now. It's a lot. Very lot. In the case of 400 days, this is a decrease in the rate by about 90%. It's a knockout. After this change, how will the game look like when someone creates long-distance connections A-B-C-...-Z with several dozen stations? The stopover at these stations can last several hundred days. The A-Z journey can last even over 1000 days. Big maps? Not at all. What if someone uses one of the train sets? There you can often find locomotives and wagons not exceeding 120 km/h. In the real world, many railway lines, even long-distance ones, do not even reach this speed on a large part of the route. In the case of freight wagons, in the most of the sets there are no faster options at all. Whereas in case someone, not looking at the industry at all, wanted to create in this game, say, the Trans-Siberian line on the map of Russia? Or on the map of Europe Orient Express lines? Or on the map of America the Transcontinental Railroad? What if, on a large map, he wanted to create a network connecting all cities? What if he wanted to build such a network in the 19th century, when the railway did not even reach 100 km/h? After all, this change in this form will cause a very large drop in income and de facto destruction of such games. Both old and new ones. I haven't done any tests because I don't have such possibility, but I'm quite sure it will be like I wrote.
**However, I have three saves from which you can check what impact this change will have.**
[Increase max cargo age TEST.zip](https://github.com/OpenTTD/OpenTTD/files/11186378/Increase.max.cargo.age.TEST.zip)

The first save is a large railway network connecting all 578 cities on a large 2048x2048 map of Europe. Save is not mine. I found it quite a while ago. Probably from the era before the cargodist was added.
I enabled cargodist and added the Base Cost Mod to lower the costs raised by inflation.
The company's profit is an average of £126M with a profitability of 85%
Kingdom and Co , mini

The second save is a medium-sized 1024x1024 map and a simulation of the first half of the 20th century. Trains run at a speed of 120 km/h. There are three companies here representing different styles of construction and play. Green builds a direct connection diagonally across the map, where the journey takes almost 300 days. The yellow company also has one line, but here it connects 18 cities. The last red company creates a complex network of connections, covering half of the map and connecting 85 cities. 
Companies profits are: £1.4M (70%), £4.9M (78%), £7.5M (44%)
Long Dist Cargo Payment TEST, mini

The third save is the same as the second, except that the red company uses trams instead of spreading the station range, and the distance effect on distributions has been reduced from 100 to 60%.

**Please check it and show here what will be the profits of the companies after this change.**

In general, I think that extending the period of rate reduction can be a very good change, because it will make replacing vehicles with better or faster ones on long journeys always make sense, but the changes should not be down, but rather up. Instead of lowering the final, lowest rate, it would be better to raise the one in the middle. A possible slight reduction in the lowest rate would not be bad either, but slight. No one will complain about such a change. However, with the change in the form you propose, you will destroy the sense of building real routes and networks not only on large maps, but even on medium-sized ones.

LC-Zorg:
Here is my comment (final part) on the originally proposed change.
https://github.com/OpenTTD/OpenTTD/pull/9002#issuecomment-974516491
It is still actual, as well as the proposed modification

ldpl:
@LC-Zorg 
Where did you find 70 days? On the chart in the PR time factor starts to diverge after about 230 transit "days" that are equivalent of 575 in-game days, that's over a year and a half while moving. And cargo doesn't age on stations so that stopover is irrelevant. And your time factor chart makes no sense, when did you suggest to start smoothing? After days1+days2 and ignore the whole third stage of price drop? Some cargoes don't even have the third stage. And it has to drop as 1/x or faster otherwise it won't fix the payment exploit.

Also, here, operating profit charts for your games, with this PR and without, I'll let you guess which one is where :p
Screenshot from 2023-04-10 03-56-40
Screenshot from 2023-04-10 03-58-02
Screenshot from 2023-04-10 04-00-58

LC-Zorg:
Ok, so now I did the tests myself...

The same map, I changed the details, because indeed that graph and these "230" led me astray.
- changed the set of trains - cars for red and yellow are much more capacious and although they have a small negative multiplier for cargo aging, in 13.1 the effect is much higher income (2-3 times more)
- changed the route of green - now it travels a much longer, natural-looking route through the land of mountains and lakes
- added a company (blue) that creates a network of connections using slow sailing ships
Long Dist Cargo Payment TEST v2
[LDCPT v2.zip](https://github.com/OpenTTD/OpenTTD/files/11216854/LDCPT.v2.zip)
In all cases, decreases in income are noticeable. In the case of red and yellow, they are acceptable and the change is not a problem in my opinion. In the case of green and blue, the drop in income is very large and devastating.

Do you think such building and playing styles should be excluded from this game?
This is still a medium sized map. What about the bigger ones?

One more example...
Slow vehicles in 19th century
Do you think that the current slow vehicle add-ons should only be loss making or that it was impossible to build longer routes with them?

LC-Zorg:
> Decisions like this require collecting feedback from large amount of players, otherwise you'll just end up limiting the game to your personal favorite playstyle.

@ldpl Do you recognize who wrote it? ;) Think about it a bit.

ldpl:
@LC-Zorg Yes, and I think there was enough research done in this case. And it only proves that point that it took you 4 attemts at hand-crafting the affected save and still it does exactly what this change supposed to prevent - ridiculous payouts for slow vehicles sent way too far. So no surprises here. But why is that a problem? You're clearly not playing for money there, so just cheat in some or make a moneymaker. That's a very normal practice in OpenTTD that pretty much every playstyle relies on. Citybuilder, coop, cargodist - you just make a moneymaker first and do whatever you want later. And even if you absolutely want to set up a game balance in a way that revards moving cargo end to end there is still a way by using cargo aging newgrf prooperty that actually works now.

LC-Zorg:
> ridiculous payouts for slow vehicles sent way too far

There are two basic problems in this sentence. First of all, you still consider low payments to be radically high, and secondly, you want to decide at what distances the player should build his transport networks.
What this change does is make it impossible to build long routes without special add-ons, which means this change is disruptive for such games. Both those that players already have and now will not be able to continue them (general rule prohibiting changing the list of add-ons in existing games) and new ones, because not all players know about all possible add-ons. Just because you and I know how to solve a problem doesn't mean every player will know and you're ruining their game. And that's the problem. YOU think that players shouldn't build like that, because you think so, because it's not your style of play, you're not interested in it and you want to destroy it. You're doing exactly what you don't like. You're telling others how to play, including telling me that I should use cheat mode. Cheating at online games? Do you like such victories? Why do you recommend it to others? And above all, don't write that this is my style of play, because it is not. Personally, out of over 1000 saves, I probably don't have a single one where this change would have a clearly negative effect. What I don't like is that this change limits the possibility of such a game - on large maps, with slow vehicles. I don't know why TrueBrain is going in this direction then, since playing on large maps will be pointless. But that's not a problem for you, because that's not the style of play you're interested in, so you've allowed yourself to destroy it...
By the way, the pathfinder settings that you care about so much might not exist for me, because I never changed anything there anyway, despite being a rather committed gamer. I suppose there are far fewer players who used that settings than those who built on big maps. I also believe that the existence of these settings can be harmful, because a player unaware of their importance can change something in them, breaking his games, and then he will not know how to undo them, so I see the point in removing them. I also don't like the fact that on each server the game mechanics can be different in this view. I would much rather have one fine-tuned setting than many different ones where the player will encounter strange, unexpected behavior every time. But you know what, I still understand the point and the need to keep these settings. Meanwhile, you're crying over something that's very niche because it's your world and someone wants to destroy it, but on the other hand you don't even want to accept that what you're doing can also destroy someone else's world. Not mine. Someone else. I am writing because I do not like this approach and such changes that, instead of expanding, only narrow the possibilities. I generally don't like this approach. I like almost all of your changes because they are positive, they actually add or improve something, but not this one. Not in a form that, apart from improving something, also spoils and limits.
Yes, a bit unkind, but sincere

ldpl:
You need to make up your mind. Are you talking about money goal game? Yes, this change stops the ridiculous situation where optimal distance on 512 map is, for examplle, 1k, on 1k map is 1k, on 2k map is 1k but on 4k map it's suddenly 8k because CS in 1995 didn't make his profit formula with 4k maps in mind. Whoevers world is ruined by having to actually pay attention to optimal profit peaks and not just going corner to corner could've shown up and during two years this change was being discussed on the github. Though I've been playing and making goal servers for over 10 years and pretty sure it won't be missed as it was a constant nuisance that limited the max size of the map that could be used for a game.

If you're not talking about money goal money doesn't matter. You can cheat it in or you spend 10 minutes and do an moneymaker, it's all the same. If there is no goal there is no victory and no cheating. You can still build long routes like you can build eyecandy landscape or race f1 cars around your town, it just won't make you money.

FLHerne:
>  In the case of green and blue, the drop in income is very large and devastating.
>
> Do you think such building and playing styles should be excluded from this game?

I appreciate the hard data in your graphs, nice work.

I've always been a 'realism'-focused player - the slow winding routes, many sailing ships etc. are my normal way of playing.

I'm not concerned about this patch. OpenTTD is not a difficult game even with that focus; I've never managed *not* to have functionally unlimited money in a few years without very harsh non-default settings and grfs. According to your numbers, even the worst-case routes still break even here, so if your network has any shorter-distance traffic it'll still turn a profit. All the better IMO if I have to work for it a little, that's realism!

As @ldpl says, provided your company doesn't lose money, the absolute amount barely matters in single-player. In multiplayer everyone is affected equally.

LC-Zorg:
@FLHerne Thanks for comment.
> I'm not concerned about this patch (...)

But what if someone only wants to build long routes? What if someone has such a game and wants to continue it? If currently these routes are profitable, as in the case of green and also blue, after this change, they will only be losses. In this case, the costs of maintaining the trains are very low, but if they were higher, the decrease in income could even exceed 200%, i.e. the loss would be the same as the current income. This may be impossible to repair and will result in the loss of all work. Newly started games in this style will also not be possible. Ok, they will, but it requires a lot of knowledge from the player what add-ons offer. Not every player will have such knowledge.

> because CS in 1995 didn't make his profit formula with 4k maps in mind.

Yes, the game and rate graph created in 1995 did not include 1k maps, let alone 4k maps. These rates were for 256x256 maps (?). Today, the maps are much larger, so the rates should fall much more slowly for playing on such maps to have a similar effect. Meanwhile, instead of slowing down this decline, you accelerate it and bring it to zero, making playing on large maps often impossible. Where is the logic here? Why were these large maps introduced to the game? Again, you're making this change for your playstyle, which isn't a bad thing in itself, but you're doing it while ignoring players who want to play differently than you. You're ruining their game just to make it better suited to your needs. Do you want only short and medium-length routes to be profitable? Then create the right add-on and add it to your games, but don't impose on others how they should play and don't destroy their existing, favorite games.

> Are you talking about money goal game? / If you're not talking about money goal money doesn't matter.

Yes, the game and rate graph created in 1995 did not include 1k maps, let alone 4k maps. These rates were for 256x256 maps (?). Today, the maps are much larger, so the rates should fall much more slowly for playing on such maps to have a similar effect. Meanwhile, instead of slowing down this decline, you accelerate it and bring it to zero, making playing on large maps often impossible. Where is the logic here? Why were these large maps introduced to the game? Again, you're making this change for your playstyle, which isn't a bad thing in itself, but you're doing it while ignoring players who want to play differently than you. You're ruining their game just to make it better suited to your needs. Do you want only short and medium-length routes to be profitable? Then create the right add-on and add it to your games, but don't impose on others how they should play and don't destroy their existing, favorite games.

From what you write, I can see that the problem is that you really don't see that someone can play differently. Not only money or not money. You write about only two styles of play, where one is based on earning as much money as possible and competing with others, and where the other is focused only on making the game look nice. And what about the third group, i.e. those who expect realism from the game? What if someone expects the game to be logical and its mechanics, though simplified, allow you to create what exists in the real world? What if, as I wrote earlier, someone wants to build really long connections on the big map? Or medium length but with slow vehicles? Why do you think this should be impossible? In the 19th century, the journey from Portland to Sacrament, California (ca. 950 km) took six days and was considered express pace. 
w-2009-94
Today you can cover the same route by car according to navigation in less than 9 hours. On the scale of this game, something that takes 100 days today would have taken over 500 before 1900. Not only that! 
goldrush-california_gold_rush_handbill jpg__400x501_q85_crop_subsampling-2_upscale
The journey from Nicaragua to California lasted "only" 35 days! Do you know how much that would be on the scale of this game? Do you think that transport companies only lost on this? If that were the case, no one would offer such connections. It was quite the opposite! It was for these companies that the gold rush that drove poor people to travel far was a real goldmine, for those who made those trips possible. [(wiki)](https://en.wikipedia.org/wiki/SS_California_(1848)) If this game is going to have some realism, at least players shouldn't lose out on this way of playing. That's how it's always been until now. I admit that a flat rate is not the best solution and replacing it with a drop is a good solution, but by lowering the payout like this, you are simply destroying both the sense of playing on large maps and the realism of the game. Realism is not only nice graphics, but also sensible economy, and you are destroying it for bigger maps.

For me, it's generally not a bad change, but it needs modification.

Andrew350:
> The journey from Nicaragua to California lasted "only" 35 days! Do you know how much that would be on the scale of this game? Do you think that transport companies only lost on this? If that were the case, no one would offer such connections. It was quite the opposite! It was for these companies that the gold rush that drove poor people to travel far was a real goldmine, for those who made those trips possible. [(wiki)](https://en.wikipedia.org/wiki/SS_California_(1848)) If this game is going to have some realism, at least players shouldn't lose out on this way of playing. That's how it's always been until now. I admit that a flat rate is not the best solution and replacing it with a drop is a good solution, but by lowering the payout like this, you are simply destroying both the sense of playing on large maps and the realism of the game. Realism is not only nice graphics, but also sensible economy, and you are destroying it for bigger maps.

OpenTTD is _**not**_ realistic. That is doubly true for the economy. The scale and economy of OpenTTD is based on gameplay progression, by incentivizing making faster vehicles to make more profit. In this model there is no reason to expect a horse-and-carriage which takes 2.5 years to travel 4k tiles will make any profit at all, quite the opposite in fact. Trying to use real life examples to argue that this is wrong is just silly since the game has virtually nothing in common with reality in this respect.

Besides, your preferred playstyle is obviously a sandbox style in which you don't care about profits anyway, so what's really the problem? You can always just cheat yourself some money so you don't have to worry about it. Or if you really care about realism, as ldpl already suggested just make a simple money maker route to subsidize the losing ones. Even in "real life" it is not uncommon for transport companies to run certain routes at a loss and use more profitable routes to offset them, often to maintain marketshare and/or keep other higher profit routes viable.

ldpl:
> Why were these large maps introduced to the game?

Because players kept asking for more tiles. And the way it was done is not by scaling the distances but by making more of the same stuff. Industry density stays the same, town density stays the same, terrain stays the same so why should it scale payments? For scaling payments there is newgrf property. 

> But what if someone only wants to build long routes? What if someone has such a game and wants to continue it? 

Let's stop worrying about imaginary needs of imaginary players. So far I've seen zero evidence playstyle like that even exist. New players tend to do the opposite and build extremely short routes and experienced players aren't bothered by some trivial money isuses. You can't even build such extremely long routes without acquiring the money for it in some way first. 

> Meanwhile, instead of slowing down this decline, you accelerate it and bring it to zero

Time factor goes to zero but it's multiplied by a distance so payments go to a constant.

> Do you want only short and medium-length routes to be profitable? Then create the right add-on and add it to your games, but don't impose on others how they should play and don't destroy their existing, favorite games.

Did you miss the motivation part of the PR? Before this change it was impossible to create such add-on. Decreasing cargo age period to shorten the optimal distance led to the exact opposite effect making it even better to move far.

And as Andrew350 game is not realistic and was never supposed to be afaict. But even if we ignore that for a second in what real universe does slowly delivering rotten food to the other end of the globe make insane profits?

TheMowgliMan:
"Imaginary needs for imaginary players."  Hmm. I don't think they are
imaginary, given that most servers I've joined have people making large
routes. Not to mention OpenTTDCoop. I also like to make long(er) routes.

Overall, this has to be adjusted to accommodate more play styles.

Finally, this basically kills ships. They were barely playable before, now
they're completely unplayable.

ldpl:
>  given that most servers I've joined have people making large routes. Not to mention OpenTTDCoop. I also like to make long(er) routes. 

On servers players are usually trying to compete, so if they know extremely long routes are brokenly profitable they'll build them. And coop has a whole "Moneymaker" article in the "Game start" section of the wiki :p (https://wiki.openttdcoop.org/Moneymaker). Also there are long routes and there are extremely long routes that are affected by this PR, are you sure you can distinguish between the two? 

> Overall, this has to be adjusted to accommodate more play styles. Finally, this basically kills ships. They were barely playable before, now they're completely unplayable.

Again, for this change has any effect whatsoever cargo needs to be moving for over a year, sometimes close to two years. You can't even send ships that far without placing an unholy amount of buoys.

LC-Zorg:
> Andrew350: OpenTTD is not realistic.

Thanks for comment.
Well, that's your point of view and your interpretation of the game. However, the fact that the game is not a simulator of reality does not mean that it should not refer to reality and in addition eliminate the possibility of such a style of play.

> Andrew350: Besides, your preferred playstyle is obviously a sandbox style

I've already written it before that this is not my style of play. It's definitely not the preferred style. :/
If you paid the slightest attention to most of what I wrote, the most important thing for me in this game is the economy. That it should be reflected in reality and that it should simply make sense. And yes, long routes make sense in reality.

> ldpl: You can't even build such extremely long routes without acquiring the money for it in some way first.

You can take more credit. But then you also have to earn something on this route to pay it back. This is another playstyle you haven't included.

> ldpl: But even if we ignore that for a second in what real universe does slowly delivering rotten food to the other end of the globe make insane profits?

Does the Silk Road mean anything to you? Pepper? Turmeric? Sugar cane? Canned fish? Freeze-dried food?

> ldpl: Also there are long routes and there are extremely long routes that are affected by this PR, are you sure you can distinguish between the two?

And that is also one of the problems with this change. Players can't always tell the difference. They may accidentally make the route too long. And then what? I wrote about this problem earlier in the previous PR ([here](https://github.com/OpenTTD/OpenTTD/pull/9002#issuecomment-974516491)). Another thing you didn't consider.

> ldpl: Again, for this change has any effect whatsoever cargo needs to be moving for over a year, sometimes close to two years. You can't even send ships that far without placing an unholy amount of buoys.

Another nonsense. Ships traveling at a speed of 20 km/h cover about 300 tiles a year. If the route is not straight, it may be 200 or even less. Is that very far in your opinion? At this distance, if there are not too many ships, 3-4 buoys will be enough, and in singleplayer games you don't even need any.

> Let's stop worrying about imaginary needs of imaginary players

So maybe we should talk with these imaginary players about their imaginary needs...

ldpl:
> If you paid the slightest attention to most of what I wrote, the most important thing for me in this game is the economy. That it should be reflected in reality and that it should simply make sense. And yes, long routes make sense in reality.

What part of openttd economy has ever made sense in real life terms? How does moving cargo from the farthest producer make any sense? Or making the most money by moving same type of cargo both ways? Or, in the context of this bug, having the fastest decaying cargo be best transported far? Or having, for example, goods make 3 times more money when transported 2k tiles or 8k tiles then 3.5k tiles? What routes take years in reality?

> You can take more credit. 
Default loan is 300k pounds. You can ofc set it to 2 billions but that's no different from cheating.

> Does the Silk Road mean anything to you? Pepper? Turmeric? Sugar cane? Canned fish? Freeze-dried food?

Game specifically defines food as one of the fastest decaying cargo. Faster than grain or maize. So you're saying we should ignore all that and just imagine that on 4k map it magically turns into pepper? Game also defines that unlike, say, plastic, food starts decaying immediately when put into vehicle. So in what logic should transporting rotting food for years make insane profits?

> 3-4 buoys will be enough, and in singleplayer games you don't even need any.

There is no good way to tell what buoy distance is sufficient in each case for ship not to get lost but openttd wiki recommends 20 tiles (https://wiki.openttd.org/en/Manual/Building%20buoys). So even if you think it's a bright idea to move food by the slowest ship that'll take around 15 buoys. For more reasonable cargo like oil that would be close to 30.

> So maybe we should talk with these imaginary players about their imaginary needs...

I consulted my imaginary players and my imaginary duck. They all said it's fine :p When real players turn up with real problems we can discuss some real solutions. Which probably include scaling payment distances, not keeping old bugs.

TheMowgliMan:
I feel like this PR contradicts some of the goals for #10606.

Eddi-z:
> I feel like this PR contradicts some of the goals for #10606.

Why would you think that? A scale patch like #10606 only can work if the game balance has a proper foundation. And this is one step in the right direction. Now you can easily tweak the cargo age property, without running into weird corner cases because of limits nobody thought of 15 years ago.

TheMowgliMan:
One of the upsides to #10606 is that it allows more realistic games. Now
it's even harder to run realistic games.

Andrew350:
>I've already written it before that this is not my style of play. It's definitely not the preferred style. :/

I took a moment and re-read the whole discussion. Apologies for mistaking this as your playstyle, I had assumed your in-game examples were a display of an effect from your personal games. 

But, this review did at least give me a chance to better summarize my view on the discussion so far:

The words "realism" and "reality" are being thrown around a lot here in a discussion about a theoretical playstyle wherein a player exclusively transports cargo across the most extreme distances possible and with very slow vehicles and expects to make a lot of money.

It is being argued that this patch makes it "impossible" to continue this hypothetical style of play, when it absolutely does not. It just means you don't make such absurdly unrealistic profits when doing so. Besides, even a relatively modest trip of ~500 tiles with just about anything should still make you more money than God if setup properly - this patch does nothing to change that. If anything this patch makes things _more realistic_ by making players pick routes which are more realistically profitable, rather than just thinking "more distance = more money" (albeit only slightly and only in the most extreme cases, but it's an improvement nonetheless).

Even in this extreme and hypothetical situation where someone wants to build an entire transport company with only such extremely long routes, the actual effects of this patch do not interfere with the ability to do so, rather only affecting the _overall profitability_ of it, which is easily overcome by other methods already mentioned earlier.

With this in mind, what _exactly_ is this patch _actually_ taking away? If it really is just the simple matter of extremely long routes no longer making as much money, that is not really a problem because that can be offset very easily by building more sensible routes and/or using the money cheat methods to ignore it. If there is something else I'm missing, however, I would gladly like to hear it, I just think it needs a different framing than "because reality", because so far that reasoning is not making much sense to me. 😄

PR #8609, opened 2021-01-25
Fix: Equalise the number of frames needed for road vehicles to traverse different radius curves

SamuXarick:
## Motivation / Problem
There is a story behind all this, before I started working on the PR.
It all began when I was interested in knowing why travelling vertically/horizontally was more rewarding than in a diagonal. It led me to create a testing scenario where I would run speed tests for ships, trains and road vehicles, traversing 1999 tiles in different straight directions, and using the timetable to count the number of days needed to traverse those tiles, with all vehicles running at the same max speed of 94 km/h.

Savegame: [vehicle speed test large.zip](https://github.com/OpenTTD/OpenTTD/files/5868924/vehicle.speed.test.large.zip)

For directions along the axis (diagonal), between the 3 vehicle types, I expected the results to match. Alas, they don't. Train took 594 days, ship took 588 days and bus took 592 days. Travelling on the way back, the results maintained consistent within each vehicle type: 594 days for train, 588 days for ship, 592 days for bus.

For the other directions (horizontal and vertical), I expected train and ship times to match, but not the bus, due to it not really having a horizontal/vertical road. I tested the bus with a long curvy zig-zag road from start to finish. Between train and ship, train took 396 days and ship took 392 days. On the way back, the results were the same: 396 for train, 392 for ship.
It is now when I look at the bus times that I find an inconsistency between itself:
- Going vertically, from North to South, it took 567 days, but on the way back, from South to North, it took 592 days.
- Going horizontally, from West to East, it took 579 days, the same on the way back, from East to West, 579 days.

Vehicle Speed Test Large, 2104-02-22

This discrepancy is what led me to investigate deep into the issue, and I find the problem is due to the different number of frames road vehicles need to go through to cross curved tiles.

Each road piece has two curves: the inner curve and the outer curve. Inner curves have fewer frames that outer curves, that's to be expected. However, different inner curves have less frames than other inner curves. The same with the outer curves, having less frames than other outer curves.

outer-inner frame count

Going back to the zig-zag curves of the speed test, and reducing the number of tiles to just 2, we get:
- Going vertically, from North to South, 27 frames total (17 for the outer curve + 10 for the inner curve)
- Going vertically, from South to North, 29 frames total (18 for the outer curve + 11 for the inner curve)
- Going horizontally, from West to East, 28 frames total (17 for the outer curve + 11 for the inner curve)
- Going horizontally, from East to West, 28 frames total (18 for the outer curve + 10 for the inner curve)

zig-zag counting

[...]

Issue #6503, opened 2016-09-01
Abnormalities in train subtile coordinates when reversing at the end of line


DorpsGek:
**__dp__ opened the ticket and wrote:**

> I was investigating train movement mechanics and noticed some abnormalities. Opening this bug to sum up what I found so far. 
> 
> I made python script(in attachment) to visualize train path in subtile coordinates. And apparently its location jumps one subtile when it reverses at the end of line (see attached image, loco lenght is 6 subtiles). From what I can tell it's a combination of several bugs:
> 1. Train goes one subtile too far before stopping. Speed also confirms it going 25->5->7...->25 on that extra subtile (that's not showing in gui though).
> 2. Reversing is wrong. It keeps coordinates unchanged and only flips direction. Because of _initial_tile_subcoord, real train position actually depends on train direction(dx,dy): real_x = pos_x + int(dx == -1) + dx * progress%, same for y. So flipping direction moves the train and x,y should be changed to compensate for it.
> 
> Just playing around with code I found that adding 1 to condition in TrainApproachingLineEnd fixes # 1. And removing TrainCheckIfLineEnds call in TrainLocoHandler somehow fixes # 2. But I didn't dig into it enough to understand what's going on there, so not sure if it's a right way to fix it (especially # 2).
> 
> Also I ran out of time I can spend on this for now. Hope someone can pick it up from where I left it ;) Also can be useful to check subpixel movement for other situations and vehicle types.
> 
> P.S. Also attaching diff that outputs coordinates for python script from game and save that I used to generate the image.

**Attachments**
- [visroute.py](https://bugs.openttd.org/task/6503/getfile/10588/visroute.py) (2.90 KiB)
- [reverse_jump_len6.png](https://bugs.openttd.org/task/6503/getfile/10589/reverse_jump_len6.png) (120.31 KiB)
- [subpixel_jump_debug.diff](https://bugs.openttd.org/task/6503/getfile/10590/subpixel_jump_debug.diff) (9.77 KiB)
- [postest_reverse.sav](https://bugs.openttd.org/task/6503/getfile/10591/postest_reverse.sav) (14.02 KiB)

**Reported version**: _1.6.1_
**Operating system**: _All_

---

DorpsGek:
**__dp__ wrote:**

[...]

PR #10597, opened 2023-04-04
Fix #10493: Fix with station stop location adjusted.

### LINKED ISSUE #10493: [Bug]: Trains on different track types can collide in end-to-end stations

TrevorShelton:
## Motivation / Problem

Regarding #10493 and #8424, trains overshoot some stopping points by an infinitesimal amount, leading to train crashing in situations where trains should not be crashing.

## Description

Train's stopping location in stations have been put back by a very slight amount.

## Limitations

#8424 is not fixed in this particular commit.

This adjustment of the stop location may theoretically cause issues regarding the collision of the back of the train, but none have been confirmed as of yet.

## Checklist for Review

### DIFF
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -294,7 +294,7 @@ int GetTrainStopLocation(StationID station_id, TileIndex tile, const Train *v, i
 
 	/* Subtract half the front vehicle length of the train so we get the real
 	 * stop location of the train. */
-	return stop - (v->gcache.cached_veh_length + 1) / 2;
+	return stop - (v->gcache.cached_veh_length) / 2 - 1;
 }
 
 


TrueBrain:
Would you mind sticking to the PR description format that is presented to you when creating a new PR? I noticed this is not your first PR where you replaced it with your own text :)
The format is designed so a few specific questions are answered, and allows us to glance over them rather quickly. Now I need to digest the information from your PR :)

Tnx!

TrueBrain:
No worries :) Just for next time :) Cheers!

rubidium42:
Just like #10598, this probably also needs a savegame bump and afterload logic. Although here the train will just not stop at the station but continue on, instead of crashing the game a bit later.
In the [attached savegame](https://github.com/OpenTTD/OpenTTD/files/11183051/poc.zip) I manually stopped Train 1 just before the stop location. If you load it and unstop it in unmodified OpenTTD, it will stop at the station one position later. With this change it will not stop at the station anymore.
 
When the train is coming from the bottom right, the stop location does seem better visually. When it is coming from the top, it seems to be at a visually worse location, as if it got moved 1 location in the wrong direction. To compare there is a second train next to it, that stopped using the old stop location, so you can compare the two behaviours.

Similar to #10570, I think the whole collision code is full of off-by-one related errors because, lets say a train is two parts long of the normal 8-length bits. Assuming it is completely on a tile, then the from part is say at location 4 and the rear part at location 12, so the front part is from [0-8) and the rear part is [8-16), where 16 is actually 0 of the next tile. If the train were going the other direction, the front part would be (16-8] and the rear part at (8-0]. So, even though they might not be technically touching the code assumes the begin and end are all inclusive. And with odd length parts this problem is compounded even more, as then you need to consider which direction to round to.
So I'm wondering if we are whacking the proverbial moles with these changes, or that the underlying logic is changed to account for the difference in inclusive/exclusive locations that get occupied by vehicles in different directions.

#10598 does make sense as a separate change, as it's just attempting to make the length the same in both cases. Though, is it actually the same? Assume we have a train perpendicular to the depot, so it is at say X=8. If going to lower X, it currently stops at 4 and going to higher X it stops at 10 (A); for these calculations I'm ignoring the whole tile that needs to be traversed in between, but that's the same in both directions. You change it to 5, but 8-5 vs 8-10 still is a different distance.
I reckon the locations where they disappear into the depot are different primarily to ensure the whole vehicle is hidden by the roof of the depot, before it goes away. So changing the 4 into a 5 or 6 might cause visible harm. And if it is crashing at position 4, then that's the same issue as I mentioned before with the off-by-ones.
I would actually argue moving the enter location of the other depots backward, the depots where you do not see the opening result in the vehicle arriving earlier and leaving closer to the main track. As such those provide a slight advantage over others.

TrevorShelton:
Yeah, you do have a point. I did sort of think about a root cause, but, come to think of it, given that #10493 and #8424 aren't the only issues related to this, there's a real possibility it's a _large_ bug, rather than just a piecemeal affair, and that deep cleaning the overall positional logic would be needed. That would probably take more than just correcting one or two returns values.... And even if the fixes do work on the game by itself, the reinterpretation of older save files could very well be causing a number of obscure happenstance issues like those you mentioned.

If a superior fix supersedes this fix, then it should probably replace it, but this could be kept up for now to be open for new commits addressing its own issues as well as being a stand-in candidate while something better is sought.

ldpl:
It's more of a design choice than a bug. TTD and thus OpenTTD vehicle positioning and movement is based on visuals, not logic. So once you start digging deeper it turns out that whatever looks better doesn't always makes sense for a continuous movement and causes all kinds of issues like #6503, #8609 and so on.

2TallTyler:
@TrevorShelton Are you interested in finishing this and your other movement PR with the saveload conversion code described by rubidium42? If not, I am going around and closing abandoned PRs to clean up a bit around here. 😄

PR #10598, opened 2023-04-04
Fix #8424: Modified Direction Depot Entry Coordinates to Try to Fix Wrongful Collision

### LINKED ISSUE #8424: Trains can crash through depots

TrevorShelton:
## Motivation / Problem

"#8424: Trains can crash through depots"

## Description

This patch appears to fix the wrongful train crashing, and it changing two directions may be a good sign because it was said the bug was happening in two directions. Counting steps from 5 to 0 also match A to F in number. No process crashing seems to be caused by this, either.

## Limitations

Further testing would be wise.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -2885,7 +2885,7 @@ static void ChangeTileOwner_Track(TileIndex tile, Owner old_owner, Owner new_own
 }
 
 static const byte _fractcoords_behind[4] = { 0x8F, 0x8, 0x80, 0xF8 };
-static const byte _fractcoords_enter[4] = { 0x8A, 0x48, 0x84, 0xA8 };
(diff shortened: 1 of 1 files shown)


rubidium42:
This requires a savegame bump and afterload logic to resolve the situation where the vehicle is at the threshold of going into the depot before this change: [oops.zip](https://github.com/OpenTTD/OpenTTD/files/11181000/oops.zip)

Issue #10028, opened 2022-09-12
[Bug]: RV overtaking fails if multiple vehicles follow closely


frosch123:
Oldest report about this: #9236 (kept this ticket, as it is more descriptive to the problem at hand)

### Version of OpenTTD

12.2

### Expected result

When a queue of road vehicles approaches a stopped/broken-down RV, they should overtake the RV one after another.

### Actual result

Depending on where the RV to overtake is stopped/broken-down, overtaking will fail completely, and all RV in queue are stuck.

The actual test that fails the overtaking attempt is:
https://github.com/OpenTTD/OpenTTD/blob/master/src/roadveh_cmd.cpp#L796
```
	/* Are there more vehicles on the tile except the two vehicles involved in overtaking */
	return HasVehicleOnPos(od->tile, od, EnumFindVehBlockingOvertake);
```

This check does not only check and fail for vehicles in front, but also for vehicles behind on the same tile.

### Steps to reproduce

1. Load attached savegame.
    [overtake_queue.zip](https://github.com/OpenTTD/OpenTTD/files/9550609/overtake_queue.zip)
2. Unpause.
    ![overtake_queue1](https://user-images.githubusercontent.com/23150643/189723531-0cf9f1d6-c4f0-4d61-89a4-309844a72800.png)
3. Observe one side succeeding with overtaking, the other side failing.
    ![overtake_queue2](https://user-images.githubusercontent.com/23150643/189723589-6e5c3657-01bb-4605-9f19-59c7396d494f.png)

LordAro:
Duplicate of #9033 & #9236 ;)

TrueBrain:
#9033 is not a duplicate of this ticket. In #9033 overtaking does happen, but ... something odd happens during it.

isacvandunem:
From what i quickly tested this seems to happen when the stopped/broken down vehicle is mid way through tiles, even though its not consistent.
I also noticed that even in the given save file, if i wait long enough, say 30 seconds or a minute the vehicles eventually start overtaking the stopped one.

Issue #9236, opened 2021-05-11
Vehicles do not pass stopped/broken down vehicles


unbeatable-101:
## Version of OpenTTD

OpenTTD Version 1.11.2 (1.11.2), macOS 11.3.1 (20E241)
## Expected result

As with slow vehicles, broken down ones get passed as well
## Actual result

The vehicle blocks traffic
## Steps to reproduce

Wait for a vehicle to break down with more vehicles behind it

https://user-images.githubusercontent.com/58151048/117745154-f87da180-b1d7-11eb-811a-220d1cdde4e0.mp4

unbeatable-101:
![Screen Shot 2021-05-11 at 14 12 12](https://user-images.githubusercontent.com/58151048/117864673-10e6ce00-b263-11eb-9fed-af730bce39dd.png)

TrueBrain:
This annoys me so much. The overtaking feels to random sometimes ... but a stopped vehicle .. why the game doesn't overtake it, I dunno .. feels like a bug to me :) But it is as old as the game itself ;)

TrueBrain:
Duplicate of #10028 (strictly seen, yours is older, but #10028 contains a bit more details.)

Issue #9033, opened 2021-04-12
RV road rage during overtaken


TrueBrain:
## Version of OpenTTD
1.11.0

## Expected result
RVs to overtaken like civil people (read: let the bus not make the mail truck brake and not let the mail truck exceed his maximum speed).

## Actual result
RVs during overtaken are one of the worst drivers. Attached movie. Few things to observe:

- Bus has to make an emergency break for a squirrel (not shown in movie)
- After bus overtakes, mail truck accelerates to max-speed of bus, as the bus just cut it off, and it is flipping the finger

## Steps to reproduce
- Build mail truck
- Build bus truck
- Send mail truck out of depot, right after, send bus truck
- Enjoy some road rage

https://user-images.githubusercontent.com/1663690/114455535-2a511780-9bdc-11eb-8901-02407342fe80.mp4

TrueBrain:
A special moment happens around ~10 seconds btw:
![image](https://user-images.githubusercontent.com/1663690/114456415-1eb22080-9bdd-11eb-891e-52d6d337c52f.png)

:)

LordAro:
This only appears to be an issue when the vehicles have close top speeds (i.e. MPS Bus & Mail truck, which have 30 & 35mph respectively (48/56kph))

I think, though am not certain, that the "special moment" happens when the overtaking vehicle crosses into the next tile, sees that it is clear, and stops overtaking. Of course at more or less the same moment the overtaken vehicle crosses into the same tile and we get the intersection of vehicles seen here. Presumably this also happens at faster overtaking speeds, but is less obvious. No idea how the vehicle is allowed to "overspeed" beyond its top speed though. Probably something to do with resolving the intersection.

PR #10602, opened 2023-04-07
Change #10028: Improved Overtaking Routines

TrevorShelton:
## Motivation / Problem

Issue #10028

## Description

This makes road vehicle smarter at overtaking through adjusting the overtaking routine and time.

## Limitations

Tweaking could theoretically make it more ideal or powerful. Feel free to test and compare with initial version.

## Checklist for review


TrevorShelton:
Huh. It looked like the white space was gone from the file, but it still complained?

Anyways, here's just some makeshift saves made on top of the demo to test.
[save.tar.gz](https://github.com/OpenTTD/OpenTTD/files/11175616/save.tar.gz)

src/roadveh_cmd.cpp
```diff
@@ -833,10 +841,10 @@ static void RoadVehCheckOvertake(RoadVehicle *v, RoadVehicle *u)
 	 *  - No barred levelcrossing
 	 *  - No other vehicles in the way
 	 */
+
```
2TallTyler: Random newline snuck in

src/roadveh_cmd.cpp
```diff
@@ -768,8 +768,16 @@ struct OvertakeData {
 static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
 {
 	const OvertakeData *od = (OvertakeData*)data;
-
-	return (v->type == VEH_ROAD && v->First() == v && v != od->u && v != od->v) ? v : nullptr;
+	TileIndexDiff    front = TileOffsByDiagDir(DirToDiagDir(od->v->direction));
+
+	/* Distinct road engine is blocking if it's within the tile or the next few of the passer as well as either it being
+	 * directionally opposed to passer, the passed pending movement, or the blocker being stopped while overtaking. */
+	return (v->type == VEH_ROAD && v->First() == v && v != od->u && v != od->v
```
2TallTyler: This should be broken up into multiple returns for readability, preferably with comments explaining each case. The ternary at the end is particularly confusing.

src/roadveh_cmd.cpp
```diff
@@ -833,10 +841,10 @@ static void RoadVehCheckOvertake(RoadVehicle *v, RoadVehicle *u)
 	 *  - No barred levelcrossing
 	 *  - No other vehicles in the way
 	 */
+
 	od.tile = v->tile;
 	if (CheckRoadBlockedForOvertaking(&od)) return;
-
-	od.tile = v->tile + TileOffsByDiagDir(DirToDiagDir(v->direction));
+	od.tile += TileOffsByDiagDir(DirToDiagDir(v->direction));
```
2TallTyler: I suspect this is wrong. We use TileOffsByDiagDir() to get the next tile in the correct direction...just incrementing to the next tile index is probably in the wrong direction. (I could be wrong if I'm misreading your usage.)

2TallTyler (commented):
This needs more explanation of the changes you've made and why. Some of these may explain themselves once you clean up that return and add comments, but why reduce the time for overtaking?

Also, we generally use the prefix "Fix" when closing issues, and I had to Google "Ameliorated" so maybe `Fix #10028: Improve vehicle overtaking` would be a better commit name. 😉

TrevorShelton:
The overtake time was reduced, if I recall, in want to compact the space an overtaking vehicle takes up, though not to the point that they'd look like they're phasing through the vehicle they're overtaking. As for the long return, I did try breaking it up with lines to make it more readable, but it probably does make more sense, is more consistent, and is easier to edit if it is broken up into such many returns. Also, I used that word because I didn't want to insinuate that everything had necessarily been fixed, but just "fix" is probably the better word here anyways. As for the other minor edits, those were probably unnecessary, but the tile offset value I put was, I thought, synonymous and maybe a slight improvement. I will revise these and try to use comments to try and explain the rest.

src/roadveh_cmd.cpp
PeterN: This isn't a behaviour change, od.tile is already initialised v->tile, this change just replaces the reassignment.

2TallTyler: Aha, I was misreading it. 🙂

src/roadveh_cmd.cpp
```diff
@@ -769,7 +769,38 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
...
+
+	/* From here, the third vehicle is blocking if one of the below is true. */
+
+	/* 1: The third vehicle is going the opposite way, so it's not wise to pass. */
+	if (v->direction == ReverseDir(od->v->direction)) return v;
+
+	/* 2: If the passee is active but not moving, then it is likely wanting to overtake as well.
+	 * It's safer to let it deal with this first. If it is moving, it is better to close passer
+	 * and passee's distance first than it is to try and overtake. */
+	if (!(od->u->vehstatus & VS_STOPPED)) return v;
```
2TallTyler: `VS_STOPPED` is only when a vehicle has been manually stopped by a player. You're probably looking for `cur_speed == 0` (same in case 3 below). 🙂

2TallTyler (commented):
Speaking of vehicle flags, you might also consider broken-down vehicles (#9236).

You may also want to look at #9033, but perhaps that's out of scope for what you're doing, since it hasn't really been diagnosed properly. 🙂

src/roadveh_cmd.cpp
TrevorShelton: For case three, it being stopped for any reason is probably ok to check for, but case two specifically checks for if the to-be-passed is ready to move, even if not moving, a.k.a not manually stopped or broken down.

src/roadveh.h
```diff
@@ -77,7 +77,7 @@ static const uint RVC_DRIVE_THROUGH_STOP_FRAME           = 11;
 static const uint RVC_DEPOT_STOP_FRAME                   = 11;
 
 /** The number of ticks a vehicle has for overtaking. */
-static const byte RV_OVERTAKE_TIMEOUT = 35;
+static const byte RV_OVERTAKE_TIMEOUT = 25;
```
2TallTyler: I don't think this should be changed without data to support it from thorough testing. Chesterton's Fence, you know? 😄

src/roadveh_cmd.cpp
```diff
@@ -769,7 +769,38 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
 {
 	const OvertakeData *od = (OvertakeData*)data;
 
-	return (v->type == VEH_ROAD && v->First() == v && v != od->u && v != od->v) ? v : nullptr;
+	/* This is the offset to tile in front of either the passer the passee, looking the passer's way. */
+	TileIndexDiff front = TileOffsByDiagDir(DirToDiagDir(od->v->direction));
```
2TallTyler, line 776: This should be defined closer to its use below. And maybe `tile_step` or similar would be a better name than `front`, which has a different meaning already relating to articulated vehicles.

src/roadveh_cmd.cpp
```diff
@@ -769,7 +769,38 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
...
+	if (v->type != VEH_ROAD || v->First() != v) return nullptr;
+
+	/* A vehicle must be a third party to block. If it is the same as either the passer
+	 * or passee, it's just part of the base overtake and there is no blockage. */
+	if (v == od->u || v == od->v) return nullptr;
+
+	/* Overtaking concerns the overtaker's tile as well as the few in front of them.
+	 * If a third party is not in these tiles, it is not blocking the overtake.
+	 * If the overtakee and overtaker are on separate tiles, look one tile further. */
+	if ( v->tile != od->v->tile
```
2TallTyler, line 788: I would split this up into two checks and use an iterator for the first, doing something like `od->v->tile + (i * next_tile)` to step ahead each time. 

The second check, if the overtaker and overtakee are on different tiles, would be outside the loop.

src/roadveh_cmd.cpp
```diff
@@ -769,7 +769,38 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
...
+	  && v->tile != od->v->tile + front
+	  && v->tile != od->v->tile + front + front
+	  && v->tile != od->u->tile + front + front) return nullptr;
+
+	/* From here, the third vehicle is blocking if one of the below is true. */
+
+	/* 1: The third vehicle is going the opposite way, so it's not wise to pass. */
+	if (v->direction == ReverseDir(od->v->direction)) return v;
+
+	/* 2: If the passee is active but not moving, then it is likely wanting to overtake as well.
```
2TallTyler, line 798: This comment doesn't make sense for the code below, which doesn't check vehicle speed. 🙂

TrevorShelton:
Will redo this with updated master and then try to add suggestions.

PR #10604, opened 2023-04-07
Fix #10578: Allow to select any version of AI/GS from GUI

### LINKED ISSUE #10578: [Bug]: No way in GUI to select a previous version of AI or GS
### Version of OpenTTD

13.0 (old bug)

### Expected result

The creator or tester of AI/GS is free to enable different versions of them at the same time for comparison and further improvement.
The player can choose a particular earlier version of AI/GS that they liked more or that simply didn't have the bugs that were evident when using it.

### Actual result

To test a previous version of AI or GS, all newer versions must be removed from the folder. Once this is done, it is not possible to re-enable the newer version - again you have to restore the deleted versions. This is cumbersome and inconsistent with what you might expect by selecting a setting that allows you to view older versions of add-ons.

Personally, I recently tested AAAHogEx, I noticed differences in behavior after the last update, but I wasn't sure if it was a coincidence or a real effect. So I wanted to run the previous version... but that turned out to be impossible*.

Recently, I also created another alternative version of my RWD script and here, too, I used several versions at the same time for comparison. There was no problem here because each of them was renamed taking into account the version. If I wanted to publish it so that the player could choose, even though the changes are rather cosmetic, I would have to publish each version as a separate add-on, thus messing up the list of add-ons.

### Steps to reproduce

If you have multiple versions of one of AI or GS:
- try to run the latest version, then select the previous one, after that select the latest one again
- try running several games at once, each with a different AI/SI version

glx22:
Closes #10578
## Motivation / Problem
Only the latest version of script is listed in GUI.
For AI it's possible to start any version using `start_ai ai.version` console command.

## Description
Add a mechanism (via Ctrl key) to list all available versions of script and not only the latest version.

Window opened with Ctrl pressed:
image

Window opened without Ctrl pressed:
image

## Checklist for review

### DIFF
--- a/src/script/script_gui.cpp
+++ b/src/script/script_gui.cpp
@@ -54,19 +54,21 @@ struct ScriptListWindow : public Window {
 	CompanyID slot;                     ///< The company we're selecting a new Script for.
 	int line_height;                    ///< Height of a row in the matrix widget.
 	Scrollbar *vscroll;                 ///< Cache of the vertical scrollbar.
+	bool show_all;                      ///< Whether to show all available versions.
 
 	/**
 	 * Constructor for the window.
 	 * @param desc The description of the window.
 	 * @param slot The company we're changing the Script for.
+	 * @param show_all Whether to show all available versions.
 	 */
-	ScriptListWindow(WindowDesc *desc, CompanyID slot) : Window(desc),
-		slot(slot)
+	ScriptListWindow(WindowDesc *desc, CompanyID slot, bool show_all) : Window(desc),
+		slot(slot), show_all(show_all)
 	{
 		if (slot == OWNER_DEITY) {
-			this->info_list = Game::GetUniqueInfoList();
+			this->info_list = this->show_all ? Game::GetInfoList() : Game::GetUniqueInfoList();
 		} else {
-			this->info_list = AI::GetUniqueInfoList();
+			this->info_list = this->show_all ? AI::GetInfoList() : AI::GetUniqueInfoList();
 		}
 
 		this->CreateNestedTree();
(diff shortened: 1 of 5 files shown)


src/script/script_gui.cpp
```diff
@@ -54,19 +54,21 @@ struct ScriptListWindow : public Window {
 	CompanyID slot;                     ///< The company we're selecting a new Script for.
 	int line_height;                    ///< Height of a row in the matrix widget.
 	Scrollbar *vscroll;                 ///< Cache of the vertical scrollbar.
+	bool show_all;                      ///< Wether to show all available versions.
```
2TallTyler:
```suggestion
	bool show_all;                      ///< Whether to show all available versions.
```

src/script/script_gui.cpp
```diff
@@ -54,19 +54,21 @@ struct ScriptListWindow : public Window {
...
 
 	/**
 	 * Constructor for the window.
 	 * @param desc The description of the window.
 	 * @param slot The company we're changing the Script for.
+	 * @param show_all Wether to show all available versions.
```
2TallTyler:
```suggestion
	 * @param show_all Whether to show all available versions.
```

src/script/script_gui.cpp
```diff
@@ -262,13 +267,14 @@ static WindowDesc _script_list_desc(
 );
 
 /**
- * Open the AI list window to chose an AI for the given company slot.
- * @param slot The slot to change the AI of.
+ * Open the Script list window to chose a script for the given company slot.
+ * @param slot The slot to change the script of.
+ * @param show_all Wether to show all available versions.
```
2TallTyler:
```suggestion
 * @param show_all Whether to show all available versions.
```

src/script/script_gui.cpp
```diff
@@ -120,11 +122,14 @@ struct ScriptListWindow : public Window {
...
 					tr.top += this->line_height;
 				}
+				StringID str = this->show_all ? STR_AI_CONFIG_NAME_VERSION : STR_JUST_RAW_STRING;
 				int i = 0;
 				for (const auto &item : *this->info_list) {
 					i++;
 					if (this->vscroll->IsVisible(i)) {
-						DrawString(tr, item.second->GetName(), (this->selected == i - 1) ? TC_WHITE : TC_ORANGE);
+						SetDParamStr(0, item.second->GetName());
+						SetDParam(1, item.second->GetVersion());
```
2TallTyler, line 131: Honest question: Is it not a problem or bad form to set a DParam that never gets consumed?

glx22: String parameters are in a fixed size array of uint64 values, so consumed or not doesn't matter.

2TallTyler: Thanks, good to know. 😃

2TallTyler (approved):

glx22:
Manual backport is probably possible if wanted. Can't be automatic because https://github.com/OpenTTD/OpenTTD/commit/b080f24bd87d4f3130ea12abbc40f8b0ad92a8cf.

LC-Zorg:
Thanks for this PR. :)
For some reason, information about the ability to use Ctrl+ is only available in English. For other languages, I don't even see that a translation is missing (99%).
Besides, I don't know if it wouldn't be better if newgrf_show_old_versions = true would also work for this window. I think if someone wants to see previous versions, they can also expect or at least accept to see older versions of AI/GS. This doesn't necessarily preclude the use of Ctrl+.

glx22:
It's not a new string, but a modified one. I don't think outdated translations are counted as missing.

PR #10606, opened 2023-04-07
Feature: Setting to scale cargo production of towns and industries

2TallTyler:
## Motivation / Problem

In #11428 I present my solution to the old daylength patch which is so popular in JGR's Patchpack.

In my conversations with players, daylength actually has two common usecases:
* Slowing down the rate of calendar and technology progression. This is handled by #11428.
* Slowing down the economy, to enable less-dense and less-frequent networks, timetabling, and other "realistic" gameplay styles. This PR is my attempt to meet this need.

Slowing down economy time has several pieces:
* Reducing industry cargo production
* Reducing town cargo production
* Slowing down vehicle running costs when trains are running empty, stabled in yards between runs, etc.
* Slowing station rating loss from infrequent service

All of these are possible with NewGRFs. However, the last two are much easier with the Base Costs GRF and #11346.

Reducing town and industry production would require the player to fork the industry and house sets of their choice and make changes to the source code before recompiling. This is problematic with closed-source GRFs and abandoned GRFs that would need decompiling into uncommented NFO.

I think it's worth offering players this control in OpenTTD itself.

JGRPP has settings that do the same thing although it's applied on top of the daylength speed and the setting value is exponential so I have no idea how to interpret it). But the survey results indicate plenty of players tweaking production to suit their play styles:
* [Industry cargo](https://survey.openttd.org/summaries/2024/wk01#jgrpp-0.56.2-game.settings.economy.industry_cargo_scale_factor)
* [Town cargo](https://survey.openttd.org/summaries/2024/wk01#jgrpp-0.56.2-game.settings.economy.town_cargo_scale_factor)

## Description

Adds two settings to scale the cargo production of:
* Towns (including company HQ)
* Industries

For houses, HQ, and basic industries, this simply scales the cargo by the selected percentage (using random to make up for rounding-to-zero losses).

Industries which use the "every 256 ticks" callback often do their own math with input cargos, and we don't want to break them by changing their output without their knowledge. We handle this by scaling the callback interval instead.

NewGRFs are unaware of the scale. They do not need to know about it, and giving them variables just gives authors footguns. Game Scripts can access settings directly so they don't need API additions.

The lower limit of cargo scaling for both industries and towns is 15%. This is a nicer number than the actual minimum (13%), which is determined by the 256-tick industry production callback. Any slower and the industry would not produce every month, leading to confusing feedback in the UI ("no production last month?"). Towns technically have no lower limit, but 15% is a nice limit where towns still produce any cargo. The actual minimum, 1%, typically has zero production even in large cities, which will confuse cargodist... So we make it 15% to match industries. 🙂 

## Limitations

* Industries which use custom text to describe their cargo needs (FIRS, Industries of the Caribbean, etc) won't be accurate. I welcome ideas from NewGRF authors for how we can help them provide descriptive helptext.
* Industries which produce passengers (Plaza as Industry, Apartment Blocks) are controlled by the Industry setting. Not out of laziness, but because this be argued either way so I chose the option with fewer edge cases. 😃 
* Houses have a cargo production callback called every 256 ticks, but since houses don't have input stockpiles, I don't bother scaling the callback interval and just scale the amount.
* I did consider full economy scaling (including cargo value scaling), but that's more opinionated than letting players control cargo production individually without changing the rest of the economy.

## Checklist for review

### DIFF
--- a/src/economy_func.h
+++ b/src/economy_func.h
@@ -15,6 +15,8 @@
 #include "cargo_type.h"
 #include "vehicle_type.h"
 #include "company_type.h"
+#include "settings_type.h"
+#include "core/random_func.hpp"
 
 void ResetPriceBaseMultipliers();
 void SetPriceBaseMultiplier(Price price, int factor);
@@ -47,6 +49,50 @@ bool AddInflation(bool check_year = true);
 inline bool EconomyIsInRecession()
 {
 	return _economy.fluct <= 0;
+}
+
+/**
+ * Scale a number by the inverse of the cargo scale setting, e.g. a scale of 25% multiplies the number by 4.
+ * @param num The number to scale.
+ * @param town Are we scaling town production, or industry production?
+ * @return The number scaled by the inverse of the cargo scale setting, minimum of 1.
+ */
+static uint ScaleByInverseCargoScale(uint num, bool town)
+{
+	uint16_t percentage = (town ? _settings_game.economy.town_cargo_scale : _settings_game.economy.industry_cargo_scale);
+
+	/* We might not need to do anything. */
+	if (percentage == 100) return num;
+
+	/* Never return 0, since we often divide by this number. */
+	return std::max((num * 100) / percentage, 1u);
+}
+
+/**
+ * Scale a number by the cargo scale setting.
+ * @param num The number to scale.
+ * @param town Are we scaling town production, or industry production?
+ * @return The number scaled by the current cargo scale setting. May be 0.
+ */
+inline uint ScaleByCargoScale(uint num, bool town)
+{
+	/* Don't bother scaling in the menu, especially since settings don't exist when starting OpenTTD and trying to read them crashes the game. */
+	if (_game_mode == GM_MENU) return num;
+
+	if (num == 0) return num;
+
+	uint16_t percentage = (town ? _settings_game.economy.town_cargo_scale : _settings_game.economy.industry_cargo_scale);
+
+	/* We might not need to do anything. */
+	if (percentage == 100) return num;
+
+	uint scaled = (num * percentage) / 100;
+
+	/* We might round down to 0, so we compensate with a random chance approximately equal to the economy scale,
+	 *  e.g. at 25% scale there's a 1/4 chance to round up to 1. */
+	if (scaled == 0 && Chance16(1, ScaleByInverseCargoScale(1, town))) return 1;
+
+	return scaled;
 }
 
 #endif /* ECONOMY_FUNC_H */
--- a/src/industry_cmd.cpp
+++ b/src/industry_cmd.cpp
@@ -1132,7 +1132,24 @@ static void ChopLumberMillTrees(Industry *i)
 
 	TileIndex tile = i->location.tile;
 	if (CircularTileSearch(&tile, 40, SearchLumberMillTrees, nullptr)) { // 40x40 tiles  to search.
-		i->produced[0].waiting = ClampTo<uint16_t>(i->produced[0].waiting + 45); // Found a tree, add according value to waiting cargo.
+		i->produced[0].waiting = ClampTo<uint16_t>(i->produced[0].waiting + ScaleByCargoScale(45, false)); // Found a tree, add according value to waiting cargo.
+	}
+}
+
+/**
+ * Helper for ProduceIndustryGoods that scales and produces cargos.
+ * @param i The industry
+ * @param scale Should we scale production of this cargo directly?
+ */
+static void ProduceIndustryGoodsHelper(Industry *i, bool scale)
+{
+	for (auto &p : i->produced) {
+		if (!IsValidCargoID(p.cargo)) continue;
+
+		uint16_t amount = p.rate;
+		if (scale) amount = ScaleByCargoScale(amount, false);
+
+		p.waiting = ClampTo<uint16_t>(p.waiting + amount);
 	}
 }
 
@@ -1155,15 +1172,23 @@ static void ProduceIndustryGoods(Industry *i)
 
 	i->counter--;
 
-	/* produce some cargo */
+	/* If using an industry callback, scale the callback interval by cargo scale percentage. */
+	if (HasBit(indsp->callback_mask, CBM_IND_PRODUCTION_256_TICKS)) {
+		if (i->counter % ScaleByInverseCargoScale(Ticks::INDUSTRY_PRODUCE_TICKS, false) == 0) {
+			IndustryProductionCallback(i, 1);
+			ProduceIndustryGoodsHelper(i, false);
+		}
+	}
+
+	/*
+	 * All other production and special effects happen every 256 ticks, and cargo production is just scaled by the cargo scale percentage.
+	 * This keeps a slow trickle of production to avoid confusion at low scale factors when the industry seems to be doing nothing for a long period of time.
+	 */
 	if ((i->counter % Ticks::INDUSTRY_PRODUCE_TICKS) == 0) {
-		if (HasBit(indsp->callback_mask, CBM_IND_PRODUCTION_256_TICKS)) IndustryProductionCallback(i, 1);
+		/* Handle non-callback cargo production. */
+		if (!HasBit(indsp->callback_mask, CBM_IND_PRODUCTION_256_TICKS)) ProduceIndustryGoodsHelper(i, true);
 
 		IndustryBehaviour indbehav = indsp->behaviour;
-		for (auto &p : i->produced) {
-			if (!IsValidCargoID(p.cargo)) continue;
-			p.waiting = ClampTo<uint16_t>(p.waiting + p.rate);
-		}
 
 		if ((indbehav & INDUSTRYBEH_PLANT_FIELDS) != 0) {
 			uint16_t cb_res = CALLBACK_FAILED;
@@ -1825,7 +1850,7 @@ static void DoCreateNewIndustry(Industry *i, TileIndex tile, IndustryType type, 
 		}
 
 		for (auto &p : i->produced) {
-			p.history[LAST_MONTH].production += p.rate * 8;
+			p.history[LAST_MONTH].production += ScaleByCargoScale(p.rate * 8, false);
 		}
 	}
 
--- a/src/object_cmd.cpp
+++ b/src/object_cmd.cpp
@@ -688,7 +688,13 @@ static void TileLoop_Object(TileIndex tile)
 	/* Top town buildings generate 250, so the top HQ type makes 256. */
 	if (GB(r, 0, 8) < (256 / 4 / (6 - level))) {
 		uint amt = GB(r, 0, 8) / 8 / 4 + 1;
+
+		/* Production is halved during recessions. */
 		if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
+
+		/* Scale by cargo scale setting. */
+		amt = ScaleByCargoScale(amt, true);
+
 		MoveGoodsToStation(CT_PASSENGERS, amt, SourceType::Headquarters, GetTileOwner(tile), stations.GetStations());
 	}
 
@@ -697,7 +703,13 @@ static void TileLoop_Object(TileIndex tile)
 	 * equations. */
 	if (GB(r, 8, 8) < (196 / 4 / (6 - level))) {
 		uint amt = GB(r, 8, 8) / 8 / 4 + 1;
+
+		/* Production is halved during recessions. */
 		if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
+
+		/* Scale by cargo scale setting. */
+		amt = ScaleByCargoScale(amt, true);
+
 		MoveGoodsToStation(CT_MAIL, amt, SourceType::Headquarters, GetTileOwner(tile), stations.GetStations());
 	}
 }
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -561,6 +561,8 @@ struct EconomySettings {
 	bool   infrastructure_maintenance;       ///< enable monthly maintenance fee for owner infrastructure
 	TimekeepingUnits timekeeping_units;      ///< time units to use for the game economy, either calendar or wallclock
 	uint16_t minutes_per_calendar_year;      ///< minutes per calendar year. Special value 0 means that calendar time is frozen.
+	uint16_t town_cargo_scale;               ///< scale cargo production of towns by this percentage.
+	uint16_t industry_cargo_scale;           ///< scale cargo production of industries by this percentage.
 };
 
 struct LinkGraphSettings {
--- a/src/table/settings/economy_settings.ini
+++ b/src/table/settings/economy_settings.ini
@@ -315,3 +315,29 @@ strval   = STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE
 pre_cb   = [](auto) { return _game_mode == GM_MENU || _settings_game.economy.timekeeping_units == 1; }
 post_cb  = ChangeMinutesPerYear
 cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.town_cargo_scale
+type     = SLE_UINT16
+flags    = SF_NO_NETWORK
+def      = DEF_CARGO_SCALE
+min      = MIN_CARGO_SCALE
+max      = MAX_CARGO_SCALE
+interval = 1
+str      = STR_CONFIG_SETTING_TOWN_CARGO_SCALE
+strhelp  = STR_CONFIG_SETTING_TOWN_CARGO_SCALE_HELPTEXT
+strval   = STR_CONFIG_SETTING_CARGO_SCALE_VALUE
+cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.industry_cargo_scale
+type     = SLE_UINT16
+flags    = SF_NO_NETWORK
+def      = DEF_CARGO_SCALE
+min      = MIN_CARGO_SCALE
+max      = MAX_CARGO_SCALE
+interval = 1
+str      = STR_CONFIG_SETTING_INDUSTRY_CARGO_SCALE
+strhelp  = STR_CONFIG_SETTING_INDUSTRY_CARGO_SCALE_HELPTEXT
+strval   = STR_CONFIG_SETTING_CARGO_SCALE_VALUE
+cat      = SC_BASIC
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -521,6 +521,34 @@ static void AdvanceHouseConstruction(TileIndex tile)
 }
 
 /**
+ * Generate cargo for a house, scaled by the current economy scale.
+ * @param t The current town.
+ * @param ct Type of cargo to generate, usually CT_PASSENGERS or CT_MAIL.
+ * @param amount The number of cargo units.
+ * @param stations Available stations for this house.
+ * @param affected_by_recession Is this cargo halved during recessions?
+ */
+static void TownGenerateCargo(Town *t, CargoID ct, uint amount, StationFinder &stations, bool affected_by_recession)
+{
+	if (amount == 0) return;
+
+	/* All production is halved during a recession (except for NewGRF-supplied town cargo). */
+	if (affected_by_recession && EconomyIsInRecession()) {
+		amount = (amount + 1) >> 1;
+	}
+
+	/* Scale by cargo scale setting. */
+	amount = ScaleByCargoScale(amount, true);
+
+	/* Actually generate cargo and update town statistics. */
+	uint moved = MoveGoodsToStation(ct, amount, SourceType::Town, t->index, stations.GetStations());
+	t->supplied[ct].new_max += amount;
+	t->supplied[ct].new_act += moved;
+}
+
+/**
+ * Tile callback function.
+ *
  * Tile callback function. Periodic tick handler for the tiles of a town.
  * @param tile been asked to do its stuff
  */
@@ -565,11 +593,8 @@ static void TileLoop_Town(TileIndex tile)
 			uint amt = GB(callback, 0, 8);
 			if (amt == 0) continue;
 
-			uint moved = MoveGoodsToStation(cargo, amt, SourceType::Town, t->index, stations.GetStations());
-
-			const CargoSpec *cs = CargoSpec::Get(cargo);
-			t->supplied[cs->Index()].new_max += amt;
-			t->supplied[cs->Index()].new_act += moved;
+			/* NewGRF-supplied town cargos are not affected by recessions. */
+			TownGenerateCargo(t, cargo, amt, stations, false);
 		}
 	} else {
 		switch (_settings_game.economy.town_cargogen_mode) {
@@ -577,18 +602,12 @@ static void TileLoop_Town(TileIndex tile)
 				/* Original (quadratic) cargo generation algorithm */
 				if (GB(r, 0, 8) < hs->population) {
 					uint amt = GB(r, 0, 8) / 8 + 1;
-
-					if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
-					t->supplied[CT_PASSENGERS].new_max += amt;
-					t->supplied[CT_PASSENGERS].new_act += MoveGoodsToStation(CT_PASSENGERS, amt, SourceType::Town, t->index, stations.GetStations());
+					TownGenerateCargo(t, CT_PASSENGERS, amt, stations, true);
 				}
 
 				if (GB(r, 8, 8) < hs->mail_generation) {
 					uint amt = GB(r, 8, 8) / 8 + 1;
-
-					if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
-					t->supplied[CT_MAIL].new_max += amt;
-					t->supplied[CT_MAIL].new_act += MoveGoodsToStation(CT_MAIL, amt, SourceType::Town, t->index, stations.GetStations());
+					TownGenerateCargo(t, CT_MAIL, amt, stations, true);
 				}
 				break;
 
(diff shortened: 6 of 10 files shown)


2TallTyler:
Why do the annotations say the function is unused? 🤔

JGRennison:
> Why do the annotations say the `ScaleByCargoScale` function is unused? thinking

You've put it in a header, but marked it as static, try inline instead.

src/economy_func.h
```diff
@@ -49,4 +51,47 @@ static inline bool EconomyIsInRecession()
...
+{
+	/* We might not need to do anything. */
+	if (_settings_game.economy.cargo_scale == 100) return num;
+
+	uint scaled;
+	if (_settings_game.economy.cargo_scale > 100) {
+		scaled = (num * 100) / _settings_game.economy.cargo_scale;
+	}
+	else {
+		scaled = (100 / _settings_game.economy.cargo_scale) * num;
```
ldpl: why are there two separate cases? it's the exact same formula but less precise, cargo_scale 51-100 will be exactly the same afaict.

LC-Zorg:
I like such changes so much... It's not controversial at all, they don't destroy anything, they don't impose anything, but they only give new possibilities. :)

I don't know how it will work in detail in combination with other sets, but in general I realy like this feature.
In my opinion, the percentage form is much more understandable than the one used in JGRPP. Maybe the only thing missing are separate settings for cities and industry. This can be good, for example, when you want to make a game based more on industry, or vice versa, on passenger transport. But a single setting isn't bad either.

src/industry_cmd.cpp
```diff
@@ -1148,6 +1148,25 @@ static void ChopLumberMillTrees(Industry *i)
...
+ * @param i The industry
+ * @param scale Should we scale production of this cargo based on the economy scale?
+ */
+static void ProduceIndustryGoodsHelper(Industry *i, bool scale)
+{
+	for (size_t j = 0; j < lengthof(i->produced_cargo_waiting); j++) {
+		int amount = i->production_rate[j];
+
+		/* We scale most cargos by economy scale, except when produced by an industry callback.
+		 * For those, the callback is called less frequently. */
```
TrueBrain: Seems this comment is not in the right place. I understand why it is here, but this will age poorly. Either it should go in the documentation of "scale", or at the caller, I think.

src/newgrf.cpp
```diff
@@ -6656,6 +6656,12 @@ bool GetGlobalVariable(byte param, uint32 *value, const GRFFile *grffile)
 			*value = _cur_year;
 			return true;
 
+		/* case 0x25: // GRFID of the grf that contains the corresponding Action3. Currently only supported for vehicles. */
```
TrueBrain: What is this comment doing here? I might be missing context, but feels a bit weird :)

src/object_cmd.cpp
```diff
@@ -685,7 +685,14 @@ static void TileLoop_Object(TileIndex tile)
 	/* Top town buildings generate 250, so the top HQ type makes 256. */
 	if (GB(r, 0, 8) < (256 / 4 / (6 - level))) {
 		uint amt = GB(r, 0, 8) / 8 / 4 + 1;
+
+		/* Production is halved during recessions. */
 		if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
+
+		/* Scale by economy scale setting. */
+		if (_settings_game.economy.cargo_scale != 100) {
```
TrueBrain: `ScaleByCargoScale` already deals with this check; no need to do it again :)

src/object_cmd.cpp
```diff
@@ -694,7 +701,14 @@ static void TileLoop_Object(TileIndex tile)
 	 * equations. */
 	if (GB(r, 8, 8) < (196 / 4 / (6 - level))) {
 		uint amt = GB(r, 8, 8) / 8 / 4 + 1;
+
+		/* Production is halved during recessions. */
 		if (EconomyIsInRecession()) amt = (amt + 1) >> 1;
+
+		/* Scale by economy scale setting. */
+		if (_settings_game.economy.cargo_scale != 100) {
```
TrueBrain: Same as above.

src/table/settings/economy_settings.ini
```diff
@@ -300,3 +300,16 @@ str      = STR_CONFIG_SETTING_INFRASTRUCTURE_MAINTENANCE
...
 cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.cargo_scale
+type     = SLE_UINT16
+flags    = SF_NO_NETWORK
+def      = 100
+min      = 15
+max      = 400
+interval = 25
```
TrueBrain: I always have a bit of issue with an interval that is contradicting the min/def. What I mean: def - n * interval != min. It is not wrong, and the system can handle it fine, but it just feels .. meh :P Do with this what-ever you want :)

src/table/settings/economy_settings.ini
```diff
@@ -300,3 +300,16 @@ str      = STR_CONFIG_SETTING_INFRASTRUCTURE_MAINTENANCE
 strhelp  = STR_CONFIG_SETTING_INFRASTRUCTURE_MAINTENANCE_HELPTEXT
 post_cb  = [](auto) { InvalidateWindowClassesData(WC_COMPANY_INFRASTRUCTURE); }
 cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.cargo_scale
+type     = SLE_UINT16
+flags    = SF_NO_NETWORK
+def      = 100
+min      = 15
```
TrueBrain: Why 15? Not 5? 1? :D

src/town_cmd.cpp
```diff
@@ -517,6 +517,50 @@ static void MakeTownHouseBigger(TileIndex tile)
...
+	/* All production is halved during a recession (except for NewGRF-supplied town cargo). */
+	if (affected_by_recession && EconomyIsInRecession()) {
+		cargo_amount = (cargo_amount + 1) >> 1;
+	}
+
+	/* If we're not producing any cargo, we can bail out now. */
+	if (cargo_amount == 0) return;
+
+	/* Maybe scale production by the chosen economy scale. */
+	if (_settings_game.economy.cargo_scale != 100) {
```
TrueBrain: Again, no need to check here too :)

src/town_cmd.cpp
```diff
@@ -517,6 +517,50 @@ static void MakeTownHouseBigger(TileIndex tile)
...
+		case TE_MAIL:
+			t->supplied[CT_MAIL].new_max += cargo_amount;
+			t->supplied[CT_MAIL].new_act += MoveGoodsToStation(CT_MAIL, cargo_amount, ST_TOWN, t->index, stations.GetStations());
+			break;
+
+		default:
+			MoveGoodsToStation(ct, cargo_amount, ST_TOWN, t->index, stations.GetStations());
+			break;
+		}
+}
```
TrueBrain, line 547: I don't actually understand where this code is coming from; it feels off.

First of, in the old code it doesn't check town_effect. So that is new. And doesn't seem to be part of this PR.
Second, `CT_PASSENGERS` and `CT_MAIL` update `supplied`. But the old code also does it for the other cargoes.

Can you elaborate a bit what is happening here exactly? Tnx :)

src/settings_type.h
```diff
@@ -532,6 +532,7 @@ struct EconomySettings {
 	uint16 town_noise_population[4];         ///< population to base decision on noise evaluation (@see town_council_tolerance)
 	bool   allow_town_level_crossings;       ///< towns are allowed to build level crossings
 	bool   infrastructure_maintenance;       ///< enable monthly maintenance fee for owner infrastructure
+	uint16 cargo_scale;                      ///< scale cargo production of towns and industries by this percentage.
```
TrueBrain: I never like variables that don't tell me their unit, especially for something like `scale`. Can I suggest we make it `cargo_scale_percentage` or something? Other ideas are welcome .. but `scale` suggests to me 0..1 :D

TrueBrain:
I don't actually have an opinion about this PR. Code-wise it looks good and clean.

So I am curious what users will think of it. And best way to find out .. is to just merge it :) Curious if other developers have a (strong) opinion one way or the other :)

src/newgrf.cpp
2TallTyler: It has nothing to do with this PR, just noting the gap and why it's not implemented here. I used it previously and JGR was kind enough to correct me. In your opinion: Keep it here, put in a separate `Doc:` PR, or cut it entirely?

src/table/settings/economy_settings.ini
2TallTyler: We need to have at least one industry production tick per economy month/minute, and inversely scale the production callback frequency by the cargo scale percentage. The lowest this setting can go without the production interval exceeding the length of a economy month/minute is 13...which seems even more arbitrary than 15. I was thinking of making it 25, but why restrict it? Thoughts?

src/table/settings/economy_settings.ini
2TallTyler: See my comment on the minimum value, but I guess I could just make this 1. I'll decide once we come to a conclusion on the next comment. 🙂

src/town_cmd.cpp
2TallTyler: Ah, this came from a JGRPP cherry-pick. I guess it supports different town effect tracking than vanilla.

2TallTyler (commented):
The main question I have for users and other developers is whether we want one setting to scale all cargo, or one each for town cargo (including HQ production) and industry cargo. JGRPP has them separated, and it would be easy for me to split them. On the other hand...more settings. 🙂

Andrew350:
> The main question I have for users and other developers is whether we want one setting to scale all cargo, or one each for town cargo (including HQ production) and industry cargo. JGRPP has them separated, and it would be easy for me to split them. On the other hand...more settings. 🙂

Disclaimer: I am not a dev or even a player who is likely to ever use this feature.

With that said, I can guarantee you that if you don't allow people to scale each option separately you will inevitably be confronted with complaints about not allowing enough control for different playstyles. And while I generally agree about the stance on not adding new settings, this feature seems major enough to overlook that and give the (more vocal) playerbase the option to finetune their settings how they see fit. This is, after all, aimed more at said vocal playerbase than the general audience, so I say just go for it. 

99% of players will never have a clue what these settings even mean anyways (just like the rest of the settings), so what's the harm in adding one more for the players who do understand it for such a (potential) major difference in gameplay? If it's not too complicated on your end (as you say it isn't), I suggest just making each option separate _now_ - if nothing else just to avoid the headache of having to justify _later_ why you didn't.

But that's just my 2 cents. 🙂

src/newgrf.cpp
TrueBrain: It is fine to keep it here, I just think it needs a bit more explaining why it is comments out; or better yet, implement it? I dunno .. I don't know what it is doing :) But having it like this in our code base is just weird :) The next person who walks into it, has nothing to trace back what it was about :D

src/town_cmd.cpp
```diff
@@ -517,6 +517,37 @@ static void MakeTownHouseBigger(TileIndex tile)
...
+
+	/* Scale by cargo scale setting. */
+	cargo_amount = ScaleByCargoScale(cargo_amount);
+
+	/* Actually generate cargo and update town statistics. */
+	uint moved = MoveGoodsToStation(ct, cargo_amount, ST_TOWN, t->index, stations.GetStations());
+
+	const CargoSpec *cs = CargoSpec::Get(ct);
+	t->supplied[cs->Index()].new_max += cargo_amount;
+	t->supplied[cs->Index()].new_act += moved;
```
TrueBrain: Just to confirm, I just don't know this stuff well enough: CargoSpec of CT_PASSENGERS returns the same index as CT_PASSENGERS?

From a coding-point-of-view, that is different in the old code and new code. Old code does `t->supplied[CT_PASSENGERS]`, and new code first looks it up via `CargoSpec` to use `cs->Index()`. I have zero clue if that has any impact.

TrueBrain:
> The main question I have for users and other developers is whether we want one setting to scale all cargo, or one each for town cargo (including HQ production) and industry cargo. JGRPP has them separated, and it would be easy for me to split them. On the other hand...more settings. 🙂

I guess the question back is: what playstyles are possible / not possible with splitting them. Although I understand where @Andrew350 is coming from, and not even disagreeing, I don't fancy adding more settings because we are afraid people will complain :) I mean, clearly it is in JGRPP, so they can just use JGRPP instead. I would rather have we tailor the vanilla game to the 99%, and that we pick something that makes sense to them; not the vocal minority.

But again, I don't understand any of this (never bothered with economy in OpenTTD), so the question back is: what playstyle does it unlock :) And, to stay within the definition of [vanilla's goal](https://github.com/OpenTTD/OpenTTD/blob/master/CONTRIBUTING.md#project-goals): can it be handled by NewGRFs instead?

src/table/settings/economy_settings.ini
TrueBrain: I am mostly checking if you did this deliberately or not :) Pick one, I am fine with what-ever. As long as it is a conscious decision, I am fine with it :)

src/town_cmd.cpp
2TallTyler: Yes, CT_PASSENGERS is the index. This code is copied from the implementation in the house production callback, whereas CT_PASSENGERS is just the hardcoded version.

2TallTyler:
Updated, fixed, rebased, etc., and I split town and industry scaling into two separate settings, since I agree that's a better choice for players — including me, probably.

> But again, I don't understand any of this (never bothered with economy in OpenTTD), so the question back is: what playstyle does it unlock :) And, to stay within the definition of [vanilla's goal](https://github.com/OpenTTD/OpenTTD/blob/master/CONTRIBUTING.md#project-goals): can it be handled by NewGRFs instead?

See the PR description for answers. 😄

src/industry_cmd.cpp
```diff
@@ -1137,6 +1137,27 @@ static void ChopLumberMillTrees(Industry *i)
...
+{
+	for (auto &p : i->produced) {
+		if (!IsValidCargoID(p.cargo)) continue;
+
+		uint16_t amount = p.rate;
+
+		/* Maybe scale production. */
+		if (amount != 0 && scale) {
+			amount = ScaleByCargoScale(amount, false);
+		}
```
PeterN: ScaleByCargoScale already shortcuts for `num == 0`...

So
```
uint16_t amount = scale ? ScaleByCargoScale(p.rate, false) : p.rate;
```
or
```
uint16_t amount = p.rate;
if (scale) amount = ScaleByCargoScale(amount, false);
```

would be fine?

src/newgrf.cpp
```diff
@@ -6636,6 +6636,16 @@ bool GetGlobalVariable(byte param, uint32_t *value, const GRFFile *grffile)
 			*value = static_cast<int32_t>(TimerGameCalendar::year);
 			return true;
 
+		/* case 0x25: // Reserved: For vehicles, this contains the GRFID of the grf that contains the corresponding Action3. */
```
PeterN: Pardon?

src/newgrf.cpp
```diff
@@ -6636,6 +6636,16 @@ bool GetGlobalVariable(byte param, uint32_t *value, const GRFFile *grffile)
...
+
+		case 0x26: // town cargo scale setting
+			*value = _settings_game.economy.town_cargo_scale_percentage;
+			return true;
+
+		case 0x27: // industry cargo scale setting
+			*value = _settings_game.economy.industry_cargo_scale_percentage;
+			return true;
```
PeterN: Why should this be exposed to NewGRFs? We are 'going behind their back' to change adjust production, but then let them know anyway?

src/settings_type.h
```diff
@@ -549,6 +549,8 @@ struct EconomySettings {
 	uint16_t town_noise_population[4];         ///< population to base decision on noise evaluation (@see town_council_tolerance)
 	bool   allow_town_level_crossings;       ///< towns are allowed to build level crossings
 	bool   infrastructure_maintenance;       ///< enable monthly maintenance fee for owner infrastructure
+	uint16_t town_cargo_scale_percentage;     ///< scale cargo production of towns by this percentage.
+	uint16_t industry_cargo_scale_percentage; ///< scale cargo production of industries by this percentage.
```
PeterN: Is `_percentage` necessary? Very long variable names... For interface scaling, which is a percentage, I went with just `gui_scale`.

src/table/settings/economy_settings.ini
PeterN: This should be documented, in which case it might be better as a constant somewhere else.

src/town_cmd.cpp
PeterN: That's a separate change and a separate commit (or even PR) then. But I don't see anything with TownEffect here so maybe it's removed already.

src/town_cmd.cpp
PeterN: There is no reason to look up the cargo spec and then use `cs->Index()`. `ct` is already that index.

src/town_cmd.cpp
```diff
@@ -516,6 +516,37 @@ static void MakeTownHouseBigger(TileIndex tile)
...
+ * Generate cargo for a house, scaled by the current economy scale.
+ * @param t The current town.
+ * @param ct Type of cargo to generate, usually CT_PASSENGERS or CT_MAIL.
+ * @param amount The number of cargo units.
+ * @param stations Available stations for this house.
+ * @param affected_by_recession Is this cargo halved during recessions?
+ */
+static void TownGenerateCargo(Town *t, CargoID ct, uint amount, StationFinder &stations, bool affected_by_recession)
+{
+	uint cargo_amount = amount;
```
PeterN: Why this assignment? You can just name the function parameter `cargo_amount`, or use `amount` through the function.

src/industry_cmd.cpp
```diff
@@ -1156,15 +1177,20 @@ static void ProduceIndustryGoods(Industry *i)
...
-	/* produce some cargo */
+	/* If using an industry callback, scale the callback interval by cargo scale percentage. */
+	if (HasBit(indsp->callback_mask, CBM_IND_PRODUCTION_256_TICKS)) {
+		if (i->counter % ScaleByInverseCargoScale(Ticks::INDUSTRY_PRODUCE_TICKS, false) == 0) {
+			IndustryProductionCallback(i, 1);
+			ProduceIndustryGoodsHelper(i, false);
+		}
+	}
+
+	/* All other production and special effects happen every 256 ticks, and cargo production is just scaled by the cargo scale percentage. */
```
PeterN: Why two different methods of scaling anyway?

src/industry_cmd.cpp
```diff
@@ -1826,7 +1852,7 @@ static void DoCreateNewIndustry(Industry *i, TileIndex tile, IndustryType type,
 		}
 
 		for (auto &p : i->produced) {
-			p.history[LAST_MONTH].production += p.rate * 8;
+			p.history[LAST_MONTH].production += ScaleByCargoScale(p.rate * 8, false);
 		}
```
PeterN, line 1854: Not actually this line. `ChopLumberMillTrees()` appears to be unaffected by any scaling as it's still executed on the fixed tick interval, and the production algorithm is not changed. The constant `45` should be scaled?

src/newgrf.cpp
frosch123: I agree, hide it from NewGRF.
Also: settings should go into GetPatchVariable(). GetGlobalVariable() is for dynamic values.

src/industry_cmd.cpp
2TallTyler: We could scale the interval for all cargos, but that makes the production quite jumpy at low scales -- vehicles sit empty for a long time waiting for the first production ticks. It annoys me in the JGRPP implementation of daylength, and I'd rather avoid it if possible.

src/industry_cmd.cpp
2TallTyler: Yes, good catch. I agree it's best to scale the produced amount, while leaving the rate of tree-cutting alone.

src/town_cmd.cpp
2TallTyler: I thought it was forbidden/frowned upon for a function to modify the arguments passed to it. Did I imagine this, or is it language-specific, perhaps?

src/newgrf.cpp
JGRennison: As an aside, putting it into GetPatchVariable would mean that you would not be able to modify the value of the settings during multiplayer games at all, because of the use of Action D.
There is otherwise no reason for the added settings to have the SF_NO_NETWORK flag.
Changing the value during a single-player game would result in it not being visible to GRFs until the next save/load or GRF reload.
This would not be ideal, given that it's quite typical that you'll want to adjust the scaling as the game develops (usually, reduce it further as the network size increases).

If these variables were exposed to GRFs, GetGlobalVariable would be more useful than GetPatchVariable.

src/town_cmd.cpp
JGRennison: amount is a value type not a reference type here, so there is nothing wrong with modifying it in the function body.
There is no effect on the caller scope.

src/newgrf.cpp
2TallTyler: I decided to not expose it to GRF. The original intention was to allow industry sets like FIRS to provide accurate helptext, but I agree that the potential for misuse might outweigh this.

PeterN:
Does this distinguish between primary and secondary industries?

If output of primary industries is reduced, then logically input of secondary industries is reduced, so already their output will lower, and then it is reduced further?

2TallTyler:
Yes. The modified scale-the-amount code only affects primary industry production, while secondary industries are not affected. NewGRF industries like ECS or Industries of the Caribbean that use stockpiling and the every-256-ticks callback produce less often, so it takes them longer to work through the stockpile. As the designer of one of these (Caribbean), that is how I would intend it to work, since the production rate is balanced against primary industry production.

Also, as @JGRennison mentioned on Discord, a similar implementation (I started with cherry-picking that commit) has been in JGRPP for several years. There, it's used to counteract the effects of daylength, typically turned up to compensate for everything else running more slowly. 😃

PeterN:
> Yes. The modified scale-the-amount code only affects primary industry production,

Okay... because the section where secondary production occurs is elsewhere and not touched by this change. Makes sense.

> Also, as @JGRennison mentioned on Discord, a similar implementation (I started with cherry-picking that commit) has been in JGRPP for several years. There, it's used to counteract the effects of daylength, typically turned up to compensate for everything else running more slowly. 😃

Lovely, but experience of a change in another build that many of us have not used doesn't mean much when it comes to code reviews.

2TallTyler:
> Lovely, but experience of a change in another build that many of us have not used doesn't mean much when it comes to code reviews.

Oh, agreed. I don't say that as a "gotcha," but I figured it was worth noting in case anyone wanted to see how it works over there. 🙂

2TallTyler:
@nielsmh As we discussed on Discord, I'll be changing the control of calendar progress speed in #11428 to "minutes per game year" units. What are your thoughts on the units for this pair of settings? Make them match calendar speed somehow, or keep them as percentages?

(Bear in mind that industry production cannot be slower than about 13% of vanilla, or some months will have no production with a callback-based industry set, leading to all sorts of usability problems. This could be fixed with averaging or something, but that's probably out of scope for now...)

nielsmh:
I think actually entering minutes for direct entry and up-down buttons makes the most sense. However, the setting itself should ideally be displayed with a unit conversion too:
- Minutes per calendar year: **12 minutes** (standard game speed)
- Minutes per calendar year: **6 minutes** (speed-up, 200% game speed)
- Minutes per calendar year: **60 minutes** (slow-down, 20% game speed)
- Minutes per calendar year: **1,440 minutes** (slow-down, 0.83% game speed)

The last one is perhaps a bit extreme, but ideally I think printing a percentage with two significant digits, even if those need to be decimal, is preferable.

As for production scaling, can it perhaps be done in the same way? "Minutes per production year"?

2TallTyler:
nodl-speed

My current build of #11428 includes a percentage, making it easy to match this percentage setting. Especially because it's limited to 13% minimum -- there's not much precision we can get here. "Minutes per production year" doesn't make much sense to me, since we don't talk about economy years anywhere in the game (by design).

src/economy_func.h
```diff
@@ -49,4 +51,50 @@ inline bool EconomyIsInRecession()
...
+{
+	uint16_t percentage = (town ? _settings_game.economy.town_cargo_scale : _settings_game.economy.industry_cargo_scale);
+
+	/* We might not need to do anything. */
+	if (percentage == 100) return num;
+
+	uint scaled = (num * 100) / percentage;
+
+	/* Never return 0, since we often divide by this number. */
+	return (scaled > 1 ? scaled : 1);
```
glx22:
```suggestion
	return max(scaled, 1);
```

michicc:
I'm not familiar with the JGR implementation, are there any additions a NewGRF could query (like some global Act D or Act 2 variable)?

I have no idea if any NewGRF does this, but in theory they could read back the produced cargo, which with this scaling is not what the NewGRF might expect. Very theoretically I guess, I'm just wondering if this scaling factor is something that a NewGRF should be able to read.

JGRennison:
> I'm not familiar with the JGR implementation, are there any additions a NewGRF could query (like some global Act D or Act 2 variable)?
> 
> I have no idea if any NewGRF does this, but in theory they could read back the produced cargo, which with this scaling is not what the NewGRF might expect. Very theoretically I guess, I'm just wondering if this scaling factor is something that a NewGRF should be able to read.

In the implementation in my branch, there are deliberately not any such variables. The scaling is intended to be as transparent as practical from the point of view of the GRF. This is to make it more likely that the scaling works correctly, and as a secondary factor, to discourage GRF attempts to undo it or apply the scaling a second time, or to refuse to work if scaling is detected.

Past month production/transported values are used by GRFs such as BSPI for stockpiling mechanics. The actual values are what is presented back to the GRF.
For example, in the case of BSPI a higher production rate leads to the stockpile being depleted faster, due to the higher rate at which cargo is transported, which is a reasonable mechanic.

michicc (approved):
Plague

PR #10608, opened 2023-04-07
Codechange: Add a property to graph windows for whether to draw dates

2TallTyler:
## Motivation / Problem

Using `this->month == 0xFF` to decide whether to draw dates below a graph window is strange.

## Description

Let's use a property instead.

Planning ahead for #10605.

## Checklist for review

### DIFF
--- a/src/graph_gui.cpp
+++ b/src/graph_gui.cpp
@@ -181,10 +181,11 @@ protected:
 	byte num_on_x_axis;
 	byte num_vert_lines;
 
-	/* The starting month and year that values are plotted against. If month is
-	 * 0xFF, use x_values_start and x_values_increment below instead. */
+	/* The starting month and year that values are plotted against. */
 	byte month;
 	Year year;
+
+	bool draw_dates = true; ///< Should we draw months and years on the time axis?
 
 	/* These values are used if the graph is being plotted against values
 	 * rather than the dates specified by month and year. */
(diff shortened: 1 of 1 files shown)


src/graph_gui.cpp
```diff
@@ -186,6 +186,8 @@ struct BaseGraphWindow : Window {
 	byte month;
 	Year year;
 
+	bool draw_dates; ///< Should we draw months and years on the time axis?
```
stormcone: Maybe I missed something, but where is this variable set to true? I only found line 887 where it is set to false.
And shouldn't lines 502 and 527 also be updated to use `draw_dates`?

2TallTyler: You didn't miss anything, I missed a few things while cherry-picking. Thanks for the heads-up.

src/graph_gui.cpp
```diff
@@ -622,6 +623,8 @@ struct OperatingProfitGraphWindow : BaseGraphWindow {
 	OperatingProfitGraphWindow(WindowDesc *desc, WindowNumber window_number) :
 			BaseGraphWindow(desc, WID_CV_GRAPH, STR_JUST_CURRENCY_SHORT)
 	{
+		this->draw_dates = true;
```
TrueBrain: What if we initialize `draw_dates` to `true` by default? Now it kinda gives the suggestion it can remain uninitialized, which is a bit weird to read.

But that makes me wonder what is going on with the other variables. Only to find out they too aren't explicitly set; just in some cases. Which all feels a bit ... odd, to me. But that is not your code.

Can't remember where OpenTTD stands on C++ initializers, but I personally really prefer to have them set explicitly in the definitions when possible. So `bool draw_dates = true`. And set it to false for the exception.

Curious what others think.

2TallTyler: Sure, makes sense to me. Should we continue explicitly set it to true for the windows which use it, just for visibility, or let it use the default?

TrueBrain: The latter, I would say. No need to set things again if they are already set.

Mainly what bothers me, is that some variables are set (like `draw_dates` for many Windows, but others only for one window.

Feels weird .. haven't bothered to actually look at the code what is going on, I have to admit. But feels like someone was a bad boy years ago, and we are now suffering :D

TrueBrain (approved):

PR #10609, opened 2023-04-08
Change: Use seconds for AI start delay

2TallTyler:
## Motivation / Problem

I was asked to break #10605 into smaller PRs for review. Most of the optional real-time units in it are indeed optional, but settings stored in days are annoying since there isn't a real-time unit which corresponds to it for a simple string swap -- the number itself would have to be converted (1 day = 2 seconds).

AI start delay doesn't need to be in days as it's a player setting to determine how much head start they get, and seconds almost make more sense.

## Description

Use seconds for AI start delay.

## Limitations

"Seconds" is 2 seconds per day, so it relies on https://github.com/OpenTTD/OpenTTD/pull/10607 for complete accuracy, doesn't count when the game is paused, and goes faster when in fast-forward. Probably not a big deal, as that's what I'd expect from any game using time as a unit.

## Checklist for review


TrueBrain:
Thanks a lot for splitting up the PR; much easier to review like that :D

I have no code-wise comments, just a practical one: should we convert the settings when loading the first time after this PR? I believe we have a version-number in the config file, and I believe there is already some boilerplate to do these kind of things .. might be nice? Otherwise people all of a sudden see their AIs start a lot faster than they expect?

2TallTyler:
Yes, although I'll need some help figuring out how to change AI settings without crashing while starting the game. 🙂 The latest commit shows what I tried.

2TallTyler:
As discussed on Discord, setting conversion will wait for you or @glx22 to change the start date from a per-AI config setting to a global setting. Then we can merge that PR and this one at the same time to avoid having to do any setting conversion, since nobody will have a chance to generate a config file with the old values. 😄

LC-Zorg:
This setting by seconds will be visible when the player selects a pace other than 100%, or it will be always like this?

2TallTyler:
Superseded by #10653.

PR #10610, opened 2023-04-08
Change: Use seconds for Linkgraph update settings

2TallTyler:
## Motivation / Problem

I was asked to break https://github.com/OpenTTD/OpenTTD/pull/10605 into smaller PRs for review. Most of the optional real-time units in it are indeed optional, but settings stored in days are annoying since there isn't a real-time unit which corresponds to it for a simple string swap -- the number itself would have to be converted (1 day = 2 seconds).

Linkgraph update settings don't need to be in days, and seconds almost make more sense.

## Description

Use seconds for linkgraph update interval and recalculation time. I changed the string names to void the old translations and added config file and game settings conversions since the units changed and 1 day = 2 seconds.

## Limitations

"Seconds" is 2 seconds per day, so it relies on https://github.com/OpenTTD/OpenTTD/pull/10607 for complete accuracy, doesn't count when the game is paused, and goes faster when in fast-forward. Probably not a big deal, as that's what I'd expect from any game using time as a unit.

## Checklist for review


src/settings.cpp
```diff
@@ -154,11 +154,12 @@ class ConfigIniFile : public IniFile {
  * location. These versions assist with situations like that.
  */
 enum IniFileVersion : uint32 {
-	IFV_0,               ///< 0  All versions prior to introduction.
-	IFV_PRIVATE_SECRETS, ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
-	IFV_GAME_TYPE,       ///< 2  PR#9515  Convert server_advertise to server_game_type.
+	IFV_0,                 ///< 0  All versions prior to introduction.
+	IFV_PRIVATE_SECRETS,   ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
+	IFV_GAME_TYPE,         ///< 2  PR#9515  Convert server_advertise to server_game_type.
+	IFV_LINKGRAPH_SECONDS, ///< 3  PR#10610 Use seconds for linkgraph update settings.
```
TrueBrain: My OCD hits. Please sync descriptions between settings and savegame. So `Store linkgraph update intervals in seconds instead of days.`

src/table/settings/linkgraph_settings.ini
```diff
@@ -37,26 +37,26 @@ startup  = false
...
 from     = SLV_183
-def      = 4
-min      = 2
-max      = 32
-interval = 2
-str      = STR_CONFIG_SETTING_LINKGRAPH_INTERVAL
+def      = 8
+min      = 4
+max      = 64
+interval = 4
```
TrueBrain: I suggest either intervals of 1s, or 5s, and a min/max that matches that. But 64 max is just weird in the next context.

src/table/settings/linkgraph_settings.ini
```diff
@@ -37,26 +37,26 @@ startup  = false
...
 
 [SDT_VAR]
 var      = linkgraph.recalc_time
 type     = SLE_UINT16
 from     = SLV_183
-def      = 16
+def      = 32
 min      = 1
-max      = 4096
+max      = 8192
```
TrueBrain: I just don't even .. what a large number :P But okay .. please also max this something more sane, like `10000` or something.

TrueBrain (commented):
For many reasons, this doesn't fit with the new timer stuff I have been working on.

LinkGraph is just .. complex. It has many assumptions on what it is doing, how busy it is, when the start the next job, etc, that it is hard to convert to something interval/timer based.

So for now, I say, let's skip this, and keep this PR as it is. I would like to sit down for it and make it a bit more understandable, but for now, this will do fine for me.

The one thing that does trouble me, is that in fast-forward the term "second" means nothing. And that keeps bringing me back to the core of my issue with link-graph: although I understand the scheduling has to be deterministic, it doesn't actually make sense. But a solution is rather complex.

So, let's go with this for now :)

src/table/settings/linkgraph_settings.ini
2TallTyler: It's now 9,000, which is 150 minutes or 100 of the maximum linkgraph recalc intervals. Hope that's even enough for you. 😉

TrueBrain (approved):

PR #10613, opened 2023-04-08
Codechange: Refactor timetable GUI

2TallTyler:
## Motivation / Problem

The timetable GUI needs some love, especially before I extend it in #10605.

## Description

* Replace magic numbers with a constant
* Better tracking of why we opened a query text input
* Un-duplicated text colour and scheduled/expected offsets
* Separated a lengthy DrawWidget() switch into separate helper functions for each widget.

## Checklist for review

### DIFF
--- a/src/timetable_gui.cpp
+++ b/src/timetable_gui.cpp
@@ -23,6 +23,7 @@
 #include "vehicle_gui.h"
 #include "settings_type.h"
 #include "timetable_cmd.h"
+#include "timetable.h"
 #include <cstdint>
 
 #include "widgets/timetable_widget.h"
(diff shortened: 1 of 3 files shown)


src/timetable_gui.cpp
```diff
@@ -438,32 +437,36 @@ struct TimetableWindow : Window {
...
+								/* Show expected arrival. */
+								this_offset = 0;
+								colour = TC_GREEN;
 							} else {
-								SetDParam(0, _date + (arr_dep[i / 2].arrival + offset) / DAY_TICKS);
-								DrawString(time.left, time.right, tr.top, STR_JUST_DATE_TINY,
-										show_late ? TC_RED : i == selected ? TC_WHITE : TC_BLACK);
+								/* Show scheduled arrival. */
+								this_offset = offset;
+								colour = show_late ? TC_RED : (i == selected ? TC_WHITE : TC_BLACK);
```
glx22: If I follow correctly, with separated abbreviation the selected state was always visible on the abbreviated part.
Now if the string is green you'll never see the selected state, same for late as the string will be full red.

2TallTyler: Ah, yes, that's correct. I might have to return to drawing the abbreviation as a separate string, as I don't know a way to draw two colours ({PUSH/POP_COLOUR} only works with colour string codes provided in the string, I think).

LordAro: Mildly unrelated, but since you're in the area, could do with splitting this switch into multiple functions. The indentation level in this function is getting a bit extreme

2TallTyler: Indeed, I'll see what I can do. 🙂

2TallTyler:
As requested, I went back to separated abbreviation to keep the selected order white, and I broke up the switch into separate functions for each widget.

src/timetable_gui.cpp
glx22: > Ah, yes, that's correct. I might have to return to drawing the abbreviation as a separate string, as I don't know a way to draw two colours ({PUSH/POP_COLOUR} only works with colour string codes provided in the string, I think).

It might be possible to introduce a new `{COLOUR}` command taking a parameter in the `SCC_BLUE`-`SCC_BLACK` range (or 0-16 range which might be better for GSText).
I see exactly how to do it, so I may try.

Edit: it's done on https://github.com/OpenTTD/OpenTTD/compare/master...glx22:OpenTTD:colour

LordAro (dismissed):
LGTM, but needs a rebase :)

LordAro (approved):

PR #10614, opened 2023-04-08
WIP: scale out TileLoop, to speed up bigger maps

TrueBrain:
## Motivation / Problem

I was bored, and wanted to do something fun.

This idea was living in my head for a while now, and when JGRennison showed me he already implemented a worker pool, it just had to be tried.

Mostly I was curious if the TileLoop can be made faster by using a worker-pool. The answer is a very clear and loud: yes.
The next, much harder, assignment will be: make all TileLoops thread-safe. They aren't. So yeah .. that is fun.
But if we go this route, it does speed up the game, for those who have more than 1 core (read: according to Steam survey, 99% of our users), especially on bigger maps.

This was btw the trigger to look into this: Steam says almost nobody has a single-core CPU anymore. So we don't have to worry that much anymore what the performance loss is on those machines; only what the gain is on multi-core CPUs :) Which is a good thing!

## Description

This is a proof-of-concept, to show that something like this will help, but requires some work on the current TileLoop implementations.

On a 4kx4k map the TileLoop goes from 17ms to 3ms on my machine, and a 4kx4k map finally becomes playable (in Debug-mode).
For Release-mode it is from 5ms to 1.5ms.

Currently it queues 1024 tile-updates on a single worker, in an attempt to balance how much we push in a std::queue and how long a worker takes to execute the job. This number is not tuned, but just a number that looked nice.

With this change, it means a TileLoop should only deal with its local Tile, and nothing else. 100% sure this currently isn't always the case, so expect issues when using this.

The main one that kept crashing was the updating of the towns coordinates. But the problem below that is of course that multiple threads updated how big the town is (as that updates the coordinates). So the current fix is .. not a fix, but a workaround.

`worker_thread` is a copy from https://github.com/JGRennison/OpenTTD-patches/blob/b776dcbd173fd09bfa1c9904a0606d4709cdb49b/src/worker_thread.cpp, with small changes (changed the function signature and there is always a worker that will be notified when something got inserted in the queue). Of course this is a dirty solution, and we should go for a more generic approach, but again: proof-of-concept. Tnx JGRennison for the work here :)

Owh, and of course: a picture, or it didn't happen.

Badly tuned batch-size:
image

Better tuned batch-size:
image

Als, as it turns out, release-builds handle a batch of 256 so quickly, that the workers had hardly anything to do. This will be one of the challenges to get such PR just right. As the batch-size will differ per CPU.
- Going too big? There is no scaling out.
- Going too small? The overhead is higher than not threading.
- There is a sweet-spot in between.

## Limitations

SO MANY! Where to begin? Just read the description, okay?

## Checklist for review

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -522,4 +522,6 @@ add_files(
     window_type.h
     zoom_func.h
     zoom_type.h
+    worker_thread.cpp
+    worker_thread.h
 )
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -34,6 +34,7 @@
 #include "landscape_cmd.h"
 #include "terraform_cmd.h"
 #include "station_func.h"
+#include "worker_thread.h"
 #include <array>
 #include <list>
 #include <set>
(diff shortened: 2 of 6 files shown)


TheMowgliMan:
I would like to see this merged.

The issue is, previous attempts to add multithreading we're rejected.

TrueBrain:
And I would like a mansion! But in case the description wasn't clear enough, but I hope it was: this is YEARS away from being finished. Not days, not weeks, not even months. This is so far WIP, that WIP needed to redefine itself to include this amount of WIP.

So don't get your hopes up.

And no, "previous attempts" weren't rejected. They didn't work. Which is something completely different. Very similar to this PR: it doesn't actually work.

TrueBrain:
CLang Thread Sanitizer reports at least 20 places where there is a "data race" (where two threads are accessing the same memory). 

Some are easy (dirty-blocks) and can be solved with a lock, others are .. interesting.

Like, `TileLoop_Trees` can plant trees on nearby tiles. This .. is problematic. Possibly we can resolve that by making an "action-to-take" queue, where the TileLoop can insert functions to execute when all TileLoops are done.

Yeah .. work to be done here :)

JGRennison:
The smoke plumes above animated vanilla industries are visually broken, and they eventually triggered a crash.
These use effect vehicles, so create entries in the global vehicle pool.

Industries in general can trigger all sort of stuff which reads and writes non tile local state from their tile loop handlers.
Possibly the whole industry tile loop handler should be queued.

ldpl:
In theory it can be solved by having two tile loop modes, concurrent and general. In concurrent mode tile can request a general run that is not parallelized. Though not sure what is the ratio of tiles that can be run fully concurrent and so if it's worth the overhead.

TrueBrain:
Going to close up this PR for now; much work ahead!

PR #9852, opened 2022-04-13
Fix #9810: 'Rebuilding' a through road stop costs money.

wholepuncher:
## Motivation / Problem

Issue #9810 and comments: Overbuilding of road stops, rail stations, and rail waypoints on pre-existing buildings of same type charge money.

## Description

For road stops:
Tiles that already have a road stop do not contribute to the build cost anymore.
Non-through road stops can be rotated by overbuilding for free.
Through road stops cannot be rotated by overbuilding since there is already an error if the underlying road is in the wrong orientation.
If no new road stops are built or rotated, the command returns  a STR_ERROR_ALREADY_BUILT error.

For rail stations:
Tiles that already have a rail station of the same orientation do not contribute to the build cost anymore, even if the station gfx is changed by the command.
Rail stations cost money to rotate since they rebuild the underlying rail as well.
If no new rail stations are built or rotated, the command returns  a STR_ERROR_ALREADY_BUILT error.

For rail waypoints:
If there already is a rail waypoint, the command returns a STR_ERROR_ALREADY_BUILT error.

These changes mean overbuilding either errors or charges only for significant changes and not for pre-existing stations.

Closes #9810 - edit 2TT

## Limitations

I decided to use STR_ERROR_ALREADY_BUILT as the error, but as seen in the comments of issue #9810, there's still a bit of inconsistency with other overbuilding building errors.
The cost calculations for road stops and rail stations use two new entire functions, one for each type.
The costs for all changed building are calculated before placements. No actual placement updating code was changed. This means the actual tile changes still happen as before.
The waypoint building interface seems to allow areas of selection, but in-game only allows single tile placements. The code change accounts for areas, but effectively only errors for waypoint overbuilding.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10618, opened 2023-04-09
Change: Allow overbuilding station and waypoint tiles

2TallTyler:
## Motivation / Problem

#9852 made overbuilding free, but blocked overbuilding without changing any tiles. This is very common when building stations using NewGRF station sets, where platforms can have different canopies, footbridges, etc., so having it blocked is frustrating.

## Description

It doesn't hurt anybody to allow overbuilding, even if graphics are the same, so I just removed the error entirely. Costs are still calculated correctly, which was the point of #9852.

Rotating road stations is still only possible for bay stations, since drive-through stations will throw an error about the underlying road.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -1261,7 +1261,6 @@ static void RestoreTrainReservation(Train *v)
 static CommandCost CalculateRailStationCost(TileArea tile_area, DoCommandFlag flags, Axis axis, StationID *station, RailType rt, std::vector<Train *> &affected_vehicles, StationClassID spec_class, byte spec_index, byte plat_len, byte numtracks)
 {
 	CommandCost cost(EXPENSES_CONSTRUCTION);
-	bool success = false;
 	bool length_price_ready = true;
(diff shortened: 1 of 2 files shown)


FLHerne:
It's explicitly intended by the author of #9852, so the commit message should perhaps be 'Change' rather than 'Fix'. I suspect it may have been overlooked by reviewers though.

> If no new rail stations are built or rotated, the command returns a STR_ERROR_ALREADY_BUILT error.
> If there already is a rail waypoint, the command returns a STR_ERROR_ALREADY_BUILT error.

I agree that allowing overbuilding is preferable, the current behaviour would be very inconvenient for building NewGRF stations.

2TallTyler:
Ah, good point. Commit title changed.

This will need backporting to 13.1 or we'll get tons of bug reports if we release without it. 😉

glx22:
No need to backport, the original change is only in master.

rubidium42 (approved):

PR #10622, opened 2023-04-10
Fix: Abort loading savegame if road vehicle is on invalid road type.

PeterN:
## Motivation / Problem

When loading savegames with missing NewGRFs, critical engine and road type information can be missing. This can result in road vehicles existing on tram tiles, which then causes an assertion.

## Description

This change detects this case during loading and aborts loading the game in road vehicles are detected on the wrong road type.

Because savegames do not store road type information about road types, vehicles, or engines, this checks if the tile has a valid road type for each road vehicle. In some cases it may pass the test, but in that case it should at least not crash (but would still be unplayable.)

## Limitations

The error message shown is only the first error, which will be about missing NewGRFs rather than the invalid road type.

## Checklist for review

### DIFF
--- a/src/saveload/vehicle_sl.cpp
+++ b/src/saveload/vehicle_sl.cpp
@@ -401,9 +401,11 @@ void AfterLoadVehicles(bool part_of_load)
 
 					rv->roadtype = Engine::Get(rv->engine_type)->u.road.roadtype;
 					rv->compatible_roadtypes = GetRoadTypeInfo(rv->roadtype)->powered_roadtypes;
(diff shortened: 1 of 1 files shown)


James103:
If a savegame was loaded in such a way so as to have tram vehicles on road tiles, would the error handling added here catch that or is an extra check needed?

PeterN:
Yes, it checks for the tile's road or tram type depending on the `RoadTramType` definition of the engine's roadtype.

rubidium42 (approved):

Issue #9609, opened 2021-10-11
[Bug]: Pathfinder prefers occupied plain track over free station track


baysupercool:
### Version of OpenTTD

1.11.2, Windows 10

### Expected result

Train picks the path through the station

### Actual result

Train always picks the passing track, and if there is already another train in the passing track, it does not pick the other track through the station. Depending on the situation this can result in both trains to become stuck forever.

### Steps to reproduce

![bug](https://user-images.githubusercontent.com/46246461/136751428-8f890d98-aa60-40ce-b3a6-951dd85a81f5.png)

1. Copy the shown track schematic. Red tiles are stations. Signals in front of the stations are one way path signals. The rest are two way path signals.
2. Buy two trains and set their orders to go non-stop to both end stations. The trains should not stop at the middle station.
3. Send the trains to the track.
4. Eventually trains get stuck facing each other. I can consistently replicate this issue.

LordAro:
Can you provide a savegame exhibiting this behaviour?

LordAro:
Went ahead and did it myself anyway. I wonder if this might be pathfinder-caching related...

![image](https://user-images.githubusercontent.com/4007992/136755864-e9aa4b23-d785-43b2-aa3d-a6607e99090f.png)

James103:
Here's my reproduction attempt:
![image](https://user-images.githubusercontent.com/37945304/136756903-70a48aab-0899-45e0-9f5e-9166b012146e.png)
Save game: [#9609 testing.zip](https://github.com/OpenTTD/OpenTTD/files/7320732/9609.testing.zip)

glx22:
Wondering if #9554 could have a similar cause.

frosch123:
The relevant pathfinder penalties are: (default values)
* pf.yapf.rail_pbs_signal_back_penalty = 15 * YAPF_TILE_LENGTH (once per signal)
* pf.yapf.rail_station_penalty = 10 * YAPF_TILE_LENGTH (per station tile)

So, currently it will prefer waiting for the reverse-pbs signal, if the station platforms are longer than 1 tile.

I think the problem here is the station penalty:
* The penalty is currently per tile, because this is part of the "pathfinder prefers platform matching train length".
* But this makes no sense, if the train passes through anyway.

So, I think there should be two different station penalties:
* A penalty per station tile, if the train will stop at the station.
* A penalty independent of platform length, if the train will not stop at the station.

glx22:
[...]

PR #10623, opened 2023-04-10
Remove: Settings for YAPF

LordAro:
## Motivation / Problem

Settings for NPF & YAPF were largely added in the beginning to help work out good defaults. That was A Long Time Ago, and we're pretty happy with the settings as they are now and are pretty confident that they should never be changed by users as it'll just cause weirdnesses and confusing behaviour.

Indeed, if you change some of them they'll crash the game due to poorly set limits.

## Description

Removes all 18 NPF settings and all 34 YAPF settings turning them into compile time constants. Largely just a straight swap of setting to constant, but a couple of tweaks to improve code where it presented itself as well.

Additionally removes a couple of similarly ancient settings for setting which PF to use.

## Limitations

Probably controversial.

Possibly some changes in behaviour to do with how maximum depot distance is handled, not sure.

## Checklist for review


src/saveload/compat/settings_sl_compat.h
```diff
@@ -95,10 +95,8 @@ const SaveLoadCompat _settings_sl_compat[] = {
 	SLC_VAR("vehicle.smoke_amount"),
 	SLC_NULL(1, SL_MIN_VERSION, SLV_159),
 	SLC_VAR("pf.roadveh_queue"),
-	SLC_VAR("pf.new_pathfinding_all"),
-	SLC_VAR("pf.yapf.ship_use_yapf"),
-	SLC_VAR("pf.yapf.road_use_yapf"),
-	SLC_VAR("pf.yapf.rail_use_yapf"),
+	SLC_NULL(1, SL_MIN_VERSION, SLV_87), // pf.new_pathfinding_all
+	SLC_NULL(3, SLV_28, SLV_87), // pf.yapf.{ship,road,rail}_use_yapf
```
TrueBrain: why SLV_87 here, and not SLV_REMOVE_PF_SETTINGS?

src/pathfinder/yapf/yapf_base.hpp
```diff
@@ -56,10 +59,10 @@ class CYapfBaseT {
 
 	NodeList             m_nodes;              ///< node list multi-container
 protected:
+	static const int MAX_SEARCH_NODES = 10000; ///< stop path-finding when this number of nodes visited
```
PeterN, line 62: Seems odd to stick this constant here when the others are defined at the top.

src/table/settings/pathfinding_settings.ini
```diff
@@ -49,33 +49,6 @@ var      = pf.roadveh_queue
...
-to       = SLV_87
-def      = true
-cat      = SC_EXPERT
-
-[SDT_BOOL]
-var      = pf.yapf.rail_use_yapf
-from     = SLV_28
-to       = SLV_87
-def      = true
-cat      = SC_EXPERT
```
TrueBrain: Why did you remove these? If I read it right, they are just to convert old settings into new settings .. but both PFs still exist. So I wouldn't expect these to be removed?

Basically now for older savegames, when someone said they didn't want YAPF, after this commit, they get YAPF anyway.

src/pathfinder/yapf/yapf_base.hpp
LordAro: I forgot to move the ship constants :)

But this constant is here (inside the class) in a vague attempt to hide it from the global scope where it's not needed (the other constants (except rail) are in cpp files)

src/table/settings/pathfinding_settings.ini
LordAro: Mostly laziness - they're extremely old compatibility settings only (removed in 2008-era), and they were getting in the way of removing the `YapfSettings` data structure. Saves still relying on it will "just" get reset to YAPF.

Sure, technically a regression, but not a significant one IMO

TrueBrain: We take so much effort in keeping old savegames as compatible as we can .. I don't feel this is the right call. It is a very small effort on our part to fix this :)

src/saveload/saveload.h
```diff
@@ -349,6 +349,7 @@ enum SaveLoadVersion : uint16 {
 
 	SLV_VELOCITY_NAUTICAL,                  ///< 305  PR#10594 Separation of land and nautical velocity (knots!)
 	SLV_CONSISTENT_PARTIAL_Z,               ///< 306  PR#10570 Conversion from an inconsistent partial Z calculation for slopes, to one that is (more) consistent.
+	SLV_REMOVE_PF_SETTINGS,                 ///< 307  PR#??? Removal of pathfinder settings
```
TrueBrain: Don't forget to update these `???` :)

src/table/settings/pathfinding_settings.ini
LordAro: I've restored the SL for the variable as well, I think that was the right thing to do?

TrueBrain (dismissed):

ldpl:
I'll just leave here what I already said on discord. 
You allowed settings to be configurable so that players can work out good defaults and now that players settled on different defaults you're just removing that.

Thankfully `rail_firstred_twoway_eol` is on by default now so that mitigates most of the damage and in that regard is even a positive change to wipe out old configs.
But there are other settings that can be useful for certain playstyles:

`max_search_nodes`. 10k nodes is not that much, can easily be reached in the game large towns with grid road layouts. And it's not impossible to reach that with trains, especially if 90 degree turns are on (which still are by default). I've personally remember helping two players to continue their games by increasing max nodes. Also, this setting affects performance so it's not unreasonable to change that on a server to prevent DoSing.

`maximum_go_to_depot_penalty`. 20 tiles isn't much and that's not counting other stuff, for an off-line depot you'd want a place to slow down that's at least as long as your trains. Max train length in the game is 64 but having penalty on 20 basically limits that playstyle to like 15 tile trains or even shorter.

`rail_station_penalty`. I've no idea why but 10% of servers at the time of scan (https://citymania.org/tools/serverstat) used a non-default value. That's not insignificant.

And that's just settings that came to my mind, same scan shows more cases of servers using non-default settings. Do you know the reason? I don't, I use default pf settings on CityMania and whatever changes Luukland/Novapolis had on the game modes I inherited can be ignored even though I'd prefer to keep them. But other people went to the trouble of changing those settings so they must have some reason. It's not good to just ignore them and drop all settings indiscriminately simply because you couldn't be arsed to fix some overflows :p

LordAro:
> You allowed settings to be configurable so that players can work out good defaults and now that players settled on different defaults you're just removing that.

No, it was so *developers* could work out defaults without having to recompile every single time. They were never intended to be altered by end users.

>But other people went to the trouble of changing those settings so they must have some reason.

Some reason is not necessarily a good reason. As we know from the whole signal debacle, there's an awful lot of cargo-culting and misleading anecdotal performance data that goes on with these things.

I'm not saying you're wrong about the the current defaults / constants being wrong though - it's definitely worth considering changing the ones you've highlighted. One might suggest that's out of scope for this PR though ;)

ldpl:
> > You allowed settings to be configurable so that players can work out good defaults and now that players settled on different defaults you're just removing that.
> 
> No, it was so _developers_ could work out defaults without having to recompile every single time. They were never intended to be altered by end users.

Developers that never play the game? :p Decisions like this require collecting feedback from large amount of players, otherwise you'll just end up limiting the game to your personal favorite playstyle.

> Some reason is not necessarily a good reason. As we know from the whole signal debacle, there's an awful lot of cargo-culting and misleading anecdotal performance data that goes on with these things.

But do you know the reason to judge whether it's good or not? It's one thing to consider all arguments and make a decision but it's another to just not even bother investigating.

> I'm not saying you're wrong about the the current defaults / constants being wrong though - it's definitely worth considering changing the ones you've highlighted. One might suggest that's out of scope for this PR though ;)

They're not universally "wrong". Some values are better suited for certain playstyles. I for sure wouldn't want `max_search_nodes` over 10k on a server but for a single player game having 100k may be alright.

LordAro:
> You allowed settings to be configurable so that players can work out good defaults and now that players settled on different defaults you're just removing that.

They were there was so _developers_ could work out defaults without having to recompile every single time. They were never intended to be altered by end users.

>But other people went to the trouble of changing those settings so they must have some reason.

Some reason is not necessarily a good reason. As we know from the whole signal debacle, there's an awful lot of cargo-culting and misleading anecdotal performance data that goes on with these things.

I'm not saying you're wrong about the the current defaults / constants being wrong though - it's definitely worth considering changing the ones you've highlighted. One might suggest that's out of scope for this PR though ;)

EmperorJake:
Just thought I'd provide a reason for regular players to adjust the pathfinder settings. 
I play on a server where the station pathfinder penalty has been reduced. A lot of our station designs rely on it for better traffic flow, as we want trains to go through empty platforms rather than waiting for the bypass track to be clear.

TrueBrain:
> Just thought I'd provide a reason for regular players to adjust the pathfinder settings. I play on a server where the station pathfinder penalty has been reduced. A lot of our station designs rely on it for better traffic flow, as we want trains to go through empty platforms rather than waiting for the bypass track to be clear.

@EmperorJake out of curiosity, what settings do you change to make that playstyle happen?

Eddi-z:
> No, it was so _developers_ could work out defaults without having to recompile every single time. They were never intended to be altered by end users.

[[citation needed]]

i'd dispute this statement. and also i'm putting my veto on this change.

aside from some situations needing to set non-default values to work properly, having the penalties available for easy reading from settings can help debugging why the pathfinder behaves weird in certain corner cases (e.g. short trains vs. long trains)

typical use cases include:
 * networks with short trains may need to increase occupied track per tile penalty to overcome switch and curve penalties on sidings
 * networks with overflow depots may use depot-reverse-penalty instead of twoway-eol (which has side effects and i hate that setting, and also it doesn't work with path signals)

EmperorJake:
> > Just thought I'd provide a reason for regular players to adjust the pathfinder settings. I play on a server where the station pathfinder penalty has been reduced. A lot of our station designs rely on it for better traffic flow, as we want trains to go through empty platforms rather than waiting for the bypass track to be clear.
> 
> @EmperorJake out of curiosity, what settings do you change to make that playstyle happen?

The only changed setting is yapf.rail_station_penalty = 100 instead of the default 1000.

Here's a common example of why we reduce it. With the default penalty the car carrier train would have waited for the tanker train to clear the central bypass track. The reduced penalty allows it to travel through the free platform without stopping, increasing traffic flow.
image

src/settings_type.h
```diff
@@ -382,82 +382,11 @@ struct ScriptSettings {
...
-	bool   ship_use_yapf;                    ///< use YAPF for ships
-	bool   road_use_yapf;                    ///< use YAPF for road
-	bool   rail_use_yapf;                    ///< use YAPF for rail
-	uint32 road_slope_penalty;               ///< penalty for up-hill slope
-	uint32 road_curve_penalty;               ///< penalty for curves
-	uint32 road_crossing_penalty;            ///< penalty for level crossing
-	uint32 road_stop_penalty;                ///< penalty for going through a drive-through road stop
-	uint32 road_stop_occupied_penalty;       ///< penalty multiplied by the fill percentage of a drive-through road stop
-	uint32 road_stop_bay_occupied_penalty;   ///< penalty multiplied by the fill percentage of a road bay
-	bool   rail_firstred_twoway_eol;         ///< treat first red two-way signal as dead end
```
frosch123, line 435: This setting should probably be kept. It's not a pathfinder setting, but rather a "remove two-way signals from the game and replace them with a magic pathfinder hint". These days defaults to "remove two-way signals, since noone used them anyway". In an ideal world this would not have been a pathfinder setting, but a different signal type.

TrueBrain (dismissed):
After some consideration and reading up on the comments, I am still going to approve this PR.

However, I do agree with @EmperorJake that there are play-styles we should support, which this PR makes impossible. But not by being able to tune internals of a pathfinder. That requires way too much in-depth knowledge of such systems for the average user, and means it is a setting for the 0.01%. I rather have a setting that is meaningful for the 99%.

As such, and @LordAro agrees, we should add a setting for such specific playstyles. In this case: "allow trains to pass through open platforms", or something. I leave it to @LordAro to either do that in this PR, or in a follow-up PR. But something like this should happen before 14.0.

The added benefit of this approach, is that it is pathfinder agnostic. And that means if we make changes to the pathfinder, or replace it all together, we can still keep the setting, and with that still support that playstyle.

Additionally, it makes clear to future-us that there is a playstyle that fancies that kind of playing.

In general, and in my strong opinion, we should always go for settings that define behaviour; not internals. This makes it much easier to explain to the user, and means we have far fewer bugs / issues because an internal setting got misunderstood.

Eddi-z:
> As such, and @LordAro agrees, we should add a setting for such specific playstyles. In this case: "allow trains to pass through open platforms", or something.

I can go with that approach, but we need some data on what that kind of settings/playstyles are actually in demand.

> I leave it to @LordAro to either do that in this PR, or in a follow-up PR. But something like this should happen before 14.0.

IMHO, that should be a separate PR, as it makes the discussion easier, but this one should be held off until that PR is ready.

2TallTyler:
I agree that a new setting should be added to allow trains to choose empty station tracks equal to or over occupied plain tracks. This is the cause of #9609.

vituscze:
I'd like to chime in as I've debugged the pathfinder in the past and submitted a couple bug fix PRs as a result. The pathfinder settings were immensely helpful there. Debugging the pathfinder is already annoying enough, being unable to change the settings on fly will make it even more painful.

There's also the issue with backwards compatibility: this PR will break a lot of old saves that used nonstandard settings. While I realize and understand that you cannot let backwards compatibility hold you back and prevent further progress, I still think you should need a compelling reason to break it. And I don't think that "powerusers using settings to fine tune their game" is good enough.

LordAro:
Well it compiles and runs and doesn't immediately crash, but I'm 90% certain there's all sorts of things wrong with it. So I made it a draft.

FLHerne:
I reduce `rail_station_penalty` for the same reason as EmperorJake above. IMO fixing that case should be a precondition of merging this patch.

I also increase `rail_shorter_platform_penalty` to 10000 so that trains never use a shorter platform, this avoids needing waypoints in cramped station layouts and removing the setting would break my existing games - long train entering a short bay platform instead of a long through one would lock the network up nicely. I'm less attached to that as it's obviously a pretty niche usecase.

PR #10624, opened 2023-04-10
Add: Some additional debugging information when chunk sizes are not as expected

LordAro:
## Motivation / Problem

Debugging SL errors is hard when they provide no information except "Invalid chunk size" with no context. Also when that message comes from one of 4 possible locations

## Description

Add some context

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/core/pool_func.hpp
+++ b/src/core/pool_func.hpp
@@ -14,7 +14,7 @@
 #include "mem_func.hpp"
 #include "pool_type.hpp"
 
-extern void NORETURN SlErrorCorruptFmt(const char *format, ...) WARN_FORMAT(1, 2);
+#include "../saveload/saveload_error.hpp" // SlErrorCorruptFmt
 
 /**
  * Helper for defining the method's signature.
@@ -157,13 +157,13 @@ DEFINE_POOL_METHOD(void *)::GetNew(size_t size)
 DEFINE_POOL_METHOD(void *)::GetNew(size_t size, size_t index)
 {
 	if (index >= Tmax_size) {
-		SlErrorCorruptFmt("%s index " PRINTF_SIZE " out of range (" PRINTF_SIZE ")", this->name, index, Tmax_size);
+		SlErrorCorruptFmt("{} index {} out of range ({})", this->name, index, Tmax_size);
 	}
 
 	if (index >= this->size) this->ResizeFor(index);
 
 	if (this->data[index] != nullptr) {
-		SlErrorCorruptFmt("%s index " PRINTF_SIZE " already in use", this->name, index);
+		SlErrorCorruptFmt("{} index {} already in use", this->name, index);
 	}
 
 	return this->AllocateItem(size, index);
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -651,11 +651,11 @@ static size_t _next_offs;
  */
 int SlIterateArray()
 {
-	int index;
-
 	/* After reading in the whole array inside the loop
 	 * we must have read in all the data, so we must be at end of current block. */
-	if (_next_offs != 0 && _sl.reader->GetSize() != _next_offs) SlErrorCorrupt("Invalid chunk size");
+	if (_next_offs != 0 && _sl.reader->GetSize() != _next_offs) {
+		SlErrorCorruptFmt("Invalid chunk size iterating array - expected to be at position {}, actually at {}", _next_offs, _sl.reader->GetSize());
+	}
 
 	for (;;) {
 		uint length = SlReadArrayLength();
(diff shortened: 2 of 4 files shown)


src/saveload/saveload.cpp
```diff
@@ -694,6 +694,7 @@ int SlIterateArray()
 			return INT32_MAX;
 		}
 
+		int index;
```
nielsmh, line 676: Is this variable used?

nielsmh (commented):
Is there any more information that is available that might be useful to include? Something like current chunk name perhaps?

src/saveload/saveload.cpp
LordAro: Yes, see the following lines :p

LordAro:
I've added what I can without altering function signatures and the like (only a couple have references to the chunk that's currently being loaded)

Have also converted usage to fmt instead of printf style. It's a bit ick due to the "extern" error in pool_func.hpp

src/core/pool_func.hpp
```diff
@@ -14,7 +14,7 @@
 #include "mem_func.hpp"
 #include "pool_type.hpp"
 
-extern void NORETURN SlErrorCorruptFmt(const char *format, ...) WARN_FORMAT(1, 2);
+#include "../saveload/saveload.h" // SlErrorCorrupt
```
rubidium42: Maybe create a new `saveload_error.h` with the `SlError` functions? Including that in here will at least not include all the file IO stuff.

src/saveload/saveload_error.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file saveload.h Functions/types related to errors from savegames. */
+
+#ifndef SAVELOAD_ERROR_HPP
+#define SAVELOAD_ERROR_HPP
+
+#include "../stdafx.h"
```
rubidium42: This include is not wanted in headers. Although I see a few headers that (erroneously I'd say) include it.

src/saveload/saveload_error.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * @tparam A The type of the first argument.
+ * @tparam Args The types of the other arguments.
+ * @note This function does never return as it throws an exception to
+ *       break out of all the saveload code.
+ */
+template <typename T, typename A, typename ... Args>
+static inline void NORETURN SlErrorCorruptFmt(const T &format, A first_arg, Args&&... other_args)
+{
+	/* The separate first_arg argument is added to aid overloading.
+	 * Otherwise the calls that do no need formatting will still use this function. */
```
rubidium42: There is no overloading of the function as far as I can see. Did you want to call this function `SlErrorCorrupt`? Or is it just sloppy-paste?

src/saveload/saveload.h
```diff
@@ -10,6 +10,7 @@
 #ifndef SAVELOAD_H
 #define SAVELOAD_H
 
+#include "saveload_error.hpp"
 #include "../fileio_type.h"
 #include "../fios.h"
 #include "../strings_type.h"
```
rubidium42: saveload_error.h includes strings_type, so it's not needed here anymore.

rubidium42 (commented):
I like that you changed this over from printf to format!

src/saveload/saveload_error.hpp
LordAro: Yup, straight copied from IConsolePrint :) I could rename this, or I could squash the args together...

src/saveload/saveload.h
LordAro: Even though this header itself uses StringID?

rubidium42: Yes. It makes little sense to include something that has already been included. It only slows things up.
Maybe it's time for me to do the old... can I compile each header with only stdafx.h, and which includes can I remove from the headers spiel again. But that's a slow and tedious process.
Having said that, it's not a huge problem to leave it either. After all, with the header protection it should skip most of the file from being processed.

LordAro: https://github.com/include-what-you-use/include-what-you-use ?

rubidium42: The question is a bit... where to draw the line? Technically this file should then also include `bitmath_func.hpp`, `cassert`, 'memory' and `stdafx.h`.
All in all it doesn't really matter. It's just that less includes help a tiny bit in building performance. Either way is ultimately fine by me.

rubidium42 (approved):

PR #10631, opened 2023-04-11
Codechange: Use std:: features for NewGRF town names.

PeterN:
## Motivation / Problem

Loading (and unloading) GRF Town Names requires manual allocation of nested listed and there are multiple levels of array indexing.

I've not tested but it looks like redefinition of a list would leak memory.

## Description

This PR replaces with std::vectors, and the text is stored as a std::string. This removes manual memory (de-)allocation and list counting. The NamePart union (which didn't really serve much purpose due to alignment) is removed as you can't easily put a std::string in a union.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -7758,8 +7758,8 @@ static void FeatureTownName(ByteReader *buf)
 		bool new_scheme = _cur.grffile->grf_version >= 7;
 
 		byte lang = buf->ReadByte();
+		StringID style = STR_UNDEFINED;
 
-		byte nb_gen = townname->nb_gen;
 		do {
 			ClrBit(lang, 7);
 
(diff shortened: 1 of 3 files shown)


src/newgrf.cpp
```diff
@@ -7758,53 +7758,51 @@ static void FeatureTownName(ByteReader *buf)
...
 	}
 
-	byte nb = buf->ReadByte();
-	grfmsg(6, "FeatureTownName: %u parts", nb);
+	townname->partlists[id].resize(buf->ReadByte());
+	grfmsg(6, "FeatureTownName: " PRINTF_SIZE " parts", townname->partlists[id].size());
 
-	townname->nbparts[id] = nb;
-	townname->partlist[id] = CallocT<NamePartList>(nb);
+	int i = 0;
```
LordAro: i & j only used for debug messages now i think? worth giving them more friendly names?

src/newgrf_townname.cpp
```diff
@@ -21,69 +21,45 @@
 
 #include "safeguards.h"
 
-static GRFTownName *_grf_townnames = nullptr;
+static std::vector<GRFTownName> _grf_townnames;
 static std::vector<StringID> _grf_townname_names;
```
LordAro, line 25: how much effort to get rid of this global? seems like it's entirely contained within `_grf_townnames`

src/newgrf_townname.h
```diff
@@ -17,29 +17,32 @@
...
-	uint16 maxprob;
-	NamePart *parts;
+	byte bitstart;               ///< Start of random seed bits to use.
+	byte bitcount;               ///< Number of bits of random seed to use.
+	uint16 maxprob;              ///< Total probability of all parts.
+	std::vector<NamePart> parts; ///< Parts
+};
+
+struct TownNameStyle
+{
```
LordAro:
```suggestion
struct TownNameStyle {
```

src/newgrf_townname.h
```diff
@@ -17,29 +17,32 @@
...
 struct NamePartList {
-	byte partcount;
-	byte bitstart;
-	byte bitcount;
-	uint16 maxprob;
-	NamePart *parts;
+	byte bitstart;               ///< Start of random seed bits to use.
+	byte bitcount;               ///< Number of bits of random seed to use.
+	uint16 maxprob;              ///< Total probability of all parts.
+	std::vector<NamePart> parts; ///< Parts
```
LordAro: probably wouldn't bother aligning with this one

src/newgrf_townname.cpp
```diff
@@ -21,69 +21,45 @@
...
 
 GRFTownName *GetGRFTownName(uint32 grfid)
 {
-	GRFTownName *t = _grf_townnames;
-	for (; t != nullptr; t = t->next) {
-		if (t->grfid == grfid) return t;
-	}
+	auto found = std::find_if(std::begin(_grf_townnames), std::end(_grf_townnames), [&grfid](const GRFTownName &t){ return t.grfid == grfid; });
+	if (found != std::end(_grf_townnames)) return &(*found);
```
LordAro: &* not particularly nice. Do we need to be returning pointers at all now? Can we do references/copy-ellision instead?

src/newgrf.cpp
PeterN: Yes, it's also possible to get the index from pointer arithmetic but that's a bit awful.

src/newgrf_townname.cpp
PeterN: _grf_townnames is per GRF, but _grf_townname_names isn't, so it would need more restructuring.

src/newgrf_townname.cpp
PeterN: Looks like it needs to return nullptr if it doesn't exist, so needs to be a pointer.

PeterN: Also this pattern is used elsewhere, it is normal as found is an iterator.

michicc: The "proper" C++ ham would be std::optional I guess.

PeterN: std::optional would return by value I think.

rubidium42 (approved):

PR #10633, opened 2023-04-11
Codechange: replace bridge cost loop by more efficient calculation

jengelh:
_This is a follow-up to https://github.com/OpenTTD/OpenTTD/pull/10403_ 

The bridge costing function has poor scalability. Looped addition is trivial and optimizable. The results from CalcBridgeLenCostFactor(n) for increasing n form a sequence which happens to have a constant-time formula. In practice, this runs at about 4 ns. The formula involves sqrt however, and since openttd wants to do without FP to avoid implementation/rounding differences, an integer unit-based implementation of sqrt will be used. Two options exist: introducing a loop again, or using bit twiddling via a log2(x) approximation, cf. https://math.stackexchange.com/questions/2469446/ .

Runtime per call of CalcBridgeLenCostFactor (i5-1135G7):

```
len      oldcode     FPU     IPU
8           4.84 ns  3.9 ns  6.6 ns
16          8.53      |      8.2
32         17.10      |      8.2
64         32.75      |      9.9
128        47.78      |      9.9
256        87.79      |      11.9
512       158.2       |      11.9
1024      382.8       |      13.8
2048      752.3       |      15.8
4096     1477        same    15.8
```

(Max map size is currently 4096...)

### DIFF
--- a/src/tunnelbridge_cmd.cpp
+++ b/src/tunnelbridge_cmd.cpp
@@ -12,6 +12,7 @@
  */
 
 #include "stdafx.h"
+#include <cmath>
 #include "newgrf_object.h"
 #include "viewport_func.h"
 #include "command_func.h"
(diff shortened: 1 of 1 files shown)


ldpl:
As max bridge len is just 4k costs can easily be pre-calculated at compile time or just put in a constant array. Though I'm not sure why bridge cost function performance is even a concern.

Issue #7941, opened 2020-01-15
Pathfinder doesn't consider blocked tiles past destination station


esigra:
After watching OpenTTD trains for many hours I conclude that the most notable pathfinding problem is that when a train has left a station but stopped (due to breakdown or waiting at a signal) with the last car on the tile after the station exit, where a signal is, the next trains sees that the whole station track is free and decides to wait for that particular station track, possibly indefinitely, even though another station track is available. Thus it blocks the whole queue of trains to the station.

This could be fixed in either of the following 2 ways:
1. The pathfinder only picks a station track for a train when it can reserve the station track AND the track beyond to the next signal.
2. Station exits are considered to have built-in signals (like depots have). This is the solution that I would prefer (even if someone would decide that then the train length capacity of stations would have to be reduced by 1 car to compensate).

MingweiSamuel:
I'm not sure exactly what you're describing, could you provide a save or image (with path reservations on)?

I wish stations had built in signals (2) but it would be a major change. Currently you can usually get the same behavior by putting two way path signals on the end(s) of your station.

esigra:
![waiting_for_free_path](https://user-images.githubusercontent.com/12582697/74112544-ead50880-4b9d-11ea-895c-2a4c5e92ca07.png)
At this station with 2 tracks traffic usually flows fine. If a station track is occupied, the next train takes the other. Problem is only when a train has left a station track but not yet the tile after with the signal, and breaks down there.

MingweiSamuel:
I personally have run into situations like this where trains get stuck deciding at PBS. I don't know if the devs would consider it a bug though. It may be possible to fix/improve based on some yapf settings.

For this situation specifically you can fix it by using block signals at the entrance instead of PBS

Eddi-z:
the problem here is that pathfinding and path reserving happens in two separate phases:
1) the pathfinding adds up penalties until the destination station is reached, and then picks the platform with the lowest penalty
2) the path reserving reserves up to that destination platform, and then a little bit further until the next safe waiting point (signal or end of track), however, this additional bit of track was not part of the pathfinder penalties earlier, so there were no penalties added for if these tiles are blocked

PR #10634, opened 2023-04-11
Codechange: Use cached name for all station/industry/town name formatting.

PeterN:
## Motivation / Problem

#7941 introduced cached names for stations, industries and towns to assist sorting performance, but they're not used elsewhere.

## Description

This change also uses the cached names when formatting regular strings, along with a test to ensure infinite recursion.

## Limitations

Industries does something with genders/cases so this might be wrong here.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -35,6 +35,7 @@
 #include "game/game_text.hpp"
 #include "network/network_content_gui.h"
 #include "newgrf_engine.h"
+#include "core/backup_type.hpp"
 #include <stack>
 
 #include "table/strings.h"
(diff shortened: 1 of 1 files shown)


src/strings.cpp
```diff
@@ -1456,7 +1456,12 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 				const Industry *i = Industry::GetIfValid(args->GetInt32(SCC_INDUSTRY_NAME));
 				if (i == nullptr) break;
 
-				if (_scan_for_gender_data) {
+				static bool use_cache = true;
+				if (use_cache) { // Use cached version if first call
+					use_cache = false;
```
rubidium42: It probably does not matter here, since `GetCachedName()` should be fairly trivial and as such not throw any exceptions.
But `AutoRestoreBackup(use_cache, false)` will also restore `use_cache` when an exception happens somewhere during the call.

rubidium42 (dismissed):

src/strings.cpp
PeterN: As that exists... probably better to use it as we have it.

rubidium42 (approved):

PR #10635, opened 2023-04-11
Fix: Inconsistent dropdown button width padding.

PeterN:
## Motivation / Problem

Three small changes which are all related...

1) Dropdown button padding does not take account of additional padding used when displaying dropdown itself, which may result in cropped text:

image

2) Some windows add padding onto size->width, but the size parameter already includes padding, resulting in double-padding.

3) WorldGen town name selection doesn't take account of option widths:

image

## Description

1) Fixed by including extra bevel padding:

image

2) Fixed by including padding before updating size output parameter.

3) Fixed by including town name style widths.

image

## Checklist for review

### DIFF
--- a/src/genworld_gui.cpp
+++ b/src/genworld_gui.cpp
@@ -591,62 +591,63 @@ struct GenerateLandscapeWindow : public Window {
 
 	void UpdateWidgetSize(int widget, Dimension *size, const Dimension &padding, Dimension *fill, Dimension *resize) override
 	{
+		Dimension d{0, (uint)FONT_HEIGHT_NORMAL};
 		const StringID *strs = nullptr;
 		switch (widget) {
 			case WID_GL_HEIGHTMAP_HEIGHT_TEXT:
 				SetDParam(0, MAX_TILE_HEIGHT);
-				*size = GetStringBoundingBox(STR_JUST_INT);
+				d = GetStringBoundingBox(STR_JUST_INT);
 				break;
 
 			case WID_GL_START_DATE_TEXT:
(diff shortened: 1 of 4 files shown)


PeterN:
To be clear it's only really wide because I happened to have a lot of town-name NewGRFs selected that have long names.

image

Of course, part 3 could be undone so that the town name dropdown button ignores the text width, and then we could change the dropdown to treat the widget width as a minimum instead.

PeterN:
Okay, so if we remove the default-sizing of dropdown buttons, and make all dropdown windows auto-size (instead of being fixed to the widget size), we get...

image

Lots of windows do still size their dropdown buttons for each option anyway, so this would need combing through to find them.

PeterN:
That is then odd behaviour as it only sizes to fix the current selection when the window is opened.

1) It'll be narrow, and selecting the Japanese set won't change that.
2) You close and re-open the window and it is now much wider because that set is selection.
3) Change back to a set with a shorter name and it stays wide.
4) Re-open the window and it's back to narrow again.

LordAro:
Why can't it resize after selection?

PeterN:
We don't normally resize windows based on the current selection, except for interface scale on the settings window.

https://user-images.githubusercontent.com/639850/231266723-16c630d6-2750-41e6-af82-706e3944a700.mp4

🤷

PeterN:
Part 3 has now been dropped. So this PR now tweaks a dropdown button's padding, and fixes the double padding.

Other bits have been or will be done in a separate PR, so I've hidden those from here.

rubidium42 (approved):

PR #7796, opened 2019-10-22
Unittests

sdcloudt:
Basically I try to reopen #7254 which is closed. One of the final notes made in that PR is that we cannot have the Makefile rely on system dependent file paths. One of the proposed ways to achieve this is to add googletest as a submodule.

For now, I believe this is the easiest way to achieve the possibility to add unittests to OpenTTD. As I do not see anyone in the near future integrate googletests through autotools, or make OpenTTD use CMake.

I think it is good to have some framework in place to add unittests to OpenTTD. Many things are difficult to test within OpenTTD, but there are some things which you can test, for example #7735 requires some URI parser which should be testable.

PR #7254, opened 2019-02-20
Codechange: introduce a few unit tests

nikolas:
There was some discussion about introducing unit tests to OpenTTD on irc. Here's a proof of concept with some unit tests for a few basic math functions.

To run these tests, you need to install [Googletest](https://github.com/google/googletest). On Debian you can do an `apt install googletest`. Then you should be able to run them by going to the `test` directory and running `./test-all.sh`. If we go further down this route we can of course set these up with the CI stuff like Azure.

PR #10636, opened 2023-04-11
Add support for unit tests, and a few unit tests

rubidium42:
## Motivation / Problem

Tests for new and existing code will make development & refactoring easier.
And will make it easier to detect regressions and unintended changes.
See #9038 and #7796. Triggered by #10570 which adds some tests as constexpr code. 

## Description

Split building of OpenTTD into a "library" and the executable, by extracting the `main()` functions into separate files. And then building the library without the `main()` function and the openttd executable is the `main()` function with the library linked. The library is not an actual library, but some construct in CMake that prevents building the same objects twice, since the library is going to be linked with the unit tests and then run.

Add catch2 as third-parties library. It does not exist as library in ubuntu 20.04, so otherwise it would require updating those. Furthermore there were issues in getting it to work with Windows and Mac OS X. By third-partying it, it will just be available and do its thing.

Add unit test from #9038/#7796 and convert constexpr tests of #10570 to a proper unit test, and integrate catch2 into the build. Add an extra executable (openttd_test) that links the new "library" and the unit tests. With the integration it automatically gets picked up with ctest, and be run in parallel with the regression tests.

## Limitations

I'm not sure whether OS/2 still builds.
I'm not sure whether the release workflows still function.
It has been a long time whack-a-moling to get this functioning, so I'm not sure whether I did the right things.

## Checklist for review


src/tests/math_func.cpp
```diff
@@ -0,0 +1,77 @@
...
+	CHECK(25 == LeastCommonMultiple(5, 25));
+	CHECK(25 == LeastCommonMultiple(25, 5));
+	CHECK(130 == LeastCommonMultiple(5, 26));
+	CHECK(130 == LeastCommonMultiple(26, 5));
+}
+
+TEST_CASE("GreatestCommonDivisorTest - Negative")
+{
+	CHECK(4 == GreatestCommonDivisor(4, -52));
+	// CHECK(3 == GreatestCommonDivisor(-27, 6)); // error - returns -3
```
2TallTyler, line 34: Commented code, and an error elsewhere?

rubidium42: I reckon an error elsewhere. It's a bit via-via-via (#9038 -> #7796 -> #7254), but there was some discussion about it before without real remedy https://github.com/OpenTTD/OpenTTD/pull/7254/files#r259590007
In any case, I copied this test straight from #9038, which copied it from #7796.

PeterN:
Had a look to see if hierarchical tests could be done, it looks like it's supported by catch2 by way of SECTIONs, but ctest itself doesn't do anything with those and only shows the top-level TEST_CASE.

PeterN (approved):
Looks fine to me.

WimLeflere:
Nice work!

PR #10641, opened 2023-04-12
Change: Make all dropdown lists extend width if necessary.

PeterN:
## Motivation / Problem

Dropdown list items can sometimes be too long for the parent widget. In most cases we try to make the parent widget wide enough, but in some cases this becomes too wide.

image

## Description

This PR removes the default-false auto_width parameter, and makes all dropdowns automatically extend in width as required.

image

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -592,9 +592,14 @@ public:
 		return this->result >= COLOUR_END ? STR_COLOUR_DEFAULT : _colour_dropdown[this->result];
 	}
 
+	uint Width() const override
+	{
+		return ScaleGUITrad(28) + WidgetDimensions::scaled.hsep_normal + GetStringBoundingBox(this->String()).width + WidgetDimensions::scaled.dropdowntext.Horizontal();
+	}
+
 	uint Height(uint width) const override
 	{
-		return std::max(FONT_HEIGHT_NORMAL, ScaleGUITrad(12) + 2);
+		return std::max(FONT_HEIGHT_NORMAL, ScaleGUITrad(12) + WidgetDimensions::scaled.vsep_normal);
 	}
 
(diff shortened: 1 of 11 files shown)


src/company_gui.cpp
```diff
@@ -592,9 +592,14 @@ class DropDownListColourItem : public DropDownListItem {
...
 
+	uint Width() const override
+	{
+		return ScaleGUITrad(28) + WidgetDimensions::scaled.hsep_normal + GetStringBoundingBox(this->String()).width + WidgetDimensions::scaled.dropdowntext.Horizontal();
+	}
+
 	uint Height(uint width) const override
 	{
-		return std::max(FONT_HEIGHT_NORMAL, ScaleGUITrad(12) + 2);
+		return std::max(FONT_HEIGHT_NORMAL, ScaleGUITrad(12) + WidgetDimensions::scaled.vsep_normal);
```
2TallTyler, line 602: Fixing scaling here doesn't really fit the commit name, could be its own commit or just a different commit title. (Unless you're planning to just squash the whole PR into one commit, of course.)

2TallTyler (commented):
Obviously we have to draw a line somewhere between strings that fit and strings that don't, but perhaps all the built-in options in English (UK) should fit?

nanes

PeterN:
Yes we could do that, but this PR no longer touches the width of the dropdown widgets, only the width of the opened dropdowns.

There are other issues with dropdown widgets, the widget system resizes based on the current value when the window is opened and requires every window to manually work out what size is really required. I think these are another PR.

src/company_gui.cpp
PeterN: Updated wording on the commit.

2TallTyler (approved):

PR #10651, opened 2023-04-13
Change: Use cstdint instead of rolling our own types.

PeterN:
## Motivation / Problem

stdafx.h has some old cruft, one of which is where we define our own sized types and min/max definitions.

## Description

This change uses cstdint instead, and removes some Haiku workarounds that may or may not be needed these days -- they're 14 years old.

This does not change how we use our named types, that would conflict with everything...

## Limitations

There's still loads of cruft in stdafx.h!

## Checklist for review

### DIFF
--- a/src/3rdparty/squirrel/include/squirrel.h
+++ b/src/3rdparty/squirrel/include/squirrel.h
@@ -32,9 +32,9 @@
 
 #include "../../../string_type.h"
 
-typedef __int64 SQInteger;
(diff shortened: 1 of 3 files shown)


LordAro:
Should be sufficent to replace the printf definitions with the `PRInx` versions - https://en.cppreference.com/w/cpp/types/integer#Format_macro_constants

TrueBrain (approved):

PR #10655, opened 2023-04-14
Change: base autosaves intervals on real time (instead of game time)

TrueBrain:
## Motivation / Problem

There are two fundamental issues with autosave:
- When fast-forwarding, it saves way too often
- When paused, it never saves

Both makes no sense. Autosaves are meant to prevent you from accidentally losing your work. The emphasis on "your" work.

## Description

To solve both issues, the autosave now works on real time. You can select every 10 / 30 / 60 / 120 minutes, which are similar to what the setting was in game-months.

When you pause, autosaving will stop. Unless you make any change to the game; then it will continue to make autosaves, even so the game is paused. Unpausing / pausing resets this mechanism.

## Limitations

In theory we could make autosave a free field, like all other settings, where you can fill in any time you want. Sadly, this component is not available in the Options menu (only in the Settings menu). The scope of this PR would explode, and as such, is considered out of scope for this PR. But, if you think we should have that, please, give it a go!

## Checklist for review


src/openttd.cpp
```diff
@@ -1441,23 +1445,31 @@ void StateGameLoop()
...
+/** Interval for regular autosaves. Initialized at zero to disable till settings are loaded. */
+static IntervalTimer<TimerGameAutosave> _autosave_interval(std::chrono::milliseconds::zero(), [](auto)
 {
-	if (_settings_client.gui.autosave == 0) return;
-	if ((_cur_month % _autosave_months[_settings_client.gui.autosave]) != 0) return;
-
 	_do_autosave = true;
 	SetWindowDirty(WC_STATUS_BAR, 0);
+
+	static FiosNumberedSaveName _autosave_ctr("autosave");
```
LordAro, line 1432: I realise it's not really changed, but does this have thread safety implications?

TrueBrain: I have no clue, but I think two things worth mentioning in this context:

- The timer runs in the game-thread; as such, this can only be accessed by that one.
- The autosave starts its own thread and copies this information in there. So when the function returns, this variable is no longer used.

So I think we are good here :)

glx22:
As I understand it, when a command happens during pause `PM_COMMAND_DURING_PAUSE` flag is set to allow saving.
But shouldn't it be unset once the save is done, to not overwrite all autosaves with the same stuff if no other command happen during the pause.

TrueBrain:
> As I understand it, when a command happens during pause `PM_COMMAND_DURING_PAUSE` flag is set to allow saving. But shouldn't it be unset once the save is done, to not overwrite all autosaves with the same stuff if no other command happen during the pause.

That is a choice we can make. Do we want to reset the flag every time we save .. I dunno. For now I was under the assumption: if you once start to build, that is it, you didn't actually want to pause to take a break, but you wanted to pause because you wanted to build.
But an argument can be made to reset it after every autosave. Downside is, that is makes the timer even more autosave specific. But I don't have an issue with that.

2TallTyler:
I have no comments on the command flags, but you can probably delete `TimerGameCalendar::Priority::AUTOSAVE`, as it should be unused now. 🙂

TrueBrain:
> I have no comments on the command flags, but you can probably delete `TimerGameCalendar::Priority::AUTOSAVE`, as it should be unused now. 🙂

Fixed!

PeterN:
Why does this needs its own Timer class specific to autosaves?

TrueBrain:
> Why does this needs its own Timer class specific to autosaves?

To quote the header-file:

```
 * Timer that represents, in real time, the time played.
 * 
 * "Time played" is defined here as the time elapsed since the game started.
 * If the game is paused, the timer stops unless a Command is executed. Than
 * the timer will continue to run.
 *
```

There is no other timer yet delivering this functionality. If you have better ideas, I am all ears :)

src/openttd.cpp
```diff
@@ -1443,23 +1447,35 @@ void StateGameLoop()
...
+	 * to make auto-saves when nothing more is changing. */
+	_pause_mode &= ~PM_COMMAND_DURING_PAUSE;
 
 	_do_autosave = true;
 	SetWindowDirty(WC_STATUS_BAR, 0);
+
+	static FiosNumberedSaveName _autosave_ctr("autosave");
+	DoAutoOrNetsave(_autosave_ctr);
+
+	_do_autosave = false;
```
glx22, line 1435: Is this global still needed ?

TrueBrain: Yup! A grep shows two usages of this variable, which this PR will not address.

src/timer/timer_game_realtime.h
```diff
@@ -0,0 +1,59 @@
...
+ * - Only tick when unpaused or when there was a Command executed recently (recently: since last autosave) (PeriodFlags::AUTOSAVE).
+ *
+ * @note The lowest possible interval is 1ms, although realistic the lowest
+ * interval is 27ms. This timer is only updated when the game-thread makes
+ * a tick, which happens every 27ms.
+ * @note Callbacks are executed in the game-thread.
+ */
+class TimerGameRealtime {
+public:
+	enum PeriodFlags {
```
PeterN, line 30: I know this is scoped already (and I didn't mention it on the other timers), but enum class?

TrueBrain: I kinda deliberately made them non-class-enums, so you can do `TimerGameRealtime::AUTOSAVE` instead of `TimerGameRealtime::PeriodFlags::AUTOSAVE`, to make the API slightly easier.

Similar thing I did for the other timers.

If we don't like that, I can change it, but it is a lot more letters for no real good reason in my opinion :) Let me know what you think!

PeterN (approved):

PR #10658, opened 2023-04-15
Codechange: Replace some p1/p2 parameter names with better names

rubidium42:
## Motivation / Problem

There are some remnants of p1/p2 naming from way way back, that are not referring to one of two points or something similar, but to just arbitrary parameters.

## Description

Replace those p1/p2 parameter names with something that describes what the parameters actually mean.

## Limitations

None; it's just parameter names.

## Checklist for review

### DIFF
--- a/src/cheat_gui.cpp
+++ b/src/cheat_gui.cpp
@@ -50,30 +50,30 @@ static int32 _money_cheat_amount = 10000000;
  * Note that the amount of money of a company must be changed through a command
  * rather than by setting a variable. Since the cheat data structure expects a
  * variable, the amount of given/taken money is used for this purpose.
- * @param p1 not used.
- * @param p2 is -1 or +1 (down/up)
+ * @param new_value not used.
+ * @param change_direction is -1 or +1 (down/up)
  * @return Amount of money cheat.
  */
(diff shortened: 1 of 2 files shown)


TrueBrain (approved):

LordAro:
Should we not be getting rid of the "unused" parameters, rather than just marking them as such?

rubidium42:
> Should we not be getting rid of the "unused" parameters, rather than just marking them as such?

They are unused for that specific instance of either the cheat or settings change callbacks. Interpret them as having `override` for a class' function.

LordAro:
Can we document that in the comment then?

rubidium42:
We can document it implements some (C-style) function interface, but... this PR is changing only the small part that hasn't been touched yet. In total there are way more functions that behave similarly that need to be annotated then. Especially for the settings, as adding it to just the ones that still refered to p1/p2 is less than 6% of the callbacks (2 of at least 33). 
That would make this PR much larger, and maybe we shouldn't be using C-style function interfaces anymore but something more C++-ish. But, in my opinion, that is out of the scope of this PR that just tries to improve the naming of parameters.

LordAro:
Doesn't matter either way, just that "Unused (used for xxx callback)" would've prevented this whole discussion :)

PR #10659, opened 2023-04-16
Codechange: Optimise FormatNumber by removing seprintf calls

LordAro:
## Motivation / Problem

Doing some profiling for some reasons, I noticed that `FormatNumber` appeared in the profile to be taking ~6% of CPU time. This felt excessive, so investigated.

Looks like FormatNumber calls `seprintf` an excessive number of times.

## Description

Initially I played around with using a basic `std::to_string(number)` and reforming the output as needed as what `FormatNumber` does currently seems unnecessarily complicated, but that quickly spiralled out of control into a "rewrite strings api to use std::string" rabbit hole.

Looked at the original function a bit closer and realised that it was calling `seprintf("%d")` for (every) single digit. This felt excessive, and was easily replaced.

With the following command line against a release build:

    ./build/openttd -X -c empty/empty.cfg -x -snull -mnull -vnull:ticks=30000 -g ./media/baseset/opntitle.dat

I get the following perf numbers,  before and after:

           3.58063 +- 0.00196 seconds time elapsed  ( +-  0.05% )
           3.48277 +- 0.00645 seconds time elapsed  ( +-  0.19% )

Similar improvement with wentbourne too

## Limitations

Technically it's lost a bit of buffer overflow protection due to the removal of the last check from within `seprintf`? I don't think that's a problem in practice though.

There's an additional issue of why the finance window is redrawn quite as often as it is - it appears that this is due to `DrawOverlappedWindow`, but the window itself isn't overlapping anything, so...

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -330,7 +330,7 @@ static char *FormatNumber(char *buff, int64 number, const char *last, const char
 	int thousands_offset = (max_digits - fractional_digits - 1) % 3;
 
 	if (number < 0) {
-		buff += seprintf(buff, last, "-");
+		if (buff != last) *buff++ = '-';
 		number = -number;
 	}
 
@@ -340,7 +340,7 @@ static char *FormatNumber(char *buff, int64 number, const char *last, const char
 		if (i == max_digits - fractional_digits) {
 			const char *decimal_separator = _settings_game.locale.digit_decimal_separator.c_str();
 			if (StrEmpty(decimal_separator)) decimal_separator = _langpack.langpack->digit_decimal_separator;
-			buff += seprintf(buff, last, "%s", decimal_separator);
+			buff = strecpy(buff, decimal_separator, last);
 		}
 
 		uint64 quot = 0;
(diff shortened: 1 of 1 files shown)


rubidium42:
In (normal) practice the overflow protection probably doesn't matter. But at some point it could be a problem, so I rather have an `if (buff != last)` or so before adding the digit to buff.

Regarding the overlapped window thing. It probably overlaps something in the main viewport that needs to be redrawn, so it redraws than, but then the windows that are on top of that, like the finance window, need to be (partly) redrawn. Think of smoke, vehicles, industry/house animations, etc.

nielsmh:
There could definitely be an argument for not using the stdlib at all for integer number formatting, and completely roll our own, if we're abusing it at the "processing a single character at a time" level anyway.

LordAro:
You're quite right, the only remaining seprintfs are for the minus sign (a single character), or just to output the separator (can easily be replaced with a strecpy)

PeterN:
Needs a bounds check even if it's rare, or better using a string system that doesn't write into a fixed length buffer. Hello rabbit-hole...

TrueBrain (approved):

LordAro:
For reference, perf numbers are now:

```
           3.46687 +- 0.00810 seconds time elapsed  ( +-  0.23% )
```

PR #10662, opened 2023-04-16
WIP: introduce WASM-runtime and move TGP into a WASM-module

TrueBrain:
## Motivation / Problem

For a while now I have been wondering if we could use WASM as a plugin platform for OpenTTD.

WASM, WebAssembly, was originally created for the web, but has its introduction into native applications for a while now. It basically is a bytecode format with some rules. The thing that makes WASM click, is that LLVM is used as backend. This means it can output shockingly efficient binaries.

The bytecode format is universal, that is to say: it doesn't care on what host it is executed. In that sense, very similar to Java (but without all the issues of Java).
There is a lot to say about WASM, but I am not going to in this WIP draft PR :)

I picked wasm-micro-runtime as WASM runtime, as it sounded nice, and has a built-in JIT. The benefit is that we can distribute a generic WASM file, which OpenTTD, via the JIT, optimizes even further for the host system when executed.

To have a nice showcase of what WASM can do, I picked the terrain generator. I could have picked anything else .. from NewGRF to AIs, to economy, or even vehicle movement. But terrain generator sounded nice.

This PR is purely meant to see what the CI thinks about this, and give a tiny bit of visibility to my work here. It is not (yet) meant to be taken seriously.

## Description

This as a proof-of-concept that you can run something like a terrain generator in an addon. In this case written in WASM.

Although WASM is fast (like, seriously fast), as this commit uses FastJIT (and not LLVM-JIT), it is about twice as slow as native C. But, in return you do get a modular system for terrain generation.

You have to compile TGP to WASM yourself. For this, install the [wasi-sdk](https://github.com/WebAssembly/wasi-sdk), and run:
`wasi-sdk-20.0/bin/clang++ -O3 -o tgp.wasm ../src/tgp_wasm.cpp -Wl,--export-all -I generated -Wl,--allow-undefined`
(mind you: I picked C++ here to write TGP in; it could have been any other language that compiles to WASM)

The conversion to a WASM module was surprisingly easy. It was mostly about creating a simple API between the WASM code and OpenTTD's internals. But as they both compile from C++, I could just share header-files and stuff.
There are a few things to be careful for .. `Map::SizeX()` compiles fine in WASM, but as it is a static method reading a static variable, it always returns 0. So I added `MapSizeX` in the API to avoid this issue.

When generating a map with this PR, it first generates it 10 times from C, and give you the results. Then it does it 10 times from WASM, and give you those results.

You can observe that the first run in WASM is a lot slower, as the JIT has to do its job.
On Debug, the WASM is a lot faster (as it is compiled with `-O3`).
On release, C is about twice as fast.

LLVM-JIT could make a change here, but when trying it gave errors compiling OpenTTD's code; haven't looked into it further.

AOT is not useful for OpenTTD, as those binaries are targetted for a specific host platform, while the idea is to have generic binaries.

Lastly, currently WASI is enabled, which is terrible. WASI allows for file-access, network-access, etc, from WASM. This would be ... scary, at best. So WASI should be disabled, and we should sandbox it ourselves. But I was too lazy to actually do this.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -25,6 +25,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+      with:
+        submodules: true
 
     - name: Setup cache
       uses: actions/cache@v3
@@ -96,6 +98,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+      with:
+        submodules: true
 
     - name: Install dependencies
       run: |
@@ -172,6 +176,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+      with:
+        submodules: true
 
     - name: Install dependencies
       env:
@@ -260,6 +266,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+      with:
+        submodules: true
 
     - name: Prepare cache key
       id: key
@@ -354,6 +362,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+      with:
+        submodules: true
 
     - name: Setup MSYS2
       uses: msys2/setup-msys2@v2
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "wasm-micro-runtime"]
+	path = wasm-micro-runtime
+	url = https://github.com/bytecodealliance/wasm-micro-runtime
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,30 @@ if(APPLE)
     )
 endif()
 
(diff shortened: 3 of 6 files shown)


nielsmh:
Two thoughts:
- Embedding a runtime that heavily depends on JIT to be efficient also makes a lack of JIT a major barrier to the game being usable on less common CPUs.
- For NewGRF support to make sense, the best approach would probably be to somehow translate the NewGRF data itself to WASM modules first, and then feed those generated WASM data to the runtime. This could also open up for WASM modules (complying with some given interface) becoming an alternative to NewGRF for mods, possibly even supplanting some of the functionality only possible in GS right now.

TrueBrain:
`wasm-micro-runtime` is a bad fit for OpenTTD. Looking into other runtimes that appear to be better.

The advantage of this runtime is that it embeds a JIT. But this is also the problem, as it only works for a select amount of targets. Less than we support. And if you go for an AOT solution, there are runtimes that are easier to embed than this.

Closing this PR up for now; we will continue with another runtime another day :)

PR #10664, opened 2023-04-16
Codechange: Use {ZEROFILL_NUM} instead of custom seprintf + {RAW_STRING}

rubidium42:
## Motivation / Problem

We have `{ZEROFILL_NUM}`, so why should we zerofill to a string ourselves and then pass that as `{RAW_STRING}`?

## Description

Change `{RAW_STRING}` into `{ZEROFILL_NUM}` for `STR_FORMAT_DATE_TINY` and `STR_FORMAT_DATE_ISO`.
Also update the translations, as nothing has changed conceptually to the strings so retranslating is not needed at all (separate commit).

## Limitations

Might be slightly more expensive due to #10659, but then again... it's rarely called.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -448,13 +448,8 @@ static char *FormatTinyOrISODate(char *buff, Date date, StringID str, const char
 	YearMonthDay ymd;
 	ConvertDateToYMD(date, &ymd);
 
-	char day[3];
-	char month[3];
-	/* We want to zero-pad the days and months */
-	seprintf(day,   lastof(day),   "%02i", ymd.day);
-	seprintf(month, lastof(month), "%02i", ymd.month + 1);
-
-	int64 args[] = {(int64)(size_t)day, (int64)(size_t)month, ymd.year};
+	/* Day and month are zero-padded with ZEROFILL_NUM, hence the two 2s. */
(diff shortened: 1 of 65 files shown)


src/strings.cpp
```diff
@@ -448,13 +448,8 @@ static char *FormatTinyOrISODate(char *buff, Date date, StringID str, const char
 	YearMonthDay ymd;
 	ConvertDateToYMD(date, &ymd);
 
-	char day[3];
-	char month[3];
 	/* We want to zero-pad the days and months */
```
TrueBrain: Isn't the comment now not a bit weird?

rubidium42: Yes because there is no zero-padding to be seen in the function, and no because it answers the question: why the twos after the day and month? Though maybe the code needs to be more pedantically obvious, so the comment isn't needed?

TrueBrain: This is just an annoying thing of our string system honestly, hard to make sure people understand this in a year or 2 :P

Suggestion: `Day and month are zero-padded with ZEROFILL_NUM, hence the two 2s`

Or something like that.

TrueBrain (approved):

PR #10667, opened 2023-04-16
Codechange: Use more std::string in fios land

rubidium42:
## Motivation / Problem

Seen `strecpy` from `std::string` to a `char`-array in #9984 and thought... better make those `char`-arrays `std::string`.

## Description

Basically replace the C-style strings with C++ style strings. Also simplify `FiosBrowseTo` as the returned name had little to do with the actual functionality of the function, and made conversion harder.

## Limitations

I didn't check the OS/2 and Windows changes.
Requires #10727 and #10728 to be merged first.

## Checklist for review


src/os/os2/os2.cpp
```diff
@@ -76,11 +76,11 @@ void FiosGetDrives(FileList &file_list)
 			fios->type = FIOS_TYPE_DRIVE;
 			fios->mtime = 0;
 #ifndef __INNOTEK_LIBC__
-			snprintf(fios->name, lengthof(fios->name),  "%c:", 'A' + disk - 1);
+			fios->name = fmt::format("{:c}:", 'A' + disk - 1);
```
LordAro: Can we get away without fmt here?
`fios->name = std::string(2, 'A' + disk - 1); fios->name[1] = ':';` ? Bit clunky, perhaps.

Similarly for other places

rubidium42: I've gone for a slightly different approacn, which should be fine allocation wise given the short string optimisations.

src/fios.cpp
```diff
@@ -56,9 +56,9 @@ bool FiosItem::operator< (const FiosItem &other) const
 	int r = false;
 
 	if ((_savegame_sort_order & SORT_BY_NAME) == 0 && (*this).mtime != other.mtime) {
-		r = (*this).mtime - other.mtime;
+		r = this->mtime - other.mtime;
 	} else {
-		r = strnatcmp((*this).title, other.title);
+		r = strnatcmp(this->title.c_str(), other.title.c_str());
```
LordAro: feels like we need a C++ version of this

src/fios.cpp
```diff
@@ -374,28 +372,27 @@ static void FiosGetFileList(SaveLoadOperation fop, fios_getlist_callback_proc *c
...
 	if ((dir = ttd_opendir(_fios_path->c_str())) != nullptr) {
 		while ((dirent = readdir(dir)) != nullptr) {
-			strecpy(d_name, FS2OTTD(dirent->d_name).c_str(), lastof(d_name));
+			std::string d_name = FS2OTTD(dirent->d_name);
 
 			/* found file must be directory, but not '.' or '..' */
 			if (FiosIsValidFile(_fios_path->c_str(), dirent, &sb) && S_ISDIR(sb.st_mode) &&
-					(!FiosIsHiddenFile(dirent) || strncasecmp(d_name, PERSONAL_DIR, strlen(d_name)) == 0) &&
-					strcmp(d_name, ".") != 0 && strcmp(d_name, "..") != 0) {
+					(!FiosIsHiddenFile(dirent) || strncasecmp(d_name.c_str(), PERSONAL_DIR, d_name.size()) == 0) &&
```
LordAro: same for str(n)casecmp, perhaps

src/fios.cpp
```diff
@@ -101,29 +101,29 @@ void FileList::BuildFileList(AbstractFileType abstract_filetype, SaveLoadOperati
...
 	}
 
 	/* If no name matches, try to parse it as number */
 	char *endptr;
-	int i = strtol(file, &endptr, 10);
-	if (file == endptr || *endptr != '\0') i = -1;
+	int i = strtol(file.data(), &endptr, 10);
+	if (file.data() == endptr || *endptr != '\0') i = -1;
 
 	if (IsInsideMM(i, 0, this->size())) return &this->at(i);
```
LordAro: I think we can do better than this nowadays? std::from_chars, or whatever?

rubidium42: `std::to_chars` is too new for our currently supported OSX. It needs at least OSX 10.15, whereas we require 10.13 right now. So arguably `std::from_chars` requires 10.15 as well.

src/fios.cpp
rubidium42: You mean a `strnatcmp` that accepts `std::string` but just calls the `const char *` version? This because `strnatcmp` is mostly implemented by calling C-style APIs in a library or the OS.

rubidium42: See #10728.

src/fios.cpp
rubidium42: See #10727.

TrueBrain (approved):
Tested on Windows; and OS/2 code looks good on paper.

PR #10668, opened 2023-04-16
Fix #10660: Sprite Font scale affected by viewport zoom level limits.

### LINKED ISSUE #10660: [Bug]: Interface including fonts and viewports doesn’t scale properly on Mac OS with retina display
### Version of OpenTTD

OpenTTD 13.1, also current nightly (20230415-master), Mac OS 13.3 (ARM)

### Expected result

Opening OpenTTD on a Mac with a retina display and setting interface scaling to 2x should cause everything, including fonts, to scale. Only the margins around the fonts changes. The scale of the actual game is also too small, including in the startup screen.

### Actual result

1x scaling, everything ‘correct’ in terms of proportions but too small to be seen on a high-resolution screen:
Screenshot 2023-04-16 at 11 51 20

2x scaling, margins around interface elements increased but nothing else, and everything still too small to comfortably read:
Screenshot 2023-04-16 at 11 51 16

If I enable ‘Auto-detect size’, it just changes to 2x scaling and looks exactly the same as when I select that manually.

### Steps to reproduce

Open OpenTTD 13.1 on Mac OS. Fiddle with the Interface scale option in Game Options.

I’m using the original_windows graphics set, if that might make a difference. @LordAro suggested this might also be helpful info:

Screenshot 2023-04-16 at 11 59 19

PeterN:
## Motivation / Problem

The maximum viewport zoom level setting inadvertently affects the scaled sprite font, since with variable interface scaling all the font_zoom related variables were merged with gui_zoom.

This issue does not affect scaleable fonts, only the built-in sprite font.

image

## Description

Add separate _font_zoom which is only used to scale the sprite font. This is not clamped to the viewport zoom level limits, so always follows interface scale.

image

## Limitations

Non-font sprites are still limited to the viewport zoom level, even when shown in a GUI. This is because there's currently no differentiation between viewport sprites and gui sprites, and also some sprites are used in both locations.

Could possibly be done but I think it's separate to this.

## Checklist for review

### DIFF
--- a/src/fontcache/spritefontcache.cpp
+++ b/src/fontcache/spritefontcache.cpp
@@ -22,6 +22,16 @@
 static const int ASCII_LETTERSTART = 32; ///< First printable ASCII letter.
 
 /**
+ * Scale traditional pixel dimensions to font zoom level, for drawing sprite fonts.
+ * @param value Pixel amount at #ZOOM_LVL_BASE (traditional "normal" interface size).
+ * @return Pixel amount at _font_zoom (current interface size).
+ */
+static int ScaleFontTrad(int value)
+{
+	return UnScaleByZoom(value * ZOOM_LVL_BASE, _font_zoom);
+}
+
+/**
  * Create a new sprite font cache.
  * @param fs The font size to create the cache for.
  */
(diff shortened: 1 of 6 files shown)


rubidium42 (approved):
It's definitely a step in the right direction.

PR #10669, opened 2023-04-16
Fix #10627: Houses subsitute specs should only be copied on first definition.

PeterN:
## Motivation / Problem

#10627 broke "ITL Houses 2.1" which redefines a house after having defined it already. Before #10627, this action would copy the substituted house specs again, after #10627 this copy only happened on first definition, but the overrides and checks were still performed, which could fail the second time.

However... this opened a can of worms, as the spec, since at least 2007 and probably earlier, states that only the first definition copies anything:

> Only the first property 08 setting copies properties; if you later change it, properties will stay. 

In addition, another step has been missed, that the church/stadium flags should not be copied.

> House flags 40 and 80 are exceptions; these flags are never set automatically.

So before #10627, substitue specs were always copied despite redefinition, and church/stadium flags were carried over too.

## Description

This PR changes OpenTTD to match the specs, so that only the first definition counts, no further substitute copying occurs, and the church/stadium flags are cleared.

This pattern makes more sense as it matches every other feature.

However this may potentially make some NewGRFs behave differently: potentially they were 'always broken' in this case.

None-the-less this change does fix the error that prevents "ITL Houses 2.1" from loading.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -2409,6 +2409,9 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
 					housespec->random_colour[2] = 0x0C;  // they stand for red, blue, orange and green
 					housespec->random_colour[3] = 0x06;
 
+					/* House flags 40 and 80 are exceptions; these flags are never set automatically. */
+					housespec->building_flags &= ~(BUILDING_IS_CHURCH | BUILDING_IS_STADIUM);
+
 					/* Make sure that the third cargo type is valid in this
 					 * climate. This can cause problems when copying the properties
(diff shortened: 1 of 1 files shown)


PeterN:
@2TallTyler In "ITL House 2.1", this affects house 103 (museum) which substitutes id 16. This is defined in landmarks.nml and tropic.nml. It may affect other IDs, but this is the one that triggers a crash.

PeterN:
Hmm, thinking about this the fix for church/stadium should be a separate commit.

But also, I wonder if later changing  property 08 should still update subst_id.

rubidium42 (approved):

PR #10670, opened 2023-04-17
Add: NewGRF persistent storage for vehicles.

PeterN:
## Motivation / Problem

People asked for persistent storage for vehicles. This could e.g. allow complex chain calculations to be performed only during the attach callback.

This bumps the savegame so take a backup...

## Description

This probably implements this, but I'm not able to test it. 

~~It likely bloats things if lots of vehicles use persistent storage, as a store allocates 1KB for the vehicle. If you have 10,000 vehicles (e.g. 1000 * 5 tile trains) that's 10MB without thinking about it.~~

PersistentStorageArray is changed to an std::vector. This allows using only as many storage slots as necessary, so the previous 1KB fixed array (256 x uint32_t) is no longer an issue.

PersistentStorage is only allocated if the vehicle uses it.

~~Two callbacks have write access enabled for PersistentStorage: CBID_VEHICLE_START_STOP_CHECK and CBID_TRAIN_ALLOW_WAGON_ATTACH. This is rudimentary and could be extended.~~

## Limitations

All the vehicle callback stuff is const, so there's a const_cast in there to get rid of that. Nasty.

~~Also PersistentStorage needs NewGRF feature, which isn't the same as vehicle type, so I was lazy there too. This is supposed to be updated on load but I didn't do that. **What does this mean? I don't remember...**~~

~~Also missing NewGRF debug GUI stuff.~~

Desync-inviting.

## Checklist for review

### DIFF
--- a/src/newgrf_engine.cpp
+++ b/src/newgrf_engine.cpp
@@ -24,6 +24,7 @@
 #include "newgrf_railtype.h"
 #include "newgrf_roadtype.h"
 #include "ship.h"
+#include "newgrf_debug.h"
 
 #include "safeguards.h"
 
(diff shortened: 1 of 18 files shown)


src/newgrf_engine.cpp
```diff
@@ -1144,11 +1164,15 @@ bool UsesWagonOverride(const Vehicle *v)
  * @param param2   Second parameter of the callback
  * @param engine   Engine type of the vehicle to evaluate the callback for
  * @param v        The vehicle to evaluate the callback for, or nullptr if it doesn't exist yet
+ * @param callback Set if Persistent Storage is readonly
```
glx22:
```suggestion
 * @param readonly Set if Persistent Storage is readonly
```

src/newgrf_engine.cpp
```diff
@@ -1160,12 +1184,18 @@ uint16_t GetVehicleCallback(CallbackID callback, uint32_t param1, uint32_t param
  * @param engine   Engine type of the vehicle to evaluate the callback for
  * @param v        The vehicle to evaluate the callback for, or nullptr if it doesn't exist yet
  * @param parent   The vehicle to use for parent scope
+ * @param callback Set if Persistent Storage is readonly
```
glx22:
```suggestion
 * @param readonly Set if Persistent Storage is readonly
```

PeterN:
Not sure if NML needs to know about this as well.

PeterN:
Closing due to lack of interest

PR #10671, opened 2023-04-17
Change title screen GUI: Move buttons to bottom edge of screen

nielsmh:
## Motivation / Problem

The title screen GUI has several oddities and wants more and more buttons, but doesn't afford structuring them well.

Allow the title screen game to be more prominent, by freeing the centre part of the viewport.

The prominence of the climate selection buttons in the title screen menu is often questioned.

## Description

image

Split the main menu into three sub-windows, positioned along the bottom edge of the screen.
- The middle window has four buttons with the theme, *Start playing*. Starting and loading single player games, joining multi player games, and accessing the scenario editor.
- The left hand window gives access to user generated content: Starting games from scenarios and height maps (those should maybe be moved into the New Game window), accessing the content downloader, and configuring NewGRF, AI, and GS.
- The right hand window offers setting up the game and getting information. This is also where a "Help" button (#7786) would go.

## Limitations

Does not look good in 640x480, the windows starts overlapping. Can we change the minimum resolution to 800x600?

Some of the strings might need a review, maybe they can be improved for the new layout.

Wishlist items: Adding snazzy graphical icons to the buttons, and maybe a bigger and fancier game logo at the top of the screen.

## Checklist for review

### DIFF
--- a/src/intro_gui.cpp
+++ b/src/intro_gui.cpp
@@ -41,6 +41,8 @@
 #include "safeguards.h"
 
 
+namespace {
+
 /**
  * A viewport command for the main menu background (intro game).
  */
@@ -95,8 +97,7 @@ struct IntroGameViewportCommand {
 	}
 };
 
-
-struct SelectGameWindow : public Window {
+struct IntroGameAnimator {
 	/** Vector of viewport commands parsed. */
 	std::vector<IntroGameViewportCommand> intro_viewport_commands;
 	/** Index of currently active viewport command. */
@@ -105,6 +106,16 @@ struct SelectGameWindow : public Window {
 	uint cur_viewport_command_time;
 	uint mouse_idle_time;
 	Point mouse_idle_pos;
+
+	IntroGameAnimator()
+	{
+		this->ReadIntroGameViewportCommands();
+
+		this->cur_viewport_command_index = (size_t)-1;
+		this->cur_viewport_command_time = 0;
+		this->mouse_idle_time = 0;
+		this->mouse_idle_pos = _cursor.pos;
+	}
 
 	/**
 	 * Find and parse all viewport command signs.
@@ -174,21 +185,7 @@ struct SelectGameWindow : public Window {
 		}
 	}
 
-	SelectGameWindow(WindowDesc *desc) : Window(desc)
-	{
-		this->CreateNestedTree();
-		this->FinishInitNested(0);
-		this->OnInvalidateData();
-
-		this->ReadIntroGameViewportCommands();
-
-		this->cur_viewport_command_index = (size_t)-1;
-		this->cur_viewport_command_time = 0;
-		this->mouse_idle_time = 0;
-		this->mouse_idle_pos = _cursor.pos;
-	}
-
-	void OnRealtimeTick(uint delta_ms) override
+	void Tick(uint delta_ms)
 	{
 		/* Move the main game viewport according to intro viewport commands. */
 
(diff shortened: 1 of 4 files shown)


src/intro_gui.cpp
```diff
@@ -259,17 +260,55 @@ struct IntroGameAnimator {
...
 
+Point PositionSelectGameWindow(Window *w)
+{
+	Point p;
+	const int edge_offset = ScaleGUITrad(20);
+	const int width = w->width;
+	const int height = w->height;
+
+	switch (w->window_number) {
+		case 0:
```
2TallTyler: Could there be an enum for these window numbers? (Honest question, I know they're not widgets but maybe we could treat them similarly.)

src/intro_gui.cpp
```diff
@@ -285,20 +324,19 @@ struct SelectGameWindow : public Window {
 	 */
 	void OnInvalidateData(int data = 0, bool gui_scope = true) override
```
2TallTyler: If this override function is empty, can it just be deleted?

src/intro_gui.cpp
```diff
@@ -333,17 +371,14 @@ struct SelectGameWindow : public Window {
...
 			int height = GetStringHeight(str, size->width);
-			if (height > 3 * FONT_HEIGHT_NORMAL) {
-				/* Don't let the window become too high. */
-				Dimension textdim = GetStringBoundingBox(str);
-				textdim.height *= 3;
-				textdim.width -= textdim.width / 2;
-				*size = maxdim(*size, textdim);
-			} else {
+			/* Don't let the window become too high - allow up to 7 lines. */
+			while (height > 7 * FONT_HEIGHT_NORMAL) {
```
2TallTyler: Could use some constants to replace magic numbers. Or at least an explanation of the `size->width * 7 / 5`: where does the 5 come from?

2TallTyler (commented):
I like this suggestion. It nicely solves the problem of limited real estate and constant rearrangement of the current menu, and shows off the title game feature nicely — which in 2023 is rare to see in a main menu.

I am slightly concerned that at higher interface size settings (2x on my 1920x1080 monitor), there's almost no margin to play the game in a smaller window than full screen without the windows overlapping. The current main menu can tolerate much, much narrower windows including my vertical monitor, without being cut off. I wonder, could the three windows rearrange themselves automatically into a vertical stack instead of simply overlapping each other? Less attractive, sure, but still usable.

PeterN:
This does not seem to be any improvement to me. It moves the important stuff (the menu items) away and makes the unimportant background game most prominent.

nielsmh:
I tried making the code stack the windows a bit if they end up overlapping.

My primary argument for making the title game more prominent on screen is that it can serve as tutorial and inspiration for novice players. It can't show you what to click to do things, but it can show what is possible to do, and exhibit tried-and-true network layouts. Having the middle of the screen obscured takes away a lot from that.

PeterN:
Hmm, the window can be moved. And at 1x interface scale at 1080p, I don't think it's really obscured:

image

nielsmh:
Yes, the window can be moved, but how many actually do that? I'm guessing that the majority of players aren't even aware that it's a window that can be moved. Having the middle of the screen obscured *by default* takes away from what a title game author can present, because there is very little way to guarantee what will be shown on screen of the stuff around the centered-on point.
The zoom of the title game also adjusts with the interface scale, so the area obscures by the original main menu window is approximately constant, but it's hard to count  on anything more than the immediate area around the window to be on screen, and it's hard to show off a station or interchange layout if you only have a bunch of edges to play with.

LC-Zorg:
> I'm guessing that the majority of players aren't even aware that it's a window that can be moved.

I think you greatly underestimate the players. ;)
I remember my beginnings with the game, I remember the title screen of version 1.4.X and I remember the station that was under the main menu. I wanted to see what was there so I just moved this window the same way like it is possible to move ALL other windows in this game. :) This behavior quickly becomes intuitive, so I don't think it's a problem for anyone. Well, unless someone doesn't know that you can move any windows in game, but then I don't think that moving the menu window will help them in the game at all. I just write what I feel.
However, I definitely understand that someone might want to watch a game that is in the background. I have that too. :) But honestly, I don't really like the proposed layout and the fact that the background becomes the most important element here. It may be interesting once or twice, but then it loses its meaning. Besides, I think that every opntitle creator takes into account the location of the menu window and adjusts their map accordingly. The only case where this is a disadvantage is vehicle tracking, but this problem can also be solved without such a quite radical rebuild. Also, these little windows with climates, while not particularly functional today, have always been a constant in this game. They can still be used as a background game switcher as well. ;)

Any way, I think that the problem of the covered game image can be solved in another way. I don't know if a vertical window would take up less space if it had the same number of buttons. I think that only reducing them can make the window smaller. However, the best solution I can imagine is to add a button that will collapse the window or possibly hide it at the bottom edge (although this is quite unexpected behavior).
What would be worth adding in this case is a button to disable / enable panning. Sometimes it can be simply annoying, and sometimes stopping the view allow to look at something that might have interested the player.

Vertical main menu 1 1

andythenorth:
Thanks, the title screen GUI has taken a different direction with OpenTTD 15, so I'm closing this to reduce open PR count :)

PR #10672, opened 2023-04-17
Change: Remove limit of objects, stations and roadstops per NewGRF.

PeterN:
## Motivation / Problem

255 objects, stations and roadstops per NewGRF.

## Description

64000+ objects, stations and roadstops per NewGRF.

Only specs for 255 different items can be defined in one action 0, but I doubt increasing this batching feature matters much.

Action 00 IDs are already an extended byte.
Action 03 IDs are now an extended byte for objects.

For stations, Action 04 only supports 256 IDs, so two additional Action 00 properties have been added to set name and class name, the same way as is already done for road stops.

Saveload is bumped so that 16 bit IDs can be stored.

## Limitations

Untested beyond starting and loading an existing object NewGRF.

## Checklist for review

### DIFF
--- a/src/base_station_base.h
+++ b/src/base_station_base.h
@@ -22,13 +22,13 @@ extern StationPool _station_pool;
 struct StationSpecList {
 	const StationSpec *spec;
 	uint32 grfid;      ///< GRF ID of this custom station
-	uint8  localidx;   ///< Station ID within GRF of station
+	uint16_t localidx; ///< Station ID within GRF of station
 };
 
 struct RoadStopSpecList {
 	const RoadStopSpec *spec;
 	uint32 grfid;      ///< GRF ID of this custom road stop
-	uint8  localidx;   ///< Station ID within GRF of road stop
+	uint16_t localidx; ///< Station ID within GRF of road stop
 };
 
 struct RoadStopTileData {
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -5643,10 +5643,10 @@ static void VehicleMapSpriteGroup(ByteReader *buf, byte feature, uint8 idcount)
 
 static void CanalMapSpriteGroup(ByteReader *buf, uint8 idcount)
 {
-	std::vector<CanalFeature> cfs;
+	std::vector<uint16_t> cfs;
 	cfs.reserve(idcount);
 	for (uint i = 0; i < idcount; i++) {
-		cfs.push_back((CanalFeature)buf->ReadByte());
+		cfs.push_back(buf->ReadExtendedByte());
 	}
 
 	uint8 cidcount = buf->ReadByte();
@@ -5674,10 +5674,10 @@ static void StationMapSpriteGroup(ByteReader *buf, uint8 idcount)
 		return;
 	}
 
-	std::vector<uint8> stations;
+	std::vector<uint16_t> stations;
 	stations.reserve(idcount);
 	for (uint i = 0; i < idcount; i++) {
-		stations.push_back(buf->ReadByte());
+		stations.push_back(buf->ReadExtendedByte());
 	}
 
 	uint8 cidcount = buf->ReadByte();
@@ -5732,10 +5732,10 @@ static void TownHouseMapSpriteGroup(ByteReader *buf, uint8 idcount)
 		return;
 	}
 
-	std::vector<uint8> houses;
+	std::vector<uint16_t> houses;
 	houses.reserve(idcount);
 	for (uint i = 0; i < idcount; i++) {
-		houses.push_back(buf->ReadByte());
+		houses.push_back(buf->ReadExtendedByte());
 	}
 
 	/* Skip the cargo type section, we only care about the default group */
@@ -5764,10 +5764,10 @@ static void IndustryMapSpriteGroup(ByteReader *buf, uint8 idcount)
 		return;
 	}
 
-	std::vector<uint8> industries;
+	std::vector<uint16_t> industries;
 	industries.reserve(idcount);
 	for (uint i = 0; i < idcount; i++) {
-		industries.push_back(buf->ReadByte());
+		industries.push_back(buf->ReadExtendedByte());
 	}
 
 	/* Skip the cargo type section, we only care about the default group */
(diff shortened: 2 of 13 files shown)


JGRennison:
You'll likely also need to extend EntityIDMapping and the various connected bits including the savegame format, (e.g something like: https://github.com/JGRennison/OpenTTD-patches/commit/43d58ce6c886f09e758b160edfdb70e1664e2887)

PeterN:
Thanks. First commit now extends the saveload (hopefully) ready for the object change.

JGRennison:
For road stops, the main trouble point is variable 68.
To support more than 255 stops, I've put the upper bits of the road stop ID in the previously unused bits 24 - 31, https://jgrennison.github.io/OpenTTD-patches/newgrf-roadstops.html#roadstop_road_stop_info_nearby_tiles.
This is a bit ugly, however does allow extending the ID range in backwards compatible way as in https://github.com/JGRennison/nml/commit/e04bee5e1517bff2be19394c5b5b1f657b1932ef#diff-23e315495dd26363695444fb44183da884c64781e40a21d410f3482d317df694. 

There are unlikely to be many existing users of this variable, so it may be more pragmatic to re-organise it more like https://jgrennison.github.io/OpenTTD-patches/newgrf-roadstops.html#roadstop_road_stop_info_nearby_tiles_v2, or to allocate a new variable ID which matches this.

PeterN:
Yeah, I'd already done the simple step of clamping to stop it overflowing.

andythenorth:
NMLC will need a patch for this.  I tried a trivial increase of block allocation to 0xFFFF (action0.py), but NMLC then throws internal error trying to allocate byes in action 3 output.

JGR has a functioning NMLC patch, but I believe it includes some compatibility mechanisms for vanilla.  
https://github.com/JGRennison/nml/commit/faa5a9278c424c05607625550b74bca543aa957e
I tried a naive cherrypick of that commit, but NMLC then throws an error for missing action7 in random action2.

PeterN:
Yea, the patch to nmlx will most likely be simpler. Ensure its padding a word around and then output an extended byte.

PeterN:
Actually I think it would make sense to make all action3s to read extended bytes, to simplify tooling.

PeterN:
Okay, so after pretending it was done, I have reworked it to extend action 3 for all features first, then bump the savegame and increase the limits.

JGRennison:
It would probably be worthwhile to add a bit to `_ttdpatch_flags`/variable 85, as otherwise GRFs which use this will have no way to correctly identify if this is supported and emit a human-readable error message if not.

glx22:
Station and road-stop var68 changes might be problematic if newgrf uses them to decide if tile graphics should be merged.

PeterN:
Yes, we will need to add extra vars to support this, but for the current vars it is clamped to avoid overflowing into the adjacent bits.

JGRPP may already extra vars for this, in which case it makes sense to copy those.

JGRennison:
I added a new variable with a slightly different format here: https://jgrennison.github.io/OpenTTD-patches/newgrf-roadstops.html#roadstop_road_stop_info_nearby_tiles_v2, it isn't allocated an ID so it's not really a big problem if what you add doesn't match though.

PeterN:
Hmm, I can't count, not sure why I left gaps for the station properties.

src/newgrf.cpp
```diff
@@ -2117,6 +2120,20 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 				break;
 			}
 
+			case 0x1B: // Minimum bridge height (not implemented)
+				buf->ReadWord();
+				buf->ReadWord();
+				buf->ReadWord();
+				buf->ReadWord();
```
JGRennison, line 2126: `break;` is missing here.

PeterN: Oops!

src/object_type.h
```diff
@@ -19,10 +19,9 @@ static const ObjectType OBJECT_STATUE       =   2;    ///< Statue in towns
 static const ObjectType OBJECT_OWNED_LAND   =   3;    ///< Owned land 'flag'
 static const ObjectType OBJECT_HQ           =   4;    ///< HeadQuarter of a player
 
-static const ObjectType NUM_OBJECTS_PER_GRF = 255;    ///< Number of supported objects per NewGRF; limited to 255 to allow extending Action3 with an extended byte later on.
-
 static const ObjectType NEW_OBJECT_OFFSET   =   5;    ///< Offset for new objects
 static const ObjectType NUM_OBJECTS         = 64000;  ///< Number of supported objects overall
+static const ObjectType NUM_OBJECTS_PER_GRF = NUM_OBJECTS; ///< Number of supported objects per NewGRF; limited to 255 to allow extending Action3 with an extended byte later on.
```
michicc: Outdated comment :)

PeterN:
Updated to fix out of date comment.

src/base_station_base.h
```diff
@@ -22,13 +22,13 @@ extern StationPool _station_pool;
 struct StationSpecList {
 	const StationSpec *spec;
 	uint32 grfid;      ///< GRF ID of this custom station
-	uint8  localidx;   ///< Station ID within GRF of station
+	uint16_t localidx; ///< Station ID within GRF of station
```
LordAro, line 25: is using the _t types something we want to start doing? imo being inconsistent is worse than updating everything all at once (if we even end up doing that)

src/object_type.h
```diff
@@ -19,10 +19,9 @@ static const ObjectType OBJECT_STATUE       =   2;    ///< Statue in towns
 static const ObjectType OBJECT_OWNED_LAND   =   3;    ///< Owned land 'flag'
 static const ObjectType OBJECT_HQ           =   4;    ///< HeadQuarter of a player
 
-static const ObjectType NUM_OBJECTS_PER_GRF = 255;    ///< Number of supported objects per NewGRF; limited to 255 to allow extending Action3 with an extended byte later on.
-
 static const ObjectType NEW_OBJECT_OFFSET   =   5;    ///< Offset for new objects
 static const ObjectType NUM_OBJECTS         = 64000;  ///< Number of supported objects overall
+static const ObjectType NUM_OBJECTS_PER_GRF = NUM_OBJECTS; ///< Number of supported objects per NewGRF
```
LordAro, line 24: does this constant need to exist anymore? as there's really no "per grf" limit, it's entirely covered by the global limit (if i'm understanding things correctly..)

PeterN: I left it in case we want to increase NUM_OBJECTS, as the per GRF limit would stay, although looking at it lots of that is limited to uint16 so it would be more involved anyway.

src/base_station_base.h
PeterN: :shrug: They are being used elsewhere, although actually in some cases that was 10 years ago...

michicc (dismissed):
Code style choices are choices, your call to merge or change them.

michicc (approved):

PR #10673, opened 2023-04-18
Fix #10665: CheckEngines should ignore wagons

### LINKED ISSUE #10665: [Bug]: NO_VEHICLES_AVAILABLE_YET calculation does not consider climate
### Version of OpenTTD

1.13.1, Windows 10

### Expected result

"No Vehicles Are Available Yet" message should appear when a new game is started before 1935 in Toyland.  Introduction date of MightyMover trucks should be fixed (not randomized).

### Actual result

"No Vehicles Are Available Yet" message appears when a new game is started before 1925 in Toyland.  Starting a game after 1925 but before 1935 creates a game with no available vehicles, until 1935 when MightyMover trucks become available.  No warning message is displayed in this case.

### Steps to reproduce

1. Start a new game Jan. 1, 1925 in Toyland.
2. "No vehicles available yet" message does not appear.
3. No vehicles are available.
4. Fast-forward to 1935, trucks are introduced.

aeonofdiscord:
## Motivation / Problem

This fixes the issue described in #10665: the "No vehicles available yet" warning doesn't pop up if the player starts a game in Toyland between 1925 and 1935.

## Description

`CheckEngines()` in `engine.cpp` fails to distinguish rail wagons from locomotives, so start dates between 1925 and 1935 (after several rail cars are introduced for use in Temperate, but before locomotives are available in Toyland) do not trigger the "no vehicles" warning. The fix explicitly ignores wagons when checking for available engines.

Closes #10665.

## Limitations

None; this is a simple fix for a minor UX issue and shouldn't cause any additional problems.

## Checklist for review

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -1246,6 +1246,9 @@ void CheckEngines()
 	for (const Engine *e : Engine::Iterate()) {
 		if (!e->IsEnabled()) continue;
 
+		/* Don't consider train wagons, we need a powered engine available. */
+		if (e->type == VEH_TRAIN && e->u.rail.railveh_type == RAILVEH_WAGON) continue;
+
 		/* We have an available engine... yay! */
 		if ((e->flags & ENGINE_AVAILABLE) != 0 && e->company_avail != 0) return;
 


src/engine.cpp
```diff
@@ -1246,6 +1246,10 @@ void CheckEngines()
 	for (const Engine *e : Engine::Iterate()) {
 		if (!e->IsEnabled()) continue;
 
+		if (e->type == VEH_TRAIN && e->u.rail.railveh_type == RAILVEH_WAGON) {
+			continue;
+		}
```
2TallTyler: Could use a comment, also the check can be one line:

```suggestion
		/* Don't consider train wagons, we need a powered engine available. */
		if (e->type == VEH_TRAIN && e->u.rail.railveh_type == RAILVEH_WAGON) continue;
```

2TallTyler (commented):
Welcome, and thanks for the contribution! 😄

I've tested and this works as advertised. Just a couple nitpicks with the code, then it looks good to go! If you can amend your commit or rebase the fix into the same, that would be ideal as we like to keep our commit history clean. 🙂

aeonofdiscord:
Ah yeah, I can add a comment but collapsing the conditional and the body into a single line would contravene the style guide.

2TallTyler:
From the control flow section of the [style guide](https://github.com/OpenTTD/OpenTTD/blob/master/CODINGSTYLE.md). 😉 

> When only a single statement is contained, the brackets can be omitted. In this case, put the single statement on the same line as the preceding keyword (if, while, etc.). Note that this is only allowed for if statements without an else clause.

PeterN (approved):

2TallTyler (approved):

Issue #10170, opened 2022-11-13
Options GUI has no scrollbar


2TallTyler:
### Version of OpenTTD

Nightly (98a2d0a), Windows 10

### Expected result

When using the default spritefont, interface scaling has to be a power of 2 to scale the font properly, so I use 2x. The Options GUI used to fit on my screen but must be taller now, so it does not.

I would expect it to have a scrollbar so I can access its contents.

### Actual result

The Options GUI extends off the bottom of my screen and has no scrollbar to reach content below.

![noscrollbar](https://user-images.githubusercontent.com/55058389/201533298-4c8d7e9e-0096-46c3-a13f-f63b11f4b717.png)

### Steps to reproduce

1. Set the interface scaling high enough that it doesn't fit on your screen
2. See that it has no scrollbar.

2TallTyler:
Resolved by #10674.

PR #10674, opened 2023-04-18
Change: Split Game options into General, Graphics, and Sound tabs.

PeterN:
## Motivation / Problem

Game Options is very large and has lots of different settings on it.

image

## Description

This change splits the Game Options window into three small tabs, "General", "Base sets" and "Video":

image
image
image

There isn't really a specific widget that could be used to represent tabs (these came in later games...) so this uses a standard button in a contrasting colour.

"Interface size" is left on the (default) General tab, as in modern terminology it fits more into a "graphics" option than a "video" option, but I don't think a separate Graphics tab is appropriate.

## Checklist for review

### DIFF
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -163,6 +163,18 @@ static const std::map<int, StringID> _scale_labels = {
 	{  500, STR_GAME_OPTIONS_GUI_SCALE_5X },
 };
 
+static const std::map<int, StringID> _volume_labels = {
+	{  0, STR_GAME_OPTIONS_VOLUME_0 },
+	{  15, STR_NULL },
+	{  31, STR_GAME_OPTIONS_VOLUME_25 },
+	{  47, STR_NULL },
+	{  63, STR_GAME_OPTIONS_VOLUME_50 },
+	{  79, STR_NULL },
+	{  95, STR_GAME_OPTIONS_VOLUME_75 },
+	{  111, STR_NULL },
+	{  127, STR_GAME_OPTIONS_VOLUME_100 },
+};
+
 struct GameOptionsWindow : Window {
 	GameSettings *opt;
 	bool reload;
@@ -178,6 +190,8 @@ struct GameOptionsWindow : Window {
 
 		this->InitNested(WN_GAME_OPTIONS_GAME_OPTIONS);
 		this->OnInvalidateData(0);
+
+		this->SetTab(WID_GO_TAB_GENERAL);
 	}
 
 	void Close() override
@@ -342,13 +356,25 @@ struct GameOptionsWindow : Window {
 				break;
 
 			case WID_GO_BASE_SFX_VOLUME:
-				DrawSliderWidget(r, 0, INT8_MAX, _settings_client.music.effect_vol, {});
+				DrawSliderWidget(r, 0, INT8_MAX, _settings_client.music.effect_vol, _volume_labels);
 				break;
 
 			case WID_GO_BASE_MUSIC_VOLUME:
-				DrawSliderWidget(r, 0, INT8_MAX, _settings_client.music.music_vol, {});
+				DrawSliderWidget(r, 0, INT8_MAX, _settings_client.music.music_vol, _volume_labels);
 				break;
 		}
+	}
+
+	void SetTab(int widget)
+	{
+		this->SetWidgetsLoweredState(false, WID_GO_TAB_GENERAL, WID_GO_TAB_GRAPHICS, WID_GO_TAB_SOUND, WIDGET_LIST_END);
+		this->LowerWidget(widget);
+
+		int pane = 0;
+		if (widget == WID_GO_TAB_GRAPHICS) pane = 1;
+		else if (widget == WID_GO_TAB_SOUND) pane = 2;
+		this->GetWidget<NWidgetStacked>(WID_GO_TAB_SELECTION)->SetDisplayedPlane(pane);
+		this->SetDirty();
 	}
 
 	void OnResize() override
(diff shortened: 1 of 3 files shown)


nielsmh:
Oh, I was about to start doing this myself too.
Personally I'd call the tabs General, Graphics, Sound, and put the grahics baseset selection on the Graphics tab.

nielsmh:
Oh, and we kind of do have tabs in the GUI: Station Details and Vehicle Details windows have buttons on the bottom to switch the lower pane of the window.

PeterN:
Yes, that's true, but they're not very prominent either (I'd forgotten about them as tabs) so maybe not the best example to follow.

nielsmh:
Definitely, the volume sliders in the current layout were just jammed in there to add _some_ way for players to adjust things from the intro screen. It was an easy place to put them without altering the layout too much.

2TallTyler:
I second @nielsmh's suggestion. General/Graphics/Sound is a pretty universal organization scheme these days, so players know where to find what they want. `Base sets` is very OpenTTD-specific.

For bonus points cleaning up the bug tracker...does this close #10170? 😃

PeterN:
Potentially, more so with the requested tab changes.

andythenorth:
Does this have the bonus effect that the window won't be cut off on some screen sizes (with no way to resize)?

LordAro:
Yellow on grey doesn't look great... Any other options?

PeterN:
> Does this have the bonus effect that the window won't be cut off on some screen sizes (with no way to resize)?

That's what #10170 referenced above is.

> Yellow on grey doesn't look great... Any other options?

It's orange. Grey on grey doesn't really work either.

PeterN:
Tweaked the order... and these are now yellow rather than orange buttons, I dunno...

Not convinced I've got the volume sliders laid out well, but they're definitely more prominent.

image
image
image

LordAro:
Since you're in the area...

I can't find it anywhere, but I definitely recall an issue where teh game options window was too wide due to the Video driver string. Splitting it within the string is probably too difficult, but perhaps it could be put on its own line, under "Current driver: "? Would help a little

2TallTyler:
`Sound/Music` could just be `Sound`, I think. I usually see music included in the Sound tab in other games.

No opinion on orange vs yellow tabs.

PeterN:
Yellow looks better I think, but perhaps the extra panel around them could go. Need to play :)

LC-Zorg:
I think I like this change. :) I don't know, I'm not sure. From my point of view, it is not really needed, the problem of too large a window can also be solved with a slider, but the advantage I see in such a layout may be the ability to add or move other options that could now fit better here. One of the options that would be worth considering would be the option to choose a font.
The yellow color is ok in my opinion. I like the updated division, it's similar to the one I was thinking to myself about (of course I won't add anything, it's just tiresome sketches again;). What I don't really like is the stretched bars in the general settings tab - these lines are a bit tiring. Might be worth shortening it? The second reservation concerns the position of the language setting. For me it seems to be the most important of all the settings. This is not an insurmountable problem, but if a player accidentally or out of curiosity sets a language completely unfamiliar to him, e.g. Arabic or Hebrew, he may have a bit of a problem with finding the setting. You can also think about adding a language selection icon (wikipedia and many other sites), which I think is quite widely known.

A bit off-topic at the end, but a bit not. I admit that I also thought (for a long time) about some changes in terms of settings in the game. One of the things that probably gnawed at me the most was the question: why are these settings divided into two windows? This makes some sense, but not always. **Maybe it would be worth moving all interface settings to the options window and leaving only the gameplay settings in the settings window?**
Anyway, I had another long-developing concept of combining all settings in one window. Interestingly, there are also the General, Graphics and Sound tabs here. :) In addition, there is also the Gameplay and Interface tab. I tried to think about it more comprehensively and the changes are related to other windows as well. I don't know if it's good, but maybe you should consider another arrangement?
This is what it looks like to me...
X++
1. Minimized main menu with a vertical layout inspired by the idea that TrueBrain presented some time ago. Add-ons, new games and settings tabs have been reduced to single buttons.
2. The settings window contains all game settings, including the "weird" ones that are not present today (in general, I don't think the game has too many settings, the impression of their excess is rather the result of not the best division of them), as well as keyboard shortcuts and, additionally, settings for online games regarding both the player and the server. The latter could be very useful for easier establishment, configuration and maintenance of online games. Today, setting up a server is something reserved for a rather small number of players, because it requires proper knowledge. I think it's worth simplifying it.
Changed color, buttons, details and assigned options are irrelevant here. It's mainly about the general concept of dividing all the settings in the game into five tabs.
3. One window for all add-ons. It is based on my earlier slightly developed concept. The GS and AI selection layout could remain almost unchanged, including the ability to save and select presets, which is currently not possible. Such an arrangement would also make sense in the case of GS if they could have dedicated functions (complex or only: conditions for the development of cities / conditions for the development of industry / taxes / restrictions / rewards / ...) and could be combined with each other.
4. One window to start a new game, both on random map, heightmap and scenario. The idea is a combination of the concept of adding presets by TrueBrain and the old concept of andythenorth, which assumed the possibility of selecting all settings for a new game in one window. The scenario editor removed from the main menu would be under the scenario tab. What you can see is the first finished sketch.

PeterN:
Shuffled things a little, including moving Language to the first general item.

image
image
image

I think the combined settings/options window needs thinking about, perhaps as a separate tab as what you have there is way more complex and overwhelming.

src/settings_gui.cpp
```diff
@@ -163,6 +163,18 @@ static const std::map<int, StringID> _scale_labels = {
 	{  500, STR_GAME_OPTIONS_GUI_SCALE_5X },
 };
 
+static const std::map<int, StringID> _volume_labels = {
```
LordAro, line 166: Why is this necessary now, when it was not previously? Does it need some doxyment?

LordAro (approved):
Excellent stuff. Ship it.

src/settings_gui.cpp
PeterN: To be honest it's not absolutely needed, but having the labels makes the two sliders look better (IMHO) when placed directly together.

PR #10675, opened 2023-04-18
Codechange: use string/fmt instead of printf for ShowInfo(F)

rubidium42:
## Motivation / Problem

Get rid of va_list, C-style strings and friends

## Description

Replace ShowInfo(F) with a variant that uses std::string and fmt over C-style string and vseprintf.

## Limitations

There's still plenty of va_list/C-style strings left, but small steps ;)
Don't know whether OS/2 still compiles.

## Checklist for review

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -57,7 +57,7 @@ extern int _debug_random_level;
 #endif
 
 char *DumpDebugFacilityNames(char *buf, char *last);
-void SetDebugString(const char *s, void (*error_func)(const char *));
+void SetDebugString(const char *s, void (*error_func)(const std::string &));
 const char *GetDebugString();
 
 /* Shorter form for passing filename and linenumber */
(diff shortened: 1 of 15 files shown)


src/debug.h
```diff
@@ -117,8 +117,8 @@ const char *GetDebugString();
 }
 
 
-void ShowInfo(const char *str);
-void CDECL ShowInfoF(const char *str, ...) WARN_FORMAT(1, 2);
+void ShowInfo(const std::string &str);
+#define ShowInfoF(format_string, ...) ShowInfo(fmt::format(FMT_STRING(format_string), ## __VA_ARGS__))
```
LordAro: Clearly should be templated instead of a define :p

rubidium42: The biggest caveat with templating that function is also losing `FMT_STRING`, which brings us compile time validation of the format string: https://fmt.dev/latest/api.html#compile-time-format-string-checks.

LordAro: True, though if I'm reading the docs correctly we get it for free with C++20. Might be a way off though

rubidium42: Yes, with C++20 you do not need the macro anymore. However, making it a template doesn't seem to do the trick: https://godbolt.org/z/r1Yvv4T9M
The macro version rightfully fails when trying to format a string as digit, but the template version fails because the format (string) isn't a constant expression even when the format is right.

PeterN (approved):

PR #10678, opened 2023-04-18
Codechange: use std::string for base media filename/warning storage

rubidium42:
## Motivation / Problem

Get rid of stredup/str_fmt/free for base media data.

## Description

Replace `char *` with `std::string` and propogate where appropriate.

## Limitations

The linked list of base sets should probably be changed to a vector, so there is no destructor needed anymore.

## Checklist for review

### DIFF
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -31,9 +31,9 @@ struct MD5File {
 		CR_NO_FILE,  ///< The file did not exist
 	};
 
-	const char *filename;        ///< filename
+	std::string filename;        ///< filename
 	uint8 hash[16];              ///< md5 sum of the file
-	const char *missing_warning; ///< warning when this file is missing
+	std::string missing_warning; ///< warning when this file is missing
 	ChecksumResult check_result; ///< cached result of md5 check
 
 	ChecksumResult CheckMD5(Subdirectory subdir, size_t max_size) const;
@@ -73,11 +73,6 @@ struct BaseSet {
 	/** Free everything we allocated */
 	~BaseSet()
 	{
-		for (uint i = 0; i < NUM_FILES; i++) {
-			free(this->files[i].filename);
-			free(this->files[i].missing_warning);
-		}
-
 		delete this->next;
 	}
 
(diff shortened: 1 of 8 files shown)


src/music.cpp
```diff
@@ -149,7 +149,7 @@ bool MusicSet::FillSetDetails(IniFile *ini, const char *path, const char *full_f
 				this->songinfo[i].filetype = MTT_STANDARDMIDI;
 			}
 
-			const char *trimmed_filename = filename;
+			const char *trimmed_filename = filename.c_str();
```
LordAro, line 152: Worth converting usage of `trimmed_filename` as well? Could use `std::string_view`

rubidium42: It wouldn't get much prettier. Primarily because `find` returns `npos`, so to iterate to the next you must first `find`and then compare the result to determine how to change the `string_view` meaning it's not a simple for loop anymore. Or am I missing something simpler that we could do?
I'd almost be tempted not keep a pointer, but rather the index into the string. Though that messes with the rest of the function where the pointer is still used.
So I rather leave that conversion/significant rewrite to another PR.

src/base_media_base.h
```diff
@@ -279,8 +274,8 @@ static const uint NUM_SONGS_AVAILABLE = 1 + NUM_SONG_CLASSES * NUM_SONGS_CLASS;
 static const uint NUM_SONGS_PLAYLIST  = 32;
 
 /* Functions to read DOS music CAT files, similar to but not quite the same as sound effect CAT files */
-char *GetMusicCatEntryName(const char *filename, size_t entrynum);
-byte *GetMusicCatEntryData(const char *filename, size_t entrynum, size_t &entrylen);
+char *GetMusicCatEntryName(const std::string &filename, size_t entrynum);
```
PeterN, line 277: Can this return std::string too? I see it's checked against nullptr.

rubidium42: To be functionally the same, it should return `std::optional<std::string>` as with just `std::string` an empty name will cause the 'song is missing' warning, where it did not do that before.
The main problem/issue with these PRs is that they quickly rabbit hole out of control if you keep unravelling the thread. Here it is returning a completely new string that does not have anything in common with the input, so that seems like a good place to stop the unravelling.

PeterN: Fair, I thought of separate PR material too.

PeterN (approved):
Looks fairly straight forward to me.

PR #10679, opened 2023-04-19
Codechange: replace printf-like functions with fmt in Squirrel

rubidium42:
## Motivation / Problem

Lot of old style printf functions in Squirrel.

## Description

Replace them with a variant that uses fmt::format when needed, or otherwise passes strings.
Also remove a function with printf related hack, that isn't used at all.

## Limitations

None.

## Checklist for review


src/3rdparty/squirrel/sqstdlib/sqstdaux.cpp
```diff
@@ -38,7 +39,7 @@ void sqstd_printcallstack(HSQUIRRELVM v)
 					src = si.source;
 				}
 			}
-			pf(v,"*FUNCTION [%s()] %s line [" OTTD_PRINTF64 "]\n",fn,src,si.line);
+			pf(v,fmt::format("*FUNCTION [{}()] {} line [{}]\n",fn,src,si.line));
```
PeterN, line 42: As this diverges from the old Squirrel more, do we need to keep the no-spaces-after-commas coding-style?

src/3rdparty/squirrel/squirrel/sqfuncstate.cpp
```diff
@@ -83,11 +84,11 @@ SQInstructionDesc g_InstrDesc[]={
...
 {
 	switch(type(o)){
-		case OT_STRING:	printf("\"%s\"",_stringval(o));break;
-		case OT_FLOAT: printf("{%f}",_float(o));break;
-		case OT_INTEGER: printf("{" OTTD_PRINTF64 "}",_integer(o));break;
-		case OT_BOOL: printf("%s",_integer(o)?"true":"false");break;
-		default: printf("(%s %p)",GetTypeName(o),(void*)_rawval(o));break; break; //shut up compiler
+		case OT_STRING:	fmt::print("\"{}\"",_stringval(o));break;
+		case OT_FLOAT: fmt::print("{{{}}}",_float(o));break;
+		case OT_INTEGER: fmt::print("{{{}}}",_integer(o));break;
```
PeterN, line 89: What is the `{{{}}}` doing that `{}` doesn't?

glx22: Enclose the output between `{` and `}` I think

rubidium42: `Format strings contain “replacement fields” surrounded by curly braces {}. Anything that is not contained in braces is considered literal text, which is copied unchanged to the output. If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}.` (https://fmt.dev/latest/syntax.html)

src/3rdparty/squirrel/sqstdlib/sqstdaux.cpp
rubidium42: I tried to make as few changes as possible to make reviews easier.
It might be useful to "autoformat" all the squirrel code once with the updated formatting rules, to see whether that really helps. Then we might also consider formatting all the code with it (semi-automatically), i.e. run it through format in a CI step so it can complain about issues.

src/3rdparty/squirrel/squirrel/sqfuncstate.cpp
PeterN: Gotcha, it was already wrapped by literal { }.

PeterN (approved):
Looks thorough to me, I can delete my own half-hearted branch that started on some of this.

PR #10680, opened 2023-04-19
Fix #10600: 'Replace Vehicles' shows numbers >999

### LINKED ISSUE #10600: [Bug]: "Replace vehicles" doesn't show number greater than 999
### Version of OpenTTD

13.0 Windows 10

### Expected result

The number of vehicles is displayed with an accuracy of one
изображение

### Actual result

Only the first thousand digit is displayed, on some fonts – just « . . . »

Tahoma (and others not monospaced):
изображение 
изображение

Consolas (and others monospaced):
изображение Consolas

Built-in openGFX sprite font:
изображение

But the second digit does not appear on the original baseset sprite font:
изображение

### Steps to reproduce

1. Create more than 999 vehicles/wagons of the same model
2. Open the corresponding replacement menu

1. Or open the wagon replacement menu in 10k.sav [999.zip](https://github.com/OpenTTD/OpenTTD/files/11154272/999.zip)

carlossss111:
## Motivation / Problem

Fixes the digit display issue in the 'Replace Vehicles' window.
Closes #10600 

## Description

Up to 6 digits can now be displayed in the 'Replace Vehicles' window.

## Limitations

Does not allow more than 6 digits.

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1026,7 +1026,15 @@ void DrawEngineList(VehicleType type, const Rect &r, const GUIEngineList &eng_li
 	int count_width = 0;
 	if (show_count) {
 		replace_icon = GetSpriteSize(SPR_GROUP_REPLACE_ACTIVE);
-		SetDParamMaxDigits(0, 3, FS_SMALL);
+
+		uint biggest_num_engines = 0;
+		for (auto i = min; i < max; i++) {
+			const auto &item = eng_list[i];
+			const uint num_engines = GetGroupNumEngines(_local_company, selected_group, item.engine_id);
+			biggest_num_engines = std::max(biggest_num_engines, num_engines);
+		}
+
+		SetDParam(0, biggest_num_engines);
 		count_width = GetStringBoundingBox(STR_JUST_COMMA, FS_SMALL).width;
 	}
 


PeterN:
There are game settings for max units of each type per player, and the limit for that is 5,000. So I think this only needs to be increased to 4 instead of 6.

glx22:
The limit only includes consists IIRC, #10600 shows there could be 10000+ wagons

PeterN:
Ah of course, sorry!

2TallTyler (dismissed):
With this change, the count width is increased but not the overall width of the window. This means that some vehicle names which currently fit, get cut off.

This PR:
changed

13.1:
vanilla

I don't see this as a problem. In 13.1, other vehicles already don't fit:
engines

PeterN:
Yeah, that was my main concern about just raising it from 3 to 6.

LordAro:
I like the idea of automatic width, but would obviously require some arbitrary upper limit. On the other hand, perhaps the padding on the right side of the number seems a bit high, could be reduced?

PeterN:
That's space left for the replace icon if the engine type is being replaced.

glx22:
Space on the left side is also for an icon ?

carlossss111:
I think it would make most sense to just increase the window width since it's only by a small amount. If people are happy with that I'll do that.

2TallTyler:
Hmm, still too narrow for me. I'm testing in the Preview build, so I don't _think_ it's saved a default window size, but maybe others can test?

narrow

TrueBrain:
> idk, something as "fundamental" as "Passenger Carriage" seems like it should fit regardless...
> 
> (yes, i know that doesn't really cover other languages anyway)

The default (no NewGRFs) is `Passenger Car`, which fits fine :)

When looking at this PR, I was like: why don't we just use the room we need, instead of always reserving XXX. So I came up with:

```cpp
		uint biggest_num_engines = 0;
		for (auto i = min; i < max; i++) {
			const auto &item = eng_list[i];
			const uint num_engines = GetGroupNumEngines(_local_company, selected_group, item.engine_id);
			biggest_num_engines = std::max(biggest_num_engines, num_engines);
		}

		SetDParamMaxDigits(0, std::to_string(biggest_num_engines).length(), FS_SMALL);
		count_width = GetStringBoundingBox(STR_TINY_BLACK_COMA).width;
```

Not sure what others think about something like this. But then at least if you don't go crazy on the amount of vehicles you have, you have more space to show actual useful stuff. And for people who go this bananas .. I am sure they know by now what a passenger carriage is and what not :D

2TallTyler:
`Passenger Car` is US English, `Passenger Carriage` is UK English. 😛 

I agree that sizing the interface to the actual largest count using your code snippet is a better idea than picking an arbitrary max. @carlossss111, your thoughts?

TrueBrain:
@carlossss111 Would you be up for making the above suggested change? If not, that is also perfectly fine ofc, just checking in :) Tnx!

TrueBrain:
No worries; I will see if I can bring it home :) Tnx for getting back to us!

glx22 (approved):
Using the needed space is better than reserving huge space just in case.

PR #10684, opened 2023-04-20
Codechange: replace error/usererror printf variant with fmt variant and rename

rubidium42:
## Motivation / Problem

The functions error/usererror using C-style printf with all the difficulty with different formatting codes for different compilers/operating systems.
Furthermore the naming of these functions is not as what the coding style prescribes.

## Description

Replace printf constructs with fmt::format.
Rename usererror to UserError, and error to FatalError to not conflict with existing function names.

Due to the location where UserError and FatalError are defined, namely stdafx.h, this means fmt/format.h needs to be included there as well. For this reason these have been moved to a separate header (error_func.h). To prevent moving NOT_REACHED (used in ~150 files) and assert (used in ~375 files) to a separate header, two helper functions have been added to offload the formatting to a .cpp file.

Including fmt::format in stdafx.h is not really an option as that causes the Emscripten build to get killed during linking and CodeQL runs into an internal compiler error.

## Limitations

Few occasions of OTTD_PRINTF & friends remaining, but together with #10675, #10678, #10679 and #10680 the remaining amount is tiny, and for another "cleanup" PR that takes care of the last small bits.

## Checklist for review

### DIFF
--- a/src/core/alloc_func.cpp
+++ b/src/core/alloc_func.cpp
@@ -9,6 +9,8 @@
 
 #include "../stdafx.h"
 
+#include "../error_func.h"
+
 #include "../safeguards.h"
 
 /**
@@ -17,7 +19,7 @@
  */
 void NORETURN MallocError(size_t size)
 {
-	error("Out of memory. Cannot allocate " PRINTF_SIZE " bytes", size);
+	FatalError("Out of memory. Cannot allocate {} bytes", size);
 }
 
 /**
@@ -26,5 +28,5 @@ void NORETURN MallocError(size_t size)
  */
 void NORETURN ReallocError(size_t size)
 {
-	error("Out of memory. Cannot reallocate " PRINTF_SIZE " bytes", size);
+	FatalError("Out of memory. Cannot reallocate {} bytes", size);
 }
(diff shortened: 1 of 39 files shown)


src/core/alloc_func.cpp
```diff
@@ -26,5 +28,5 @@ void NORETURN MallocError(size_t size)
  */
 void NORETURN ReallocError(size_t size)
 {
-	error("Out of memory. Cannot reallocate " PRINTF_SIZE " bytes", size);
+	FatalError("Out of memory. Cannot reallocate {} bytes", size);
```
PeterN, line 31: I think `PRINTF_SIZE` in stdafx.h is no longer needed.

rubidium42: Since some of the other PRs got merged it's indeed not needed anymore. I've got a (local) branch that removes this one and its friends in stdafx.h together with the final bits. However, that depends on this and the other PRs.

But since it conflicted with landscape.cpp I had to manually rebase, so I removed `PRINTF_SIZE` as well.

PeterN (approved):

PR #9577, opened 2021-09-26
Feature: Multi-tile depots

J0anJosep:
## Motivation / Problem
Replacing thousands of old steam trains with monorail replacements is tiresome.
Also, if trains have a "Go to depot X", several trains may be waiting for entering a depot if one train is already entering or leaving the depot.

## Description
This PR allows joining several depot tiles for the same vehicle type into one single depot. For joining them, I have tried to use the same criteria as for joining station tiles.

Replacements with different railtypes are enabled with the option "allow replacing rail vehicles with incompatible railtype". The same applies for road vehicles. Then, users can build a monorail depot tile joined to a normal rail type depot and replacements are much easier.
replacement

If trains get stuck when trying to enter a depot, add more tiles to the depot for easing it.
depot_interaction

#8480 adds extended depots (which are not included here) and also contains the commits of this PR.

## Limitations

If a hangar or depot has several tiles, user cannot control in which tile a new vehicle is bought. (On the other hand, the hangar window shows all vehicles in all hangars of the airport: I see this as something good, as in this way it is more difficult to miss stopped aircraft in an airport.)

I don't know whether this feature can be accepted. It adds more settings and getting used to this feature takes time.

The PR touches thousands of lines of code. There are some things I am not happy with yet and I don't have much time to revise it. Code reviews of this PR can help me to improve it, as it will hint me which parts I need to work on.

## Possible To-do (discarded by now)

- Check exploit when rearranging train chains. (add cost for moving vehicles inside the depot/move train chain to some tile to make the exploit more difficult). This can also happen when transferring cargo from one train to another using a station. But as loading/unloading takes time, it is difficult to exploit it.
- When buying a new train vehicle and appending it to an existing chain, keep the tile of the existing chain. Add no cost.
- When issuing a move train chain command, add a cost for vehicles that teleport (distance + heightlevelincrease).
Note: leaving the depot at a speed of 0 is a penalty to consider, but maybe not enough for steep landscapes.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? **No** (label: 'backport requested')
* This PR touches english.txt or translations? **Yes** Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? **Yes** (label 'savegame upgrade')
* This PR affects the GS/AI API? **Yes** (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
[...]

PR #10689, opened 2023-04-21
Add: New rectangle place method: fixed and limited side lengths.

J0anJosep:
## Motivation / Problem

This PR is related to #9577 

There is no place method for selecting a rectangle of tiles where one side has a fixed length and the other side has a variable and limited length.

In other words, to make it possible to build 7 ship depots, one next to the other, it should be possible to draw a rectangle with a fixed size of 2 in one side (as ship depots have a length of 2 tiles) and then move the mouse to select the desired tiles (a 2x7 rectangle).
The other side, the one that is not fixed, is limited to the depot spread limit.

This commit adds this new place method.

## Checklist for review

### DIFF
--- a/src/tilehighlight_func.h
+++ b/src/tilehighlight_func.h
@@ -26,6 +26,8 @@ void VpStartDragging(ViewportDragDropSelectionProcess process);
 void VpStartPlaceSizing(TileIndex tile, ViewportPlaceMethod method, ViewportDragDropSelectionProcess process);
 void VpSetPresizeRange(TileIndex from, TileIndex to);
 void VpSetPlaceSizingLimit(int limit);
+void VpSetPlaceFixedSize(byte fixed_size);
+void VpResetFixedSize();
 
 void UpdateTileSelection();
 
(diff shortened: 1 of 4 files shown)


2TallTyler (commented):
Sorry, I think I failed to explain myself fully. Each PR needs to stand on its own without needing a future PR to have a purpose. This commit should probably be in the PR that adds multi-tile depots of the same rail/roadtype, if I'm correct that this would be its first use. Until then, it's orphaned code, so it should stay with the commit which gives it purpose. 

The other PR you just opened for using DepotIDs is perfect.

J0anJosep:
> Sorry, I think I failed to explain myself fully. Each PR needs to stand on its own without needing a future PR to have a purpose. This commit should probably be in the PR that adds multi-tile depots of the same rail/roadtype, if I'm correct that this would be its first use. Until then, it's orphaned code, so it should stay with the commit which gives it purpose.
> 
> The other PR you just opened for using DepotIDs is perfect.

Ok. But this commit could also be used for click-and-drag for building several docks. So it could be used for other purposes, not only depots.

2TallTyler:
Yes, absolutely, but that functionality would need to be implemented in the same PR — which you could absolutely add here, if you like.

Issue #8054, opened 2020-04-02
Automatic spacing in timetables not working


arbulgazar:
Automatic spacing introduced since version 1.4.0 is not working and `ctrl-Start Date` clicking does not result in start date being offset for the vehicles sharing the same order.
![Screenshot from 2020-04-02 17-38-20](https://user-images.githubusercontent.com/12510654/78276450-f3e1b680-750a-11ea-927d-fa93f861c80f.png)
## Version of OpenTTD
openttd-1.10.0-linux-ubuntu-bionic-amd64.deb
## Expected result
The start date of each vehicle is increased by the "gap number" so that each vehicle starts the same number of days after the last.
## Actual result
Only the first vehicle in the list (the one that "Start Date" was applied to) gets the start date. No other vehicles in the shared order get the start date. Vehicles end up in a usual catarpillar-style stacking after any amount of time.
## Steps to reproduce
From [wiki (see last section)](https://wiki.openttd.org/Timetable):

1. Set up a vehicle and plot the route using the Orders tab.
2. When complete, click on the Timetable button and click Autofill.
3. Start the vehicle and let it complete the route and fill in the timetable.
4. Stop the vehicle as soon as the last item has filled in the timetable.
5. On the Orders tab, click Skip until the next station is the first station in the list.
6. Unclick Autofill and add a little slack into the timetable to account for filling up at stations. Since this is generally the first time you visit each station, there are no passengers to load, and you need to increase the time so that the vehicle can load up.
7. Once the timetable is complete, decide how many vehicles you want on that route.
8. Clone the initial vehicle using ctrl-click. The cloned vehicles will share the order list and timetable.
9. Once you have all the clones, select the first vehicle and ctrl-click on Start Date.
10. Set a start date a few days into the future. Ideally, the vehicle should arrive at the first station before this time.
11. Release all the vehicles. They will bunch up at the first station but should leave in regular intervals until they are all spaced out.

arbulgazar:
Adding a save file with a minimalistic example, but this problem is consistent across all games.
[timetable_bug_save.zip](https://github.com/OpenTTD/OpenTTD/files/4422447/timetable_bug_save.zip)

arbulgazar:
[...]

PR #10690, opened 2023-04-21
Fix #10289: Don't silently fail when setting timetable start dates

### LINKED ISSUE #10289: Setting timetable start dates for shared orders silently fails if not fully timetabled
### Version of OpenTTD

13.0-beta2

### Expected result

When setting timetable start dates for multiple vehicle with shared orders, Ctrl+clicking the Start Date button spaces out vehicles along the route. 

If the entire route is not timetabled, we don’t know the full duration and therefore can’t space our vehicles evenly. 

In this case I would expect it to either spread out vehicles using the known incomplete total, or throw an error message to the player. 

### Actual result

The command fails silently and only sets the start date of the selected vehicle, leading to misunderstandings like #8054. 

@nchappe explains in #8054:
> In the code the exact condition to set the timetable start date for all the vehicles with the same shared orders when the set start date button is pressed is v->orders->IsCompleteTimetable() && _ctrl_pressed, which means that if some orders are not explicitly timetabled, Ctrl+Click silently falls back to setting the timetable start date of a single vehicle, with no user feedback.
This is explained in the relevant tooltip ("Ctrl+Click distributes all vehicles sharing this order evenly from the given date based on their relative order, if the order is completely timetabled"), but I still find it a bit counter-intuitive, and I suspect this is the root of this issue.

### Steps to reproduce

1. Create more than one vehicle with shared orders
2. Timetable most but not all of their orders
3. Ctrl+click Start Date to space them out over the route
4. Notice that only the selected vehicle gets its start date changed

2TallTyler:
## Motivation / Problem

Copied from #10289:

### Expected result

When setting timetable start dates for multiple vehicle with shared orders, Ctrl+clicking the Start Date button spaces out vehicles along the route. 

If the entire route is not timetabled, we don’t know the full duration and therefore can’t space our vehicles evenly. 

In this case I would expect it to either spread out vehicles using the known incomplete total, or throw an error message to the player. 

### Actual result

The command fails silently and only sets the start date of the selected vehicle, leading to misunderstandings like #8054. 

@nchappe explains in #8054:
> In the code the exact condition to set the timetable start date for all the vehicles with the same shared orders when the set start date button is pressed is v->orders->IsCompleteTimetable() && _ctrl_pressed, which means that if some orders are not explicitly timetabled, Ctrl+Click silently falls back to setting the timetable start date of a single vehicle, with no user feedback.
This is explained in the relevant tooltip ("Ctrl+Click distributes all vehicles sharing this order evenly from the given date based on their relative order, if the order is completely timetabled"), but I still find it a bit counter-intuitive, and I suspect this is the root of this issue.

### Steps to reproduce

1. Create more than one vehicle with shared orders
2. Timetable most but not all of their orders
3. Ctrl+click Start Date to space them out over the route
4. Notice that only the selected vehicle gets its start date changed

## Description

Throw an error instead of silently failing.

The `CmdSetTimetableStart()` command is already supposed to throw an error if the timetable is not complete, but the check is being performed in the GUI click handler before calling the command, which incorrectly passes the `timetable_all` parameter as false.

I've removed the incorrectly-placed check and added a descriptive error to tell the player why it's not working.

Closes #10289.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -5110,6 +5110,7 @@ STR_ERROR_NO_BUOY                                               :{WHITE}There is
 STR_ERROR_CAN_T_TIMETABLE_VEHICLE                               :{WHITE}Can't timetable vehicle...
 STR_ERROR_TIMETABLE_ONLY_WAIT_AT_STATIONS                       :{WHITE}Vehicles can only wait at stations
 STR_ERROR_TIMETABLE_NOT_STOPPING_HERE                           :{WHITE}This vehicle is not stopping at this station
+STR_ERROR_TIMETABLE_INCOMPLETE                                  :{WHITE}... timetable is incomplete
 
 # Sign related errors
 STR_ERROR_TOO_MANY_SIGNS                                        :{WHITE}... too many signs
--- a/src/timetable_cmd.cpp
+++ b/src/timetable_cmd.cpp
@@ -306,7 +306,7 @@ CommandCost CmdSetTimetableStart(DoCommandFlag flags, VehicleID veh_id, bool tim
 	if (start_date < 0 || start_date > MAX_DAY) return CMD_ERROR;
 	if (start_date - _date > MAX_TIMETABLE_START_YEARS * DAYS_IN_LEAP_YEAR) return CMD_ERROR;
 	if (_date - start_date > DAYS_IN_LEAP_YEAR) return CMD_ERROR;
-	if (timetable_all && !v->orders->IsCompleteTimetable()) return CMD_ERROR;
+	if (timetable_all && !v->orders->IsCompleteTimetable()) return CommandCost(STR_ERROR_TIMETABLE_INCOMPLETE);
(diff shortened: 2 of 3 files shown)


PeterN:
Looks like this is this all client-side? Do AIs have the ability to do this timetabling?

rubidium42:
> Looks like this is this all client-side? Do AIs have the ability to do this timetabling?

No(t yet).

2TallTyler:
Right, the other option is for the command itself to return an error. Any preference?

PeterN:
I think it's best for the command to fail instead of silently doing something other than asked.

Can still be done client-side as well if that is not too much duplication and allows a cleaner experience.

2TallTyler:
Thanks. I'll look into that later today or tomorrow.

2TallTyler:
On further investigation, the command is already supposed to throw an error if the timetable is not complete, but the check is being performed before calling the command and incorrectly passing the `timetable_all` parameter as false.

I've removed the incorrectly-placed check and kept the descriptive error to tell the player why it's not working.

PeterN (approved):
Looks good to me.

PR #8480, opened 2021-01-02
Feature: Extended depots

J0anJosep:
(Depends on #9577)

## Motivation / Problem
It is strange a train with a length of 6 tiles can get out from a 1-tile depot.
Or that you can have 300 road vehicles in a single tile depot.
Or that 300 ships can be stopped in the same depot tile.

## Description
This PR introduces a new depot tile for rail, road and water transport: extended depots.

Standard depots are the legacy depots of OpenTTD. They should work the same as before.

Extended depots have limitations related to the amount of vehicles they can handle. **It is possible to join extended depots with other extended or standard depots.** 

**How to use it:**
- In settings, you can choose which depot types the human players can use (search "depot types").

## Limitations
This is for testing purpose.
There are several things to be improved and I need feedback: wording, functionality, implementation,...

It needs testing and feedback.
**Please, expect some crashes. Report them, please.**
Try also to crash it in:
- Multiplayer using NewGRFs
- Replacing/autorenewing/autoreplacing

## To consider
- Consider which new graphics are needed and possible NewGRF support.
- Platform length for stations and depots can be stored in the map array, so pathfinders can benefit from it. (I consider it out of scope by now but it could be done.)
- Related with previous item: Station::GetPlatformLength is changed to a generic GetPlatformLength. I am not really sure if this is the way to go. Help will be appreciated in this one.
- Hangars: airports with multiple hangars currently have different depot windows for different hangar tiles and you can build aircraft in each of those tiles; with the current implementation of multi-tile depots, the hangars of an airport share the same window and newly built aircraft is placed only in the first hangar. This works, but it is not the current vanilla behavior. Check with the community/developers whether this is acceptable or not. (Having in mind tileable airports, the new behavior may be a better approach than the current one.)
- How these changes affect AI API? Right now AI are able to build small depots, independently from the settings. Keeping this behavior is the easiest way to deal with AIs and big depots. Exposing big depots to AI is not an easy task, as they should deal with cross-replacements and max train length in platform depot.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* **label savegame upgrade**
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

[...]

PR #10691, opened 2023-04-21
Change: Add Depots and DepotIDs for airports with hangars.

J0anJosep:
## Motivation / Problem

This PR is related to #9577 

If depots can have more than one tile (as intended with #9577), then the depot window number should be a DepotID (as the TileIndex corresponding to a depot won't be unique anymore).

Then, depot structs for hangars should also be available for consistency and dealing with order backups later and "Go To Hangar" orders. This implies that some old specific code for goto hangar orders will be removed, as some cases will now be dealt with standard cases involving depot construction/demolition and when loading old savegames.

Also, when an airport is removed, their goto depot orders are also removed. This could be a problem, as the immediate removal of these orders is not always desired (rebuilding an airport). In subsequent commits, when a depot is deleted, it will stay for a while as a ghost (to keep orders associated to it) and eventually will be removed forever (behaving in a similar way to station removal).

If modular airports or other changes to airports are made, it is possible that storing a DepotID, as this PR suggests, may be useful.

## Description

- Each airport with at least a hangar tile creates a depot object. The airport stores a pointer to the depot.
- "Go to hangar" orders store the DepotID instead of the StationID.

## Checklist for review

### DIFF
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -41,6 +41,7 @@
 #include "framerate_type.h"
 #include "aircraft_cmd.h"
 #include "vehicle_cmd.h"
+#include "depot_base.h"
 
 #include "table/strings.h"
 
@@ -136,7 +137,7 @@ static StationID FindNearestHangar(const Aircraft *v)
 		if (v->current_order.IsType(OT_GOTO_STATION) ||
 				(v->current_order.IsType(OT_GOTO_DEPOT) && (v->current_order.GetDepotActionType() & ODATFB_NEAREST_DEPOT) == 0)) {
 			last_dest = Station::GetIfValid(v->last_station_visited);
-			next_dest = Station::GetIfValid(v->current_order.GetDestination());
+			next_dest = Station::GetIfValid(GetTargetDestination(v->current_order, true));
 		} else {
 			last_dest = GetTargetAirportIfValid(v);
 			next_dest = Station::GetIfValid(v->GetNextStoppingStation().value);
@@ -407,9 +408,10 @@ ClosestDepot Aircraft::FindClosestDepot()
 		if (station == INVALID_STATION) return ClosestDepot();
 
 		st = Station::Get(station);
+		assert(st->airport.hangar != nullptr);
 	}
 
-	return ClosestDepot(st->xy, st->index);
+	return ClosestDepot(st->xy, st->airport.hangar->index);
 }
 
 static void CheckIfAircraftNeedsService(Aircraft *v)
@@ -424,13 +426,13 @@ static void CheckIfAircraftNeedsService(Aircraft *v)
 	 * we don't want to consider going to a depot too. */
 	if (!v->current_order.IsType(OT_GOTO_DEPOT) && !v->current_order.IsType(OT_GOTO_STATION)) return;
 
-	const Station *st = Station::Get(v->current_order.GetDestination());
+	const Station *st = Station::Get(GetTargetDestination(v->current_order, true));
 
 	assert(st != nullptr);
 
 	/* only goto depot if the target airport has a depot */
 	if (st->airport.HasHangar() && CanVehicleUseStation(v, st)) {
-		v->current_order.MakeGoToDepot(st->index, ODTFB_SERVICE);
+		v->current_order.MakeGoToDepot(st->airport.hangar->index, ODTFB_SERVICE);
 		SetWindowWidgetDirty(WC_VEHICLE_VIEW, v->index, WID_VV_START_STOP);
 	} else if (v->current_order.IsType(OT_GOTO_DEPOT)) {
 		v->current_order.MakeDummy();
@@ -892,7 +894,7 @@ static bool AircraftController(Aircraft *v)
 		/* Jump into our "holding pattern" state machine if possible */
 		if (v->pos >= afc->nofelements) {
 			v->pos = v->previous_pos = AircraftGetEntryPoint(v, afc, DIR_N);
-		} else if (v->targetairport != v->current_order.GetDestination()) {
+		} else if (v->targetairport != GetTargetDestination(v->current_order, true)) {
 			/* If not possible, just get out of here fast */
 			v->state = FLYING;
 			UpdateAircraftCache(v);
@@ -1449,7 +1451,7 @@ static void AircraftLandAirplane(Aircraft *v)
 void AircraftNextAirportPos_and_Order(Aircraft *v)
 {
 	if (v->current_order.IsType(OT_GOTO_STATION) || v->current_order.IsType(OT_GOTO_DEPOT)) {
-		v->targetairport = v->current_order.GetDestination();
+		v->targetairport = GetTargetDestination(v->current_order, true);
 	}
 
 	const Station *st = GetTargetAirportIfValid(v);
@@ -1539,7 +1541,7 @@ static void AircraftEventHandler_InHangar(Aircraft *v, const AirportFTAClass *ap
 		return;
 
 	/* We are leaving a hangar, but have to go to the exact same one; re-enter */
-	if (v->current_order.IsType(OT_GOTO_DEPOT) && v->current_order.GetDestination() == v->targetairport) {
+	if (v->current_order.IsType(OT_GOTO_DEPOT) && GetTargetDestination(v->current_order, true) == v->targetairport) {
 		VehicleEnterDepot(v);
 		return;
 	}
@@ -1548,7 +1550,7 @@ static void AircraftEventHandler_InHangar(Aircraft *v, const AirportFTAClass *ap
 	if (AirportHasBlock(v, &apc->layout[v->pos], apc)) return;
 
 	/* We are already at the target airport, we need to find a terminal */
-	if (v->current_order.GetDestination() == v->targetairport) {
+	if (GetTargetDestination(v->current_order, true) == v->targetairport) {
 		/* FindFreeTerminal:
 		 * 1. Find a free terminal, 2. Occupy it, 3. Set the vehicle's state to that terminal */
 		if (v->subtype == AIR_HELICOPTER) {
@@ -1599,7 +1601,7 @@ static void AircraftEventHandler_AtTerminal(Aircraft *v, const AirportFTAClass *
 		case OT_GOTO_STATION: // ready to fly to another airport
 			break;
 		case OT_GOTO_DEPOT:   // visit hangar for servicing, sale, etc.
-			go_to_hangar = v->current_order.GetDestination() == v->targetairport;
+			go_to_hangar = GetTargetDestination(v->current_order, true) == v->targetairport;
 			break;
 		case OT_CONDITIONAL:
 			/* In case of a conditional order we just have to wait a tick
@@ -2103,7 +2105,7 @@ static bool AircraftEventHandler(Aircraft *v, int loop)
 		/* Check the distance to the next destination. This code works because the target
 		 * airport is only updated after take off and not on the ground. */
 		Station *cur_st = Station::GetIfValid(v->targetairport);
-		Station *next_st = v->current_order.IsType(OT_GOTO_STATION) || v->current_order.IsType(OT_GOTO_DEPOT) ? Station::GetIfValid(v->current_order.GetDestination()) : nullptr;
+		Station *next_st = Station::GetIfValid(GetTargetDestination(v->current_order, true));
 
 		if (cur_st != nullptr && cur_st->airport.tile != INVALID_TILE && next_st != nullptr && next_st->airport.tile != INVALID_TILE) {
 			uint dist = DistanceSquare(cur_st->airport.tile, next_st->airport.tile);
@@ -2169,18 +2171,7 @@ void UpdateAirplanesOnNewStation(const Station *st)
 		if (!v->IsNormalAircraft() || v->targetairport != st->index) continue;
 		assert(v->state == FLYING);
 
-		Order *o = &v->current_order;
-		/* The aircraft is heading to a hangar, but the new station doesn't have one,
-		 * or the aircraft can't land on the new station. Cancel current order. */
-		if (o->IsType(OT_GOTO_DEPOT) && !(o->GetDepotOrderType() & ODTFB_PART_OF_ORDERS) && o->GetDestination() == st->index &&
-				(!st->airport.HasHangar() || !CanVehicleUseStation(v, st))) {
-			o->MakeDummy();
-			SetWindowWidgetDirty(WC_VEHICLE_VIEW, v->index, WID_VV_START_STOP);
-		}
 		v->pos = v->previous_pos = AircraftGetEntryPoint(v, ap, rotation);
 		UpdateAircraftCache(v);
 	}
-
-	/* Heliports don't have a hangar. Invalidate all go to hangar orders from all aircraft. */
-	if (!st->airport.HasHangar()) RemoveOrderFromAllVehicles(OT_GOTO_DEPOT, st->index, true);
 }
--- a/src/depot_map.h
+++ b/src/depot_map.h
@@ -43,16 +43,21 @@ inline bool IsDepotTile(Tile tile)
 	return IsRailDepotTile(tile) || IsRoadDepotTile(tile) || IsShipDepotTile(tile) || IsHangarTile(tile);
 }
 
+extern DepotID GetHangarIndex(TileIndex t);
+
 /**
  * Get the index of which depot is attached to the tile.
  * @param t the tile
- * @pre IsRailDepotTile(t) || IsRoadDepotTile(t) || IsShipDepotTile(t)
+ * @pre IsDepotTile(t)
  * @return DepotID
  */
 inline DepotID GetDepotIndex(Tile t)
 {
-	/* Hangars don't have a Depot class, thus store no DepotID. */
-	assert(IsRailDepotTile(t) || IsRoadDepotTile(t) || IsShipDepotTile(t));
+	assert(IsDepotTile(t));
+
+	/* Hangars don't store depot id on m2. */
+	if (IsTileType(t, MP_STATION)) return GetHangarIndex(t);
+
 	return t.m2();
 }
 
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -650,7 +650,7 @@ TileIndex Order::GetLocation(const Vehicle *v, bool airport) const
 
 		case OT_GOTO_DEPOT:
 			if (this->GetDestination() == INVALID_DEPOT) return INVALID_TILE;
-			return (v->type == VEH_AIRCRAFT) ? Station::Get(this->GetDestination())->xy : Depot::Get(this->GetDestination())->xy;
+			return Depot::Get(this->GetDestination())->xy;
 
 		default:
 			return INVALID_TILE;
@@ -683,6 +683,28 @@ uint GetOrderDistance(const Order *prev, const Order *cur, const Vehicle *v, int
 	if (prev_tile == INVALID_TILE || cur_tile == INVALID_TILE) return 0;
 	return v->type == VEH_AIRCRAFT ? DistanceSquare(prev_tile, cur_tile) : DistanceManhattan(prev_tile, cur_tile);
 }
+
+/**
+ * Get the station or depot index associated to an order of a vehicle.
+ * For aircraft, it will return the index of the associated station, even for go to hangar orders.
+ * @param o Order to check.
+ * @param is_aircraft Whether the order is of an aircraft vehicle.
+ * @return index associated to a station or depot, or INVALID_STATION.
+ */
+DestinationID GetTargetDestination(const Order &o, bool is_aircraft)
+{
+	DestinationID destination_id = o.GetDestination();
+	switch (o.GetType()) {
+		case OT_GOTO_STATION:
+			return destination_id;
+		case OT_GOTO_DEPOT:
+			assert(Depot::IsValidID(destination_id));
+			return is_aircraft ? GetStationIndex(Depot::Get(destination_id)->xy) : destination_id;
+		default:
+			return INVALID_STATION;
+	}
+}
+
 
 /**
  * Add an order to the orderlist of a vehicle.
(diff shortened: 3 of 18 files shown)


src/depot_map.h
```diff
@@ -43,16 +43,21 @@ static inline bool IsDepotTile(Tile tile)
...
  * @return DepotID
  */
 static inline DepotID GetDepotIndex(Tile t)
 {
-	/* Hangars don't have a Depot class, thus store no DepotID. */
-	assert(IsRailDepotTile(t) || IsRoadDepotTile(t) || IsShipDepotTile(t));
+	assert(IsDepotTile(t));
+
+	/* Hangars don't store depot id on m2. */
+	if (IsTileType(t, MP_STATION)) return GetHangarIndex(t);
```
rubidium42, line 59: Why not make hangars proper depots with the depot index in their tile? That would make things more uniform.

src/order_cmd.cpp
```diff
@@ -725,6 +725,29 @@ uint GetOrderDistance(const Order *prev, const Order *cur, const Vehicle *v, int
...
+/**
+ * Get the station or depot index associated to an order of a vehicle.
+ * For aircraft, it will return the index of the associated station, even for go to hangar orders.
+ * @param v Vehicle
+ * @param o Order to check.
+ * @return index associated to a station or depot, or INVALID_.
+ */
+DestinationID GetTargetDestination(const Vehicle *v, const Order *o)
+{
+	assert(v != nullptr && o != nullptr);
```
rubidium42: If they may not be nullptr, why not pass them as reference?

src/saveload/station_sl.cpp
```diff
@@ -645,6 +645,7 @@ class SlStationNormal : public DefaultSaveLoadHandler<SlStationNormal, BaseStati
 		SLE_CONDVAR(Station, airport.layout,             SLE_UINT8,                 SLV_145, SL_MAX_VERSION),
 		    SLE_VAR(Station, airport.flags,              SLE_UINT64),
 		SLE_CONDVAR(Station, airport.rotation,           SLE_UINT8,                 SLV_145, SL_MAX_VERSION),
+		SLE_CONDVAR(Station, airport.depot_id,           SLE_UINT16,       SLV_MTILE_DEPOTS_ADD_HANGARS, SL_MAX_VERSION),
```
rubidium42: Why not use a pointer for depot? That would require introducing a REF_DEPOT just like REF_STATION or REF_TOWN, but it would allow for mostly automatic saveload checking of the depot's validity.

src/station.cpp
```diff
@@ -717,6 +718,33 @@ Money AirportMaintenanceCost(Owner owner)
 	return total_cost >> 3;
 }
 
+/**
+ * Create or delete the hangar corresponding to this airport.
+ * @param create if true, create a depot associated to this airport;
+ *               if false, delete the hangar corresponding to this airport.
+ */
+void Airport::SetHangar(bool create)
```
rubidium42: Why not split this up in two separate functions? An `AddHangar` and `RemoveHangar`. All I see is hardcoded `true` or `false` parameters, so it sounds like there is little need for it to be the same function.

src/vehicle.cpp
```diff
@@ -1529,6 +1529,7 @@ void VehicleEnterDepot(Vehicle *v)
 	}
 	SetWindowDirty(WC_VEHICLE_VIEW, v->index);
 
+	DepotID depot_id = GetDepotIndex(v->tile);
```
rubidium42: Why determine it here, when it's only used in one place later on within an if-statement and all kinds of other checks? Seems to me it's only needed when coming to that particular comparison, or does this have important side effects?

src/aircraft_cmd.cpp
```diff
@@ -399,9 +400,11 @@ ClosestDepot Aircraft::FindClosestDepot()
 		if (station == INVALID_STATION) return ClosestDepot();
 
 		st = Station::Get(station);
+		assert(st->airport.depot_id != INVALID_DEPOT);
 	}
 
-	return ClosestDepot(st->xy, st->index);
+	assert(st != nullptr);
```
rubidium42: `st` can't be `nullptr` here, because if the branch wasn't taken `st` isn't `nullptr` and in the branch you already dereference `st`, so it can't be `nullptr` after that either.

rubidium42 (commented):
I like the direction/idea, but in shouldn't hangars be proper "stand alone" depots? So, a depot order always has a DepotID, and a depot is not a MP_STATION tile, etc.

src/aircraft_cmd.cpp
J0anJosep: Ok

src/vehicle.cpp
J0anJosep: On subsequent commits of #9577, depot_id is to be used more times. In fact, when the depot window numbers are changed, depot_id is used in line 1536 and probably more times through these lines.
As I have been rebasing this changes for some time, it was easier to store the depot_id here to resolve conflicts.
I can do as you suggest, but it is just adding more work.

Just let me know whether I should do the change you proposed.

src/depot_map.h
J0anJosep: The depot id for rail, road and ship depots is stored in m2, as the station id.
As airport hangars are part of the airport, they should store the station id and the depot id.
When I wrote this, there was no available space for storing it in the tile. Storing the depot_id of hangars in m8 is now possible, but then it is not fully uniform, as some depot_ids will be stored in m2 and some others in m8.

I also thought about storing it in m8 some time ago, but as it is now (using GetHangarIndex) seems ok to me. It may look ugly, but having it under Station::Airport::DepotID is useful enough. Even if the depot_id was also stored in the map array, it is usually better to have it under Airport::DepotID, as it is something aircraft will need while they probably have a pointer to the corresponding station. Moreover, while toying with the idea of modular airports, the more free bits on the map array, the better.

J0anJosep:
> I like the direction/idea, but in shouldn't hangars be proper "stand alone" depots? So, a depot order always has a DepotID, and a depot is not a MP_STATION tile, etc.

Hangars are kind of hardcoded depots on MP_STATION tiles. They cannot be fully stand alone depots, as the current airports are FSM. Even if we had modular airports, I still think that hangars should be part of their airport stations. Otherwise, aircraft pathfinding would be a nightmare.

src/order_cmd.cpp
J0anJosep: Ok!

src/saveload/station_sl.cpp
J0anJosep: Done. Thanks!

src/aircraft_cmd.cpp
```diff
@@ -1586,7 +1588,7 @@ static void AircraftEventHandler_AtTerminal(Aircraft *v, const AirportFTAClass *
 			return;
 		default:  // orders have been deleted (no orders), goto depot and don't bother us
 			v->current_order.Free();
-			go_to_hangar = Station::Get(v->targetairport)->airport.HasHangar();
+			go_to_hangar = true;
```
2TallTyler: What behavior does this change? Does it try a different airport? Maybe the comment above this could better explain what happens in this case. 🙂

src/aircraft_cmd.cpp
```diff
@@ -2147,18 +2149,7 @@ void UpdateAirplanesOnNewStation(const Station *st)
...
 		assert(v->state == FLYING);
 
-		Order *o = &v->current_order;
-		/* The aircraft is heading to a hangar, but the new station doesn't have one,
-		 * or the aircraft can't land on the new station. Cancel current order. */
-		if (o->IsType(OT_GOTO_DEPOT) && !(o->GetDepotOrderType() & ODTFB_PART_OF_ORDERS) && o->GetDestination() == st->index &&
-				(!st->airport.HasHangar() || !CanVehicleUseStation(v, st))) {
-			o->MakeDummy();
-			SetWindowWidgetDirty(WC_VEHICLE_VIEW, v->index, WID_VV_START_STOP);
-		}
```
2TallTyler, line 2179: Why is this no longer needed? (Just for our reference when reviewing, no need to add comments in the code itself.)

src/aircraft_cmd.cpp
```diff
@@ -2147,18 +2149,7 @@ void UpdateAirplanesOnNewStation(const Station *st)
...
 		v->pos = v->previous_pos = AircraftGetEntryPoint(v, ap, rotation);
 		UpdateAircraftCache(v);
 	}
-
-	/* Heliports don't have a hangar. Invalidate all go to hangar orders from all aircraft. */
-	if (!st->airport.HasHangar()) RemoveOrderFromAllVehicles(OT_GOTO_DEPOT, st->index, true);
```
2TallTyler, line 2185: Same as above, help us (me) understand why this isn't needed anymore.

src/saveload/saveload.h
```diff
@@ -353,6 +353,9 @@ enum SaveLoadVersion : uint16 {
 	SLV_LINKGRAPH_SECONDS,                  ///< 308  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	SLV_AI_START_DATE,                      ///< 309  PR#10653 Removal of individual AI start dates and added a generic one.
 
+	SLV_MTILE_DEPOTS_ADD_HANGARS,           ///< 310  PR#9577 Multi-tile depots. Add DepotIDs to airports that have a hangar.
```
2TallTyler: It's not necessary to mention multi-tile depots, as this change doesn't affect that yet.

Also, I've previously done only one savegame bump per PR, with multiple commits using it. Not sure if that's best practice, though, so maybe someone more experienced can chime in. 😛 

```suggestion
	SLV_HANGER_DEPOT_IDS,                   ///< 310  PR#10691 Add DepotIDs to airports that have a hangar.
```

src/aircraft_cmd.cpp
J0anJosep: Old goto hangar orders stored the StationID. It was possible to create a goto hangar order to a station having a hangar, demolish the airport and rebuilt it with another airport that had no hangar. Thus, when rebuilding an airport that had no hangar, all goto hangar orders were to be removed.

Adding a depot struct to airports when they have a hangar, allows hangar orders to store a DepotID. Most of the special cases you had to deal with hangar orders, can now be handled by the standard way (rail, road and water transport code).

With this commit, old goto depot orders to airports that don't have a hangar are removed on loading a game.

Also, when an airport is removed, their goto depot orders are also removed. This could be a problem, as the immediate removal of these orders is not always desired (rebuilding an airport). In subsequent commits, when a depot is deleted, it will stay for a while as a ghost (to keep orders associated to it) and eventually will be removed forever (behaving in a similar way to station removal).

src/aircraft_cmd.cpp
J0anJosep: I answered on another place. I will add some comment to the description of the PR as well.

src/saveload/saveload.h
J0anJosep: Ok. I will remove that mention. Anyway, I think it is ok to keep it, as some time from now maybe somebody reading the code won't know these changes are related to PR#9577. I will change this anyway.

I've been previously told each commit should work on its own. So I need one savegame bump per commit. I will keep it with two separated savebumps if I am not told otherwise.

src/aircraft_cmd.cpp
J0anJosep: This is a no change, if you read subsequent lines of code.

I will create another PR suggesting this.

2TallTyler: Thanks for the other PR. That sort of cleanup is helpful.

src/aircraft_cmd.cpp
2TallTyler: Ah, so instead this is handled by the change to `RemoveOrderFromAllVehicles()`?

J0anJosep: Yes. It is called during actions such as removing an airport, which deletes the associated depot, which calls RemoveOrderFromAllVehicles() in Depot::~Depot().

2TallTyler:
CI checks (and/or GitHub itself) were broken for a few days, but appear to be fixed now. If you rebase and force-push, it should trigger them to run.

src/depot_map.h
J0anJosep: I am not sure whether to store the depot index in the hangar tile. I agree that it would make things sightly more uniform, but I would prefer not to use those bits.
Any suggestion/preference about which approach should I take?

src/saveload/afterload.cpp
```diff
@@ -2770,6 +2748,43 @@ bool AfterLoadGame()
...
 	}
 
+	if (IsSavegameVersionBefore(SLV_ADD_DEPOTS_TO_HANGARS)) {
+		for (Station *st : Station::Iterate()) {
+			if ((st->facilities & FACIL_AIRPORT) && st->airport.HasHangar()) {
+				/* Add a built-in hangar for some airport types. */
+				assert(Depot::CanAllocateItem());
+				st->airport.AddHangar();
+			} else {
+				/* If airport has no hangar, remove old go to hangar orders
```
2TallTyler, line 2780: How would this case happen?

src/depot_map.h
2TallTyler: Your reasoning makes sense to me, unless @rubidium42 disagrees.

src/order_cmd.cpp
```diff
@@ -2020,7 +2019,8 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
 				} else {
 					Aircraft *a = Aircraft::From(v);
 					DestinationID destination = a->current_order.GetDestination();
-					if (a->targetairport != destination) {
+					Depot *hangar = Station::Get(a->targetairport)->airport.hangar;
```
2TallTyler: Should this be `Station::GetIfValid()` since the next line checks if `hanger != nullptr`?

src/saveload/afterload.cpp
```diff
@@ -2763,6 +2763,36 @@ bool AfterLoadGame()
...
+				order->SetDestination(st->airport.hangar->index);
+			}
+		}
+
+		for (Aircraft *a : Aircraft::Iterate()) {
+			if (!a->IsNormalAircraft()) continue;
+			if (!a->current_order.IsType(OT_GOTO_DEPOT)) continue;
+
+			Station *st = Station::Get(a->current_order.GetDestination());
+			if (st->airport.hangar != nullptr) {
```
2TallTyler: The loop above asserts that `st->airport.hangar != nullptr`. Is it possible here?

src/saveload/afterload.cpp
```diff
@@ -2763,6 +2763,36 @@ bool AfterLoadGame()
...
+				if (!order->IsType(OT_GOTO_DEPOT)) continue;
+				StationID station_id = order->GetDestination();
+				assert(Station::IsValidID(station_id));
+				Station *st = Station::Get(station_id);
+				assert(st->airport.hangar != nullptr);
+				order->SetDestination(st->airport.hangar->index);
+			}
+		}
+
+		for (Aircraft *a : Aircraft::Iterate()) {
```
2TallTyler: It would be nice to have a brief comment here explaining why we need to iterate through the aircraft list twice.

2TallTyler (commented):
Hi @J0anJosep! Sorry for dropping the ball on reviewing this PR. I have some more time on my hands in the foreseeable future, and if you're still willing, I'd like to continue the review process to eventually get #8480 merged into OpenTTD.

This PR looks good to me -- quality code, good comments, logical asserts, etc. It's clear you've done your planning and testing. If you rebase and address my comments (and promise to fix any bugs we find when merging big projects like this 😃), I'm willing to approve this and move on to your next PRs. That is, assuming @rubidium42 or others have no review items, and somebody else is willing to hit the merge button.

src/saveload/afterload.cpp
glx22: Heliport ;)

J0anJosep:
> Hi @J0anJosep! Sorry for dropping the ball on reviewing this PR. I have some more time on my hands in the foreseeable future, and if you're still willing, I'd like to continue the review process to eventually get #8480 merged into OpenTTD.
> 
> This PR looks good to me -- quality code, good comments, logical asserts, etc. It's clear you've done your planning and testing. If you rebase and address my comments (and promise to fix any bugs we find when merging big projects like this 😃), I'm willing to approve this and move on to your next PRs. That is, assuming @rubidium42 or others have no review items, and somebody else is willing to hit the merge button.

Thank you.
I will check your suggestions and update the PR soon.
I hope to have some free time this week or the next one.

src/saveload/afterload.cpp
J0anJosep: I will merge the two iterations. One was for the common lists of orders and the other one for the current order of the aircraft.

src/saveload/afterload.cpp
J0anJosep: Technically, it shouldn't be possible. But I am not sure all savegames are bug free and I know there have been some bugs related to hangar orders in the past. I will change the "if" with an assert, just in case.

I could remove the assert too, but I would keep it just in case.

2TallTyler: Well, they should match, at least. Keep in mind that assertions are used for development and testing only and are disabled in release builds. So if you expect users to load corrupted savegames, an assertion won't prevent this.

2TallTyler:
Hi @J0anJosep, I hope you're doing well. Whenever you have time, I'm happy to keep working with you on this. 😃

J0anJosep:
> Hi @J0anJosep, I hope you're doing well. Whenever you have time, I'm happy to keep working with you on this. 😃

I'll get in contact with you soon. Probably next week.

frosch123:
The PR description says:
> the current behavior is that any new built aircraft is built in the first hangar, so it is a no change/not a drawback.

This is not true? You can currently open two hangar windows for international airports, and build aircraft independently.
#9577 mentions this as well, though I do not quite understand how it solves it:
* Depot/hangar order should not care about the exact target tile.
* A vehicle entering a depot for service, should exit the same depot tile, and not exit from a different joined one. From this PR description it does not sound like it does?
* Manually stopping / buying... I have no idea :)

J0anJosep:
> > the current behavior is that any new built aircraft is built in the first hangar, so it is a no change/not a drawback.
> This is not true? You can currently open two hangar windows for international airports, and build aircraft independently. 
Yes. I changed the description. By a comment in the code, I assumed all aircraft ended up in the first hangar

With the current behavior, you can build aircraft on any of the two hangars and this will be lost with this PR. Anyway, the fact that currently you can select a hangar where to buy the aircraft vehicles but it is not possible to select a specific hangar as a "Go to depot" order seems like an inconsistency on the current behavior.

To me, it is more consistent that opening a hangar window shows all vehicles on the hangars of the airport. This way it is more difficult to miss some stopped vehicles.

In the future, new aircraft could be created to the least crowded hangar, but this is out of scope right now.

>     * Depot/hangar order should not care about the exact target tile.
No changes about that. The aircraft will go to any of the hangar tiles for servicing. All the hangar tiles of the same airport share the same DepotID.
 
>     * A vehicle entering a depot for service, should exit the same depot tile, and not exit from a different joined one. From this PR description it does not sound like it does?
Existing aircraft does not "teleport" from one hangar tile to another one. The "teleporting" problem with multi-tile depots arises only when a train or road vehicle is replaced and the new vehicle is incompatible with the depot tile (and the new vehicle is moved to a compatible depot tile).

>     * Manually stopping / buying... I have no idea :)
I don't understand which problem are you hinting here. I will gladly discuss it, but this PR is just for introducing DepotIDs for hangars.

frosch123:
I tried this PR. Hangars still seem to be separate: they show different vehicles and I can buy new aircraft in one or the other. What am I missing?

src/saveload/saveload.h
```diff
@@ -383,6 +383,7 @@ enum SaveLoadVersion : uint16_t {
 	SLV_GROUP_NUMBERS,                      ///< 336  PR#12297 Add per-company group numbers.
 	SLV_INCREASE_STATION_TYPE_FIELD_SIZE,   ///< 337  PR#12572 Increase size of StationType field in map array
 	SLV_ROAD_WAYPOINTS,                     ///< 338  PR#12572 Road waypoints
+	SLV_ADD_DEPOTS_TO_HANGARS,              ///< XXX  PR#10691 Add depots to airports that have a hangar.
```
2TallTyler:
```suggestion
	SLV_ADD_DEPOTS_TO_HANGARS,              ///< 339  PR#10691 Add depots to airports that have a hangar.
```

src/station.cpp
```diff
@@ -713,6 +714,44 @@
...
+/**
+ * Create a hangar on the airport.
+ */
+void Airport::AddHangar()
+{
+	assert(this->hangar == nullptr);
+	assert(Depot::CanAllocateItem());
+	assert(this->GetNumHangars() > 0);
+	Station *st = Station::GetByTile(this->GetHangarTile(0));
+	this->hangar = new Depot(this->GetHangarTile(0));
```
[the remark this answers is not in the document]
2TallTyler: @J0anJosep Should `Station::~Station()` include a `delete this->hangar`?

src/saveload/saveload.h
```diff
@@ -385,6 +385,8 @@ enum SaveLoadVersion : uint16_t {
 	SLV_ROAD_WAYPOINTS,                     ///< 338  PR#12572 Road waypoints
 	SLV_ADD_DEPOTS_TO_HANGARS,              ///< XXX  PR#10691 Add depots to airports that have a hangar.
 
+	SLV_DEPOTID_IN_HANGAR_ORDERS,           ///< 320  PR#10691 Go to hangar orders store the DepotID instead of StationID.
```
2TallTyler:
```suggestion
	SLV_DEPOTID_IN_HANGAR_ORDERS,           ///< 340  PR#10691 Go to hangar orders store the DepotID instead of StationID.
```

2TallTyler (commented):
Let's get this done so you can move on to the follow-up PRs. 😃

src/saveload/saveload.h
LordAro: I don't follow why there's two savegame bumps

J0anJosep: > I don't follow why there's two savegame bumps

**One is for adding the depot structs for airports. (first commit)
The other one is for changing the way goto hangar orders are handled. (second commit)**

If it is better to have only one savegame bump, I can change it, but I was told in the past to keep individual commits with its corresponding savegame bump.

The values XXX and 320, of course, would need to be updated, but I will do it when the PR were to be accepted. I have rebased it so many times, I got tired to do it at some point.

Just let me know what is best in this case.

src/saveload/saveload.h
J0anJosep: Ok. I will update those numbers. After like more than 50 rebases related with savegame bumps, I got tired to do so.

J0anJosep:
> @J0anJosep Should Station::~Station() include a delete this->hangar?
Removal of hangar-depots is handled in RemoveAirport. While CleaningPools, nothing is to be done.

I don't think it should be handled on ~Station(), as it happens some time after RemoveAirport is called, so it has no airport nor hangar. I don't know how to silence the warning, though.

I have also checked how are Depots handled in takeovers and bankruptcy of companies. On takeovers, no change is needed. On bankruptcy and complete removal of the company items, RemoveAirport is called, so it is already handled.

src/saveload/saveload.h
2TallTyler: My recollection of past conversations included the idea that each commit should be able to stand alone. An easy way to avoid this debate is to squash when merging. 🙂 

(Curious what other devs think)

src/station.cpp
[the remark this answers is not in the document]
2TallTyler: > Sorry, but I couldn't comment it above.
> Removal of hangar-depots is handled in RemoveAirport. While CleaningPools, nothing is to be done.
> 
> I don't think it should be handled on ~Station(), as it happens some time after RemoveAirport is called, so it has no airport nor hangar. I don't know how to silence the warning, though.
> 
> I have also checked how are Depots handled in takeovers and bankruptcy of companies. On takeovers, no change is needed. On bankruptcy and complete removal of the company items, RemoveAirport is called, so it is already handled.

Ah, and `delete this->hangar;` shouldn't be moved to the destructor because the airport can be rebuilt to add or remove a hanger without the airport itself being deleted.

I can silence the warning.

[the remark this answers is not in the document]
2TallTyler: Actually, can it be moved to the destructor?

I am not fully clear what persists after a station is demolished to allow a new station to take the name (and orders) of a recently-demolished station. Perhaps you can investigate. 🙂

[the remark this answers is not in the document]
J0anJosep: I think it shouldn't be moved to the destructor.

If the airport is deleted, the hangar is immediately deleted as well, **but the station will still exist for some time**, and the station will keep its name until it is destroyed (or the station could have other facilities so the station won't be removed at all).

There would be other ways to deal with the hangars: we could create a depot for each station, independently of whether the station has an airport, so no need to worry to remove the hangar. But I don't like it: creating or removing the hangars when needed seems like a better approach to me.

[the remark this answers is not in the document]
2TallTyler: Fair enough. I will silence the warning.

PR #10694, opened 2023-04-22
Fix: Violation of strict weak ordering in TownRatingSorter and GenericEngineValueVsRunningCostSorter

JGRennison:
## Motivation / Problem

Violation of strict weak ordering in TownRatingSorter if two unrated town names compare equal in TownNameSorter.
Violation of strict weak ordering in GenericEngineValueVsRunningCostSorter if an engine is compared with itself.

Platforms which rely on std::sort comparators obeying the strict weak ordering restrictions (i.e. MacOS) may misbehave and/or crash if this is not followed. In particular a < b and b < a may not both be true.

Crash report identifying issue: https://www.tt-forums.net/viewtopic.php?p=1260523#p1260523

## Description

Don't use !comparator(a, b) to reverse order, because this incorrectly handles the equal case, use comparator(b, a).

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -300,7 +300,7 @@ static bool EnginePowerVsRunningCostSorter(const GUIEngineListItem &a, const GUI
 	 * since we want consistent sorting.
 	 * Also if both have no power then sort with reverse of running cost to simulate
 	 * previous sorting behaviour for wagons. */
-	if (v_a == 0 && v_b == 0) return !EngineRunningCostSorter(a, b);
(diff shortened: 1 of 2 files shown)


glx22 (dismissed):

PeterN:
Could this issue also affect `return !EngineRunningCostSorter(a, b);` in `build_vehicle_gui.cpp`?

JGRennison:
EngineRunningCostSorter should be less of a problem because it falls back to sorting on Engine::list_position which is unique for each engine ID, but it is still not strictly correct.

glx22 (approved):

Issue #1076, opened 2007-07-25
Glitch when o


DorpsGek:
**_Zuu_ opened the ticket and wrote:**

> Version: Win32 nightly: r10673
> 
> === How to reproduce ===
> 1 Open rail-toolbar. 
> 2 Enable patch for linking landscape toolbar with construction toolabars.
> 3 Press A to open rail+landscape.
> 
> You now have a nice glitch. The rail-bar is moved, but it's drawing rectangle is not updated. Or how to put it. 
> 
> This glitch can be achived with the link-patch turned on permanently. You just have to close the landscape box and move back the rail-box and then press A. The glitch do not happen if you open the rail-box with the F7 shortcut nor if you press the rail-box-button.

**Reported version**: _trunk_
**Operating system**: _All_

---

DorpsGek:
**_Rubidium_ closed the ticket.**

**Reason for closing: _Fixed_**

> In r10685

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/1076_

PR #10700, opened 2023-04-23
Codechange: Split dates and timers into Economy and Calendar time

2TallTyler:
## Motivation / Problem

In preparation for NotDaylength (#11428), we need to track calendar and economy dates separately.

## Description

For now, just split the dates into two systems and keep them synchronized. No part of this PR is visible to the player.

Calendar time is used for technology and time-of-year changes, including:
- Vehicle, airport, station, object introduction and obsolescence
- Vehicle aging, depreciation, reliability, and renewal
- NewGRF variables for visual styles or behavior based on year or time of year (e.g. variable snow line)
- Inflation, since it is tied to original game years. One interpretation of inflation is that it compensates for faster and higher capacity vehicles, another is that it compensates for more established companies. Each of these point to a different choice of calendar versus economy time, but we have to pick one so we follow a previous decision to tie inflation to original TTD game years.

Economy time is used for the regular pace of the game, including:
- Industry and house production/consumption
- Industry production changes, closure, and spawning
- Town growth
- Company age and financial statistics
- Vehicle financial statistics
- Payment intervals for running and maintenance costs, loan interest, etc.
- Cargo payment "time" calculation
- Local authority and station ratings change intervals
- LogCommandExecution

## Limitations

These remain only on calendar time, because with economy time not displayed anywhere, it is generally useless information. Counter-arguments welcome. 🙂 

- `getdate` command

## Checklist for review


src/date_type.h
```diff
@@ -32,6 +32,10 @@
 
 static const int SECONDS_PER_DAY   = 2;   ///< approximate seconds per day, not for precise calculations
 
+static const int DAYS_IN_ECONOMY_MONTH  = 30; ///< economy days per econony month -- at 2 seconds a day, this equals 1 minute per month
+static const int MONTHS_IN_ECONOMY_YEAR = 12; ///< months in economy economy year
+static const int DAYS_IN_ECONOMY_YEAR   = DAYS_IN_ECONOMY_MONTH * MONTHS_IN_ECONOMY_YEAR; ///< days per economy year
```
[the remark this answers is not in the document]
2TallTyler: Oops. This gets used a lot in later PRs. Okay to leave it in, or add it then?

src/date.cpp
```diff
@@ -25,14 +25,26 @@
...
+Month     _cur_month;  ///< Current calendar month (0..11)
+Date      _date;       ///< Current calendar date in days (day counter)
+DateFract _date_fract; ///< Fractional part of the calendar day
+
+Year      _cur_economy_year;   ///< Current economy year
+Month     _cur_economy_month;  ///< Current economy month (0..11)
+Date      _economy_date;       ///< Current economy date in days
+DateFract _economy_date_fract; ///< Fractional part of the economy day
+
+uint64 _tick_counter;  ///< Ever incrementing (and sometimes wrapping) tick counter for setting off various events
```
PeterN: Perhaps this should be a struct?

```
struct Dates { Year year, Month month, Date date, DateFract date_fract };
Dates _calendar;
Dates _economy;
```

But better names...

src/date_type.h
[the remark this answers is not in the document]
TrueBrain: Please remove; who knows when the other PR will land, better to not have these things now :)

(also, it would remove all changes in this file, if I get this right :D)

src/aircraft_cmd.cpp
```diff
@@ -1556,6 +1557,7 @@ static void AircraftEventHandler_AtTerminal(Aircraft *v, const AirportFTAClass *
 			if (v->subtype == AIR_HELICOPTER && apc->num_helipads > 0) {
 				/* an excerpt of ServiceAircraft, without the invisibility stuff */
 				v->date_of_last_service = _date;
```
TrueBrain: Please also rename `date_of_last_service`, so it is more clear which is what.

src/cheat_gui.cpp
```diff
@@ -115,11 +115,16 @@ static int32 ClickChangeDateCheat(int32 new_value, int32 change_direction)
 	/* Change the date. */
 	SetDate(new_date, _date_fract);
 
+	/* Change economy date. */
+	ConvertDateToYMD(_economy_date, &ymd);
+	new_date = ConvertYMDToDate(new_value, ymd.month, ymd.day);
+	SetEconomyDate(new_date, _economy_date_fract);
```
TrueBrain: please don't reuse `new_date`; it makes reading the code more confusing.

src/company_base.h
```diff
@@ -77,7 +77,8 @@ struct CompanyProperties {
 
 	std::array<Owner, MAX_COMPANY_SHARE_OWNERS> share_owners; ///< Owners of the shares of the company. #INVALID_OWNER if nobody has bought them yet.
 
-	Year inaugurated_year;           ///< Year of starting the company.
+	Year inaugurated_year;           ///< Calendar year the company was started.
+	Year inaugurated_economy_year;   ///< Economy year the company was started.
```
TrueBrain: let's assume #10709 lands, meaning you can leave out this new variable :)

src/engine.cpp
```diff
@@ -1112,9 +1112,9 @@ void EnginesMonthlyLoop()
 	}
 }
 
-static IntervalTimer<TimerGameCalendar> _engines_monthly({TimerGameCalendar::MONTH, TimerGameCalendar::Priority::ENGINE}, [](auto)
+static IntervalTimer<TimerGameCalendar> _calendar_engines_monthly({TimerGameCalendar::MONTH, TimerGameCalendar::Priority::ENGINE}, [](auto)
 {
-	EnginesMonthlyLoop();
+	EnginesCalendarMonthlyLoop();
```
TrueBrain: My OCD hits on the difference in where "Calendar" goes. `_calendar_engines_monthly` vs `EnginesCalendarMonthlyLoop`. Pick one please :) Not both.

src/industry_cmd.cpp
```diff
@@ -2986,11 +2986,11 @@ static IntervalTimer<TimerGameCalendar> _industries_daily({TimerGameCalendar::DA
 	InvalidateWindowData(WC_INDUSTRY_DIRECTORY, 0, IDIWD_PRODUCTION_CHANGE);
 });
 
-static IntervalTimer<TimerGameCalendar> _industries_monthly({TimerGameCalendar::MONTH, TimerGameCalendar::Priority::INDUSTRY}, [](auto)
+static IntervalTimer<TimerGameEconomy> _economy_industries_monthly({TimerGameEconomy::MONTH, TimerGameEconomy::Priority::INDUSTRY}, [](auto)
 {
 	Backup<CompanyID> cur_company(_current_company, OWNER_NONE, FILE_LINE);
 
-	_industry_builder.MonthlyLoop();
+	_industry_builder.EconomyMonthLoop();
```
TrueBrain: `Month` -> `Monthly`, for consistency.

src/news_gui.cpp
```diff
@@ -1009,9 +1009,9 @@ void NewsLoop()
 
 	static byte _last_clean_month = 0;
```
TrueBrain: Please fix `byte` -> `Month` while we are here.

src/saveload/saveload.h
```diff
@@ -353,6 +353,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_LINKGRAPH_SECONDS,                  ///< 308  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	SLV_AI_START_DATE,                      ///< 309  PR#10653 Removal of individual AI start dates and added a generic one.
 
+	SLV_ECONOMY_DATE,                       ///< 310  PR#XXXXX Separate calendar time and economy time.
```
TrueBrain: You can fix this too :D

src/timer/timer_game_economy.cpp
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+template<>
+void TimerManager<TimerGameEconomy>::Elapsed(TimerGameEconomy::TElapsed delta)
+{
+	assert(delta == 1);
+
+	/* Tick counter gets updated here, not in calendar timer. */
+	_tick_counter++;
```
TrueBrain: No longer needed after rebase :)

src/timer/timer_game_economy.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	/* Check if we reached the maximum year, decrement dates by a year */
+	if (_cur_economy_year == MAX_YEAR + 1) {
+		int days_this_year;
+
+		_cur_economy_year--;
+		days_this_year = IsLeapYear(_cur_economy_year) ? DAYS_IN_LEAP_YEAR : DAYS_IN_YEAR;
+		_economy_date -= days_this_year;
+		for (Vehicle *v : Vehicle::Iterate()) v->ShiftDates(-days_this_year);
+		for (LinkGraph *lg : LinkGraph::Iterate()) lg->ShiftDates(-days_this_year);
```
TrueBrain, line 141: This should either be done in Economy or in Calendar; not both.

src/timetable.h
```diff
@@ -13,7 +13,7 @@
 #include "date_type.h"
 #include "vehicle_type.h"
 
-static const uint8 MAX_TIMETABLE_START_YEARS = 15; ///< The maximum start date offset, in years.
+static const uint8 MAX_TIMETABLE_START_YEARS = 15; ///< The maximum start date offset, in economy years.
```
TrueBrain: Maybe also change the type to `Year` while at it? (not sure if anything breaks because of it)

src/aircraft_cmd.cpp
2TallTyler: I'll make a separate PR later, following the approach of #10721. I guess I'll also need to do the same for `v->build_year` since NewGRF can get that too. 😕 (Writing it here so I don't forget! 😛)

src/company_base.h
```diff
@@ -76,7 +76,7 @@ struct CompanyProperties {
 	TileIndex location_of_HQ;        ///< Northern tile of HQ; #INVALID_TILE when there is none.
 	TileIndex last_build_coordinate; ///< Coordinate of the last build thing by this company.
 
-	TimerGameCalendar::Year inaugurated_year; ///< Year of starting the company.
+	TimerGameCalendar::Year inaugurated_year; ///< Economy year of starting the company.
```
TrueBrain: Changing the comment, not the type? :)

src/industry.h
```diff
@@ -88,9 +89,9 @@ struct Industry : IndustryPool::PoolItem<&_industry_pool> {
 	mutable std::string cached_name; ///< NOSAVE: Cache of the resolved name of the industry
 
 	Owner founder;                 ///< Founder of the industry
-	TimerGameCalendar::Date construction_date; ///< Date of the construction of the industry
+	TimerGameCalendar::Date construction_date; ///< Calendar date of the construction of the industry
 	uint8 construction_type;       ///< Way the industry was constructed (@see IndustryConstructionType)
-	TimerGameCalendar::Date last_cargo_accepted_at[INDUSTRY_NUM_INPUTS]; ///< Last day each cargo type was accepted by this industry
+	TimerGameEconomy::Date last_cargo_accepted_at[INDUSTRY_NUM_INPUTS]; ///< Last calendar day each cargo type was accepted by this industry
```
TrueBrain: Mismatch comment / type.

src/saveload/misc_sl.cpp
```diff
@@ -78,6 +78,8 @@ static const SaveLoad _date_desc[] = {
 	    SLEG_VAR("date_fract",             TimerGameCalendar::date_fract,             SLE_UINT16),
 	SLEG_CONDVAR("tick_counter",           TimerGameTick::counter,           SLE_FILE_U16 | SLE_VAR_U64,  SL_MIN_VERSION, SLV_U64_TICK_COUNTER),
 	SLEG_CONDVAR("tick_counter",           TimerGameTick::counter,           SLE_UINT64,                  SLV_U64_TICK_COUNTER, SL_MAX_VERSION),
+	SLEG_CONDVAR("economy_date",           TimerGameCalendar::date,          SLE_INT32,                   SLV_ECONOMY_DATE, SL_MAX_VERSION),
+	SLEG_CONDVAR("economy_date_fract",     TimerGameCalendar::date_fract,    SLE_UINT16,                  SLV_ECONOMY_DATE, SL_MAX_VERSION),
```
TrueBrain: Nasty typo; think you want to store the `TimerGameEconomy` here :D

TrueBrain (commented):
Scary how much easier this is to review now :)

I haven't looked yet if I agree with which timers moved to where; will do that soon (if no other developer jumps in to do so :D)

src/aircraft_cmd.cpp
```diff
@@ -340,7 +340,7 @@ CommandCost CmdBuildAircraft(DoCommandFlag flags, TileIndex tile, const Engine *
 		v->SetServiceInterval(Company::Get(_current_company)->settings.vehicle.servint_aircraft);
 
 		v->date_of_last_service = TimerGameEconomy::date;
-		v->build_year = u->build_year = TimerGameCalendar::year;
+		v->build_year = u->build_year = TimerGameEconomy::year;
```
PeterN: Doing this will make NewGRFs that change graphics or other properties depending on when the vehicle was built or last serviced not necessarily do what the author intended.

2TallTyler: For now, economy and calendar dates are always kept in sync. Before adding the feature that changes this (real-time mode), I'll need to do something like #10721 for both date of last service and build year.

glx22:
Hmm when debugging regression with mingw something is definitely wrong.
image
Finance window pops every day and seems to contain garbage. (Regression AI gets the same broken info most likely)

And for some reasons script debug window is empty (might not be related).

src/timer/timer_game_economy.cpp
```diff
@@ -0,0 +1,174 @@
...
+ * @param ymd  the year, month and day to write to
+ */
+/* static */ void TimerGameEconomy::ConvertDateToYMD(TimerGameEconomy::Date date, YearMonthDay *ymd)
+{
+	TimerGameCalendar::YearMonthDay calendar_ymd;
+	calendar_ymd.day = ymd->day;
+	calendar_ymd.month = ymd->month;
+	calendar_ymd.year = ymd->year;
+	TimerGameCalendar::ConvertDateToYMD(date, &calendar_ymd);
+}
```
JGRennison: This does not look like it is doing the right thing.
*ymd is not actually written to here.

Starting a new game with this PR seems to result in TimerGameEconomy::year/month having silly values.

src/saveload/saveload.h
```diff
@@ -352,6 +352,7 @@ enum SaveLoadVersion : uint16 {
 
 	SLV_EXTEND_VEHICLE_RANDOM,              ///< 310  PR#10701 Extend vehicle random bits.
 	SLV_EXTEND_ENTITY_MAPPING,              ///< 311  PR#10672 Extend entity mapping range.
+	SLV_ECONOMY_DATE,                       ///< 312  PR#10700 Split calendar and economy timers and dates.
```
TrueBrain: Strictly seen we don't need this new version, as introducing a new saveload field "just works". But I guess we are so used to it .. so it is fine.

src/timer/timer_game_economy.h
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+/** @file timer_game_economy.h Definition of the game-economy-timer */
+
+#ifndef TIMER_GAME_ECONOMY_H
+#define TIMER_GAME_ECONOMY_H
+
+/**
+ * Timer that is increased every 27ms, and counts towards economy time units, expressed in days / months / years.
+ * It also updates _tick_counter.
```
TrueBrain: Lies! :D

src/cheat_gui.cpp
```diff
@@ -112,13 +113,14 @@ static int32 ClickChangeDateCheat(int32 new_value, int32 change_direction)
 	TimerGameCalendar::Date new_date = TimerGameCalendar::ConvertYMDToDate(new_value, ymd.month, ymd.day);
 
 	/* Shift cached dates before we change the date. */
-	for (auto v : Vehicle::Iterate()) v->ShiftDates(new_date - TimerGameCalendar::date);
-	LinkGraphSchedule::instance.ShiftDates(new_date - TimerGameCalendar::date);
+	for (auto v : Vehicle::Iterate()) v->ShiftDates(new_date - TimerGameEconomy::date);
+	LinkGraphSchedule::instance.ShiftDates(new_date - TimerGameEconomy::date);
```
TrueBrain: This is now a bit weird; I know it will be fixed in later PRs, but you calculate `new_date` based on `Calendar`, then use `Economy` to do the delta. And yes, of course they are both set to the same value later, but it just reads a bit weird.

Not sure it needs fixing, just observing.

TrueBrain:
Some random questions you might or might not have answers for. Not all related to this PR, but worth mentioning nevertheless.

- Timetable remains on Calendar. How is that going to work? If you freeze the calendar, timetables are useless? Or what are the plans here exactly :)
- `LogCommandExecution` doesn't log Economy date. Might be useful to do so too. There are other similar situations, like crashlog.
- `getdate` console command only return Calendar date. I would expect it to return both. But I can imagine this being done in a later PR, when they actually differ.
- `_newgrf_profile_end_date` is a bit weird. I think the intention is that it is against `tick_counter`. But for some reason it is written against `date`. Needs a bit looking into to .. but having a profiler look at Calendar is just weird :P (and yes, not for this PR)
- `DeleteStateLinks` uses `Calendar`, which feels a bit off. I guess it should be similar to linkgraph?
- Network heavily relies on `Calendar`, for desync debugging etc. Will be a bit awkward with one of the other PRs that allows freezing the calendar :D
- In `station_cmd.cpp` we compare `date_of_last_service` with `Calendar::date`. This seems wrong, given the first changed to `Economy`?

So in summary, if I get this right, after this PR, `Calendar` is used for:
- Introduction dates
- Age
- Construction date

These kind of things. Everything else is now `Economy`. I think it would be good if we write in the comments of the different timers for what they are meant to be used. Just so it is more clear for future-us what the division between the two is. And that requires some very clear rules, to not allow for fluffiness :)

For example, that `inaugurated_year` is Economy is a bit surprising given the other entries. You would think it also to be a `Calendar`.

Another example, inflation is typical considered an economy-based something. It makes perfect sense that it is still based on `Calendar` when to be active or not, but I think it requires some clear comments in the timers how/what/why :)

src/cheat_gui.cpp
2TallTyler: Yes, this will only ever happen when calendar and economy dates are synced (not in real-time mode, which also means 100% calendar progress speed), so either date type is somewhat wrong (the dates being shifted are economy dates). I could change this calculation to calendar dates, it gives identical results.

In real-time mode, we never shift dates and only the calendar date is affected by the date cheat, which should fix a lot of strange ~~bugs~~ quirks related to the cheat such as company finance data.

src/saveload/saveload.h
2TallTyler: No? Is there a way to copy data to the new fields, instead of doing it in afterload? 

I also suspect we need it for the case of loading scenarios, also handled here in afterload. I'm not sure if we will eventually allow changing real-time mode in Scenario Editor, will have to look into that in the future.

2TallTyler (commented):
Some answers to your questions:

- Timetable are stored in Ticks, except for the Start Date feature which now uses the economy date. In real-time mode it will use an offset in seconds (seconds until start). So this has not changed, unless I've missed a cherry-pick. Timetables work fine with frozen calendar speed in my other PRs in this project.
- I'm not up to speed with logging commands and crashlogs. Maybe those should be a separate PR? (as of now, calendar and economy are the same)
- Same as above for getdate command. 
- I'm also planning on doing a separate PR for NewGRF and AI/GS API additions.
- Thanks for handling NewGRF profiling separately. 😃 
- Yes, the date types in `station_cmd` for Linkgraph stuff was indeed missed in a cherry-pick somewhere. Should be fixed now.

I've added some comments to the timer files that explain the use cases for each, and will update the PR description with my notes from #10605, as requested.

TrueBrain:
I think this PR is mostly okay as it is now, but I also think we should wait till we did some other stuff first. We already know this would be work that requires a lot of other changes first to truly become reviewable and easy to maintain.
Mainly what you notice, especially after splitting up the original PR, that we had a lot of technical debt around these areas. And they are now biting in our ass. I am afraid that if we merge this too soon, it is tempted to continue functional work on it, while the fundamentals are not ready yet. So I suggest we first resolve the technical debt till a level this PR becomes easy to validate, and continue from there.

The list of things I would like to see done before we do this PR (but this is my opinion, so .. take that for what it is worth :D):

- Something like #1076, so we get warnings / errors when we assign a Calendar thing to an Economy thing and visa versa. That is going to avoid a lot of bugs and weird issues down the line.
- Resolve the weirdness around link-graph, and use a proper timer for that.
- Split network away from either clock, and make sure that all makes a bit more sense.
- Date of last service mess; figure out what we actually want there.

To be clear, I really appreciate all the work that has gone into this so far, and our codebase is truly becoming better (in my opinion) so we can work towards this goal. But brrr, we had a lot of technical debt :D

2TallTyler:
Sounds good to me!

2TallTyler:
@TrueBrain: As far as tech debt goes, #11142 and #11124 should clear up things somewhat. Other network stuff seems to be based on economy dates rolling over, and should stay that way.

This leaves #10761 in your hands, and then the big scary monster I've been leaving for last... Linkgraph. To be honest, I have no idea how to approach this. @michicc explained on Discord that these updates are spaced out to avoid the computing load of doing them all at once, so it clearly doesn't need to remain tied to any sort of date. I'm just not sure how best to restructure this. Maybe a conversation we can have on Discord sometime. 🙂

2TallTyler:
Rebased to try and stay current with master. NOT guaranteed to work -- once we finally conquer our tech debt I'm going to have to go through the code with a fine-tooth comb (again) and test every commit to ensure everything works properly. 😃

src/saveload/afterload.cpp
```diff
@@ -1428,12 +1434,12 @@ bool AfterLoadGame()
...
-		for (Company *c : Company::Iterate()) c->inaugurated_year += ORIGINAL_BASE_YEAR;
-		for (Industry *i : Industry::Iterate())  i->last_prod_year   += ORIGINAL_BASE_YEAR;
+		for (Company *c : Company::Iterate())    c->inaugurated_year += static_cast<TimerGameEconomy::Year>(ORIGINAL_BASE_YEAR);
+		for (Industry *i : Industry::Iterate())  i->last_prod_year   += static_cast<TimerGameEconomy::Year>(ORIGINAL_BASE_YEAR);
 
 		for (Vehicle *v : Vehicle::Iterate()) {
-			v->date_of_last_service += DAYS_TILL_ORIGINAL_BASE_YEAR;
-			v->build_year += ORIGINAL_BASE_YEAR;
+			v->date_of_last_service += static_cast<TimerGameEconomy::Year>(DAYS_TILL_ORIGINAL_BASE_YEAR);
+			v->build_year += static_cast<TimerGameEconomy::Year>(ORIGINAL_BASE_YEAR);
```
TrueBrain: These kind of casts shouldn't exist :D I like StrongTypeDef .. shows when you are cheating :D

src/saveload/industry_sl.cpp
```diff
@@ -214,7 +214,7 @@ struct INDYChunkHandler : ChunkHandler {
 			auto &a = i->accepted[j];
 			a.cargo = SlIndustryAccepted::old_cargo[j];
 			a.waiting = SlIndustryAccepted::old_waiting[j];
-			a.last_accepted = SlIndustryAccepted::old_last_accepted[j];
+			a.last_accepted = static_cast<TimerGameEconomy::Date>(SlIndustryAccepted::old_last_accepted[j]);
```
TrueBrain: Also an odd one :)

src/saveload/afterload.cpp
```diff
@@ -3241,6 +3247,13 @@ bool AfterLoadGame()
 		}
 	}
 
+	if (IsSavegameVersionBefore(SLV_ECONOMY_DATE)) {
+		/* Set vehicle build dates provided to NewGRF. */
+		for (Vehicle *v : Vehicle::Iterate()) {
+			v->build_year_newgrf = v->build_year;
```
TrueBrain: Hmm, no cast? Guess Year is not strongly typed yet?

src/toolbar_gui.cpp
```diff
@@ -1150,8 +1150,6 @@ void SetStartingYear(TimerGameCalendar::Year year)
 {
 	_settings_game.game_creation.starting_year = Clamp(year, CalendarTime::MIN_YEAR, CalendarTime::MAX_YEAR);
 	TimerGameCalendar::Date new_date = TimerGameCalendar::ConvertYMDToDate(_settings_game.game_creation.starting_year, 0, 1);
-	/* If you open a savegame as scenario there may already be link graphs.*/
-	LinkGraphSchedule::instance.ShiftDates(new_date - TimerGameCalendar::date);
```
TrueBrain, line 1108: Why this removal?

src/saveload/afterload.cpp
2TallTyler: Yeah, strange that it builds fine. It is casted now, though. 🙂

src/toolbar_gui.cpp
2TallTyler: This is in `SetStartingYear()`, which changes TimerGameCalendar::date and used to shift linkgraph dates. But linkgraphs are now in TimerGameEconomy, so they don't care about the calendar date.

TrueBrain: But in this PR they are kept in sync, right?

src/timer/timer_game_common.h
```diff
@@ -0,0 +1,292 @@
...
+	static constexpr typename TimerGame<T>::Date INVALID_DATE = -1; ///< Representation of an invalid date
+};
+
+/**
+ * Converts a Date to a Year, Month & Day.
+ * @param date the date to convert from
+ * @param ymd  the year, month and day to write to
+ */
+template <class T>
+/* static */ void TimerGame<T>::ConvertDateToYMD(TimerGame<T>::Date date, YearMonthDay *ymd)
```
TrueBrain: I came to realise this might be problematic, as I am not sure how often this is now generated in the code. But I also realised we can still move this to a .cpp file, it just needs a bit of help. I will thinking with it later today to give you an example.

src/saveload/afterload.cpp
```diff
@@ -733,6 +734,11 @@ bool AfterLoadGame()
 	 * must be done before loading sprites as some newgrfs check it */
 	TimerGameCalendar::SetDate(TimerGameCalendar::date, TimerGameCalendar::date_fract);
 
+	/* Determine economy dates based on calendar date */
+	if (IsSavegameVersionBefore(SLV_ECONOMY_DATE)) {
+		TimerGameEconomy::SetDate(static_cast<int32_t>(TimerGameCalendar::date), TimerGameCalendar::date_fract);
+	}
```
TrueBrain, line 743: This should not be done conditionally.

In the savegame, only `date` is stored, not `year`. This is deduced on load again by this unconditional statement as on line 735.

TrueBrain (commented):
In `newgrf.cpp`, around line 9932 is `LoadNewGRF`, which siwtches the time temporary for loading in network games. I guess this should also do the same trick for economy?

Anyway, with this PR, we are really going for your version of NoDL, and there "is no way back". Are we actually sure that is the right way to go? Just double-checking :)

src/saveload/afterload.cpp
2TallTyler: Ah, yes. It needs to be conditional so that we always set call SetDate() but if no economy date is stored, we use the calendar date instead of the economy date.

src/timer/timer_game_calendar.cpp
```diff
@@ -10,12 +10,20 @@
...
  */
 
+/**
+ * Calendar time is used for technology and time-of-year changes, including:
+ * - Vehicle, airport, station, object introduction and obsolescence
+ * - NewGRF variables for visual styles or behavior based on year or time of year (e.g. variable snow line)
+ * - Inflation, since it is tied to original game years. One interpretation of inflation is that it compensates for faster and higher capacity vehicles,
+ *   another is that it compensates for more established companies. Each of these point to a different choice of calendar versus economy time, but we have to pick one
+ *   so we follow a previous decision to tie inflation to original TTD game years.
+ */
```
TrueBrain, line 21: should be in the header-file, honestly, as doxygen comment on the class I think?

src/timer/timer_game_common.h
```diff
@@ -0,0 +1,205 @@
...
+	ACCUM_NOV = ACCUM_OCT + 31,
+	ACCUM_DEC = ACCUM_NOV + 30,
+};
+
+/** Number of days to pass from the first day in the year before reaching the first of a month. */
+static constexpr uint16_t _accum_days_for_month[] = {
+	ACCUM_JAN, ACCUM_FEB, ACCUM_MAR, ACCUM_APR,
+	ACCUM_MAY, ACCUM_JUN, ACCUM_JUL, ACCUM_AUG,
+	ACCUM_SEP, ACCUM_OCT, ACCUM_NOV, ACCUM_DEC,
+};
```
TrueBrain: These can now be moved into the .cpp. That was kinda the point of moving the function, not? :)

src/timer/timer_game_common.h
```diff
@@ -0,0 +1,205 @@
...
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file timer_game_common.h Definition of the common class inherited by both calendar and economy timers. */
+
+#ifndef TIMER_GAME_COMMON_H
+#define TIMER_GAME_COMMON_H
+
+#include "stdafx.h"
```
TrueBrain: We shouldn't include stdafx.h in header-files, I think. I think we don't do that anywhere? All cpp files already do, as first line :)

src/timer/timer_game_economy.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * - Industry and house production/consumption
+ * - Industry production changes, closure, and spawning
+ * - Town growth
+ * - Company age and periodical finance stats
+ * - Vehicle age and profit statistics, both individual and group
+ * - Vehicle aging, depreciation, reliability, and renewal
+ * - Payment intervals for running and maintenance costs, loan interest, etc.
+ * - Cargo payment "time" calculation
+ * - Local authority and station ratings change intervals
+ */
```
TrueBrain, line 24: Same here, move to header file?

rubidium42:
Regarding `LogCommandExecution`, I think that should be on economy date. Why? If you want to rerun the commands to try to reproduce a desync, the "stable" clock should be used and not the slowed down or even stopped clock. After all, with a stopped clock at some point date_fract will overflow, making it impossible to uniquely determine the moment to replay the command.

2TallTyler:
Good point, I've made that change. Just as a note, TimerGameCalendar::date_fract should never overflow even in Frozen mode, since in that mode it stops incrementing. But that's equally useless for debugging.

TrueBrain (dismissed):

TrueBrain (approved):

PR #10706, opened 2023-04-24
Codechange: move all date-related variables inside the timer

TrueBrain:
## Motivation / Problem

While reading #10700, it became apparent that we have a few date-related globals that are now part of `TimerGameCalendar`, but are stored in `date.cpp`. To make code more readable and understandable, move them inside the class.

## Description

Simple search/replace for most part. Took some extra effort to make sure `date_func.h` is only included when actually needed.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10712, opened 2023-04-24
Codechange: move tick-counter into TimerGameTick

TrueBrain:
## Motivation / Problem

During #10706 I also wanted to move `_tick_counter`, but it required a bit more attention.

Either way, the counter should be part of `TimerGameTick`, as that almost says what it does: tick!

## Description

Move `_tick_counter` into `TimerGameTick`. There are a few observations:

The counter is mostly used for animation, so `TimerGameTick::counter` makes total sense. However, there are three cases where it is not used for that:

- Station's big tick.
- Vehicle travel time (used for link-graph).
- Town good distribution with `TCGM_BITCOUNT`.

The second (vehicle travel time) is a non-issue; it is just used by the link-graph for distribution. Any timer works there, so `TimerGameTick` is as good as any other.

The first and third are a bit weird. In this PR, they are fine in `TimerGameTick`. But looking at #10700, they would make more sense to be in `TimerGameEconomy`, as they are economy related. So it might be good in that PR to move them there.

## Checklist for review


src/station_cmd.cpp
```diff
@@ -3979,16 +3980,18 @@ void OnTick_Station()
...
-		 * at the same time. */
-		if ((_tick_counter + st->index) % STATION_ACCEPTANCE_TICKS == 0) {
+		/* Spread out big-tick over STATION_ACCEPTANCE_TICKS ticks. */
+		if ((TimerGameTick::counter + st->index) % STATION_ACCEPTANCE_TICKS == 0) {
 			/* Stop processing this station if it was deleted */
 			if (!StationHandleBigTick(st)) continue;
+		}
+
+		/* Spread out station animation over STATION_ACCEPTANCE_TICKS ticks. */
+		if ((TimerGameTick::counter + st->index) % STATION_ACCEPTANCE_TICKS == 0) {
```
2TallTyler, line 3994: Duplicate code, looks like a rebase error?

2TallTyler (commented):
Monotonic, now there's a fancy word 😃 (fine by me though).

src/station_cmd.cpp
TrueBrain: No, I split the animation from the big tick. I even updated the comment to be different :D

This so it is easier to move the first to economy later.

2TallTyler (approved):

PR #10716, opened 2023-04-24
Change: Set baseset description & video driver info size after initial sizing.

PeterN:
## Motivation / Problem

Baseset description and video driver info widgets do not get sized properly as their height is determined before the final width is known. This results in more vertical lines being used than necessary.

image

## Description

This is split off from #10674 as the scope of that kept increasing...

This PR updates the sizes of these widgets after the initial sizing is complete, so the final width is now known.

image

These images are huge but not side-by-side so it's not immediately obvious. Spacing underneath the baseset description :-)

## Limitations

<s>As the window height is changed after initial sizing, it's not quite centred by default any more...</s>

## Checklist for review


PeterN:
Added parameter to reinit to allow repositioning to default, and now it is centred.

TrueBrain (approved):
Sure. Not exactly sure why you want to reposition it, but in general this sounds like a good improvement :)

PR #10717, opened 2023-04-24
Codechange: Define Date/Year/Month/Day within TimerGameCalendar class

2TallTyler:
## Motivation / Problem

 #10700 splits time into two sets of variables. It would be awfully convenient if they were different data types so the compiler could catch mistakes.

## Description

- Move the definitions for these data types to within the TimerGameCalendar class
- Replace all references to the old data types with the new types, including two in `script_industry.cpp`
- Also includes a sneaky commit to change some Months in `graph_gui.cpp` from byte to the correct type 😉 

## Checklist for review

### DIFF
--- a/src/date_type.h
+++ b/src/date_type.h
@@ -10,14 +10,10 @@
 #ifndef DATE_TYPE_H
 #define DATE_TYPE_H
 
+#include "timer/timer_game_calendar.h"
 
-typedef int32  Date;      ///< The type to store our dates in
-typedef uint16 DateFract; ///< The fraction of a date we're in, i.e. the number of ticks since the last date changeover
 typedef int32  Ticks;     ///< The type to store ticks in
 
-typedef int32  Year;  ///< Type for the year, note: 0 based, i.e. starts at the year 0.
-typedef uint8  Month; ///< Type for the month, note: 0 based, i.e. 0 = January, 11 = December.
-typedef uint8  Day;   ///< Type for the day of the month, note: 1 based, first day of a month is 1.
 
 /**
  * 1 day is 74 ticks; TimerGameCalendar::date_fract used to be uint16 and incremented by 885. On
@@ -49,11 +45,11 @@ static const int INDUSTRY_CUT_TREE_TICKS  = INDUSTRY_PRODUCE_TICKS * 2; ///< cyc
  */
 
 /** The minimum starting year/base year of the original TTD */
-static const Year ORIGINAL_BASE_YEAR = 1920;
+static const TimerGameCalendar::Year ORIGINAL_BASE_YEAR = 1920;
 /** The original ending year */
-static const Year ORIGINAL_END_YEAR  = 2051;
+static const TimerGameCalendar::Year ORIGINAL_END_YEAR  = 2051;
 /** The maximum year of the original TTD */
-static const Year ORIGINAL_MAX_YEAR  = 2090;
+static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
 
 /**
  * Calculate the number of leap years till a given year.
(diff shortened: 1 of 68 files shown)


src/timer/timer_game_calendar.h
```diff
@@ -75,11 +75,18 @@ class TimerGameCalendar {
...
+	using Month = uint8; ///< Type for the month, note: 0 based, i.e. 0 = January, 11 = December.
+	using Day   = uint8; ///< Type for the day of the month, note: 1 based, first day of a month is 1.
+
+	static void SetDate(TimerGameCalendar::Date date, TimerGameCalendar::DateFract fract);
 
-	static Year year; ///< Current year, starting at 0.
+	static Year year;   ///< Current year, starting at 0.
 	static Month month; ///< Current month (0..11).
-	static Date date; ///< Current date in days (day counter).
+	static Date date;   ///< Current date in days (day counter).
```
TrueBrain: What are these random spaces?

Please don't do this .. it is such a nightmare. For example, why align 3 of the 4? That makes so little sense ...

2TallTyler: Year/Month/Day align with each other. 🙂

src/date_type.h
```diff
@@ -7,17 +7,14 @@
 
 /** @file date_type.h Types related to the dates in OpenTTD. */
 
+#include "timer/timer_game_calendar.h"
```
TrueBrain: I don't expect this include here. I would expect all soruce-files to include `timer_game_calendar.h` when they need access to Year/Month/Day from calendar. That way, if we add economy, it is also more clear which source files make use of what timers.

I know, a lot of work to get the includes right .. been there, got the t-shirt. But it does make more sense :)

src/date_type.h
```diff
@@ -104,13 +101,13 @@ static const Year MAX_YEAR  = 5000000;
...
-	Day   day;    ///< Day (1..31)
+	TimerGameCalendar::Year  year;   ///< Year (0...)
+	TimerGameCalendar::Month month;  ///< Month (0..11)
+	TimerGameCalendar::Day   day;    ///< Day (1..31)
 };
 
-static const Year  INVALID_YEAR  = -1; ///< Representation of an invalid year
-static const Date  INVALID_DATE  = -1; ///< Representation of an invalid date
+static const TimerGameCalendar::Year  INVALID_YEAR  = -1; ///< Representation of an invalid year
+static const TimerGameCalendar::Date  INVALID_DATE  = -1; ///< Representation of an invalid date
```
TrueBrain: So many silly alignments that now no longer make sense. Nothing you have to fix, but just to show how silly those space-alignments actually are.

src/network/network.cpp
```diff
@@ -1040,7 +1040,7 @@ void NetworkGameLoop()
 		/* Log the sync state to check for in-syncedness of replays. */
 		if (TimerGameCalendar::date_fract == 0) {
 			/* We don't want to log multiple times if paused. */
-			static Date last_log;
+			static TimerGameCalendar::Date last_log;
```
TrueBrain, line 1043: lol, is this timer based on TimerGameCalendar? Oy .. that will be a problem :D But this is why we mark these things .. not for this PR, but in later PRs if we make GameCalendar scalable, this needs to be addressed. I guess it should actually be `TimerGameTick`, but again, for another PR :)

src/timer/timer_game_calendar.h
```diff
@@ -7,11 +7,11 @@
 
 /** @file timer_game_calendar.h Definition of the game-calendar-timer */
 
+#include "stdafx.h"
```
TrueBrain: Please never ever include anything before the header protection.

For all header files: copyright, `/** */`, header-protection, includes, rest :)

src/timer/timer_game_calendar.h
TrueBrain: But Yeah / Month / Date / DateFract doesn't. Not sure what DateFract did to you .. but yeah, please just don't. Either align everything in a single block, or don't align. This is just adding spaces for the fact of adding spaces :)

TrueBrain (commented):
In general I think this is a good improvement on readability and intention of developer. Especially with Economy on the horizon.

Some minor comments, mostly the fact you include a timer in `date_type.h`. That feels wrong :)

src/date_type.h
2TallTyler: This is needed for consts like `ORIGINAL_BASE_YEAR`, but it does have the unfortunate side effect of cascading down to files which include `date_type.h`. I'll make sure they include `timer_game_calendar.h` directly.

src/network/network.cpp
2TallTyler: Yes, for sure. I'll leave this comment unresolved so we can see it later.

2TallTyler:
I found quite a few includes that previously had access to `date_type.h` cascaded through several headers, and needed direct access once I changed things around.

TrueBrain (approved):

PR #10718, opened 2023-04-25
Cleanup: remnants of OTTD_PRINTF, str_fmt and vseprintf

rubidium42:
## Motivation / Problem

Small bits and pieces of OTTD_PRINTF macros with environment specific formatting parameters, str_fmt and vseprintf remaining with single uses.

## Description

Replace them with fmt::format and remove the macros, str_fmt and vseprintf.

## Limitations

Breaks patches that use them.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -385,7 +385,8 @@ DEF_CONSOLE_CMD(ConSave)
 	}
 
 	if (argc == 2) {
-		char *filename = str_fmt("%s.sav", argv[1]);
+		std::string filename = argv[1];
+		filename += ".sav";
 		IConsolePrint(CC_DEFAULT, "Saving map...");
 
 		if (SaveOrLoad(filename, SLO_SAVE, DFT_GAME_FILE, SAVE_DIR) != SL_OK) {
@@ -393,7 +394,6 @@ DEF_CONSOLE_CMD(ConSave)
 		} else {
 			IConsolePrint(CC_INFO, "Map successfully saved to '{}'.", filename);
 		}
-		free(filename);
 		return true;
 	}
 
--- a/src/newgrf_debug_gui.cpp
+++ b/src/newgrf_debug_gui.cpp
@@ -403,7 +403,7 @@ struct NewGRFInspectWindow : Window {
 		offset -= this->vscroll->GetPosition();
 		if (offset < 0 || offset >= this->vscroll->GetCapacity()) return;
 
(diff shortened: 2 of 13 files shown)


src/console_cmds.cpp
```diff
@@ -385,15 +385,15 @@ DEF_CONSOLE_CMD(ConSave)
 	}
 
 	if (argc == 2) {
-		char *filename = str_fmt("%s.sav", argv[1]);
+		std::string filename = argv[1];
+		filename += "s.sav";
```
glx22:
```suggestion
		filename += ".sav";
```

LordAro: well that's not right...

src/newgrf_debug_gui.cpp
```diff
@@ -470,9 +463,9 @@ struct NewGRFInspectWindow : Window {
 				if (!avail) continue;
 
 				if (HasVariableParameter(niv->var)) {
-					this->DrawString(r, i++, "  %02x[%02x]: %08x (%s)", niv->var, param, value, niv->name);
+					this->DrawString(r, i++, fmt::format("  {:02x}[{:02x}]: {:08x} ({})", niv->var, param, value, niv->name));
```
glx22, line 466: Isn't there a thing with `0x` included in the `02` counting ?

rubidium42: If you use the `{:#02x}` modifier it is, as the `#` denotes adding the `0x`. With `{:02x}` it's equivalent to `%02x`.

src/strings.cpp
```diff
@@ -381,7 +382,7 @@ static char *FormatZerofillNumber(char *buff, int64 number, int64 count, const c
 
 static char *FormatHexNumber(char *buff, uint64 number, const char *last)
 {
-	return buff + seprintf(buff, last, "0x" OTTD_PRINTFHEX64, number);
+	return strecpy(buff, fmt::format("0x{:X}", number).c_str(), last);
```
rubidium42, line 385: The more efficient way would have been:
```
       auto [ptr, ec] = std::to_chars(buff, const_cast<char *>(last), number, 16);
       if (ec == std::errc()) buff = ptr;

       /* std::to_chars does not zero-terminate. */
       *buff = '\0';
       return buff;
```
However, that fails to build on OS X because it is not available until OS X 10.15.
So I have reverted to a simpler, though slightly less efficient manner. Which should be fine as it's only used in NewGRF error messages right now.

PeterN (approved):

PR #10719, opened 2023-04-25
Feature: opt-in survey when exiting a game

TrueBrain:
Binaries for testing available here:
https://www.openttd.org/downloads/openttd-branches/pr10719/latest

## Motivation / Problem

As mentioned in #10632 we would like to add an opt-in survey system to OpenTTD, for us to see how the game is being used.

There are many ways to play OpenTTD. There are tons of settings, you can load extensions, from AIs to NewGRFs. They all modify how the game plays.

From a developer perspective, we don't actually know what settings are heavily used, and which are never used. As such, we often find ourself in a discussion about whether something is actually useful. The people we talk most about, are not our average user; as it goes, the most vocal people are not the commoner, but more likely the niche. And although we like to promote as many playstyles as possible, it is easy to focus on those vocal people, instead of the 90%.

From a AI / NewGRF perspective, we don't actually know how many people play a game with my extension. And as such it is difficult to judge what is actually "popular", and what extensions are never really used. This means it is tricky on BaNaNaS to suggest any content.

For both there is a simple solution we have been putting off for a long time now: automated surveys. (one could also call it telemetry, but survey sounds a bit more like what our intent is).

First of all, of course survey will always be opt-in. But to gain a better understanding how OpenTTD is being used, we are considering adding survey.

There are a few groundrules:

- No personal data or data to track a specific user with.
- The user has to be able to validate the information that is being sent.
- Minimize the moments we send survey results.

## Description

On first start-up, the game will ask if you want to participate in our automated survey. You have to opt-in, and can easily opt-out (via the Options) at any time.

When opt-in, whenever you exit a game, a JSON blob will be send to the survey server hosted by OpenTTD. This JSON blob contains information that gives a global picture of the game just played:
- What settings were used
- How many humans vs AIs
- How long the game has been played
- Basic information about the OS / CPU

All this information is kept very generic, so there is no chance we send private information to our survey server. Nothing in the JSON blob could identify you as a person; it mostly tells about the game played. At any time you can see what the JSON blob includes, by pressing the "Preview Survey Results" button in-game.

The HTML part that belongs to this PR can be found here: https://github.com/OpenTTD/survey-web/pull/1

image
image
image

## Work in progress

Currently the JSON is collected, but that is it. This means there are a few things left to do:

- [x] Add the opt-in question on first start.
- [x] Allow an easy way to opt-out again (via Options?).
- [x] Link to the privacy statement or something, explaining what is in the JSON blob, and what we do with the information.
- [x] Add JSON library to Emscripten.
- [x] Generate SavegameID when loading a scenario.
- [x] Add more randomness for the SavegameID, by doing a sleep and using the clocks nanoseconds; that should be rather random.
- [x] Generate secrets per release in the CI. (a secret will be a JWT token signed by the survey server, which includes things like what OpenTTD version etc)
- [x] Implement backend: collector (the part that receives the surveys and stores it in a bucket)
- [ ] ~Implement backend: analyzer (the part that looks at all the surveys and takes conclusions based on it)~ (postponed till we have more data)
- [ ] ~Implement backend: compactor (the part that takes the surveys of a period, and compresses them (to save on disk-space)~

## Limitations

If you don't have nlohmann-json available during compile-time, you get from cmake an ENCOURAGE warning, but otherwise it compiles just fine. Additionally, in-game there is no mention of a survey in those cases (as you can't participate anyway).

Emscripten can't make HTTPS connections (yet), so it doesn't participate in survey.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1041,6 +1041,14 @@ STR_GAME_OPTIONS_GUI_SCALE_3X                                   :3x
 STR_GAME_OPTIONS_GUI_SCALE_4X                                   :4x
 STR_GAME_OPTIONS_GUI_SCALE_5X                                   :5x
 
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_FRAME                       :{BLACK}Automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY                             :{BLACK}Participate in automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_TOOLTIP                     :{BLACK}When enabled, OpenTTD will transmit a survey when leaving a game
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_LINK                        :{BLACK}About survey and privacy
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_LINK_TOOLTIP                :{BLACK}This opens a browser with more information about the automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_PREVIEW                     :{BLACK}Preview survey result
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_PREVIEW_TOOLTIP             :{BLACK}Show the survey result of the current running game
+
 STR_GAME_OPTIONS_GRAPHICS                                       :{BLACK}Graphics
 
 STR_GAME_OPTIONS_REFRESH_RATE                                   :{BLACK}Display refresh rate
@@ -2401,6 +2409,13 @@ STR_NETWORK_ASK_RELAY_TEXT                                      :{YELLOW}Failed 
 STR_NETWORK_ASK_RELAY_NO                                        :{BLACK}No
 STR_NETWORK_ASK_RELAY_YES_ONCE                                  :{BLACK}Yes, this once
 STR_NETWORK_ASK_RELAY_YES_ALWAYS                                :{BLACK}Yes, don't ask again
+
+STR_NETWORK_ASK_SURVEY_CAPTION                                  :Participate in automated survey?
+STR_NETWORK_ASK_SURVEY_TEXT                                     :Would you like to participate in the automated survey?{}OpenTTD will transmit a survey when leaving a game.{}You can change this at any time under "Game Options".
+STR_NETWORK_ASK_SURVEY_PREVIEW                                  :Preview survey result
+STR_NETWORK_ASK_SURVEY_LINK                                     :About survey and privacy
+STR_NETWORK_ASK_SURVEY_NO                                       :No
+STR_NETWORK_ASK_SURVEY_YES                                      :Yes
 
 STR_NETWORK_SPECTATORS                                          :Spectators
 
@@ -4654,10 +4669,11 @@ STR_TEXTFILE_WRAP_TEXT_TOOLTIP                                  :{BLACK}Wrap the
 STR_TEXTFILE_VIEW_README                                        :{BLACK}View readme
 STR_TEXTFILE_VIEW_CHANGELOG                                     :{BLACK}Changelog
 STR_TEXTFILE_VIEW_LICENCE                                       :{BLACK}Licence
-###length 3
+###length 4
 STR_TEXTFILE_README_CAPTION                                     :{WHITE}{STRING} readme of {RAW_STRING}
 STR_TEXTFILE_CHANGELOG_CAPTION                                  :{WHITE}{STRING} changelog of {RAW_STRING}
 STR_TEXTFILE_LICENCE_CAPTION                                    :{WHITE}{STRING} licence of {RAW_STRING}
+STR_TEXTFILE_SURVEY_RESULT_CAPTION                              :{WHITE}Preview of survey result
 
 
 # Vehicle loading indicators
--- a/src/misc.cpp
+++ b/src/misc.cpp
@@ -31,9 +31,11 @@
 #include "town_kdtree.h"
 #include "viewport_kdtree.h"
 #include "newgrf_profiling.h"
+#include "3rdparty/md5/md5.h"
 
 #include "safeguards.h"
 
+std::string _savegame_id; ///< Unique ID of the current savegame.
 
 extern TileIndex _cur_tileloop_tile;
 extern void MakeNewgameSettingsLive();
@@ -55,6 +57,40 @@ void InitializeCompanies();
 void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
+
+/**
+ * Generate an unique ID.
+ *
+ * It isn't as much of an unique ID as we would like, but our random generator
+ * can only produce 32bit random numbers.
+ * That is why we combine InteractiveRandom with the current (steady) clock.
+ * The first to add a bit of randomness, the second to ensure you can't get
+ * the same unique ID when you run it twice from the same state at different
+ * times.
+ *
+ * This makes it unlikely that two users generate the same ID for different
+ * subjects. But as this is not an UUID, so it can't be ruled out either.
+ */
+std::string GenerateUid(std::string_view subject)
+{
+	auto current_time = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now().time_since_epoch()).count();
+	std::string coding_string = fmt::format("{}{}{}", InteractiveRandom(), current_time, subject);
+
+	Md5 checksum;
+	uint8 digest[16];
+	checksum.Append(coding_string.c_str(), coding_string.length());
+	checksum.Finish(digest);
+
+	return MD5SumToString(digest);
+}
+
+/**
+ * Generate an unique savegame ID.
+ */
+void GenerateSavegameId()
+{
+	_savegame_id = GenerateUid("OpenTTD Savegame ID");
+}
 
 void InitializeGame(uint size_x, uint size_y, bool reset_date, bool reset_settings)
 {
--- a/src/network/core/http_winhttp.cpp
+++ b/src/network/core/http_winhttp.cpp
@@ -131,7 +131,8 @@ void NetworkHTTPRequest::WinHttpCallback(DWORD code, void *info, DWORD length)
 
 			/* If there is any error, we simply abort the request. */
 			if (status_code >= 400) {
-				Debug(net, 0, "HTTP request failed: status-code {}", status_code);
+				/* No need to be verbose about rate limiting. */
+				Debug(net, status_code == HTTP_429_TOO_MANY_REQUESTS ? 1 : 0, "HTTP request failed: status-code {}", status_code);
 				this->finished = true;
 				this->callback->OnFailure();
 				return;
@@ -242,7 +243,9 @@ void NetworkHTTPRequest::Connect()
 	if (data.empty()) {
 		WinHttpSendRequest(this->request, WINHTTP_NO_ADDITIONAL_HEADERS, 0, WINHTTP_NO_REQUEST_DATA, 0, 0, reinterpret_cast<DWORD_PTR>(this));
 	} else {
-		WinHttpSendRequest(this->request, L"Content-Type: application/x-www-form-urlencoded\r\n", -1, const_cast<char *>(data.c_str()), static_cast<DWORD>(data.size()), static_cast<DWORD>(data.size()), reinterpret_cast<DWORD_PTR>(this));
+		/* When the payload starts with a '{', it is a JSON payload. */
+		LPCWSTR content_type = StrStartsWith(data, "{") ? L"Content-Type: application/json\r\n" : L"Content-Type: application/x-www-form-urlencoded\r\n";
+		WinHttpSendRequest(this->request, content_type, -1, const_cast<char *>(data.c_str()), static_cast<DWORD>(data.size()), static_cast<DWORD>(data.size()), reinterpret_cast<DWORD_PTR>(this));
 	}
 }
 
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -18,6 +18,7 @@
 #include "network_content.h"
 #include "network_server.h"
 #include "network_coordinator.h"
+#include "network_survey.h"
 #include "../gui.h"
 #include "network_udp.h"
 #include "../window_func.h"
@@ -38,6 +39,7 @@
 #include "../timer/timer.h"
 #include "../timer/timer_window.h"
 #include "../timer/timer_game_calendar.h"
+#include "../textfile_gui.h"
 
 #include "../widgets/network_widget.h"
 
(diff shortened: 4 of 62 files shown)


src/network/network_survey.cpp
```diff
@@ -0,0 +1,356 @@
...
+ * @param table The settings table to convert.
+ * @param object The object to get the settings from.
+ */
+static void SettingsTableToJson(nlohmann::json &survey, const SettingTable &table, void *object)
+{
+	char buf[512];
+	for (auto &desc : table) {
+		const SettingDesc *sd = GetSettingDesc(desc);
+		/* Skip any old settings we no longer save/load. */
+		if (!SlIsObjectCurrentlyValid(sd->save.version_from, sd->save.version_to)) continue;
```
glx22, line 113: Maybe also skip settings not different than default value ?

TrueBrain: That would mean we need to keep a backend-side store of default values; I see what you mean, and it would save (a lot of) storage, but it would also complicate the flow of information .. for every new release we also have to tell how settings are doing etc.

Meh. Both have their merits :)

krysclarke:
Caveat: I haven't looked at the code for this PR.

What's stopping a person scripting the running and stopping of OTTD to spam your server with multiple survey results to increase the popularity of their chosen settings by reporting multiple responses?
The 'popularity' could be rated on how many hours of gameplay each setting has been used for, instead of the number of responses.
Generating a unique device ID that possible, but how determined a spammer do you want to defend against?
At somewhat of an extreme, an actor could record the JSON blob sent, and write a utility to spam that repeatedly to the survey server, so I'd suggest that each blob has a unique ID (probably incorporating a timestamp), as well as server-side processing to defend against timestamp and gameplay duration manipulation (e.g. a transmission timestamp that's within the last 5 minutes when received by the survey server is considered 'valid'; two JSON blobs received from the same device ID, say 1 minute apart, with suitably adjusted timestamps but with gameplay duration of >10minutes for the 2nd report, would be treated as 'invalid').

What if I changed a game setting during play?  (e.g. Infrastructure costs can be toggled on an off, at any time)
How will that be reported in the survey?
Ideally, I feel it should send information about all values that were used during that session and the time they were used for.

src/network/network_survey.cpp
TrueBrain: Gave this some thought, and I rather have a maintenance free system, with as little moving parts as I can. So, I think we should go for sending all settings :)

We compress survey results in the backend, which is doing an amazing job (1:100 in early benchmarks). So it is just a bit of bandwidth we waste; and with ~20KiB, that shouldn't be an issue.

src/lang/english.txt
```diff
@@ -1041,6 +1041,14 @@ STR_GAME_OPTIONS_GUI_SCALE_3X                                   :3x
 STR_GAME_OPTIONS_GUI_SCALE_4X                                   :4x
 STR_GAME_OPTIONS_GUI_SCALE_5X                                   :5x
 
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_FRAME                       :{BLACK}Automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY                             :{BLACK}Participate in automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_TOOLTIP                     :{BLACK}When enabled, OpenTTD will transmit an survey when leaving a game
```
michicc: `a survey`

src/lang/english.txt
```diff
@@ -2401,6 +2409,11 @@ STR_NETWORK_ASK_RELAY_NO                                        :{BLACK}No
 STR_NETWORK_ASK_RELAY_YES_ONCE                                  :{BLACK}Yes, this once
 STR_NETWORK_ASK_RELAY_YES_ALWAYS                                :{BLACK}Yes, don't ask again
 
+STR_NETWORK_ASK_SURVEY_CAPTION                                  :{WHITE}Participate in automated survey?
+STR_NETWORK_ASK_SURVEY_TEXT                                     :{BLACK}Would you like to participate in the automated survey?{}OpenTTD will transmit an survey when leaving a game.{}See "Game Options" for more information and privacy statement.
```
michicc: `a survey`

src/network/network_survey.cpp
```diff
@@ -0,0 +1,409 @@
...
+		std::vector<std::string> flags;
+		if (HasBit(c->flags, GCF_SYSTEM)) flags.push_back("system");
+		if (HasBit(c->flags, GCF_UNSAFE)) flags.push_back("unsafe");
+		if (HasBit(c->flags, GCF_STATIC)) flags.push_back("static");
+		if (HasBit(c->flags, GCF_COMPATIBLE)) flags.push_back("compatible");
+		if (HasBit(c->flags, GCF_COPY)) flags.push_back("copy");
+		if (HasBit(c->flags, GCF_INIT_ONLY)) flags.push_back("init_only");
+		if (HasBit(c->flags, GCF_RESERVED)) flags.push_back("reserved");
+		if (HasBit(c->flags, GCF_INVALID)) flags.push_back("invalid");
+		grf["flags"] = flags;
```
frosch123: I think the only flag that makes sense in the survey is `GCF_STATIC`. The rest is internal state during NewGRF initialisation.
```suggestion
		grf["is_static"] = HasBit(c->flags, GCF_STATIC);
```

src/saveload/saveload.h
```diff
@@ -353,6 +353,8 @@ enum SaveLoadVersion : uint16 {
 	SLV_LINKGRAPH_SECONDS,                  ///< 308  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	SLV_AI_START_DATE,                      ///< 309  PR#10653 Removal of individual AI start dates and added a generic one.
 
+	SLV_SAVEGAME_UUID,                      ///< 310  PR#????? Add an unique UUID to every savegame (used to deduplicate surveys).
```
```suggestion
	SLV_SAVEGAME_UUID,                      ///< 310  PR#10719 Add a unique UUID to every savegame (used to deduplicate surveys).
```

src/misc.cpp
```diff
@@ -56,6 +58,41 @@ void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
 
+/**
+ * Generate an UUID.
```
rubidium42: This does not conform to the UUID standard, so I'd also not call it that way. Having said that, I doubt you want to generate an actual UUID with all complications of the specification associated with that. So like our speed is in km-ish/h, this is more a unique-ish ID.
By the way, I'm fine with this implementation just not the name. It's good enough for what we're trying to do.

src/misc.cpp
```diff
@@ -56,6 +58,41 @@ void InitializeCheats();
...
+ *
+ * It isn't as much of an UUID as we would like, but our random generator can
+ * only produce 32bit random numbers.
+ * That is why we combine InteractiveRandom with the current (steady) clock.
+ * The first to add a bit of randomness, the second to ensure you can't get
+ * the same UUID when you run it twice from the same state.
+ */
+std::string GenerateUuid(std::string_view subject)
+{
+	auto current_time = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now().time_since_epoch()).count();
```
rubidium42: I'd use nanoseconds instead of milliseconds. It'll give you more random-ish bits, unless the underlying clock is in milliseconds though I reckon that chance is pretty low.

src/misc.cpp
```diff
@@ -56,6 +58,41 @@ void InitializeCheats();
...
+	uint8 digest[16];
+	checksum.Append(coding_string.c_str(), coding_string.length());
+	checksum.Finish(digest);
+
+	std::string hex_output;
+	for (int i = 0; i < 16; i++) {
+		hex_output += fmt::format("{:02x}", digest[i]);
+	}
+
+	return hex_output;
```
rubidium42:
```suggestion
	return MD5SumToString(digest);
```

src/misc.cpp
```diff
@@ -56,6 +58,41 @@ void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
 
+/**
+ * Generate an UUID.
+ *
+ * It isn't as much of an UUID as we would like, but our random generator can
+ * only produce 32bit random numbers.
+ * That is why we combine InteractiveRandom with the current (steady) clock.
+ * The first to add a bit of randomness, the second to ensure you can't get
```
rubidium42, line 67: Technically the clock does not ensure it; if you achieve to start the same savegame within 1 millisecond...
In any case, the MD5 checksum afterwards undoes any ensuring you do before that.
Since both `InteractiveRandom` and `steady_clock` are essentially time based (`InteractiveRandom` gets seeded by time), it is plausible that two processes starting at the exact same time yield the same result. So I'd just phrase the comment slightly more cautiously. What about the following:
```
This doesn't create a UUID as per the specifications (RFC 4122). Not because it is complex,
but because it leaks identifiable information, i.e. the IEEE address / MAC address. This makes
the specific computer from which multiple reports were sent identifiable.
It does create a string of 32 hexadecimal digits that are based on the subject, current time,
and the `InteractiveRandom()`, which is a pseudo-random number generator seeded with the time
OpenTTD was started plus the number of times `InteractiveRandom()` was called. These inputs
make it highly unlikely that the same identifier is generated twice, while not containing any
identifiable information. 
Finally this data is passed through a MD5 hash to make it very difficult to get back the
original input adding another layer against potentially leaking identifiable information.
```

src/network/core/http_winhttp.cpp
```diff
@@ -131,7 +131,8 @@ void NetworkHTTPRequest::WinHttpCallback(DWORD code, void *info, DWORD length)
 
 			/* If there is any error, we simply abort the request. */
 			if (status_code >= 400) {
-				Debug(net, 0, "HTTP request failed: status-code {}", status_code);
+				/* status_code == 429 means we are rate-limited; no need to be verbose about that. */
+				Debug(net, status_code == 429 ? 1 : 0, "HTTP request failed: status-code {}", status_code);
```
rubidium42: Magic number; should be a static const, probably in http.h.

src/network/network_gui.cpp
```diff
@@ -2514,3 +2517,131 @@ void ShowNetworkAskRelay(const std::string &server_connection_string, const std:
...
+
+	SurveyResultTextfileWindow(TextfileType file_type) : TextfileWindow(file_type)
+	{
+		this->lines.clear();
+		auto result = _survey.CreatePayload(NetworkSurveyHandler::Reason::PREVIEW, true);
+
+		/* Put string in C-style string. */
+		this->text = ReallocT(this->text, result.size() + 1);
+		memcpy(this->text, result.c_str(), result.size());
+		this->text[result.size()] = '\0';
```
rubidium42: Not that I'm fond of it... but... why not free and stredup? Now you're essentially hiding the C-style string allocation.
But alas... #10772 would make this whole problem go away I reckon.

src/os/windows/survey_win.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "../../safeguards.h"
+
+void SurveyOS(nlohmann::json &json)
+{
+	_OSVERSIONINFOA os;
+	os.dwOSVersionInfoSize = sizeof(os);
+	GetVersionExA(&os);
+
+	json["os"] = "Windows";
+	json["release"] = fmt::format("{}.{}.{} ({})", (int)os.dwMajorVersion, (int)os.dwMinorVersion, (int)os.dwBuildNumber, os.szCSDVersion);
```
rubidium42: Is the casting really needed?

src/network/network_survey.cpp
```diff
@@ -0,0 +1,409 @@
...
+	for (GRFConfig *c = _grfconfig; c != nullptr; c = c->next) {
+		auto grfid = fmt::format("{:08x}", BSWAP32(c->ident.grfid));
+		auto &grf = survey[grfid];
+
+		std::string md5sum;
+		for (int i = 0; i < 16; i++) {
+			md5sum += fmt::format("{:02x}", c->ident.md5sum[i]);
+		}
+
+		grf["md5sum"] = md5sum;
```
rubidium42:
```suggestion
		grf["md5sum"] = MD5SumToString(c->ident.md5sum);
```

src/network/network_survey.cpp
```diff
@@ -0,0 +1,409 @@
...
+		auto &game = survey["game"];
+		game["ticks"] = TimerGameTick::counter;
+		game["time"] = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - _switch_mode_time).count();
+		SurveyCompanies(game["companies"]);
+		SurveySettings(game["settings"]);
+		SurveyGrfs(game["grfs"]);
+		SurveyGameScript(game["game_script"]);
+	}
+
+	return survey.dump(for_preview ? 4 : -1);
```
rubidium42: It's not obvious what the 4 or -1 do from the code.
`static const int PREVIEW_INDENT_LEVEL = 4;` or something similar?

src/network/network_survey.cpp
```diff
@@ -0,0 +1,409 @@
...
+{
+	survey["network"] = _networking ? (_network_server ? "server" : "client") : "no";
+	if (_current_language != nullptr) {
+		std::string language_basename(_current_language->file);
+		auto e = language_basename.rfind(PATHSEPCHAR);
+		if (e != std::string::npos) {
+			language_basename.erase(0, e + 1);
+		}
+
+		survey["language"] = language_basename;
```
rubidium42: Can't you assign a `std::string_view`? That would save you from moving characters around in `language_basename`.

src/network/network_survey.cpp
```diff
@@ -0,0 +1,409 @@
...
+#endif
+		;
+	survey["endian"] =
+#if (TTD_ENDIAN == TTD_LITTLE_ENDIAN)
+			"little"
+#else
+			"big"
+#endif
+		;
+	survey["dedicated"] =
```
rubidium42: How important is it whether the client was compiled as dedicated server? If you just install the standard OpenTTD, then this will be "no" and we'd still have to look to the video driver for the actual information.
If it is important, then maybe rename it to "dedicated_build", to make it clear that it's about the choice made at building the game not when starting the game.

src/misc.cpp
TrueBrain: I think you have been overthinking this, and I am not going to put that huge block of comments there :D

I changed UUID to "unique ID", and left most of the wording. The idea is just that this isn't as unique as we would like; ideally we switch to libuuid, but that is something for another PR :)

>  it is plausible that two processes starting at the exact same time yield the same result

That is very much not true. It is VERY unlikely this happens. Being able to load a savegame at the exact same millisecond (and now nanosecond) while you also started OpenTTD at the exact same time? Yeah, you are not going to make that in any realistic scenario :)
These kind of functions are not meant to avoid abuse (as you can just change the source code and make it return `42` at all times). They are merely meant that 2 users in different places in the world have a very very low probability to generate the same identifier :) I added that in the comment of the function now!

PS: the md5 is not meant to make this non-identifiable, as all inputs are already not identifiable. But it is meant to output a known-length string :) That is all :P

src/network/network_gui.cpp
TrueBrain: There is a reason I started #10772, yes :) But it depends on #10776 .. and down the rabbithole we go :) Will be fine before this gets undrafted :)

src/network/network_survey.cpp
TrueBrain: Personally, don't actually care. But it was in the crashlog .. so someone once thought it was important :P `dedicated_build` it is :)

(I think it is good to keep it, on second thought .. shows us a bit how often people make a dedicated server build, as in: is it worth supporting the option. I think it is, and this would make it statistical :D)

TrueBrain:
This PR is now ready for review. Backend is implemented and online.

Test binaries can be found here: https://www.openttd.org/downloads/openttd-branches/pr10719/latest . Please make sure you have a version from 2023-05-11 or later.

For more information on the survey-keys, please see https://github.com/OpenTTD/survey-web/pull/4 (and the README changes there-in).

src/network/network_gui.cpp
```diff
@@ -2428,7 +2430,8 @@ struct NetworkAskRelayWindow : public Window {
 	{
 		if (widget == WID_NAR_TEXT) {
 			*size = GetStringBoundingBox(STR_NETWORK_ASK_RELAY_TEXT);
-			size->height = GetStringHeight(STR_NETWORK_ASK_RELAY_TEXT, size->width - WidgetDimensions::scaled.frametext.Horizontal()) + WidgetDimensions::scaled.frametext.Vertical();
+			size->width += WidgetDimensions::scaled.frametext.Horizontal();
```
glx22: Seems `-` became `+`

src/network/network_gui.cpp
```diff
@@ -2514,3 +2517,119 @@ void ShowNetworkAskRelay(const std::string &server_connection_string, const std:
...
+	{
+		this->parent = parent;
+		this->InitNested(0);
+	}
+
+	void UpdateWidgetSize(int widget, Dimension *size, const Dimension &padding, Dimension *fill, Dimension *resize) override
+	{
+		if (widget == WID_NAS_TEXT) {
+			*size = GetStringBoundingBox(STR_NETWORK_ASK_SURVEY_TEXT);
+			size->width += WidgetDimensions::scaled.frametext.Horizontal();
```
glx22, line 2536: Maybe same as above comment.

src/os/unix/survey_unix.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file survey_osx.cpp OSX implementation of OS-specific survey information. */
```
rubidium42: Sloppy paste

src/os/windows/survey_win.cpp
```diff
@@ -0,0 +1,31 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file survey_osx.cpp OSX implementation of OS-specific survey information. */
```
rubidium42: Sloppy paste

src/saveload/saveload.h
```diff
@@ -353,6 +353,7 @@ enum SaveLoadVersion : uint16 {
 	SLV_EXTEND_VEHICLE_RANDOM,              ///< 310  PR#10701 Extend vehicle random bits.
 	SLV_EXTEND_ENTITY_MAPPING,              ///< 311  PR#10672 Extend entity mapping range.
 	SLV_DISASTER_VEH_STATE,                 ///< 312  PR#10798 Explicit storage of disaster vehicle state.
+	SLV_SAVEGAME_UUID,                      ///< 313  PR#10719 Add an unique ID to every savegame (used to deduplicate surveys).
```
rubidium42: UUID -> ID?

src/lang/english.txt
```diff
@@ -1041,6 +1041,14 @@ STR_GAME_OPTIONS_GUI_SCALE_3X                                   :3x
 STR_GAME_OPTIONS_GUI_SCALE_4X                                   :4x
 STR_GAME_OPTIONS_GUI_SCALE_5X                                   :5x
 
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_FRAME                       :{BLACK}Automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY                             :{BLACK}Participate in automated survey
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_TOOLTIP                     :{BLACK}When enabled, OpenTTD will transmit a survey when leaving a game
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_LINK                        :{BLACK}About survey and privacy
+STR_GAME_OPTIONS_PARTICIPATE_SURVEY_LINK_TOOLTIP                :{BLACK}This opens a browser to https://survey.openttd.org/privacy with more information about the automated survey
```
```suggestion
STR_GAME_OPTIONS_PARTICIPATE_SURVEY_LINK_TOOLTIP                :{BLACK}This opens a browser to {RAW_STRING} with more information about the automated survey
```
and fill with `NETWORK_SURVEY_DETAILS_LINK` (but I'm not sure it's possible for tooltips with the window system)

Anyway the URI is wrong here.

src/textfile_type.h
```diff
@@ -18,7 +18,9 @@ enum TextfileType {
 	TFT_CHANGELOG,          ///< NewGRF changelog
 	TFT_LICENSE,            ///< NewGRF license
 
-	TFT_END,
+	TFT_END, // This marker is used to generate the above three buttons in sequence by various of places in the code.
```
rubidium42: Having end not meaning end might trip someone up in the future. Maybe introduce `TFT_CONTENT_BEGIN`/`TFT_CONTENT_END` for the current iterations, that are clearly not meant to be used for all files?

src/network/network_gui.cpp
TrueBrain: Yeah, this fixes an annoying bug; I could have PRed that separately btw .. the original behaviour is broken :P The width needs to increased, not decreased! :D

src/network/network_survey.cpp
```diff
@@ -0,0 +1,400 @@
...
+	return "";
+#else
+	nlohmann::json survey;
+
+	survey["schema"] = NETWORK_SURVEY_VERSION;
+	survey["reason"] = reason;
+
+	/* id and key are used to authenticate surveys results and are not meant
+	 * to be previewed by the user. */
+	if (!for_preview) {
```
rubidium42: Hiding things you are going to send doesn't feel right to me, if you add the UI to preview it to be open and clear about what it about to be sent.
Side note: do we mention the URL this gets uploaded to, and should we maybe show that as well in the preview UI?

TrueBrain: Security vs transparency is always a battle. I have no problems making the `id` visible, but they `key` I rather don't show. But I can show: `key: <redacted>`, if that works for you?

Of course it is trivial for anyone with knowledge to figure out the key anyway, but at least it makes it more difficult for the other 90% :)

src/lang/english.txt
glx22: Seems the only way to handle that is `OnTooltip()`

TrueBrain: The URL should never been in here. It isn't helpful (as you can't copy it) anyway :P So I just removed the URL :)

glx22: Even simpler :)

src/network/network_survey.cpp
TrueBrain: image

ID visible, key redacted. Good enough?

Edit: changed it into `(redacted)` instead. Just because that looks way better :D

rubidium42 (approved):

PR #10722, opened 2023-04-26
Change: [CI] Automatically disable asserts when building a stable release

LordAro:
## Motivation / Problem

Remembering to turn off asserts for official releases is often a very last minute thing that has been forgotten entirely in the past on several occasions (most recently, 13.0)

## Description

Use the fact that when we build releases we already have a pre-written .ottdrev available. Parse that and turn off the cmake option as appropriate

## Limitations

Not fully tested:

```
 $ echo -e "foo\tbar\tbaz\tquix\tqneez\t1\tflibble" > .ottdrev
 $ IS_STABLE_TAG=$(if [ "$(cat .ottdrev | cut -f 6 -d$'\t')" = "1" ]; then echo "OFF"; else echo "ON"; fi)
 $ echo $IS_STABLE_TAG
OFF
 $ echo -e "foo\tbar\tbaz\tquix\tqneez\t0\tflibble" > .ottdrev
 $ IS_STABLE_TAG=$(if [ "$(cat .ottdrev | cut -f 6 -d$'\t')" = "1" ]; then echo "OFF"; else echo "ON"; fi)
 $ echo $IS_STABLE_TAG
ON

```

## Checklist for review

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -104,6 +104,9 @@ jobs:
 
     - name: Build
       run: |
+        # Automatically disable asserts for stable releases
+        IS_STABLE_TAG=$(if [ "$(cat .ottdrev | cut -f 6 -d$'\t')" = "1" ]; then echo "OFF"; else echo "ON"; fi)
+
         mkdir -p build
         cd build
 
@@ -112,6 +115,7 @@ jobs:
           -DCMAKE_TOOLCHAIN_FILE=/vcpkg/scripts/buildsystems/vcpkg.cmake \
           -DCMAKE_BUILD_TYPE=RelWithDebInfo \
           -DOPTION_PACKAGE_DEPENDENCIES=ON \
+          -DOPTION_USE_ASSERTS=${IS_STABLE_TAG} \
           # EOF
         echo "::endgroup::"
 
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -92,6 +92,9 @@ jobs:
 
     - name: Build arm64
       run: |
+        # Automatically disable asserts for stable releases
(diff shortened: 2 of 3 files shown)


cmake/Options.cmake
```diff
@@ -52,11 +52,32 @@ function(set_options)
...
+                        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+                        ERROR_QUIET
+        )
+        if(NOT ${TAG} STREQUAL "")
+            string(REGEX REPLACE "^[0-9.]+$" "" STABLETAG "${TAG}")
+            if(STABLETAG STREQUAL "")
+                set(DEFAULT_OPTION_USE_ASSERTS OFF)
+            endif()
+        endif()
+    endif()
```
TrueBrain: I like the approach, I very much dislike the location. Should be outside this function for sure (which is fine in CMake, you just need to do this before this function is called).

TrueBrain (dismissed):
Assuming you finish testing :) Let's be sure it actually works as advertised :D

TrueBrain:
One note: I was thinking about Linux distros that build OpenTTD. Currently, they can just build the tag, and the right asserts yes/no come out. After this PR, they have to flip that switch themselves.

I am not sure what to think about that .. good .. bad .. but it is different. Just wanted to note that :)

TrueBrain:
Another approach to this would be to use a Workflow to create a release, instead of doing it via the release page.

That way, the workflow can check what kind of release this is, and make the correct changes in the form of a commit before making the actual release.

As we recently found out, besides asserts, we also have to set the CMake project version, which we tend to forget.

Another thing that comes to mind when I think along these ways, that the workflow can also make a changelog.txt draft, push that in a PR, and waits for it to be merged.

Dunno .. just thinking out loud here :)

LordAro:
This probably ultimately isn't the right direction to go

PR #10723, opened 2023-04-26
Change: [CI] Use a custom name for matrix runs

TrueBrain:
## Motivation / Problem

By default, GitHub adds all arguments of the matrix between (). This is fine sometimes, but in other times it becomes a very lengthy line.

## Description

With this commit, we decide what is between those (), making it a lot more readable.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -69,24 +69,29 @@ jobs:
         echo "::endgroup::"
 
   linux:
-    name: Linux
-
     strategy:
       fail-fast: false
       matrix:
         include:
-        - compiler: clang
+        - name: Clang
+          compiler: clang
           cxxcompiler: clang++
-          libsdl: libsdl2-dev
-        - compiler: gcc
+          libraries: libsdl2-dev
+        - name: GCC - SDL2
+          compiler: gcc
           cxxcompiler: g++
-          libsdl: libsdl2-dev
-        - compiler: gcc
+          libraries: libsdl2-dev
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-build.yml
```diff
@@ -69,24 +69,29 @@ jobs:
...
 
   linux:
-    name: Linux
-
     strategy:
       fail-fast: false
       matrix:
         include:
-        - compiler: clang
+        - name: Ckang
```
glx22: oups

.github/workflows/ci-build.yml
```diff
@@ -349,6 +352,8 @@ jobs:
           - msystem: MINGW32
             arch: i686
 
+    name: MingW (${{ matrix.arch }})
```
LordAro:
```suggestion
    name: MinGW (${{ matrix.arch }})
```

rubidium42:
Is the configuration of which checks are required and which aren't required in the repository, or is that something separately? I can't find it, so I'm assuming it's the latter.

If it's the former, then that needs to be updated as it is awaiting a lot of required "checks" that will never be run. 

If it's the latter, then someone needs to change the "required checks configuration" and all PRs after this gets merged need to be rebased to pass these changed checks, right?

TrueBrain:
After this PR is approved, I need to change some GitHub settings so the CI required labels change; and yes, after that all PRs need to rebase :)

rubidium42 (approved):

PR #10725, opened 2023-04-26
Codechange: [CMake] Switch GRF generation from NFO to NML

glx22:
## Motivation / Problem
NFO is hard to read and hard to write.

## Description
All NFO has been rewritten to NML (direct translation).
I modified CreateGRF.cmake to follow preprocessor output format for file:line tracking (nmlc nicely supports it).
Needed to rework md5 handling for orig_extra.grf as nmlc generates format 2, while the old grfcodec command was generating format 1. Now instead of calculating md5 of the whole file, it's using the md5 generated during build of the grf.

## Limitations
It's a direct translation from NFO, so not always the best choice in nml (mainly multiple identical climate checks in some files).
Also the OTTD version checks could probably be rewritten.

## Checklist for review


frosch123:
What would the workflow for adding a new GUI sprite or Action5-type look like in the future?
Would one have to release an updated NML before merging the feature into OTTD?

glx22:
No, `nmlc` warns with `Exceeding the limit of {:d} sprites for sprite replacement type '{}', extra sprites may be ignored`, but still happily compiles. And `nmlc` is actually ran with `--quiet` so nothing will be shown.

michicc:
If I read the nml code correctly, adding a new Action 5 type would need a NML patch. Now, I have no idea how often that would be a problem, as (going by the NewGRF specs wiki) the last new A5 type was with OTTD 1.5.

While I don't have a strong opinion on this, I kinda like the fact that right now this is a total non-issue as grfcodec itself does not validate anything.

PR #10726, opened 2023-04-27
Fix: Moving cargo definition slots prevents normal town/industry behaviour.

PeterN:
## Motivation / Problem

<s>NewGRFs can define additional passenger or mail type cargoes (with TE_PASSENGERS/TE_MAIL or CC_PASSENGERS/CC_MAIL) but they don't do a lot.</s>

NewGRFs can move passengers from CT_PASSENGERS or mail from CT_MAIL and everything looks fine in the UI but it isn't produced by towns, and aircraft cannot carry mail (if there was any to move...)

Subsidies...

## Description

<s>This changeset creates a list of passenger and mail type cargoes, and allows towns to produce the first available type.

Aircraft with mail use the first available mail type.</s>

This changeset creates a cargo translation map between base climate cargo types and the active cargo types, and uses the translation map when determining what cargo to accept/produce for towns/default houses/default industries.

Vehicles already used a cargo translation lookup, which is now switched to use the map, and the hardcoded CT_MAIL for aircraft is now mapped as well.

This change no longer changes the game's behaviour when multiple cargoes providing town effect or cargo classes are installed.

## Limitations

WIP!

## Checklist for review


PeterN:
Possibly cargo classes / town effect should have more impact on what happens with cargo produced by towns. While individual houses have acceptance, they don't have production settings.

andythenorth:
Does this cover the case where grfs defining cargo have to place them in specific slots (IDs), otherwise house acceptance will be 'strange'?

https://github.com/andythenorth/firs/blob/main/src/economies/basic_arctic.py#L6

PeterN:
Yes, this is pretty much all it does right now.

PeterN:
This would 'break' cargo NewGRFs that rely on the current behaviour (Improved Town Industries is one I've been told about) but actually their behaviour is already broken for aircraft -- and it's possible to be fixed by the NewGRF implementing the town production callback.

PeterN:
Superseded by #11719

PR #10727, opened 2023-04-27
Codechange: introduce C++ string accepting case insensitive string comparisons

rubidium42:
## Motivation / Problem

https://github.com/OpenTTD/OpenTTD/pull/10667#discussion_r1170586829

## Description

Introduce `StrEqualsIgnoreCase`, `StrCompareIgnoreCase`, `StrStartsWithIgnoreCase` and `StrEndsWithIgnoreCase` that accept `std::string_view`, so it can be used with both C-style strings and `std::string`.
Replace instances of `strcasecmp` and `strncasecmp` with one of the new functions.
Add unit tests, due to the peculiarities that can occur when working with `std:string_view` and passing that to C-style functions.

## Limitations

Maybe Haiku, Cygwin or OS/2 failing to build?

## Checklist for review

### DIFF
--- a/src/string.cpp
+++ b/src/string.cpp
@@ -368,6 +368,59 @@ bool StrEndsWith(const std::string_view str, const std::string_view suffix)
 	return str.compare(str.size() - suffix_len, suffix_len, suffix, 0, suffix_len) == 0;
 }
 
+/** Case insensitive implementation of the standard character type traits. */
+struct CaseInsensitiveCharTraits : public std::char_traits<char> {
+	static bool eq(char c1, char c2) { return toupper(c1) == toupper(c2); }
+	static bool ne(char c1, char c2) { return toupper(c1) != toupper(c2); }
+	static bool lt(char c1, char c2) { return toupper(c1) <  toupper(c2); }
+
+	static int compare(const char *s1, const char *s2, size_t n)
+	{
+		while (n-- != 0) {
+			if (toupper(*s1) < toupper(*s2)) return -1;
+			if (toupper(*s1) > toupper(*s2)) return 1;
+			++s1; ++s2;
+		}
+		return 0;
+	}
+
+	static const char *find(const char *s, int n, char a)
+	{
+		while (n-- > 0 && toupper(*s) != toupper(a)) {
+			++s;
+		}
+		return s;
+	}
+};
+
+/** Case insensitive string view. */
+typedef std::basic_string_view<char, CaseInsensitiveCharTraits> CaseInsensitiveStringView;
+
+/**
+ * Compares two string( view)s, while ignoring the case of the characters.
+ * @param str1 The first string.
+ * @param str2 The second string.
+ * @return Less than zero if str1 < str2, zero if str1 == str2, greater than
+ *         zero if str1 > str2. All ignoring the case of the characters.
+ */
+int StrCompareIgnoreCase(const std::string_view str1, const std::string_view str2)
+{
+	CaseInsensitiveStringView ci_str1{ str1.data(), str1.size() };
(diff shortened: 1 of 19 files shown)


src/string.cpp
```diff
@@ -368,6 +368,35 @@ bool StrEndsWith(const std::string_view str, const std::string_view suffix)
...
+ * @param str2 The second string.
+ * @return Less than zero if str1 < str2, zero if str1 == str2, greater than
+ *         zero if str1 > str2. All ignoring the case of the characters.
+ */
+int StrCompareIgnoreCase(const std::string_view str1, const std::string_view str2)
+{
+	size_t len = std::min(str1.size(), str2.size());
+	int res = strncasecmp(str1.data(), str2.data(), len);
+	if (res != 0) return res;
+
```
PeterN: If `strncasecmp` returns 0, aren't they considered equal?

Oh, it's comparing part of the string, for whatever reason.

rubidium42: Due to `std::string_view` you're not guaranteed a '\0', so we must limit it manually to the shortest and compare. When they are equal, then the difference in size will determine the result.

Having said that str(n)casecmp is essentially broken for comparison as it compares ASCII characters when we feed it UTF-8. To properly solve that, we have to go the way of strnatcmp, which uses ICU/OSX/Windows specific APIs for sorting accounting for all these things.

But as stop-gap I can implement a pure C++ variant of strcasecmp.

src/string.cpp
```diff
@@ -368,6 +381,70 @@ bool StrEndsWith(const std::string_view str, const std::string_view suffix)
...
+ * @return True iff the end of the string is the same as the suffix, ignoring case.
+ */
+bool StrEndsWithIgnoreCase(std::string_view str, const std::string_view suffix)
+{
+	if (str.size() < suffix.size()) return false;
+	return StrEqualsIgnoreCase(str.substr(str.size() - suffix.size()), suffix);
+}
+
+/** Case insensitive implementation of the standard character type traits. */
+struct ci_char_traits : public std::char_traits<char> {
```
LordAro: why not just write it out? 
```suggestion
struct case_insensitive_char_traits : public std::char_traits<char> {
```

src/string.cpp
```diff
@@ -368,6 +381,70 @@ bool StrEndsWith(const std::string_view str, const std::string_view suffix)
...
+	{
+		while( n-- != 0 ) {
+			if (toupper(*s1) < toupper(*s2)) return -1;
+			if (toupper(*s1) > toupper(*s2)) return 1;
+			++s1; ++s2;
+		}
+		return 0;
+	}
+
+	static const char *find(const char *s, int n, char a) {
```
LordAro:
```suggestion
	static const char *find(const char *s, int n, char a)
	{
```

src/string.cpp
```diff
@@ -368,6 +381,70 @@ bool StrEndsWith(const std::string_view str, const std::string_view suffix)
...
+
+/** Case insensitive implementation of the standard character type traits. */
+struct ci_char_traits : public std::char_traits<char> {
+	static bool eq(char c1, char c2) { return toupper(c1) == toupper(c2); }
+	static bool ne(char c1, char c2) { return toupper(c1) != toupper(c2); }
+	static bool lt(char c1, char c2) { return toupper(c1) <  toupper(c2); }
+
+	static int compare(const char *s1, const char *s2, size_t n)
+	{
+		while( n-- != 0 ) {
```
LordAro:
```suggestion
		while(n-- != 0) {
```

LordAro (commented):
More or less fine by me

LordAro (approved):
Sure lol

PR #10733, opened 2023-04-28
Codechange: use standard int types

rubidium42:
## Motivation / Problem

Using custom typedef for types that are standard types.
Mostly `(u)intXX` that can become `(u)intXX_t`, but also `WChar` becoming `char32_t`.

## Description

Automatic replacements; script in commit message.
Manual removal of the `typedef`s from stdafx.h.

## Limitations

Breaks quite some patches.

## Checklist for review


TrueBrain:
I think this would help for new developers to join in, as this makes this project more like other C++ projects. Similar to the efforts of converting to `std::string` etc; it just makes things easier for new joiners.

But, it is a total patch-killer, and as such, I think it is good to involve patchpacks; either by supplying seds or something, so they can go through a similar conversion.

rubidium42:
> I think this would help for new developers to join in, as this makes this project more like other C++ projects. Similar to the efforts of converting to `std::string` etc; it just makes things easier for new joiners.
> 
> But, it is a total patch-killer, and as such, I think it is good to involve patchpacks; either by supplying seds or something, so they can go through a similar conversion.

The probably should wait to just before branching 14, or at least just before the first beta. Regarding the sed, that's in the second paragraph of the automatic change commit of this PR.
And yes, the commits need to get a better description before this gets merged. But alas, it's more proof-of-work and something to get the discussion going than it already being polished to get to be merged.

TrueBrain:
> The probably should wait to just before branching 14, or at least just before the first beta. Regarding the sed, that's in the second paragraph of the automatic change commit of this PR. And yes, the commits need to get a better description before this gets merged. But alas, it's more proof-of-work and something to get the discussion going than it already being polished to get to be merged.

Well, what I more meant is: we should actual talk to patchpack owners; that kind of "involving". If for them this means it is a HUGE hassle to keep up-to-date with upstream, we shouldn't do this.

PeterN:
I would like to see this if possible -- I had done global search & replaces to see what it looks like :-)

However we can at least start to use the correct types piecemeal for future changes?

TrueBrain (approved):
Everyone I talked to signed off on this, as in:

@JGRennison  indicated it doesn't bother him, and he will deal with it.
@PeterN, myself and OP would like to see this change.
Nobody else spoke up against it.

In general I think this is a good improvement, as everyone these days assume `uint32_t`. And as these days it works on all OSes, why not convert the rest to it so we can no longer use `uint32` at all. This also makes things just slightly easier for new contributors, as it complies with modern expectations of a codebase.

And anyone else that runs a patchpack, there is enough information in this PR to do this on your own code too, saving you a lot of trouble. But can't deny, this will hurt a bit.

But so yeah, let's do this!

(you most likely have to check if this needs a rebase; just wanted to give the approval to make a point :D )

PR #10734, opened 2023-04-28
Feature: [NewGRF] Allow higher max speeds for ships

Kuhnovic:
## Motivation / Problem

I recently made this fix: https://github.com/OpenTTD/OpenTTD/pull/10695 . In the original bug report was a link to a forum thread about the Ekranoplans https://www.tt-forums.net/viewtopic.php?t=39188 . These are Soviet era ground effect ships that can go up to 500 km/h, but the max speed of ships in openttd is 128 km/h. This PR changes that.

## Description

I added a new uint16 value that allows higher ship speeds to be set. I also added a byte to set the acceleration, because with the default acceleration fast ships take a very long time to accelerate, which looks a little odd.

## Limitations

More a limitation of the dev: this is the first time I've worked on anything NewGRF-related. Please let me know if I forgot something.

I've never created a NewGRF and obviously there are no NewGRFs available that contain these new values. For testing I just hacked the code in newgrf.cpp:1596 to multiply the ship speed by e.g. 4, and set the acceleration value to a high value. It would be great if somebody with a bit of newgrf-making experience could verify if it works.

**NOTE TO SELF: https://newgrf-specs.tt-wiki.net/wiki/Action0/Vehicles/Ships update this page when this PR is merged.**

## Checklist for review

### DIFF
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -3273,6 +3273,13 @@ bool AfterLoadGame()
 		}
 	}
 
+	if (IsSavegameVersionBefore(SLV_SHIP_ACCELERATION)) {
+		/* NewGRF acceleration information was added to ships. */
+		for (Ship *s : Ship::Iterate()) {
+			if (s->acceleration == 0) s->acceleration = ShipVehInfo(s->engine_type)->acceleration;
+		}
+	}
+
 	for (Company *c : Company::Iterate()) {
 		UpdateCompanyLiveries(c);
 	}
--- a/src/ship_cmd.cpp
+++ b/src/ship_cmd.cpp
@@ -415,8 +415,7 @@ static bool CheckShipLeaveDepot(Ship *v)
 static uint ShipAccelerate(Vehicle *v)
 {
 	uint speed;
-
-	speed = std::min<uint>(v->cur_speed + 1, v->vcache.cached_max_speed);
+	speed = std::min<uint>(v->cur_speed + v->acceleration, v->vcache.cached_max_speed);
 	speed = std::min<uint>(speed, v->current_order.GetMaxSpeed() * 2);
 
 	/* updates statusbar only if speed have changed to save CPU time */
@@ -711,6 +710,8 @@ static void ShipController(Ship *v)
 
 	const uint number_of_steps = ShipAccelerate(v);
 	for (uint i = 0; i < number_of_steps; ++i) {
+		if (ShipMoveUpDownOnLock(v)) return;
+
 		GetNewVehiclePosResult gp = GetNewVehiclePos(v);
 		if (v->state != TRACK_BIT_WORMHOLE) {
 			/* Not on a bridge */
@@ -915,6 +916,7 @@ CommandCost CmdBuildShip(DoCommandFlag flags, TileIndex tile, const Engine *e, V
 		v->sprite_cache.sprite_seq.Set(SPR_IMG_QUERY);
 		v->random_bits = Random();
 
+		v->acceleration = svi->acceleration;
 		v->UpdateCache();
 
 		if (e->flags & ENGINE_EXCLUSIVE_PREVIEW) SetBit(v->vehicle_flags, VF_BUILT_AS_PROTOTYPE);
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -87,6 +87,8 @@ Engine::Engine(VehicleType type, EngineID base)
 		if (type == VEH_ROAD) this->u.road.tractive_effort = 0x4C;
 		/* Aircraft must have INVALID_CARGO as default, as there is no property */
 		if (type == VEH_AIRCRAFT) this->info.cargo_type = INVALID_CARGO;
+		/* Ships must have a non-zero acceleration. */
+		if (type == VEH_SHIP) this->u.ship.acceleration = 1;
 		/* Set visual effect to the default value */
 		switch (type) {
 			case VEH_TRAIN: this->u.rail.visual_effect = VE_DEFAULT; break;
--- a/src/engine_type.h
+++ b/src/engine_type.h
@@ -67,6 +67,7 @@ struct RailVehicleInfo {
 struct ShipVehicleInfo {
 	byte image_index;
 	byte cost_factor;
+	uint8_t acceleration;    ///< Acceleration (1 unit = 1/3.2 mph per tick = 0.5 km-ish/h per tick)
 	uint16_t max_speed;      ///< Maximum speed (1 unit = 1/3.2 mph = 0.5 km-ish/h)
 	uint16_t capacity;
 	byte running_cost;
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1586,7 +1586,7 @@ static ChangeInfoResult ShipVehicleChangeInfo(uint engine, int numinfo, int prop
 				svi->cost_factor = buf->ReadByte();
 				break;
 
-			case PROP_SHIP_SPEED: // 0x0B Speed (1 unit is 0.5 km-ish/h)
+			case PROP_SHIP_SPEED: // 0x0B Speed (1 unit is 0.5 km-ish/h). Use 0x23 to achieve higher speeds.
 				svi->max_speed = buf->ReadByte();
 				break;
 
@@ -1712,6 +1712,14 @@ static ChangeInfoResult ShipVehicleChangeInfo(uint engine, int numinfo, int prop
 
 			case 0x22: // Callback additional mask
 				SB(ei->callback_mask, 8, 8, buf->ReadByte());
+				break;
+
+			case 0x23: // Speed (1 unit is 0.5 km-ish/h)
+				svi->max_speed = buf->ReadWord();
+				break;
+
+			case 0x24: // Acceleration (1 unit is 0.5 km-ish/h per tick)
+				svi->acceleration = std::max<uint8_t>(1, buf->ReadByte());
 				break;
 
 			default:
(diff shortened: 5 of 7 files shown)


PeterN:
Not sure that how this was done for Road Vehicles is a good example... nor do I know why it was done that way.

Just adding a uint16 to set max_speed seems to make more sense to me.

rubidium42:
> Not sure that how this was done for Road Vehicles is a good example... nor do I know why it was done that way.
> 
> Just adding a uint16 to set max_speed seems to make more sense to me.

I agree that setting the maximum speed as a new uint16 property, just like the long format introduction (1A vs 00 for ships), would be better for the future. Similarly, why not expose the acceleration of the vehicle to NewGRFs as well? Aircraft has such a thing, though maybe use a more sensible unit?

Kuhnovic:
@PeterN @rubidium42 The property is now an uint16, and the acceleration is also added as a property. The acceleration is 0.5 km/h per tick, which is still not a really sensible unit. These numbers are basically dictated by the integer math that control ships, not much I can do about it without more high-impact changes.

src/ship_cmd.cpp
```diff
@@ -704,6 +703,8 @@ static void ShipController(Ship *v)
 
 	const uint number_of_steps = ShipAccelerate(v);
 	for (uint i = 0; i < number_of_steps; ++i) {
+		if (ShipMoveUpDownOnLock(v)) return;
```
Kuhnovic, line 713: A ship is only considered inside a lock if it's precisely on a specific pixel within the tile. Without this extra check fast ships might skip over that pixel and the ship will just pass through the lock without waiting (and without changing z height).

rubidium42: I would almost consider adding a speed limit for a lock. Going through a lock at > 100 km/h seems a bit unsafe.
Having said that, do ships also bump into docks at maximum speed? Oh... darn it, rabbit hole.

Kuhnovic: Their speed within the lock is just the up/down travel speed of the dock. This is independent of the ship's speed, which makes sense: a lock just let's water seep in or out at a certain rate.

But yes, a ship can come racing towards a lock at 1000 km/h and then grinds to a halt in the blink of an eye. Same thing with ships arriving at docks. This is not a problem for slow ships but it looks a little funny for faster ships.

PeterN: Speed limits for locks (and docks in also) is a nice idea, but out of scope of this change I think.

(And might be better served as not a property of boats themselves)

src/newgrf_properties.h
```diff
@@ -41,10 +41,10 @@ enum PropertyID {
 	PROP_ROADVEH_SHORTEN_FACTOR                 = 0x23, ///< Shorter vehicles
 
 	PROP_SHIP_COST_FACTOR                       = 0x0A, ///< Purchase cost
-	PROP_SHIP_SPEED                             = 0x0B, ///< Max. speed: 1 unit = 1/3.2 mph = 0.5 km-ish/h
 	PROP_SHIP_CARGO_CAPACITY                    = 0x0D, ///< Capacity
 	PROP_SHIP_RUNNING_COST_FACTOR               = 0x0F, ///< Yearly runningcost
 	PROP_SHIP_CARGO_AGE_PERIOD                  = 0x1D, ///< Number of ticks before carried cargo is aged
+	PROP_SHIP_SPEED                             = 0x23, ///< Max. speed: 1 unit = 1/3.2 mph = 0.5 km-ish/h
```
rubidium42: I'm no NewGRF expert, but I fear this change breaks NewGRFs that use `PROP_SHIP_SPEED` as in the NewGRFs that would simply be encoded as `(0x)0B`. In other words, I think the value of the constant may not be changed.
Having said that, I do not think it is a problem to leave it at the old value as the callback will return a 14 or 15 bit value from the NewGRF. That even makes me wonder whether the issue there was with fast ships might have been due to a ship defining 256 units via this callback.

Kuhnovic: Is this enum made available outside of the OpenTTD code? If so then indeed this could lead to problems with existing NewGRFs. I'm no NewGRF expert either, I hope someone with a bit more domain knowledge kan weigh in here. The main reason I changed it was that I needed it elsewhere in the code instead of the "legacy" speed, and I followed the documentation of the PropertyID enum which says: `Currently the list only contains properties which are used more than once in the code. I.e. they are available for callback 0x36.`

Are you saying that we could also just change the byte to a uint16 or increase the max allowed speed? I can image that that would give issues with existing NewGRFs, but again, no expert... it might work?

michicc: `available for callback 0x36` means that NewGRF expect that callback to be run for prop 0B to get the speed (and GRF is all raw numbers, there are no symbolic constants), not prop 23.

Now while there might be existing NewGRFs that wrongly try to return too large nonsense values in CB 36, it might be safe to just use the full callback result. In fact, I think there are things already that just say "use CB 36 if you need a value bigger than prop X allows".

2TallTyler:
I'd be interested in seeing this in 14.0 alongside the new ship pathfinder, if we have a qualified reviewer (which I am not) to sign off on it.

rubidium42:
I've resolved the merge conflict and reverted the change in the value of `PROP_SHIP_SPEED` as that value is essentially defined via the NewGRF callback specification. I haven't tested it though.

2TallTyler:
@zephyris Would you be interested in dusting off your old Ekranoplan GRF to test this? (or making the source available for @Kuhnovic to test 🙂)

I've added this to the 14.0 milestone so hopefully it doesn't slip through the cracks, but somebody has to test it and say if it works and makes sense. 😉

zephyris:
Yeah, of course. I'll have to do some digging though. The code may have been lost to an openttdcoop hole in the cloud, but I should have all original sprites and reverse engineering the code should be easy enough.

I've just moved house, so no Internet right now, feel free to remind me in a week or so if no progress.

zephyris:
Yeah, of course. I'll have to do some digging though. The code may have been lost to an openttdcoop hole in the cloud, but I should have all original sprites and reverse engineering the code should be easy enough.

I've just moved house, so no Internet right now, feel free to remind me in a week or so if no progress.

zephyris:
Ok, I've got my ekranoplan grf essentially revived. I haven't got a nice `nfo` source as it's essentially just a demo grf with no reasonable stats, so if you want an nfo file to work from then it's best to just decompile the grf file here: https://www.tt-forums.net/viewtopic.php?p=723787#p723787

I've also grabbed the sprites and set them up for a nicer `nml` source. Currently skeletal vehicle stats, but nicely templated and commented. I'll polish this and put it up on my openttd-newgrfs repo at some point - but let me know if you'd like a copy to work on in the meantime.

zephyris:
Another day, another bump. I've finished NML recode of the ekranoplan newGRF, available here: https://github.com/zephyris/openttd-newgrfs/tree/main/ekranoplan

Currently, their speeds are all 127 km/h, but there are comments with the correct speed for each vehicle.

zephyris:
[ekranoplan.zip](https://github.com/OpenTTD/OpenTTD/files/14067181/ekranoplan.zip)
This is a test grf, coding a late-game ekranoplan (intro year 2029). There's two versions, one coded with `0B FE` (127 km/h) for speed and one with `23 14 05` (650 km/h).

rubidium42:
> This is a test grf, coding a late-game ekranoplan (intro year 2029). There's two versions, one coded with `0B FE` (127 km/h) for speed and one with `23 14 05` (650 km/h).

ekro

The NewGRF works! Only caveat is that it needs almost 256 tiles to get up to speed with the slow acceleration it got now (not configured). Ship 1 is the "slow" one, Ship 2 is the "fast" one.

zephyris:
Awesome. This is a test newGRF with three variants, one now with `23 14 05` and `24 FE` for high speed and super-high acceleration. [ekranoplan2.zip](https://github.com/OpenTTD/OpenTTD/files/14072485/ekranoplan2.zip)

Assuming this works as expected, this looks decent from a newGRF author perspective. Just one query: What's the default acceleration? ie. default value for `0x24`. Is it `1`? That doesn't allow _slower_ acceleration than default ships, not that I think that slower acceleration would realistically ever be needed.

rubidium42:
The super-high acceleration is really noticeable.
Roughly speaking for a ~250 long route, one leg of the "slow" one equals 3 legs of the fast one with slow acceleration and equals 5 legs of the fast one with fast acceleration.

For slower acceleration we would need to get fractional speeds for ships. That's going to be a huge hassle, so I do not think making acceleration any slower would be a good idea.

src/saveload/afterload.cpp
```diff
@@ -3273,6 +3273,13 @@ bool AfterLoadGame()
 		}
 	}
 
+	if (IsSavegameVersionBeforeOrAt(SLV_CALENDAR_SUB_DATE_FRACT)) {
```
TrueBrain: Please create a new `SLV_` for this PR; no need to piggyback on another version.

Kuhnovic:
The default acceleration is indeed 1 unit per tick, in some kind of bizarre unit that I can't remember. IMO we don't need any slower than that, but we do need higher acceleration for fast ships to look a bit more realistic. 

@zephyris you might have overdone the acceleration in your newgrf a bit though :P . It is a lot of fun to play around with!

zephyris:
That was deliberate! I wanted to check that a single byte for acceleration was appropriate. Back of the envelope calculation says yes, but seeing it in-game is the real test.

Just need some nml support now, so I don't have to hack through decompiled nfo to get the high speeds set :)

2TallTyler (approved):
Code looks good to me, and Ekranoplan seems like an excellent stress-test to prove that it works for NewGRF authors. 😃

PR #10739, opened 2023-04-29
Codechange: remove need for stredup/free from GRFConfig/GRFFile

rubidium42:
## Motivation / Problem

Some manual filename/error memory management in GRFConfig and GRFFile, and especially the long term vision of replacing C-style constructs with C++ style constructs.

## Description

Replace GRFConfig's error pointer with a `std::unique_ptr`, which contrary to what some comments imply is not copied.
Replace GRFConfig and GRFFile's filename `char *` with `std::string`, and replace `stredup` with assigments. Furthermore replace some `strcmp` with `==` from `std::string`.

Replace the `seprintf` style MD5 checksum printing with `std::string` and using `FormatArrayAsHex`. This is primarily done to ease the migration to C++ style string formatting, which was needed to prevent some additions of `.c_str()` in the PR.

## Limitations

None that I am aware of.
Well, maybe the MD5 checksum should be it's own PR, but then I'm stacking PRs on top of eachother again making my life harder, and all in all this isn't that big I think.

## Checklist for review

### DIFF
--- a/src/gamelog.cpp
+++ b/src/gamelog.cpp
@@ -114,36 +114,34 @@ void Gamelog::Reset()
 }
 
 /**
- * Prints GRF ID, checksum and filename if found
- * @param buf The location in the buffer to draw
+ * Adds the GRF ID, checksum and filename if found to the output iterator
+ * @param outputIterator The iterator to add the GRF info to.
  * @param last The end of the buffer
  * @param grfid GRF ID
  * @param md5sum array of md5sum to print, if known
  * @param gc GrfConfig, if known
- * @return The buffer location.
  */
-static char *PrintGrfInfo(char *buf, const char *last, uint grfid, const uint8 *md5sum, const GRFConfig *gc)
+template <typename T>
+static void AddGrfInfo(T &outputIterator, uint grfid, const uint8 *md5sum, const GRFConfig *gc)
 {
-	char txt[40];
-
 	if (md5sum != nullptr) {
+		char txt[40];
 		md5sumToString(txt, lastof(txt), md5sum);
-		buf += seprintf(buf, last, "GRF ID %08X, checksum %s", BSWAP32(grfid), txt);
+		fmt::format_to(outputIterator, "GRF ID {:08X}, checksum {}", BSWAP32(grfid), txt);
 	} else {
-		buf += seprintf(buf, last, "GRF ID %08X", BSWAP32(grfid));
+		fmt::format_to(outputIterator, "GRF ID {:08X}", BSWAP32(grfid));
 	}
 
 	if (gc != nullptr) {
-		buf += seprintf(buf, last, ", filename: %s (md5sum matches)", gc->filename);
+		fmt::format_to(outputIterator, ", filename: {} (md5sum matches)", gc->filename);
 	} else {
 		gc = FindGRFConfig(grfid, FGCM_ANY);
 		if (gc != nullptr) {
-			buf += seprintf(buf, last, ", filename: %s (matches GRFID only)", gc->filename);
+			fmt::format_to(outputIterator, ", filename: {} (matches GRFID only)", gc->filename);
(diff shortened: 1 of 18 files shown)


src/gamelog.cpp
```diff
@@ -134,11 +134,8 @@ void GamelogReset()
  */
 static char *PrintGrfInfo(char *buf, const char *last, uint grfid, const uint8 *md5sum, const GRFConfig *gc)
 {
-	char txt[40];
-
 	if (md5sum != nullptr) {
-		md5sumToString(txt, lastof(txt), md5sum);
-		buf += seprintf(buf, last, "GRF ID %08X, checksum %s", BSWAP32(grfid), txt);
+		buf += seprintf(buf, last, "GRF ID %08X, checksum %s", BSWAP32(grfid), Md5sumToString(md5sum).c_str());
```
TrueBrain: Maybe a stupid question, but does the result of `Md5sumToString` live long enough for `seprintf` to be safe to use the `c_str`?

I never know in these statements when the `std::string` is being deleted again, and if that is well defined in C++ to start with.

src/string_func.h
```diff
@@ -85,7 +85,7 @@ static inline size_t ttd_strnlen(const char *str, size_t maxlen)
 	return t - str;
 }
 
-char *md5sumToString(char *buf, const char *last, const uint8 md5sum[16]);
+std::string Md5sumToString(const uint8 md5sum[16]);
```
LordAro:
```suggestion
std::string MD5sumToString(const uint8 md5sum[16]);
```
?

lower case 'd' looks odd

src/gamelog.cpp
rubidium42: It should be fine, but I've got a better solution... rewrite gamelog's print to use `fmt::format_to` instead, so `c_str()` isn't needed anymore.

src/newgrf.cpp
```diff
@@ -7150,9 +7148,10 @@ static void GRFLoadError(ByteReader *buf)
 	}
 
 	/* For now we can only show one message per newgrf file. */
-	if (_cur.grfconfig->error != nullptr) return;
+	if (_cur.grfconfig->error.get() != nullptr) return;
```
michicc:
```suggestion
	if (_cur.grfconfig->error) return;
```
`operator bool` of a `unique_ptr` checks if it has a value assigned.

src/newgrf_config.cpp
```diff
@@ -63,8 +62,7 @@ GRFConfig::GRFConfig(const GRFConfig &config) :
 {
 	MemCpyT<uint8>(this->original_md5sum, config.original_md5sum, lengthof(this->original_md5sum));
 	MemCpyT<uint32>(this->param, config.param, lengthof(this->param));
-	if (config.filename != nullptr) this->filename = stredup(config.filename);
-	if (config.error != nullptr) this->error = new GRFError(*config.error);
+	if (config.error.get() != nullptr) this->error = std::make_unique<GRFError>(*config.error);
```
michicc: See earlier comment.

src/newgrf.cpp
PeterN: We've traditionally deliberately avoided nullptr-is-false tests, I know it's not exactly the same but the appearance is similar. Do we want to revise that?

In CODINGSTYLE.md: `* Don't treat non-flags as flags: use "if (char_pointer != nullptr && *char_pointer != '\0')", not "if (char_pointer && *char_pointer)".`

Pretty sure you can also do `if (_cur.grfconfig->error != nullptr) return;` without the get() to stick with current coding style.

rubidium42: I wasn't aware that they special-cases the `==` and `!=` operators to allow `nullptr`. Changed it to just compare to `nullptr`.

PeterN (approved):

PR #10743, opened 2023-04-29
Codechange: Pass more std::string to StringFilter::AddLine()

2TallTyler:
## Motivation / Problem

In https://github.com/OpenTTD/OpenTTD/pull/10519#discussion_r1181112446 @rubidium42 suggested changing some `.c_str()` conversions to just passing std::strings, now that we have a suitable `StringFilter::AddLine()` which accepts them.

## Description

Convert all the easy C-strings to std::string.

Does not depend on #10519.

## Limitations

I'm not sure the two conversions from `inline const char *GetCachedName() const` to `inline const std::string &GetCachedName() const` is correct, because my understanding of `*` vs `&` is...limited. 🙂 

## Checklist for review

### DIFF
--- a/src/town.h
+++ b/src/town.h
@@ -125,11 +125,11 @@ struct Town : TownPool::PoolItem<&_town_pool> {
 
 	void UpdateVirtCoord();
 
-	inline const char *GetCachedName() const
+	inline const std::string &GetCachedName() const
 	{
-		if (!this->name.empty()) return this->name.c_str();
+		if (!this->name.empty()) return this->name;
 		if (this->cached_name.empty()) this->FillCachedName();
-		return this->cached_name.c_str();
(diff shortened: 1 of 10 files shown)


rubidium42 (dismissed):

src/town.h
```diff
@@ -124,11 +124,11 @@ struct Town : TownPool::PoolItem<&_town_pool> {
 
 	void UpdateVirtCoord();
 
-	inline const char *GetCachedName() const
+	inline const std::string &GetCachedName() const
 	{
 		if (!this->name.empty()) return this->name.c_str();
```
glx22: works but `c_str()` is useless now :)

rubidium42 (approved):

PR #10745, opened 2023-04-30
Codechange: use more C++ constructs in strgen

rubidium42:
## Motivation / Problem

Long term goal to go towards more C++-ish code, though primarily getting rid of stredup/calloc.

## Description

Use `std::filesystem::path` and friends over `stredup` C-strings.
Use `std::string` over `stredup` C-strings.
Use `std::vector` over `CallocT`-ed arrays and linked lists.
Use `std::array` where applicable.
Use streams over file handled for file access.
Replace custom hashing lookup with `std::unordered_map`.
Fix a memory leak (CmdPair.v would be `stredup`-ed, but never `free`d).

## Limitations

Requires a newer MacOS deployment target due to amongst others `std::filesystem::path`. 

## Checklist for review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ if (EMSCRIPTEN)
 endif()
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.13)
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
 
 # Use GNUInstallDirs to allow customisation
 # but set our own default data and bin dir
--- a/src/strgen/strgen.cpp
+++ b/src/strgen/strgen.cpp
@@ -15,18 +15,12 @@
 #include "../strings_type.h"
 #include "../misc/getoptdata.h"
 #include "../table/control_codes.h"
+#include "../3rdparty/fmt/std.h"
 
 #include "strgen.h"
 
-
-#if !defined(_WIN32) || defined(__CYGWIN__)
-#include <unistd.h>
-#include <sys/stat.h>
-#endif
-
-#if defined(_WIN32) || defined(__WATCOMC__)
-#include <direct.h>
-#endif /* _WIN32 || __WATCOMC__ */
+#include <filesystem>
+#include <fstream>
 
 #include "../table/strgen_tables.h"
 
@@ -75,7 +69,7 @@ void NORETURN FatalErrorI(const std::string &msg)
 
 /** A reader that simply reads using fopen. */
 struct FileStringReader : StringReader {
-	FILE *fh; ///< The file we are reading.
+	std::ifstream input_stream;
 
 	/**
 	 * Create the reader.
@@ -84,22 +78,15 @@ struct FileStringReader : StringReader {
 	 * @param master      Are we reading the master file?
 	 * @param translation Are we reading a translation?
 	 */
-	FileStringReader(StringData &data, const char *file, bool master, bool translation) :
-			StringReader(data, file, master, translation)
+	FileStringReader(StringData &data, const std::filesystem::path &file, bool master, bool translation) :
+			StringReader(data, file.generic_string().c_str(), master, translation)
 	{
-		this->fh = fopen(file, "rb");
-		if (this->fh == nullptr) FatalError("Could not open {}", file);
-	}
-
-	/** Free/close the file. */
-	virtual ~FileStringReader()
-	{
-		fclose(this->fh);
+		this->input_stream.open(file, std::ifstream::binary);
 	}
 
 	char *ReadLine(char *buffer, const char *last) override
 	{
-		return fgets(buffer, ClampTo<uint16_t>(last - buffer + 1), this->fh);
+		return this->input_stream.getline(buffer, last - buffer) ? buffer : nullptr;
 	}
 
 	void HandlePragma(char *str) override;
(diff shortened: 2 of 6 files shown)


src/strgen/strgen.cpp
```diff
@@ -485,78 +396,74 @@ int CDECL main(int argc, char *argv[])
...
 			master_reader.ParseFile();
 
 			for (int i = 0; i < mgo.numleft; i++) {
 				data.FreeTranslation();
 
-				const char *translation = replace_pathsep(mgo.argv[i]);
-				const char *file = strrchr(translation, PATHSEPCHAR);
-				FileStringReader translation_reader(data, translation, false, file == nullptr || strcmp(file + 1, "english.txt") != 0);
+				std::filesystem::path lang_file = mgo.argv[i];
+				FileStringReader translation_reader(data, lang_file, false, lang_file.filename() != "english.txt");
```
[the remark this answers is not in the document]
rubidium42: The whole idea of strgen is that we can inject paths as command-line arguments.

src/strgen/strgen.cpp
```diff
@@ -485,78 +396,74 @@
...
```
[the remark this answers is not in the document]
rubidium42: The whole idea of strgen is that we can inject paths as command-line arguments. As such, I have dismissed the alert.

TrueBrain:
@michicc if I remember correctly, you found some juicy links to issues with `std::filesystem` and MacOS, and why it was better to not touch that yet with a long stick. I just can't find those URLs anymore, and I have no clue if they apply here. You happen to remember? :)

michicc:
I don't think there are any "juice" issues today. `std::filesystem` is strictly 10.15+ though and generally the least supported part of C++17 it feels like. E.g. you need to link an extra lib for some (older) gcc and clang versions, but not for current ones anymore, and there's no automatic for this.

rubidium42:
I'm not quite sure about the `<< fmt::format(...` code. There could be more elegant code where `fmt::print` allows writing to `ostreams`, but that requires adding `ostream` support which is in #10850.

src/strgen/strgen.cpp
```diff
@@ -184,103 +171,75 @@ void FileStringReader::HandlePragma(char *str)
...
-	if (f2 == nullptr) return false;
-
-	FILE *f1 = fopen(n1, "rb");
-	if (f1 == nullptr) {
-		fclose(f2);
-		FatalError("can't open {}", n1);
-	}
+	/* Check for equal size, but ignore the error code for cases when a file does not exist. */
+	std::error_code error_code;
+	if (std::filesystem::file_size(path1, error_code) != std::filesystem::file_size(path2, error_code)) return false;
```
LordAro, line 178: what happens if both of them don't exist?

src/strgen/strgen.cpp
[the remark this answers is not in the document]
LordAro: isn't the issue with these that access is unchecked? I realise it's unnecessary for our purposes, but perhaps some basic validity checks would make it go away? https://codeql.github.com/codeql-query-help/cpp/cpp-path-injection/#recommendation

src/strgen/strgen_base.cpp
```diff
@@ -38,50 +38,29 @@ static const CmdStruct *ParseCommandString(const char **str, char *param, int *a
...
-LangString::LangString(const char *name, const char *english, size_t index, int line) :
-		name(stredup(name)), english(stredup(english)), translated(nullptr),
-		hash_next(0), index(index), line(line), translated_case(nullptr)
-{
-}
-
-/** Free everything we allocated. */
-LangString::~LangString()
+LangString::LangString(const std::string &name, const std::string &english, size_t index, int line) :
+		name(name), english(english), index(index), line(line)
```
LordAro, line 55: Do we not normally put the `:` on the newline, rather than before?

```suggestion
LangString::LangString(const std::string &name, const std::string &english, size_t index, int line)
		: name(name), english(english), index(index), line(line)
```

rubidium42:
```
src$ grep -R ':$' .|grep ::|grep -v 3rdparty|grep -v case|wc -l
75
src$ grep -R '       : ' .|grep -v 3rdparty|wc -l
56
```
So... the `:` at the end of the line seems more prevalent than at the begin of the line. But there is a clear lack of consensus about it.

src/strgen/strgen.cpp
rubidium42: Lets first discuss when that happens. The second parameter is the file it just finished writing and just closed. So, you must have removed the file between writing finished and this call. That is a super small window, so... essentially never happening. That's why it was a fatal error.

What happens... well it will open the files and either error out (exception) or just read an empty file and return true. In the latter case both files are the same and it tries to remove the now-removed temporary file. If it were to return false, it would remove the other missing file and then move the temporary file to the missing file.

So, I'd say: perfect is the enemy of good. This works unless you remove the build folder while building, of when memory corruption is triggered in the computer.

src/strgen/strgen.cpp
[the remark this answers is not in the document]
rubidium42: From the suggestions:
- Do not allow more than a single “.” character.
  - Given CMake uses full paths, so checking for this means a builder is not allowed to have a `.` in the build path. I do not think we should add this limitation.
- Do not allow directory separators such as “/” or “\” (depending on the filesystem).
  - Given CMake uses full paths, this breaks building with CMake.
- Do not rely on simply replacing problematic sequences such as “../”. For example, after applying this filter to “…/…//” the resulting string would still be “../”.
  - There is no reliance on removing "../". Doesn't seem very useful to me, as then you can't easily run strgen without providing the full path which would be a nuisance.
- Ideally use a whitelist of known good patterns.
  - Can you provide me with a list of the paths of all the builders? Or are we going to "just" check that it contains "lang" as whitelist; the latter seems pretty pointless.

So, what check can I implement? Just checking for "lang" in the filename seems very pointless and that yields the question... why? And the answer: to silence a code scanner on a warning that we will not comply to and we tried to silence, instead of just dismissing it with a proper comment.

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if (EMSCRIPTEN)
 endif()
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.13)
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
```
LordAro, line 26: What sort of issues will this cause? Will the website need updating?

rubidium42: I can't find any reference to 10.13 on openttd.org, except for OS version of an old bug report and some IRC log discussion about 10.13 being deprecated by Azure.

The only issue this might cause is that people need to upgrade their macOS version, and that we do not support Apple hardware from before the 2012 models anymore when running macOS. Because Apple does not support older hardware with 10.15. And 10.15 support has already been discontinued by Apple.

LordAro (approved):

PR #10746, opened 2023-04-30
Codechange: Don't use macros for DAYS_TILL and friends

2TallTyler:
## Motivation / Problem

We use macros for date lookups. 

I needed to change this for my ongoing work with time, but it's no longer necessary. Might as well clean it up, though.

## Description

- Convert the macros in `date_type.h` into inline functions.

## Limitations

- I have not measured performance changes.
- There might be better ways to do this.

## Checklist for review

### DIFF
--- a/src/date_type.h
+++ b/src/date_type.h
@@ -52,31 +52,21 @@ static const TimerGameCalendar::Year ORIGINAL_END_YEAR  = 2051;
 static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
 
 /**
- * Calculate the number of leap years till a given year.
- *
- * Each passed leap year adds one day to the 'day count'.
- *
- * A special case for the year 0 as no year has been passed,
- * but '(year - 1) / 4' does not yield '-1' to counteract the
- * '+1' at the end of the formula as divisions round to zero.
- *
- * @param year the year to get the leap years till.
- * @return the number of leap years.
- */
-#define LEAP_YEARS_TILL(year) ((year) == 0 ? 0 : ((year) - 1) / 4 - ((year) - 1) / 100 + ((year) - 1) / 400 + 1)
-
-/**
  * Calculate the date of the first day of a given year.
  * @param year the year to get the first day of.
  * @return the date.
  */
-#define DAYS_TILL(year) (DAYS_IN_YEAR * (year) + LEAP_YEARS_TILL(year))
+static inline TimerGameCalendar::Date DateAtStartOfYear(TimerGameCalendar::Year year)
+{
+	uint number_of_leap_years = (year == 0) ? 0 : ((year - 1) / 4 - (year - 1) / 100 + (year - 1) / 400 + 1);
+
+	return (DAYS_IN_YEAR * year) + number_of_leap_years;
+}
 
 /**
- * The offset in days from the 'TimerGameCalendar::date == 0' till
- * 'TimerGameCalendar::ConvertYMDToDate(ORIGINAL_BASE_YEAR, 0, 1)'
+ * The date of the first day of the original base year.
  */
-#define DAYS_TILL_ORIGINAL_BASE_YEAR DAYS_TILL(ORIGINAL_BASE_YEAR)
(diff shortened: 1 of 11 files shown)


src/date.cpp
```diff
@@ -25,22 +25,25 @@
 
 #include "safeguards.h"
 
-#define M(a, b) ((a << 5) | b)
+static uint16 PackMonthDateHelper(uint8 a, uint8 b)
+{
+	return (a << 5) | b;
+}
+
 static const uint16 _month_date_from_year_day[] = {
```
PeterN: This is not really an improvement. It would be better to have a 2 value struct and just use initialization to fill it, skipping stuffing it into a single 16 bit value.

src/date_type.h
```diff
@@ -92,8 +102,11 @@ static const TimerGameCalendar::Year DEF_END_YEAR = ORIGINAL_END_YEAR - 1;
  */
 static const TimerGameCalendar::Year MAX_YEAR  = 5000000;
 
-/** The number of days till the last day */
-#define MAX_DAY (DAYS_TILL(MAX_YEAR + 1) - 1)
+/** Calculate the number of days till the last day, starting at day 0. */
+static inline TimerGameCalendar::Day MaxDay()
```
PeterN: `static const TimerGameCalendar::Day MAX_DAY = DaysTill(MAX_YEAR + 1) - 1;`
may work... may need constexpr. Not sure though.

2TallTyler: Yes, it works. `constexpr` does not.

src/date.cpp
2TallTyler: I'm going to leave this one alone for now...

src/date_type.h
```diff
@@ -63,20 +63,25 @@ static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
...
+	return (year == 0) ? 0 : ((year - 1) / 4 - (year - 1) / 100 + (year - 1) / 400 + 1);
+}
 
 /**
  * Calculate the date of the first day of a given year.
  * @param year the year to get the first day of.
  * @return the date.
  */
-#define DAYS_TILL(year) (DAYS_IN_YEAR * (year) + LEAP_YEARS_TILL(year))
+static inline TimerGameCalendar::Year DaysTill(TimerGameCalendar::Year year)
```
rubidium42: I do not think you want to return years here. I guess it should return `Date`, and maybe rename it to `DateAtStartOfYear` or something similar?

src/date_type.h
```diff
@@ -63,20 +63,25 @@ static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
  * @param year the year to get the leap years till.
  * @return the number of leap years.
  */
-#define LEAP_YEARS_TILL(year) ((year) == 0 ? 0 : ((year) - 1) / 4 - ((year) - 1) / 100 + ((year) - 1) / 400 + 1)
+static inline TimerGameCalendar::Year LeapYearsTill(TimerGameCalendar::Year year)
```
rubidium42: I do not think that counting the number of leap years should return a `Year`. Having said that, maybe just move this functionality into the `DaysTill` function. And have it as `uint number_of_leap_years = ...`.

src/date_type.h
```diff
@@ -92,8 +97,8 @@ static const TimerGameCalendar::Year DEF_END_YEAR = ORIGINAL_END_YEAR - 1;
  */
 static const TimerGameCalendar::Year MAX_YEAR  = 5000000;
 
-/** The number of days till the last day */
-#define MAX_DAY (DAYS_TILL(MAX_YEAR + 1) - 1)
+/** The number of days till the last day, starting at day 0. */
+static const TimerGameCalendar::Day MAX_DAY = DaysTill(MAX_YEAR + 1) - 1;
```
rubidium42: `Day` is `uint8`, `MAX_YEAR` is way bigger, and so is number of days in a year. I suggest renaming to `MAX_DATE`.

2TallTyler: Ah yes, meant this to be Date but confused the two. Problems with submitting PRs after a long day of chasing the bottom of rabbitholes. 😄

rubidium42 (approved):

PR #10747, opened 2023-04-30
Feature: [Linux] RTL support with modern versions of ICU

### LINKED ISSUE #6666: Town name list in Arabic
**_PaulC_ opened the ticket and wrote:**

> Is this a bug with the Arabic/Egypt language? In the town name list, populations are shown with a closing bracket (and space?) before the number.

**Attachments**
- [Arabic.png](https://bugs.openttd.org/task/6666/getfile/10959/Arabic.png) (197.51 KiB)

**Reported version**: _trunk_
**Operating system**: _All_

---

### LINKED ISSUE #6922: Replace ICU ParagraphLayout with something else?
OpenTTD currently uses the ICU ParagraphLayout API for doing
internationalized wordwrapping. This usage is a problematic, since
ICU deprecated and removed their layout API for being buggy and
unmaintained. The ParagraphLayout API that OpenTTD uses is still present
in ICU, but it relies on the removed Layout API. This gap can be be
filled using the icu-le-hb library, which implements the layout API on
top of the Harfbuzz library.

However, it seems icu-le-hb is more of a proof of concept and not really
maintained either. Additionally, because ParagraphLayout is in ICU and
icu-le-hb depends on ICU, this gives an undesirable circular dependency,
which is problematic for at least Debian, since it complicates updating
to new ICU versions,.

See https://bugs.debian.org/894159 for more info on this, in particular
https://bugs.debian.org/894159#45 and onwards. The Debian version of
this issue is https://bugs.debian.org/897233

It seems it would be beneficial to move away from the ICU
ParagraphLayout API, so it can be removed from Debian (OpenTTD is the
only user in Debian), but also because it has indeed proven to be buggy
in the past (OpenTTD talks about ICU-related crashes, though this was
before the icu-le-hb implementation).

In the above linked bug, the following was written:

> > Another solution is of course to disable ParagraphLayout. László also
> > asked if OpenTTD, being the only user of this API, could migrate to
> > another solution. I've discussed this with OpenTTD upstream yesterday,
> > and they were already aware of the layout API removal and have been
> > casually looking at Harfbuzz and Pango as a replacement, but they do not
> > see an easy solution yet. ParagraphLayout seems to fit their usecase
> > quite neatly: they need internationalized word-wrapping of text (e.g.
> > also supporting right-to-left locales). Harfbuzz does not seem to offer
> > that, and Pango seems heavy-handed (and might not be easy to adapt to
> > OpenTTD's SDL renderer, and might not be portable enough).
> This is bad to read. I had the hope there's an easy solution and/or the
> replacement might be already in the making.
>
> > Neither me or upstream has much experience in this field, perhaps you
> > have a different suggestion for an alternative?
> I don't know any other alternative. Only OpenTTD uses the Paragraph Layout
> API and it makes me wonder what other solution the other projects use? I
> may think other games like Lincity-NG[3] also need internationalized text
> placement and/or LibreOffice still need to handle this as well. Do these
> have an alternative solution?

Since the above was written, the ICU-in-Debian folks have asked about
any progress on the OpenTTD side of things, since they really like us to
remove this dependency from OpenTTD.

This has been discussed on IRC on a few occasions. Some observations:
 - The ICU layout stuff was removed in ICU 58.
 - Pango does layouting, but does not handle sorting, collation, etc. So
   when using Pango for layout, we'd need ICU anyway. Pango is fairly
   heavy-handed, depending on (I think) glib and gdk rendering as well.
 - IIRC Harfbuzz handles rendering of glyphs and lines, but not
   directionality detection and multi-line layouting.
 - icu-le-hb seems to be mostly unmaintained, and might be meant more of
   an example implementation than a production library. I believe it is
   mostly an API translation layer from HB to ICU Layout, without (much)
   code that actually does something.
 - ICU seems to be generally buggy too, so if we can stop depending on
   (parts of) it, that could be a feature. It seems we'll still need it
   for collating, but it could be that the ParagraphLayout code is the
   most buggy part (wishful thinking)?
 - Vendoring icu-le-hb in OpenTTD does not seem to work, since we depend
   on ICU ParagraphLayout which depends on ICU Layout (now provided by
   icu-le-hb). Vendoring both icu-le-hb and the ParagraphLayout code
   from ICU could solve this.
 - Taking the code from icu-le-hb and ICU ParagraphLayout and creating a
   new library from that might be an interesting approach. The code can
   then probably be cleaned up from there, since the ICU Layout API is
   no longer required, and can be removed internally. It might also not
   be needed to preserve the ParagraphLayout API exactly, if we feel
   that should be cleaned up (to be closer to the HB API for example),
   we can also do that.
 - It might be interesting to see what other software does (e.g.
   LibreOffice can probably do this thing as well). Does everyone use
   Pango? Or do they implement their own word-wrapping on top of HB
   layouting?

Did anyone make any progress on this subject in the last few months? @LordAro, @TrueBrain?

TrueBrain:
## Motivation / Problem

With ICU 58+, we no longer have RTL support on Linux. This is becoming a bit annoying, as we can't keep using ICU 50 to build our releases.

Debian, Ubuntu, Flatpak, and many others already have this problem, and their versions do not support RTL. Our official Linux release does, but purely as we use a very old ICU to build against.

So it is about time. Switch to ICU + harfbuzz, meaning we no longer need ICU-lx (which supplied the RTL support).

Fixes #6922. A 5 year old ticket.
Fixes #6666.

## Description

This means we have RTL support again with ICU 58+. It makes use of:
- ICU for bidi-itemization
- ICU for script-itemization
- OpenTTD for style-itemization
- harfbuzz for shaping

image
image
image
image
image

## Work-in-progress

So far the Windows (Uniscribe) and Linux (ICU + harfbuzz) look identical for single-line entries.

- [x] Support multi-line
- [x] IME input (the char-to-glyph mapping returns `nullptr` atm)
- [x] Crashlog showing harfbuzz version
- [x] Remove ICU from comments where it is no longer ICU-only
- [x] Split fallback layouter to its own file
- [x] Support custom OpenTTD icons (like train-icon, etc)

## Limitations

Some translations aren't perfect, which means some things look really odd. So often it seems like the shaping goes wrong, but in reality it is the translation who is to blame. So for now, we are comparing Windows (uniscribe) with this PR, and accept it if both render things the same.

## Checklist for review


src/gfx_layout_icu.cpp
```diff
@@ -0,0 +1,311 @@
...
+			}
+
+			this->total_advance = advance;
+
+			this->positions.push_back(advance + x);
+			this->positions.push_back(0);
+
+			hb_buffer_destroy(buf);
+			hb_font_destroy(hbfont);
+		}
```
michicc: Not sure I would have put the constructor inline, but that is purely a style choice.

TrueBrain: As I mentioned on Discord, this whole function needs to move to its own function anyway, so that will be fixed :) This was just me being lazy :D

src/gfx_layout_icu.cpp
```diff
@@ -0,0 +1,311 @@
...
+			hb_font_set_scale(hbfont, this->font->fc->GetFontSize() * FONT_SCALE, this->font->fc->GetFontSize() * FONT_SCALE);
+
+			auto buf = hb_buffer_create();
+			hb_buffer_add_utf16(buf, reinterpret_cast<uint16 *>(buff), length, run.start, run.length);
+
+			hb_buffer_set_direction(buf, run.direction == UBIDI_RTL ? HB_DIRECTION_RTL : HB_DIRECTION_LTR);
+			hb_buffer_set_script(buf, hb_script_from_string(uscript_getShortName(run.script), -1));
+			hb_buffer_set_language(buf, hb_language_from_string(_current_language->isocode, -1));
+
+			hb_shape(hbfont, buf, nullptr, 0);
```
michicc: Does this just work for our special sprite glyphs (i.e. train/truck/etc icons)? At least for Uniscribe and CoreText I had to include overrides for them to fix-up the positions and advances.

src/CMakeLists.txt
```diff
@@ -33,6 +33,12 @@ add_files(
     CONDITION SSE_FOUND
 )
 
+add_files(
+    gfx_layout_icu.cpp
+    gfx_layout_icu.h
+    CONDITION WITH_ICU_LX
```
glx22: I know it's replaced in following commit, but I'm not sure it's a working condition check at cmake level.

TrueBrain: It isn't. Will replace with `ICU_lx_FOUND`, which is. But yeah, it is being removed anyway :D

src/gfx_layout_icu.cpp
```diff
@@ -5,170 +5,523 @@
...
+#include <hb.h>
+#include <hb-ft.h>
+
+#include <deque>
 
 #include "safeguards.h"
 
-/* Implementation details of LEFontInstance */
+/** harfbuzz doesn't use floats, so we need a value to scale position with to get sub-pixel precision. */
+constexpr int FONT_SCALE = 64.0;
```
michicc: `int` and `64.0`? :)

michicc (approved):
In the end, nobody here can really verify the layout results, so if it looks alright enough, it is definitely better than nothing at all.

matthijskooijman:
Seems this PR got omitted from the changelog, but it is quite a relevant change IMHO. Maybe good to add it retroactively?

TrueBrain:
In the changelog we tend to only mention things that have an impact on users. In this case, for them nothing changed: RTL worked with 13.0 and with 14.0, with little observable difference.

The only change really is for someone like yourself. And we informed you by closing your ticket about this :D we don't really have a good way to "talk" to distro maintainer in that sense.

So personally I don't see a reason to add this to the changelog itself. And as this was the last outdated package that we depended on, I also hope this is the last time it would have been interesting to distro maintainer :D

matthijskooijman:
> In the changelog we tend to only mention things that have an impact on users. In this case, for them nothing changed: RTL worked with 13.0 and with 14.0, with little observable difference.

Ah, that makes sense. Not super-convenient for me and maybe worth reconsidering in the future (mentioning dependency changes or licensing changes/additions of 3rd party code is probably not problematic for regular users, but would help packagers), but no need to retroactively fix anything then.

> And we informed you by closing your ticket about this

Yeah, so I put a cryptic note on my own TODO list, but then was a bit confused when it was not mentioned in the changelog. Now I know why :-)

> And as this was the last outdated package that we depended on, I also hope this is the last time it would have been interesting to distro maintainer :D

I don't want to burst your bubble, but I'm pretty sure this is not going to be the last dependency change ;-p

PR #10753, opened 2023-05-01
Codechange: Move calendar date functions inside TimerGameCalendar

2TallTyler:
## Motivation / Problem

We store calendar date typedefs and variables inside TimerGameCalendar.

There are various date helper functions like `ConvertDateToYMD()` which only apply to calendar dates, and I'll soon be introducing TimerGameEconomy which needs its own implementations of these functions.

The cleanest way to make room for these, I think, is to move all these calendar date functions inside TimerGameCalendar (and later add the counterpart TimerGameEconomy methods inside that class).

## Description

- Move everything from `date.cpp` and `date_func.h` into TimerGameCalendar.
- Move YearMonthDay struct into TimerGameCalendar, since TimerGameEconomy will need a different implementation.
- Fix #includes to reflect the new home of these functions, and add direct `date_type.h` includes that must have been indirect before, and broke when I moved things.
- Delete `date.cpp` and `date_func.h`.

I will squash all these commits when merging.

## Checklist for review


src/console_cmds.cpp
```diff
@@ -29,7 +29,7 @@
 #include "strings_func.h"
 #include "viewport_func.h"
 #include "window_func.h"
-#include "date_func.h"
+#include "timer/timer_game_calendar.h"
 #include "timer/timer_game_calendar.h"
```
TrueBrain: include it once, include it twice! :D

src/engine.cpp
```diff
@@ -17,7 +17,7 @@
 #include "strings_func.h"
 #include "core/random_func.hpp"
 #include "window_func.h"
-#include "date_func.h"
+#include "timer/timer_game_calendar.h"
```
TrueBrain: You really love those double includes :)

src/graph_gui.cpp
```diff
@@ -18,6 +18,7 @@
 #include "window_func.h"
 #include "date_type.h"
 #include "timer/timer_game_calendar.h"
+#include "timer/timer_game_calendar.h"
```
TrueBrain: Yeah, I will just stop commenting on this now, okay? :)

TrueBrain (approved):

PR #10755, opened 2023-05-01
Feature: Setting to disallow level crossings with competitors

### LINKED ISSUE #10311: [Feature request]: Add a setting to prevent crossing tracks with other players roads
### Version of OpenTTD

13.0+

### Expected result

There is a setting that prohibits building tracks on someone else's roads and building a road on someone else's tracks. Useful especially for online games, but also for offline games with some AI.

### Actual result

This is a problem that has probably always existed. It existed even before the era of multiplayer games, although it mainly applies to such games.
Practically in every online game, there are cases where some players build tracks on someone else's roads. Often they do it out of ignorance or intellectual limitations (they can't predict the consequences), but just as often they do it fully consciously or even intentionally to destroy other players vehicles. Along with blocking ports and ships, this is the most common form of trolling in this game. In a game without add-ons, the player has no sensible way to protect himself against this. He can try to build one-way roads everywhere, but it just doesn't look good.

**1st example**
The purple player has entered into a conflict or just for fun destroys the green player's road vehicles for no reason.
Why? Because he can.
purple vs  green

**2nd example**
The red player, paying no attention to others, built his tracks on the road used by the orange player. As you can see, the blue player was able to build a bridge over the road. Even if he doesn't do it maliciously, paying attention to everyone, explaining what it leads to is simply tiring.
red vs  orange vs  blue

The most sensible solution that would solve this problem seems to be adding an option that will prohibit crossing roads with tracks of different companies. Possibly also crossing tracks with public roads. A similar setting is available for the construction of through stops.
settings for level crossing
+Allow crossing of tracks with public roads
+Allow crossing of tracks with other players roads

Both could be enabled by default.

### Steps to reproduce

If you play on public servers, you are probably familiar with this problem and have seen situations similar to the above many times.
If not, it may take several hours to notice the situation. Seemingly rare, but given the gameplay characteristics, for someone who plays on public servers, such situations happen in practically every game.

mrmbernardi:
## Motivation / Problem

See https://github.com/OpenTTD/OpenTTD/issues/10311

Players can construct level crossings with other player's roads/rails and cause collisions.

## Description

I added a setting which prevents players from constructing level crossings on other player's (or AI's) roads/rails.

It doesn't affect infrastructure constructed by towns or scripts.

Closes #10311 (edit 2TT)

## Limitations

Perhaps we should think more about the implications of preventing towns from constructing level crossings.

On one hand, collisions can still occur. On the other, it would be a bit cheesy to block towns from growing with just one rail.
Also, it's harder to grief with the roads towns build.

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -507,6 +507,11 @@ CommandCost CmdBuildSingleRail(DoCommandFlag flags, TileIndex tile, RailType rai
 			/* Level crossings may only be built on these slopes */
 			if (!HasBit(VALID_LEVEL_CROSSING_SLOPES, tileh)) return_cmd_error(STR_ERROR_LAND_SLOPED_IN_WRONG_DIRECTION);
 
+			if (!_settings_game.construction.crossing_with_competitor && _current_company != OWNER_DEITY) {
+				CommandCost ret = CheckTileOwnership(tile);
+				if (ret.Failed()) return ret;
+			}
+
 			CommandCost ret = EnsureNoVehicleOnGround(tile);
 			if (ret.Failed()) return ret;
 
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1426,6 +1426,9 @@ STR_CONFIG_SETTING_PLANE_CRASHES_NONE                           :None*
 STR_CONFIG_SETTING_PLANE_CRASHES_REDUCED                        :Reduced
 STR_CONFIG_SETTING_PLANE_CRASHES_NORMAL                         :Normal
 
+STR_CONFIG_SETTING_CROSSING_WITH_COMPETITOR                     :Allow level crossings with roads or rails owned by competitors: {STRING2}
+STR_CONFIG_SETTING_CROSSING_WITH_COMPETITOR_HELPTEXT            :Allow construction of level crossings on roads or rails owned by competitors
+
 STR_CONFIG_SETTING_STOP_ON_TOWN_ROAD                            :Allow drive-through road stops on roads owned by towns: {STRING2}
 STR_CONFIG_SETTING_STOP_ON_TOWN_ROAD_HELPTEXT                   :Allow construction of drive-through road stops on roads owned by towns
 STR_CONFIG_SETTING_STOP_ON_COMPETITOR_ROAD                      :Allow drive-through road stops on roads owned by competitors: {STRING2}
--- a/src/road_cmd.cpp
+++ b/src/road_cmd.cpp
@@ -737,6 +737,11 @@ CommandCost CmdBuildRoad(DoCommandFlag flags, TileIndex tile, RoadBits pieces, R
 				return_cmd_error(STR_ERROR_LAND_SLOPED_IN_WRONG_DIRECTION);
 			}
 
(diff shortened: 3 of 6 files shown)


src/saveload/saveload.h
```diff
@@ -352,6 +352,7 @@ enum SaveLoadVersion : uint16 {
 	SLV_MORE_CARGO_AGE,                     ///< 307  PR#10596 Track cargo age for a longer period.
 	SLV_LINKGRAPH_SECONDS,                  ///< 308  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	SLV_AI_START_DATE,                      ///< 309  PR#10653 Removal of individual AI start dates and added a generic one.
+	SLV_COMPETITOR_CROSSINGS,               ///< 310  PR#todo  Added setting for disabling level crossings with competitors
```
2TallTyler: You don't need this. It's used to detect when the player loads a savegame created before this feature was added, so you can do things in `afterload.cpp` to fix their game. You don't do this, so you can remove this line entirely.

2TallTyler (commented):
Please use the PR template, it's there for a reason to help us review your proposal more easily. 🙂 

Also, we typically only use an issue number in the commit when fixing something. A better title for this might be `Feature: Setting to disallow level crossings with competitors`.

src/saveload/saveload.h
mrmbernardi: Thanks, I've removed this change and the `from` line in world_setings.ini

src/rail_cmd.cpp
```diff
@@ -506,6 +506,11 @@ CommandCost CmdBuildSingleRail(DoCommandFlag flags, TileIndex tile, RailType rai
 			/* Level crossings may only be built on these slopes */
 			if (!HasBit(VALID_LEVEL_CROSSING_SLOPES, tileh)) return_cmd_error(STR_ERROR_LAND_SLOPED_IN_WRONG_DIRECTION);
 
+			if(!_settings_game.construction.crossing_with_competitor && _current_company != OWNER_DEITY) {
```
2TallTyler: Missing space between `if` and `(`

2TallTyler (commented):
Otherwise looks good to me.

src/rail_cmd.cpp
mrmbernardi: Fixed

2TallTyler (dismissed):
Technology cannot fully prevent griefing, but we can make it less fun.

It's another setting, but it does rhyme with the "allow building road stops" setting so there's precedent for this sort of thing. 

Other developers are more than welcome to disagree with me. I won't use this myself, but see no good reason not to add it. 😄

bjornwarmedal:
I agree with 2TallTyler. It's not a setting I would necessarily use as a social contract generally handles this kind of griefing, but I definitely see where it would be useful.

2TallTyler (dismissed):

mrmbernardi:
I've rebased the branch. One point in favour I'd like to highlight is that this setting is analogous to the existing "Allow drive-through road stops on roads owned by competitors" setting.

2TallTyler:
Sorry, I created a conflict. 😢 I'll re-approve once you rebase.

2TallTyler (approved):

PR #10756, opened 2023-05-02
Codechange: generify ClampTo<type>

rubidium42:
## Motivation / Problem

We have `ClampToI32` and `ClampToU16`.
We also have lots of variations of `min(<value>, <highest_value_of_type>)` or `Clamp(<value>, <lowest_value_of_type>, <highest_value_of_type>)`. If you then want that function to return the requested type, the parameters become of that type as well, performing potential down casts.
For example `Clamp<uint8_t>(257, 0, 255)` will yield 2. Similarly `ClampToU16` with `MIN_INT64` will yield 65535,
or `ClampToI32` yields -1 for `MAX_UINT64`.

## Description

Introduce a generic `ClampTo<integer_type>` function that handles all the nitty gritty details. Like `uint64_t` to `int64_t` will not cast the `uint64_t` value just to `int64_t`, it will return the minimum of MAX_INT64 and the given value, alleviating the problem with casting where the very large numbers become negative.
After that replace `ClampToI32` and `ClampToU16` with the new function, and introduce the function on more places where a similar idiom is used.

## Limitations

There could be more places that could benefit, that I might have overlooked.

## Checklist for review


src/core/math_func.hpp
```diff
@@ -124,6 +124,57 @@ static inline uint ClampU(const uint a, const uint min, const uint max)
...
+	if constexpr (std::is_same<To, From>::value) {
+		/* If we clamp to the same type, just do nothing. */
+		return value;
+	}
+
+	if constexpr (std::numeric_limits<To>::is_signed) {
+		/* The output is a signed number. */
+		if constexpr (std::numeric_limits<From>::is_signed) {
+			/* Input is signed, so use int64_t as intermediate type. */
+			return static_cast<To>(std::clamp<int64_t>(value,
```
PeterN: Is std::clamp<From>() enough, instead of promoting to 64 bits?

rubidium42: Sadly not, or at least not always. I think I have looked for some (simple) template logic to determine the bigger of the `To` and `From` types, but couldn't come up with something elegant.
```
src/script/../core/math_func.hpp: In instantiation of ‘constexpr To ClampTo(From) [with To = int; From = long int]’:
src/script/script_info.cpp:149:43:   required from here
third/src/script/../core/math_func.hpp:154:37: error: no matching function for call to ‘clamp(long int&, int, int)’
  154 |    return static_cast<To>(std::clamp(value,
      |                           ~~~~~~~~~~^~~~~~~
  155 |      std::numeric_limits<To>::lowest(), std::numeric_limits<To>::max()));
      |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/core/math_func.hpp
```diff
@@ -125,38 +125,54 @@ static inline uint ClampU(const uint a, const uint min, const uint max)
...
+ * for the return type.
  * @see Clamp(int, int, int)
  */
-static inline int32 ClampToI32(const int64 a)
+template <typename To, typename From>
+constexpr To ClampTo(From value)
 {
-	return static_cast<int32>(Clamp<int64>(a, INT32_MIN, INT32_MAX));
-}
+	if constexpr (!std::numeric_limits<To>::is_integer || !std::numeric_limits<From>::is_integer) {
```
glx22: Seems MinGW requires `#include <limits>`

src/core/math_func.hpp
rubidium42: I think I found a solution, although it does make it a more difficult to understand.

src/core/math_func.hpp
```diff
@@ -125,38 +128,61 @@ static inline uint ClampU(const uint a, const uint min, const uint max)
...
-static inline uint16 ClampToU16(const uint64 a)
-{
-	/* MSVC thinks, in its infinite wisdom, that int min(int, int) is a better
-	 * match for min(uint64, uint) than uint64 min(uint64, uint64). As such we
-	 * need to cast the UINT16_MAX to prevent MSVC from displaying its
-	 * infinite loads of warnings. */
-	return static_cast<uint16>(std::min(a, static_cast<uint64>(UINT16_MAX)));
+	if constexpr (std::is_same<To, From>::value) {
+		/* If we clamp to the same type, just do nothing. */
+		return value;
```
LordAro: I wonder if this should be an error. It seems to me that attempting to Clamp a variable to the same type is always a logic error somewhere, and could be hiding some other problem?

rubidium42: I guess that probably also holds for cases where `sizeof(From) < sizeof(To)`.

Though I wonder whether it should directly become an error. As this way it allows you to write (templated) functions that have the insurance that it will not trigger an overflow while only injecting the actual clamping logic into the function when it is actually necessary.

Could you give an actual code suggestion how you would implement that check in a manner that gives a clear error message to the developer?

LordAro: static_assert would do, no?

```c++
static_assert(!std::is_same<To, From>::value, "Should not try to clamp to the same type")
```

or similar

rubidium42: I'm not fully convinced yet it's a good idea to assert this. It would require removing some of the clamps, and for some of them keeping them might be better/safer; https://github.com/OpenTTD/OpenTTD/pull/10756/commits/aa546907af084319dc7b56c9ef869b1075934402

newgrf_industries: the callback should return 16 bits; now it requires removing the clamp, but when we change the waiting value in the future things will break while we could have guarded against it in advance.
station_cmd: even nastier; the code will just work without the clamping, also in the future but it will cause bits to get messed up, yielding potentially weird station ratings. Keeping the clamp would prevent an issue if the time since pickup becomes larger.
vehicle_gui: adding and subtracting ages of vehicles... that's going to take a long while to get into problems. So removing the clamp here seems fine.

I could even imagine, especially in the API towards NewGRFs, that we conciously add clamping regardless of the type and let the compiler, based on the actual types at that moment to determine whether a clamp is needed or not. Always the same logic, so less chance of things going awry.

PeterN: How harmful is an unnecessary clamp? It can be difficult to know if the compiler has promoted an operation up to int, or a type-change later on somewhere else could have undesirable effects.

I feel an unnecessary clamp is less harmful than a missing clamp that was not originally put in place as it was unnecessary at the time.

rubidium42: The `ClampTo` can be written in such a manner that if the "to" type can hold all values of "from", e.g. from `uint16_t` to `int32_t` to simply nothing. `ClampTo` would likely get optimized away with even low optimisation levels.
`int` to `uint32_t` would at most be optimized to a `max(0, value)`, so one comparison, and then it's up to the compiler to optimize that away if it has more knowledge.

PeterN: Yes, I was thinking more in terms of a rationale for prohibiting clamping to the same type, rather than the resulting compiled code which more than likely would be optimized away.

"We shouldn't do it because it's not strictly necessary" is quite weak when it comes to C++ :-)

rubidium42: I changed it back to just do nothing when no actual clamping is needed. That indeed feels like the best way forward by allowing adding `Clamp` pre-emptively to prevent issues at a later time. In a sense it would be documenting in actual code what the behaviour of the result is expected to be, regardless of the current input type.

src/company_gui.cpp
```diff
@@ -2650,8 +2650,8 @@ struct CompanyWindow : Window
 			default: NOT_REACHED();
 
 			case WID_C_GIVE_MONEY: {
-				Money money = (Money)(std::strtoull(str, nullptr, 10) / _currency->rate);
-				uint32 money_c = Clamp(ClampToI32(money), 0, 20000000); // Clamp between 20 million and 0
+				uint32_t money = ClampTo<uint32_t>(std::strtoull(str, nullptr, 10) / _currency->rate);
+				uint32 money_c = std::min(money, 20000000U); // Maximum is 20 million
 
 				Command<CMD_GIVE_MONEY>::Post(STR_ERROR_CAN_T_GIVE_MONEY, money_c, (CompanyID)this->window_number);
```
PeterN: Probably a bit too much patch-creep, but can we get rid of `uint32 money_c` here and just keep it as `money`?

src/highscore.cpp
```diff
@@ -132,7 +132,7 @@ void SaveToHighScore()
 		for (i = 0; i < SP_SAVED_HIGHSCORE_END; i++) {
 			for (hs = _highscore_table[i]; hs != endof(_highscore_table[i]); hs++) {
 				/* First character is a command character, so strlen will fail on that */
-				byte length = std::min(sizeof(hs->company), StrEmpty(hs->company) ? 0 : strlen(&hs->company[1]) + 1);
+				byte length = ClampTo<byte>(std::min(sizeof(hs->company), StrEmpty(hs->company) ? 0 : strlen(&hs->company[1]) + 1));
```
PeterN, line 135: Is this in the wrong commit?

src/company_gui.cpp
rubidium42: I think so... but we'll have to wait on the compilers for the final verdict.

src/highscore.cpp
rubidium42: Yes

PeterN (approved):

PR #10758, opened 2023-05-02
Fix: crash in emscripten when saving games

TrueBrain:
## Motivation / Problem

Saving in emscripten said BOOM!

## Description

Don't allocate 128KB on stack, but rather on the heap.

## Limitations

The real reason this was noticed on emscripten, is most likely because the stack is nearly full to start with. Than we really shouldn't be allocating these big blocks on the stack, but there might be more crashes on the horizon here!

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -2582,6 +2582,7 @@ struct ZlibLoadFilter : LoadFilter {
 /** Filter using Zlib compression. */
 struct ZlibSaveFilter : SaveFilter {
 	z_stream z; ///< Stream state we are writing to.
+	byte fwrite_buf[MEMORY_CHUNK_SIZE]; ///< Buffer for writing to the file.
 
 	/**
 	 * Initialise this filter.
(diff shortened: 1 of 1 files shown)


src/saveload/saveload.cpp
```diff
@@ -2582,6 +2582,7 @@ struct ZlibLoadFilter : LoadFilter {
 /** Filter using Zlib compression. */
 struct ZlibSaveFilter : SaveFilter {
 	z_stream z; ///< Stream state we are writing to.
+	byte fwrite_byte[MEMORY_CHUNK_SIZE]; // output buffer
```
PeterN: Doxygen or not? Same with below.

PeterN (dismissed):

src/saveload/saveload.cpp
```diff
@@ -2582,6 +2582,7 @@ struct ZlibLoadFilter : LoadFilter {
 /** Filter using Zlib compression. */
 struct ZlibSaveFilter : SaveFilter {
 	z_stream z; ///< Stream state we are writing to.
+	byte fwrite_byte[MEMORY_CHUNK_SIZE]; ///< Buffer for writing to the file.
```
LordAro: Bikeshedding! Why `fwrite_byte` ? There are lots of bytes :) fwrite_buf seems more appropriate

LordAro: Also, is it worth just sticking this in the base class?

PeterN: It's a different size depending on the filter, so probably not.

PeterN (approved):

PR #10761, opened 2023-05-04
Codechange: make TimerGameCalendar Date and Year types strongly typed

TrueBrain:
## Motivation / Problem

With #10700 on the horizon, I am a bit afraid we will mix up Calendar with Economy, and get some weird buggy behaviour because of it. So I was wondering: can we make Date strongly typed, to avoid that issue.

This PR is a proof-of-concept for that. Curious what developers think.

And of course, by making it strongly typed, there is no actual functional change. It is just meant to assist the developer in not making silly mistakes. This was already fruitful, looking at all the PRs that came from this PR, with weird mistakes and boo-boos we made in our code-base.

## Description

I went for a slightly different approach than `TileIndex`, as that requires defining a struct. I was more interested in using `using` to create a new type. Got some inspiration from https://github.com/anthonywilliams/strong_typedef to make that happen.

Initially, I made the conversion to `int32` implicit (so it auto-cases to `int32` if it wants to). This had as nasty side-result that you could assign a `Year` to a `Date` and the other way around. As the compiler, as clever as it is, casted the one first to an int, to assign it to the other. This is kinda defeating the purpose of having strong-types to prevent people mixing up Calendar and Economy.

So, making it explicit is the way to go. Sadly, that showed tons of places in our code where we take some ... freedom in terms of using a date/year to use as integer. The weirdest things. A lot of small PRs came from this to address most of these.

## Limitations

Although performance shouldn't be impacted, as this is just compiler-hints etc, it is not a guarantee that there won't be any if used wrongly. Especially when disabling all optimization, some performance degradation is to be expected.

## Checklist for review

### DIFF
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -10,6 +10,7 @@
 #ifndef MATH_FUNC_HPP
 #define MATH_FUNC_HPP
 
+#include "strong_typedef_type.hpp"
 
 /**
  * Returns the absolute value of (scalar) variable.
@@ -162,7 +163,7 @@ static inline uint ClampU(const uint a, const uint min, const uint max)
  * for the return type.
  * @see Clamp(int, int, int)
  */
-template <typename To, typename From>
+template <typename To, typename From, std::enable_if_t<std::is_integral<From>::value, int> = 0>
 constexpr To ClampTo(From value)
 {
 	static_assert(std::numeric_limits<To>::is_integer, "Do not clamp from non-integer values");
@@ -214,6 +215,15 @@ constexpr To ClampTo(From value)
 }
 
 /**
+ * Specialization of ClampTo for #StrongType::Typedef.
+ */
+template <typename To, typename From, std::enable_if_t<std::is_base_of<StrongTypedefBase, From>::value, int> = 0>
+constexpr To ClampTo(From value)
+{
+	return ClampTo<To>(static_cast<typename From::BaseType>(value));
+}
+
+/**
  * Returns the (absolute) difference between two (scalar) variables
  *
  * @param a The first scalar
@@ -254,10 +264,14 @@ static inline bool IsInsideBS(const T x, const size_t base, const size_t size)
  * @param max The maximum of the interval
  * @see IsInsideBS()
  */
-template <typename T>
+template <typename T, std::enable_if_t<std::disjunction_v<std::is_convertible<T, size_t>, std::is_base_of<StrongTypedefBase, T>>, int> = 0>
 static constexpr inline bool IsInsideMM(const T x, const size_t min, const size_t max) noexcept
 {
-	return (size_t)(x - min) < (max - min);
+	if constexpr (std::is_base_of_v<StrongTypedefBase, T>) {
+		return (size_t)(static_cast<typename T::BaseType>(x) - min) < (max - min);
+	} else {
+		return (size_t)(x - min) < (max - min);
+	}
 }
 
 /**
--- a/src/core/strong_typedef_type.hpp
+++ b/src/core/strong_typedef_type.hpp
@@ -10,74 +10,199 @@
 #ifndef STRONG_TYPEDEF_TYPE_HPP
 #define STRONG_TYPEDEF_TYPE_HPP
 
-/** Non-templated base for #StrongTypedef for use with type trait queries. */
+#include "../3rdparty/fmt/format.h"
+
+/** Non-templated base for #StrongType::Typedef for use with type trait queries. */
 struct StrongTypedefBase {};
 
-/**
- * Templated helper to make a type-safe 'typedef' representing a single POD value.
- * A normal 'typedef' is not distinct from its base type and will be treated as
- * identical in many contexts. This class provides a distinct type that can still
- * be assign from and compared to values of its base type.
- *
- * @note This is meant to be used as a base class, not directly.
- * @tparam T Storage type
- * @tparam Tthis Type of the derived class (i.e. the concrete usage of this class).
- */
-template <class T, class Tthis>
-struct StrongTypedef : StrongTypedefBase {
-	using Type = T;
+namespace StrongType {
+	/**
+	 * Mix-in which makes the new Typedef comparable with itself and its base type.
+	 */
+	struct Compare {
+		template <typename TType, typename TBaseType>
+		struct mixin {
+			friend constexpr bool operator ==(const TType &lhs, const TType &rhs) { return lhs.value == rhs.value; }
+			friend constexpr bool operator ==(const TType &lhs, const TBaseType &rhs) { return lhs.value == rhs; }
 
-	T value{}; ///< Backing storage field.
+			friend constexpr bool operator !=(const TType &lhs, const TType &rhs) { return lhs.value != rhs.value; }
+			friend constexpr bool operator !=(const TType &lhs, const TBaseType &rhs) { return lhs.value != rhs; }
 
-	debug_inline constexpr StrongTypedef() = default;
-	debug_inline constexpr StrongTypedef(const StrongTypedef &o) = default;
-	debug_inline constexpr StrongTypedef(StrongTypedef &&o) = default;
+			friend constexpr bool operator <=(const TType &lhs, const TType &rhs) { return lhs.value <= rhs.value; }
+			friend constexpr bool operator <=(const TType &lhs, const TBaseType &rhs) { return lhs.value <= rhs; }
 
-	debug_inline constexpr StrongTypedef(const T &value) : value(value) {}
+			friend constexpr bool operator <(const TType &lhs, const TType &rhs) { return lhs.value < rhs.value; }
+			friend constexpr bool operator <(const TType &lhs, const TBaseType &rhs) { return lhs.value < rhs; }
 
-	debug_inline constexpr Tthis &operator =(const Tthis &rhs) { this->value = rhs.value; return static_cast<Tthis &>(*this); }
-	debug_inline constexpr Tthis &operator =(Tthis &&rhs) { this->value = std::move(rhs.value); return static_cast<Tthis &>(*this); }
-	debug_inline constexpr Tthis &operator =(const T &rhs) { this->value = rhs; return static_cast<Tthis &>(*this); }
+			friend constexpr bool operator >=(const TType &lhs, const TType &rhs) { return lhs.value >= rhs.value; }
+			friend constexpr bool operator >=(const TType &lhs, const TBaseType &rhs) { return lhs.value >= rhs; }
 
-	explicit constexpr operator T() const { return this->value; }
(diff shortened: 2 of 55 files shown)


TrueBrain:
I shifted gears, and went for a "I want to use `using` to define a strongly typed type`. I got some inspiration, and that is easily doable. Means we don't need to define a struct to create a new type, but we can just use `using`.

The trick here is that to avoid 2 `NewStrongTypedef` of the same base-type being equal, that we tag them with an unique tag. This works wonderful.

Sadly, I noticed that the compiler was sneaky converting everything back and forth between `int32` for us, so you could assign a year to a date. To solve this, I made the conversion to `int32` explicit, and boy, this showed a lot of places were we just depend on that. Not surprising, given we came from a world where you could. But plenty of places that are ... weird. Like `% 7` on the date.

Anyway, much more interested in the overall idea of this PR, if we like defining such type this strongly. Does it actually help?

TrueBrain:
Another update. I decided to go for Mixins, after the idea of https://github.com/anthonywilliams/strong_typedef.

What is nice about this, is that it means `TileIndex` can be converted too, and we deduplicate defining a new type by a lot.
The downside is that the line to define a type becomes a bit lengthy :)

2TallTyler:
@TrueBrain A lot of this PR is way over my head with templating and casting, but you are absolutely right that we'll need this before #10700. Maybe some more experienced developers can weigh in on this? I would be eternally grateful. ❤️

2TallTyler:
The same way you make `SetDParam()` take strong-typed date arguments, it might be helpful (if that's possible) to do the same with settings. That could save a lot of casting, I think.

src/network/network_server.cpp
```diff
@@ -1838,7 +1838,7 @@ static IntervalTimer<TimerGameCalendar> _network_daily({TimerGameCalendar::DAY,
 	if (!_network_server) return;
 
 	NetworkAdminUpdate(ADMIN_FREQUENCY_DAILY);
-	if ((TimerGameCalendar::date % 7) == 3) NetworkAdminUpdate(ADMIN_FREQUENCY_WEEKLY);
+	if (((int32_t)TimerGameCalendar::date % 7) == 3) NetworkAdminUpdate(ADMIN_FREQUENCY_WEEKLY);
```
TrueBrain: TODO: this should be fixed more properly, by adding `WEEK` and `QUARTER` similar as `DAY` in `TimerGameCalendar`, so a timer could be used here too.

src/saveload/oldloader_sl.cpp
```diff
@@ -416,7 +416,7 @@ static bool FixTTOEngines()
 			if (TimerGameCalendar::date >= e->intro_date && HasBit(e->info.climates, 0)) {
 				e->flags |= ENGINE_AVAILABLE;
 				e->company_avail = MAX_UVALUE(CompanyMask);
-				e->age = TimerGameCalendar::date > e->intro_date ? (TimerGameCalendar::date - e->intro_date) / 30 : 0;
+				e->age = TimerGameCalendar::date > e->intro_date ? (int32_t)(TimerGameCalendar::date - e->intro_date) / 30 : 0;
```
TrueBrain: We kinda need DateToMonth or something here too. The `/ 30` is unclear as it is.

Edit: or just unmagicify the `30` with a constant. Might be sufficient too.

src/table/settings/world_settings.ini
```diff
@@ -135,9 +135,9 @@ var      = game_creation.ending_year
 type     = SLE_INT32
 from     = SLV_ENDING_YEAR
 flags    = SF_GUI_0_IS_SPECIAL
-def      = DEF_END_YEAR
-min      = MIN_YEAR
-max      = MAX_YEAR - 1
+def      = (int32_t)DEF_END_YEAR
+min      = (int32_t)MIN_YEAR
+max      = (int32_t)MAX_YEAR - 1
```
TrueBrain: See if we can render these casts automatically

src/saveload/oldloader_sl.cpp
TrueBrain: Not for this PR. It needs a cast either way, so this is fine.

src/core/math_func.hpp
```diff
@@ -254,10 +261,14 @@ static inline bool IsInsideBS(const T x, const size_t base, const size_t size)
  * @param max The maximum of the interval
  * @see IsInsideBS()
  */
-template <typename T>
+template<typename T, std::enable_if_t<std::disjunction_v<std::is_convertible<T, size_t>, std::is_base_of<StrongTypedefBase, T>>, int> = 0>
```
2TallTyler:
```suggestion
template <typename T, std::enable_if_t<std::disjunction_v<std::is_convertible<T, size_t>, std::is_base_of<StrongTypedefBase, T>>, int> = 0>
```
Other uses in this file have a space, while cppreference has no space. We could at least be consistent with ourselves. 😄

src/landscape.cpp
```diff
@@ -787,7 +787,7 @@ void RunTileLoop()
 		_tile_type_procs[GetTileType(tile)]->tile_loop_proc(tile);
 
 		/* Get the next tile in sequence using a Galois LFSR. */
-		tile = (tile >> 1) ^ (-(int32_t)(tile & 1) & feedback);
+		tile = (static_cast<uint32_t>(tile) >> 1) ^ (-(int32_t)(static_cast<uint32_t>(tile) & 1) & feedback);
```
2TallTyler, line 790: The changes in this file deal with tiles, not dates. Should it be a separate commit or maybe a separate PR?

(Same goes for changes in `newgrf_house.cpp`, `newgrf_industries.cpp`, `newgrf_town.cpp`, `town_cmd.cpp`, `yapf_node.hpp`, and three files in `saveload/*`, just to make a nice list for you of what I found. 😉)

src/newgrf_engine.cpp
```diff
@@ -597,7 +597,7 @@ static uint32_t VehicleGetVariable(Vehicle *v, const VehicleScopeResolver *objec
 			}
 
 		case 0x4B: // Long date of last service
-			return v->date_of_last_service_newgrf;
+			return static_cast<int32_t>(v->date_of_last_service);
```
2TallTyler:
```suggestion
			return static_cast<int32_t>(v->date_of_last_service_newgrf);
```

src/script/api/script_basestation.cpp
```diff
@@ -63,5 +63,5 @@
 {
 	if (!IsValidBaseStation(station_id)) return ScriptDate::DATE_INVALID;
 
-	return (ScriptDate::Date)::BaseStation::Get(station_id)->build_date;
+	return (ScriptDate::Date)(int32_t)::BaseStation::Get(station_id)->build_date;
```
2TallTyler, line 66: It feels weird to cast to an `int32_t` and then immediately to a `ScriptDate::Date`. Can we not do this in one step? (Honest question, I have no idea how ScriptDates work, nor casting in general. 😅)

src/timer/timer_game_calendar.h
```diff
@@ -76,10 +77,12 @@ class TimerGameCalendar {
...
 	};
 
-	using Date = int32_t; ///< The type to store our dates in
+	/** The type to store our dates in. */
+	using Date = StrongType::Typedef<int32_t, struct DateTag, StrongType::Explicit, StrongType::Compare, StrongType::Integer>;
+
 	using DateFract = uint16_t; ///< The fraction of a date we're in, i.e. the number of ticks since the last date changeover
 
-	using Year = int32_t; ///< Type for the year, note: 0 based, i.e. starts at the year 0.
+	using Year = StrongType::Typedef<int32_t, struct YearTag, StrongType::Explicit, StrongType::Compare, StrongType::Integer>; ///< Type for the year, note: 0 based, i.e. starts at the year 0.
```
2TallTyler:
```suggestion
	/** Type for the year, note: 0 based, i.e. starts at the year 0. */
	using Year = StrongType::Typedef<int32_t, struct YearTag, StrongType::Explicit, StrongType::Compare, StrongType::Integer>;
```
If a block comment is good enough for Date, it's good enough for this. 😉 

...is your OCD contagious? 😛

src/script/api/script_basestation.cpp
TrueBrain: Sadly, no. The first case changes the StrongTypeDef to an integer, the second to the Script-specific type. It is a bit annoying, but it is what it is. Endless casts in scripts are just .. what they are :P

src/landscape.cpp
TrueBrain: No, sadly, these where it runs an operator on the StrongTypeDef, like `&`, `>>`, etc, has to be done here. Before this PR, that was a valid operator. But I think it has very little definition .. what type does `date >> 1` return .. so I revoked access to all those kind of operators.

As a downside, it means that the `TileIndex` also no longer can use them. And those you listed. So they have to be in this PR!

2TallTyler: Okay! Maybe a separate commit, then? ("No" is a valid answer. 😃)

TrueBrain: That would be really really difficult to make that a sane and compilable commit :) I rather keep it like this. It is only a few lines :)

2TallTyler (dismissed):
I don't understand template stuff well enough to review it, but the casting all looks correct to me. 

If someone else reviews the template stuff, I can continue working on #10700 tomorrow! Thanks again for taking this on! 😄

src/core/strong_typedef_type.hpp
```diff
@@ -10,74 +10,199 @@
...
+			friend constexpr bool operator ==(const TType &lhs, TCompatibleType rhs) { return lhs.value == (TBaseType)rhs; }
+			friend constexpr bool operator !=(const TType &lhs, TCompatibleType rhs) { return lhs.value != (TBaseType)rhs; }
+
+			friend constexpr bool operator <=(const TType &lhs, TCompatibleType rhs) { return lhs.value <= (TBaseType)rhs; }
+			friend constexpr bool operator <(const TType &lhs, TCompatibleType rhs) { return lhs.value < (TBaseType)rhs; }
+			friend constexpr bool operator >=(const TType &lhs, TCompatibleType rhs) { return lhs.value >= (TBaseType)rhs; }
+			friend constexpr bool operator >(const TType &lhs, TCompatibleType rhs) { return lhs.value > (TBaseType)rhs; }
+
+			friend constexpr TType operator +(const TType &lhs, TCompatibleType rhs) { return { (TBaseType)(lhs.value + rhs) }; }
+			friend constexpr TType operator -(const TType &lhs, TCompatibleType rhs) { return { (TBaseType)(lhs.value - rhs) }; }
```
michicc: The templates in general look as reasonable as they can I think.

For this specific area, was the C-style cast an explicit choice, i.e. do you want the `reinterpret_cast` case to included?

2TallTyler (approved):

PR #10762, opened 2023-05-04
Fix: no_http_content_downloads and use_relay_service as private settings

TrueBrain:
## Motivation / Problem

While working on the automated opt-in survey, I noticed two settings which I believe should be kept private, and not send in the survey. Additionally, we also don't want them when people send in crash reports.

## Description

So, I moved both settings to the "private.cfg" we have for information like this.

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -154,10 +154,11 @@ public:
  * location. These versions assist with situations like that.
  */
 enum IniFileVersion : uint32 {
-	IFV_0,                 ///< 0  All versions prior to introduction.
-	IFV_PRIVATE_SECRETS,   ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
-	IFV_GAME_TYPE,         ///< 2  PR#9515  Convert server_advertise to server_game_type.
-	IFV_LINKGRAPH_SECONDS, ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
+	IFV_0,                                                 ///< 0  All versions prior to introduction.
+	IFV_PRIVATE_SECRETS,                                   ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
+	IFV_GAME_TYPE,                                         ///< 2  PR#9515  Convert server_advertise to server_game_type.
+	IFV_LINKGRAPH_SECONDS,                                 ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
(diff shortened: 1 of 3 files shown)


src/settings.cpp
```diff
@@ -154,10 +154,11 @@ class ConfigIniFile : public IniFile {
...
 enum IniFileVersion : uint32 {
-	IFV_0,                 ///< 0  All versions prior to introduction.
-	IFV_PRIVATE_SECRETS,   ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
-	IFV_GAME_TYPE,         ///< 2  PR#9515  Convert server_advertise to server_game_type.
-	IFV_LINKGRAPH_SECONDS, ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
+	IFV_0,                                                 ///< 0  All versions prior to introduction.
+	IFV_PRIVATE_SECRETS,                                   ///< 1  PR#9298  Moving of settings from openttd.cfg to private.cfg / secrets.cfg.
+	IFV_GAME_TYPE,                                         ///< 2  PR#9515  Convert server_advertise to server_game_type.
+	IFV_LINKGRAPH_SECONDS,                                 ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
+	IFV_NETWORK_PRIVATE_SETTINGS,                          ///< 4  PR#???   Move no_http_content_downloads / use_relay_service to private settings.
```
```suggestion
	IFV_NETWORK_PRIVATE_SETTINGS,                          ///< 4  PR#10762 Move no_http_content_downloads / use_relay_service to private settings.
```

rubidium42:
I wonder whether this would hinder debugging/bug reports. For example OpenTTD crashing when there is no TURN-connection possible. I'd not run into that problem as I'd happily get the TURN-connection going. Similarly with downloading some (specific) content from the content server instead of HTTP.
On the other hand, if they provide a proper backtrace/crash dump then it would probably be obviously from those where the problem lies.

TrueBrain:
There is so much happening before a TURN connection is tried, that knowing the value of this setting doesn't help us debug problems. As you said, the backtrack needs to hint to it.

Similar for http; it silently falls back to TCP if http fails. So here too, the backtrack needs to give the hints; not this setting.

Nevertheless, both settings are private settings, as they are personal and are not influencing the game itself in any way. Similar for the other private settings.

As it goes: do we want to know ad "it might be useful"? Sure. Should we know or care about any of the private settings? No :)

rubidium42 (approved):

PR #10768, opened 2023-05-05
Codechange: Clean up scrollbars by adding iteration

PeterN:
## Motivation / Problem

There are about 5 million different way we iterate through visible scrollbar elements. They are mostly very C-style, and often involve casts due to data type conversion issues.

## Description

Implement a range iterator to allow iteration through the visible elements that a scrollbar refers to.

## Checklist for review

### DIFF
--- a/src/widget_type.h
+++ b/src/widget_type.h
@@ -682,6 +682,12 @@ public:
 	}
 
 	/**
+	 * Get the end of the visible valid elements.
+	 * @return index to the end of the visible valid elements.
+	 */
+	inline int End() const { return std::min<int>(this->GetPosition() + this->GetCapacity(), this->GetCount()); }
+
+	/**
 	 * Checks whether given current item is visible in the list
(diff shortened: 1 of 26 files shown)


src/widget_type.h
```diff
@@ -17,6 +17,25 @@
 #include "gfx_type.h"
 #include "window_type.h"
 
+/* Range Iterator Helper. This should be somewhere else, and documented. Where? */
+template <class Iter>
```
rubidium42: Isn't this just a simple version of `(std::)span` (core/span_type.hpp)?

PeterN: Perhaps, now trying to work out how to use that.

PeterN: Unless I was trying it wrong, our span only seems to support iterators that are pointers.

PeterN:
Mostly implemented by #12190

PR #10769, opened 2023-05-06
Codechange: get rid of C-style allocation/free in script API

rubidium42:
## Motivation / Problem

Mix and match of C and C++ style code.

## Description

Slowly pulling at the thread removing one manually allocated/managed string/object after the other.
This includes redoing the API wrapper layer to use `std::string`/`std::optional<std::string>` instead of `char *` and `const char *` (the former you had to free, the latter you didn't free), and as such getting rid of the `SQAutoFreePtrs` construct.
It gets rid of all `stredups`, almost all `seprintf` and `strecpy` (the remaining ones are for ScriptText and called from `FormatStrings`, so that is left as is for decoupling). Also gets rid of all `free`/`MallocT` and friends, except for the allocator that Squirrel uses.

## Limitations

Squirrel itself was not touched, however that has its own string type so where needed wrappers.
Some `.c_str()`s were added at the fringes, for example in the crashlog code to not pull rewriting that to C++ strings into this PR.
This builds further upon #10767.

## Checklist for review


src/script/api/script_road.cpp
```diff
@@ -379,7 +379,7 @@ static bool NormaliseTileOffset(int32 *tile)
 		return false;
 }
 
-/* static */ SQInteger ScriptRoad::CanBuildConnectedRoadParts(ScriptTile::Slope slope_, Array<> existing, TileIndex start_, TileIndex end_)
+/* static */ SQInteger ScriptRoad::CanBuildConnectedRoadParts(ScriptTile::Slope slope_, Array<> &&existing, TileIndex start_, TileIndex end_)
```
TrueBrain, line 382: This one is a bit scary, and hard to judge from the context of this PR if it is actually correct. Can it actually be used successful from a script?

src/script/script_info_dummy.cpp
```diff
@@ -54,52 +52,50 @@ void Script_CreateDummyInfo(HSQUIRRELVM vm, const char *type, const char *dir)
 	NOT_REACHED();
 }
 
+static std::vector<std::string> EscapeQuotesAndSlashesAndSplitOnNewLines(const std::string &message)
```
TrueBrain, line 61: missing doxygen

src/script/squirrel_helper.hpp
```diff
@@ -52,48 +39,58 @@ namespace SQConvert {
 	template <> struct Return<Money>        { static inline int Set(HSQUIRRELVM vm, Money res)       { sq_pushinteger(vm, res); return 1; } };
 	template <> struct Return<TileIndex>    { static inline int Set(HSQUIRRELVM vm, TileIndex res)   { sq_pushinteger(vm, (int32)res.value); return 1; } };
 	template <> struct Return<bool>         { static inline int Set(HSQUIRRELVM vm, bool res)        { sq_pushbool   (vm, res); return 1; } };
-	template <> struct Return<char *>       { static inline int Set(HSQUIRRELVM vm, char *res)       { if (res == nullptr) sq_pushnull(vm); else { sq_pushstring(vm, res, -1); free(res); } return 1; } };
-	template <> struct Return<const char *> { static inline int Set(HSQUIRRELVM vm, const char *res) { if (res == nullptr) sq_pushnull(vm); else { sq_pushstring(vm, res, -1); } return 1; } };
+	template <> struct Return<char *>       { /* Do not use char *, use std::optional<std::string> instead. */ };
+	template <> struct Return<const char *> { /* Do not use const char *, use std::optional<std::string> instead. */ };
```
TrueBrain, line 43: why not just remove them? Or otherwise `= delete`? For sure the comment will be missed ... `assert_compile(false)`?

rubidium42: Removing them means it will fall back to the `void *` variant, so if there's any patch(pack) with custom API's, they will just succeed compiling and fail when running. So I left the tempated structs alive, but not the function that they would actually use. In other words, that patch(pack) will now fail to compile because the `Set` function cannot be found.

I doubt `= delete` works for templated structs. As far as I am aware that is for automatically generated functions.

`static_assert(false);` simply fails the build, even when it's not used.

src/script/api/script_road.cpp
rubidium42: It works if I put the following into regression.nut:
```
	local array = [AIMap.GetTileIndex(0, -1), AIMap.GetTileIndex(0, 1)]
	print("    CanBuildConnectedRoadParts():  " + AIRoad.CanBuildConnectedRoadParts(AITile.SLOPE_FLAT, array, AIMap.GetTileIndex(-1, 0), AIMap.GetTileIndex(1, 0)));
```
Works in the sense of: it does not crash.
Side note: I could not find any direct users of the function on github.

Now for the kicker... the same code in master fails because it says it expects `integer` instead of `array`. This can be seen in the automatically generated ai_road.hpp.sq. But I have not changed the code that generates `array` code in this branch?!?

rubidium42: By the way, the result of that bit of code with this PR is the same as with the releases/13 branch. The generated ai_road.hpp.sq is practically the same in release/13 compared to this PR except the `ST_AI` -> `ScriptType::AI` change, so apparantly I unknowingly fixed a bug in master?

TrueBrain: haha; great, tnx for checking :) It was scary looking :)

src/script/squirrel_helper.hpp
TrueBrain: then at the very least a `NOT_REACHED`? This is now too hidden that we removed an implementation, in my opinion :)

TrueBrain: Owh, it is a struct, lolz; sorry, that took a bit too long. I see what you mean with `static_assert` too :D Yeah, okay, fine .. I still don't like we leave code like this behind, but it is what it is :)

TrueBrain (approved):

PR #10770, opened 2023-05-06
Change: Replace ScriptLog data array with std::deque.

PeterN:
## Motivation / Problem

ScriptLog's data is stored in a manually managed ring-buffer, which is not iterable and requires manual memory management.

It is also passed around as a void pointer (to avoid exposing internals presumably) which means casting is involved.

## Description

Change to the log data to use a std::deque instead, and store as such in the script storage.

Due to cyclic header dependency this requires moving the data types used by ScriptLog out of the ScriptLog class.

## Checklist for review

### DIFF
--- a/src/script/api/script_log.cpp
+++ b/src/script/api/script_log.cpp
@@ -8,6 +8,7 @@
 /** @file script_log.cpp Implementation of ScriptLog. */
 
 #include "../../stdafx.h"
+#include "script_log_types.hpp"
 #include "script_log.hpp"
 #include "../../core/alloc_func.hpp"
 #include "../../debug.h"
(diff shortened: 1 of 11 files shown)


src/script/script_storage.hpp
```diff
@@ -54,7 +56,7 @@ friend class ScriptObject;
 	RailType rail_type;              ///< The current railtype we build.
 
 	void *event_data;                ///< Pointer to the event data storage.
-	void *log_data;                  ///< Pointer to the log data storage.
+	ScriptLogTypes::LogData log_data;     ///< Log data storage.
```
rubidium42: Just remove the remainder of the spaces as well.

src/script/script_instance.hpp
```diff
@@ -95,7 +96,7 @@ class ScriptInstance {
 	/**
 	 * Get the log pointer of this script.
 	 */
-	void *GetLogPointer();
+	ScriptLogTypes::LogData &GetLogPointer();
```
rubidium42: Maybe change all the `GetLogPointer` to `GetLogData`?

src/script/api/script_log.cpp
```diff
@@ -18,75 +19,46 @@
...
-	if (log->used != log->count) log->used++;
+	/* Limit the log to 400 lines. */
+	if (logdata.size() >= 400U) logdata.pop_front();
 
-	/* Free last message, and write new message */
-	free(log->lines[log->pos]);
-	log->lines[log->pos] = stredup(message);
-	log->type[log->pos] = level;
+	auto &line = logdata.emplace_back();
+	line.text = message;
```
rubidium42: Maybe something like the following to not have to copy unneeded text? On the other hand, it's relatively unlikely for there to be newlines in the message so it might also be overkill/unneeded complication.
```
char *newline = strchr(message, '\n');
line.text = std::string(message, 0, newline == nullptr ? strlen(message) : newline - message);
```
Not that it matters much as with #10769 we'd be getting a `std::string&` as parameter, and then it's just `message.substr(0, message.find_first_of('\n'));`

rubidium42 (dismissed):

rubidium42 (approved):

PR #10772, opened 2023-05-06
Codechange: replace C-style strings with C++-style strings in textfile

TrueBrain:
## Motivation / Problem

For an upcoming PR (survey addition), I want to reuse the textfile GUI. But the content in that PR won't come from a file, but from memory.

Sadly, the current code uses a C-style approach, which made it really tricky to combine these two worlds. So, the easy solution: convert the current code to more C++.

## Description

Exactly what you expect: C-style strings to C++-style strings.

## Limitations

I need to split up this PR before it becomes reviewable.

## Checklist for review

### DIFF
--- a/src/strings_func.h
+++ b/src/strings_func.h
@@ -261,9 +261,9 @@ public:
 
 	/**
 	 * Get the next string to search through.
-	 * @return The next string or nullptr if there is none.
+	 * @return The next string or nullopt if there is none.
 	 */
-	virtual const char *NextString() = 0;
+	virtual std::optional<std::string_view> NextString() = 0;
 
 	/**
 	 * Get the default (font) size of the string.
--- a/src/textfile_gui.cpp
+++ b/src/textfile_gui.cpp
@@ -70,11 +70,6 @@ TextfileWindow::TextfileWindow(TextfileType file_type) : Window(&_textfile_desc)
 	this->hscroll->SetStepSize(10); // Speed up horizontal scrollbar
 }
 
-/* virtual */ TextfileWindow::~TextfileWindow()
-{
-	free(this->text);
-}
-
 /**
  * Get the total height of the content displayed in this window, if wrapping is disabled.
  * @return the height in pixels
@@ -199,9 +194,9 @@ void TextfileWindow::SetupScrollbars(bool force_reflow)
 	return FS_MONO;
 }
 
-/* virtual */ const char *TextfileWindow::NextString()
+/* virtual */ std::optional<std::string_view> TextfileWindow::NextString()
 {
-	if (this->search_iterator >= this->lines.size()) return nullptr;
+	if (this->search_iterator >= this->lines.size()) return std::nullopt;
 
 	return this->lines[this->search_iterator++].text;
 }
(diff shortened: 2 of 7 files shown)


src/textfile_gui.cpp
```diff
@@ -344,48 +339,48 @@ static void Xunzip(byte **bufp, size_t *sizep)
 	FILE *handle = FioFOpenFile(textfile, "rb", dir, &filesize);
 	if (handle == nullptr) return;
 
-	this->text = ReallocT(this->text, filesize);
-	size_t read = fread(this->text, 1, filesize, handle);
+	char *buf = MallocT<char>(filesize);
```
LordAro, line 342: Replace this stuff with new/delete while you're at it?

TrueBrain: I tried to minimise this PR, as my first attempts for removing this alloc were a nightmare .. the decompress functions also do magic, it really wants to realloc, and brrrr .. so I rather keep that for another day. Ideally while fixing fread too.

nielsmh:
Possibly related: #7786
(I should just finish that one already.)

src/strings_func.h
```diff
@@ -15,6 +15,7 @@
 #include "gfx_type.h"
 #include "core/bitmath_func.hpp"
 #include "vehicle_type.h"
+#include <optional>
```
rubidium42: `optional` got places into `stdafx.h` by the precompiled headers PR, so it's not needed here anymore.

src/textfile_gui.cpp
```diff
@@ -344,48 +339,48 @@ static void Xunzip(byte **bufp, size_t *sizep)
...
+		if (c == '\t' || c == '\r') c = ' ';
 	}
 
 	/* Check for the byte-order-mark, and skip it if needed. */
-	char *p = this->text + (strncmp(u8"\ufeff", this->text, 3) == 0 ? 3 : 0);
-
-	/* Make sure the string is a valid UTF-8 sequence. */
-	StrMakeValidInPlace(p, this->text + filesize, SVS_REPLACE_WITH_QUESTION_MARK | SVS_ALLOW_NEWLINE);
+	if (StrStartsWith(this->text, u8"\ufeff")) this->text = this->text.erase(0, 3);
+	this->text = StrMakeValid(this->text, SVS_REPLACE_WITH_QUESTION_MARK | SVS_ALLOW_NEWLINE);
```
rubidium42: At lin 360 you start by copying the string, then at line 369 the erase essentially copies the string (to move it 3 positions forward), before copying it a third time for `StrMakeValid`.
What I suggest is moving the `free` to after the `StrMakeValid` and at line 360 create a `string_view`. At line 369 you can then `remove_prefix(3)`. The removes two copies.

src/textfile_gui.cpp
```diff
@@ -344,48 +339,48 @@ static void Xunzip(byte **bufp, size_t *sizep)
...
-	this->text[filesize] = '\0';
+	this->text = std::string(buf, filesize);
+	free(buf);
 
 	/* Replace tabs and line feeds with a space since StrMakeValidInPlace removes those. */
-	for (char *p = this->text; *p != '\0'; p++) {
-		if (*p == '\t' || *p == '\r') *p = ' ';
+	for (auto &c : this->text) {
+		if (c == '\t' || c == '\r') c = ' ';
 	}
```
rubidium42: Could be done with `std::transform`, although that might not be as clear/terse as this code.

TrueBrain: You answered yourself! :D

src/textfile_gui.cpp
TrueBrain: As mentioned on IRC, this is not as easy as you make it sound :) The replacing of `\t` and `\r` is a bit of an issue. You can't do that on a `string_view` obviously, and it needs to be done before `StrMakeValid`. So as a compromise I can remove at least one copy, by doing the BOM earlier, as you suggested :) But I do not see how to avoid 2 copies, without doing the `\t` and `\r` replacement in the C-world again (which I kinda want to avoid :D).

Ideally we had a `StrMakeValidInPlace` that works on `std::string`, but we don't (yet?).

Open to other suggestions, ofc :)

src/textfile_gui.cpp
```diff
@@ -344,48 +339,45 @@ static void Xunzip(byte **bufp, size_t *sizep)
 	FILE *handle = FioFOpenFile(textfile, "rb", dir, &filesize);
 	if (handle == nullptr) return;
 
-	this->text = ReallocT(this->text, filesize);
-	size_t read = fread(this->text, 1, filesize, handle);
+	char *buf = MallocT<char>(filesize);
+	size_t read = fread(buf, 1, filesize, handle);
 	fclose(handle);
 
 	if (read != filesize) return;
```
rubidium42: `buf` leaks here :(

rubidium42 (approved):

PR #10773, opened 2023-05-06
Fix: Engine age is in months, not days

2TallTyler:
## Motivation / Problem

Engine age is in months, but has been incorrectly stored as a Date for a long time.

## Description

Make it a generic int32, the same type as the current Date, but generic.

## Limitations

When calculating engine reliability, we use a uint32 to avoid repeated signed/unsigned comparison issues. This shouldn't be a problem as long as `engine->age` is never less than 0, which should not be possible.

## Checklist for review

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -611,8 +611,8 @@ void CalcEngineReliability(Engine *e, bool new_month)
 		re = Engine::Get(re->info.variant_id);
 	}
 
-	uint age = re->age;
-	if (new_month && re->index > e->index && age != MAX_DAY) age++; /* parent variant's age has not yet updated. */
+	uint32 age = re->age;
+	if (new_month && re->index > e->index && age != INT32_MAX) age++; /* parent variant's age has not yet updated. */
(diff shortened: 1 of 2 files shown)


src/engine.cpp
```diff
@@ -611,8 +611,8 @@ void CalcEngineReliability(Engine *e, bool new_month)
 		re = Engine::Get(re->info.variant_id);
 	}
 
-	uint age = re->age;
-	if (new_month && re->index > e->index && age != MAX_DAY) age++; /* parent variant's age has not yet updated. */
+	uint32 age = re->age;
```
rubidium42, line 614: re->age is of type `int32`, why make it `uint32` here?

2TallTyler: We compare this local variable against a variety of duration phases, which are all unsigned. This PR came about when trying to fix signed/unsigned comparison annotations in #10746, and I wanted to avoid that headache again.

rubidium42 (approved):

PR #10776, opened 2023-05-06
Codechange: C++-ify the Layouter and related functions

TrueBrain:
## Motivation / Problem

While working on another PR (survey addition) I ran into #10772, which showed me that you cannot give `string_view` to the Layouter. This PR sets out to fix that issue.

## Description

If you do the Layouter, you also have to do the functions that use the Layouter, which are mostly Draw functions. This also means we no longer need an overloaded function for both `const char *` and `std::string`, as `const char *` can safely be casted to `std::string_view` for all our use-cases.

Also made `GetLayouter` a bit more solid, by switching to `std::string_view` (instead of only a `const char *` and repeat the logic to stop at the right position). This means the constructor of `Layouter` is now fully in control over the snippet `GetLayouter` is converting, instead of having identical logic in both to ensure they are in sync.

Another quirk was found with `GetCharAtPosition`, which returns a pointer inside a string. This is a bit annoying to deal with, so instead, it now returns an index inside the string. The caller can handle it fine to put 1 and 1 back together.

As a bonus, reworked `GetCharPosInString` to be much more readable. Especially when looking for the full line-length it just wasted CPU for no good reason (anymore).

## Limitations

Any call to these Draw functions with `const char *` as parameter, which are most, will now implicitly do a `strlen` to convert to a `std::string_view`. This can cause some minor slowdown.
The biggest issue here is that `GetString` doesn't return `std::string`, causing most of the conversions. Fixing that function to be more C++ would resolve most of this issue completely.

`GetCharPosInString` still has a `const char *`, which has to be within the `std::string_view` given in the first parameter. This is not ideal, but the code using this function ... well, they deserve their own PR to clean up.

I couldn't test MacOS, which is the only consumer of `GetCharAtPosition`. I am also not sure when this function is exactly used, but will do my best to poke someone to validate the functionality :)

## Checklist for review

### DIFF
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -642,7 +642,7 @@ static int DrawLayoutLine(const ParagraphLayouter::Line &line, int y, int left, 
  * @return In case of left or center alignment the right most pixel we have drawn to.
  *         In case of right alignment the left most pixel we have drawn to.
  */
-int DrawString(int left, int right, int top, const char *str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
+int DrawString(int left, int right, int top, std::string_view str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
 {
 	/* The string may contain control chars to change the font, just use the biggest font for clipping. */
 	int max_height = std::max({FONT_HEIGHT_SMALL, FONT_HEIGHT_NORMAL, FONT_HEIGHT_LARGE, FONT_HEIGHT_MONO});
@@ -678,28 +678,6 @@ int DrawString(int left, int right, int top, const char *str, TextColour colour,
  * @return In case of left or center alignment the right most pixel we have drawn to.
  *         In case of right alignment the left most pixel we have drawn to.
  */
-int DrawString(int left, int right, int top, const std::string &str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
-{
-	return DrawString(left, right, top, str.c_str(), colour, align, underline, fontsize);
-}
-
-/**
- * Draw string, possibly truncated to make it fit in its allocated space
- *
- * @param left   The left most position to draw on.
- * @param right  The right most position to draw on.
- * @param top    The top most position to draw on.
- * @param str    String to draw.
- * @param colour Colour used for drawing the string, for details see _string_colourmap in
- *               table/palettes.h or docs/ottd-colourtext-palette.png or the enum TextColour in gfx_type.h
- * @param align  The alignment of the string when drawing left-to-right. In the
- *               case a right-to-left language is chosen this is inverted so it
- *               will be drawn in the right direction.
- * @param underline Whether to underline what has been drawn or not.
- * @param fontsize The size of the initial characters.
- * @return In case of left or center alignment the right most pixel we have drawn to.
- *         In case of right alignment the left most pixel we have drawn to.
- */
 int DrawString(int left, int right, int top, StringID str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
 {
 	char buffer[DRAW_STRING_BUFFER];
@@ -713,7 +691,7 @@ int DrawString(int left, int right, int top, StringID str, TextColour colour, St
  * @param maxw maximum string width
  * @return height of pixels of string when it is drawn
  */
-int GetStringHeight(const char *str, int maxw, FontSize fontsize)
+int GetStringHeight(std::string_view str, int maxw, FontSize fontsize)
 {
 	Layouter layout(str, maxw, TC_FROMSTRING, fontsize);
 	return layout.GetBounds().height;
@@ -765,7 +743,7 @@ Dimension GetStringMultiLineBoundingBox(StringID str, const Dimension &suggestio
  * @param suggestion Suggested bounding box.
  * @return Bounding box for the multi-line string, may be bigger than \a suggestion.
  */
-Dimension GetStringMultiLineBoundingBox(const char *str, const Dimension &suggestion)
+Dimension GetStringMultiLineBoundingBox(std::string_view str, const Dimension &suggestion)
 {
 	Dimension box = {suggestion.width, (uint)GetStringHeight(str, suggestion.width)};
 	return box;
@@ -787,7 +765,7 @@ Dimension GetStringMultiLineBoundingBox(const char *str, const Dimension &sugges
  *
  * @return If \a align is #SA_BOTTOM, the top to where we have written, else the bottom to where we have written.
  */
-int DrawStringMultiLine(int left, int right, int top, int bottom, const char *str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
+int DrawStringMultiLine(int left, int right, int top, int bottom, std::string_view str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
 {
 	int maxw = right - left + 1;
 	int maxh = bottom - top + 1;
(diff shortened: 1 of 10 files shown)


src/gfx.cpp
```diff
@@ -946,22 +887,26 @@ uint GetStringListWidth(const StringID *list, FontSize fontsize)
  * @param start_fontsize Font size to start the text with.
  * @return Upper left corner of the glyph associated with the character.
  */
-Point GetCharPosInString(const char *str, const char *ch, FontSize start_fontsize)
+Point GetCharPosInString(std::string_view str, const char *ch, FontSize start_fontsize)
 {
+	/* Ensure "ch" is inside "str" or at the exact end. */
+	assert(ch >= str.data() && (ch - str.data()) <= static_cast<int>(str.size()));
```
rubidium42: `int` -> `ptrdiff_t`?
An alternative to think about: `assert(IsInsideBS(ch, str.data(), str.size() + 1));`

src/gfx_layout.cpp
```diff
@@ -123,65 +126,55 @@ static inline void GetLayouter(Layouter::LineCacheItem &line, const char *&str,
...
-		/* Scan string for end of line */
-		const char *lineend = str;
-		for (;;) {
-			size_t len = Utf8Decode(&c, lineend);
-			if (c == '\0' || c == '\n') break;
-			lineend += len;
+	while (true) {
+		auto line_length = str.find_first_of('\n');
+		if (line_length == std::string_view::npos) {
+			line_length = str.size();
```
rubidium42: `std::string_view`'s (and `std::string`'s) `substr` function allow `::npos` as their `count` parameter, which will mean "until the end of the string". So, as I understand it you can just leave this whole if-statement out and run `substr` regardless whether `line_length` is `::npos` or not.

src/gfx_layout.cpp
```diff
@@ -191,7 +184,14 @@ Layouter::Layouter(const char *str, int maxw, TextColour colour, FontSize fontsi
 			if (l == nullptr) break;
 			this->push_back(std::move(l));
 		}
-	} while (c != '\0');
+
+		/* Go to the next line, or break out if this was the last line. */
+		if (line_length == str.size()) {
```
rubidium42: If removing the `if` because `substr` allows `::npos`, then this check needs to change to check to `::npos`.

src/gfx_layout.cpp
```diff
@@ -191,7 +184,14 @@ Layouter::Layouter(const char *str, int maxw, TextColour colour, FontSize fontsi
...
+			break;
+		} else {
+			str = str.substr(line_length + 1);
```
rubidium42: There is no need to have this in an `else` branch because the `if` already terminates execution of the loop.

src/gfx_layout.cpp
```diff
@@ -280,17 +282,18 @@ const char *Layouter::GetCharAtPosition(int x) const
 				size_t index = run.GetGlyphToCharMap()[i];
 
 				size_t cur_idx = 0;
-				for (const char *str = this->string; *str != '\0'; ) {
-					if (cur_idx == index) return str;
+				int char_index = 0;
+				for (auto str = this->string.begin(); str < this->string.end(); char_index++) {
```
rubidium42: I'm personally not really fond of comparing iterators with `<`. Primarily because the specification does not mention `<` or `>`, only `==` and `!=`. Conceptually, with reverse iterators, linked lists, deques and anything where data might not be retrieved in ascending order, you have to hope it's implemented in a manner that compares the index and not the memory location.

src/gfx.cpp
TrueBrain: ~~it would need to become `assert(IsInsideBS(ch, str.data(), str.data() + str.size() + 1));` .. it gets so hard to read :P I will just make it two asserts .. this is already annoyingly hard to read :P Stupid pointers ... :D~~

Nevermind, read the `IsInsideBS` wrong :P

Edit: `IsInsideBS` second parameter is a `size_t`, which is a bit odd. It seems that it should be `T`, but I don't feel comfortable to change that. So going with the current version. It would be a good idea to use `IsInsideBS`, but .. yeah .. not risking it :D

src/gfx_layout.cpp
TrueBrain: Cool, tnx :D

src/gfx_layout.cpp
TrueBrain: Where my scare is, that `Utf8Consume` might consume more bytes that `string` is large, and as such, it never actually hits `end()`.

On Linux, a `std::string_view::iterator` is just a plain `const char *`, and no bound-checks are done when manipulating it.

So I fully agree with you on any other type, that you should never do `< end()`. But here .. I am a bit scared.
But I guess it is unfounded scareness, as we already know the strings are valid Utf8, so it should never happen that it misses `end()`.

So okay, let's be proper, and use `!=`. But I am still a tiny bit scared :D

src/gfx_layout.cpp
```diff
@@ -123,65 +126,51 @@ static inline void GetLayouter(Layouter::LineCacheItem &line, const char *&str,
...
-			size_t len = Utf8Decode(&c, lineend);
-			if (c == '\0' || c == '\n') break;
-			lineend += len;
-		}
+	while (true) {
+		auto line_length = str.find_first_of('\n');
+		auto str_line = str.substr(0, line_length);
 
-		LineCacheItem& line = GetCachedParagraphLayout(str, lineend - str, state);
+		LineCacheItem& line = GetCachedParagraphLayout(str_line, state);
```
rubidium42:
```suggestion
		LineCacheItem &line = GetCachedParagraphLayout(str_line, state);
```
It's already wrong... but if you're changing it (and I found more tiny things...)

src/gfx_layout.cpp
```diff
@@ -210,58 +207,58 @@ Dimension Layouter::GetBounds()
 
 /**
  * Get the position of a character in the layout.
- * @param ch Character to get the position of.
+ * @param ch Character to get the position of. Must be a iterator of the string passed to the constructor.
```
rubidium42:
```suggestion
 * @param ch Character to get the position of. Must be an iterator of the string passed to the constructor.
```

src/gfx_layout.cpp
```diff
@@ -191,7 +180,15 @@ Layouter::Layouter(const char *str, int maxw, TextColour colour, FontSize fontsi
...
 		}
-	} while (c != '\0');
+
+		/* Break out if this was the last line. */
+		if (line_length == std::string_view::npos) {
+			break;
+		}
+
+		/* Go to the next line. */
+		str = str.substr(line_length + 1);
```
rubidium42: Conceptually the following might be better as it does not involve a copy. Though I reckon the compiler might be smart enough to not create the copy. On the other hand, `substr` will throw an exception when you pass it an invalid position and `remove_suffix` does not. So maybe the `substr` variant is better as that implicitly performs bound checks.
```suggestion
		str.remove_suffix(line_length + 1);
```

TrueBrain: A quick godbolt later, and it seems `remove_prefix` (prefix btw, not suffix :P) saves a few opcodes; but `string_view` is cheap, so the difference is rather low. And indeed, optimizer fixes any difference between them :) So will change, just because it looks nicer :) I have more places I can use this :)

(I thought "remove_prefix" removed a prefix-string .. didn't expect it to remove just characters :D TIL!)

rubidium42 (approved):

PR #10778, opened 2023-05-06
Change: use precompiled headers to speed up compilation

rubidium42:
## Motivation / Problem

Compilation is slow; precompiled headers make it faster.

## Description

Enable precompiled headers in cmake for `stdafx.h`, `3rdparty/fmt/format.h` and `table/strings.h`, though that required a bump of minimum cmake version.

```
real    12m2.645s
user    43m25.293s
sys     2m39.714s
```
=>
```
real    9m35.804s
user    33m50.140s
sys     2m19.130s
```

## Limitations

18 years ago it seemed to have caused problems, but compilers have probably improved quite a bit since then.

## Scope questions

Should this PR also include the common STL headers into stdafx.h?
Should stdafx.h (ancient MSVC naming) be renamed to the currently more common pch.h?

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -88,7 +88,7 @@ jobs:
         - name: GCC - Dedicated
           compiler: gcc
           cxxcompiler: g++
-          extra-cmake-parameters: -DOPTION_DEDICATED=ON -DCMAKE_CXX_FLAGS_INIT="-DRANDOM_DEBUG"
+          extra-cmake-parameters: -DOPTION_DEDICATED=ON -DCMAKE_CXX_FLAGS_INIT="-DRANDOM_DEBUG" -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON
           # Compile without SDL / SDL2, as that should compile fine too.
 
     name: Linux (${{ matrix.name }})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.16)
 
 if(NOT BINARY_NAME)
     set(BINARY_NAME openttd)
@@ -6,6 +6,7 @@ endif()
 
 project(${BINARY_NAME}
     VERSION 14.0
+    LANGUAGES CXX
 )
 
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
@@ -149,6 +150,8 @@ if(NOT OPTION_DEDICATED)
     endif()
 endif()
 if(APPLE)
+    enable_language(OBJCXX)
+
     find_package(Iconv)
 
     find_library(AUDIOTOOLBOX_LIBRARY AudioToolbox)
@@ -242,6 +245,12 @@ if(MSVC)
     # Add DPI manifest to project; other WIN32 targets get this via ottdres.rc
     target_sources(openttd PRIVATE "${CMAKE_SOURCE_DIR}/os/windows/openttd.manifest")
 endif()
+
+target_precompile_headers(openttd_lib
+    PRIVATE
+    src/stdafx.h
+    src/3rdparty/fmt/format.h
+)
 
 add_subdirectory(${CMAKE_SOURCE_DIR}/bin)
 add_subdirectory(${CMAKE_SOURCE_DIR}/src)
--- a/src/strgen/strgen.cpp
+++ b/src/strgen/strgen.cpp
@@ -17,7 +17,6 @@
 
 #include "strgen.h"
 
-#include <exception>
 
 #if !defined(_WIN32) || defined(__CYGWIN__)
 #include <unistd.h>
--- a/src/base_consist.h
+++ b/src/base_consist.h
@@ -12,7 +12,6 @@
 
 #include "order_type.h"
 #include "timer/timer_game_calendar.h"
-#include <string>
 
 /** Various front vehicle properties that are preserved when autoreplacing, using order-backup or switching front engines within a consist. */
 struct BaseConsist {
--- a/src/bitmap_type.h
+++ b/src/bitmap_type.h
@@ -10,7 +10,6 @@
 #ifndef BITMAP_TYPE_HPP
 #define BITMAP_TYPE_HPP
 
-#include <vector>
 
 /** Represents a tile area containing containing individually set tiles.
  * Each tile must be contained within the preallocated area.
--- a/src/cargotype.h
+++ b/src/cargotype.h
@@ -17,7 +17,6 @@
 #include "landscape_type.h"
 #include "core/bitmath_func.hpp"
 #include "core/span_type.hpp"
-#include <vector>
 
 /** Globally unique label of a cargo type. */
 typedef uint32 CargoLabel;
--- a/src/command_type.h
+++ b/src/command_type.h
@@ -13,7 +13,6 @@
 #include "economy_type.h"
 #include "strings_type.h"
 #include "tile_type.h"
-#include <vector>
 
 struct GRFFile;
 
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -17,7 +17,6 @@
 #include "timer/timer_game_calendar.h"
 #include "settings_type.h"
 #include "group.h"
-#include <string>
 #include <array>
 
 /** Statistics about the economy. */
--- a/src/console_gui.cpp
+++ b/src/console_gui.cpp
@@ -23,7 +23,6 @@
 #include "timer/timer.h"
 #include "timer/timer_window.h"
 #include <deque>
-#include <string>
 
 #include "widgets/console_widget.h"
 
--- a/src/core/kdtree.hpp
+++ b/src/core/kdtree.hpp
@@ -11,8 +11,6 @@
 #define KDTREE_HPP
 
 #include "../stdafx.h"
-#include <vector>
-#include <limits>
 
 /**
  * K-dimensional tree, specialised for 2-dimensional space.
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -10,8 +10,6 @@
 #ifndef MATH_FUNC_HPP
 #define MATH_FUNC_HPP
 
-#include <limits>
-#include <type_traits>
 
 /**
  * Returns the absolute value of (scalar) variable.
--- a/src/core/overflowsafe_type.hpp
+++ b/src/core/overflowsafe_type.hpp
@@ -12,7 +12,6 @@
 
 #include "math_func.hpp"
 
-#include <limits>
 
 #ifdef __has_builtin
 #	if __has_builtin(__builtin_add_overflow) && __has_builtin(__builtin_sub_overflow) && __has_builtin(__builtin_mul_overflow)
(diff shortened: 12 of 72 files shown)


src/strgen/strgen.cpp
```diff
@@ -269,6 +269,7 @@ struct HeaderFileWriter : HeaderWriter, FileWriter {
 		fprintf(this->fh, "/* This file is automatically generated. Do not modify */\n\n");
 		fprintf(this->fh, "#ifndef TABLE_STRINGS_H\n");
 		fprintf(this->fh, "#define TABLE_STRINGS_H\n");
+		fprintf(this->fh, "typedef uint32_t StringID;\n");
```
TrueBrain: Why this change? Can't find it in the description either. Feel a bit weird, in this PR?

TrueBrain: ah, found the answer myself, as otherwise you cannot precompile that header. But this feels like the wrong solution? Shouldn't we include some header that defines this type?

rubidium42: I tried header inclusion, specifically `#include "strings_type.h"` and `#include "../strings_type.h"` (because it's as if it is in the table folder for the normal compile). But both give an error that it can't find the file because it's resolving the name relative to `<user defined path>/generated/table/strings.h`. So slapping a couple of `../`s to it would also not reliably get it to the right path.
And at that point the question is... how much effort are we going to put into something relatively brittle when just adding a single typedef works just fine.

glx22: Aren't we also trying to replace `typedef` with `using` ?

TrueBrain: The manual suggest to include `strings_type.h` in the precompiled header as well, to resolve that issue.

I do not think it is advised to define `StringID` twice in our source code; that is just asking for trouble. So I agree that a `../` is not a good path, but than dismissing: so we should do a typedef like this, seems too short-sighted :) I am sure we can find something else that is elegant and sweet :)

CMakeLists.txt
```diff
@@ -243,6 +246,13 @@ if(MSVC)
     target_sources(openttd PRIVATE "${CMAKE_SOURCE_DIR}/os/windows/openttd.manifest")
 endif()
 
+target_precompile_headers(openttd_lib
+    PRIVATE
+    src/stdafx.h
+    src/3rdparty/fmt/format.h
+    "${CMAKE_BINARY_DIR}/generated/table/strings.h"
```
glx22: Maybe
```suggestion
    [["table/strings.h"]]
```
?

src/strgen/strgen.cpp
PeterN: Yeah, as glx said, `using StringID = uint32_t;` if it is needed.

TrueBrain:
One of the things we have to be mindful of, is that with precompiled headers, you don't actually have to include those files anymore. I don't know if this is actually a problem or not, but we could mitigate that by having a single CI job that validates with precompiled-headers disabled (CMake supplies for this).

On the other hand, we could go all-out here, and just remove any includes we have for anything we have in the precompiled headers.

TrueBrain:
Bit weird .. with `gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0`  I see no difference between with this PR and without this PR. With clang on the other hand, I see comparable difference as in OP. No clue why there is such a sharp contrast between those two. It also doesn't matter: gcc didn't become slower.

src/strgen/strgen.cpp
glx22: Including `strings_type.h` works for me, but I also have the CMakeLists.txt change I suggested

rubidium42: With glx's suggestion including "strings_type.h" seems to do the trick.

rubidium42:
Including all of STL that was already used in the source code in `stdafx.h` does not seem to have an appreciable effect on the total build time; just a few seconds for my test case. However, it might on others and will reduce build failures because some STL header is required for some platforms and not on another; with this it essentially has all of them all the time.
```
real    9m32.967s
user    33m47.277s
sys     2m21.270s
```

TrueBrain:
Did some benchmarking with some precompiled header-files, to get a bit of feel for things.

|   |   |
| - | - |
| No precompiled headers | 22m |
| `stdafx.h` | 19m |
| `stdafx.h` + `strings.h` | 21m |
| `stdafx.h` + `fmt.h` | 14m |
| `stdafx.h` + `fmt.h` + `strings.h` | 15m |
| All three with STL included | 14m |
| `stdafx.h` + `fmt.h` with STL | 13m |
| `stdafx.h` + `fmt.h` with STL + removed include-statements for stdafx/fmt** | 13m |

(**: as I am lazy, I also applied precompiled headers to strgen/settingsgen; that was just easier to `sed` the headers out of the files)

So my findings (and we need confirmation from others for it to be meaningful findings):

The bad:

- precompiling `strings.h` increases compile-time slightly (at least, never was faster; tried multiple times)

The good:

- precompiling `stdafx.h` is a modern decrease in compile-time
- precompiling `fmt.h` is a huge decrease in compile-time
- adding STL to stdafx is a small decrease

The pointless:

- removing the include statements from our code-base made no meaningful difference.

TrueBrain:
I redid my benchmarking above, but this time with a much more stable `perf stat`, instead of `time`. Gave much more sane numbers that were repeatable.

Based on that, I would say, let's make it ourselves easy:

- We always add STL library to `stdafx.h`. Some are just .. not used often, but this makes the rule really simple.
- We enable precompiled headers for `fmt.h` and `stdafx.h`.
- But not for `strings.h`. It is a bit of an odd duck: it is not really what is normally considered a precompiled header, it doesn't actually decrease compile-times all that much, and I am a bit afraid that it will give issues really quick, as it is one of those files that is more likely to change than the other two. `fmt.h` and `stdafx.h` are rare events. `strings.h` changes every time we add a string.

I have no opinion about stdafx.h and pch.h.

glx22:
Did some measuring with MSVC (using NINJA_STATUS env var to show elapsed time)
master with cl:
```
  [796/797 -- 402.064]Linking CXX executable openttd.exe
  [797/797 -- 412.391]Linking CXX executable openttd_test.exe
```
master with clang-cl:
```
  [795/797 -- 557.915]Linking CXX executable openttd.exe
  [796/797 -- 558.515]Building CXX object CMakeFiles\openttd_test.dir\src\tests\test_main.cpp.obj
  [797/797 -- 569.596]Linking CXX executable openttd_test.exe
```
PR with cl:
```
  [800/801 -- 270.160]Linking CXX executable openttd.exe
  [801/801 -- 279.215]Linking CXX executable openttd_test.exe
```
PR with clang-cl:
```
  [800/801 -- 373.914]Linking CXX executable openttd.exe
  [801/801 -- 377.432]Linking CXX executable openttd_test.exe
```

rubidium42:
I've also run it without the `table/strings.h` change:
```
real    9m24.805s
user    33m29.428s
sys     2m15.345s
```

With some grains of salt because I'm looking at full compilation time, compared to the number of operations TrueBrain researched. Furthermore this is on 10 year old hardware.
| Combination | Instructions | Delta |
| - | - | - |
| No precompiled headers | 722s | 100% |
| `stdafx.h` + `fmt.h` with STL | 564s | 78.1% |
| `stdafx.h` + `fmt.h` + `strings.h` | 575s | 79.6% |
| `stdafx.h` + `fmt.h` + `strings.h` with STL | 572s | 79.2% |

So it looks like `stdafx.h` + `fmt.h` with STL is fastest for me, although by a relatively small margin. So it is clear that both 'stdafx.h' and 'fmt.h' should be in the mix, and the rest is less clear.
A reason why adding STL does not have a huge impact might be because `fmt.h` already includes a huge amount, such as algorithm, cmath, cstdint, cstring, cstdio, exception, functional, iterator, memory, string, type_traits, vector and string_view. So adding them to `stdafx.h` does not necessarily add a lot of improvement as `fmt.h` already took care of them. However, putting them in `stdafx.h` and including that everywhere will mean it is quite likely that the code will keep compiling without precompiled headers... though with a significantly reduced performance.

Removing the STL includes from the other files should indeed not have much of a benefit for compilation, as it would see that it already loaded it and skip all the complicated parsing associated with the file. However, why keep includes around that are not strictly needed anymore?

So the main question is: how much do we want to compromise non-precompiled header builds? I'd say not that much, in which case precompiling `stdafx.h` and `fmt.h` without messing with STL manually will give us likely the best of both world when speaking about performance, with the occasional compilation failure when precompiled headers are turned off.

.github/workflows/ci-build.yml
```diff
@@ -256,10 +256,23 @@ jobs:
...
+        - name: Windows (windows-2019 / x86)
+          arch: x86
+          os: windows-latest
+        - name: Windows (windows-2019 / x64)
+          arch: x64
+          os: windows-latest
+        - name: Windows (windows-latest / x86) without precompiled headers
+          arch: x64
+          os: windows-latest
+          extra-cmake-parameters: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=true
```
TrueBrain: Normally it is `ON` for CMake to enable things. (and not `true`).

.github/workflows/ci-build.yml
```diff
@@ -256,10 +256,25 @@ jobs:
...
+        - name: windows-latest / x86
+          arch: x86
+          os: windows-latest
+        - name: windows-latest / x64
+          arch: x64
+          os: windows-latest
+        - name: windows-2019 / x86
+          arch: x86
+          os: windows-2019
+        - name: windows-2019 / x64
```
TrueBrain: personally, I wouldn't do this `name` entry. Without it, and having the `name: Windows (..)` as it was, saves a field we can make mistakes with.

TrueBrain: owh, no, I see why you did this. Maybe we are better off adding this to the GCC Dedicated run. It is already a special run to check these kinds of things, and saves us from this trouble?

rubidium42: Though how would you see the difference between the one with and the one without precompiled headers if it's only showing arch and OS?

TrueBrain:
I think we should at least include the STL headers in stdafx that are also in fmt. Otherwise we will constantly have PRs that break on the "no precompiled headers" runner. My reasoning:

- I start using `std::vector` in a file that doesn't include it
- I have no compile errors, as they are included via precompiled
- The file doesn't include fmt

This only breaks on CI. I would find that annoying. But I might be overthinking this :D

.github/workflows/ci-build.yml
rubidium42: > owh, no, I see why you did this. Maybe we are better off adding this to the GCC Dedicated run. It is already a special run to check these kinds of things, and saves us from this trouble?

The reason I choose MSVC is because GCC, and Clang on Linux, will happily accept it in our current codebase to not include `optional` when using `std::optional`, whereas MSVC just fails to build. That's the whole reason I went for MSVC, and the x86 variant is because that's most picky about `size_t` to `int` conversions.

Though indeed making GCC Dedicate the one without precompiled headers is the easiest, but will it catch most of the issues? That's the conundrum I've been thinking about.

TrueBrain: I could argue that nobody is going to run MSVC without precompiled headers (as it is on by default on all new MSVC projects anyway), while I am sure there will be some Linux variants where it doesn't work (looking at you GCC 11) :P

I would go for our sanity, and keep the CI file as simple as we can .. it is already scary complex :)

TrueBrain (approved):

PR #10780, opened 2023-05-07
Cleanup: Remove unnecessary hangar check.

J0anJosep:
## Motivation

Cleanup code. Removing a check that doesn't belong to the case and is performed anyway later.
In case of doubt, just read line 1593. The rest of the code right now doesn't use this boolean.


2TallTyler (approved):
Yep, makes sense to me. The check is repeated three lines later and handles it better there.

PR #10781, opened 2023-05-07
Codechange: Rely on Scrollbar::SetPosition to clamp.

PeterN:
## Motivation / Problem

Manually clamping scrollbar bounds before calling `SetPosition()` is doubling up work that the function already does.

## Description

Remove the clamping and let SetPosition() handle it.
Also uses return value of SetPosition() instead of testing whether the position should be changed.
Also changes the script debug autoscroll condition to  make it clearer, i.e. "last row >= size". Probably separate really but it was right there and kinda related...

## Checklist for review

### DIFF
--- a/src/script/script_gui.cpp
+++ b/src/script/script_gui.cpp
@@ -860,11 +860,10 @@ struct ScriptDebugWindow : public Window {
 		/* Detect when the user scrolls the window. Enable autoscroll when the
 		 * bottom-most line becomes visible. */
 		if (this->last_vscroll_pos != this->vscroll->GetPosition()) {
-			this->autoscroll = this->vscroll->GetPosition() >= log.size() - this->vscroll->GetCapacity();
+			this->autoscroll = this->vscroll->GetPosition() + this->vscroll->GetCapacity() >= (int)log.size();
 		}
 		if (this->autoscroll) {
-			int scroll_pos = std::max<int>(0, (int)log.size() - this->vscroll->GetCapacity());
-			if (this->vscroll->SetPosition(scroll_pos)) {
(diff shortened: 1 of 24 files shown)


stormcone:
The first commit says: Scrollbar methods new accept size_t.
Maybe: new = now? :)

PeterN:
Argh! I'll wait to see if MinGW likes this one first :)

PeterN:
Fixed commit typo, not that matters if squashed.

src/script/script_gui.cpp
```diff
@@ -860,11 +860,10 @@ struct ScriptDebugWindow : public Window {
 		/* Detect when the user scrolls the window. Enable autoscroll when the
 		 * bottom-most line becomes visible. */
 		if (this->last_vscroll_pos != this->vscroll->GetPosition()) {
-			this->autoscroll = this->vscroll->GetPosition() >= log.size() - this->vscroll->GetCapacity();
+			this->autoscroll = this->vscroll->GetPosition() + this->vscroll->GetCapacity() >= (int)log.size();
 		}
 		if (this->autoscroll) {
-			int scroll_pos = std::max<int>(0, (int)log.size() - this->vscroll->GetCapacity());
-			if (this->vscroll->SetPosition(scroll_pos)) {
+			if (this->vscroll->SetPosition((int)log.size())) {
```
rubidium42, line 866: Maybe `SetPosition` should also accept `size_t` like the other ones, as done in the first commit?

rubidium42 (approved):

src/script/script_gui.cpp
PeterN: I thought about that, but that then means the automatic range clamping won't work properly for negative values.

PR #10783, opened 2023-05-07
Change: Harmonized tree placement window with other construction windows

Kuhnovic:
## Motivation / Problem

I've always found the tree placement window odd looking compared to other construction-related windows. This is an attempt to fix that.

## Description

I know that the previous "window with separate gray buttons" style is one that's being used in multiple other places (Depot orientation, rail station creation, to name a few). I just don't think it's a good fit for this particular window.

Here's a before and after:
image

And this is what it looks like in the Scenario Editor:

image

## Limitations

The GUI style is quite inconsistent, some window use this "tightly connected" style where other windows use the aforementioned "window with separate gray buttons" style. This change doesn't really fix that, maybe I'm just adding fuel to the fire.

## Checklist for review

### DIFF
--- a/src/tree_gui.cpp
+++ b/src/tree_gui.cpp
@@ -269,15 +269,13 @@ static NWidgetBase *MakeTreeTypeButtons(int *biggest_index)
 	byte cur_type = type_base;
 
 	NWidgetVertical *vstack = new NWidgetVertical(NC_EQUALSIZE);
-	vstack->SetPIP(0, 1, 0);
 
 	for (int row = 0; row < num_rows; row++) {
 		NWidgetHorizontal *hstack = new NWidgetHorizontal(NC_EQUALSIZE);
-		hstack->SetPIP(0, 1, 0);
 		vstack->Add(hstack);
 		for (int col = 0; col < num_columns; col++) {
 			if (cur_type > type_base + type_count) break;
-			NWidgetBackground *button = new NWidgetBackground(WWT_PANEL, COLOUR_GREY, WID_BT_TYPE_BUTTON_FIRST + cur_type);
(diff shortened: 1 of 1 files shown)


PeterN:
But the current window 'fits' with other windows where you get to chose type

image

2TallTyler (dismissed):
Our GUI style is all over the place, but I see no reason why we can't try to improve it. I don't imagine anyone is brave/crazy enough to take on a comprehensive, top-down approach, so why not do it piecemeal? It can't get _worse_. 😛

PeterN:
Also green background for choosing something which is predominantly green seems like a bad idea to me.

2TallTyler:
Trees will generally have a green background in-game too, typically placed on grass (at least in temperate/toyland).

PeterN:
Also, this is a classic TTD window :-)

image

Kuhnovic:
Haha I knew this was going to cause some controversy :P 

@PeterN Not all "choose a type" windows use the gray buttons, for example:
image

It just goes to show how inconsistent things currently are

PeterN:
Yup, that one was added later, not a TTD original.

Kuhnovic:
image
Just quickly changed the Depot selection window (just locally, I did not commit any code). IMO it looks much cleaner. Other windows will not be as easy to adapt to this style though, especially the create railway station window.

For this PR i'll stick to the Tree selection window, don't want to blow up the scope too much.

TrueBrain:
Meh; bikeshedding, always difficult, as 10 people 11 opinions :)

To just make a call here, I think we should keep it grey. For two reasons:

1) the argument: green trees on a green background, is a very strong one for me. It just makes it harder to see and understand.
2) in general, I think having a more neutral colour for these things is very beneficial. And grey is more neutral than green.

So all in all, I think we should stick with the current colour. I do appreciate the attempt to harmonize though :) If you find other places, please do feel the waters :) Cheers!

PR #10788, opened 2023-05-08
Allow control dedicated server refresh rate.

axet:
## Motivation / Problem

Dedicated servers on old hardware can not run at full speed 30ms / 33.33 fps. It eats 100% for a process.

For example Raspberry PI 1 B, can run at maximum 0.55x game speed which equivalent of 16 simulation rate fps.

## Description

Limiting simulation fps to 15 keep sever load under 80% and keep game speed constant. Feels better for a client.

Helps a lot on low end hardware.

## Limitations

This is optional feature not affecting current server state. Since refresh_rate set to default value of 60, will change nothing. Dedicated server will be affected only by lowering refresh_rate value.

## Checklist for review

### DIFF
--- a/src/video/dedicated_v.h
+++ b/src/video/dedicated_v.h
@@ -28,6 +28,19 @@ public:
 	bool ToggleFullscreen(bool fullscreen) override;
 	const char *GetName() const override { return "dedicated"; }
 	bool HasGUI() const override { return false; }
+
+	std::chrono::steady_clock::duration GetGameInterval() override
+	{
+		/* Infinite speed, as quickly as you can. */
+		if (_game_speed == 0) return std::chrono::microseconds(0);
+
+		long min = MILLISECONDS_PER_TICK * 1000;
+		long ms = 1000000 / _settings_client.gui.refresh_rate;
+		if (ms < min)
+			ms = min;
+
+		return std::chrono::microseconds(ms * 100 / _game_speed);
+	}
 };
 
 /** Factory for the dedicated server video driver. */
--- a/src/video/video_driver.hpp
+++ b/src/video/video_driver.hpp
@@ -307,7 +307,7 @@ protected:
 	 */
 	void SleepTillNextTick();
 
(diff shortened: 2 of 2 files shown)


src/video/dedicated_v.h
```diff
@@ -28,6 +28,19 @@ class VideoDriver_Dedicated : public VideoDriver {
...
 	const char *GetName() const override { return "dedicated"; }
 	bool HasGUI() const override { return false; }
+
+	std::chrono::steady_clock::duration GetGameInterval() override
+	{
+		/* Infinite speed, as quickly as you can. */
+		if (_game_speed == 0) return std::chrono::microseconds(0);
+
+		long min = MILLISECONDS_PER_TICK * 1000;
+		long ms = 1000000 / _settings_client.gui.refresh_rate;
```
TrueBrain, line 38: You seem to mix up two concepts of the game. We have two threads: draw-thread and game-thread. The first is limited by `refresh_rate`, and often expressed in Hz. Like 60Hz for most people these days. The second is limited by `MILLISECONDS_PER_TICK`, which is the same for every game and (currently) not changable. This is set to 27ms per tick.

This change mixes up those two concepts. Please don't do that.

TrueBrain (commented):
I think this PR is a bad idea; it solves a person itch on a very local level. And this is ignoring that you mix up drawing vs game.

What you basically want is that you can alter `_game_speed` in multiplayer. But this isn't as simple as it might seem. Currently we left that disabled, as there are a few things to figure out and consider first. For example: if you can slow it down, can you also speed it up? And what do we do with lag-detection in that case? (as it is currently based on ticks).

All this is being worked on via various of other PRs, and will take a while longer till we get there. But this is not one of those steps.

So sorry, going to close this PR. Feel free to dive into speeding up / slowing down `_game_speed`, and see what is involved in getting that functional. It mostly means more splitting out of timers to make sure they don't assume a game-tick is 27ms. As a lot of code currently does.

PR #10793, opened 2023-05-08
Fix: Support more than 256 stations/waypoints/roadstops per class.

PeterN:
## Motivation / Problem

It was already possible to define more than 256 per class, but not possible to use them as the index used in GUI and passed through commands was limited to a byte.

Objects are handled differently, by passing a global instead of class & class index, so do not have the same issue.

Loading lots of NewGRFs could cause more than 256 specs to be defined per class. This would then wrap around in the UI and now allow use of all specs.

468 waypoints are loaded, but only 256 are visible.

Screenshot 2023-05-08 181227

## Description

Change the index to be uint16_t instead of a byte. While 65536 may be reachable it's not a problematic limit currently.

Same set of NewGRFs are loading which gives 468 waypoints:

Screenshot 2023-05-08 180950

## Checklist for review

### DIFF
--- a/src/rail_gui.cpp
+++ b/src/rail_gui.cpp
@@ -53,7 +53,7 @@
 static RailType _cur_railtype;               ///< Rail type of the current build-rail toolbar.
 static bool _remove_button_clicked;          ///< Flag whether 'remove' toggle-button is currently enabled
 static DiagDirection _build_depot_direction; ///< Currently selected depot direction
-static uint16_t _waypoint_count = 1;             ///< Number of waypoint types
+static uint16_t _waypoint_count = 1;         ///< Number of waypoint types
 static uint16_t _cur_waypoint_type;          ///< Currently selected waypoint type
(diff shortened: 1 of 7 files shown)


src/rail_gui.cpp
```diff
@@ -53,8 +53,8 @@
 static RailType _cur_railtype;               ///< Rail type of the current build-rail toolbar.
 static bool _remove_button_clicked;          ///< Flag whether 'remove' toggle-button is currently enabled
 static DiagDirection _build_depot_direction; ///< Currently selected depot direction
-static byte _waypoint_count = 1;             ///< Number of waypoint types
-static byte _cur_waypoint_type;              ///< Currently selected waypoint type
+static uint16_t _waypoint_count = 1;             ///< Number of waypoint types
```
glx22: As you fixed comment alignment for next line, that's inconsistent ;)

rubidium42 (approved):

PR #10794, opened 2023-05-08
Fix: Incorrect y-position of monospace glyphs.

PeterN:
## Motivation / Problem

OpenTTD sprite font renderer expects glyphs to be a specific position within the individual sprites.

The current alignment causes glyphs to be cropped below the baseline.

## Description

Fixed by change the sprite offset to point to the (normal) top row of pixels, as sprite font renderer expects.

This may cause accented characters on the first row of text to be cropped, but they are less likely to be rendered.

## Limitations

This does not change monospace fonts with OpenGFX in use.

An alternative is to set the monospace size to 12 pixels instead of 10, but that causes a different set of issues.

## Checklist for review


TrueBrain (approved):
I am sure you know what you are doing :P

PeterN:
It's quite common in things in that from OpenGFX that dimensions and offsets don't exactly fit the original. In this case the monospace glyphs are entirely ours anyway, but still.

We could special-case this in the game to do something different for monospace (which isn't needed for truetype fonts) or make the glyphs offsets match a bit better.

PR #10782, opened 2023-05-07
Add: NewGRF string code "9A 21" to display force from textstack.

PeterN:
## Motivation / Problem

Cannot display an additional force value in extra engine text, because there's no NewGRF string code for it.

## Description

Add "9A 21" which takes an unsigned <s>word (2 bytes)</s> <b>dword (4 bytes)</b> value from the textstack and formats as the user's force units.

## Limitations

Not tested.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10795, opened 2023-05-08
Change: Replace multiply/shift with single factor for units conversion.

PeterN:
## Motivation / Problem

Working out a good multiply and shift combination for unit conversion is awkward, has the potential to overflow, and results in inaccuracies (although they are at least deterministic.)

## Description

Instead, switch all conversions to use one double-precision factor instead. These conversions are only used for display purposes so the cross-platform issues with floats is not a problem here.

This change also allows use of decimals in any unit conversion (although doesn't actually make use of this), so that the code is the same for each unit being formatted.

Lastly, when formatting {FORCE}, the internal Newton value was preconverted to kN by dividing by 1000, which lost precision for other units. This has been removed, leaving the conversion solely to the units system.

## Limitations

Some of the conversions are replaced as in by performing the multiply and shift, others are replaced with closer real-world conversion.

Therefore it's possible that some values of some units don't result in the same value as before. In particular the power-to-weight conversions were previously a bit vague.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -658,8 +658,7 @@ static const char *ParseStringChoice(const char *b, uint form, char **dst, const
 
 /** Helper for unit conversion. */
 struct UnitConversion {
-	int multiplier; ///< Amount to multiply upon conversion.
-	int shift;      ///< Amount to shift upon conversion.
+	double factor; ///< Amount to multiply or divide upon conversion.
 
 	/**
 	 * Convert value from OpenTTD's internal unit into the displayed value.
@@ -669,7 +668,9 @@ struct UnitConversion {
 	 */
 	int64 ToDisplay(int64 input, bool round = true) const
 	{
-		return ((input * this->multiplier) + (round && this->shift != 0 ? 1 << (this->shift - 1) : 0)) >> this->shift;
+		return round
+			? (int64_t)std::round(input * this->factor)
+			: (int64_t)(input * this->factor);
 	}
 
 	/**
@@ -681,7 +682,9 @@ struct UnitConversion {
 	 */
 	int64 FromDisplay(int64 input, bool round = true, int64 divider = 1) const
 	{
-		return ((input << this->shift) + (round ? (this->multiplier * divider) - 1 : 0)) / (this->multiplier * divider);
+		return round
+			? (int64_t)std::round(input / this->factor / divider)
+			: (int64_t)(input / this->factor / divider);
 	}
 };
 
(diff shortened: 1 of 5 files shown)


PeterN:
The switch from passing kN to N to {FORCE} affects #10782 -- currently that PR doesn't specify which units to use, so it would be nice to say for sure which way that should go.

src/strings.cpp
```diff
@@ -697,63 +700,64 @@ struct UnitsLong {
...
 
 /** Unit conversions for velocity. */
 static const Units _units_velocity[] = {
-	{ {       1,  0}, STR_UNITS_VELOCITY_IMPERIAL,     0 },
-	{ {     103,  6}, STR_UNITS_VELOCITY_METRIC,       0 },
-	{ {    1831, 12}, STR_UNITS_VELOCITY_SI,           0 },
-	{ {   37888, 16}, STR_UNITS_VELOCITY_GAMEUNITS,    1 },
-	{ { 7289499, 23}, STR_UNITS_VELOCITY_KNOTS,        0 },
+	{ { 1.0      }, STR_UNITS_VELOCITY_IMPERIAL,  0 },
+	{ { 1.609344 }, STR_UNITS_VELOCITY_METRIC,    0 },
```
rubidium42, line 709: This is not going to cause the issue with the SH125 (?) or something not going 200 km/h? Or whatever the issue used to be when someone messed with the (displayed) speeds a long time ago?

PeterN: Shows 201 km/h in both 13.1 and with this PR.

I think the issue you refer to was lower speed NewGRFs... possibly around 50 to 60km/h.

I've checked with DBSetXL and everything seems to be the same.

src/strings.cpp
```diff
@@ -697,63 +700,64 @@ struct UnitsLong {
...
 };
 
 /** Unit conversions for volume. */
 static const UnitsLong _units_volume[] = {
-	{ {4227,  4}, STR_UNITS_VOLUME_SHORT_IMPERIAL, STR_UNITS_VOLUME_LONG_IMPERIAL },
-	{ {1000,  0}, STR_UNITS_VOLUME_SHORT_METRIC,   STR_UNITS_VOLUME_LONG_METRIC   },
-	{ {   1,  0}, STR_UNITS_VOLUME_SHORT_SI,       STR_UNITS_VOLUME_LONG_SI       },
+	{ {  264.172 }, STR_UNITS_VOLUME_SHORT_IMPERIAL, STR_UNITS_VOLUME_LONG_IMPERIAL, 0 },
+	{ { 1000.0   }, STR_UNITS_VOLUME_SHORT_METRIC,   STR_UNITS_VOLUME_LONG_METRIC  , 0 },
+	{ {    1.0   }, STR_UNITS_VOLUME_SHORT_SI,       STR_UNITS_VOLUME_LONG_SI      , 0 },
```
rubidium42: Bit inconsistent with the comma and the spaces compared to the rest of the change.

rubidium42 (dismissed):

rubidium42 (approved):

PR #10799, opened 2023-05-09
Change: Try to better preserve train orders/info when re-arranging vehicle chains.

michicc:
## Motivation / Problem

When a train is composed of e.g. two engines (doubled-headed) and the first engine is deleted/moved away, we currently throw away the consist information (unit number, order list etc), even though the resulting chain is still a fully valid train.
Player experience could be improved when re-arranging trains by preserving as much as possible.

## Description

Transfer consist information from the old front engine to the new front engine by copying statistics and order list.

## Limitations

It is a change in behaviour.

## Checklist for review

### DIFF
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -1322,8 +1322,14 @@ CommandCost CmdMoveRailVehicle(DoCommandFlag flags, VehicleID src_veh, VehicleID
 			DeleteNewGRFInspectWindow(GSF_TRAINS, src->index);
 			SetWindowDirty(WC_COMPANY, _current_company);
 
-			/* Delete orders, group stuff and the unit number as we're not the
-			 * front of any vehicle anymore. */
+			if (src_head != nullptr && src_head->IsFrontEngine()) {
+				/* Cases #?b: Transfer order, unit number and other stuff
+				 * to the new front engine. */
+				src_head->orders = src->orders;
+				if (src_head->orders != nullptr) src_head->AddToShared(src);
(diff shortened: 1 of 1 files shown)


PeterN:
Bound to be a spacebar-heating effect here, but preserving the info seems correct to me.

PeterN (approved):

src/train_cmd.cpp
```diff
@@ -1322,10 +1322,15 @@ CommandCost CmdMoveRailVehicle(DoCommandFlag flags, VehicleID src_veh, VehicleID
...
+			if (src_head != nullptr && src_head->IsFrontEngine()) {
+				/* Cases #?b: Transfer order, unit number and other stuff
+				 * to the new front engine. */
+				src_head->orders = src->orders;
+				if (src_head->orders != nullptr) src_head->AddToShared(src);
+				src_head->CopyVehicleConfigAndStatistics(src);
+			}
+			/* Remove stuff not valid anymore for non-front engines. */
 			DeleteVehicleOrders(src);
-			RemoveVehicleFromGroup(src);
```
glx22, line 1328: I know it's merged but I don't think `group_id` is copied (at least `CopyVehicleConfigAndStatistics()` doesn't seem to copy it) so this might be a problem

michicc: Look at the first line of the removed function. It is a no-op for a non-front vehicle, and we are in the case that src is not a front vehicle anymore.

andythenorth:
This would be a small but much appreciated QoL thing :)

(I use a lot of double headed trains)

PR #10801, opened 2023-05-10
Codechange: rework Gamelog changes from union to classes

rubidium42:
## Motivation / Problem

Usage of manually managed memory within Gamelog, and the last user of `SLE_STR(`(i.e. non std::string text in the savegame).

## Description

Change the union in gamelog to use a class with a lot of sub classes, like for vehicles, and therefor `std::unique_ptr`s.
Rewrite saveload to use these classes, especially because the element name for the union is not needed anymore. I have chosen to retain the old names in the savegame for compatibility reasons.
Replace `uint8 text` and `char *text` with `std::string`.

## Limitations

Sadly an atomic change as due to the classes the internal names of field change, and it's every inefficient to do the conversion per sub type.

## Checklist for review

### DIFF
--- a/src/gamelog.cpp
+++ b/src/gamelog.cpp
@@ -32,11 +32,6 @@ extern byte   _sl_minor_version;    ///< the minor savegame version, DO NOT USE!
 
 Gamelog _gamelog; ///< Gamelog instance
 
-LoggedChange::~LoggedChange()
-{
-	if (this->ct == GLCT_SETTING) free(this->setting.name);
-}
-
 Gamelog::Gamelog()
 {
 	this->data = std::make_unique<GamelogInternalData>();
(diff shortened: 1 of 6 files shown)


src/gamelog.cpp
```diff
@@ -354,25 +327,20 @@ void Gamelog::PrintDebug(int level)
...
 		this->current_action = &this->data->action.emplace_back();
 		this->current_action->at = this->action_type;
 		this->current_action->tick = TimerGameTick::counter;
 	}
 
-	LoggedChange *lc = &this->current_action->change.emplace_back();
-	lc->ct = ct;
-
-	return lc;
+	this->current_action->change.emplace_back().swap(change);
```
PeterN: Is this `emplace_back().swap()` a standard pattern?

src/saveload/gamelog_sl.cpp
```diff
@@ -301,22 +334,25 @@ class SlGamelogAction : public DefaultSaveLoadHandler<SlGamelogAction, LoggedAct
 		SlSetStructListLength(la->change.size());
 
 		for (auto &lc : la->change) {
-			assert((uint)lc.ct < GLCT_END);
-			SlObject(&lc, this->GetDescription());
+			assert((uint)lc->ct < GLCT_END);
```
PeterN: Is this cast still needed given both sides use the same type?

PeterN (commented):
Hopefully my earlier work helped rather than hindered...

src/gamelog.cpp
rubidium42: No, I was having trouble figuring out how to do it right and thought... I'll fix this later. Well, later it was... a bit too late but still.

PeterN:
Hmm, use uint32 or uint32_t, etc?

PeterN (approved):

PR #10804, opened 2023-05-10
Fix #10467: [YAPF] Trains do not use all available tracks

### LINKED ISSUE #10467: [Bug]: Trains do not use all available tracks
### Version of OpenTTD

Flatpak 13.0

### Expected result

Trains use all available "waiting tracks".

### Actual result

Trains only use some "waiting tracks". Instead trains stop on the entrance of the yard.
train-PF-test
train-PF-test2

### Steps to reproduce

1. load one of the savegames and unpause
2. wait until trains have left the depot.
3. observe only some tracks being used although more are free.
[train-PF-issue.zip](https://github.com/OpenTTD/OpenTTD/files/10713497/train-PF-issue.zip)

mrmbernardi:
## Motivation / Problem
 

See #10467 and #9609

There is an issue in the pathfinder regarding the cost associated with blocking trains. The cost can be so low that trains may not use all the bays in a waiting bay. This happens because the path through an occupied bay can have a lower cost than the path through free bays if the occupying train is short enough and the free bays have higher penalties due to additional corners. This can happen in common designs (see #10467). The end result is that trains will wait before the bays and cause blockages.

#9609 is similar, the train refuses to consider driving through the station as it sees a lower cost path trying to drive through the blocking train and a deadlock arises. 

## Description

I have introduced an 'impatience' factor to the cost of reserved tiles. This factor is proportional to how long the train has been waiting at the signal. The longer it waits, the more it considers further paths around the blocking train. This resolves the deadlock in the first save game of the issue and the deadlock pictured in the second issue.

## Limitations

- The parameters of the impatience factor may need to be tuned.  
- The cost imposed by the impatience factor is proportional to how many reserved tiles the path crosses through. This doesn't really make sense. If a block section ahead is blocked, a train should not tolerate a longer detour just because the block is occupied by a longer train. However this is a limitation with the original YAPF algorithm and perhaps should be dealt with in a separate issue. (I have a fix for it here: PR #10807)

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_costrail.hpp
+++ b/src/pathfinder/yapf/yapf_costrail.hpp
@@ -156,12 +156,18 @@ public:
 		if (n.m_num_signals_passed >= m_sig_look_ahead_costs.size() / 2) return 0;
 		if (!IsPbsSignal(n.m_last_signal_type)) return 0;
 
+		const Train *v = Yapf().GetVehicle();
+		assert(v != nullptr);
+		assert(v->type == VEH_TRAIN);
+		/* the longer a train waits, the more it will consider longer routes around blocking trains */
+		int impatience = std::max(0, v->wait_counter - DAY_TICKS / 2);
+
 		if (IsRailStationTile(tile) && IsAnyStationTileReserved(tile, trackdir, skipped)) {
-			return Yapf().PfGetSettings().rail_pbs_station_penalty * (skipped + 1);
+			return Yapf().PfGetSettings().rail_pbs_station_penalty * (skipped + 1) + impatience;
 		} else if (TrackOverlapsTracks(GetReservedTrackbits(tile), TrackdirToTrack(trackdir))) {
 			int cost = Yapf().PfGetSettings().rail_pbs_cross_penalty;
 			if (!IsDiagonalTrackdir(trackdir)) cost = (cost * YAPF_TILE_CORNER_LENGTH) / YAPF_TILE_LENGTH;
-			return cost * (skipped + 1);
+			return cost * (skipped + 1) + impatience;
 		}
 		return 0;
 	}


Eddi-z:
I don't really like the wait time approach, because even if it prevents this deadlock, the situation will probably repeat over and over again, leading to long wait times every time. The underlying issue will be hidden and lead to a terribly inefficent network.

Eddi-z:
Taking inspiration from a previous discussion about the pathfinder penalty, we should maybe introduce a logical setting:

"Optimize dynamic track allocation for: [short trains|long trains]"

mrmbernardi:
> Taking inspiration from a previous discussion about the pathfinder penalties, we should maybe introduce a logical setting:
> 
> "Optimize dynamic track allocation for: {short trains|long trains}"

The dichotomy behind short trains and long trains doesn't make much sense to me personally. It doesn't matter how long a train is, you can't drive through it either way. Also reserved tiles are what's actually counted rather than the actual train length, so short trains at the beginning of a big block will be equivalent to long trains. 

I think the real solution is this: https://github.com/OpenTTD/OpenTTD/pull/10807

Eddi-z:
here are a few random thoughts i had (not containing any judgement about whether they are good)

 * we already have an "impatience" setting: trains reversing after waiting too long
 * other things the pathfinder is not good at:
   * avoiding broken down trains
   * differentiating between crossing paths or following other trains
   * deciding between two equal paths where one of them might need to be reserved in the near future

about the train length thing: most commonly (over the past 10-ish years) the issue has come up because people tried a thing with 1 tile trains. when a train waits at a signal, it only reserves the tiles it's currently occupying, which obviously depends on the train length, and with 1-tile trains this usually means the reservation penalty doesn't outweigh even the tiniest detour.

mrmbernardi:
I have rebased the branch, however I still think https://github.com/OpenTTD/OpenTTD/pull/10807 is the better fix for this issue.

zakimaksyutov:
I just bumped into similar (same) issue - on a high-throughput loading bay trains sometimes stop while there are still empty tracks available. If it is the same, then it is happening with long (7-tile) trains. I haven't looked at implementation but hope that the fix will address long trains as well.

https://gaming.stackexchange.com/questions/404258/debugging-high-throughput-station-why-does-a-train-stop-when-there-is-a-path

(have a save at exactly this moment as well)

zakimaksyutov:
I looked at the implementation (thank you for PR!). I don't think it will fix my issue. The goal for a train not to stop in the first place.

It does fix the deadlock issue though.

PR #10810, opened 2023-05-12
Codechange: use StringBuilder to create town names

rubidium42:
## Motivation / Problem

Another step in the ongoing slow migration to C++ .

## Description

Use the `StringBuilder` instead of the `strecpy`-esque parameters to create the town names.
Not sure this is the way to go, or whether they should just create the town names to their own strings, and then including the strings later.

## Limitations

Changing `StringBuilder` over to actually become a `std::string` is beyond the scope of this, as that requires replacing all the `GetString(buf, <StringID>, last)` calls with the `std::string` returning `GetString` version.
Adds quite a bit of cruft to force the town name building into the `StringBuilder`. Are there any good ideas how to make it better?

## Checklist for review

### DIFF
--- a/src/strings_internal.h
+++ b/src/strings_internal.h
@@ -20,7 +20,8 @@
  * extra functions to ease the migration from char buffers to std::string.
  */
 class StringBuilder {
-	char **current; ///< The current location to add strings
+	char **current; ///< The current location to add strings.
+	char *start;   ///< The begin of the string.
 	const char *last; ///< The last element of the buffer.
 
 public:
@@ -36,7 +37,7 @@ public:
 	 * @param start The start location to write to.
 	 * @param last  The last location to write to.
 	 */
-	StringBuilder(char **start, const char *last) : current(start), last(last) {}
+	StringBuilder(char **start, const char *last) : current(start), start(*start), last(last) {}
 
 	/* Required operators for this to be an output_iterator; mimics std::back_insert_iterator, which has no-ops. */
 	StringBuilder &operator++() { return *this; }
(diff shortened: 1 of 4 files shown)


PeterN:
Is the NewGRF feature to remove characters actually used?

rubidium42:
> Is the NewGRF feature to remove characters actually used?

I'm not sure. The same functionality that removes characters from the end is reused in the Czech town name generator though.

TrueBrain:
#10884 is pending, which would mean this PR doesn't need the helper?

rubidium42:
> #10884 is pending, which would mean this PR doesn't need the helper?

Sadly not, as the Czech town name generation uses the remove some characters from the back functionality.
Furthermore without the helper everything needs to be done in one big commit, and without the helper we would go to `std::back_insert_iterator` and that makes adding strings harder, as the back insert iterator works character-for-character. But you can't just define it as `+= other string`.

Also, with this PR we are still using a char buffer for the string building. Changing that is for another PR. This is mostly to disentangle a lot of stuff now, so at a later point we could eventually go to `using StringBuilder = std::back_insert_iterator<string>`. But that's still a few PRs away.

src/strings_internal.h
```diff
@@ -122,19 +132,28 @@ class StringBuilder {
 	}
 
 	/**
-	 * Add a string using the strecpy/strecat-esque calling signature.
-	 * @param function The function to pass the current and last location to,
-	 *                 that will then return the new current location.
+	 * Get the index in the string at the current moment.
```
TrueBrain: Suggestion: `Get the current index in the string`. The `current moment` feels a bit weird, like it can magically change when you look at it long enough :)

TrueBrain (commented):
I find myself literally incapable of actually reviewing this (too many very similar changes make my head go BRRRRRRR), so I just skimmed it over, and it looks okay. I assume you tested this for various of different townnames?

One minor comment, as ofc, we can't just approve things here :P

rubidium42:
> I find myself literally incapable of actually reviewing this (too many very similar changes make my head go BRRRRRRR), so I just skimmed it over, and it looks okay. I assume you tested this for various of different townnames?

Yes, I tested things. Especially those using weird logic, like the English and Czech ones by checking whether town name list on a huge map with many towns is the same before and after the change.

TrueBrain (dismissed):

TrueBrain (approved):

PR #10814, opened 2023-05-13
Codechange: replace C-style (stredup) chat completion with std::string_view

rubidium42:
## Motivation / Problem

Replacement of C-style strings with C++ style constructs.

## Description

Replace `stredup` with a `string_view`. Especially as the `stredup` is only used to later make two strings by inserting a `\0` somewhere. So making that two `string_view`s to the origin string would work just a good.

## Checklist for review

### DIFF
--- a/src/network/network_chat_gui.cpp
+++ b/src/network/network_chat_gui.cpp
@@ -323,18 +323,16 @@ struct NetworkChatWindow : public Window {
 	 * Find the next item of the list of things that can be auto-completed.
 	 * @param item The current indexed item to return. This function can, and most
 	 *     likely will, alter item, to skip empty items in the arrays.
-	 * @return Returns the char that matched to the index.
+	 * @return Returns the view that matched to the index.
 	 */
-	const char *ChatTabCompletionNextItem(uint *item)
+	std::optional<std::string> ChatTabCompletionNextItem(uint *item)
 	{
-		static char chat_tab_temp_buffer[64];
-
 		/* First, try clients */
 		if (*item < MAX_CLIENT_SLOTS) {
 			/* Skip inactive clients */
 			for (NetworkClientInfo *ci : NetworkClientInfo::Iterate(*item)) {
 				*item = ci->index;
-				return ci->client_name.c_str();
+				return ci->client_name;
 			}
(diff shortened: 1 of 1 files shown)


src/network/network_chat_gui.cpp
```diff
@@ -346,95 +344,85 @@ struct NetworkChatWindow : public Window {
...
 		bool second_scan = false;
 
-		item = 0;
-
 		/* Copy the buffer so we can modify it without damaging the real data */
-		pre_buf = (_chat_tab_completion_active) ? stredup(_chat_tab_completion_buf) : stredup(tb->buf);
-
-		tb_buf  = ChatTabCompletionFindText(pre_buf);
-		tb_len  = strlen(tb_buf);
+		std::string_view pre_buf = _chat_tab_completion_active ? std::string_view(_chat_tab_completion_buf) : std::string_view(tb->buf);
```
rubidium42, line 381: Manual creation of `std::string_view` is needed, otherwise a copy of `_chat_tab_completion_buf` is created, and `pre_buf` would to point to freed memory after this line. For `tb->buf` it's not technically needed, until that ofcourse gets changed to `std::string`, so better safe than sorry.

src/network/network_chat_gui.cpp
```diff
@@ -346,95 +344,85 @@ struct NetworkChatWindow : public Window {
...
 				if (!second_scan) {
-					size_t offset;
-					size_t length;
+					std::string_view view;
 
 					/* If we are completing at the begin of the line, skip the ': ' we added */
-					if (tb_buf == pre_buf) {
-						offset = 0;
-						length = (tb->bytes - 1) - 2;
+					if (tb_buf.data() == pre_buf.data()) {
```
rubidium42: Yes, comparing the pointers is needed. Otherwise something like 'Hi: Hi:<tab complete>' is going to trigger this branch, as both `tb_buf` and `pre_buf` have the same content... though they're not pointing to the same location in the original string.

michicc: One could argue that this would be a perfectly good code comment 😄

michicc (approved):

PR #10815, opened 2023-05-13
Codechange: rework NewGRFProfiler to use ticks instead of calendar-days

TrueBrain:
## Motivation / Problem

We are planning to allow things like freezing the calendar, which makes this variable a bit problemetic. 

## Description

So instead, suggest to the user how many ticks there are in a calendar day, and let them figure out how many ticks they want.

Additionally, use a TimeoutTimer for this, instead of an end-date variable which is checked in an IntervalTimer.

## Limitations

There are a few odd things about the newgrf profiler. For example, when you start it in the middle of a day, and you asked for a single game-day, it would in fact run 1.5 game-days. By using `ticks`, this problem is also resolved. But it is a bit hard to guess the intention behind this functionality .. was it intended to stop on an exact day?

## Checklist for review

### DIFF
--- a/src/newgrf_profiling.cpp
+++ b/src/newgrf_profiling.cpp
@@ -14,14 +14,12 @@
 #include "spritecache.h"
 #include "walltime_func.h"
 #include "timer/timer.h"
-#include "timer/timer_game_calendar.h"
 #include "timer/timer_game_tick.h"
 
 #include <chrono>
 
 
 std::vector<NewGRFProfiler> _newgrf_profilers;
-TimerGameCalendar::Date _newgrf_profile_end_date;
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/newgrf_profiling.cpp
```diff
@@ -156,17 +156,29 @@ uint32 NewGRFProfiler::FinishAll()
...
+	_profiling_finish_timeout.Reset(ticks);
+}
+
+/**
+ * Abort the timeout timer, so the timer callback is never called.
+ */
+/* static */ void NewGRFProfiler::AbortTimer()
+{
+	_profiling_finish_timeout.Abort();
+}
```
TrueBrain, line 184: I did consider just making `_profiling_finish_timeout` public, and letting the console commands call them directly. Saves a few wrapper functions. On the other hand, nothing wrong with wrapper functions.

PeterN:
I would imagine that running for 1.5 days instead of 1 is unintentional, it doesn't let you easily compare two different runs.

rubidium42 (approved):

PR #10827, opened 2023-05-14
Fix: padding miscalculated for the network-relay and bootstrap-question GUI

TrueBrain:
## Motivation / Problem

image

## Description

image

Basically, the intention of the code was to fit every sentence on a line. But because the width got reduced by padding, it always put at least one sentence on two lines.

I noticed this for the relay, and I know (as I wrote it :D) that it was not the intention there. I noticed the same on the Bootstrap GUI .. I think the similar motivation holds there, but not 100% sure.

## Checklist for review

### DIFF
--- a/src/bootstrap_gui.cpp
+++ b/src/bootstrap_gui.cpp
@@ -102,7 +102,8 @@ public:
 	{
 		if (widget == WID_BEM_MESSAGE) {
 			*size = GetStringBoundingBox(STR_MISSING_GRAPHICS_ERROR);
-			size->height = GetStringHeight(STR_MISSING_GRAPHICS_ERROR, size->width - WidgetDimensions::scaled.frametext.Horizontal()) + WidgetDimensions::scaled.frametext.Vertical();
+			size->width += WidgetDimensions::scaled.frametext.Horizontal();
+			size->height += WidgetDimensions::scaled.frametext.Vertical();
 		}
 	}
 
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -2428,7 +2428,8 @@ struct NetworkAskRelayWindow : public Window {
 	{
 		if (widget == WID_NAR_TEXT) {
 			*size = GetStringBoundingBox(STR_NETWORK_ASK_RELAY_TEXT);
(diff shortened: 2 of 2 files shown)


PeterN:
Although this is miscalculated, I don't think a very wide error window is very nice.

This probably wants something similar to the settings window where the width and height of the multiline string are determined after the rest of the window is sized -- but in that case there are plenty of other widgets to provide a decent size, not so much here.

TrueBrain:
Yeah, this is far from ideal; so that is why I limited this PR on fixing the bug, rather than fixing the GUI. I am not sure what a good solution is .. sounds like something for another day :)

src/bootstrap_gui.cpp
```diff
@@ -102,7 +102,8 @@ class BootstrapErrorWindow : public Window {
 	{
 		if (widget == WID_BEM_MESSAGE) {
 			*size = GetStringBoundingBox(STR_MISSING_GRAPHICS_ERROR);
-			size->height = GetStringHeight(STR_MISSING_GRAPHICS_ERROR, size->width - WidgetDimensions::scaled.frametext.Horizontal()) + WidgetDimensions::scaled.frametext.Vertical();
+			size->width += WidgetDimensions::scaled.frametext.Horizontal();
+			size->height = GetStringHeight(STR_MISSING_GRAPHICS_ERROR, size->width) + WidgetDimensions::scaled.frametext.Vertical();
```
PeterN: At this point, size already has the string height too, so...
```suggestion
size->height += WidgetDimensions::scaled.frametext.Vertical();
```

src/network/network_gui.cpp
```diff
@@ -2428,7 +2428,8 @@ struct NetworkAskRelayWindow : public Window {
 	{
 		if (widget == WID_NAR_TEXT) {
 			*size = GetStringBoundingBox(STR_NETWORK_ASK_RELAY_TEXT);
-			size->height = GetStringHeight(STR_NETWORK_ASK_RELAY_TEXT, size->width - WidgetDimensions::scaled.frametext.Horizontal()) + WidgetDimensions::scaled.frametext.Vertical();
+			size->width += WidgetDimensions::scaled.frametext.Horizontal();
+			size->height = GetStringHeight(STR_NETWORK_ASK_RELAY_TEXT, size->width) + WidgetDimensions::scaled.frametext.Vertical();
```
PeterN: Same as above.

PeterN (changes requested):

PeterN (approved):

Issue #10390, opened 2023-01-22
[Bug]: GS has no autonomous way to remove industries


andythenorth:
### Version of OpenTTD

13-ish

### Expected result

GS has one or more methods to remove an industry from the map, which are _entirely_ within the control of the GS.

### Actual result

There are multiple routes to remove industries, none are inherently reliable.

### Use GSTile.DemolishTile()

See #10383 - this method can't remove industries currently, which is a bug.  This issue might be resolve by resolving #10383.

DemolishTile() does work if the player has enabled magic bulldozer via the cheat menu.  GS has no way to enable / disable cheats.

### Side-channel signalling to grf industries

The GS can use GSIndustry.SetControlFlags() to signal intents to an industry.  These can be interpreted by grf industries, and used to trigger the industry closure mechanism (via monthly industry callback 35).  

Control flags are currently: `
enum | IndustryControlFlags {  INDCTL_NO_PRODUCTION_DECREASE,  INDCTL_NO_PRODUCTION_INCREASE,  INDCTL_NO_CL }`

There is no appropriate flag for 'closure requested' or 'closure required'.  However a GS and grf can collaborate to abuse one of the other flags.

We could add further flags.

### Forced closure of an industry

This would be via monthly industry callback 35 (or the random equivalent). We delegated to grf the right to refuse closure, and I'm not proposing changing that.

### Forced removal of all of a type from the map

This would primarily be of use when modifying the map at game start, but there might be other uses.

Assuming #10383 was resolved, industries could be removed by DemolishTile().  Mass removal could be done by creating a GSIndustryList() instance, and filtering for industry type, then valuating with the tile index for each industry instance. Then running a foreach loop and calling DemolishTile on each instance.  This would consume a command for every instance, and commands are limited / expensive.

A method to mass-remove all of a type might be useful.

### Steps to reproduce

I can provide a test GS, but it's not much use really except as evidence for the findings above.

If we have proposed solutions, I can write a GS to test them.

Issue #10383, opened 2023-01-21
[Bug]: GSTile.DemolishTile() doesn't demolish some entities


andythenorth:
### Version of OpenTTD

~13 approx

### Expected result

The comment for GSTile.DemolishTile() is `Destroy everything on the given tile.`

Per https://docs.openttd.org/gs-api/classGSTile.html#a5eb630e5a2cc7bcb352c76c4d819cb40

Expected result is that everything on the tile is cleared.

### Actual result

Tested in both deity mode and company mode, at least the following are not demolished:

- all industries
- houses which appear to have a protected building flag set (needs confirmed)
- at least some objects, including default transmitters (needs confirmed if these have a protected flag set)
- there may be others, the ones above are what I found

Either:
- the comment is wrong and needs updated
-  or the method should demolish the items above (preferred).  Possibly an extra optional parameter might be required, to allow GS to respect items intended to be unremovable, or demolish the entire tile.

I can provide a test GS to repro if needed.

### Steps to reproduce

Write a GS that tries to clear tiles containing the entities listed above.
Observe that those entities are not cleared.

James103:
Does the function work as intended if the magic bulldozer cheat is enabled?

andythenorth:
> Does the function work as intended if the magic bulldozer cheat is enabled?

Yes.

mrmbernardi:
Can you provide a test GS?

PeterN:
Comment is correct but lacking context.
Basically the context is demolish a tile vs remove bits from a tile.
i.e. you can remove individual road pieces, or demolish the whole tile.
It should definitely not be expected that it will remove otherwise unremovable things.

PR #10834, opened 2023-05-15
Feature: [GS] GSTile::DemolishTile() can now demolish everything

mrmbernardi:
## Motivation / Problem

See #10383 and #10390

It's not possible for a GS to demolish certain objects such as industries.

## Description

I have added a boolean parameter `force` to GSTile::DemolishTile(). When this is set,  all objects on the tile will be destroyed.

To facilitate this, I have added an extra `DoCommandFlag` called `DC_MAGIC_BULLDOZER`. Deities can set this flag in order to have the magic bulldozer functionality apply to `CmdLandscapeClear`.

## Limitations

Since scripts can't directly set flags on commands, I had to add another entry in the `Commands` enum pointing to `CmdLandscapeClear` with a new `CMD_MAGIC_BULLDOZER` flag to set the `DC_MAGIC_BULLDOZER`. This is in line with how flags are set on other commands, but this is the first time we have two entries for the same command as far as I can see.

## Checklist for review

### DIFF
--- a/src/command_type.h
+++ b/src/command_type.h
@@ -190,6 +190,7 @@ enum Commands : uint16_t {
 	CMD_BUILD_SINGLE_RAIL,            ///< build a single rail track
 	CMD_REMOVE_SINGLE_RAIL,           ///< remove a single rail track
 	CMD_LANDSCAPE_CLEAR,              ///< demolish a tile
+	CMD_LANDSCAPE_MAGIC_CLEAR,        ///< demolish a tile with magic bulldozer behaviour
 	CMD_BUILD_BRIDGE,                 ///< build a bridge
 	CMD_BUILD_RAIL_STATION,           ///< build a rail station
 	CMD_BUILD_TRAIN_DEPOT,            ///< build a train depot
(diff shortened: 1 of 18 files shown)


glx22:
Changing function signature requires adding compatibility layer for previous versions.

BTW I'm not a fan of the signature change as it affects both AI and GS, I think it would be better to add a scoped mode to temporary switch GS into magic bulldozer enabled.

PeterN:
To be fair the author did ask about signatures but I was only thinking about internal/networking signatures and forgot about the API signature...

glx22:
Yeah for internal/network it's fine (the new command flag is a good idea too), but for API it must be though very carefully.

Maybe another option is to have 2 functions, one with only one parameter for AIs, internally calling the two parameters version, I think `@api` tag can handle that.

PeterN:
I think the scoped mode idea is the best way.

mrmbernardi:
Can someone give me an example of a scoped mode? I'm not sure what it is but I'd be happy to fix this PR up.

glx22:
You can check script_testmode.[ch]pp, script_execmode.[ch]pp or script_companymode.[ch]pp for implementation, and https://github.com/OpenTTD/OpenTTD/blob/master/regression/regression/main.nut#L392 for usage examples.

mrmbernardi:
Hi all, I've added the scoped mode.

src/script/api/script_tile.cpp
```diff
@@ -278,10 +278,15 @@
...
 {
-	EnforceDeityOrCompanyModeValid(false);
 	EnforcePrecondition(false, ::IsValidTile(tile));
-
-	return ScriptObject::Command<CMD_LANDSCAPE_CLEAR>::Do(tile);
+	if(GetMagicBulldozerMode()) {
+		EnforceDeityMode(false);
+		return ScriptObject::Command<CMD_LANDSCAPE_MAGIC_CLEAR>::Do(tile);
+	}
+	else {
```
glx22:
```suggestion
	} else {
```

src/script/api/script_tile.cpp
```diff
@@ -278,10 +278,15 @@
 
 /* static */ bool ScriptTile::DemolishTile(TileIndex tile)
 {
-	EnforceDeityOrCompanyModeValid(false);
 	EnforcePrecondition(false, ::IsValidTile(tile));
-
-	return ScriptObject::Command<CMD_LANDSCAPE_CLEAR>::Do(tile);
+	if(GetMagicBulldozerMode()) {
```
glx22:
```suggestion
	if (GetMagicBulldozerMode()) {
```

src/landscape.cpp
```diff
@@ -655,6 +655,8 @@ void ClearSnowLine()
  */
 CommandCost CmdLandscapeClear(DoCommandFlag flags, TileIndex tile)
 {
+	if ((flags & DC_MAGIC_BULLDOZER) &&_current_company != OWNER_DEITY) return CMD_ERROR;
```
glx22:
```suggestion
	if ((flags & DC_MAGIC_BULLDOZER) != 0 &&_current_company != OWNER_DEITY) return CMD_ERROR;
```

src/tile_cmd.h
```diff
@@ -208,4 +209,9 @@ static inline bool ClickTile(TileIndex tile)
 	return proc(tile);
 }
 
+static inline bool IsMagicBulldozeCommand(DoCommandFlag flags)
+{
+	return _cheats.magic_bulldozer.value || (flags & DC_MAGIC_BULLDOZER);
```
glx22:
```suggestion
	return _cheats.magic_bulldozer.value || (flags & DC_MAGIC_BULLDOZER) != 0;
```

src/script/api/game_changelog.hpp
```diff
@@ -78,6 +78,9 @@
  * API removals:
  * \li GSError::ERR_PRECONDITION_TOO_MANY_PARAMETERS, that error is never returned anymore.
  *
+ * Other changes:
+ * \li GSTile::DemolishTile takes one extra parameter to enable magic bulldozer behaviour
```
glx22: Not true anymore, but the new mode should be listed in API additions instead.

src/command_type.h
```diff
@@ -190,6 +190,7 @@ enum Commands : uint16 {
 	CMD_BUILD_SINGLE_RAIL,            ///< build a single rail track
 	CMD_REMOVE_SINGLE_RAIL,           ///< remove a single rail track
 	CMD_LANDSCAPE_CLEAR,              ///< demolish a tile
+	CMD_LANDSCAPE_MAGIC_CLEAR,        ///< demolish a tile with magic bulldozer behaviour
```
PeterN, line 193: As magic bulldozer is a command flag set by a scope mode, I don't think adding a new command should be necessary.

glx22: Probably needs handling of mode in `ScriptObject::DoCommandPrep()` and a way to add the flag during execution (I don't think any script call ever touched the flags)

PeterN: Presumably Test/Exec scope mode also sets the appropriate flags for that, so it should be similar to that?

glx22: Test/Exec just set the `estimate_only` boolean, which is then passed through `Command::Unsafe()` and `Command::Execute()`. `DC_EXEC` handling is done in `Command::Execute()`.

I used `Command::` to simplify, it's actually the templated command stuff.

Oh and the flags are extracted from `CommandTraits`.

src/script/api/script_tile.cpp
mrmbernardi: Changed

src/script/api/script_tile.cpp
mrmbernardi: Changed

src/landscape.cpp
mrmbernardi: Changed

src/tile_cmd.h
mrmbernardi: Changed

src/script/api/game_changelog.hpp
mrmbernardi: Oops that was embarrassing. Thanks for the catch.

src/command_type.h
mrmbernardi: Yes currently the extra command is needed to set the flag via the traits.

Should I try and change the template code instead?

mrmbernardi:
Rebased

mrmbernardi:
Rebased again

PR #10835, opened 2023-05-16
Codechange: Use std::vector for GRFConfig lists.

PeterN:
## Motivation / Problem

GRFConfig lists are stored as C-style linked lists.

~~And often it is difficult to know if a function accepts a pointer to a list (GRFConfig *) or a pointer to an individual config (GRFConfig *).~~

## Description

Instead, use a `std::vector` of `std::unique_ptr`s for `GRFConfigList`.

~~There is still some faffing about with iterators to work out 'index' for GUI purposes, but some manual management is now gone.~~

~~Some conversions from linked-list to std algorithms are missing or not optimal, as it takes some effort to work out what the mix of pointers and pointers to pointers is actually intended to do.~~

~~To do:~~

* [x] ~~Standardize on parameter naming, differentiating between lists and individual configs.~~
* [x] ~~Fix initial loading which still uses new/delete and makes a copy, instead of doing it in-place.~~
* [x] ~~Fix doxygen which probably refers to pointers and heads and tails.~~

## Checklist for review

### DIFF
--- a/src/gfxinit.cpp
+++ b/src/gfxinit.cpp
@@ -144,6 +144,31 @@ void CheckExternalFiles()
 	if (!error_msg.empty()) ShowInfoI(error_msg);
 }
 
+/**
+ * Get GRFConfig for the default extra graphics.
+ * @return Managed pointer to default extra GRFConfig.
+ */
+static std::unique_ptr<GRFConfig> GetDefaultExtraGRFConfig()
+{
+	auto gc = std::make_unique<GRFConfig>("OPENTTD.GRF");
+	gc->palette |= GRFP_GRF_DOS;
+	FillGRFDetails(*gc, false, BASESET_DIR);
+	ClrBit(gc->flags, GCF_INIT_ONLY);
+	return gc;
+}
+
+/**
+ * Get GRFConfig for the baseset extra graphics.
+ * @return Managed pointer to baseset extra GRFConfig.
+ */
+static std::unique_ptr<GRFConfig> GetBasesetExtraGRFConfig()
+{
+	auto gc = std::make_unique<GRFConfig>(BaseGraphics::GetUsedSet()->GetOrCreateExtraConfig());
+	if (gc->param.empty()) gc->SetParameterDefaults();
+	ClrBit(gc->flags, GCF_INIT_ONLY);
+	return gc;
+}
+
 /** Actually load the sprite tables. */
 static void LoadSpriteTables()
 {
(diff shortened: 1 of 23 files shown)


src/gfxinit.cpp
```diff
@@ -215,9 +214,8 @@ static void LoadSpriteTables()
 	FillGRFDetails(extra, false, BASESET_DIR);
 	ClrBit(extra->flags, GCF_INIT_ONLY);
 
-	extra->next = top;
-	master->next = extra;
-	_grfconfig = master;
+	_grfconfig.insert(_grfconfig.begin(), *extra);
```
rubidium42: What about something like `GRFConfig *master = _grfconfig.emplace(_grfconfig.begin(), master_filename);` around like 197?
Saves a copy here and a manual delete later on.
And if/when tackling `GRFConfig *extra`, the `.c_str()` isn't needed anymore since `GRFConfig` accepts `std::string`.

btzy:
It sounds like you want a `std::forward_list` instead of a `std::list`?

PeterN:
Now switched to using `std::vector<std::shared_ptr<GRFConfig>>` which is more easily managed.

src/network/core/network_game_info.cpp
```diff
@@ -230,17 +230,16 @@ void SerializeNetworkGameInfo(Packet &p, const NetworkServerGameInfo &info, bool
...
 		 * selected in the NewGRF GUI and not the ones that are used due
 		 * to the fact that they are in [newgrf-static] in openttd.cfg */
-		const GRFConfig *c;
 		uint count = 0;
 
 		/* Count number of GRFs to send information about */
-		for (c = info.grfconfig; c != nullptr; c = c->next) {
+		for (const auto &c : info.grfconfig) {
 			if (!HasBit(c->flags, GCF_STATIC)) count++;
 		}
```
rubidium42: `uint count = std::ranges::count_if(info.grfconfig, [](const auto &c){ return !HasBit(c->flags, GCF_STATIC); });`

src/newgrf_config.cpp
```diff
@@ -537,51 +515,24 @@ bool GRFFileScanner::AddFile(const std::string &filename, size_t basepath_length
...
-			for (pd = &_all_grfs; (d = *pd) != nullptr; pd = &d->next) {
-				if (c->ident.grfid == d->ident.grfid && c->ident.md5sum == d->ident.md5sum) added = false;
-				/* Because there can be multiple grfs with the same name, make sure we checked all grfs with the same name,
-				 *  before inserting the entry. So insert a new grf at the end of all grfs with the same name, instead of
-				 *  just after the first with the same name. Avoids doubles in the list. */
-				if (StrCompareIgnoreCase(c->GetName(), d->GetName()) <= 0) {
-					stop = true;
-				} else if (stop) {
-					break;
-				}
```
rubidium42, line 560: Just for the next time reviewing: sorting is *also* done by `DoScanNewGRFFiles`, making all attempts to sort here kinda pointless.

src/newgrf_config.cpp
```diff
@@ -537,51 +515,24 @@ bool GRFFileScanner::AddFile(const std::string &filename, size_t basepath_length
...
-			}
-			if (added) {
-				c->next = d;
-				*pd = c;
-			}
+		if (std::ranges::none_of(_all_grfs, [&c](const auto &gc) { return c->ident.grfid == gc->ident.grfid && c->ident.md5sum == gc->ident.md5sum; })) {
+			grfconfig = _all_grfs.emplace_back(std::move(c)).get();
```
rubidium42: This isn't needed. The result of `.get()` on the unique_ptr will not change when moving. Otherwise putting unique_ptrs into vectors would be an enormous hassle.

src/saveload/newgrf_sl.cpp
```diff
@@ -91,11 +91,11 @@ struct NGRFChunkHandler : ChunkHandler {
 
 		int index = 0;
 
-		for (GRFConfig *c = _grfconfig; c != nullptr; c = c->next) {
+		for (const auto &c : _grfconfig) {
 			if (HasBit(c->flags, GCF_STATIC) || HasBit(c->flags, GCF_INIT_ONLY)) continue;
 			this->SaveParameters(*c);
 			SlSetArrayIndex(index++);
-			SlObject(c, description);
+			SlObject(&*c, description);
```
rubidium42: Maybe, so it's consistent with `LoadCommon`.
```suggestion
			SlObject(c.get(), description);
```

src/script/api/script_newgrf.cpp
```diff
@@ -28,11 +28,8 @@ ScriptNewGRFList::ScriptNewGRFList()
...
 
-	for (auto c = _grfconfig; c != nullptr; c = c->next) {
-		if (!HasBit(c->flags, GCF_STATIC) && c->ident.grfid == grfid) {
-			return true;
-		}
-	}
+	auto it = std::ranges::find_if(_grfconfig, [grfid](const auto &c) { return !HasBit(c->flags, GCF_STATIC) && c->ident.grfid == grfid; });
+	if (it != std::end(_grfconfig)) return true;
 
 	return false;
```
rubidium42: Maybe use `std::ranges_any_of` instead?

src/settings.cpp
```diff
@@ -1123,23 +1122,15 @@ static GRFConfig *GRFLoadConfig(const IniFile &ini, const char *grpname, bool is
...
-				SetDParamStr(1, gc->filename);
-				ShowErrorMessage(STR_CONFIG_ERROR, STR_CONFIG_ERROR_DUPLICATE_GRFID, WL_CRITICAL);
-				duplicate = true;
-				break;
-			}
-		}
-		if (duplicate) {
-			delete c;
+		auto found = std::ranges::find_if(list, [&c](const auto &gc) { return gc->ident.grfid == c->ident.grfid; });
+		if (found != std::end(list)) {
```
rubidium42, line 1130:
```suggestion
		auto found = std::ranges::any_of(list, [&c](const auto &gc) { return gc->ident.grfid == c->ident.grfid; });
		if (found) {
```

PeterN: Can't for this one as `found` is used inside the if-block.

PeterN:
This has switched to using `std::vector<std::unique_ptr<GRFConfig>>` which is generally a bit better.

This does mean that the GRFConfigs are duplicated for local network server info however. Possibly that could use a different layout as most of `GRFConfig` is not actually needed there.

src/newgrf_gui.cpp
```diff
@@ -1520,30 +1485,26 @@ struct NewGRFWindow : public Window, NewGRFScanCallback {
...
+		}
+
+		auto entry = (ins_pos >= 0 && static_cast<size_t>(ins_pos) < std::size(this->actives))
+			? std::next(std::begin(this->actives), ins_pos)
+			: std::end(this->actives);
 
-		/* Insert GRF config to configuration list. */
-		c->next = *entry;
-		*entry = c;
+		/*/ Copy GRF details from scanned list. */
```
rubidium42:
```suggestion
		/* Copy GRF details from scanned list. */
```

rubidium42 (approved):

PR #10836, opened 2023-05-16
Codechange: Replace all uses of SmallMap with std::map

PeterN:
## Motivation / Problem

SmallMap.

## Description

Replace all uses of SmallMap with std::map. Some other adjustments were needed to make this possible:

1) Replacing a pointer with a unique_ptr in FontColourMap.
2) Swapping key/value around for SocketList due to const constraints (SocketList was modifying its keys), 

## Limitations

Given the changes to SocketList, this probably shouldn't be a map in the first place?

## Checklist for review


src/fios.h
```diff
@@ -16,6 +16,7 @@
 #include "gamelog.h"
 #include "network/core/tcp_content_type.h"
 #include "timer/timer_game_calendar.h"
+#include <map>
```
TrueBrain: I think we should add this to `stdafx.h`. As the fact it works for some platforms and not for others is just annoying, and we shouldn't want to deal with :)

PeterN: Seems reasonable. We did it for vector already :)

LordAro:
> [Codechange: Use std::map for gamelog's tempraroy grf name lookup.](https://github.com/OpenTTD/OpenTTD/pull/10836/commits/7e7b18f5bc18bf81eb32bfd5c198e934d0459ad2)

:)

src/fios.h
```diff
@@ -26,7 +25,7 @@ enum SaveLoadInvalidateWindowData {
 	SLIWD_FILTER_CHANGES,        ///< The filename filter has changed (via the editbox)
 };
 
-typedef SmallMap<uint, CompanyProperties *> CompanyPropertiesMap;
+using CompanyPropertiesMap = std::map<uint, CompanyProperties *>;
```
rubidium42: Rabbit hole-ish: `std::unique_ptr`?

rubidium42 (commented):
I'd make the static initialisation and the flip of the socket list separate commits. The rest could be squashed into one commit I'd say.

src/fios.h
PeterN: Good point, I did it for the font map one already.

src/os/macosx/string_osx.cpp
```diff
@@ -103,7 +103,7 @@ class CoreTextParagraphLayout : public ParagraphLayouter {
 				/* Extract font information for this run. */
 				CFRange chars = CTRunGetStringRange(run);
 				auto map = fontMapping.begin();
-				while (map < fontMapping.end() - 1 && map->first <= chars.location) map++;
+				while (std::next(map) != fontMapping.end() && map->first <= chars.location) map++;
```
rubidium42: Wouldn't this be equal to `auto map = fontMapping.upper_bound(chars.location)`?

PeterN:
Reorganized a bit. Fixes are done first, then conversion to map, then switching some things from raw pointer to unique_ptr.,

rubidium42 (approved):

PR #10841, opened 2023-05-18
Codechange: use std::array for MD5 hash storage

rubidium42:
## Motivation / Problem

With the MD5 hash/digest/checksum in a uint8/byte array throughout the code, the fact that a MD5 hash is 16 bytes got spread around a lot. Furthermore it being a C-style array meant using memcmp/memcpy instead of letting the compiler generate the appropriate code.
Also `MD5SumToString` was basically a simple wrapper to convert the uint8 C-style array into a span, which with `std::array` will be created automatically removing the need for the wrapper.

## Description

Replace `uint8 md5sum[16]` and similar with a new class `MD5Hash` that extends `std::array` and add `^=` to easily X-or different hashes.
Replace `memcpy` with assignment, `memcmp` with `==` and `MD5SumToString` with `FormatAsHexString`.

## Limitations

None?

## Checklist for review


src/3rdparty/md5/md5.h
```diff
@@ -53,6 +53,25 @@
 #ifndef MD5_INCLUDED
 #define MD5_INCLUDED
 
+/** The number of bytes in a MD5 hash. */
+static const size_t MD5_HASH_BYTES = 16;
+
+/**
+ * Container for storing a MD5 hash/checksum/digest.
+ */
+struct MD5Hash : public std::array<byte, MD5_HASH_BYTES> {
```
michicc: Strictly speaking, you wouldn't even need a class, but one can definitely argue if the code below is in fact better or worse:
```c++
using MD5Hash = std::array<byte, MD5_HASH_BYTES>;

inline MD5Hash &operator^=(MD5Hash &lhs, const MD5Hash &rhs)
{
    for (size_t i = 0; i < lhs.size(); i++) lhs[i] ^= rhs[i];
    return lhs;
}
````

Comment (PeterN) [reply src/3rdparty/md5/md5.h]: I think that is neater and more extensible -- no extra class indeed.

**Comment by PeterN** [comment]
Probably every `sizeof(md5)` and `lengthof(md5)` should be replaced with `md5.size()`

File: src/settings.cpp [hunk 2]
```diff
@@ -978,12 +979,12 @@ static GRFConfig *GRFLoadConfig(IniFile &ini, const char *grpname, bool is_stati
 		has_grfid = DecodeHexText(filename, grfid_buf, lengthof(grfid_buf));
 		if (has_grfid) {
 			filename += 1 + 2 * lengthof(grfid_buf);
-			has_md5sum = DecodeHexText(filename, md5sum, lengthof(md5sum));
+			has_md5sum = DecodeHexText(filename, md5sum.data(), lengthof(md5sum));
```
TrueBrain: Is lengthof actually doing here what we expect? Shouldn't that be size()?

TrueBrain (approved):

PR #10843, opened 2023-05-19
Codechange: migrate away from C-style Getstring

rubidium42:
## Motivation / Problem

Migration to more C++-style strings.

## Description

Replace `GetString(buffer, str, lastof(buffer))` with something like `std::string message = GetString(str)`.

In some cases, where there is filtering happening on the string afterwards to remove/replace certain characters, just pass it through `StrMakeValid` to remove these characters (color/font size) or replace (\r \n to space). This currently happened only for the first character at the front, but now it will remove any color/font encoding character.

## Limitations

There is more to be changed, but lets keep it in reviewable portions.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -356,10 +356,6 @@ CommandCost CheckTileOwnership(TileIndex tile)
  */
 static void GenerateCompanyName(Company *c)
 {
-	/* Reserve space for extra unicode character. We need to do this to be able
-	 * to detect too long company name. */
-	char buffer[(MAX_LENGTH_COMPANY_NAME_CHARS + 1) * MAX_CHAR_LENGTH];
-
 	if (c->name_1 != STR_SV_UNNAMED) return;
 	if (c->last_build_coordinate == 0) return;
 
@@ -367,6 +363,7 @@ static void GenerateCompanyName(Company *c)
 
 	StringID str;
 	uint32 strp;
+	std::string name;
 	if (t->name.empty() && IsInsideMM(t->townnametype, SPECSTR_TOWNNAME_START, SPECSTR_TOWNNAME_LAST + 1)) {
 		str = t->townnametype - SPECSTR_TOWNNAME_START + SPECSTR_COMPANY_NAME_START;
 		strp = t->townnameparts;
@@ -377,8 +374,8 @@ verify_name:;
 			if (cc->name_1 == str && cc->name_2 == strp) goto bad_town_name;
 		}
 
-		GetString(buffer, str, lastof(buffer));
-		if (Utf8StringLength(buffer) >= MAX_LENGTH_COMPANY_NAME_CHARS) goto bad_town_name;
+		name = GetString(str);
+		if (Utf8StringLength(name) >= MAX_LENGTH_COMPANY_NAME_CHARS) goto bad_town_name;
 
 set_name:;
 		c->name_1 = str;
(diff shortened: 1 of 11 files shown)


src/company_cmd.cpp
```diff
@@ -356,17 +356,14 @@ CommandCost CheckTileOwnership(TileIndex tile)
...
-	char buffer[(MAX_LENGTH_COMPANY_NAME_CHARS + 1) * MAX_CHAR_LENGTH];
-
 	if (c->name_1 != STR_SV_UNNAMED) return;
 	if (c->last_build_coordinate == 0) return;
 
 	Town *t = ClosestTownFromTile(c->last_build_coordinate, UINT_MAX);
 
 	StringID str;
 	uint32 strp;
+	std::string name;
```
LordAro, line 366: if i'm not mistaken, this doesn't need to be declared anywhere beyond where it's used

rubidium42: My compiler disagrees / complains otherwise:
```
src/company_cmd.cpp: In function ‘void GenerateCompanyName(Company*)’:
src/company_cmd.cpp:379:1: error: jump to label ‘set_name’ [-fpermissive]
  379 | set_name:;
      | ^~~~~~~~
src/company_cmd.cpp:400:8: note:   from here
  400 |   goto set_name;
      |        ^~~~~~~~
src/company_cmd.cpp:376:15: note:   skips initialization of ‘std::string name’
  376 |   std::string name = GetString(str);
      |               ^~~~
```

I do agree that it shouldn't be needed, but that would require rewriting the function to not use `goto`. That's another battle to fight somewhere in the future.

LordAro: Ah. Lol.

src/gfx.cpp
```diff
@@ -681,9 +681,7 @@ int DrawString(int left, int right, int top, std::string_view str, TextColour co
  */
 int DrawString(int left, int right, int top, StringID str, TextColour colour, StringAlignment align, bool underline, FontSize fontsize)
 {
-	char buffer[DRAW_STRING_BUFFER];
-	GetString(buffer, str, lastof(buffer));
-	return DrawString(left, right, top, buffer, colour, align, underline, fontsize);
+	return DrawString(left, right, top, GetString(str), colour, align, underline, fontsize);
```
LordAro, line 685: worth marking these functions as inline? I imagine the compiler probably is already, but...

rubidium42: Don't you have to move the functions to the header to make them inline? Then we'd be adding a dependency on `strings_func.h` to `gfx_func.h`.
I'm not sure whether it's worth doing that. I'm personally in favour of letting the compiler decide (during LTO) whether to inline it, instead of the whole spiel with adding hints that it might/will ignore and increasing compilation time due to the dependency on `strings_func.h`.

src/industry_gui.cpp
```diff
@@ -1547,7 +1535,7 @@ class IndustryDirectoryWindow : public Window {
 
 		for (byte j = 0; j < lengthof(i->produced_cargo); j++) {
 			if (i->produced_cargo[j] == CT_INVALID) continue;
-			cargos.push_back({ i->produced_cargo[j], i->last_month_production[j], cargo_suffix[j].text, ToPercent8(i->last_month_pct_transported[j]) });
+			cargos.push_back({ i->produced_cargo[j], i->last_month_production[j], cargo_suffix[j].text.c_str(), ToPercent8(i->last_month_pct_transported[j]) });
```
PeterN, line 1538: Could CargoSuffix::text be std::string or is that creeping too much?

rubidium42: Well... `CargoSuffix::text` has become `std::string` in this PR.

However, I did, in the end, not change `CargoInfo::suffix` to `std::string`, as that is just a "pointer" to the data and as such a needless copy. I did try making it `std::string_view`, but then I ran into `SetDParamStr` that doesn't accept `std::string_view`, so I stopped and undid the `std::string_view` change as it became a bit too rabbit-holey.

PeterN: Whoops, yes, that's the one I meant :-)

And yes, that definitely sounds like creeping too much!

PeterN (approved):
Looks alright to me.

PR #10849, opened 2023-05-20
Add: Ctrl+click Rename on VehicleView to rename the group

fmang:
## Motivation / Problem

I like to assign a group to all my vehicles so that they all appear named and can be easily managed in bulks after I create duplicate vehicles. Duplicating specifically-named vehicles copies the name, but then duplicated vehicles cannot be easily renamed at once.

Currently, I have to open the vehicle list window, create a group, move my vehicle into it, for every new line. This is tedious enough that I sometimes give up, but then I cannot tell which vehicle is which in vehicle lists.

## Description

In the vehicle view window, the user can name a specific vehicle using the pencil icon. I have extended its behavior so that shift-clicking renames its group instead. If a vehicle has not group, a new group is automatically created containing the vehicle and vehicles sharing its orders.

This feature makes it trivial to create groups for new bus lines without even having to open the vehicle list window. Since unnamed vehicles inherit the name of their group, ~~shift-clicking~~ ctrl-clicking the pencil icon will still have the effect of renaming the vehicle, which makes this approach intuitive in my opinion.

This feature should not negatively impact anyone’s existing workflow since it can hardly be accidentally triggered, and brings no clutter to the UI except for one longer tooltip.

## Limitations

~~I assumed that when the user creates a group for a vehicle, they always expect vehicles sharing orders to follow.~~ Only the current vehicle is added to the new group.

Clicking Default in the query string window when creating a new group cancels the creation instead of creating a default-named group. I suspect this is a behavior of ShowQueryString.

Cancelling the renaming will still create a group if the vehicle was ungrouped, pretty much like cancelling a group creation leaves a default-named group.

## Checklist for review

Nothing special.

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -21,6 +21,7 @@
 #include "train.h"
 #include "aircraft.h"
 #include "depot_map.h"
+#include "group_cmd.h"
 #include "group_gui.h"
 #include "strings_func.h"
 #include "vehicle_func.h"
@@ -2840,6 +2841,15 @@ private:
 	};
 	bool mouse_over_start_stop = false;
 
+	/** Possible contexts in which ShowQueryString may be called. */
+	enum class QueryType {
+		RenameVehicle,
+		RenameGroup,
+	};
+
+	/** Context in which ShowQueryString was last called. */
+	QueryType query;
+
 	/**
 	 * Display a plane in the window.
 	 * @param plane Plane to show.
@@ -3088,15 +3098,39 @@ public:
 		DrawString(tr.left, tr.right, CenterBounds(tr.top, tr.bottom, FONT_HEIGHT_NORMAL), str, text_colour, SA_HOR_CENTER);
 	}
 
+	/** Reuse the same logic as VehicleGroupWindow::ShowRenameGroupWindow, but in the vehicle view window context. */
+	void ShowRenameGroupWindow(bool empty)
+	{
+		const Vehicle *v = Vehicle::Get(this->window_number);
+		/* Show empty query for new groups */
+		StringID str = STR_EMPTY;
+		if (!empty) {
+			SetDParam(0, v->group_id);
+			str = STR_GROUP_NAME;
+		}
+		this->query = QueryType::RenameGroup;
+		ShowQueryString(str, STR_GROUP_RENAME_CAPTION, MAX_LENGTH_GROUP_NAME_CHARS, this, CS_ALPHANUMERAL, QSF_ENABLE_DEFAULT | QSF_LEN_IN_CHARS);
+	}
+
 	void OnClick(Point pt, int widget, int click_count) override
 	{
 		const Vehicle *v = Vehicle::Get(this->window_number);
 
 		switch (widget) {
 			case WID_VV_RENAME: { // rename
-				SetDParam(0, v->index);
-				ShowQueryString(STR_VEHICLE_NAME, STR_QUERY_RENAME_TRAIN_CAPTION + v->type,
-						MAX_LENGTH_VEHICLE_NAME_CHARS, this, CS_ALPHANUMERAL, QSF_ENABLE_DEFAULT | QSF_LEN_IN_CHARS);
+				if (_ctrl_pressed) {
+					if (v->group_id == DEFAULT_GROUP) {
+						// Add the vehicle to a new group. The callback will then open the window to rename the group.
+						Command<CMD_ADD_VEHICLE_GROUP>::Post(STR_ERROR_GROUP_CAN_T_ADD_VEHICLE, CcRenameVehicleGroup, NEW_GROUP, v->index, false);
+					} else {
+						ShowRenameGroupWindow(false);
+					}
+				} else {
+					this->query = QueryType::RenameVehicle;
+					SetDParam(0, v->index);
+					ShowQueryString(STR_VEHICLE_NAME, STR_QUERY_RENAME_TRAIN_CAPTION + v->type,
+							MAX_LENGTH_VEHICLE_NAME_CHARS, this, CS_ALPHANUMERAL, QSF_ENABLE_DEFAULT | QSF_LEN_IN_CHARS);
+				}
 				break;
 			}
 
@@ -3193,7 +3227,17 @@ public:
 	{
 		if (str == nullptr) return;
 
-		Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + Vehicle::Get(this->window_number)->type, this->window_number, str);
+		const Vehicle *v = Vehicle::Get(this->window_number);
+		switch (this->query) {
+			case QueryType::RenameVehicle:
+				Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + v->type, v->index, str);
+				break;
+			case QueryType::RenameGroup:
+				Command<CMD_ALTER_GROUP>::Post(STR_ERROR_GROUP_CAN_T_RENAME, AlterGroupMode::Rename, v->group_id, 0, str);
+				break;
+			default:
+				NOT_REACHED();
+		}
 	}
 
 	void OnMouseOver(Point pt, int widget) override
@@ -3352,6 +3396,16 @@ void StopGlobalFollowVehicle(const Vehicle *v)
 	}
 }
 
+/**
+ * When naming the group of an ungrouped vehicle, the CMD_ADD_VEHICLE_GROUP is
+ * called with this callback to create the new group and add the vehicle. Then,
+ * this callback must open the *Rename group* box to rename the new group.
+ */
+void CcRenameVehicleGroup(Commands cmd, const CommandCost &result, GroupID new_group, GroupID, VehicleID veh_id, bool)
+{
+	VehicleViewWindow *w = (VehicleViewWindow*) FindWindowById(WC_VEHICLE_VIEW, veh_id);
+	if (w != nullptr) w->ShowRenameGroupWindow(true);
+}
 
 /**
  * This is the Callback method after the construction attempt of a primary vehicle
--- a/src/group_cmd.h
+++ b/src/group_cmd.h
@@ -43,5 +43,6 @@ DEF_CMD_TRAIT(CMD_SET_GROUP_LIVERY,          CmdSetGroupLivery,         0, CMDT_
 
 void CcCreateGroup(Commands cmd, const CommandCost &result, GroupID new_group, VehicleType vt, GroupID parent_group);
 void CcAddVehicleNewGroup(Commands cmd, const CommandCost &result, GroupID new_group, GroupID, VehicleID veh_id, bool);
+void CcRenameVehicleGroup(Commands cmd, const CommandCost &result, GroupID new_group, GroupID, VehicleID veh_id, bool);
 
 #endif /* GROUP_CMD_H */
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -4255,10 +4255,10 @@ STR_VEHICLE_COMMAND_STARTED                                     :{GREEN}Started
 STR_VEHICLE_DETAILS_CAPTION                                     :{WHITE}{VEHICLE} (Details)
 
 ###length VEHICLE_TYPES
-STR_VEHICLE_DETAILS_TRAIN_RENAME                                :{BLACK}Name train
-STR_VEHICLE_DETAILS_ROAD_VEHICLE_RENAME                         :{BLACK}Name road vehicle
-STR_VEHICLE_DETAILS_SHIP_RENAME                                 :{BLACK}Name ship
-STR_VEHICLE_DETAILS_AIRCRAFT_RENAME                             :{BLACK}Name aircraft
+STR_VEHICLE_DETAILS_TRAIN_RENAME                                :{BLACK}Name train. Ctrl+Click names its group
+STR_VEHICLE_DETAILS_ROAD_VEHICLE_RENAME                         :{BLACK}Name road vehicle. Ctrl+Click names its group
+STR_VEHICLE_DETAILS_SHIP_RENAME                                 :{BLACK}Name ship. Ctrl+Click names its group
+STR_VEHICLE_DETAILS_AIRCRAFT_RENAME                             :{BLACK}Name aircraft. Ctrl+Click names its group
 
 STR_VEHICLE_INFO_AGE_RUNNING_COST_YR                            :{BLACK}Age: {LTBLUE}{STRING2}{BLACK}   Running Cost: {LTBLUE}{CURRENCY_LONG}/yr
 STR_VEHICLE_INFO_AGE                                            :{COMMA} year{P "" s} ({COMMA})
--- a/src/network/network_command.cpp
+++ b/src/network/network_command.cpp
@@ -82,7 +82,8 @@ static constexpr auto _callback_tuple = std::make_tuple(
 	&CcBuildIndustry,
 	&CcStartStopVehicle,
 	&CcGame,
-	&CcAddVehicleNewGroup
+	&CcAddVehicleNewGroup,
+	&CcRenameVehicleGroup
 );
 
 #ifdef SILENCE_GCC_FUNCTION_POINTER_CAST


src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
...
+				Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + v->type, v->index, str);
+				break;
+			case QueryType::RenameGroup:
+				if (v->group_id == DEFAULT_GROUP)
+					Command<CMD_ADD_VEHICLE_GROUP>::Post(STR_ERROR_GROUP_CAN_T_ADD_VEHICLE, NEW_GROUP, v->index, true /* add shared */);
+				if (v->group_id != DEFAULT_GROUP) // Not guaranteed if CMD_ADD_VEHICLE_GROUP fails.
+					Command<CMD_ALTER_GROUP>::Post(STR_ERROR_GROUP_CAN_T_RENAME, AlterGroupMode::Rename, v->group_id, 0, str);
+				break;
+			default:
+				; /* Not supposed to happen. */
```
2TallTyler: We have a handy tool for this, and it can help you test if we ever use `QueryType::None` by throwing an assertion if that happens. 😃 

```suggestion
				NOT_REACHED();
```

src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
...
 
-		Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + Vehicle::Get(this->window_number)->type, this->window_number, str);
+		const Vehicle *v = Vehicle::Get(this->window_number);
+		switch (query) {
+			case QueryType::RenameVehicle:
+				Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + v->type, v->index, str);
+				break;
+			case QueryType::RenameGroup:
+				if (v->group_id == DEFAULT_GROUP)
+					Command<CMD_ADD_VEHICLE_GROUP>::Post(STR_ERROR_GROUP_CAN_T_ADD_VEHICLE, NEW_GROUP, v->index, true /* add shared */);
```
2TallTyler: This comment should be either on its own line or at the end of the line, not inside a function call. It would also benefit from being a full sentence to expand on what you mean. 🙂

src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
...
 		if (str == nullptr) return;
 
-		Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + Vehicle::Get(this->window_number)->type, this->window_number, str);
+		const Vehicle *v = Vehicle::Get(this->window_number);
+		switch (query) {
+			case QueryType::RenameVehicle:
+				Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + v->type, v->index, str);
+				break;
+			case QueryType::RenameGroup:
+				if (v->group_id == DEFAULT_GROUP)
```
2TallTyler: `If` blocks need {} around the contained statements.

src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
...
+					Command<CMD_ADD_VEHICLE_GROUP>::Post(STR_ERROR_GROUP_CAN_T_ADD_VEHICLE, NEW_GROUP, v->index, true /* add shared */);
+				if (v->group_id != DEFAULT_GROUP) // Not guaranteed if CMD_ADD_VEHICLE_GROUP fails.
```
2TallTyler: Shouldn't this be `else`? Like above, it also needs {} to contain its statement.

src/vehicle_gui.cpp
```diff
@@ -2839,6 +2840,16 @@ struct VehicleViewWindow : Window {
 	};
 	bool mouse_over_start_stop = false;
 
+	/** Possible contexts in which ShowQueryString may be called. */
+	enum class QueryType {
+		None, ///< Invalid pseudo-context.
```
2TallTyler: Why do we need this? There shouldn't be a path where this value could ever be used.

src/vehicle_gui.cpp
```diff
@@ -3093,9 +3104,20 @@ struct VehicleViewWindow : Window {
 
 		switch (widget) {
 			case WID_VV_RENAME: { // rename
-				SetDParam(0, v->index);
-				ShowQueryString(STR_VEHICLE_NAME, STR_QUERY_RENAME_TRAIN_CAPTION + v->type,
-						MAX_LENGTH_VEHICLE_NAME_CHARS, this, CS_ALPHANUMERAL, QSF_ENABLE_DEFAULT | QSF_LEN_IN_CHARS);
+				if (_shift_pressed) {
```
2TallTyler: `Shift+click` is almost exclusively used for estimating cost. We use `Ctrl+click` for alternate functions.

2TallTyler (dismissed):
I would love a more convenient way to add groups. 😃 

That said, I'm not convinced this is the right way to do it. I'm fine with it being a "hidden feature" with a modifier key, but I would not expect the Rename button to do anything with groups.

We already have a contextual menu that opens a psuedo-group window of vehicles that share orders.

quasigroup

Inside that psuedo-group, JGRPP adds a "Create group from list" option to its "Manage list" dropdown ([old commit](https://github.com/JGRennison/OpenTTD-patches/commit/ca9f6411cce7dbabaa2224ba58bb68acd6fd5848), would need rewriting).

creategroupfromlist

Obviously, that's several more clicks than your proposed solution, and they do add up. So I wonder if this "Show all vehicles that share this schedule" button could be added to the main vehicle window (or perhaps moved entirely) and given the Ctrl+click hidden feature to create a new group proposed by this PR. To me, that would make a lot more sense than hijacking the Rename button.

src/vehicle_gui.cpp
glx22: Can't be `else`, if adding to group failed it will still be `DEFAULT_GROUP`

src/vehicle_gui.cpp
```diff
@@ -3093,9 +3104,20 @@ struct VehicleViewWindow : Window {
...
+					query = QueryType::RenameGroup;
```
glx22:
```suggestion
					this->query = QueryType::RenameGroup;
```

src/vehicle_gui.cpp
```diff
@@ -3093,9 +3104,20 @@ struct VehicleViewWindow : Window {
...
+					StringID str = STR_EMPTY;
+					if (v->group_id != DEFAULT_GROUP) {
+						SetDParam(0, v->group_id);
+						str = STR_GROUP_NAME;
+					}
+					ShowQueryString(str, STR_GROUP_RENAME_CAPTION, MAX_LENGTH_GROUP_NAME_CHARS, this, CS_ALPHANUMERAL, QSF_ENABLE_DEFAULT | QSF_LEN_IN_CHARS);
+				} else {
+					query = QueryType::RenameVehicle;
```
glx22:
```suggestion
					this->query = QueryType::RenameVehicle;
```

src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
 	{
 		if (str == nullptr) return;
 
-		Command<CMD_RENAME_VEHICLE>::Post(STR_ERROR_CAN_T_RENAME_TRAIN + Vehicle::Get(this->window_number)->type, this->window_number, str);
+		const Vehicle *v = Vehicle::Get(this->window_number);
+		switch (query) {
```
glx22:
```suggestion
		switch (this->query) {
```

src/vehicle_gui.cpp
```diff
@@ -3192,7 +3214,22 @@ struct VehicleViewWindow : Window {
...
+				if (v->group_id == DEFAULT_GROUP)
+					Command<CMD_ADD_VEHICLE_GROUP>::Post(STR_ERROR_GROUP_CAN_T_ADD_VEHICLE, NEW_GROUP, v->index, true /* add shared */);
+				if (v->group_id != DEFAULT_GROUP) // Not guaranteed if CMD_ADD_VEHICLE_GROUP fails.
+					Command<CMD_ALTER_GROUP>::Post(STR_ERROR_GROUP_CAN_T_RENAME, AlterGroupMode::Rename, v->group_id, 0, str);
+				break;
+			default:
+				; /* Not supposed to happen. */
+		}
+
+		query = QueryType::None;
```
glx22:
```suggestion
		this->query = QueryType::None;
```
but as mentioned in another comment `QueryType::None` is most likely unneeded.

src/vehicle_gui.cpp
rubidium42: Actually... `else` is "fine", or rather... this bit of code has a major problem and multiplayer behaves as if there is an else.
In multiplayer `Post` is essentially asynchronous (and think of it as running several game ticks later), so if it makes a new group it won't rename it because the creation of the group has not yet happened when doing this check.

fmang:
Thank you everyone for the comments.

@2TallTyler Before I apply the suggestions, I’d like your verdict on the idea. The vehicle sharing orders list you mention could be a good option indeed if the Create group function was available upstream, but it has one severe limitation in my case: it requires another vehicle sharing orders to exist in order to be accessible. That could be lifted if the button was made available even if the vehicle does not share orders with any other vehicle, in which case the list would contain only the one vehicle. To be clear, I make groups even for single vehicles so that duplicating them works smoothly.

Adding a *set group* icon next to the pencil icon could be an alternative option. To go to the full length, that button could let the user select an existing group, or create a new one, but that requires much more effort for a use case I can’t imagine.

glx22:
Less hidden grouping, i.e. grouping from the vehicle window itself, seems to be a good idea. Of course with less functionality than the manage group window, but a simple "move to group" button inside vehicle window would be a improvement.

src/vehicle_gui.cpp
PeterN: In case it is not clear, this code will not work in multiplayer. Posting commands is asynchronous, and the command will not have been executed by the time you check v->group_id again.

fmang: I’ve noticed Post accepts a callback to perform further processing after the command executes. If it accepts a custom state through a lambda, I’ll use that. Otherwise I’ll resort to adjusting the commands to perform both operations as a single command, or a adding a state somewhere in the vehicle window to store the new name.

michicc: Command callbacks can have various argument formats. One of them allows you to get all the command parameters in addition to the command result. For example you can look at `CcAddVehicleNewGroup`.

2TallTyler:
I have no argument against making the psuedo-group available even if the list would only contain one vehicle. If that's what is needed, go for it.

It would make sense to use the existing psuedo-group button instead of creating a new one just for official groups.

src/vehicle_gui.cpp
fmang: @michicc That wouldn’t be enough. I need to store the result of ShowQueryString (i.e. the new group name) and reuse it in the callback. CMD_ADD_VEHICLE_GROUP does not take that kind of argument, which is why I’m thinking a lambda capture would be well-suited.

fmang:
By combining the idea of making the pseudo-list of vehicles sharing orders always available, adding it to the vehicle view window, and reusing for managing groups, how about making its control+click open the vehicle list window and preselect the vehicle group? When the vehicle has a group, that would be a great way of having a “See group” feature will all the power of the vehicle list window. However, if the vehicle has no group, selecting the default group is not helpful. Automatically creating a group in that context would be unexpected, but I guess for that case the “Create group” action from pseudo-list would be the right way.

However, even after some more thought, I still think my current proposal holds. For a workflow where every vehicle is exclusively named through groups, and where groups are mainly used for naming, the current Rename button is useless, and renaming the group is the only way to properly rename the vehicle. In that case, groups are basically shared names, so extending the Rename button does not feel like a hijack to me at all.

fmang:
I’ve looked into the way callback works and they look quite difficult to use. I could copy the behavior of group_gui.cpp and trigger CMD_CREATE_GROUP (or in this case CMD_ADD_VEHICLE_GROUP) when the button is clicked, and from the callback of the command open the query string window. However, I’d get the same unwanted behavior as for group creation from VehicleGroupWindow: cancelling the group creation leaves an unnamed group that should not have been created in the first place. I have considered other options, the best being adding a dedicated command, but in reality what we’d need are stateful callbacks (lambdas).

I will be closing this pull request for now and will focus on adding the *Create group* action, but the above issue applies there too. I would need some guidance.

@2TallTyler Is replacing the command callback type by an std::function something the development team considers?

fmang:
Never mind, the *Create group* action would miss one important use case: renaming. Most often, I rename vehicle groups after changing their orders, in which case I have the vehicle view window open, but rarely the vehicle groups one.

I will apply everyone’s suggestions, and mimic the current callback flow from the vehicle group window.

btzy:
A a high level, I think the main debatable bit is whether the game should be made opinionated that vehicles sharing orders should also share a group.  If we merge the feature as proposed (i.e. that Ctrl-clicking the pencil button will move all vehicles sharing orders into the same group), then we're being opinionated, and so it follows that a number of actions that previously only modifies shared orders should also modify the group of a vehicle.  For example, from the orders list of a vehicle, if we are holding the "Go to" tool and Ctrl-click another vehicle, we will end up sharing orders with that other vehicle - should we also change the group of the current vehicle to match the group of that vehicle then?

I don't think we should be opinionated about whether vehicles sharing orders should also share a group.

btzy:
Actually, with the two extra buttons from my previous comment, we could make the green button normally create a group with only a the current vehicle (assuming that the current vehicle has no group yet), and Ctrl-clicking the green button would create a group containing all vehicles sharing orders with the current vehicle.  This is more consistent as most tools that support actions for shared orders hide those functionalities behind a Ctrl-click too.

btzy:
Or maybe the buttons could be better placed in the bottom bar instead of the title bar.

fmang:
@btzy Thank you for your feedback!

You made good points regarding groups and shared orders, and I must admit automatically adding vehicles sharing orders to the new group is probably a mistake since it breaks some workflows. I’ll rectify the pull request to add only the current vehicle. This still fits my use case though: I always name my vehicles through groups before Ctrl+duplicating them, so the new groups would only contain one vehicle anyway.

The one opinion I’m fervent about is that vehicles inherit the name of their group, which is why I find it odd that naming a vehicle’s  group is so much more complicated than naming the vehicle itself. In fact, shared orders are irrelevant here: it’s all about changing the name that appears on vehicles, and using the same button for that same effect, though the action behind the scenes is different. The automatic group creation feels natural to me because it’s the only way to achieve the goal of arbitarily naming a vehicle through a group.

In parallel, regardless of remaning, we seem to share the idea that it would be nice to be able to open the group window from the vehicle view window. However, I do not agree that the *Open group* button (the green button) should create a group for ungrouped vehicles. If we’re to open the group window, the Ungrouped vehicles meta-group should be selected instead. In any case, the group window should automatically scroll to and highlight the vehicle. From there, the user could easily drag it and drop into onto an existing group, or onto the New group icon.

What initially brought the topic of the *Vehicles with shared orders* list window is the extra action in JGRPP to create a group from that list, but that’s possibly misleading. I actually don’t think we need that window accessible directly from the vehicle window. Since it’s about shared orders, it makes sense to open it from the Orders window.

2TallTyler:
I'm skeptical of the concern that vehicles sharing orders might not be part of the same group. What usecase would there be for adding some, but not all, shared-order vehicles to a group?

btzy:
@fmang I think I saw the green button as more of a _Groupify_ button rather than an _Open Group_ button.  In that context, clicking it when the vehicle is ungrouped should make a new group and put it in there, but clicking it when the vehicle is already in a group is a sort of failure situation since a vehicle can't be in multiple groups at once, and so the game highlights the problem by opening the group window to that particular group.  But yeah what you're proposing for the green button makes sense to me too.

I personally find it odd that the "rename" button renames the group instead of the vehicle, but then I never give my vehicles individual names, and I only group vehicles on important routes.  I think it is because every other button on the vehicle GUI does things _about this vehicle_ (e.g. go to _this vehicle_'s location, send _this vehicle_ to depot, stop/start _this vehicle_), but your proposed action on the pencil button is another layer deeper (rename the _group of this vehicle_).  I guess it might be reasonable for someone like you who uses groups heavily.

@2TallTyler One reason could be having multiple models of rolling stock in the same route, to simulate real life where companies don't just retire all their rolling stock and replace them with one single new model at the same time.  These would all share orders, but could be in different groups to facilitate separate vehicle replacement regimes.  Another situation could be having separate trains transporting pax and mail on the same route.

fmang:
@btzy What you’re saying about grouping only important routes is very close to what I’ve ended up doing. Since groups are time-consuming and individual vehicle names are unflexible, I mostly gave up naming my vehicles at all. But then I get confused about what each vehicle does in a vehicle list, so I eventually name my main routes when such issue arises.

This Reddit post quite describes the problem:
https://www.reddit.com/r/openttd/comments/5sait6/how_do_you_guys_name_andor_group_your_vehicles/

Let’s call these the 4 stages of vehicle naming:

1. Naming vehicles individually.
2. Naming vehicles through groups because individual names get hard to manage.
3. Giving up on groups because they are time-consuming, and leaving all vehicles unnamed.
4. Creating a few groups anyway because having too many unnamed vehicles gets confusing.

If by any chance you’ve experienced the same stages, don’t you think that Ctrl+renaming would have let you stick at stage 2?

I would love an *Open group* button too, but I wouldn’t use it anywhere near as much as Ctrl+Rename.

btzy:
Not really.  I'm not concerned about naming apart from some special long-distance routes, and so grouping for me is mainly a tool for easier management (e.g. vehicle replacement; finding a vehicle on a certain route).

I usually have the "Group by" dropdown set to "Shared orders" in all the vehicle list windows, and that is usually good enough to find the correct vehicles when looking at the list of vehicles stopping at a given station.

2TallTyler:
In my opinion, this was superseded by #10890. This is still a few clicks easier, but it seems to be a niche use that doesn't seem to be worth the extra complication. 

Regarding the opinionation of whether groups == shared orders, there's been occasional discussion about eventually converting shared order groups to lines that vehicles are assigned to, like in Transport Fever. This is another reason that I'm hesitant to add this feature, because if we do people will expect us to support it forever. 

I am closing this PR to clean things up around here, but if others disagree and want this PR to happen, please don't be shy. 😄

fmang:
All right! Thank you for considering the feature.

PR #10851, opened 2023-05-20
Change: Use "Via-Destination-Source" as default station cargodist display

2TallTyler:
## Motivation / Problem

When using Cargo Distribution, waiting cargo at the station has several display options to help the player understand its destination. The setting changes the nesting order.

When a station has many destinations, as in a typical passenger network, the default "Source-Via-Destination" is absolutely useless for determining where cargo needs to go and by extension, which routes may need more capacity.

station-source

## Description

We often suggest players use Via-Destination-Source instead to better understand their network.

station-via

Let's make that the default for new players.

Existing players will not have this setting change. It only affects new installations where no config file exists.

If cargo distribution is set to manual, there is no nesting if all cargo is from that station...

no-cargodist

...and if cargo is transferred from elsewhere, it says "from" regardless of the setting.

no-cargodist-transferred

## Checklist for review


EmperorJake:
It wasn't the default already? I assumed it was since it’s what I have it set to 99% of the time

PeterN:
This will only affect players who have don't already have a game configuration. Worse, the setting does not appear in the settings window -- I didn't know it was configurable.

PeterN:
I wonder what's more useful, Via-Destination-Source or Destination-Via-Source?

Brickblock1:
> I wonder what's more useful, Via-Destination-Source or Destination-Via-Source?

Imo Via-Destination-source is a lot better as Destination is at times useless

2TallTyler:
Destination is pretty much useless if there is more than one route to it, as is common with passenger CargoDist networks. Via tells you the next hop, which is usually what you want to know when trying to determine which route capacity is responsible for the hundreds or thousands of passengers waiting. 😛

PeterN (approved):

PR #10854, opened 2023-05-21
Codechange: replace (f)printf/f(puts) with fmt::print

rubidium42:
## Motivation / Problem

Old style printf formatting, that is being replaced by C++-style formatting.

## Description

Simply replace `printf` with `fmt::print`, `fprintf(stdout` with `fmt::print(`, `fprintf(fh` with `fmt::print(fh`, `puts("...")` with `fmt::print("...\n")` and `fputs("...", fh)` with `fmt::print(fh, "...")`, and in some cases replace it with a more appropriate one like `Debug(...`. These replacements are all paired with the obligatory formatting string updates.
Furthermore add safeguards to prevent these functions from being used in the future.

## Limitations

None as far as I know. Well.. strgen/settingsgen always print `$Revision$` as version, which is wrong. But that's for a future PR, as changing that would also mean recompiling the settings and languages more often during development.

## Checklist for review


src/debug.cpp
```diff
@@ -127,19 +127,19 @@ void DebugPrint(const char *level, const std::string &message)
...
 		if (f == nullptr) return;
 
-		fprintf(f, "%s\n", message.c_str());
+		fmt::print(f, "{}\n", message);
 		fflush(f);
 #endif
 	} else {
 		std::string msg = fmt::format("{}dbg: [{}] {}\n", GetLogPrefix(), level, message);
-		fputs(msg.c_str(), stderr);
+		fmt::print(stderr, "{}", msg);
```
glx22: Could merge with `fmt::format` line

glx22 (approved):

PR #10858, opened 2023-05-22
Codechange: Use is valid cargo helper functions.

PeterN:
## Motivation / Problem

Lots of tests against CT_INVALID or INVALID_CARGO. One or both of these are potentially wrong, but that's for a later date.

## Description

Replace tests with `IsValidCargoID()` or `IsValidCargoType()` (now used once, as opposed to never).

## Checklist for review

### DIFF
--- a/src/script/api/script_cargolist.cpp
+++ b/src/script/api/script_cargolist.cpp
@@ -31,7 +31,7 @@ ScriptCargoList_IndustryAccepting::ScriptCargoList_IndustryAccepting(IndustryID 
 	Industry *ind = ::Industry::Get(industry_id);
 	for (uint i = 0; i < lengthof(ind->accepts_cargo); i++) {
 		CargoID cargo_id = ind->accepts_cargo[i];
-		if (cargo_id != CT_INVALID) {
+		if (::IsValidCargoID(cargo_id)) {
 			this->AddItem(cargo_id);
 		}
 	}
@@ -44,7 +44,7 @@ ScriptCargoList_IndustryProducing::ScriptCargoList_IndustryProducing(IndustryID 
 	Industry *ind = ::Industry::Get(industry_id);
 	for (uint i = 0; i < lengthof(ind->produced_cargo); i++) {
 		CargoID cargo_id = ind->produced_cargo[i];
-		if (cargo_id != CT_INVALID) {
+		if (::IsValidCargoID(cargo_id)) {
 			this->AddItem(cargo_id);
 		}
 	}
--- a/src/script/api/script_vehicle.cpp
+++ b/src/script/api/script_vehicle.cpp
@@ -75,7 +75,7 @@
 {
 	EnforceCompanyModeValid(VEHICLE_INVALID);
 	EnforcePrecondition(VEHICLE_INVALID, ScriptEngine::IsBuildable(engine_id));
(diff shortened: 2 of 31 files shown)


src/script/api/script_industry.hpp
```diff
@@ -246,7 +246,7 @@ class ScriptIndustry : public ScriptObject {
 	 * @param industry_id The index of the industry.
 	 * @param cargo_type The cargo to query, or CT_INVALID to query latest of all accepted cargoes.
 	 * @pre IsValidIndustry(industry_id).
-	 * @pre IsValidCargo(cargo_type) || cargo_type == CT_INVALID.
+	 * @pre IsValidCargo(cargo_type) || !IsValidCargoID(cargo_type).
```
2TallTyler: This looks odd. Requiring a cargo to be either valid or invalid... Maybe the old comment was better, to specify that if it was invalid, it was a known value instead of...what else would it be? Undefined?

src/script/api/script_vehicle.cpp
```diff
@@ -75,7 +75,7 @@
 {
 	EnforceCompanyModeValid(VEHICLE_INVALID);
 	EnforcePrecondition(VEHICLE_INVALID, ScriptEngine::IsBuildable(engine_id));
-	EnforcePrecondition(VEHICLE_INVALID, cargo == CT_INVALID || ScriptCargo::IsValidCargo(cargo));
+	EnforcePrecondition(VEHICLE_INVALID, !IsValidCargoID(cargo) || ScriptCargo::IsValidCargo(cargo));
```
2TallTyler: Same as my comment above, although the other one is documentation so it's probably more critical for it to make sense to humans.

src/script/api/script_industry.hpp
PeterN: Yeah, it's weird and makes no mention of a 3rd state that is neither valid or invalid... Dunno!

rubidium42: Look at line 247. CT_INVALID is a special case in the API. So replacing this with `!IsValidCargoID(cargo_type)` is not the same, as it also allows other garbage that isn't `CT_INVALID`. After all, the script can pass any (integer) garbage as parameter.

2TallTyler: Maybe negate the helper function to `IsInvalidCargoID()`? Or rename CT_INVALID to something else to identify it as a special case and not just a cargo that is not valid...CT_NONE, perhaps?

glx22: `CT_INVALID` works well for a special case, and changing stuff like that in the API would need compatibility layers everywhere.

PeterN: I've reverted this documentation change. The behaviour was never changed.

src/script/api/script_cargolist.cpp
```diff
@@ -31,7 +31,7 @@ ScriptCargoList_IndustryAccepting::ScriptCargoList_IndustryAccepting(IndustryID
 	Industry *ind = ::Industry::Get(industry_id);
 	for (uint i = 0; i < lengthof(ind->accepts_cargo); i++) {
 		CargoID cargo_id = ind->accepts_cargo[i];
-		if (cargo_id != CT_INVALID) {
+		if (IsValidCargoID(cargo_id)) {
```
glx22:
```suggestion
		if (::IsValidCargoID(cargo_id)) {
```
IIRC we use `::` when calling non script functions

glx22 (approved):

PR #10859, opened 2023-05-22
Codechange: use fmt::format and time conversions over "custom" implementation

rubidium42:
## Motivation / Problem

Formatting using `strftime` and lots of special casing around it to get it to work on all platforms with no warnings.
Formatting into a buffer and then into something else when talking about time.

## Description

Replace the `strftime` with `fmt::format` or move it into another `fmt::format`/`fmt::print`. 
Essentially inline the call to either `fmt::gmtime` or `fmt::localtime` with `time(nullptr)` as paramater.

## Limitations

I would like to have used `std::chrono` itself, but `std::chrono::system_clock::now()` seems to be the hardware clock, not the actual local time. That means that time is essentially neither UTC nor local time, and you have to mess with time zones.
With C++20 there is a `std::chrono::utc_clock` and things to convert to local time, but we do not have that yet.

CodeQL is likely going to warn about using the "unsafe" `gmtime`/`localtime`, but the actual implementation in fmt is quite nice, as it first tries `gmtime_r`, then `gmtime_s` and finally `gmtime` in the compilation. Arguably `gmtime` will be optimised away, but not in debug mode... so it lingers in the object file without actually being called when `gmtime_r` exists. The same holds for `localtime`.

## Checklist for review


TrueBrain (approved):
I always assume you tested this (at least, most of them), so YOLO!

PR #10861, opened 2023-05-23
Codechange: Use correct invalid cargo constants.

PeterN:
## Motivation / Problem

Both CT_INVALID and INVALID_CARGO have the value 0xFF but refer to different things. They have been mixed up in the past.

## Description

Rename INVALID_CARGO to INVALID_CARGO_BITNUM, and use CT_INVALID consistently, unless INVALID_CARGO_BITNUM is actually intended.

## Checklist for review

### DIFF
--- a/src/cargotype.cpp
+++ b/src/cargotype.cpp
@@ -44,7 +44,7 @@ void SetupCargoForClimate(LandscapeID l)
 	/* Reset and disable all cargo types */
 	for (CargoID i = 0; i < lengthof(CargoSpec::array); i++) {
 		*CargoSpec::Get(i) = {};
-		CargoSpec::Get(i)->bitnum = INVALID_CARGO;
+		CargoSpec::Get(i)->bitnum = INVALID_CARGO_BITNUM;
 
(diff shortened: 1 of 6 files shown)


LordAro:
Perhaps rename CT_INVALID to `CARGO_TYPE_INVALID` while you're at it?

PeterN:
I considered renaming it but it does appear in the script API, so I'm not sure it's feasible (but maybe that's a separate enum)

Also it's used in a lot of tables which would no longer be (barely) aligned ;)

But that possibility is one of the reasons for using `IsValidCargoID()` everywhere. There are other changes to cargo types I have in the works as well which involve separating CargoID from CT_xxx, but small steps.

PeterN:
Oh and as the warnings (that didn't show for me) say, there's also the CargoID / CargoType discrepancy. I guess I should fix that before this change :-)

2TallTyler (approved):
I haven't verified that you've used the right type for each use, but I assume you (and the warnings you mention) have. 🙂

PR #10862, opened 2023-05-23
Codechange: Base CargoArray off std::array

PeterN:
## Motivation / Problem

CargoArray has a C array as a member and implements operators to allow access to it. 

It also uses MemSet to initialize itself.

## Description

Remove the array member and base CargoArray off std::array instead. This allows use of std::array's standard operators, and also allows value initialization instead of using MemSet.

The member functions that loop through the array have been replaced with iterator-based code instead.,

Second commit makes use of CargoArray::GetCount() where I could see counting being done.

## Checklist for review

### DIFF
--- a/src/train_gui.cpp
+++ b/src/train_gui.cpp
@@ -322,10 +322,8 @@ int GetTrainDetailsWndVScroll(VehicleID veh_id, TrainDetailsWindowTabs det_tab)
 	int num = 0;
 
 	if (det_tab == TDW_TAB_TOTALS) { // Total cargo tab
-		CargoArray act_cargo{};
 		CargoArray max_cargo{};
 		for (const Vehicle *v = Vehicle::Get(veh_id); v != nullptr; v = v->Next()) {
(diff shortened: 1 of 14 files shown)


src/train_gui.cpp
```diff
@@ -322,19 +322,14 @@ int GetTrainDetailsWndVScroll(VehicleID veh_id, TrainDetailsWindowTabs det_tab)
 	int num = 0;
 
 	if (det_tab == TDW_TAB_TOTALS) { // Total cargo tab
-		CargoArray act_cargo;
-		CargoArray max_cargo;
+		CargoArray act_cargo{};
```
rubidium42: Since we're only counting what's in `max_cargo`, is `act_cargo` actually needed? Can't it be removed completely?

PeterN: Yeah I didn't really look at what it was doing with it :)

rubidium42 (approved):

PR #10871, opened 2023-05-25
Codechange: replace C-style idioms with C++-style for file name generation

rubidium42:
## Motivation / Problem

C-style strings.

## Description

Use C++ idioms over 'char *buffer, const char *last` for some things related to file names.

## Checklist for review

### DIFF
--- a/src/saveload/oldloader.cpp
+++ b/src/saveload/oldloader.cpp
@@ -24,6 +24,8 @@
 
 static const int TTO_HEADER_SIZE = 41;
 static const int TTD_HEADER_SIZE = 49;
+/** The size of the checksum in the name/header of the TTD/TTO savegames. */
+static const int HEADER_CHECKSUM_SIZE = 2;
 
 uint32 _bump_assert_value;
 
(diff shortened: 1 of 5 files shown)


src/saveload/oldloader.cpp
```diff
@@ -24,6 +24,8 @@
 
 static const int TTO_HEADER_SIZE = 41;
 static const int TTD_HEADER_SIZE = 49;
+/** The size of the checksum in the name/header of the TTD/TTO savegames. */
+static const int HEADER_CHECKSUM_SIZE = 2;
```
LordAro, line 28: length, rather than size?

src/saveload/oldloader.cpp
```diff
@@ -314,16 +290,13 @@ bool LoadOldSaveGame(const std::string &file)
...
 
-	if (f == nullptr) {
-		*title = '\0';
-		return;
-	}
-
-	DetermineOldSavegameType(f, title, last);
-
+	std::string name;
+	std::tie(std::ignore, name) = DetermineOldSavegameTypeAndName(f);
```
LordAro, line 299: why ignore? shouldn't it be validated (or at least asserted?)

src/saveload/oldloader.cpp
rubidium42: Does that mean `TTO/D_HEADER_LENGTH` too? After all, the checksum is part of the header.
I made it consistent with the names that were already there.

src/saveload/oldloader.cpp
rubidium42: The type has no value when getting the savegame name; the name is always valid. 
You could think to return something like "broken" when the type is invalid. However, that is practically what the function already returns, with slightly more information about the reason.

LordAro (approved):

PR #10872, opened 2023-05-25
Change: Make cargo flow legend only show defined cargo.

PeterN:
## Motivation / Problem

Cargo flow legend window has a button for every possible cargo type (64) instead of only the currently defined cargo types. (11 on temperate by default.)

image

## Description

Show buttons only for the currently defined cargo types instead, removing the large mostly-empty grid of cargo buttons when not using a complex industry set.

image

Caveat: Cargo types are now first drawn down by column instead of across by rows, and in "sort" order instead of index order. This happens to match the cargo legend on the minimap.

## Checklist for review

### DIFF
--- a/src/linkgraph/linkgraph_gui.cpp
+++ b/src/linkgraph/linkgraph_gui.cpp
@@ -474,30 +474,32 @@ NWidgetBase *MakeSaturationLegendLinkGraphGUI(int *biggest_index)
 
 NWidgetBase *MakeCargoesLegendLinkGraphGUI(int *biggest_index)
 {
-	static const uint ENTRIES_PER_ROW = CeilDiv(NUM_CARGO, 5);
-	NWidgetVertical *panel = new NWidgetVertical(NC_EQUALSIZE);
-	NWidgetHorizontal *row = nullptr;
-	for (uint i = 0; i < NUM_CARGO; ++i) {
-		if (i % ENTRIES_PER_ROW == 0) {
-			if (row) panel->Add(row);
-			row = new NWidgetHorizontal(NC_EQUALSIZE);
+	uint num_cargo = static_cast<uint>(_sorted_cargo_specs.size());
+	static const uint ENTRIES_PER_COL = 5;
+	NWidgetHorizontal *panel = new NWidgetHorizontal(NC_EQUALSIZE);
+	NWidgetVertical *col = nullptr;
+
+	for (uint i = 0; i < num_cargo; ++i) {
+		if (i % ENTRIES_PER_COL == 0) {
+			if (col != nullptr) panel->Add(col);
+			col = new NWidgetVertical(NC_EQUALSIZE);
 		}
 		NWidgetBackground * wid = new NWidgetBackground(WWT_PANEL, COLOUR_GREY, i + WID_LGL_CARGO_FIRST);
 		wid->SetMinimalSize(25, 0);
 		wid->SetMinimalTextLines(1, 0, FS_SMALL);
 		wid->SetFill(1, 1);
 		wid->SetResize(0, 0);
-		row->Add(wid);
+		col->Add(wid);
 	}
 	/* Fill up last row */
-	for (uint i = 0; i < 4 - (NUM_CARGO - 1) % 5; ++i) {
+	for (uint i = num_cargo; i < Ceil(num_cargo, ENTRIES_PER_COL); ++i) {
 		NWidgetSpacer *spc = new NWidgetSpacer(25, 0);
 		spc->SetMinimalTextLines(1, 0, FS_SMALL);
 		spc->SetFill(1, 1);
 		spc->SetResize(0, 0);
-		row->Add(spc);
+		col->Add(spc);
 	}
-	panel->Add(row);
+	panel->Add(col);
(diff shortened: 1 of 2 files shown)


LordAro:
How does it look with something like FIRS?

PeterN:
Almost the same:

image

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -474,30 +474,32 @@ NWidgetBase *MakeSaturationLegendLinkGraphGUI(int *biggest_index)
...
 		wid->SetMinimalSize(25, 0);
 		wid->SetMinimalTextLines(1, 0, FS_SMALL);
 		wid->SetFill(1, 1);
 		wid->SetResize(0, 0);
-		row->Add(wid);
+		col->Add(wid);
 	}
-	/* Fill up last row */
-	for (uint i = 0; i < 4 - (NUM_CARGO - 1) % 5; ++i) {
+	/* Fill up last col */
```
2TallTyler: `column`?

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -474,30 +474,32 @@ NWidgetBase *MakeSaturationLegendLinkGraphGUI(int *biggest_index)
...
+	for (uint i = 0; i < 4 - (num_cargo - 1) % 5; ++i) {
```
2TallTyler: Could one or both of these magic numbers be based on `ENTRIES_PER_COLUMN`?

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -728,11 +730,13 @@ void LinkGraphLegendWindow::OnClick(Point pt, int widget, int click_count)
  */
 void LinkGraphLegendWindow::OnInvalidateData(int data, bool gui_scope)
 {
+	if (this->num_cargo != _sorted_cargo_specs.size()) {
+		this->Close();
```
2TallTyler, line 726: Close? Not re-init or something? Or are we in some sort of failure state if we get here?

2TallTyler (commented):
I love it. The current window is so unnecessarily huge. 

What is the sort we're using now? Sorted by initial value, like in the payment rates graph?

src/linkgraph/linkgraph_gui.cpp
PeterN: It needs to be deleted and reopened, not just reinited, and I'm not sure on the semantics of opening the current window while closing it at the same time.

And the only way that this could happen if someone is changing NewGRF mid-game which for industries and cargo types is going to break anyway, so I don't think it's important enough to be worth dealing with.

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -474,30 +474,32 @@ NWidgetBase *MakeSaturationLegendLinkGraphGUI(int *biggest_index)
...
-			if (row) panel->Add(row);
-			row = new NWidgetHorizontal(NC_EQUALSIZE);
+	size_t num_cargo = _sorted_cargo_specs.size();
+	static const uint ENTRIES_PER_COL = 5;
+	NWidgetHorizontal *panel = new NWidgetHorizontal(NC_EQUALSIZE);
+	NWidgetVertical *col = nullptr;
+
+	for (uint i = 0; i < num_cargo; ++i) {
+		if (i % ENTRIES_PER_COL == 0) {
+			if (col) panel->Add(col);
```
rubidium42:
```suggestion
			if (col != nullptr) panel->Add(col);
```

src/linkgraph/linkgraph_gui.cpp
rubidium42: Closing sounds perfectly fine to me, given it's due to the basically unsupported changing of NewGRFs.

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -474,30 +474,32 @@ NWidgetBase *MakeSaturationLegendLinkGraphGUI(int *biggest_index)
...
 		wid->SetMinimalSize(25, 0);
 		wid->SetMinimalTextLines(1, 0, FS_SMALL);
 		wid->SetFill(1, 1);
 		wid->SetResize(0, 0);
-		row->Add(wid);
+		col->Add(wid);
 	}
 	/* Fill up last row */
-	for (uint i = 0; i < 4 - (NUM_CARGO - 1) % 5; ++i) {
+	for (uint i = num_cargo; i < Ceil(num_cargo, ENTRIES_PER_COL); ++i) {
```
rubidium42, line 495: Ugh... `Ceil` isn't the most descriptive name of what that function does. Though that's outside of the scope of this PR, but I first thought... how can this function?!? Without having seen the implementation/documentation of the function.
It's actually a `RoundToNextMultiple`, and that would have made reviewing this bit a lot easier.

rubidium42 (dismissed):

rubidium42 (dismissed):

PeterN:
Needs reviewing again, in the last push I removed a few checks for disabled widgets, and fixed setting the initial widget lowered state. Should've been separate but I fixed up by habit.

2TallTyler (approved):

PR #10875, opened 2023-05-26
Feature: Highlight waypoint tiles when building/selecting distant waypoint.

PeterN:
## Motivation / Problem

When distant-joining waypoints it's not obvious from the map which is the correct waypoint, you need to check the signs.

## Description

Similar to rail stations, highlight the waypoint tiles when selecting with distant join, or when adjacent to a waypoint.

Feature backport from JGRPP, with some stuff about road waypoints ripped out, so it's quite a bit different.

image

image

## Checklist for review

### DIFF
--- a/src/rail_gui.cpp
+++ b/src/rail_gui.cpp
@@ -422,6 +422,7 @@ struct BuildRailToolbarWindow : Window {
 	void Close() override
 	{
 		if (this->IsWidgetLowered(WID_RAT_BUILD_STATION)) SetViewportCatchmentStation(nullptr, true);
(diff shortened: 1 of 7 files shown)


TrueBrain (approved):
I really do not like the big "Coverage" button, but I also don't have suggestions.

Tested it a bit, does what it says it does, and I kinda like the feature on its own. So yeah, let's go!

PeterN:
FWIW, without this change you still get that big button, however it's just an empty panel that does nothing instead.

TrueBrain:
> FWIW, without this change you still get that big button, however it's just an empty panel that does nothing instead.

Haha, that is fair :)

PR #10876, opened 2023-05-26
Fix d9a04ba446: Ensure all MD5Hash are initialized.

PeterN:
## Motivation / Problem

In some places an MD5Hash is xored with another hash, without ensure the original array is clear. This results in false MD5 mismatches.

## Description

Ensure all MD5Hashes are value initialized. Not all instances need to be initialized as often they are copied or written to, but doing all ensures no surprises.

## Checklist for review

### DIFF
--- a/src/3rdparty/md5/md5.h
+++ b/src/3rdparty/md5/md5.h
@@ -57,20 +57,20 @@
 static const size_t MD5_HASH_BYTES = 16;
 
 /** Container for storing a MD5 hash/checksum/digest. */
-using MD5Hash = std::array<byte, MD5_HASH_BYTES>;
+struct MD5Hash : std::array<byte, MD5_HASH_BYTES> {
+	MD5Hash() : std::array<byte, MD5_HASH_BYTES>{} {}
 
-/**
- * Exclusively-or one hash into another hash.
- * @param lhs The hash to exclusively-or into.
- * @param rhs The hash to exclusively-or with.
- * @return Reference to \c lhs hash.
- */
-inline MD5Hash &operator^=(MD5Hash &lhs, const MD5Hash &rhs)
-{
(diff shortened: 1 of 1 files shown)


glx22 (dismissed):

TrueBrain:
I mentioned it on Discord, but this feels like this will go wrong a lot more often in the future. It isn't all that clear you need to initialize MD5Hash. Makes me wonder if we shouldn't just make it a class and put an explicit ctor on it. But that might just be me being annoyed by PODs and C++ :P

PeterN:
Okay, as @TrueBrain mentioned, I have now replaced the original commit by this which gives MD5Hash a default constructor that value initializes itself, so anyone using MD5Hash doesn't need to care about the underlying container.

glx22:
I think it would make sense to also put the xor operator inside the struct.

TrueBrain:
I would say that comment, well, both of them, aged like milk :D :D :D :D :)

(sorry, couldn't resist poking fun :P)

src/3rdparty/md5/md5.h
```diff
@@ -57,20 +57,20 @@
...
-	for (size_t i = 0; i < lhs.size(); i++) lhs[i] ^= rhs[i];
-	return lhs;
-}
-
+struct MD5Hash : std::array<byte, MD5_HASH_BYTES> {
+	MD5Hash() : std::array<byte, MD5_HASH_BYTES>{} {}
+
+	/**
+	 * Exclusively-or the given hash into this hash.
+	 * @param other The other has.
```
TrueBrain: has? HASH! :)

PeterN: I can't even just copy & paste Rubidium's code without getting critiqued! ;-)

TrueBrain (approved):

PR #10879, opened 2023-05-26
Change: Include style when getting font name.

PeterN:
## Motivation / Problem

With Freetype, family name and style name are separate members, and our GetFontName() only returns the family name. This means that the `font` console command and the survey data does not show the full detail:

image

## Description

This is fixed for Freetype by changing `GetFontName()` to return a std::string, and then including both family name and style name for Freetype.

With Win32 the style does not seem to be separated so it is already included.

I'm unsure how this shows with OSX. 

image

## Checklist for review

### DIFF
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -134,7 +134,7 @@ public:
 	 * Get the name of this font.
 	 * @return The name of the font.
 	 */
-	virtual const char *GetFontName() = 0;
+	virtual std::string GetFontName() = 0;
 
 	/**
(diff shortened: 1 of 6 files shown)


TrueBrain (dismissed):

PeterN:
The OSX fontcache does already include the style as part of the font name.

PeterN:
After some testing, this also fixes a bug where changing one font will reset the other fonts something else, because the style information is missing.

glx22 (approved):

PR #10890, opened 2023-05-29
Feature: Create group of vehicles from manage vehicle list button.

PeterN:
## Motivation / Problem

Often requested, often referred to JGRPP... the ability to create a group of vehicles from a vehicle list.

## Description

This PR adds a 'Create Group' option to 'Manage list' dropdown menu.

`CMD_ADD_VEHICLE_GROUP` is modified to accept a VehicleListIdentifier which is used if the single VehicleID is set to INVALID_VEHICLE. (This could probably be improved by having a VehicleListIdentifier that refers to only the single vehicle.)

image

(Not based on the version in JGRPP as that does not use the current command system.)

## Limitations

Vehicles in the list that are already in a group will also be moved to a new group. They mayn't be desirable?

## Checklist for review

### DIFF
--- a/src/group_cmd.cpp
+++ b/src/group_cmd.cpp
@@ -512,48 +512,63 @@ static void AddVehicleToGroup(Vehicle *v, GroupID new_g)
  * @param add_shared Add shared vehicles as well.
  * @return the cost of this operation or an error
  */
-std::tuple<CommandCost, GroupID> CmdAddVehicleGroup(DoCommandFlag flags, GroupID group_id, VehicleID veh_id, bool add_shared)
+std::tuple<CommandCost, GroupID> CmdAddVehicleGroup(DoCommandFlag flags, GroupID group_id, VehicleID veh_id, bool add_shared, const VehicleListIdentifier &vli)
 {
-	Vehicle *v = Vehicle::GetIfValid(veh_id);
 	GroupID new_g = group_id;
+	if (!Group::IsValidID(new_g) && !IsDefaultGroupID(new_g) && new_g != NEW_GROUP) return { CMD_ERROR, INVALID_GROUP };
 
-	if (v == nullptr || (!Group::IsValidID(new_g) && !IsDefaultGroupID(new_g) && new_g != NEW_GROUP)) return { CMD_ERROR, INVALID_GROUP };
+	VehicleList list;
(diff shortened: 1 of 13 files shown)


PeterN:
Alright, where am I supposed to put an EndianBufferWriter/Reader to be shared by multiple _cmd.h files...

rubidium42:
> Alright, where am I supposed to put an EndianBufferWriter/Reader to be shared by multiple _cmd.h files...

`vehiclelist_cmd.h` and then include that in the other _cmd.h files?

src/group_cmd.cpp
```diff
@@ -512,48 +512,63 @@ static void AddVehicleToGroup(Vehicle *v, GroupID new_g)
...
+		if (!GenerateVehicleSortList(&list, vli) || list.empty()) return { CMD_ERROR, INVALID_GROUP };
+	} else {
+		Vehicle *v = Vehicle::GetIfValid(veh_id);
+		if (v == nullptr) return { CMD_ERROR, INVALID_GROUP };
+		list.push_back(v);
+	}
+
+	VehicleType vtype = list.front()->type;
+	for (const Vehicle *v : list) {
+		if (v->owner != _current_company || !v->IsPrimaryVehicle()) return { CMD_ERROR, INVALID_GROUP };
```
rubidium42, line 531: Most of the checks in this function are just to prevent bad actors (assume all information is invalid), or issues when the state changes in between, right?
However, I think I can, as user, trigger this error when I get the list of vehicles of an oil rig. If you're the only company it works fine, but when another company joins you're in trouble. It might be hard to see that there is a vehicle from another company in there, so would it be useful to have a specific error message in case the owner is not the current company hinting about the fact that there are vehicles from another company in the list?

Or should it maybe just remove the vehicle of other companies from the list, so it's easier to work with for shared stations?

rubidium42 (commented):
The rest looks okay to me.

src/group_cmd.cpp
PeterN: Yeah, ignoring wrong owners seems like the best way.

PeterN: Well, turns out there is no "Manage Vehicles" menu on a shared station...

rubidium42 (approved):

PR #10709, opened 2023-04-24
Remove: buying/selling/owning company shares

TrueBrain:
## Motivation / Problem

Right, where to start.

First off, a personal opinion: this feature is just the worst. 

Let me explain.

Shares, as implemented, is just very poor, and makes no sense at all. You can buy shares of another company, and sell it. But when you sell it, nobody else has to buy it. So you either generate money from thin air, or push it in the aether. That is not how that works.
But also .. why 25%? Why not .. any percentage? Why not an amount?
From an economical point of view, it is just very .. weird.

Codewise, it is just a constant stream of problems. So many bug-reports over the years, either because we calculate the shares with the company value, or we don't, or it is mismatching, or or or or .. it is just a nightmare.

Mostly, you can generate infinite money with this. #7691 as example.

Now, what does it add to the player?

ABSOLUTELY NOTHING.

In fact, we have it disabled by default. Additionally, you could buy out an AI company in full, but not a human company ..
It is just so weird.

And the list goes on ... and on ...

It is just a bad feature.

So that gives us two choices: keep continue to support this bad feature, in the hope someone improves it. Or say: fuck it, let's remove it, and if someone comes up with a better idea to do shares, let's add that. This PR goes for the second.

## Description

Remove the option to buy/sell shares. Bankruptcy is still in there, of course.

Mind you: as savegames contain table-headers these days, there is no need for a savegame bump. Simply no longer loading/saving the entries is sufficiently. There is need for a savegame bump between releases, to give users early feedback something is not right, but that is covered already by other PRs.

## Limitations

The money you invested is lost.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10891, opened 2023-05-30
Change: Simplify advertisement by removing small and medium options

ldpl:
I got inspired by #10709 /s

## Motivation / Problem

While having a way to spend money to increase ratings isn't bad in general the way it currently works is far from perfect. As it's very unclear what advertisement does and which stations are affected. Also vanilla game lacks a very useful function to set up a continuous advertisement. And improving it is complicated by the fact that there are 3 advertisement options for historical reasons. So, for example, making advertisement range configurable would require 3 settings, showing zone highlight needs 3 colours and so on. And while it can probably be done question is whether it is worth it and what does having 3 options even add to the game. As it currently is you have to plan for future advertisement when you build the station. And smaller advertisement options are cheaper so it encourages players to destroy towns to place stations closer. That's hardly a good thing and even large adv radius is usually too small imo. So I don't see any value in clogging the interface with options that provide questionable incentives.

## Description

This PR removes small and medium advertisement options and renames "large advertisement" to just "advertisement". That hardly affects the gameplay as you can still do advertisement in all the same cases, it just takes more money. And custom patchpacks that provide advertisement automation never even bothered to include small and medium options afaik and I don't remember anyone missing them.

## Limitations

- It totally breaks the game again.
- I forgot to deal with script api so drafting this for now.

## Checklist for review

* **This PR affects the GS/AI API? (label 'needs review: Script API')**
    * **ai_changelog.hpp, gs_changelog.hpp need updating.**
    * **The compatibility wrappers (compat_*.nut) need updating.**

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3519,20 +3519,16 @@ STR_LOCAL_AUTHORITY_ACTIONS_TOOLTIP                             :{BLACK}List of 
 STR_LOCAL_AUTHORITY_DO_IT_BUTTON                                :{BLACK}Do it
 STR_LOCAL_AUTHORITY_DO_IT_TOOLTIP                               :{BLACK}Carry out the highlighted action in the list above
 
-###length 8
-STR_LOCAL_AUTHORITY_ACTION_SMALL_ADVERTISING_CAMPAIGN           :Small advertising campaign
-STR_LOCAL_AUTHORITY_ACTION_MEDIUM_ADVERTISING_CAMPAIGN          :Medium advertising campaign
-STR_LOCAL_AUTHORITY_ACTION_LARGE_ADVERTISING_CAMPAIGN           :Large advertising campaign
+###length 6
+STR_LOCAL_AUTHORITY_ACTION_ADVERTISING_CAMPAIGN                 :Advertising campaign
 STR_LOCAL_AUTHORITY_ACTION_ROAD_RECONSTRUCTION                  :Fund local road reconstruction
 STR_LOCAL_AUTHORITY_ACTION_STATUE_OF_COMPANY                    :Build statue of company owner
 STR_LOCAL_AUTHORITY_ACTION_NEW_BUILDINGS                        :Fund new buildings
 STR_LOCAL_AUTHORITY_ACTION_EXCLUSIVE_TRANSPORT                  :Buy exclusive transport rights
 STR_LOCAL_AUTHORITY_ACTION_BRIBE                                :Bribe the local authority
 
-###length 8
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_SMALL_ADVERTISING            :{PUSH_COLOUR}{YELLOW}Initiate a small local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a small radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_MEDIUM_ADVERTISING           :{PUSH_COLOUR}{YELLOW}Initiate a medium local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a medium radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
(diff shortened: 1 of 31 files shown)


orudge:
From a financial perspective it seems unfortunate to remove the lower levels (which I have used when I recently played a proper OpenTTD game for the first time in eons!). But then I've generally always appreciated OpenTTD offering new options but not generally removing old ones...

orudge:
> Lowering it even further needs careful balance considerations, especially considering automated advertisement is an option in patch packs.

Surely that's the concern of the patch pack author, not the main OpenTTD project?

PeterN:
So you are removing something because it makes your custom client harder?

PeterN:
That was my takeaway from skim reading your rationale, as you talk about things that aren't in vanilla.

src/lang/english.txt
```diff
@@ -3519,20 +3519,16 @@ STR_LOCAL_AUTHORITY_ACTIONS_TOOLTIP                             :{BLACK}List of
...
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_MEDIUM_ADVERTISING           :{PUSH_COLOUR}{YELLOW}Initiate a medium local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a medium radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_LARGE_ADVERTISING            :{PUSH_COLOUR}{YELLOW}Initiate a large local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating in a large radius around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+###length 6
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_ADVERTISING            :{PUSH_COLOUR}{YELLOW}Initiate a local advertising campaign, to attract more passengers and cargo to your transport services.{}Provides a temporary boost to station rating around the town centre.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
```
2TallTyler, line 3531: Wrong spacing

2TallTyler (commented):
I'm all for making PRs as small as possible, but if the point of this is to enable new features like automatic advertising and something with zoning (previewing the area the advertising affects?), it should be proposed alongside those things. Maybe not in the same PR but opened simultaneously.

Unlike #10709, multiple radii doesn't cost much to maintain and doesn't allow an exploit. If it's a barrier to additional features, let's see those features as justification for this change. 😃

ldpl:
@PeterN I'm talking things that can be added to vanilla. In cmclient they're either already done or can't be added because of compatibility.

@2TallTyler That "exploit" got way more attention than it ever deserved :p Also, unlike #10709 this doesn't completely remove a gameplay mechanic ;)

Also on its own this PR is useful for:
- Simplifiying authority window and reducing it's height.
- Removing the encouragement to place stations too close to towns and so the need to destroy them and also reliance on perfect industry spawn to get the station within the small advertisement radius.

Notts90:
I think a stronger reason is required for removing functionality. As it stands I’d recommend leaving advertising as three separate options.

TrueBrain:
I too have a hard time seeing the point of this PR. How it reads to me:

- I want to remove two options so my patchpack becomes easier
- Maybe, if I ever feel like it, I might PR an addition which makes it deserve the removal of the two options (paraphrasing what you mentioned on Discord here)

That motivation makes little sense. I agree with @2TallTyler , make this commit part of the PR that adds the automation, and then you might be on to something. But on its own, this just doesn't make the cut (for me).

And the fencing with #10709 is silly. Not all things are equal. #10709 isn't a "now we can remove whatever we want if we feel like it because there is precedent!". No, every removal still deserves scrutiny, like we always do.

Personally, I always like the options here, as especially when I am low on money, it always feels like progress to slowly see bigger advertisements pop up. That just feels nice. Call me nostalgic.

So in short, my opinion: if you are up for actually improving advertisements, have at it! But make this commit part of that PR please :)

ldpl:
Ok, screw it. Those who want automatic advertisement can already use CityMania patchpack (https://citymania.org/downloads). And who want better gameplay can play on CityMania servers as they don't allow small and medium advertisement.

PR #10893, opened 2023-05-30
Codechange: replace C-style idioms with C++-style for OS specific file calls

rubidium42:
## Motivation / Problem

The internal API for the different OSes to implement checking whether a file is a root, a valid file or the free disk space use C-style strings.
Complicated function returning values by reference, instead of just using `std::optional`.

## Description

Replace `const char *` with `const std::string &`.
Split `FiosGetDescText` into `FiosGetCurrentPath` and `FiosGetDiskFreeSpace`.
Let them return values instead of using out parameter references.

## Checklist for review

### DIFF
--- a/src/fios.cpp
+++ b/src/fios.cpp
@@ -39,7 +39,6 @@ extern bool FiosIsRoot(const std::string &path);
 extern bool FiosIsValidFile(const std::string &path, const struct dirent *ent, struct stat *sb);
 extern bool FiosIsHiddenFile(const struct dirent *ent);
 extern void FiosGetDrives(FileList &file_list);
-extern bool FiosGetDiskFreeSpace(const char *path, uint64 *tot);
 
 /* get the name of an oldstyle savegame */
 extern void GetOldSaveGameName(const std::string &file, char *title, const char *last);
@@ -128,16 +127,11 @@ const FiosItem *FileList::FindItem(const std::string_view file)
 }
 
 /**
- * Get descriptive texts. Returns the path and free space
- * left on the device
- * @param path string describing the path
- * @param total_free total free space in megabytes, optional (can be nullptr)
- * @return StringID describing the path (free space or failure)
+ * Get the current path/working directory.
  */
-StringID FiosGetDescText(const char **path, uint64 *total_free)
+std::string FiosGetCurrentPath()
 {
-	*path = _fios_path->c_str();
-	return FiosGetDiskFreeSpace(*path, total_free) ? STR_SAVELOAD_BYTES_FREE : STR_ERROR_UNABLE_TO_READ_DRIVE;
+	return *_fios_path;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/fios.cpp
```diff
@@ -128,16 +127,11 @@ const FiosItem *FileList::FindItem(const std::string_view file)
...
 /**
- * Get descriptive texts. Returns the path and free space
- * left on the device
- * @param path string describing the path
- * @param total_free total free space in megabytes, optional (can be nullptr)
- * @return StringID describing the path (free space or failure)
+ * Get the current path/working directory.
  */
-StringID FiosGetDescText(const char **path, uint64 *total_free)
+std::string FiosGetCurrentPath()
```
TrueBrain, line 132: not a `string_view`?

src/os/os2/os2.cpp
```diff
@@ -92,32 +92,21 @@ void FiosGetDrives(FileList &file_list)
...
-	{
-		struct statvfs s;
-
-		if (statvfs(path, &s) != 0) return false;
-		free = (uint64)s.f_frsize * s.f_bavail;
-	}
-#endif
-	if (tot != nullptr) *tot = free;
-	return true;
+	if (statvfs(path.c_str(), &s) == 0) return (uint64_t)s.f_frsize * s.f_bavail;
```
TrueBrain: `static_cast` while at it?

src/os/windows/win32.cpp
```diff
@@ -213,16 +213,16 @@ bool FiosIsHiddenFile(const struct dirent *ent)
...
 
-bool FiosGetDiskFreeSpace(const char *path, uint64 *tot)
+std::optional<uint64_t> FiosGetDiskFreeSpace(const std::string &path)
 {
 	UINT sem = SetErrorMode(SEM_FAILCRITICALERRORS);  // disable 'no-disk' message box
 
 	ULARGE_INTEGER bytes_free;
 	bool retval = GetDiskFreeSpaceEx(OTTD2FS(path).c_str(), &bytes_free, nullptr, nullptr);
-	if (retval && tot != nullptr) *tot = bytes_free.QuadPart;
-
```
TrueBrain, line 223: Maybe leave a newline to give visual separation between the getting of diskspace and setting of error mode.

TrueBrain (commented):
As always, I assume you tested this :)

src/fios.cpp
rubidium42: No; it gets passed to `FiosGetDiskFreeSpace` which in its turn calls C-style OS APIs which want `'\0'`-terminated strings, which `std::string_view`s are not.

TrueBrain (dismissed):

src/os/os2/os2.cpp
```diff
@@ -92,40 +92,27 @@ void FiosGetDrives(FileList &file_list)
...
 }
 
-bool FiosGetDiskFreeSpace(const char *path, uint64 *tot)
+std::optional<uint64_t> FiosGetDiskFreeSpace(const std::string &path)
 {
 #ifndef __INNOTEK_LIBC__
 	struct diskfree_t free;
 	char drive = path[0] - 'A' + 1;
 
 	if (tot != nullptr && _getdiskfree(drive, &free) == 0) {
```
glx22:
```suggestion
	if (_getdiskfree(drive, &free) == 0) {
```

PeterN (approved):

PR #10895, opened 2023-05-31
Codechange: replace seprintf with fmt::format/std::to_string

rubidium42:
## Motivation / Problem

C-style strings.

## Description

Replace some `seprintf`s with `fmt::format` or `std::to_string`.
Replace some `const char *` parameters with `const std::string&` to reduce uses of `.c_str()`.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -2269,9 +2269,7 @@ DEF_CONSOLE_CMD(ConNewGRFProfile)
 				started++;
 
 				if (!grfids.empty()) grfids += ", ";
-				char grfidstr[12]{ 0 };
-				seprintf(grfidstr, lastof(grfidstr), "[%08X]", BSWAP32(pr.grffile->grfid));
-				grfids += grfidstr;
+				fmt::format_to(std::back_inserter(grfids), "[{%:08X}]", BSWAP32(pr.grffile->grfid));
 			}
 		}
 		if (started > 0) {
--- a/src/framerate_gui.cpp
+++ b/src/framerate_gui.cpp
@@ -1058,7 +1058,7 @@ void ConPrintFramerate()
 		"AI/GS scripts total",
 		"Game script",
 	};
-	char ai_name_buf[128];
+	std::string ai_name_buf;
 
 	static const PerformanceElement rate_elements[] = { PFE_GAMELOOP, PFE_DRAWING, PFE_VIDEO };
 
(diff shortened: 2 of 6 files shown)


src/console_cmds.cpp
```diff
@@ -2269,9 +2269,7 @@ DEF_CONSOLE_CMD(ConNewGRFProfile)
 				started++;
 
 				if (!grfids.empty()) grfids += ", ";
-				char grfidstr[12]{ 0 };
-				seprintf(grfidstr, lastof(grfidstr), "[%08X]", BSWAP32(pr.grffile->grfid));
-				grfids += grfidstr;
+				fmt::format_to(std::back_inserter(grfids), "[{%:08X}]", BSWAP32(pr.grffile->grfid));
```
LordAro, line 2272: Doesn't hex formatting do something strange, compared to printf-format ?

rubidium42: That's when there's a `#` involved to add `0x`. Without the `#` no `0x` get added, just like `%X` doesn't get a `0x` added automatically. They weird thing is that when you specify the length of the format, that includes the `0x` if `#` modifier is added to the `{:X}`.

src/framerate_gui.cpp
```diff
@@ -1077,11 +1077,11 @@ void ConPrintFramerate()
 	for (PerformanceElement e = PFE_FIRST; e < PFE_MAX; e++) {
 		auto &pf = _pf_data[e];
 		if (pf.num_valid == 0) continue;
-		const char *name;
+		std::string_view name;
 		if (e < PFE_AI0) {
 			name = MEASUREMENT_NAMES[e];
 		} else {
-			seprintf(ai_name_buf, lastof(ai_name_buf), "AI %d %s", e - PFE_AI0 + 1, GetAIName(e - PFE_AI0)),
+			ai_name_buf = fmt::format("AI {} {}", e - PFE_AI0 + 1, GetAIName(e - PFE_AI0));
```
glx22, line 1084: Is using `ai_name_buf` still really needed ?

rubidium42: If we do not want to copy `MEASUREMENT_NAMES[e]` into a `std::string` each time, then yes. Otherwise `name` could become `std::string` with a technically unneeded copy. After all, the `std::string` must outlive the `std::string_view`.

LordAro (approved):

PR #10898, opened 2023-05-31
Codechange: use C++ idioms for the clipboard handling

rubidium42:
## Motivation / Problem

C-style string handling.

## Description

Buffer + bool return value to `std::optional<std::string>`.

## Limitations

Not tested on Apple nor OS/2.

## Checklist for review

### DIFF
--- a/src/os/os2/os2.cpp
+++ b/src/os/os2/os2.cpp
@@ -154,27 +154,25 @@ void ShowOSErrorBox(const char *buf, bool system)
 	WinTerminate(hab);
 }
 
-bool GetClipboardContents(char *buffer, const char *last)
+std::optional<std::string> GetClipboardContents()
 {
 /* XXX -- Currently no clipboard support implemented with GCC */
 #ifndef __INNOTEK_LIBC__
 	HAB hab = 0;
 
-	if (WinOpenClipbrd(hab))
-	{
-		const char *text = (const char*)WinQueryClipbrdData(hab, CF_TEXT);
+	if (WinOpenClipbrd(hab)) {
+		const char *text = (const char *)WinQueryClipbrdData(hab, CF_TEXT);
 
-		if (text != nullptr)
-		{
-			strecpy(buffer, text, last);
+		if (text != nullptr) {
+			std::string result = text;
 			WinCloseClipbrd(hab);
-			return true;
+			return result;
 		}
 
 		WinCloseClipbrd(hab);
 	}
 #endif
-	return false;
+	return std::nullopt;
 }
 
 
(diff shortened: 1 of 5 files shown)


rubidium42:
@glx22, could you check whether the Windows changes still remain functional?

glx22:
I can paste (tested only in the console but if it works there...)

src/os/os2/os2.cpp
```diff
@@ -154,27 +154,25 @@ void ShowOSErrorBox(const char *buf, bool system)
...
 {
 /* XXX -- Currently no clipboard support implemented with GCC */
 #ifndef __INNOTEK_LIBC__
 	HAB hab = 0;
 
-	if (WinOpenClipbrd(hab))
-	{
-		const char *text = (const char*)WinQueryClipbrdData(hab, CF_TEXT);
+	if (WinOpenClipbrd(hab)) {
+		const char *text = (const char *)WinQueryClipbrdData(hab, CF_TEXT);
```
TrueBrain, line 164: `static_cast` if you are changing it anyway? Or is it a `const_cast`? Can't tell :D

src/os/windows/win32.cpp
```diff
@@ -433,26 +433,19 @@ void DetermineBasePaths(const char *exe)
...
+	GlobalUnlock(cbuf);
+	CloseClipboard();
 
-		if (out_len == 0) return false;
-	} else {
-		return false;
-	}
-
-	return true;
+	if (result.empty()) return std::nullopt;
```
TrueBrain, line 447: This line made me wonder: do we actually need `std::optional`, or is an empty string the same thing?

I am fine with both, just double-checking we actually need an `std::optional` here.

src/os/os2/os2.cpp
rubidium42: According to the specification it returns a `ULONG`. No idea whether that can be `static_cast` or `const_cast`. So, I rather leave it as is and just fix the coding style mistakes.

src/os/windows/win32.cpp
rubidium42: For Windows and Apple it seems to consider empty strings as "no content available". For Unix/OS/2 it returns a nullptr if there is no content available, and it seemingly happily returns an empty string.
So, I'm just keeping it functionally the same. I don't know whether that's the best thing, but figuring that out over multiple platforms is out of scope for me.

src/os/os2/os2.cpp
TrueBrain: You are lucky this is code that hasn't been compiled in months / years .. @orudge : you had to intention to maintain OS/2 port, but is that still realistic? I don't mind etc, but as we also don't have a CI, I am starting to worry more and more if it actually still works :)

TrueBrain (approved):

PR #10905, opened 2023-06-02
Codechange: Sprite mapping for objects doesn't involve cargo types.

PeterN:
## Motivation / Problem

Objects have a default sprite group and an optional purchase list sprite group. There is no need to pretend that one of these is a cargo ID.

## Description

Remove TranslateCargo call from ObjectMapSpriteGroup and check for the valid value directly.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -5883,7 +5883,7 @@ static void ObjectMapSpriteGroup(ByteReader *buf, uint8 idcount)
 				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


TrueBrain (dismissed):
The `1` is as magic as the `0`, so not for this PR to fix as far as I am concerned, but it would help if we can replace them with some more informative information :)

TrueBrain (approved):

PR #10909, opened 2023-06-02
Codechange: use C++ strings/paths to resolve links in tars

rubidium42:
## Motivation / Problem

The `AddFile` function of the `TarScanner` uses a lot of C-style string mangling to achieve its goal.

## Description

Replace strecpy/strecat with string assignment and concatenation.
Replace manual concatenation and normalisation of link path names with usage of std::filesystem::path's functions.

## Limitations

Added an extra function to safely get a string from the tar header as that might be garbage, so `strlen` can easily overflow.
Doing `std::string(random_buffer, lengthof(buffer))` will yield a string of the buffer's size, so `.empty()` does not do what it's supposed to do. 
So, either count a priori and copy only what's needed, or copy everything and reduce the size later on. I chose for the former.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -26,6 +26,7 @@
 #endif
 #include <sys/stat.h>
 #include <sstream>
+#include <filesystem>
 
 #include "safeguards.h"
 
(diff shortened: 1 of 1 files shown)


src/fileio.cpp
```diff
@@ -392,15 +393,17 @@ static void TarAddLink(const std::string &srcParam, const std::string &destParam
...
  * @param name Filename to process.
  */
-static void SimplifyFileName(char *name)
+static void SimplifyFileName(std::string &name)
 {
-	/* Force lowercase */
-	strtolower(name);
-
-	/* Tar-files always have '/' path-separator, but we want our PATHSEPCHAR */
+	std::transform(name.begin(), name.end(), name.begin(), [](char c) {
```
LordAro: seems mildly overkill, why not a for loop? iirc you did that for SanitizeFilename elsewhere

rubidium42: It's what you get when looking for a function to get lower case characters. But I changed it back to a for loop.

LordAro (approved):

PR #10910, opened 2023-06-03
Fix: track "memory installed" for surveys less precisely

TrueBrain:
## Motivation / Problem

It turns out, for Windows and Linux having the exact memory allows for easy tracing of an individual. That is exactly against the idea of the survey. And honestly, we don't need this precision.

## Description

Memory isn't all that precise for Linux and Windows; basically, you see the actual memory minus some value (things like VRAM etc are deduced from it).

To also allow representing things like 512 MiB of RAM, we use a simple algorithm to find what is the most likely nicest way to present the memory.

As bonus, also moved the concurrency to their OS functions; much cleaner code that way (but at the cost of a bit of duplication).

## Checklist for review

### DIFF
--- a/src/network/network_survey.cpp
+++ b/src/network/network_survey.cpp
@@ -298,6 +298,33 @@ static void SurveyGameScript(nlohmann::json &survey)
 	survey = fmt::format("{}.{}", Game::GetInfo()->GetName(), Game::GetInfo()->GetVersion());
 }
 
+/**
+ * Change the bytes of memory into a textual version rounded up to the biggest unit.
+ *
+ * For example, 16751108096 would become 16 GiB.
+ *
+ * @param memory The bytes of memory.
+ * @return std::string A textual representation.
+ */
+std::string SurveyMemoryToText(uint64_t memory)
+{
+	memory = memory / 1024; // KiB
+	memory = CeilDiv(memory, 1024); // MiB
+
+	/* Anything above 512 MiB we represent in GiB. */
+	if (memory > 512) {
+		return fmt::format("{} GiB", CeilDiv(memory, 1024));
+	}
+
+	/* Anything above 64 MiB we represent in a multiplier of 128 MiB. */
+	if (memory > 64) {
+		return fmt::format("{} MiB", Ceil(memory, 128));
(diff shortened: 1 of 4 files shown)


PeterN:
Less precisely?

TrueBrain:
> Less precisely?

Don't ask me! You are the native speaker :P

src/network/network_survey.cpp
```diff
@@ -329,6 +355,7 @@ std::string NetworkSurveyHandler::CreatePayload(Reason reason, bool for_preview)
 		auto &info = survey["info"];
 		SurveyOS(info["os"]);
 		info["os"]["hardware_concurrency"] = std::thread::hardware_concurrency();
+		info["os"]["memory"] = MemoryToText(info["os"]["memory"]);
```
PeterN: I guess there's some magic going on here by the json stuff, to be able to pass a 64bit value without casts and return a std::string into the same place.

TrueBrain: Absolutely! If you like, I can avoid it, but this is what nlohmann-json allows just fine :)

(in a bit more detail, this is just templating. By calling the function I ask the item to return an `uint64_t`, so it returns that internal value. nlohmann-json does validate there was actually an integer in there btw).

TrueBrain: It annoyed me enough that I fixed it. Much more proper now.

src/network/network_survey.cpp
```diff
@@ -298,6 +298,32 @@ static void SurveyGameScript(nlohmann::json &survey)
...
 
+/**
+ * Change the bytes of memory into a textual version rounded up to the biggest unit.
+ *
+ * For example, 16751108096 would become 16 GiB.
+ *
+ * @param memory The bytes of memory.
+ * @return std::string A textual representation.
+ */
+std::string SurveyMemoryToText(uint64_t memory) {
```
rubidium42:
```suggestion
std::string SurveyMemoryToText(uint64_t memory)
{
```

rubidium42 (approved):

LordAro (approved):

PR #10914, opened 2023-06-03
Feature: allow to do a hostile takeover of an AI company (in singleplayer)

TrueBrain:
## Motivation / Problem

With the removal of the share-system, you could no longer make an AI disappear in a single player game. At least, not without going into the console.

## Description

Add a nice little button `Hostile Takeover` to allow you to take over any AI (in singleplayer) at any moment in time. But be careful: it can be really expensive if the AI is doing well.

I reused the bankruptcy flow, as it is very identical; just the amount (and text) is different.

image

## Limitations

Finding a good value for what to pay is a bit tricky. The normal `CalculateCompanyValue(.., false)` doesn't work (this is what you pay when you go bankrupt), as the loan is "forgiven" there. Which is silly in a hostile takeover.
Additionally, I also added a penalty for companies that are doing well, as otherwise it is really cheap to buy out an AI early on, and even cheaper later in the game.
But, this is all based on an opinion .. it needs some input, before or after merging, what an actual good value is.

## Checklist for review

### DIFF
--- a/src/company_func.h
+++ b/src/company_func.h
@@ -19,7 +19,7 @@ bool MayCompanyTakeOver(CompanyID cbig, CompanyID small);
 void ChangeOwnershipOfCompanyItems(Owner old_owner, Owner new_owner);
 void GetNameOfOwner(Owner owner, TileIndex tile);
 void SetLocalCompany(CompanyID new_company);
-void ShowBuyCompanyDialog(CompanyID company);
+void ShowBuyCompanyDialog(CompanyID company, bool hostile_takeover);
 void CompanyAdminUpdate(const Company *company);
 void CompanyAdminBankrupt(CompanyID company_id);
 void UpdateLandscapingLimits();
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -2230,6 +2230,12 @@ static const NWidgetPart _nested_company_widgets[] = {
 						NWidget(NWID_SPACER), SetFill(0, 1),
 						NWidget(NWID_HORIZONTAL), SetPIP(0, 4, 0),
 							NWidget(NWID_SPACER), SetFill(1, 0),
+							NWidget(NWID_SELECTION, INVALID_COLOUR, WID_C_SELECT_HOSTILE_TAKEOVER),
+								NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_C_HOSTILE_TAKEOVER), SetDataTip(STR_COMPANY_VIEW_HOSTILE_TAKEOVER_BUTTON, STR_COMPANY_VIEW_HOSTILE_TAKEOVER_TOOLTIP),
+							EndContainer(),
+						EndContainer(),
+						NWidget(NWID_HORIZONTAL), SetPIP(0, 4, 0),
+							NWidget(NWID_SPACER), SetFill(1, 0),
 							NWidget(NWID_SELECTION, INVALID_COLOUR, WID_C_SELECT_GIVE_MONEY),
 								NWidget(WWT_PUSHTXTBTN, COLOUR_GREY, WID_C_GIVE_MONEY), SetDataTip(STR_COMPANY_VIEW_GIVE_MONEY_BUTTON, STR_COMPANY_VIEW_GIVE_MONEY_TOOLTIP),
 							EndContainer(),
@@ -2332,6 +2338,13 @@ struct CompanyWindow : Window
 				wi->SetDisplayedPlane(plane);
 				reinit = true;
 			}
+			/* Enable/disable 'Hostile Takeover' button. */
+			plane = ((local || _local_company == COMPANY_SPECTATOR || !c->is_ai || _networking) ? SZSP_NONE : 0);
+			wi = this->GetWidget<NWidgetStacked>(WID_C_SELECT_HOSTILE_TAKEOVER);
+			if (plane != wi->shown_plane) {
+				wi->SetDisplayedPlane(plane);
+				reinit = true;
+			}
 
 			/* Multiplayer buttons. */
 			plane = ((!_networking) ? (int)SZSP_NONE : (int)(local ? CWP_MP_C_PWD : CWP_MP_C_JOIN));
@@ -2397,6 +2410,7 @@ struct CompanyWindow : Window
 			case WID_C_RELOCATE_HQ:
 			case WID_C_VIEW_INFRASTRUCTURE:
 			case WID_C_GIVE_MONEY:
+			case WID_C_HOSTILE_TAKEOVER:
 			case WID_C_COMPANY_PASSWORD:
 			case WID_C_COMPANY_JOIN:
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_VIEW_HQ_BUTTON).width);
@@ -2404,6 +2418,7 @@ struct CompanyWindow : Window
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_RELOCATE_HQ).width);
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_INFRASTRUCTURE_BUTTON).width);
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_GIVE_MONEY_BUTTON).width);
+				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_HOSTILE_TAKEOVER_BUTTON).width);
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_PASSWORD).width);
 				size->width = std::max(size->width, GetStringBoundingBox(STR_COMPANY_VIEW_JOIN).width);
 				break;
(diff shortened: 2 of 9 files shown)


src/lang/english.txt
```diff
@@ -3774,6 +3776,7 @@ STR_COMPANY_VIEW_PRESIDENT_S_NAME_QUERY_CAPTION                 :Manager's Name
 STR_COMPANY_VIEW_GIVE_MONEY_QUERY_CAPTION                       :Enter the amount of money you want to give
 
 STR_BUY_COMPANY_MESSAGE                                         :{WHITE}We are looking for a transport company to take-over our company.{}{}Do you want to purchase {COMPANY} for {CURRENCY_LONG}?
+STR_BUY_COMPANY_HOSTILE_TAKEOVER                                :{WHITE}You are about to do a hostile takeover of {COMPANY}.{}{}You will get all their assets, pay off their loan and two years of profit. The total comes down to {CURRENCY_LONG} currently.{}{}Do you want to continue the hostile takeover?
```
James103: From the message:
> You will get all their assets, pay off their loan and two years of profit.

That might be interpreted as receiving two years of profit, but the actual cost includes two years of profit. Therefore, another way that part could be worded is:
> You will get all their assets in exchange for paying off their loan plus two years of profit.

PeterN:
So a stop-gap until an unbroken shares system with hostile takeovers is implemented?

TrueBrain:
> So a stop-gap until an unbroken shares system with hostile takeovers is implemented?

In sorts, yes. At least it is easier to maintain than the shares system was, and it fills that gap of players that like an AI for the first few years, but then want to get ride of them.

src/lang/english.txt
TrueBrain: I really appreciate these comments, tnx :) Also cleaned up the next sentence a bit. Much better now!

nielsmh:
The purchase price should at the very least also include the value of fixed assets too, so you can't just get fifty brand new Yate Haugan, that never made a delivery yet, and thus have zero profit, for free.

TrueBrain:
> The purchase price should at the very least also include the value of fixed assets too, so you can't just get fifty brand new Yate Haugan, that never made a delivery yet, and thus have zero profit, for free.

They are.

nielsmh:
> > The purchase price should at the very least also include the value of fixed assets too, so you can't just get fifty brand new Yate Haugan, that never made a delivery yet, and thus have zero profit, for free.
> 
> They are.

Make sure that's reflected in the GUI too, right now the message shown does not mention anything about that.

TrueBrain:
> > > The purchase price should at the very least also include the value of fixed assets too, so you can't just get fifty brand new Yate Haugan, that never made a delivery yet, and thus have zero profit, for free.
> > 
> > 
> > They are.
> 
> Make sure that's reflected in the GUI too, right now the message shown does not mention anything about that.

Yes boss. Wait, you are not my boss :P

Anyway, suggestions? I am having a hard time wording it clearly, so could use some suggestions :)

Andrew350:
I thought "all their assets" made that pretty clear?

TrueBrain:
Tnx @nielsmh , talked on Discord to find a much better sentence. `In a hostile takeover of {COMPANY} you will purchase all fixed assets, pay off all loans, and pay two years worth of profits.{}{}The total is estimated to be {CURRENCY_LONG}.{}{}Do you want to continue this hostile takeover?`

Will update the screenshot soon.

FLHerne:
> I thought "all their assets" made that pretty clear?

It says "*get* all their assets, in exchange for paying off their loan plus two years of profits".

"all their assets in exchange for their book value, plus paying off any loan plus two years of profits", except that's too long.

To be honest I don't see why it's important to explain the cost at all. It is what it is, the approximate range of the company's value is fairly obvious, and no other estimated cost has a similar explanation.

Does it even have to be described as a "Hostile Takeover"? It's an AI, we don't know what the management thinks about the issue. Could just as well be a voluntary merger.

My concrete suggestion (off the top of my head):
----
[Buy Company]

You are about to buy Fast Transport Inc.
All infrastructure, vehicles and other assets will become part of your company, Anglian Logistics.
Estimated Cost: £131,445

[Buy Company] [Cancel]

TrueBrain:
I really like "Hostile Takeover", just because it is a bit tongue in cheek towards what you are doing. In reverse, "buy" gives the suggestion the AI is always fine with it. So I would rather keep this wording. Not everything has to be all that serious in OpenTTD :)

As for the rest of the text, I think the new text is fine, not? Being a bit clear it is not the same as a bankruptcy never hurts anyone.

michicc:
Doing some more bike painting: "fixed assets" might be the correct financial term, but I would very much think that many players will interpret is as excluding vehicles (as the move and are not fixed in place).

nielsmh:
Additional thought, which probably does not need to be part of this PR: A game script should be able to command the closure of a company and transfer of all its assets to someone else. Possibly with flags for what assets to transfer and what to void. When a GS does that, the GS should also be responsible for handling any payment it deems is required.

Eddi-z:
Afair the old feature had a protection period of 5 years before a complete takeover was possible, should this one have this as well?

TrueBrain:
> Afair the old feature had a protection period of 5 years before a complete takeover was possible, should this one have this as well?

I did not see such need, given this is an AI and in single player only. If you want to do a hostile takeover earlier .. go nuts, I would say. Additionally, the costs are significantly higher than a share-buyout was, so that hopefully helps too.

src/company_func.h
```diff
@@ -19,7 +19,7 @@ bool MayCompanyTakeOver(CompanyID cbig, CompanyID small);
 void ChangeOwnershipOfCompanyItems(Owner old_owner, Owner new_owner);
 void GetNameOfOwner(Owner owner, TileIndex tile);
 void SetLocalCompany(CompanyID new_company);
-void ShowBuyCompanyDialog(CompanyID company);
+void ShowBuyCompanyDialog(CompanyID company, bool hostile_takeover);
```
Eddi-z, line 22: Maybe this could be an enum: {Bankrupcy, Hostile, ...}?

TrueBrain: When a third setting arrives, sure. But for now that seems a bit premature. Unless others agree ofc!

src/company_gui.cpp
```diff
@@ -2753,10 +2775,29 @@ struct BuyCompanyWindow : Window {
...
+	 */
+	IntervalTimer<TimerWindow> rescale_interval = {std::chrono::seconds(3), [this](auto) {
+		/* Value can't change when in bankruptcy. */
+		if (!this->hostile_takeover) return;
+
+		const Company *c = Company::Get((CompanyID)this->window_number);
+		auto new_value = CalculateHostileTakeoverValue(c);
+		if (new_value != this->company_value) {
+			this->company_value = new_value;
+			this->SetDirty();
```
rubidium42: Not sure if it is really worth bothering about, as the chance is probably relatively small. However, if the company value increases the string height might increase too, causing the last bit of the message to not be shown. Maybe call `ReInit` when the number of digits in the company value increases?
Or maybe it's always a concern, if the digits are not a fixed width.

src/company_gui.cpp
```diff
@@ -2788,8 +2829,12 @@ static WindowDesc _buy_company_desc(
...
+ * @param hostile_takeover Whether this is a hostile takeover.
  */
-void ShowBuyCompanyDialog(CompanyID company)
+void ShowBuyCompanyDialog(CompanyID company, bool hostile_takeover)
 {
-	AllocateWindowDescFront<BuyCompanyWindow>(&_buy_company_desc, company);
+	BuyCompanyWindow *w = (BuyCompanyWindow *)BringWindowToFrontById(WC_BUY_COMPANY, company);
+	if (w == nullptr) {
+		new BuyCompanyWindow(&_buy_company_desc, company, hostile_takeover);
+	}
```
rubidium42:
```suggestion
	if (BringWindowToFrontById(WC_BUY_COMPANY, company) == nullptr) {
		new BuyCompanyWindow(&_buy_company_desc, company, hostile_takeover);
	}
```
Not sure this is conceptually better, but the casting to `BuyCompanyWindow *` seems unneeded as nothing is done with it later on that requires the type to be anything other than `Window *`. At least replace the cast with a `static_cast`, instead of the C-style cast.

src/economy.cpp
```diff
@@ -136,13 +133,62 @@ Money CalculateCompanyValue(const Company *c, bool including_loan)
...
+{
+	Money value = CalculateCompanyAssetValue(c);
+
+	value += c->current_loan;
+	/* Negative balance is basically a loan. */
+	if (c->money < 0) {
+		value += -c->money;
+	}
+
+	for (int i = 0; i < 4; i++) {
```
rubidium42: Maybe rename `i` to `quarter`, so it's more self-descriptive that you are essentially iterating over 4 quarters, i.e. a year. Now it was not trivial to figure out from the code alone.

src/economy.cpp
```diff
@@ -136,13 +133,62 @@ Money CalculateCompanyValue(const Company *c, bool including_loan)
...
+/**
+ * Calculate what you have to pay to take over a company.
+ *
+ * This is different from bankruptcy and company value, and involves a few
+ * more parameters to make it more realistic.
+ *
+ * You have to pay for:
+ * - The value of all the assets in the company.
+ * - The loan the company has (the investors really want their money back).
+ * - The profit for the next two years (if positive) based on the last year.
```
rubidium42:
```suggestion
 * - The profit for the next two years (if positive) based on the previous four quarters.
```
"last year" in itself is a bit confusing, lets assume today is June 6th 2023. Does it mean June 6th 2022 up to including June 5th 2023, or January 1st 2022 up to including December 31st 2023. But here it's actually like April 1st 2022 up to including March 31st 2022.

src/company_gui.cpp
TrueBrain: This is a copy/paste of one of the many other places we do it like this. So I didn't actually want to touch that .. but it is such an old way of doing shit :) I will see what I can do :)

rubidium42 (approved):

PR #10915, opened 2023-06-03
Fix: when syncing width of GUI items, take padding into account

TrueBrain:
## Motivation / Problem

image

Minor misalignment on some buttons, where the intention was to make all those buttons the same size.

## Description

While at it, fixed all places that showed similar pattern. Basically, `size` already has `padding` included, so if you recalculate it, add `padding` again.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -1922,17 +1922,17 @@ struct CompanyInfrastructureWindow : Window
 			case WID_CI_RAIL_DESC: {
 				uint lines = 1; // Starts at 1 because a line is also required for the section title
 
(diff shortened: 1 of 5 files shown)


PeterN:
Some of this isn't necessary, but only because the default padding for some widget types (WWT_TEXT, WWT_LABEL and WWT_EMPTY) is zero anyway.

rubidium42 (approved):
Seems okay to me. Adding an extra `0` when not currently needed doesn't seem like a big problem, and might be useful when someone decides those should get a padding other than 0.

PR #10920, opened 2023-06-03
Add: [Linux] change default scroll mode to non-mouse-lock

### LINKED ISSUE #10140: [Bug]: SDL2 / Wayland: Move viewport with mouse position locked does not lock mouse

TrueBrain:
## Motivation / Problem

Not all SDL backends support mouse locking, of "warping". This means users are left confused why the navigation is so shitty, and have to find a certain setting to make it better.

I think it would be better if we guide the user towards this goal.
This most likely is the root cause of #10140, as Wayland doesn't support warping at all.

Fixes #10140.

## Description

Wayland doesn't support mouse warping, X11 only for native systems (so not for remote desktop, WSLg, etc), and emscripten neither without complications. All these cannot offer a mouse-lock.

So the amount of systems that do offer it, are less and less than those that do offer it. Enough reason to flip the switch here.

Initially we tried to detect whether warping was supported, but this turns out to be nearly impossible. Detection for Wayland is simple, but detecting whether the X11 backend actually supports warping ... no. So this solution is simpler.

This will confuse a few players, especially those on X11 and have everything working. But we were enable to find a proper solution that works for all; and we rather have 100% of the players able to play the game, than N% of the players having it the way they are used to.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1558,7 +1558,7 @@ STR_CONFIG_SETTING_LINKGRAPH_COLOURS_GREY_TO_RED                :Grey to red
 STR_CONFIG_SETTING_LINKGRAPH_COLOURS_GREYSCALE                  :Greyscale
 
 STR_CONFIG_SETTING_SCROLLMODE                                   :Viewport scroll behaviour: {STRING2}
-STR_CONFIG_SETTING_SCROLLMODE_HELPTEXT                          :Behaviour when scrolling the map
+STR_CONFIG_SETTING_SCROLLMODE_HELPTEXT                          :Behaviour when scrolling the map. The "mouse position locked" options don't work on all systems, like web-based versions, touchscreens, Linux with Wayland, and others
 ###length 4
 STR_CONFIG_SETTING_SCROLLMODE_DEFAULT                           :Move viewport with RMB, mouse position locked
 STR_CONFIG_SETTING_SCROLLMODE_RMB_LOCKED                        :Move map with RMB, mouse position locked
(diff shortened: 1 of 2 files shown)


src/video/sdl2_v.cpp
```diff
@@ -623,6 +623,22 @@ void VideoDriver_SDL_Base::LoopOnce()
...
 void VideoDriver_SDL_Base::MainLoop()
 {
+	/* We smoketest if the backend supports actually moving the
+	 * mouse. There is really no other way to detect it, as even
+	 * if, say, X11 reports it did it, X11-servers like remote
+	 * desktops or WSLg don't actually execute it. So the best
+	 * way to learn is to try. */
+	if (_settings_client.gui.scroll_mode == VSM_VIEWPORT_RMB_FIXED || _settings_client.gui.scroll_mode == VSM_MAP_RMB_FIXED) {
+		int x_before, x_after, y;
+		SDL_GetMouseState(&x_before, &y);
```
frosch123: SDL_GetMouseState always returns (0,0) for me on native X11.

If using SDL_GetGlobalMouseState and SDL_WarpMouseGlobal instead, it works.

frosch123 (approved):

PR #10921, opened 2023-06-04
Codechange: introduce (temporary) string builder and use it for formatting numbers

rubidium42:
## Motivation / Problem

The strings system uses a (limited) char buffer to write the data strings to. This has to be changed to std::string.

## Description

Introduce a `StringBuilder` that essentially a functional `std::back_insert_iterator<std::string>` with a number of convenience functions for the migration. At some point the `StringBuilder` will have to be replaced by `std::back_insert_iterator<std::string>` or just aliasing it as such.
This is the first step of rewriting the string system, and originally part of #10810 but I'm trying to get it merged in some smaller steps now.

## Limitations

Not the complete migration, but just a step towards moving to std::string.

## Checklist for review

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -424,6 +424,7 @@ add_files(
     stringfilter_type.h
     strings.cpp
     strings_func.h
+    strings_internal.h
     strings_type.h
     subsidy.cpp
     subsidy_base.h
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -43,6 +43,8 @@
 #include "table/strings.h"
 #include "table/control_codes.h"
 
+#include "strings_internal.h"
+
 #include "safeguards.h"
 
 std::string _config_language_file;                ///< The file (name) stored in the configuration.
@@ -315,16 +317,15 @@ void SetDParamStr(uint n, const std::string &str)
 
 /**
  * Format a number into a string.
- * @param buff      the buffer to write to
+ * @param builder   the string builder to write to
  * @param number    the number to write down
  * @param last      the last element in the buffer
  * @param separator the thousands-separator to use
  * @param zerofill  minimum number of digits to print for the integer part. The number will be filled with zeros at the front if necessary.
  * @param fractional_digits number of fractional digits to display after a decimal separator. The decimal separator is inserted
  *                          in front of the \a fractional_digits last digit of \a number.
- * @return till where we wrote
  */
-static char *FormatNumber(char *buff, int64 number, const char *last, const char *separator, int zerofill = 1, int fractional_digits = 0)
+static void FormatNumber(StringBuilder &builder, int64 number, const char *separator, int zerofill = 1, int fractional_digits = 0)
 {
 	static const int max_digits = 20;
 	uint64 divisor = 10000000000000000000ULL;
@@ -332,7 +333,7 @@ static char *FormatNumber(char *buff, int64 number, const char *last, const char
 	int thousands_offset = (max_digits - fractional_digits - 1) % 3;
 
 	if (number < 0) {
-		if (buff != last) *buff++ = '-';
+		builder += '-';
 		number = -number;
 	}
 
@@ -342,7 +343,7 @@ static char *FormatNumber(char *buff, int64 number, const char *last, const char
 		if (i == max_digits - fractional_digits) {
 			const char *decimal_separator = _settings_game.locale.digit_decimal_separator.c_str();
 			if (StrEmpty(decimal_separator)) decimal_separator = _langpack.langpack->digit_decimal_separator;
-			buff = strecpy(buff, decimal_separator, last);
+			builder += decimal_separator;
 		}
 
 		uint64 quot = 0;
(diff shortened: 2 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -423,6 +423,7 @@ add_files(
     stringfilter.cpp
     stringfilter_type.h
     strings.cpp
+    strings_internal.h
```
TrueBrain: The alphabet says: no! :D

src/strings_internal.h
```diff
@@ -0,0 +1,123 @@
...
+	/**
+	 * Create the builder of an external buffer.
+	 * @param start The start location to write to.
+	 * @param last  The last location to write to.
+	 */
+	StringBuilder(char *start, const char *last) : current(start), last(last) {}
+
+	/* Required operators for this to be an output_iterator; mimicks std::back_insert_iterator. */
+	StringBuilder &operator++() { return *this; }
+	StringBuilder operator++(int) { return *this; }
```
TrueBrain, line 41: Shouldn't there be some ++ on the right side?

src/strings_internal.h
```diff
@@ -0,0 +1,123 @@
...
+	{
+		return this->operator+=(str.c_str());
+	}
+
+	/**
+	 * Encode the given Utf8 character into the output buffer.
+	 * @param c The character to encode.
+	 * @return true iff there was enough space and the character got added.
+	 */
+	bool Utf8Encode(WChar c)
```
TrueBrain, line 93: Why does this as only function return a bool? And of the three cases I saw above, only 2 check thee result and the third doesn't? Shouldn't this just silently ignore adding like the rest does?

See (1)

src/strings.cpp
```diff
@@ -496,21 +487,17 @@ static char *FormatGenericCurrency(char *buff, const CurrencySpec *spec, Money n
...
 	 * Here, it can can be either 1 (suffix) or 2 (both prefix and suffix).
 	 * The only remaining value is 1 (prefix), so everything that is not 0 */
-	if (spec->symbol_pos != 0) buff = strecpy(buff, spec->suffix.c_str(), last);
+	if (spec->symbol_pos != 0) builder += spec->suffix;
 
 	if (negative) {
-		if (buff + Utf8CharLen(SCC_POP_COLOUR) > last) return buff;
-		buff += Utf8Encode(buff, SCC_POP_COLOUR);
-		*buff = '\0';
+		builder.Utf8Encode(SCC_POP_COLOUR);
```
TrueBrain, line 499: (1)

src/strings.cpp
```diff
@@ -467,18 +460,16 @@ static char *FormatGenericCurrency(char *buff, const CurrencySpec *spec, Money n
 
 	/* convert from negative */
 	if (number < 0) {
-		if (buff + Utf8CharLen(SCC_PUSH_COLOUR) > last) return buff;
-		buff += Utf8Encode(buff, SCC_PUSH_COLOUR);
-		if (buff + Utf8CharLen(SCC_RED) > last) return buff;
-		buff += Utf8Encode(buff, SCC_RED);
-		buff = strecpy(buff, "-", last);
+		if (!builder.Utf8Encode(SCC_PUSH_COLOUR)) return;
+		if (!builder.Utf8Encode(SCC_RED)) return;
```
TrueBrain, line 464: (1)

src/strings_internal.h
rubidium42: https://en.cppreference.com/w/cpp/language/operator_incdec
The first is the prefix one, so ++builder. The second is the postfix one, so builder++.
So, I think I got both covered unless I'm not understanding your comment.

src/CMakeLists.txt
rubidium42: https://en.wikipedia.org/wiki/I_before_E_except_after_C :D

src/strings_internal.h
rubidium42: It returns a bool so it can remain functionally the same as it was before. In the future, when we move to std::string, no function will (silently) truncate the string anymore, so this can return void and even `Remaining()` can go away. But not in this step; first move without functional changes to something where we can replace the `char[]` with `std::string`.

The case that doesn't check is due to me performing some "optimisations"; otherwise you would have:
```
void function(some params) {
   ...
   if (negative) {
      if (!builder.Utf8Encode(SCC_POP_COLOUR)) return;
   }
   /* Implicit return due to function ending. */
}
```

src/strings_internal.h
TrueBrain: You return `*this`. So if I do `StringBuilder a; a++; a++; a++`. I get 3 times the same pointer back? Or am I missing something? I miss the actual changing of anything in or with `*this` :)

src/strings_internal.h
TrueBrain: What I find a bit odd about this choice, is that you only did it for `Utf8Encode`. Things like `+=` etc did change functional, and do silently drop the addition. So what is it that makes `Utf8Encode` special, is basically what I was asking. Don't any of the other cases need to return early (for now) if the string was full?

src/strings_internal.h
rubidium42: In short: `std::back_insert_iterator` does it this way.
Longer explanations: https://en.cppreference.com/w/cpp/iterator/back_insert_iterator/operator%2B%2B. The functions need to exist for it to be considered an `output_iterator`, but like the `std::back_insert_iterator` it does not do anything as the assignments itself already move the iterator. And it needs to be an `output_iterator` for `fmt::format_to` to function.

src/strings_internal.h
TrueBrain: To put Discord here, I took another look at the code: the rest already silently ignored when a string was full. So that is unchanged with your commits.

I guess they did check the length and returned early for `Utf8Encode`, as otherwise you can end up in situations where you inserted only part of the codepoint in the string, which is bad. But you don't do that anyway. And when I look at the code, if the string is full after `SCC_RED`, it will happily run the rest of the function, trying to keep on adding to the string.

So I would suggest to already extend that to this function. If `SCC_PUSHCOLOR` already doesn't fit: who cares, let it run on for those 2 more functions.

Already prepares you for the future, and does away with the special exception :) My 2 cents .. I can't see the whole picture here ofc :)

TrueBrain: And now I talked myself in a corner ... `SCC_RED` might not fit, but the `-` added after that might. So that would be very wrong. No I take it back, your code is correct.

rubidium42: No, `+=` does not change functionally as it replaces a `strecpy` with a (now hidden) `strecpy`. Both truncate(d) silently.

`Utf8Encode` is a different beast, as that function in itself is not even doing bounds checking. So, in many places where it is used there is a "is there enough space to put this encoded character" check, which does then decide what to do. I moved that "is there enough space" check into `Utf8Encode` and exposed the result, so things can remain functionally the same while performing the migration.

src/strings_internal.h
TrueBrain: Please add this tiny piece of information to the comment. You now only note it mimics (misspelled btw) the iterator, but maybe add: `mimics std::back_insert_iterator, which does nothing for these operators`. (well, my phrasing sucks, but okay :D) Prevents future-us thinking: is this a bug? And go down this rabbit hole again :)

And tnx for the explanation :)

TrueBrain (approved):

PR #10923, opened 2023-06-04
Codechange: use std::array and std::string for high scores

rubidium42:
## Motivation / Problem

High score code using `GetString` that writes to a `char[]`.

## Description

Replace `memmove`/`memset` with `std::move_backward`/`std::fill`.
Replace the C-style array of high scores with `std::array`, to easier loop over it.
Replace the C-style string with `std::string`.
Use `std::unique_ptr` with our `FileDeleter` for the `FILE *`, so bailing out of the functions is easier as you do not need to close the file pointer in each of the paths.
Simplify some of the code where applicable.

## Limitations

Still very much relies on an ancient format. I'm not changing that, which means the file writing and reading can be somewhat cumbersome.

## Checklist for review

### DIFF
--- a/src/highscore.cpp
+++ b/src/highscore.cpp
@@ -12,6 +12,7 @@
 #include "company_base.h"
 #include "company_func.h"
 #include "cheat_func.h"
+#include "fileio_func.h"
 #include "string_func.h"
 #include "strings_func.h"
 #include "table/strings.h"
@@ -19,7 +20,7 @@
 
 #include "safeguards.h"
 
-HighScore _highscore_table[SP_HIGHSCORE_END][5]; ///< various difficulty-settings; top 5
+HighScoresTable _highscore_table; ///< Table with all the high scores.
 std::string _highscore_file; ///< The file to store the highscore data in.
 
 static const StringID _endgame_perf_titles[] = {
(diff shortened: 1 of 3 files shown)


src/highscore.cpp
```diff
@@ -85,95 +89,82 @@ static bool HighScoreSorter(const Company * const &a, const Company * const &b)
...
+
+	/* Does not iterate through the complete array!. */
+	for (int i = 0; i < SP_SAVED_HIGHSCORE_END; i++) {
+		for (HighScore &hs : _highscore_table[i]) {
+			/* This code is weird and old fashioned to keep compatibility with the old high score files. */
+			byte name_length = ClampTo<byte>(hs.name.size());
+			if (fwrite(&name_length, sizeof(name_length), 1, fp.get()) != 1 || // Write the string length of the name
+					fwrite(hs.name.data(), name_length, 1, fp.get()) > 1 || // Yes... could be 0 bytes too
+					fwrite(&hs.score, sizeof(hs.score), 1, fp.get()) != 1 ||
+					fwrite("  ", 2, 1, fp.get()) != 1) { // Used to be hs.title, not saved anymore; compatibility
```
LordAro, line 136: this is horrid, but not your problem

src/highscore_gui.cpp
```diff
@@ -184,7 +184,7 @@ struct HighScoreWindow : EndGameHighScoreBaseWindow {
 
 	void OnPaint() override
 	{
-		const HighScore *hs = _highscore_table[this->window_number];
+		auto &hs = _highscore_table[this->window_number];
```
LordAro: const?

LordAro (approved):
Can't see anything obvious

src/highscore.cpp
rubidium42: Indeed it's horrid, but something with avoiding rabbit holes.
Like... there are 4 high score tables that are saved, but only one gets set and shown (the custom one).
So the code could do with an even larger cleaning, but that's indeed way beyond the scope of this.

PR #10884, opened 2023-05-27
Remove: obsolete NewGRF text unprinting.

PeterN:
## Motivation / Problem

TTDPatches "unprint" facility does not seem to serve a useful purpose in OpenTTD.

## Description

Remove it.

## Limitations

:shrug:

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #10925, opened 2023-06-04
Codechange: let FormatString use StringBuilder

rubidium42:
## Motivation / Problem

The strings system uses a (limited) char buffer to write the data strings to. This has to be changed to std::string.

## Description

Use the `StringBuilder` in `FormatString`, add some migration helpers so `GetStringWithArgs` can still remain as it is.

## Limitations

Depends on #10884 to get merged.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -171,11 +171,11 @@ void CopyOutDParam(uint64 *dst, const char **strings, StringID string, int num)
 	}
 }
 
-static char *StationGetSpecialString(char *buff, int x, const char *last);
+static void StationGetSpecialString(StringBuilder &builder, int x);
 static char *GetSpecialTownNameString(char *buff, int ind, uint32 seed, const char *last);
 static char *GetSpecialNameString(char *buff, int ind, StringParameters *args, const char *last);
 
-static char *FormatString(char *buff, const char *str, StringParameters *args, const char *last, uint case_index = 0, bool game_script = false, bool dry_run = false);
+static void FormatString(StringBuilder &builder, const char *str, StringParameters *args, uint case_index = 0, bool game_script = false, bool dry_run = false);
 
 struct LanguagePack : public LanguagePackHeader {
 	char data[]; // list of strings
@@ -251,14 +251,20 @@ char *GetStringWithArgs(char *buffr, StringID string, StringParameters *args, co
 			}
 			break;
 
-		case TEXT_TAB_GAMESCRIPT_START:
-			return FormatString(buffr, GetGameStringPtr(index), args, last, case_index, true);
+		case TEXT_TAB_GAMESCRIPT_START: {
+			StringBuilder builder(buffr, last);
+			FormatString(builder, GetGameStringPtr(index), args, case_index, true);
+			return builder.GetEnd();
+		}
 
 		case TEXT_TAB_OLD_NEWGRF:
 			NOT_REACHED();
 
-		case TEXT_TAB_NEWGRF_START:
-			return FormatString(buffr, GetGRFStringPtr(index), args, last, case_index);
+		case TEXT_TAB_NEWGRF_START: {
+			StringBuilder builder(buffr, last);
+			FormatString(builder, GetGRFStringPtr(index), args, case_index);
+			return builder.GetEnd();
+		}
 
 		default:
 			break;
@@ -271,7 +277,9 @@ char *GetStringWithArgs(char *buffr, StringID string, StringParameters *args, co
 		FatalError("String 0x{:X} is invalid. You are probably using an old version of the .lng file.\n", string);
 	}
 
-	return FormatString(buffr, GetStringPtr(string), args, last, case_index);
+	StringBuilder builder(buffr, last);
+	FormatString(builder, GetStringPtr(string), args, case_index);
+	return builder.GetEnd();
 }
 
 char *GetString(char *buffr, StringID string, const char *last)
@@ -280,6 +288,20 @@ char *GetString(char *buffr, StringID string, const char *last)
 	_global_string_params.offset = 0;
 	return GetStringWithArgs(buffr, string, &_global_string_params, last);
 }
+
+/**
+ * Get a parsed string with most special stringcodes replaced by the string parameters.
+ * @param builder     The builder of the string.
+ * @param string      The ID of the string to parse.
+ * @param args        Arguments for the string.
+ * @param case_index  The "case index". This will only be set when FormatString wants to print the string in a different case.
+ * @param game_script The string is coming directly from a game script.
+ */
+void GetStringWithArgs(StringBuilder &builder, StringID string, StringParameters *args, uint case_index = 0, bool game_script = false)
+{
+	builder.AddViaStreCallback([&](auto buff, auto last) { return GetStringWithArgs(buff, string, args, last, case_index, game_script); });
+}
+
 
 /**
  * Resolve the given StringID into a std::string with all the associated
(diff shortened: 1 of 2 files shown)


src/strings.cpp
```diff
@@ -281,6 +289,20 @@ char *GetString(char *buffr, StringID string, const char *last)
 	return GetStringWithArgs(buffr, string, &_global_string_params, last);
 }
 
+/**
+ * Migration helper to get a parsed string with most special stringcodes replaced by the string parameters.
```
TrueBrain: I would leave out "Migration helper" .. who knows, maybe you get bored tomorrow and never continue. Which is fine. But then this comment makes little sense :D

src/strings.cpp
```diff
@@ -806,18 +828,18 @@ static std::vector<const char *> _game_script_raw_strings;
...
  * @param dry_run True when the argt array is not yet initialized.
  */
-static char *FormatString(char *buff, const char *str_arg, StringParameters *args, const char *last, uint case_index, bool game_script, bool dry_run)
+static void FormatString(StringBuilder &builder, const char *str_arg, StringParameters *args, uint case_index, bool game_script, bool dry_run)
 {
 	uint orig_offset = args->offset;
 
 	/* When there is no array with types there is no need to do a dry run. */
 	if (args->HasTypeInformation() && !dry_run) {
+		StringBuilder dry_run_builder = builder;
```
TrueBrain, line 850: Codewise this feels off. The line above says `if not dry-run`, after which you make a `dry-run` variable. But also: why? The old code didn't do it?

As you can see, I don't actually understand the code, so it might be easily explained :)

src/strings.cpp
```diff
@@ -981,31 +1003,31 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
 					char *p = input + Utf8Encode(input, args->GetTypeAtOffset(offset));
 					*p = '\0';
 
-					/* Now do the string formatting. */
-					char buf[256];
+					/* Now do the string formatting; we need the first character, so that plus '\0' is enough. */
```
TrueBrain: "we need the first character"? I have no clue what you try to say here :)

src/strings.cpp
```diff
@@ -1625,38 +1618,34 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
-	if ((x & FACIL_BUS_STOP)   && (buff + Utf8CharLen(SCC_BUS)   < last)) buff += Utf8Encode(buff, SCC_BUS);
-	if ((x & FACIL_DOCK)       && (buff + Utf8CharLen(SCC_SHIP)  < last)) buff += Utf8Encode(buff, SCC_SHIP);
-	if ((x & FACIL_AIRPORT)    && (buff + Utf8CharLen(SCC_PLANE) < last)) buff += Utf8Encode(buff, SCC_PLANE);
-	*buff = '\0';
-	return buff;
+	if (x & FACIL_TRAIN)      builder.Utf8Encode(SCC_TRAIN);
+	if (x & FACIL_TRUCK_STOP) builder.Utf8Encode(SCC_LORRY);
+	if (x & FACIL_BUS_STOP)   builder.Utf8Encode(SCC_BUS);
+	if (x & FACIL_DOCK)       builder.Utf8Encode(SCC_SHIP);
+	if (x & FACIL_AIRPORT)    builder.Utf8Encode(SCC_PLANE);
```
TrueBrain: While at it, can we drop the silly alignment spaces here? :)

src/strings.cpp
```diff
@@ -1625,38 +1618,34 @@ static char *FormatString(char *buff, const char *str_arg, StringParameters *arg
...
+static void StationGetSpecialString(StringBuilder &builder, int x)
 {
-	if ((x & FACIL_TRAIN)      && (buff + Utf8CharLen(SCC_TRAIN) < last)) buff += Utf8Encode(buff, SCC_TRAIN);
-	if ((x & FACIL_TRUCK_STOP) && (buff + Utf8CharLen(SCC_LORRY) < last)) buff += Utf8Encode(buff, SCC_LORRY);
-	if ((x & FACIL_BUS_STOP)   && (buff + Utf8CharLen(SCC_BUS)   < last)) buff += Utf8Encode(buff, SCC_BUS);
-	if ((x & FACIL_DOCK)       && (buff + Utf8CharLen(SCC_SHIP)  < last)) buff += Utf8Encode(buff, SCC_SHIP);
-	if ((x & FACIL_AIRPORT)    && (buff + Utf8CharLen(SCC_PLANE) < last)) buff += Utf8Encode(buff, SCC_PLANE);
-	*buff = '\0';
-	return buff;
+	if (x & FACIL_TRAIN)      builder.Utf8Encode(SCC_TRAIN);
```
TrueBrain: While at it, I wonder if we should make those bit-checks explicit, as I think we normally do that? `((x & FACIL_XXX) != 0)` stuff.

src/strings_internal.h
```diff
@@ -118,6 +118,17 @@ class StringBuilder {
...
 		return (ptrdiff_t)(this->last - this->current);
 	}
+
+	/**
+	 * Migration helper function to add a string using the old/to be deprecated
+	 * method of: new_destination = function(destination, last).
+	 * @param function The function to pass the current and last location to,
+	 *                 that will then return the new current location.
+	 */
+	void AddOldStyle(std::function<char*(char*, const char*)> function)
```
TrueBrain: I know you have the intention to remove this eventually, but the name is a bit odd. `AddCString`? I dunno .. but `Old` is not a word that is helping for context what this does :)

TrueBrain (commented):
Nothing code-wise, all about the structure of things; sorry about that. But it is really hard to follow sometimes, when you are not deep into this stuff :)

src/strings.cpp
rubidium42: Added some comments in the code what is happening/why it is happening, but it that doesn't make it show as outdated here.

src/strings.cpp
```diff
@@ -806,18 +828,26 @@ static std::vector<const char *> _game_script_raw_strings;
 
 /**
  * Parse most format codes within a string and write the result to a buffer.
- * @param buff    The buffer to write the final string to.
+ * @param builder The string builder to write the final string to.
  * @param str_arg The original string with format codes.
  * @param args    Pointer to extra arguments used by various string codes.
- * @param last    Pointer to just past the end of the buff array.
  * @param dry_run True when the argt array is not yet initialized.
```
TrueBrain: lol; `argt`. Sorry, not your fault, but typo :P If there is nothing else, you can ignore this comment :)

src/strings.cpp
```diff
@@ -806,18 +828,26 @@ static std::vector<const char *> _game_script_raw_strings;
...
  */
-static char *FormatString(char *buff, const char *str_arg, StringParameters *args, const char *last, uint case_index, bool game_script, bool dry_run)
+static void FormatString(StringBuilder &builder, const char *str_arg, StringParameters *args, uint case_index, bool game_script, bool dry_run)
 {
 	uint orig_offset = args->offset;
 
 	/* When there is no array with types there is no need to do a dry run. */
```
TrueBrain: This comment .. seems wrong or outdated. I guess by now it is more what you wrote in the blob below.

`args` isn't initialized, and this wasn't a dry-run, so call ourselves again to do a dry-run, so `args` becomes initialized.

I would say, it is almost better to remove this comment, honestly.

TrueBrain (dismissed):

TrueBrain (approved):

PR #10928, opened 2023-06-04
Fix: disable hardware acceleration when GPU driver crashed the game last attempt

### LINKED ISSUE #10394: [Crash]: Fatal application Failure on startup (graphics driver crash)
### Version of OpenTTD

12.2, windows 10

### Steps to reproduce

attempted to start the game from steam
have tested different versions of OTTD(Including nightly) and reinstalling same error and crash

### Upload crash files

[crash20230122121903.zip](https://github.com/OpenTTD/OpenTTD/files/10474394/crash20230122121903.zip)
[crash20230122122619.zip](https://github.com/OpenTTD/OpenTTD/files/10474418/crash20230122122619.zip)

### LINKED ISSUE #10463: [Crash]: When I Opened the Game, my screen became flash, then game shut off.
### Version of OpenTTD

1.13.0 Win10

### Steps to reproduce

Just open the game.

### Upload crash files

[Crash.zip](https://github.com/OpenTTD/OpenTTD/files/10704630/Crash.zip)

### LINKED ISSUE #10571: [Bug]: (Steam) Black screen on startup, can hear background noise.  Cannot get past.
### Version of OpenTTD

13.0, Windows 10

### Expected result

I have uninstalled and reinstalled multiple times to no avail.  And google doesn't seem to be very helpful...

### Actual result

I hit play in steam.  
  
The game starts up and creates a black screen, though I can hear the vehicle noises that the background creates from the main menu.  
  
I can only hit windows key to get out of it and close the game from steam.

### Steps to reproduce

I don't know how to reproduce nor figure out how to get it to stop.

TrueBrain:
## Motivation / Problem

We have a slow but steady report of "game crashes on startup" over the last few months. All these cases come down to:
- We start the OpenGL driver
- The driver crashes for one reason or the other

This leaves the user confused, and they have to dig through tons of things to find out they have to add `-v win32` to start the game. Only then they can disable hardware acceleration.

I was wondering: can't we detect if we crashed, and on next start, disable hardware acceleration for the user?

Fixes #10463 
Fixes #10394
Fixes #10571

## Description

I build this PR on the assumption that `Start()` causes the driver crash. A few backtraces show this is the case, but a few others we cannot decode; so I am not sure this covers all cases. It might still happen that a crash happens when we actually start drawing for the first time, but we can always extend this code to that later on. For example, by waiting for the first tick to remove the detection file.

Either way, this PR creates a file `hwaccel.dat` when starting with hardware acceleration. As soon as the first drawing tick finishes, we remove it again. If we try to initialize the video driver with acceleration while that file exist, it means we must have crashed last time. We then disable hardware acceleration.

PS: some changes recently caused those video-driver errors to no longer show up, as they weren't marked critical (any more). I fixed that while I was here. No clue what happened exactly; it used to work, it no longer does, it does now again. *shrug*

## Limitations

Ideally, we find out what gets those drivers to crash; but that has turned out to be very difficult, and often just: known driver bug. Some people seem to never update their drivers, other GPUs haven't release a new drivers in months, and just randomly crash.

## Checklist for review

### DIFF
--- a/src/driver.cpp
+++ b/src/driver.cpp
@@ -16,7 +16,14 @@
 #include "video/video_driver.hpp"
 #include "string_func.h"
 #include "table/strings.h"
+#include "fileio_func.h"
 #include <sstream>
+
+#ifdef _WIN32
+# include <windows.h>
+#else
+# include <unistd.h>
+#endif /* _WIN32 */
 
 #include "safeguards.h"
 
@@ -31,6 +38,8 @@ std::string _ini_musicdriver;        ///< The music driver a stored in the confi
 
 std::string _ini_blitter;            ///< The blitter as stored in the configuration file.
 bool _blitter_autodetected;          ///< Was the blitter autodetected or specified by the user?
+
+static const std::string HWACCELERATION_TEST_FILE = "hwaccel.dat"; ///< Filename to test if we crashed last time we tried to use hardware acceleration.
 
 /**
  * Get a string parameter the list of parameters.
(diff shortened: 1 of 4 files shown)


glx22 (approved):
Can't really test myself, but it makes sense.

PR #10931, opened 2023-06-04
Codechange: use std::string for the screenshot name/path

rubidium42:
## Motivation / Problem

The screenshot name/path are a C-style string.

## Description

Replace them with `std::string`.
Then use assignment instead of `strecpy`/`strecat`, and `fmt::format` instead of `seprintf`.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/screenshot.cpp
+++ b/src/screenshot.cpp
@@ -40,8 +40,8 @@ static const char * const HEIGHTMAP_NAME  = "heightmap";  ///< Default filename 
 std::string _screenshot_format_name;  ///< Extension of the current screenshot format (corresponds with #_cur_screenshot_format).
 uint _num_screenshot_formats;         ///< Number of available screenshot formats.
 uint _cur_screenshot_format;          ///< Index of the currently selected screenshot format in #_screenshot_formats.
-static char _screenshot_name[128];    ///< Filename of the screenshot file.
-char _full_screenshot_name[MAX_PATH]; ///< Pathname of the screenshot file.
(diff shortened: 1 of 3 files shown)


src/screenshot.cpp
```diff
@@ -668,43 +668,40 @@ static void LargeWorldCallback(void *userdata, void *buf, uint y, uint pitch, ui
...
 	const char *screenshot_dir = crashlog ? _personal_dir.c_str() : FiosGetScreenshotDir();
 
 	for (uint serial = 1;; serial++) {
-		if (seprintf(_full_screenshot_name, lastof(_full_screenshot_name), "%s%s", screenshot_dir, _screenshot_name) >= (int)lengthof(_full_screenshot_name)) {
-			/* We need more characters than MAX_PATH -> end with error */
-			_full_screenshot_name[0] = '\0';
-			break;
-		}
+		_full_screenshot_path = screenshot_dir;
+		_full_screenshot_path += _screenshot_name;
```
TrueBrain: Just curious: `fmt::format("{}{}", screenshot_dir, _screenshot_name)` is a bit more readable, I think. But I can imagine that the current format is faster? Is there a reason for one over the other?

rubidium42: I'm not sure which one's faster, as that might highly depend on the situation. Though screenshot file names are not a place where we should really focus on the performance, but rather readability and functionality.

TrueBrain (approved):

PR #10932, opened 2023-06-04
Fix: crash when window can't be placed on low resolution screens.

### LINKED ISSUE #8869: Setting resolution really small in openttd.cfg crashes the game
## Version of OpenTTD
1.11.0-RC1

## Expected result
No crash, even if a user is being silly.

## Actual result
Crash.

## Steps to reproduce
In your config, set:
`resolution = 1,1`
Start the game
Enjoy the crash!

## Backtrace

```
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff70d1859 in __GI_abort () at abort.c:79
#2  0x00007ffff70d1729 in __assert_fail_base (fmt=0x7ffff7267588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
    assertion=0x55555674503d "min <= max",
    file=0x555556745000 "OpenTTD/src/core/math_func.hpp", line=79,
    function=<optimized out>) at assert.c:92
#3  0x00007ffff70e2f36 in __GI___assert_fail (assertion=0x55555674503d "min <= max",
    file=0x555556745000 "OpenTTD/src/core/math_func.hpp", line=79,
    function=0x555556744fe0 "T Clamp(T, T, T) [with T = int]") at assert.c:101
#4  0x00005555560ba63e in Clamp<int> (a=0, min=0, max=-12)
    at OpenTTD/src/core/math_func.hpp:79
#5  0x00005555560b1371 in Clamp (a=0, min=0, max=-12)
    at OpenTTD/src/core/math_func.hpp:103
#6  0x00005555560b68ea in EnsureVisibleCaption (w=0x5555588a3220, nx=-12, ny=0)
    at OpenTTD/src/window.cpp:2127
#7  0x00005555560b6d26 in ResizeWindow (w=0x5555588a3220, delta_x=0, delta_y=0, clamp_to_screen=true)
    at OpenTTD/src/window.cpp:2177
#8  0x00005555560b3ea2 in Window::ReInit (this=0x5555588a3220, rx=0, ry=0)
    at OpenTTD/src/window.cpp:1022
#9  0x00005555560b9f55 in ReInitAllWindows () at OpenTTD/src/window.cpp:3466
#10 0x0000555555d9f606 in LoadStringWidthTable (monospace=false)
    at OpenTTD/src/gfx.cpp:1286
#11 0x0000555555fee021 in CheckForMissingGlyphs (base_font=true,
    searcher=0x5555583b4070 <CheckForMissingGlyphs(bool, MissingGlyphSearcher*)::pack_searcher>)
    at OpenTTD/src/strings.cpp:2143
#12 0x0000555555ee3e72 in LoadIntroGame (load_newgrfs=false)
    at OpenTTD/src/openttd.cpp:359
#13 0x0000555555ee5485 in openttd_main (argc=4, argv=0x7fffffffe038)
    at OpenTTD/src/openttd.cpp:830
#14 0x0000555555c1ad7b in main (argc=4, argv=0x7fffffffe038)
    at OpenTTD/src/os/unix/unix.cpp:265
```

### LINKED ISSUE #10530: [Crash]: Game crashes when closing the window if it is narrower than the confirmation message
### Version of OpenTTD

13.0, Windows 11

### Steps to reproduce

1. Start a game.
2. Resize the game window to be smaller than the confirmation popup that you get when closing the game window (with the X on the corner of the window, not ingame menu)
3. Close the window.
4. Crash occurs.

See https://github.com/JGRennison/OpenTTD-patches/issues/490 which was the initial report, and contains a probable fix.

### Upload crash files

[OpenTTD.zip](https://github.com/OpenTTD/OpenTTD/files/10869828/OpenTTD.zip)

TrueBrain:
## Motivation / Problem

Fixes #10530
Fixes #8869

When a resolution is too low, windows can't be placed anymore with `Clamp`, causing crashes.

## Description

@JGRennison created a `Softclamp`, which tolerates min > max, which is the actual problem here.

I cherry-picked the related commits, and made some minor changes for it to work for us.

Related commits:
- [cf26d48c9](https://github.com/JGRennison/OpenTTD-patches/commit/cf26d48c9b3f9097bc36509c82cbba9aa5166674) (partially cherry-picked, and fixed SoftClamp to return the midpoint in all cases)
- [36700631](https://github.com/JGRennison/OpenTTD-patches/commit/36700631aa7466e82449ed1597264e3d1f965e72)

## Checklist for review

### DIFF
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -84,6 +84,32 @@ static inline T Clamp(const T a, const T min, const T max)
 }
 
 /**
+ * Clamp a value between an interval.
+ *
+ * This function returns a value which is between the given interval of
+ * min and max. If the given value is in this interval the value itself
+ * is returned otherwise the border of the interval is returned, according
+ * which side of the interval was 'left'.
+ *
+ * @note If the min value is greater than the max, return value is the average of the min and max.
+ * @param a The value to clamp/truncate.
+ * @param min The minimum of the interval.
+ * @param max the maximum of the interval.
+ * @returns A value between min and max which is closest to a.
+ */
+template <typename T>
+static inline T SoftClamp(const T a, const T min, const T max)
+{
+	if (min > max) {
+		using U = std::make_unsigned_t<T>;
+		return min - (U(min) - max) / 2;
+	}
+	if (a <= min) return min;
+	if (a >= max) return max;
+	return a;
+}
+
+/**
  * Clamp an integer between an interval.
  *
  * This function returns a value which is between the given interval of
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -1140,8 +1140,8 @@ struct QueryWindow : public Window {
 	void FindWindowPlacementAndResize(int def_width, int def_height) override
 	{
 		/* Position query window over the calling window, ensuring it's within screen bounds. */
-		this->left = Clamp(parent->left + (parent->width / 2) - (this->width / 2), 0, _screen.width - this->width);
-		this->top = Clamp(parent->top + (parent->height / 2) - (this->height / 2), 0, _screen.height - this->height);
+		this->left = SoftClamp(parent->left + (parent->width / 2) - (this->width / 2), 0, _screen.width - this->width);
+		this->top = SoftClamp(parent->top + (parent->height / 2) - (this->height / 2), 0, _screen.height - this->height);
 		this->SetDirty();
 	}
 
--- a/src/tests/math_func.cpp
+++ b/src/tests/math_func.cpp
@@ -120,3 +120,13 @@ TEST_CASE("ClampTo")
 	/* max uint64_t gets clamped to max int64_t. */
 	CHECK(std::numeric_limits<int64_t>::max() == ClampTo<int64_t>(std::numeric_limits<uint64_t>::max()));
 }
+
+
+TEST_CASE("SoftClamp")
+{
+	/* Special behaviour of soft clamp returning the average of min/max when min is higher than max. */
+	CHECK(1250 == SoftClamp(0, 1500, 1000));
+	int million = 1000 * 1000;
+	CHECK(1250 * million == SoftClamp(0, 1500 * million, 1000 * million));
+	CHECK(0 == SoftClamp(0, 1500 * million, -1500 * million));
+}


src/core/math_func.hpp
```diff
@@ -83,6 +83,29 @@ static inline T Clamp(const T a, const T min, const T max)
...
 
+/**
+ * Clamp a value between an interval.
+ *
+ * This function returns a value which is between the given interval of
+ * min and max. If the given value is in this interval the value itself
+ * is returned otherwise the border of the interval is returned, according
+ * which side of the interval was 'left'.
+ *
+ * @note If the min value is greater than the max, return value is the average of the min and max.
```
PeterN, line 94: What's the rationale for this? Just "something is better than crashing"?

TrueBrain: Yup, exactly that! No reason to crash like this, as there is nothing actually wrong with it not fitting on the screen. So finding some value is better than saying: fuck this shit, I am out.

So yeah :)

PeterN (dismissed):

src/core/math_func.hpp
```diff
@@ -83,6 +83,29 @@ static inline T Clamp(const T a, const T min, const T max)
...
+ * @param a The value to clamp/truncate.
+ * @param min The minimum of the interval.
+ * @param max the maximum of the interval.
+ * @returns A value between min and max which is closest to a.
+ */
+template <typename T>
+static inline T SoftClamp(const T a, const T min, const T max)
+{
+	if (min > max) return (min + max) / 2;
```
rubidium42: This is broken: https://www.youtube.com/watch?v=sBtAGxBh-XI&t=278s but C++20 to the rescue!

TrueBrain: These comments always puzzle me .. do you want us to address the problem? Or was it just meant as fun: hihi, strictly seen this can break! 

For our use-cases the current is ofc fine; but yeah .. let me know how you meant this comment, and we take it from there :)

TrueBrain: Right, so you added a unit-test that broke the CI, what this comment was referencing. Please, PLEASE, use a bit more words when you try to say something :) It really saves us a lot of time :P

Anyway, I fixed it .. bit rude to not fix it yourself, but okay :D :) (just kidding ofc; I am fine with fixing it)

PS: this still breaks in weird cases, like a very large negative number as max, and a very large positive number as min. But that is far outside of the usecase for this function. We could `assert` it. Any further improvements, as far as I know, get really complicated with even more corner cases to cover ..

TrueBrain: @JGRennison might be worth for you to know too, that `SoftClamp` could overflow and return invalid midpoints; I fixed it in this PR :)

rubidium42: In my opinion a function that is as generic as this one is, should also work right in reasonably conceivable cases.

After all, you add it here to be reused in many other places and based on the comment you will not know that it is broken in some cases. To make matters weirder, with any type T that is smaller than `int` this function will return the expected value als the math gets promoted to `int` first.

At the location where you currently use it, it's indeed reasonable to consider it to not be a problem because one value is always 0. Though if this starts to get used throughot the code I rather this known issue solved, than to find out about it in some weird bug report/crash where you are wondering why `SoftClamp` returns "garbage" in some cases.

TrueBrain: After finally finding somewhere what the C++ midpoint implementation was:

```
constexpr Integer midpoint(Integer a, Integer b) noexcept {
  using U = make_unsigned_t<Integer>;
  return a>b ? a-(U(a)-b)/2 : a+(U(b)-a)/2;
}
```

It has a nice point: making the left side unsigned also solves the corner case of extremely large `min` vs extremely small `max`. So added that to the mix. And added another test, to actually notice that scenario :) So it should now be 100% bug-free. Or something.

rubidium42 (approved):

PR #10938, opened 2023-06-04
Feature: Option to disable activate-on-release behaviour of toolbar dropdown buttons.

PeterN:
## Motivation / Problem

A common problem for new users is they don't realise the main toolbar buttons bring up a dropdown menu as it disappears straight after clicking.

Therefore they cannot find how to change railtypes nor find a few more features probably.

## Description

This change introduces an option to change this behaviour (by default) so that clicking and releasing simply opens the dropdown menu.

This is probably game-breaking for existing users but if the default value maintains existing behaviour then it won't benefit new users at all.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1690,6 +1690,9 @@ STR_CONFIG_SETTING_SCROLLMODE_LMB                               :Move map with L
 STR_CONFIG_SETTING_SMOOTH_SCROLLING                             :Smooth viewport scrolling: {STRING2}
 STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT                    :Control how the main view scrolls to a specific position when clicking on the smallmap or when issuing a command to scroll to a specific object on the map. If enabled, the viewport scrolls smoothly, if disabled it jumps directly to the targeted spot
 
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT                  :Traditional toolbar menu selection behaviour: {STRING2}
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT_HELPTEXT         :Choose selection behaviour of toolbar menus. If off, toolbar menus will stay open until a selection is made. If on, toolbar menus will activate when the mouse button is released.
+
 STR_CONFIG_SETTING_MEASURE_TOOLTIP                              :Show a measurement tooltip when using various build-tools: {STRING2}
 STR_CONFIG_SETTING_MEASURE_TOOLTIP_HELPTEXT                     :Display tile-distances and height differences when dragging during construction operations
 
--- a/src/settingentry_gui.cpp
+++ b/src/settingentry_gui.cpp
@@ -669,6 +669,7 @@ SettingsContainer &GetSettingsTree()
 				general->Add(new SettingEntry("gui.window_snap_radius"));
 				general->Add(new SettingEntry("gui.window_soft_limit"));
 				general->Add(new SettingEntry("gui.right_click_wnd_close"));
+				general->Add(new SettingEntry("gui.toolbar_dropdown_autoselect"));
 			}
 
 			SettingsPage *viewports = interface->Add(new SettingsPage(STR_CONFIG_SETTING_INTERFACE_VIEWPORTS));
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -185,6 +185,7 @@ struct GUISettings {
 	uint8_t  scrollwheel_multiplier;           ///< how much 'wheel' per incoming event from the OS?
 	bool   timetable_arrival_departure;      ///< show arrivals and departures in vehicle timetables
 	RightClickClose  right_click_wnd_close;  ///< close window with right click
+	bool   toolbar_dropdown_autoselect;      ///< should toolbar dropdown buttons autoselect when releasing the mouse button
 	bool   pause_on_newgame;                 ///< whether to start new games paused or not
 	SignalGUISettings signal_gui_mode;       ///< select which signal types are shown in the signal GUI
 	SignalCycleSettings cycle_signal_types;  ///< Which signal types to cycle with the build signal tool.
--- a/src/table/settings/gui_settings.ini
+++ b/src/table/settings/gui_settings.ini
@@ -147,6 +147,14 @@ strhelp  = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_HELPTEXT
 strval   = STR_CONFIG_SETTING_RIGHT_MOUSE_WND_CLOSE_NO
 cat      = SC_BASIC
 
+[SDTC_BOOL]
+var      = gui.toolbar_dropdown_autoselect
+flags    = SettingFlag::NotInSave, SettingFlag::NoNetworkSync
+def      = false
+str      = STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT
+strhelp  = STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT_HELPTEXT
+cat      = SC_BASIC
+
 ; We might need to emulate a right mouse button on mac
 [SDTC_VAR]
 ifdef    = __APPLE__
--- a/src/toolbar_gui.cpp
+++ b/src/toolbar_gui.cpp
@@ -116,7 +116,7 @@ public:
  */
 static void PopupMainToolbarMenu(Window *w, WidgetID widget, DropDownList &&list, int def)
 {
-	ShowDropDownList(w, std::move(list), def, widget, 0, true);
+	ShowDropDownList(w, std::move(list), def, widget, 0, _settings_client.gui.toolbar_dropdown_autoselect);
 	SndClickBeep();
 }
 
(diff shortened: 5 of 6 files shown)


src/lang/english.txt
```diff
@@ -1568,6 +1568,9 @@ STR_CONFIG_SETTING_SCROLLMODE_LMB                               :Move map with L
 STR_CONFIG_SETTING_SMOOTH_SCROLLING                             :Smooth viewport scrolling: {STRING2}
 STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT                    :Control how the main view scrolls to a specific position when clicking on the smallmap or when issuing a command to scroll to a specific object on the map. If enabled, the viewport scrolls smoothly, if disabled it jumps directly to the targeted spot
 
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_CLOSE                       :Toolbar dropdowns activate on release: {STRING2}
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_CLOSE_HELPTEXT              :Control whether tthe main toolbar dropdowns activate their default function when released, or if they require an additional selection.
```
TrueBrain: tthe? :D Drop the t please :) Just one, not both.

TrueBrain: This sentence is really hard to read, and even after reading it 4 times, I am not exactly sure what it does :P Maybe dumb it down a bit?

TrueBrain (commented):
I like the idea of this PR :)

Does it still mean that if you click it once, it selects the default? Or do you always have to click twice?

PeterN:
You can click-drag-release to select an option with only one click, but simply clicking on the button will just leave the dropdown open, and require another click.

TrueBrain:
> You can click-drag-release to select an option with only one click, but simply clicking on the button will just leave the dropdown open, and require another click.

Okay, current players will find that annoying; as would I. But for new players this will be so much better. So I am wondering if we can make that a bit easier on both parties .. but I still like the idea of this PR a lot :)

2TallTyler:
I wonder if we should introduce a framework to do what the survey opt-in did, but to ask users to choose their preference when a new setting is introduced. Something like:

```
We've added a new setting. Here's what it does. Would you to try the new default or use the current/legacy behavior?

[use new default] [keep current setting]
```

It might be annoying, but might also cut down on "where did my signals go?" comments. 😃 And we already sort of do this with new engine previews. 😛

TrueBrain:
The other suggestion discussed on Discord the other day, was to use the fact the configuration file is versioned. So if there is an existing config, keep the old setting. And if there is none, use the new.

I like both ideas; one is just more work than the other :P In general, a "What's new" dialog could also be good :)

Kuhnovic:
I like the idea, it's definitely going to be helpful for new players. Keep in mind that we probably need to write a whole bunch of tooltips for these menu items since you can now actually hover over them.

> You can click-drag-release to select an option with only one click, but simply clicking on the button will just leave the dropdown open, and require another click.

Reading this I don't think it's going to be that annoying to players that are used to the old behavior, since it sounds like the old way of mousepress-drag-mouserelease still works. The only thing that would be different to us oldtimers is if we misclicked on the main toolbar button and suddenly the menu would stay, but one click would solve that.

Idea 1: hitting the close-all-windows-hotkey should probably close these dropdown menu's too.

Idea 2: maybe the dropdown should close the moment the user moves the mouse cursor outside of it. Not sure if that would feel natural or not, but it might be worth a try.

EmperorJake:
While the old behaviour was always completely natural to me, I'm surprised at how many new players are confused by it so I would support this change being enabled by default. 
As I said I would use this new behaviour even as a veteran player, because selecting a railtype from the bottom of a long list is very difficult when using a trackpad, especially if the list is long enough to have a scroll bar.

Also +1 on the delete hotkey closing any open dropdowns.

Andrew350:
So I messed around with this on the online preview for about 20 minutes (awesome feature by the way) and had some thoughts.

Overall I found this very intuitive even as a veteran player, mostly because just about every other game in existence for the last 25 years or so has done drop-down/pop-out menus like this. It did trip up my muscle memory at first, but it was actually very easy to adapt - for the most part. I still find it jarring for things where I'm used to doing a one-click action, for example saving a game or auto-selecting the most used railtype (which by the way, this makes that setting basically useless), because it takes one more click than before in those cases. 

There are just certain buttons I have engrained into my head that only need one quick click to do the action I want and it is tough to break that habit 😛 But for everything else where the usual click-and-drag method was necessary I actually find this a welcome change and am 1000% certain it is better for new players. I'm not sure if there's some way of doing a sort of hybrid system (perhaps something like Kuhnovic's suggestion) to try to preserve the best of both worlds, but if not I still think this PR is a better default option.

Whether the aforementioned inconvenience of losing the use of those one click actions will be welcomed by other seasoned players is another story...but I am at least willing to give it a chance, and those who don't can always change it back 🙂

Also, the delete key does indeed close the drop-downs which is nice, however clicking on the same menu button again does not which I feel is unintuitive, but maybe that's just an oversight. Also thinking if the "right click to close windows" setting is on you should be able to close dropdowns on right click as well, if that's possible.

EmperorJake:
I've given this a test now, and I'd say a change that absolutely needs to be made is for clicking the button again to make the menu disappear. This behaviour would be consistent with most other programs/OS that use dropdowns like this, and so it's what I'm used to doing. And if the dropdown has only one item it shouldn't need a dropdown at all, but instead behave like a simple button like it did previously.

But that's all I can complain about, it's a good change and it makes navigating a long list of railtypes more convenient on either a mouse or trackpad. 

Also I'd argue that most-used railtype doesn't become useless, as I make heavy use of the A for autorail hotkey.

PeterN:
Yeah, not disappearing when pressing the button again is a bug, looking into it :-)

2TallTyler:
See also: #11338.

mdias:
This PR was just brought to my attention as it appears I've made basically a duplicate of it (#11338).

It seems that we've mostly implemented it the same way, but I managed to fix the "click button again for the dropdown to disappear". Please look into the `window.cpp` change on my PR for that.

If you decide to incorporate that change please let me know and I'll close my PR.

2TallTyler:
Both @PeterN and @mdias have implemented this feature as separate PRs. At this point I think it's fair game for either of you (or anyone else) to borrow code from the other and update their PR with one perfect solution, that we can get into the upcoming 14.0 release. 😄

mdias:
Since @PeterN's PR is considerably older than mine, I think it's fair for his to go in.
I also don't mind if he incorporates my fix to the button behaving properly when clicked again mentioned by [a previous comment](https://github.com/OpenTTD/OpenTTD/pull/10938#issuecomment-1590375501).

nielsmh:
This has been requested on the forums also: https://www.tt-forums.net/viewtopic.php?t=91878

AnthonyDiGirolamo:
New player here, would 100% love this. I'm on a laptop with a touchscreen (GPD Win Max 2) and tapping to open a toolbar menu doesn't work since it immediately disappears again. The Android port seems to do what I'd prefer by default.

mdias:
I'm not sure what needs doing at this point to merge it.
If there's something depending on me, please let me know.

src/lang/english.txt
```diff
@@ -1690,6 +1690,9 @@ STR_CONFIG_SETTING_SCROLLMODE_LMB                               :Move map with L
 STR_CONFIG_SETTING_SMOOTH_SCROLLING                             :Smooth viewport scrolling: {STRING2}
 STR_CONFIG_SETTING_SMOOTH_SCROLLING_HELPTEXT                    :Control how the main view scrolls to a specific position when clicking on the smallmap or when issuing a command to scroll to a specific object on the map. If enabled, the viewport scrolls smoothly, if disabled it jumps directly to the targeted spot
 
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT                  :Toolbar dropdowns automatically activate: {STRING2}
```
2TallTyler: This is a hard thing to explain, so what about a different approach of not trying to fully explain it? Something like `Quick toolbar dropdowns` or even `Legacy toolbar dropdowns?

src/lang/english.txt
```diff
@@ -1690,6 +1690,9 @@ STR_CONFIG_SETTING_SCROLLMODE_LMB                               :Move map with L
...
+STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT_HELPTEXT         :Control whether the main toolbar dropdowns activate their default function when the mouse button is released. If not enabled then the dropdown menu will stay open so a selection can be made.
```
```suggestion
STR_CONFIG_SETTING_TOOLBAR_DROPDOWN_AUTOSELECT_HELPTEXT         :Enable legacy main toolbar behavior. If enabled, clicking on a main toolbar button will activate its default function while holding the mouse button will open the dropdown, selecting whichever options is hovered over when released. If disabled, clicking on a main toolbar button will open the dropdown menu without needing to hold the mouse button.
```

PeterN: I think "legacy" is a bit strong, the current way is not being deprecated.

2TallTyler: Traditional?

PeterN: IMHO this text is too wordy, and is wrong. We can omit unnecessary details but if we include them it should be correct. (The dropdown menus are always displayed when clicking, they just close again after a brief time.)

2TallTyler (approved):

PR #10947, opened 2023-06-04
Codechange: convert C-style GetTownName API to std::string returning API

rubidium42:
## Motivation / Problem

C-style string APIs.

## Description

Replace the `strecpy`-esque C-style API with one that returns `std::string`.

## Limitations

None

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -207,9 +207,7 @@ void Town::InitializeLayout(TownLayout layout)
 
 void Town::FillCachedName() const
 {
-	char buf[MAX_LENGTH_TOWN_NAME_CHARS * MAX_CHAR_LENGTH];
-	char *end = GetTownName(buf, this, lastof(buf));
-	this->cached_name.assign(buf, end);
+	this->cached_name.assign(GetTownName(this));
 }
 
 /**
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -207,7 +207,7 @@ void Town::InitializeLayout(TownLayout layout)
 
 void Town::FillCachedName() const
 {
-	this->cached_name.assign(GetTownName(this));
+	this->cached_name = GetTownName(this);
 }
 
 /**
--- a/src/townname.cpp
+++ b/src/townname.cpp
@@ -58,18 +58,18 @@ static void GetTownName(StringBuilder &builder, const TownNameParams *par, uint3
 }
 
 /**
- * Fills buffer with specified town name
- * @param buff buffer start
- * @param par town name parameters
- * @param townnameparts 'encoded' town name
- * @param last end of buffer
- * @return pointer to terminating '\0'
+ * Get the town name for the given parameters and parts.
+ * @param par Town name parameters.
+ * @param townnameparts 'Encoded' town name.
+ * @return The town name.
  */
-char *GetTownName(char *buff, const TownNameParams *par, uint32 townnameparts, const char *last)
+std::string GetTownName(const TownNameParams *par, uint32 townnameparts)
 {
-	StringBuilder builder(&buff, last);
+	char buffer[DRAW_STRING_BUFFER];
+	char *state = buffer;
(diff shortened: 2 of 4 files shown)


src/town_cmd.cpp
```diff
@@ -207,9 +207,7 @@ void Town::InitializeLayout(TownLayout layout)
 
 void Town::FillCachedName() const
 {
-	char buf[MAX_LENGTH_TOWN_NAME_CHARS * MAX_CHAR_LENGTH];
-	char *end = GetTownName(buf, this, lastof(buf));
-	this->cached_name.assign(buf, end);
+	this->cached_name.assign(GetTownName(this));
```
LordAro: Why assign, rather than the standard = ?

src/townname.cpp
```diff
@@ -107,28 +104,19 @@ char *GetTownName(char *buff, const Town *t, const char *last)
...
 		for (const Town *t : Town::Iterate()) {
 			/* We can't just compare the numbers since
 			 * several numbers may map to a single name. */
-			const char *buf = t->name.empty() ? nullptr : t->name.c_str();
-			if (buf == nullptr) {
-				GetTownName(buf2, t, lastof(buf2));
-				buf = buf2;
-			}
-			if (strcmp(buf1, buf) == 0) return false;
+			if (name == (t->name.empty() ? GetTownName(t) : t->name)) return false;
```
LordAro: This condition feels confusing

src/town_cmd.cpp
rubidium42:
```suggestion
	this->cached_name = GetTownName(this);
```
Probably because I made some `TextBuf` changes before, which uses `Assign` so I didn't really think about it?

src/townname.cpp
rubidium42: The alternative would be something like the following:
```suggestion
			if (t->name.empty()) {
				if (name == GetTownName(t)) return false;
			} else {
				if (name == t->name) return false;
			}
```
Although I could do a `std::string town_name = t->name.empty() ? GetTownName(t) : t->name;` and then compare on that. But that would copy the name, which seems kinda pointless to me.

So the main question is: is the suggestion better than the existing PR or not? And if that's not the case, how could it be made less confusing without the need to copy the town name?

LordAro: The suggestion seems fine to me. Could use std::string_view with the alternative?

TrueBrain: Be careful that we don't sit too much on the premature optimization; even in your suggestion, -O1 already doesn't make the copy anymore. If you put `const` in front of it, GCC already doesn't with -O0 (but clang and MSVC do). But if you make it a `const std::string &`, none of them make a copy :) But that ofc is just a `std::string_view` at that point :)

(sorry, I had some fun with godbolt :P I am always curious how much we have to keep those things in mind :))

rubidium42: > The suggestion seems fine to me. Could use std::string_view with the alternative?

That would leave a dangling `std::string_view`, since `GetTownName` returns a `std::string` and that would go out of scope after the statement, i.e. after the view has been assigned. It will likely work fine most of the time, since the use is just after the release though.

TrueBrain: Just from an academical point of view, and as I like sharing things, I made two godbolts, one with:

```
        if (compare == (right.empty() ? GetTown() : right)) return 1;
```

The other with:

```
        std::string test = right.empty() ? GetTown() : right;
        if (compare == test) return 1;
```

Did you know that the ternary operator already creates a new `std::string` in both cases? So both are nearly identical in result. Especially to how many `std::string`s are created :) So yeah, that solution would just have been fine :)

Your new solution works fine too ofc, just has 2 compares instead of one. Same amount of strings are allocated in all 3 solutions. Compilers are fun :D Right, enough nitpicking on a few lines of code :P

PS: personally I think the second option here is way more readable than the other two.

LordAro (approved):

PR #10953, opened 2023-06-06
Add: Station variable 6B to get extended station id of nearby tiles

kiwitreekor:
## Motivation / Problem
 * Recently station id was extended to 16 bit, but station variable 0x68 is still not available to access station ids over 255.

## Description
 * Adds a new variable to allow access to station IDs greater than 255.

## Checklist for review

### DIFF
--- a/src/newgrf_roadstop.cpp
+++ b/src/newgrf_roadstop.cpp
@@ -171,6 +171,23 @@ uint32_t RoadStopScopeResolver::GetVariable(byte variable, [[maybe_unused]] uint
 			return ssl.grfid;
 		}
 
+		/* 16 bit road stop ID of nearby tiles */
+		case 0x6B: {
+			TileIndex nearby_tile = GetNearbyTile(parameter, this->tile);
+
+			if (!IsRoadStopTile(nearby_tile)) return 0xFFFFFFFF;
+			if (!IsCustomRoadStopSpecIndex(nearby_tile)) return 0xFFFE;
+
+			uint32_t grfid = this->st->roadstop_speclist[GetCustomRoadStopSpecIndex(this->tile)].grfid;
+
+			const RoadStopSpecList ssl = BaseStation::GetByTile(nearby_tile)->roadstop_speclist[GetCustomRoadStopSpecIndex(nearby_tile)];
+			if (ssl.grfid == grfid) {
+				return ssl.localidx;
+			}
+
+			return 0xFFFE;
+		}
+
 		case 0xF0: return this->st == nullptr ? 0 : this->st->facilities; // facilities
 
 		case 0xFA: return ClampTo<uint16_t>((this->st == nullptr ? TimerGameCalendar::date : this->st->build_date) - CalendarTime::DAYS_TILL_ORIGINAL_BASE_YEAR); // build date
--- a/src/newgrf_station.cpp
+++ b/src/newgrf_station.cpp
@@ -376,6 +376,22 @@ TownScopeResolver *StationResolverObject::GetTown()
 			return ssl.grfid;
 		}
 
+		case 0x6B: { // 16 bit Station ID of nearby tiles
+			TileIndex nearby_tile = GetNearbyTile(parameter, this->tile);
+
+			if (!HasStationTileRail(nearby_tile)) return 0xFFFFFFFF;
+			if (!IsCustomStationSpecIndex(nearby_tile)) return 0xFFFE;
+
+			uint32_t grfid = this->st->speclist[GetCustomStationSpecIndex(this->tile)].grfid;
+
+			const StationSpecList ssl = BaseStation::GetByTile(nearby_tile)->speclist[GetCustomStationSpecIndex(nearby_tile)];
+			if (ssl.grfid == grfid) {
+				return ssl.localidx;
+			}
+
+			return 0xFFFE;
+		}
+
 		/* General station variables */
 		case 0x82: return 50;
 		case 0x84: return this->st->string_id;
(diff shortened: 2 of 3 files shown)


PeterN:
Would it make sense to include a test (or bit) for whether it is in the same NewGRF?

src/newgrf_roadstop.cpp
```diff
@@ -171,6 +171,17 @@ uint32 RoadStopScopeResolver::GetVariable(byte variable, uint32 parameter, bool
...
 		}
 
+		/* 16 bit Station ID of nearby tiles */
+		case 0x6B: { 
+			TileIndex nearby_tile = GetNearbyTile(parameter, this->tile);
+
+			if (!HasStationTileRail(nearby_tile)) return 0xFFFF;
+			if (!IsCustomStationSpecIndex(nearby_tile)) return 0;
+
+			const StationSpecList ssl = BaseStation::GetByTile(nearby_tile)->speclist[GetCustomStationSpecIndex(nearby_tile)];
```
JGRennison: This should use IsRoadStopTile, IsCustomRoadStopSpecIndex, RoadStopSpecList, roadstop_speclist and GetCustomRoadStopSpecIndex instead of the rail station functions/types.

kiwitreekor: Fixed!

kiwitreekor:
> Would it make sense to include the test / bits for whether it is in the same NewGRF?

It can be checked by bit 8 of variable 68 or variable 6A, so I think it's not necessary

frosch123:
Other variables return a 32bit value 0xFFFFFFFF for "no station"/"no roadstop". I guess the same value should be used here?

I see that the old var 68 also returns the local-index for other grfs and 0 for the default station. I kind of prefer the behavior of industries (indvar 60, indtilevar 62) returning some distinct invalid value for them.
Alternatively the house behavior (house var 66) also sounds sane, which adds a bit "other grf".

The distinct value makes it easy to switch in a NewGRF depending on known tiles, and not accidentially confuse other GRF's tiles for your own.
The "other grf" bit is kind of harder: In NFO you could just choose to not mask the bit. But in NML you would always have to  check a separate "same grf" variable, which is easily forgotten and technically already available via var 68.

kiwitreekor:
Now I changed var6b to return 0xFFFFFFFF if the selected tile isn't a station tile, 0xFFFE if the selected tile is from other grf/default station, like industry var 60.

2TallTyler:
PR checker is failing for trailing whitespace.

kiwitreekor:
now fixed trailing whitespace and rebased to current master...

2TallTyler:
Needs a re-review from a qualified reviewer. I'm adding this to the 14.0 milestone so (hopefully) it doesn't slip through the cracks.

src/newgrf_roadstop.cpp
```diff
@@ -171,6 +171,22 @@ uint32 RoadStopScopeResolver::GetVariable(byte variable, uint32 parameter, bool
 			return ssl.grfid;
 		}
 
+		/* 16 bit road stop ID of nearby tiles */
+		case 0x6B: {
+			TileIndex nearby_tile = GetNearbyTile(parameter, this->tile);
+			uint32 grfid = this->st->roadstop_speclist[GetCustomRoadStopSpecIndex(this->tile)].grfid;
```
michicc: Could/should be moved to right before usage, to save some CPU cycles unless actually needed.

src/newgrf_station.cpp
```diff
@@ -376,6 +376,21 @@ TownScopeResolver *StationResolverObject::GetTown()
 			return ssl.grfid;
 		}
 
+		case 0x6B: { // 16 bit Station ID of nearby tiles
+			TileIndex nearby_tile = GetNearbyTile(parameter, this->tile);
+			uint32 grfid = this->st->speclist[GetCustomStationSpecIndex(this->tile)].grfid;
```
michicc: As above.

src/newgrf_roadstop.cpp
```diff
@@ -171,6 +171,22 @@ uint32 RoadStopScopeResolver::GetVariable(byte variable, uint32 parameter, bool
...
+
+			if (!IsRoadStopTile(nearby_tile)) return 0xFFFFFFFF;
+			if (!IsCustomRoadStopSpecIndex(nearby_tile)) return 0xFFFE;
+
+			const RoadStopSpecList ssl = BaseStation::GetByTile(nearby_tile)->roadstop_speclist[GetCustomRoadStopSpecIndex(nearby_tile)];
+			if (ssl.grfid != grfid) {
```
TrueBrain: Here it reads `ssl.grfid != grfid`, but in the station variant it reads `ssl.grfid == grfid`.

I know very little about NewGRFs, but it looks to me that one of the two isn't right. Couldn't tell you which of the two.

src/newgrf_roadstop.cpp
```diff
@@ -171,6 +171,22 @@ uint32 RoadStopScopeResolver::GetVariable(byte variable, uint32 parameter, bool
 			return ssl.grfid;
 		}
 
+		/* 16 bit road stop ID of nearby tiles */
```
TrueBrain, line 174: Is the "16 bit" part important? As we don't mention how many bits for any of the other?

src/newgrf_roadstop.cpp
JGRennison: This one is the incorrect one. The GRF local ID has no meaning to other GRFs.

TrueBrain:
I fixed the comments and rebased the PR, in an effort to get it in 14.0. Hope you don't mind @kiwitreekor . 

Can't just the PR itself, but hopefully someone else can now say: yes, or what else needs taken care of first :)

michicc (approved):

PR #10959, opened 2023-06-06
Codechange: use std::filesystem::path for the language file's path

rubidium42:
## Motivation / Problem

Initially due to `seprintf` that I wanted to replace with `std::string`. Then figuring out quite a lot of places only want the filename or stem (filename without extension), with their own implementations using `strrchr`. That would be quite complicated to write with `std::string` but trivial with `std::filesystem::path`.

## Description

Replace the char buffer for the language metadata with `std::filesystem::path`.
Replace C-style string logic to get the filename with `.filename()`.
Replace C-style string logic to get the filename without extension with `.stem()`.
Remove the full path from the filename in the crashlog.

## Limitations

Having difficulty with Windows. `path.c_str()` is probably `wchar_t *`, so using that around it troublesome. It also means that we require `path.string()` to convert to `std::string` which happens automatically on other platforms.
Furthermore regression fails because it seems to pick another language for some weird reason... but only on Windows?!?

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -42,6 +42,7 @@
 
 #include "table/strings.h"
 #include "table/control_codes.h"
+#include "3rdparty/fmt/std.h"
 
 #include "strings_internal.h"
 
@@ -1827,7 +1828,7 @@ bool ReadLanguagePack(const LanguageMetadata *lang)
 {
 	/* Current language pack */
 	size_t len = 0;
-	std::unique_ptr<LanguagePack, LanguagePackDeleter> lang_pack(reinterpret_cast<LanguagePack *>(ReadFileToMem(lang->file, len, 1U << 20).release()));
+	std::unique_ptr<LanguagePack, LanguagePackDeleter> lang_pack(reinterpret_cast<LanguagePack *>(ReadFileToMem(lang->file.string(), len, 1U << 20).release()));
 	if (!lang_pack) return false;
 
 	/* End of read data (+ terminating zero added in ReadFileToMem()) */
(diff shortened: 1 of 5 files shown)


PeterN (dismissed):

rubidium42:
At the moment this PR is essentially dead in the water, The warnings are kind-of bogus, as no other compiler including the 64-bit variant of the same compiler has any problem with it. Also without the filesystem header the compiler has no problem with it. So something weird is going on in the compiler, that I can't influence except for disabling said warning for the x86 runs with that compiler.

src/strings.cpp
```diff
@@ -2042,8 +2042,7 @@ static void GetLanguageList(const char *path)
 void InitializeLanguagePacks()
 {
 	for (Searchpath sp : _valid_searchpaths) {
-		std::string path = FioGetDirectory(sp, LANG_DIR);
-		GetLanguageList(path.c_str());
+		GetLanguageList(FioGetDirectory(sp, LANG_DIR));
```
LordAro, line 2045: this function being named `Get` but returning nothing is... unusual. `Set` seems better, ironically.

src/strings.cpp
```diff
@@ -1827,7 +1828,7 @@ bool ReadLanguagePack(const LanguageMetadata *lang)
 {
 	/* Current language pack */
 	size_t len = 0;
-	std::unique_ptr<LanguagePack, LanguagePackDeleter> lang_pack(reinterpret_cast<LanguagePack *>(ReadFileToMem(lang->file, len, 1U << 20).release()));
+	std::unique_ptr<LanguagePack, LanguagePackDeleter> lang_pack(reinterpret_cast<LanguagePack *>(ReadFileToMem(lang->file.string(), len, 1U << 20).release()));
```
LordAro, line 1831: this was horrifying the last time i saw this line, and it's horrifying still

LordAro (approved):

PR #10967, opened 2023-06-07
Codechange: use std::string for Windows' ISO code mangling

rubidium42:
## Motivation / Problem

Use of strecpy.

## Description

Pass ISO code as std::string, replace `strecpy` with `=` and `strcmp` with `==`.

## Limitations

None

## Checklist for review

### DIFF
--- a/src/os/windows/win32.cpp
+++ b/src/os/windows/win32.cpp
@@ -541,23 +541,21 @@ const char *GetCurrentLocale(const char *)
 
 static WCHAR _cur_iso_locale[16] = L"";
 
-void Win32SetCurrentLocaleName(const char *iso_code)
+void Win32SetCurrentLocaleName(std::string iso_code)
 {
 	/* Convert the iso code into the format that windows expects. */
-	char iso[16];
-	if (strcmp(iso_code, "zh_TW") == 0) {
-		strecpy(iso, "zh-Hant", lastof(iso));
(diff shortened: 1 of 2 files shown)


src/os/windows/win32.cpp
```diff
@@ -541,23 +541,21 @@ const char *GetCurrentLocale(const char *)
...
+	if (iso_code == "zh_TW") {
+		iso_code = "zh-Hant";
+	} else if (iso_code == "zh_CN") {
+		iso_code = "zh-Hans";
 	} else {
 		/* Windows expects a '-' between language and country code, but we use a '_'. */
-		strecpy(iso, iso_code, lastof(iso));
-		for (char *c = iso; *c != '\0'; c++) {
-			if (*c == '_') *c = '-';
+		for (char &c : iso_code) {
```
LordAro, line 553: Can you use string.replace here?

rubidium42: No, string.replace only has ranges (iterator-iterator, pos-count) as first parameter, not "some character" to replace.

There is a `std::replace` that could potentially be used: `std::replace(iso_code.begin(), iso_code.end(), '_', '-');` or even shorter with C++20: `std::ranges::replace(iso_code, '_', '-');`.

LordAro: Eh, leave it

LordAro (approved):

PR #10972, opened 2023-06-08
Codechange: move saveload string fixing code to saveload

rubidium42:
## Motivation / Problem

Looking at `str_fix_scc_encoded` and wondering how that is somehow generic, primarily because it still works using C-style string logic when working on a `std::string`.

## Description

Move the code, rename the function and pass the `std::string`. Then simplify the code where applicable, although it's still not free of C-style logic but that's a bit of a catch 22.
When the string is corrupt, just bail out and let the later call to `StrMakeValid` handle that.
Only overwrite the characters where needed.
Do not try to '\0' terminate the string, as that is already the case due to it being `std::string`.

## Limitations

Still some C-style Utf8 character messing, but that's likely going to be tricky to solve and definitely out of scope.

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -890,6 +890,27 @@ static inline size_t SlCalcStdStringLen(const void *ptr)
 	return len + SlGetArrayLength(len); // also include the length of the index
 }
 
+
+/**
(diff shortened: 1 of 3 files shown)


LordAro (approved):
Seems reasonable

PR #10976, opened 2023-06-09
Fix #10975: Clear a non-head engine's name

### LINKED ISSUE #10975: [Bug]: Train name wrongly marked as non-unique
### Version of OpenTTD

13.1, Windows 11

### Expected result

"Name must be unique" error only occurs when there is a train existing with the same name.

### Actual result

"Name must be unique" error occurs even when the train with the same name does not exist anymore. This seemingly continues even when fast-forwarded for a long time.

### Steps to reproduce

1. Buy two trains and name them say "A" and "B"
2. Drag the train "A" and couple it on train "B". Now there is only one train, "B".
3. Buy another train.
4. Try to set the name of this new train as "A".
5. The message "Can't name train... Name must be unique" is displayed.

merni-ns:
## Motivation / Problem

Fixes #10975.

## Description
When reconfiguring a vehicle such that it is no longer the head of the rake, clear the name of such a vehicle. This allows other trains to use that name. Anyway, only whole trains (not parts of them) should have a name.

## Limitations
In earlier behaviour, when you joined two trains, both having a name, into one train, the non-head engine still retained its name in a hidden way, and if you separated that engine again, it would show its old name again. Now, it just shows the train number. I believe this is acceptable, because anyway the orders of such an engine have been cleared.

## Checklist for review

### DIFF
--- a/src/train_cmd.cpp
+++ b/src/train_cmd.cpp
@@ -1327,6 +1327,7 @@ CommandCost CmdMoveRailVehicle(DoCommandFlag flags, VehicleID src_veh, VehicleID
 			/* Remove stuff not valid anymore for non-front engines. */
 			DeleteVehicleOrders(src);
 			src->unitnumber = 0;
+			src->name.clear();
 		}
 
 		/* We weren't a front engine but are becoming one. So


2TallTyler (approved):
Makes sense to me.

rubidium42:
Just a question... how does this function/change when you set the name of a train, sell it and then buy it again. I seem to remember that in those cases the train name remains the same.

Is that situation maybe technically another case of #10975?

PR #10979, opened 2023-06-09
Fix: Don't restore backed up vehicle name if it's no longer unique.

PeterN:
## Motivation / Problem

While testing #10976 it was noted that names can be restored with order backups, and I found that it restoring the name does not check for uniqueness.

## Description

Fix by checking `IsUniqueVehicleName()` before restoring the name. If it is not unique then the default name is used instead.

## Checklist for review

### DIFF
--- a/src/order_backup.cpp
+++ b/src/order_backup.cpp
@@ -18,6 +18,7 @@
 #include "station_map.h"
 #include "order_cmd.h"
 #include "group_cmd.h"
+#include "vehicle_func.h"
 
 #include "safeguards.h"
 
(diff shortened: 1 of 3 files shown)


src/order_backup.cpp
```diff
@@ -83,7 +84,7 @@ void OrderBackup::DoRestore(Vehicle *v)
 		InvalidateWindowClassesData(WC_STATION_LIST, 0);
 	}
 
-	v->CopyConsistPropertiesFrom(this);
+	if (IsUniqueVehicleName(this->name)) v->CopyConsistPropertiesFrom(this);
```
rubidium42: I would expect only the name to not be restored, not all the other flags. Maybe a parameter whether the name must be copied defaulted to `true`?

PeterN: Oops. Messed this up didn't I :/

rubidium42 (approved):

PR #10984, opened 2023-06-10
Fix #10982: No help text for gamelog command

### LINKED ISSUE #10982: [Bug]: No help text for `gamelog` command
### Version of OpenTTD

Since c7a9637834a96722be5a524c4c1e8932a45e3461

### Expected result

Running `help gamelog` should display help text for the `gamelog` command.

### Actual result

Running `help gamelog` does the same thing as running just `gamelog`, meaning there's no help text.
See: https://github.com/OpenTTD/OpenTTD/blob/2097719f269edf2e42ada3ebe7854534d27b4e11/src/console_cmds.cpp#L2106

### Steps to reproduce

Run the command: `help gamelog`

def4alt:
## Motivation / Problem

See #10982.

## Description

I have added a short description of the `gamelog` command when calling `help gamelog`:

> Print logged potentially unsafe actions. Usage: 'gamelog'.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -2105,6 +2105,11 @@ DEF_CONSOLE_CMD(ConListSettings)
 
 DEF_CONSOLE_CMD(ConGamelogPrint)
 {
+	if (argc == 0) {
+		IConsolePrint(CC_HELP, "Print logged fundamental changes to the game since the start. Usage: 'gamelog'.");
+		return true;
+	}
+
 	_gamelog.PrintConsole();
 	return true;
 }
--- a/src/gamelog.cpp
+++ b/src/gamelog.cpp
@@ -5,7 +5,7 @@
  * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
  */
 
-/** @file gamelog.cpp Definition of functions used for logging of important changes in the game */
+/** @file gamelog.cpp Definition of functions used for logging of fundamental changes to the game */
 
 #include "stdafx.h"
 #include "saveload/saveload.h"
--- a/src/gamelog.h
+++ b/src/gamelog.h
@@ -5,7 +5,7 @@
  * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
  */
 
-/** @file gamelog.h Functions to be called to log possibly unsafe game events */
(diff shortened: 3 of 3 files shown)


glx22 (dismissed):
Code is fine.
Maybe another reviewer has something to say about the message itself.

LordAro:
I'm not sure gamelog has anything particular to do with "unsafe" actions, it includes things like savegame version updates too. "changes to the game state/environment" perhaps?

def4alt:
I based on the description of the header file of `gamelog`:

https://github.com/OpenTTD/OpenTTD/blob/2097719f269edf2e42ada3ebe7854534d27b4e11/src/gamelog.h#L7-L9

And also on the print function of `gamelog`, where these events are called as 'actions':

https://github.com/OpenTTD/OpenTTD/blob/2097719f269edf2e42ada3ebe7854534d27b4e11/src/gamelog.cpp#L147-L168

TrueBrain:
Ha, that is some poorly worded doxygen string you found there :) It always reminds me that most of us aren't native English speakers, and that what sounded correct when writing, isn't when you read it back years later :P

"Unsafe" here gives the impression the log itself could cause corruption or demolish your machine. That of course isn't the case. It tries to reference the fact the actions might indicate something that could indicate the reason for a broken game, or a crashing game, etc. So it is about the action, not about the logging of the action.

But "possibly unsafe" could reference the "log" part or the "game events" part :D So yeah ... worded poorly :)

Additionally, although the original intent of the gamelog was to tell people off for removing GRFs, I think it is weird to also call it like that. It does much more, like also at what tick a user upgraded their OpenTTD version. That is never harmful.

Anyway, my analysis here isn't relevant; my suggestion to address it might :P

For the console text, I would suggest something like:
`Print stored actions about fundamental changes in the game since start`

And maybe we should fix the doxygen text too, into something like:
`Functions to be called to indicate a fundamental change to the game`

Sorry for nitpicking this; but we also know nobody is going to change the help-text in the next 10 years :P I am curious about your thought to these suggestions!

def4alt:
Maybe it would be better to write something like this `Functions to be called to log fundamental changes to the game`, cause we're not actually changing anything ;)

And also what do you think about this help text `Print logged fundamental changes to the game since the start`?

TrueBrain:
> Maybe it would be better to write something like this `Functions to be called to log fundamental changes to the game`, cause we're not actually changing anything ;)
> 
> And also what do you think about this help text `Print logged fundamental changes to the game since the start`?

Sounds good to me!

LordAro (approved):

PR #11000, opened 2023-06-12
Fix #10987: Double-close of dropdown stopped land-info tool working as default.

### LINKED ISSUE #10987: [Bug]: Clicking the land info button in the main toolbar no longer activates the land info tool without using the drop down
### Version of OpenTTD

master

### Expected result

Clicking the land info button in the main toolbar activates the land info tool without using the drop down

### Actual result

Clicking the land info button in the main toolbar no longer activates the land info tool without using the drop down

### Steps to reproduce

Clicking the land info button in the main toolbar

PeterN:
## Motivation / Problem

Clicking and releasing on the query toolbar icon is meant to select the land-info tool.

This did not work as during closing a window, OnFocusLost() is called, which then closes the window again. These two calls toggled the land-info tool one and off in the same action.

## Description

Resolve by not calling Window::Close in OnFocusLost() if the window is already closing.

## Checklist for review

### DIFF
--- a/src/console_gui.cpp
+++ b/src/console_gui.cpp
@@ -336,7 +336,7 @@ struct IConsoleWindow : Window
 		VideoDriver::GetInstance()->EditBoxGainedFocus();
 	}
 
-	void OnFocusLost() override
+	void OnFocusLost(bool closing) override
 	{
 		VideoDriver::GetInstance()->EditBoxLostFocus();
 	}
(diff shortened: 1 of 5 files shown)


TrueBrain (approved):

LordAro:
I wonder if there might be some scenario where there's a "window property" of "auto closes on losing focus", rather than it being manually done in the OnFocusLost function. Might be TMWFTLB

PeterN:
As there are only two windows that do it, and it would still need to track whether it is already closing, I don't think it would gain much.

PR #11001, opened 2023-06-13
Feature: Transparency option for cost and income indicators

zephyris:
## Motivation / Problem
The cost and income indicators (the rising red and green text) from vehicle income, construction costs, vehicle purchase, etc. are always visible.

This can become problematic. All other signs/text/labels can be hidden, either by transparency options (invisible signs, vehicle loading indicators) or the settings dropdown (notably town names). As far as I could tell, there was no way to hide the cost and income indicators - therefore added it as a transparency option.

This option is particularly for players who want clean interfaces or beauty screenshots. It is also useful for late game construction where (especially on fast forward) you can end up building in a cloud of green vehicle income text.

## Description

This feature adds a show/hide of cost and income indicators as a transparency option, as another transparency toggle button to the transparency control window. When transparent, no cost/income text. When not transparent, normal cost/income text. It toggles along with all other transparency with the normal `X` transparency hotkey. It can be locked, like the other transparency options, with `Ctrl+click`.

This implementation imitates handling of the loading/unloading indicators via a transparency button.

Now we can have beautiful screenshots and clean, text free, viewports!

## Limitations

I believe this is a fully functional minimal feature, however:

There is no dedicated hotkey. `Ctrl+<number>` is used for transparency toggling of individual features, but `1`-`9` are already in use. `Ctrl+0` could be used, but I thought was unintuitive.

`x` toggles transparency of all features including the cost and income indicators. This is a change in behaviour, which might be undesirable? eg. for a new player when building in transparent mode, no cost would be shown.

It is not addressed here, but a setting, similar to the loading/unloading indicators, could be added. Alternatively, cost and income indicator transparency could default to locked.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2650,7 +2650,7 @@ STR_TRANSPARENT_BUILDINGS_TOOLTIP                               :{BLACK}Toggle t
 STR_TRANSPARENT_BRIDGES_TOOLTIP                                 :{BLACK}Toggle transparency for bridges. Ctrl+Click to lock
 STR_TRANSPARENT_STRUCTURES_TOOLTIP                              :{BLACK}Toggle transparency for structures like lighthouses and antennas. Ctrl+Click to lock
 STR_TRANSPARENT_CATENARY_TOOLTIP                                :{BLACK}Toggle transparency for catenary. Ctrl+Click to lock
-STR_TRANSPARENT_LOADING_TOOLTIP                                 :{BLACK}Toggle transparency for loading indicators. Ctrl+Click to lock
+STR_TRANSPARENT_TEXT_TOOLTIP                                    :{BLACK}Toggle transparency for loading and cost/income text. Ctrl+Click to lock
 STR_TRANSPARENT_INVISIBLE_TOOLTIP                               :{BLACK}Set objects invisible instead of transparent
 
 # Linkgraph legend window
--- a/src/texteff.cpp
+++ b/src/texteff.cpp
@@ -118,10 +118,10 @@ void DrawTextEffects(DrawPixelInfo *dpi)
 {
 	/* Don't draw the text effects when zoomed out a lot */
 	if (dpi->zoom > ZOOM_LVL_OUT_8X) return;
-
+	if (IsTransparencySet(TO_TEXT)) return;
 	for (TextEffect &te : _text_effects) {
 		if (te.string_id == INVALID_STRING_ID) continue;
-		if (te.mode == TE_RISING || (_settings_client.gui.loading_indicators && !IsTransparencySet(TO_LOADING))) {
+		if (te.mode == TE_RISING || _settings_client.gui.loading_indicators) {
 			CopyInDParam(te.params);
 			ViewportAddString(dpi, ZOOM_LVL_OUT_8X, &te, te.string_id, te.string_id - 1, STR_NULL);
 		}
--- a/src/transparency.h
+++ b/src/transparency.h
@@ -28,7 +28,7 @@ enum TransparencyOption {
 	TO_BRIDGES,    ///< bridges
 	TO_STRUCTURES, ///< other objects such as transmitters and lighthouses
 	TO_CATENARY,   ///< catenary
-	TO_LOADING,    ///< loading indicators
+	TO_TEXT,       ///< loading and cost/income text
 	TO_END,
 	TO_INVALID,    ///< Invalid transparency option
 };
--- a/src/transparency_gui.cpp
+++ b/src/transparency_gui.cpp
@@ -50,7 +50,7 @@ public:
 			case WID_TT_BRIDGES:
 			case WID_TT_STRUCTURES:
 			case WID_TT_CATENARY:
-			case WID_TT_LOADING: {
+			case WID_TT_TEXT: {
 				uint i = widget - WID_TT_BEGIN;
 				if (HasBit(_transparency_lock, i)) DrawSprite(SPR_LOCK, PAL_NONE, r.left + WidgetDimensions::scaled.fullbevel.left, r.top + WidgetDimensions::scaled.fullbevel.top);
 				break;
@@ -58,7 +58,7 @@ public:
 			case WID_TT_BUTTONS: {
 				const Rect fr = r.Shrink(WidgetDimensions::scaled.framerect);
 				for (uint i = WID_TT_BEGIN; i < WID_TT_END; i++) {
-					if (i == WID_TT_LOADING) continue; // Do not draw button for invisible loading indicators.
+					if (i == WID_TT_TEXT) continue; // Loading and cost/income text has no invisibility button.
 
 					const Rect wr = this->GetWidget<NWidgetBase>(i)->GetCurrentRect().Shrink(WidgetDimensions::scaled.fullbevel);
 					DrawFrameRect(wr.left, fr.top, wr.right, fr.bottom, COLOUR_PALE_GREEN,
@@ -90,7 +90,7 @@ public:
 					break;
 				}
 			}
-			if (i == WID_TT_LOADING || i == WID_TT_END) return;
+			if (i == WID_TT_TEXT|| i == WID_TT_END) return;
 
 			ToggleInvisibility((TransparencyOption)(i - WID_TT_BEGIN));
 			if (_settings_client.sound.click_beep) SndPlayFx(SND_15_BEEP);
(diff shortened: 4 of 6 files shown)


zephyris:
True... Easy to implement as an alternative.

src/texteff.cpp
```diff
@@ -125,8 +125,8 @@ void DrawTextEffects(DrawPixelInfo *dpi)
 
 	for (TextEffect &te : _text_effects) {
 		if (te.string_id == INVALID_STRING_ID) continue;
-		if (te.mode == TE_RISING || (_settings_client.gui.loading_indicators && !IsTransparencySet(TO_LOADING))) {
-			ViewportAddString(dpi, ZOOM_LVL_OUT_8X, &te, te.string_id, te.string_id - 1, STR_NULL, te.params_1, te.params_2);
-		}
+		if (te.mode == TE_RISING && IsTransparencySet(TO_COSTINCOME)) continue;
+		if (te.mode != TE_RISING && _settings_client.gui.loading_indicators && IsTransparencySet(TO_LOADING)) continue;
```
rubidium42: I think the logic got changed.

Before the `loading_indicators` had to be on and it had to be not-transparent to get/keep it shown.
Now you check `loading_indicators` to continue but you inverted the transparent bit.

It might be difficult to see the effect in-game, as with `loading_indicators` disabled no new ones will be made. However, the current ones won't be updated either. So, if you enable `loading_indicators`, put some vehicles in wait-for-load, wait until they are waiting and the indicator is showing. Then pause and disable the loading indicators. When you unpause the indicator remains and won't be updated. I would expect the indicator to disappear immediately instead, like they do in master.

zephyris:
```suggestion
		if (te.mode == TE_RISING && IsTransparencySet(TO_COSTINCOME)) continue;
		if (te.mode != TE_RISING && (!_settings_client.gui.loading_indicators || IsTransparencySet(TO_LOADING))) continue;
```

You're absolutely right, silly mistake. This is the correct logic.

(I guess I need to commit this to the branch, then squash to a single tidy commit? Still very much learning git...)

src/transparency_gui.cpp
```diff
@@ -55,10 +55,15 @@ class TransparenciesWindow : public Window
 				if (HasBit(_transparency_lock, i)) DrawSprite(SPR_LOCK, PAL_NONE, r.left + WidgetDimensions::scaled.fullbevel.left, r.top + WidgetDimensions::scaled.fullbevel.top);
 				break;
 			}
+			case WID_TT_COSTINCOME: {
+				uint i = widget - WID_TT_BEGIN;
+				if (HasBit(_transparency_lock, i)) DrawSprite(SPR_LOCK, PAL_NONE, r.left + WidgetDimensions::scaled.fullbevel.left, r.top + WidgetDimensions::scaled.fullbevel.top);
+				break;
+			}
```
rubidium42: I guess you copied this from the five lines before. The simpler/shorter solution would be to remove the `{` from the line with `case WID_TT_LOADING:` and add `case WID_TT_COSTINCOME: {` on the next line. Then the rest of this block is not needed.
I would suggest the code change, but GitHub seems to not allow changing unaltered lines when making a suggestion.

zephyris: Thank you - this was my naivety and not fully understanding the case syntax (coming from Python which doesn't need breaks!)

rubidium42 (dismissed):

LordAro:
I agree with @llugo tbh, I don't think this deserves a new button/setting

zephyris:
On digesting @llugo and @LordAro 's comments, I think I agree... a separate transparency button is overkill. Merge the loading/unloading indicator and cost/income transparency to both use the existing loading/unloading indicator transparency button? Do you think this demands a new icon?

src/transparency.h
```diff
@@ -28,7 +28,7 @@ enum TransparencyOption {
 	TO_BRIDGES,    ///< bridges
 	TO_STRUCTURES, ///< other objects such as transmitters and lighthouses
 	TO_CATENARY,   ///< catenary
-	TO_LOADING,    ///< loading indicators
+	TO_TEXT,       ///< text - loading indicators, cost/income text
```
2TallTyler:
```suggestion
	TO_TEXT,       ///< loading and cost/income text
```

src/lang/english.txt
```diff
@@ -2650,7 +2650,7 @@ STR_TRANSPARENT_BUILDINGS_TOOLTIP                               :{BLACK}Toggle t
 STR_TRANSPARENT_BRIDGES_TOOLTIP                                 :{BLACK}Toggle transparency for bridges. Ctrl+Click to lock
 STR_TRANSPARENT_STRUCTURES_TOOLTIP                              :{BLACK}Toggle transparency for structures like lighthouses and antennas. Ctrl+Click to lock
 STR_TRANSPARENT_CATENARY_TOOLTIP                                :{BLACK}Toggle transparency for catenary. Ctrl+Click to lock
-STR_TRANSPARENT_LOADING_TOOLTIP                                 :{BLACK}Toggle transparency for loading indicators. Ctrl+Click to lock
+STR_TRANSPARENT_TEXT_TOOLTIP                                    :{BLACK}Toggle transparency for loading and cost/income indicators. Ctrl+Click to lock
```
```suggestion
STR_TRANSPARENT_TEXT_TOOLTIP                                    :{BLACK}Toggle transparency for loading and cost/income text. Ctrl+Click to lock
```

src/transparency_gui.cpp
```diff
@@ -50,15 +50,15 @@ class TransparenciesWindow : public Window
...
+			case WID_TT_TEXT: {
 				uint i = widget - WID_TT_BEGIN;
 				if (HasBit(_transparency_lock, i)) DrawSprite(SPR_LOCK, PAL_NONE, r.left + WidgetDimensions::scaled.fullbevel.left, r.top + WidgetDimensions::scaled.fullbevel.top);
 				break;
 			}
 			case WID_TT_BUTTONS: {
 				const Rect fr = r.Shrink(WidgetDimensions::scaled.framerect);
 				for (uint i = WID_TT_BEGIN; i < WID_TT_END; i++) {
-					if (i == WID_TT_LOADING) continue; // Do not draw button for invisible loading indicators.
+					if (i == WID_TT_TEXT) continue; // Do not draw button for invisible text (loading and cost/income) indicators.
```
2TallTyler:
```suggestion
					if (i == WID_TT_TEXT) continue; // Loading and cost/income text has no invisibility button.
```

src/widgets/transparency_widget.h
```diff
@@ -22,7 +22,7 @@ enum TransparencyToolbarWidgets {
 	WID_TT_BRIDGES,                  ///< Bridges transparency toggle button.
 	WID_TT_STRUCTURES,               ///< Object structure transparency toggle button.
 	WID_TT_CATENARY,                 ///< Catenary transparency toggle button.
-	WID_TT_LOADING,                  ///< Loading indicators transparency toggle button.
+	WID_TT_TEXT,                     ///< Text - loading indicators and cost/income - transparency toggle button.
```
2TallTyler:
```suggestion
	WID_TT_TEXT,                     ///< Loading and cost/income text transparency toggle button.
```

zephyris:
Simple revision, simplifies the change so that cost/income indicator visibility is toggled with the existing loading indicator visibility button. Changed the tooltip to indicate that it toggles both (and changed the name of the tooltip string and widget to reflect this new functionality). I think the icon of a vehicle is still _OK_, as cost/income and loading are all strongly vehicle-associated.

src/texteff.cpp
```diff
@@ -121,7 +121,7 @@ void DrawTextEffects(DrawPixelInfo *dpi)
 
 	for (TextEffect &te : _text_effects) {
 		if (te.string_id == INVALID_STRING_ID) continue;
-		if (te.mode == TE_RISING || (_settings_client.gui.loading_indicators && !IsTransparencySet(TO_LOADING))) {
+		if ((te.mode == TE_RISING && !IsTransparencySet(TO_TEXT)) || (_settings_client.gui.loading_indicators && !IsTransparencySet(TO_TEXT))) {
```
2TallTyler: Minor detail, but this conditional is somewhat confusing because of the short-circuit evaluation and double check of transparency.
```suggestion
		if (IsTransparencySet(TO_TEXT)) continue;
		if (te.mode == TE_RISING || _settings_client.gui.loading_indicator) {
```

2TallTyler (commented):
Looks pretty good to me, just a few minor details for code readability. 😃

src/texteff.cpp
zephyris: Much nicer :)

src/texteff.cpp
```diff
@@ -121,7 +121,8 @@ void DrawTextEffects(DrawPixelInfo *dpi)
```
2TallTyler:
```suggestion
if (IsTransparencySet(TO_TEXT)) return;

```

2TallTyler (commented):
Oof, I'm sorry, my suggestion for handling `IsTransparencySet()` wasn't quite right. It should return early before the for loop on line 121, rather than pointlessly continuing through the entire loop.

Change that, and I'll approve (as I was about to when I realized my error). 😳

src/texteff.cpp
zephyris: No worries! It's a very good point.

2TallTyler (approved):

PR #11002, opened 2023-06-13
Fix: when a string consumes more parameters than allowed, nullptr is attempted to be formatted

rubidium42:
## Motivation / Problem

When the enforcement of `{STRING}` to not take any parameters is enabled, `0` aka `nullptr` is returned for any `{RAW_STRING}` within that string.
Most common case is `STR_JUST_STRING` that includes e.g. `STR_AI_DEBUG_NAME_AND_VERSION`.

## Description

Go into the '(invalid RAW_STRING parameter)' branch of formatting when that happens, instead of SIG_SEGV.

## Limitations

Doesn't solve the issue that `STR_JUST_STRING` gets used in this way. However, in any case not crashing out is the better solution. Even when the underlying issues get solved. That also allows us to actually enable checking that `{STRING}` does not consume any parameters.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -1074,7 +1074,9 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 
 			case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
 				const char *raw_string = (const char *)(size_t)args->GetInt64(SCC_RAW_STRING_POINTER);
-				if (game_script && std::find(_game_script_raw_strings.begin(), _game_script_raw_strings.end(), raw_string) == _game_script_raw_strings.end()) {
+				/* raw_string can be(come) nullptr when the parameter is out of range and 0 is returned instead. */
+				if (raw_string == nullptr ||
+						(game_script && std::find(_game_script_raw_strings.begin(), _game_script_raw_strings.end(), raw_string) == _game_script_raw_strings.end())) {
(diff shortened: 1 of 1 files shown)


glx22 (approved):

src/strings.cpp
```diff
@@ -1074,7 +1074,9 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 
 			case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
 				const char *raw_string = (const char *)(size_t)args->GetInt64(SCC_RAW_STRING_POINTER);
-				if (game_script && std::find(_game_script_raw_strings.begin(), _game_script_raw_strings.end(), raw_string) == _game_script_raw_strings.end()) {
+				/* raw_string can be(come) nullptr when the parameter is out of range and 0 is returned instead. */
+				if (raw_string == nullptr ||
+						(game_script && std::find(_game_script_raw_strings.begin(), _game_script_raw_strings.end(), raw_string) == _game_script_raw_strings.end())) {
```
PeterN, line 1079: Isn't passing `nullptr` to std::find here just going to look for `nullptr` in `_game_script_raw_strings` and return `.end()`?

i.e. the special case stops an unnecessary search but doesn't actually fix anything.

PeterN: Ah, if but not if !game_script...

PR #11007, opened 2023-06-14
Codechange: make creating temporary StringParameters easier

rubidium42:
## Motivation / Problem

The repetitiveness of:
```
int64 args_array[] = { this->index };
StringParameters tmp_params(args_array);
```

## Description

Introduce a sub class of StringParameters that has its own buffer.
Introduce a templated method that can accept any number of arguments that creates and fills the appropriate "StringParameters" instance.
Use that templated function instead, so the above code becomes `auto tmp_params = MakeParameters(this->index);`.

As added bonus, stuff like:
```
int64 args_array[] = {(int64)(size_t)e->name.c_str()};
StringParameters tmp_params(args_array);
```
becomes `auto tmp_params = MakeParameters(e->name);`, by implementing `SetParam` for `const std::string&` and `const char *`.

## Limitations

None I can think of, except that there are a few other places where this new class could be used. However, those are for a later PR.

String pointers/references are still cast to `size_t` and then `uint64_t`. Solving that is, again, for a later PR.

I would have liked to use `std::array` and rely on copy-elision. That does work with GCC, but neither Clang nor MSVC does copy-elision for `MakeParameters` making that route troublesome. So, I've gone with `std::vector` so it can just move a pointer around.

## Checklist for review

### DIFF
--- a/src/industry_cmd.cpp
+++ b/src/industry_cmd.cpp
@@ -2431,8 +2431,7 @@ void Industry::RecomputeProductionMultipliers()
 
 void Industry::FillCachedName() const
 {
-	int64 args_array[] = { this->index };
-	StringParameters tmp_params(args_array);
+	auto tmp_params = MakeParameters(this->index);
 	this->cached_name = GetStringWithArgs(STR_INDUSTRY_NAME, tmp_params);
 }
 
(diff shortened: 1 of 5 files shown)


LordAro (commented):
Is there any particular reason all the AllocatedStringParameters variables can't be marked as const?

rubidium42:
> Is there any particular reason all the AllocatedStringParameters variables can't be marked as const?

Most places require `StringParameters &`, because either they run `GetInt64` which updates the `offset` variable within `StringParameters` or run `SetParam` which also updates the object.

LordAro (approved):

PR #11010, opened 2023-06-14
Change: [Actions] Use notarytool for notarization instead of gon

orudge:
## Motivation / Problem

We are currently using `gon` to handle notarization of OpenTTD. Apple has sent us an e-mail instructing us that `altool` (which `gon` currently uses) will no longer be supported from November 2023.

Fortunately, Apple released a new notarization tool a while ago which is quite straightforward to use, so use that instead.

## Description

Removes dependency on `gon`, uses the built in `notarytool` provided by Xcode. I've tested this on my own fork with our real Apple details and notarization appears to have been successful.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -162,19 +162,19 @@ jobs:
         rm -f bundles/*.sha256
         echo "::endgroup::"
 
-    - name: Install gon
-      env:
-        HOMEBREW_NO_AUTO_UPDATE: 1
-        HOMEBREW_NO_INSTALL_CLEANUP: 1
-      run: |
-        brew tap mitchellh/gon
-        brew install mitchellh/gon/gon
-
     - name: Notarize
       env:
         AC_USERNAME: ${{ secrets.APPLE_DEVELOPER_APP_USERNAME }}
         AC_PASSWORD: ${{ secrets.APPLE_DEVELOPER_APP_PASSWORD }}
+        AC_TEAM_ID: ${{ secrets.APPLE_DEVELOPER_TEAM_ID }}
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-macos.yml
```diff
@@ -162,19 +162,13 @@ jobs:
...
-        brew tap mitchellh/gon
-        brew install mitchellh/gon/gon
-
     - name: Notarize
       env:
         AC_USERNAME: ${{ secrets.APPLE_DEVELOPER_APP_USERNAME }}
         AC_PASSWORD: ${{ secrets.APPLE_DEVELOPER_APP_PASSWORD }}
+        AC_TEAM_ID: ${{ secrets.APPLE_DEVELOPER_TEAM_ID }}
       run: |
+        xcrun notarytool store-credentials --apple-id "$AC_USERNAME" --password "$AC_PASSWORD" --team-id "$AC_TEAM_ID" openttd
```
TrueBrain: please use `${NNN}` instead of `$NNN`. Avoids common mistakes when concatenating strings (not a problem in this case :)).

TrueBrain: This will now fail for people who don't have those secrets, read: everyone but us. Shouldn't we silent ignore errors, like we did before?

Sadly, an `if:` in the workflow itself doesn't work, as that cannot check if a secret exists .. so in bash an `if` statement seems the way to go?

orudge: The workflow fails anyway if you don’t have a survey secret key set up, or something, I discovered. 🙂

I can add a bash if check though.

TrueBrain (approved):

PR #11012, opened 2023-06-15
Change: Allow aqueduct to use foundation

### LINKED ISSUE #10995: [Bug]: Aqueduct construction on slopes inconsistent with other bridges and tunnels
### Version of OpenTTD

Latest master, also in 13.1

### Expected result

After comparing aqueducts to tunnels, which show similar highlighting behavior, I noticed that a corner on the opposite tunnel end will automatically get terraformed if it's the wrong slope. Aqueducts don't do this, even though the highlighted area suggest you can actually build the aqueduct.

### Actual result

image
If I select this side of the water the aqueduct is correctly highlighted, and it builds fine when you click.

image
However, hovering over the tile next to it highlights the aqueduct. But it can't be built due to the wrong slope at the other end. Clicking results in the "land sloped in wrong direction" message. Here I'd expect some automatic terraforming similar to bridges and tunnels.

### Steps to reproduce

1. Generate a map with enough water
2. Do what I did in the screenshots ;)

marcossouza1232:
## Motivation / Problem

fix #10995 

## Description

Removes aqueduct restriction, allowing automatic slope correction while constructing aqueducts.

## Limitations

I don't know the side effects, because I don't know why they disabled this functionality for aqueducts.

### DIFF
--- a/src/tunnelbridge_cmd.cpp
+++ b/src/tunnelbridge_cmd.cpp
@@ -408,7 +408,7 @@ CommandCost CmdBuildBridge(DoCommandFlag flags, TileIndex tile_end, TileIndex ti
 	} else {
 		/* Build a new bridge. */
 
-		bool allow_on_slopes = (_settings_game.construction.build_on_slopes && transport_type != TRANSPORT_WATER);
+		bool allow_on_slopes = (_settings_game.construction.build_on_slopes );
 
 		/* Try and clear the start landscape */
 		CommandCost ret = Command<CMD_LANDSCAPE_CLEAR>::Do(flags, tile_start);


src/tunnelbridge_cmd.cpp
```diff
@@ -408,7 +408,7 @@ CommandCost CmdBuildBridge(DoCommandFlag flags, TileIndex tile_end, TileIndex ti
 	} else {
 		/* Build a new bridge. */
 
-		bool allow_on_slopes = (_settings_game.construction.build_on_slopes && transport_type != TRANSPORT_WATER);
+		bool allow_on_slopes = (_settings_game.construction.build_on_slopes );
```
2TallTyler, line 411: Extra space after conditional

2TallTyler (commented):
The commit checker is failing because you didn't follow the message format specified in the coding style. 

It should be `Change: Allow aqueduct to use foundation` or similar and then note in the PR description that it closes #10995. It's not a fix because the original choice seems to be intentional.

Also, this does not terraform the end tile, it builds the aqueduct on a foundation. That is a change, since we don't allow flat canals to be built on foundations. A proper "fix" might be to terraform the end slope to allow building the aqueduct without a foundation.

src/tunnelbridge_cmd.cpp
romw314, line 411:
```suggestion
		bool allow_on_slopes = (_settings_game.construction.build_on_slopes);
```

2TallTyler:
This doesn't seem like the right approach, but if anyone feels differently feel free to comment, re-open, etc. I am closing this PR because it appears to be abandoned.

PR #11015, opened 2023-06-15
Codechange: make SQString::Create that supports std::string and use that

rubidium42:
## Motivation / Problem

Number of places doing `.c_str()` to then (implicitly) run `strlen` to determine the length within the Squirrel API.

## Description

Add `std::string` supporting function to create `SQString` that passes the string size to the function that actually creates the object.

## Limitations

None.

## Checklist for review


rubidium42:
> changing from c_str to data without any changes to length/size makes me uncomfortable

1: since C++11 `.data()` and `.c_str()` are defined/specified to be the same. The specifications literally says: "The returned array is null-terminated, that is, data() and c_str() perform the same function.".
2: we are already passing the string's size there, unless the size is overridden by the caller. So conceptually `.data()` and `.size()` seem better than `.c_str()` and `.size()`.

LordAro (approved):

PR #11018, opened 2023-06-16
Remove: OS/2 port

orudge:
## Motivation / Problem

The OS/2 port doesn't currently build, CI is... challenging... to set up, and sadly I don't really have the time or resources to improve these at present.

## Description

Back in the day, OpenTTD was ticking along quietly, just getting on with things. Then one day a certain nameless developer inflicted OS/2 support upon OpenTTD, thanks to the wonderful compiler Open Watcom. As a result of this superpower, OpenTTD went from strength to strength. Over time it evolved, with the excellent work of Paul Smedley fixing things up to build with GCC. OpenTTD could even still be built on OS/2 when we moved to CMake. But alas, all good things must come to an end, and for the official OS/2 port, I feel that time has come.

Technically there is little stopping the current code from working on OS/2, but my own build environment is problematic. Ideally I need to buy a copy of [ArcaOS](https://www.arcanoae.com/arcaos/) which has a relatively modern software repository (based on RPM no less), but for some reason I don't feel *quite* that dedicated to OS/2. I daresay I could even then fashion some sort of GitHub Actions build system support out of it. Maybe one day I will. But that day is not today.

Goodbye, old friend. 😢

## Limitations

Quite a few, hence the removal. Oh, you didn't mean OS/2?

## Checklist for review


TrueBrain (approved):
So long and tnx for all the fish!

psmedley:
Shame... just found this after being nudged to build a newer openttd for OS/2

PR #11022, opened 2023-06-16
Fix #11016: Use after free in network invalid packet error path

### LINKED ISSUE #11016: [Bug]: Use after free in network invalid packet error path
### Version of OpenTTD

master

### Expected result

No bugs, etc

### Actual result

CloseConnection in NetworkGameSocketHandler::HandlePacket calls `delete this`.
`this` is still used afterwards in NetworkGameSocketHandler::HandlePacket.
In the case of clients, `this` then gets used to try to send an error to the server in `ClientNetworkGameSocketHandler::Receive` which fairly reliably causes a crash.

### Steps to reproduce

See above and/or inject a packet fault.

JGRennison:
## Motivation / Problem

ClientNetworkGameSocketHandler/ServerNetworkGameSocketHandler::CloseConnection called `delete this`.
However they could still be used afterwards, causing crashes, etc.

## Description

Instead of calling `delete this`, register the handler for deferred deletion, and do the deletions at a later safe point.

The logging in the path where this type of crash occurred was also incorrect/misleading (HandlePacket), so that is also fixed.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/network/core/tcp_admin.cpp
+++ b/src/network/core/tcp_admin.cpp
@@ -47,7 +47,13 @@ NetworkRecvStatus NetworkAdminSocketHandler::HandlePacket(Packet *p)
 {
 	PacketAdminType type = (PacketAdminType)p->Recv_uint8();
 
-	switch (this->HasClientQuit() ? INVALID_ADMIN_PACKET : type) {
+	if (this->HasClientQuit()) {
+		Debug(net, 0, "[tcp/admin] Received invalid packet from '{}' ({})", this->admin_name, this->admin_version);
+		this->CloseConnection();
+		return NETWORK_RECV_STATUS_MALFORMED_PACKET;
+	}
+
+	switch (type) {
 		case ADMIN_PACKET_ADMIN_JOIN:             return this->Receive_ADMIN_JOIN(p);
 		case ADMIN_PACKET_ADMIN_QUIT:             return this->Receive_ADMIN_QUIT(p);
 		case ADMIN_PACKET_ADMIN_UPDATE_FREQUENCY: return this->Receive_ADMIN_UPDATE_FREQUENCY(p);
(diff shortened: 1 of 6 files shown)


glx22 (dismissed):

src/network/core/tcp_game.cpp
```diff
@@ -197,3 +200,17 @@ NetworkRecvStatus NetworkGameSocketHandler::Receive_SERVER_MOVE(Packet *p) { ret
...
+	_deferred_deletions.push_back(this);
+	this->is_pending_deletion = true;
+}
+
+/* static */ void NetworkGameSocketHandler::ProcessDeferredDeletions()
+{
+	for (NetworkGameSocketHandler *cs : _deferred_deletions) {
+		delete cs;
+	}
+	_deferred_deletions.clear();
```
rubidium42: Can't `_deferred_deletions` be a `std::vector<std::unique_ptr<NetworkGameSocketHandler>>` so this becomes just a `_deferred_deletions.clear()` without manually deleting objects?

JGRennison: Wouldn't that be a bit confusing, as unique_ptr implies a different ownership model than what is actually used for the various socket handlers?

rubidium42: Not sure if it's confusing. To me it looks like `_deferred_deletions` owns the handlers once they are passed into it, but if it confuses others then keeping it pointers that are manually `delete`d is fine by me as well.

JGRennison: OK, I'll update the PR with this a bit later.

rubidium42 (approved):

src/network/core/tcp_admin.cpp
```diff
@@ -47,7 +47,13 @@ NetworkRecvStatus NetworkAdminSocketHandler::HandlePacket(Packet *p)
 {
 	PacketAdminType type = (PacketAdminType)p->Recv_uint8();
```
PeterN, line 48: If the client has quit, should we even receive the packet type? Maybe place the HasClientQuit() check above here?

src/network/core/tcp_game.cpp
```diff
@@ -63,9 +65,15 @@ NetworkRecvStatus NetworkGameSocketHandler::HandlePacket(Packet *p)
 {
 	PacketGameType type = (PacketGameType)p->Recv_uint8();
 
+	if (this->HasClientQuit()) {
```
PeterN, line 68: Same as above

src/network/core/tcp_admin.cpp
JGRennison: Recv_uint8 itself can set HasClientQuit to true.
The result is 0 (i.e. not valid) whether HasClientQuit was set before or during the call.

Putting the check below is to avoid logging a "received invalid packet type 0" message, when no such packet was ever received.

PeterN: Ah ha

PR #11024, opened 2023-06-16
Fix #10993: Crash log when font caches not initialised

### LINKED ISSUE #10993: [Bug]: Crash log no longer works for crashes which occur before FontCache::InitializeFontCaches
### Version of OpenTTD

Since #10836 / f454ec8d63be9442e8cb03f0db072e66f2332fca

### Expected result

Crash log works as far as is reasonably practical

### Actual result

InitializeFontCaches is now called fairly late into the initialisation process due to the changes in #10836, but the crash log assumes that they are always present, and so attempts to dereference them, resulting in no useful output at all.

Possible fix: https://github.com/JGRennison/OpenTTD-patches/commit/aa8b60da150e72f0b2956b6913f0982fd9d03da6 ?

### Steps to reproduce

Synthesise a crash somewhere early (e.g. in InitializeLanguagePacks).

JGRennison:
## Motivation / Problem

#10993, the crash logger no longer works before font caches are initialised due to #10836.

## Description

Fix above.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/crashlog.cpp
+++ b/src/crashlog.cpp
@@ -183,10 +183,10 @@ void CrashLog::LogConfiguration(std::back_insert_iterator<std::string> &output_i
 			" Medium: {}\n"
 			" Large:  {}\n"
 			" Mono:   {}\n\n",
-			FontCache::Get(FS_SMALL)->GetFontName(),
-			FontCache::Get(FS_NORMAL)->GetFontName(),
-			FontCache::Get(FS_LARGE)->GetFontName(),
-			FontCache::Get(FS_MONO)->GetFontName()
+			FontCache::GetName(FS_SMALL),
+			FontCache::GetName(FS_NORMAL),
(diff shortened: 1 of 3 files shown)


src/crashlog.cpp
```diff
@@ -177,16 +177,25 @@ void CrashLog::LogConfiguration(std::back_insert_iterator<std::string> &output_i
 			VideoDriver::GetInstance() == nullptr ? "none" : VideoDriver::GetInstance()->GetInfoString()
 	);
 
+	auto log_font = [&](FontSize fs) -> std::string {
```
PeterN: Can't this be a regular function?

glx22: Like `static std::string FontCache::GetName(FontSize fs)` maybe

glx22 (approved):

PeterN:
Thanks! 👍

PR #11041, opened 2023-06-18
Codechange: move to std::string over stredup + free

rubidium42:
## Motivation / Problem

Use of `stredup` + `free`.

## Description

`FS2OTTD` already returns a `std::string`, so just use that and pass its `.data()` to the command line parsing instead of `stredup`-ing the `.c_str()` of that string.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/os/windows/win32_main.cpp
+++ b/src/os/windows/win32_main.cpp
@@ -58,10 +58,8 @@ int APIENTRY WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLi
 
 	CrashLog::InitialiseCrashLog();
 
-	/* Convert the command line to UTF-8. We need a dedicated buffer
-	 * for this because argv[] points into this buffer and this needs to
-	 * be available between subsequent calls to FS2OTTD(). */
-	char *cmdline = stredup(FS2OTTD(GetCommandLine()).c_str());
+	/* Convert the command line to UTF-8. */
+	std::string cmdline = FS2OTTD(GetCommandLine());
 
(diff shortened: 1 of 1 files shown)


src/os/windows/win32_main.cpp
```diff
@@ -76,7 +74,7 @@ int APIENTRY WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLi
 	SetRandomSeed(GetTickCount());
 
 	char *argv[64]; // max 64 command line arguments
-	int argc = ParseCommandLine(cmdline, argv, lengthof(argv));
+	int argc = ParseCommandLine(cmdline.data(), argv, lengthof(argv));
```
glx22, line 77: Maybe `ParseCommandLine()` could be modified to work with `std::string` directly.

rubidium42: It definitely could be, but... then you rather want to do it properly. Like returning a vector of strings. That would mean changing `openttd_main` to accept that vector instead, which means `getopt` code needs to be made string-aware and that could potentially cause changes in strgen/settingsgen.
That's something you rather want to tackle from the other side, so first make `getopt` string aware, then a vector of strings to `openttd_main` and then rewriting this function.

That's quite a significant thing to do, so I rather first finish the string parameters stuff which already is part of the eradicate stredup "project". The string parameters are waiting on reviews, before I can continue further, so I'm tackling the remaining uses of `stredup` besides the string parameters in quite minimalistic style.

glx22 (approved):

PR #11052, opened 2023-06-21
Codechange: use vector with struct for string parameter backups

rubidium42:
## Motivation / Problem

Use of two or three parameters to retain the backup, and as such making it harder to generalize.
Furthermore the use of `stredup` and the associated manual memory management.

## Description

Add StringParameterBackup struct that contains either a `uint64_t` or `std::string`, a `std::vector` for it and the needed functions for it to create a backup or restore it.
Convert the different users of the parameter backups to the new type/functions.
Remove the old functions.

## Limitations

None that I can think of, except a bit more memory usage. But in the end, when next PRs get merged safer and simpler code differentiating between strings and integers.

## Checklist for review

### DIFF
--- a/src/error_gui.cpp
+++ b/src/error_gui.cpp
@@ -74,24 +74,10 @@ static WindowDesc _errmsg_face_desc(
  * @param data The data to copy.
  */
 ErrorMessageData::ErrorMessageData(const ErrorMessageData &data) :
-	is_critical(data.is_critical), textref_stack_grffile(data.textref_stack_grffile), textref_stack_size(data.textref_stack_size),
+	is_critical(data.is_critical), params(data.params), textref_stack_grffile(data.textref_stack_grffile), textref_stack_size(data.textref_stack_size),
 	summary_msg(data.summary_msg), detailed_msg(data.detailed_msg), extra_msg(data.extra_msg), position(data.position), face(data.face)
 {
 	memcpy(this->textref_stack, data.textref_stack, sizeof(this->textref_stack));
-	memcpy(this->decode_params, data.decode_params, sizeof(this->decode_params));
-	memcpy(this->strings,       data.strings,       sizeof(this->strings));
-	for (size_t i = 0; i < lengthof(this->strings); i++) {
-		if (this->strings[i] != nullptr) {
-			this->strings[i] = stredup(this->strings[i]);
-			this->decode_params[i] = (size_t)this->strings[i];
-		}
-	}
-}
-
-/** Free all the strings. */
-ErrorMessageData::~ErrorMessageData()
-{
-	for (size_t i = 0; i < lengthof(this->strings); i++) free(this->strings[i]);
 }
 
 /**
@@ -118,9 +104,6 @@ ErrorMessageData::ErrorMessageData(StringID summary_msg, StringID detailed_msg, 
 	this->position.x = x;
 	this->position.y = y;
 
-	memset(this->decode_params, 0, sizeof(this->decode_params));
-	memset(this->strings, 0, sizeof(this->strings));
-
 	if (textref_stack_size > 0) MemCpyT(this->textref_stack, textref_stack, textref_stack_size);
 
 	assert(summary_msg != INVALID_STRING_ID);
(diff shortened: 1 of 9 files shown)


src/strings.cpp
```diff
@@ -152,49 +152,56 @@ void SetDParamMaxDigits(size_t n, uint count, FontSize size)
...
+void CopyInDParam(const span<const StringParameterBackup> backup)
 {
-	for (int i = 0; i < num; i++) SetDParam(i, src[i]);
+	for (size_t i = 0; i < backup.size(); i++) {
+		const auto &value = backup[i];
+		if (std::holds_alternative<std::string>(value)) {
+			_global_string_params.SetParam(i, std::get<std::string>(value));
+		} else {
+			_global_string_params.SetParam(i, std::get<uint64_t>(value));
+		}
```
TrueBrain, line 168: You could use the `std::visit` here, and avoid having to use the `holds_alternative`. I don't know if we already defined any form of `overloaded`, as that is kinda mandatory to keep it readable :) But it would mean it becomes something like:

```
std::visit(overloaded{
            [](auto arg) {
                        _global_string_params.SetParam(i, arg);
            }
}
```

And the compiler will deal with the different types.

No clue if it is an improvement :)

rubidium42: The `overloaded` bit isn't even needed. It does look prettier now!

rubidium42:
> I could be slightly worried that introducing the variant type could be a performance problem, depending on how often it would be used. Some strings are certainly queried a lot.

I wouldn't know; it's only for cases where an actual backup of the string parameters is made, that is restored at some later point (tooltip, news, errors). Would having a struct with `std::optional<uint64_t> data` and `std::optional<std::string> string` be much better? That feels like way more boilerplate code that needs to be written to achieve approximately the same, while using more memory.


> Another thought: Combine the string id with its parameter vector, as well as an (optional) cached formatted version of the string, and the language the string was formatted with (for invalidating the cache). It would use more memory, but could avoid a number of string formatting calls. (It could also result in stale strings in a few situations, e.g. if stations or vehicles are renamed, which might be difficult to catch.)

I'm not sure whether that's going to work nicely when there are multiple format calls with different string ids with the same parameter vector. See for example the news code.

rubidium42:
So essentially the question is: is https://github.com/OpenTTD/OpenTTD/compare/master...rubidium42:string-backup-struct better or worse than this PR?

rubidium42:
Changed this over to the `struct` variant over `std::variant` variant (https://github.com/OpenTTD/OpenTTD/compare/master...rubidium42:string-backup-variant)

src/error_gui.cpp
```diff
@@ -137,14 +120,9 @@ void ErrorMessageData::CopyOutDParams()
...
 
-	/* Reset parameters */
-	for (size_t i = 0; i < lengthof(this->strings); i++) free(this->strings[i]);
-	memset(this->decode_params, 0, sizeof(this->decode_params));
-	memset(this->strings, 0, sizeof(this->strings));
-
 	/* Get parameters using type information */
 	if (this->textref_stack_size > 0) StartTextRefStackUsage(this->textref_stack_grffile, this->textref_stack_size, this->textref_stack);
-	CopyOutDParam(this->decode_params, this->strings, this->detailed_msg == INVALID_STRING_ID ? this->summary_msg : this->detailed_msg, lengthof(this->decode_params));
+	CopyOutDParam(this->params, 20, this->detailed_msg == INVALID_STRING_ID ? this->summary_msg : this->detailed_msg);
```
PeterN, line 125: Magic 20, presumably that's just some limit that was set elsewhere before.

src/misc_gui.cpp
```diff
@@ -1098,14 +1093,14 @@ void ShowQueryString(StringID str, StringID caption, uint maxsize, Window *paren
...
-	uint64 params[10];       ///< local copy of #_global_string_params
+	std::vector<StringParameterBackup> params; ///< local copy of #_global_string_params
 	StringID message;        ///< message shown for query window
 
 	QueryWindow(WindowDesc *desc, StringID caption, StringID message, Window *parent, QueryCallbackProc *callback) : Window(desc)
 	{
 		/* Create a backup of the variadic arguments to strings because it will be
 		 * overridden pretty often. We will copy these back for drawing */
-		CopyOutDParam(this->params, lengthof(this->params));
+		CopyOutDParam(this->params, 10);
```
PeterN, line 1103: Magic 10, as above I suppose.

PeterN (approved):

src/error_gui.cpp
rubidium42: Yes, error.h:34/35 to be precise. In the end only the needed amount of parameters should be copied, but that's for later.

src/misc_gui.cpp
rubidium42: Yes, misc_gui.cpp:1101 to be precise.

PR #11057, opened 2023-06-22
Codechange: improve name of method and reduce casting

rubidium42:
## Motivation / Problem

There is a `GetInt64` and `GetInt32` function in `StringParameters`. From the name it's immediately not obvious that it changes the state within the object. Also many places cast to another type after calling either `GetInt32` or `GetInt64`.

## Description

Split the offset updating and type validation from the actual returning of the data, so only the actual returning of the data needs to be duplicated.

Rename `GetInt64`/`GetInt32` to `GetNextParameter` to more clearly show it has side effects, but also template the expected return type so there is no need to cast after calling `GetNextParameter`. Something which happened relatively often.
While doing this, also improve the returned types to what is actually used.

Remove the old `GetInt64`/`GetInt32`.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -72,26 +72,28 @@ void StringParameters::PrepareForNextRun()
 
 
 /**
- * Read an int64 from the argument array. The offset is increased
- * so the next time GetInt64 is called the next value is read.
+ * Get the next parameter from our parameters.
+ * This updates the offset, so the next time this is called the next parameter
+ * will be read.
+ * @return The pointer to the next parameter.
  */
-int64 StringParameters::GetInt64()
+StringParameter *StringParameters::GetNextParameterPointer()
 {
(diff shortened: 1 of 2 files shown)


LordAro (approved):
I'm still not sure the naming of the function improves things, but I don't have any better alternatives

src/strings.cpp
```diff
@@ -1098,7 +1100,7 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 				break;
 
 			case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
-				const char *raw_string = (const char *)(size_t)args.GetInt64();
+				const char *raw_string = (const char *)(size_t)args.GetNextParameter<size_t>();
```
PeterN, line 1103: Casting size_t to size_t here.

PR #11060, opened 2023-06-22
Fix: Layouter reverse-mapping to characters

nielsmh:
## Motivation / Problem

Mapping back from a glyph to a character index in the source string should work, even with multiple run and when formatting codes are involved.

## Description

Have visual runs keep track of their offsets in the input text.
Count formatting codes the same way in both forward and reverse mapping.

## Limitations

Not thoroughly tested recently, but this fixed something when I was working on #7786 originally. (Reverse-mapping is relevant for hit-testing mouse clicks against individual character glyphs.)

## Checklist for review

### DIFF
--- a/src/gfx_layout.cpp
+++ b/src/gfx_layout.cpp
@@ -207,6 +207,19 @@ Dimension Layouter::GetBounds()
 }
 
 /**
+ * Test whether a character is a non-printable formatting code
+ */
+static bool IsConsumedFormattingCode(WChar ch)
+{
+	if (ch >= SCC_BLUE && ch <= SCC_BLACK) return true;
+	if (ch == SCC_PUSH_COLOUR) return true;
+	if (ch == SCC_POP_COLOUR) return true;
+	if (ch >= SCC_FIRST_FONT && ch <= SCC_LAST_FONT) return true;
+	// All other characters defined in Unicode standard are assumed to be non-consumed.
+	return false;
+}
+
+/**
  * Get the position of a character in the layout.
  * @param ch Character to get the position of. Must be an iterator of the string passed to the constructor.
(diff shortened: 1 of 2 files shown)


src/gfx_layout_fallback.cpp
```diff
@@ -109,7 +109,7 @@ class FallbackParagraphLayout : public ParagraphLayouter {
  * @param char_count The number of characters in this run.
  * @param x          The initial x position for this run.
  */
-FallbackParagraphLayout::FallbackVisualRun::FallbackVisualRun(Font *font, const WChar *chars, int char_count, int x) :
+FallbackParagraphLayout::FallbackVisualRun::FallbackVisualRun(Font *font, const WChar *chars, int char_count, int char_ofs, int x) :
```
LordAro: missing doxyment. Also, ofs short for offset ? That seems overly terse

src/gfx_layout.cpp
```diff
@@ -206,6 +206,20 @@ Dimension Layouter::GetBounds()
...
+/**
+ * Test whether a character is a non-printable formatting code
+ */
+static bool IsConsumedFormattingCode(WChar ch)
+{
+	if (ch >= SCC_BLUE && ch <= SCC_BLACK) return true;
+	if (ch == SCC_PUSH_COLOUR) return true;
+	if (ch == SCC_POP_COLOUR) return true;
+	if (ch >= SCC_FIRST_FONT && ch <= SCC_LAST_FONT) return true;
+	if (!IsPrintable(ch)) return false;
```
glx22: Printable or not seem to not matter here (both cases return false)

rubidium42 (approved):

PR #11061, opened 2023-06-22
Fix #11054: Prevent translation of currency codes.

### LINKED ISSUE #11054: [Bug]: font (A-OTF Shin Go Pr6N) not working
### Version of OpenTTD

1.13.3, windows 10

### Expected result

the font to work

### Actual result

some fonts do not work. trying to use *A-OTF Shin Go Pr6N* doesn't work but using *MS Gothic* it does work. when using *A-OTF Shin Go Pr6N* it seems to be using a fallback but not telling me, putting in random keyboard smashing to the font options results in no difference to the font

my openttd.cfg:
```ini
small_font = A-OTF Shin Go Pr6N
medium_font = A-OTF Shin Go Pr6N
large_font = A-OTF Shin Go Pr6N
mono_font = A-OTF Shin Go Pr6N
small_size = 12
medium_size = 14
large_size = 18
mono_size = 14
small_aa = true
medium_aa = true
large_aa = true
mono_aa = true
```

when using *A-OTF Shin Go Pr6N* and trying to read readme with japanese text (likely just the fallback font not supporting the characters):
0x3072 is ひ
```
Unknown debug level 'freetype=1'
dbg: [sprite] 1244 extra sprites, 1244 from baseset, 0 from fallback
dbg: [sprite] 1244 extra sprites, 1244 from baseset, 0 from fallback
dbg: [sprite] 1244 extra sprites, 1244 from baseset, 0 from fallback
dbg: [fontcache] Font is missing glyphs to display char 0x3072 in mono font size
dbg: [fontcache] Font is missing glyphs to display char 0x3072 in mono font size
dbg: [fontcache] Font is missing glyphs to display char 0x3072 in mono font size
dbg: [fontcache] Font is missing glyphs to display char 0x3072 in mono font size
```

### Steps to reproduce

use A-OTF Shin Go Pr6N or seemingly any other type of shin go font, or probably other fonts too

PeterN:
## Motivation / Problem

As per #11054, translation of currency codes by translators into currency symbols can cause issues using some language files as the currency symbol may not be available in the fonts available for that language's script.

Without this change, after selecting Japanese translation:
image

## Description

Most languages stick with the 3-letter latin currency codes in the name string, however some translations are... clever... and use the currency symbol instead. Whilst this may look nice, it can cause issues with fonts as some scripts have a specific limited set of fonts which do not include these symbols.

Instead, hard code the currency code list and add it when drawing the currency name.

With this change, after selecting Japanese translation:
image

## Limitations

1) This may not fully solve the problem, but did offer an improvement in my testing.
2) Also, bulk language update.
3) Hardcoding (in the CurrencySpec table, at least!) is a bit meh.

## Checklist for review


src/settings_gui.cpp
```diff
@@ -216,13 +216,19 @@ struct GameOptionsWindow : Window {
...
-				for (uint i = 0; i < CURRENCY_END; items++, i++) {
+				for (const CurrencySpec &cur : _currency_specs) {
+					int i = &cur - _currency_specs;
 					if (i == CURRENCY_CUSTOM) continue;
-					list.emplace_back(new DropDownListStringItem(*items, i, HasBit(disabled, i)));
+					if (cur.code.empty()) {
+						list.emplace_back(new DropDownListStringItem(cur.name, i, HasBit(disabled, i)));
+					} else {
+						SetDParam(0, cur.name);
+						SetDParamStr(1, cur.code.c_str());
```
rubidium42: The `.c_str()` is not needed with `SetDParamStr`. Like in line 319.

src/settings_gui.cpp
```diff
@@ -763,7 +779,7 @@ static const NWidgetPart _nested_game_options_widgets[] = {
 				EndContainer(),
 
 				NWidget(WWT_FRAME, COLOUR_GREY), SetDataTip(STR_GAME_OPTIONS_CURRENCY_UNITS_FRAME, STR_NULL),
-					NWidget(WWT_DROPDOWN, COLOUR_GREY, WID_GO_CURRENCY_DROPDOWN), SetMinimalSize(100, 12), SetDataTip(STR_JUST_STRING, STR_GAME_OPTIONS_CURRENCY_UNITS_DROPDOWN_TOOLTIP), SetFill(1, 0),
+					NWidget(WWT_DROPDOWN, COLOUR_GREY, WID_GO_CURRENCY_DROPDOWN), SetMinimalSize(100, 12), SetDataTip(STR_JUST_STRING3, STR_GAME_OPTIONS_CURRENCY_UNITS_DROPDOWN_TOOLTIP), SetFill(1, 0),
```
rubidium42: Isn't `STR_JUST_STRING2` enough? You have a string with 2 (additional) parameters.
`STR_JUST_STRING(0)` needs `SetDParam(0, STR_something)`. `STR_JUST_STRING1` then has 1 extra parameter, so sets parameter 0 to the string ID and parameter 1 is the first parameter to that string.

src/settings_gui.cpp
PeterN: Good point, it was needed with DropDownListParamStringItem :-)

src/settings_gui.cpp
PeterN: Hmm, not sure, parameters 0, 1, and 2 are used for WID_GO_CURRENCY_DROPDOWN, which can be either just the currency name (for custom or NewGRF-overridden currencies), or for the STR_GAME_OPTIONS_CURRENCY_CODE string code which takes additional currency name and currency code as parameters.

rubidium42 (approved):

PR #11063, opened 2023-06-23
Codechange: Make DropDownListStringItem preformat and remove other implementations.

PeterN:
## Motivation / Problem

Having to choose between DropDownListStringItem, DropDownListCharStringItem, and DropDownListParamStringItem depending on whether to draw a StringID, a raw string, or a StringID with extra parameters was needlessly complex.

## Description

Instead, allow passing a StringID or raw string to DropDownListStringItem. This will preformat the StringID into a raw string, and can therefore accept parameters via the normal SetDParam mechanism.

This also means that strings no longer need to be formatted on every draw.

## Limitations

Trades a little memory for CPU I guess, but dropdownlists are fairly short-lived.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -588,14 +588,9 @@ static const LiveryClass _livery_class[LS_END] = {
 	LC_ROAD, LC_ROAD,
 };
 
-class DropDownListColourItem : public DropDownListItem {
+class DropDownListColourItem : public DropDownListStringItem {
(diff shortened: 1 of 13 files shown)


rubidium42 (approved):
For the DropDownListParamStringItems you're likely even saving memory, as there was a buffer of 80 bytes for parameters of which one or two were used.

PR #11068, opened 2023-06-24
Feature: Set default company secondary colour for new games

M3Henry:
## Motivation / Problem

The primary company colour for new games is a config option. However the secondary colour always assumes the same as the primary colour. It would be nice to be able to set both independently.

## Description

Add secondary company colour to config and use that in new games if set. Allowed options are:

- "Same as Primary CC" _(default)_
- "_Colour_"...
 

## Limitations

Options could include:
- "Random (not same as Primary CC)"
- "Completely Random"
- "Complementary of Primary CC"
- "Shade / Tint of Primary CC"

### Other things that might be interesting?

* Generate a full set of random 1CC/2CC liveries: by group and/or type
* Also use setting for companies started during multiplayer
* This only sets the Global secondary company colour, going more granular might be good, or might not be.

## Checklist for review

- This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)


src/lang/english.txt
```diff
@@ -1416,6 +1436,9 @@ STR_CONFIG_SETTING_INFRASTRUCTURE_MAINTENANCE_HELPTEXT          :When enabled, i
 STR_CONFIG_SETTING_COMPANY_STARTING_COLOUR                      :Starting company colour: {STRING2}
 STR_CONFIG_SETTING_COMPANY_STARTING_COLOUR_HELPTEXT             :Choose starting colour for the company
 
+STR_CONFIG_SETTING_COMPANY_STARTING_COLOUR_SECONDARY            :Starting company secondary colour: {STRING2}
+STR_CONFIG_SETTING_COMPANY_STARTING_COLOUR_SECONDARY_HELPTEXT   :Choose starting secondary colour for the company
```
2TallTyler: Might save some confusion:

```suggestion
STR_CONFIG_SETTING_COMPANY_STARTING_COLOUR_SECONDARY_HELPTEXT   :Choose starting secondary colour for the company, if using a NewGRF that enables it
```

2TallTyler (commented):
Do you plan to add the other colour options (tint, complementary, etc) in this PR, or leave them for later? I think this stands alone, but don't want to be hasty with an approval. 😃

src/lang/english.txt
M3Henry: I think it's good as it is for now. My guess is that it would be unused.

2TallTyler (dismissed):

src/openttd.cpp
```diff
@@ -870,12 +871,17 @@ static void MakeNewGameDone()
...
 	 * COLOUR_END corresponds to Random colour */
+
 	if (_settings_client.gui.starting_colour != COLOUR_END) {
 		c->colour = _settings_client.gui.starting_colour;
 		ResetCompanyLivery(c);
 		_company_colours[c->index] = (Colours)c->colour;
 	}
 
+	if (_settings_client.gui.starting_colour_secondary != COLOUR_END && HasBit(_loaded_newgrf_features.used_liveries, LS_DEFAULT)) {
+		Command<CMD_SET_COMPANY_COLOUR>::Post(LS_DEFAULT, 0, (Colours)_settings_client.gui.starting_colour_secondary);
```
rubidium42:
```suggestion
		Command<CMD_SET_COMPANY_COLOUR>::Post(LS_DEFAULT, false, (Colours)_settings_client.gui.starting_colour_secondary);
```
The second parameter is a `bool`, so it should be `false`.

2TallTyler (approved):

PR #11073, opened 2023-06-25
Codechange: Pass face index as font os_handle for FreeType fonts.

PeterN:
## Motivation / Problem

Fontconfig/Freetype fallback detection lists suitable fonts but ignores the font index for font collection files. This means that font tested and used may not actually be the font filtered by fontconfig.

## Description

Pass face index as font os_handle for FreeType fonts. This allows fallback font detection to test the specific face within the font rather instead of only the first.

Fallback font detection currently passes the font filename. An alternative solution is to pass the font name and style (which is how regular fonts are set up) but this would mean the that we then search for the font again, by name & style.

## Checklist for review

### DIFF
--- a/src/fontcache/freetypefontcache.cpp
+++ b/src/fontcache/freetypefontcache.cpp
@@ -141,7 +141,9 @@ void LoadFreeTypeFont(FontSize fs)
 	FT_Face face = nullptr;
 
 	/* If font is an absolute path to a ttf, try loading that first. */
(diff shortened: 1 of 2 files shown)


PeterN:
Hmm, I also tried the alternative solution, selecting by family & style to avoid the magic integer pointer, and it seems to that changing the attributes to return also affects the font list's order. So ultimately the selected font is different, but neither way is obviously correct.

rubidium42 (approved):

PR #11074, opened 2023-06-27
Add: show fallback sprites with debuglevel sprite=4

Eddi-z:
## Motivation / Problem

When trying to update a baseset, only the number of missing sprites is shown, not which sprites are missing, which may make the process unnecessarily complicated.

## Description

This adds debuglevel `sprite=4` which lists all sprites taken from the fallback grf. while here, bump lru debug messages to level 5, because they seem to be rather spammy and unhelpful for this usecase.

## Limitations

The original baseset makes intentional use of the fallback, which is a lot spammier than it needs to be.

## Checklist for review

### DIFF
--- a/src/spritecache.cpp
+++ b/src/spritecache.cpp
@@ -205,7 +205,10 @@ uint GetSpriteCountForFile(const std::string &filename, SpriteID begin, SpriteID
 	for (SpriteID i = begin; i != end; i++) {
 		if (SpriteExists(i)) {
 			SpriteCache *sc = GetSpriteCache(i);
-			if (sc->file == file) count++;
+			if (sc->file == file) {
+				count++;
+				Debug(sprite, 4, "Sprite: {}", i);
+			}
 		}
 	}
 	return count;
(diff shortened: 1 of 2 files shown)


src/spritecache.cpp
```diff
@@ -205,7 +205,10 @@ uint GetSpriteCountForFile(const std::string &filename, SpriteID begin, SpriteID
 	for (SpriteID i = begin; i != end; i++) {
 		if (SpriteExists(i)) {
 			SpriteCache *sc = GetSpriteCache(i);
-			if (sc->file == file) count++;
+			if (sc->file == file) {
+				count++;
+				Debug(sprite, 4, "Sprite: {}", i);
```
LordAro, line 210: Feels like this message could be more descriptive. "Fallback sprite in use: " perhaps?

Eddi-z: Might be, but i decided against it because this function seems to be written for a more generic purpose, even though i could only find this one case where it's being used

LordAro: Yeah, that's the only place it's used. I'd rename & move it, tbh.

At the very least, you could add a corresponding Debug "header" message where it's called

Eddi-z: Ok, did that.

LordAro (approved):

PR #11076, opened 2023-06-27
Changing exclusive transport rights

bjornwarmedal:
## Motivation / Problem

As a relatively new player myself and in talks with very new players it's become evident that buying exclusive transport rights works very differently from the player expectation. The current behaviour is that exclusive rights can be bought while another company already has them, causing that company to lose them. The expectation is that buying exclusive transport rights for 12 months will ensure that your company indeed does have exclusive transport rights for 12 months.

In multiplayer games where exclusive rights are disabled, or where a social contract restricts this abuse, this change will have no impact. However, in more competitively inclined multiplayer games this change makes exclusive transport rights a more dynamic and balanced tool for both defensive and offensive play.

## Description

Two changes are made here:
* When a company has bought exclusive transport rights they can neither be extended by that company nor purchased by any other until the 12 month period is over.
* If a company currently own exclusive transport rights and some *other* company successfully bribes the local authority then the exclusive rights contract becomes void, making them available for purchase again.

In the current state a smaller company can never keep exclusive transport rights if a bigger company decides to have them. The bigger company will buy them back until the smaller company either gives up or becomes bankrupt. This means that buying exclusive transport rights is not a viable option for the smaller company neither to protect its own business nor to sabotage for a larger company.

With this change the smaller company can pre-emptively buy exclusive transport rights in its favoured town(s) knowing that the bigger company has to take the risk of being discovered paying bribes in order to gain a foothold in the market. The smaller company may also have an actual chance to inconvenience the larger company in towns where the latter has much business.

This new way of doing it also -- maybe most importantly -- makes the Buy exclusive transport rights functionality more intuitive. When you buy the rights you can expect to have the contract for 12 months. It doesn't make sense that the local authority would sign a deal with one actor and then wave that deal away and sign the same with some other actor the next moment; that's just not how contracts work. Unless foul play (i.e. bribery) is involved, of course.

## Limitations

Exclusive transport rights could work in a thousand different ways. This is one way of doing it. I stand by that this is a big improvement over the current functionality, but I don't believe it to be a perfect solution. Such a thing probably doesn't exist.

The descriptions for buying rights or bribing have not been updated. This should be done, but as the current description of the former already fails to set the correct expectation with players I haven't spent time on it for now.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -3242,6 +3242,7 @@ static CommandCost TownActionBuyRights(Town *t, DoCommandFlag flags)
 {
 	/* Check if it's allowed to buy the rights */
 	if (!_settings_game.economy.exclusive_rights) return CMD_ERROR;
+	if (t->exclusivity != INVALID_COMPANY) return CMD_ERROR;
 
 	if (flags & DC_EXEC) {
 		t->exclusive_counter = 12;
@@ -3292,6 +3293,10 @@ static CommandCost TownActionBribe(Town *t, DoCommandFlag flags)
 			}
 		} else {
 			ChangeTownRating(t, RATING_BRIBE_UP_STEP, RATING_BRIBE_MAXIMUM, DC_EXEC);
+			if (t->exclusivity != _current_company && t->exclusivity != INVALID_COMPANY) {
+				t->exclusivity = INVALID_COMPANY;
+				t->exclusive_counter = 0;
+			}
 		}
 	}
 	return CommandCost();
@@ -3334,7 +3339,7 @@ TownActions GetMaskOfTownActions(CompanyID cid, const Town *t)
 			if (cur == TACT_BRIBE && (!_settings_game.economy.bribe || t->ratings[cid] >= RATING_BRIBE_MAXIMUM)) continue;
 
 			/* Is the company not able to buy exclusive rights ? */
-			if (cur == TACT_BUY_RIGHTS && !_settings_game.economy.exclusive_rights) continue;
+			if (cur == TACT_BUY_RIGHTS && (!_settings_game.economy.exclusive_rights || t->exclusive_counter != 0)) continue;
 
 			/* Is the company not able to fund buildings ? */
 			if (cur == TACT_FUND_BUILDINGS && !_settings_game.economy.fund_buildings) continue;
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3539,8 +3539,8 @@ STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_LARGE_ADVERTISING            :{PUSH_COLOUR}{Y
 STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_ROAD_RECONSTRUCTION          :{PUSH_COLOUR}{YELLOW}Fund the reconstruction of the urban road network.{}Causes considerable disruption to road traffic for up to 6 months.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
 STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_STATUE_OF_COMPANY            :{PUSH_COLOUR}{YELLOW}Build a statue in honour of your company.{}Provides a permanent boost to station rating in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
 STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_NEW_BUILDINGS                :{PUSH_COLOUR}{YELLOW}Fund the construction of new buildings in the town.{}Provides a temporary boost to town growth in this town.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 1 year's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
-STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_BRIBE                        :{PUSH_COLOUR}{YELLOW}Bribe the local authority to increase your rating, at the risk of a severe penalty if caught.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
+STR_LOCAL_AUTHORITY_ACTION_TOOLTIP_EXCLUSIVE_TRANSPORT          :{PUSH_COLOUR}{YELLOW}Buy 1 year's exclusive transport rights in town.{}Town authority will not allow passengers and cargo to use your competitors' stations. A successful bribe from a competitor will cancel this contract.{}{POP_COLOUR}Cost: {CURRENCY_LONG}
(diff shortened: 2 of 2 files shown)


src/town_cmd.cpp
```diff
@@ -3243,7 +3243,7 @@ static CommandCost TownActionBuyRights(Town *t, DoCommandFlag flags)
 	/* Check if it's allowed to buy the rights */
 	if (!_settings_game.economy.exclusive_rights) return CMD_ERROR;
 
-	if (flags & DC_EXEC) {
+	if (flags & DC_EXEC && t->exclusivity == INVALID_COMPANY) {
```
nielsmh: This is not right, this will cause the command to seem to succeed, despite a precondition having failed. The calling function (`CmdDoTownAction` on line 3375) will then think the action has succeeded, and subtract the cost from the company attempting the action.

You need to make a check _before_ this if statement, where you fail the command if the precondition of nobody else having exclusivity fails.

bjornwarmedal: Would that be on line 3244?

The possibility to buy the rights is disabled in the GUI when it can't be done, but I guess that doesn't stop an AI from trying.

nielsmh: It's also a potential race condition if two players just happen to click the button on the exact same tick. Then one will properly succeed, but the other will get nothing but still lose their money.

ldpl:
This is even worse than it currently is. Bought rights in competitor starting town with first income or even loan - gg. Failed bribe to take back rights on your own - gg. But I guess as long as there is a setting to disable this nonsense it doesn't matter, no one should use it anyway aside from bullying AIs. The best way to do any kind of blocking mechanic in the game is not to do it at all.

src/town_cmd.cpp
bjornwarmedal: This race condition is the current behaviour. If two companies spend money on rights both will lose them but only one will walk away with the contract.

I would say that this race condition is sufficiently improbable as to not really care about, but I'll definitely fix it if someone can tell me how.

bjornwarmedal:
> This is even worse than it currently is. Bought rights in competitor starting town with first income or even loan - gg. Failed bribe to take back rights on your own - gg.

If you can borrow enough money at the start of the game to buy exclusive rights you can:
* buy them pre-emptively in your starting town.
* have enough money to just move to a different town.

Early in the game there isn't as much infrastructure in play.

>  But I guess as long as there is a setting to disable this nonsense it doesn't matter, no one should use it anyway aside from bullying AIs.

I've seen similar arguments in other places. I read it as "I think exclusive rights is mean and shouldn't be used at all." and I quite honestly don't see how that's an argument either for or against this change. As I said in the description:

> In multiplayer games where exclusive rights are disabled, or where a social contract restricts this abuse, this change will have no impact. However, in more competitively inclined multiplayer games this change makes exclusive transport rights a more dynamic and balanced tool for both defensive and offensive play.

Exclusive transport rights -- like road reconstruction -- is not *meant* to be kind. I love kind games as much as the next guy and play like that 99% of the time. Exclusive transport rights are for fiercely competitive games, last-man-standing kind of play.

src/town_cmd.cpp
TrueBrain: Also, it is good to remember people make customised clients. So every check done in the GUI also nees to be done on command level, as bypassing the GUI is simple (and will be done). And we kinda want that a command does the same during testing vs execution. As a bit of background.

I think if you move the check one line up in its own if with a CMD_ERROR return, you are good already. But didn't look indepth enough to see if that causes you other trouble :)

See line 3201 for an example.

ldpl:
>     * buy them pre-emptively in your starting town.
You buy pre-emptively, competitor spends it on moneymaker, gg.

>     * have enough money to just move to a different town.
Competitor starts in your town, pulls ahead while you move and buys rights in your new town. 

Basically, it creates a situation where the first one to move loses.

> I've seen similar arguments in other places. I read it as "I think exclusive rights is mean and shouldn't be used at all." and I quite honestly don't see how that's an argument either for or against this change. 

It's neither. All is fine as long as it can be disabled.

> Exclusive transport rights -- like road reconstruction -- is not _meant_ to be kind. I love kind games as much as the next guy and play like that 99% of the time. Exclusive transport rights are for fiercely competitive games, last-man-standing kind of play.

It's not about being kind, it's about having a meaningful gameplay. Having exclusive rights forces you to focus on rigorously watching them and prolonging every year/buying back just so you can play your transport game game normally. Why?

bjornwarmedal:
> It's not about being kind, it's about having a meaningful gameplay. Having exclusive rights forces you to focus on rigorously watching them and prolonging every year/buying back just so you can play your transport game game normally. Why?

That's not my experience with the current implementation, but I think we can agree that you just don't like the concept at all either way :)

zephyris:
I think these suggestions are good. Original TTD behaviour is IMO simplistic and doesn't feel well thought through - realistically they were included in original TTD as a quick little extra for a LAN game or for griefing an AI. Your proposal looks like a more thought-through design.

If not already possible, it'd be good to have a setting which removes aggressive town interaction - bribes, exclusives, road reconstruction, etc. Probably everything except fund new buildings.

src/town_cmd.cpp
bjornwarmedal: Fixed.

```
diff --git a/src/town_cmd.cpp b/src/town_cmd.cpp
index d6446bf27..6891504b9 100644
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -3242,8 +3242,9 @@ static CommandCost TownActionBuyRights(Town *t, DoCommandFlag flags)
 {
        /* Check if it's allowed to buy the rights */
        if (!_settings_game.economy.exclusive_rights) return CMD_ERROR;
+       if (t->exclusivity != INVALID_COMPANY) return CMD_ERROR;
 
-       if (flags & DC_EXEC && t->exclusivity == INVALID_COMPANY) {
+       if (flags & DC_EXEC) {
                t->exclusive_counter = 12;
                t->exclusivity = _current_company;

```

2TallTyler:
I like this change. The whole exclusivity feature is flawed but this makes it more viable for the type of cutthroat games with friends that you describe. If you can change the helptexts to reflect the new behavior, I'd like to approve this. 😄 

@zephyris: Those settings already exist to disable exclusivity and road reconstruction. There isn't one for bribery since its only use prior to this PR is to improve local authority ratings. That could be added, but is probably out of scope for this PR.

2TallTyler (changes requested):
You might consider adding a `Can be cancelled by a successful bribe" to the end of the exclusive transport tooltip.

In any case, you shouldn't change other languages besides `english.txt` in a PR; they can only be changed in the web translator. Please remove those changes and rebase everything into one commit, then I plan to approve. 😃

bjornwarmedal:
> You might consider adding a `Can be cancelled by a successful bribe" to the end of the exclusive transport tooltip.
> 
> In any case, you shouldn't change other languages besides `english.txt` in a PR; they can only be changed in the web translator. Please remove those changes and rebase everything into one commit, then I plan to approve. smiley

Alright, I've retracted the changes to english_*.txt and swedish.txt, and added a sentence to the exclusive rights help text :)

2TallTyler (approved):
Thanks for your hard work! 😄

PR #11080, opened 2023-06-28
Fix: select wants a count, not FD of STDIN + 1

rubidium42:
## Motivation / Problem

Looking at code and wondering why `select` should get `STDIN + 1`.

It coincidentally works because `STDIN` is defined as `0`, but `select` specifies the parameter as number of file descriptors. But if someone were to change it to any other file descriptor, this would not work anymore.

## Description

Just pass `1` instead of `STDIN + 1`.

## Checklist for review

### DIFF
--- a/src/video/dedicated_v.cpp
+++ b/src/video/dedicated_v.cpp
@@ -162,7 +162,7 @@ static bool InputWaiting()
 	FD_SET(STDIN, &readfds);
 
 	/* don't care about writefds and exceptfds: */
-	return select(STDIN + 1, &readfds, nullptr, nullptr, &tv) > 0;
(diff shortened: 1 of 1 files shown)


LordAro (approved):

michicc:
https://man7.org/linux/man-pages/man2/select.2.html seems to somewhat disagree:
```
       nfds   This argument should be set to the highest-numbered file
              descriptor in any of the three sets, plus 1.  The
              indicated file descriptors in each set are checked, up to
              this limit (but see BUGS).
```

frosch123 (changes requested):
I think this is wrong.

rubidium42:
Looks like I was being misled by earlier bits of the manual.

PR #11082, opened 2023-06-28
Codechange: let IConsoleCmdExec accept std::string

rubidium42:
## Motivation / Problem

`.c_str()` in a few places when calling `IConsoleCmdExec`.

## Description

Change the parameter from `const char *` to `const std::string &`.

## Limitations

The internals of `IConsoleCmdExec` have not been updated to C++ string functions; that is something to do at a later moment. It does mean some `.c_str()`s within that function though. There is a net reduction in number of `.c_str()` calls in the code base though.

## Checklist for review

### DIFF
--- a/src/console.cpp
+++ b/src/console.cpp
@@ -225,7 +225,7 @@ static void IConsoleAliasExec(const IConsoleAlias *alias, byte tokencount, char 
 				break;
 
 			case ';': // Cmd separator; execute previous and start new command
-				IConsoleCmdExec(alias_buffer.c_str(), recurse_count);
+				IConsoleCmdExec(alias_buffer, recurse_count);
 
 				alias_buffer.clear();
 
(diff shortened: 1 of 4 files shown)


src/console.cpp
```diff
@@ -300,24 +300,24 @@ void IConsoleCmdExec(const char *cmdstr, const uint recurse_count)
 	bool longtoken = false;
 	bool foundtoken = false;
 
-	if (cmdstr[0] == '#') return; // comments
+	if (command_string[0] == '#') return; // comments
 
-	for (cmdptr = cmdstr; *cmdptr != '\0'; cmdptr++) {
+	for (cmdptr = command_string.c_str(); *cmdptr != '\0'; cmdptr++) {
```
LordAro, line 305: Can this for loop (& the one below) be cleaned up a bit more?

rubidium42: It can probably be cleaned up a bit more, though arguably rethinking/redesigning the whole function with std::string(_view) functions would yield a better result than doing some tiny improvements.

And that's in my opinion out of scope for this PR, as I just want to not add more `.c_str()`s when changing other callers of `IConsoleCmdExec` to use std::string, to get rid of `strecpy`.

LordAro (approved):

PR #11083, opened 2023-06-29
Codechange: use std::getline to read strings from stdin

rubidium42:
## Motivation / Problem

The wish to get rid of `strecpy`.
Also noticing that a dedicated server under Windows may only pass about 200 characters, compared to about 1000 characters for the other platforms, thus being inconsistent since mid 2008.

## Description

Use `std::getline` to get a `std::string` from the console, instead of `fgets` or `ReadFile`.
Use `std::string` for the buffer, instead of a allocated C-string buffer.
Use string's `find_last_not_of` to strip trailing `\r\n` characters.

## Limitations

Windows is not yet tested.

## Checklist for review

### DIFF
--- a/src/video/dedicated_v.cpp
+++ b/src/video/dedicated_v.cpp
@@ -23,6 +23,7 @@
 #include "../saveload/saveload.h"
 #include "../thread.h"
 #include "../window_func.h"
+#include <iostream>
 #include "dedicated_v.h"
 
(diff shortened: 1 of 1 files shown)


glx22 (approved):
Tested on windows (`openttd -D`), no change of behaviour so it's fine for me.

PR #11085, opened 2023-06-29
Codechange: introduce new type and functions for StringParameter backups

rubidium42:
## Motivation / Problem

The current way of storing backups of StringParameters is quite error prone, and requires manual management of memory. Especially in case strings are involved.

## Description

Introduce a new way of storing backups by just returning a vector with a structure containing the data or string value. This removes the need for manual memory management, and having to think about that all throughout the code.
The introduction of the vector also makes it possible to only allocate/copy the elements that are required, instead of some arbitrary amount.

## Limitations

The code is not used yet; there is #11052 that introduces the users, but that seems to be a too large change to get reviewed.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -197,6 +197,56 @@ void CopyOutDParam(uint64 *dst, const char **strings, StringID string, int num)
 	}
 }
 
+/**
+ * Copy the parameters from the backup into the global string parameter array.
+ * @param backup The backup to copy from.
+ */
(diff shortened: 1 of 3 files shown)


PeterN:
Where are these functions used?

rubidium42:
> Where are these functions used?

Nowhere yet (or #11052), as I wrote in the limitations. That's because implementing them and using them seems to be "too big to get approved" (in a reasonable time). Short one commit PRs seem to get approved way faster.

PR #11086, opened 2023-06-30
Codechange: use std::string instead of a temporary buffer for iconv calls

rubidium42:
## Motivation / Problem

Noticing a weird `strecpy` in the `iconv` code. It is copying the input to the output buffer, and then runs iconv which modifies the output buffer again.
After that, the passed buffer is converted to `std::string` in the return.

## Description

Just directly allocate the `std::string` that is going to be returned, with a guess how much bytes will be needed.

If `iconv` fails, return just the input string. It might not be encoded correctly, but conversion failed anyhow so... do what the callers do and return the string that hasn't been converted yet.

Finally just resize the string to what was actually required.

All in all making `FS2OTTD` and `OTTD2FS` reentrant from our code's perspective; especially not reusing the same buffer when two calls are made to `FS2OTTD` at the same time.

## Limitations

None that I'm aware of regarding this code.

Though... glibc has it's own `iconv`, but we do not use it; we only use `iconv` for `APPLE`. Is that right/correct? Shouldn't we run `iconv` on other platforms as well? Or am I overlooking something? (Not that it really matters for me as my system is UTF-8).

## Checklist for review

### DIFF
--- a/src/os/unix/unix.cpp
+++ b/src/os/unix/unix.cpp
@@ -124,29 +124,30 @@ static const char *GetLocalCode()
  * Convert between locales, which from and which to is set in the calling
  * functions OTTD2FS() and FS2OTTD().
  */
-static const char *convert_tofrom_fs(iconv_t convd, const char *name, char *outbuf, size_t outlen)
+static std::string convert_tofrom_fs(iconv_t convd, const std::string &name)
 {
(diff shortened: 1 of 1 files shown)


PeterN (approved):
Looks okay -- replaces the externally allocated temporary output buffer with a simple return value.

PR #11089, opened 2023-06-30
Autorail / autoroad tools can start dragging from invalid tiles

Kuhnovic:
## Motivation / Problem

When you build a section of rail, that section will be built until you hit a tile that it can't build on. In that case the section will stop there, and the valid start of the rail section is still built. This is quite a nice feature especially if you are creating very long stretches of track, or if you accidentally release your mouse while hovering over say an industry tile.

Unfortunately this does not work the same when you start on a tile that can't be built on. It has happened to me more than once that I wanted to create a long stretch of track starting from a tile right next to a station, but I had accidentally clicked the station tile. You then get an error message and nothing is built.

The same problem applies for roads / trams.

## Description

This change fixes that. If the first x tiles can't be built on then those tiles will simply be skipped. The player will only see an error if the entire section is unbuildable. 

Here are some example cases where it can be helpful:
openttd_SteYK7SXRb

## Limitations

None

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -875,8 +875,8 @@ static CommandCost ValidateAutoDrag(Trackdir *trackdir, TileIndex start, TileInd
  * @param railtype railroad type normal/maglev (0 = normal, 1 = mono, 2 = maglev), only used for building
  * @param track track-orientation
  * @param remove remove tracks?
- * @param auto_remove_signals false = build up to an obstacle, true = fail if an obstacle is found (used for AIs), only used for building
- * @param fail_on_obstacle false = error on signal in the way, true = auto remove signals when in the way, only used for building
+ * @param auto_remove_signals false = error on signal in the way, true = auto remove signals when in the way, only used for building
+ * @param fail_on_obstacle false = build starting from and up to an obstacle, true = fail if an obstacle is found (used for AIs)
  * @return the cost of this operation or an error
  */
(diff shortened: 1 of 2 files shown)


src/rail_cmd.cpp
```diff
@@ -900,7 +900,7 @@ static CommandCost CmdRailTrackHelper(DoCommandFlag flags, TileIndex tile, TileI
 			last_error = ret;
 			if (last_error.GetErrorMessage() != STR_ERROR_ALREADY_BUILT && !remove) {
 				if (fail_on_obstacle) return last_error;
-				break;
+				if (had_success) break; // Keep going if we haven't constructed any road yet, skipping the start of the drag
```
2TallTyler:
```suggestion
				if (had_success) break; // Keep going if we haven't constructed any rail yet, skipping the start of the drag
```

2TallTyler (commented):
I really like this. Works as advertised, just one copy-and-paste to fix.

2TallTyler (approved):

rubidium42:
To what extent does this break AIs/GSes? I can imagine them wanting the build to fail if something else built on the tile they want to start from.

Kuhnovic:
> To what extent does this break AIs/GSes? I can imagine them wanting the build to fail if something else built on the tile they want to start from.

AFAIK it's not going to affect them. If the fail_on_obstacle parameter of CmdRailTrackHelper is set to true then it will just return the first error it encounters. This happens before the new code is ever reached. According to the documentation fail_on_obstacle is in fact used by AIs.

Same thing for CmdBuildLongRoad, but there the parameter is called is_ai.

PR #11091, opened 2023-07-01
Fix #11087: Disable base graphics/sound dropdown outside main menu (and change tooltips)

### LINKED ISSUE #11087: [Bug]: Base sound selection disabled
### Version of OpenTTD

1.13.0, Windows 11

### Expected result

I can select an installed base sound set from the game menu

### Actual result

All base sound sets except for "NoSound" are disabled.

image

### Steps to reproduce

1. Install from installer `openttd-13.3-windows-win64` on Windows 11
2. Download OpenSFX through "Check Online Content" in the main menu
3. Start a new game
4. Click on the gear in the top bar
5. Open the drop down under "Base sound sets"

merni-ns:
## Motivation / Problem

Fixes #11087, which applies to base graphics as well as sounds. Also, adds a note in tooltips that those settings can be changed only in the main menu, to avoid confusing players.

## Description
As suggested by @LordAro in #11087:
- Disables the dropdown for base graphics and base sound set selection, outside the main menu.
- Adds a note to the tooltip about the reason.
- The disabling of individual items is no longer needed, so is removed.
- This allows us to remove a redundant parameter of `BuildSetDropDownList`. After doing that, `BuildMusicSetDropDownList` just becomes an alias of `BuildSetDropDownList<BaseMusic>` so the former is also removed.

## Limitations

Players can't see a list of the installed base graphic/sound sets during a game anymore.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1059,12 +1059,12 @@ STR_GAME_OPTIONS_REFRESH_RATE_ITEM                              :{NUM}Hz
 STR_GAME_OPTIONS_REFRESH_RATE_WARNING                           :{WHITE}Refresh rates higher than 60Hz might impact performance.
 
 STR_GAME_OPTIONS_BASE_GRF                                       :{BLACK}Base graphics set
-STR_GAME_OPTIONS_BASE_GRF_TOOLTIP                               :{BLACK}Select the base graphics set to use
+STR_GAME_OPTIONS_BASE_GRF_TOOLTIP                               :{BLACK}Select the base graphics set to use (cannot be changed in-game, only from the main menu)
 STR_GAME_OPTIONS_BASE_GRF_STATUS                                :{RED}{NUM} missing/corrupted file{P "" s}
 STR_GAME_OPTIONS_BASE_GRF_DESCRIPTION_TOOLTIP                   :{BLACK}Additional information about the base graphics set
 
 STR_GAME_OPTIONS_BASE_SFX                                       :{BLACK}Base sounds set
-STR_GAME_OPTIONS_BASE_SFX_TOOLTIP                               :{BLACK}Select the base sounds set to use
+STR_GAME_OPTIONS_BASE_SFX_TOOLTIP                               :{BLACK}Select the base sounds set to use (cannot be changed in-game, only from the main menu)
 STR_GAME_OPTIONS_BASE_SFX_DESCRIPTION_TOOLTIP                   :{BLACK}Additional information about the base sounds set
 
 STR_GAME_OPTIONS_BASE_MUSIC                                     :{BLACK}Base music set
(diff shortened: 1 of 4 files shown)


src/lang/english.txt
```diff
@@ -1059,12 +1059,12 @@ STR_GAME_OPTIONS_REFRESH_RATE_ITEM                              :{NUM}Hz
 STR_GAME_OPTIONS_REFRESH_RATE_WARNING                           :{WHITE}Refresh rates higher than 60Hz might impact performance.
 
 STR_GAME_OPTIONS_BASE_GRF                                       :{BLACK}Base graphics set
-STR_GAME_OPTIONS_BASE_GRF_TOOLTIP                               :{BLACK}Select the base graphics set to use
+STR_GAME_OPTIONS_BASE_GRF_TOOLTIP                               :{BLACK}Select the base graphics set to use (cannot be changed in-game, only in the main menu)
```
LordAro: Bikeshedding: "only from the main menu"

Also, is "in-game" usually hyphenated?

merni-ns: > Also, is "in-game" usually hyphenated?

https://github.com/OpenTTD/OpenTTD/blob/master/src/lang/english.txt#L3308 is the only occurrence of in-game/ingame in `english.txt` and it is hyphenated.

src/settings_gui.cpp
```diff
@@ -291,11 +291,11 @@ struct GameOptionsWindow : Window {
 				break;
 
 			case WID_GO_BASE_GRF_DROPDOWN:
-				list = BuildSetDropDownList<BaseGraphics>(selected_index, (_game_mode == GM_MENU));
+				list = BuildSetDropDownList<BaseGraphics>(selected_index, true);
 				break;
 
 			case WID_GO_BASE_SFX_DROPDOWN:
-				list = BuildSetDropDownList<BaseSounds>(selected_index, (_game_mode == GM_MENU));
+				list = BuildSetDropDownList<BaseSounds>(selected_index, true);
```
rubidium42: With this change all callers of `BuildSetDropDownList` pass `true`, so that parameter can be removed and `BuildSetDropDownList`'s internal logic can be simplified a bit.

merni-ns: Done.

src/settings_gui.cpp
```diff
@@ -75,24 +75,19 @@ static uint GetCurrentResolutionIndex()
 static void ShowCustCurrency();
 
 template <class T>
-static DropDownList BuildSetDropDownList(int *selected_index, bool allow_selection)
+DropDownList BuildSetDropDownList(int *selected_index)
```
glx22: Maybe use a reference and not a pointer here? The code seems to assume it's never `nullptr`.
(But it might be a rabbit hole)

merni-ns: It was already like that and doesn't seem to have caused any issues so far :P 

I would prefer to leave that to a future PR as it is unrelated to this change.

glx22 (approved):

PR #11093, opened 2023-07-01
Change: [CI] upload releases to new CDN

TrueBrain:
## Motivation / Problem

To reduce cost, we are switching from AWS S3 to Cloudflare R2. This means another way of uploading binaries to our CDN.

## Description

Change the way we upload binaries to our CDN, and do this via a reusing-workflow. This means we have one place that defines how uploads needs to be done, which works for all other projects.

As the reusing workflow needs an artifact to know what to upload, another artifact is created. But by now we have so many artifacts that are unused by GOG and Steam, also change those two to be more specific in which artifacts they want.

It is really hard to test GOG / Steam; so I am going for a "it looks fine" review on those two.

Note: this work has to be merged together with https://github.com/OpenTTD/workflows/pull/34 and friends, while a change in the infra is pushed through.

## Checklist for review

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,8 +76,8 @@ jobs:
     with:
       version: ${{ needs.source.outputs.version }}
 
-  upload-aws:
-    name: Upload (AWS)
+  upload-cdn:
+    name: Upload (CDN)
     needs:
     - source
     - docs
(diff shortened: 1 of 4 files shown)


glx22 (approved):
Nothing obviously wrong (and steam upload will be tested with nightly)

PR #11094, opened 2023-07-01
Codechange: use Textbuf directly, instead via several virtual functions in Window

rubidium42:
## Motivation / Problem

In the console and querystring GUI there are duplicated functions to get certain information from `Textbuf`. These functions are only called from the MacOS specific code. Why not just pass the `Textbuf` and use that directly?

Furthermore some of the functions are overly complicated for what they do. For example `GetMarkedText` returns a pointer (which could be `nullptr`) and a length (built from end - start offsets from the complete string). The in the MacOS code that pointer is taken and the length is added to is. So in essence: `char *end = (text + markpos) + (markpos - markend)`; this can be much simpler by just accessing the fields directly.

## Description

Change the `Window` functions to just one version that returns the `Textbuf`.
Move the `GetText()` to `Textbuf` from `Querystring` and `IConsoleWindow`.
Remove some unneeded indirections, but just getting `caretpos`, `markpos` and `markend` directly from `Textbuf` instead of the current (sometimes overly complex) functions.

## Limitations

I can't test it.

## Checklist for review

### DIFF
--- a/src/console_gui.cpp
+++ b/src/console_gui.cpp
@@ -280,22 +280,9 @@ struct IConsoleWindow : Window
 		}
 	}
 
-	const char *GetFocusedText() const override
+	Textbuf *GetFocusedTextbuf() const override
 	{
-		return _iconsole_cmdline.buf;
-	}
-
-	const char *GetCaret() const override
-	{
-		return _iconsole_cmdline.buf + _iconsole_cmdline.caretpos;
-	}
-
-	const char *GetMarkedText(size_t *length) const override
-	{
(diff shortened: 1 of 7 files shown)


TrueBrain:
@andythenorth mind testing this PR, see if console and querystrings still work as expected? Especially test things like moving left/right with cursor to see if the cursor updates correctly please. Tnx :)

andythenorth:
I built and tested this on macOS 12.6.8 (Monterey). 
image

Tried and failed to trigger any unexpected behaviour in:
- console
- text entry UI widget
- cursor

TrueBrain (approved):

PR #11102, opened 2023-07-02
Change: Only show platform stopping location in orders when other than default

2TallTyler:
## Motivation / Problem

The order window has a lot of text.

## Description

We don't draw loading/unloading instructions unless they are different than the default `Load if available / Unload if accepted`. Let's do the same with the platform stopping location: only draw the text if it's different than the default chosen by the player in their GUI settings.

stop_location
(In this screenshot, the player default is middle.)

None of this inhibits the player cycling through the stopping locations by clicking the order text. I updated the Settings helptext to clarify this.

## Limitations

New players may not realize they can change the stopping location just from looking at the order window. The feature is purely cosmetic, so I don't see a problem with this.

## Checklist for review

### DIFF
--- a/src/order_gui.cpp
+++ b/src/order_gui.cpp
@@ -287,7 +287,12 @@ void DrawOrderString(const Vehicle *v, const Order *order, int order_index, int 
 					SetDParam(4, order->IsAutoRefit() ? STR_ORDER_AUTO_REFIT_ANY : CargoSpec::Get(order->GetRefitCargo())->name);
 				}
 				if (v->type == VEH_TRAIN && (order->GetNonStopType() & ONSF_NO_STOP_AT_DESTINATION_STATION) == 0) {
-					SetDParam(5, order->GetStopLocation() + STR_ORDER_STOP_LOCATION_NEAR_END);
+					/* Only show the stopping location if other than the default chosen by the player. */
+					if (order->GetStopLocation() != (OrderStopLocation)(_settings_client.gui.stop_location)) {
+						SetDParam(5, order->GetStopLocation() + STR_ORDER_STOP_LOCATION_NEAR_END);
+					} else {
+						SetDParam(5, STR_EMPTY);
+					}
 				}
 			}
 			break;
--- a/src/table/settings/gui_settings.ini
+++ b/src/table/settings/gui_settings.ini
@@ -600,6 +600,7 @@ interval = 1
 str      = STR_CONFIG_SETTING_STOP_LOCATION
 strhelp  = STR_CONFIG_SETTING_STOP_LOCATION_HELPTEXT
 strval   = STR_CONFIG_SETTING_STOP_LOCATION_NEAR_END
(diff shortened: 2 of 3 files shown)


M3Henry:
I like the idea, but when I tried it, it wasn't immediately obvious that I could click in the empty space after the text to change it from the default. And for a new user it's not obvious that they _shouldn't_ click there if they want to select an order.

- Perhaps empty `[ ]` would do?
- Add another drop-down, to make it consistent with the load type?

PeterN:
This doesn't really fit any existing UI and may be fiddly to implement, but could it show the text for the current line the mouse cursor is over? So it's hidden most of the time, unless you go to select or click on it and then it's more obviously somewhere to click to change it.

llugo:
> So it's hidden most of the time, unless you go to select or click on it and then it's more obviously somewhere to click to change it.

There is quite a lot of "hidden stuff" which is available through a ctrl-click. So a ctrl-click on an existing order to change platform stopping location might be more in-line with current UI/UX?

src/order_gui.cpp
```diff
@@ -287,7 +287,12 @@ void DrawOrderString(const Vehicle *v, const Order *order, int order_index, int
 					SetDParam(4, order->IsAutoRefit() ? STR_ORDER_AUTO_REFIT_ANY : CargoSpec::Get(order->GetRefitCargo())->name);
 				}
 				if (v->type == VEH_TRAIN && (order->GetNonStopType() & ONSF_NO_STOP_AT_DESTINATION_STATION) == 0) {
-					SetDParam(5, order->GetStopLocation() + STR_ORDER_STOP_LOCATION_NEAR_END);
+					/* Only show the stopping location if other than middle. */
```
rubidium42:
```suggestion
					/* Only show the stopping location if other than the configured default. */
```

src/table/settings/gui_settings.ini
```diff
@@ -585,6 +585,7 @@ interval = 1
 str      = STR_CONFIG_SETTING_STOP_LOCATION
 strhelp  = STR_CONFIG_SETTING_STOP_LOCATION_HELPTEXT
 strval   = STR_CONFIG_SETTING_STOP_LOCATION_NEAR_END
+post_cb  = [](auto) { MarkWholeScreenDirty(); }
```
rubidium42: Isn't `SetWindowClassesDirty(WC_VEHICLE_ORDERS)` more to the point? After all, only those parts of the UI need to be updated, not the rest of the screen. Especially when there are no orders shown.

2TallTyler:
>I like the idea, but when I tried it, it wasn't immediately obvious that I could click in the empty space after the text to change it from the default. And for a new user it's not obvious that they shouldn't click there if they want to select an order.

Nothing special happens if you click there. It only changes the stop location if you click twice.

Adding another dropdown probably makes the most sense with the current UI style, but...oof. The code is not designed for that, and an additional button would be quite visually busy.

rubidium42 (approved):

PR #11103, opened 2023-07-03
Codechange: support string temporaries as DParam/Remove: stredup

rubidium42:
## Motivation / Problem

Having to keep local variables around to prevent string temporaries with the string parameter system.
The existence of `stredup`/manual memory management, especially to validate whether a game string parameter is valid.

## Description

Add a `std::unique_ptr<std::string>` to retain a copy of a string in the `StringParameter`. Not directly a `std::string`, as we need a mechanism to know whether the value is used and given temporaries are rarely used the overhead of `std::unique_ptr` over `std::optional` is negligable compared to the memory usage.

Use this new support for temporaries to not have to `stredup` string parameters, and to not have to maintain an array of fields to `free` later.

Remove `stredup`.

Use this new support for temporaries to remove the local variables in callers of `SetDParamStr`.

## Checklist for review

### DIFF
--- a/src/script/api/script_text.cpp
+++ b/src/script/api/script_text.cpp
@@ -248,8 +248,6 @@ void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output,
 
 const std::string Text::GetDecodedText()
 {
-	const std::string &encoded_text = this->GetEncodedText();
-
-	::SetDParamStr(0, encoded_text);
+	::SetDParamStr(0, this->GetEncodedText());
 	return ::GetString(STR_JUST_RAW_STRING);
 }
--- a/src/script/api/script_text.cpp
+++ b/src/script/api/script_text.cpp
@@ -158,7 +158,7 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
 	return this->_SetParam(k, vm);
 }
 
-const std::string ScriptText::GetEncodedText()
+std::string ScriptText::GetEncodedText()
 {
 	static StringIDList seen_ids;
 	int param_count = 0;
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -859,8 +859,6 @@ uint ConvertDisplaySpeedToKmhishSpeed(uint speed, VehicleType type)
 	return _units_velocity[GetVelocityUnits(type)].c.FromDisplay(speed * 16, true, 10);
 }
 
-static std::vector<const char *> _game_script_raw_strings;
-
 /**
  * Parse most format codes within a string and write the result to a buffer.
  * @param builder The string builder to write the final string to.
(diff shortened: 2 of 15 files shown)


PeterN (dismissed):

src/strings.cpp
```diff
@@ -363,6 +363,18 @@ void SetDParamStr(size_t n, const std::string &str)
...
+/**
+ * This function is used to "bind" the std::string to a OpenTTD dparam slot.
+ * Contrary to the other \c SetDParamStr functions, this moves the string into
+ * the parameter slot.
+ * @param n slot of the string
+ * @param str string to bind
+ */
+void SetDParamStr(size_t n, std::string &&str)
+{
+	_global_string_params.SetParam(n, std::move(str));
```
PeterN, line 375: Is it valid to use std::move here? str might not be a temporary.

src/strings.cpp
```diff
@@ -986,26 +994,14 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 						sub_args.SetParam(i++, param);
 					} else {
 						s++; // skip the leading \"
-						char *g = stredup(s);
-						g[p - s - 1] = '\0'; // skip the trailing \"
-
-						sub_args_need_free[i] = true;
-						sub_args.SetParam(i++, (uint64)(size_t)g);
-						_game_script_raw_strings.push_back(g);
+						sub_args.SetParam(i++, std::string(s, p - s - 1)); // also skip the trailing \".
```
glx22, line 997: ~~suggestion~~
						~~sub_args.SetParam(i++, std::string(s, p - s)); // also skip the trailing \".~~
~~Last char vs size I think.~~

Actually I was wrong, I built the PR to check.

src/script/api/script_text.cpp
```diff
@@ -248,8 +248,6 @@ void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output,
 
 const std::string Text::GetDecodedText()
 {
-	const std::string &encoded_text = this->GetEncodedText();
-
-	::SetDParamStr(0, encoded_text);
+	::SetDParamStr(0, this->GetEncodedText());
```
glx22, line 251: For yet unknown reason, MSVC uses `void SetDParamStr(size_t n, const std::string &str)` for this call.

End of SetDParamStr call 
image

And just before GetString call 
image

src/strings.cpp
rubidium42: https://godbolt.org/z/nhr8d74ch

If there is only a `std::string &&`, then passing non-rvalues will yield a compilation error... unless you explicitly use `std::move`.
If there is only a `const std::string &`, then thing will just be as the were before.
With this change there is a `const std::string &` and `std::string &&`, which means that for rvalues the latter is called which does the `std::move` to make it implicit that it should be passed on. The former remains as is.

So, the only way to pass a `std::string` that should not be copied to `SetDParamStr` is when it's either a (temporary) rvalue, or when you explicitly `std::move` it. It's practically the same construct that is used for copy vs move constructors/assignments.

src/script/api/script_text.cpp
rubidium42: `GetEncodedText` returned `const std::string`, which means MSVC does not see it as a temporary/rvalue. By making `GetEncodedText` just return `std::string`
Note that returning `const std::string` was common practice before C++11, however since it has become less common as it hinders move semantics. Also, the language has become more strict what must be done with respect to copy elision. This copy elision seems to have been one of the reasons to return const objects.

PeterN (approved):

PR #11106, opened 2023-07-04
Fix: [Emscripten] switch to URL for content-service that supports WebSocket

TrueBrain:
## Motivation / Problem

In the old days, content.openttd.org and bananas-server.openttd.org ended up on the same route. But with a recent migration, they do not. content.openttd.org only serves the custom TCP protocol, and bananas-server.openttd.org only serves the HTTP(S).

Websockets use HTTPS, and as such, should be routed via the latter.

## Description

Update the URLs to work with new infra.

Additionally, `installer.cdn` is being deprecated, so move it to another URL that works (for now). But honestly, this part should be removed and the bootstrapping should be made functional instead. That would mean it would automatically pick up on new OpenGFX, instead of this custom work. Emscripten is the last target not supporting bootstrapping. But .. not for this PR!

## Checklist for review

### DIFF
--- a/os/emscripten/pre.js
+++ b/os/emscripten/pre.js
@@ -2,7 +2,7 @@ Module.arguments.push('-mnull', '-snull', '-vsdl:relative_mode');
 Module['websocket'] = { url: function(host, port, proto) {
     /* openttd.org hosts a WebSocket proxy for the content service. */
     if (host == "content.openttd.org" && port == 3978 && proto == "tcp") {
-        return "wss://content.openttd.org/";
+        return "wss://bananas-server.openttd.org/";
     }
 
     /* Everything else just tries to make a default WebSocket connection.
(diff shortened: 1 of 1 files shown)


os/emscripten/pre.js
```diff
@@ -41,7 +41,7 @@ Module.preRun.push(function() {
         /* Check if the OpenGFX baseset is already downloaded. */
         if (!FS.analyzePath(content_download_dir + '/baseset/opengfx-0.6.0.tar').exists) {
             window.openttd_downloaded_opengfx = true;
-            FS.createPreloadedFile(content_download_dir + '/baseset', 'opengfx-0.6.0.tar', 'https://installer.cdn.openttd.org/emscripten/opengfx-0.6.0.tar', true, true);
+            FS.createPreloadedFile(content_download_dir + '/baseset', 'opengfx-0.6.0.tar', 'https://binaries.openttd.org/installer/emscripten/opengfx-0.6.0.tar', true, true);
```
glx22, line 44: There's no more recent version to pick ?

TrueBrain: Nope; see my bla in the description :)

glx22 (approved):

PR #11109, opened 2023-07-05
Add: [Emscripten] support for bootstrapping

glx22:
## Motivation / Problem
Emscripten build comes with an outdated opengfx-0.6.0 baseset and as it gets it via the old installer method (which was since replaced by bootstrapping), it will never get a recent version (well user can update it manually via content download).

## Description
Implement bootstrap mechanism for emscripten builds.
Doesn't show any confirmation window and just downloads a baseset if none was present.
A message is shown during the download and the page is reloaded once downloading is done.

## Checklist for review

### DIFF
--- a/os/emscripten/pre.js
+++ b/os/emscripten/pre.js
@@ -30,24 +30,6 @@ Module.preRun.push(function() {
 
     Module.addRunDependency('syncfs');
     FS.syncfs(true, function (err) {
-        /* FS.mkdir() tends to fail if parent folders do not exist. */
-        if (!FS.analyzePath(content_download_dir).exists) {
-            FS.mkdir(content_download_dir);
-        }
-        if (!FS.analyzePath(content_download_dir + '/baseset').exists) {
-            FS.mkdir(content_download_dir + '/baseset');
-        }
-
-        /* Check if the OpenGFX baseset is already downloaded. */
-        if (!FS.analyzePath(content_download_dir + '/baseset/opengfx-0.6.0.tar').exists) {
-            window.openttd_downloaded_opengfx = true;
-            FS.createPreloadedFile(content_download_dir + '/baseset', 'opengfx-0.6.0.tar', 'https://binaries.openttd.org/installer/emscripten/opengfx-0.6.0.tar', true, true);
-        } else {
-            /* Fake dependency increase, so the counter is stable. */
-            Module.addRunDependency('opengfx');
-            Module.removeRunDependency('opengfx');
-        }
-
         Module.removeRunDependency('syncfs');
     });
 
@@ -72,6 +54,23 @@ Module.preRun.push(function() {
 
     window.openttd_abort = function() {
         window.openttd_syncfs(Module.onAbort);
+    }
+
+    window.openttd_bootstrap = function(current, total) {
+        Module.onBootstrap(current, total);
+    }
+
+    window.openttd_bootstrap_failed = function() {
+        Module.onBootstrapFailed();
+    }
+
+    window.openttd_bootstrap_reload = function() {
+        window.openttd_syncfs(function() {
+            Module.onBootstrapReload();
+            setTimeout(function() {
+                location.reload();
+            }, 1000);
+        });
     }
 
     window.openttd_server_list = function() {
@@ -124,12 +123,4 @@ Module.preRun.push(function() {
        Module['websocket']['url'] = func;
        return ret;
    }
-});
-
-Module.postRun.push(function() {
-    /* Check if we downloaded OpenGFX; if so, sync the virtual FS back to the
-     * IDBFS so OpenGFX is stored persistent. */
-    if (window['openttd_downloaded_opengfx']) {
-        FS.syncfs(false, function (err) { });
-    }
 });
--- a/os/emscripten/pre.js
+++ b/os/emscripten/pre.js
@@ -52,24 +52,26 @@ Module.preRun.push(function() {
     });
 
     window.openttd_syncfs_shown_warning = false;
-    window.openttd_syncfs = function() {
+    window.openttd_syncfs = function(callback) {
         /* Copy the virtual FS to the persistent storage. */
-        FS.syncfs(false, function (err) { });
+        FS.syncfs(false, function (err) {
+            /* On first time, warn the user about the volatile behaviour of
+             * persistent storage. */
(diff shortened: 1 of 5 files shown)


os/emscripten/shell.html
```diff
@@ -171,6 +173,13 @@
           document.getElementById("message").innerHTML = "Preparing game ...";
         },
 
+        onBootstrap: function() {
+          document.getElementById("canvas").style.display = "none";
+
+          document.getElementById("title").innerHTML = "Missing graphics";
```
TrueBrain: Should we name it "base graphics" here instead of "graphics"?

src/bootstrap_gui.cpp
```diff
@@ -278,6 +278,59 @@ class BootstrapAskForDownloadWindow : public Window, ContentCallback {
...
+
+	void OnReceiveContentInfo(const ContentInfo *ci) override
+	{
+		if (this->downloading) return;
+		/* And once the meta data is received, start downloading it. */
+		_network_content_client.Select(ci->id);
+		_network_content_client.DownloadSelectedContent(this->total_files, this->total_bytes);
+		this->downloading = true;
+	}
+
```
TrueBrain, line 336: Should we define an OnFailure callback too, in case download fails?

os/emscripten/pre.js
```diff
@@ -72,6 +54,14 @@ Module.preRun.push(function() {
         Module.onAbort();
     }
 
+    window.openttd_bootstrap = function() {
+        Module.onBootstrap();
+    }
+
+    window.openttd_reload = function() {
+        location.reload();
```
TrueBrain: Maybe update the status too, to say: finished, reloading?

glx22: Reloading will happen faster than the update of the message I think.

os/emscripten/shell.html
glx22: I copied the string from language file.
But it's a good idea.

src/bootstrap_gui.cpp
glx22: `ContentCallback` doesn't have `OnFailure`

src/video/sdl2_v.cpp
```diff
@@ -613,7 +613,11 @@ void VideoDriver_SDL_Base::LoopOnce()
 		 * the stack to be unwound, the code after MainLoop() in
 		 * openttd_main() is never executed. */
 		EM_ASM(if (window["openttd_syncfs"]) openttd_syncfs());
-		EM_ASM(if (window["openttd_exit"]) openttd_exit());
+		if (_game_mode == GM_BOOTSTRAP) {
+			EM_ASM(if (window["openttd_reload"]) openttd_reload());
+		} else {
+			EM_ASM(if (window["openttd_exit"]) openttd_exit());
+	}
```
TrueBrain: hihi, this bracket ran away from the rest! :P (I love nitpicking)

os/emscripten/pre.js
```diff
@@ -30,32 +30,11 @@ Module.preRun.push(function() {
...
-
         Module.removeRunDependency('syncfs');
     });
 
     window.openttd_syncfs_shown_warning = false;
-    window.openttd_syncfs = function() {
-        /* Copy the virtual FS to the persistent storage. */
-        FS.syncfs(false, function (err) { });
-
+    window.openttd_syncfs_warn = function() {
```
TrueBrain: I think we should make this function call `FS.syncfs`, and take as parameter a callback to call when things are done.

This because there is now a lot of duplication going on, where there doesn't have to be.

We could even collapse this function with `openttd_syncfs`, and have the callback default to `undefined`, and not being called in that case. This means `openttd_exit` will look something like:

```
window.openttd_exit = function() {
  window.openttd_sync(() => {
    Module.onExit();
  });
}
```

Minus any typos, ofc.

src/bootstrap_gui.cpp
TrueBrain: Last we spoke you indicated it crashes when downloading failed. Did you manage to address this part? What does it currently do if the user fails to download OpenGFX?

src/bootstrap_gui.cpp
```diff
@@ -278,6 +278,69 @@ class BootstrapAskForDownloadWindow : public Window, ContentCallback {
...
+#include "network/network.h"
+#include "network/network_content.h"
+#include "openttd.h"
+#include "video/video_driver.hpp"
+
+class BootstrapEmscripten : public ContentCallback {
+	bool downloading = false;
+	uint total_files;
+	uint total_bytes;
+	uint downloaded_bytes;
```
TrueBrain: please initialize these values. At least ,how I read it, `downloaded_bytes` is never actually set to zero, but printed nevertheless :)

total_files and total_bytes strictly seen are initialized before use, but it might just be good to assign values to them, to make perfectly clear they are always initialized correctly.

src/bootstrap_gui.cpp
TrueBrain: From what I read, you could use `this->IsCancelled()` here to find out if there will be a retry if `bytes < 0`. As if it is true, there will not be a retry, and this is the end of it. If it is false, it will fall back to the fallback, and so there will be another round.

This at least would allow us to tell the user bootstrapping failed, and they are on their own. I think.

glx22: `IsCancelled()` is not a member of `ContentCallback`.
And I just tried win64 exe with blocked port 3978, it seems to crash too (without crash report because bootstrapping happens early), but I need to confirm in debugger.

TrueBrain: Turns out you need to monitor `OnConnect` to know whether it failed. Don't ask me why :P But that is what the normal boostrap does :)

TrueBrain:
As this PR has been stale for a while now, and I kinda want to have it in, I spend some time fixing the remaining issues. For visibility, I put them here in a commit:

https://github.com/OpenTTD/OpenTTD/commit/ab6079e28b4808e29bb9f73183f305ab7c3fb007

Basically, the game now shows a proper error when the base graphics can't be downloaded (instead of a crash).

The rest are just bikeshedding things. I was a bit confused with `onReload`, as it should only be called by bootstrap. So I renamed it to `onBootstrapReload`. Stuff like that. Nothing fancy.

TrueBrain (approved):

PR #11112, opened 2023-07-05
Codechange: rename cargo aging days to periods, as they are not really days

rubidium42:
## Motivation / Problem

Cargo days or days in transit are a big misnomer, causing confusion. Especially when day length gets introduced.

They are the number of cargo ageing calls/periods that have been done. The length of the period is defined by the vehicle NewGRF, with a default of 2.5 original game days (185 ticks).

## Description

Rename `days_in_transit` to `periods_in_transit`, `transit_days` to `transit_periods` and `_cargo_days` to `_cargo_periods`.
This includes a savegame bump to rename the field in the savegame.

## Limitations

Maybe `periods` on itself is not clear enough, but I tried to solve that mostly with documentation in the code of cargopacket. Going for `cargo_periods` will make the current `cargo_days_in_transit` (equivalent to person-hours) a bit weird, and the same holds with `cargo_aging_periods` which would mean `cargo_cargo_aging_periods_in_transit`, which is equivally weird.

## Checklist for review


PeterN:
I also spent quite a long time looking at these variables and wondering what the better name would be, also the SaveLoad dilemma.

Additionally I don't think the CargoSpec data should be in a two-entry array -- yes they are related, but there is not really anything about their usage that is array-like.

TrueBrain (approved):
It is at least better. Not sure if it is good, but it isn't worse. So yeah, let's go for this :)

PR #11116, opened 2023-07-07
Change: [CI] rework preview flow and use Cloudflare Pages to publish

TrueBrain:
## Motivation / Problem

Currently previews are published on AWS, and charges an insane amount of money to serve that. So, as with other services, we are migrating these things to Cloudflare.

## Description

We publish every PR on their own Cloudflare Pages branch. Other parts of the infrastructure makes sure that we get a pretty URL like `https://preview.openttd.org/prNNNN/`.

Additionally, went for a much simpler workflow; we used to check if you were a core-developer before the preview label was accepted, as we weren't sure that translators couldn't slap on the label. But honestly, the amount of code required to handle that is not worth it.

Example of successful deployment: https://github.com/TrueBrain/OpenTTD/pull/27 . And because the preview URL is not deployed yet (as it awaits merging of this PR), you can see the result here: https://preview-preview.openttd.org/pr27/

## Limitations

All current deployed previews will have to be remade; but this is as simple as removing the preview label and adding it back again.

## Checklist for review


.github/workflows/preview-build.yml
```diff
@@ -0,0 +1,107 @@
...
+        # Ensure we use the latest version of npm; the one we get with current
+        # emscripten doesn't allow running "npx wrangler" as root.
+        npm install -g npm
+
+    - name: Publish preview
+      uses: cloudflare/pages-action@v1
+      with:
+        apiToken: ${{ secrets.PREVIEW_CLOUDFLARE_API_TOKEN }}
+        accountId: ${{ secrets.PREVIEW_CLOUDFLARE_ACCOUNT_ID }}
+        projectName: ${{ vars.PREVIEW_CLOUDFLARE_PROJECT_NAME }}
```
glx22, line 105: For secrets they have `required: true` but what happens if this var is not set?

TrueBrain: Sadly, you can't define if vars should be set yes/no. But as they are provisioned together, you can be sure they are (via the Infrastructure-as-Code that provisions them).

One of those minor missing things in GitHub .. you can say inputs and secrets should be there, but you can't tell (yet?) if vars should be there.

Mainly why the secrets are required, is to make sure forks don't execute the code (the action will fail instantly). And that is all that really matters here :)

glx22 (approved):

PR #11117, opened 2023-07-07
Codechange: Unnecessry ptr incrementation, max_sprite_size constant

DexterIV:
## Motivation / Problem

First pr to the repo, trying to get familiar with it. Spotted a small "bug". Also added a constexpr value to not calculate the size over and over. Doesnt fix any bugs/issues, but makes code cleaner.

## Description

pixel was a pointer created in a for loop on each iteration, incrementing its pointer at the end of the loop was an unnecessary operation. converted to reference so its less error prone. other solution could be to define it outside the for loop and leave the incrementation. 

## Limitations

none

## Checklist for review

* The bug fix is important enough to be backported? (label: 'backport requested') - no
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md) - no
* This PR affects the save game format? (label 'savegame upgrade') - no
    * ai_changelog.hpp, gs_changelog.hpp need updating. - no
    * newgrf_debug_data.h may need updating. - no


src/spriteloader/grf.cpp
```diff
@@ -191,23 +192,22 @@ bool DecodeSingleSprite(SpriteLoader::Sprite *sprite, SpriteFile &file, size_t f
...
-			byte *pixel = &dest[i * bpp];
+			byte &pixel = dest[i * bpp];
 
 			if (colour_fmt & SCC_RGB) {
-				sprite->data[i].r = *pixel++;
-				sprite->data[i].g = *pixel++;
-				sprite->data[i].b = *pixel++;
+				sprite->data[i].r = pixel++;
+				sprite->data[i].g = pixel++;
+				sprite->data[i].b = pixel++;
```
nielsmh: These are not equivalent.

`*pixel++` reads the value pointed to, then increments the pointer to point at the next value.
Your change makes it increment the pointed-to value and does not move the pointer.

nielsmh (changes requested):

src/spriteloader/grf.cpp
DexterIV: fixed

DexterIV:
commit from wrong acc, will recreate

PR #11125, opened 2023-07-10
Fix #11115: Focus the abandon game/exit game windows

### LINKED ISSUE #11115: [Bug]: Window focus for keyboard shortcuts doesn't respect draw order/priority of windows
### Version of OpenTTD

openttd-20230515-master-g389b66bd16-windows-win64

### Expected result

When a new window pops up, eg. the "Abandon Game" dialog, then it should be the target of keyboard shortcuts instead of any already open windows.

### Actual result

New windows don't seem to correctly gain the focus of keyboard shortcuts, eg. hitting enter can accept/clear/close a window behind a new "Abandon Game" dialog.

### Steps to reproduce

1. Start a new game.
2. Select "Save Game" from the floppy disk icon dropdown, this opens the "Save Game" dialog.
3. Select "Abandon Game" from the floppy disk dropdown, this opens the "Abandon Game" warning dialog.
4. Hit enter or return.
This results in the game being saved (ie. enter in the "Save Game" dialog) rather than abandoning the game (ie. enter in "Abandon Game" dialog). Clicking on the "Abandon Game" dialog before hitting enter results in abandoning the game, as expected, pointing to a keyboard focus issue.

merni-ns:
## Motivation / Problem

Fixes #11115.

## Description

- Add a parameter to `ShowQuery` which gives focus to the query window created by the function. By default it is `false` to preserve current behaviour.
- Use that parameter to focus the "Abandon game" and "Exit game" confirmation queries.
- Also, the query created by `ShowQuery` is not modal even though it is described as modal by the comment. So that comment is fixed.

## Limitations

Opening the abandon/exit query windows will now remove focus on whatever window is currently focussed.

## Checklist for review

### DIFF
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -1218,15 +1218,16 @@ static WindowDesc _query_desc(
 );
 
 /**
- * Show a modal confirmation window with standard 'yes' and 'no' buttons
+ * Show a confirmation window with standard 'yes' and 'no' buttons
  * The window is aligned to the centre of its parent.
  * @param caption string shown as window caption
  * @param message string that will be shown for the window
  * @param parent pointer to parent window, if this pointer is nullptr the parent becomes
  * the main window WC_MAIN_WINDOW
  * @param callback callback function pointer to set in the window descriptor
+ * @param focus whether the window should be focussed (by default false)
  */
-void ShowQuery(StringID caption, StringID message, Window *parent, QueryCallbackProc *callback)
+void ShowQuery(StringID caption, StringID message, Window *parent, QueryCallbackProc *callback, bool focus)
 {
 	if (parent == nullptr) parent = GetMainWindow();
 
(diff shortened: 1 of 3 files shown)


src/misc_gui.cpp
```diff
@@ -1218,15 +1218,16 @@ static WindowDesc _query_desc(
...
 /**
- * Show a modal confirmation window with standard 'yes' and 'no' buttons
+ * Show a confirmation window with standard 'yes' and 'no' buttons
  * The window is aligned to the centre of its parent.
  * @param caption string shown as window caption
  * @param message string that will be shown for the window
  * @param parent pointer to parent window, if this pointer is nullptr the parent becomes
  * the main window WC_MAIN_WINDOW
  * @param callback callback function pointer to set in the window descriptor
+ * @param focus whether the window should be focussed (by default false)
```
2TallTyler, line 1228:
```suggestion
 * @param focus whether the window should be focused (by default false)
```

2TallTyler (commented):
We could probably bikeshed about what other windows could or should be focused...but let's not. The new default parameter is a nice way to handle this, I think.

src/misc_gui.cpp
merni-ns: https://www.dictionary.com/browse/focus Both forms seem to be correct :)

2TallTyler: 🤷

2TallTyler (approved):

PR #11129, opened 2023-07-12
Codechange: make date-related variables constexpr

TrueBrain:
## Motivation / Problem

I noticed during #10761 that the compiler sometimes complained, as it couldn't do the math I wanted it to do compile-time. Turns out, we need to hint a bit more that we want things done compile-time when it comes to date constants.

## Description

This hints the compiler even more to result math-related questions compile-time.

## Checklist for review

### DIFF
--- a/src/date_type.h
+++ b/src/date_type.h
@@ -45,18 +45,18 @@ static const int INDUSTRY_CUT_TREE_TICKS  = INDUSTRY_PRODUCE_TICKS * 2; ///< cyc
  */
 
 /** The minimum starting year/base year of the original TTD */
-static const TimerGameCalendar::Year ORIGINAL_BASE_YEAR = 1920;
+static constexpr TimerGameCalendar::Year ORIGINAL_BASE_YEAR = 1920;
 /** The original ending year */
-static const TimerGameCalendar::Year ORIGINAL_END_YEAR  = 2051;
+static constexpr TimerGameCalendar::Year ORIGINAL_END_YEAR = 2051;
 /** The maximum year of the original TTD */
-static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
+static constexpr TimerGameCalendar::Year ORIGINAL_MAX_YEAR = 2090;
 
 /**
  * Calculate the date of the first day of a given year.
  * @param year the year to get the first day of.
  * @return the date.
  */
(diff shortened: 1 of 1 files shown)


src/date_type.h
```diff
@@ -45,18 +45,18 @@ static const int INDUSTRY_CUT_TREE_TICKS  = INDUSTRY_PRODUCE_TICKS * 2; ///< cyc
...
-static const TimerGameCalendar::Year ORIGINAL_MAX_YEAR  = 2090;
+static constexpr TimerGameCalendar::Year ORIGINAL_MAX_YEAR = 2090;
 
 /**
  * Calculate the date of the first day of a given year.
  * @param year the year to get the first day of.
  * @return the date.
  */
-static inline TimerGameCalendar::Date DateAtStartOfYear(TimerGameCalendar::Year year)
+static constexpr TimerGameCalendar::Date DateAtStartOfYear(TimerGameCalendar::Year year)
```
2TallTyler, line 59: Can this really be a constexpr? It takes a parameter. (I may be not understanding what a constexpr does 🙂)

TrueBrain: It hints to the compiler to resolve things compile-time. In this case we have a few lines lower:

`DAYS_TILL_ORIGINAL_BASE_YEAR = DateAtStartOfYear(ORIGINAL_BASE_YEAR);`

Which, before this PR, was not resolved compile-time (but might, with optimizations). After this PR, it is always. And for places that use `DateAtStartOfYear` that cannot be done compile-time, that is okay too, and it will be done runtime.

I ran into this with making things strong-typed, as I was making some other things `constexpr` too. There it was yelling at me that `DAYS_TILL_ORIGINAL_BASE_YEAR` wasn't a constexpr, but it really can be (I mean, all information is there compile-time). Those things that mostly become important if you start doing `if constexpr ()` stuff, where you assume all stuff is resolved compile-time :)

In newer C++ there is also `consteval`. This forced the compiler to error-out if it cannot be resolved compile-time. This function cannot be `consteval`. A lot of the variables can :) But, it is in C++20, so we are not using that :D

I think I explained this correctly .. but honestly, it is just a bit weird ... `const` vs `constexpr` was already weird .. but with `consteval`, it became even messier :D

2TallTyler: Thanks for the detailed explanation! 😄

2TallTyler (approved):

PR #11126, opened 2023-07-10
Fix: Build road/tram stop windows did not set WindowDesc::ini_key

JGRennison:
## Motivation / Problem

#10144 added a save/restore default window size button to the road/tram build station stop windows, but did not set WindowDesc::ini_key.
Saved default window sizes were therefore not persisted in the configuration file.

## Description

Set WindowDesc::ini_key for the above WindowDesc instances, such that these are persisted in the configuration file.

## Limitations

N/A

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11131, opened 2023-07-12
Add: workflow and script for checking ini_key issues in WindowDesc entries

glx22:
## Motivation / Problem
WindowDesc::ini_key is used for saving prefered sticky state and size.
As #11126 showed, it's easy to forget to set WindowDesc::ini_key.
It's also easy to set it while not needed, and also to use duplicate ini_key values.

## Description
Make sure we catch potential issues before merge.
For each WindowDesc, the script gets name, ini_key (if any) and the nested widget array.
It validates uniqueness of ini_key value, then it checks if the nested widget array contains `WWT_DEFSIZEBOX` or `WWT_STICKYBOX` and validates ini_key is set only when required.

## Limitations
~~Current source has issues and the script will find them. So they need to be fixed first.~~

## Checklist for review

### DIFF
--- a/.github/windowdesc-ini-key.py
+++ b/.github/windowdesc-ini-key.py
@@ -1,0 +1,50 @@
+"""
+Script to scan the OpenTTD source-tree for ini_key issues in WindowDesc entries.
+"""
+
+import glob
+import os
+import re
+import sys
+
+
+def scan_source_files(path, ini_keys=None):
+    if ini_keys is None:
+        ini_keys = set()
+
+    errors = []
+
+    for new_path in glob.glob(f"{path}/*.cpp"):
+        if os.path.isdir(new_path):
+            errors.append(scan_source_files(new_path, ini_keys))
+            continue
+
+        with open(new_path) as fp:
+            output = fp.read()
+
+        for (name, ini_key, widgets) in re.findall(r"^static WindowDesc ([a-zA-Z0-9_]*).*?, (?:\"(.*?)\")?.*?,(?:\s+(.*?),){6}", output, re.S|re.M):
+            if ini_key:
+                if ini_key in ini_keys:
+                    errors.append(f"{new_path}: {name} ini_key is a duplicate")
+                ini_keys.add(ini_key)
+            widget = re.findall("static const (?:struct )?NWidgetPart " + widgets + ".*?(?:WWT_(DEFSIZE|STICKY)BOX.*?)?;", output, re.S)[0]
+            if widget and not ini_key:
+                errors.append(f"{new_path}: {name} has WWT_DEFSIZEBOX/WWT_STICKYBOX without ini_key")
+            if ini_key and not widget:
+                errors.append(f"{new_path}: {name} has ini_key without WWT_DEFSIZEBOX/WWT_STICKYBOX")
+
+    return errors
+
+
+def main():
+    errors = scan_source_files("src")
+
+    if errors:
+        print("\n".join(errors))
+        sys.exit(1)
+
+    print("OK")
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/workflows/windowdesc-ini-key.yml
+++ b/.github/workflows/windowdesc-ini-key.yml
@@ -1,0 +1,24 @@
+name: WindowDesc ini_key
+
+on:
+  pull_request:
+    branches:
+    - master
+
+concurrency:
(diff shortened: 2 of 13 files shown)


glx22:
Before and after merge of #11126
image
image

.github/windowdesc-ini_key.py
```diff
@@ -0,0 +1,48 @@
...
+def scan_source_files(path, errors, ini_keys):
+    for new_path in glob.glob(f"{path}/*.cpp"):
+        if os.path.isdir(new_path):
+            errors.append(scan_source_files(new_path, errors, ini_keys))
+            continue
+
+        with open(new_path) as fp:
+            output = fp.read()
+
+        for windowdesc in re.findall(r"^static WindowDesc ([a-zA-Z0-9_]*).*?, (?:\"(.*?)\")?.*?,(?:\s+(.*?),){6}", output, re.S|re.M):
```
TrueBrain: `for (name, ini_key, widgets) ..` maybe?

.github/windowdesc-ini_key.py
```diff
@@ -0,0 +1,48 @@
+"""
+Script to scan the OpenTTD source-tree for ini_key issues in WindowDesc entries.
+"""
+
+import glob
+import os
+import re
+import sys
+
+def scan_source_files(path, errors, ini_keys):
```
TrueBrain: What you mostly see here, is that `ini_keys=None`, and the recursive call gives `ini_keys`. And then a line in this function:

```
if ini_keys is None:
  ini_keys = set()
```

This way, the outside doesn't know the inside of this function.

.github/windowdesc-ini_key.py
```diff
@@ -0,0 +1,48 @@
...
+                errors.append(f"{new_path}: {name} has WWT_DEFSIZEBOX/WWT_STICKYBOX without ini_key")
+            if ini_key and not widget:
+                errors.append(f"{new_path}: {name} has ini_key without WWT_DEFSIZEBOX/WWT_STICKYBOX")
+
+
+def main():
+    errors = []
+    ini_keys = set()
+
+    scan_source_files("src", errors, ini_keys)
```
TrueBrain: better to create `errors` inside the function, and return it. That is a lot cleaner Python-wise.

Don't worry about the "copies" you make for the recursive calls; it is really not worth mentioning in this context :)

.github/workflows/windowdesc_ini_key.yml
```diff
@@ -0,0 +1,22 @@
+name: WindowDesc ini_key
+
+on:
+  pull_request:
```
TrueBrain: missing `branches` part, as we would only want this to run when `master` is targetted (I would think)

.github/workflows/windowdesc_ini_key.yml
```diff
@@ -0,0 +1,22 @@
...
+    name: WindowDesc ini_key issues
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Check for ini_key issues in WindowDesc entries
+      run: |
+        set -ex
```
TrueBrain: Please use `shell: bash` instead of any `set` inside a shell; let GitHub take care of the complexity called "bash" :D

.github/windowdesc-ini_key.py
```diff
@@ -0,0 +1,48 @@
+"""
+Script to scan the OpenTTD source-tree for ini_key issues in WindowDesc entries.
```
TrueBrain: this filename is a bit confusing, especially in a Python context. Either use `_` or `-`, but mixing them both seems off. Python prefers `_`, as you cannot include a `-` file :)

TrueBrain (commented):
Minor nitpicking on the Python; sorry, occupational hazard :)

.github/windowdesc-ini_key.py
glx22: Other scripts use `-`, so I think I'll follow them.

.github/workflows/windowdesc_ini_key.yml
glx22: I copied other workflows, so something to fix in them too.

glx22 (commented):
Will rework python functions.

TrueBrain (approved):

PR #11133, opened 2023-07-13
Change: Don't save industry history if cargo slot isn't used.

PeterN:
## Motivation / Problem

Industries have 16 slots which may produce cargo, and we save two months of production/transported history for each slot for each industry.

In many cases (dependent on NewGRFs), 14 of the 16 slots won't actually ever be used as cargo slot is CT_INVALID.

## Description

Skip saving history if the slot is CT_INVALID by setting the saved length to 0.

This avoids saving history of 16 slots per industry when in many cases (NewGRF dependent) only a couple are used.

Does not need a saveload bump as the loading code is already told how much data there is.

## Limitations

As savegames are compressed, this doesn't particularly save much space... on a 4096x4096 map with high industries the resultant file size difference is negli:ble.

Vanilla: 24,766,936 bytes
Patched: 24,760,648 bytes

This may be of more benefit with #10541, but... no idea.

## Checklist for review

### DIFF
--- a/src/saveload/industry_sl.cpp
+++ b/src/saveload/industry_sl.cpp
@@ -67,6 +67,12 @@ public:
 
 	void Save(Industry::ProducedCargo *p) const override
 	{
+		if (!IsValidCargoID(p->cargo)) {
+			/* Don't save any history if cargo slot isn't used. */
+			SlSetStructListLength(0);
+			return;
+		}
+
 		SlSetStructListLength(p->history.size());
 
 		for (auto &h : p->history) {


PeterN:
Oops, this should use IsValidCargoId() :-)

rubidium42 (approved):

ldpl:
Did a bit of testing, seems to be a slight improvement overall. Though uncompressed size doesn't add up, I'd expect about 32 * 2 * 14 = 896 bytes per industry reduction based on the description of this PR, but got 119 instead.
Screenshot from 2023-07-14 15-00-21

Timing patch:
[savetiming.txt](https://github.com/OpenTTD/OpenTTD/files/12049735/savetiming.txt)
Command: `./openttd -G 1 -g -v null:ticks=55000 && ls -lsa ~/.openttd/save/autosave/exit.sav`

PR #11136, opened 2023-07-14
Change: Allow DEL key to close windows when focused on a textbox

Kuhnovic:
## Motivation / Problem

The feature to close all windows using the DEL key is great. I found that it's especially helpful for new players. If they get a bit overwhelmed by the amount of menu's then you can panic-press DEL to get back to the main viewport.

However, when you are focused on a textbox, the keypress event of the DEL key is "stolen" and the hotkey corresponding to DEL does not work. It makes sense because deleting characters is more important, but it does make the user experience a bit less consistent.

## Description

After my changes the DEL will delete a character if this is possible, i.e. if there is at least one character after the caret. If not the keypress action "passes through" the function which triggers the hotkey associated with the DEL key (close all windows by default).

**I am aware that the hotkey can be changed to a key that doesn't get swallowed by the textbox. But I'm used to the DEL key now and I am stubborn.**

## Limitations

- Pressing and holding DEL will cause the entire string to be deleted one character at a time, and then the window suddenly closes.
- Right now this only works for DEL and BACKSPACE. For consistency we probably want something similar for other keys. First I want to see what the general opinion on this is.

## Checklist for review

### DIFF
--- a/src/textbuf.cpp
+++ b/src/textbuf.cpp
@@ -476,6 +476,7 @@ HandleKeyPressResult Textbuf::HandleKeyPress(WChar key, uint16 keycode)
 		case WKC_BACKSPACE: case WKC_DELETE:
 		case WKC_CTRL | WKC_BACKSPACE: case WKC_CTRL | WKC_DELETE:
 			edited = this->DeleteChar(keycode);
+			if (!edited) return HKPR_NOT_HANDLED;
 			break;
 
 		case WKC_LEFT: case WKC_RIGHT: case WKC_END: case WKC_HOME:


rubidium42:
I'm not so sure whether the window closing when pressing DEL one time too many is a better user experience than allowing DEL to (eventually) close all windows. After all, they might need to be pressing DEL hysterically before it closes all windows.

Having said that, just DEL for closing windows might have been a bad choice... but then both DEL and Ctrl-DEL are taken  by the textbox. Shift-DEL does work, but then I would expect Ctrl-DEL for the stronger behaviour given Ctrl is usually used for stronger/alternative behaviour in other parts of the game. Though that's is already taken by deleting all messages. And too bad that Alt is not really useful since that gets taken by some window managers, so doesn't even arrive into OpenTTD, especially when clicking.

PeterN:
Changing the behaviour of pressing DEL depending one what's in the textbox is a bad idea.

2TallTyler:
I agree with the above. I like the effort to improve the user experience, but we have enough inconsistency already. 😉 

We have a lot of stale PRs, so I am closing this one. I hope there are no hard feelings. If someone feels strongly in support of this PR, feel free to re-open and we can continue debating. 🙂

PR #11140, opened 2023-07-16
Fix #10983: [AdminPort] Correct order of messages

### LINKED ISSUE #10983: [Bug]: Openttd's admin port sends Update event first after creation of the company if it posses the same ID as deleted company
### Version of OpenTTD

13.2 (also happened in previous version but I do not remember my previous version :( )

### Expected result

I expect that after creating company OpenTTD AdminPort's is going to send `ADMIN_PACKET_SERVER_COMPANY_INFO` first and `ADMIN_PACKET_SERVER_COMPANY_UPDATE` afterwards to inform AdminPort's client about updates.

### Actual result

Openttd is sending `ADMIN_PACKET_SERVER_COMPANY_UPDATE` before `ADMIN_PACKET_SERVER_COMPANY_INFO` if you create company with Id that was already used in the past by removed company. 

If I create company which Id was not used before then `ADMIN_PACKET_SERVER_COMPANY_INFO` is send first.

I have a feeling that not all information about company are removed from memory after company removal.

### Steps to reproduce
1. Send a ADMIN_PACKET_ADMIN_UPDATE_FREQUENCY message with `ADMIN_UPDATE_COMPANY_INFO` having value `ADMIN_FREQUENCY_AUTOMATIC`
2. Create a Company. Let's assume it's id is `1`
3. Execute `reset_company 1`
4. Create a company again
5. `ADMIN_PACKET_SERVER_COMPANY_UPDATE` is send first

shoter:
## Motivation / Problem

This pull request would close #10983 .
It solves problem of admin port messages arriving in un-logical order. 

## Description

The only logical order in my opinion when it comes to receiving admin port messages about company is as follows:

- ADMIN_PACKET_SERVER_COMPANY_NEW (once)
- ADMIN_PACKET_SERVER_COMPANY_INFO (once)
- ADMIN_PACKET_SERVER_COMPANY_UPDATE ( can be received many times afterwards)

Before this PR order could look like this for *newly* created company:

- ADMIN_PACKET_SERVER_COMPANY_UPDATE (0-2 times)
- ADMIN_PACKET_SERVER_COMPANY_INFO (once)
- ADMIN_PACKET_SERVER_COMPANY_NEW (once)
- ADMIN_PACKET_SERVER_COMPANY_UPDATE (many)

In my opinion that order does not make any sense at all. This PR solves it.

Additionally second parameter to `NetworkAdminCompanyInfo` function was removed due to it always being `true`. We are always creating new company when sending company info.

## Limitations

From my personal tests it seems that problem is solved.
Additionally I think that ADMIN_PACKET_SERVER_COMPANY_NEW  and ADMIN_PACKET_SERVER_COMPANY_INFO  should be merged. They are always sent together and `NEW` packet does not provide any other information that `INFO` would not provide.

Order of messages for admin port is changed. It is logical change in my opinion, however this change might have unforseen consequences to any admin port client which expects unlogical messages order. Maybe upping `NETWORK_GAME_ADMIN_VERSION` to 5 would be a good idea?

## Checklist for review

I messed my branch and I needed to recreate PR #11092

### DIFF
--- a/src/network/network_admin.h
+++ b/src/network/network_admin.h
@@ -103,7 +103,7 @@ void NetworkAdminClientInfo(const NetworkClientSocket *cs, bool new_client = fal
 void NetworkAdminClientUpdate(const NetworkClientInfo *ci);
 void NetworkAdminClientQuit(ClientID client_id);
 void NetworkAdminClientError(ClientID client_id, NetworkErrorCode error_code);
-void NetworkAdminCompanyInfo(const Company *company, bool new_company);
+void NetworkAdminCompanyNew(const Company *company);
 void NetworkAdminCompanyUpdate(const Company *company);
 void NetworkAdminCompanyRemove(CompanyID company_id, AdminCompanyRemoveReason bcrr);
 
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -888,6 +888,11 @@ CommandCost CmdCompanyCtrl(DoCommandFlag flags, CompanyCtrlAction cca, CompanyID
 				break;
 			}
 
+			/* Send new companies, before potentially setting the password. Otherwise,
+			 * the password update could be sent when the company is not yet known. */
+			NetworkAdminCompanyNew(c);
+			NetworkServerNewCompany(c, ci);
+
 			/* This is the client (or non-dedicated server) who wants a new company */
 			if (client_id == _network_own_client_id) {
 				assert(_local_company == COMPANY_SPECTATOR);
@@ -906,8 +911,6 @@ CommandCost CmdCompanyCtrl(DoCommandFlag flags, CompanyCtrlAction cca, CompanyID
 
 				MarkWholeScreenDirty();
 			}
-
-			NetworkServerNewCompany(c, ci);
 			break;
 		}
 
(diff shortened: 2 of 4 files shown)


src/network/network_admin.h
```diff
@@ -103,7 +103,7 @@ void NetworkAdminClientInfo(const NetworkClientSocket *cs, bool new_client = fal
 void NetworkAdminClientUpdate(const NetworkClientInfo *ci);
 void NetworkAdminClientQuit(ClientID client_id);
 void NetworkAdminClientError(ClientID client_id, NetworkErrorCode error_code);
-void NetworkAdminCompanyInfo(const Company *company, bool new_company);
+void NetworkAdminCompanyInfo(const Company *company);
```
rubidium42:
```suggestion
void NetworkAdminCompanyNew(const Company *company);
```
Renaming this function will make it much clearer in the places where it is called what the intent is: notifying the admins that a new company was started.

rubidium42 (commented):
I agree that the new order seems a lot more logical.

I doubt that a different order is going to cause a lot of problems. Especially since they are in a more logical order. However, the current version is 3, so that would mean bumping it to 4. On the other hand, I wonder whether many of the libraries already support version 3, which got introduced about three months ago. So if it's a problem, it's probably quite easy for them to just assume the correct order for version 3 given there aren't many servers on the main development branch.
Given the change causing version 3 won't be backported, this must not be backported either... or at least it must not be if the order of the packets is deemed to be important for versioning of the packets.

I agree with the sentiment that having a `NEW` and `INFO` packet that are always sent together is not very useful. However, what should we then do? Moving the content of `INFO` to `NEW` and deprecate `INFO` (the packet ID won't be removed), but that means moving logic around in the receivers. We could only send the `INFO` packet and rename it in our code to `NEW`, though that means there will be no original `NEW` which something could have depended on.

Conceptually the original location for sending the `NEW` and `INFO` packet is more or less the right location. However, it should not be sending out the updates before. Now you're going to have an `INFO` with some intermediate president name, and potentially some other not-yet-up-to-date data. So if a script only used the `INFO` that would be wrong. Though achieving not sending the updates before the `INFO` is sent, is probably way more complex than this relatively simple solution.

rubidium42 (approved):

PR #11141, opened 2023-07-16
Add: [Script] Game script control of industry production level.

michicc:
## Motivation / Problem

Game scripts can override industry production in a limited way. They can prohibit production increases and/or decreases and prevent closure. But they can't actually change production, just hold it in place.

This half-way control isn't very useful if a GS wants to implement custom industry production rules.

## Description

This PR adds GS API methods to get and set the production multiplier of an industry. NewGRFs are notified by a new flag in var 47 (GameScript control status). Setting the production multiplier does not prohibit industry closure, GS can use the existing control flag for this.

A script can optionally let OTTD generate a production change news message if needed.

## Limitations

I'm not sure about the actual GS API. Maybe it would be better to use a magic prod_level value to restore the default production behaviour instead of the need to clear the control flag.

## Checklist for review

### DIFF
--- a/src/command_type.h
+++ b/src/command_type.h
@@ -246,6 +246,7 @@ enum Commands : uint16_t {
 	CMD_INDUSTRY_SET_FLAGS,           ///< change industry control flags
 	CMD_INDUSTRY_SET_EXCLUSIVITY,     ///< change industry exclusive consumer/supplier
 	CMD_INDUSTRY_SET_TEXT,            ///< change additional text for the industry
+	CMD_INDUSTRY_SET_PRODUCTION,      ///< change industry production
 
 	CMD_SET_COMPANY_MANAGER_FACE,     ///< set the manager's face of the company
 	CMD_SET_COMPANY_COLOUR,           ///< set the colour of the company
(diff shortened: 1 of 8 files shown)


andythenorth:
Works for me in initial test.

andythenorth:
Tested the version with news text option.  Works, but the default news texts aren't going to be appropriate in most contexts 😄 
image

andythenorth:
With the news flag set to `false`, seems a news message is triggered appropriately by the game without intervention?

`                GSIndustry.SetProductionLevel(industry, GSIndustry.GetProductionLevel(industry) + 8, false);
                // we issue the news message ourselves as of Jul 2023, this may or may not be needed depending on https://github.com/OpenTTD/OpenTTD/pull/11141
                GSNews.Create(GSNews.NT_GENERAL, "Cabbage", GSCompany.COMPANY_INVALID, GSNews.NR_INDUSTRY, industry);
`

image

michicc:
Now with a custom news message option.

2TallTyler (commented):
I agree, the docstring should explain more explicitly what argument a GS author should pass if no news message is desired.

michicc:
So something like `If set to true and the production changed, generate a production change news message. If set to false, no news message is shown.`?

> If `show_news` is `false`, `null` (or similar) must be passed as the parameter for `custom_news`.

If `show_news` is false, any custom news text is simply ignored, so it can be null or some text. Unless you found some actual code bug where a message is still displayed.

andythenorth:
> If `show_news` is false, any custom news text is simply ignored, so it can be null or some text. Unless you found some actual code bug where a message is still displayed.
I verified that doesn't happen, `show_news=false` works as expected :)

2TallTyler (approved):

PR #11142, opened 2023-07-16
Feature: Setting to automatically restart server based on hours played

2TallTyler:
## Motivation / Problem

We have a setting for network games to automatically restart upon reaching a certain year. That will never happen if calendar time is frozen.

## Description

Add a new setting for a timer that counts down hours played, and resets after that.

## Limitations

This was originally a change before @TrueBrain suggested we have both, so blame him for more code to maintain. 😉 

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -33,6 +33,7 @@
 #include "../timer/timer.h"
 #include "../timer/timer_game_calendar.h"
 #include "../timer/timer_game_economy.h"
+#include "../timer/timer_game_realtime.h"
 #include <mutex>
 #include <condition_variable>
 
@@ -1491,29 +1492,6 @@ void NetworkUpdateClientInfo(ClientID client_id)
 	NetworkAdminClientUpdate(ci);
 }
 
-/** Check if we want to restart the map */
-static void NetworkCheckRestartMap()
-{
-	if (_settings_client.network.restart_game_year != 0 && TimerGameCalendar::year >= _settings_client.network.restart_game_year) {
-		Debug(net, 3, "Auto-restarting map: year {} reached", TimerGameCalendar::year);
-
-		_settings_newgame.game_creation.generation_seed = GENERATE_NEW_SEED;
-		switch(_file_to_saveload.abstract_ftype) {
-			case FT_SAVEGAME:
-			case FT_SCENARIO:
-				_switch_mode = SM_LOAD_GAME;
-				break;
-
-			case FT_HEIGHTMAP:
-				_switch_mode = SM_START_HEIGHTMAP;
-				break;
-
-			default:
-				_switch_mode = SM_NEWGAME;
-		}
-	}
-}
-
 /** Check if the server has autoclean_companies activated
  * Two things happen:
  *     1) If a company is not protected, it is closed after 1 year (for example)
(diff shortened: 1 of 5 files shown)


TrueBrain:
I like this PR, but not in the way you might think. Btw, for settings conversion, see #11143. That is, if that gets accepted. I tried to work out there how to do that.

Anyway, for this PR:

I think it is the wrong approach, but accidentally shows a good improvement. Let me explain.

I think we have two types of users: 
1) users that just want to restart the map every NN hours, like this PR does. Like: every 24h it is reset. That sounds like a total awesome things to do.
2) users that want to play till in-game year 2050, because there the scoreboard shows up. After that, the server restarts (or a few days after, or what-ever).

The first is more for long-term servers like reddit etc. I totally see this PR being of added value to them.
The second is more for competitive play, where people start about on the same time. This PR would make that gametype more difficult.

So ... and I know this is a big ask .. shouldn't we have both settings? With day-length patch, the setting of this PR is going to make much more sense for that first group. But there will always be people in that second group .. and they will have to understand that if you freeze the calendar, the setting will never work. But this new one does. Best of both worlds?

What do you think?

2TallTyler:
Sure, that makes sense. This becomes a `Feature:` PR, to add new functionality for when frozen calendar time makes the existing setting useless.

I think I'll also change the time units to hours. Minutes is probably too much precision, given than an hour is only five years in-game. On the other hand, it doesn't really hurt to use minutes, except that users desiring a longer time have to do some math. Anyone feel strongly that it should stay minutes?

2TallTyler:
A relevant discussion from the forums, where a poster wants to increase the autoclean interval beyond the current limit. This PR would let them do that.
https://www.tt-forums.net/viewtopic.php?p=1264059

src/network/network_server.cpp
```diff
@@ -1813,12 +1791,63 @@ void NetworkServer_Tick(bool send_frame)
...
+static void NetworkCheckRestartMapYear()
+{
+	/* If setting is 0, this feature is disabled. */
+	if (_settings_client.network.restart_game_year == 0) return;
+
+	if (TimerGameCalendar::year >= _settings_client.network.restart_game_year) {
+		Debug(net, 3, "Auto-restarting map: year {} reached", TimerGameCalendar::year);
+		RestartMapHelper();
+	}
+}
```
TrueBrain, line 1844: Please move this function to just above the actual timer. Locality kinda helps in this case :)

src/network/network_server.cpp
```diff
@@ -1813,12 +1791,63 @@ void NetworkServer_Tick(bool send_frame)
 	}
 }
 
+/** Helper function to restart the map. */
+static void RestartMapHelper()
```
TrueBrain: `Helper` feels a bit awkward .. why not just call it for what it is? `NetworkRestartMap` or something?

src/settings_type.h
```diff
@@ -318,6 +318,7 @@ struct NetworkSettings {
 	uint8_t       max_companies;                            ///< maximum amount of companies
 	uint8_t       max_clients;                              ///< maximum amount of clients
 	TimerGameCalendar::Year restart_game_year;            ///< year the server restarts
+	uint16_t      restart_hours;                            ///< number of hours to run the server before automatic restart
```
TrueBrain: Why not `std::chrono::hours` as type?

src/network/network_server.cpp
```diff
@@ -1813,12 +1791,63 @@ void NetworkServer_Tick(bool send_frame)
...
+	Debug(net, 3, "Auto-restarting map: {} hours played", _settings_client.network.restart_hours);
+	RestartMapHelper();
+});
+
+/** Reset the automatic network restart time interval. */
+void ChangeNetworkRestartTime()
+{
+	if (!_network_server) return;
+
+	_network_restart_map_timer.SetInterval({ std::chrono::hours(_settings_client.network.restart_hours), TimerGameRealtime::UNPAUSED }, true);
```
TrueBrain: The part I am doubting, is the `true` part. This means if you change the setting while a game is running, the counter resets. I am not sure that is the expected behaviour. But as you reuse this function to reset the timer on game-start, it is kinda tricky.

Maybe send the `bool reset` as parameter, and reset it when you start a new game, but not when you change the setting. And yes, that can mean if you change the setting the game all of a sudden resets :P This is where I am not sure if my idea works. But if we document the setting as "time since start", that would kinda make sense.

Now I think about it, the `year` based setting has the same. When you change it just before year change, it will directly reset. So I guess indeed, not resetting is the better thing to do when changing the setting. I think :P

src/network/network_server.cpp
```diff
@@ -1813,12 +1791,63 @@ void NetworkServer_Tick(bool send_frame)
...
+	}
+}
+
+/** Timer to restart a network server automatically based on real-time hours played. Initialized at zero to disable until settings are loaded. */
+static IntervalTimer<TimerGameRealtime> _network_restart_map_timer({std::chrono::milliseconds::zero(), TimerGameRealtime::UNPAUSED}, [](auto)
+{
+	if (!_network_server) return;
+
+	/* If setting is 0, this feature is disabled. */
+	if (_settings_client.network.restart_hours == 0) return;
```
TrueBrain, line 1817: Strictly seen you don't need this, as an interval of zero means this callback never triggers. It also doesn't hurt.

TrueBrain (commented):
Overall this looks nice; just a few bits and pieces :)

TrueBrain:
> A relevant discussion from the forums, where a poster wants to increase the autoclean interval beyond the current limit. This PR would let them do that. https://www.tt-forums.net/viewtopic.php?p=1264059

I don't see any changes to anything autoclean related? What am I missing?

TrueBrain:
Owh, and one more: title says `dedicated server`, but the code says it works for just `server`. Do we want to emphasis `dedicated` here, or just omit it?

src/settings_type.h
2TallTyler: Tried this, got a `'parse': is not a member of 'fmt::v10::formatter<U,Char,void>` error on line 2555 of	`src\3rdparty\fmt\core.h`. 😕

2TallTyler (commented):
> I don't see any changes to anything autoclean related? What am I missing?

You're missing my lack of reading comprehension. Indeed, this has nothing to do with autoclean. 🙄 

>Owh, and one more: title says `dedicated server`, but the code says it works for just `server`. Do we want to emphasis `dedicated` here, or just omit it?

...Hmm, I guess I don't know what difference there is between the two. I don't intend any emphasis, and have only tested on a standard client hosting a game, so I've amended the title.

src/table/settings/network_settings.ini
```diff
@@ -241,6 +242,16 @@ min      = CalendarTime::MIN_YEAR
...
 
+[SDTC_VAR]
+var      = network.restart_hours
+type     = SLE_UINT16
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_0_IS_SPECIAL | SF_NETWORK_ONLY
+def      = 0
+min      = 0
+max      = UINT16_MAX
+interval = 1
+post_cb  = [](auto) {ChangeNetworkRestartTime(false); }
```
glx22:
```suggestion
post_cb  = [](auto) { ChangeNetworkRestartTime(false); }
```

src/settings_type.h
TrueBrain: Yes ofc you silly goose. There is a bit more work that needs to be done for this :D But only stating what line in fmt fails is not useful; you need to check what line causes that :)

What I am guessing, is that you see the error from `Debug()`. You have to tell him how to print time. There are several ways, the easiest is to use `count()`.

Either way, our settings-code can't handle `std::chrono`, so let this PR not be the first to introduce that. Bit sad, but it is what it is.

TrueBrain (approved):

TrueBrain:
The only thing I wonder about, but I think it is fine:

- The calendar based restart works even on a loaded savegame.
- This restart counts from the moment you start.

An alternative would be to hook it onto the economy timer, and use that for restart. That way it would be a restart after N hours of play.

In reality, I do not think it matters at all, as I suspect only servers use this, where the owner just wants a new game every 23 hours. And they never load in a map, and expect the restart to continue counting.

2TallTyler:
I agree that servers are probably the only users of this, since it is only accessible from the command line. And once they do restart using this function, the timer starts from 0 as expected and it's a new game, not a savegame.

PR #11143, opened 2023-07-17
Fix: store autosave settings under the new names

TrueBrain:
## Motivation / Problem

While looking for a way to allow aliases for `OMANY`, I noticed autosave was storing the wrong settings values. And that we have a way to allow aliases.

## Description

Instead of storing the old `monthly`, `yearly`, etc, store it in minutes.

This keeps the original setting in tact, and renames the new one. This allows switching back between old and new versions without annoying warnings.

Based on comments and feedback, this PR grew a bit bigger to supply a framework for future conversions too. Also makes it a bit more readable :)

## Limitations

The GUI doesn't allow for anything between the predefined settings. This means if the user changes the value in the config to something other than 0/10/30/60/120, it will work, till the Game Options are opened again.

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -161,6 +161,8 @@ enum IniFileVersion : uint32 {
 	IFV_LINKGRAPH_SECONDS,                                 ///< 3  PR#10610 Store linkgraph update intervals in seconds instead of days.
 	IFV_NETWORK_PRIVATE_SETTINGS,                          ///< 4  PR#10762 Move no_http_content_downloads / use_relay_service to private settings.
 
+	IFV_AUTOSAVE_RENAME,                                   ///< 5  PR#11143 Renamed values of autosave to be in minutes.
+
 	IFV_MAX_VERSION,       ///< Highest possible ini-file version.
 };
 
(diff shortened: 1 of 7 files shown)


PeterN:
This will make my 'allow any value' patch a lot simpler 😍

src/settings.cpp
```diff
@@ -1262,6 +1264,27 @@ void LoadFromConfig(bool startup)
...
+						case 0: _settings_client.gui.autosave_interval = std::chrono::minutes::zero(); break;
+						case 1: _settings_client.gui.autosave_interval = std::chrono::minutes(10); break;
+						case 2: _settings_client.gui.autosave_interval = std::chrono::minutes(30); break;
+						case 3: _settings_client.gui.autosave_interval = std::chrono::minutes(60); break;
+						case 4: _settings_client.gui.autosave_interval = std::chrono::minutes(120); break;
+						default: break;
+					}
+				}
+			}
+		}
```
PeterN, line 1335: Should we check if the new setting already exists?

Not sure, but I think if you load a new (and change the value) then old then new version it will reset your `autosave_interval` to whatever `autosave` maps to?

TrueBrain: Yes and no. It depends :D

Configuration-files have a version, and this check ofc validates if you are loading a configuration with an older version. So what can happen is:

- You have a v4 config.
- You load a client with v5 support. It converts the `autosave` from v4 to v5.
- You now load a client with v4 support. It will resave the config as a v4.
- Now if you load a client with v5 support again, it will convert the `autosave` again, as that is what the version asks.

I am not sure this is a bad thing. How ever you slice it, if people use the same configuration with multiple versions, you will always have these weird edge-cases to resolve. This is the simplest way to approach it. What it means that if someone changes the autosave in the v4 config, it will always have an effect on v5.

If you would be to check if `autosave_interval` already exists, it won't convert it again. This can be good .. this can be bad.

So far, for all the other settings, we opted for not checking if the new setting exists. The setting in the older config wins from the newer.

PeterN: Problem with this is it will override the setting in the new version even if you didn't change the setting in the old version.

And I do often load an old version just to check behaviour of something else.

So I'm not sure I like the current way...

TrueBrain: Takes a bit more code, so I generalized the approach a bit. There was one other setting that did the same, so converted that. That makes this PR a bit bigger than the exact context of the PR, but helps future-us in not having this talk again to figure out what we actually want :)

PeterN (approved):

PR #11144, opened 2023-07-17
Codechange: wire "load" support into all OMANY and MMANY for settings

TrueBrain:
## Motivation / Problem

While working on #11143 I had some arguments with the code in respect to `load` and OMANY.

## Description

This PR fixes so that `load` works for all `OMANY` (and `MMANY`). Nobody actually uses it, so you can argue if this PR should be merged, but the current code was silly.

While at it, also reworked the logic slightly, that it first tries to use the lookup, next the conversion, and only after that check if it is a digit. The original code did the digit first, which meant you can't do a lookup starting with a number.

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -171,18 +171,23 @@ const uint16 INIFILE_VERSION = (IniFileVersion)(IFV_MAX_VERSION - 1); ///< Curre
  * @param str the current value of the setting for which a value needs found
  * @param len length of the string
  * @param many full domain of values the ONEofMANY setting can have
+ * @param many_cnvt callback procedure when loading value mechanism fails
  * @return the integer index of the full-list, or -1 if not found
  */
-size_t OneOfManySettingDesc::ParseSingleValue(const char *str, size_t len, const std::vector<std::string> &many)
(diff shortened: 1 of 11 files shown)


src/settings.cpp
```diff
@@ -171,19 +171,24 @@ const uint16 INIFILE_VERSION = (IniFileVersion)(IFV_MAX_VERSION - 1); ///< Curre
...
-
 	size_t idx = 0;
 	for (auto one : many) {
 		if (one.size() == len && strncmp(one.c_str(), str, len) == 0) return idx;
 		idx++;
 	}
 
+	/* We failed our lookup; try the conversion function next. */
+	idx = many_cnvt(str);
+	if (idx != (size_t)-1) return idx;
```
rubidium42, line 187:
```suggestion
	if (many_cnvt != nullptr) {
		idx = many_cnvt(str);
		if (idx != (size_t)-1) return idx;
	}
```
As far as I can see `many_cnvt` can be `nullptr`.

TrueBrain:
The more I look at this PR, the less I think we should do it. The new stuff isn't used, and one could argue we shouldn't be using this anymore, in favour of renaming settings to remain backwards (and forwards) compatible between versions. So I am going to close this up for now!

PR #11153, opened 2023-07-25
Fix: crash when not passing parameter for -n

rubidium42:
## Motivation / Problem

Run `openttd -n` -> crash.

## Description

The crash happens for at least two years. However, practically this command line parameter has been broken since its introduction. It proclaimed to have an optional parameter, but later checking for it to be `nullptr` or not. Actually, in the begin it also set a boolean, which it checked first before checking for `nullptr` at the later location. In other words, even though it said the parameter is optional, it has never been optional if you wanted to get into that functionality.

So, I made the parameter non-optional (line 475).

Also looking at the comment/help message, there is no requirement for passing an IP-address, just a hostname is good enough. So I took the 'host' labelling from the man page.
Since the man page needed to be updated anyhow, I added the other missing parameters to it, renamed player to company (a change from a decade or more ago) and updated the date.

## Limitations

Not really. Unless you consider `openttd -n` now showing the help message and exiting as a deviation from the pre 12 behaviour. Since 12 it crashes to desktop though, so I do not think it's a real problem.

## Checklist for review

### DIFF
--- a/docs/openttd.6
+++ b/docs/openttd.6
@@ -1,6 +1,6 @@
 .\"                                      Hey, EMACS: -*- nroff -*-
 .\" Please adjust this date whenever revising the manpage.
-.Dd October 13, 2014
+.Dd August 12, 2023
 .Dt OPENTTD 6
 .Os
 .Sh NAME
@@ -8,7 +8,7 @@
 .Nd open source clone of the Microprose game \(lqTransport Tycoon Deluxe\(rq
 .Sh SYNOPSIS
 .Nm
-.Op Fl efhx
+.Op Fl efhQxX
 .Op Fl b Ar blitter
 .Op Fl c Ar config_file
 .Op Fl d Op Ar level | Ar cat Ns = Ns Ar lvl Ns Op , Ns Ar ...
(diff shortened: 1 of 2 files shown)


docs/openttd.6
```diff
@@ -141,11 +147,12 @@ see
 for a full list.
 .It Fl x
 Do not automatically save to config file on exit.
+.It Fl X
+Do not use global folders to search for files.
 .El
 .Sh SEE ALSO
 .Lk https://wiki.openttd.org "Wiki"
 (includes community maintained manual),
-.Lk https://forum.openttd.org "Forum",
```
TrueBrain: You say this link doesn't exist in your description, but that is not true. That link does (and has for many years) exactly what you expect: redirect you to our forums. If that doesn't work for you, let me know, as it really should :)

Whether to remove it or not, I don't actually care :)

TrueBrain (dismissed):

docs/openttd.6
rubidium42: Well, I'm sure that the link did not work as it should have when I checked all the links in the man page. It could have been some transient issue though; I didn't investigate much further as I didn't really care whether it still worked or not, just that there weren't broken links in the man page.

TrueBrain: Just to check: now you know what you know, you still want to leave it out? Again, I don't really care, as long as we do it for the right reasons :D

PS: the singular form as it is here, is correct. Your description mentions `forums.openttd.org`; that subdomain does not exist. Maybe there things broke down :)

TrueBrain (approved):

PR #11170, opened 2023-08-06
Feature: Count the number of towns and cities in town directory

2TallTyler:
## Motivation / Problem

There is no way to see how many towns and/or cities are on a map. The town directory only lists the total population.

I've seen this feature requested several times on Discord.

## Description

directory

List the number of towns and cities above the total population.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3510,7 +3510,9 @@ STR_TOWN_DIRECTORY_NONE                                         :{ORANGE}- None 
 STR_TOWN_DIRECTORY_TOWN                                         :{ORANGE}{TOWN}{BLACK} ({COMMA})
 STR_TOWN_DIRECTORY_CITY                                         :{ORANGE}{TOWN}{YELLOW} (City){BLACK} ({COMMA})
 STR_TOWN_DIRECTORY_LIST_TOOLTIP                                 :{BLACK}Town names - click on name to centre main view on town. Ctrl+Click opens a new viewport on town location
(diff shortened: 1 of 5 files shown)


glx22:
Code says they are counted separately so the total is `50 + 2`

Eddi-z:
What i would suggest is to add a filter for cities, and then display in the title how many items are filtered (in the example this would be "Towns (2/52)", similar things could then be done for industry lists and stuff as well.

TrueBrain:
As with others, I am not really a fan of this change. Mostly, why towns and not industries? And what value do we expect from it? Visually it is also very bloaty, for information you don't actually need a lot.

I like the suggestion to put in the caption how many items are shown. So `Towns (2 / 52)` and `Industries (4 / 109)`. But that indeed requires filtering. And without filtering, `Towns (52)`, doesn't give you the information you tried to address.

So I guess we will have to start with some kind of filter system? Dunno, also feels like a lot of work for something this simple.

Which means I go back to the motivation. In this PR the motivation is .. lacking :D "Information is not there, I added it" is not actually a motivation. So when I think about it, I would like to know what my cities are. That currently requires scrolling through all towns, and spotting them. Basically, filtering seems like the primary use-case here to me.

But I should put this in a question: why did you want this information to be displayed, other than: "because"? :D That might help me help you finding other solutions for this problem, either visually or functionality-wise.

2TallTyler:
The motivation is repeated feature requests from players, on Discord. I don't really need the feature myself, and like you say, it does seem like a lot more work to do properly. So, others can feel free to pick this up, but I'm done for now. 😃

PR #11069, opened 2023-06-25
Fix #11067, ed83c4b: Don't start competitors during map generation

glx22:
Fixes #11067.

## Motivation / Problem
When "Interval between starting of competitors" is 0, competitors are started during map generation while human company is started after map generation.

## Description
Start the competitors after map generation, and after human company startup.
I put the start in `OnTick_Companies()` because when game is just created the timer is fired which will immediately start as many AIs as needed and reset the timer for ~10 minutes. Then on timeout a new AI is started if needed and `OnTick_Companies()` will start the remaining as needed if more than one disappeared during the delay.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11185, opened 2023-08-11
Fix #11179, a979d9c: Don't start more competitors than allowed

### LINKED ISSUE #11179: [Bug]: Max competitors is ignored when AI interval is zero
### Version of OpenTTD

ba95764f1c on Windows 10, 64-bit (via Steam)

### Expected result

When "Interval between starting of competitors" is zero, "Maximum no. competitors" is respected.

### Actual result

When "Interval between starting of competitors" is zero, "Maximum no. competitors" is not respected properly and is instead treated as if it was set to the maximum number of companies in total (less one for the human player).

### Steps to reproduce

1. Set "Interval between starting of competitors" to zero.
2. Set "Maximum no. competitors" to any positive value.
3. Start a new game.
4. Fast-forward.
5. Every 10 minutes of game time, a new AI is spun up, even if that would place the total AI count over the limit set by "Maximum no. competitors."

glx22:
Fixes #11179

## Motivation / Problem
In #11069, I forgot a check (should have tested more) and that allowed to start more competitors than `max_no_competitors` when interval between starts is 0.

## Description
Respect `max_no_competitors` setting when starting as much as needed competitors with interval between starts being 0.

I also changed a variable type to remove some casts.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -600,12 +600,12 @@ TimeoutTimer<TimerGameTick> _new_competitor_timeout(0, []() {
 	if (_networking && Company::GetNumItems() >= _settings_client.network.max_companies) return;
 
 	/* count number of competitors */
-	uint n = 0;
+	uint8_t n = 0;
 	for (const Company *c : Company::Iterate()) {
 		if (c->is_ai) n++;
 	}
 
-	if (n >= (uint)_settings_game.difficulty.max_no_competitors) return;
+	if (n >= _settings_game.difficulty.max_no_competitors) return;
 
 	/* Send a command to all clients to start up a new AI.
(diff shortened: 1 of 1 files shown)


2TallTyler (commented):
Why are AI companies and companies in networking games treated differently? The gap seems to be allowing more than the limit of multiple human companies in an offline game...but that's not possible, is it?

glx22:
In network games there's a max number of companies (AI and human), while in an offline game there's only one human player (the first company).
So there are 2 limits to check before starting a new AI, first the max number of companies if network game, then the max number of AIs for network and offline.

2TallTyler (approved):

PR #11190, opened 2023-08-12
Codechange: make explicit when a TileIndex is cast to its basetype 

TrueBrain:
This PR includes #10761 as it depends on it.

## Motivation / Problem

StrongTypeDef allowed, before #10761 already, implicit conversion to `uint32_t` for `TileIndex`, because of the many many many cases this was needed. But because of #10761, the amount of places this is still happening is reduced by a lot. This is mostly because functions like `SetDParam` now accept a `StrongTypeDef`, and knows how to deal with it.

This made me wonder: can we make `TileIndex`'s conversion explicit, so we don't accidentally assign incompatible types to each other just because the compiler used an implicit conversion.

## Description

Turns out, the answer is: yes, we can. And this PR is the result of that.

`dest_tile` is misused by disasters to store an ID. Not part of this PR to ... "fix" that issue.

The `TileIndexDiff` is a bit weird; that is not normally how we calculate a `TileIndexDiff`. So changed it to what others do too.

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1177,7 +1177,7 @@ struct BuildVehicleWindow : Window {
 	{
 		this->vehicle_type = type;
 		this->listview_mode = tile == INVALID_TILE;
-		this->window_number = this->listview_mode ? (int)type : (int)tile;
+		this->window_number = this->listview_mode ? (int)type : static_cast<uint32_t>(tile);
 
 		this->sel_engine = INVALID_ENGINE;
 
@@ -1212,7 +1212,11 @@ struct BuildVehicleWindow : Window {
 
 		this->details_height = ((this->vehicle_type == VEH_TRAIN) ? 10 : 9);
 
-		this->FinishInitNested(tile == INVALID_TILE ? (int)type : (int)tile);
+		if (tile == INVALID_TILE) {
+			this->FinishInitNested(type);
+		} else {
+			this->FinishInitNested(tile);
+		}
 
 		this->querystrings[WID_BV_FILTER] = &this->vehicle_editbox;
 		this->vehicle_editbox.cancel_button = QueryString::ACTION_CLEAR;
@@ -1891,7 +1895,7 @@ void ShowBuildVehicleWindow(TileIndex tile, VehicleType type)
 	 *  so if tile == INVALID_TILE (Available XXX Window), use 'type' as unique number.
 	 *  As it always is a low value, it won't collide with any real tile
 	 *  number. */
-	uint num = (tile == INVALID_TILE) ? (int)type : (int)tile;
+	uint num = (tile == INVALID_TILE) ? (int)type : static_cast<uint32_t>(tile);
 
 	assert(IsCompanyBuildableVehicleType(type));
 
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -389,7 +389,7 @@ set_name:;
 			SetDParam(1, STR_NEWS_COMPANY_LAUNCH_DESCRIPTION);
 			SetDParamStr(2, cni->company_name);
 			SetDParam(3, t->index);
-			AddNewsItem(STR_MESSAGE_NEWS_FORMAT, NT_COMPANY_INFO, NF_COMPANY, NR_TILE, c->last_build_coordinate, NR_NONE, UINT32_MAX, cni);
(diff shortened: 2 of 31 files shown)


src/build_vehicle_gui.cpp
```diff
@@ -1212,7 +1212,7 @@ struct BuildVehicleWindow : Window {
 
 		this->details_height = ((this->vehicle_type == VEH_TRAIN) ? 10 : 9);
 
-		this->FinishInitNested(tile == INVALID_TILE ? (int)type : (int)tile);
+		this->FinishInitNested(tile == INVALID_TILE ? (int)type : tile);
```
PeterN: This looks odd, as the types of both sides are the same.

TrueBrain: Ugh, yes. We mix and matched this stuff way too much :D Lemme fix!

src/newgrf_object.cpp
```diff
@@ -305,7 +305,7 @@ static uint32_t GetCountAndDistanceOfClosestInstance(byte local_id, uint32_t grf
 	switch (variable) {
 		/* Relative position. */
 		case 0x40: {
-			uint offset = this->tile - this->obj->location.tile;
+			uint offset = static_cast<uint32_t>(this->tile - this->obj->location.tile);
```
2TallTyler: Could this work instead? TileX() and TileY() already take a TileIndex as an argument.

```suggestion
			TileIndex offset = this->tile - this->obj->location.tile;
```

src/company_cmd.cpp
```diff
@@ -389,7 +389,7 @@ set_name:;
 			SetDParam(1, STR_NEWS_COMPANY_LAUNCH_DESCRIPTION);
 			SetDParamStr(2, cni->company_name);
 			SetDParam(3, t->index);
-			AddNewsItem(STR_MESSAGE_NEWS_FORMAT, NT_COMPANY_INFO, NF_COMPANY, NR_TILE, c->last_build_coordinate, NR_NONE, UINT32_MAX, cni);
+			AddNewsItem(STR_MESSAGE_NEWS_FORMAT, NT_COMPANY_INFO, NF_COMPANY, NR_TILE, static_cast<uint32_t>(c->last_build_coordinate), NR_NONE, UINT32_MAX, cni);
```
2TallTyler, line 392: I investigated overloading this to pass a TileIndex directly, and found we already have one! Try `AddTileNewsItem()` instead? (I see this PR adds a cast there, but one less cast is one less cast!)

src/industry_cmd.cpp
```diff
@@ -151,7 +151,7 @@ Industry::~Industry()
 	for (TileIndex tile_cur : this->location) {
 		if (IsTileType(tile_cur, MP_INDUSTRY)) {
 			if (GetIndustryIndex(tile_cur) == this->index) {
-				DeleteNewGRFInspectWindow(GSF_INDUSTRYTILES, tile_cur);
+				DeleteNewGRFInspectWindow(GSF_INDUSTRYTILES, static_cast<uint32_t>(tile_cur));
```
2TallTyler, line 154: This function shows up a lot in this PR. Could it be templated to accept a TileIndex without casting each time it's called?

src/company_cmd.cpp
TrueBrain: Yes and no .. the `AddTileNewsItem` sets `NF_NO_TRANSPARENT | NF_SHADE | NF_THIN` while we want `NF_COMPANY`. So you can't actually use that function without loss of functionality :( And I didn't want to make more changes than needed in this PR.

src/industry_cmd.cpp
TrueBrain: It is a balance. I didn't want to silently case the strong-typed values for the newgrf inspect window, as honestly, that felt a bit wrong in these cases. It depends on the enum what the type of the second thing is, and that deserves its own proper solution. But not for this PR.

Similar can be argued for the window-related functions, but there were just too many of them to  do it like that.

So yeah, meh. I don't want to sparkles templated functions everywhere; for SetDParam and window_number I can see the argument. Here .. I was much more hesitant. So that is why I went for this approach. I rather not add templates functions to just cast away the strong typing when possible. "Work by exception" approach :)

Let me know if you disagree. It is an opinion, not a black/white thing :D

2TallTyler: I don't know if I'm qualified to have an opinion, but looking through code with casts everywhere can be confusing. And we know the timeline of "we'll give it a proper solution in another PR someday." 😛 

I wonder if a function overload would clean up the function calls to cast in one central place, or if that's just half-assing the eventual correct solution.

In any case, none of it bothers me enough to prevent an approval. If you decide to change it, I'll re-approve.

2TallTyler (approved):

PR #11191, opened 2023-08-12
Codechange: use std::variant instead of using bitflags in the value

TrueBrain:
## Motivation / Problem

While working on #11190 I found that `InitializeViewport` used some weird tactics to support a Vehicle viewport and a Tile viewport. Let's modernize it a bit!

## Description

Use `std::variant` to know whether we want to see a Vehicle or a Tile. This works, as `TileIndex` is using `StrongTypeDef` and as such, is another type from `VehicleID` (which is just an integer).

## Checklist for review


PeterN (approved):
oooh shiny modern!

PR #11192, opened 2023-08-12
Codechange: be more type-specific about types in NPFs queue

TrueBrain:
## Motivation / Problem

While working on #11190 the compiler (rightfully) complained a `TileIndex` was shoved in an `uint32_t`, and it didn't like that.

## Description

Turns out, all `key1` references are in fact always a `TileIndex`, and this is assumed in several places. So be more correct, and make it in fact a `TileIndex`. `key2` is always `Trackdir`. Also rename `key1` and `key2`, as .. weird names.

And just to be clear, this queue stuff in NPF is also used by rivers. Explaining the one change in `landscape.cpp`.

## Checklist for review

### DIFF
--- a/src/pathfinder/npf/npf.cpp
+++ b/src/pathfinder/npf/npf.cpp
@@ -132,20 +132,20 @@ static uint NPFDistanceTrack(TileIndex t0, TileIndex t1)
 
 /**
  * Calculates a hash value for use in the NPF.
- * @param key1 The TileIndex of the tile to hash
- * @param key2 The Trackdir of the track on the tile.
+ * @param tile The TileIndex of the tile to hash
+ * @param dir The Trackdir of the track on the tile.
  *
  * @todo Think of a better hash.
  */
-static uint NPFHash(uint key1, uint key2)
+static uint NPFHash(TileIndex tile, Trackdir dir)
 {
 	/* TODO: think of a better hash? */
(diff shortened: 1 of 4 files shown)


PeterN:
Main question, rather than fiddling with variable types, is there a stl container that can be used instead of the custom queue?

TrueBrain:
Most likely, but that would take far more time than I am willing to spend on this atm :) NPF is old .. Rivers is unknown to me .. testing takes a lot of time. I rather focus on getting the strong typedef in atm :)

michicc (dismissed):

PeterN:
Might as well fix the type of key2 to be Trackdir as well then?

src/pathfinder/npf/npf.cpp
```diff
@@ -137,7 +137,7 @@ static uint NPFDistanceTrack(TileIndex t0, TileIndex t1)
  *
  * @todo Think of a better hash.
  */
-static uint NPFHash(uint key1, uint key2)
+static uint NPFHash(TileIndex key1, uint key2)
 {
 	/* TODO: think of a better hash? */
 	uint part1 = TileX(key1) & NPF_HASH_HALFMASK;
```
PeterN: I can't comment on the line in question, but the cast on line 147 in `assert(IsValidTile((TileIndex)key1));` is no longer needed.

PeterN (approved):

PR #11208, opened 2023-08-18
Fix: [MSVC] generate static libraries / executables correctly with vcpkg -static target

TrueBrain:
## Motivation / Problem

When building for MSVC I was often warned that `/MDd` (dynamic) was overwriting `/MT` (static), which surprised me, as we should be `/MT`. Now this was never the problem with our current libraries, and things were linked in statically anyway, but when adding `breakpad` for testing, it became apparent that we were actually compiling `/MDd`. But with the vcpkg target `x64-windows-static` I expected `/MTd`.

## Description

Since CMake 3.15 (our minimum is 3.16) there is a CMake variable https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html to indicate whether `/MD` or `/MT` should be used. Seems in more recent MSVC and CMake this is finally handled properly, and MSVC was happily compiling the binary dynamic.

By setting it correctly, the `/MTd` is correctly added instead of the `/MDd` (well, it is `-MTd` and `-MDd`, but who is counting). The best way to validate this is to make an error in a file, and check the `cl.exe` line.

In case you wonder if it is actually the intention to build static, please see https://github.com/OpenTTD/OpenTTD/blob/master/cmake/CompileFlags.cmake#L8 . We already tried .. but that no longer works for the latest CMake / MSVC combo.

## Checklist for review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,6 +247,14 @@ set_target_properties(openttd PROPERTIES OUTPUT_NAME "${BINARY_NAME}")
 if(MSVC)
     # Add DPI manifest to project; other WIN32 targets get this via ottdres.rc
     target_sources(openttd PRIVATE "${CMAKE_SOURCE_DIR}/os/windows/openttd.manifest")
+
+    # If target -static is used, switch our project to static (/MT) too.
+    # If the target ends on -static-md, it will remain dynamic (/MD).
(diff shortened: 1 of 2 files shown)


glx22 (approved):
Using the proper way is clearly better than the hacky way (especially when the hacky no longer works)

PR #11209, opened 2023-08-18
Remove: stack-traces from crash.log on all OSes

TrueBrain:
## Motivation / Problem

This might be a bit controversial, but bear with me.

For all OSes, we do an attempt to include the strack-trace in the crash.log. However, for any official release, whether it is a nightly, a beta, or a stable release, there isn't enough information available in the binary to produce any meaningful strack-trace. It is always a mention of `openttd` followed by a pointer in memory. So that isn't all that helpful.

The only place it is helpful, is during development. There it can help you quickly find where it crashed. But with #11202 I set out to create a crash.dmp on all platforms, which is a minidump with tons of extra information about the crash. Using this in combination with, for example, gdb, gives you much more information.

In short, I am of the opinion that the strack-trace in the crash.log does more harm than good, and we have tons of crash.logs on our issue-tracker to proof that. In general, we need glx to actually tell what the stack-trace was before we can look into it. (which I am also set to correct, and #11202 is the first move towards that).

## Description

As advertised, remove the code related to stack-traces in crash.log.

And yes, the windows driver had two: one for MSVC, the other for mingw.

## Limitations

On Linux it is slightly more useful than on other platforms, as Linux does retain some symbol information. So you get information like ` [05] ./openttd(_ZN10MainWindow8OnHotkeyEi+0x1f5) [0x55ab108495d5]` on release builds from it, which is not totally useless.

## Checklist for review


TrueBrain:
Enough reasons to keep at least Linux variant. Going to close this, and approach some parts in a different way.

PR #11212, opened 2023-08-18
Remove: [Win32] register values in crash.log

TrueBrain:
## Motivation / Problem

On crash, Windows prints out the current register values in the crash.log. This on its own doesn't really mean anything, and is also stored in the crash.dmp that comes with the log. Here it is easier to delve into the "why did it crash".

To unify how the crash.log looks between OSes, I had a choice: add registers to the other OSes, or remove it from Windows. But as the amount of value these registers add to debugging problems is so low, and a crash.dmp also contains all this, removing it is the easier way.

## Description

Removal of the register values.

This is part of my attempt to make crash.log and crash.dmp more useful and unified over all platforms.

## Checklist for review

### DIFF
--- a/src/crashlog.cpp
+++ b/src/crashlog.cpp
@@ -99,11 +99,6 @@ void CrashLog::LogCompiler(std::back_insert_iterator<std::string> &output_iterat
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


glx22 (approved):
Registers are indeed not very useful without the actual assembly code.

PR #11221, opened 2023-08-20
Codechange: Break up date_type.h into timer classes

2TallTyler:
## Motivation / Problem

As of #10761, dates and years are strongly-typed. #10700 will split time into calendar and economy dates, but lots of useful functions like `DateToYear()` and constants like `DAYS_IN_YEAR`, `MAX_DATE`, etc., will need to be used by both types. This would require lots of casting. Let's not do that.

## Description

In #10700, I will introduce BaseTime and TimerGameBase classes to store lots of useful things, which CalendarTime, EconomyTime, TimerGameCalendar and TimerGameEconomy will inherit. (Note: Outside a future Real Time Mode, TimerGameEconomy will mirror TimerGameCalendar, so it needs access to all those calendar-based functions.)

#10753 started this direction, but stopped short of what's happening in this PR. Let's finish the job and empty out (and then delete) `date_type.h`. This includes moving tick stuff to a new Tick class inside the TimerGameTick file.

## Checklist for review


src/timer/timer_game_calendar.h
```diff
@@ -109,6 +109,70 @@ class TimerGameCalendar {
...
+
+	static constexpr Year INVALID_YEAR = -1; ///< Representation of an invalid year
+	static constexpr Date INVALID_DATE = -1; ///< Representation of an invalid date
 };
 
+/** The date of the first day of the original base year. */
+static constexpr TimerGameCalendar::Date CALENDAR_DAYS_TILL_ORIGINAL_BASE_YEAR = TimerGameCalendar::DateAtStartOfYear(TimerGameCalendar::ORIGINAL_BASE_YEAR);
+
+/** The date of the last day of the max year. */
+static constexpr TimerGameCalendar::Date MAX_CALENDAR_DATE = TimerGameCalendar::DateAtStartOfYear(TimerGameCalendar::MAX_YEAR + 1) - 1;
```
TrueBrain: This unbalance of what variable is scoped inside `TimerGameCalendar` and which is not, will drive us absolutely mental. You have to check the source every time to know which is where. Simplicity and consistency of code is important for future-us; surprises like this really makes the difference between "this is easy to work with" and "wtf happened here" :D :)

So I would suggest either of the two (but there might be other suggestions):

(1) Move all constants outside of the class, and all use the `CALENDAR_` prefix.

This would ofc mean that later overloading and moving to a base-class will be more difficult. But you stepped into that one already with having 2 defined outside; you will need to redefine them anyway. So that is just a result of your current choice :D

(2) Create a new class `Calendar` (in the same file), where you put all the constants in. So that you get `Calendar::INVALID_YEAR`, and `Calendar::MAX_DATE`.

It is fine to leave the functions in `TimerGameCalendar`, as that way the second class is just a namespace. So on that note, it doesn't even have to be a class or struct, but it can also just be a namespace definition. Just a namespace you can't inherit, and as it was your goal to make it easier to split economy, a class (or struct) might be more to the point.

PS: this would also mean we have a nice place to leave a comment why those two classes aren't merged. That is also not clear now, why those two are not inside the class, and many people will try over the next few years :P

src/timer/timer_game_tick.h
```diff
@@ -14,8 +14,7 @@
 
 #include <chrono>
 
-/** Estimation of how many ticks fit in a single second. */
-static const uint TICKS_PER_SECOND = 1000 / MILLISECONDS_PER_TICK;
+typedef int32_t Ticks; ///< The type to store ticks in
```
TrueBrain: why define this outside of the class? And please use `using` when possible.

src/timer/timer_game_calendar.h
2TallTyler: I went with option 2, a new CalendarTime class (named to avoid confusion with economy.cpp when we have EconomyTime later).

I also switched to precomputed magic numbers for `DAYS_TILL_ORIGINAL_BASE_YEAR` and `MAX_DATE`, because defining those outside the class was worse, IMO.

Ticks also got their own class, for consistency and since I had to change all their references anyway.

src/timer/timer_game_calendar.h
```diff
@@ -13,6 +13,78 @@
...
+
+	/** The fraction of a date we're in, i.e. the number of ticks since the last date changeover. */
+	using DateFract = uint16_t;
+
+	/** Type for the year, note: 0 based, i.e. starts at the year 0. */
+	using Year = StrongType::Typedef<int32_t, struct YearTag, StrongType::Compare, StrongType::Integer>;
+	/** Type for the month, note: 0 based, i.e. 0 = January, 11 = December. */
+	using Month = uint8_t;
+	/** Type for the day of the month, note: 1 based, first day of a month is 1. */
+	using Day = uint8_t;
```
TrueBrain, line 47: Why did you also move the types? That still means you can't use the functions. That was not what my suggestion was about :)

I would suggest: leave the types and functions in TimerGameCalendar, and ONLY have the constants in this class. And move it below TimerGameCalendar. This way you don't have `701266`, but you can just do `TimerGameCalendar::DateAtStartOfYear(ORIGINAL_BASE_YEAR)`, which was kinda the point of my idea :D

src/aircraft_cmd.cpp
```diff
@@ -2095,7 +2095,7 @@ static bool AircraftEventHandler(Aircraft *v, int loop)
 	return true;
 }
 
-bool Aircraft::Tick()
+bool Aircraft::Ticks()
```
TrueBrain: Why this change? It does a single tick, so plural feels weird? Or just a mass rename? :D

2TallTyler: Oops, that's from a mass rename.

TrueBrain (approved):
Nice :)

Issue #8049, opened 2020-03-29
Suggestion: Allow Game Scripts to set the max loan of a company.


James103:
## Version of OpenTTD
As of 1.10.0-RC1

## Expected result
Game Scripts should be able to set the loan amount of a company independent of any restrictions like the loan interval and bank balance. For example, a Game Script should be able to set the loan of a company to £1,234,567 without changing the bank balance (which is both greater than the max loan and not a multiple of the loan interval).

An entirely new function designed for this purpose could be implemented to allow for this behavior to occur while taking the loan amount and company as parameters, or alternatively `GSCompany.SetLoanAmount` could be expanded to take an additional optional parameter for the company ID. If the additional parameter is supplied and there's no company scope, then the proposed behavior is executed instead of the current behavior.

## Actual result
Game Scripts are unable to set the loan of a company to any value that is not a multiple of the loan interval (£10,000) or that is greater than the max loan.

## Steps to reproduce
1. Run the following code in a Game Script without company mode scope:
```
GSCompany.SetLoanAmount(...) // Fails (no company scope)
```
2. Run the following code in a Game Script with company mode scope:
```
GSCompany.SetLoanAmount(1234567) // Fails (bigger than max loan)
GSCompany.SetLoanAmount(123456) // Fails (not a multiple of 10000)
GSTile.DemolishTile(...) // Spend some money
GSCompany.SetLoanAmount(0) // Fails (not enough money to set the loan)
```

LordAro:
I don't see a particular reason why you'd want to set a non-multiple-of-10000 loan amount. To avoid complicating the SetLoadAmount code, perhaps a better solution would be to add a new SetMaxLoanAmount GS-only function, that takes a company ID as parameter

James103:
Note that OpenTTD handles hacked loan amounts in a way that still (mostly) makes sense. Paying off a loan between 1 and 9999 still works correctly (even in the error handler), unless you control-click to pay off the loan. This will set the loan to a negative number, resulting in loan interest adding to your bank balance instead of subtracting.

James103:
The part about setting loans that aren't multiples of 10000 can be dropped as it is not as important as being able to set a loan higher than the max loan via a Game Script. I would still like it if a Game Script can set the max loan of a company to any non-negative multiple of 10,000 (this includes zeroing the max loan). I have changed the title to reflect this.

nielsmh:
[...]

PR #11224, opened 2023-08-22
Add: [GS] Allow to set max loan for each company separately

ldpl:
## Motivation / Problem

Seems to be quite a popular request. Making separate max loan can allow to use it as a reward for goal completion or story progression. Also, to implement some kind of bank credit rating.

## Description

- Adds `GSCompany.SetMaxLoanAmountForCompany` to set max loan for the company.
- Adds `GSCompany.ResetMaxLoanAmountForCompany` to switch it back to default (global) max loan.
- Changes `GSCompany.GetMaxLoanAmount` to return individual loans in company mode and global in deity.

## Limitations

- Doesn't provide API compatibility for `GSCompany.GetMaxLoanAmount` as I think new behavior is more adequate wrt this new feature instead of just returning global goal. 

## Checklist for review

* **This PR affects the GS/AI API? (label 'needs review: Script API')**
    * **ai_changelog.hpp, game_changelog.hpp need updating.**

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -96,6 +96,16 @@ void Company::PostDestructor(size_t index)
 }
 
 /**
+ * Calculate the max allowed loan for this company.
+ * @return the max loan amount.
+ */
+Money Company::GetMaxLoan() const
+{
+	if (this->max_loan == COMPANY_MAX_LOAN_DEFAULT) return _economy.max_loan;
+	return this->max_loan;
+}
+
+/**
  * Sets the local company and updates the settings that are set on a
  * per-company basis to reflect the core's state in the GUI.
  * @param new_company the new company
--- a/src/command_type.h
+++ b/src/command_type.h
@@ -253,6 +253,7 @@ enum Commands : uint16_t {
 
 	CMD_INCREASE_LOAN,                ///< increase the loan from the bank
 	CMD_DECREASE_LOAN,                ///< decrease the loan from the bank
+	CMD_SET_COMPANY_MAX_LOAN,         ///< sets the max loan for the company
 
 	CMD_WANT_ENGINE_PREVIEW,          ///< confirm the preview of an engine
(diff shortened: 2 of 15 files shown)


James103:
Note: this PR fixes https://github.com/OpenTTD/OpenTTD/issues/8049 for real.

src/company_cmd.cpp
```diff
@@ -95,6 +95,15 @@ void Company::PostDestructor(size_t index)
 	InvalidateWindowData(WC_ERRMSG, 0);
 }
 
+/**
+ * Calculate the max allowed loan for this company.
+ * @return the max loan amount.
+ */
+Money Company::GetMaxLoan() const {
```
PeterN: Line break before `{`

2TallTyler:
Code looks fine to me, with the caveat that I haven't touched script code very much.

Is there a GS that implements this, for testing?

ldpl:
Oops, forgot to attach the GS 
[testcompanymaxloan.zip](https://github.com/OpenTTD/OpenTTD/files/13652042/testcompanymaxloan.zip)

2TallTyler (commented):
I have tested this with the provided GS and found one small issue:

When a company's loan is changed, the company GUI is redrawn to update the `Maximum Loan:` status, but it does not check if the `Borrow £10,000` should be disabled. So if I have a loan of £290,000 and a limit of £300,000, and then my limit is changed to £10,000, the button is still clickable. It gives an error, but should be disabled. 🙂 

Otherwise, it seems to work as expected. Fix that issue and rebase, and I will approve.

ldpl:
> When a company's loan is changed, the company GUI is redrawn to update the `Maximum Loan:` status, but it does not check if the `Borrow £10,000` should be disabled. So if I have a loan of £290,000 and a limit of £300,000, and then my limit is changed to £10,000, the button is still clickable. It gives an error, but should be disabled. 🙂

Fixed

2TallTyler (approved):

PR #11225, opened 2023-08-22
Fix #11161: Apply cargo filter to shared groups.

### LINKED ISSUE #11161: [Bug]: Vehicle list cargo filter ignored when grouping by shared orders
### Version of OpenTTD

13.4

### Expected result

Using the cargo filter dropdown in a vehicle group only shows vehicles which can load that cargo.

### Actual result

oil

All vehicles in the selected group are shown, regardless of what they can carry.

This is true for all types of vehicles, not just RVs.

### Steps to reproduce

1. Buy some vehicles fitted for different cargos (for example, Oil and Passengers)
2. Filter the list to show only vehicles that carry Oil
3. Observe that the Passengers vehicles disappear from the list
4. Group the vehicle list by shared orders
5. Observe that the Passengers vehicles reappear

Durchbruchswagen:
## Motivation / Problem
Fix #11161 

Closes #11294

## Description
To solve the problem I had to somehow apply a cargo filter on groups of vehicles, unfortunately currently used filter only works for groups of size one (so basically single vehicle). I workaround it by first sorting all vehicles by shared orders and then gathering them into vehicle groups of size one. After that I applied a cargo filter and was left with vector of filtered vehicle groups of size one, sorted by their shared orders so I iterated over this vector gathering those single vehicles into groups.

## Limitations
During testing I noticed that even though shared groups don't show vehicles that don't carry a specified cargo type when you inspect one of those groups you see all vehicles that share that certain order, even those that don't carry specified cargo type.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -204,22 +204,33 @@ void BaseVehicleListWindow::BuildVehicleList()
 	} else {
 		/* Sort by the primary vehicle; we just want all vehicles that share the same orders to form a contiguous range. */
 		std::stable_sort(this->vehicles.begin(), this->vehicles.end(), [](const Vehicle * const &u, const Vehicle * const &v) {
+			if (u->FirstShared() == v->FirstShared()) {
+				return u->cargo_type < v->cargo_type;
+			}
 			return u->FirstShared() < v->FirstShared();
 		});
 
+		for (auto it = this->vehicles.begin(); it != this->vehicles.end(); ++it) {
+			this->vehgroups.emplace_back(it, it + 1);
+		}
+
+		this->FilterVehicleList();
+
 		uint max_num_vehicles = 0;
 
-		VehicleList::const_iterator begin = this->vehicles.begin();
-		while (begin != this->vehicles.end()) {
-			VehicleList::const_iterator end = std::find_if_not(begin, this->vehicles.cend(), [first_shared = (*begin)->FirstShared()](const Vehicle * const &v) {
-				return v->FirstShared() == first_shared;
-			});
+		GUIVehicleGroupList::iterator begin = this->vehgroups.begin();
+		GUIVehicleGroupList::iterator end = begin;
+		std::advance(end, 1);
 
-			this->vehgroups.emplace_back(begin, end);
-
+		while (begin != this->vehgroups.end()) {
+			while (end != this->vehgroups.end() && begin->vehicles_begin[0]->FirstShared() == end->vehicles_begin[0]->FirstShared()) {
+				begin->vehicles_end = end->vehicles_end;
+				end = this->vehgroups.erase(end);
+			}
 			max_num_vehicles = std::max<uint>(max_num_vehicles, static_cast<uint>(end - begin));
 
 			begin = end;
+			std::advance(end, 1);
 		}
 
 		this->unitnumber_digits = CountDigitsForAllocatingSpace(max_num_vehicles);


src/vehicle_gui.cpp
```diff
@@ -203,23 +203,34 @@ void BaseVehicleListWindow::BuildVehicleList()
 		this->FilterVehicleList();
 	} else {
 		/* Sort by the primary vehicle; we just want all vehicles that share the same orders to form a contiguous range. */
-		std::stable_sort(this->vehicles.begin(), this->vehicles.end(), [](const Vehicle * const &u, const Vehicle * const &v) {
+		std::stable_sort(this->vehicles.begin(), this->vehicles.end(), [](const Vehicle * const &u, const Vehicle * const &v){
```
2TallTyler:
```suggestion
		std::stable_sort(this->vehicles.begin(), this->vehicles.end(), [](const Vehicle * const &u, const Vehicle * const &v) {
```

src/vehicle_gui.cpp
```diff
@@ -203,23 +203,34 @@ void BaseVehicleListWindow::BuildVehicleList()
...
+			if(u->FirstShared() == v->FirstShared()){
```
2TallTyler:
```suggestion
			if (u->FirstShared() == v->FirstShared()) {
```

src/vehicle_gui.cpp
```diff
@@ -203,23 +203,34 @@ void BaseVehicleListWindow::BuildVehicleList()
...
+
+		uint max_num_vehicles = 0;
 
-			this->vehgroups.emplace_back(begin, end);
+		GUIVehicleGroupList::iterator begin = this->vehgroups.begin();
+		GUIVehicleGroupList::iterator end = begin;
+		std::advance(end, 1);
 
+		while (begin != this->vehgroups.end()) {
+			while(end != this->vehgroups.end() && begin->vehicles_begin[0]->FirstShared()==end->vehicles_begin[0]->FirstShared()){
```
2TallTyler:
```suggestion
			while (end != this->vehgroups.end() && begin->vehicles_begin[0]->FirstShared() == end->vehicles_begin[0]->FirstShared()) {
```

src/vehicle_gui.cpp
```diff
@@ -203,23 +203,34 @@ void BaseVehicleListWindow::BuildVehicleList()
...
+				begin->vehicles_end = end->vehicles_end;
+				end=this->vehgroups.erase(end);
```
2TallTyler:
```suggestion
				end = this->vehgroups.erase(end);
```

2TallTyler (commented):
No comments on the actual fix, but there are a bunch of missing spaces. 🙂

TrueBrain:
What you purpose is a fine solution. What triggered me a bit was the limitation. I would expect that if you filter the list of shared orders, you will see the full group when only one vehicle matches. I would not expect the group to be filtered itself. This depends a bit on your definition of the filtering .. I would expect the filtering to be done after-the-fact, while you suggest to do it before-the-fact. This is just an opinion, and there isn't a right or wrong here.

I am curious, what method do you prefer? As an alternative, I wrote https://github.com/OpenTTD/OpenTTD/pull/11294 . Which does what I suggest. I am very curious to your opinion in the matter.

Durchbruchswagen:
So I checked your solution and I think that I like yours a little bit more than mine, to be honest. Besides being less awkward than mine,  filtering after-the-fact seems more "intuitive" because if someone further inspects the vehicle group they are going to see the entire group and by that I mean they are going to see even those vehicles that do not carry specified cargo but share the same order. 
Small little nitpick to your solution. In `CargoFilter` after you set `match` to `true` there is no way that it is going to turn back to `false` so you may as well `return true` at this point.

PR #11227, opened 2023-08-25
Codechange: Invalid depot value is unrelated to invalid station value.

J0anJosep:
## Motivation / Problem
Depot indexes and Station indexes are unrelated although both of them are represented with UINT16 (and UINT16_MAX being used as the invalid value for both of them).

### DIFF
--- a/src/depot_type.h
+++ b/src/depot_type.h
@@ -10,12 +10,10 @@
 #ifndef DEPOT_TYPE_H
 #define DEPOT_TYPE_H
 
-#include "station_type.h"
-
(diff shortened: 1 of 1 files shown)


TrueBrain:
Please don't modify the PR template this much; it really helps us in the review process to leave the layout the same. If a chapter remains empty, so be it. Tnx :)

TrueBrain (approved):

J0anJosep:
> Please don't modify the PR template this much; it really helps us in the review process to leave the layout the same. If a chapter remains empty, so be it. Tnx :)

Sorry!

PR #11228, opened 2023-08-25
Doc: Improve the example given for fall throughs.

J0anJosep:
## Motivation / Problem

In CODYNGSTYLE.md it states that "All fall throughs must be documented, using a FALLTHROUGH define/macro", but in the code provided they are missing.

## Description

Add `FALLTHROUGH;` in the given code and remove the unnecessary comments.

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -180,7 +180,7 @@ switch (a) {
 
 	case 1:
 		DoSomething();
-		/* FALL THROUGH */
+		FALLTHROUGH;
 
 	case 2:
 		DoMore();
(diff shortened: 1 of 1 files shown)


CODINGSTYLE.md
```diff
@@ -181,6 +181,7 @@ switch (a) {
 	case 1:
 		DoSomething();
 		/* FALL THROUGH */
```
TrueBrain: Now this comment can also be removed, as the macro alone is more than clear enough :)

J0anJosep: Ok!

J0anJosep: I was about to comment the same and ask whether should I remove it :)

TrueBrain (approved):

PR #11231, opened 2023-08-27
Fix #11230: Group list widget layout failed, if dropdown widgets were higher than the font height.

### LINKED ISSUE #11230: [Bug]: Sort by button is misaligned with OpenGFX2.
### Version of OpenTTD

20230630-master-f082c47a4f9

### Expected result

The vehicle list shouldn't have a hole in it! But with OpenGFX2 under the "sort by" button there's a hole....

### Actual result

opengfx2_listbug

When using OpenGFX2 a hole appears... Related to the size of sprite 4942? Because the "Sort by" button is too small vertically. But that doesn't quite make sense, because the "Group by" button seems too small vertically too.

Not sure if this should count as a base set or OpenTTD bug, but I suspect assumptions about sprite sizes shouldn't give holes in windows.

### Steps to reproduce

Just download OpenGFX2 via BaNaNaS and set as the base set... Open any vehicle list and you'll see the gap.

frosch123:
## Motivation / Problem

#11230 

## Description

Our lord insisted that an ugly layout is still better than a glitchy layout:
image

## Checklist for review


TrueBrain:
Guess my only concern would be how a baseset author is ever going to figure out they are doing something bad .. suggestions for that, by any chance?

frosch123:
This is only a drop in the ocean of concerns a baseset author has to account for :)
Try replacing the sprites for foundations (V453000 probably still has sweaty nightmares), toy factory or bubble generator.

TrueBrain (approved):

PR #11233, opened 2023-08-27
Codechange: [MacOS] use backtrace() instead of our custom variant

TrueBrain:
## Motivation / Problem

As mentioned in the comment, we only did it ourselves as we once were compatible with versions before 10.5. But that time has long gone. So let's update the code to a bit more modern approach.

## Description

Remove custom backtrace code, and reuse the exact same code as Linux has been using.

## Limitations

The output is slightly different, but not in a way that should be any problem to any reader.

## Checklist for review


glx22 (approved):
Assuming you tested it :)

PR #11232, opened 2023-08-27
Change: store crash logs in JSON format

TrueBrain:
## Motivation / Problem

A few things are coming together:

- I want to transmit crashes to us automatically (opt-in, ofc) when they happen. This to better react to bugs in our software. By the looks, there are very few crashes going on, but often we depend on users reporting them, and it can be a bit of a struggle to get the right files delivered to us. By automating this, I hope to make us a bit more responsive to fuck-ups, especially just after a release.
- Survey code looks up various of system-information, and the crashlog does too. But in the way the crashlog was written, it was nearly impossible to reuse that code. High on my wish-list was still a way to fix that, and deduplicate this information.
- Crash logs were wildly different between OSes, mostly for no real reason.

In an attempt to prepare for the first and address the other two, I created this PR.

## Description

The first commit contains moving of some code around, to make the next commit easier. GitHub doesn't show this well, but if you use `git diff HEAD^^..HEAD^ --color-moved` on this PR, you see with colour that it is moved code.

Instead of generating a `crash.log`, a crash now generates a `crash.json.log` (the extra `.log` is needed, as you are not allowed to upload a `.json` to GitHub .. so to make it easier for manual uploads, add another `.log`). Initially I wanted to convert the JSON back to a log-like file, but I realised that wouldn't actually add value. A human can read the JSON just as well as a flat file, and skipping another conversion sounds like a win.

During a crash, a short summary is printed to the console, as especially on Linux we are kinda used to seeing the stacktrace there.

The Crashlog contains slightly more information than the Survey. Most noticeable, the stacktrace (duh), the gamelog and recent news. None of them are needed for the Survey.

Tested on the following platforms:
- x64-linux
- x64-windows
- x64-macos

Example JSON file:
[crash20230827121137.json.txt](https://github.com/OpenTTD/OpenTTD/files/12447937/crash20230827121137.json.txt)

<details><summary>In-line example</summary>

```
{
    "crash": {
        "id": 11,
        "reason": "Segmentation fault"
    },
    "date": "2023-08-27 12:11:37 (UTC)",
    "game": {
        "companies": {
            "0": {
                "infrastructure": {
                    "airport": 0,
                    "rail": 0,
                    "road": 0,
                    "signal": 0,
                    "station": 0,
                    "tram": 0,
                    "water": 0
                },
                "type": "human",
                "vehicles": {
                    "aircraft": 0,
                    "roadveh": 0,
                    "ship": 0,
                    "train": 0
                }
            }
        },
[...]

PR #11236, opened 2023-08-27
Change: [Win32] unify the crashlog handler with the other OSes

TrueBrain:
## Motivation / Problem

Windows didn't use the base crashlog handler for most part, and does everything itself. This is a bit annoying when we want to change things, as you have to remember to change it in two places. So while working on #11232 this was a real pita.

So, in this PR I set out to fix that discrepancy, and make sure Windows is much more like the rest.

## Description

The main difference is that Windows delayed creating a `crash.sav`. This is the only OS that does so, both MacOS and Linux create a `crash.sav` immediately. Otherwise, other PRs before this PR already made sure Windows is much more like the rest.

While at it, make the crash text a bit more readable, and sync this with MacOS. It just read a bit awkward.

## Limitations

There is an argument to be made that a `crash.sav` should not be made when a crashlog is generated, but only on demand. As it is not unlikely creating a savegame causes other problems. After all, we are in a broken state. But we also noticed this often leads to users not having a `crash.sav` to add to a ticket, as they simply didn't read the dialog, or didn't understand what it was saying.

In the long run, I rather focus on something JGRPP did, and detect if we crash during the crash handler, and handle that case more proper. That will be done in a follow-up PR, for all OSes (and not only for Windows).

## Checklist for review

### DIFF
--- a/src/os/windows/crashlog_win.cpp
+++ b/src/os/windows/crashlog_win.cpp
@@ -270,12 +270,7 @@ static LONG WINAPI ExceptionHandler(EXCEPTION_POINTERS *ep)
 
 	CrashLogWindows *log = new CrashLogWindows(ep);
 	CrashLogWindows::current = log;
-	auto output_iterator = std::back_inserter(log->crashlog);
-	log->FillCrashLog(output_iterator);
-	log->WriteCrashDump();
-	log->WriteCrashLog();
-	log->WriteScreenshot();
-	log->SendSurvey();
+	log->MakeCrashLog();
 
 	/* Close any possible log files */
 	CloseConsoleLogIfActive();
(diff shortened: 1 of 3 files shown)


src/os/windows/crashlog_win.cpp
```diff
@@ -403,11 +392,11 @@ static INT_PTR CALLBACK CrashDialogFunc(HWND wnd, UINT msg, WPARAM wParam, LPARA
...
 
-			/* Add path to crash.log and crash.dmp (if any) to the crash window text */
+			/* Add path to all files to the crash window text */
 			size_t len = wcslen(_crash_desc) + 2;
 			len += wcslen(convert_to_fs(CrashLogWindows::current->crashlog_filename, filenamebuf, lengthof(filenamebuf))) + 2;
 			len += wcslen(convert_to_fs(CrashLogWindows::current->crashdump_filename, filenamebuf, lengthof(filenamebuf))) + 2;
-			len += wcslen(convert_to_fs(CrashLogWindows::current->screenshot_filename, filenamebuf, lengthof(filenamebuf))) + 1;
+			len += wcslen(convert_to_fs(CrashLogWindows::current->screenshot_filename, filenamebuf, lengthof(filenamebuf))) + 2;
 
 			static wchar_t text[lengthof(_crash_desc) + 3 * MAX_PATH * 2 + 7];
```
glx22:
```suggestion
			static wchar_t text[lengthof(_crash_desc) + 3 * MAX_PATH * 2 + 8];
```
to match the change in `len`calculation I think. (Though we already double the paths max length).

TrueBrain: I misunderstood what the `* 2` was doing .. it is to reserve space for the wchar .. (UTF-16 and all) .. oops. Reverting my change here :)

glx22 (approved):

PR #11238, opened 2023-08-27
Change: recover when possible from crashes during a crash

TrueBrain:
## Motivation / Problem

When, for example, creating of a savegame fails during the crash handler, everything stops and the user is presented with what-ever the OS decides. But in most cases, we don't actually mind that much that the savegame or, for example, the screenshot failed to be created. These issues should just be ignored and the crashlog should do as much as possible.

## Description

By clever use of `setjmp` and Windows SEH, we can protect certain blocks, that if they cause a segfault or something, we capture that, clean it up, and return in the crashlog process.

I went with the function name `TryExecute`, which takes a `std::function` that should return a boolean. The function itself should also return a boolean. Now the rules are simple:

If the function returns true and caused no problem, `TryExecute` returns true.
If the function returns false and caused no problem, `TryExecute` returns false (indicating that the function failed).
If the function caused an exception, `TryExecute` returns false (indicating that the function failed).

This way, if the `TryExecute` return true, it means everything was fine. This allows for the caller to handle what he wants to be done with problems.

This solution is heavily inspired by the work of @JGRennison in JGRPP; the main difference is that there a `char *` is returned, and there is a checkpoint system to ensure it is reset to a known-good-state if an exception happens. Here however I made it a bit more generic that I could also attach `WriteSavegame` to it, and leave the handling of the buffer to the caller. Otherwise, the implementation and idea behind it is very similar, although I did clean up some things, named things slightly different, etc, to hopefully pass easier in our review process.

## Limitations

I did not invest time in adding this functionality to `FillCrashLog`, because with #11232 that is going to change significantly enough that it was not worth my time and effort. In other words: I couldn't find a simple way to remember and restore a `back_insert_iterator`, which made that change blow up in my face. So postponing till #11232 :) This PR was mostly to prevent issues with savegames anyway (as #11236 took that away from the Windows users).

## Checklist for review

### DIFF
--- a/src/os/windows/crashlog_win.cpp
+++ b/src/os/windows/crashlog_win.cpp
@@ -25,6 +25,8 @@
 
 #if defined(_MSC_VER)
 #	include <dbghelp.h>
+#else
+#	include <setjmp.h>
 #endif
 
 #ifdef WITH_UNOFFICIAL_BREAKPAD
@@ -32,6 +34,21 @@
 #endif
 
 #include "../../safeguards.h"
+
+/** Exception code used for custom abort. */
+static constexpr DWORD CUSTOM_ABORT_EXCEPTION = 0xE1212012;
+
+/**
+ * Forcefully try to terminate the application.
+ *
+ * @param exit_code The exit code to return.
+ */
+static void NORETURN ImmediateExitProcess(uint exit_code)
+{
+	/* TerminateProcess may fail in some special edge cases; fall back to ExitProcess in this case. */
+	TerminateProcess(GetCurrentProcess(), exit_code);
+	ExitProcess(exit_code);
+}
 
 /**
  * Windows implementation for the crash logger.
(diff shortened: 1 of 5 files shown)


src/os/windows/crashlog_win.cpp
```diff
@@ -55,21 +72,65 @@ class CrashLogWindows : public CrashLog {
...
+#if defined(_MSC_VER)
+	/* virtual */ bool TryExecute(std::string_view section_name, std::function<bool()> &&func) override
+	{
+		this->try_execute_active = true;
+		bool res;
+
+		__try {
+			res = func();
+		} __except (EXCEPTION_EXECUTE_HANDLER) {
+			fmt::print("Something went wrong when attempting to fill {} section of the crash log.\n", section_name);
```
glx22, line 90: Printing won't show much in typical release run.

TrueBrain: True; but it also isn't all that important. The handler of the 'return false' will make it more visible. But for debugging at least it was useful.

Better suggestions? Or just leave it as it is?

glx22: Handler of the 'return false' is also and fmt::print it seems.

TrueBrain: Somewhat. And there is still a bit of work to do here. But basically, it should reset the filename, so it doesn't show up in the error report. That is a bit the point of this `TryExcept`. But the current handling of all this is a bit awkward, but also not for this PR, I think :)

glx22: Yeah this PR doesn't make it worse

TrueBrain: (mainly, MacOS currently always prints the filename, even if it is still empty .. so much stuff to fix here :D)

glx22 (approved):

PR #11240, opened 2023-08-27
Change: [Win32] don't allocate 192KiB of memory on the stack on crash

TrueBrain:
## Motivation / Problem

While looking at the Windows crashlog dialog, I noticed it was allocating a lot of memory on the stack. So I looked at JGRPP, and got some inspiration how to solve it properly.

## Description

Heap is out of the question, as it might be corrupted. Allocating this much on stack is silly.

So instead, allocate virtual pages to write the information in.

While at it, also added the savegame filename to the dialog, and as they are (now) always set, remove any conditional around them. This simplifies the code a lot, as now it can be printed by a single snprint.

## Limitations

Where JGRPP took the approach of: better allocate too much memory than run out of a buffer, I went for: let's do the math correctly, and make sure all buffers fit, while making them nice and snugly. This means the buffers are on the byte of the correct size.
JGRPP's approach might be more robust; although I am pretty sure my math adds up, and should be safe. I hope.

## Checklist for review


src/os/windows/crashlog_win.cpp
```diff
@@ -462,44 +462,51 @@ static INT_PTR CALLBACK CrashDialogFunc(HWND wnd, UINT msg, WPARAM wParam, LPARA
...
+			wchar_t *crashlog_buf = crash_desc_buf + crash_desc_buf_length;
+			wchar_t *filename_buf = crashlog_buf + crashlog_length;
+			char *crashlog_dos_nl = reinterpret_cast<char *>(filename_buf + filename_buf_length * filename_count);
+
+			/* Convert unix -> dos newlines because the edit box only supports that properly. */
+			const char *crashlog_unix_nl = CrashLogWindows::current->crashlog.data();
+			char *p = crashlog_dos_nl;
 			char32_t c;
-			while ((c = Utf8Consume(&unix_nl)) && p < lastof(dos_nl) - 4) { // 4 is max number of bytes per character
+			while ((c = Utf8Consume(&crashlog_unix_nl))) {
```
TrueBrain, line 492: The removal of the `p <` check is safe, as `crashlog_dos_nl` is calculated to be at least as wide as `crashlog_unix_nl`. So it is not possible for the latter to return a character while the first can't fit it.

(please confirm my claim, ofc :D )

glx22: And `\n` are counted twice to make room for `\r` in `crashlog_dos_nl`. I think I can confirm your claim :).

src/os/windows/crashlog_win.cpp
```diff
@@ -462,44 +462,51 @@ static INT_PTR CALLBACK CrashDialogFunc(HWND wnd, UINT msg, WPARAM wParam, LPARA
...
-			wchar_t filenamebuf[MAX_PATH * 2];
-			wchar_t crash_msgW[65536];
-			/* Convert unix -> dos newlines because the edit box only supports that properly :( */
-			const char *unix_nl = CrashLogWindows::current->crashlog.data();
-			char dos_nl[65536];
-			char *p = dos_nl;
+			 * buffer in MB_TO_WIDE is not large enough (512 chars).
+			 * Use VirtualAlloc to allocate pages for the buffer to avoid overflowing the stack.
+			 * Avoid the heap in case the crash is because the heap became corrupted. */
+			const size_t total_length = crash_desc_buf_length * sizeof(wchar_t) + \
```
glx22: `\` at end of line are only for preprocessor stuff like macros, the compiler knows the C++ line ends at the next `;`

glx22 (approved):

PR #11243, opened 2023-08-28
Increase the number of maximum cargoes there can be in a single game

Altitud49:
This feature shouldmake it easier to develop complex economy sets (notably AXIS) without going through the hassle of trying to know what supply chains shuld be let down in favor of others or merging cargoes that could have merit being split up.

Note that I didn't knpw what I was doing for half the time I was developing this feature 😞 .

### DIFF
--- a/src/cargo_type.h
+++ b/src/cargo_type.h
@@ -62,7 +62,7 @@ enum CargoType {
 	CT_FIZZY_DRINKS = 11,
 
 	NUM_ORIGINAL_CARGO = 12,
-	NUM_CARGO       = 64,   ///< Maximal number of cargo types in a game.
+	NUM_CARGO       = 128,   ///< Maximal number of cargo types in a game.
 
 	CT_AUTO_REFIT   = 0xFD, ///< Automatically choose cargo type when doing auto refitting.
 	CT_NO_REFIT     = 0xFE, ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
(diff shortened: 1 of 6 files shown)


TrueBrain:
Your first contribution, nice!

Sadly, you didn't pick the simplest subject .. simply increasing the amount of bits to increase the amount of cargo possible is not as easy as it sounds. When you reach the 64bit barrier, any increase has a significant penalty in several places. For example, each station all of a sudden consumes roughly 10 KiB more memory. But also read/writes to such variables are much more expensive CPU wise, as on most CPUs it takes several instructions to manipulate 128bit fields.

A bit where this balances on, is: what is next? Someone thinks 128 cargoes isn't enough, and adds even more bits? Where do we stop adding bits .. it is important to note that all games suffer for the few games that actually use them.

So for us this is difficult to balance. A much better, but much much more complicated, solution might be to not use a bitfield, but some other form of storage to only store the amount of cargo actually used. But quickly this requires even more memory / CPU to pull off, and so what we are left with is 64 cargoes.

I hope you understand the complexity here, and if you want to weigh in, you are more than welcome!


PS: for next time, and I hope there is one, please follow the Pull Request template presented to you when you were opening a Pull Request. We put great effort in streamlining the review process .. just throwing that away is not nice :D But no harm done, you have the most important bits in your text currently :)

Altitud49:
Oh okay sorry... Thank you for your time ! 😅

PR #11250, opened 2023-08-31
Fix: [Script] GSAdmin.Send() could generate invalid JSON

TrueBrain:
## Motivation / Problem

Given the following script-snippet:

```
GSAdmin.Send({ test = GSAdmin });
```

This would send over the Admin Port: `{ `.

## Description

A bit of a silly bug: not checking the return value.

## Checklist for review

### DIFF
--- a/src/script/api/script_admin.cpp
+++ b/src/script/api/script_admin.cpp
@@ -129,7 +129,10 @@
 	}
 
 	std::string json;
-	ScriptAdmin::MakeJSON(vm, -1, SQUIRREL_MAX_DEPTH, json);
+	if (!ScriptAdmin::MakeJSON(vm, -1, SQUIRREL_MAX_DEPTH, json)) {
+		sq_pushinteger(vm, 0);
+		return 1;
+	}
 
 	if (json.length() > NETWORK_GAMESCRIPT_JSON_LENGTH) {
 		ScriptLog::Error("You are trying to send a table that is too large to the AdminPort. No data sent.");


src/tests/script_admin_json.cpp
```diff
@@ -0,0 +1,104 @@
...
+	sq_pop(engine->GetVM(), 1);
+
+	CHECK(admin->expected_call == 0);
+}
+
+TEST_CASE("Squirrel -> JSON conversion")
+{
+	/* Create a fake admin socket that can validate our JSON. */
+	ServerNetworkAdminSocketHandler::CanAllocateItem();
+	TestServerNetworkAdminSocketHandler *a = new TestServerNetworkAdminSocketHandler();
```
rubidium42: I would use `std::unique_ptr`. Then you don't need to `delete` is at the end of the test.

rubidium42 (commented):
This looks quite reasonable to me.

However, for more limited testing I can imagine splitting up ScriptAdmin::Send into two functions. One that does all the error checking and creation of the string/return values, and then Send would call that function, in case there is no error (non-empty `std::optional<std::string>`), it sends the string over the network. In either case it'll return the return value from the inner function.
This inner function can then be unit tested, however not directly using the Squirrel API. So we need someway for writing a small bit of Squirrel code for tests that generates the appropriate `HSQUIRRELVM` that can be passed to the object under test. Then you can test the inner function for return value, potentially set error messages and whether the string has been set or not.
The actual sending of the script must then still be tested manually, but if the unit test of the inner test is correct, then sending JSON to the admin port either always works or never works (i.e. you flipped the logic whether to send data on the admin port).

I'm not sure whether it is really worth all that effort though for this function. Unit testing MakeJSON on the other hand seems better, although I could imagine that might be possible from within Squirrel land if we were to have an GS API test like we have for the AI API. Though we need to make sure that the return value for "not valid to send" and "sent correctly to anyone that wanted it" (i.e. might not have been actually sent) are different, as they seem to be the same right now.

glx22 (approved):

PR #11251, opened 2023-08-31
Codechange: [Script] use nlohmann for Squirrel <-> JSON conversion

TrueBrain:
## Motivation / Problem

We rolled our custom JSON generator for `GSAdmin.Send` / `ScriptEventAdminPort`. Now we have nlohmann, switch to that.

## Description

A relative straight-forward conversion. Few things worth nothing:

- `max_depth` was a bit weird, counting down. Instead, it is now `depth`, counting up. That way a backtrace shows better at what depth we were.
- GSAdmin.Send() couldn't send more than ~1400 bytes of JSON. This was a (rather old) limitation that each packet couldn't consume more than one packet. This limitation has been solved a long time ago, but never made it into GSAdmin.Send(). So while at it, also removed that restriction. For incoming JSON there is still a restriction, but increased the value of that to 9000. Just a random number. Open for better suggestions.
- There was a restriction that a string was not allowed to exceed 254 characters. I am not sure where this restriction comes from, or why it was there. I just removed it. Felt like a weird limitation for not actual reason.
- The custom JSON generator added spaces in all kinds of places. nlohmann doesn't (unless told to). The spaces makes no sense to me, as you want this to be as compact as possible. It doesn't have to be human readable. That is something for the other side of the admin-port to handle / do.

## Checklist for review

### DIFF
--- a/src/network/core/config.h
+++ b/src/network/core/config.h
@@ -60,7 +60,7 @@ static const uint NETWORK_REVISION_LENGTH           =   33;           ///< The m
 static const uint NETWORK_PASSWORD_LENGTH           =   33;           ///< The maximum length of the password, in bytes including '\0' (must be >= NETWORK_SERVER_ID_LENGTH)
 static const uint NETWORK_CLIENT_NAME_LENGTH        =   25;           ///< The maximum length of a client's name, in bytes including '\0'
 static const uint NETWORK_RCONCOMMAND_LENGTH        =  500;           ///< The maximum length of a rconsole command, in bytes including '\0'
-static const uint NETWORK_GAMESCRIPT_JSON_LENGTH    = COMPAT_MTU - 3; ///< The maximum length of a gamescript json string, in bytes including '\0'. Must not be longer than COMPAT_MTU including header (3 bytes)
+static const uint NETWORK_GAMESCRIPT_JSON_LENGTH    = 9000;           ///< The maximum length of a receiving gamescript json string, in bytes including '\0'.
 static const uint NETWORK_CHAT_LENGTH               =  900;           ///< The maximum length of a chat message, in bytes including '\0'
 static const uint NETWORK_CONTENT_FILENAME_LENGTH   =   48;           ///< The maximum length of a content's filename, in bytes including '\0'.
 static const uint NETWORK_CONTENT_NAME_LENGTH       =   32;           ///< The maximum length of a content's name, in bytes including '\0'.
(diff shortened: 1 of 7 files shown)


src/network/core/config.h
```diff
@@ -60,7 +60,7 @@ static const uint NETWORK_REVISION_LENGTH           =   33;           ///< The m
 static const uint NETWORK_PASSWORD_LENGTH           =   33;           ///< The maximum length of the password, in bytes including '\0' (must be >= NETWORK_SERVER_ID_LENGTH)
 static const uint NETWORK_CLIENT_NAME_LENGTH        =   25;           ///< The maximum length of a client's name, in bytes including '\0'
 static const uint NETWORK_RCONCOMMAND_LENGTH        =  500;           ///< The maximum length of a rconsole command, in bytes including '\0'
-static const uint NETWORK_GAMESCRIPT_JSON_LENGTH    = COMPAT_MTU - 3; ///< The maximum length of a gamescript json string, in bytes including '\0'. Must not be longer than COMPAT_MTU including header (3 bytes)
+static const uint NETWORK_GAMESCRIPT_JSON_LENGTH    = 9000;           ///< The maximum length of a receiving gamescript json string, in bytes including '\0'.
```
glx22, line 63: Is the constant still used ?

TrueBrain: Yes. As mentioned in description of the PR: for incoming JSON messages. We still have an upperbound on all receive-strings there.

src/script/api/script_admin.cpp
```diff
@@ -28,89 +28,73 @@
...
-				if (!first) data += ", ";
-				if (first) first = false;
-
-				std::string key;
-				std::string value;
-
-				/* Store the key + value */
-				bool res = MakeJSON(vm, -2, max_depth - 1, key) && MakeJSON(vm, -1, max_depth - 1, value);
+				/* Squirrel ensure the key is a string. */
+				assert(sq_gettype(vm, -2) == OT_STRING);
```
PeterN, line 69: Safe to assume this assert can't be triggered by user scripts?

PeterN (approved):
Looks okay, haven't tested (wouldn't know how) but less is more, and clearly using json library is better than implementing one badly...

src/script/api/script_admin.cpp
TrueBrain: Yeah, that is what this comment tries to say. A table/dict in Squirrel always has strings as keys (by definition). So there is no need here to check other types. But I did wanted to add the assert to make sure the claim Squirrel makes it actually correct :)

PR #11258, opened 2023-09-03
Change: Replace fixed length _grf_text array with vector.

PeterN:
## Motivation / Problem

On x86-64, `_grf_text` is a 20MB array. This was due to increasing the number of permitted IDs without realising the ID is used as an array index.

## Description

IDs are allocated sequentially by OpenTTD, not by NewGRFs, so it is possible to replace the array with a vector and allocate as necessary. The limit is still enforced, though is unlikely to be reached.

Additionally reshuffle GRFTextEntry for better alignment. This only seems to help 32bit x86, but shouldn't hurt other architectures.

## Limitations

An assert has been added for out-of-bounds access. I have assumed the NewGRFs cannot interfere and pass IDs directly.

## Checklist for review

### DIFF
--- a/src/newgrf_text.cpp
+++ b/src/newgrf_text.cpp
@@ -636,7 +636,7 @@ const char *GetGRFStringFromGRFText(const GRFTextWrapper &text)
 /**
  * Get a C-string from a stringid set by a newgrf.
  */
-const char *GetGRFStringPtr(uint16_t stringid)
+const char *GetGRFStringPtr(uint32_t stringid)
 {
 	assert(stringid < _grf_text.size());
 	assert(_grf_text[stringid].grfid != 0);
--- a/src/newgrf_text.cpp
+++ b/src/newgrf_text.cpp
@@ -64,15 +64,14 @@ enum GRFExtendedLanguages {
  * since it is NOT SUPPOSED to happen.
  */
 struct GRFTextEntry {
+	GRFTextList textholder;
+	StringID def_string;
 	uint32_t grfid;
 	uint16_t stringid;
-	StringID def_string;
-	GRFTextList textholder;
 };
 
 
-static uint _num_grf_texts = 0;
-static GRFTextEntry _grf_text[TAB_SIZE_NEWGRF];
+static std::vector<GRFTextEntry> _grf_text;
 static byte _currentLangID = GRFLX_ENGLISH;  ///< by default, english is used.
 
 /**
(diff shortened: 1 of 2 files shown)


src/newgrf_text.cpp
```diff
@@ -561,27 +560,22 @@ StringID AddGRFString(uint32_t grfid, uint16_t stringid, byte langid_to_add, boo
...
-			break;
-		}
+	auto it = std::find_if(std::begin(_grf_text), std::end(_grf_text), [&grfid, &stringid](const GRFTextEntry &grf_text) { return grf_text.grfid == grfid && grf_text.stringid == stringid; });
+	if (it == std::end(_grf_text)) {
+		/* Too many strings allocated, return empty */
+		if (_grf_text.size() == TAB_SIZE_NEWGRF) return STR_EMPTY;
+
+		/* We didn't find our stringid and grfid in the list, allocate a new id */
+		_grf_text.emplace_back();
+		it = std::end(_grf_text) - 1;
```
glx22:
```suggestion
		it = _grf_text.emplace(std::end(_grf_text));
```
Maybe ?

PeterN: Thanks, that's the construct I was trying to find but failing...

src/newgrf_text.cpp
```diff
@@ -645,6 +638,7 @@ const char *GetGRFStringFromGRFText(const GRFTextWrapper &text)
  */
 const char *GetGRFStringPtr(uint16_t stringid)
 {
+	assert(_grf_text.size() > stringid);
```
TrueBrain: Doesn't the inverse make more sense? Stringid < size? As I think it should be >= otherwise, not?

It is late, pretty sure I am wrong about the =. The order still confuses my brain :D

src/newgrf_text.cpp
```diff
@@ -561,27 +560,21 @@ StringID AddGRFString(uint32_t grfid, uint16_t stringid, byte langid_to_add, boo
...
 
-	uint id;
-	for (id = 0; id < _num_grf_texts; id++) {
-		if (_grf_text[id].grfid == grfid && _grf_text[id].stringid == stringid) {
-			break;
-		}
-	}
+	auto it = std::find_if(std::begin(_grf_text), std::end(_grf_text), [&grfid, &stringid](const GRFTextEntry &grf_text) { return grf_text.grfid == grfid && grf_text.stringid == stringid; });
+	if (it == std::end(_grf_text)) {
+		/* Too many strings allocated, return empty */
```
TrueBrain: missing dot at end of sentence :)

src/newgrf_text.cpp
```diff
@@ -561,27 +560,21 @@ StringID AddGRFString(uint32_t grfid, uint16_t stringid, byte langid_to_add, boo
...
+		if (_grf_text.size() == TAB_SIZE_NEWGRF) return STR_EMPTY;
 
-	/* Too many strings allocated, return empty */
-	if (id == lengthof(_grf_text)) return STR_EMPTY;
+		/* We didn't find our stringid and grfid in the list, allocate a new id */
```
TrueBrain: missing dot at end of sentence :)

TrueBrain (approved):

PR #11259, opened 2023-09-03
Codechange: Use begin/end of nwidget parts instead of begin/length.

PeterN:
## Motivation / Problem

Attempt to avoid using `lengthof()` as a legacy macro, and improve/simplify widget construction.

Constructing widgets needs to keep track of how many widget parts are in the array, how many have been consumed, and then uses this information to increase the start and reduce the length.

## Description

Instead, pass the beginning and ending of each nwidget part array. Unlike remaining length, the end pointer never changes. We only have to increment the begin pointer, and ensure it is less than the end pointer.

Similar principle to our C-style string handling using end instead of length.

## Checklist for review

### DIFF
--- a/src/widget_type.h
+++ b/src/widget_type.h
@@ -1296,8 +1296,8 @@ static inline NWidgetPart NWidgetFunction(NWidgetFunctionType *func_ptr)
 	return part;
 }
 
-NWidgetContainer *MakeNWidgets(const NWidgetPart *parts, int count, int *biggest_index, NWidgetContainer *container);
-NWidgetContainer *MakeWindowNWidgetTree(const NWidgetPart *parts, int count, int *biggest_index, NWidgetStacked **shade_select);
+NWidgetContainer *MakeNWidgets(const NWidgetPart *nwid_begin, const NWidgetPart *nwid_end, int *biggest_index, NWidgetContainer *container);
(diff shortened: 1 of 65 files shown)


src/widget_type.h
```diff
@@ -1296,8 +1296,8 @@ static inline NWidgetPart NWidgetFunction(NWidgetFunctionType *func_ptr)
 	return part;
 }
 
-NWidgetContainer *MakeNWidgets(const NWidgetPart *parts, int count, int *biggest_index, NWidgetContainer *container);
-NWidgetContainer *MakeWindowNWidgetTree(const NWidgetPart *parts, int count, int *biggest_index, NWidgetStacked **shade_select);
+NWidgetContainer *MakeNWidgets(const NWidgetPart *nwid_begin, const NWidgetPart *nwid_end, int *biggest_index, NWidgetContainer *container);
+NWidgetContainer *MakeWindowNWidgetTree(const NWidgetPart *parts, const NWidgetPart *nwid_end, int *biggest_index, NWidgetStacked **shade_select);
```
TrueBrain: nitpick: you renamed `parts` to `nwid_begin` in the source-file, but not in the header-file :)

TrueBrain (approved):

PR #11262, opened 2023-09-04
Change: Allow vehicles in the build-vehicle-window to be filtered based on their cargo type

Kuhnovic:
## Motivation / Problem

With large train sets it can be annoying to switch back and forth between filters to find vehicles that support the cargo you are interested in. I often find myself switching between different sorting methods / filters which can get tedious.

## Description

Thanks to @2TallTyler we have a filter textbox in the build vehicle window. It already supported filtering by name and newgrf text, and cargo types have now been added as well. This includes cargo types that can be refitted to.

## Checklist for review

### DIFF
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1366,7 +1366,7 @@ struct BuildVehicleWindow : Window {
 		return CargoAndEngineFilter(&item, filter_type);
 	}
 
-	/** Filter by name and NewGRF extra text */
+	/** Filter by name, cargo type and NewGRF extra text */
 	bool FilterByText(const Engine *e)
 	{
 		/* Do not filter if the filter text box is empty */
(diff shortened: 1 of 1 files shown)


2TallTyler:
I respectfully disagree with this change. My original filter intentionally omits cargo refits, for two reasons:

1. We already have a dropdown cargo filter
2. If the filter is too broad, the results won't be as useful

That said, one limitation of the existing dropdown that is not shared by a text filter is the ability to filter multiple cargos, say for a vehicle that refits between two cargos in its orders. I'm not convinced this is a strong enough reason to do this, but welcome counterarguments. 😃

andythenorth:
Wonder if we can / should(?) have a dropdown widget which will also filter? Usually this would be done via a combo box widget 😄

src/build_vehicle_gui.cpp
```diff
@@ -1377,6 +1377,12 @@ struct BuildVehicleWindow : Window {
 		SetDParam(0, PackEngineNameDParam(e->index, EngineNameContext::PurchaseList));
 		this->string_filter.AddLine(GetString(STR_ENGINE_NAME));
 
+		/* Filter by cargo type. This includes cargo types that the engine can be refitted to. */
+		const CargoTypes refit_mask = GetUnionOfArticulatedRefitMasks(e->index, true) &_standard_cargo_mask;
```
PeterN, line 1381: Missing space between `&` and `_standard_cargo_mask`.

PeterN:
I agree with 2TallTyler, this would cause "collisions" in matches due to partially matching by cargo type when not intended -- and there would be two ways of filtering for cargo, only the original way would be able to perform a refit.

And if the drop down list did support multiple selections, that would prevent the automatic refit from working.

Kuhnovic:
Alright, I'm all for a democratic approach to this, no hard feeling ;). If the majority thinks that this is a bad idea then let's just not do it. I'll keep the PR open for a bit of additional discussion, if there is none then let's just close this PR :)

frosch123:
In office software you usually have a list of "recent files". But I cannot think of a nice way to add this to filters :)
* Sort the dropdown by recently-used instead of alphabetic. => awful
* Add 3 most-recent cargos to top of dropdown, then rest alphabetic. => weird
* Add 3 small buttons with cargo icon of most recent cargo filters next to dropdown. => maybe?

2TallTyler:
> I'll keep the PR open for a bit of additional discussion, if there is none then let's just close this PR :)

There hasn't been any additional discussion, so let's close this. 🙂

PR #11263, opened 2023-09-05
Feature: Procedural tree growth

ldpl:
## Motivation / Problem

Trees add a lot of entropy to the map array drastically increasing the size of the save file. Which is a big problem in multiplayer and other cases of sending or archiving saves. To a point that sometimes players have to remove trees entirely.
Continues what I started in #10018.

## Description

This PR almost seamlessly adds a new procedural tree growth mode and tree types. When using it tree growth and counter are decided by pre-computed random arrays and are not stored on the map. Freshly planted trees still use old growth algorithm only switching to the procedural growth then their state happens to align with it.

Compared to master my tests show about 15-40% (depending on settings) reduction of savegame size with tree spreading disabled and about 5% reduction with spreading enabled (on 4k maps). I didn't make precise performance measurements but quick tests show no noticeable difference on the framerate counter.

Savegame bump is required as it extends the growth value with new option but additional conversion is not needed as all old values are still perfectly valid and will switch to procedural growth as the game progresses.

Python script I used to pre-compute growth arrays:
[gen_trees.py.txt](https://github.com/OpenTTD/OpenTTD/files/12528049/gen_trees.py.txt)

## Limitations

- I still had to store 1 random bit in the map array to indicate the tile loop state and synchonize it with drawing routine. Removing it is probably possible but will require some meddling with the tile loop logic and so I decided against trying to do that in this PR.
- Changing tree growth setting changes the trees. I could technically bake in the procedural state on change but decided it's not worth complicating the code.
- New trees still have a lot of randomness greatly reducing the effect of this PR when spreading is enabled.

## Checklist for review

    * ~~ai_changelog.hpp, game_changelog.hpp need updating.~~

### DIFF
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -352,4 +352,20 @@ constexpr int RoundDivSU(int a, uint b)
 
 uint32_t IntSqrt(uint32_t num);
 
+
+/**
+ * Simple 32 bit to 32 bit hash.
+ * From MurmurHash3.
+ */
+inline uint32_t SimpleHash32(uint32_t h)
+{
+	h ^= h >> 16;
+	h *= 0x85ebca6b;
+	h ^= h >> 13;
+	h *= 0xc2b2ae35;
+	h ^= h >> 16;
+
+	return h;
+}
+
 #endif /* MATH_FUNC_HPP */
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -54,9 +54,15 @@ enum ExtraTreePlacement {
 /** Determines when to consider building more trees. */
 byte _trees_tick_ctr;
 
+static const uint TREE_GROWTH_PROCEDURAL = 7;  ///< Magic value for the tree growth stage that signifies that tree growth is determined procedurally, not stored in the map array.
+
 static const uint16_t DEFAULT_TREE_STEPS = 1000;             ///< Default number of attempts for placing trees.
 static const uint16_t DEFAULT_RAINFOREST_TREE_STEPS = 15000; ///< Default number of attempts for placing extra trees at rainforest in tropic.
 static const uint16_t EDITOR_TREE_DIV = 5;                   ///< Game editor tree generation divisor factor.
+
+/* Pre-randomized tree growth cycles (for procedural growth). */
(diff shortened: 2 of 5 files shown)


src/core/math_func.hpp
```diff
@@ -371,4 +371,20 @@ static inline int DivAwayFromZero(int a, uint b)
 
 uint32_t IntSqrt(uint32_t num);
 
+
+/**
+ * Simple 32 bit to 32 bit hash.
+ * From MurmurHash3.
+ */
+inline uint32_t SimpleHash32(uint32_t h)
```
2TallTyler, line 360: Needs doxygen for parameter and return value.

src/tree_cmd.cpp
```diff
@@ -134,24 +140,47 @@ static void PlantTreesOnTile(TileIndex tile, TreeType treetype, uint count, uint
...
 		case LT_TROPIC:
 			switch (GetTropicZone(tile)) {
-				case TROPICZONE_NORMAL:  return (TreeType)(seed * TREE_COUNT_SUB_TROPICAL / 256 + TREE_SUB_TROPICAL);
-				case TROPICZONE_DESERT:  return (TreeType)((seed > 12) ? TREE_INVALID : TREE_CACTUS);
-				default:                 return (TreeType)(seed * TREE_COUNT_RAINFOREST / 256 + TREE_RAINFOREST);
+				case TROPICZONE_NORMAL:  return TREE_RANDOM_TROPIC_NORMAL;
+				case TROPICZONE_DESERT:  return ((seed > 12) ? TREE_INVALID : TREE_CACTUS);
+				default:                 return TREE_RANDOM_TROPIC_RAINFOREST;
 			}
+		default: return TREE_RANDOM_TOYLAND;
```
2TallTyler: While you're here, I think this would make more sense as an explicit case with a default NOT_REACHED() to follow most other OpenTTD code.

src/tree_cmd.cpp
```diff
@@ -134,24 +140,47 @@ static void PlantTreesOnTile(TileIndex tile, TreeType treetype, uint count, uint
...
+		case TREE_RANDOM_TROPIC_RAINFOREST:
+			return (TreeType)(tile_hash % TREE_COUNT_RAINFOREST + TREE_RAINFOREST);
+		case TREE_RANDOM_TOYLAND:
+			return (TreeType)(tile_hash % TREE_COUNT_TOYLAND + TREE_TOYLAND);
 		default:
-			return (TreeType)(seed * TREE_COUNT_TOYLAND / 256 + TREE_TOYLAND);
+			return type;
 	}
 }
-
```
2TallTyler, line 154: I think you still want this newline between functions. 😉

src/tree_cmd.cpp
```diff
@@ -515,6 +544,20 @@ struct TreeListEnt : PalSpriteID {
 	byte x, y;
 };
 
+static bool CanPlantExtraTrees(TileIndex tile)
+{
+	return ((_settings_game.game_creation.landscape == LT_TROPIC && GetTropicZone(tile) == TROPICZONE_RAINFOREST) ?
```
2TallTyler: Can this long ternary be split up?

src/tree_cmd.cpp
```diff
@@ -515,6 +544,20 @@ struct TreeListEnt : PalSpriteID {
...
+		(_settings_game.construction.extra_tree_placement == ETP_SPREAD_ALL || _settings_game.construction.extra_tree_placement == ETP_SPREAD_RAINFOREST) :
+		_settings_game.construction.extra_tree_placement == ETP_SPREAD_ALL);
+}
+
+uint8_t PickRandomGrowth(uint32_t tile_hash, bool can_plant_extra, uint32_t cycle) {
```
2TallTyler: Needs doxygen.

src/tree_cmd.cpp
```diff
@@ -515,6 +544,20 @@ struct TreeListEnt : PalSpriteID {
 	byte x, y;
 };
 
+static bool CanPlantExtraTrees(TileIndex tile)
```
2TallTyler, line 551: Needs doxygen.

src/tree_map.h
```diff
@@ -29,7 +29,14 @@ enum TreeType {
 	TREE_CACTUS       = 0x1B, ///< a cactus for the 'desert part' on a sub-tropical map
 	TREE_SUB_TROPICAL = 0x1C, ///< tree on a sub-tropical map, non-rainforest, non-desert
 	TREE_TOYLAND      = 0x20, ///< tree on a toyland map
+	TREE_RANDOM_TEMPERATE = 0xFA,  // TODO
```
2TallTyler: TODO needs doing 😄

src/tree_cmd.cpp
```diff
@@ -726,86 +782,126 @@ static void TileLoop_Trees(TileIndex tile)
...
+	}
 
-		default:
-			AddTreeGrowth(tile, 1);
-			break;
+	if (spread) {
+		TreeType treetype = GetProceduralTreeType(tile, tile_hash);
+
+		/* We never spread while changing the current tile so it's ok to overwrite and only mark the new tile dirty later. */
+		tile += TileOffsByDir((Direction)(Random() & 7));
```
2TallTyler: Should this 7 be TREE_GROWTH_PROCEDURAL?

src/tree_cmd.cpp
```diff
@@ -726,86 +782,126 @@ static void TileLoop_Trees(TileIndex tile)
...
+	}
+
+	if (clear) {
+		/* All trees died, change type into MP_CLEAR */
+		switch (GetTreeGround(tile)) {
+			case TREE_GROUND_SHORE: MakeShore(tile); break;
+			case TREE_GROUND_GRASS: MakeClear(tile, CLEAR_GRASS, GetTreeDensity(tile)); break;
+			case TREE_GROUND_ROUGH: MakeClear(tile, CLEAR_ROUGH, 3); break;
+			case TREE_GROUND_ROUGH_SNOW: {
+				uint density = GetTreeDensity(tile);
```
2TallTyler, line 888: This could be moved up and used for `case TREE_GROUND_GRASS`

src/tree_cmd.cpp
```diff
@@ -726,86 +782,126 @@ static void TileLoop_Trees(TileIndex tile)
...
-						}
-						break;
+			case 6: // fina)l stage of tree destruction
+				if (!can_plant_extra) {
+					/* if trees can't spread just plant a new one to prevent deforestation */
+					SetTreeGrowth(tile, 0);
+				} else if (count > 1) {
+					/* more than one tree, delete it */
+					AddTreeCount(tile, -1);
+					SetTreeGrowth(tile, 3);
```
2TallTyler, line 868: I see a lot of 3s in reference to growth amounts (and a lot of magic numbers generally, but I understand it's hard to avoid these in math-heavy functions without making the code less readable). Could you look for opportunities to make consts or enums where they make sense, as a favor to reviewers and anyone working in this area in the future?

src/tree_cmd.cpp
```diff
@@ -726,86 +782,126 @@ static void TileLoop_Trees(TileIndex tile)
...
-					default: // snow or desert
-						if (_settings_game.game_creation.landscape == LT_TROPIC) {
-							MakeClear(tile, CLEAR_DESERT, GetTreeDensity(tile));
-						} else {
-							uint density = GetTreeDensity(tile);
-							MakeClear(tile, CLEAR_GRASS, 3);
-							MakeSnow(tile, density);
-						}
-						break;
+			case 6: // fina)l stage of tree destruction
```
2TallTyler: `final` typo?

src/tree_cmd.cpp
```diff
@@ -54,10 +54,16 @@ enum ExtraTreePlacement {
 /** Determines when to consider building more trees. */
 byte _trees_tick_ctr;
 
+static const uint TREE_GROWTH_PROCEDURAL = 7;  ///< Magic value to signal that tree growth is determined procedurally.
```
2TallTyler: I'm going to need a bit more clarification on this magic number. The comment reads like it's a setting value, but you didn't add to the settings... maybe explain what it's compared with to help me understand?

src/tree_cmd.cpp
ldpl: I didn't write this function, just moved it around but whatever.

src/tree_cmd.cpp
ldpl: 3 here is just the number of growth stage, there is nothing particularly special about it except that it lasts a bit longer than other on average. And other stages are even less special so that would be a pretty silly enum like GROWING_1, GROWING_2, GROWING_3...

src/tree_cmd.cpp
ldpl: Moving it up would mean an extra map array access even in cases when density is not used. Unless you believe compiler to be smart enough to optimize it back, I sure don't :p.
Also, this is not a new code, I just moved it around.

src/tree_cmd.cpp
ldpl: No, that's direction, Again, not a new code. Though I guess this one is pretty straightforward to improve a bit.

src/tree_cmd.cpp
ldpl: Idk how to explain it any better. There are 3 bits in the map array for growth stage. Values 0-6 were used before and still are. Value 7 was unused, now it means that growth stage is not stored.
By no means should this be configurable.

PeterN:
"Feature" is overstating this somewhat. It's somewhere between "Codechange" and "Change".

src/tree_cmd.cpp
```diff
@@ -515,6 +546,33 @@ struct TreeListEnt : PalSpriteID {
...
+}
+
+/**
+ * Picks predefined tree growth state based on input parameters.
+ * @param tile_hash The hash of the tile.
+ * @param can_plant_extra Whether new trees can be planted (result of CanPlantExtraTrees call).
+ * @param counter Time counter.
+ * @return tree growth state (bits 0..1 - tree count, 2..4 - growth stage).
+ */
+static uint8_t PickRandomGrowth(uint32_t tile_hash, bool can_plant_extra, uint32_t counter) {
```
PeterN: Newline before `{`

src/tree_cmd.cpp
```diff
@@ -134,24 +140,49 @@ static void PlantTreesOnTile(TileIndex tile, TreeType treetype, uint count, uint
...
+		default: NOT_REACHED();
+	}
+}
 
+/**
+ * Get a random TreeType for the given tile based on a given seed
+ *
+ * This function returns a random TreeType which can be placed on the given tile.
+ * The seed for randomness must be less or equal 256, use #GB on the value of Random()
+ * to get such a value.
```
PeterN: What is this comment referring to?

src/tree_cmd.cpp
```diff
@@ -166,7 +197,7 @@ static void PlaceTree(TileIndex tile, uint32_t r)
 	TreeType tree = GetRandomTreeType(tile, GB(r, 24, 8));
 
 	if (tree != TREE_INVALID) {
-		PlantTreesOnTile(tile, tree, GB(r, 22, 2), std::min<byte>(GB(r, 16, 3), 6));
+		PlantTreesOnTile(tile, tree, 0, 7);
```
PeterN: What is this 7?

ldpl:
@PeterN Idk, this is essentially adding a new functionality though, even if from used POV it's for the most part only visible as savegame size reduction. How about "Add:"?

Also, this one annoying bit of randomness left is hurting my ocd, I'll look into removing it.

PR #11264, opened 2023-09-05
Add: Use http_proxy environment variable for curl

aldot:
## Motivation / Problem

The content is attempted to be downloaded bypassing my proxy.

With the tweak, the proxy is used proper:
```
$ echo $http_proxy
http://proxy.loc:3128
$ strace -ff -econnect -- ./openttd -x -d net=9
...
dbg: [grf] No MD5 checksum specified for: game18.mid (in /home/b/.local/share/openttd/content_download/baseset/descent_soundtrack-1.2/descent.obm)
dbg: [grf] No MD5 checksum specified for: game18.mid (in /home/b/.local/share/openttd/content_download/baseset/descent_soundtrack-1.2/descent.obm)
...
dbg: [net] Network revision name: 20230904-master-m9d2920e9c5
dbg: [net] Using proxy from environment http_proxy=http://proxy.loc:3128
* processing: https://binaries.openttd.org/bananas
strace: Process 157709 attached
[pid 157709] +++ exited with 0 +++
*   Trying 192.168.3.46:3128...
[pid 157699] connect(45, {sa_family=AF_INET, sin_port=htons(3128), sin_addr=inet_addr("192.168.3.46")}, 16) = -1 EINPROGRESS (Operation now in progress)
* Connected to proxy.loc (192.168.3.46) port 3128
* CONNECT tunnel: HTTP/1.1 negotiated
* allocate connect buffer
* Establish HTTP proxy tunnel to binaries.openttd.org:443
> CONNECT binaries.openttd.org:443 HTTP/1.1
Host: binaries.openttd.org:443
User-Agent: OpenTTD/20230904-master-m9d2920e9c5
Proxy-Connection: Keep-Alive

< HTTP/1.1 200 Connection established

```

**# It looks odd that the descent soundtrack file seems to be checked twice?**
Is this something which could be optimized to just a single check? Entirely different thing though, didn't look..

## Description

curl supports using a proxy set in the environment variable "http_proxy". Let's use that for the content downloading IFF requested.

Older versions of OpenTTD (both UNIX and Windows) just hung when attempting to download content when the network BOFH did not allow direct HTTP connections but the windows system proxy was setup properly, since OpenTTD ignored the proxy settings.

## Limitations

Due to lack of a windows system, i did not check if the windows version handles HTTP(S) traffic properly by using the system's proxy settings.

## Checklist for review

### DIFF
--- a/src/network/core/http_curl.cpp
+++ b/src/network/core/http_curl.cpp
@@ -180,6 +180,16 @@ void HttpThread()
 		});
 		curl_easy_setopt(curl, CURLOPT_XFERINFODATA, request->callback);
 
+		/* Pass the proxy to curl explicitly. */
+		if (const char *proxy = std::getenv("http_proxy")) {
+			if (std::getenv("no_proxy") != NULL) {
+				Debug(net, 3, "no_proxy environment variable set, ignoring http_proxy");
+			} else {
+				Debug(net, 3, "Using proxy from environment http_proxy={}", proxy);
+				curl_easy_setopt(curl, CURLOPT_PROXY, proxy);
+			}
+		}
+
 		/* Perform the request. */
 		CURLcode res = curl_easy_perform(curl);
 


2TallTyler:
Thanks for your contribution!

Please go back and fill out the PR template, to explain why you think this change should be made. That helps us review it more effectively. 😃

TrueBrain:
According to libcurls documentation it should automatically pick up on all the env-vars. If that doesn't work for you, that is something to look into.

We will need to do some tests to see what is going on. But ideally we avoid doing this ourselves, as it is a bit annoying to do properly.

For example, there is not only http_proxy, but also https_proxy, which have some interaction with each other. There is also all_proxy, and all three can be lower or uppercase. Yeah .. ideally we leave this to libcurl to handle properly :)

So yes, the first question is: doesn't this already work, and why is this PR needed? Did we forget something else when setting up libcurl?

aldot:
## Motivation / Problem

The content is attempted to be downloaded bypassing my proxy.

With the tweak, the proxy is used proper:
```
$ echo $http_proxy
http://proxy.loc:3128
$ strace -ff -econnect -- ./openttd -x -d net=9
...
dbg: [grf] No MD5 checksum specified for: game18.mid (in /home/b/.local/share/openttd/content_download/baseset/descent_soundtrack-1.2/descent.obm)
dbg: [grf] No MD5 checksum specified for: game18.mid (in /home/b/.local/share/openttd/content_download/baseset/descent_soundtrack-1.2/descent.obm)
...
dbg: [net] Network revision name: 20230904-master-m9d2920e9c5
dbg: [net] Using proxy from environment http_proxy=http://proxy.loc:3128
* processing: https://binaries.openttd.org/bananas
strace: Process 157709 attached
[pid 157709] +++ exited with 0 +++
*   Trying 192.168.3.46:3128...
[pid 157699] connect(45, {sa_family=AF_INET, sin_port=htons(3128), sin_addr=inet_addr("192.168.3.46")}, 16) = -1 EINPROGRESS (Operation now in progress)
* Connected to proxy.loc (192.168.3.46) port 3128
* CONNECT tunnel: HTTP/1.1 negotiated
* allocate connect buffer
* Establish HTTP proxy tunnel to binaries.openttd.org:443
> CONNECT binaries.openttd.org:443 HTTP/1.1
Host: binaries.openttd.org:443
User-Agent: OpenTTD/20230904-master-m9d2920e9c5
Proxy-Connection: Keep-Alive

< HTTP/1.1 200 Connection established

```

**# It looks odd that the descent soundtrack file seems to be checked twice?**
Is this something which could be optimized to just a single check? Entirely different thing though, didn't look..


## Description

curl supports using a proxy set in the environment variable "http_proxy". Let's use that for the content downloading IFF requested.

Older versions of OpenTTD (both UNIX and Windows) just hung when attempting to download content when the network BOFH did not allow direct HTTP connections but the windows system proxy was setup properly, since OpenTTD ignored the proxy settings.

## Limitations

Due to lack of a windows system, i did not check if the windows version handles HTTP(S) traffic properly by using the system's proxy settings.


## Checklist for review

TrueBrain:
Just as a mental note, a summary of what has been talked about on Discord:

- `https_proxy` and `all_proxy` work as expected without this PR.
- `http_proxy` shouldn't work, as in the nightly (and once released, in 14.0 and later) all our connections are https, and no longer http.
- `no_proxy` as introduced in this PR is not according to spec; it should be a list of domains for which to ignore the proxy variables. But libcurl all handles this already, and doesn't require action from us.

All in all, this PR isn't actually needed, and OP was mostly confused with http vs https (as 13.4 and older use http, and only nightly/14.0+ use https).

It was discussed whether we should document the env-variables somewhere; on one hand the argument is: that would be more verbose. On the other hand: but these are default assumptions one can have of an application, and are part of libraries we use. We also don't document env-vars of other libraries (like SDL, fontconfig, etc etc).

I have no opinion on the latter; I am just marking down what has been talked about :)

aldot:
Ok, so i retract this patch.
Thanks for your patience (and switching to HTTPS  for > 13.4 :)

Let me mention that 13.4 on windows with correct proxy settings is unable to download any content if the windows box is disallowed to directly connect to the peer HTTP servers.

Looking forward to a fresh windows build ;-)
thanks, again, and cheers,

PR #11272, opened 2023-09-08
Codechange: Use std::array/std::unique_ptr for Persistent/TemporaryStorageArray.

PeterN:
## Motivation / Problem

Reducing use of C-style memory allocation and handling, memcpy, memset, sizeof, etc.

## Description

Replace the arrays/pointers in Persistent/TemporaryStorageArray with std::arrays instead. This allows for default blank initialization, and use of assignment operators to copy/reset the data. The constructors and destructors are no longer necessary.

Additionally the method "ResetToZero()" was never called, so is removed.

## Checklist for review

### DIFF
--- a/src/newgrf_storage.h
+++ b/src/newgrf_storage.h
@@ -131,17 +131,12 @@ struct PersistentStorageArray : BasePersistentStorageArray {
  */
 template <typename TYPE, uint SIZE>
 struct TemporaryStorageArray {
-	TYPE storage[SIZE]; ///< Memory to for the storage array
-	uint16_t init[SIZE];  ///< Storage has been assigned, if this equals 'init_key'.
-	uint16_t init_key;    ///< Magic key to 'init'.
+	using StorageType = std::array<TYPE, SIZE>;
+	using StorageInitType = std::array<uint16_t, SIZE>;
 
-	/** Simply construct the array */
-	TemporaryStorageArray()
-	{
-		memset(this->storage, 0, sizeof(this->storage)); // not exactly needed, but makes code analysers happy
-		memset(this->init, 0, sizeof(this->init));
-		this->init_key = 1;
-	}
+	StorageType storage{}; ///< Memory for the storage array
+	StorageInitType init{}; ///< Storage has been assigned, if this equals 'init_key'.
+	uint16_t init_key{1}; ///< Magic key to 'init'.
 
 	/**
 	 * Stores some value at a given position.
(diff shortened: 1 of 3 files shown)


src/newgrf_storage.h
```diff
@@ -65,26 +65,10 @@ struct BasePersistentStorageArray {
...
-		free(this->prev_storage);
-	}
-
-	/** Resets all values to zero. */
-	void ResetToZero()
-	{
-		memset(this->storage, 0, sizeof(this->storage));
-	}
+	StorageType storage{}; ///< Memory for the storage array
+	std::unique_ptr<StorageType> prev_storage{}; ///< Memory to store "old" states so we can revert them on the performance of test cases for commands etc.
```
TrueBrain: "on the performance of test cases" .. it took me a while to parse that sentence :D Maybe reword it a bit to read a bit more natural? "so we can revert after running commands in test-mode" or something? Dunno, but "performance" threw me off here.

src/newgrf_storage.h
```diff
@@ -104,10 +88,9 @@ struct PersistentStorageArray : BasePersistentStorageArray {
 
 		/* We do not have made a backup; lets do so */
 		if (AreChangesPersistent()) {
-			assert(this->prev_storage == nullptr);
-		} else if (this->prev_storage == nullptr) {
-			this->prev_storage = MallocT<TYPE>(SIZE);
-			memcpy(this->prev_storage, this->storage, sizeof(this->storage));
+			assert(!this->prev_storage);
+		} else if (!this->prev_storage) {
+			this->prev_storage = std::make_unique<StorageType>(this->storage);
```
TrueBrain, line 93: Why exactly a unique-pointer? If I read this correct, you kinda use it as an `std::optional`? Ofc you can use an unique-ptr as an optional, but initial I was thinking you actually need the pointer. But you just want to know if the object is set yes/no?

src/newgrf_storage.h
PeterN: Not gonna lie, I never read all that comment as I didn't write it either...

src/newgrf_storage.h
PeterN: As far as I know, using std::optional<> would (may?) mean prev_storage always takes up space even if it's not used. Given this 1KB each, using std::unique_ptr<> instead made sense to me.

You are right about not specifically needing the semantics of a pointer though.

TrueBrain (approved):

PR #11274, opened 2023-09-08
Fix: Cargo payment can no longer be exploited by "teleportation".

ldpl:
## Motivation / Problem

Cargo payment is calculated between station signs, but it can freely move within (and with) the station enabling various exploits.
- Station sign manipulation.
Screenshot from 2023-09-09 00-57-21
- Station spreading abuse
Screenshot from 2023-09-09 00-49-56
- Station walking abuse. (Move station by building and removing parts. Cargo keeps the source location.)

## Description

This pr makes it so that only cargo movement with the vehicle is counted toward the payment. It is achieved by storing source position as a direction(vector) when on station. Thus it is the same regardless of where it was loaded in the vehicle and keeps the same accumulated distance. When in the vehicle it's still stored as location so is not affected by vehicle path. Conversion between location and direction is happening when loading and unloading using position of the vehicle as a reference point. Distance to the industry is not taken into account and station sign position no longer has any effect.

Another way to view the accumulated value is as a sum of vectors for each vehicle route (between loading and unloading tile). And it's calculated along the vehicle route: `start1 - end1 + start2 - end2 +...` so each `+ start` makes location(point), each `- end` makes direction(vector).

Should not have any significant effect on regular games that don't abuse station spread.

Save I used to test exploits: [teleport-exploit.zip](https://github.com/OpenTTD/OpenTTD/files/12567985/teleport-exploit.zip)

## Limitations

- There is still some minor abusing possible by reversing long trains in place.
- Cheesing the distance between industries is still possible by placing stations further than they need to be.

## Checklist for review

    * ~~ai_changelog.hpp, game_changelog.hpp need updating.~~

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1096,7 +1096,7 @@ static uint DeliverGoodsToIndustry(const Station *st, CargoID cargo_type, uint n
  * @param num_pieces amount of cargo delivered
  * @param cargo_type the type of cargo that is delivered
  * @param dest Station the cargo has been unloaded
- * @param source_tile The origin of the cargo for distance calculation
+ * @param distance The distance cargo travelled
  * @param periods_in_transit Travel time in cargo aging periods
  * @param company The company delivering the cargo
  * @param src_type Type of source of cargo (industry, town, headquarters)
@@ -1104,7 +1104,7 @@ static uint DeliverGoodsToIndustry(const Station *st, CargoID cargo_type, uint n
  * @return Revenue for delivering cargo
  * @note The cargo is just added to the stockpile of the industry. It is due to the caller to trigger the industry's production machinery
  */
-static Money DeliverGoods(int num_pieces, CargoID cargo_type, StationID dest, TileIndex source_tile, uint16_t periods_in_transit, Company *company, SourceType src_type, SourceID src)
+static Money DeliverGoods(int num_pieces, CargoID cargo_type, StationID dest, uint distance, uint16_t periods_in_transit, Company *company, SourceType src_type, SourceID src)
 {
 	assert(num_pieces > 0);
 
@@ -1131,7 +1131,7 @@ static Money DeliverGoods(int num_pieces, CargoID cargo_type, StationID dest, Ti
 	st->town->received[cs->town_effect].new_act += accepted_total;
 
 	/* Determine profit */
-	Money profit = GetTransportedGoodsIncome(accepted_total, DistanceManhattan(source_tile, st->xy), periods_in_transit, cargo_type);
+	Money profit = GetTransportedGoodsIncome(accepted_total, distance, periods_in_transit, cargo_type);
 
 	/* Update the cargo monitor. */
 	AddCargoDelivery(cargo_type, company->index, accepted_total - accepted_ind, src_type, src, st);
@@ -1225,15 +1225,16 @@ CargoPayment::~CargoPayment()
  * Handle payment for final delivery of the given cargo packet.
  * @param cp The cargo packet to pay for.
  * @param count The number of packets to pay for.
+ * @param distance The distance cargo travelled.
  */
-void CargoPayment::PayFinalDelivery(const CargoPacket *cp, uint count)
+void CargoPayment::PayFinalDelivery(const CargoPacket *cp, uint count, uint distance)
 {
 	if (this->owner == nullptr) {
 		this->owner = Company::Get(this->front->owner);
 	}
 
 	/* Handle end of route payment */
-	Money profit = DeliverGoods(count, this->ct, this->current_station, cp->GetSourceXY(), cp->GetPeriodsInTransit(), this->owner, cp->GetSourceType(), cp->GetSourceID());
+	Money profit = DeliverGoods(count, this->ct, this->current_station, distance, cp->GetPeriodsInTransit(), this->owner, cp->GetSourceType(), cp->GetSourceID());
 	this->route_profit += profit;
 
 	/* The vehicle's profit is whatever route profit there is minus feeder shares. */
@@ -1244,15 +1245,16 @@ void CargoPayment::PayFinalDelivery(const CargoPacket *cp, uint count)
  * Handle payment for transfer of the given cargo packet.
  * @param cp The cargo packet to pay for; actual payment won't be made!.
  * @param count The number of packets to pay for.
+ * @param location TileIndex if the point packet is unloaded at.
  * @return The amount of money paid for the transfer.
  */
-Money CargoPayment::PayTransfer(const CargoPacket *cp, uint count)
+Money CargoPayment::PayTransfer(const CargoPacket *cp, uint count, TileIndex location)
 {
 	Money profit = -cp->GetFeederShare(count) + GetTransportedGoodsIncome(
 			count,
 			/* pay transfer vehicle the difference between the payment for the journey from
 			 * the source to the current point, and the sum of the previous transfer payments */
-			DistanceManhattan(cp->GetSourceXY(), Station::Get(this->current_station)->xy),
+			DistanceManhattan(cp->GetMovement(), location),
 			cp->GetPeriodsInTransit(),
 			this->ct);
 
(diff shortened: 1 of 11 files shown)


TrueBrain:
So basically this is a bug report + solution at once, and it seems to address more than one problem. That makes it pretty hard to untangle, and what is actually meant. Additionally, the code isn't all that clear, `CargoPacket` in general already isn't all that clear, so it will be a bit hit&miss about what the intentions are etc.

First off, code-wise: we already have an union that differentiates between different states the CargoPacket can be in. And it is incredibly complex to understand what is actually meant. By using a single variable that can contain two different units depending on where the cargo is, you increase that complexity by a lot. I would suggest we look for a solution where we do not store different units in a single variable, as the technical debt only increases with this approach.
There are a few more bits and ends, but this is my main concern with this patch. As mentioned on Discord, `direction` seems like a bad name, but that is bikeshedding.

But mostly, this PR made me wonder: what do we actually pay for? And this is a more zoomed-out question that most likely nobody wants to answer, but how I see it there are two ways to look at this:

Method one:
Cargo payment is done based on the amount of tiles the cargo moved looking at it from industry to industry. So not how it is currently, from the first station that accepted it to the last station where it is delivered, but really from industry to industry. The tile that produced the cargo all the way to the tile that consumed the cargo. This would make for an easy to explain, easy to implement payment system.

The downside: any form of "teleportation" is pure profit: the faster you get cargo from A to B, the more money you get. One can argue that is just how the game is. It was never meant to be balanced like this. And as such, competitive MP is just not something we are aiming for. Maybe we need an extension system to allow for different modes. But the game this game is based on, meant to do it like that. 
A way to address this could be to not allow "teleportation" of cargo like this, either by charging for it, putting a penalty on it, etc.

This does resolve the issue of "where the station sign is" completely.

Method two:
Cargo payment is done based on the tiles traveled in a vehicle. Basically, where you pickup the cargo and drop it off creates a vector. You add up all the vectors from source to destination. You get paid for the distance of this final vector.

But, this is also one that is very difficult to explain to any end-user. As most of them wouldn't even know what a mathematical vector is, and just see their payment being different from what they expect.

This does resolve the issue of "teleportation" completely.

---

Now for this PR, if I look at what I think it tries to address, and basically it is a bit of everything. Let's put down some notation, as otherwise we will go mental. Say, we live in a 1D world.

- There is a producing industry at tile 100.
- There is a consuming industry at tile 200.

Now method one argues payment should be done based on `[100, 200]`, meaning 101 tiles (as with the `[]` notation, it means the values are inclusive).

Now we add stations:

Say we have a station at tile `[110, 120]`, and one at `[180, 190]`. The label for the first is at 110, for the second at 190.

When I move cargo from `[120, 180]`, I get paid for `[110, 190]`. Currently, this is getting paid for a trip of 61 tiles for the price of 81 tiles. If we add in a transfer station, it gets a bit more weird:

We add a transfer station from `[125, 175]`. You get paid for `[110, 190]`, but the actual travel is from `[120, 125]` and `[175, 180]`. This creates a lot of money really quick.

----

What I think this PR tries to address are several things:

First, use the actual tile where loading / unloading is done, instead of the location of the station sign. This on its own seems like a solution for the station-sign exploits.

But there is more in this PR. It also seems to track a delta of how much cargo "teleportation" happened in stations. And here it gets a bit vague .. it seems this PR tracks how much station-teleportation has been done. Or at least, possibly unintentionally, seems to be the effect. To use the words from the description:

`start1 - end1 + start2 - end2 ..`

So once cargo arrives at a station, all that is left is the amount of tiles it moved, and it is ignored how much the cargo "teleported" in a station. This seems to be an attempt to address the issue, and pay you for less tiles moved than the actual Manhattan distance.

If I understand this correctly, the "location" is basically a fictive point on the map that ignores all the "teleportation" happening. Not saying that is a bad thing, just trying to grasp what the intention of this PR is.

So given I understood it all correctly (otherwise, blame yourself), to hint for better wording and variables etc, basically you create a vector of each leg. Add the vectors together, and get the Manhattan from start to finish, and you have the "distance traveled". When a vehicle leaves a station you store the location. When it arrives you calculate the vector based on that, and add the vectors together. All this talk about "conversion" makes it unneeded more complex and hard to understand. You calculate a vector. You don't "convert" two tiles into a vector.

Either way, although I understand the need to address all these kind of exploits, I seriously wonder if this is the correct approach to the problem. It is rather complex to explain to any user how this exactly works; given the amount of text I need to wrap my head around it, is an example of the complexity.
I could argue this is just part of station-spread. Set it too high, and you create these possibilities. Keep it low, and the amount of abuse one can do with them is limited. I am finding it difficult to balance the complexity this introduces versus the benefit it delivers.

Mostly, what worries me, is that the Manhattan distance of this final vector is a rather arbitrary distance traveled. Sure, it is how much you effectively moved towards the goal, but it also feels really really wrong.

----

tldr:

This PR is overly complex, and contains several changes. I would suggest to:

- Make a separate PR to change that we use loading / unloading tile instead of station-sign-location for cargo processing. This solves both station-walking abuse as station-sign-location abuse.
- I would drop the vector-based approach to calculate distance traveled, as I don't see a sane way to explain it to the user. I do not think it is worth address the station-spread abuse, as that is literally a setting and part of the game.

ldpl:
> If I understand this correctly, the "location" is basically a fictive point on the map that ignores all the "teleportation" happening. Not saying that is a bad thing, just trying to grasp what the intention of this PR is.

Yes, you seem to understand it correctly. It's possible to view point as a vector from (0, 0) and so drop the "convertion" part of the explanation as math stays the same anyway. And with some static casts I can probably use TileIndexDiff for it instead of TileIndex.

> Make a separate PR to change that we use loading / unloading tile instead of station-sign-location for cargo processing. This solves both station-walking abuse as station-sign-location abuse.

It doesn't solve station walking. Also if you only use endpoints for final payment you need to somehow extend CargoPacket to differentiate them from transfers.

> I would drop the vector-based approach to calculate distance traveled, as I don't see a sane way to explain it to the user. 

For most users explanaition is the same as before: payment ~= distance, no zig-zags. If they know about teleportation, well, then they'll know it no longer works. And those few who want to know exacly can wrap their head around vectors. Cargo is only aging in vehicles, so calculate distance in vehicles as well.

> I do not think it is worth address the station-spread abuse, as that is literally a setting and part of the game. 
It basically makes high station spread unusable in a competitive game. And setting it low doesn't really eliminate the issue completely as

src/economy.cpp
```diff
@@ -1244,15 +1245,16 @@ void CargoPayment::PayFinalDelivery(const CargoPacket *cp, uint count)
...
  */
-Money CargoPayment::PayTransfer(const CargoPacket *cp, uint count)
+Money CargoPayment::PayTransfer(const CargoPacket *cp, uint count, TileIndex location)
 {
 	Money profit = -cp->GetFeederShare(count) + GetTransportedGoodsIncome(
 			count,
 			/* pay transfer vehicle the difference between the payment for the journey from
 			 * the source to the current point, and the sum of the previous transfer payments */
-			DistanceManhattan(cp->GetSourceXY(), Station::Get(this->current_station)->xy),
+			DistanceManhattan(cp->GetMovement(), Station::Get(this->current_station)->xy),
```
TrueBrain: Guess you also meant to change the station-xy here, as otherwise transfers are still paid in the old system.

TrueBrain (commented):
Fun fact, this PR introduces a new exploit:

image

Lot of money for 2 stations that are next to each other :) In general, what goes wrong here, is to use a TileIndex for TileIndexDiffC information. That simply isn't going to work.

src/economy.cpp
ldpl: Right, I think I fixed it.

ldpl:
> Fun fact, this PR introduces a new exploit:

Can you explain what exactly did you do here? Maybe attach a save. Screenshot doesn't really tell much, I can do the same in master.

> In general, what goes wrong here, is to use a TileIndex for TileIndexDiffC information. That simply isn't going to work.

TileIndexDiffC is not necessary here, regular TileIndexDiff will do and I actually used that at first but #11190 replaced TileIndexDiff usage with just TileIndex. And I don't understand how to get TileIndexDiff from two tiles anymore, everything I try just gives some weird type errors.

TrueBrain:
Going to close this PR in favour of #11281 and #11283, as they both together result in this PR (give or take a few other variable names, and a few bug-fixes).

#11281 and #11283 contain a lot more in-code comments and a more elaborate PR description, hopefully helping the review of the next developer. It took me a good two days to figure out what this PR exactly did, and that it didn't impact any parts negatively. It also required more than a few PRs to clean up CargoPacket, as it suffered a bit of years of changes, meaning more than a few things were no longer what they said they were.

Basically, although this PR mostly does what it says it does, it took a bit more code-wise effort to make it maintainable for the years to come. This sometimes happens with these borrowed-deep piece of code OpenTTD has.

In result, the underlying idea of this PR, to vectorize the movement of cargo, gives a nice balance between punishing cheaters while not impacting normal players all too much. It is a novel idea, worth implementing.

PR #11276, opened 2023-09-09
Codechange: remove loaded_at_xy from CargoPacket as it was unused

TrueBrain:
## Motivation / Problem

In 2fee030a26996a167adbf28f96fca8c4fd749005 the last usage of `LoadedAtXY` was removed, and unbeknown to the author, left for a whole strain of code that can be removed from our codebase.

## Description

Removing `LoadedAtXY` leads to removing a lot of `loaded_at_xy` (or sometimes called `load_place`) throughout the code. This also cleans up a very odd `union` in `CargoPacket`, which should make everyone really happy.

## Limitations

This PR doesn't need a savegame bump, although for simplicity I did do it anyway:

- Savegames are header-based: if you present a savegame which still has `loaded_at_xy`, it is just ignored by this PR.
- If you load a savegame made after this PR in an older OpenTTD, it will not find `loaded_at_xy`, and as such not set it. Sadly, it was never explicitly set, so it will get some random value. But as it isn't used, that doesn't matter.

The only place that is a bit of an argument, is the compatibility loader. As before we had headers, we did need to tell the system when fields needed to be skipped. So that could be pinned to the version that changed the savegame to contain headers.

But, the second reason above was enough reason for me to say: no, let's just bump the savegame. Versions are cheap, and we prevent someone in a few weeks / months go: when I load this game in an older OpenTTD, I have uninitialized memory!!! or something. Simplicity comes first. But you can disagree on this, so let me know if you do :)

## Checklist for review

### DIFF
--- a/src/cargoaction.h
+++ b/src/cargoaction.h
@@ -77,19 +77,17 @@ public:
 
 /** Action of loading cargo from a station onto a vehicle. */
 class CargoLoad : public CargoMovement<StationCargoList, VehicleCargoList> {
-protected:
-	TileIndex load_place; ///< TileIndex to be saved in the packets' loaded_at_xy.
 public:
-	CargoLoad(StationCargoList *source, VehicleCargoList *destination, uint max_move, TileIndex load_place) :
-			CargoMovement<StationCargoList, VehicleCargoList>(source, destination, max_move), load_place(load_place) {}
+	CargoLoad(StationCargoList *source, VehicleCargoList *destination, uint max_move) :
+			CargoMovement<StationCargoList, VehicleCargoList>(source, destination, max_move) {}
 	bool operator()(CargoPacket *cp);
 };
 
 /** Action of reserving cargo from a station to be loaded onto a vehicle. */
 class CargoReservation : public CargoLoad {
 public:
-	CargoReservation(StationCargoList *source, VehicleCargoList *destination, uint max_move, TileIndex load_place) :
-			CargoLoad(source, destination, max_move, load_place) {}
+	CargoReservation(StationCargoList *source, VehicleCargoList *destination, uint max_move) :
+			CargoLoad(source, destination, max_move) {}
 	bool operator()(CargoPacket *cp);
 };
 
--- a/src/saveload/vehicle_sl.cpp
+++ b/src/saveload/vehicle_sl.cpp
@@ -577,7 +577,6 @@ static uint32_t _cargo_source_xy;
 static uint16_t _cargo_count;
 static uint16_t _cargo_paid_for;
 static Money  _cargo_feeder_share;
-static uint32_t _cargo_loaded_at_xy;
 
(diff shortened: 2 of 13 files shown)


src/cargoaction.h
```diff
@@ -77,19 +77,17 @@ class CargoTransfer : public CargoMovement<VehicleCargoList, StationCargoList> {
 
 /** Action of loading cargo from a station onto a vehicle. */
 class CargoLoad : public CargoMovement<StationCargoList, VehicleCargoList> {
-protected:
-	TileIndex load_place; ///< TileIndex to be saved in the packets' loaded_at_xy.
 public:
-	CargoLoad(StationCargoList *source, VehicleCargoList *destination, uint max_move, TileIndex load_place) :
-			CargoMovement<StationCargoList, VehicleCargoList>(source, destination, max_move), load_place(load_place) {}
```
ldpl, line 84: Removing loaded_at_xy is better be done after #11274 as it needs load_place here and in CargoReservation.

TrueBrain: Yeah, let's not build these kind of constraints. Who knows when #11274 lands, while this PR is trivially correct.

Even more, it would be good for #11274 to reintroduce these variables, as `load_place` is a weird name. More commonly we use `load_xy` or `load_tile`.

ldpl: I guess #11274 can reintroduce them as `location` like in other actions as it not always load or unload.

src/saveload/vehicle_sl.cpp
```diff
@@ -699,7 +698,6 @@ class SlVehicleCommon : public DefaultSaveLoadHandler<SlVehicleCommon, Vehicle>
 		SLE_CONDVAR(Vehicle, profit_last_year,      SLE_INT64,                   SLV_65, SL_MAX_VERSION),
 		SLEG_CONDVAR("cargo_feeder_share", _cargo_feeder_share, SLE_FILE_I32 | SLE_VAR_I64,  SLV_51,  SLV_65),
 		SLEG_CONDVAR("cargo_feeder_share", _cargo_feeder_share, SLE_INT64,                   SLV_65,  SLV_68),
-		SLEG_CONDVAR("cargo_loaded_at_xy", _cargo_loaded_at_xy, SLE_UINT32,                  SLV_51,  SLV_68),
```
PeterN, line 702: Is this right for loading a [SLV_15..SLV_68) savegame?

TrueBrain: See `compat/vehicle_sl_compat.h` a bit higher:

`SLC_NULL(4, SLV_51, SLV_68),`

I think that should do it. Basically, SLV_68 is from WAY before we added headers to the savegame, so that is handled fully by the compat layer. There it now says to ignore the bytes for that game.

PeterN (approved):

PR #11281, opened 2023-09-10
Fix: base cargo payment on load/unload tile, instead of station sign location

TrueBrain:
## Motivation / Problem

Cargo payment is based on the location of a station, which is the same as the location of the sign of the station. The problem with this approach is that the location of the sign can be manipulated in such way you get a nice advantage when moving cargo around.

This can happen both on the loading side, as on the unloading side. To put it in an image (savegames also attached):

image
image

In both images, the vehicles made a single delivery. As you can see, one of the two made far more money than the other. Why? Well, we have to zoom out a bit:

image
image

These savegames were made with a much higher than normal station-spread value, but this just to show the effect better. On the default value of 12, it is already noticable: you can get 2x 12 tiles "free" movement, for which you get paid.

The Pull Request #11274 set out to address this and other problems, but the PR is very hard to read / review. So while looking into what is actually going on, I decided to roll my own PR. Initially I did request the author of #11274 to make this PR, but I have seen no sign of this happening. And well .. sometimes it is just easier to not bother, and do it yourself. Mostly as doing it yourself means a bit better understanding of what is going on.

[Sign-payment-unload.zip](https://github.com/OpenTTD/OpenTTD/files/12568193/Sign-payment-unload.zip)
[Sign-payment-load.zip](https://github.com/OpenTTD/OpenTTD/files/12568194/Sign-payment-load.zip)

PS: just to be clear, this fixes the issue where station location is used for payment. It does not address teleportation at all. And yes, #11274 can be applied on top of this PR to mitigate that issue. I deliberately left that out of this PR, to ease up review. This PR is basically a part of #11274 (although I made some other choices to increase readability and understandably of the code; for example, I went with `current_tile` instead of `location`, as the former is used a lot more in our code than the latter.).

PPS: I go here from the assumption people deliberately were abusing this mechanism, but of course the inverse is also true: there might be people that got paid less than they were due, because the station sign was closer to the destination. In those cases, this PR also fixes their game, giving them more money for their routes.

## Description

There are two fixes in this PR, and as such, two commits. But I will squash this into master, as from a user perspective it fixes the same thing. In other words: the commits are purely to ease up review.

The first commit fixes the issue for the destination. This is very easy, as all we need is the actual location we are dropping off the cargo, instead of the station location. To make things a bit more readable and understandable, I decided to move the `GetDistance` inside the CargoPacket, meaning `source_xy` no longer has to be announced to the outside world. This makes future additions, like #11274 etc, a bit easier.

The second commit fixes the issue for the source. This is less trivial, as when a CargoPacket is created, it is moved to the station. But at that moment we don't know what tile will be used for loading. As such, on creation of a CargoPacket the source tile is now kept as INVALID_TILE. And on first load, the actual tile used for loading is set. After this the value is never changed again, meaning we can track distance as normal.

## Limitations

Initially I assumed having different `source_xy` for a single station would increase the amount of CargoPackets drastically. But after some testing with various of scenarios and savegames, I can only conclude that this isn't really noticeable. On very big games there is a slight increase (of about 5%), but not a huge amount (which I expected). For example, one of the openttdcoop games uses ~950 cargopackets before this PR, and ~1000 after.
I think this is due to cargo aging (but I haven't verified that): when the first piece of cargo is loaded, it also ages. So when the next piece of cargo is loaded, it can't be merged. So having another `source_xy` doesn't actually matter.

No savegame conversions are needed; current cargo will just use their current source_xy, which might still be based on the station location. But new cargo that is generated will have the fixes, so slowly, over the next few game-months, the old cargo is replaced by the new, and payment will be fixed. This PR (deliberately, as it is non-trivial code that might break) takes no attempt to retroactively fix the issue for older savegames. That said, in theory, one could walk all existing CargoPackets, and see if they are still in their first station, and reset source_xy. If someone wants to, I can make it happen. But I rather say: let's give older savegames the pleasure of still getting higher payment till new cargo is generated.

## Checklist for review

### DIFF
--- a/src/cargoaction.cpp
+++ b/src/cargoaction.cpp
@@ -107,7 +107,7 @@ bool CargoDelivery::operator()(CargoPacket *cp)
 {
 	uint remove = this->Preprocess(cp);
 	this->source->RemoveFromMeta(cp, VehicleCargoList::MTA_DELIVER, remove);
-	this->payment->PayFinalDelivery(cp, remove);
+	this->payment->PayFinalDelivery(cp, remove, this->current_tile);
 	return this->Postprocess(cp, remove);
 }
 
--- a/src/cargoaction.cpp
+++ b/src/cargoaction.cpp
@@ -120,6 +120,7 @@ bool CargoLoad::operator()(CargoPacket *cp)
 {
 	CargoPacket *cp_new = this->Preprocess(cp);
 	if (cp_new == nullptr) return false;
+	cp_new->SetSourceXY(this->current_tile);
 	this->source->RemoveFromCache(cp_new, cp_new->Count());
 	this->destination->Append(cp_new, VehicleCargoList::MTA_KEEP);
 	return cp_new == cp;
@@ -134,6 +135,7 @@ bool CargoReservation::operator()(CargoPacket *cp)
 {
 	CargoPacket *cp_new = this->Preprocess(cp);
 	if (cp_new == nullptr) return false;
+	cp_new->SetSourceXY(this->current_tile);
 	this->source->reserved_count += cp_new->Count();
 	this->source->RemoveFromCache(cp_new, cp_new->Count());
 	this->destination->Append(cp_new, VehicleCargoList::MTA_LOAD);
--- a/src/cargoaction.h
+++ b/src/cargoaction.h
@@ -38,10 +38,11 @@ public:
 /** Action of final delivery of cargo. */
 class CargoDelivery : public CargoRemoval<VehicleCargoList> {
 protected:
+	TileIndex current_tile; ///< Current tile cargo delivery is happening.
 	CargoPayment *payment; ///< Payment object where payments will be registered.
 public:
-	CargoDelivery(VehicleCargoList *source, uint max_move, CargoPayment *payment) :
-			CargoRemoval<VehicleCargoList>(source, max_move), payment(payment) {}
+	CargoDelivery(VehicleCargoList *source, uint max_move, CargoPayment *payment, TileIndex current_tile) :
+			CargoRemoval<VehicleCargoList>(source, max_move), current_tile(current_tile), payment(payment) {}
 	bool operator()(CargoPacket *cp);
 };
 
--- a/src/cargoaction.h
+++ b/src/cargoaction.h
@@ -78,17 +78,19 @@ public:
 
 /** Action of loading cargo from a station onto a vehicle. */
 class CargoLoad : public CargoMovement<StationCargoList, VehicleCargoList> {
+protected:
+	TileIndex current_tile; ///< Current tile cargo loading is happening.
 public:
-	CargoLoad(StationCargoList *source, VehicleCargoList *destination, uint max_move) :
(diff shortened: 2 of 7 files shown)


2TallTyler (commented):
The code looks fine to me, but I wonder if "the tile location of the vehicle when loaded" is the right tile to choose. I expect we'll get some complaints from users who build large stations, and silly discussions about forcing vehicles to use the furthest platform... In any case, the exploit isn't eliminated, just made smaller. Is it possible to store the locations of the industry, house, or industry HQ that generate and receive the cargo instead?

TrueBrain:
> The code looks fine to me, but I wonder if "the tile location of the vehicle when loaded" is the right tile to choose. I expect we'll get some complaints from users who build large stations, and silly discussions about forcing vehicles to use the furthest platform... In any case, the exploit isn't eliminated, just made smaller. Is it possible to store the locations of the industry, house, or industry HQ that generate and receive the cargo instead?

Ghehe, funny you mention that, as that was what I originally wanted to do. But I also realised that would be a complete different way of doing payments than it is currently, and you get in the discussion of: shouldn't we do this in an extension? So I walked back from it, and tried to stay as close as possible to the current way it works.

The other thing that came to mind when writing this, was to always use the front vehicle, so cargo loaded in a vehicle would always be from the same place, not from different station tiles. But in the end, after some playtesting, the difference really isn't noticeable. And the amount of "abuse" possible is really really limited. So I just went with the easier solution as suggested by @ldpl in the original PR.

In the end, I think exploiting how we do payment will always remain a possibility; this, even more in combination with #11283 just reduces the ability for people to make lots of money.

2TallTyler:
To me, getting paid for the distance between producer and consumer is how I'd assume the game works if I didn't know better. We were bold in #10596, so I'd argue that we continue that pattern. 😛

I would happily approve a PR changing payments to industry/house/HQ tiles, if it were opened. But if you're strongly against it, this PR is still an improvement.

TrueBrain:
> To me, getting paid for the distance between producer and consumer is how I'd assume the game works if I didn't know better. We were bold in #10596, so I'd argue that we continue that pattern. 😛
> 
> I would happily approve a PR changing payments to industry/house/HQ tiles, if it were opened. But if you're strongly against it, this PR is still an improvement.

It is about a place and a time for everything. #11283 basically fixes that "teleportation" no longer works, and in a sense, cargo teleports from industry to station-tile, and wouldn't be counted anyway. So such change would be negated by #11283.

Now one could argue if #11283 should happen, or that we should just say: fuck it, you get paid from industry to industry, and that is it. But it does allow for silly exploits where people make tons of money.

So yeah, I think the current approach is, for now at least, the safest. It gives very little buzz, as the change isn't noticeable by people who played by the rules. It removes most of the exploitable surface, making competitive play a bit more realistic. But in long term, we should look into making it possible to have payments done in an extension, so people can pick what economy style they like most.

In short, I had your opinion on Saturday. By now, I am like: let's not tilt too many people by being too radical here, and let's just plug the holes :)

2TallTyler (approved):
Yes, that conversation certainly opens some rabbit-holes about which industry tile "should" generate or consume cargo... Let's just do this for now. 🙂

Issue #10324, opened 2023-01-06
[Bug]: Airports are not the related object / parent of Airport Tiles


PikkaBird:
### Version of OpenTTD

13 rc1

### Expected result

Airports and airport tiles derive much of their spec from industries and industry tiles, except in one important respect; whereas an industry is the related object of its tiles, airport tiles have no related object.

This means that although airports have access to station variables including date of construction and cargo information, it is impossible to use those variables to change the appearance of the airport, as the tiles cannot access them. It also makes reusing tiles for multiple rotations a hassle, as they cannot simply access the rotation variable to figure out where they are in the layout.

Please add airports as the related object of airport tiles.

### Actual result

It's broken yo.

### Steps to reproduce

Try to create an airport newgrf that does anything interesting, and discover you can't.

PR #11282, opened 2023-09-10
Feature: [NewGRF] Related Act2 objects for airports and airport tiles.

michicc:
## Motivation / Problem

NewGRF airports are made out of airport tiles to control on-map appearance, similar to the industry <-> industry tile relation.

Unfortunately, airport tiles don't have a related Act2 object, so it is not possible to reuse tiles between different airports in one GRF. Airports itself don't have a related object either, even though most other similar features (rail stations, industries, towns) refer to the town.

## Description

There's no indication neither in the source code nor in the specs wiki that leaving out the related object was a deliberate design decision.

As such, this PR adds the airport as the related object of the airport tile and the town as the related object of the airport. This mirrors the structure for industries and industry tiles.

While we're at it, also add the missing NewGRF inspection window for airports, which can be reached from the existing inspection window for airport tiles. As the station window is not specific just to airports, I opted not to add a direct inspection button there.

## Limitations

No test GRF was made.

## Checklist for review

### DIFF
--- a/src/table/newgrf_debug_data.h
+++ b/src/table/newgrf_debug_data.h
@@ -494,7 +494,7 @@ static const NICallback _nic_airporttiles[] = {
 
 class NIHAirportTile : public NIHelper {
 	bool IsInspectable(uint index) const override        { return AirportTileSpec::Get(GetAirportGfx(index))->grf_prop.grffile != nullptr; }
-	uint GetParent(uint index) const override            { return GetInspectWindowNumber(GSF_FAKE_TOWNS, Station::GetByTile(index)->town->index); }
+	uint GetParent(uint index) const override            { return GetInspectWindowNumber(GSF_AIRPORTS, GetStationIndex(index)); }
 	const void *GetInstance(uint index)const override    { return nullptr; }
 	const void *GetSpec(uint index) const override       { return AirportTileSpec::Get(GetAirportGfx(index)); }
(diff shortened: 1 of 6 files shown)


src/table/newgrf_debug_data.h
```diff
@@ -680,7 +731,7 @@ static const NIFeature * const _nifeatures[] = {
 	&_nif_industry,     // GSF_INDUSTRIES
 	nullptr,               // GSF_CARGOES (has no "physical" objects)
 	nullptr,               // GSF_SOUNDFX (has no "physical" objects)
-	nullptr,               // GSF_AIRPORTS (feature not implemented)
+	&_nif_airport,      // GSF_AIRPORTS
```
TrueBrain, line 734: What kind of weird alignment did they pick here? You are consistent, I give you that, but this is weird :D (please leave it as it is :P It just triggered my OCD)

JGRennison: It's a side effect of the bulk renaming NULL to nullptr

TrueBrain: Ha! Tnx for that, now my OCD is much less triggered :D

TrueBrain (approved):

PikkaBird:
Does this close #10324? I'll try it out once the next nightly is published, thanks. :D

PR #11283, opened 2023-09-10
Fix: only count distance traveled in vehicles for cargo payment

TrueBrain:
## Motivation / Problem

Given this map:
image

Where most stations are like this:
image

One can abuse the fact that cargo in stations are instantly (and free of charge) moved between tiles. What happens here is this:

- Cargo is picked up from the coal mine
- It is delivered in the first station. The first station is station-spread connected to the second.
- The second station is right next to the first station. So the truck goes from the first to the second station by driving 4 tiles (2 station tiles + 2 road tiles)
- This continues all the way down to close by the destination
- There it is picked up and delivered

Now something really funny happens here: the incoming per vehiclechain is the same as having a single truck drive all the way. As the loading/unloading takes about as long the travel time. But: the throughput increases by a lot. So you can make a lot more money this way .. about ten times as much in this setup.

This PR sets out to address that, and is basically #11274 implemented on top of #11281. It only contains a lot more documentation and description, to make much more clear what is going on. Also, I picked a few things differently. The end result is the same: after this PR, you can no longer misuse the free labour given to you by station workers.

[teleport-profit.zip](https://github.com/OpenTTD/OpenTTD/files/12569699/teleport-profit.zip)

Closes #11274.

## Description

To combat this misuse, #11274 found a clever approach: create a vector of distance moved in vehicles, and use that for payment. This boils down to:

- Get the vector of each leg the cargo traveled
- Add those vectors up
- Get the Manhattan distance of this total vector
- This is your distance traveled.

This works really well to resolve this teleportation. However, there is an edgecase: big networks might take the cargo the wrong way first to bring it back in, using a lot of this station teleportation along the way. It can be that the vector ends up much larger than the actual distance between source and destination. For that reason, the smallest of these two is used: either the (Manhattan) size of the vector, or the Manhattan distance between source and destination.

To cheap out on variables (and struct-size), `cargo_movement` is a vector in station, but it is a virtual point when in a vehicle. As we have to calculate the vector of a leg, which is A - B, we add A to the vector when leaving a station, and remove B when entering a station. That way, `cargo_movement` is only a vector when in a station. When it is in a vehicle, it doesn't actually have a meaning.

The variable `in_vehicle` is introduced purely to run `assert` on. As such, it is only compiled in when `WITH_ASSERT` is enabled. It is not meant to be used otherwise. It is just there to ensure `cargo_movement` is in the right state.

For loading older savegames, choices have to be made. I picked one, but you sure can have another. I am open for that dialog.

Although 5 bytes are added to CargoPacket, the actual size (on a 64-bit machine) isn't changed, as there were 5 bytes free to be used: 27 used. Now 32 used. As we still need access to `source_xy`, so we can calculate the Manhattan distance between source and destination, `cargo_movement` is added as extra field.

## Limitations

None that I know of, although this could use some extra testing etc. I did a lot of testing with all kind of different savegames, and all seem to be good. The attached savegame also shows the teleport problem is solved.

## Checklist for review

### DIFF
--- a/src/cargopacket.h
+++ b/src/cargopacket.h
@@ -39,14 +39,26 @@ extern SaveLoadTable GetCargoPacketDesc();
  */
 struct CargoPacket : CargoPacketPool::PoolItem<&_cargopacket_pool> {
 private:
+	/* A mathematical vector from (0,0). */
+	struct Vector {
+		int16_t x;
+		int16_t y;
+	};
+
 	uint16_t count{0}; ///< The amount of cargo in this packet.
 	uint16_t periods_in_transit{0}; ///< Amount of cargo aging periods this packet has been in transit.
 
 	Money feeder_share{0}; ///< Value of feeder pickup to be paid for on delivery of cargo.
 
 	TileIndex source_xy{INVALID_TILE}; ///< The origin of the cargo.
+	Vector travelled{0, 0}; ///< If cargo is in station: the vector from the unload tile to the source tile. If in vehicle: an intermediate value.
+
 	SourceID source_id{INVALID_SOURCE}; ///< Index of industry/town/HQ, INVALID_SOURCE if unknown/invalid.
 	SourceType source_type{SourceType::Industry}; ///< Type of \c source_id.
+
+#ifdef WITH_ASSERT
+	bool in_vehicle{false}; ///< NOSAVE: Whether this cargo is in a vehicle or not.
+#endif /* WITH_ASSERT */
 
 	StationID first_station{INVALID_STATION}; ///< The station where the cargo came from first.
 	StationID next_hop{INVALID_STATION}; ///< Station where the cargo wants to go next.
@@ -83,22 +95,51 @@ public:
 	}
 
 	/**
-	 * Set the origin of the packet.
+	 * Update for the cargo being loaded on this tile.
 	 *
-	 * Can only be set once.
+	 * When a CargoPacket is created, it is moved to a station. But at that
+	 * moment in time it is not known yet at which tile the cargo will be
+	 * picked up. As this tile is used for payment information, we delay
+	 * setting the source_xy till first pickup, getting a better idea where
+	 * a cargo started from.
 	 *
-	 * When a packet is created, it is moved to a station. But at that moment
-	 * in time it is not known yet at which tile the cargo will be picked up.
-	 * As this tile is used for payment information, we delay setting the
-	 * source_xy till first pickup.
+	 * Further more, we keep track of the amount of tiles the cargo moved
+	 * inside a vehicle. This is used in GetDistance() below.
 	 *
 	 * @param tile Tile the cargo is being picked up from.
 	 */
-	void SetSourceXY(TileIndex tile)
+	void UpdateLoadingTile(TileIndex tile)
 	{
 		if (this->source_xy == INVALID_TILE) {
 			this->source_xy = tile;
 		}
+
+#ifdef WITH_ASSERT
+		assert(!this->in_vehicle);
+		this->in_vehicle = true;
+#endif /* WITH_ASSERT */
+
+		/* We want to calculate the vector from tile-unload to tile-load. As
+		 * we currently only know the latter, add it. When we know where we unload,
+		 * we subtract is, giving us our vector (unload - load). */
+		this->travelled.x += TileX(tile);
+		this->travelled.y += TileY(tile);
+	}
+
+	/**
+	 * Update for the cargo being unloaded on this tile.
+	 *
+	 * @param tile Tile the cargo is being dropped off at.
+	 */
+	void UpdateUnloadingTile(TileIndex tile)
+	{
+#ifdef WITH_ASSERT
+		assert(this->in_vehicle);
+		this->in_vehicle = false;
+#endif /* WITH_ASSERT */
+
+		this->travelled.x -= TileX(tile);
+		this->travelled.y -= TileY(tile);
 	}
 
 	/**
@@ -188,7 +229,36 @@ public:
 	inline uint GetDistance(TileIndex current_tile) const
 	{
 		assert(this->source_xy != INVALID_TILE);
-		return DistanceManhattan(this->source_xy, current_tile);
+#ifdef WITH_ASSERT
+		assert(this->in_vehicle);
+#endif /* WITH_ASSERT */
+
+		/* Distance is always requested when the cargo is still inside the
+		 * vehicle. So first finish the calculation for travelled to
+		 * become a vector. */
+		auto local_travelled = travelled;
+		local_travelled.x -= TileX(current_tile);
+		local_travelled.y -= TileY(current_tile);
+
+		/* Cargo-movement is a vector that indicates how much the cargo has
+		 * actually traveled in a vehicle. This is the distance you get paid
+		 * for. However, one could construct a route where this vector would
+		 * be really long. To not overpay the player, cap out at the distance
+		 * between source and destination.
+		 *
+		 * This way of calculating is to counter people moving cargo for free
+		 * and instantly in stations, where you deliver it in one part of the
+		 * station and pick it up in another. By using the actual distance
+		 * traveled in a vehicle, using this trick doesn't give you more money.
+		 *
+		 * However, especially in large networks with large transfer station,
+		 * etc, one could actually make the route a lot longer. In that case,
+		 * use the actual distance between source and destination.
+		 */
+
+		uint distance_travelled = abs(local_travelled.x) + abs(local_travelled.y);
+		uint distance_source_dest = DistanceManhattan(this->source_xy, current_tile);
+		return std::min(distance_travelled, distance_source_dest);
 	}
 
 	/**
@@ -427,7 +497,7 @@ public:
 
 	template<MoveToAction Tfrom, MoveToAction Tto>
 	uint Reassign(uint max_move);
-	uint Return(uint max_move, StationCargoList *dest, StationID next_station);
+	uint Return(uint max_move, StationCargoList *dest, StationID next_station, TileIndex current_tile);
 	uint Unload(uint max_move, StationCargoList *dest, CargoPayment *payment, TileIndex current_tile);
 	uint Shift(uint max_move, VehicleCargoList *dest);
 	uint Truncate(uint max_move = UINT_MAX);
--- a/src/saveload/cargopacket_sl.cpp
+++ b/src/saveload/cargopacket_sl.cpp
@@ -76,6 +76,42 @@
 	if (IsSavegameVersionBefore(SLV_181)) {
 		for (Vehicle *v : Vehicle::Iterate()) v->cargo.KeepAll();
 	}
+
+	/* Before this version, we didn't track how far cargo actually traveled in vehicles. Make best-effort estimates of this. */
+	if (IsSavegameVersionBefore(SLV_CARGO_TRAVELLED)) {
+		/* Update the cargo-traveled in stations as if they arrived from the source tile. */
+		for (Station *st : Station::Iterate()) {
+			for (size_t i = 0; i < NUM_CARGO; i++) {
+				GoodsEntry *ge = &st->goods[i];
+				for (auto it = ge->cargo.Packets()->begin(); it != ge->cargo.Packets()->end(); it++) {
+					for (CargoPacket *cp : it->second) {
+						if (cp->source_xy != INVALID_TILE && cp->source_xy != st->xy) {
+							cp->travelled.x = TileX(cp->source_xy) - TileX(st->xy);
+							cp->travelled.y = TileY(cp->source_xy) - TileY(st->xy);
+						}
+					}
+				}
+			}
+		}
+
+		/* Update the cargo-traveled in vehicles as if they were loaded at the source tile. */
+		for (Vehicle *v : Vehicle::Iterate()) {
+			for (auto it = v->cargo.Packets()->begin(); it != v->cargo.Packets()->end(); it++) {
+				if ((*it)->source_xy != INVALID_TILE) {
+					(*it)->UpdateLoadingTile((*it)->source_xy);
+				}
+			}
+		}
+	}
+
+#ifdef WITH_ASSERT
+	/* in_vehicle is a NOSAVE; it tells if cargo is in a vehicle or not. Restore the value in here. */
+	for (Vehicle *v : Vehicle::Iterate()) {
+		for (auto it = v->cargo.Packets()->begin(); it != v->cargo.Packets()->end(); it++) {
+			(*it)->in_vehicle = true;
+		}
+	}
+#endif /* WITH_ASSERT */
 }
 
 /**
@@ -97,6 +133,8 @@ SaveLoadTable GetCargoPacketDesc()
 		SLE_VAR(CargoPacket, feeder_share,    SLE_INT64),
 		SLE_CONDVAR(CargoPacket, source_type,     SLE_UINT8,  SLV_125, SL_MAX_VERSION),
 		SLE_CONDVAR(CargoPacket, source_id,       SLE_UINT16, SLV_125, SL_MAX_VERSION),
+		SLE_CONDVAR(CargoPacket, travelled.x, SLE_INT16, SLV_CARGO_TRAVELLED, SL_MAX_VERSION),
+		SLE_CONDVAR(CargoPacket, travelled.y, SLE_INT16, SLV_CARGO_TRAVELLED, SL_MAX_VERSION),
 	};
 	return _cargopacket_desc;
 }
--- a/src/cargoaction.cpp
+++ b/src/cargoaction.cpp
@@ -120,7 +120,7 @@ bool CargoLoad::operator()(CargoPacket *cp)
 {
 	CargoPacket *cp_new = this->Preprocess(cp);
 	if (cp_new == nullptr) return false;
-	cp_new->SetSourceXY(this->current_tile);
+	cp_new->UpdateLoadingTile(this->current_tile);
 	this->source->RemoveFromCache(cp_new, cp_new->Count());
 	this->destination->Append(cp_new, VehicleCargoList::MTA_KEEP);
 	return cp_new == cp;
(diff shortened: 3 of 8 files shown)


src/cargopacket.h
```diff
@@ -44,10 +44,16 @@ struct CargoPacket : CargoPacketPool::PoolItem<&_cargopacket_pool> {
 
 	Money feeder_share{0}; ///< Value of feeder pickup to be paid for on delivery of cargo.
 
-	TileIndex source_xy{0}; ///< The origin of the cargo.
+	TileIndex source_xy{INVALID_TILE}; ///< The origin of the cargo.
+	TileIndexDiffC cargo_movement{0, 0}; ///< A vector (from 0,0) of how far the cargo got actually transported in a vehicle. Is a vector when in station. Is a temporary variable when in vehicle.
```
ldpl: `cargo_` prefix is redundant imo as it a member of CargoPacket. You don't have `cargo_source_id` or `cargo_count`, so why `cargo_movement`?

src/saveload/afterload.cpp
```diff
@@ -3241,6 +3241,50 @@ bool AfterLoadGame()
...
+					/* We don't know where the train was last, so just assume we were picking it up at the source. */
+					(*it)->UpdateLoadingTile(first_station->xy);
+				} else {
+					auto first_station = Station::Get(first_station_id);
+					auto last_station = Station::Get(last_station_id);
+
+					/* Act like we loaded it at the first station and unloaded at the last. */
+					(*it)->UpdateLoadingTile(first_station->xy);
+					(*it)->UpdateUnloadingTile(last_station->xy);
+					(*it)->UpdateLoadingTile(last_station->xy);
```
ldpl: Especially noticeable here, two calls with last_station->xy just cancel out each other. (see the next comment, github places them in weird order, at least for me)

src/saveload/afterload.cpp
```diff
@@ -3241,6 +3241,50 @@ bool AfterLoadGame()
 		}
 	}
 
+	if (IsSavegameVersionBefore(SLV_CARGO_MOVEMENT)) {
+		/* Find a reasonable source and travel distance for cargo in vehicles. */
```
ldpl: Why do all that logic to try (and possibly fail) to restore source location from station ids when you already have it in `source_xy`?

src/saveload/afterload.cpp
```diff
@@ -3241,6 +3241,50 @@ bool AfterLoadGame()
...
+				}
+			}
+		}
+
+		for (Station *st : Station::Iterate()) {
+			for (size_t i = 0; i < NUM_CARGO; i++) {
+				GoodsEntry *ge = &st->goods[i];
+				for (auto it = ge->cargo.Packets()->begin(); it != ge->cargo.Packets()->end(); it++) {
+					for (CargoPacket *cp : it->second) {
+						cp->AfterLoadBeforeCargoMovement();
```
ldpl: AfterLoadBeforeCargoMovement is a copy of this whole loop, you're basically running the whole conversion quadratically. You just need to update `cp->cargo_movement` here.

src/saveload/cargopacket_sl.cpp
```diff
@@ -17,6 +17,24 @@
 
 #include "../safeguards.h"
 
+/**
+ * Savegame conversion for version before cargo_movement was added.
+ */
+/* static */ void CargoPacket::AfterLoadBeforeCargoMovement()
+{
+	for (Station *st : Station::Iterate()) {
```
ldpl: here, same loop as in afterload

ldpl (commented):
While I'm still not convinced that extending travel distance with spreading has any practical exploits as you still need to run vehicle the whole way and having transfers makes it much worse that running directly it's still have some niche uses in theory. So fixing it preemptively is probably fine as that doesn't seem to add that much overhead. On the other hand, typical big mp game seems to have 10-30k cargo packets, adding 40-120k of savegame data (pre commression) is not critical but still somewhat significant, so may be worth postponing this addition until it's clear how impactful this theoretical exploit actually is.

src/cargopacket.h
TrueBrain: In contrast to the 'source' prefixes. But I also didn't really like it, and I would like 'vector" in the name. 'traveled_vector' maybe.

src/saveload/afterload.cpp
TrueBrain: Good point! I found that function a bit later .. I can't access the vector here, so it needs to move. Bit weird code now, need to clean it up a bit.

src/saveload/cargopacket_sl.cpp
```diff
@@ -17,6 +17,24 @@
...
+			GoodsEntry *ge = &st->goods[i];
+			for (auto it = ge->cargo.Packets()->begin(); it != ge->cargo.Packets()->end(); it++) {
+				for (CargoPacket *cp : it->second) {
+					/* For cargo-movement, estimate it as the distance of the current station from the source. */
+					cp->cargo_movement = TileIndexToTileIndexDiffC(cp->source_xy, st->xy);
+				}
+			}
+		}
+	}
+}
```
TrueBrain: Reminder to self: dont cause desyncs, actually save the vector and restore in_vehicle. Silly goose.

src/saveload/afterload.cpp
TrueBrain: `source_xy` is not always a valid value, when loading older savegames. As such, it is good to test some other methods too. But I do see I never test for source_xy, which might be good to see if that is a valid tile first, and try the others after.

ldpl: Weird, but it is currently used for cargo payments so maybe it's fine? Even if you want to fix that it may be better to put that conversion in a version that introduced the issue as it's not really related to the teleport fix. And you need a sane source_xy for payment check anyway.

TrueBrain:
> While I'm still not convinced that extending travel distance with spreading has any practical exploits as you still need to run vehicle the whole way and having transfers makes it much worse that running directly it's still have some niche uses in theory. So fixing it preemptively is probably fine as that doesn't seem to add that much overhead. On the other hand, typical big mp game seems to have 10-30k cargo packets, adding 40-120k of savegame data (pre commression) is not critical but still somewhat significant, so may be worth postponing this addition until it's clear how impactful this theoretical exploit actually is.

That was a bit rambling, but I think I get what you mean. 120KiB on a savegame with 30k cargopackets really is not significant in any terms. And it is not as niche as you make it out. It happens really easily if you have a few proper transfer stations of max size around your map, which you use to navigate cargo (and especially passengers).

So yeah, given it is a very low impact (to address the issue), and keeps things fair both ways, let's go with it. No need to only have fairness work one way :)

(further more, I noticed it really helped with debugging, so it is also pretty nice for that).

src/cargopacket.h
```diff
@@ -39,15 +39,27 @@ extern SaveLoadTable GetCargoPacketDesc();
...
+		int16_t y;
+	};
+
 	uint16_t count{0}; ///< The amount of cargo in this packet.
 	uint16_t periods_in_transit{0}; ///< Amount of cargo aging periods this packet has been in transit.
 
 	Money feeder_share{0}; ///< Value of feeder pickup to be paid for on delivery of cargo.
 
 	TileIndex source_xy{INVALID_TILE}; ///< The origin of the cargo.
+	Vector distance_traveled{0, 0}; ///< A vector (from 0,0) of how far the cargo got actually transported in a vehicle. Is a vector when in station. Is a temporary variable when in vehicle.
```
ldpl: it's not really a distance though. At least, I'd expect "distance travelled" to be a total length of the path taken. Also, British spelling seems to be "travelled".
P.S. "Displacement" seems to be a good word for this.

TrueBrain: Bikeshedding, but yeah. "distance" is only one of the two pieces of information in a vector. Going to drop "distance", so just "travelled" (and sure, as we write colour, I will write travelled).

PeterN:
I suppose you can 'exploit' this by making further-than-necessary stops/platforms, but that's not much different than the current version of making the station signs as far apart as possible...

frosch123:
## int16_t too small?
If I read this right, `CargoPacket::travelled` accumulates the number of tiles travelled (signed). It uses `int16_t` for that.
This will at least not work for the huge-map patches.

How about inverting the accumulation? Sum the distance travelled within stations: `tile_for_loading` - `tile_of_last_unload`.
* The result should be the same: `distance(source, destination) = travel_in_train + travel_in_station`
* Just the numbers are smaller.

## Train reversing
I am not quite sure how reversing trains affect this. Do long trains have to travel less distance than short trains, if they reverse at terminal stations?
Should the "travel distance" exclude distance "travelled" by reversing trains?

## Off-topic
Out of scope for this PR, but I'll write it anyway:
* One could also account for "distance(producer, first-loading-tile)" and "distance(final-unloading-tile, accepter)".
* This would allow adding fees / negative payment / ..., if you build stations far from industries.
* Though town producers/accepters have no specific tile.

TrueBrain:
> I suppose you can 'exploit' this by making further-than-necessary stops/platforms, but that's not much different than the current version of making the station signs as far apart as possible...

Yup.

TrueBrain:
> ## int16_t too small?
> 
> If I read this right, `CargoPacket::travelled` accumulates the number of tiles travelled (signed). It uses `int16_t` for that. This will at least not work for the huge-map patches.
> 
> How about inverting the accumulation? Sum the distance travelled within stations: `tile_for_loading` - `tile_of_last_unload`.
> 
>     * The result should be the same: `destination - source = travel_vector_in_train + travel_vector_in_station`
> 
>     * Just the numbers are smaller.
> 
> 
> ## Train reversing
> 
> I am not quite sure how reversing trains affect this. Do long trains have to travel less distance than short trains, if they reverse at terminal stations? Should the "travel distance" exclude distance "travelled" by reversing trains?
> ## Off-topic
> 
> Out of scope for this PR, but I'll write it anyway:
> 
>     * One could also account for "distance(producer, first-loading-tile)" and "distance(final-unloading-tile, accepter)".
> 
>     * This would allow adding fees / negative payment / ..., if you build stations far from industries.
> 
>     * Though town producers/accepters have no specific tile.

int16_t: I did look at this, but can't remember my conclusion. Os the length of the vector in stations actually the invert of the length of the vector travelled. Will look into this again (and write down the conclusion :P)

Considering reversing a teleport would be the next step I guess. But I would say, not for this PR.

As for your offtopic, yes, I went through the same motions. Also it isn't fully trivial which Station accepted your cargo in the current code .. requires getting some values in other layers then they life now. But I do agree with the idea, and 2TallTyler had similar remarks. But yeah, not this PR :)

ldpl:
> If I read this right, `CargoPacket::travelled` accumulates the number of tiles travelled (signed). It uses `int16_t` for that. This will at least not work for the huge-map patches.
> 
> How about inverting the accumulation? Sum the distance travelled within stations: `tile_for_loading` - `tile_of_last_unload`.

As transfers aren't limited it seems possible to overflow int16 in both travelled and teleported vector cases. But overflowing travelled vector is pointless as you just decrease the distance you get paid for. Not that it matters as it requires transporting stuff far beoyond the map size and it will be capped anyway. Overflowing teleported vector is actually more promising for the exploit as you can get 64 teleported tiles for 1 tile of movement. Though after 1024 transfers it probably doesn't matter either. And considering how much effort it would require to overflow I don't think any additional checks are even necessary here.
Patches that allow more than int16 map side would need to use larger type but don't think that's something vanilla should worry about.

> I am not quite sure how reversing trains affect this. Do long trains have to travel less distance than short trains, if they reverse at terminal stations? Should the "travel distance" exclude distance "travelled" by reversing trains?

As I mentioned in #11274 reversing still gives some "unfair" money but it's less that could be done prior to this with sign manipulation. And as endpoint check prevents accumulating reversing movement endlessly anyway it doesn't seem like any significant exploiting is possible here, there are better ways to use long trains.

>     * One could also account for "distance(producer, first-loading-tile)" and "distance(final-unloading-tile, accepter)".
> 
>     * This would allow adding fees / negative payment / ..., if you build stations far from industries.

Not counting distance between industries and loading tile is a good thing imo, as allows more flexibility with station placement. Gaining some distance with extended station placement is of no concern as usually players aren't forced to use specific industries and can just chose ones that have required distance in between. Same with teleport fees. I just don't see how forcing players into a smaller area would add anything to the game. Station catchment already just a few tiles anyway.

src/saveload/cargopacket_sl.cpp
```diff
@@ -76,6 +76,53 @@
...
+				auto first_station_id = (*it)->first_station;
+				if (first_station_id != INVALID_STATION) {
+					auto first_station = Station::Get(first_station_id);
+					(*it)->UpdateLoadingTile(first_station->xy);
+					continue;
+				}
+
+				/* And some games also don't have a first-station. Act like it loaded at the current tile.
+				 * This is always wrong, but the best we can do. */
+				(*it)->UpdateLoadingTile(v->tile);
```
frosch123: According to some comments in the source `Vehicle::tile` is zero for aircraft in flight. `Vehicle::x_pos` is kind of valid in flight, but may be outside of the map, if aircraft is in landing-pattern of an airport near the map border.

src/cargopacket.h
```diff
@@ -188,7 +229,34 @@ struct CargoPacket : CargoPacketPool::PoolItem<&_cargopacket_pool> {
...
+		 * vehicle. So first finish the calculation for travelled to
+		 * become a vector. */
+		auto local_travelled = travelled;
+		local_travelled.x -= TileX(current_tile);
+		local_travelled.y -= TileY(current_tile);
+
+		/* Cargo-movement is a vector (from 0,0) that indicates how much the
+		 * cargo has actually traveled in a vehicle. This is the distance you
+		 * get paid for. However, one could construct a route where this vector
+		 * would be really long. To not underpay the player, cap out at the
```
frosch123: s/underpay/overpay/?

src/cargopacket.h
```diff
@@ -39,15 +39,27 @@ extern SaveLoadTable GetCargoPacketDesc();
...
+		int16_t y;
+	};
+
 	uint16_t count{0}; ///< The amount of cargo in this packet.
 	uint16_t periods_in_transit{0}; ///< Amount of cargo aging periods this packet has been in transit.
 
 	Money feeder_share{0}; ///< Value of feeder pickup to be paid for on delivery of cargo.
 
 	TileIndex source_xy{INVALID_TILE}; ///< The origin of the cargo.
+	Vector travelled{0, 0}; ///< A vector (from 0,0) of how far the cargo travelled in a vehicle. Is a vector when in station. Is a temporary variable when in vehicle.
```
frosch123: Not sure what you mean with "from 0,0".
I got confused about the orientation of the vector. It is "source - destination", I would have expected the opposite :)
So:
* If inside station: Vector from current location to source tile.
* If inside vehicle: Intermediate value.

src/cargopacket.h
```diff
@@ -188,7 +229,34 @@ struct CargoPacket : CargoPacketPool::PoolItem<&_cargopacket_pool> {
 	inline uint GetDistance(TileIndex current_tile) const
 	{
 		assert(this->source_xy != INVALID_TILE);
-		return DistanceManhattan(this->source_xy, current_tile);
+		assert(this->in_vehicle);
```
frosch123, line 233: Using `in_vehicle` outside of `#ifdef WITH_ASSERT` scares me, even though it is in an `assert`.

src/saveload/cargopacket_sl.cpp
```diff
@@ -76,6 +76,53 @@
...
+			}
+		}
+	}
+
+	/* in_vehicle is a NOSAVE; it tells if cargo is in a vehicle or not. Restore the value in here. */
+	for (Vehicle *v : Vehicle::Iterate()) {
+		for (auto it = v->cargo.Packets()->begin(); it != v->cargo.Packets()->end(); it++) {
+			(*it)->in_vehicle = true;
```
frosch123, line 111: Missing `#ifdef WITH_ASSERT`.

src/cargopacket.h
TrueBrain: We do this in other places too. But I will add an extra guard, no harm in doing do :)

src/cargopacket.h
TrueBrain: The "from 0,0" is from a pedantic discussion that a vector needs an origin. If unspecified it is ofc 0,0, so the pedanticness confused you :D

Ps, the opposite of 0,0 is 0,0 :P what you spotted is the direction is "negative", as in, it goed "the wrong way". All doesnt matter, I will word it better :)

src/saveload/cargopacket_sl.cpp
TrueBrain: Hmm.. for some reason I missed some earlier code, that already restores `source_xy` for old games. And that already uses `Vehicle::tile`. So this has been broken for a long time for aircrafts.

Mind you, this is only a problem for really old (think: TTO-old) games. The profit they make in their first leg will be weird compared to what it should pay. I am not actually sure it is worth writing a lot of code to prevent that, and maybe fail. I cannot actually test any of it :P

The reason I did do another cascade here of checking if the `source_xy` and `first_station` are valid, as I am not actually sure if anything between `SLV_44` and now broke any of this. But maybe we could just say: fuck that, let's assume `source_xy` is already restored in previous code, as before this PR it is also assumed to be valid.

Edit: Yeah, after reading it through, I just dropped most of this code, and only look at `source_xy`. We don't document our saveload versions enough to see if there are any possibly issues as a result, but I am going to assume everyone did their job right :P

src/saveload/cargopacket_sl.cpp
```diff
@@ -76,6 +76,42 @@
...
+
+		/* Update the cargo-traveled in vehicles as if they were loaded at the source tile. */
+		for (Vehicle *v : Vehicle::Iterate()) {
+			for (auto it = v->cargo.Packets()->begin(); it != v->cargo.Packets()->end(); it++) {
+				if ((*it)->source_xy != INVALID_TILE) {
+					(*it)->UpdateLoadingTile((*it)->source_xy);
+				}
+			}
+		}
+	}
```
TrueBrain, line 105: Did some more testing for the different vehicle-types (RV, Train, Ship and Aircraft). This now all works how I expect it, although I didn't test really old savegames.

What you see is for vehicles that are already on their way, that their payout is identical to before this PR (which is expected). Vehicles that are still loading, have their payout altered. Especially noticable when using transfer stations, ofc.

So all seems to work as advertised :)

frosch123 (approved):

PR #11286, opened 2023-09-11
Codechange: simplify splitting of CargoPacket

TrueBrain:
## Motivation / Problem

While working on CargoPackets, and extending it, I noticed how easy it was to forget to copy the right entries, but also how much work was involved for Split. Small refactor to address that, making the code more clear.

## Description

Introduce a new ctor for backwards compatibility, so no lines throughout the code have to change if something in CargoPacket is being changed.
Additionally, change the Split ctor to just take the second cargo as parameter, and let the ctor deal with copying the right fields.

PS: while at it, remove a note that was no longer true, and confusing.

## Checklist for review

### DIFF
--- a/src/saveload/oldloader_sl.cpp
+++ b/src/saveload/oldloader_sl.cpp
@@ -710,7 +710,7 @@ static bool LoadOldGood(LoadgameState *ls, int num)
 	SB(ge->status, GoodsEntry::GES_ACCEPTANCE, 1, HasBit(_waiting_acceptance, 15));
 	SB(ge->status, GoodsEntry::GES_RATING, 1, _cargo_source != 0xFF);
 	if (GB(_waiting_acceptance, 0, 12) != 0 && CargoPacket::CanAllocateItem()) {
-		ge->cargo.Append(new CargoPacket(GB(_waiting_acceptance, 0, 12), _cargo_periods, (_cargo_source == 0xFF) ? INVALID_STATION : _cargo_source, INVALID_STATION, 0),
+		ge->cargo.Append(new CargoPacket(GB(_waiting_acceptance, 0, 12), _cargo_periods, (_cargo_source == 0xFF) ? INVALID_STATION : _cargo_source, INVALID_TILE, 0),
 				INVALID_STATION);
 	}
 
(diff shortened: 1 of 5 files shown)


src/saveload/oldloader_sl.cpp
```diff
@@ -1353,7 +1353,7 @@ bool LoadOldVehicle(LoadgameState *ls, int num)
 		if (_cargo_count != 0 && CargoPacket::CanAllocateItem()) {
 			StationID source =    (_cargo_source == 0xFF) ? INVALID_STATION : _cargo_source;
 			TileIndex source_xy = (source != INVALID_STATION) ? Station::Get(source)->xy : (TileIndex)0;
-			v->cargo.Append(new CargoPacket(_cargo_count, _cargo_periods, source, INVALID_STATION, source_xy));
+			v->cargo.Append(new CargoPacket(_cargo_count, _cargo_periods, source, source_xy, 0));
```
2TallTyler, line 1356: 0 is feeder share, I think, but we lose the first station? Why?

src/saveload/oldloader_sl.cpp
```diff
@@ -710,7 +710,7 @@ static bool LoadOldGood(LoadgameState *ls, int num)
 	SB(ge->status, GoodsEntry::GES_ACCEPTANCE, 1, HasBit(_waiting_acceptance, 15));
 	SB(ge->status, GoodsEntry::GES_RATING, 1, _cargo_source != 0xFF);
 	if (GB(_waiting_acceptance, 0, 12) != 0 && CargoPacket::CanAllocateItem()) {
-		ge->cargo.Append(new CargoPacket(GB(_waiting_acceptance, 0, 12), _cargo_periods, (_cargo_source == 0xFF) ? INVALID_STATION : _cargo_source, INVALID_STATION, 0),
+		ge->cargo.Append(new CargoPacket(GB(_waiting_acceptance, 0, 12), _cargo_periods, (_cargo_source == 0xFF) ? INVALID_STATION : _cargo_source, INVALID_TILE, 0),
```
2TallTyler, line 713: With the change to INVALID_TILE...is this built upon #11281? Or is this a mistake?

src/saveload/oldloader_sl.cpp
TrueBrain: confusingly, `source` is `first_station`. The `next_station` is removed, as it is always `INVALID_STATION`. It was really confusing code. Now it is simpler:

`count, period, first_station, source_xy, feeder_share`, for all of them. No optional parameters.

src/saveload/oldloader_sl.cpp
TrueBrain: `next_station` is removed from the ctor, and `feeder_share` is no longer optional. So the diff shows a bit weird, and the original was wrong. It set the `source_xy` to zero, instead of `INVALID_TILE`.

This code was really old, and really annoying to understand.

2TallTyler (approved):

PR #11290, opened 2023-09-13
Codechange: vendor the nlohmann-json library

TrueBrain:
## Motivation / Problem

As nlohmann-json is mandatory, it is important that everyone can find and use it easily. Sadly, we see too many reports of people struggling with it. Often related to their build-setup, like MacOS cmake issues, but we also see reports of people who followed the guide, and where it still breaks. All solveable, but at a certain point I just read enough time: "nlohmann-json issues", that I am like: why don't we just vendor the library.

Important to note, the library is a single-header-file.

## Description

Vendor (as in, add it to our repository) the nlohmann-json library version 3.11.2, in `src/3rdparty`. The version is mentioned in the header, so it is easy to check if we need to update or not.

This removes a bunch of customization from different places, so in fact makes things easier for us.

This is in similar idea as `fmt`, `catch2`, `opengl`, `squirrel`, ...

## Checklist for review


TrueBrain:
Oops .. Preview won't work as it uses the `master` workflow (for security reasons). So it still assumes nlohmann patch needs to be applied.

glx22:
Build failures are weird, `3rdparty` should already be in the include path.

TrueBrain:
> Build failures are weird, `3rdparty` should already be in the include path.

Build failures are my mistake; I was lazy, and didn't test it without deinstalling my system variant. Update will arrive soon :)

2TallTyler (approved):
Makes sense to me. I don't know the build process well enough to know if you've missed anything, but CI likes it so 🤷

Issue #11291, opened 2023-09-13
[Bug]: Windows IME: Length unit mismatch using result of ImmGetCompositionString for GCS_CURSORPOS


JGRennison:
### Version of OpenTTD

master

### Expected result

ImmGetCompositionString always returns a value in bytes, the other call sites for ImmGetCompositionString have a comment about this.

### Actual result

At https://github.com/OpenTTD/OpenTTD/blob/master/src/video/win32_v.cpp#L366
`LONG caret_bytes = ImmGetCompositionString(hIMC, GCS_CURSORPOS, nullptr, 0);`

However `caret_bytes` appears to be used as a length in sizeof(wchar_t).
At https://github.com/OpenTTD/OpenTTD/blob/master/src/video/win32_v.cpp#L368
`for (const wchar_t *c = str.c_str(); *c != '\0' && *caret != '\0' && caret_bytes > 0; c++, caret_bytes--) {`

This appears to be a possible cause for this crash: https://github.com/JGRennison/OpenTTD-patches/issues/596

At https://github.com/OpenTTD/OpenTTD/blob/master/src/video/win32_v.cpp#L356-L357
`str` also appears to be unnecessarily large

### Steps to reproduce

I don't have a working IME setup on Windows for reproduction steps

glx22:
It's very easy to reproduce:
* Start a newgame
* Switch keyboard to japanese IME
* Try to rename a town with non latin characters

michicc:
The code in win32_v.cpp seems to be fine, expect that the comment on line 365 is simply misleading. `GCS_CURSORPOS` is one if  the oddball things that are in fact not bytes but chars, see for example https://github.com/microsoft/VCSamples/blob/9e1d4475555b76a17a3568369867f1d7b6cc6126/VC2008Samples/International/IME/IMEEdit.cpp#L386 in combination with https://github.com/microsoft/VCSamples/blob/9e1d4475555b76a17a3568369867f1d7b6cc6126/VC2008Samples/International/IME/IMEEdit.cpp#L826

As Japanese input definitely worked in the past (it was my go to test-IME when writing all this), it very much looks like some oversight in some query string changes or `std::string` conversion.

JGRennison:
As far as I can see the direct issue in the title (ImmGetCompositionString) is not an issue after all, and the original problem is fixed by https://github.com/OpenTTD/OpenTTD/pull/11831

PR #11293, opened 2023-09-13
Fix: marked text was not updated during text deletion

glx22:
## Motivation / Problem
* open the console
* switch keyboard to japanese IME
* enter non latin text
* trigger https://github.com/OpenTTD/OpenTTD/blob/master/src/gfx.cpp#L883

It crashes because `markend` can be outside of the string after text deletion.

## Description
Properly update marked text start and end during text deletion, like it was already done for caret.

## Checklist for review

### DIFF
--- a/src/textbuf.cpp
+++ b/src/textbuf.cpp
@@ -247,14 +247,21 @@ void Textbuf::DeleteText(uint16_t from, uint16_t to, bool update)
 	this->bytes -= to - from;
 	this->chars -= c;
 
-	/* Fixup caret if needed. */
-	if (this->caretpos > from) {
-		if (this->caretpos <= to) {
-			this->caretpos = from;
+	auto fixup = [&](uint16_t &pos) {
+		if (pos <= from) return;
+		if (pos <= to) {
+			pos = from;
 		} else {
-			this->caretpos -= to - from;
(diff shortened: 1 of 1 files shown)


TrueBrain (dismissed):
That is most likely one of the more complex ways I have seen a function do stuff, but given it is a copy/paste of an earlier if, let's go with it :)

glx22:
Now I think I should write a function as the "same" thing is done at least 3 times :)

michicc:
Specifically for #11291, https://gist.github.com/michicc/59a204d99d0321ae4592b4b9d32ad93a would do it, too.

michicc:
Thinking more, I think both is needed, as `Textbuf::InsertString` also has a code path that does not call `DeleteText` but only `DiscardMarkedText`.

glx22:
Reduced copy/paste.

> Thinking more, I think both is needed, as `Textbuf::InsertString` also has a code path that does not call `DeleteText` but only `DiscardMarkedText`.

`DiscardMarkedText` calls `DeleteText`, so I don't think it's needed to touch `Textbuf::DiscardMarkedText`.

michicc:
Eh, yes, temporary brain fart.

michicc (approved):

PR #11294, opened 2023-09-13
Fix: also apply cargo filters on shared groups in vehicle listing

### LINKED ISSUE #11161: [Bug]: Vehicle list cargo filter ignored when grouping by shared orders

TrueBrain:
## Motivation / Problem

#11161 notices that the vehicle listing doesn't filter on cargos when you group by shared orders. This PR set out to address this issue.

Closes #11225 and closes #11161.

## Description

#11225 correctly identifies that the current code doesn't filter on anything but the first vehicle in the list. As solution, they purpose to first build the list, filter it, then group for shared orders. A fine solution. With one thing I don't like, which is open for opinion:

When there is 1 vehicle matching in a shared-order group, I expect to see the whole group, not only the vehicle that matches.

But as this can be open for discussion, what filtering when grouping actually means, I created this PR to have that discussion. I will ping in #11225 to collect thought.

Image for clarification, with this PR, you get this:

image

With #11225, you won't see the Livestock truck.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -199,8 +199,6 @@ void BaseVehicleListWindow::BuildVehicleList()
 			max_unitnumber = std::max<uint>(max_unitnumber, (*it)->unitnumber);
 		}
 		this->unitnumber_digits = CountDigitsForAllocatingSpace(max_unitnumber);
-
-		this->FilterVehicleList();
 	} else {
 		/* Sort by the primary vehicle; we just want all vehicles that share the same orders to form a contiguous range. */
 		std::stable_sort(this->vehicles.begin(), this->vehicles.end(), [](const Vehicle * const &u, const Vehicle * const &v) {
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
I agree with your reasoning in https://github.com/OpenTTD/OpenTTD/pull/11225#issuecomment-1718102698 that the entire collapsed group of vehicles should be shown if at least one fulfills the cargo filter, rather than showing only a subset of vehicles. Like you said, both are correct, but I agree with the approach in this PR. 🙂

PR #11296, opened 2023-09-14
Codechange: Pass by reference and use emplace-at-end for CargoSummary.

PeterN:
## Motivation / Problem

* Passing a pointer but not checking it's null.
* Using emplace_back() and then end() - 1 to get an iterator to a new element.

## Description

* Replace pointer with reference, no need to check null.
* Use emplace(end()) pattern to directly get iterator to new element.

## Checklist for review

### DIFF
--- a/src/train_gui.cpp
+++ b/src/train_gui.cpp
@@ -280,8 +280,7 @@ static void GetCargoSummaryOfArticulatedVehicle(const Train *v, CargoSummary &su
 
 		auto item = std::find(std::begin(summary), std::end(summary), new_item);
 		if (item == std::end(summary)) {
-			summary.emplace_back();
-			item = std::end(summary) - 1;
+			item = summary.emplace(std::end(summary));
 			item->cargo = new_item.cargo;
 			item->subtype = new_item.subtype;
 			item->capacity = 0;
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):

src/train_gui.cpp
```diff
@@ -278,10 +278,9 @@ static void GetCargoSummaryOfArticulatedVehicle(const Train *v, CargoSummary *su
 		new_item.subtype = GetCargoSubtypeText(v);
 		if (!IsValidCargoID(new_item.cargo) && new_item.subtype == STR_EMPTY) continue;
 
-		auto item = std::find(summary->begin(), summary->end(), new_item);
-		if (item == summary->end()) {
-			summary->emplace_back();
-			item = summary->end() - 1;
+		auto item = std::find(std::begin(summary), std::end(summary), new_item);
+		if (item == std::end(summary)) {
+			item = summary.emplace(std::end(summary));
```
LordAro, line 283: emplace_back returns a result since C++17, no real need for emplace + end ?

PeterN: Yes but `item` (misnamed) is an iterator, which is not what `emplace_back()` returns. This is why the original code used `emplace_back()` and then recalculated the iterator from `end() - 1`.

If you only need the object you inserted emplace_back is simpler, but emplace/end is a better pattern for this (suggested to me by glx 2 weeks ago, https://github.com/OpenTTD/OpenTTD/pull/11258#discussion_r1314328624) when you have an iterator.

PR #11305, opened 2023-09-16
Remove: stdarg.h/cstdarg dependency

rubidium42:
## Motivation / Problem

va_arg, va_end, va_list, va_start and friends are simply the C way of doing arbitrary length parameter lists.
In C++ we have templates with parameter packs and fold expressions, that can do the same in fewer lines.

## Description

Replace the last uses of va_arg in `SetWidgetsDisabledState` and `SetWidgetsLoweredState` with parameter packs and fold expressions.
There is one instance where `SetWidgetsDisabledState` was called with a single widget, so it should just use `SetWidgetDisabledState` coincidentally which has swapped parameters.
Then remove the include of stdarg.h/cstdarg from everywhere.

## Limitations

This won't get fully rid of stdarg.h/cstdarg, unless we want to be very aggressive in removing external libraries, as at least zlib, freetype and SDL2 use/include stdarg.h in their headers. As such, we cannot easily add it to the safeguards header to prevent adding it in the future. 
However, I think adding this header in future PRs will cause some frowns, so it's unlikely that it'll return.

## Checklist for review

### DIFF
--- a/src/airport_gui.cpp
+++ b/src/airport_gui.cpp
@@ -107,9 +107,7 @@ struct BuildAirToolbarWindow : Window {
 		if (!gui_scope) return;
 
 		bool can_build = CanBuildVehicleInfrastructure(VEH_AIRCRAFT);
-		this->SetWidgetsDisabledState(!can_build,
-			WID_AT_AIRPORT,
-			WIDGET_LIST_END);
+		this->SetWidgetDisabledState(WID_AT_AIRPORT, !can_build);
 		if (!can_build) {
 			CloseWindowById(WC_BUILD_STATION, TRANSPORT_AIR);
 
--- a/src/3rdparty/squirrel/sqstdlib/sqstdstring.cpp
+++ b/src/3rdparty/squirrel/sqstdlib/sqstdstring.cpp
@@ -1,7 +1,6 @@
 /* see copyright notice in squirrel.h */
 #include <squirrel.h>
 #include <sqstdstring.h>
-#include <stdarg.h>
 
 #define scstrchr strchr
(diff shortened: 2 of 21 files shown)


src/airport_gui.cpp
```diff
@@ -107,9 +107,7 @@ struct BuildAirToolbarWindow : Window {
 		if (!gui_scope) return;
 
 		bool can_build = CanBuildVehicleInfrastructure(VEH_AIRCRAFT);
-		this->SetWidgetsDisabledState(!can_build,
-			WID_AT_AIRPORT,
-			WIDGET_LIST_END);
+		this->SetWidgetDisabledState(WID_AT_AIRPORT, !can_build);
```
michicc, line 110: Swapped arguments?

glx22, line 110: Inverted value and widgets id I think

glx22: Ah no it's a different function, with the reverted order (silly thing)

rubidium42: Oh, forgot to mention/split that... why use the function for multiple widgets when you call it for just one? So I let it use the non-plural version.

TrueBrain (approved):

src/airport_gui.cpp
TrueBrain: As mentioned in chat:

```
void SetWidgetDisabledState(byte widget_index, bool disab_stat)
void SetWidgetsDisabledState(bool disab_stat, Args... widgets)
```

Someone should be hang. But not something to fix in this PR :)

PR #11306, opened 2023-09-17
Enable more warnings and fix them

frosch123:
## Motivation / Problem

Quite a lot of "unused" warnings are disabled, hiding bugs like #11299.

## Description

Enable previously disabled warnings:
* "Wunused-parameter"
* "Wunused-but-set-parameter"
* "Wunused-variable"
* "Wunused-but-set-variable"

There were actually no reports about unused variables, all changes are about parameters.

The commits are split into
* silence warnings: parameters are intentionally unused.
* fix warnings: parameters should be used, but were not.
* remove parameters which are unneeded on both caller and callee side.

For silencing warnings there are different options:
* If the parameter is unused currently, but likely to be used in the future, `[[maybe_unused]]` is added. Examples for this are GUI methods like `UpdateWidgetSize` and `OnClick`: adding/changing widgets may use more parameters.
* If the parameter is only used for assertions, `[[maybe_unused]]` is added.
* Otherwise the parameter name is elided.

## Limitations

I only found settings for gcc.
Where are the warnings for MSVC?

## Checklist for review

### DIFF
--- a/cmake/CompileFlags.cmake
+++ b/cmake/CompileFlags.cmake
@@ -38,7 +38,11 @@ macro(compile_flags)
     set(IS_STABLE_RELEASE "$<AND:$<NOT:$<CONFIG:Debug>>,$<NOT:$<BOOL:${OPTION_USE_ASSERTS}>>>")
 
     if(MSVC)
-        add_compile_options(/W3)
+        add_compile_options(
+            /W3
+            /w34100 # 'identifier' : unreferenced formal parameter
+            /w34189 # 'identifier' : local variable is initialized but not referenced
+        )
         if(MSVC_VERSION GREATER 1929 AND MSVC_VERSION LESS 1937 AND CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
             # Starting with version 19.30 (fixed in version 19.37), there is an optimisation bug, see #9966 for details
             # This flag disables the broken optimisation to work around the bug
@@ -60,8 +64,6 @@ macro(compile_flags)
             -Winit-self
             -Wnon-virtual-dtor
 
-            # Often parameters are unused, which is fine.
-            -Wno-unused-parameter
             # We use 'ABCD' multichar for SaveLoad chunks identifiers
             -Wno-multichar
 
@@ -71,22 +73,6 @@ macro(compile_flags)
             #  compiler all happy.
             -fno-strict-aliasing
         )
-
-        # When we are a stable release (Release build + USE_ASSERTS not set),
-        # assertations are off, which trigger a lot of warnings. We disable
-        # these warnings for these releases.
-        if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-            add_compile_options(
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-variable>"
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-but-set-parameter>"
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-but-set-variable>"
-            )
-        else()
-            add_compile_options(
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-variable>"
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-parameter>"
-            )
-        endif()
 
         # Ninja processes the output so the output from the compiler
         # isn't directly to a terminal; hence, the default is
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -280,11 +280,10 @@ void CommandHelperBase::LogCommandExecution(Commands cmd, StringID err_message, 
 /**
  * Prepare for the test run of a command proc call.
  * @param cmd_flags Command flags.
- * @param tile Tile of command execution.
  * @param[in,out] cur_company Backup of current company at start of command execution.
  * @return True if test run can go ahead, false on error.
  */
-bool CommandHelperBase::InternalExecutePrepTest(CommandFlags cmd_flags, TileIndex tile, Backup<CompanyID> &cur_company)
+bool CommandHelperBase::InternalExecutePrepTest(CommandFlags cmd_flags, TileIndex, Backup<CompanyID> &cur_company)
 {
 	/* Always execute server and spectator commands as spectator */
 	bool exec_as_spectator = (cmd_flags & (CMD_SPECTATOR | CMD_SERVER)) != 0;
@@ -314,7 +313,7 @@ bool CommandHelperBase::InternalExecutePrepTest(CommandFlags cmd_flags, TileInde
  * @param[in,out] cur_company Backup of current company at start of command execution.
  * @return True if test run can go ahead, false on error.
  */
-std::tuple<bool, bool, bool> CommandHelperBase::InternalExecuteValidateTestAndPrepExec(CommandCost &res, CommandFlags cmd_flags, bool estimate_only, bool network_command, Backup<CompanyID> &cur_company)
+std::tuple<bool, bool, bool> CommandHelperBase::InternalExecuteValidateTestAndPrepExec(CommandCost &res, CommandFlags cmd_flags, bool estimate_only, bool network_command, [[maybe_unused]] Backup<CompanyID> &cur_company)
 {
 	BasePersistentStorageArray::SwitchMode(PSM_LEAVE_TESTMODE);
 	SetTownRatingTestMode(false);
@@ -354,7 +353,7 @@ std::tuple<bool, bool, bool> CommandHelperBase::InternalExecuteValidateTestAndPr
  * @param[in,out] cur_company Backup of current company at start of command execution.
  * @return Final command result.
  */
-CommandCost CommandHelperBase::InternalExecuteProcessResult(Commands cmd, CommandFlags cmd_flags, const CommandCost &res_test, const CommandCost &res_exec, Money extra_cash, TileIndex tile, Backup<CompanyID> &cur_company)
+CommandCost CommandHelperBase::InternalExecuteProcessResult(Commands cmd, CommandFlags cmd_flags, [[maybe_unused]] const CommandCost &res_test, const CommandCost &res_exec, Money extra_cash, TileIndex tile, Backup<CompanyID> &cur_company)
 {
 	BasePersistentStorageArray::SwitchMode(PSM_LEAVE_COMMAND);
 
--- a/src/command_func.h
+++ b/src/command_func.h
@@ -257,7 +257,7 @@ public:
 protected:
 	/** Helper to process a single ClientID argument. */
 	template <class T>
-	static inline void SetClientIdHelper(T &data)
+	static inline void SetClientIdHelper([[maybe_unused]] T &data)
 	{
 		if constexpr (std::is_same_v<ClientID, T>) {
 			if (data == INVALID_CLIENT_ID) data = CLIENT_ID_SERVER;
@@ -333,7 +333,7 @@ protected:
 
 	/** Helper to process a single ClientID argument. */
 	template <class T>
-	static inline bool ClientIdIsSet(T &data)
+	static inline bool ClientIdIsSet([[maybe_unused]] T &data)
 	{
 		if constexpr (std::is_same_v<ClientID, T>) {
 			return data != INVALID_CLIENT_ID;
@@ -350,7 +350,7 @@ protected:
 	}
 
 	template<class Ttuple>
-	static inline Money ExtractAdditionalMoney(Ttuple &values)
+	static inline Money ExtractAdditionalMoney([[maybe_unused]] Ttuple &values)
 	{
 		if constexpr (std::is_same_v<std::tuple_element_t<1, Tret>, Money>) {
 			return std::get<1>(values);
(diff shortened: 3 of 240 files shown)


glx22:
For MSVC it's line 41 in CompileFlags.cmake (the `/W3` enabling warnings up to level 3), we could try `/W4` for more warnings (including https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100 and https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4189) or `/Wall` to enable warnings up to level 4 plus https://learn.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default

glx22:
Just tested master with `/W4` and I get 20211 warnings:
* the expected https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100
* a lot of https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127
* one https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4201 in a header (so it's reported many times)
* some https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4245
* many https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4310 in headers (they are actually level 4 even if the page title says level3)
* 3 https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4389
* some https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4456
* many https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4458
* 2 https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4611
* 3 https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4701
* some https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4702
* 7 https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4706
* 2 C4815 not even documented

Maybe it would be better to just use `/W3nnnn` to change the level of the warnings we want from 4 to 3 and stay with `/W3`. We can also silence unwanted warnings with `/Wdnnnn` if we go to `/W4`

cmake/CompileFlags.cmake
```diff
@@ -72,22 +70,6 @@ macro(compile_flags)
...
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-but-set-parameter>"
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-but-set-variable>"
-            )
-        else()
-            add_compile_options(
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-variable>"
-                "$<${IS_STABLE_RELEASE}:-Wno-unused-parameter>"
-            )
-        endif()
-
```
TrueBrain, line 90: We don't actually have a mechanic to test releases without asserts before the actual release, and I guess this is why they exist. These warnings would happen so late in our release process, that it is a lot of effort to correct for them.

This change kinda implies we should add to the CI a run that has asserts disabled, I guess. Without that, this is a tricky change (as warnings can go unnoticed for a long time).

TrueBrain: https://github.com/OpenTTD/OpenTTD/pull/11309

Would address that concern.

glx22:
For MSVC, `/W4` would generate way too many warnings, so for me the best solution is to replace line 41 with
```
        add_compile_options(
            /W3
            /w34100 # 'identifier' : unreferenced formal parameter
            /w34189 # 'identifier' : local variable is initialized but not referenced
        )
```

src/train_gui.cpp
```diff
@@ -22,12 +22,11 @@
 
 /**
  * Callback for building wagons.
- * @param cmd Unused.
  * @param result The result of the command.
  * @param new_veh_id ID of the ne vehicle.
  * @param tile   The tile the command was executed on.
  */
-void CcBuildWagon(Commands cmd, const CommandCost &result, VehicleID new_veh_id, uint, uint16_t, CargoArray, TileIndex tile, EngineID, bool, CargoID, ClientID)
+void CcBuildWagon(Commands, const CommandCost &result, VehicleID new_veh_id, uint, uint16_t, CargoArray, TileIndex tile, EngineID, bool, CargoID, ClientID)
```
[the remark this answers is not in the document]
aldot: GCC has an option to moan about conspicuous frame sizes, fwiw:

-Wframe-larger-than=511 or, at your discretion -Wframe-larger-than=16384 or whatever you deem sensible

src/vehicle_gui.cpp
```diff
@@ -3374,11 +3373,10 @@
 
 /**
  * This is the Callback method after the construction attempt of a primary vehicle
- * @param cmd unused
  * @param result indicates completion (or not) of the operation
  * @param new_veh_id ID of the new vehicle.
  */
-void CcBuildPrimaryVehicle(Commands cmd, const CommandCost &result, VehicleID new_veh_id, uint, uint16_t, CargoArray)
+void CcBuildPrimaryVehicle(Commands, const CommandCost &result, VehicleID new_veh_id, uint, uint16_t, CargoArray)
```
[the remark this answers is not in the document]
TrueBrain: We should do something with this some day, but I guess not for this PR? :D

src/command.cpp
```diff
@@ -280,11 +280,10 @@ void CommandHelperBase::LogCommandExecution(Commands cmd, StringID err_message,
 /**
  * Prepare for the test run of a command proc call.
  * @param cmd_flags Command flags.
- * @param tile Tile of command execution.
  * @param[in,out] cur_company Backup of current company at start of command execution.
  * @return True if test run can go ahead, false on error.
  */
-bool CommandHelperBase::InternalExecutePrepTest(CommandFlags cmd_flags, TileIndex tile, Backup<CompanyID> &cur_company)
+bool CommandHelperBase::InternalExecutePrepTest(CommandFlags cmd_flags, TileIndex, Backup<CompanyID> &cur_company)
```
rubidium42, line 286: `TileIndex tile` can be removed completely.

src/video/sdl2_opengl_v.cpp
```diff
@@ -50,7 +50,7 @@ static OGLProc GetOGLProcAddressCallback(const char *proc)
 
 bool VideoDriver_SDL_OpenGL::CreateMainWindow(uint w, uint h, uint flags)
 {
-	return this->VideoDriver_SDL_Base::CreateMainWindow(w, h, SDL_WINDOW_OPENGL);
+	return this->VideoDriver_SDL_Base::CreateMainWindow(w, h, flags | SDL_WINDOW_OPENGL);
```
TrueBrain, line 53: this was is funny .. `flags` is actually never used from the top layer, but only exists to have optional flags to the base layer. No caller ever actually sets any flags. Just a bit poor OO design (I think I did this? :D )

Edit: nothing to be fixed for this PR tbh, but this fix is equally "bad" as marking the flags as "not-used" :D

src/command_func.h
```diff
@@ -359,7 +359,7 @@ struct CommandHelper<Tcmd, Tret(*)(DoCommandFlag, Targs...), true> : protected C
 		}
 	}
 
-	static Tret Execute(StringID err_message, CommandCallback *callback, bool my_cmd, bool estimate_only, bool network_command, TileIndex tile, std::tuple<Targs...> args)
+	static Tret Execute(StringID err_message, CommandCallback *callback, bool, bool estimate_only, bool network_command, TileIndex tile, std::tuple<Targs...> args)
```
rubidium42, line 362: `bool my_cmd` can be removed completely.

src/ship_cmd.cpp
```diff
@@ -624,7 +624,7 @@ bool IsShipDestinationTile(TileIndex tile, StationID station)
 	for (DiagDirection d = DIAGDIR_BEGIN; d != DIAGDIR_END; d++) {
 		TileIndex t = tile + TileOffsByDiagDir(d);
 		if (!IsValidTile(t)) continue;
-		if (IsDockTile(t) && GetStationIndex(t) == station && IsValidDockingDirectionForDock(t, d)) return true;
+		if (IsDockTile(t) && GetStationIndex(t) == station && IsDockWaterPart(t)) return true;
```
TrueBrain, line 627: If you plan to "merge with rebase", this commit message needs a bit of work. Such long sentences don't help anyone :P

``` 
Cleanup: docking tiles method names and parameters lost their meaning

This was caused by so many refactors, that it no longer makes sense.

```

src/saveload/saveload.cpp
```diff
@@ -1767,7 +1767,7 @@ std::vector<SaveLoad> SlTableHeader(const SaveLoadTable &slt)
 			}
 
 			while (true) {
-				uint8_t type;
+				uint8_t type = 0;
```
TrueBrain, line 1770: Here too, if you plan to "merge with rebase", this is not a fix. This really is a codechange. Fix is for user-facing issues, not for compiler-facing issues :P

src/core/pool_type.hpp
```diff
@@ -375,7 +374,7 @@ struct Pool : PoolBase {
 		 * @note when this function is called, PoolItem::Get(index) == nullptr.
 		 * @note it's called only when !CleaningPool()
 		 */
-		static inline void PostDestructor(size_t index) { }
+		static inline void PostDestructor([[maybe_unused]] size_t index) { }
```
rubidium42, line 377: Why `[[maybe_unused]]` over removing the parameter name?

src/fontcache/spritefontcache.h
```diff
@@ -23,16 +23,16 @@ class SpriteFontCache : public FontCache {
...
 	SpriteFontCache(FontSize fs);
 	~SpriteFontCache();
-	virtual void SetUnicodeGlyph(char32_t key, SpriteID sprite);
-	virtual void InitializeUnicodeGlyphMap();
-	virtual void ClearFontCache();
-	virtual const Sprite *GetGlyph(GlyphID key);
-	virtual uint GetGlyphWidth(GlyphID key);
-	virtual bool GetDrawGlyphShadow();
-	virtual GlyphID MapCharToGlyph(char32_t key) { assert(IsPrintable(key)); return SPRITE_GLYPH | key; }
+	void SetUnicodeGlyph(char32_t key, SpriteID sprite) override;
```
TrueBrain, line 26: And same here, not a fix, in my opinion :)

src/network/core/http_curl.cpp
```diff
@@ -174,7 +174,7 @@ void HttpThread()
 		 * up to a second before this callback is called. There is little we can
 		 * do about this. */
 		curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 0L);
-		curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, +[](void *userdata, curl_off_t dltotal, curl_off_t dlnow, curl_off_t ultotal, curl_off_t ulnow) -> int {
+		curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, +[](void *userdata, curl_off_t /*dltotal*/, curl_off_t /*dlnow*/, curl_off_t /*ultotal*/, curl_off_t /*ulnow*/) -> int {
```
rubidium42, line 177: Is commenting out the parameter names intentional here, over removal?

src/game/game_gui.cpp
```diff
@@ -367,12 +367,14 @@ struct GSConfigWindow : public Window {
 
 	void OnDropdownSelect(int widget, int index) override
 	{
+		if (widget >= 0) return;
```
rubidium42, line 370: Why not make a sentinel `WID_GSC_DROPDOWN = -1` or something similar in widgets/game_widgets.h. Then use that here and in the calls to create the drop downs? Seems better than the implicit assumption that the widget for the callback is always going to be some negative value for those cases.

src/newgrf_gui.cpp
```diff
@@ -453,6 +453,7 @@ struct NewGRFParametersWindow : public Window {
 
 	void OnDropdownSelect(int widget, int index) override
 	{
+		if (widget >= 0) return;
```
rubidium42, line 456: Maybe create a sentinel enum value instead in the widget enumeration?

src/script/script_gui.cpp
```diff
@@ -526,12 +526,14 @@ struct ScriptSettingsWindow : public Window {
 
 	void OnDropdownSelect(int widget, int index) override
 	{
+		if (widget >= 0) return;
```
rubidium42, line 529: Maybe create a sentinel enum value instead in the widget enumeration?

src/core/pool_type.hpp
frosch123: This is a method in a base-class. The parameter is required for documentation, including the "@param".

src/network/core/http_curl.cpp
frosch123: I did not want to force people look up the curl documentation, just to figure out what all that stuff is. But maybe you would have to anyway :)

TrueBrain: I kinda like what you did here honestly

TrueBrain (approved):

PR #11308, opened 2023-09-17
Codechange: introduce and use function to raise and dirty a set of widgets when they are lowered

rubidium42:
## Motivation / Problem

Same code exists in multiple locations, so why not introduce a function for that and reuse it.
Furthermore a sentinel value to denote end-of-list is a C-style construct.

## Description

Introduce `RaiseWidgetWhenLowered` and `RaiseWidgetsWhenLowered` to check and update multiple locations in one call.

## Limitations

None I can think of.

## Checklist for review

### DIFF
--- a/src/genworld_gui.cpp
+++ b/src/genworld_gui.cpp
@@ -884,16 +884,10 @@ struct GenerateLandscapeWindow : public Window {
 
 	void OnTimeout() override
 	{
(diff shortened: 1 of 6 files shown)


2TallTyler (approved):
The two uses that iterate through widgets could use a bit more love in the future, but no need to do it here, I guess.

PR #11312, opened 2023-09-17
Codechange: replace WIDGET_LIST_END with more specific variants

rubidium42:
## Motivation / Problem

Primarily that `WIDGET_LIST_END` is documented to be the sentinel value for the last element of a vararg list, but it is used as sentinel value for the clicked widget in a window.

## Description

Two `WIDGET_LIST_END` removals that were forgotten in a previous commit.

Add `INVALID_WID_XXX` values to the appropriate widgets enumerations, and use that as sentinel value for the last action/clicked widget.

## Limitations

I attempted to make `last_user_action` of the appropriate widget enumeration type, but that causes many many warnings about missing cast. That is because the widget parameter of the functions in the window class is just `int` and not the actual widget enumeration.
In a perfecter world, the widget (index) parameters would be of the appropriate type. However, that'll require templating which is far beyond the scope of this story.

These `INVALID_WID_XXX` sentinel enumerators end up in the GS API. I doubt it is really worth the effort to not add these there.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -2616,8 +2616,7 @@ struct VehicleDetailsWindow : Window {
 		/* Disable service-scroller when interval is set to disabled */
 		this->SetWidgetsDisabledState(!IsVehicleServiceIntervalEnabled(v->type, v->owner),
 			WID_VD_INCREASE_SERVICING_INTERVAL,
-			WID_VD_DECREASE_SERVICING_INTERVAL,
-			WIDGET_LIST_END);
+			WID_VD_DECREASE_SERVICING_INTERVAL);
 
 		StringID str = v->ServiceIntervalIsCustom() ?
 			(v->ServiceIntervalIsPercent() ? STR_VEHICLE_DETAILS_PERCENT : STR_VEHICLE_DETAILS_DAYS) :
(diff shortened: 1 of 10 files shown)


src/vehicle_gui.cpp
```diff
@@ -2658,8 +2657,7 @@ struct VehicleDetailsWindow : Window {
 					WID_VD_DETAILS_TRAIN_VEHICLES,
 					WID_VD_DETAILS_CAPACITY_OF_EACH,
 					WID_VD_DETAILS_TOTAL_CARGO,
-					widget,
-					WIDGET_LIST_END);
+					widget);
```
glx22: `widget` is one of the 4 widgets, it's not really needed to raise it twice.

rubidium42: Good point. I've removed it.

PeterN:
Would a single INVALID_WIDGET work instead of having to define one for each window?

rubidium42:
> Would a single INVALID_WIDGET work instead of having to define one for each window?

Well, it did in the past with a different name. Though that would mean that the data for `last_user_action` would be coming from two distinct sources, i.e. the window specific enumeration and some global magic constant.
In my opinion it would be better if the windows are going to use the actual widget enumeration as type for widget callbacks, but that's far beyond the scope of this PR.

PeterN (approved):

PR #11314, opened 2023-09-18
Fix: Don't produce invalid cargo.

PeterN:
## Motivation / Problem

Production for secondary industries did not always check that the produced cargo type isn't valid -- it was only checked for some NewGRF industries using callback-based production.

## Description

Check if cargo is valid for all (I think) industry cargo production. This does not check that the cargo is the correct type that the industry is expecting, only that it is not an invalid cargo type.

This also disables the tropical lumber mill from chopping trees if the industry's first cargo slot is invalid -- it may be preferable to continue chopping trees.

## Limitations

The issue of producing invalid cargo is mostly theoretical, and these checks don't really add much to a normal game. I don't have any sample savegames where it occurs, but I believe I have seen it in the past.

Issues could probably only be caused by changes to NewGRFs (either by the user (in which case who cares) or the author) or perhaps loading incompatible NewGRFs concurrently.

However this change has more impact if (the to-be-updated) #10726 PR is applied.

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1169,9 +1169,10 @@ static void TriggerIndustryProduction(Industry *i)
 		}
 	} else {
 		for (auto ita = std::begin(i->accepted); ita != std::end(i->accepted); ++ita) {
-			if (ita->waiting == 0) continue;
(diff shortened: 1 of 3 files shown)


TrueBrain (approved):
Sure.

PR #11316, opened 2023-09-18
Codechange: Remove unused parameter for DropDownListItem::Height()

aldot:
Allegedly several "width" were set to unused in #11306 so remove the now superfluous remaining ones.

## Motivation / Problem

Warnings about unused parameters, code cleanup.

## Description

Built successfully, make test PASSED.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -597,7 +597,7 @@ public:
 		return ScaleGUITrad(28) + WidgetDimensions::scaled.hsep_normal + GetStringBoundingBox(this->String()).width + WidgetDimensions::scaled.dropdowntext.Horizontal();
 	}
 
-	uint Height(uint) const override
+	uint Height() const override
 	{
 		return std::max(FONT_HEIGHT_NORMAL, ScaleGUITrad(12) + WidgetDimensions::scaled.vsep_normal);
 	}
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -475,7 +475,7 @@ struct GameOptionsWindow : Window {
 						Dimension string_dim;
 						int width = ddli->Width();
 						string_dim.width = width + padding.width;
-						string_dim.height = ddli->Height(width) + padding.height;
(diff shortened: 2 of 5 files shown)


LordAro:
I don't understand why this PR exists, it appears to duplicate a (very specific) part of #11306

PeterN:
#11306 removes the parameter name, assuming the parameter itself is used somewhere. It's not.

I probably intended it to be used to allow wrapping lists at some point, but instead we now just make lists wider.

So we should either implement wrapping which requires this parameter, or just remove the unused parameter.

Either way it needs an update :-)

LordAro:
Aha, that makes much more sense. Slightly further update needed to remove the unused variable :)

frosch123:
@aldot Now there are multiple warnings about unused variables. You can remove even more code :)

aldot:
On 20 September 2023 21:28:56 CEST, frosch ***@***.***> wrote:
>@aldot Now there are multiple warnings about unused variables. You can remove even more code :)

I think DSE will do its work and then the unused Param would be removed. But as long as the API has this param (if it would be a public API), we have to somehow retain the landing pads because one class said so, so i think it would be way cleaner to just bite the bullet and remove the param. IMO. I can see how it would make sense to calc in, but it's just not done, right now. But, the binary not being a library, I think we already grok that fact and it's optimised away, so TBH the patch is not really needed. All it would do really is to document that the param is not handled, even to a human. Not sure if that's desirable.. I'd have hoped so.

PeterN:
@aldot I'm not sure what your comment is related to, you seem to misinterpreted.

With this change, now that width is not passed to the Height() method, there is an unnecessary local variable assigned that can also be removed. As you have replied via email I presume you have not viewed the  annotations from the CI here:

https://github.com/OpenTTD/OpenTTD/pull/11316/files

PeterN:
@aldot will you address the issues that need resolving in this?

PeterN (approved):

PR #11324, opened 2023-09-20
Add enum items for magic values

frosch123:
## Motivation / Problem

@rubidium42 suggested in #11306 to add enum items for a number of -1 constants.
There were a few more places using plain numbers for widgets.

## Description

Replace raw numbers with:
* Existing widget enum items.
* Newly added widget enum items.

Bonus: #11306 actually broke the NewGRF preset dropdown. Fixed here.

## Checklist for review

### DIFF
--- a/src/widgets/settings_widget.h
+++ b/src/widgets/settings_widget.h
@@ -64,6 +64,8 @@ enum GameSettingsWidgets {
 	WID_GS_RESTRICT_TYPE,      ///< Label upfront to the type drop-down box to restrict the list of settings to show
 	WID_GS_RESTRICT_DROPDOWN,  ///< The drop down box to restrict the list of settings
 	WID_GS_TYPE_DROPDOWN,      ///< The drop down box to choose client/game/company/all settings
+
+	WID_GS_SETTING_DROPDOWN = -1, ///< Dynamically created dropdown for changing setting value.
 };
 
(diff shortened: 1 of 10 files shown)


src/widgets/settings_widget.h
```diff
@@ -64,6 +64,8 @@ enum GameSettingsWidgets {
 	WID_GS_RESTRICT_TYPE,      ///< Label upfront to the type drop-down box to restrict the list of settings to show
 	WID_GS_RESTRICT_DROPDOWN,  ///< The drop down box to restrict the list of settings
 	WID_GS_TYPE_DROPDOWN,      ///< The drop down box to choose client/game/company/all settings
+
+	WID_GS_SETTING_DROPDOWN = -1, ///< Dynamically created dropdown for changing setting value.
```
TrueBrain, line 68: does it actually need to be `-1`, or can we just do the next number?

frosch123: My understanding is:
* Positive numbers must be valid widgets. Code in widget/window.cpp will use them as indices into arrays without further bounds checking.
* Negative numbers are checked in widget/window.cpp and return early.

Examples:
* https://github.com/OpenTTD/OpenTTD/blob/master/src/window.cpp#L295
* https://github.com/OpenTTD/OpenTTD/blob/master/src/window.cpp#L726

TrueBrain (approved):

PR #11335, opened 2023-09-25
Codechange: Add missing override specifiers.

PeterN:
## Motivation / Problem

Methods incorrectly marked `virtual` and/or not marked `override`.

## Description

This change marks methods that override base classes with `override`. Many of these were found with a linting tool. There may be some missing.

## Checklist for review

### DIFF
--- a/cmake/CompileFlags.cmake
+++ b/cmake/CompileFlags.cmake
@@ -63,6 +63,7 @@ macro(compile_flags)
             -Wformat=2
             -Winit-self
             -Wnon-virtual-dtor
+            -Wsuggest-override
 
             # We use 'ABCD' multichar for SaveLoad chunks identifiers
             -Wno-multichar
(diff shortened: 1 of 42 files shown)


LordAro:
Is there a warning we could enable to avoid these? Feels like there could

PeterN:
> > Is there a warning we could enable to avoid these? Feels like there could
> 
> Good point. There is -Wsuggest-override. I will add that and update this PR (which of course will fail with the flag because there are more missing...)

Eh, 3rdparty screws it up of course :(

LordAro:
Add -Wno-suggest-override to 3rdparty/CMakeLists ?

PeterN:
> Add -Wno-suggest-override to 3rdparty/CMakeLists ?

Don't think that is possible, the subdirectories add files to the main project and is it all built together. And besides the bits that link to squirrel from OpenTTD would still trip up.

We have precedent with previous changes that add override, so I'll just do it as a separate commit.

frosch123 (approved):
We have plenty of modifies in 3rdparty/squirrel for all kind of warnings.

PR #11338, opened 2023-09-26
GUI Setting to control whether main toolbar dropdown menus close instantly or not

mdias:
I find that needing to have the mouse button pressed to navigate the dropdown menus on the main toolbar is very annoying so I decided to create a setting to allow users like me to use normal clicks in these menus.

Do note that this initially created a visual problem on the toolbar buttons when we tried to click them after it had their dropdown menu open already. The change in window.cpp was done to fix this.

I couldn't find any issues with this change, but I'm also not very familiar with all the things that could happen in the gui of the game.


2TallTyler:
See also: #10938.

mdias:
I've updated the PR to rebased on latest.

PeterN:
#10938 instead

PR #11342, opened 2023-09-30
Fix #10763, Fix #11168: Display variant groups that consist only of unavailable sub-groups

### LINKED ISSUE #10763: [Bug]: Wrong list of vehicles to replace with vehicles groups
### Version of OpenTTD

n20230430

### Expected result

You are able to autoreplace vehicles inside the group the same way you replace vehicles not in the group

### Actual result

Wagons in group are not represented in the list.
image
image
If the wagon is removed from the group, it becomes available
image

There is also a problem with engines, but it looks a bit different
the first entry in the group disappears from the list
image
but the second one becomes disabled
image

image
image

### Steps to reproduce

[xussrset-dmus.zip](https://github.com/OpenTTD/OpenTTD/files/11401404/xussrset-dmus.zip)
start a new game, build a DMU (DR1A for example, try to autoreplace wagons)

### LINKED ISSUE #11168: [Bug]: Vehicle group not shown if it consist of unavailable subgroups with available vehicles
### Version of OpenTTD

n20230729

### Expected result

A group is available

### Actual result

A group is unavailable, because its subgroups are unavailable, while vehicles in subgroup are available
If you move the subgroup out of the group, subgroup becomes available

### Steps to reproduce

1) start a game in 1970-th with xussr-electric
[xussr-electric.zip](https://github.com/OpenTTD/OpenTTD/files/12251258/xussr-electric.zip)
2) try to build VL10

You can't, because VL10 series group is not represented in the list
VL10 series group consist of a subgroup TEVZ, that is vehicle with no climates, but there are engines VL10 inside this subgroup 
 
I've moved VL10U to the VL10 Series group to illustrate, what was intended
image

merni-ns:
## Motivation / Problem
Fixes #10763 and fixes #11168.

Essentially, the code added the immediate parent of a variant vehicle to the list, but not higher levels.

## Description

Loop through and add all the parents of every variant vehicle up to the top level, so that even when a group consists only of unavailable sub-groups, the top-level group is still displayed when its second-level children are available.

The example provided by @George-VB in #11168 now produces the following list when game is started in 1.1.1970 with default settings.
image

And the example in #10763:
image

## Limitations
Not well tested as I cannot write NewGRF nor do I know the actual intended output. Perhaps @George-VB could help with testing.

## Checklist for review

### DIFF
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -179,7 +179,13 @@ class ReplaceVehicleWindow : public Window {
 			if (side == 1 && eid == this->sel_engine[0]) flags |= EngineDisplayFlags::Shaded;
 			list.emplace_back(eid, e->info.variant_id, flags, 0);
 
-			if (side == 1 && e->info.variant_id != INVALID_ENGINE) variants.push_back(e->info.variant_id);
+			if (side == 1) {
+				EngineID parent = e->info.variant_id;
+				while (parent != INVALID_ENGINE) {
+					variants.push_back(parent);
+					parent = Engine::Get(parent)->info.variant_id;
+				}
+			}
 			if (eid == this->sel_engine[side]) selected_engine = eid; // The selected engine is still in the list
 		}
 
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1422,7 +1422,14 @@ struct BuildVehicleWindow : Window {
 			list.emplace_back(eid, e->info.variant_id, e->display_flags, 0);
 
 			if (rvi->railveh_type != RAILVEH_WAGON) num_engines++;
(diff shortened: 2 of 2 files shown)


merni-ns (commented):
Ideally #11343 would be merged before this PR.
https://github.com/merni-ns/OpenTTD/blob/a80d8f6353d8c0a4bb4582afe8bf91d421d4c0aa/src/build_vehicle_gui.cpp#L1174
and
https://github.com/merni-ns/OpenTTD/blob/a80d8f6353d8c0a4bb4582afe8bf91d421d4c0aa/src/build_vehicle_gui.cpp#L1425
can be deleted once that is merged.

PeterN (commented):
This looks good to me, but needs #11343 to prevent a bad NewGRF creating an infinite loop.

PeterN (approved):

PR #11343, opened 2023-09-30
Fix: Check for engine variant loops during NewGRF initialization.

PeterN:
## Motivation / Problem

Invalid NewGRFs could set up an engine variant loop that never ends. This was checked for in some places that evaluated variants, but not all. In most cases this would result in the engines not appearing, but could potentially cause an infinite loop and crash.

This issue would also affect #11342.

## Description

Instead, during NewGRF initialization detect loops and remove invalid variants before setting display flags.

## Checklist for review


LordAro (approved):
Seems reasonable

Issue #8846, opened 2021-03-11
Updating add-ons in the NewGRF settings window don't take into account the compatibility parameter


LC-Zorg:
## Version of OpenTTD
1.10.3
## Expected result
It should only be possible to upgrade a add-on to a newer version if both old and new are compatible.
## Actual result
The game allows you to update the add-on to a version that is incompatible, which may cause problems.
For example if you update Wasteland from 0.4.0 (version 7) to 0.7.0 (version 100), the new version has the old version settings after the update.

It is about an update before the game starts. The newgrf_developer_tools and similar settings have no effect on the occurrence of the error. The error also occurs with the default settings.
![Incompatibile versions](https://user-images.githubusercontent.com/67662046/110845086-7d1c7400-82aa-11eb-818d-d0710c694574.png)

## Steps to reproduce
0. Download Wasteland 0.7.0alpha1 - [here](https://www.tt-forums.net/viewtopic.php?p=1242094#p1242094)
1. Start the game (any version)
2. Open the NewGRF settings window
3. For the clean list, select Wasteland 0.4.0 (or earlier)
4. Press Upgrade
- The add-on will be updated to 0.7.0 which is incompatibile with any pervious version
- Version 0.7.0 will still be shown also on the inactive side
- The replaced version will not be reverted to the inactive side
5. Open the add-on's settings
- New version will have incorrect settings

LordAro:
Actual bug, in my view: NewGRF settings are not reset (or at least migrated) after upgrading a NewGRF

Andrew350:
So I'd just like to call attention to this again and maybe expand on the issue, as I've since had two more reports in the span of about 3 months claiming my NewGRF is broken, when in fact it is this bug causing the issues, and this last time it has genuinely broken someone's game.

In addition to the parameters window not being updated (as described by the OP), any new parameters added by the newer version will simply be set to a 0 value when starting a new game. This can cause major problems if these new parameters are not supposed to ever have a 0 value, like in my case, it will completely break the gameplay leading to reports that the NewGRF is broken (which it now is). Normally this should never be possible, as newer versions are set as incompatible, but as shown here the "upgrade" button disregards that completely allowing users to unknowingly break their own game.

[...]

Issue #10549, opened 2023-03-07
[Bug]: Default parameters in NewGRF not used when an active GRF is upgraded


cmcaine:
### Version of OpenTTD

13.0, Windows 10

### Expected result

A GRFs default values should be loaded if no value was specified by the user.

### Actual result

When parameters are added in a new version of a GRF, the new parameters take the value 0 rather than their default.

(at least in this case)

### Steps to reproduce

1. Get "Sailing Ships 0.62" from BaNaNaS and install it as an active GRF. This GRF has no parameters.
2. Activate it.
3. Close OpenTTD
4. Remove Sailing Ships from the content_download folder
5. Add ss.grf (a newer version of Sailing Ships GRF that has parameters) to your /newgrf/ folder
6. Start the game, click new game, newgrf settings, set parameters for "Sailing Ships (tweaked) 0.72" and observe that all parameters have value 0 (if you want to observe that the value really is 0, start a new game in 1705 and check if a ship called Schooner is available, if it is, then the "Ship introduction date" parameter is not at its default value (1: Historical)).
7. Click "reset". Observe that all parameters now have named values

[ss.zip](https://github.com/OpenTTD/OpenTTD/files/10909374/ss.zip)

The source for ss.zip can be found here: https://github.com/cmcaine/OpenTTD-Sailing-Ships commit bbaf839a613133ef7bdcd4522eb7b8749564a247

James103:
See also https://github.com/OpenTTD/OpenTTD/issues/8846

frosch123:
#8846 was fixed in #11348.
If a NewGRF is defined as "incompatible" it will now reset *all* parameters to defaults.

I am not sure whether there is a good method to only set new parameters to default. In the "load savegame/scenario with compatible NewGRF" case, there is definitely not.

But the "upgrade NewGRF preset" feature could be more confortable.

2TallTyler:
As discussed on Discord, I believe #11348 resolves this issue. When a NewGRF author adds or rearranges parameters, they will need to bump the compatible version. Then, a user upgrading the NewGRF will have all parameters set to the defaults. (Not just the new ones; existing choices will be lost and need to be re-selected by the player.)

Andrew350:
Quote from #11348:

>If a NewGRF changes or adds parameters with non-zero defaults, this is (currently) always a savegame-incompatible change.

[...]

PR #11347, opened 2023-10-02
Support parameters for base graphics

frosch123:
## Motivation / Problem

OpenGFX2 features a number of parameters to configure detail looks:
image

Currently this requires an extra NewGRF as work-around.

## Description

* Base graphics parameters are the NewGRF parameters of the Extra-GRF in the baseset.
* The settings window is accessible from the game options, next to the baseset selection.
image
* The settings are stored in openttd.cfg in a separate section. When saving openttd.cfg the section is reset, so all items in the section are always written consistently. Future versions of OpenTTD can savely add more items to the section. 
```
[graphicsset]
name = OpenGFX2 Classic
shortname = 6F676678
extra_version = 0
extra_params = 0 1 0 0
```

* When downloading a new version of the baseset, the "min compatible version" from action14 is checked. The parameters are kept, if the versions are reported compatible. Otherwise all settings are reset to default values.

## Limitations

This PR works with OpenGFX2_Classic-0.1.

The current version 0.21 appears to be broken: It has parameters, but they do nothing. @zephyris any idea why?

## Checklist for review

### DIFF
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -237,12 +237,22 @@ enum BlitterType {
 	BLT_32BPP,      ///< Base set has both 8 bpp and 32 bpp sprites.
 };
 
+struct GRFConfig;
+
 /** All data of a graphics set. */
 struct GraphicsSet : BaseSet<GraphicsSet, MAX_GFT, true> {
+private:
+	mutable std::unique_ptr<GRFConfig> extra_cfg; ///< Parameters for extra GRF
+public:
 	PaletteType palette;       ///< Palette of this graphics set
 	BlitterType blitter;       ///< Blitter of this graphics set
 
+	GraphicsSet();
+	~GraphicsSet();
+
 	bool FillSetDetails(const IniFile &ini, const std::string &path, const std::string &full_filename);
+	GRFConfig *GetExtraConfig() const { return this->extra_cfg.get(); }
+	GRFConfig &GetOrCreateExtraConfig() const;
 
 	static MD5File::ChecksumResult CheckMD5(const MD5File *file, Subdirectory subdir);
 };
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -170,9 +170,6 @@ protected:
 	 */
 	static const char *GetExtension();
 public:
-	/** The set as saved in the config file. */
-	static std::string ini_set;
-
 	/**
 	 * Determine the graphics pack that has to be used.
 	 * The one with the most correct files wins.
@@ -207,7 +204,6 @@ public:
 	static bool HasSet(const ContentInfo *ci, bool md5sum);
 };
 
-template <class Tbase_set> /* static */ std::string BaseMedia<Tbase_set>::ini_set;
 template <class Tbase_set> /* static */ const Tbase_set *BaseMedia<Tbase_set>::used_set;
 template <class Tbase_set> /* static */ Tbase_set *BaseMedia<Tbase_set>::available_sets;
 template <class Tbase_set> /* static */ Tbase_set *BaseMedia<Tbase_set>::duplicate_sets;
@@ -252,6 +248,12 @@ struct GraphicsSet : BaseSet<GraphicsSet, MAX_GFT, true> {
 /** All data/functions related with replacing the base graphics. */
 class BaseGraphics : public BaseMedia<GraphicsSet> {
 public:
+	/** Values loaded from config file. */
+	struct Ini {
+		std::string name;
+	};
+	static inline Ini ini_data;
+
 };
 
 /** All data of a sounds set. */
@@ -261,6 +263,9 @@ struct SoundsSet : BaseSet<SoundsSet, 1, true> {
 /** All data/functions related with replacing the base sounds */
 class BaseSounds : public BaseMedia<SoundsSet> {
 public:
+	/** The set as saved in the config file. */
+	static inline std::string ini_set;
+
 };
 
 /** Maximum number of songs in the 'class' playlists. */
@@ -307,6 +312,9 @@ struct MusicSet : BaseSet<MusicSet, NUM_SONGS_AVAILABLE, false> {
 /** All data/functions related with replacing the base music */
 class BaseMusic : public BaseMedia<MusicSet> {
 public:
+	/** The set as saved in the config file. */
+	static inline std::string ini_set;
+
 };
 
 #endif /* BASE_MEDIA_BASE_H */
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -254,6 +254,7 @@ public:
 	bool FillSetDetails(const IniFile &ini, const std::string &path, const std::string &full_filename);
 	GRFConfig *GetExtraConfig() const { return this->extra_cfg.get(); }
 	GRFConfig &GetOrCreateExtraConfig() const;
+	bool IsConfigurable() const;
 	void CopyCompatibleConfig(const GraphicsSet &src);
 
 	static MD5File::ChecksumResult CheckMD5(const MD5File *file, Subdirectory subdir);
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -188,7 +188,9 @@ public:
 
 	static Tbase_set *GetAvailableSets();
 
-	static bool SetSet(const std::string &name);
+	static bool SetSet(const Tbase_set *set);
+	static bool SetSetByName(const std::string &name);
+	static bool SetSetByShortname(uint32_t shortname);
 	static void GetSetsList(std::back_insert_iterator<std::string> &output_iterator);
 	static int GetNumSets();
 	static int GetIndexOfUsedSet();
@@ -251,6 +253,7 @@ public:
 	/** Values loaded from config file. */
 	struct Ini {
 		std::string name;
+		uint32_t shortname;                 ///< unique key for base set
 	};
 	static inline Ini ini_data;
 
--- a/src/base_media_base.h
+++ b/src/base_media_base.h
@@ -96,6 +96,7 @@ struct BaseSet {
 	}
 
 	bool FillSetDetails(const IniFile &ini, const std::string &path, const std::string &full_filename, bool allow_empty_filename = true);
+	void CopyCompatibleConfig([[maybe_unused]] const T &src) {}
 
 	/**
 	 * Get the description for the given ISO code.
@@ -253,6 +254,7 @@ public:
 	bool FillSetDetails(const IniFile &ini, const std::string &path, const std::string &full_filename);
 	GRFConfig *GetExtraConfig() const { return this->extra_cfg.get(); }
 	GRFConfig &GetOrCreateExtraConfig() const;
+	void CopyCompatibleConfig(const GraphicsSet &src);
 
 	static MD5File::ChecksumResult CheckMD5(const MD5File *file, Subdirectory subdir);
 };
@@ -264,6 +266,8 @@ public:
 	struct Ini {
 		std::string name;
 		uint32_t shortname;                 ///< unique key for base set
+		uint32_t extra_version;             ///< version of the extra GRF
+		std::vector<uint32_t> extra_params; ///< parameters for the extra GRF
 	};
 	static inline Ini ini_data;
 
--- a/src/gfxinit.cpp
+++ b/src/gfxinit.cpp
@@ -192,18 +192,7 @@ static void LoadSpriteTables()
 	ClrBit(master->flags, GCF_INIT_ONLY);
 
 	/* Baseset extra graphics */
-	GRFConfig *extra = new GRFConfig(used_set->files[GFT_EXTRA].filename.c_str());
-
-	/* We know the palette of the base set, so if the base NewGRF is not
-	 * setting one, use the palette of the base set and not the global
-	 * one which might be the wrong palette for this base NewGRF.
-	 * The value set here might be overridden via action14 later. */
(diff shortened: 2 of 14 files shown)


src/base_media_base.h
```diff
@@ -239,19 +238,36 @@ enum BlitterType {
...
 /** All data of a graphics set. */
 struct GraphicsSet : BaseSet<GraphicsSet, MAX_GFT, true> {
+private:
+	mutable std::unique_ptr<GRFConfig> extra_cfg; ///< Parameters for extra GRF
+public:
 	PaletteType palette;       ///< Palette of this graphics set
 	BlitterType blitter;       ///< Blitter of this graphics set
 
 	bool FillSetDetails(struct IniFile *ini, const std::string &path, const std::string &full_filename);
+	GRFConfig *GetExtraConfig(bool load=true) const;
```
PeterN: Spaces... `bool load = true`

src/gfxinit.cpp
```diff
@@ -364,6 +354,47 @@ bool GraphicsSet::FillSetDetails(IniFile *ini, const std::string &path, const st
...
+	const GRFConfig* cfg = this->GetExtraConfig();
+	/* This check is more strict than the one for NewGRF Settings.
+	 * There are no legacy basesets with parameters, but without Action14 */
+	return cfg && !cfg->param_info.empty();
+}
+
+void GraphicsSet::CopyCompatibleConfig(const GraphicsSet &src)
+{
+	const GRFConfig* src_cfg = src.GetExtraConfig(false);
+	if (!src_cfg || src_cfg->num_params == 0) return;
```
PeterN: `src_cfg != nullptr`

src/gfxinit.cpp
```diff
@@ -364,6 +354,47 @@ bool GraphicsSet::FillSetDetails(IniFile *ini, const std::string &path, const st
...
+	GRFConfig* dest_cfg = this->GetExtraConfig();
+	if (!dest_cfg || !dest_cfg->IsCompatible(src_cfg->version)) return;
```
PeterN: dest_cfg != nullptr

src/settings.cpp
```diff
@@ -981,6 +985,36 @@ static bool DecodeHexText(const char *pos, uint8_t *dest, size_t dest_size)
...
 }
 
+/**
+ * Load BaseGraphics set selection and configuration.
+ */
+static void GraphicsSetLoadConfig(IniFile &ini)
+{
+	if (IniGroup *group = ini.GetGroup("misc"); group != nullptr) {
+		/* Load old setting first. */
+		if (IniItem *item = group->GetItem("graphicsset"); item != nullptr && item->value) BaseGraphics::ini_data.name = *item->value;
```
PeterN: Interesting syntax inside if that I've not seen before.

src/settings.cpp
```diff
@@ -1159,6 +1193,25 @@ static void SaveVersionInConfig(IniFile &ini)
 	group->GetOrCreateItem("ini_version").SetValue(std::to_string(INIFILE_VERSION));
 }
 
+/**
+ * Save BaseGraphics set selection and configuration.
+ */
+static void GraphicsSetSaveConfig(IniFile &ini)
+{
+	const GraphicsSet *used_set = BaseGraphics::GetUsedSet();
+	if (used_set) {
```
PeterN: `if (used_set != nullptr) {`

or better,

`if (used_set == nullptr) return;`

src/settings.cpp
```diff
@@ -1159,6 +1193,25 @@ static void SaveVersionInConfig(IniFile &ini)
...
+		ini.RemoveGroup("graphicsset");
+		IniGroup *group = ini.GetGroup("graphicsset");
+		group->GetOrCreateItem("name").SetValue(used_set->name);
+		group->GetOrCreateItem("shortname").SetValue(fmt::format("{:08X}", BSWAP32(used_set->shortname)));
+		const GRFConfig *extra_cfg = used_set->GetExtraConfig(false);
+		if (extra_cfg && extra_cfg->num_params > 0) {
```
PeterN: As above.

src/settings_gui.cpp
```diff
@@ -601,6 +602,16 @@ struct GameOptionsWindow : Window {
 				break;
 			}
 
+			case WID_GO_BASE_GRF_PARAMETERS: {
+				auto *used_set = BaseGraphics::GetUsedSet();
+				if (!used_set || !used_set->IsConfigurable()) break;
```
PeterN: `!= nullptr`

PeterN (changes requested):
Nice

src/base_media_base.h
```diff
@@ -239,19 +238,36 @@ enum BlitterType {
...
 	static MD5File::ChecksumResult CheckMD5(const MD5File *file, Subdirectory subdir);
 };
 
 /** All data/functions related with replacing the base graphics. */
 class BaseGraphics : public BaseMedia<GraphicsSet> {
 public:
+	/** Values loaded from config file. */
+	struct Ini {
+		std::string name;
+		uint32_t shortname;                 //!< unique key for base set
```
LordAro:
```suggestion
		uint32_t shortname;                 ///< unique key for base set
```

etc

src/gfxinit.cpp
```diff
@@ -364,6 +354,47 @@ bool GraphicsSet::FillSetDetails(IniFile *ini, const std::string &path, const st
...
+			default: break;
+		}
+		FillGRFDetails(this->extra_cfg.get(), false, BASESET_DIR);
+	}
+	return this->extra_cfg.get();
+}
+
+bool GraphicsSet::IsConfigurable() const
+{
+	const GRFConfig* cfg = this->GetExtraConfig();
```
LordAro:
```suggestion
	const GRFConfig *cfg = this->GetExtraConfig();
```

etc

src/settings.cpp
```diff
@@ -981,6 +985,36 @@ static bool DecodeHexText(const char *pos, uint8_t *dest, size_t dest_size)
...
+	}
+	if (IniGroup *group = ini.GetGroup("graphicsset", false); group != nullptr) {
+		/* Load new settings. */
+		if (IniItem *item = group->GetItem("name"); item != nullptr && item->value) BaseGraphics::ini_data.name = *item->value;
+		if (IniItem *item = group->GetItem("shortname"); item != nullptr && item->value && item->value->size() == 8) {
+			BaseGraphics::ini_data.shortname = BSWAP32(std::strtoul(item->value->c_str(), nullptr, 16));
+		}
+		if (IniItem *item = group->GetItem("extra_version"); item != nullptr && item->value) BaseGraphics::ini_data.extra_version = std::strtoul(item->value->c_str(), nullptr, 10);
+		if (IniItem *item = group->GetItem("extra_params"); item != nullptr && item->value) {
+			auto& extra_params = BaseGraphics::ini_data.extra_params;
```
LordAro:
```suggestion
			auto &extra_params = BaseGraphics::ini_data.extra_params;
```

etc

src/settings.cpp
frosch123: https://youtu.be/tsG95Y-C14k?t=114

src/settings.cpp
```diff
@@ -53,6 +54,9 @@
 ClientSettings _settings_client;
 GameSettings _settings_game;     ///< Game settings of a running game or the scenario editor.
 GameSettings _settings_newgame;  ///< Game settings for new games (updated from the intro screen).
+BaseGraphics::Ini BaseGraphics::ini_data;
+std::string BaseSounds::ini_set;
+std::string BaseMusic::ini_set;
```
PeterN: Does it work to have these as static inline inside the class definitions (like how window hotkeys are set up), instead of dumped here?

src/settings.cpp
LordAro: https://en.cppreference.com/w/cpp/language/if#If_statements_with_initializer

2TallTyler:
The first two commits could be their own PR (or two). The second commit in particular would fix #8846 and #10549, if I'm reading everything correctly.

zephyris:
Very nice, thank you!

Re. OpenGFX2 v0.1 vs v0.21 - it's my fault, see: https://github.com/zephyris/opengfx2/blob/7be1c3e7cf42d0eb1cebef52b162eb1763c99ba3/baseset/nml/extra/extra-header.pnml#L257-L297
I was trying to work out how to have the base set parameters (client side) and a settings grf (newgrf/server defined) coexist nicely. Without a way for me to set base set parameters for testing I gave up and hardcoded some defaults for now!

PeterN:
Speaking of networking, presumably this doesn't introduce a method of baseset-desyncs?

frosch123:
@PeterN There is already a loop hole for desyncs.
* If NewGRF check for presence of static NewGRF, the static NewGRF are disabled. (desync safe)
* If NewGRF check for presence of the baseset extra GRF, OpenTTD happily does that, and this can potentially desync.
* Some NewGRF do this to match sprites for different basesets.
* The same applies to reading parameters from other NewGRF: OpenTTD catches it for static GRF, but allows it for the baseset extra GRF.

For me it's a case of: Tolerate as long as noone causes desyncs. Fix on first occassion someone does something bad.

src/gfxinit.cpp
```diff
@@ -347,6 +337,17 @@ void GfxLoadSprites()
...
+GraphicsSet::GraphicsSet()
+	: BaseSet<GraphicsSet, MAX_GFT, true>{}, palette{}, blitter{}
+{
+	// only here due to incomplete types
+}
+
+GraphicsSet::~GraphicsSet()
+{
+	// only here due to incomplete types
+}
```
PeterN, line 349: Why are these necessary now? Also, prefer member initialization of `palette` and `blitter` instead of initialization list.

src/settings.cpp
```diff
@@ -1159,6 +1190,24 @@ static void SaveVersionInConfig(IniFile &ini)
...
+	if (used_set == nullptr) return;
+	ini.RemoveGroup("graphicsset");
+	IniGroup *group = ini.GetGroup("graphicsset");
+	group->GetOrCreateItem("name").SetValue(used_set->name);
+	group->GetOrCreateItem("shortname").SetValue(fmt::format("{:08X}", BSWAP32(used_set->shortname)));
+	const GRFConfig *extra_cfg = used_set->GetExtraConfig();
+	if (extra_cfg != nullptr && extra_cfg->num_params > 0) {
+		group->GetOrCreateItem("extra_version").SetValue(fmt::format("{}", extra_cfg->version));
+		group->GetOrCreateItem("extra_params").SetValue(GRFBuildParamList(extra_cfg));
+	}
```
PeterN: Could perhaps do with some extra newlines to break it into more readable sections...

src/gfxinit.cpp
frosch123: The destructor of `std::unique_ptr` needs a complete type, so it can call the destructor of the held type.
I do not want to include `newgrf_config.h` in `base_media_base.h`, so I have to put `~GraphicsSet` into some `.cpp` file.
For some reason MSVC also needs the constructor in a `.cpp` file.

PeterN: Ah, so something like this might be helpful:
`/* Explicit constructor/destructor required due to incomplete type used in unique_ptr type. */`

glx22:
Test build available at https://cdn.openttd.org/openttd-branches/pr11347/

zephyris:
Thanks @glx22. I've tested it with OpenGFX2 v0.1 and my latest OpenGFX2 dev build. Looks like it works exactly as expected.
* It's not 100% intuitive how to apply the changes, ie. that the settings change on exit from the settings window, but I think it is fine as is.
* The settings window is wider than before - presumably wider by the parameters button width? Should probably be narrowed down.
* I haven't tested the `action14` support.
* Looks like these changes also fix an old bug. Previously OpenTTD ignored default values for parameter values set in nml `grf {}` block, this behaves correctly now.

PeterN:
>     * Looks like these changes also fix an old bug. Previously OpenTTD ignored default values for parameter values set in nml `grf {}` block, this behaves correctly now.

What's the issue number for the old bug? It might have been closed by related change #11348?

zephyris:
It hasn't got a issue number, this comes from me trying to get OpenGFX2 parameters working as expected over the last week or so. It's only now, testing with this PR, that it's clear that it was OpenTTD defaulting to 0 for parameter values set in nml grf {} block (and not an nml syntax error or something else).

PeterN:
Oh, okay... OpenTTD doesn't support baseset parameters without this change, so I'm not sure why you'd consider it a separate bug that baseset parameters were not supported...

zephyris:
My naivety then :)  Moot point anyway, this PR handles it all.

PeterN (dismissed):
Let's not let this go stale

src/settings.cpp
```diff
@@ -1153,6 +1188,27 @@ static void SaveVersionInConfig(IniFile &ini)
...
+/**
+ * Save BaseGraphics set selection and configuration.
+ */
+static void GraphicsSetSaveConfig(IniFile &ini)
+{
+	const GraphicsSet *used_set = BaseGraphics::GetUsedSet();
+	if (used_set == nullptr) return;
+
+	ini.RemoveGroup("graphicsset");
+	IniGroup *group = ini.GetGroup("graphicsset");
```
PeterN: `GetGroup()` will no longer create a group.

```suggestion
	IniGroup &group = ini.GetOrCreateGroup("graphicsset");
	group.Clear();
```

PeterN (approved):

PR #11348, opened 2023-10-02
Fix #8846: When upgrading NewGRF presets, copy NewGRF parameters only if the NewGRF are compatible.

### LINKED ISSUE #8846: Updating add-ons in the NewGRF settings window don't take into account the compatibility parameter

frosch123:
## Motivation / Problem

#8846

## Description

When upgrading NewGRF the old parameters do not necessarily fit to the new version.
* NewGRF specify a version and minimum-compatible-version for savegame-compatibility.
* If a NewGRF changes or adds parameters with non-zero defaults, this is (currently) always a savegame-incompatible change. (parameters are not changed, when loading savegames/scenarios with compatible NewGRF)
* However, there are other savegame-incompatible changes, which do not invalidate parameters.

As such, this PR is on the "annoying, but safe" side. When upgrading NewGRF:
* parameters are only copied, if they are definitely compatible. (no false negatives)
* parameters are reset to defaults, if they are potentially incompatible. (false positives exist)

## Limitations

This PR does not implement #10549:
* Upgrading NewGRFs in presets uses the same logic as finding compatible NewGRF for savegames/scenarios.
* Detecting "new parameters" is not possible in all cases, if the original NewGRF is not present. (only if they use new registers at the end)

## Checklist for review

### DIFF
--- a/src/newgrf_config.cpp
+++ b/src/newgrf_config.cpp
@@ -73,7 +73,6 @@ GRFConfig::GRFConfig(const GRFConfig &config) :
 void GRFConfig::CopyParams(const GRFConfig &src)
 {
 	this->num_params = src.num_params;
-	this->num_valid_params = src.num_valid_params;
 	this->param = src.param;
 }
 
--- a/src/newgrf_config.cpp
+++ b/src/newgrf_config.cpp
@@ -67,6 +67,14 @@ GRFConfig::GRFConfig(const GRFConfig &config) :
 }
 
 /**
+ * Return whether this NewGRF can replace an older version of the same NewGRF.
+ */
+bool GRFConfig::IsCompatible(uint32_t old_version) const
+{
+	return this->min_loadable_version <= old_version && old_version <= this->version;
+}
+
+/**
  * Copy the parameter information from the \a src config.
  * @param src Source config.
  */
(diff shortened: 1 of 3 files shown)


LordAro (approved):
Seems legit

2TallTyler:
Would this fix #10549 by refusing to upgrade those NewGRFs, assuming the NewGRF author has correctly marked the new version as incompatible? (I could be misunderstanding the use case involved in both that issue and this PR.)

frosch123:
It would not "refuse" to update them, but reset their configuration to "default parameters".

PeterN (approved):

PR #11349, opened 2023-10-02
Codechange: Simplify usage of GRFFile cargo_map.

PeterN:
## Motivation / Problem

Weird flow control when resolving default cargo uses unnecessary assignment and comparison of local variable `cargo_map_for_first_refittable` instead of directly using `file->cargo_map`.

## Description

Just use `file->cargo_map` directly. And also make it a std::array...

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -8850,7 +8850,7 @@ void ResetPersistentNewGRFData()
  */
 static void BuildCargoTranslationMap()
 {
-	memset(_cur.grffile->cargo_map, 0xFF, sizeof(_cur.grffile->cargo_map));
+	std::fill(std::begin(_cur.grffile->cargo_map), std::end(_cur.grffile->cargo_map), uint8_t{UINT8_MAX});
 
 	for (CargoID c = 0; c < NUM_CARGO; c++) {
 		const CargoSpec *cs = CargoSpec::Get(c);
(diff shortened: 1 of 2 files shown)


src/newgrf.cpp
```diff
@@ -8850,7 +8850,7 @@ void ResetPersistentNewGRFData()
  */
 static void BuildCargoTranslationMap()
 {
-	memset(_cur.grffile->cargo_map, 0xFF, sizeof(_cur.grffile->cargo_map));
+	std::fill(std::begin(_cur.grffile->cargo_map), std::end(_cur.grffile->cargo_map), uint8_t{UINT8_MAX});
```
frosch123: There is also `std::array::fill`

PeterN: There is. Should we keep it generic or keep it short? I found that so far all our uses of this particular `std::fill` pattern are used on C-arrays, and `std::array::fill` isn't used at all. I do however have some other WIP patches that use this pattern that could use the shorter form if we want.

src/newgrf.cpp
```diff
@@ -9068,20 +9068,13 @@ static void CalculateRefitMasks()
...
-				if (file == nullptr) file = e->GetGRF();
-				if (file != nullptr && file->grf_version >= 8 && file->cargo_list.size() != 0) {
-					cargo_map_for_first_refittable = file->cargo_map;
-				}
-			}
-
-			if (cargo_map_for_first_refittable != nullptr) {
+			const GRFFile *file = _gted[engine].defaultcargo_grf;
+			if (file == nullptr) file = e->GetGRF();
+			if (file != nullptr && file->grf_version >= 8 && file->cargo_list.size() != 0) {
```
rubidium42, line 9073:
```suggestion
			if (file != nullptr && file->grf_version >= 8 && !file->cargo_list.empty()) {
```
It's already existing code that got re-indented. So, it might also be something for a later PR that looks for other `size() == 0` vs `size() != 0` cases.

rubidium42 (approved):

PR #11341, opened 2023-09-30
Feature: Use real-time "wallclock" timekeeping units

2TallTyler:
## Motivation / Problem

This is a step towards configurable calendar progression speed (#11428). It requires and is built atop #10700.

In #11428, we will be able to freeze calendar time completely. We therefore need to introduce a new way to display financial data, industry and town production rates, etc.

## Description

@nielsmh's brilliant suggestion was to change the tick speed slightly so that time more closely matches real-world units: if a day is two seconds, we can make an economic month have 30 days consistently (meaning a year is now 360 economy days) and be one minute long. Then we can use these real-world units in the GUI, leaving calendar time only for things like technology and variable snowline.

This PR builds atop #10700, which decouples economy and calendar time. It adds a new setting for timekeeping units. Calendar is the standard OpenTTD gameplay and players notice no change.

In Wallclock mode, economy time ticks at its own consistent rate. Economy months are 30 days and become a Minute. Economy years have 360 days and become a Period. The economy starts at Period 1, like in Transport Fever 2's frozen time mode. The full economy date is never shown -- just the period portion of it.

There are some oddities requiring more explanation:

### Timetables

#11435 added a mode to display timetables in seconds. This PR forces the game to use that mode.

### Graphs

OpenTTD graphs are shown in quarters. These are now three minutes long, forming a 12-minute year. Since again, we have no fixed dates to show the player, these are relative to now.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, game_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11359, opened 2023-10-08
Change: Don't set vehicle on time if timetable not started

2TallTyler:
## Motivation / Problem

This is a change required for #11341, which can be a separate PR.

When a vehicle has not started its timetable, it cannot be late. Pressing the button currently has no effect on the vehicle's actions, and there is no effect visible to the player.

Under the hood, it sets `lateness_counter` which is then reset as soon as the vehicle gets to its first order and starts its timetable.

## Description

- If the current vehicle has not started its timetable, return a descriptive error message. This is an improvement over silently doing nothing without warning the player.
- If iterating through other vehicles in the group, ignore those which have not started their timetables, because their `lateness_counter` will always be a meaningless 0.

Again, this has no effect now besides not setting `lateness_counter` needlessly, but it matters in my current approach to #11341.

## Checklist for review

### DIFF
--- a/src/timetable_cmd.cpp
+++ b/src/timetable_cmd.cpp
@@ -221,6 +221,10 @@ CommandCost CmdSetVehicleOnTime(DoCommandFlag flags, VehicleID veh, bool apply_t
 	Vehicle *v = Vehicle::GetIfValid(veh);
 	if (v == nullptr || !v->IsPrimaryVehicle() || v->orders == nullptr) return CMD_ERROR;
 
+	/* A vehicle can't be late if its timetable hasn't started.
+	 * If we're setting all vehicles in the group, we handle that below. */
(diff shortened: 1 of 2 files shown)


src/timetable_cmd.cpp
```diff
@@ -221,19 +221,28 @@ CommandCost CmdSetVehicleOnTime(DoCommandFlag flags, VehicleID veh, bool apply_t
 	Vehicle *v = Vehicle::GetIfValid(veh);
 	if (v == nullptr || !v->IsPrimaryVehicle() || v->orders == nullptr) return CMD_ERROR;
 
+	/* A vehicle can't be late if its timetable hasn't started. */
+	if (!HasBit(v->vehicle_flags, VF_TIMETABLE_STARTED)) return CommandCost(STR_ERROR_TIMETABLE_NOT_STARTED);
```
PeterN: Feels like this should be conditional upon `apply_to_group`. If that is set then it probably doesn't matter if the first vehicle hasn't started, some of the others in the group may have?

PeterN (approved):

PR #11363, opened 2023-10-09
Change: Filter duplicate paths from valid search path list.

PeterN:
## Motivation / Problem

`_valid_searchpaths` can contain the same path multiple times. This may mean that files in these search paths are also scanned multiple times, maybe including an MD5 check.

## Description

Test if `_valid_searchpaths` already contains a search path before adding it. This leaves `_valid_searchpaths` as a vector instead of directly using a set, in case search order matters.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -84,6 +84,8 @@ static bool IsValidSearchPath(Searchpath sp)
 static void FillValidSearchPaths(bool only_local_path)
 {
 	_valid_searchpaths.clear();
+
+	std::set<std::string> seen{};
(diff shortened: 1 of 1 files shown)


src/fileio.cpp
```diff
@@ -97,7 +99,11 @@ static void FillValidSearchPaths(bool only_local_path)
 			}
 		}
 
-		if (IsValidSearchPath(sp)) _valid_searchpaths.emplace_back(sp);
+		if (IsValidSearchPath(sp)) {
+			if (seen.count(_searchpaths[sp])) continue;
```
TrueBrain: We tend to be explicit about our comparisons .. so `!= 0`?

TrueBrain (approved):

PR #11366, opened 2023-10-13
Fix: Closing network relay popup without pressing a button left connection in limbo.

PeterN:
## Motivation / Problem

When connecting to a server that requires a relay, a popup is shown to the user, if this popup is closed without pressing an action button (e.g. by clicking on the X or right-click close) this leaves the game connection in limbo, as it cannot proceed. It will eventually time out.

## Description

Treat closing the window the same as clicking on "No".

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -2413,6 +2413,12 @@ struct NetworkAskRelayWindow : public Window {
 		this->InitNested(0);
 	}
 
+	void Close(int data = 0) override
+	{
+		if (data == NRWCD_UNHANDLED) _network_coordinator_client.ConnectFailure(this->token, 0);
+		this->Window::Close();
+	}
+
 	void UpdateWidgetSize(int widget, Dimension *size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension *fill, [[maybe_unused]] Dimension *resize) override
 	{
 		if (widget == WID_NAR_TEXT) {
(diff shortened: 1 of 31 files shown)


src/network/network_gui.cpp
```diff
@@ -2413,6 +2413,26 @@ struct NetworkAskRelayWindow : public Window {
 		this->InitNested(0);
 	}
 
+	void Close([[maybe_unused]] int data = 0) override
```
TrueBrain: not "maybe" unused. Very used :)

What I find difficult to see now, is whether `WID_NAR_YES_ONCE` for example is not also `0`. How do I know that this default value is not used?

Maybe we should use `-1` as default?

PeterN: Probably passing the widget index here is not correct. Defaulting to -1 would be a design pattern change from how OnInvalidateData is used. 

Given we also need to handle CloseWindowByClass() which should not issue ConnectFailure, and is definitely not a widget, I think it makes sense to revert the last commit here and then pass handled/unhandled state instead.

TrueBrain: Still this pesky `maybe_unused` that is annoying me; nice solution otherwise :)

PeterN: Hmm, well, I followed the pattern used elsewhere maybe they are wrong too?

TrueBrain: We really should only add it in places where it is unused, in my opinion. Adding it to places where it is used, is weird?

The parent can still have it, and all other definitions of `Close`, but here it is clearly used? Feels weird to keep it. Before you know it, we just add it everywhere :D

TrueBrain (approved):

PR #11369, opened 2023-10-15
Codechange: Replace refit option list C-array with std::map.

PeterN:
## Motivation / Problem

Refit options list uses a C-style array of std::vectors, where the array size is NUM_CARGO, but the index is not CargoID. This means a separate counter is used in various places to get the correct index.

Indices are used in various places.

## Description

Replace with std::map of std::vector. Potentially this could be a std::multimap, however the tree-like nature of display favours the former construct.

Iteration is used instead of indices.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -550,18 +550,18 @@ struct RefitOption {
 	}
 };
 
-typedef std::vector<RefitOption> SubtypeList; ///< List of refit subtypes associated to a cargo.
+using RefitOptions = std::map<CargoID, std::vector<RefitOption>, CargoIDComparator>; ///< Available refit options (subtype and string) associated with each cargo type.
 
 /**
  * Draw the list of available refit options for a consist and highlight the selected refit option (if any).
- * @param list  List of subtype options for each (sorted) cargo.
- * @param sel   Selected refit cargo-type in the window
+ * @param refits Available refit options for each (sorted) cargo.
+ * @param sel   Selected refit option in the window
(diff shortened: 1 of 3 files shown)


src/vehicle_gui.cpp
```diff
@@ -774,45 +766,24 @@ struct RefitWindow : public Window {
...
-
-				/* Hide subtypes if sel[0] does not match */
-				if (this->sel[0] != (int)i && refit.subtype != 0xFF) continue;
-
+		for (const auto &pair : refit_list) {
+			for (const RefitOption &refit : pair.second) {
 				if (row == click_row) {
-					this->sel[0] = i;
-					this->sel[1] = j;
+					this->selected_refit = &refit;
```
rubidium42, line 792: This feels icky to me. Maybe I'm overlooking something, but now  `this->selected_refit` is a pointer to something in `refit_list`. So anything that touched `this->refit_list` should ensure `this->selected_refit` remains valid (or at best gets set to `nullptr`. Given that `BuildRefitList` clears it, doesn't `OnInvalidateData`'s call to `BuildRefitList` cause its subsequent call to `RefreshScrollbar` to read the memory that got freed by `BuildRefitList`?

PeterN: Ok... `selected_refit` is now always revalidated in `BuildRefitList()`. This means the separate code path in `OnInit()` is not necessary.

Alternatively I could keep a copy of the selected item instead of a pointer to it.

rubidium42 (approved):

PR #11373, opened 2023-10-15
Fix: Tile slope missing from road stops varact2 variable 0x42

JGRennison:
## Motivation / Problem

Tile slope part was missing from road stops varact2 variable 0x42.

## Description

Add tile slope part to road stops varact2 variable 0x42.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/newgrf_roadstop.cpp
+++ b/src/newgrf_roadstop.cpp
@@ -90,7 +90,7 @@ uint32_t RoadStopScopeResolver::GetVariable(byte variable, [[maybe_unused]] uint
 			return 2;
 
 		/* Terrain type */
-		case 0x42: return this->tile == INVALID_TILE ? 0 : GetTerrainType(this->tile, TCX_NORMAL); // terrain_type
+		case 0x42: return this->tile == INVALID_TILE ? 0 : (GetTileSlope(this->tile) << 8 | GetTerrainType(this->tile, TCX_NORMAL));
 
 		/* Road type */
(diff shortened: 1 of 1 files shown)


glx22 (approved):

rubidium42:
Just a final check question; is this supposed to make it behave the same as https://jgrennison.github.io/OpenTTD-patches/newgrf-roadstops.html#roadstop_terrain_type?
If so, then it doesn't as OpenTTD writes either 0, 1, 2 or 4 in the first 8 bits whereas that documentation suggests setting specific bits, e.g. `1 << 4` for on/above snowline. Although I wonder whether those specifications are intended that way, since I would find it awkward to be on a tile that is both normal, desert and rainforest.

glx22:
It's using the same method as https://github.com/OpenTTD/OpenTTD/blob/07d6415715059b9c98208b18792cd4102193ba02/src/newgrf_object.cpp#L315

JGRennison:
> Just a final check question; is this supposed to make it behave the same as https://jgrennison.github.io/OpenTTD-patches/newgrf-roadstops.html#roadstop_terrain_type? If so, then it doesn't as OpenTTD writes either 0, 1, 2 or 4 in the first 8 bits whereas that documentation suggests setting specific bits, e.g. `1 << 4` for on/above snowline. Although I wonder whether those specifications are intended that way, since I would find it awkward to be on a tile that is both normal, desert and rainforest.

The table at that link is incorrect, I'll have it fixed shortly

PR #11377, opened 2023-10-16
Change: Always allow expanding towns in Scenario Editor to build new roads

### LINKED ISSUE #11374: [Bug]: Cannot expand towns in Scenario Editor when town road construction disallowed
### Version of OpenTTD

13.4

### Expected result

In Scenario Editor, when I open a town window and press Expand, the town gets bigger, building more roads if needed.

If I have disabled the setting `Towns are allowed to build roads`, the town expands anyway, because according to the documentation of the setting, it is `always allowed when generating world / in SE`: https://github.com/OpenTTD/OpenTTD/blob/master/src/settings_type.h#L546

### Actual result

If I have disabled town road construction, the town does not get larger and instead gives a warning popup: `The town will not build roads. You can enable building of roads via Settings->Environment->Towns`.

I guess there might be a use case for expanding a town without building roads, so the resolution for this issue might be to fix the setting documentation.

### Steps to reproduce

1. Open Scenario Editor
2. Open Settings menu and disable `Towns are allowed to build roads`
3. Build a town
4. Click on the town name to open the town window
5. Click Expand
6. Note that the town does not grow and you get an error message

2TallTyler:
## Motivation / Problem

As described in #11374, the setting `Towns are allowed to build roads` blocks expansion in Scenario Editor and shows a warning about this behavior. The setting documentation says it doesn't apply in Scenario Editor.

## Description

Change the behavior to match the documentation, so that the Expand button in Scenario Editor constructs roads regardless of the `Towns are allowed to build roads` setting.

Closes #11374.

## Limitations

[Spacebar heating](https://xkcd.com/1172/), etc.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -4866,7 +4866,6 @@ STR_ERROR_TOO_CLOSE_TO_EDGE_OF_MAP_SUB                          :{WHITE}... too 
 STR_ERROR_TOO_CLOSE_TO_ANOTHER_TOWN                             :{WHITE}... too close to another town
 STR_ERROR_TOO_MANY_TOWNS                                        :{WHITE}... too many towns
 STR_ERROR_NO_SPACE_FOR_TOWN                                     :{WHITE}... there is no more space on the map
-STR_ERROR_TOWN_EXPAND_WARN_NO_ROADS                             :{WHITE}The town will not build roads. You can enable building of roads via Settings->Environment->Towns
 STR_ERROR_ROAD_WORKS_IN_PROGRESS                                :{WHITE}Road works in progress
 STR_ERROR_TOWN_CAN_T_DELETE                                     :{WHITE}Can't delete this town...{}A station or depot is referring to the town or a town owned tile can't be removed
 STR_ERROR_STATUE_NO_SUITABLE_PLACE                              :{WHITE}... there is no suitable place for a statue in the centre of this town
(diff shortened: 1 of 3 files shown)


frosch123:
Actually, JGRPP integrates this patch ( https://www.tt-forums.net/viewtopic.php?f=33&t=68894 ), so maybe scenario authors would use that instead.

zephyris:
[Fixing a bug always breaks a workflow](https://xkcd.com/1172/).
Surely the 'proper' solution is better scenario editor tools for towns. Three buttons: expand town, expand roads, expand buildings.

PeterN (approved):
Sames a bit of repetition too

EmperorJake:
This was an unnecessary change. All it does is remove choice from the player, and benefits no-one. https://old.reddit.com/r/openttd/comments/1e87vuy/towns_building_roads_in_editor/

LordActive:
It's a shame that a playstyle is completely killed because of a short line (in brackets even) inside the documentation. I think the settings were fine as they were.

Town doesn't expand: Either build the roads yourself or enable road building yourself, simple as. Don't see how it's a problem, this didn't make anything different for people who want to build maps/towns the "conventional" way in SE.

The only thing this change has done was to hurt people who like to make maps like Cindini (which I find to be the greatest map ever made and I'm sure I'm not the only RP-oriented player to hold that opinion).

PR #11378, opened 2023-10-17
Change: Decouple town cargo production from cargo type.

PeterN:
## Motivation / Problem

Towns are hardcoded to produce CT_PASSENGERS and CT_MAIL. If those cargo types are not available, weird things happen.

## Description

This change adds a Town Production Effect property that tells a town to produce a cargo as if it were passengers or mail. No adjustment is made in case of multiple candidate cargos, you will just get higher production.

This is distinct from Town Acceptance Effect (nee Town Effect) which controls what happens to cargo received by a town. This separating is necessary to not break existing NewGRFs, and to allow flexibility.

This is also applied for generating passenger subsidies, although instead of creating more subsidies, that simply picks one at random.

This means the game no longer particularly cares about CT_PASSENGERS or CT_MAIL and goes with the town production effect.

## Limitations

The NewGRF variables for town production still deal only with CT_PASSENGERS/CT_MAIL, apparently these are meant to be avoided anyway.

## Checklist for review

### DIFF
--- a/src/cargotype.h
+++ b/src/cargotype.h
@@ -33,6 +33,20 @@ enum TownAcceptanceEffect : byte {
 	NUM_TAE = TAE_END, ///< Amount of town effects.
 };
 
+/** Town effect when producing cargo. */
+enum TownProductionEffect : byte {
+	TPE_NONE, ///< Town will not produce this cargo type.
+	TPE_PASSENGERS, ///< Cargo behaves passenger-like for production.
+	TPE_MAIL, ///< Cargo behaves mail-like for production.
+	NUM_TPE,
+
+	/**
+	 * Invalid town production effect. Used as a sentinel to indicate if a NewGRF has explicitly set an effect.
+	 * This does not 'exist' after cargo types are finalised.
+	 */
+	INVALID_TPE,
+};
+
 /** Cargo classes. */
 enum CargoClass {
 	CC_NOAVAILABLE  = 0,       ///< No cargo class has been specified
@@ -65,6 +79,7 @@ struct CargoSpec {
 
 	bool is_freight;                 ///< Cargo type is considered to be freight (affects train freight multiplier).
 	TownAcceptanceEffect town_acceptance_effect; ///< The effect that delivering this cargo type has on towns. Also affects destination of subsidies.
+	TownProductionEffect town_production_effect{INVALID_TPE}; ///< The effect on town cargo production.
 	uint8_t callback_mask;             ///< Bitmask of cargo callbacks that have to be called
 
 	StringID name;                   ///< Name of this type of cargo.
@@ -170,6 +185,9 @@ struct CargoSpec {
 	 * @return an iterable ensemble of all valid CargoSpec
 	 */
 	static IterateWrapper Iterate(size_t from = 0) { return IterateWrapper(from); }
+
+	/** List of cargo specs for each Town Product Effect. */
+	static std::array<std::vector<const CargoSpec *>, NUM_TPE> town_production_cargoes;
 
 private:
 	static CargoSpec array[NUM_CARGO]; ///< Array holding all CargoSpecs
--- a/src/cargotype.h
+++ b/src/cargotype.h
@@ -21,16 +21,16 @@
 typedef uint32_t CargoLabel;
 
 /** Town growth effect when delivering cargo. */
-enum TownEffect : byte {
-	TE_BEGIN = 0,
-	TE_NONE = TE_BEGIN, ///< Cargo has no effect.
-	TE_PASSENGERS,      ///< Cargo behaves passenger-like.
-	TE_MAIL,            ///< Cargo behaves mail-like.
-	TE_GOODS,           ///< Cargo behaves goods/candy-like.
-	TE_WATER,           ///< Cargo behaves water-like.
-	TE_FOOD,            ///< Cargo behaves food/fizzy-drinks-like.
-	TE_END,             ///< End of town effects.
-	NUM_TE = TE_END,    ///< Amount of town effects.
+enum TownAcceptanceEffect : byte {
+	TAE_BEGIN = 0,
+	TAE_NONE = TAE_BEGIN, ///< Cargo has no effect.
+	TAE_PASSENGERS, ///< Cargo behaves passenger-like.
+	TAE_MAIL, ///< Cargo behaves mail-like.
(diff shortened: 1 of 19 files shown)


PeterN:
Improved Town Industries replaces the slot used by CT_MAIL with waste. Because this is not mail, the town produces it anyway.

13.4:
image
This PR:
image

OTIS sets multiple cargo types to TE_PASSENGERS/TE_MAIL

13.4:
image
This PR:
image

Eddi-z:
> This will probably break some NewGRFs as towns will unexpectedly start producing more cargo types...

I'm not sure whether this automatically disqualifies the approach.

2TallTyler:
I agree. My creative abuse in Improved Town Industries is not necessarily behavior we need to continue to support. 😉

Andrew350:
Is it possible to apply a preview tag to a draft PR? Would be nice to play with this a bit to see the effects first hand 🙂

glx22:
It's possible to add preview label, but preview can't access content server. You would only test changes without newgrf and I think there's nothing to see in this case.

Andrew350:
Hmm, ok. I could have sworn there was a way to add NewGRFs to the web version but maybe I misremembered 🙂

glx22:
Added preview label (seems they can actually access content)

PeterN:
This is now updated to use a separate property from Town Effect, which avoids the main issue of behaviour change with existing NewGRFs.

2TallTyler:
I posted this on Discord, but for the record: I like this proposal a lot. We already allow houses to have custom production using a callback, but in practice this is rarely used to its full potential. This is because industry GRFs are typically the ones which define cargos and cargo chains, and do not include houses. Most industry GRF authors do not have their own house set, and vice versa, so there is little coordination between sets.

This PR gives industry GRFs a sensible tool to make towns produce cargos, without any footguns that I see and certainly with less issues than trying to do both. I did both in Improved Town Industries 1 and required players to use both it and ITL Houses, and it was such a nightmare that I overwrite ITI 1 with ITI 2, which does the mail ID slot kludge. (ITI 1 was recently resurrected as [ITI Classic](https://github.com/2TallTyler/iti-classic), which bundles both industries and houses into one GRF file, to appease players who asked for it).

ITI 2 will need an update if this PR is merged, but I've already prepared [a PR](https://github.com/2TallTyler/improved_town_industries/pull/104) so as long as we update NML before this patch is released, it's no problem. 🙂

src/cargotype.h
```diff
@@ -22,16 +22,16 @@
...
+enum TownAcceptanceEffect : byte {
+	TAE_BEGIN = 0,
+	TAE_NONE = TAE_BEGIN, ///< Cargo has no effect.
+	TAE_PASSENGERS,      ///< Cargo behaves passenger-like.
+	TAE_MAIL,            ///< Cargo behaves mail-like.
+	TAE_GOODS,           ///< Cargo behaves goods/candy-like.
+	TAE_WATER,           ///< Cargo behaves water-like.
+	TAE_FOOD,            ///< Cargo behaves food/fizzy-drinks-like.
+	TAE_END,             ///< End of town effects.
+	NUM_TAE = TAE_END,    ///< Amount of town effects.
```
rubidium42: This made the alignment weird; I guess, you'd want to remove the extra spaces all together.

src/town.h
```diff
@@ -73,8 +73,8 @@ struct Town : TownPool::PoolItem<&_town_pool> {
 	int16_t ratings[MAX_COMPANIES];  ///< ratings of each company for this town
 
 	TransportedCargoStat<uint32_t> supplied[NUM_CARGO]; ///< Cargo statistics about supplied cargo.
-	TransportedCargoStat<uint16_t> received[NUM_TE];    ///< Cargo statistics about received cargotypes.
-	uint32_t goal[NUM_TE];                              ///< Amount of cargo required for the town to grow.
+	TransportedCargoStat<uint16_t> received[NUM_TAE];    ///< Cargo statistics about received cargotypes.
+	uint32_t goal[NUM_TAE];                              ///< Amount of cargo required for the town to grow.
```
rubidium42: More messed up alignment.

src/town_cmd.cpp
```diff
@@ -518,6 +518,43 @@ static void AdvanceHouseConstruction(TileIndex tile)
...
+	for (const CargoSpec *cs : _town_production_cargoes[tpe]) {
+		CargoID cid = cs->Index();
+		uint32_t r = Random();
+
+		/* Make a bitmask with up to 32 bits set, one for each potential pax. */
+		int genmax = (rate + 7) / 8;
+		uint32_t genmask = (genmax >= 32) ? 0xFFFFFFFF : ((1 << genmax) - 1);
+
+		/* Mask random value by potential pax and count number of actual pax. */
+		uint amt = CountBits(r & genmask);
```
rubidium42, line 586: Wow... what a convoluted way of essentially doing `RandomRange`, though I acknowledge that it's just moved code.
It's something like: `uint amt = RandomRange(1 + std::min(32, (rate + 7) / 8));`, right? The +1 because `RandomRange`'s limit is not included in the resulting values.

PeterN: Must admit I didn't try to understand what it wants to do!

rubidium42: Never mind-ish. It is getting a random number is the range, but instead of a even chance there is a binominal spread. In other words, the middle of the range is way more likely than the extremities of the range.
It would have been better if that got wrapped up in a separate function in random_func, but for now I'd say moving it is fine.

PeterN:
Updated to remove NewGRF properties added on the sly. That can come as a separate Feature.

PeterN:
Needed to update as #11719 (which depends on this) conflicts with master :)

TrueBrain (approved):
I have no argument against this, and I know you will fix problems that may come from this. So let's go!

2TallTyler:
It's already in the milestone, but I'd just like to note that #11947 is also required in order to not break Improved Town Industries. (I'll need to add the flags, but that's easy.) Totally biased, because it's my GRF, but if we break it I'll get all the bug reports. 😛

TrueBrain:
> It's already in the milestone, but I'd just like to note that #11947 is also required in order to not break Improved Town Industries. (I'll need to add the flags, but that's easy.) Totally biased, because it's my GRF, but if we break it I'll get all the bug reports. 😛

Well, I have some excellent news for you! You can review that PR yourself! Isn't it absolutely amazing? :D

PR #11379, opened 2023-10-18
Change: Use CARGO_LIST to show station cargo acceptance changes.

PeterN:
## Motivation / Problem

When determining station cargo acceptance changes for new alerts, only changes for up to 2 cargo types can be reported. The process also creates a (two-entry) list of cargo types and uses different strings depending on if there's 1 or 2 cargos.

## Description

Instead, replace the two different strings with one string using `{CARGO_LIST}`. This accepts a bit-mask of cargo types, so can report on all station cargo acceptance changes.

Because the process now uses bit-masks exclusively, simple logical operators can be use to determine the changes, instead of running through a loop.

This simplifies construction of the news message and allows for more than two changes to be show in one line.

## Limitations

The presentation of the list of cargo is slightly different, being purely comma separated.

With the old approach, the news message is more natural "Station no longer accepts Passengers and Mail".
With this new approach, the news message becomes "Station no longer accepts: Passnegers, Mail".

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -518,16 +518,16 @@ CargoTypes GetEmptyMask(const Station *st)
 }
 
 /**
- * Items contains the two cargo names that are to be accepted or rejected.
- * msg is the string id of the message to display.
+ * Add news item for when a station changes which cargoes it accepts.
+ * @param st Station of cargo change.
+ * @param cargoes Bit mask of cargo types to list.
+ * @param reject True iff the station rejects the cargo types.
  */
-static void ShowRejectOrAcceptNews(const Station *st, uint num_items, CargoID *cargo, StringID msg)
+static void ShowRejectOrAcceptNews(const Station *st, CargoTypes cargoes, bool reject)
 {
-	for (uint i = 0; i < num_items; i++) {
-		SetDParam(i + 1, CargoSpec::Get(cargo[i])->name);
-	}
-
 	SetDParam(0, st->index);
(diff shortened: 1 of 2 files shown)


src/station_cmd.cpp
```diff
@@ -518,16 +518,15 @@ CargoTypes GetEmptyMask(const Station *st)
 }
 
 /**
- * Items contains the two cargo names that are to be accepted or rejected.
- * msg is the string id of the message to display.
+ * Add news item for cargo accepted or rejected changes of a station.
+ * @param st Station of cargo change.
+ * @param cargoes Bit mask of cargo types to list.
+ * @param msg String IDk for either accepts or rejects message.
```
2TallTyler:
```suggestion
 * @param msg String ID to show, either adding or removing cargos from the station's acceptance.
```

This might make more sense as a `bool remove` with the String IDs specified here, instead of being passed from `UpdateStationAcceptance()`.

src/station_cmd.cpp
```diff
@@ -518,16 +518,15 @@ CargoTypes GetEmptyMask(const Station *st)
 }
 
 /**
- * Items contains the two cargo names that are to be accepted or rejected.
- * msg is the string id of the message to display.
+ * Add news item for cargo accepted or rejected changes of a station.
```
2TallTyler:
```suggestion
 * Add news item when a station changes which cargos it accepts.
```

2TallTyler (approved):

PR #11380, opened 2023-10-18
Codechange: Use iterators and/or range-for on cargo related loops.

PeterN:
## Motivation / Problem

Lots of loops from 0 (there's no CARGO_BEGIN) through to NUM_CARGO used as array indices, which can instead be iterated.

## Description

Use explicit iteration, or range-for, where possible to avoid using CargoID as an array indices.

This doesn't cover all occurrences, mostly those that loop and then use CargoSpec(cid), or those that simply don't care about the CargoID.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -496,8 +496,8 @@ CargoTypes GetAcceptanceMask(const Station *st)
 {
 	CargoTypes mask = 0;
 
-	for (CargoID i = 0; i < NUM_CARGO; i++) {
-		if (HasBit(st->goods[i].status, GoodsEntry::GES_ACCEPTANCE)) SetBit(mask, i);
+	for (auto it = std::begin(st->goods); it != std::end(st->goods); ++it) {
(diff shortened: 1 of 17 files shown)


src/station_cmd.cpp
```diff
@@ -496,8 +496,8 @@ CargoTypes GetAcceptanceMask(const Station *st)
 {
 	CargoTypes mask = 0;
 
-	for (CargoID i = 0; i < NUM_CARGO; i++) {
-		if (HasBit(st->goods[i].status, GoodsEntry::GES_ACCEPTANCE)) SetBit(mask, i);
+	for (auto it = std::begin(st->goods); it != std::end(st->goods); ++it) {
+		if (HasBit(it->status, GoodsEntry::GES_ACCEPTANCE)) SetBit(mask, it - std::begin(st->goods));
```
rubidium42: Maybe use `std::distance`? Also around line 515.

rubidium42 (approved):

PR #11389, opened 2023-10-22
Cleanup: Initial untangling of town_cmd.cpp

2TallTyler:
## Motivation / Problem

While working on the Scenario Editor features suggested in #11377, I discovered that `town_cmd.cpp` is a hot mess.

Before I can even consider refactoring town growth code, the entire file needs some cleanup.

## Description

This is a first pass at cleaning up the file. Each task is a separate commit for ease of review. I plan to rebase when merging to keep the history clear.

Yes, my last documentation commit got a bit out of hand; adding missing doxygen turned into fixing grammar and then fixing punctuation... 😬 I hope it's not too hard to review. 😃 

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -481,14 +481,14 @@ static void RemoveNearbyStations(Town *t, TileIndex tile, BuildingFlags flags)
 }
 
 /**
- * Helper function for house completion stages progression
- * @param tile TileIndex of the house (or parts of it) to "grow"
+ * Helper function for house construction stage progression.
+ * @param tile TileIndex of the house (or parts of it) to construct.
  */
-static void MakeSingleHouseBigger(TileIndex tile)
+static void AdvanceSingleHouseConstruction(TileIndex tile)
 {
 	assert(IsTileType(tile, MP_HOUSE));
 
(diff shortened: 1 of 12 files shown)


src/town_cmd.cpp
```diff
@@ -249,7 +249,7 @@ static TownDrawTileProc * const _town_draw_tile_procs[1] = {
  */
 static inline DiagDirection RandomDiagDir()
 {
-	return (DiagDirection)(3 & Random());
+	return (DiagDirection)(Random() & (DIAGDIR_END - 1));
```
rubidium42: I know the code is technically correct... but the trickery with `DIAGDIR_END` and then doing a bitmask tingles my spider senses. What if DIAGDIR_END would become 6 (because we use hexagonal tiles). Then this code becomes horribly broken. However, just using modulo might not be what you really want to express.
You actually want to say: `return (DiagDirection)RandomRange(DIAGDIR_END);` because you just want a random number that is in the range of valid DiagDirection values.

rubidium42 (commented):
Rest seems okay to me.

rubidium42 (approved):

PR #8952, opened 2021-04-05
Social rich presence plugins

nielsmh:
## Motivation / Problem
There's long been interest in being able to "broadcast" your status of playing OpenTTD and possibly being in a multiplayer server, to friends on Discord and similar social platforms. This is even more relevant with the recent Steam release, where features that could allow friends to join your current game, or you to directly send invites, would be popular.

## Description
This PR adds a generic interface for OpenTTD to load a "social plugin", an external dynamic library that implements whatever integration. This PR does not integrate with anything in itself. This level of indirection is to avoid having to potentially link OpenTTD directly with proprietary libraries that might not be compatible with GPL.

A proof-of-concept integration for Discord is here: https://github.com/nielsmh/ottd-discord

## Limitations
Can only load one social plugin currently, it might be better to support loading and updating any number of social plugins, for example to have Steam and Discord integrations active at the same time.

Needs a per-platform loader for the social plugins, that finds and loads the dynamic libraries. Currently only implemented for Win32.

Support for game invites is not complete, some more specification is needed there.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* ~~The bug fix is important enough to be backported? (label: 'backport requested')~~
* ~~This PR affects the save game format? (label 'savegame upgrade')~~
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* ~~This PR affects the NewGRF API? (label 'needs review: NewGRF')~~

Issue #7735, opened 2019-09-11
Protocol handler to join network games


nielsmh:
It could be some good UI to support a protocol handler to join network games.
A protocol handler could allow players to make links to a server directly from a website or social service and have players join through that. It would also make the browser at http://servers.openttd.org/ more accessible.

## URI scheme
Example of what an OpenTTD server link could look like:
`openttd://127.0.0.1:3981?version=1.9.3&password=asdf&join=4&companypassword=qwerty`

Suggested parts:
- `openttd:` because a scheme name is required
- `//` because everyone else does it
- `127.0.0.1:3981` IP address (or hostname) and port
- `?` make a HTTP-like query string
- `version=1.9.3` could in the future allow having a meta-handler that selects between multiple installed versions of the game for the required one
- `password=` server password, leave out if not required (prompt user if server needs one or the provided is wrong)
- `join=` join a company (leave out to spectate)
- `join=new` create a new company on connect, if possible
- `companypassword=` password to join company, or to set on the new company created (spectate if provided password is wrong)

## Implementation
Every OS has its own way of registering URI handlers, the installers/packaging would have to handle that.

The implementation in game would be a commandline parameter such as `-U <uri>`. The game should perhaps ask the player for confirmation before beginning to connect, so they don't accidentally connect to random servers.

sdcloudt:
Working on it.

JMcKiern:
I was looking into this over the weekend. I have a working parser but I couldn't figure out how to call [`NetworkClientConnectGame()`](https://github.com/OpenTTD/OpenTTD/blob/faaec445d669a222718c2634eb059c44c00a43b7/src/network/network.cpp#L664) at the right time. I tried calling it before [`VideoDriver::GetInstance()->MainLoop()`](https://github.com/OpenTTD/OpenTTD/blob/a38122e6f21374440f5d644b5d54b94ebfac4300/src/openttd.cpp#L858) to see what would happen but it didn't seem to work. Maybe there is some network initialization that needs to be performed beforehand?

Presumably, the main menu needs to load first and *then* this function needs to be called. 
Any ideas on how to "queue" this join game command?

Is there perhaps a function similar to [`ShowQuery()`](https://github.com/OpenTTD/OpenTTD/blob/13cc8a0ceec90def39cbcb84135a0bf039793a6f/src/misc_gui.cpp#L1260) that would allow passing of the game server info to the callback function?

nielsmh:
[...]

PR #11391, opened 2023-10-23
Implement social platform support

yeah-its-gloria:
## Motivation / Problem
Telling others IP addresses or invite codes do be inconvenient.

## Description
This adds an API interface for OpenTTD that can load an arbitrary number of social interfaces through DLLs.

## Limitations
Joining without being inside a game isn't possible yet; for Steam there's things like [ISteamApps::GetLaunchCommandLine](https://partner.steamgames.com/doc/api/ISteamApps#GetLaunchCommandLine), however, this has to be explicitly enabled by whoever controls the Steam release of this game on Steamworks first.
The implementation featuring Steam is also somewhat basic and barebones, e.g Steam generally likes being given a `status` or `steam_display` field for its presence display in the friend list. This, however, would require localization efforts, both within OpenTTD and on OpenTTD's Steamworks page; [ISteamFriends::SetRichPresence](https://partner.steamgames.com/doc/api/ISteamFriends#SetRichPresence) elaborates further on this.

## Checklist for review
Despite introducing new components to the options file primarily, this PR does no changes to existing content.

This would largely replace the (seemingly abandoned) efforts of PR #8952.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+/.cache
 /.vs
+/.vscode
 /build*
 CMakeSettings.json
 docs/aidocs/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-/.cache
 /.vs
-/.vscode
 /build*
 CMakeSettings.json
 docs/aidocs/*
--- a/src/network/social/loader.cpp
+++ b/src/network/social/loader.cpp
@@ -1,0 +1,52 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file network_chat_gui.cpp GUI for handling chat messages. */
+
+#include "../../stdafx.h"
+#include "../network_func.h"
+#include "loader.h"
+
+void callback(const char* serverName) {
+	NetworkClientConnectGame(serverName, COMPANY_SPECTATOR);
+}
+
+SocialPlatformLoader::SocialPlatformLoader() {
+	LoadSocialPlatforms(this->plugins);
+
+	for (SocialPlatformPlugin plugin : plugins) {
+		plugin.initialize(callback, &plugin.userdata);
+	}
+}
+
+void SocialPlatformLoader::Shutdown() {
+	for (SocialPlatformPlugin plugin : plugins) {
+		plugin.shutdown(plugin.userdata);
+	}
+}
+
+void SocialPlatformLoader::RunDispatch() {
+	for (SocialPlatformPlugin plugin : plugins) {
+		plugin.dispatch(plugin.userdata);
+	}
+}
+
+void SocialPlatformLoader::NewState(eventCode event, void* parameter) {
+	for (SocialPlatformPlugin plugin : plugins) {
+		plugin.newState(event, parameter, plugin.userdata);
+	}
+}
+
+SocialPlatformLoader* SocialPlatformLoader::GetInstance() {
+	static SocialPlatformLoader* loader = nullptr;
+
+	if (loader == nullptr) {
+		loader = new SocialPlatformLoader();
+	}
+
+	return loader;
+}
--- a/src/network/social/loader.cpp
+++ b/src/network/social/loader.cpp
@@ -16,7 +16,9 @@ void callback(const char* serverName) {
 }
 
 SocialPlatformLoader::SocialPlatformLoader() {
+#if !defined(__EMSCRIPTEN__)
 	LoadSocialPlatforms(this->plugins);
+#endif
 
 	for (SocialPlatformPlugin plugin : plugins) {
 		plugin.initialize(callback, &plugin.userdata);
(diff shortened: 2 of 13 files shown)


TrueBrain:
Although I really appreciate this effort, there is a reason why we didn't implement this yet. Most of the conversation has been in https://github.com/OpenTTD/OpenTTD/pull/8952 about this.

Mainly, Steam SDK doesn't allow linking against GPL. So that makes for a bit of a pita how we integrate into it. In the above mentioned PR we started the work to make this possible .. by basically no linking against the SDK directly. How this would look, is that when you compile OpenTTD, you won't see anything about any social plugins.

But when you put a magic DLL / so file in a magic folder, all of a sudden Steam integration starts to work. It is a bit .. evading the rules, if you will, but it seems the default way to comply with all licenses involved. But it would also mean OpenTTD would start shipping a non-GPL part, to make this happen (in a certain extend, we already do with content on BaNaNaS).

During shipping of OpenTTD on Steam, we will ofc add this file to the right place, so it works there by default.

I wrote much more about this here: https://gist.github.com/TrueBrain/bccb44cf813cb556613dcb89889bd357

yeah-its-gloria:
After some Discussion:tm: on Discord I've gotten around to implementing a plugin interface and a [Steamworks plugin implementation](https://github.com/yeah-its-gloria/ottd-social-steam).

src/network/social/loader.h
```diff
@@ -0,0 +1,44 @@
...
+
+	OTTD_Social_Initialize initialize;
+	OTTD_Social_Shutdown shutdown;
+	OTTD_Social_Dispatch dispatch;
+	OTTD_Social_NewState newState;
+
+	void* userdata;
+};
+
+class SocialPlatformLoader : public ZeroedMemoryAllocator {
```
nielsmh: You don't _need_ to use ZeroedMemoryAllocator as base class, and I think it might actually be Undefined Behavior to use it with a class that includes non-trivially-constructible members such as a std::vector.

src/network/social/loader.cpp
```diff
@@ -0,0 +1,52 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file network_chat_gui.cpp GUI for handling chat messages. */
```
nielsmh: Remember to update the doc comments.

src/network/social/loader.h
```diff
@@ -0,0 +1,44 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file network_client.h Client part of the network protocol. */
```
nielsmh: Remember to update the doc comments.

src/os/windows/social_win.cpp
```diff
@@ -0,0 +1,53 @@
...
+				(OTTD_Social_NewState)GetProcAddress(library, "OTTD_Social_NewState"),
+
+				nullptr
+			};
+
+			if (plugin.initialize == nullptr || plugin.shutdown == nullptr || plugin.dispatch == nullptr || plugin.newState == nullptr) {
+				FreeLibrary(library);
+			}
+
+			plugins.push_back(plugin);
```
nielsmh: This should probably be in an else block, to not add failed libraries to the collection.

nielsmh (commented):
Keep in mind that there needs to be at least one plugin that can be distributed under GPL terms for a patch to be acceptable. So one for Steam isn't enough.

I think the plugin-API you have designed might be a bit too basic and might not be able to support all the features some integrations will want. Remember to describe memory ownership and validity duration of pointers passed.

2TallTyler:
> Keep in mind that there needs to be at least one plugin that can be distributed under GPL terms for a patch to be acceptable. So one for Steam isn't enough.

A possible GPL-licensed plugin to include by default could be #7735.

.gitignore
```diff
@@ -1,4 +1,6 @@
+/.cache
 /.vs
+/.vscode
```
TrueBrain: Please put these in your global gitignore, and not here. Your choice of IDE is a private business :D

src/os/windows/social_win.cpp
```diff
@@ -0,0 +1,53 @@
...
+			std::string library_path = search_dir + FS2OTTD(find_data.cFileName);
+			HMODULE library = LoadLibraryW(OTTD2FS(library_path).c_str());
+			if (library == nullptr) {
+				continue;
+			}
+
+			SocialPlatformPlugin plugin = {
+				library,
+
+				(OTTD_Social_Initialize)GetProcAddress(library, "OTTD_Social_Initialize"),
```
glx22: We have a `DllLoader` class in win32.h to handle DLLs, you can see how it's used in crashlog_win.cpp (and it takes care of all the casting stuff).
I think it should be possible to make `SocialPlatformPlugin` a derived class of `DllLoader`.

yeah-its-gloria:
PR #11628 is doing a much better job of what I was trying to do, so I'm closing it as such. Wishing everyone involved there good luck and all :)

PR #11393, opened 2023-10-24
Fix #11392: Make neutral stations behave as industry-exclusivity company.

### LINKED ISSUE #11392: [Bug]: SetExclusiveConsumer, SetExclusiveSupplier do not work correctly with oil rigs. Gamescript.
### Version of OpenTTD

13.4

### Expected result

1. I expected that if you create an exclusive supplier and consumer for an oil rig, then only this company will be able to deliver and pick up products from the oil rig.

### Actual result

About SetExclusiveConsumer. Neutral stations do not work, any company cannot take oil if the industry has a set exclusive consumer. It only works if you build a separate station (trucks, trains, etc.). 

About SetExclusiveSupplier. This is completely ignored. There is no difference whether it is a neutral station or a separate one, any company can supply passengers and make a profit.

### Steps to reproduce

1. I launch the local server.
2. I use the event GSEvent.ET_INDUSTRY_OPEN
3. Funding a new oil rig
4. Setting up an exclusive supplier and consumer by using GSIndustry.SetExclusiveConsumer(), GSIndustry.SetExclusiveSupplier().
5. I create two companies and try to get oil and supply passengers.

PeterN:
## Motivation / Problem

Industry exclusivity ignores neutral stations.

## Description

When testing station ownership for exclusivity, pretend a neutral station's owner is the same as its exclusivity consumer/supplier.

This means that goods can still be transfer between station and industry when exclusivity is setup, but this doesn't actually check who is consuming or supplying.

So we also check during (un)load that the vehicle owner matches the exclusivity company too.

## Limitations

This is NOT TESTED.

## Checklist for review

### DIFF
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -1635,10 +1635,14 @@ static void LoadUnloadVehicle(Vehicle *front)
 	StationID last_visited = front->last_station_visited;
 	Station *st = Station::Get(last_visited);
 
+	/* If this is a neutral station, check its industry for exclusivity. */
+	bool can_load   = front->owner == st->GetExclusivityConsumer();
+	bool can_unload = front->owner == st->GetExclusivitySupplier();
+
 	StationIDStack next_station = front->GetNextStoppingStation();
 	bool use_autorefit = front->current_order.IsRefit() && front->current_order.GetRefitCargo() == CT_AUTO_REFIT;
 	CargoArray consist_capleft{};
-	if (_settings_game.order.improved_load && use_autorefit ?
+	if (can_load && _settings_game.order.improved_load && use_autorefit ?
 			front->cargo_payment == nullptr : (front->current_order.GetLoadType() & OLFB_FULL_LOAD) != 0) {
 		ReserveConsist(st, front,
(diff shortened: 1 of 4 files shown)


src/economy.cpp
```diff
@@ -1747,7 +1751,7 @@ static void LoadUnloadVehicle(Vehicle *front)
 		}
 
 		/* Do not pick up goods when we have no-load set or loading is stopped. */
-		if (front->current_order.GetLoadType() & OLFB_NO_LOAD || HasBit(front->vehicle_flags, VF_STOP_LOADING)) continue;
+		if (!can_load || front->current_order.GetLoadType() & OLFB_NO_LOAD || HasBit(front->vehicle_flags, VF_STOP_LOADING)) continue;
```
nielsmh, line 1754: Am I reading right that this makes any competitor vehicle that attempts to load finish its loading instantly?
My worry is a denial-of-service where a competitor blocks the helipad with a full load order that will (obviously) never complete.

Another potential DoS in the same category is making a vehicle wait at the helipad with a timetable minimum wait time.

Issue #11350, opened 2023-10-04
[Bug]: OpenTTD doesn't cleanly exit when no graphics set are found


planetmaker:
### Version of OpenTTD

7145b7060aa00c3dbd60e05c51d7c275f016f825 (master/HEAD)

### Expected result

I expected OpenTTD to complain about missing base graphics and exit cleanly or to offer downloading them

### Actual result

OpenTTD remained in limbo:

```
[planetmaker:~/ottd/openttd/build] [base] master* ± git show
commit 7145b7060aa00c3dbd60e05c51d7c275f016f825 (HEAD -> master, origin/master, origin/HEAD)
[planetmaker:~/ottd/openttd/build] [base] master* ± ./openttd 
Error: Failed to find a graphics set. Please acquire a graphics set for OpenTTD. See section 1.4 of README.md.
warning: queue 0x4955220 destroyed while proxies still attached:
  wl_callback@44 still attached
  wl_surface@42 still attached
```

(but it didn't exit)

Backtrace as follows:

```
futex_wait (private=<optimized out>, expected=12, futex_word=0x27a4ae4 <_http_cv+36>) at ../sysdeps/nptl/futex-internal.h:146
Downloading source file /usr/src/debug/glibc-2.38-4.fc39.x86_64/nptl/../sysdeps/nptl/futex-internal.h
146       int err = lll_futex_timed_wait (futex_word, expected, NULL, private);                                                                            
(gdb) bt
#0  futex_wait (private=<optimized out>, expected=12, futex_word=0x27a4ae4 <_http_cv+36>) at ../sysdeps/nptl/futex-internal.h:146
#1  futex_wait_simple (private=<optimized out>, expected=12, futex_word=0x27a4ae4 <_http_cv+36>) at ../sysdeps/nptl/futex-internal.h:177
#2  __GI___pthread_cond_destroy (cond=0x27a4ac0 <_http_cv>) at pthread_cond_destroy.c:53
#3  0x00007f30ffadc13d in __gthread_cond_destroy (__cond=<optimized out>)
    at /usr/src/debug/gcc-13.2.1-3.fc39.x86_64/obj-x86_64-redhat-linux/x86_64-redhat-linux/libstdc++-v3/include/x86_64-redhat-linux/bits/gthr-default.h:885
#4  std::__condvar::~__condvar() (this=<optimized out>, __in_chrg=<optimized out>)
[...]

PR #11398, opened 2023-10-25
Fix: OpenTTD can fail to exit on an error due to mutex locks in threads.

PeterN:
## Motivation / Problem

The game can crash but fail to exit due to waiting on a mutex lock.

## Description

This could be fixed by ensuring that all resources are properly released... or we can just call `_exit()` instead `exit()`. This is already done in other places so shouldn't be too controversial...

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -121,7 +121,7 @@ void UserErrorI(const std::string &str)
 	EM_ASM(if (window["openttd_abort"]) openttd_abort());
 #endif
 
-	exit(1);
+	_exit(1);
 }
 
 /**


PeterN:
This also fixes #11350

rubidium42 (approved):
Nothing wrong with the PR itself, but related...

I reckon the `exit()` in strgen/settingsgen are not affected because there are is no mutex there.
However, it is the `exit()` in `OSOpenBrowser` in  os/unix/unix.cpp is fine? It is for the forked process, so doesn't forking also fork the mutexes and as such run into the same potential problem?
Something similar happens in `DedicatedFork()` in dedicated.cpp, but I think that is safe as it is ran really early during startup and as such unlikely that the offending mutexes have been created/locked yet.

PR #11404, opened 2023-10-28
Change: Try stopping extmidi player with SIGINT first.

PeterN:
## Motivation / Problem

On Linux with extmidi music playback, stopping playback or switching to the next track may leave stuck notes. This is because it's down to the external player to turn those notes off, and we kill the process with SIGTERM which can interrupt this.

## Description

Instead, try sending SIGINT first, and if that fails fall back to SIGTERM then SIGKILL. This can give the player a chance to issue MIDI note-off commands.

This fixes at least pmidi leaving stuck notes on.

## Checklist for review

### DIFF
--- a/src/music/extmidi.cpp
+++ b/src/music/extmidi.cpp
@@ -130,22 +130,37 @@ void MusicDriver_ExtMidi::DoPlay()
 	}
 }
 
+/**
+ * Try to end child process with kill/waitpid for up to 1 second.
+ * @param pid The process ID to end.
+ * @param signal The signal type to send.
+ * @return True if the process has been ended.
+ */
+static bool KillWait(pid_t &pid, int signal)
+{
+	/* First try to stop for about a second;
+	 * 1 seconds = 1000 milliseconds, 50 ms per cycle => 20 cycles. */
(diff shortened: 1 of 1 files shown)


src/music/extmidi.cpp
```diff
@@ -130,23 +130,32 @@ void MusicDriver_ExtMidi::DoPlay()
...
 
+	return false;
+}
+
+void MusicDriver_ExtMidi::DoStop()
+{
+	if (this->pid <= 0) return;
+
+	/* First try to gracefully stop for about five seconds;
+	 * 5 seconds = 5000 milliseconds, 10 ms per cycle => 500 cycles. */
```
rubidium42: Not sure how to solve this exactly, but... with this change the magic 500 and 10 in `DoKillWait` are not documented in the function, but rather here even though they are essentially implementation details of how to get to the 5 seconds. On the other hand, writing down it is going to gracefully try to interrupt, then terminate and if all else fails kill the driver belongs to this function. So I guess the comment ought to be split over the two functions.

rubidium42 (approved):

Issue #11403, opened 2023-10-27
[Bug]: Code contains 3rdparty software without simple opt-out


pemensik:
### Version of OpenTTD

master, 7b4ecb52415f16fd4c776559e75729a2ddad7909

### Expected result

[Fedora Packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling) contains a rule about bundling libraries. If they can be omitted, they should be omitted. After looking into src/3rdparty directory, I have found quite a lot of bundled libraries.

Problem is they cannot be avoided when the same library is present on the system. Such as when it is packaged in linux distribution. At least I haven't found simple way to do it.

### Actual result

If the system can provide appropriate libraries, they should be used from the system.

Fedora has packages for:
- fmt
- json (nlohmann/json.hpp)
- squirrel (modified)

From what I have understood it is not possible to use vanilla squirrel language, but it should be possible for fmt and json to use system library. If it needs specific version, CMake can solve it. I think 3rdparty should be used only when system does not provide them.

### Steps to reproduce

- remove src/3rdparty/fmt and src/3rdparty/nlohmann
- remove references to them from src/3rdparty/CMakeLists.txt
- the game should build if json-devel and fmt-devel are installed.

TrueBrain:
I am happy to read that Fedora has a packaging guide, but honestly .. we want to make it as easy as possible for as many people as possible to get going in OpenTTD. We do not tailor to a single distribution. And this means we sometimes vendor libraries, just because it is more practical for most people; not just a select few.

Vendoring a library is always a balance, and a discussion. First off, we only vendor libraries that are mandatory for building OpenTTD. All optional libraries are (on Linux anyway) using system libraries. This because if you have a library that is too old on your system, you can still compile the game; just without that optional library. lzma is a nice example of this: although you really want this, and it is almost considered required, we still don't vendor it, as you can build without. And if you don't have it, building just works. With a big fat warning you really want lzma, but the game will run fine without .. except for loading most savegames :P

So for the required libraries, we always first try to use vcpkg (for MacOS and Windows) and system libraries (for Linux). But sometimes this just isn't practical. Popular distros tend to lag behind in terms of what version they ship; and you can't expect us to support version (insert one from 3 years ago), just because there is a popular distro that doesn't ship with anything more recent.

So for those cases we vendor the library. And always after discussion.

[...]

PR #11405, opened 2023-10-28
WIP: Allow using system provided libraries

pemensik:
Try detecting fmt and nlohmann/json, also catch2. If they have compatible version, allow using them from the system.

Until it is proved to work well, make it non-default and needs to be enabled by OPTION_PREFER_SYSLIBS=ON.

## Motivation / Problem

Proposes fix for issue https://github.com/OpenTTD/OpenTTD/issues/11403. Allow using tested system libraries if they are compatible. Make it done only when manually requested to not disturb experience for existing builders.

## Description

Adds new options to check fmt, nlohmann and catch2. Moves directories on src/3rdparty in order to be able to use include directives to use ``<fmt/format.h>`` instead of relative path to bundled directory. Still backwards compatible.

## Limitations

At least Fedora Rawhide or 39 cannot provide both fmt in recent enough version and Catch2 in old enough version. I would say that should be solved by upgrading bundled version to Catch 3, but that would require further modifications of tests.

## Checklist for review


pemensik:
Hmm, I am not sure where to fix include paths used in Enscripten builds. I expect there should be parameter somewhere to add appropriate paths if it had not catched those from CMake.

TrueBrain:
To keep comments in the same place, wrote in #11403 why we shouldn't be doing this.

But again, tnx for delving into this!

PR #11406, opened 2023-10-29
Include AppStream metadata info into source distribution

pemensik:
Taken from Flathub package at:
https://github.com/flathub/org.openttd.OpenTTD

With few changes to add more url links and use SPDX license identifier.

## Motivation / Problem
Openttd is a nice graphical application, but lacks AppStream metainfo file in source archive. I think it would be nice to include it directly with the game, so GUI frontends like GNOME Software can show nice information about the game.

## Checklist for review

### DIFF
--- a/media/org.openttd.OpenTTD.metainfo.xml
+++ b/media/org.openttd.OpenTTD.metainfo.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--Copyright 2018-2020 Flathub maintainers-->
+<component type="desktop-application">
+  <id>org.openttd.OpenTTD.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-only</project_license>
+  <name>OpenTTD</name>
+  <summary>Open source simulation game based upon Transport Tycoon Deluxe</summary>
+  <description>
+    <p>
+      OpenTTD is an open source simulation game based upon the popular
+      Microprose game "Transport Tycoon Deluxe", written by Chris Sawyer. It
+      attempts to mimic the original game as closely as possible while
+      extending it with new features.
+    </p>
+  </description>
+  <launchable type="desktop-id">openttd.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://www.openttd.org/screenshots/1.4-02-opengfx-1920x1200.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://www.openttd.org/screenshots/1.9-darkuk-1.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://www.openttd.org/screenshots/1.9-darkuk-2.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://www.openttd.org/screenshots/1.9-darkuk-3.png</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://www.openttd.org/</url>
+  <url type="bugtracker">https://github.com/OpenTTD/OpenTTD/issues</url>
+  <url type="help">https://wiki.openttd.org/en/Manual/</url>
+  <developer_name>The OpenTTD team</developer_name>
+  <update_contact>info@openttd.org</update_contact>
+  <provides>
+    <binary>openttd</binary>
+  </provides>
+  <releases>
+    <release version="13.4" date="2023-08-05"/>
+    <release version="13.3" date="2023-06-11"/>
+    <release version="13.2.1" date="2023-06-11"/>
+    <release version="13.2" date="2023-06-10"/>
+    <release version="13.1" date="2023-04-10"/>
+    <release version="13.0" date="2023-02-05"/>
+    <release version="12.2" date="2022-04-02"/>
+    <release version="12.1" date="2021-11-08"/>
+    <release version="12.0" date="2021-10-17"/>
+    <release version="1.11.2" date="2021-05-03"/>
+    <release version="1.11.1" date="2021-04-18"/>
+    <release version="1.11.0" date="2021-04-01"/>
+    <release version="1.10.3" date="2020-08-09"/>
+    <release version="1.10.2" date="2020-06-01"/>
+    <release version="1.10.1" date="2020-04-13"/>
+    <release version="1.10.0" date="2020-04-01"/>
+  </releases>
+  <content_rating type="oars-1.1"/>
+</component>
--- a/media/org.openttd.OpenTTD.metainfo.xml
+++ b/media/org.openttd.OpenTTD.metainfo.xml
@@ -39,5 +39,7 @@
   </provides>
   <!-- TODO: generate releases from the source automatically -->
   <!-- <releases/> -->
-  <content_rating type="oars-1.1"/>
+  <content_rating type="oars-1.1">
+    <content_attribute id="social-chat">intense</content_attribute>
+  </content_rating>
 </component>
--- a/media/org.openttd.OpenTTD.metainfo.xml
+++ b/media/org.openttd.OpenTTD.metainfo.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--Copyright 2018-2020 Flathub maintainers-->
+<!--Authors:
+  2018-2020 Flathub maintainers
+  2023 Petr Mensik
+-->
 <component type="desktop-application">
   <id>org.openttd.OpenTTD.desktop</id>
   <metadata_license>CC0-1.0</metadata_license>
--- a/media/org.openttd.OpenTTD.metainfo.xml
+++ b/media/org.openttd.OpenTTD.metainfo.xml
@@ -37,23 +37,7 @@
   <provides>
     <binary>openttd</binary>
   </provides>
-  <releases>
-    <release version="13.4" date="2023-08-05"/>
-    <release version="13.3" date="2023-06-11"/>
-    <release version="13.2.1" date="2023-06-11"/>
-    <release version="13.2" date="2023-06-10"/>
-    <release version="13.1" date="2023-04-10"/>
-    <release version="13.0" date="2023-02-05"/>
-    <release version="12.2" date="2022-04-02"/>
-    <release version="12.1" date="2021-11-08"/>
-    <release version="12.0" date="2021-10-17"/>
-    <release version="1.11.2" date="2021-05-03"/>
-    <release version="1.11.1" date="2021-04-18"/>
-    <release version="1.11.0" date="2021-04-01"/>
-    <release version="1.10.3" date="2020-08-09"/>
-    <release version="1.10.2" date="2020-06-01"/>
-    <release version="1.10.1" date="2020-04-13"/>
-    <release version="1.10.0" date="2020-04-01"/>
-  </releases>
+  <!-- TODO: generate releases from the source automatically -->
+  <!-- <releases/> -->
   <content_rating type="oars-1.1"/>
 </component>
--- a/cmake/InstallAndPackage.cmake
+++ b/cmake/InstallAndPackage.cmake
@@ -90,6 +90,10 @@ if(UNIX AND NOT APPLE)
                     ${CMAKE_BINARY_DIR}/media/${BINARY_NAME}.desktop
             DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/applications
             COMPONENT menu)
+    install(FILES
+                    ${CMAKE_SOURCE_DIR}/media/org.openttd.OpenTTD.metainfo.xml
+            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/metainfo
+            COMPONENT docs)
 endif()
 
 if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")


pemensik:
Specifications for the format: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

TrueBrain:
Adding a file for just adding a file doesn't actually do anything. After installation, no other software will know about this file, as it only exists in our source repository. So I am not sure what good adding this file does?

PS: the license indicator is wrong. We are GPLv2, not "or later" :)

pemensik:
The file itself does not do anything, sure. But it can be used by GUI tools presenting the package before installing. Such as GNOME Software. Fedora package adds it downstream at https://src.fedoraproject.org/rpms/openttd/blob/rawhide/f/openttd.spec#_94. Usually this file is contained in sources with the application itself. It is recommended to send it upstream, especially as it can include different language descriptions also. Linux packages should all install these with GUI applications.

Example screenshot how this does look in Gnome Software (Czech language): https://pemensik.fedorapeople.org/gnome-software-openttd-2023-10-29%2012-54-35.png

media/org.openttd.OpenTTD.metainfo.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--Copyright 2018-2020 Flathub maintainers-->
```
TrueBrain: They cannot be the copyright holder if it is included in our source-tree. So you will have to check with them if they are willing to transfer the copyright to us.

media/org.openttd.OpenTTD.metainfo.xml
```diff
@@ -0,0 +1,59 @@
...
+    <release version="12.1" date="2021-11-08"/>
+    <release version="12.0" date="2021-10-17"/>
+    <release version="1.11.2" date="2021-05-03"/>
+    <release version="1.11.1" date="2021-04-18"/>
+    <release version="1.11.0" date="2021-04-01"/>
+    <release version="1.10.3" date="2020-08-09"/>
+    <release version="1.10.2" date="2020-06-01"/>
+    <release version="1.10.1" date="2020-04-13"/>
+    <release version="1.10.0" date="2020-04-01"/>
+  </releases>
```
TrueBrain: This `<release>` blob is impractical for us to maintain. Please remove it.

media/org.openttd.OpenTTD.metainfo.xml
```diff
@@ -0,0 +1,59 @@
...
+  <content_rating type="oars-1.1"/>
```
TrueBrain: We do not have any rating done.  So what does this mean? Seems a bit weird to add.

TrueBrain:
> The file itself does not do anything, sure. But it can be used by GUI tools presenting the package before installing. Such as GNOME Software. Fedora package adds it downstream at https://src.fedoraproject.org/rpms/openttd/blob/rawhide/f/openttd.spec#_94. Usually this file is contained in sources with the application itself. It is recommended to send it upstream, especially as it can include different language descriptions also. Linux packages should all install these with GUI applications.
> 
> Example screenshot how this does look in Gnome Software (Czech language): https://pemensik.fedorapeople.org/gnome-software-openttd-2023-10-29%2012-54-35.png

I understand what it does; my point is that it doesn't do anything in our repo, so that makes it a bit weird to have in our repo. We tend to only have files that we actively do something with. All the other files are for other places to deal with, like distros etc.

For example the `openttd.desktop` file is installed when you build on Linux. Files that aren't used by us, tend to get outdated quickly, and sometimes takes years before someone notices it doesn't work. So we are always a bit reluctant to add these kind of files. If the file is as useful as you say, I would expect we could do something more with it. Like with `openttd.desktop`. Isn't that possible?

media/org.openttd.OpenTTD.metainfo.xml
pemensik: The file is under CC0-1.0 license, so this should be just about having authors specified. But I have created issue for that, so they can react any way.

TrueBrain: Copyright and license are two different things :) They can license it under CC0, but they still hold the copyright :) A bit silly for a file like this, ofc, but they added the copyright, not me :P Tnx for following up on it!

media/org.openttd.OpenTTD.metainfo.xml
pemensik: That can be specified at https://hughsie.github.io/oars/

I think it contains unmoderated chat in multiplayer mode, so it should have following spec:

```
 <content_rating type="oars-1.1">
    <content_attribute id="social-chat">intense</content_attribute>
  </content_rating>
```

TrueBrain: We tend to stay away from any rating, as we never did any official rating. So I rather just not mention it at all, than to start filling in a random rating system with the information we think is valid :)

pemensik:
> For example the `openttd.desktop` file is installed when you build on Linux. Files that aren't used by us, tend to get outdated quickly, and sometimes takes years before someone notices it doesn't work. So we are always a bit reluctant to add these kind of files. If the file is as useful as you say, I would expect we could do something more with it. Like with `openttd.desktop`. Isn't that possible?

Yes, it should be actually installed just like openttd.desktop, at least on UNIX like systems. It should be installed into ``$PREFIX/share/metainfo/``. I haven't found how is openttd.desktop installed, so I did not add installation script as well. I am not sure which all tools can process such file, especially if installed from source locally. But it should be part of the installation anyway.

TrueBrain:
https://github.com/OpenTTD/OpenTTD/blob/master/cmake/InstallAndPackage.cmake#L89

There you go! :)

media/org.openttd.OpenTTD.metainfo.xml
ghisvail: > They can license it under CC0, but they still hold the copyright

"The person who associated a work with this deed has dedicated the work to the public domain by **waiving all of his or her rights to the work worldwide under copyright law**, including all related and neighboring rights, to the extent allowed by law." [CC0-1.0](https://creativecommons.org/publicdomain/zero/1.0/deed)

media/org.openttd.OpenTTD.metainfo.xml
glx22: It might be possible to auto-fill it by parsing changelog when configuring/building

TrueBrain: Please don't. It will break. We will not find out :) We have been around this bush a few times already ;)

ghisvail: These are automatically added by the Flatpak update bot. It's okay to drop them and only maintain the latest release entry.

media/org.openttd.OpenTTD.metainfo.xml
ghisvail: > A bit silly for a file like this

Your judgement call put aside, the copyright header was more of an authorship statement than anything. Copyright is waived as per the quote of the CC0 above. Feel free to remove it.

media/org.openttd.OpenTTD.metainfo.xml
pemensik: Understood, having extra release tags is impractical. But what if we had a tool to generate it instead? Just like shell script:

```
#!/bin/sh
# Generate metainfo releases from git tags
#

# <release version="1.10.1" date="2020-04-13"/>
#
VERSIONS=${1:-10}

get_date() {
	git log -1 --pretty=format:%as $@
}

get_tags() {
	git tag -l | grep -Ev -- '-(RC|beta).*' | sort -r --version-sort
}

print_release() {
	local VER="$1"
	DATE=$(get_date $VER)
	echo "<release version=\"$VER\" date=\"$DATE\"/>"
}

if [ -n "$VERSIONS" ] && [ "$VERSIONS" != all ]; then
	get_tags | head -${VERSIONS} | while read VER; do print_release $VER; done
else
	get_tags | while read VER; do print_release $VER; done
fi
```

Would that be acceptable instead?

TrueBrain: The issue here too is that we have to maintain these things. And we are having a hard time as it is to maintain our current release version etc. We often even forget to the the simplest thing upon release, causing issues like 13.2 / 13.3 / 13.4.

Especially for files that won't show up in a short period of time after release that they are borked, it is very annoying to throw any automation against it.

But maybe more of a question: what value does it add to see the version history in a file like this? If it is just to help GNOME etc to show OpenTTD, why would it want to know the history? (honest question, to be clear)

ghisvail:
> Adding a file for just adding a file doesn't actually do anything. After installation, no other software will know about this file, as it only exists in our source repository. So I am not sure what good adding this file does?
> 
> PS: the license indicator is wrong. We are GPLv2, not "or later" :)

My mistake. Let me suggest dropping its mention in the copyright header model from this [section](https://github.com/OpenTTD/OpenTTD/blob/master/COPYING.md#how-to-apply-these-terms-to-your-new-programs) to avoid confusion.

media/org.openttd.OpenTTD.metainfo.xml
ghisvail: > So I rather just not mention it at all

Flathub requires a rating to pass validation of the AppStream metadata.

> than to start filling in a random rating system with the information we think is valid :)

The specification is versioned, used universally in the Flatpak ecosystem and fits in a [simple webpage](https://hughsie.github.io/oars/generate.html). We can call that many things, but random it is not.

ghisvail: > That can be specified at https://hughsie.github.io/oars/
> 
> I think it contains unmoderated chat in multiplayer mode, so it should have following spec:
> 
> ```
>  <content_rating type="oars-1.1">
>     <content_attribute id="social-chat">intense</content_attribute>
>   </content_rating>
> ```

I have never used OpenTTD online, so I could not verify this. If that's the case, then indeed the OARS metadata need an update.

media/org.openttd.OpenTTD.metainfo.xml
TrueBrain: Btw, if you really need an entry in the releases list, you could do something similar as with `openttd.desktop` (which needs the actual binary name), and have a single entry which is replaced with the current version the game is building for.

This would be a bit of work, as you need to do something similar for `openttd.desktop` in `media/CMakeLists.txt`, and update `cmake/scripts/Desktop.cmake` to also be able to replace the version. But that way you could have the latest version in the list there, in an automated way, that doesn't require any real maintenance from our side.

But if we can just not have the list, that would work too. Fully depends on what it is actually used for :)

media/org.openttd.OpenTTD.metainfo.xml
TrueBrain: So we gathered so far that the rating entry exists so this file can be used to pass validation on Flathub. This isn't clear from the context of this file, but that is fixable.

Sadly, the more pressing issue is that OpenTTD hasn't officially made any comments on how the game is rated. And I feel uncomfortable to do that with a line in a PR that is about a Linux-only file. I guess that should be done more clearly, and more prominent.

Further more, there are many rating systems in the world. We are confronted with most of them when publishing the game on stores. This further complicates the situation, that we have to make sure those ratings are in sync with each other. I guess it was about to happen sooner or later that we need to figure out a few of those rating systems, and put somewhere what the results are. So everyone can align on the same, and avoid exactly what happened in this short thread.

Which means we have two choices for this PR: either we remove the rating, and this PR becomes easier. Or we keep the rating, and we need to figure out (in another PR) how to rate OpenTTD for a few of these rating systems and publish the results somewhere.
This stuff is never easy :(

pemensik: It could be solved by making extra chat plugin, which would not be enabled by default. If some file existed in filesystem, it would enable in-game chat. That way parent-administrator could leave children safe with the game, if they would lack permission to install the extension (remove or install tested file). But I expect children below 13 years are not likely to play this game often. At least compile-time option would be nice for the chat.

ghisvail: >  This stuff is never easy :(

Hence why this [generator](https://hughsie.github.io/oars/generate.html) exists: to save time from upstream developers.

Now, that's a lot of text to say "we don't care". So let's not lose further time on this.

rubidium42: What is the goal of the rating? How pedantic does one need to be when making the rating?

I would say you forgot moderate money-gambling; for example attempting to bribe local authorities which is practically gambling with virtual money on a successful bribe.
And drugs-alcohol due to FIRS having alcohol as cargo, which is essentially consumed by the human units.
And drugs-tobacco due to Industries of the Caribbean having tobacco as cargo. (https://newgrf-specs.tt-wiki.net/wiki/CargoTypes)
And violence-fantasy as you can deliberately crash trains/vehicles and kill many human units.
And sex-nudity as Long road vehicle 4 has vehicle with imagery that could be recognised as naked women, so there's even a version with those images cut (http://george.zernebok.net/newgrf/downloads.html)

And to be honest... intense desecration is feasible with a NewGRF. Just have an industry that accepts passengers, when passengers arrive they are put in a pen and some animals start feasting on them until they are gone. Does that mean OpenTTD should be rated like that, since it is possible with readily downloadable graphics sets?

In any case, I would rather not self-rate the application as it will open a huge can of worms of people who see things differently. In which the developers are caught in the middle as they cannot please both sides, so if it doesn't need to be provided for anything that the OpenTTD developers provide I rather not add this point of contention.

TrueBrain: > It could be solved by making extra chat plugin, which would not be enabled by default. If some file existed in filesystem, it would enable in-game chat. That way parent-administrator could leave children safe with the game, if they would lack permission to install the extension (remove or install tested file). But I expect children below 13 years are not likely to play this game often. At least compile-time option would be nice for the chat.

I never thought about that, but you make a very good point. And it might surprise you how young our players can be .. at least, some of the older players always pride themselves with telling their N-year-old is playing OpenTTD :P Tnx for this sharp remark, it is a bit of an eye-opener (might sound silly to some, but I am just being honest here).

media/org.openttd.OpenTTD.metainfo.xml
pemensik: Yes, at least one release record is mandatory for strictly validated metainfo file. Problem is we need also the release date, not only current version. Having only single record would be a bit poor choice, but better than nothing at all. Ideally the release tag should contain short summary about what is new in that version.

TrueBrain: Short summary we don't have. Like, literally, we don't have it :P We make a short news post poking fun of things, but that is about as far as it gets :D

There is `REV_ISODATE`, which is the date of the current build you are running. For a release that will be the date of the release.

media/org.openttd.OpenTTD.metainfo.xml
pemensik: Of course local games are safe for any children and I expect that would be most common. For online play it would be nice, if chat could be hidden on online game. Maybe just messages from the server shown. You can still make some rude images or words using railroads, but I expect that would not allow unwanted chat with the kid. Current unchecked chat allows sexual harassment among other thing unfortunately.

pemensik: > What is the goal of the rating? How pedantic does one need to be when making the rating?

I think this is a best effort thing. What people responsible for the release agree upon.
> 
> I would say you forgot moderate money-gambling; for example attempting to bribe local authorities which is practically gambling with virtual money on a successful bribe. And drugs-alcohol due to FIRS having alcohol as cargo, which is essentially consumed by the human units. And drugs-tobacco due to Industries of the Caribbean having tobacco as cargo. (https://newgrf-specs.tt-wiki.net/wiki/CargoTypes) And violence-fantasy as you can deliberately crash trains/vehicles and kill many human units. And sex-nudity as Long road vehicle 4 has vehicle with imagery that could be recognised as naked women, so there's even a version with those images cut (http://george.zernebok.net/newgrf/downloads.html)

Well, good point. With online content made by 3rd parties and downloaded in the game, it might require extensions to declare how they are rated themselves. I am not aware there is a way to declare account a child one, with age limit, on a Linux desktop. Similar to what Android offers. Might be good to have downloading of new extensions disabled by administrator decision as well.
> 
> And to be honest... intense desecration is feasible with a NewGRF. Just have an industry that accepts passengers, when passengers arrive they are put in a pen and some animals start feasting on them until they are gone. Does that mean OpenTTD should be rated like that, since it is possible with readily downloadable graphics sets?
That would be why even Steam disables extensions in family locked accounts. I would say we should rate what is contained in the game itself, not their extensions. But some future decision might design some concept, how the game should/can protect underage players.
> 
> In any case, I would rather not self-rate the application as it will open a huge can of worms of people who see things differently. In which the developers are caught in the middle as they cannot please both sides, so if it doesn't need to be provided for anything that the OpenTTD developers provide I rather not add this point of contention.

This is not an option already. Flatpak already requires some definition for a reason. I am quite sure Steam does as well, even though there is already OpenTTD without any adult specifications. I would say every current app store has similar thing. Pretending we do not do won't work. Stating explicitly it is only for adults would, but should not be necessary.

TrueBrain:
As mentioned in earlier comments, and we talked this over between some devs a few times more, it seems we won't be able to provide a metainfo that is actually useful for consumers.

Having a full set of "releases" is pretty useful, but not something we can control from this repository. Every consumer would need to extend it with a full set of their releases (whether that is flatpak, fedora, appimage, ...).

And more importantly, we are unwilling to rate our game, and as such, this metainfo provided here needs to have no mention of a rating, which makes it impractical to use for consumers.

For those reasons alone, I think it is best that we just don't add a metainfo in our upstream, and leave it to downstreams to supply it. Although I would like some synchronization between those, if not alone to avoid small mistakes like the right license, right text, etc, it seems we will not find a balance here that works for all of us.

Despite that, I do appreciate you trying to upstream this; having the conversation is always good!

PR #11408, opened 2023-10-29
Fix: allow same length server password in the UI

rubidium42:
## Motivation / Problem

When starting a server through the UI, the maximum length of the password you can set in the dialogue is 19 characters.  However, the password via the settings or console may be 32 characters.

## Description

Replace `20` with `NETWORK_PASSWORD_LENGTH`.

## Limitations

It's still inconsistent when a message is shown to the user that the administrators might read the password. Now it is when setting a company password, or when being asked for the server's and company's password.
Adding the warning when setting the server's password used to be in this PR, but that seems to not be liked.

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -1037,7 +1037,7 @@ struct NetworkStartServerWindow : public Window {
 			case WID_NSS_SETPWD: // Set password button
 				this->widget_id = WID_NSS_SETPWD;
 				SetDParamStr(0, _settings_client.network.server_password);
-				ShowQueryString(STR_JUST_RAW_STRING, STR_NETWORK_START_SERVER_SET_PASSWORD, 20, this, CS_ALPHANUMERAL, QSF_NONE);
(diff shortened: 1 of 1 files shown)


PeterN:
If you're running a server you are the administrator, so I'm not sure that makes sense.

rubidium42:
> If you're running a server you are the administrator, so I'm not sure that makes sense.

Though there might be more than one administrator, and those can also read the password!
But now it's consistent with joining the server telling you that the administrators might be able to read the server's password. After all, one of the administrators set the password and they even have shared the password with you (if you're allowed to join).

nielsmh:
Agree that masking the server password creation is probably more counter-productive that it prevents any serious shoulder-surfing attacks. At most it might prevent some griefing at LAN parties, and I don't think that's a terribly common situation.

rubidium42:
> Agree that masking the server password creation is probably more counter-productive that it prevents any serious shoulder-surfing attacks. At most it might prevent some griefing at LAN parties, and I don't think that's a terribly common situation.

It has nothing to do with masking the password. The passwords are never masked. It's just that you see a warning that the password might be read by administrators when setting a company password, or when being asked for the company's or server's password. I made it consistent to also show it when setting the server's password, but that's changed "now".

2TallTyler (approved):
I have no thoughts on masking the password, but with your latest force-push this PR just fixes the max length, which is a no-brainer.

PR #11426, opened 2023-11-03
Fix #11402: Make string filter locale-aware.

### LINKED ISSUE #11402: [Bug]: Search fields are case sensitive in some languages
### Version of OpenTTD

OpenTTD 20231026

### Expected result

Search fields are not case sensitive, just like in English
изображение
изображение
изображение

### Actual result

At least in Cyrillic, the search fields are case sensitive.
изображение
изображение
изображение

### Steps to reproduce

1. Switch language to Russian/Belarusian/Ukrainian/Bulgarian
2. Try searching for something in «aviable trains» using lowercase text, for example паровоз / Паровоз

michicc:
## Motivation / Problem

String filters are currently not doing a locale-aware string match, which leads to issues as described in #11402 .

## Description

Use appropriate Win32/OSX/ICU functions to build a locale-aware `strstr` function, which then makes the string filter work as expected. Tested with the Russian example from the issue on all three platforms.

As locale-aware matching might not be suitable for every string filter usage, I've added an additional parameter to `StringFilter` to keep using the locale-unaware match. Currently, I've only enabled this for the script debug window break string filter.

## Limitations

There might be some other uses of `StringFilter` that should keep using the old matching.

I've only tested the Russian example from the issue report.

## Checklist for review

### DIFF
--- a/src/os/macosx/string_osx.cpp
+++ b/src/os/macosx/string_osx.cpp
@@ -336,6 +336,31 @@ int MacOSStringCompare(std::string_view s1, std::string_view s2)
 	return (int)CFStringCompareWithOptionsAndLocale(cf1.get(), cf2.get(), CFRangeMake(0, CFStringGetLength(cf1.get())), flags, _osx_locale.get()) + 2;
 }
 
+/**
+ * Search if a string is contained in another string using the current locale.
+ *
+ * @param str String to search in.
+ * @param value String to search for.
+ * @param case_insensitive Search case-insensitive.
+ * @return 1 if value was found, 0 if it was not found, or -1 if not supported by the OS.
+ */
+int MacOSStringContains(const std::string_view str, const std::string_view value, bool case_insensitive)
+{
+	static bool supported = MacOSVersionIsAtLeast(10, 5, 0);
+	if (!supported) return -1;
+
+	CFStringCompareFlags flags = kCFCompareLocalized | kCFCompareWidthInsensitive;
+	if (case_insensitive) flags |= kCFCompareCaseInsensitive;
+
+	CFAutoRelease<CFStringRef> cf_str(CFStringCreateWithBytes(kCFAllocatorDefault, (const UInt8 *)str.data(), str.size(), kCFStringEncodingUTF8, false));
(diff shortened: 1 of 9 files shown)


src/os/macosx/string_osx.cpp
```diff
@@ -336,6 +336,31 @@ int MacOSStringCompare(std::string_view s1, std::string_view s2)
...
 }
 
+/**
+ * Search if a string is contained in another string.
+ *
+ * @param str String to search in.
+ * @param value String to search for.
+ * @param case_insensitive Search case-insensitive.
+ * @return 1 if value was found, 0 if it was not found, or -1 if not supported by the OS.
+ */
```
TrueBrain, line 346: Why only document the OSX variant? :D

src/stringfilter.cpp
```diff
@@ -118,9 +118,16 @@ void StringFilter::AddLine(const char *str)
...
+			if (this->locale_aware) {
+				if (match_case ? StrNaturalContains(str, ws.start) : StrNaturalContainsIgnoreCase(str, ws.start)) {
+					ws.match = true;
+					this->word_matches++;
+				}
+			} else {
+				if ((match_case ? strstr(str, ws.start) : strcasestr(str, ws.start)) != nullptr) {
+					ws.match = true;
+					this->word_matches++;
+				}
```
LordAro, line 130: ternery inside an if is... a thing. I was thinking maybe a function pointer, but locale/notlocale has different signatures

src/string_func.h
```diff
@@ -39,6 +39,8 @@ void StrTrimInPlace(std::string &str);
 [[nodiscard]] int StrCompareIgnoreCase(const std::string_view str1, const std::string_view str2);
 [[nodiscard]] bool StrEqualsIgnoreCase(const std::string_view str1, const std::string_view str2);
 [[nodiscard]] int StrNaturalCompare(std::string_view s1, std::string_view s2, bool ignore_garbage_at_front = false);
+[[nodiscard]] bool StrNaturalContains(const std::string_view str, const std::string_view value);
+[[nodiscard]] bool StrNaturalContainsIgnoreCase(std::string_view str, const std::string_view value);
```
LordAro: why is this not const?

src/stringfilter.cpp
michicc: Well, the ternary was already there, I just adapted to it 😄

rubidium42 (approved):

PR #11428, opened 2023-11-03
Feature: Setting for minutes per calendar year

2TallTyler:
## Motivation / Problem

This is the last step of NotDaylength. It builds atop #10700 and #11341.

## Description

Add a setting to control how fast calendar time progresses, by setting "minutes per year". The default is 12, the minimum is 0 (special value to indicate Frozen calendar progression) and the maximum is 10,080, which is one real week. It can't go much slower without overflowing TimerGameCalendar::date_fract. This means that the fastest progression is one minute per year, 12 times faster than default.

Using this setting requires the player to use real-time units (#11341), which is only available when starting a new game. Once in that mode, the speed of calendar progression can be freely changed at any time.

## Limitations

* When in a game and not using real-time units, it's not possible to change this setting, but in the settings window the buttons to change minutes per year are not greyed out -- they just don't do anything when you click them.

## To do after this PR

* Change the multiplayer server list to show the amount of real time a game has been played, instead of how many years have passed.
* Add a file to docs/ to explain the new time system, with a section for players and a note about how it works on the back-end. Add a link to this in the new Help dialogue, and mention it in tooltips.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1487,6 +1487,13 @@ STR_CONFIG_SETTING_TIMEKEEPING_UNITS_HELPTEXT                   :Select the time
 STR_CONFIG_SETTING_TIMEKEEPING_UNITS_CALENDAR                   :Calendar
 STR_CONFIG_SETTING_TIMEKEEPING_UNITS_WALLCLOCK                  :Wallclock
 
+STR_CONFIG_SETTING_MINUTES_PER_YEAR                             :Minutes per year: {STRING2}
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_HELPTEXT                    :Choose the number of minutes in a calendar year. The default is 12 minutes. Set to 0 to stop calendar time from changing. This setting does not affect the economic simulation of the game, and is only available when using wallclock timekeeping.
+
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE                       :{NUM}
+###setting-zero-is-special
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_FROZEN                      :0 (calendar time frozen)
+
 STR_CONFIG_SETTING_AUTORENEW_VEHICLE                            :Autorenew vehicle when it gets old: {STRING2}
 STR_CONFIG_SETTING_AUTORENEW_VEHICLE_HELPTEXT                   :When enabled, a vehicle nearing its end of life gets automatically replaced when the renew conditions are fulfilled
 
--- a/src/settings_table.cpp
+++ b/src/settings_table.cpp
@@ -22,6 +22,7 @@
 #include "news_func.h"
 #include "window_func.h"
 #include "company_func.h"
+#include "timer/timer_game_calendar.h"
 #if defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA)
 #define HAS_TRUETYPE_FONT
 #include "fontcache.h"
@@ -500,7 +501,46 @@ static void ChangeTimekeepingUnits(int32_t)
 		UpdateAllServiceInterval(0);
 	}
 
+	/* If we are using calendar timekeeping, "minutes per year" must be default. */
+	if (!TimerGameEconomy::UsingWallclockUnits(true)) {
+		_settings_newgame.economy.minutes_per_calendar_year = CalendarTime::DEF_MINUTES_PER_YEAR;
+	}
+
 	InvalidateWindowClassesData(WC_GAME_OPTIONS, 0);
+}
+
+/**
+ * Callback after the player changes the minutes per year.
+ * @param new_value The intended new value of the setting, used for clamping.
+ */
+static void ChangeMinutesPerYear(int32_t new_value)
+{
+	/* We don't allow setting Minutes Per Year below default, unless it's to 0 for frozen calendar time. */
+	if (new_value < CalendarTime::DEF_MINUTES_PER_YEAR) {
+		int clamped;
+
+		/* If the new value is 1, we're probably at 0 and trying to increase the value, so we should jump up to default. */
+		if (new_value == 1) {
+			clamped = CalendarTime::DEF_MINUTES_PER_YEAR;
+		} else {
+			clamped = CalendarTime::FROZEN_MINUTES_PER_YEAR;
+		}
+
+		/* Override the setting with the clamped value. */
+		if (_game_mode == GM_MENU) {
+			_settings_newgame.economy.minutes_per_calendar_year = clamped;
+		} else {
+			_settings_game.economy.minutes_per_calendar_year = clamped;
+		}
+	}
+
+	/* If the setting value is not the default, force the game to use wallclock timekeeping units.
+	 * This can only happen in the menu, since the pre_cb ensures this setting can only be changed there, or if we're already using wallclock units.
+	 */
+	if (_game_mode == GM_MENU && (_settings_newgame.economy.minutes_per_calendar_year != CalendarTime::DEF_MINUTES_PER_YEAR)) {
+		_settings_newgame.economy.timekeeping_units = TKU_WALLCLOCK;
+		InvalidateWindowClassesData(WC_GAME_OPTIONS, 0);
+	}
 }
 
 /**
--- a/src/settings_type.h
+++ b/src/settings_type.h
@@ -558,6 +558,7 @@ struct EconomySettings {
 	bool   allow_town_level_crossings;       ///< towns are allowed to build level crossings
 	bool   infrastructure_maintenance;       ///< enable monthly maintenance fee for owner infrastructure
 	TimekeepingUnits timekeeping_units;      ///< time units to use for the game economy, either calendar or wallclock
+	uint16_t minutes_per_calendar_year;      ///< minutes per calendar year. Special value 0 means that calendar time is frozen.
 };
 
 struct LinkGraphSettings {
--- a/src/table/settings/economy_settings.ini
+++ b/src/table/settings/economy_settings.ini
@@ -10,6 +10,7 @@
 [pre-amble]
 static void TownFoundingChanged(int32_t new_value);
 static void ChangeTimekeepingUnits(int32_t new_value);
+static void ChangeMinutesPerYear(int32_t new_value);
 
 static const SettingVariant _economy_settings_table[] = {
 [post-amble]
@@ -293,4 +294,19 @@ str      = STR_CONFIG_SETTING_TIMEKEEPING_UNITS
 strval   = STR_CONFIG_SETTING_TIMEKEEPING_UNITS_CALENDAR
 strhelp  = STR_CONFIG_SETTING_TIMEKEEPING_UNITS_HELPTEXT
 post_cb  = ChangeTimekeepingUnits
+cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.minutes_per_calendar_year
+type     = SLE_UINT16
+flags    = SF_GUI_0_IS_SPECIAL | SF_NO_NETWORK
+def      = CalendarTime::DEF_MINUTES_PER_YEAR
+min      = CalendarTime::FROZEN_MINUTES_PER_YEAR
+max      = CalendarTime::MAX_MINUTES_PER_YEAR
+interval = 1
+str      = STR_CONFIG_SETTING_MINUTES_PER_YEAR
+strhelp  = STR_CONFIG_SETTING_MINUTES_PER_YEAR_HELPTEXT
+strval   = STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE
+pre_cb   = [](auto) { return _game_mode == GM_MENU || _settings_game.economy.timekeeping_units == 1; }
+post_cb  = ChangeMinutesPerYear
 cat      = SC_BASIC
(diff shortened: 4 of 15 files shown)


src/table/settings/economy_settings.ini
```diff
@@ -9,6 +9,8 @@
 
 [pre-amble]
 static void TownFoundingChanged(int32_t new_value);
+static void UseRealtimeUnits(int32_t new_value);
+void ChangeCalendarProgressSpeed(int32_t new_value);
```
glx22: My guess is the annotations are related to this line.

src/settings_table.h
```diff
@@ -41,4 +41,6 @@ static const uint GAME_DIFFICULTY_NUM = 18;
 extern const std::array<std::string, GAME_DIFFICULTY_NUM> _old_diff_settings;
 extern uint16_t _old_diff_custom[GAME_DIFFICULTY_NUM];
 
+void ChangeCalendarProgressSpeed(int32_t);
```
glx22: Or this one.

src/genworld_gui.cpp
```diff
@@ -375,6 +389,38 @@ static DropDownList BuildTownNameDropDown()
 	return list;
 }
 
+/**
+ * Set string parameters before drawing the number of minutes per year.
+ * @param max Set the maximum value, to determine widget size.
+ */
+static void SetMinutesPerYearParams(bool max = false)
+{
+	/* When resizing widgets, the longest string is 1% of default. */
```
TrueBrain: I do not understand this comment. 1% of default, but you multiply the default with 100? Did you mean: the default is 1% of the longest string? I dunno, sorry :)

src/genworld_gui.cpp
```diff
@@ -375,6 +389,38 @@ static DropDownList BuildTownNameDropDown()
...
+		if (percent > 0) {
+			/* Only draw a percent if it's above 0%. */
+			SetDParam(0, STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_PERCENT);
+			SetDParam(1, minutes);
+			SetDParam(2, percent);
+		} else {
+			SetDParam(0, STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE);
+			SetDParam(1, minutes);
+		}
+	}
```
TrueBrain: This if-cascade is rather hard to read, and hard to figure out what it actually does. As in: I get what the code does, but why it is this complicated I do not really understand? Some suggestions:

Start with the exception and early return:

`if (minutes == frozen) { SetDParam(..); return; }`

That way, we got that one out of the way.

Next, there is something complicated about sometimes returning a percentage, sometimes a value. Also there is `SETTING` vs `SETTINGS`, where the latter has a `{BLACK}`, but the first doesn't. But both are used here.

Also, I am not sure why it is important to add the `(default)` here? I don't think we do it anywhere else?

I would reason: the default is 100%, right? So you can just always show that: `12 minutes (100%)`. I think that is enough reinforcement that that is the default?

Lastly, although I get that when percentage goes < 0.5%, you no longer want to draw %, but that feels weird too. Maybe simpler rules:

`if percentage >= 1`: `N minutes (M%)`
`if percentage == 0`: `N minutes (<1%)`

That would simplify this function a lot, and makes things a bit easier for the end-user too?

src/lang/english.txt
```diff
@@ -1462,6 +1464,19 @@ STR_CONFIG_SETTING_NEVER_EXPIRE_VEHICLES_HELPTEXT               :When enabled, a
...
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_HELPTEXT                    :Choose the number of minutes in a calendar year. The default is 12 minutes. Set to 0 to freeze calendar time completely. This setting does not affect the economic simulation of the game, and is only available when using real-time units.
+
+# Value strings used in Settings GUI
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE                       :{NUM} minutes
+###setting-zero-is-special
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_FROZEN                      :Frozen
+
+# Value strings used in Generate World GUI
+STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_DEFAULT              :{BLACK}{NUM} minute{P 0 "" s} (default)
+STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_PERCENT              :{BLACK}{NUM} minute{P 0 "" s} ({NUM}% speed)
```
TrueBrain: `(90% speed)` feels off. `(90% gamespeed)`? `(90%)`? Dunno .. "speed" feels weird. But I lack good suggestions :)

TrueBrain:
Reviewed most of your work, it looks good to me. And I think this has been open long enough for anyone who strongly disagrees to speak up; guess there aren't any :P (we will find out after we merge ;) ).

I suggest the next step is, before we merge any of these 3 PRs, is to have a test-game, much like with ship pathfinder, to find out if we missed anything. For that I suggest we freeze the calendar, as that is the most extreme case, or set it very slow. But that allows us to test if we moved everything to the economy timer what should be there.

If that test-game is a success, and after you cleaned up the last few, mostly minor, comments, I say we merge the stuff.

Would you be up to setup such a game?

TrueBrain:
Not sure if the screenshot is still up to date, but that "real time" button stands out.

Maybe make it a dropdown, where you can select either of the two time units. Makes it also more verbose what the options are, instead of having to press a button and find out.

2TallTyler:
speed

Rebased and updated from review comments.

src/lang/english.txt
```diff
@@ -1465,6 +1467,19 @@ STR_CONFIG_SETTING_TIMEKEEPING_UNITS_HELPTEXT                   :Select the time
...
+# Value strings used in Settings GUI
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE                       :{NUM} minutes
+###setting-zero-is-special
+STR_CONFIG_SETTING_MINUTES_PER_YEAR_FROZEN                      :Frozen
+
+# Value strings used in Generate World GUI
+STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_FROZEN               :{BLACK}Frozen
+STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_PERCENT              :{BLACK}{NUM} minute{P 0 "" s} ({NUM}% game speed)
+STR_CONFIG_SETTINGS_MINUTES_PER_YEAR_VALUE_ONE_PERCENT          :{BLACK}{NUM} minute{P 0 "" s} (<1% game speed)
+
```
TrueBrain, line 1496: I think you forgot to remove some strings? :)

TrueBrain: Nevermind, GitHub interface playing tricks on me.

Guess the next question is: weren't you going to drop the %? :)

2TallTyler: Yes, I haven't gotten to it yet. 😉

2TallTyler:
speed

Per a discussion on Discord yesterday, I removed the (nnn% speed) on the value string, so it only shows minutes. The intended use of the percentage was to help the player find the default value, but there are other (better) ways to find that:
1. Read the tooltip
2. Change Timekeeping units to Calendar (which resets Minutes per year to default)
3. Click the value to open the custom text input, and click the Default button

src/timer/timer_game_realtime.cpp
```diff
@@ -53,12 +53,18 @@ void TimeoutTimer<TimerGameRealtime>::Elapsed(TimerGameRealtime::TElapsed delta)
 	}
 }
 
+/**
+ * Every-tick handler for real-time timers.
+ * @return Always false, since this value is required by the template but not used for anything by this timer.
```
TrueBrain: This reasoning is very hard to parse. Why not keep it simple, and just say, as general rule for all `Elapsed`: `true iff the time progressed`, and that means this function would also just return true. Easier to explain, less vague statements that only makes sense when you know what you were struggling with that the time of writing :)

src/timer/timer_manager.h
```diff
@@ -77,9 +77,12 @@ class TimerManager {
 	 * The implementation per type is different, but they all share a similar goal:
 	 *   Call the Elapsed() method of all active timers.
 	 *
+	 * There is an optional return value that timers may use to signal if they have "ticked".
+	 *
 	 * @param value The amount of time that has elapsed.
+	 * @return A boolean value to signal that something has "ticked," not used by all timers.
```
TrueBrain: Here too, I would make it simpler: `True iff time has progressed`.

Btw, `a boolean value` is totally wasting bytes. That it is a boolean, we can see, and that it is a value, is obvious :)

src/timer/timer_game_calendar.cpp
```diff
@@ -93,28 +93,33 @@ void TimeoutTimer<TimerGameCalendar>::Elapsed(TimerGameCalendar::TElapsed trigge
 	}
 }
 
+/**
+ * Every-tick handler for possibly advancing calendar time.
+ * @note The return value is used to inform other functions whether we advanced TimerGameCalendar::date_fract.
+ * @return True if we advanced TimerGameCalendar::date_fract, otherwise false if we didn't change anything or only advanced TimerGameCalendar::sub_date_fract.
```
TrueBrain: Here too (in combination with the other comments of mine). You now appear to assign a different return "meaning" for each class type. That is not ideal.

Better just make the return value mean something in one place, and not something different for each implementation of the template. Basically, there shouldn't be a need to have a comment per template implementation, and the comment at the definition should explain what is going on.

src/timer/timer_manager.h
```diff
@@ -77,9 +77,12 @@ class TimerManager {
 	 * The implementation per type is different, but they all share a similar goal:
 	 *   Call the Elapsed() method of all active timers.
 	 *
+	 * There is an optional return value that timers may use to signal if they have "ticked".
```
TrueBrain: This makes no sense. Return values aren't optional.

src/timer/timer_game_tick.cpp
```diff
@@ -50,14 +50,20 @@ void TimeoutTimer<TimerGameTick>::Elapsed(TimerGameTick::TElapsed delta)
 	}
 }
 
+/**
+ * Every-tick handler for game tick timers.
+ * @return True iff time has progressed.
+ */
```
TrueBrain: No need for this comment (and the other cpp files); you already have that at the template definition in the header-file :)

src/settings_table.cpp
```diff
@@ -500,9 +501,40 @@ static void ChangeTimekeepingUnits(int32_t)
...
+	/* We don't allow setting Minutes Per Year below default, unless it's to 0 for frozen calendar time. */
+	if (_game_mode == GM_MENU) {
+		if (_settings_newgame.economy.minutes_per_calendar_year < CalendarTime::DEF_MINUTES_PER_YEAR) {
+			_settings_newgame.economy.minutes_per_calendar_year = CalendarTime::FROZEN_MINUTES_PER_YEAR;
+		}
+	} else {
+		if (_settings_game.economy.minutes_per_calendar_year < CalendarTime::DEF_MINUTES_PER_YEAR) {
+			_settings_game.economy.minutes_per_calendar_year = CalendarTime::FROZEN_MINUTES_PER_YEAR;
+		}
+	}
```
TrueBrain, line 535: The problem here is, I think, that once it is 0, you can't get it off 0 again? Can't you use the `new_value` (which is the parameter) to figure out which direction we are going?

Honestly, I need to test this, I am just guessing behaviour by looking at code. Terrible way to judge functionality.

src/settings_type.h
```diff
@@ -558,6 +558,7 @@ struct EconomySettings {
 	bool   allow_town_level_crossings;       ///< towns are allowed to build level crossings
 	bool   infrastructure_maintenance;       ///< enable monthly maintenance fee for owner infrastructure
 	TimekeepingUnits timekeeping_units;      ///< time units to use for the game economy, either calendar or wallclock
+	uint16_t minutes_per_calendar_year;      ///< minutes per calendar year. Special value 0 is frozen.
```
TrueBrain: `is frozen` -> `means frozen`?

src/table/settings/economy_settings.ini
```diff
@@ -294,3 +295,18 @@ strval   = STR_CONFIG_SETTING_TIMEKEEPING_UNITS_CALENDAR
 strhelp  = STR_CONFIG_SETTING_TIMEKEEPING_UNITS_HELPTEXT
 post_cb  = ChangeTimekeepingUnits
 cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.minutes_per_calendar_year
+type     = SLE_UINT16
+flags    = SF_GUI_0_IS_SPECIAL | SF_NO_NETWORK
+def      = 12
```
TrueBrain: Why not use the enum values here? (for all 3)

src/settings_table.cpp
2TallTyler: Your guess was correct, as was your suggestion. 🙂

TrueBrain (approved):

Meatplay:
Is there a technical reason why the setting is clamped to 12 and one can not speed up time?

Not sure if this is the correct place to ask, but I was hoping this feature would allow implementing a game script which introduces a tech tree. So speeding up time (or skipping time directly via script not cheat) on some condition would be helpful. Since the calender is separated from economy now this should not have major sideeffects from my understanding.

PR #11431, opened 2023-11-04
Fix cda6f24f: don't ignore binary-dir if it happens to be working-dir

TrueBrain:
## Motivation / Problem

Emscripten (previews) broke with cda6f24f.

## Description

Some of our code ignores the `SP_WORKING_DIR` for some actions, which means that if, for example, your `SP_BINARY_DIR` is the same as your `SP_WORKING_DIR`, neither is scanned.

Instead, only add `SP_WORKING_DIR` if it is unique.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -87,6 +87,8 @@ static void FillValidSearchPaths(bool only_local_path)
 
 	std::set<std::string> seen{};
 	for (Searchpath sp = SP_FIRST_DIR; sp < NUM_SEARCHPATHS; sp++) {
+		if (sp == SP_WORKING_DIR) continue;
+
 		if (only_local_path) {
 			switch (sp) {
 				case SP_WORKING_DIR:      // Can be influence by "-c" option.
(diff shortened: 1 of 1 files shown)


PeterN (approved):

src/fileio.cpp
```diff
@@ -87,6 +87,8 @@ static void FillValidSearchPaths(bool only_local_path)
 
 	std::set<std::string> seen{};
 	for (Searchpath sp = SP_FIRST_DIR; sp < NUM_SEARCHPATHS; sp++) {
+		if (sp == SP_WORKING_DIR) continue;
+
 		if (only_local_path) {
 			switch (sp) {
 				case SP_WORKING_DIR:      // Can be influence by "-c" option.
```
glx22, line 94: This line is now useless

glx22 (commented):
Yes, I know it's already merged.

src/fileio.cpp
TrueBrain: No, it is not. It is the reason this PR works.

Why do you think it is useless?

TrueBrain: ow, wow, GitHub isn't all that clear it wasn't about the changed text. Right, that explains a bit more.

Yes, that line is now not doing anything. But I am a bit in doubt if we should remove it, as .. it does give information, I am just not sure what information :P

PR #11435, opened 2023-11-04
Feature: Mode to display timetable in seconds

2TallTyler:
## Motivation / Problem

In OpenTTD, timetables use calendar dates for arrival and departures, as well as the timetable start date. This makes no sense when the calendar date never changes (and we never show players the economy date). Instead, in Real-Time Mode (#11341) we display these as an offset in seconds from the current time.

This feature can actually be implemented separately and used outside Real-Time Mode, to be useful to more people and to make the PR smaller and easier to review.

## Description

Replace the boolean "timetable in ticks" setting with a drop-down to select days, seconds, or ticks.

With NotDaylength, the calendar date never changes (and we never show players the economy date) so we will force this new setting to show seconds.

For what it's worth, I am a regular user of timetables in both in vanilla OpenTTD and JGRPP using Scheduled Dispatch and the Wall Clock feature. So as I'm working on this feature, I'm thinking about my ideal use case, not just shrugging off timetables entirely as broken and not worthy of attention.

## Limitations

* When timetabling in Ticks, arrival/departure times and start date use dates. This is unchanged from current behavior but there's no way to set travel time in Ticks and show arrival/departure times in seconds. I don't expect this to be a problem for many people.

## Checklist for review

### DIFF
--- a/src/base_consist.h
+++ b/src/base_consist.h
@@ -18,8 +18,8 @@ struct BaseConsist {
 	std::string name;                   ///< Name of vehicle
 
 	/* Used for timetabling. */
-	uint32_t current_order_time;               ///< How many ticks have passed since this order started.
-	int32_t lateness_counter;                  ///< How many ticks late (or early if negative) this vehicle is.
+	TimerGameTick::Ticks current_order_time;    ///< How many ticks have passed since this order started.
+	TimerGameTick::Ticks lateness_counter;      ///< How many ticks late (or early if negative) this vehicle is.
 	TimerGameTick::TickCounter timetable_start; ///< At what tick of TimerGameTick::counter the vehicle should start its timetable.
 
 	uint16_t service_interval;            ///< The interval for (automatic) servicing; either in days or %.
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -255,6 +255,10 @@ STR_UNITS_HEIGHT_IMPERIAL                                       :{DECIMAL}{NBSP}
 STR_UNITS_HEIGHT_METRIC                                         :{DECIMAL}{NBSP}m
 STR_UNITS_HEIGHT_SI                                             :{DECIMAL}{NBSP}m
 
+STR_UNITS_DAYS                                                  :{COMMA}{NBSP}day{P "" s}
+STR_UNITS_SECONDS                                               :{COMMA}{NBSP}second{P "" s}
+STR_UNITS_TICKS                                                 :{COMMA}{NBSP}tick{P "" s}
+
 # Common window strings
 STR_LIST_FILTER_TITLE                                           :{BLACK}Filter:
 STR_LIST_FILTER_OSKTITLE                                        :{BLACK}Enter one or more keywords to filter the list for
@@ -1656,8 +1660,12 @@ STR_CONFIG_SETTING_ADVANCED_VEHICLE_LISTS_HELPTEXT              :Enable usage of
 STR_CONFIG_SETTING_LOADING_INDICATORS                           :Use loading indicators: {STRING2}
 STR_CONFIG_SETTING_LOADING_INDICATORS_HELPTEXT                  :Select whether loading indicators are displayed above loading or unloading vehicles
 
-STR_CONFIG_SETTING_TIMETABLE_IN_TICKS                           :Show timetable in ticks rather than days: {STRING2}
-STR_CONFIG_SETTING_TIMETABLE_IN_TICKS_HELPTEXT                  :Show travel times in time tables in game ticks instead of days
+STR_CONFIG_SETTING_TIMETABLE_MODE                               :Time units for timetables: {STRING2}
+STR_CONFIG_SETTING_TIMETABLE_MODE_HELPTEXT                      :Select the time units used for vehicle timetables
+###length 3
+STR_CONFIG_SETTING_TIMETABLE_MODE_DAYS                          :Days
+STR_CONFIG_SETTING_TIMETABLE_MODE_SECONDS                       :Seconds
+STR_CONFIG_SETTING_TIMETABLE_MODE_TICKS                         :Ticks
 
 STR_CONFIG_SETTING_TIMETABLE_SHOW_ARRIVAL_DEPARTURE             :Show arrival and departure in timetables: {STRING2}
 STR_CONFIG_SETTING_TIMETABLE_SHOW_ARRIVAL_DEPARTURE_HELPTEXT    :Display anticipated arrival and departure times in timetables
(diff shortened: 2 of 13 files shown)


src/timetable.h
```diff
@@ -18,6 +18,11 @@ static const TimerGameCalendar::Year MAX_TIMETABLE_START_YEARS = 15; ///< The ma
 
 TimerGameTick::TickCounter GetStartTickFromDate(TimerGameCalendar::Date start_date);
 TimerGameCalendar::Date GetDateFromStartTick(TimerGameTick::TickCounter start_tick);
+enum class TimetableMode : uint8_t {
```
PeterN, line 19: I think this should go before the function definitions, and be separated by a blank line.

src/timetable_gui.cpp
```diff
@@ -47,15 +49,35 @@ struct TimetableArrivalDeparture {
...
+/**
+ * Get the correct number of ticks depending on the timetable mode.
+ * @return The number of ticks per day or second, depending on which time units are displayed.
+ */
+static inline TimerGameTick::Ticks TicksPerDayOrSecond()
+{
+	if (_settings_client.gui.timetable_mode == TimetableMode::Seconds) return Ticks::TICKS_PER_SECOND;
+
+	return Ticks::DAY_TICKS;
+}
```
PeterN: This function seems to be used for two purposes:

1) to convert from ticks to days or seconds (but not for display, as that's handled above).
2) to set the threshhold for lateness.

I think the second case should be handled in a separate function (something like `TimetableLatenessThreshold()`), and this function should be able to used to convert a value to/from days, seconds, or ticks (return 1)

If you do that then there are some special cases further on that no longer need to be special cases.

src/timetable_gui.cpp
```diff
@@ -490,19 +533,28 @@ struct TimetableWindow : Window {
...
+				SetDParam(1, GetDateFromStartTick(v->timetable_start));
+				DrawString(tr, STR_TIMETABLE_STATUS_START_AT_DATE);
+			}
 		} else if (!HasBit(v->vehicle_flags, VF_TIMETABLE_STARTED)) {
-			/* We aren't running on a timetable yet, so how can we be "on time"
-			 * when we aren't even "on service"/"on duty"? */
+			/* We aren't running on a timetable yet. */
 			DrawString(tr, STR_TIMETABLE_STATUS_NOT_STARTED);
-		} else if (v->lateness_counter == 0 || (!_settings_client.gui.timetable_in_ticks && v->lateness_counter / Ticks::DAY_TICKS == 0)) {
+		} else if (v->lateness_counter == 0 || (_settings_client.gui.timetable_mode != TimetableMode::Ticks && v->lateness_counter / TicksPerDayOrSecond() == 0)) {
```
PeterN: Special-case that could be removed if `TicksPerDayOrSecond()` also handled `TimetableMode::Ticks`.

src/timetable_gui.cpp
```diff
@@ -571,7 +630,9 @@ struct TimetableWindow : Window {
 
 				if (order != nullptr) {
 					uint time = (selected % 2 != 0) ? order->GetTravelTime() : order->GetWaitTime();
-					if (!_settings_client.gui.timetable_in_ticks) time /= Ticks::DAY_TICKS;
+					if (_settings_client.gui.timetable_mode != TimetableMode::Ticks) {
+						time /= TicksPerDayOrSecond();
+					}
```
PeterN: Special-case that could be removed if `TicksPerDayOrSecond()` also handled `TimetableMode::Ticks`.

src/timetable_gui.cpp
```diff
@@ -667,7 +728,7 @@ struct TimetableWindow : Window {
 			}
 
 			case WID_VT_CHANGE_TIME:
-				if (!_settings_client.gui.timetable_in_ticks) val *= Ticks::DAY_TICKS;
+				if (_settings_client.gui.timetable_mode != TimetableMode::Ticks) val *= TicksPerDayOrSecond();
```
PeterN: Special-case that could be removed if `TicksPerDayOrSecond()` also handled `TimetableMode::Ticks`.

PeterN:
Also "Use seconds for timetables" implies days/ticks are removed (maybe they are later...) but for now this is "Add seconds for timetable display"

src/timetable_gui.cpp
```diff
@@ -47,12 +49,58 @@ struct TimetableArrivalDeparture {
...
+ */
+bool VehicleIsAboveLatenessThreshold(const Vehicle *v)
+{
+	switch (_settings_client.gui.timetable_mode) {
+		case TimetableMode::Days:
+			return v->lateness_counter > Ticks::DAY_TICKS;
+		case TimetableMode::Seconds:
+			return v->lateness_counter > Ticks::TICKS_PER_SECOND;
+		case TimetableMode::Ticks:
+			return v->lateness_counter == 0;
```
PeterN: Is this correct? Previously it used the ` > Ticks::DAY_TICKS` condition, now it shows as late if it's ... not late?

PeterN (dismissed):

src/base_consist.h
```diff
@@ -18,8 +18,8 @@ struct BaseConsist {
 	std::string name;                   ///< Name of vehicle
 
 	/* Used for timetabling. */
-	uint32_t current_order_time;               ///< How many ticks have passed since this order started.
-	int32_t lateness_counter;                  ///< How many ticks late (or early if negative) this vehicle is.
+	TimerGameTick::Ticks current_order_time;    ///< How many ticks have passed since this order started.
```
PeterN, line 21: This is saved as uint32, but you've changed it to an int32 instead. The data type is the same size so it'll appear to be fine, but some changes for saveload will be necessary.

src/timetable_gui.cpp
```diff
@@ -556,8 +638,15 @@ struct TimetableWindow : Window {
...
-				break;
+				if (_settings_client.gui.timetable_mode == TimetableMode::Seconds) {
+					this->query_widget = WID_VT_START_DATE;
+					this->change_timetable_all = _ctrl_pressed;
+					ShowQueryString(STR_EMPTY, STR_TIMETABLE_START_SECONDS_QUERY, 6, this, CS_NUMERAL, QSF_ACCEPT_UNCHANGED);
+					break;
+				} else {
+					ShowSetDateWindow(this, v->index, TimerGameCalendar::date, TimerGameCalendar::year, TimerGameCalendar::year + MAX_TIMETABLE_START_YEARS, ChangeTimetableStartCallback, reinterpret_cast<void*>(static_cast<uintptr_t>(_ctrl_pressed)));
+					break;
+				}
```
rubidium42: The else is not needed, as the if branch already leaves the case.
Or you have to move the `break;` outside of the if-else statement.

src/lang/english.txt
```diff
@@ -4576,10 +4582,13 @@ STR_TIMETABLE_STATUS_ON_TIME                                    :{BLACK}This veh
 STR_TIMETABLE_STATUS_LATE                                       :{BLACK}This vehicle is currently running {STRING1} late
 STR_TIMETABLE_STATUS_EARLY                                      :{BLACK}This vehicle is currently running {STRING1} early
 STR_TIMETABLE_STATUS_NOT_STARTED                                :{BLACK}This timetable has not yet started
-STR_TIMETABLE_STATUS_START_AT                                   :{BLACK}This timetable will start at {STRING1}
+STR_TIMETABLE_STATUS_START_AT_DATE                              :{BLACK}This timetable will start at {STRING1}
+STR_TIMETABLE_STATUS_START_IN_SECONDS                           :{BLACK}This timetable will start in {COMMA} seconds
+
+STR_TIMETABLE_START                                             :{BLACK}Start Timetable
+STR_TIMETABLE_START_TOOLTIP                                     :{BLACK}Select a starting point of this timetable. Ctrl+Click distributes all vehicles sharing this order evenly from the given date based on their relative order, if the order is completely timetabled
```
rubidium42: Very not sure as I'm not native English, but maybe 'starting moment' instead of 'starting point'?
Also, there's another 'date' lingering in this string, should that become something like 'moment' as well?

rubidium42 (approved):

PR #11439, opened 2023-11-04
Change: Use sensible default settings for Cargo Distribution

2TallTyler:
## Motivation / Problem

On the OpenTTD Discord, new players ask how to give passengers destinations on a near-daily basis. We have a pinned message showing them how to change cargo distribution settings to do this. This currently requires showing advanced settings.

I have not played a modern transport game where passengers do not have destinations. Even Railroad Tycoon 3 (2003) had this feature.

Some longtime OpenTTD and TTD players hate cargo distribution. These are the players most able to tweak settings to their liking, and they already have OpenTTD installs with their configuration files set accordingly, so this would not affect them.

## Description

Change the default settings, for first-time installations of OpenTTD, to enable cargo distribution for passengers, mail, and valuables. Specifically:
- Passengers: Symmetric
- Mail: Asymmetric
- Armoured: Asymmetric
- Others: Manual

Also, change the setting visibility to Basic so these settings can be found more easily.

## Limitations

It's getting cold where I live, so a flamewar would keep me warm. 😛 

## Checklist for review

### DIFF
--- a/src/table/settings/linkgraph_settings.ini
+++ b/src/table/settings/linkgraph_settings.ini
@@ -64,7 +64,7 @@ var      = linkgraph.distribution_pax
 type     = SLE_UINT8
 from     = SLV_183
 flags    = SF_GUI_DROPDOWN
-def      = DT_MANUAL
+def      = DT_SYMMETRIC
 min      = DT_MIN
 max      = DT_MAX
 interval = 1
@@ -72,13 +72,14 @@ str      = STR_CONFIG_SETTING_DISTRIBUTION_PAX
 strval   = STR_CONFIG_SETTING_DISTRIBUTION_MANUAL
 strhelp  = STR_CONFIG_SETTING_DISTRIBUTION_PAX_HELPTEXT
 extra    = offsetof(LinkGraphSettings, distribution_pax)
+cat      = SC_BASIC
 
 [SDT_VAR]
 var      = linkgraph.distribution_mail
 type     = SLE_UINT8
 from     = SLV_183
 flags    = SF_GUI_DROPDOWN
-def      = DT_MANUAL
+def      = DT_ASYMMETRIC
 min      = DT_MIN
 max      = DT_MAX
 interval = 1
@@ -86,13 +87,14 @@ str      = STR_CONFIG_SETTING_DISTRIBUTION_MAIL
 strval   = STR_CONFIG_SETTING_DISTRIBUTION_MANUAL
 strhelp  = STR_CONFIG_SETTING_DISTRIBUTION_MAIL_HELPTEXT
 extra    = offsetof(LinkGraphSettings, distribution_mail)
+cat      = SC_BASIC
 
 [SDT_VAR]
 var      = linkgraph.distribution_armoured
 type     = SLE_UINT8
 from     = SLV_183
 flags    = SF_GUI_DROPDOWN
-def      = DT_MANUAL
+def      = DT_ASYMMETRIC
 min      = DT_MIN
 max      = DT_MAX
 interval = 1
@@ -100,6 +102,7 @@ str      = STR_CONFIG_SETTING_DISTRIBUTION_ARMOURED
 strval   = STR_CONFIG_SETTING_DISTRIBUTION_MANUAL
 strhelp  = STR_CONFIG_SETTING_DISTRIBUTION_ARMOURED_HELPTEXT
 extra    = offsetof(LinkGraphSettings, distribution_armoured)
+cat      = SC_BASIC
 
 [SDT_VAR]
 var      = linkgraph.distribution_default
@@ -114,6 +117,7 @@ str      = STR_CONFIG_SETTING_DISTRIBUTION_DEFAULT
 strval   = STR_CONFIG_SETTING_DISTRIBUTION_MANUAL
 strhelp  = STR_CONFIG_SETTING_DISTRIBUTION_DEFAULT_HELPTEXT
 extra    = offsetof(LinkGraphSettings, distribution_default)
+cat      = SC_BASIC
 
 [SDT_VAR]
 var      = linkgraph.accuracy


nielsmh:
Arguments against:
- Transfer behavior looks really weird if you look closely, since cargo does not actually have destinations while loaded on vehicles, only while waiting at stations, and receive new destinations every time they are transferred to a station.
- Requires more network capacity than not playing with automatic distribution, meaning it might make the game more difficult: Vehicle may make less per cargo unit transported per distance.
- The "transfer" "income" indicators don't indicate money actually going into the company balance, prompting players to be confused to why they see something that looks like income but isn't actually.
- Transfer legs with effective negative distance, or very slow compared to previous legs, might end up making negative attributed value to vehicles, prompting players to be confused to why their vehicles (e.g. buses at the other end of an airline) accumulate more negative income than their running cost, and despite transporting lots of cargo.

PikkaBird:
I think "hate cargo distribution" is a bit strong (and a bit well-poisony - hatred implies irrationality or a loss of perspective), but I'm certainly a long-time player who used, and encouraged others to use, cargo distribution for many years, but now prefer to leave it turned off.

It's often struck me that cargodist is 80% imagination - people like the *idea* of little sims beetling around their virtual world, and cargodist plays into that fantasy. But the reality is that the feature is conceptually broken (because it's trying to simultaneously shape *and* react to what the player builds), and players who try to satisfy cargodist demands on any but the simplest networks quickly find themselves chasing their own tail. In the end, if you want to *imagine* little sims making connections and riding trains more than one stop, you can do that perfectly well without cargodist at all.

We do often hear from players wanting to give passengers destinations, and to that end I would not oppose moving the settings up from Advanced to Basic. But equally often we hear from players who have been advised to turn cargodist on, not fully understanding what it does, and are now confused that their vehicles are failing to load the hundreds of passengers at their stations. Turning cargo distribution on by default would not be an unobtrusive "sensible" change, and it would certainly not make the game more accessible or understandable for new players.

I don't disagree with making cargodist settings more visible or accessible, but, given the feature's shortcomings, I still feel it's important that players not be subjected to it unless they make a deliberate decision to turn it on.

Eddi-z:
Potential reasons for symmetric/asymmetric:

1. most people return to their home after going to work/shopping/holiday (= symmetric)
2. most homes receive more mail than they send out (= asymmetric)

on the whole, i agree with this change. but it's only a first step.

therealbungus:
I feel this might be something better suited for an option in the new game menu, alongside terrain generation method, town/industry levels, max map height, etc, instead of a default setting. PikkaBird has a very good point about it not making the game more accessible or understandable to new players. When you look up guides, videos, or anything about how to play this game as a new player, pretty much nothing at all mentions cargodist mechanics. The mechanic isn't something that is immediately apparent or explained, and players who expect typical Transport Tycoon Deluxe behavior would be thrown off. 

Cargodist is essential for all of my games that I play because it adds more depth, but I would definitely not want to force it on a new player, and would recommend it only after they wanted a little more depth in the game. As it stands now, the option is very hidden and basically requires direct instruction to enable it for someone who doesn't know how. Moving this option to the new game menu would be a good compromise between not setting it as default, but making it more accessible to a vast majority of people.

DefinitelyNotRau117:
I think this setting should be “in a visible place", probably even so that you can quickly switch cargodist for pax/mail right in the game, without having to go through those… massive and scary settings window. But for beginners, it’s worth leaving cargodist turned off, just like hidden block signals.
Cargodist will make "casual" passenger systems less profitable, which will even more complicate life for those who do not know how to earn their first dollars. Also sometimes "cost: 1234" appear and will definitely raise some questions. As far as I remember, this does not bring real negative income, but it just looks bad, like something that should not be in a "simply working system". More or less experienced players will just ignore that, but for beginners it may be a sign that they are doing something wrong.
изображение

PeterN:
But if cargodist is going to cause confusion and misunderstanding, then I'm not sure it should even be moved out of Advanced.

therealbungus:
> But if cargodist is going to cause confusion and misunderstanding, then I'm not sure it should even be moved out of Advanced.

I believe it would only cause confusion and misunderstanding if set as a default setting for new games.

> On the OpenTTD Discord, new players ask how to give passengers destinations on a near-daily basis. We have a pinned message showing them how to change cargo distribution settings to do this. This currently requires showing advanced settings.

Moving this out of advanced settings would tackle the core issue which this PR was raised to address. Putting it next to the map generation settings is a decent compromise to make it more visible, because cargodist is a setting you usually enable or disable for an entire game, not something you would usually turn on or off in the middle of the game.

zephyris:
I've digested this a bit since my first comment, followed some of the chat on discord and stretched my memory (and dug through tt-forums) back to remember when cargodist was new... And I've changed my mind a bit. I'd forgotten about the old competing strategies for adding cargo destinations to the game. At the time I thought cargo _dest_ was more intuitive than cargo _dist_, in still think that's the case. There's a definite risk of confusion, cargo _dest_ is closer to how the world works, and while cargo _dist_ is a decent practical half-solution, it's perhaps not good enough to be a default... Now I mostly want a cargo _dest_ implementation to play with and see how it feels!

2TallTyler:
The people have spoken. No CargoDist today. 😃

ketsuban:
> It's often struck me that cargodist is 80% imagination - people like the idea of little sims beetling around their virtual world, and cargodist plays into that fantasy. [...] In the end, if you want to imagine little sims making connections and riding trains more than one stop, you can do that perfectly well without cargodist at all.

Maybe this doesn't reflect everyone else's experiences, but I find it hard to imagine that people are actually riding trains in a realistic fashion when the UI explicitly shows me everyone who got on the train at Worcester Foregate Street getting off again at Worcester Shrub Hill despite the train going to London Paddington.

> PikkaBird has a very good point about it not making the game more accessible or understandable to new players. When you look up guides, videos, or anything about how to play this game as a new player, pretty much nothing at all mentions cargodist mechanics.

This is a chicken-and-egg problem: nobody is going to make guides for new players explaining cargo distribution as long as cargo distribution is an off-by-default setting that newbies can just ignore, because the only people who turn it on are self-selecting for willingness to relearn part of the game on their own.

PikkaBird:
> This is a chicken-and-egg problem: nobody is going to make guides for new players explaining cargo distribution as long as cargo distribution is an off-by-default setting that newbies can just ignore, because the only people who turn it on are self-selecting for willingness to relearn part of the game on their own.

Back in the day, new players used to get directed to the OpenTTDCoop Way to Build Dumb Shit with Presignals™ all the time, despite it requiring changing a setting that _isn't even in the gui_. Eventually, that style of play became so commonplace that the config setting was changed to the Coop-friendly setting by default. So maybe someone needs to be the change they want to see, and create the newbie-friendly guide to cargodist?

PR #11440, opened 2023-11-04
Codechange: Add unit-test for nwidget parts.

PeterN:
## Motivation / Problem

NWidgetPart lists describing window layouts may have invalid parts that currently don't do anything and are ignored. This leads to confusion when creating or adjust window layouts.

## Description

This PR adds a unit test that creates a nested widget tree from the NWidgetPart list of every WindowDesc, using the same functions as used by the game.

As this is somewhat inside GUI code, this requires access to the FontCache and the SpriteCache, so this PR also contains a Mock FontCache implementation and a Mock SpriteCache implementation.

Finally, some exceptions are inserted in the method that processes the NWidgetPart lists to replace the nullptr checks that ignore the invalid parts. We don't normally use exceptions, but here they are necessary otherwise the unit tests are aborted.

## Limitations

This is quite a bit of extra code just to implement some sanity checks for things that are currently ignored...

## Checklist for review

### DIFF
--- a/src/spritecache_internal.h
+++ b/src/spritecache_internal.h
@@ -36,6 +36,7 @@ static inline bool IsMapgenSpriteID(SpriteID sprite)
 	return IsInsideMM(sprite, SPR_MAPGEN_BEGIN, SPR_MAPGEN_END);
 }
 
+void *AllocSprite(size_t mem_req);
 SpriteCache *AllocateSpriteCache(uint index);
 
 #endif /* SPRITECACHE_INTERNAL_H */
--- a/src/spritecache_internal.h
+++ b/src/spritecache_internal.h
@@ -1,0 +1,41 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file spritecache_internal.h Internal functions to cache sprites in memory. */
+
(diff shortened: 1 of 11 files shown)


src/widget.cpp
```diff
@@ -3117,64 +3117,60 @@ static const NWidgetPart *MakeNWidget(const NWidgetPart *nwid_begin, const NWidg
 
 			case WPT_RESIZE: {
 				NWidgetResizeBase *nwrb = dynamic_cast<NWidgetResizeBase *>(*dest);
-				if (nwrb != nullptr) {
-					assert(nwid_begin->u.xy.x >= 0 && nwid_begin->u.xy.y >= 0);
-					nwrb->SetResize(nwid_begin->u.xy.x, nwid_begin->u.xy.y);
-				}
+				if (unlikely(nwrb == nullptr)) throw std::runtime_error("WPT_RESIZE requires NWidgetResizeBase");
```
rubidium42, line 3120: Maybe a bit of bike shedding, but... might `std::domain_error` be a better exception to throw?

> It may be used by the implementation to report domain errors, that is, situations where the inputs are outside of the domain on which an operation is defined.

I'd say defining resize things for some object that doesn't resize is a domain error.

src/spritecache_internal.h
```diff
@@ -0,0 +1,42 @@
...
+#include "spriteloader/spriteloader.hpp"
+
+#include "table/sprites.h"
+
+/* These declarations are internal to spritecache but need to be exposed for unit-tests. */
+
+struct SpriteCache {
+	void *ptr;
+	size_t file_pos;
+	SpriteFile *file;    ///< The file the sprite in this entry can be found in.
```
rubidium42, line 26: Often you remove the extra spaces when changing/moving the lines. Was that intended to happen here too?

rubidium42 (approved):

src/widget.cpp
PeterN: I'm not that familiar with exceptions (any other exceptions we throw seem to be custom) and didn't know that domain_error exists.

If it's a better match for all of these then I can easily change it.

src/spritecache_internal.h
PeterN: Didn't really think of it, just copy & paste...

PR #11447, opened 2023-11-06
Change: Show empty string drop down entries as divider.

PeterN:
## Motivation / Problem

Some drop down lists contain empty strings to separate sections of the list. We also have a specific way to create drop down list divider, but this is only used (afaik) in two places, the main tool bar settings menu and the currency units menu.

A blank string is used instead for the save menu and the help menu.

## Description

Instead of complicating drop down list item creation, make `DropDownListStringItem::Draw` call the base `DropDownListItem::Draw` (which handles the divider) if the string is empty, and make the item unselectable for good measure.

No screenshots as I can't show toolbar menus without holding a mouse button, and that stops my printscreen from working :)

## Checklist for review

### DIFF
--- a/src/widgets/dropdown.cpp
+++ b/src/widgets/dropdown.cpp
@@ -48,10 +48,14 @@ uint DropDownListStringItem::Width() const
 	return GetStringBoundingBox(this->String()).width + WidgetDimensions::scaled.dropdowntext.Horizontal();
 }
 
-void DropDownListStringItem::Draw(const Rect &r, bool sel, Colours) const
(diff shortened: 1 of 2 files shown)


rubidium42 (approved):
Sounds like a good plan to unify this.

2TallTyler:
Can these now be cleaned up, or is skipping enum values still needed to indicate a divider?
https://github.com/OpenTTD/OpenTTD/blob/master/src/toolbar_gui.cpp#L388C1-L410C3

PeterN:
That's still needed, as the unused enum value is still assigned to the divider item -- if the list was to be built "manually" then you could skip the enum values that way,

PR #11452, opened 2023-11-09
Codechange: Replace AllocatedStringParameters with std::array based type

JGRennison:
## Motivation / Problem

All uses of AllocatedStringParameters are with a compile-time constant.
Use of a dynamically allocated buffer on the heap is unnecessary and increases overhead, particularly due to frequent use as a temporary.

## Description

Replace AllocatedStringParameters with a std::array-based equivalent.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -58,7 +58,7 @@ TextDirection _current_text_dir; ///< Text direction of the currently selected l
 std::unique_ptr<icu::Collator> _current_collator;    ///< Collator for the language currently in use.
 #endif /* WITH_ICU_I18N */
 
-AllocatedStringParameters _global_string_params(20);
+ArrayStringParameters<20> _global_string_params;
 
 /**
  * Prepare the string parameters for the next formatting run. This means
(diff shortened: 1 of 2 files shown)


rubidium42:
Digging back into my memory I seem to remember having run into the same compilation errors, which is why I eventually went for `std::vector`. I reckon something does a `std::move` or something similar, which invalidates the `span` pointing to the array.

JGRennison:
Some compilers don't seem to do the copy elision as expected so move assignment/construction need to update the span.

src/strings.cpp
```diff
@@ -1478,7 +1478,7 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 				} else if (_scan_for_gender_data) {
 					/* Gender is defined by the industry type.
 					 * STR_FORMAT_INDUSTRY_NAME may have the town first, so it would result in the gender of the town name */
-					auto tmp_params = AllocatedStringParameters();
+					auto tmp_params = ArrayStringParameters<0>();
```
LordAro, line 1481: Is a zero length array even valid? Bit strange regardless...

rubidium42: The specification specifically mention a zero length array, and how it should behave. I'd say it's better to leave the special case in the standard library, than having to implement a special string parameters class for when there are no parameters at all.

rubidium42 (approved):

PR #11461, opened 2023-11-15
Codechange: Move RectPadding::zero definition to header.

PeterN:
## Motivation / Problem

RectPadding::zero is declared const in a header but its definition is in a source file. This means that users of it do not necessarily know its values, possibly affecting inlining/optimization.

## Description

Move the definition to the header file so compilers are able to see the definition wherever it is used. This potentially allows better for inlining/optimization.

## Checklist for review

### DIFF
--- a/src/core/geometry_func.cpp
+++ b/src/core/geometry_func.cpp
@@ -48,5 +48,3 @@ Rect BoundingRect(const Rect &r1, const Rect &r2)
 
 	return r;
 }
-
-const RectPadding RectPadding::zero = {0, 0, 0, 0};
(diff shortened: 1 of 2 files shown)


2TallTyler (commented):
This loses the initialization value. I guess it defaults to 0, but is that best practice? (Honest question, I'm self-taught and don't know. 😛)

PeterN:
> This loses the initialization value. I guess it defaults to 0, but is that best practice? (Honest question, I'm self-taught and don't know. 😛)

Yes, it's now default initialization. Whether that is best practice or not I don't know.

2TallTyler (approved):
If it works, it works.

PR #11468, opened 2023-11-16
Codechange: Use ticks for timetable start date

2TallTyler:
## Motivation / Problem

In OpenTTD, days are two seconds long. In #11435, we show timetables in seconds, so unless we want to restrict players to using multiples of two, we can't use days for timetable-related things.

Currently, timetables use ticks for everything except the start date, mostly avoiding this problem.

## Description

We use the tick counter to set an absolute tick when the timetable starts. For clarity, I've created a new TickCounter type for this and converted the other use of an absolute tick (this can be a separate PR if we like the approach).

We convert `BaseConsist->timetable_start` from a date to this absolute tick, and handle saveload conversion.

The player sees no changes with this PR, with the user interface continuing to use a date. I also kept the validation in `CmdSetTimetableStart()` using days. We don't need to worry about invalid dates in the command, but they still show up in the GUI so they must continue to be valid dates.

## Limitations

There is an existing rounding error (confirmed in 13.4) when a vehicle reaches its first order and the timetable GUI shows "This vehicle is currently running N days early" -- it fails to account for `TimerGameCalendar::date_fract` and rounds down. Fixing this is out of scope for this PR.

## Checklist for review

### DIFF
--- a/src/timetable.h
+++ b/src/timetable.h
@@ -16,6 +16,9 @@
 
 static const TimerGameCalendar::Year MAX_TIMETABLE_START_YEARS = 15; ///< The maximum start date offset, in years.
 
+TimerGameTick::TickCounter GetStartTickFromDate(TimerGameCalendar::Date start_date);
+TimerGameCalendar::Date GetDateFromStartTick(TimerGameTick::TickCounter start_tick);
+
 void ShowTimetableWindow(const Vehicle *v);
 void UpdateVehicleTimetable(Vehicle *v, bool travelling);
 void SetTimetableParams(int param1, int param2, TimerGameTick::Ticks ticks);
--- a/src/timetable_cmd.cpp
+++ b/src/timetable_cmd.cpp
@@ -22,6 +22,40 @@
 #include "safeguards.h"
 
 /**
+ * Get the TimerGameTick::TickCounter tick of a given date.
+ * @param start_date The date when the timetable starts.
+ * @return The first tick of this date.
+ */
+TimerGameTick::TickCounter GetStartTickFromDate(TimerGameCalendar::Date start_date)
+{
+	/* Calculate the offset in ticks from the current date. */
+	TimerGameTick::Ticks tick_offset = (start_date - TimerGameCalendar::date).base() * Ticks::DAY_TICKS;
+
+	/* Compensate for the current date_fract. */
+	tick_offset -= TimerGameCalendar::date_fract;
+
+	/* Return the current tick plus the offset. */
+	return TimerGameTick::counter + tick_offset;
+}
+
+/**
+ * Get a date from a given start tick of timetable.
+ * @param start_tick The TimerGameTick::TickCounter when the timetable starts.
+ * @return The date when we reach this tick.
+ */
+TimerGameCalendar::Date GetDateFromStartTick(TimerGameTick::TickCounter start_tick)
+{
+	/* Calculate the offset in ticks from the current counter tick. */
+	TimerGameTick::Ticks tick_offset = start_tick - TimerGameTick::counter;
+
+	/* Compensate for the current date_fract. */
+	tick_offset += TimerGameCalendar::date_fract;
+
+	/* Return the current date plus the offset in days. */
+	return TimerGameCalendar::date + (tick_offset / Ticks::DAY_TICKS);
+}
+
+/**
  * Change/update a particular timetable entry.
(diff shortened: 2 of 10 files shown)


src/timetable_cmd.cpp
```diff
@@ -300,10 +300,10 @@ static bool VehicleTimetableSorter(Vehicle * const &a, Vehicle * const &b)
  * @param flags Operation to perform.
  * @param veh_id Vehicle ID.
  * @param timetable_all Set to set timetable start for all vehicles sharing this order
- * @param start_date The timetable start date.
+ * @param start_ticks Ticks until the timetable starts.
  * @return The error or cost of the operation.
  */
-CommandCost CmdSetTimetableStart(DoCommandFlag flags, VehicleID veh_id, bool timetable_all, TimerGameCalendar::Date start_date)
+CommandCost CmdSetTimetableStart(DoCommandFlag flags, VehicleID veh_id, bool timetable_all, TimerGameTick::Ticks start_ticks)
```
PeterN: I think this means the actual start time varies depending on network client latency. Not sure if that's a problem, but does make it a bit non-deterministic.

2TallTyler: Is `TimerGameCalendar::day_fract` not synced between clients?

2TallTyler: Or are you saying, send a start time in ticks but don't apply date_fract until we get inside the command?

PeterN: I'm saying that (I think) sending a relative ticks means it's relative to when the server receives the command, not relative to when the client sends the command. With the current absolute date that isn't the case (because it's absolute.)

2TallTyler: What about if we revert to sending the date (and eventually a `bool extra_second` once I add seconds mode) in the command call, but then actually store the start date as ticks? Would that be safe?

(Who actually decides when the command runs? The server?)

PeterN: Yes, the server decides. For now you could send as Date and the command can convert to ticks internally.

PeterN: Or, if the concept of "absolute ticks" exists, use that...

2TallTyler: There is an absolute tick counter, but it loops, so probably wouldn't be suitable.

JGRennison: TimerGameTick::counter is 64 bits so looping isn't a huge concern.
You could rebase all the values when it's near wrapping if this was really a concern.

PeterN: 64 bits is enough for 2^64 / 74 / 365 = 682 trillion years, I don't think wrapping is an issue.

Okay, so actually `timetable_start` could be also be an absolute tick, avoiding logic changes to how that works (no need to decrement every tick.) — it simply changes from Date to ticks. (But if making it relative internally helps you later than that's fine too.)

2TallTyler:
I'll do some thinking about absolute ticks, but in the meantime, we can go back to passing the date as a date. How do we feel about #11435 passing a `bool extra_second` to the command? Okay, or too messy where we'd want to rewrite stuff to use absolute ticks?

PeterN:
I'd prefer sending absolute ticks (which won't need to be changed again) instead of Date + bool (which will need changing when economy & calendar time are split)

2TallTyler:
The time split would just require TimerGameCalendar to become TimerGameEconomy, it's not a big change at all. It's in this commit, alongside a bunch of date GUI stuff: https://github.com/OpenTTD/OpenTTD/pull/10700/commits/eef786dfb574b41d1564fdff790bf3f1c1333388

PeterN:
Yeah, my point being that adding `bool extra_second` temporarily is a bit odd.

src/timetable.h
```diff
@@ -16,6 +16,40 @@
...
+{
+	/* Calculate the offset in ticks from the current counter tick. */
+	TimerGameTick::Ticks tick_offset = start_tick - TimerGameTick::counter;
+
+	/* Compensate for the current date_fract. */
+	tick_offset += TimerGameCalendar::date_fract;
+
+	/* Return the current date plus the offset in days. */
+	return TimerGameCalendar::date + (tick_offset / Ticks::DAY_TICKS);
+}
```
PeterN: Do these need to be inline functions? I think splitting definition and declaration is fine for these.

PeterN (approved):

PR #11471, opened 2023-11-18
Fix: Extra space allocated to container-within-container may not get allocated to children.

PeterN:
## Motivation / Problem

Extra space allocated to a container-within-container layout may not get allocated to the children of the container, as a shortcut incorrectly assumes that the current smallest size is the same as the size allocated to the child widgets. This is not always the case if NC_EQUALSIZE is used.

## Description

Always derive additional length from contained widgets instead of from the container, as the container's minimal length may have been adjusted by an NC_EQUALSIZE parent container.

This change essentially just removes a shortcut in favour of always performing a loop to get the real currently allocated size.

## Limitations

This does not actually affect any current layouts so there's no simple way to demonstrate this...

## Checklist for review

### DIFF
--- a/src/widget.cpp
+++ b/src/widget.cpp
@@ -1593,15 +1593,9 @@ void NWidgetHorizontal::AssignSizePosition(SizingType sizing, int x, int y, uint
 	assert(given_width >= this->smallest_x && given_height >= this->smallest_y);
 
 	/* Compute additional width given to us. */
-	uint additional_length = given_width;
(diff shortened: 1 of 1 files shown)


PeterN:
image

This is where I discovered the issue. There are two horizontal containers here, the outer one is marked NC_EQUALSIZE and contains the text label with fill_x = 1, the other containers another container with 3 buttons, one of which is marked as fill_x = 1.

During allocation of space, the outer container correctly allocates the same width to both its children (the text and the child container), but without this fix, the child container does not allocate that space to its own children.

image

2TallTyler (approved):

PR #11474, opened 2023-11-19
Codechange: Simplify setting minimum width for game options dropdown lists.

PeterN:
## Motivation / Problem

Looping to find maximum width when a function exists to do that already.

## Description

Height isn't necessary as drop down controls already have sufficient height, so we can use GetDropDownListDimension() to get the width.

## Checklist for review

### DIFF
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -445,17 +445,8 @@ struct GameOptionsWindow : Window {
 
 			default: {
 				int selected;
-				DropDownList list = this->BuildDropDownList(widget, &selected);
-				if (!list.empty()) {
-					/* Find the biggest item for the default size. */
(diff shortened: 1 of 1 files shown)


2TallTyler (commented):
Just checking... We no longer care about the string height? Also, do we need the `selected` that does nothing?

PeterN:
> Just checking... We no longer care about the string height? Also, do we need the `selected` that does nothing?

In the commit message:
> Height isn't necessary as drop down controls already have sufficient height, so we can use GetDropDownListDimension() to get the width.

And `selected` is an out-parameter, so it's needed even though we don't use it here.

2TallTyler (approved):
Ah, didn't see the commit message on mobile. 🙂

PR #11480, opened 2023-11-21
Fix: Restore original cargo legend 'blob' dimensions.

PeterN:
## Motivation / Problem

"Fix" in the losest terms... An off-by-one in a previous life made the 'blob' too narrow:

OpenTTD:

image

TTD:

image

## Description

Change the calculate from `(FONT_HEIGHT - 1) * 8 / 5` to `FONT_HEIGHT * 9 / 6`:

image

## Limitations

Duplicated arithmetic but it can't be a single constant...

## Checklist for review

### DIFF
--- a/src/graph_gui.cpp
+++ b/src/graph_gui.cpp
@@ -901,7 +901,7 @@ struct PaymentRatesGraphWindow : BaseGraphWindow {
 	void OnInit() override
 	{
 		/* Width of the legend blob. */
-		this->legend_width = (GetCharacterHeight(FS_SMALL) - ScaleGUITrad(1)) * 8 / 5;
+		this->legend_width = GetCharacterHeight(FS_SMALL) * 9 / 6;
 	}
 
 	void UpdateExcludedData()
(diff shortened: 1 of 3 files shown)


2TallTyler (commented):
Why can't it be a `const int CARGO_LEGEND_BLOB_WIDTH = FONT_HEIGHT_SMALL * 9 / 6`?

`this->legend.width` and the like are already `FONT_HEIGHT_SMALL`, so it's the same calculation.

PeterN:
Because FONT_HEIGHT_SMALL is not a constant, it's a sneaky macro to call GetCharacterHeight(FS_SMALL) that was used to simplify updating code long ago...

PeterN:
But good call, there's no reason we can't search & replace all those :-)

2TallTyler:
Scope creep!

Do you want to do that first, or merge this and add that to the "later" list?

2TallTyler (dismissed):

2TallTyler (approved):

PR #11487, opened 2023-11-23
Feature: Show cargo icon on cargo filter lists.

PeterN:
## Motivation / Problem

Cargo icons are only visible in the station view window, when stations have waiting cargo. Painstakingly drawn 10 pixel high pixel art masterpieces are being denied to us.

## Description

Add cargo icons to cargo filter dropdown lists, and revel in beauty of them.

Cargo icons have various widths so we have a function to get the largest and make sure all items use the same width.
<s>Cargo icons have various widths so they are drawn at the end of the item. Other positions were tested and were less visually coherent.</s>

This also introduces the CRTP (Curiously Recursive Template Pattern) redesign for drop down list items, which allows them to be composed of various parts instead of to a fixed design, but unfortunately means they are templates to are implemented in the head filters.

<s>https://github.com/OpenTTD/OpenTTD/assets/639850/00fe6c55-118b-4eff-86a5-38de324e123e</s>

image

## Checklist for review

### DIFF
--- a/src/widgets/dropdown_type.h
+++ b/src/widgets/dropdown_type.h
@@ -12,60 +12,199 @@
 
 #include "../window_type.h"
 #include "../gfx_func.h"
-#include "table/strings.h"
+#include "../gfx_type.h"
+#include "../string_func.h"
+#include "../strings_func.h"
+#include "../table/strings.h"
+#include "../window_gui.h"
 
 /**
- * Base list item class from which others are derived. If placed in a list it
- * will appear as a horizontal line in the menu.
+ * Base list item class from which others are derived.
  */
 class DropDownListItem {
 public:
-	int result;  ///< Result code to return to window on selection
-	bool masked; ///< Masked and unselectable item
+	int result; ///< Result value to return to window on selection.
+	bool masked; ///< Masked and unselectable item.
+	bool shaded; ///< Shaded item, affects text colour.
 
-	DropDownListItem(int result, bool masked) : result(result), masked(masked) {}
+	explicit DropDownListItem(int result, bool masked = false, bool shaded = false) : result(result), masked(masked), shaded(shaded) {}
 	virtual ~DropDownListItem() = default;
 
-	virtual bool Selectable() const { return false; }
-	virtual uint Height() const { return GetCharacterHeight(FS_NORMAL); }
+	virtual bool Selectable() const { return true; }
+	virtual uint Height() const { return 0; }
 	virtual uint Width() const { return 0; }
-	virtual void Draw(const Rect &r, bool sel, Colours bg_colour) const;
+
+	virtual void Draw(const Rect &full, const Rect &, bool, Colours bg_colour) const
+	{
+		if (this->masked) GfxFillRect(full, _colour_gradient[bg_colour][5], FILLRECT_CHECKER);
+	}
+
+	TextColour GetColour(bool sel) const
+	{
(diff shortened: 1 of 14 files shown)


src/widgets/dropdown_type.h
```diff
@@ -12,61 +12,209 @@
...
+		if (this->checked) {
+			DrawStringMultiLine(r.WithWidth(this->dim.width, rtl), STR_JUST_CHECKMARK, this->GetColour(sel), SA_CENTER, false, TFs);
+		}
+		this->TBase::Draw(full, r.Indent(this->dim.width + WidgetDimensions::scaled.hsep_wide, rtl), sel, bg_colour);
+	}
+};
+
+/* Commonly used drop down list items. */
+using DropDownListDividerItem = DropDownDivider<DropDownListItem>;
+using DropDownListStringItem = DropDownString<DropDownDivider<DropDownListItem>>;
```
glx22: I don't fully get why `DropDownDivider` as base of `DropDownString`

PeterN: Because the old base DropDownListItem provided the divider itself. The alternative is to undo #11447, and explicitly create a DropDownListDividerItem. That may not be too much hassle now.

DefinitelyNotRau117:
Good idea, it will really improve the clarity of filter lists! 
But on the “very wide” windows a problem may arise with “fictitiously misalighed” icons when placing icons after text. To be honest, I didn’t even notice these icons on your screenshot…
изображение
What about placing an icon before the text? This will help to visually connect “icon-cargoname” pairs without having to refocus to the other end of the window.
изображение

2TallTyler:
We discussed icons before the text on Discord. It didn't look great because icons are all different sizes, so their alignment with text looks bad. 

I'd rather see the dropdown clamped to a suitable width (using a spacer to fill the empty space) to avoid the problem. 🙂

PeterN:
I wasn't making this up...

> Cargo icons have various widths so they are drawn at the end of the item. Other positions were tested and were less visually coherent.

image

planetmaker:
Arguably the icons don't match well in width. However that's a task for the (New)GRF authors to take care of. The list with the icons on the left still looks better to me than with the icons on the right with the huge separation from the text.

Could the maximum size of the icons be chosen as icon column width and all icons be centered wrt the icon column instead of left-aligned to make it look better?

PeterN:
At the end of the day, I prefer it on the right.

LordAro:
Add a setting ;)

But I do agree that slightly less of a gap would be better

PeterN:
image

PeterN:
And with a checkmark, which is also on the left to match the menu items on the settings toolbar menu:

image

PeterN:
Okay, now changed to put the icon on the left, along with a few other changes to make that easier to perform.

src/widgets/dropdown_type.h
```diff
@@ -12,61 +12,200 @@
...
+ * @tparam TFs Font size.
+ * @tparam TEnd Position string at end if true, or start if false.
+ */
+template<class TBase, FontSize TFs = FS_NORMAL, bool TEnd = false>
+class DropDownString : public TBase {
+	std::string string; ///< String to be drawn.
+	Dimension dim; ///< Dimensions of string.
+public:
+	template <typename... Args>
+	explicit DropDownString(StringID string, Args&&... args) : TBase(std::forward<Args>(args)...)
```
rubidium42, line 86: Oh, nice idea/trick when using CRTP.

rubidium42 (approved):

PR #11489, opened 2023-11-24
Feature: Randomize direction of rail vehicle on build based on probability callback.

PeterN:
## Motivation / Problem

NewGRF authors have be known to implement randomisation of engine direction for variety. An engine may be built facing forward or backward.

Unfortunately this requires using a random bit to record whether the vehicle is reverse, and it requires duplicating the sprites to be drawn, along with probably duplicating act3/2/1 chains.

This is quite a lot of complexity, considering the game already has the ability to let uses flip engines in the depot manually.

## Description

<s>Implement automatic randomization using 2 bits of the ExtraFlags property.</s>

Implement automatic randomization using a new callback, CBID_VEHICLE_BUILD_PROBABILITY.

To use this callback, test for var 0C = 0x0162 and var 10 = 0x00, and return a value between 0...100 decimal to indicate the probability of being reversed.

This allows NewGRF authors to indicate that the game should randomly flip rail vehicles on build, without needing to use random bits nor duplicate sprites to handle it themselves.

<s>Bits 4 and 5 of ExtraFlags are used to create a 2-bit value ranging from 0-3, resulting a random probability of 0%, 16%, 33% and 50%.</s>

If said randomization decides the vehicle should be reverse, then the appropriate VRF_REVERSE_DIRECTION bit is set. This can still be toggled by the player.

Authors do not need to use a random bit to store this state as it is a built-in flag, and importantly they do not need to duplicate sprites.

<s>This is effective only if the vehicle is built as a single unit.</s>

image

## Limitations

This uses 2 bits from ExtraFlags, which could possibly be considered abuse, However adding a new property would prevent compatibility with previous game versions which will simply ignore these flags.

## Checklist for review

### DIFF
--- a/src/newgrf_callbacks.h
+++ b/src/newgrf_callbacks.h
@@ -282,6 +282,9 @@ enum CallbackID {
 
 	/** Called to determine the engine name to show. */
 	CBID_VEHICLE_NAME                    = 0x161, // 15 bit callback
+
+	/** Called to determine probability during build. */
+	CBID_VEHICLE_BUILD_PROBABILITY       = 0x162, // 15 bit callback
 };
 
 /**
--- a/src/newgrf_engine.cpp
+++ b/src/newgrf_engine.cpp
@@ -1192,6 +1192,20 @@ int GetEngineProperty(EngineID engine, PropertyID property, int orig_value, cons
 	return orig_value;
 }
 
+/**
+ * Test for vehicle build probablity type.
+ * @param v Vehicle whose build probability to test.
+ * @param type Build probability type to test for.
+ * @returns True iff the probability result says so.
+ */
+bool TestVehicleBuildProbability(Vehicle *v, EngineID engine, BuildProbabilityType type)
+{
+	uint16_t p = GetVehicleCallback(CBID_VEHICLE_BUILD_PROBABILITY, std::underlying_type<BuildProbabilityType>::type(type), 0, engine, v);
+	if (p == CALLBACK_FAILED) return false;
+
+	const uint16_t PROBABILITY_RANGE = 100;
+	return p + RandomRange(PROBABILITY_RANGE) >= PROBABILITY_RANGE;
+}
 
 static void DoTriggerVehicle(Vehicle *v, VehicleTrigger trigger, uint16_t base_random_bits, bool first)
 {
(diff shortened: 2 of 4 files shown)


andythenorth:
I am not watching PRs very much (nor getting notifications), but if this is merged, I am happy to update spec tracker, ping me on discord? https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status

2TallTyler:
I see three use cases for randomly reversed locomotives:

1. The engine is double-ended but is visually asymmetrical (perhaps in small ways like exhaust stacks or other greebles) and you want to randomize appearance. For example, most UK and European diesels and electric locomotives. Set to 50%.
2. The engine is single-ended but sometimes runs backwards. For example, tank engines, some U.S. road-switcher locomotives. Set to a lower probability, depending on taste.
3. The engine is single-ended but _when operated in multiple_, the leading engine always faces forwards while trailing engines may face either direction. For example, every every non-steam locomotive in the world including every U.S. freight train. This isn't currently possible.

I propose that this PR should support the third case. Is there a way to randomly flip, _unless_ it's the first vehicle in the chain?

ldpl:
Having is only as 2-bit probability value feels kind of arbitrary.
I think this is better be implemented via callback that can be run after vehicle creation and return whether is should be reversed.
Not sure if it's possible to give that callback enough context to implement third of @2TallTyler cases but at least it should be able to do whatever probability newgrf author wishes and even easter eggs like that drunk workers thing.

PeterN:
It is very possible to flip multi-head engines, it is explicitly disabled here because it is also disabled for manually flipping in the UI.

It may even be desirable and possible to allow articulated parts to be automatically flipped -- if the author doesn't want that they can just not set the flags.

Brickblock1:
Any reason not to have a flag to controll it for players as well?
Could sometimes be nice to prevent  flipping when the sprites aren't meant for it, and the option to do it for dual-ended and articulated units would be a nice extra addition.

PeterN:
What does "when the sprites aren't meant for it" mean?

Brickblock1:
I probably should have specified that I meant that lights would possibly be incorrect but I could see it being of use just to prevent flipping for any reason.

PeterN:
Does that benefit a player?

PeterN:
Now written as a callback-only approach which gives more flexibility to authors.

2TallTyler:
I guess there's no way for a callback to know if it's the first vehicle in a chain. But then again, this wouldn't really work even for the former property-based system.

rubidium42 (approved):

ldpl:
Figured it's super simple to make a test grf for this with grf-py so here it is.
[pr11489-test.zip](https://github.com/OpenTTD/OpenTTD/files/13504983/pr11489-test.zip)

andythenorth:
Newgrf spec done https://newgrf-specs.tt-wiki.net/wiki/Callbacks#Probability_of_reversing_rail_vehicle_on_build_.28162.29

andythenorth:
NML PR https://github.com/OpenTTD/nml/pull/313

PR #11477, opened 2023-11-20
Fix 6065122: Missing SetFill on worldgen spacer.

PeterN:
## Motivation / Problem

Missing SetFill() on top-right spacer of world gen layout caused incorrect spacing if the dropdown menus are taller than 'normal'. This is the case with OpenGFX2, which is why this was not noticed with the TTD baseset.

image

## Description

Add SetFill() to spacer widget to allow it to be resized and match height.

image

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, game_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11490, opened 2023-11-24
Codechange: Build dropdown toolbar menus from explicit lists of strings.

PeterN:
## Motivation / Problem

As pointed out by @2TallTyler in https://github.com/OpenTTD/OpenTTD/pull/11447#issuecomment-1798608476, the toolbar dropdown lists are a bit opaque and require skipping enum values.

## Description

Build toolbar dropdowns by explicitly listing the strings instead of counting from the starting string. Empty divider options are marked by STR_NULL (which is simple to test for).

Because the divider is explicitly created instead of relying on a blank string, the special-casing used in DropDownListStringItem (implemented by #11477) is not necessary.

Return values are consecutive but dividers do not increment the value.

## Checklist for review

### DIFF
--- a/src/toolbar_gui.cpp
+++ b/src/toolbar_gui.cpp
@@ -200,14 +200,19 @@ static void PopupMainToolbMenu(Window *w, int widget, DropDownList &&list, int d
  * Pop up a generic text only menu.
  * @param w Toolbar
  * @param widget Toolbar button
- * @param string String for the first item in the menu
- * @param count Number of items in the menu
+ * @param strings List of strings for each item in the menu
  */
-static void PopupMainToolbMenu(Window *w, int widget, StringID string, int count)
+static void PopupMainToolbMenu(Window *w, int widget, const std::initializer_list<StringID> &strings)
 {
 	DropDownList list;
-	for (int i = 0; i < count; i++) {
-		list.push_back(std::make_unique<DropDownListStringItem>(string + i, i, false));
+	int i = 0;
+	for (StringID string : strings) {
+		if (string == STR_NULL) {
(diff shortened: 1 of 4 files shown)


src/toolbar_gui.cpp
```diff
@@ -1047,7 +1061,7 @@ static CallBackFunction MenuClickForest(int index)
 
 static CallBackFunction ToolbarMusicClick(Window *w)
 {
-	PopupMainToolbMenu(w, _game_mode == GM_EDITOR ? (int)WID_TE_MUSIC_SOUND : (int)WID_TN_MUSIC_SOUND, STR_TOOLBAR_SOUND_MUSIC, 1);
+	PopupMainToolbMenu(w, _game_mode == GM_EDITOR ? (int)WID_TE_MUSIC_SOUND : (int)WID_TN_MUSIC_SOUND, {STR_TOOLBAR_SOUND_MUSIC});
```
2TallTyler: Why the `(int)` casting when a ternary is used, here and in `ToolbarHelpClick()`? Other uses don't seem to need it.

2TallTyler (commented):
The reversal of #11447 is missing the change to `widgets/dropdown_type.h`. Not sure if that's intended.

PeterN:
> The reversal of #11447 is missing the change to `widgets/dropdown_type.h`. Not sure if that's intended.

Not intended

src/toolbar_gui.cpp
PeterN: Two different enums can't be used in a ternary. The others do variably cast to byte, uint or int which is inconsistent, but I'm not touching this in this PR.

2TallTyler (approved):

PR #11491, opened 2023-11-25
Fix: Add shading to river slopes

zephyris:
Adds a 32bpp shading to the river rapids/slopes to make them more visible. Fixes #9031
slopeshadig

## Motivation / Problem
Water slopes on rivers are not readily visible as sloped, making it confusing why ships cannot sail through them. 

## Description
This fix adds 32bpp sloped river water sprites to the orig_extra.grf. By using the 8bpp sprite as a mask, this adds shading to the slope while retaining the animation and 8bit style. Requires a bump of original graphics grfs to use container version 2.

## Limitations
Does not work with 8bpp blitters.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, game_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11492, opened 2023-11-25
Change: When building NewGRFs, use grfid (from grfcodec) to calculate NewGRF MD5 hash.

PeterN:
## Motivation / Problem

#11491 adds 32bpp sprites to the extra data NewGRF, which requires it to be built as container v2, but our build scripts do not correctly calculate the MD5 hash for container v2 NewGRFs.

`Baseset.cmake` uses CMake's built in MD5 function to calculate the hash of `orig_extra.grf`. However hash calculation only works for v1 container grfs.

The tool `grfid` from grfcodec can be used to calculate both v1 and v2 hashes, however grfcodec isn't currently a build dependency.

## Description

<s>Vendor grfid (with some changes due to our updated Md5 implementation and use of fmt, and use it in the build process.

Switch to NewGRF container v2. A side effect of this is that the files are a about 10% larger.</s>

Update the build system to use grfid from grfcodec to calculate the hash of the custom NewGRFs. This requires a bit more effect than before as MD5 was a nice cmake built-in. The hash is written to a file which is stored in the source tree to enable OpenTTD to be compiled if grfcodec is not present.

## Limitations

<s>I removed the copyright message on grfid -- it's my name but I never put that there, so...</s>

## Checklist for review


src/grfid/grfid.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+		lseek(this->fd, 0, SEEK_SET);
+
+#ifndef _WIN32
+		this->file_buffer = static_cast<uint8_t *>(mmap(NULL, this->file_length, PROT_READ, MAP_PRIVATE, fd, 0));
+		if (this->file_buffer == nullptr) return false;
+#else
+		this->file_buffer = new uint8_t[this->file_length];
+		if (this->file_buffer == nullptr) return false;
+		size_t file_read = read(this->fd, this->file_buffer, this->file_length);
```
glx22: MSVC warnings are valid (the function expects a 32bit unsigned int) and `file_length` is 64bit.
But `lseek` return is a 32bit signed value, so an explicit cast should be fine.

PeterN: Who would ever have a 2GB file... but okay, unlikely for a NewGRF :D

media/baseset/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ if(GRFCODEC_FOUND)
 endif()
 
 set(BASESET_EXTRAGRF_FILE ${CMAKE_CURRENT_SOURCE_DIR}/orig_extra.grf)
+set(BASESET_EXTRAGRF_HASH ${CMAKE_CURRENT_SOURCE_DIR}/orig_extra.grf.md5)
```
glx22:
```suggestion

if (HOST_BINARY_DIR)
    include(${HOST_BINARY_DIR}/grfid.cmake)
endif()
```

media/baseset/CMakeLists.txt
```diff
@@ -38,17 +39,25 @@ foreach(BASESET_SOURCE_FILE IN LISTS BASESET_SOURCE_FILES)
 
     get_target_property(LANG_SOURCE_FILES language_files LANG_SOURCE_FILES)
 
+    add_custom_command(
+        OUTPUT ${BASESET_EXTRAGRF_HASH}
+        COMMAND grfid -m ${BASESET_EXTRAGRF_FILE} > ${BASESET_EXTRAGRF_HASH}
+        DEPENDS grfid ${BASESET_EXTRAGRF_FILE}
+    )
+
```
glx22: And I'd extract this outside of the `foreach`

glx22:
Might be simpler to generate md5 file only if cmake has found grfcodec (using grfid that comes with it) and store the md5 file in repo for when grfcodec is not found (like we do for the grf themselves).

TrueBrain:
What is a bit missing in the motivation, is why you want to do this.

I kinda deduce that you want to upgrade from v1 to v2? Maybe for the history add a few words in the description why that is something we would like.

Not disgreeing, to be clear, I was just missing the reason why :)

PeterN:
Now updated to just use grfid from grfcodec only when building the NewGRFs as well. Still a bit of wrangling going on but at least there's no code being duplicated.

cmake/scripts/CreateGRF.cmake
```diff
@@ -55,4 +58,18 @@ if(RESULT)
...
+if(RESULT)
+    if(NOT RESULT MATCHES "^[0-9]*$")
+        message(FATAL_ERROR "Failed to run GRFID (${RESULT}), please check GRFID_EXECUTABLE variable")
+    endif()
+    message(FATAL_ERROR "GRFID failed")
+endif()
+
+# Trim hash from grfid
+string(STRIP ${GRFID_HASH} GRFID_HASH)
+file(WRITE ${CMAKE_SOURCE_DIR}/${GRF_SOURCE_FOLDER_NAME}.grf.hash ${GRFID_HASH})
```
glx22: `${CMAKE_SOURCE_DIR}` is current working dir for script ran with `cmake -P`
But 
```suggestion
file(WRITE ${GRF_BINARY_FILE}.hash ${GRFID_HASH})
```
should work without needing the manual copy.

PeterN: Oops, that was me trying things that didn't do anything, and not reverting it :)

cmake/CreateGrfCommand.cmake
```diff
@@ -42,9 +43,14 @@ function(create_grf_command)
...
+    # For convenience, if you want to only test building the GRF
     add_custom_target(${GRF_SOURCE_FOLDER_NAME}.grf
             DEPENDS
             ${GRF_BINARY_FILE}
     )
+
+    add_custom_target(${GRF_SOURCE_FOLDER_NAME}.grf.hash
+            DEPENDS
+            ${GRF_BINARY_FILE}
+    )
```
glx22: This target will do the same as `${GRF_SOURCE_FOLDER_NAME}.grf` target, I don't think it's needed to add it.

glx22 (approved):

Issue #11488, opened 2023-11-24
[Bug]: spritecache excessive size


tamlin-mike:
### Version of OpenTTD

1.13.4-64, Win10-64

### Expected result

A reasonable amount of sprite cache being allocated.

### Actual result

768 MB (8.05*10^6 bytes) of spritecache is allocated on startup.

Three quarter of a gigabyte. For OpenTTD's default tileset.

On entry of `GfxInitSpriteCache`, `_sprite_cache_size` is 128, which while quite excessive already (considering the default is still claimed in the code to be 4MB), this value is quadrupled by running on a 32bpp display.

`IniLoadSettings` updates it to 128. Since I have not modified anything, I didn't bother digging deeper into it than setting a data breakpoint and find where it was modified from.

Assuming a normal (by todays standards) 32bpp output device;

1. The comment for `_sprite_cache_size` in spritecache.cpp claims it's number of MB, which it's clearly not. It's mega _pixels_. Well, it's 66% of the megapixels actually allocated.
2. In `GfxInitSpriteCache()`, `_allocated_sprite_cache_size` (and a few local variables holding the same value) are all uint. Should `_sprite_cache_size` ever be > 512, the code would overflow `target_size` and therefore `_allocated_sprite_cache_size`.
4. Actually, since the code is growing it by _another_ 50%, it would cause meltdown with `_sprite_cache_size` >= 342, I think.

### Steps to reproduce

1. Run openttd.exe.
2. Done.

TrueBrain:
We are a bit puzzled .. what is the actual problem you found here? Why is the default not "reasonable"? By which standard is that measured?
The tone of your whole post reads very passive aggressive, like we made the worst sin possible, but I fail to see what we actually did wrong :) So I will classify it as a language barrier :) But please do help us out understanding what you actually find troubling.

Anyway, to answer your questions:

1. The `_sprite_cache_size` comment is written in a time of 8bpp. With 32bpp, it is no longer true. But it is also not in mega pixels. It is in the amount of bytes a pixel consumes. So Mega Pixels Bytes? MPB! Anyway, this means it is 1/8th of the amount of pixels, not 66% :) Or, if you add the extra 50% we might allocate: 1/8th + 1/16th = 3/16th of the amount of pixels. So yeah, the comment in the code is not correct; this could be fixed.
[...]

PR #11495, opened 2023-11-26
Change: Simplify sprite cache memory management 

JGRennison:
## Motivation / Problem

Rather dated memory management style in the sprite cache.
Poor performance whenever the cache reached the size limit.

## Description

Remove custom allocator, memory management, compaction, etc.
~~Use std::vector for sprite data~~ Use std::unique_ptr for sprite data
Perform LRU cache eviction in a single pass. Previously each evicted sprite had an O(N) cost, which tended to O(N^2) overall.

Also fix DupSprite not properly copying the control_flags field.

## Limitations

Doesn't change what is being cached.
Doesn't reduce total sprite memory requirements.
Doesn't substantially change the cache eviction policy.

## Checklist for review

### DIFF
--- a/src/spritecache.cpp
+++ b/src/spritecache.cpp
@@ -8,11 +8,8 @@
 /** @file spritecache.cpp Caching of sprites. */
 
 #include "stdafx.h"
-#include "random_access_file_type.h"
 #include "spriteloader/grf.hpp"
 #include "spriteloader/makeindexed.h"
-#include "gfx_func.h"
-#include "error.h"
 #include "error_func.h"
 #include "strings_func.h"
 #include "zoom_func.h"
@@ -24,7 +21,6 @@
 #include "spritecache_internal.h"
 
 #include "table/sprites.h"
-#include "table/strings.h"
 #include "table/palette_convert.h"
 
 #include "safeguards.h"
@@ -34,6 +30,8 @@ uint _sprite_cache_size = 4;
 
 
 static std::vector<SpriteCache> _spritecache;
+static size_t _spritecache_bytes_used = 0;
+static uint32_t _sprite_lru_counter;
 static std::vector<std::unique_ptr<SpriteFile>> _sprite_files;
 
 static inline SpriteCache *GetSpriteCache(uint index)
@@ -96,18 +94,6 @@ SpriteFile &OpenCachedSpriteFile(const std::string &filename, Subdirectory subdi
 	}
 	return *file;
 }
-
-struct MemBlock {
-	size_t size;
-	uint8_t data[];
-};
-
-static uint _sprite_lru_counter;
-static MemBlock *_spritecache_ptr;
-static uint _allocated_sprite_cache_size = 0;
-static int _compact_cache_counter;
-
-static void CompactSpriteCache();
 
 /**
  * Skip the given amount of sprite graphics data.
@@ -641,7 +627,6 @@ bool LoadNextSprite(SpriteID load_index, SpriteFile &file, uint file_sprite_id)
 	uint8_t grf_type = file.ReadByte();
 
 	SpriteType type;
-	void *data = nullptr;
 	uint8_t control_flags = 0;
 	if (grf_type == 0xFF) {
 		/* Some NewGRF files have "empty" pseudo-sprites which are 1
@@ -692,7 +677,6 @@ bool LoadNextSprite(SpriteID load_index, SpriteFile &file, uint file_sprite_id)
 	sc->file = &file;
 	sc->file_pos = file_pos;
 	sc->length = num;
-	sc->ptr = data;
 	sc->lru = 0;
 	sc->id = file_sprite_id;
 	sc->type = type;
@@ -710,7 +694,7 @@ void DupSprite(SpriteID old_spr, SpriteID new_spr)
 
 	scnew->file = scold->file;
 	scnew->file_pos = scold->file_pos;
-	scnew->ptr = nullptr;
+	scnew->ClearSpriteData();
 	scnew->id = scold->id;
 	scnew->type = scold->type;
 	scnew->warned = false;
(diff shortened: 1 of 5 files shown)


src/spritecache_internal.h
```diff
@@ -20,15 +20,55 @@
...
+	size_t size{};
+
+public:
+	bool IsEmpty() const { return this->buffer == nullptr; }
+	void *GetData() const { return static_cast<void *>(this->buffer.get()); }
+	size_t GetSize() const { return this->size; }
+
+	void Allocate(size_t size)
+	{
+		this->buffer.reset(new byte[size]);
```
PeterN: `this->buffer = std::make_unique<byte[]>(size);`

TrueBrain:
From what I can tell, you silently remove the piece of code that made sure OpenTTD never allocates all available physical memory for the spritecache. The current code does this "50%" trick, where it over-allocates the sprite-cache buffer by 50% to ensure that fits, before it changes to it actual size.
Basically, it creates a dynamic upperbound to the sprite-cache size, based on the actual physical memory available.

There is some merit in that solution, as for computers with low amount of memory, it avoids OpenTTD being the reason of starvation (or worse, swapping).

That said, I am not saying it is a bad thing to remove that, or that we should keep it. Just highlighting that piece of code now appears to be removed :)

src/spritecache.cpp
```diff
@@ -1045,9 +896,8 @@ void GfxClearSpriteCache()
 void GfxClearFontSpriteCache()
 {
 	/* Clear sprite ptr for all cached font items */
-	for (uint i = 0; i != _spritecache_items; i++) {
-		SpriteCache *sc = GetSpriteCache(i);
-		if (sc->type == SpriteType::Font && sc->ptr != nullptr) DeleteEntryFromSpriteCache(i);
+	for (SpriteCache &sc : _spritecache) {
+		if (sc.type  == SpriteType::Font && !sc.data.IsEmpty()) sc.ClearSpriteData();
```
TrueBrain: nitpick: two spaces before `==` instead of one :P

src/spritecache.cpp
```diff
@@ -669,203 +649,125 @@ bool LoadNextSprite(int load_index, SpriteFile &file, uint file_sprite_id)
...
-/* make sure it's a power of two */
-static_assert((sizeof(size_t) & (sizeof(size_t) - 1)) == 0);
-
-static inline MemBlock *NextBlock(MemBlock *block)
-{
-	return (MemBlock*)((byte*)block + (block->size & ~S_FREE_MASK));
-}
-
-static size_t GetSpriteCacheUsage()
+static void DeleteEntriesFromSpriteCache(size_t target)
```
TrueBrain: nitpick: maybe add a doxygen comment here what `target` does. One can read it as: the size, the count, the amount to remove, the size to remove, etc. It is a bit ambiguous. But nothing a comment cannot fix :)

LordAro:
Should this close #11488 ?

src/spritecache_internal.h
```diff
@@ -20,15 +20,55 @@
 
 /* These declarations are internal to spritecache but need to be exposed for unit-tests. */
 
+class SpriteDataBuffer {
```
TrueBrain: nitpick: don't we have an STL container for this? It feels like something that should already exist in the STL :P

Not saying we should use it btw; I am fine with this. I was just looking a bit too deep into this PR, as I was kinda curious, and got dragged away :P

nielsm suggests maybe a `std::vector<char>`?

PeterN: Using std::vector would remove the need for the custom `SpriteDataBuffer` class, at the expense of an extra memory used internally by vector, and time spent initializing the buffer when the vector's size is set.

It would be smaller code though, which is nice.

src/spritecache.cpp
```diff
@@ -669,203 +649,133 @@ bool LoadNextSprite(int load_index, SpriteFile &file, uint file_sprite_id)
...
-		SpriteCache *sc = GetSpriteCache(i);
-		if (sc->type != SpriteType::Recolour && sc->ptr != nullptr && sc->lru < cur_lru) {
-			cur_lru = sc->lru;
-			best = i;
-		}
+	_spritecache_bytes_used -= this->data.size();
+	if (ptr != nullptr) {
+		assert(static_cast<void *>(_last_sprite_allocation.data()) == ptr);
+		this->data = std::move(_last_sprite_allocation);
+		_spritecache_bytes_used += this->data.size();
```
PeterN: vector has a swap() method which might be useful rather than messing about with std::move (which technically makes the moved object invalid, I believe.)

src/tests/mock_spritecache.cpp
```diff
@@ -17,14 +17,14 @@
 
 static bool MockLoadNextSprite(int load_index)
 {
-	static Sprite *sprite = (Sprite *)AllocSprite(sizeof(*sprite));
+	void *ptr = AllocSprite(sizeof(Sprite));
```
PeterN: Hmm, this was specifically static to use the same buffer and avoid allocating thousands of dummy sprites, but of course that's not compatible when using a vector. Oh well, that doesn't really matter much. :)

src/spritecache.cpp
JGRennison: std::move does not make the moved object invalid.
It isn't required to do a move in the other direction as well.

PeterN: Okay, most moved objects are in a "valid but unspecified state", but vector is specifically valid and empty, apparently.

PeterN:
Turns out that using a std::vector there can add quite a bit of memory...

master:
> dbg: [sprite] Increasing sprite cache to 129024 items (5160960 bytes)

pr11495:
> dbg: [sprite] Increasing sprite cache to 128666 items (7205296 bytes)

Also is there a way to guide the capacity of the _spritecache vector? I'm assuming this is system specific, but a lot of memory is wasted if you are just above a threshold...

dbg: [sprite] Increasing sprite cache to 1 items (56 bytes)
dbg: [sprite] Increasing sprite cache to 2 items (112 bytes)
dbg: [sprite] Increasing sprite cache to 4 items (224 bytes)
dbg: [sprite] Increasing sprite cache to 8 items (448 bytes)
dbg: [sprite] Increasing sprite cache to 16 items (896 bytes)
dbg: [sprite] Increasing sprite cache to 32 items (1792 bytes)
dbg: [sprite] Increasing sprite cache to 64 items (3584 bytes)
dbg: [sprite] Increasing sprite cache to 128 items (7168 bytes)
dbg: [sprite] Increasing sprite cache to 256 items (14336 bytes)
dbg: [sprite] Increasing sprite cache to 512 items (28672 bytes)
dbg: [sprite] Increasing sprite cache to 1024 items (57344 bytes)
dbg: [sprite] Increasing sprite cache to 2048 items (114688 bytes)
dbg: [sprite] Increasing sprite cache to 4096 items (229376 bytes)
dbg: [sprite] Increasing sprite cache to 8192 items (458752 bytes)
dbg: [sprite] Increasing sprite cache to 16384 items (917504 bytes)
dbg: [sprite] Increasing sprite cache to 32768 items (1835008 bytes)
dbg: [sprite] Increasing sprite cache to 65536 items (3670016 bytes)
dbg: [sprite] Increasing sprite cache to 131072 items (7340032 bytes)

JGRennison:
I've added a call to shrink_to_fit() after all GRFs are loaded.

With a non-STL container, allocation policy can be controlled arbitrarily, however the default exponential behaviour of std::vector generally gives a reasonable trade-off between reallocation costs and unused capacity. Previous PR reviews have seemed keen to prefer STL containers where at all possible.
The memory cost of the index vector is small compared to the actual sprite data.

src/fontcache/freetypefontcache.cpp
```diff
@@ -303,7 +303,8 @@ const Sprite *FreeTypeFontCache::InternalGetGlyph(GlyphID key, bool aa)
 	}
 
 	GlyphEntry new_glyph;
-	new_glyph.sprite = BlitterFactory::GetCurrentBlitter()->Encode(spritecollection, SimpleSpriteAlloc);
+	SimpleSpriteAllocator allocator;
+	new_glyph.sprite = BlitterFactory::GetCurrentBlitter()->Encode(spritecollection, allocator);
```
michicc: Depending on the colour of your bikeshed, you could also make this (and the other lines) `new_glyph.sprite = BlitterFactory::GetCurrentBlitter()->Encode(SimpleSpriteAllocator{}, allocator);`

PeterN: The `allocator` parameter would have to be `const` to allow this, and then the `Allocate()` method would also have to be `const`, which I'm not sure is desirable.

michicc: Hmm, yes. Or I think you could make it a `&&` :)

PeterN: Tricky given it's passed between functions, sometimes from a pointer (optional references aren't a thing afaik.)

src/spritecache.cpp
```diff
@@ -43,17 +43,8 @@ static inline SpriteCache *GetSpriteCache(uint index)
...
-		uint items = Align(index + 1, 1024);
-
-		Debug(sprite, 4, "Increasing sprite cache to {} items ({} bytes)", items, items * sizeof(*_spritecache));
-
-		_spritecache = ReallocT(_spritecache, items);
-
-		/* Reset the new items and update the count */
-		memset(_spritecache + _spritecache_items, 0, (items - _spritecache_items) * sizeof(*_spritecache));
-		_spritecache_items = items;
+	if ((size_t)index >= _spritecache.size()) {
```
TrueBrain: /me browses through PR

Finds this one place where there is still a C-cast

Takes a stick and points.

Look.

Here :P


(sorry :D)

src/spritecache_internal.h
```diff
@@ -21,22 +21,31 @@
...
-	bool warned;         ///< True iff the user has been warned about incorrect use of this sprite
-	byte control_flags;  ///< Control flags, see SpriteCacheCtrlFlags
+	std::vector<byte> data{};
+	size_t file_pos{};
+	SpriteFile *file{};    ///< The file the sprite in this entry can be found in.
+	uint32_t id{};
+	uint32_t lru{};
+	SpriteType type{};     ///< In some cases a single sprite is misused by two NewGRFs. Once as real sprite and once as recolour sprite. If the recolour sprite gets into the cache it might be drawn as real sprite which causes enormous trouble.
+	bool warned{};         ///< True iff the user has been warned about incorrect use of this sprite
+	byte control_flags{};  ///< Control flags, see SpriteCacheCtrlFlags
```
TrueBrain: I know this is fine and all, but it still looks really odd to me to see it like this. I find the explicit style of saying `= false` and `= 0` much more verbose what the intention was. For vectors, sure, we don't have a choice. But PODs .. it is not my preference. But I am also not going to not approve it because of it :)

PeterN:
Due to changes over time, this is now pared down a bit, sprite storage is now in a `std::unique_ptr`, but no changes are required to `_sprite_cache` itself any more.

It should be noted that when the sprite cache is at its limit, the new LRU eviction algortihm is superior to that in master. A game with `sprite_cache_size_px` of just 1 is actually playable, whereas master will basically freeze need to be killed.

rubidium42:
Won't this change just crash the game when the configured memory limit more than the actual memory limit? Instead of limiting itself to at most 66.7% of the available memory?
On the other hand... now it won't allocate a few hundred MBs of memory when it's not going to be used because you don't use extra zoom or high resolution sprites.

Though, if we were to implement checks and reduce the cache in case of the `new` for a sprite's allocation failure, would that solve the crashing due to lack of memory somewhere later? It might be more troublesome than a potential crash due to the user have too little memory to play the game.

So yeah, I've got mixed feelings about whether crashing is better and whether adding complicated magic to prevent crashing will just delay the inevitable by a few milliseconds until some other allocation pushes the game over its limit. I'm currently tending slightly towards crashing though.

rubidium42 (approved):

PeterN:
Any other thoughts on the memory implications? I'm not really sure what problems were solved by how it was done before vs what could happen with this change.

PR #11496, opened 2023-11-26
Change: Set height of larger heightmap/scenario worldgen buttons by text height.

PeterN:
## Motivation / Problem

The world-gen windows use larger text buttons at the bottom and specifically for the Generate button.

These sizes are specified in pixels and are scaled by the interface scale. This in general works fine, unless the specified font size is different from normal, which makes the size of the buttons disproportionate.

It also introduces arbitrary pixel numbers in the window layouts that don't necessarily match anything.

The height of most text buttons is determined by font size, so why not set the height of these larger text buttons by font size too?

image

## Description

Set height of larger heightmap/scenario worldgen buttons by `SetMinimalTextLines()` instead of `SetMinimalSize()`

This allows the text button height to scale properly by font size.

And it also makes a bit more semantic sense. "The button is 2 text lines heigh" vs "The button is max of 1 text line, or 20 pixels high".

image

## Checklist for review


2TallTyler (approved):
Makes sense to me.

PR #11499, opened 2023-11-26
Codechange: Use a few less NUM_CARGO loops.

PeterN:
## Motivation / Problem

Reducing the use of 0 - NUM_CARGO loops. for-loops that index into an array which can use range-for or std::max_element.

## Description

Replace some loops with range-for or std::max_element.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -1373,8 +1373,8 @@ static bool VehicleCargoSorter(const Vehicle * const &a, const Vehicle * const &
 	for (v = b; v != nullptr; v = v->Next()) diff[v->cargo_type] -= v->cargo_cap;
 
 	int r = 0;
-	for (CargoID i = 0; i < NUM_CARGO; i++) {
-		r = diff[i];
+	for (uint d : diff) {
+		r = d;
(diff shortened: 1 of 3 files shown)


src/vehicle_gui.cpp
```diff
@@ -1373,8 +1373,8 @@ static bool VehicleCargoSorter(const Vehicle * const &a, const Vehicle * const &
 	for (v = b; v != nullptr; v = v->Next()) diff[v->cargo_type] -= v->cargo_cap;
 
 	int r = 0;
-	for (CargoID i = 0; i < NUM_CARGO; i++) {
-		r = diff[i];
+	for (uint d : diff) {
+		r = d;
 		if (r != 0) break;
 	}
```
rubidium42, line 1379: Couldn't this be something along the line of the following?
```
auto it = std::find_if(diff.begin(), diff.end(), [](auto d) { return d != 0; });
return it != diff.end() ? *it < 0 : VehicleNumberSorter(a, b);
```
It's sort of a recurring pattern with the capacities as well, though there the code is smaller than it's here.

rubidium42 (approved):

PR #11505, opened 2023-11-27
Codechange: Template input buffer type in audio mixer.

PeterN:
## Motivation / Problem

The `mix_int16` and `mix_int8_to_int16` audio mixer functions are quite long and are very similar, save for a type difference and a shift factor.

## Description

Add the type and the shift factor as template parameters to `mix_int16` (int16 is the output type here which does not change) to reduce code duplication.

## Checklist for review

### DIFF
--- a/src/mixer.cpp
+++ b/src/mixer.cpp
@@ -63,13 +63,17 @@ static int RateConversion(T *b, int frac_pos)
 	return ((b[0] * ((1 << 16) - frac_pos)) + (b[1] * frac_pos)) >> 16;
 }
 
+template <typename T>
 static void mix_int16(MixerChannel *sc, int16_t *buffer, uint samples, uint8_t effect_vol)
 {
+	/* Shift required to get sample value into range for the data type. */
+	const uint SHIFT = sizeof(T) * CHAR_BIT;
+
 	if (samples > sc->samples_left) samples = sc->samples_left;
 	sc->samples_left -= samples;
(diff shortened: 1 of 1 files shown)


src/mixer.cpp
```diff
@@ -78,16 +79,16 @@ static void mix_int16(MixerChannel *sc, int16_t *buffer, uint samples, uint8_t e
 	if (frac_speed == 0x10000) {
 		/* Special case when frac_speed is 0x10000 */
 		do {
-			buffer[0] = Clamp(buffer[0] + (*b * volume_left  >> 16), -MAX_VOLUME, MAX_VOLUME);
-			buffer[1] = Clamp(buffer[1] + (*b * volume_right >> 16), -MAX_VOLUME, MAX_VOLUME);
+			buffer[0] = Clamp(buffer[0] + (*b * volume_left  >> TShift), -MAX_VOLUME, MAX_VOLUME);
```
rubidium42: Could even replace `TShift` with `std::numeric_limits<T>::digits` I think. Maybe just as the 'default' for `TShift` in the template, or not at all as it's more complicated.

rubidium42 (dismissed):

src/mixer.cpp
PeterN: That seems like a good idea. There's no reason for it to be configurable if it can be automatically determined from the type.

rubidium42 (approved):

PR #11511, opened 2023-11-28
Fix ab1a4c6c: Crash if the "No Music" set is loaded because there is no current set_index.

PeterN:
## Motivation / Problem

Crash if the "No Music" set is loaded because there is no current set_index.

`active_playlist` is empty in this case but I did not bounds check.

## Description

Add a helper function to get the current set_index which which checks it is in range.

## Checklist for review

### DIFF
--- a/src/music_gui.cpp
+++ b/src/music_gui.cpp
@@ -87,6 +87,7 @@ struct MusicSystem {
 	void PlaylistClear();
 
 private:
+	uint GetSetIndex();
 	void SetPositionBySetIndex(uint set_index);
 	void ChangePlaylistPosition(int ofs);
 	int playlist_position;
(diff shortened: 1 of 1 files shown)


src/music_gui.cpp
```diff
@@ -191,14 +192,25 @@ void MusicSystem::SetPositionBySetIndex(uint set_index)
...
 
+/**
+ * Get set index from current playlist position.
+ * @return current set index, or UINT_MAX if nothing is selected.
+ */
+uint MusicSystem::GetSetIndex()
+{
+	return static_cast<size_t>(this->playlist_position) < this->active_playlist.size()
+		? this->active_playlist[this->playlist_position].set_index
+		: UINT_MAX;
```
2TallTyler, line 203:
```suggestion
	if (static_cast<size_t>(this->playlist_position) < this->active_playlist.size()) return this->active_playlist[this->playlist_position].set_index;	
	
	return UINT_MAX;
```
Using a ternary here seems unnecessarily hard to read.

PeterN: Does it? I ... actually think your suggestion is harder to read.

2TallTyler: It doesn't really matter to me. 🤷

2TallTyler (approved):

PR #11517, opened 2023-11-29
Fix #11516: Changing interface scale makes resizeable windows larger but not smaller.

### LINKED ISSUE #11516: [Bug]: Changing interface size has inconsistent effects on window sizes
### Version of OpenTTD

Latest nightly, Windows 10 & 11

### Expected result

Open windows should scale along with interface size when the interface size setting is changed. Doubling interface size should double window sizes, halving it should halve window sizes. 

### Actual result

Windows with a fixed size, eg game options, behave as expected. Resizeable windows do not:
When scaling up the interface, a resizable window is only increased in size if it would be too small to fit all of the UI elements at the increased UI scale. Otherwise, the window size does not change.
When scaling down the interface, resizable windows stay the same size.

### Steps to reproduce

1. Start a new game with the interface at 1x size.
2. Open a few resizable windows, eg. A vehicle window, the train list window, etc., at their default size. 
3. Change the interface size to 2x, and these windows will increase in size.
4. Change the interface size to 1x, and these windows will stay the same size - which is now effectively giant as the UI elements have halved in size.

PeterN:
## Motivation / Problem

As per #11516, changing interface scale make resizeable windows larger, but not smaller. This is because rather than special handling on scale change, the window size is constrained by its minimum size which necessarily makes it larger.

## Description

Instead, explicitly scale window size by interface scale changes during window `ReInit()`. This requires the current scale to be stored and updated per window, as resizes can be nested and only the first pass should apply the scale change.

This also fixes a bug that `LoadStringWidthTable()` issues a `ReInit()` itself, so changing interface scale ended up doing two unnecessary passes. This was just wasteful before, but would now mean the scaling change is applied too early.

## Checklist for review

### DIFF
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -1391,8 +1391,6 @@ void LoadStringWidthTable(bool monospace)
 			_stringwidth_table[fs][i] = GetGlyphWidth(fs, i + 32);
 		}
 	}
-
-	ReInitAllWindows(false);
 }
 
 /**
--- a/src/settings_table.cpp
+++ b/src/settings_table.cpp
@@ -270,6 +270,7 @@ static void ZoomMinMaxChanged(int32_t)
 		_gui_zoom = _settings_client.gui.zoom_min;
 		UpdateCursorSize();
 		LoadStringWidthTable();
+		ReInitAllWindows(true);
 	}
 }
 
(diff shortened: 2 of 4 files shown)


PeterN:
Here is the behaviour in master:

[Screencast from 2023-12-06 21-48-51.webm](https://github.com/OpenTTD/OpenTTD/assets/639850/adc8afa3-2eb5-42a9-9d52-67fef03e972c)

Here is the behaviour with this PR applied:

[Screencast from 2023-12-06 21-50-32.webm](https://github.com/OpenTTD/OpenTTD/assets/639850/7bc8654b-a6ed-44a4-8ceb-319c7e3602d2)

2TallTyler (approved):

PR #11518, opened 2023-11-29
Feature: (-tte) Add zoom level buttons to sprite aligner.

PeterN:
## Motivation / Problem

Sprite Aligner is hard to use; it shows the sprite at the current GUI zoom level, and offsets are always adjusted and shown for the maximum zoom level.

## Description

Add extra zoom buttons to the sprite aligner window to allow selecting which sprite zoom level to display.

Offsets are scaled by the selected zoom level, and using the adjustment buttons takes this into account.

Additionally the selected sprite number is drawn in write in the sprite number list.

Lastly, the window is now resizeable.

image

## Checklist for review


src/widgets/newgrf_debug_widget.h
```diff
@@ -40,6 +40,7 @@ enum SpriteAlignerWidgets {
 	WID_SA_RESET_REL,   ///< Reset relative sprite offset
 	WID_SA_CENTRE,      ///< Toggle centre sprite.
 	WID_SA_CROSSHAIR,   ///< Toggle crosshair.
+	WID_SA_ZOOM,
```
rubidium42: Might be useful to mention that this takes `ZOOM_LVL_COUNT` entries, so when the new button gets added the 2x zoom button doesn't "click" the new button.
Maybe even codify it by not putting it at the end of the enumeration, and already adding the right offset for the `WID` after it.

src/newgrf_debug_gui.cpp
```diff
@@ -1025,6 +1032,13 @@ struct SpriteAlignerWindow : Window {
 				this->SetWidgetLoweredState(widget, SpriteAlignerWindow::crosshair);
 				this->SetDirty();
 				break;
+
+			default:
+				if (IsInsideBS(widget, WID_SA_ZOOM, ZOOM_LVL_END)) {
```
rubidium42, line 1042:
```suggestion
				if (IsInsideBS(widget, WID_SA_ZOOM, ZOOM_LVL_COUNT)) {
```
For some reason the enumeration having both end and count, seems to imply that begin might not be at 0. But then, the trick with adding the enumeration to the base widget might fail when begin is non-zero.

rubidium42 (approved):

PR #11519, opened 2023-11-29
Fix: Nightly build failures for macos and linux

glx22:
## Motivation / Problem
For a week now, macos and linux nightlies fail to build.
For both it happens during the "Install dependencies" step.
For macos the error is
```
  Error: You are using macOS 11.
  We (and Apple) do not provide support for this old version.
  It is expected behaviour that some formulae will fail to build in this old version.
  It is expected behaviour that Homebrew will be buggy and slow.
  Do not create any issues about this on Homebrew's GitHub repositories.
  Do not create any issues even if you think this message is unrelated.
  Any opened issues will be immediately closed without response.
  Do not ask for help from Homebrew or its maintainers on social media.
  You may ask for help in Homebrew's discussions but are unlikely to receive a response.
  Try to figure out the problem yourself and submit a fix as a pull request.
  We will review it but may or may not accept it.
```
For linux it's during vcpkg run (upstream may have changed stuff again)

## Description
For macos just switch runner from `macos-11` to `macos-12`.

For linux I reworked our vcpkg flow. We used to clone vcpkg repo into our container, and manually upgrade packages, as a way to always get up-to-date version of our dependencies, but that comes with the risk of having broken stuff from upstream. And the cache was not updated with the upgraded version, making it quite useless anyway.
So I decided to instead use vcpkg from the runner, it changes only when runner image is updated (like our other release workflows) so we don't need to manually update the packages as a new cache is created when a new image is released, and still get frequent enough upgrades of our dependencies.

## Checklist for review

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -17,6 +17,11 @@ jobs:
       # manylinux2014 is based on CentOS 7, but already has a lot of things
       # installed and preconfigured. It makes it easier to build OpenTTD.
       image: quay.io/pypa/manylinux2014_x86_64
+      volumes:
+      - /usr/local/share/vcpkg:/vcpkg
+      env:
+        ImageOS: ${{ env.ImageOS }}
+        ImageVersion: ${{ env.ImageVersion }}
 
     steps:
     - name: Download source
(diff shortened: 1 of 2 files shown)


LordAro (approved):
Seems reasonable

PR #11520, opened 2023-11-30
Remove: ZOOM_LVL_COUNT

rubidium42:
## Motivation / Problem

Looking at #11518 and falling in a rabbit hole.

`ZoomLevel` is the only enumeration with a COUNT and END. The logic of the COUNT implies that BEGIN could be non-zero, but all but two uses of zoom level assume that BEGIN is zero, making the separate count only confusing.

## Description

Replace `ZOOM_LVL_COUNT` with `ZOOM_LVL_END`, and remove the ` - ZOOM_LVL_BEGIN` from the two cases where it existed.

## Limitations

Changing `ZOOM_LVL_BEGIN` to be non-zero will be slightly more work, but having code that looks like it should work but doesn't is in my opinion worse than having code that doesn't imply that changing `ZOOM_LVL_BEGIN` is fine.

## Checklist for review


PeterN (approved):
Looks fine, I don't think it makes sense to ever start from non-zero.

PR #11526, opened 2023-11-30
Fix ddd609c: Reverse build probability callback was not called for articulated parts.

PeterN:
## Motivation / Problem

Reverse build probability callback is intended to work regardless of the restrictions on ctrl-click in the depot, but it turns out that reversing articulated parts was missing.

## Description

Also call the reverse build probability for articulated rail vehicles. Default probability is 0% so nothing changes if the NewGRF does not implement this.

## Checklist for review

### DIFF
--- a/src/articulated_vehicles.cpp
+++ b/src/articulated_vehicles.cpp
@@ -431,6 +431,7 @@ void AddArticulatedParts(Vehicle *first)
 
 		if (flip_image) v->spritenum++;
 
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
With articulated vehicles there's no way for the player to un-flip, but I guess that's between them and the GRF author.

PeterN:
Yes, manual flipping has a different set of constraints (we don't know what the author has done!) but as this controlled by the author anyway, that is not a concern. Disallowing the callback would be adding limits for limits sake.

PR #11533, opened 2023-12-03
Change: Hide unused cargos from vehicle cargo filter.

PeterN:
## Motivation / Problem

The vehicle list cargo filter shows all cargo types. When an industry set introduces a lot of cargo types, this be very long, and may contain cargo types that aren't relevant.

image

## Description

Instead, filter the cargo filter (hah) to only show the cargos carried by vehicles listed by the (pre-cargo-filtered) vehicle list.

The full list is used when determining widget size so that the allocated width is always sufficient.

image

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -191,6 +191,14 @@ void BaseVehicleListWindow::BuildVehicleList()
 
 	GenerateVehicleSortList(&this->vehicles, this->vli);
 
+	CargoTypes used = 0;
+	for (const Vehicle *v : this->vehicles) {
+		for (const Vehicle *u = v; u != nullptr; u = u->Next()) {
+			if (u->cargo_cap > 0) SetBit(used, u->cargo_type);
+		}
+	}
+	this->used_cargoes = used;
+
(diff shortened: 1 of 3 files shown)


src/vehicle_gui.cpp
```diff
@@ -371,7 +379,7 @@ StringID BaseVehicleListWindow::GetCargoFilterLabel(CargoID cid) const
 	}
 }
```
2TallTyler, line 381: Needs doxygen to explain what "full" means

2TallTyler (commented):
I like this, but I do wonder about the edge-case of vehicles that refit. There aren't currently any vehicles that carry it, but might be in the future...

The main issue that I see is a player opens the window, it builds the list of vehicles, and then the vehicles change — either by refit or the player builds a new vehicle. They open the dropdown to filter by this new cargo, but the dropdown has old data and doesn't show the new cargo type.

Maybe this is solved already by something invalidating data and rebuilding the GUI, but it's worth checking, IMHO.

2TallTyler (approved):

PR #11542, opened 2023-12-04
Change: Revert pressed-button offset shifting introduced in r2161.

PeterN:
## Motivation / Problem

Buttons shift their content when pressed by 1 (scaled) pixel both horizontally and vertically. This was added over 18 years ago in r2161 (https://github.com/OpenTTD/OpenTTD/commit/081ff2428202ae8c184e9cb25bafdb1bf6a214a6), but did not apply to all buttons. Since then we reworked the GUI system with nested widgets, automatically sized layouts and (usually) applying standardisation.

For a long time with 2x or 4x GUI, the shifting was still only 1 pixel. With the advent of non-integer interface scaling, the shift offset was scaled as well, and this meant that images would leak over buttons -- so button sizes were automatically increased to compensate.

Incidentally, the station, depot, road stop and object pickers never had this shifting applied. So it was inconsistent.

image

## Description

Remove the pressed-button offset completely. The first stage is very brief and simply sets the unscaled 'pressed' WidgetDimension to zero along with updating the imgbtn dimensions, which results in the desired effect of pressed buttons not being shifted.

The larger second commit removes all the code that actually applies the offset -- this is also done on the basis that if there is no offset being applied, pretending there is going to lead to mistakes.

This restores the original game behaviour for pressed buttons.

image

## Limitations

Haters gunna hate?

## Checklist for review

### DIFF
--- a/src/bridge_gui.cpp
+++ b/src/bridge_gui.cpp
@@ -234,7 +234,7 @@ public:
 				for (int i = this->vscroll->GetPosition(); this->vscroll->IsVisible(i) && i < (int)this->bridges.size(); i++) {
 					const BuildBridgeData &bridge_data = this->bridges.at(i);
 					const BridgeSpec *b = bridge_data.spec;
-					DrawSpriteIgnorePadding(b->sprite, b->pal, tr.WithWidth(this->icon_width, rtl), false, SA_HOR_CENTER | SA_BOTTOM);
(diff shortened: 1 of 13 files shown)


2TallTyler (approved):
I've always thought the offset looks strange. Between inconsistent use and dealing with the padding with GUI scaling, I think its time has come.

Plus, as you wrote:
>This restores the original game behaviour

PR #11543, opened 2023-12-04
Add: [CI] Test grf generation

glx22:
## Motivation / Problem
We never really validate the grf generation process.

## Description
Install grfcodec for one of the linux target so we can test grf generation.
Also check the tracked files are not modified by grf generation.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -93,6 +93,7 @@ jobs:
         - name: GCC - Dedicated
           compiler: gcc
           cxxcompiler: g++
+          libraries: grfcodec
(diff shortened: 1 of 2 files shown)


rubidium42:
Might it be useful to forcefully remove the generated files before building? So if something fails there is a definite difference in the files on disk, instead of the files not being written while there is a non-error exit code and thus things fail without being noticed.

glx22:
Yeah when failing `CreateGRF.cmake` should set exit code, but it's easy to remove the existing output just to make sure any failure is highly visible. We always overwrite the existing output at the end anyway, so removing it at start is not different.

rubidium42 (approved):

PR #11544, opened 2023-12-05
Fix #10452: Limit river generation max_search_nodes

### LINKED ISSUE #10452: [Bug]: Long stalls during river generation on heightmap
### Version of OpenTTD

13.0, Windows 11 Pro

### Expected result

Faster river generation

### Actual result

Slow river generation with stalls.

### Steps to reproduce

Use this image to generate heightmap and create a 4096x4096 map with many rivers:
Windowed (fullscreen) bug

Long stalls happen during river generation.

SamuXarick:
Prevent long stalls during river generation by capping the maximum search nodes to be the same as that defined in NPF.

## Motivation / Problem
See https://github.com/OpenTTD/OpenTTD/issues/10452#issuecomment-1432078250

Terraforming occurs during river generation which causes issues for the pathfinding. If the tile the pathfinder is working with is no longer flat, a long stall occurs due to the pathfinding doing an exhaustive search around that tile, trying to connect a river path to it and ultimately failing in the end, leaving a river disconnected somewhere along the way from the spring to the sea.

The stall can last up to 3-5 minutes until all paths are tried.
https://github.com/OpenTTD/OpenTTD/assets/43006711/cbb18259-51cc-4f18-9b4e-3248334f4f11

## Description
This PR fixes the stall, by preventing the pathfinder from initializing max_search_nodes as 0 (= infinite). Since AyStar is used, I created a constant with the default value equal to that of both NPF and YAPF and have it referred.

## Limitations
This does not fix the river becoming disconnected due to terraforming.

## Checklist for review

### DIFF
--- a/src/pathfinder/npf/aystar.cpp
+++ b/src/pathfinder/npf/aystar.cpp
@@ -301,4 +301,7 @@ void AyStar::Init(Hash_HashProc hash, uint num_buckets)
 	 *  When that one gets full it reserves another one, till this number
 	 *  That is why it can stay this high */
 	this->openlist_queue.Init(102400);
+
+	/* Set a reasonable default limit */
+	this->max_search_nodes = AYSTAR_DEF_MAX_SEARCH_NODES;
 }
--- a/src/pathfinder/npf/aystar.h
+++ b/src/pathfinder/npf/aystar.h
@@ -20,6 +20,8 @@
 #include "../../tile_type.h"
 #include "../../track_type.h"
 
+static const int AYSTAR_DEF_MAX_SEARCH_NODES = 10000; ///< Reference limit for #AyStar::max_search_nodes
+
 /** Return status of #AyStar methods. */
 enum AystarStatus {
 	AYSTAR_FOUND_END_NODE, ///< An end node was found.
--- a/src/pathfinder/pathfinder_type.h
+++ b/src/pathfinder/pathfinder_type.h
@@ -11,6 +11,7 @@
 #define PATHFINDER_TYPE_H
 
 #include "../tile_type.h"
+#include "npf/aystar.h"
 
 /** Length (penalty) of one tile with NPF */
 static const int NPF_TILE_LENGTH = 100;
--- a/src/table/settings/pathfinding_settings.ini
+++ b/src/table/settings/pathfinding_settings.ini
@@ -168,7 +168,7 @@ cat      = SC_EXPERT
 [SDT_VAR]
 var      = pf.npf.npf_max_search_nodes
 type     = SLE_UINT
-def      = 10000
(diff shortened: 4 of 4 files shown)


src/landscape.cpp
```diff
@@ -1308,6 +1308,7 @@ static void BuildRiver(TileIndex begin, TileIndex end, TileIndex spring, bool ma
 	finder.FoundEndNode = River_FoundEndNode;
 	finder.user_target = &end;
 	finder.user_data = &user_data;
+	finder.max_search_nodes = _settings_game.pf.npf.npf_max_search_nodes;
```
2TallTyler:
```suggestion
	finder.max_search_nodes = 10000;
```

2TallTyler (commented):
This seems reasonable.

I don't think it makes sense for players to be able to change this limit, so it could be hardcoded to 10,000 (the default for both NPF and YAPF).

src/landscape.cpp
PeterN: npf_max_search_nodes shouldn't ever be 0, so should it have been `_settings_newgame` instead of `_settings_game`?

While I agrree about configurability, slapping a 10000 in there seems wrong.

SamuXarick: `_settings_game` seems appropriate. Or at least, it's what's being used in the other functions related to river generation.

src/pathfinder/npf/aystar.h
```diff
@@ -20,6 +20,8 @@
 #include "../../tile_type.h"
 #include "../../track_type.h"
 
+static const int AYSTAR_MAX_SEARCH_NODES = 12312; ///< Reference limit for #AyStar::max_search_nodes
```
2TallTyler: Why not 10,000 like the setting default?

Also, this is not truly the max, it's the default setting for the max. Maybe a better name would be "AYSTAR_DEF_MAX_SEARCH_NODES".

src/pathfinder/npf/aystar.cpp
```diff
@@ -301,4 +301,9 @@ void AyStar::Init(Hash_HashProc hash, uint num_buckets)
 	 *  When that one gets full it reserves another one, till this number
 	 *  That is why it can stay this high */
 	this->openlist_queue.Init(102400);
+
+	/* Limit the number of nodes expanded by default
+	 *  This is a preventive method as this should never be left unattended
```
2TallTyler: What does this comment mean? The river pathfinder doesn't change it later, so it seems the comment is wrong? You could just write "Set a reasonable default limit."

src/pathfinder/npf/aystar.h
SamuXarick: oh, my bad... lol I missed that. Thx for pointing out.

2TallTyler (approved):

PR #11548, opened 2023-12-05
Fix #10511: Delay 'go to nearest depot' orders

### LINKED ISSUE #10511: [Bug]: Go to nearest road vehicle depot constantly running the pathfinder
### Version of OpenTTD

20230224-master-gc73b88ddca

### Expected result

I'm not sure yet, but maybe give up calling the pathfinder?

### Actual result

Road vehicle has only 1 order, and it's "go to nearest depot", but there is no depot, only a giant road network. It's constantly calling the pathfinder.

### Steps to reproduce

[go to nearest road vehicle depot.zip](https://github.com/OpenTTD/OpenTTD/files/10832374/go.to.nearest.road.vehicle.depot.zip)

SamuXarick:
## Motivation / Problem
https://github.com/OpenTTD/OpenTTD/issues/10511 

A lone Road Vehicle with a "go to nearest depot" order on a large road network and no depot nearby causes constant calls to the pathfinder every tick, hindering game performance.

This is happening in `ProcessOrders` and not in `IndividualRoadVehicleController` as it's been in the previous reported cases with similar symptoms. I think that the ideal solution here would be to call the pathfinder upon reaching the end of a segment, as that would in theory limit the number of calls significantly. But in this case, it really wants to know what's the destination of the order before handling it over to the `IndividualRoadVehicleController`.

I don't know how to do that, so I thought of a different approach.

## Description
Delay the nearest depot order search for a day if the destination has been set to zero, which happens when it has already attempted to do so and failed to find a valid destination.

Closes #10511.

## Limitations
It may break other "legitimate" uses for `v->dest_tile == 0`

## Checklist for review

### DIFF
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -1961,6 +1961,10 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
 			}
 
 			if (v->current_order.GetDepotActionType() & ODATFB_NEAREST_DEPOT) {
+				/* If the vehicle can't find its destination, delay its next search.
+				 * In case many vehicles are in this state, use the vehicle index to spread out pathfinder calls. */
+				if (v->dest_tile == 0 && TimerGameCalendar::date_fract != (v->index % Ticks::DAY_TICKS)) break;
+
 				/* We need to search for the nearest depot (hangar). */
 				ClosestDepot closestDepot = v->FindClosestDepot();
 


glx22:
I like the idea, using same limitation as in CheckIfXXXNeedsService. But not a fan of the implementation.
For me it would make more sense to add a `max_penalty` parameter, and a virtual function to get the proper value. This virtual function can then also be used in CheckIfXXXNeedsService.

glx22 (dismissed):
Squash required, but final diff looks good to me.

SamuXarick:
Converted to draft. Currently, the solution is unviable.

The cost for a train reversing when departing from a dead-end station is YAPF_INFINITE_PENALTY = 100000 and the default value for penalty is 2000. Meaning that 'go to nearby train depot' would always fail, even with a depot just next door after the reverse.

https://github.com/OpenTTD/OpenTTD/assets/43006711/b6010516-156e-4e6a-918b-32e4c6009275

SamuXarick:
A better savegame with 74 vehicles going to nearest depot.
[go to nearest road vehicle depot x74.zip](https://github.com/OpenTTD/OpenTTD/files/13693960/go.to.nearest.road.vehicle.depot.x74.zip)

image

src/order_cmd.cpp
```diff
@@ -1961,34 +1961,37 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
...
-					/* If there is no depot in front, reverse automatically (trains only) */
-					if (v->type == VEH_TRAIN && closestDepot.reverse) Command<CMD_REVERSE_TRAIN_DIRECTION>::Do(DC_EXEC, v->index, false);
-
-					if (v->type == VEH_AIRCRAFT) {
-						Aircraft *a = Aircraft::From(v);
-						if (a->state == FLYING && a->targetairport != closestDepot.destination) {
-							/* The aircraft is now heading for a different hangar than the next in the orders */
-							AircraftNextAirportPos_and_Order(a);
+				/* Try to delay the search for a bit if the previous destination is invalid. */
+				if (v->dest_tile != 0 || TimerGameTick::counter % Ticks::DAY_TICKS == v->index % Ticks::DAY_TICKS) {
```
2TallTyler: ~~You could save some math with `TimerGameCalendar::date_fract == v->index`~~

glx22: Would still need a modulo for `v->index` I think.

2TallTyler: Oh, right. `TimerGameCalendar::date_fract == (v->index % Ticks::DAY_TICKS)` ?

2TallTyler:
@SamuXarick This looks fine to me, but can you test your demo savegame and verify it still works, and I haven't given you some wrong code? 😉

SamuXarick:
> @SamuXarick This looks fine to me, but can you test your demo savegame and verify it still works, and I haven't given you some wrong code? 😉

`TimerGameCalendar::date_fract` worked the same as if it was `TimerGameTick::counter % Ticks::DAY_TICKS`.
One starts from 0 and goes up to 73. On the 74th, it becomes 0 again. The values I'm interested are in the range 0-73.
The other starts from 0 and goes to infinity. Using % 74 assures I am also getting the desired values in the range 0-73.

If there's some other functionality I'm not aware that could skew this check, then I must be aware.

2TallTyler (dismissed):

src/order_cmd.cpp
```diff
@@ -1961,34 +1961,37 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
...
-					/* If there is no depot in front, reverse automatically (trains only) */
-					if (v->type == VEH_TRAIN && closestDepot.reverse) Command<CMD_REVERSE_TRAIN_DIRECTION>::Do(DC_EXEC, v->index, false);
-
-					if (v->type == VEH_AIRCRAFT) {
-						Aircraft *a = Aircraft::From(v);
-						if (a->state == FLYING && a->targetairport != closestDepot.destination) {
-							/* The aircraft is now heading for a different hangar than the next in the orders */
-							AircraftNextAirportPos_and_Order(a);
+				/* Try to delay the search for a bit if the previous destination is invalid. */
+				if (v->dest_tile != 0 || TimerGameCalendar::date_fract == (v->index % Ticks::DAY_TICKS)) {
```
PeterN: Please explain, as a comment, why `v->index` is involved in the timing decision.

SamuXarick: I was looking for a way to spread the calls to pathfinder evenly and not concentrate all of them in the same tick. `TimerGameCalendar::date_fract == (v->index % Ticks::DAY_TICKS)` seemed to be a good approach. As each vehicle has a unique index, there will only be a repeat every 74 vehicles in the pool reducing the chances of that happening.

src/order_cmd.cpp
```diff
@@ -1961,34 +1961,37 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
...
-
-					/* If there is no depot in front, reverse automatically (trains only) */
-					if (v->type == VEH_TRAIN && closestDepot.reverse) Command<CMD_REVERSE_TRAIN_DIRECTION>::Do(DC_EXEC, v->index, false);
-
-					if (v->type == VEH_AIRCRAFT) {
-						Aircraft *a = Aircraft::From(v);
-						if (a->state == FLYING && a->targetairport != closestDepot.destination) {
-							/* The aircraft is now heading for a different hangar than the next in the orders */
-							AircraftNextAirportPos_and_Order(a);
+				/* Try to delay the search for a bit if the previous destination is invalid. */
```
2TallTyler: A possible answer to @PeterN's request. 🙂 

```suggestion
				/* If the vehicle can't find its destination, delay its next search.
				 * In case many vehicles are in this state, use the vehicle index to spread out pathfinder calls.*/
```

src/order_cmd.cpp
```diff
@@ -1961,34 +1961,38 @@ bool UpdateOrderDest(Vehicle *v, const Order *order, int conditional_depth, bool
...
-					if (v->type == VEH_TRAIN && closestDepot.reverse) Command<CMD_REVERSE_TRAIN_DIRECTION>::Do(DC_EXEC, v->index, false);
-
-					if (v->type == VEH_AIRCRAFT) {
-						Aircraft *a = Aircraft::From(v);
-						if (a->state == FLYING && a->targetairport != closestDepot.destination) {
-							/* The aircraft is now heading for a different hangar than the next in the orders */
-							AircraftNextAirportPos_and_Order(a);
+				/* If the vehicle can't find its destination, delay its next search.
+				 * In case many vehicles are in this state, use the vehicle index to spread out pathfinder calls. */
+				if (v->dest_tile != 0 || TimerGameCalendar::date_fract == (v->index % Ticks::DAY_TICKS)) {
```
glx22:
```suggestion
				if (v->dest_tile == 0 && TimerGameCalendar::date_fract != (v->index % Ticks::DAY_TICKS)) break;
```
should work too without needing to touch indentation.
(Maybe `return false` instead of break, but I didn't analyse all the function)

glx22 (approved):

PeterN:
What happens to the road vehicle in this state? Does it just carry on driving randomly?

glx22:
Yes for at most a day, it's usually a user error if the vehicle can't determine what is the nearest depot, and the change is just so it has less impact on gameloop.
Before the change the search would happen every tick.

PR #11558, opened 2023-12-07
Fix #11551: Link graph job results were incorrectly applied

### LINKED ISSUE #11551: [Bug]: Cargo distribution is broken
### Version of OpenTTD

Nightly 20231204

### Expected result

Normal cargodist behaviour: when setting distribution mode to either symmetric or asymmetric, passengers will have a particular destination and route themselves through other stations if necessary to get to that destination.

Working in nightly 20231121.

working

### Actual result

Passengers behave as though distribution mode if set to manual: they don't care where they're going and will board the first vehicle (in my test case bus) that arrives at the station.

Seems to be broken from nightly 20231204 onwards.

broken

### Steps to reproduce

See images above - I just made a simple passenger loop with four bus stops and four buses. Each bus only serves two stops; each stop is only served by two buses. Overkill setup but it makes it clear whether cargodist is working or not.

JGRennison:
## Motivation / Problem

#11551 caused by #10314 (aab580e0a)

## Description

Edge flows were incorrectly restricted because the restricted instead of unrestricted last update date was checked.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/linkgraph/linkgraphjob.cpp
+++ b/src/linkgraph/linkgraphjob.cpp
@@ -138,7 +138,7 @@ LinkGraphJob::~LinkGraphJob()
 				 * from the new flows. This avoids flow cycles between old and
 				 * new flows. */
 				while (!erased.IsEmpty()) ge.flows.erase(erased.Pop());
-			} else if ((*lg)[node_id][dest_id].last_restricted_update == CalendarTime::INVALID_DATE) {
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
Good catch. Thanks.

glx22 (approved):

PeterN (approved):

PR #11564, opened 2023-12-08
Fix: Prevent overflow when calculating max town noise

SamuXarick:
## Motivation / Problem
- Max town noise could overflow if the population was high enough.

## Description
- Clamp max town noise to the minimum value between the calculated result and UINT16_MAX.

## Checklist for review


PeterN:
This is a perfect example of randomly moving things without looking at the overall logic flow.

And the motivation smells of X/Y problem.

PeterN:
Move ALL of the code from L2518-2535 to L2553, that way there is no need to have extra uninitialized variables in the outer scope, and all of the DC_EXEC guarded actions are done in the same block.

PeterN:
In fact, that won't work either, my mistake :dagger:

PeterN (approved):

PR #11565, opened 2023-12-09
Fix: AirportGetNearestTown incorrectly assumed first TileIterator result was origin.

PeterN:
## Motivation / Problem

AirportGetNearestTown uses the first result of TileIterator and assumes it is the origin tile. However not all TileIterators are equal, and some do not start at the top-corner, so the perimeter check was wrong.

AirportTileTableIterator iterates the tiles in the airport spec, which is made up from a list of offsets. These can start anywhere within the layout.

As the perimeter is used to limit the range of tiles that are checked, if it is wrong then some tiles will be missed, which can result in the wrong town being returned. This then leads on to station noise level being applied/removed from the wrong town.

## Description

As the caller already has the origin tile, pass that and use it to determine the perimeter instead.

This issue was found by asserting that TileX(cur_tile) and TileY(cur_tile) were within the perimeter bounds.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -2303,18 +2303,19 @@ uint8_t GetAirportNoiseLevelForDistance(const AirportSpec *as, uint distance)
  * Finds the town nearest to given airport. Based on minimal manhattan distance to any airport's tile.
  * If two towns have the same distance, town with lower index is returned.
  * @param as airport's description
+ * @param tile origin tile (top corner of the airport)
  * @param it An iterator over all airport tiles
  * @param[out] mindist Minimum distance to town
  * @return nearest town to airport
  */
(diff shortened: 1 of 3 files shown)


src/station_cmd.cpp
```diff
@@ -2303,25 +2303,28 @@ uint8_t GetAirportNoiseLevelForDistance(const AirportSpec *as, uint distance)
  * Finds the town nearest to given airport. Based on minimal manhattan distance to any airport's tile.
  * If two towns have the same distance, town with lower index is returned.
  * @param as airport's description
+ * @param tile origin tile (top-corner)
```
TrueBrain: Nitpick, just for us in 10 years, mind adding "top corner" of what? :)

TrueBrain (approved):

PR #11568, opened 2023-12-09
Codechange: Add 'changed' return value to SetDisplayedPlane, and simplify callers.

PeterN:
## Motivation / Problem

* Repetitive code to get a widget, get its current plane, then change the plane if it's changed, and record that it's changed.
* Completing widget initialization, and then calling SetDisplayedPlane and forcing a ReInit again in the constructor.

## Description

* Make SetDisplayedPlane return whether the plane changed or not. This means callers don't need to refer to the widget more than once not remember the old value.
* Call SetDisplayPlace between 2-part widget initialization instead.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -2259,54 +2259,24 @@ struct CompanyWindow : Window
 			bool reinit = false;
 
 			/* Button bar selection. */
-			int plane = local ? 0 : SZSP_NONE;
-			NWidgetStacked *wi = this->GetWidget<NWidgetStacked>(WID_C_SELECT_BUTTONS);
-			if (plane != wi->shown_plane) {
-				wi->SetDisplayedPlane(plane);
-				this->InvalidateData();
-				reinit = true;
-			}
+			reinit |= this->GetWidget<NWidgetStacked>(WID_C_SELECT_BUTTONS)->SetDisplayedPlane(local ? 0 : SZSP_NONE);
 
 			/* Build HQ button handling. */
-			plane = (local && c->location_of_HQ == INVALID_TILE) ? CWP_VB_BUILD : CWP_VB_VIEW;
-			wi = this->GetWidget<NWidgetStacked>(WID_C_SELECT_VIEW_BUILD_HQ);
-			if (plane != wi->shown_plane) {
-				wi->SetDisplayedPlane(plane);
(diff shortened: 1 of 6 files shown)


LordAro:
I like the simplification, don't like the use of `|=` to avoid shortcircuit (I assume it's intentional) . How about `reinit = ... || reinit;` ?

PeterN:
> I like the simplification, don't like the use of `|=` to avoid shortcircuit (I assume it's intentional) . How about `reinit = ... || reinit;` ?

Confused, `|=` is a pattern used all over the place...

src/widget.cpp
```diff
@@ -1492,10 +1492,13 @@ NWidgetCore *NWidgetStacked::GetWidgetFromPos(int x, int y)
 /**
  * Select which plane to show (for #NWID_SELECTION only).
  * @param plane Plane number to display.
+ * @return true iif the shown plane changed.
```
rubidium42: `iif` -> `iff` I reckon.

src/company_gui.cpp
```diff
@@ -2259,54 +2259,24 @@ struct CompanyWindow : Window
...
-			plane = ((!_networking) ? (int)SZSP_NONE : (int)(local ? CWP_MP_C_PWD : CWP_MP_C_JOIN));
-			wi = this->GetWidget<NWidgetStacked>(WID_C_SELECT_MULTIPLAYER);
-			if (plane != wi->shown_plane) {
-				wi->SetDisplayedPlane(plane);
-				reinit = true;
-			}
-			this->SetWidgetDisabledState(WID_C_COMPANY_JOIN,   c->is_ai);
+			reinit |= this->GetWidget<NWidgetStacked>(WID_C_SELECT_MULTIPLAYER)->SetDisplayedPlane((!_networking) ? (int)SZSP_NONE : (int)(local ? CWP_MP_C_PWD : CWP_MP_C_JOIN));
+
+			this->SetWidgetDisabledState(WID_C_COMPANY_JOIN,  c->is_ai);
```
rubidium42: You removed one of the three spaces; is two spaces intentional?

PeterN: Intentional, but... I should have remove all but one.

rubidium42 (approved):

PR #11572, opened 2023-12-10
Change: Hide bevel for resizeable sparse layout windows.

PeterN:
## Motivation / Problem

On resizeable sparse layout windows, the resize box sticks out and looks strange, and is too close to the nearest button.

image

## Description

Hide the resizebox's bevel for sparse and graph windows. This avoids the clash, but keeps the general layout of the sparse window. When clicked, the button is still highlighted to show that it is active.

The bevel is controlled with widget_data; default value 0 draws the bevel, 1 hides it.

image

## Checklist for review


2TallTyler (commented):
I like the look of this very much.

But I'm concerned about the magic 0s and 1s that define behavior. It won't be obvious when reading `*_gui.cpp` files how we're hiding the bevel.

Could this use enum flags instead? Something like `WIDGET_SHOW_BEVEL = 0` as the default value, and `WIDGET_HIDE_BEVEL = 1` every time we want to hide it?

2TallTyler (approved):

PR #11574, opened 2023-12-11
Codechange: Use town zone constants instead of magic numbers

SamuXarick:
## Motivation / Problem
Numbers are being used for town zones, but they already have names associated.

## Description
Use town zone constants instead of magic numbers.

## Checklist for review

### DIFF
--- a/src/road_cmd.cpp
+++ b/src/road_cmd.cpp
@@ -1937,6 +1937,8 @@ static const Roadside _town_road_types[][2] = {
 	{ ROADSIDE_STREET_LIGHTS, ROADSIDE_PAVED }
 };
 
+static_assert(lengthof(_town_road_types) == HZB_END);
+
 static const Roadside _town_road_types_2[][2] = {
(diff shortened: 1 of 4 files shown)


src/road_cmd.cpp
```diff
@@ -1929,15 +1929,15 @@ static Foundation GetFoundation_Road(TileIndex tile, Slope tileh)
 	}
 }
 
-static const Roadside _town_road_types[][2] = {
+static const Roadside _town_road_types[HZB_END][2] = {
```
rubidium42: Not necessarily bad, but there's a tiny caveat in the specifications IIRC. If we make `HZB_END` bigger, it will just fill the remainder with default initialized values. So, maybe leaving `HZB_END` out of this, but adding `static_assert(lengthof(_town_road_types) == HZB_END);` after the table would be safer?
For now I would not immediately update the PR, but rather await some consensus on this.

2TallTyler: ^^ I agree with this.

2TallTyler (approved):

PR #11583, opened 2023-12-14
Codechange: Use MakeParameters with GetNextParameter in FormatString instead of StringParameters subspan

JGRennison:
## Motivation / Problem

This is primarily to avoid GS strings via SCC_ENCODED being able to trigger the span `offset + count <= size()` bounds assertion via StringParameter's `StringParameters(StringParameters &parent, size_t size)` constructor when args is too short.
GetNextParameter handles this gracefully.

These would otherwise each need a `game_script && size > args.GetDataLeft()` type guard, as used in SCC_STRINGn.

An example way to trigger this is to use the GS: "Neighbours are important" v16, and look in a town window.

## Description

Use MakeParameters with GetNextParameter instead of unguarded use of StringParameters subspan constructor.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -1231,7 +1231,7 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 					}
 
 					default: {
-						StringParameters tmp_params(args, 1);
+						auto tmp_params = MakeParameters(args.GetNextParameter<int64_t>());
 						GetStringWithArgs(builder, cargo_str, tmp_params);
 						break;
 					}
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):

src/strings.cpp
```diff
@@ -1405,7 +1405,7 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
 			case SCC_DEPOT_NAME: { // {DEPOT}
 				VehicleType vt = args.GetNextParameter<VehicleType>();
 				if (vt == VEH_AIRCRAFT) {
-					StringParameters tmp_params = StringParameters(args, 1);
+					auto tmp_params = MakeParameters(args.GetNextParameter<StationID>());
```
2TallTyler, line 1408: `StationID` and not `DepotID`?

JGRennison: VEH_AIRCRAFT means airports/airport hangars, which are referenced by StationID, see the definition of STR_FORMAT_DEPOT_NAME_AIRCRAFT.

2TallTyler: Ah, right, thanks.

PR #11589, opened 2023-12-15
Fix: Crash if squirrel compatibility scripts cannot be parsed.

PeterN:
## Motivation / Problem

If squirrel API initialization fails in RegisterAPI(), e.g. to due compatibility script initialization, then OpenTTD will crash as it has no handling for this case.

## Description

Fortunately RegisterAPI() cleans up after itself and deletes the engine. We just need to do is exit early instead of assuming the engine is ready.

## Checklist for review


glx22 (dismissed):
The test could also be `if(this->IsDead())` but checking `this->engine` works too.

glx22 (approved):

PR #11593, opened 2023-12-16
Feature: Include scalable OpenTTD TrueType fonts by default.

michicc:
## Motivation / Problem

Thanks to the recently created [custom-made TTF fonts by @zephyris](https://github.com/zephyris/openttd-ttf), it might be time to take the next step in the GUI refresh, especially with the variable GUI zoom.

## Description

With this PR, the TTF fonts will be loaded by default if no user-configured font is present and the new option to prefer the sprite font is off. Additional, there's a now option to enable font AA for all font sizes at once. The AA option defaults to on, the prefer sprite to off.

~~As @zephyris hasn't made a monospace font so far, I've included the Apache licensed Droid Sans Mono font for this.~~

The code was prepared to render our icons from the private-use area from a TTF, ~~but as of right now the fonts do not include them.~~

## Limitations

Quite a lot of stuff to figure out, mostly how best to include the font files and how to apply updates to them. I'm also not sure if the location I chose is really the proper place for them.

## Checklist for review


WenSimEHRP:
The fonts mentioned in this PR currently only contains latin charactors. I'm afraid that for players not using latin languages, (for example, CJK) forcing the game to use these fonts might cause a display problem. If I remember correctly, a new installed openttd game uses the font from the os, and it usually contains full support for other charactors.

Another possibility I assumed is that the game will use the defult font from the system if the font doesn't support the charactors the language uses, but I'm not sure about that.

michicc:
This PR does not change the font search behaviour of OTTD, if the currently active font can't display the language, it will search for a better font as before.

src/fontcache/freetypefontcache.cpp
```diff
@@ -157,40 +193,39 @@ void LoadFreeTypeFont(FontSize fs)
...
-					found = charmap;
-				}
-			}
-
-			if (found != nullptr) {
-				error = FT_Set_Charmap(face, found);
-				if (error == FT_Err_Ok) goto found_face;
-			}
+		if (!LoadFont(fs, face, font_name, settings->size)) {
+			ShowInfo("Unable to use '{}' for {} font, FreeType reported error 0x{:X}, using sprite font instead", font_name, FontSizeToName(fs), error);
```
glx22, line 198: `error` is always `FT_Err_Ok` here

src/os/windows/font_win32.cpp
```diff
@@ -300,7 +296,8 @@ void Win32FontCache::ClearFontCache()
 	WORD glyphs[2] = { 0, 0 };
 	GetGlyphIndicesW(this->dc, chars, key >= 0x010000U ? 2 : 1, glyphs, GGI_MARK_NONEXISTING_GLYPHS);
 
-	return glyphs[0] != 0xFFFF ? glyphs[0] : 0;
+	if (glyphs[0] != 0xFFFF) return glyphs[0];
+	return fallback && key >= SCC_SPRITE_START && key <= SCC_SPRITE_END ? this->parent->MapCharToGlyph(key)  : 0;
```
rubidium42:
```suggestion
	return fallback && key >= SCC_SPRITE_START && key <= SCC_SPRITE_END ? this->parent->MapCharToGlyph(key) : 0;
```

zephyris:
Overall, great. Originally, I wanted to make the fonts to simplify sprite font generation... But they work directly in OTTD better than I expected!

Packaging TTFs with OTTD. 
* Yes please! 
* Is there anything licensing I should think about? I've put the open font licence in the metadata for now, but can add an alternate.

TTF as default.
* At 1x zoom without anti aliasing the fonts should almost perfectly match the original graphics, so a decent substitution.
* When using a 1x zoom only base set, GUI can be used for nice pixel doubling scaling. TTFs as default would break that nice behaviour.
* Sprite based fonts allow per pixel tweaking. But, IIRC, TTFs can have embedded bitmap imprints to allow per pixel tweaking - that could be used for any final polish. 

Anti-aliased as default.
* Anti-aliased definitely works best for full flexibility of GUI scaling. With antialiasing the fonts work OK at fractional and integer scales.
* IMO it looks best not antialiased but only at an integer GUI scaling - but that is to match the GUI icons (which aren't antialiased) and will vary based on base set.

What are the private unicode ranges for OTTD? And is it just the bus, lorry, train, aeroplane and boat icons? Or anything else?

michicc:
> What are the private unicode ranges for OTTD? And is it just the bus, lorry, train, aeroplane and boat icons? Or anything else?

```
	SCC_LESS_THAN        = SCC_SPRITE_START + 0x3C,
	SCC_GREATER_THAN     = SCC_SPRITE_START + 0x3E,
	SCC_UP_ARROW         = SCC_SPRITE_START + 0xA0,
	SCC_DOWN_ARROW       = SCC_SPRITE_START + 0xAA,
	SCC_CHECKMARK        = SCC_SPRITE_START + 0xAC,
	SCC_CROSS            = SCC_SPRITE_START + 0xAD,
	SCC_RIGHT_ARROW      = SCC_SPRITE_START + 0xAF,
	SCC_TRAIN            = SCC_SPRITE_START + 0xB4,
	SCC_LORRY            = SCC_SPRITE_START + 0xB5,
	SCC_BUS              = SCC_SPRITE_START + 0xB6,
	SCC_PLANE            = SCC_SPRITE_START + 0xB7,
	SCC_SHIP             = SCC_SPRITE_START + 0xB8,
	SCC_SUPERSCRIPT_M1   = SCC_SPRITE_START + 0xB9,
	SCC_SMALL_UP_ARROW   = SCC_SPRITE_START + 0xBC,
	SCC_SMALL_DOWN_ARROW = SCC_SPRITE_START + 0xBD,
```

`SCC_SPRITE_START` is 0xE200.

Ufiby:
>What are the private unicode ranges for OTTD? And is it just the bus, lorry, train, aeroplane and boat icons? Or anything else? 

`HOUSE` HOUSE

michicc:
The simple default AA change is out, and the game options window now has two options for the sprite font and for a global font AA state.

This global AA state for all font sizes is to simplify UI and will enable AA even if the individual AA setting is off. Global AA on has precedence over size specific AA off.

zephyris:
Simplified AA settings is very reasonable.

Some thought should probably be put into fallback routes. Eg. If using sprite font and a character is missing then fall back to the TTF in the first instance, then a system default otherwise. Also, handle large/medium/small fallback separately. Might belong in a different PR.

PeterN:
I think the font still needs work -- it looks pretty bad at 1x interface scale, and by default shows up at 12px / 8px for medium / small instead of 10px / 6px, so it's not directly equivalent to the sprite font.

michicc:
The 12 pixel height is indirectly coming form the font. To explain:
Fonts can specify `Lowest_Rec_PPEM` in the `head` table, which is a suggestion for the minimal font size. OTTD takes this into account when no explicit size is given by taking the larger value of `default size for this type` and `Lowest_Rec_PPEM + diff between default size of this type and FS_SMALL`.
For the OpenTTD-Sans this value is set to 8. As FS_SMALL is 6 and FS_NORMAL 10, the value is 8 + 6 - 10 == 12 (so this takes precedence over 10).

To not have this, the new fonts need to declare 6 as the recommended minimum size.

michicc:
Bumped the font files to match release 0.3 and replaced the mono font with the one from zephyris.

For the small font, I've currently included the normal variant and not the all-caps version, but this is certainly up for debate.

src/fontcache.cpp
```diff
@@ -143,13 +144,52 @@ void SetFont(FontSize fontsize, const std::string &font, uint size, bool aa)
...
+			font_name = "OpenTTD-Small.ttf";
+			break;
+		case FS_LARGE:
+			font_name = "OpenTTD-Serif.ttf";
+			break;
+		case FS_MONO:
+			font_name = "OpenTTD-Mono.ttf";
+			break;
+
+		default: return;
```
TrueBrain: `return` or `NOT_REACHED()`?

src/fontcache.cpp
```diff
@@ -143,13 +144,52 @@ void SetFont(FontSize fontsize, const std::string &font, uint size, bool aa)
...
+	}
+
+	// Find font file.
```
TrueBrain: nitpick: old-style comments. Please use `/* */` :)

src/os/macosx/font_osx.cpp
```diff
@@ -298,6 +298,40 @@ const Sprite *CoreTextFontCache::InternalGetGlyph(GlyphID key, bool use_aa)
 	return new_glyph.sprite;
 }
 
+static CTFontDescriptorRef LoadFontFromFile(const std::string &font_name)
+{
+	if (!MacOSVersionIsAtLeast(10, 6, 0)) return nullptr;
```
TrueBrain, line 303: We haven't support 10.6 or older in years. Do we need these checks?

I think we currently aim 10.13.

src/settings_gui.cpp
```diff
@@ -50,6 +50,10 @@
 #include "safeguards.h"
 
 
+#if defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA)
+#define HAS_TRUETYPE_FONT
```
TrueBrain: nitpick: missing tab between `#` and `define`

src/os/macosx/string_osx.cpp
```diff
@@ -191,8 +194,10 @@ static CTRunDelegateCallbacks _sprite_font_callback = {
 
 		/* Install a size callback for our special sprite glyphs. */
 		for (ssize_t c = last; c < i.first; c++) {
-			if (buff[c] >= SCC_SPRITE_START && buff[c] <= SCC_SPRITE_END) {
+			if (buff[c] >= SCC_SPRITE_START && buff[c] <= SCC_SPRITE_END && i.second->fc->MapCharToGlyph(buff[c], false) == 0) {
```
TrueBrain, line 197: When reading the PR, both description and code, I cannot figure out why only MacOS needs the fallback set to false in one instance. I am not actually sure what problem that actually solves, and why only for MacOS. No comment also explains that to me in a way I understand.

Mind adding some snippets somewhere to make that a bit more clear, or tell me: look, it is here, read (which is also fine :D ).

michicc: The existing comment does already explain it, but not in a very obvious way. I can expand that.

I.e., we need to install the size callback for *sprite* glyphs, but not for font glyphs. And the font might or might not contain the special use chars.

TrueBrain: That still doesn't actually tell me why only MacOS needs this :) That is the part I am most confused about :)

michicc: Because the other OS's just work differently? Only CoreText has the concept of a custom size callback.

TrueBrain: Haha, that is fair :) I can accept that as answer :)

src/fontcache.h
```diff
@@ -108,9 +108,10 @@ class FontCache {
 	/**
 	 * Map a character into a glyph.
 	 * @param key The character.
+	 * @param fallback Allow fallback to the parent font.
 	 * @return The glyph ID used to draw the character.
 	 */
-	virtual GlyphID MapCharToGlyph(char32_t key) = 0;
+	virtual GlyphID MapCharToGlyph(char32_t key, [[maybe_unused]] bool fallback = true) = 0;
```
TrueBrain: Suggestion: `fallback` -> `allow_fallback`

src/os/macosx/font_osx.cpp
michicc: Well, I just kept the test that was already in `LoadCoreTextFont`.

src/fontcache/freetypefontcache.cpp
```diff
@@ -41,7 +41,7 @@ class FreeTypeFontCache : public TrueTypeFontCache {
 	FreeTypeFontCache(FontSize fs, FT_Face face, int pixels);
 	~FreeTypeFontCache();
 	void ClearFontCache() override;
-	GlyphID MapCharToGlyph(char32_t key) override;
+	GlyphID MapCharToGlyph(char32_t key, [[maybe_unused]] bool allow_fallback = true) override;
```
TrueBrain: nitpick: is this `[[maybe_unused]]` really needed here?

Same with the one in fontcache.h

TrueBrain (dismissed):
I cannot think of a reason not to approve this, although I don't really understand the context of this PR. But you tested it, and you have enough experience with this to know what is right and wrong.

As for the location of the files, it is where I would have put them too. So that all seems fine.

I am sure we will get font updates over the next few weeks / months / years; we can just update them as they come in.

Tnx a lot (both to @zephyris and @michicc) for this work; most excellent!

TrueBrain (approved):

PR #11600, opened 2023-12-18
Change: Scale sprites to requested highest resolution level.

PeterN:
## Motivation / Problem

The setting "Highest resolution sprites to use" is ignored if a graphics set only has high resolution sprites. The high resolution sprites are loaded anyway and used as-is.

image

## Description

Sprites from graphics sets which only provide high resolution sprites are now scaled up from scaled down versions, replacing the high resolution versions.

image

## Limitations

Graphics sets that use "odd" offsets may end up with shifting sprites. This also happens when they're viewed at 2x or 4x zoomed out levels, but because they're zoomed out it's less noticeable.

Not sure if this is a rounding issue with offsets, or the sets are "doing it wrong."

## Checklist for review

### DIFF
--- a/src/spritecache.cpp
+++ b/src/spritecache.cpp
@@ -397,6 +397,12 @@ static bool ResizeSprites(SpriteLoader::SpriteCollection &sprite, uint8_t sprite
 		if (!HasBit(sprite_avail, zoom)) ResizeSpriteOut(sprite, zoom);
 	}
 
+	/* Upscale to desired sprite_min_zoom if provided sprite only had zoomed in versions. */
+	if (first_avail < _settings_client.gui.sprite_zoom_min) {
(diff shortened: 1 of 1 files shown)


PeterN:
Example of weirdness due to offsets scaling. These images are taken at the exact same viewport scroll position yet they appear shifted, and "gridlines" appear. Offsets are -127px instead of -124px,

zBase with 4x sprites:
image
zBase with 1x sprites:
image

2TallTyler:
If there are no grid lines drawn, offsetting the sprites such that it appears this way is probably a problem. It only looks fine because we're used to grid lines...otherwise that kind of misalignment would look very wrong indeed.

PeterN:
As far as I know this misalignment is caused by the graphics set itself. 

This is a regular 4x zoomed out (commonly called 1x zoom) shot, also showing the grid lines that shouldn't be there;

image

2TallTyler (approved):

PR #11603, opened 2023-12-18
Add: AI/GS Time Mode to choose between economy (default) and calendar time

2TallTyler:
## Motivation / Problem

#10700 splits off economy and calendar time. #11341 makes these time units not follow each other, and #11428 allows them to tick at different rates.

It's time we break the news to Game Scripts and AI.

## Description

The PR adds a Time Mode for scripts (both AI and GS) to operate in, which modifies what ScriptDate returns. This includes calls from other script methods (like ScriptSubsidy) to ScriptDate.

The default time mode is economy time. Existing scripts need a default, and the most crucial information for GS and AI is knowing how time in the game is progressing to grow towns, build new train lines, give the player new goals, etc. Since calendar time can be frozen in #11428, that's unsuitable as a way of scripts to keep track of time. More about this in Limitations.

(I previously opened #11588 and ran into issues with compat files, so @rubidium42 and @glx22 suggested this route instead.)

## Limitations

- If a script looks for a specific calendar date rather than using dates only to measure the passing of time (I can't think of any examples, but a hypothetical example would be building industries based on real-world economic history) will behave strangely. They won't crash, but when OpenTTD is using real-time units, economy starts in year 1 and the script will react accordingly. I am not aware of any ways to split time without the potential of breaking scripts in this way. 😢 
- I don't have a GS to test calendar mode, so it is not tested. I welcome help with this.

## Checklist for review

### DIFF
--- a/src/script/api/script_timemode.cpp
+++ b/src/script/api/script_timemode.cpp
@@ -1,0 +1,29 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file script_timemode.cpp Implementation of ScriptTimeMode. */
+
+#include "../../stdafx.h"
+#include "script_timemode.hpp"
+
+#include "../../safeguards.h"
+
+ScriptTimeMode::ScriptTimeMode(bool calendar)
+{
+	this->last_time_mode = ScriptObject::IsCalendarTimeMode();
+	ScriptObject::SetTimeMode(calendar);
+}
+
+ScriptTimeMode::~ScriptTimeMode()
+{
+	ScriptObject::SetTimeMode(this->last_time_mode);
+}
+
+/* static */ bool ScriptTimeMode::IsCalendarMode()
+{
+	return ScriptObject::IsCalendarTimeMode();
+}
--- a/src/script/api/script_timemode.hpp
+++ b/src/script/api/script_timemode.hpp
@@ -1,0 +1,45 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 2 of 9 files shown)


src/script/api/script_timemode.cpp
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/** @file script_timemode.cpp Implementation of ScriptTimeMode. */
+
+#include "../../stdafx.h"
+#include "script_timemode.hpp"
+
+#include "../../safeguards.h"
+
+ScriptTimeMode::ScriptTimeMode(bool calendar)
```
glx22, line 15: Constructors for squirrel objects need a special handling, it should be like AsyncMode I think.

2TallTyler: It can't work like ScriptCompanyMode? That's what I copied for this application.

glx22: I tested and it works, now I'm wondering why AsyncMode uses the complex way.

src/script/api/script_timemode.hpp
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	~ScriptTimeMode();
+
+	/**
+	 * Set if the script is running in calendar time or economy time mode.
+	 * Calendar time is used by OpenTTD for technology like vehicle introductions and expiration, and variable snowline. It can be sped up or slowed down by the player.
+	 * Economy time always runs at the same pace and handles things like cargo production, everything related to money, etc.
+	 * @param Calendar Should we use calendar time mode? (Set to false for economy time mode.)
+	 */
+	static void SetTimeMode(bool calendar);
```
glx22: Mode stuff is usually scoped (and authors are used to this by now I think), this means to change mode you open a new scope and create the object.
So `SetTimeMode()` is not the way to go as it implies authors need to store the old state themselves.

2TallTyler: Ah, so like async mode after all?

glx22: Like any other mode actually.
To switch mode you create an instance of XXXMode class, and the previous mode is restored when instance is destroyed.
Exec and Test modes don't take parameters (the class name is explicit), while Company mode takes a company ID and Async takes the async state. But they all switch modes for as long the instance is in scope.
It's possible to do things like
```
/* do stuff as deity */
{
  local c0 = GSCompanyMode(0);
  /* do stuff as company 0 */
  {
    local c = GSCompanyMode(1);
    /* do stuff as company 1 */
  }
  /* do more stuff as company 0 */
  {
    local c = GSCompanyMode(2);
    /* do stuff as company 2 */
  }
}
/* do stuff as deity */
```
It will be the same for TimeMode.

2TallTyler: Ah, I think I have that already. I just need to delete ScriptTimeMode::SetTimeMode() and ScriptTimeMode::IsCalendarTime() so script authors can use the object constructor/destructor instead?

glx22: ScriptTimeMode::IsCalendarTime() will be needed for ScriptDate :)
(Similar to ScriptCompanyMode::IsValid() and ScriptCompanyMode::IsDeity())

2TallTyler: Okay, let me know if I got it right this time. 😉  Thanks for all your help with this.

glx22:
Code seems fine.
Don't forget to update ai_changelog.hpp and game_changelog.hpp.
Currently regression doesn't test `AIDate` at all, and I'm not sure it would see a difference if it tested both modes.

2TallTyler:
Changelogs updated. Let me know if there's more detail I should add.

Regression would have to use a game with wallclock timekeeping mode in order to see any difference between modes. That might be a separate PR (and it's well beyond my abilities, I suspect).

glx22 (approved):

Issue #11601, opened 2023-12-18
[Bug]: Compilation fails when DEBUG_DUMP_COMMANDS is defined


JGRennison:
### Version of OpenTTD

master

### Expected result

Compilation doesn't fail when DEBUG_DUMP_COMMANDS is defined

### Actual result

Compilation fails due to:
* Strong typedef/date changes (`Date` is no longer a type, and these don't work directly with sscanf anyway)
* CommandPacket no longer has a tile field

The use of DEBUG_DUMP_COMMANDS is mentioned in docs/desync.md and the top of src/network/network_func.h

### Steps to reproduce

Compile with DEBUG_DUMP_COMMANDS

JGRennison:
Also, the use of `std::from_chars` to parse the command data from the sscanf result appears to be incorrect?
It suggests that of each pair of hex characters being parsed into a byte, only the first is actually parsed.

PR #11607, opened 2023-12-19
Fix: Compilation with DEBUG_DUMP_COMMANDS enabled

LordAro:
## Motivation / Problem

#11601

## Description

Fix compile as appropriate -

* Fix from_chars to include both characters of the hex byte
* Fix next_date usage
* Stop trying to use CommandPacket::tile (just leave it as unused?)

## Limitations

I don't actually know what I'm doing here, but it does compile now. Probably needs @2TallTyler / @JGRennison and/or someone with actual commands.log experience to weigh in...

In particular, I feel like tile could probably be removed from the output entirely - feels like a bigger job though, idk

## Checklist for review

### DIFF
--- a/src/network/network.cpp
+++ b/src/network/network.cpp
@@ -1055,21 +1055,21 @@ void NetworkGameLoop()
 #ifdef DEBUG_DUMP_COMMANDS
 		/* Loading of the debug commands from -ddesync>=1 */
 		static FILE *f = FioFOpenFile("commands.log", "rb", SAVE_DIR);
-		static Date next_date = 0;
+		static TimerGameCalendar::Date next_date(0);
 		static uint32_t next_date_fract;
 		static CommandPacket *cp = nullptr;
 		static bool check_sync_state = false;
 		static uint32_t sync_state[2];
 		if (f == nullptr && next_date == 0) {
 			Debug(desync, 0, "Cannot open commands.log");
-			next_date = 1;
+			next_date = TimerGameCalendar::Date(1);
 		}
 
 		while (f != nullptr && !feof(f)) {
 			if (TimerGameCalendar::date == next_date && TimerGameCalendar::date_fract == next_date_fract) {
 				if (cp != nullptr) {
 					NetworkSendCommand(cp->cmd, cp->err_msg, nullptr, cp->company, cp->data);
-					Debug(desync, 0, "Injecting: {:08x}; {:02x}; {:02x}; {:08x}; {:06x}; {} ({})", TimerGameCalendar::date, TimerGameCalendar::date_fract, (int)_current_company, cp->cmd, cp->tile, FormatArrayAsHex(cp->data), GetCommandName(cp->cmd));
+					Debug(desync, 0, "Injecting: {:08x}; {:02x}; {:02x}; {:08x}; {} ({})", TimerGameCalendar::date, TimerGameCalendar::date_fract, (int)_current_company, cp->cmd, FormatArrayAsHex(cp->data), GetCommandName(cp->cmd));
 					delete cp;
 					cp = nullptr;
 				}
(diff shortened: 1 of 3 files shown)


src/network/network.cpp
```diff
@@ -1109,8 +1109,11 @@ void NetworkGameLoop()
 				int company;
 				uint cmd;
 				char buffer[256];
-				int ret = sscanf(p, "%x; %x; %x; %x; %x; %x; %255s", &next_date, &next_date_fract, &company, &cmd, &cp->err_msg, &cp->tile, buffer);
+				uint32_t next_date_raw;
```
2TallTyler, line 1112: TimerGameCalendar::Date is an int32_t. Same in the two `else if` blocks below.

JGRennison:
When building locally it still complains about about the TileIndex, however this does not prevent the sscanf from working correctly.

```
/home/jgr/openttd/trunk/src/network/network.cpp: In function ‘void NetworkGameLoop()’:
/home/jgr/openttd/trunk/src/network/network.cpp:1114:75: warning: format ‘%x’ expects argument of type ‘unsigned int*’, but argument 8 has type ‘TileIndex*’ {aka ‘StrongType::Typedef<unsigned int, TileIndexTag, StrongType::Compare, StrongType::Integer, StrongType::Compatible<int>, StrongType::Compatible<long int> >*’} [-Wformat=]
 1114 |                                 int ret = sscanf(p, "%x; %x; %x; %x; %x; %x; %255s", &next_date_raw, &next_date_fract, &company, &cmd, &cp->err_msg, &tile, buffer);
      |                                                                          ~^                                                                          ~~~~~
      |                                                                           |                                                                          |
      |                                                                           unsigned int*                                                              TileIndex* {aka StrongType::Typedef<unsigned int, TileIndexTag, StrongType::Compare, StrongType::Integer, StrongType::Compatible<int>, StrongType::Compatible<long int> >*}

```

The dummy TileIndex can be omitted entirely by changing the sscanf string to use %*x instead of %x, as in:

```
int ret = sscanf(p, "%x; %x; %x; %x; %x; %*x; %255s", &next_date_raw, &next_date_fract, &company, &cmd, &cp->err_msg, buffer);
assert(ret == 6);
```

If the the dummy TileIndex is left in, the assert needs to be changed to `assert(ret == 7);`

Otherwise, it seems to work correctly when loading a commands.log file.

LordAro:
Strange, didn't get a warning here. Ah well

Thoughts on just removing tile from the output entirely? If it's not being read in, is there any need for it to be there? I suppose it could help with manual debugging as it more clearly separates it out from the rest of the command parameters?

rubidium42:
> Thoughts on just removing tile from the output entirely? If it's not being read in, is there any need for it to be there? I suppose it could help with manual debugging as it more clearly separates it out from the rest of the command parameters?

Seems fine to me. It's in `LogCommandExecution` which is called in the same function that calls `::NetworkSendCommand`. `::NetworkSendCommand` does not get `tile` passed, so I'm fairly certain that `tile` is not needed in the desync logs.

src/network/network.cpp
rubidium42: There's no hexadecimal format for signed integers, so either the format needs to be changed to '%d' and the output of the debug commands needs to be changed to not be written in hexadecimal, or we accept the debug commands are written hexadecimal and read them as unsigned numbers.

LordAro:
I've gone with explicit int32_t casts for Date, and removed tile entirely from the desync debug output

rubidium42 (approved):

LordAro:
@JGRennison can you confirm this all works now?

JGRennison:
> @JGRennison can you confirm this all works now?

It seems to work fine here

PR #6998, opened 2018-12-29
Feature #4115: default company colour setting

GabdaZM:
Works only in single player.

PR #11612, opened 2023-12-21
Change: remove sorting weights when generating a company colour

EmperorJake:
## Motivation / Problem

When starting a new company, a company colour is selected at random. However, it isn't really random because the game prioritises certain colours over others. The result is that the first five companies that are founded will always end up with the same five colours (assuming they're not changed manually). 

image

This is because the game splits company colour options into 3 priorities and sorts them accordingly when generating a new company colour, and only selects lower priority colours once all the higher priority ones have been used up. The result is that when all the company slots are filled, the colours always end up in a similar order, with red, green, pink, orange, and blue at the top and brown, grey, and white near the middle or bottom.

This is an arbitrary choice which seems to be inherited from TTD. Company colour is a personal preference, and since https://github.com/OpenTTD/OpenTTD/pull/6998 players can select their preferred colours in the settings, giving certain colours priority becomes even less relevant.

## Description

This PR removes a few lines of code which prioritises some colours over others when generating a company colour. It does not affect the code which prevents similar company colours from appearing next to each other, in fact that code ostensibly becomes more useful. 

This means that when starting a new game, a company colour is selected at random from all 16 colour options instead of just five. All subsequent companies will also receive a random colour, but the game will still prevent similar colours from appearing next to each other.

image

## Limitations

Some players might prefer starting with a bright colour but don't care for a specific one, they might be annoyed that they could be given brown or grey instead.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -406,8 +406,6 @@ bad_town_name:;
 	}
 }
 
-/** Sorting weights for the company colours. */
-static const byte _colour_sort[COLOUR_END] = {2, 2, 3, 2, 3, 2, 3, 2, 3, 2, 2, 2, 3, 1, 1, 1};
 /** Similar colours, so we can try to prevent same coloured companies. */
 static const Colours _similar_colour[COLOUR_END][2] = {
 	{ COLOUR_BLUE,       COLOUR_LIGHT_BLUE }, // COLOUR_DARK_BLUE
@@ -443,15 +441,6 @@ static Colours GenerateCompanyColour()
 	for (uint i = 0; i < 100; i++) {
 		uint r = Random();
 		Swap(colours[GB(r, 0, 4)], colours[GB(r, 4, 4)]);
-	}
-
-	/* Bubble sort it according to the values in table 1 */
-	for (uint i = 0; i < COLOUR_END; i++) {
-		for (uint j = 1; j < COLOUR_END; j++) {
-			if (_colour_sort[colours[j - 1]] < _colour_sort[colours[j]]) {
-				Swap(colours[j - 1], colours[j]);
-			}
-		}
 	}
 
 	/* Move the colours that look similar to each company's colour to the side */


EmperorJake:
I left TTD running, it took a while for 4 AI companies to start but it indeed follows the same pattern 
image

PeterN:
I think this weighting is designed to choose colours that are most distinct from each other where possible.

EmperorJake:
That was probably more important back in the days of TTD and CRT monitors. If it's still considered important, perhaps the code for avoiding similar colours could be tweaked to be more strict and compensate? I did notice that brown and grey look similar for instance, but that combination can be seen in both the before and after screenshots.

2TallTyler:
I'm definitely getting Chesterton's Fence vibes here.

PeterN:
Once it gets beyond a certain point, they tend to start looking similar to existing colours. This is a downside of upping the limit from 7 to 15 companies.

CRT vs LCD does not have much bearing on colour fidelity (if anything early LCDs were worse.)

TrueBrain:
The motivation given in the PR is rather thin. I am not even sure it is an actual motivation, if I am being honest :) It reads like this PR mostly exists "because we can" rather than having a reason. So the following might miss an implied point, not written in this PR :)

Personally, it is a "no" for me on this PR.

The fact the game picks a colour that has a high contrast with each other for the first few companies makes total sense. And, maybe not less important, the colour also has a high contrast with other game elements, like the minimap, graphs, etc.
This means that if I play solo, I know I will have a colour that I can easily distinct on the minimap, and other game elements. As I don't actually care which colour my company has, this makes me rather happy.
I am always considering "new player" experience here, as any expert player can fix it one way or the other. If a company of a new player would get one of those colours you can't actually see on the minimap, that would degrade their experience.

Additionally, if another company starts, either because I am having a small multiplayer game or I added once again an AI to just be disappointment in a few gameyears, at least I know their colour has a high contrast. And as I can't change the colour of an AI, that is important to me. It is also the reason I never really like big multiplayer games, as I struggle seeing what company is actually making that crazy amount of profit, as I can't figure it out based on the colour. And I have to disable/enable companies via the Legend to find out. Not the best experience. Not at all.

Next, this code exists for a reason. Someone thought about this, years and years ago, and said: yeah, let's do high-contrast colours first, and fall back to less clear colours next, simply as our palette doesn't have any others. This is a limitation of the palette picked, and one we have to live with :( Just reversing that code without a really good argumentation feels like a step backwards.

Lastly, who does it actually hurt that colours are assigned like this? To use the argument made in the PR: anyone can just change their colour to any of the other ones anyway. So what is the harm in having this sorting order? This hooks back in that the motivation isn't all that clear. Sorry :)

It seems the PR is suggesting every game has 15 companies, and that this is the argument the sorting order makes no sense (anymore). But I would like to reject that statement. With 14.0 we can proof it, but I wanna bet most games (like: 95% of the games) never have more than a few companies. Most games are played offline, for example. So there might be an implied bias here for the motivation of this PR. But I am guessing here :)

Anyway, enough reasons for me to say "no".

That said, what I do think would actually be a good change, is if we find out a way to make more company colours with high contrast. Or fix the game elements that make it important to have high contrast colours as company colours. There are many places where a few of the current colours just look .... bad. If we would fix those places, and make all colours a high contrast to each other, that would be the moment I would say: yeah, time to drop this sorting order. But until then, in my opinion, there is a very good reason to keep things the way they are.


Sorry, many words, but maybe that sparks a discussion :)

EmperorJake:
I'm closing this PR as it's become clear, after constructive discussion, that this feature was intended and was coded by Sawyer himself into the game for a good reason. I will be pursuing other, better methods improving the company colour situation, such as a colour randomisation button and improvements to the NewCC GRF. Thanks for considering.

PR #11614, opened 2023-12-22
Fix #11442: "default" colour in group colour window is not updated when changing master colour

### LINKED ISSUE #11442: [Bug]: "default" colour in group colour window is not updated when changing master colour
### Version of OpenTTD

20231104-master

### Expected result

The default colour for a group of vehicles should change when the master colour [for that vehicle type] is changed.

### Actual result

The default colour in the livery window remains whatever the master colour was when the group was created and/or the group colour last changed to default. This error persists through opening and closing the window, and saving and loading the game.

Seeness Transport, 1976-10-18

### Steps to reproduce

1) create a vehicle group, and do not change its colour.
2) change the master colour.
3) note the colour in the vehicle group window remains the old master colour.

Darragh-Hazell:
## Motivation / Problem

As described in issue #11442. When setting the default company livery colour, previously created groups that use the default company livery will still display the default colour as being the one automatically set at the start of the game.

This is purely an interface issue which can prove distracting and/or confusing to the user due to the visible inconsistency it creates.

## Description

Under company_gui.cpp the DrawWidget method of SelectCompanyLiveryWIndow is responsible for rendering the default colour square preview in the groups window. 

The company default livery can be conditionally passed, to draw_livery, instead of the group livery. This is done with a constant boolean, livery_set, from HasBit on the group’s livery.in_use 0 bit. livery_set is also passed to the def parameter.

Furthermore, both branches of the if else statement of DrawWidget use the company constant, it was moved to be before the if statement so that both conditionals had it within their scope.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -924,9 +924,10 @@ public:
 			y += this->line_height;
 		};
 
+		const Company *c = Company::Get((CompanyID)this->window_number);
+
 		if (livery_class < LC_GROUP_RAIL) {
 			int pos = this->vscroll->GetPosition();
-			const Company *c = Company::Get((CompanyID)this->window_number);
 			for (LiveryScheme scheme = LS_DEFAULT; scheme < LS_END; scheme++) {
 				if (_livery_class[scheme] == this->livery_class && HasBit(_loaded_newgrf_features.used_liveries, scheme)) {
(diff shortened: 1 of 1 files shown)


PeterN (approved):
Looks good to me.

PR #11616, opened 2023-12-22
Change: Support custom transparent palette remaps.

PeterN:
## Motivation / Problem

When 32bpp was introduced, we left a note that custom transparency remaps are no longer supported, and left it with a TODO comment.

This is feature that relies on 8bpp palette remaps to change the colour of existing pixels. In the picture below, the station roofs should have a green tint, and when in 8bpp they do.

image

It was never done...

## Description

... until now.

This change implements a lookup table to convert 32bpp to 8bpp palette indexes. This can then have the palette remap applied, and be redrawn in the viewport. To reduce the size of the lookup table, some bits of the 32bpp value are dropped, but due to the nature of converting to an 8bpp palette, this is not really noticeable, especially given it's remapped anyway.

So finally this (limited use) feature can be supported with 32bpp blitters:

image

## Checklist for review

### DIFF
--- a/src/blitter/40bpp_anim.cpp
+++ b/src/blitter/40bpp_anim.cpp
@@ -11,6 +11,7 @@
 #include "../zoom_func.h"
 #include "../settings_type.h"
 #include "../video/video_driver.hpp"
+#include "../palette_func.h"
 #include "40bpp_anim.hpp"
 #include "common.hpp"
 
@@ -234,10 +235,6 @@ inline void Blitter_40bppAnim::Draw(const Blitter::BlitterParams *bp, ZoomLevel 
 					break;
 
 				case BM_TRANSPARENT:
-					/* TODO -- We make an assumption here that the remap in fact is transparency, not some colour.
-					 *  This is never a problem with the code we produce, but newgrfs can make it fail... or at least:
-					 *  we produce a result the newgrf maker didn't expect ;) */
-
 					/* Make the current colour a bit more black, so it looks like this image is transparent */
 					src_n += n;
 					if (src_px->a == 255) {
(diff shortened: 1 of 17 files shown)


src/palette.cpp
```diff
@@ -26,6 +27,112 @@ byte _colour_gradient[COLOUR_END][8];
 
 static std::recursive_mutex _palette_mutex; ///< To coordinate access to _cur_palette.
 
+/*
```
TrueBrain: Make it a doxygen comment?

src/palette.cpp
```diff
@@ -26,6 +27,112 @@ byte _colour_gradient[COLOUR_END][8];
...
+inline uint CrunchColour(uint c)
+{
+	return (c & PALETTE_BITS_MASK) | PALETTE_BITS_OR;
+}
+
+/**
+ * Calculate distance between two colours.
+ * @param col1 First colour.
+ * @param r2 Red component of second colour.
+ * @param g2 Gren component of second colour.
```
TrueBrain: Gren, the new colour. Please make it green :)

src/palette.cpp
```diff
@@ -26,6 +27,112 @@ byte _colour_gradient[COLOUR_END][8];
...
+ * @param b2 Blue component of second colour.
+ * @returns Euclidean distance between col1 and col2.
```
TrueBrain: nitpick: "between first and second colour"?

src/blitter/40bpp_anim.cpp
```diff
@@ -234,18 +235,14 @@ inline void Blitter_40bppAnim::Draw(const Blitter::BlitterParams *bp, ZoomLevel
...
 					/* Make the current colour a bit more black, so it looks like this image is transparent */
 					src_n += n;
 					if (src_px->a == 255) {
 						src_px += n;
 						do {
 							/* If the anim buffer contains a color value, the image composition will
-							 * only look at the RGB brightness value. As such, we can simply darken the
-							 * RGB value to darken the anim color. */
+							* only look at the RGB brightness value. As such, we can simply darken the
+							* RGB value to darken the anim color. */
```
TrueBrain: Your editor broke the sanity again! :P

TrueBrain (commented):
Only nitpicks; looks good to me!

src/blitter/40bpp_anim.cpp
PeterN: We need a commit-checker to test for "\t\* "...!

TrueBrain (approved):

ldpl:
What I don't like about this PR is that it's kind of a backwards approach to the problem. IMO, logical way would be to introduce 32bpp recolours (LUTs), then make a 8bpp recolor to LUT conversion for compatibility. What this essentially does is allowing 32bpp recolors but in a horrible form of palette mapping. 

Also, why is it only supporting transparent remaps and not regular remaps? Or is that part of some other PR?

P.S. Using floats in CalculateColourDistance probably hurts performance.

PeterN:
Go on then.

PeterN:
> Also, why is it only supporting transparent remaps and not regular remaps? Or is that part of some other PR?

Because regular remaps were implemented 17 years ago, and you've been using them all the time.

Issue #9784, opened 2022-01-09
The GUI for changing company color scheme is confusing


TinCanTech:
### Version of OpenTTD

12.1 .deb from official TTD site + Linux Ubuntu 20.04

### Expected result

I want to change company colour scheme during the game.

### Actual result

It is possible to select a colour sceme before starting a game via settings.
Once the game is started it is not possible to change colour scheme.

### Steps to reproduce

Start a new game in v12.1 and try to change company colour scheme.

James103:
Click on the little 'face' in the toolbar to open the company info window.
In the company info window, click "Colour Scheme".
Try changing the Standard Livery from there. Does it update? (It does for me)

The setting you were most likely trying to change in game is "Starting company color", but that only affects new games and when you start a new (human) company in an existing game.

TinCanTech:
Hi, I have tested all sorts of things from single player to LAN game.  Once the game starts it is not possible to select any other colour scheme.  It's like they do not exist in the save game or something.

TinCanTech:
Even starting a brand new game, once the game has started I can no longer change scheme.

LordAro:
As James says, it's not clear what you've tried and what isn't working. See the tutorial for how you change your company colour in-game:

https://wiki.openttd.org/en/Manual/Transport%20company + https://wiki.openttd.org/en/Archive/Manual/Company%20Colours

TinCanTech:
I have done the following:
1. Installed OpenTTD from official`.deb` version 12.1 for Ubuntu 20.04
2. I start a new single player game with no NewGFX loaded.
3. Once the game starts it is not possible to change colour scheme.

Although, I tried lots of other settings, to determine if I had done something wrong, with these 3 basic steps above, it is possible to reproduce the problem.

In the past, I used different colour schemes to label and rank different groups of vehicles.  For example, all my passenger vehicles were white scheme, while my Oil vehicles were Purple.

LordAro:
It is still not clear what you are doing. Which buttons are you pressing, what is not happening? Are you able to select different colour schemes but the vehicles aren't actually changing? Are you not able to select anything at all?

Vehicle groups can have their liveries changed via the group gui ( https://wiki.openttd.org/en/Manual/Vehicle%20groups#using-groups ) but I believe you need a NewGRF that supports that loaded before it is available (this is unchanged)

TinCanTech:
[...]

PR #11617, opened 2023-12-23
Change: Company Colour Scheme layout changes.

PeterN:
## Motivation / Problem

Feedback from Company Colour Schemes window.

* Position of colour selection dropdowns make them look like filters instead of actions.
* Tooltip of vehicle group buttons are the same as the vehicle livery scheme buttons.
* "Standard Livery" doesn't mean much.
* Blank groups are empty.
* The window is larger than necessary due to SetMinimalSize()

image
image

## Description

* Move dropdowns to bottom of window, to match most other windows that have actions at the bottom.
* Add per-vehicle type tooltips for the group colour buttons.
* Rename to "Default Livery".
* Show text if selected vehicle type group is empty.
* Remove SetMinimalSize() where the widget dimensions are already calculated from text size.

image
image

## Checklist for review


TrueBrain (dismissed):
"To show" gives me the suggestion they are there just not showing, but I fail to find better wording, so I guess it is also the best I can come up with :D

Awesome work btw, these improvements really help for a n00b like me :)

2TallTyler:
Closes #9784?

Also, not feedback that should block merging this -- but I wonder if a dropdown on each row would be more obvious than selecting a row and then changing it with the dropdown on the bottom of the window. Or maybe that would look too busy.

EmperorJake:
The "no groups to show" string gets cut off in certain cases 
image

PeterN:
image

Slight adjustment. Empty group text now ends with "are set up" rather than "to show", it's now black so that it's easier to read, and it spans the whole row instead of just the left column.

TrueBrain (approved):

Issue #953, opened 2007-06-27
Engine smoke


DorpsGek:
**_SirkoZ_ opened the ticket and wrote:**

> With the introduction of realistic acceleration, the locomotives (especially diesels) in OpenTTD don't smoke at all or very little - so I have written a patch a long ago that visually makes things much more appealing in that aspect. The diesel_smoke and electric sparks are dependent on train weight, speed and power, steam puffs though are fixed although all three types are configurable via "Configure Patches". 
> 
> Don't worry about possible problems due to the smoke causing problems with many vehicles' sprites on the map - as we all know - when the game lacks vehicle slots (and we can see this in as early as TT_Original) the first thing that goes are the smoke "vehicles" (sprites). The patch has also been tested on a few maps with up to 1000 trains (some of them OpenTTD coop's).
> 
> Version: all builds up to now, patch source - r10348.
> 
> File included.

**Attachments**
- [diesel_smoke_sz_2-0b_r10348_0.patch](https://bugs.openttd.org/task/953/getfile/1124/diesel_smoke_sz_2-0b_r10348_0.patch) (4.69 KiB)

**Reported version**: _0.5.2_
**Operating system**: _All_

---

DorpsGek:
**_Rubidium_ closed the ticket.**

**Reason for closing: _Duplicate_**

> 

---
###### _This comment was imported from FlySpray: https://bugs.openttd.org/task/953_

PR #11624, opened 2023-12-25
Change: set amount of smoke/sparks to "realistic" by default

EmperorJake:
## Motivation / Problem

The "realistic" smoke/sparks setting has been in the game for a very long time, since at least https://github.com/OpenTTD/OpenTTD/issues/953. The original motivation for creating this patch was that the original smoke didn't work well with realistic acceleration enabled, and this is by far the most noticeable with diesel engines. In fact there is so little diesel smoke with the original setting, that I have questioned whether the engines were coded correctly.

For a while now (can't find the exact commit) realistic acceleration has been the default, therefore it would make sense to set the amount of smoke/sparks to "realistic" by default as well. 

Furthermore, the smoke/sparks setting cannot be controlled client-side in multiplayer for technical reasons (to prevent desyncs), and server owners often forget to change this setting as it is a relatively unimportant one among many other important settings.

## Description

This PR changes the setting for vehicle smoke/sparks to "realistic" by default. This is better suited to the game's default setting of realistic acceleration, and it simply looks better in various ways. 

Steam: more smoke puffs while accelerating and while at speed
Diesel: generates smoke the whole time the engine is under load, instead of only a brief puff when accelerating
Electric: generates more sparks while the engine is under load

Also, it's hard to avoid the "r-word" when it's right there in the name of the setting ;)

## Limitations

It's not how TTD looked/behaved (but then neither is realistic acceleration)

More smoke/sparks sprites contributes towards the vehicle limit. Any performance impact should be pretty negligible though.

As mentioned before, the setting cannot be changed client-side in multiplayer for technical reasons.

## Checklist for review


2TallTyler (approved):
Since realistic acceleration is the default, we should use the compatible smoke/sparks to go with it.

In my mind, original acceleration and smoke/sparks exists for backwards-compatibility and people who want the game to look like TTD. Both fully valid, but these don't need to be the default for new players.

PeterN:
Is this just waiting for someone to object?

PR #11625, opened 2023-12-25
Codechange: Store and pass Colours in Colours type.

PeterN:
## Motivation / Problem

Values from the Colours enum are often stored as a byte and passed as an integer, and may require casts to be used as a Colours again. This is just messy.

## Description

Store and pass colours in Colours type instead. This reduces casts, some magic numbers, and introduces a bit of type-safety, as well as reducing ambiguity.

## Checklist for review

### DIFF
--- a/src/company_gui.cpp
+++ b/src/company_gui.cpp
@@ -582,7 +582,7 @@ static const LiveryClass _livery_class[LS_END] = {
 template <SpriteID TSprite = SPR_SQUARE>
 class DropDownListColourItem : public DropDownIcon<DropDownString<DropDownListItem>> {
 public:
-	DropDownListColourItem(int colour, bool masked) : DropDownIcon<DropDownString<DropDownListItem>>(TSprite, PALETTE_RECOLOUR_START + (colour % COLOUR_END), colour < COLOUR_END ? _colour_dropdown[colour] : STR_COLOUR_DEFAULT, colour, masked)
+	DropDownListColourItem(int colour, bool masked) : DropDownIcon<DropDownString<DropDownListItem>>(TSprite, GENERAL_SPRITE_COLOUR(colour % COLOUR_END), colour < COLOUR_END ? _colour_dropdown[colour] : STR_COLOUR_DEFAULT, colour, masked)
 	{
 	}
 };
@@ -647,7 +647,12 @@ private:
 			list.push_back(std::make_unique<DropDownListColourItem<>>(i, HasBit(used_colours, i)));
 		}
 
-		byte sel = (default_livery == nullptr || HasBit(livery->in_use, primary ? 0 : 1)) ? (primary ? livery->colour1 : livery->colour2) : default_col;
+		byte sel;
+		if (default_livery == nullptr || HasBit(livery->in_use, primary ? 0 : 1)) {
+			sel = primary ? livery->colour1 : livery->colour2;
+		} else {
+			sel = default_col;
+		}
 		ShowDropDownList(this, std::move(list), sel, widget);
 	}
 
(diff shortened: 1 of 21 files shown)


LordAro (commented):
Casting is normally `(Colours)foo` rather than `Colours(foo)`

src/newgrf.cpp
```diff
@@ -2405,10 +2405,10 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
 					housespec->grf_prop.local_id = hid + i;
 					housespec->grf_prop.subst_id = subs_id;
 					housespec->grf_prop.grffile = _cur.grffile;
-					housespec->random_colour[0] = 0x04;  // those 4 random colours are the base colour
-					housespec->random_colour[1] = 0x08;  // for all new houses
-					housespec->random_colour[2] = 0x0C;  // they stand for red, blue, orange and green
-					housespec->random_colour[3] = 0x06;
+					housespec->random_colour[0] = COLOUR_RED; // these 4 random colours are the base colour
```
2TallTyler: Not your doing, but an inline comment wrapping over two lines and trying to describe four lines is very creative!

Maybe fix that. 😉

src/saveload/oldloader_sl.cpp
```diff
@@ -465,10 +465,10 @@ static void FixTTOCompanies()
 	}
 }
 
-static inline byte RemapTTOColour(byte tto)
+static Colours RemapTTOColour(Colours tto)
```
2TallTyler: Lost its `inline` - intentional?

src/company_gui.cpp
```diff
@@ -582,7 +582,7 @@ static const LiveryClass _livery_class[LS_END] = {
 template <SpriteID TSprite = SPR_SQUARE>
 class DropDownListColourItem : public DropDownIcon<DropDownString<DropDownListItem>> {
 public:
-	DropDownListColourItem(int colour, bool masked) : DropDownIcon<DropDownString<DropDownListItem>>(TSprite, PALETTE_RECOLOUR_START + (colour % COLOUR_END), colour < COLOUR_END ? _colour_dropdown[colour] : STR_COLOUR_DEFAULT, colour, masked)
+	DropDownListColourItem(int colour, bool masked) : DropDownIcon<DropDownString<DropDownListItem>>(TSprite, GENERAL_SPRITE_COLOUR(colour % COLOUR_END), colour < COLOUR_END ? _colour_dropdown[colour] : STR_COLOUR_DEFAULT, colour, masked)
```
2TallTyler, line 585: `GENERAL_SPRITE_COLOUR` shows up several times but doesn't appear to originate in this PR. Is it a macro? (It looks like a constant but takes arguments like a function or constructor...)

src/newgrf.cpp
```diff
@@ -2502,7 +2502,7 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
 				break;
 
 			case 0x17: // Four random colours to use
-				for (uint j = 0; j < 4; j++) housespec->random_colour[j] = buf->ReadByte();
+				for (uint j = 0; j < 4; j++) housespec->random_colour[j] = Colours(buf->ReadByte() & 0xF);
```
2TallTyler: New magic number without explanation...?

src/company_gui.cpp
PeterN: It's a macro that does "PALETTE_RECOLOUR_START + (colour)", it is a bit unrelated but I changed it for consistency.

src/saveload/oldloader_sl.cpp
PeterN: Hmm, probably but not relevant to this change I think.

src/newgrf.cpp
PeterN: Just a mask. I've switched it to `GB(x, 0, 4)` which is still fairly magic, but is at least used in other places so it's consistent.

Colours is a currently uint8_t but only values 0-15 are valid colours, and 255 is explicitly INVALID_COLOUR, which doesn't make sense to have here. Currently the code doesn't range check, which would end up treating a non-recolour sprite as a recolour sprite.

PeterN:
Oops, while looking around I realised I missed `byte colour` on objects, so I'll tackle that later.

PeterN:
> Casting is normally `(Colours)foo` rather than `Colours(foo)`

Now using C++ style explicit `static_cast<Colours>(foo)`

2TallTyler (approved):

PR #11627, opened 2023-12-25
Codechange: replace NULL with nullptr

rubidium42:
## Motivation / Problem

Someone commenting on `NULL` in a diff, made me wonder if there's still any `NULL` left. There is.

## Description

Replace `NULL` with `nullptr`, in two commits; one for squirrel and one for the rest.

## Limitations

There are still some `NULL`s in the code, but that's in printf output, comments and json.hpp defines.
```
./saveload/compat/cheat_sl_compat.h:26: SLC(1, SL_MIN_VERSION, SLV_TABLE_CHUNKS), // Need to be two NULL fields. See Load_CHTS().
./saveload/compat/cheat_sl_compat.h:30: SLC(1, SL_MIN_VERSION, SLV_TABLE_CHUNKS), // Need to be two NULL fields. See Load_CHTS().
./saveload/compat/cheat_sl_compat.h:36: SLC(1, SL_MIN_VERSION, SLV_TABLE_CHUNKS), // Need to be two NULL fields. See Load_CHTS().
./saveload/saveload.h:708:      uint16_t length;                ///< Length of the NULL field.
./saveload/saveload.cpp:310:     * during NULLing; especially those that try to get
./stdafx.h:145:#        pragma warning(disable: 6011)   // code analyzer: Dereferencing NULL pointer 'pfGetAddrInfo': Lines: 995, 996, 998, 999, 1001
./fontcache.cpp:67:             return "[NULL]";
./3rdparty/squirrel/squirrel/sqvm.cpp:1583:             case OT:                        printf("NULL"); break;
./3rdparty/squirrel/sqstdlib/sqstdaux.cpp:56:                                   pf(v,fmt::format("[{}] NULL\n",name));
./3rdparty/nlohmann/json.hpp:2198:    #elif defined(NULL)
./3rdparty/nlohmann/json.hpp:2199:        #define JSON_HEDLEY NULL
./3rdparty/nlohmann/json.hpp:2203:#elif defined(NULL)
./3rdparty/nlohmann/json.hpp:2204:    #define JSON_HEDLEY NULL
./3rdparty/catch2/catch.hpp:2314:    // Specialised comparison functions to handle equality comparisons between ints and pointers (NULL deduces as an int)
```

The Microsoft API uses `NULL`, which [also allows nullptr](https://devblogs.microsoft.com/oldnewthing/20180307-00/?p=98175). We were already using `nullptr` in some places so I made it consistent.

## Checklist for review

[...]

PR #11628, opened 2023-12-26
Feature: Plugin framework for Social Integration with Steam, Discord, GOG, etc

TrueBrain:
## Motivation / Problem

This is another approach to both #11391 and #8952 .
Closes #11391 
Closes #8952

Quality games in 2023 have integration with at least Steam, to show in more detail what a player is up to, and make it trivial to join their game (assuming it is open to join, ofc). OpenTTD is lacking this functionality.

The main hurdle has always been licensing; Steamworks SDK (and Discord SDK, and GOG SDK) are closed-source, and don't play nice with our GPLv2 license. So a simple integration was long ago established to be impossible, as that would give us a license-hell nobody wanted to deal with.

So in #8952 it was already suggested to create a plugin-API, where the API definition is licensed under MIT. This means that there can be other projects, with different licenses, that implement integration with platforms like Steam, Discord, GOG, ...
By doing it this way, it is very similar to content on BaNaNaS. Most of it is not licensed under GPLv2, yet any user can download it, and amend their game with all the goodies BaNaNaS content brings.

#8952 stalled in development, and a few months ago we got a similar PR #11391, trying to do the same thing. When looking at it this morning I was like: this can be done in a few days, and is actually not that much code. It is much more about figuring out how to deliver it to the clients, and how to make the experience as smooth as possible. So .. here we are.

## Description

Update:
Plugins are now validated whether they are signed by us. This, as otherwise we are afraid people will make misuse of this interface, and start running coin-miners or what-not on machine of users, by distributing a malicious but seemly useful plugin via the forums or whatnot.

Most of this work is heavily based and/or inspired by #8952, but with a few twists. First off, it assumes the plugins are distributed via BaNaNaS (more on this later). Second, it moves the code out of the "network" and into the global space, as it is not network specific. Lastly, by rewriting it myself I could clean up some bits to make it work with the code in 2023 (that PR was created in 2021).

I liked the approach of #8952:
- A plugin exports a single "init" function, which fills a table with pointers for OpenTTD to use. This is in contrast to #11391, which went for the route to export all functions OpenTTD might call. The benefit of the first approach is that a single plugin can implement several API versions, and staying compatible is a problem for the plugin; not for OpenTTD. In the second approach, we would have to add compatibility layers, to ensure older plugins still work.
- #11391 is event-based, and per event there is a payload. This makes for a bit of an awkward definition, and a lot can go wrong here. #8952 is much more verbose, and announces each function with their parameters (where the functions are just events). I liked this descriptive API definition much more.

This is a preference, and neither is actually better or worse than the other. There are a few other things that both did a bit weird in my opinion, so I took the opportunity to straiten that out too.

That all said, what is this actually doing:

On startup, the `social_integration` folder is scanned much like any other folder in OpenTTD. It follows the same rules, same searchpaths, etc. For every plugin found (they always end on `-social.(so|dll|dylib)`), it is loaded and the function `SocialPlugin_v1_GetInfo` and `SocialPlugin_v1_Init` is looked up. 

`GetInfo` is called first, which returns things like the name of the plugin, the social platform it is for, etc.
After, `Init` is called, which initializes the plugin. One of three things can happen:
- The plugin fails to initialize
- The plugin fails to initialize because, for example, Steam is not running
- The plugin is initialized

Now every tick the `RunCallbacks` is called. When it returns `false`, the plugin is unloaded. When OpenTTD quits, all plugins are unloaded.

Events that happen, like someone starting a single player, or joining a server, triggers an event (a function callback), to inform the plugin. The `RunCallbacks` is guaranteed to be called in the GameThread. All other callbacks (both directions) are undefined where they run, and no assumptions should be made. It is up to the plugin to be thread-safe for the functions called by OpenTTD (and up to OpenTTD to be thread-safe for the callbacks it gives the plugin).

The API itself is under `3rdparty/openttd_social_plugin_api`. This is done as `3rdparty` is where code lives under another license than GPL-v2. And the API definition is licensed under MIT, allowing plugins to use any license they like.

Social Presence plugins can be distributed in two ways:
- For Steam, GOG, etc, we prepackage them in the distribution.
- Via BaNaNaS (for later PRs)

This last one needs a bit more words:

Only developers are allowed to upload new content in this category on BaNaNaS. This is because these are executables we distribute, and so they have to be hold to the same standards as the game itself. Source in a Version Control System, build via a predefined CI workflow, and no human interaction during distribution. This to make the chance of someone sneaking in something bad very very small.
If anyone would be able to upload a plugin, it would be trivial for anyone to upload a coinminer, or worse. Other BaNaNaS content is protected from that, as we load those files ourselves and validate they are sane. But as the plugins are executables, this is impractical.

The upside of this approach is that we can update the plugins outside a normal OpenTTD release cycle; this means that if an SDK breaks something, we can quickly act, and don't have to wait for the next release. Also it means that older OpenTTD versions get updates for Steam, Discord, GOG, ... integration.

The downside is that I had to figure out how to pack both the plugin library as their dependency (all SDKs give their own library) in a way that OpenTTD could easily load. I succeeded on both Linux, MacOS, and Windows to do so.

Example of Steam and Discord plugins:
- https://github.com/OpenTTD/discord-social
- https://github.com/OpenTTD/steam-social
- https://github.com/OpenTTD/gog-galaxy-social

This PR deliberately is only adding a very small plugin interface. Later PRs can extend on it. But getting everything setup and working is already a huge task; so to not let it linger for a longer time because "oeh, we can also do NNN", the scope is made as small as possible, while still being useful.

For example, BaNaNaS support is prepared, but not part of this PR. This is because we need to figure things out like: there should only be one plugin for a social platform active; but what is "latest"?
For now, people can download plugins manually, and we will ship it with Steam / GOG automatically. Still a benefit to the user, while not getting overwhelmed with all the details we need to figure out for the next step.

Screenshot 2023-12-29 202541
Screenshot 2023-12-29 203120
Screenshot 2023-12-29 205104

## Limitations

- We, as developers, need to agree on this approach. As mentioned above, it is exactly the same as other BaNaNaS content, so this should be fine by anyones standard, but it is important to note that those plugins will not be under GPLv2 license.
- This is not supported on Emscripten, as it cannot communicate with Steam, Discord, GOG, ... anyway.
- Linux Distributions that want to distribute the plugins by compiling them theirselves, would need to sign them and add their public key to the list of keys in the game, before they can be loaded. As such, I suspect no distribution to do this.

## TODO list

- [x] Add to crashlog what libraries are loaded
- [x] Add to GUI somewhere what libraries are loaded
- [x] Create Discord Plugin
- [x] Create Steam Plugin
- [x] Create GOG Plugin
- [x] Track what the player is doing
  - [x] In main menu
  - [x] In scenario editor
  - [x] In singleplayer
  - [x] In multiplayer
- [x] Support Windows
- [x] Support Linux
- [x] Support MacOS

## Checklist for review


nielsmh:
Regarding
> Allow joining someone's game if public (via invite-code only)
I think that might be complex enough to warrant its own PR. See the considerations I've made in #7735 about it. Especially the part with being able to launch the game if not running, and being able to handle multiple installed versions.

nielsmh:
I remember one of the reasons I put much of the code under `network/` in #8952 was that (at the time) tracking information about what network game the player was in was actually quite difficult.  After joining a game, most of the information was thrown away, so I needed to integrate tighter for that, but that might be different now.

Either way, I think tracking information about the company the player is in is quite useful (company name, number, colour, possibly fleet size and general financials, and if it's a company shared with multiple players) and is better to add sooner rather than later.

Another idea for a Free plugin would be a very basic one that simply writes all of the information to a plain text file, which could then be parsed out by whatever the user wants to set up. That could for example be a script for an IRC client or a text overlay for a live stream via OBS.

TrueBrain:
I gave it some thought, I am going to limit this PR to only show where in the game you are: menu, singleplayer, multiplayer, scenario editor, without any further detail.

It will already be a lot of working getting that done, and by creating a focus it is much more likely this PR actually finishes. All the other (great) ideas we have can be made (also by others) after this PR, also increasing the ability to review such PR :)

nielsmh:
Possibly another way to handle the plugin startup would be to have an exported function per API version, which could possibly build on top of each other.

So this initial implementation supports `OpenTTD_SocialPluginInit1` and when an extension to support company info, server name etc. is ready it would support both `OpenTTD_SocialPluginInit1` and `OpenTTD_SocialPluginInit2` where the game calls both functions in series in a plugin, if supported, and the structures passed to the `2` version only concern the new/changed features in that version.

TrueBrain:
> Possibly another way to handle the plugin startup would be to have an exported function per API version, which could possibly build on top of each other.
> 
> So this initial implementation supports `OpenTTD_SocialPluginInit1` and when an extension to support company info, server name etc. is ready it would support both `OpenTTD_SocialPluginInit1` and `OpenTTD_SocialPluginInit2` where the game calls both functions in series in a plugin, if supported, and the structures passed to the `2` version only concern the new/changed features in that version.

I went a slightly different route, with a similar idea.

TrueBrain:
I scoped down the PR even further, and this PR will not contain BaNaNaS integration.

That is to say, it is prepared for it, files are in a location BaNaNaS can write to, etc. But the distribution for now will be automatically via Steam / GOG, and otherwise players have to download the plugin themselves and put it in the right folder.

Who knows, maybe I find some more time before 14.0 releases to get BaNaNaS integration done; but I think it is more important to have something working, than to have everything done. It is just too much, and BaNaNaS integration has some hurdles to get over. So yeah .. scoping it down :)

src/os/windows/library_loader_win.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "../../library_loader.h"
+
+#include "../../safeguards.h"
+
+static std::string GetLoadError()
+{
+	auto error = GetLastError();
+	std::string message;
+	char buffer[512];
+	if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, error,
```
nielsmh: I worry that this might end up returning garbage for some locales. The error message might not be in English and the current ANSI codepage for the thread might not be UTF-8 (that might not even be supported on the Windows version the user is running), so this could end up generating mojibake when a non-UTF8 non-English string is treated as UTF-8 further down.

TrueBrain: This is already existing code in our codebase; I just copied it as the other versions are baked into other flows (like network error handling). So it already stood the test of time.

Further more, the result is only used in Debug statements to allow figuring out why a library failed loading. So I am not too fuzzed about it.

src/lang/english.txt
```diff
@@ -1076,6 +1078,19 @@ STR_GAME_OPTIONS_BASE_MUSIC                                     :{BLACK}Base mus
...
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_TITLE                            :{BLACK}{RAW_STRING} ({RAW_STRING})
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_PLATFORM                         :{BLACK}Platform:
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE                            :{BLACK}Plugin state:
+
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_RUNNING                    :{GREEN}Running
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_FAILED                     :{RED}Failed to initialize
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_PLATFORM_NOT_RUNNING       :{ORANGE}{RAW_STRING} not running
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_UNLOADED                   :{RED}Unloaded
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_DUPLICATE                  :{RED}Duplicated plugin
+STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_UNSUPPORTED_API            :{RED}API not supported
```
nielsmh: Maybe it will be clearer as "Version not supported" or "Unsupported API version"?

src/library_loader.h
```diff
@@ -0,0 +1,115 @@
...
+	void CloseLibrary();
+
+	/**
+	 * Get a symbol from the library.
+	 *
+	 * Should set has_error / error if any error occurred.
+	 *
+	 * @param symbol_name The name of the symbol to get.
+	 */
+	void *GetSymbol(const std::string &symbol_name);
```
nielsmh: Would this make more sense as a string_view for the parameter? Since it's generally static string constants and not constructed values used in calls.

src/social_integration.h
```diff
@@ -0,0 +1,84 @@
...
+
+	SocialIntegrationPlugin(const std::string &basepath) : basepath(basepath) {}
+};
+
+class SocialIntegration {
+public:
+	/**
+	 * Get the list of loaded social integration plugins.
+	 */
+	static std::vector<SocialIntegrationPlugin *> GetPlugins();
```
nielsmh, line 42: I think this can return a `const std::vector<SocialIntegrationPlugin *> &` instead, to avoid copying the vector every time someone wants the list just to inspect it.

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api_v1.h
```diff
@@ -0,0 +1,154 @@
...
+	 *
+	 * Values are like:
+	 * - Steam
+	 * - Discord
+	 * - GOG Galaxy
+	 * - ...
+	 */
+	const char *social_platform;
+
+	const char *name; ///< Full name of the plugin. UTF-8, nul-terminated. The plugin is and remains the owner of the memory.
```
rubidium42, line 60: Has any consideration been made to support translated strings?

Here `social_platform` seems to be most like the base set's name, and this `name` feels more like a base set's description. The base set's descriptions can be translated.

I'm aware that it is (currently) only for the name, as `social_platform` and `version` definitely shouldn't be translated I'd say. So, it might not be worth it. I just want to make sure a consideration has been made about it and it's not forgotten.

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api_v1.h
```diff
@@ -0,0 +1,154 @@
...
+	 * As there can only be one plugin active for each Social Platform, this
+	 * value is used to determine which plugin to use.
+	 *
+	 * Values are like:
+	 * - Steam
+	 * - Discord
+	 * - GOG Galaxy
+	 * - ...
+	 */
+	const char *social_platform;
```
rubidium42, line 58: Maybe we should set up a wiki page that is the definitive source of social platform names, so when two people are working on XYZ they will both call it XYZ instead of one calling in Xyz. As that would defeat the whole preventing of duplicates. And by not putting it in the source code directly, later additions can still be easily managed.
Also maybe check for duplicate social platform names case insensitively?

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api.h
```diff
@@ -0,0 +1,37 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/* Although all the source-files created by OpenTTD are licensed under the
+ * GPL-v2, this file is an exception. This file is part of the API for
+ * social integration plugins, and licensed under MIT, to allow for non-free
```
rubidium42:
```suggestion
 * social integration plugins, and licensed under the MIT license, to allow for non-free
```
and the few other places where this pre-amble is added.

src/os/windows/library_loader_win.cpp
```diff
@@ -0,0 +1,61 @@
...
+		MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), buffer, sizeof(buffer), NULL) == 0) {
```
rubidium42:
```suggestion
	if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, error,
		MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), buffer, sizeof(buffer), nullptr) == 0) {
```
Also, shouldn't we pass the lang ID from the language file here (`_current_language->winlangid`), so we get a translated error message back from Windows?

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api_v1.h
TrueBrain: Yes, I did consider it! :)

Name/version of the Plugin are just that: names. Not a description, like with basesets. It is more to allow for competitive implementations of a single plugin, so we can distinguish them. And names like that don't need translation if you ask me.

Similar to social_platform, Steam is called Steam in any language. Equally, I suspect that a plugin name will be named the same in every language. For example, the current names are `Steam Integration`  and `Discord Integration` :) I have such an imagination :P

I did consider just removing `name` all together, as it doesn't actually add value. But again, it might be useful if we have more than one competing implementation of the same social platform.

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api_v1.h
TrueBrain: Solved the last part. First part, sure, we can do that. Leaving open as I have yet to do that :)

Now done:
https://wiki.openttd.org/en/Development/Social%20Integration

src/os/windows/library_loader_win.cpp
TrueBrain: We don't in the other cases we use this function in a nearly identical way (which also use `NULL` btw :P Not for this PR to fix).

I think we talked about it during the development of the network variant of this function, and considered it to be better to use the system language, as then it matches the console language. Not sure, but I believe we spend time on this :)

Either way, if we want this, we should fix all instances of this function; so going to push that out of scope of this PR, if you don't mind :)

src/library_loader.h
TrueBrain: It cannot be a `std::string_view`, as in the usage, the strings need to be nul-terminated. Otherwise all implementations need to copy the string into a `std::string` to make it nul-terminated. Which ofc is a bit defeating the purpose.

This sadly happens when you interface with C-functions :(

So I think this is the best approach for us; if you have better suggestions, do let me know :)

src/os/windows/library_loader_win.cpp
rubidium42: Since https://github.com/OpenTTD/OpenTTD/pull/11627 that changed; we now use `nullptr` in those cases.

src/library_loader.h
nielsmh: Aren't string literals in the form of `"foo"` specified to generate a char array like `{ 'f', 'o', 'o', '\0' }` i.e. the terminator is guaranteed to be part of the literal? But yes it would not give the guarantee if the strings were for some reason actual views of larger strings.

TrueBrain:
Also finished the GOG plugin. Hard to test, as I cannot see my own rich presence (?!). So I am not sure if it simply isn't working, or that I just cannot see it. And I don't have any friends on GOG. So to properly test it, I need someone to befriend me on GOG, so they can see if it works. Not the most friendliest way of testing this, but okay.

Other than that, all ready and done.

src/3rdparty/openttd_social_integration_api/LICENSE
```diff
@@ -0,0 +1,20 @@
+ Copyright 2023 OpenTTD project
```
rubidium42: Should this year (and in the headers of the API) be reconsidered now it's 2024?

src/openttd.cpp
```diff
@@ -1009,6 +1012,28 @@ bool SafeLoad(const std::string &filename, SaveLoadOperation fop, DetailedFileTy
...
+static void UpdateSocialIntegration(GameMode game_mode)
+{
+	switch (game_mode) {
+		case GM_BOOTSTRAP:
+		case GM_MENU:
+			SocialIntegration::EventEnterMainMenu();
+			break;
+
+		case GM_NORMAL:
+			if (_network_server) {
```
rubidium42:
```suggestion
			if (_networking) {
```
Or is playing as a client considered single player?

src/os/unix/library_loader_unix.cpp
```diff
@@ -0,0 +1,68 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file library_loader_win.cpp Implementation of the LibraryLoader for Linux / MacOS */
```
rubidium42:
```suggestion
/** @file library_loader_unix.cpp Implementation of the LibraryLoader for Linux / MacOS */
```

src/library_loader.h
```diff
@@ -0,0 +1,115 @@
+/*
```
rubidium42: I think the library loader should at least be its own commit, if not its own PR.

rubidium42 (commented):
Also a tiny potential suggestion, depending on whether bananas integration is going to happen any time soon: a short manual on a wiki page listing known plugins and how to install them. That is linked from the social page when there are no plugins available.

src/library_loader.h
TrueBrain: Sounds like a good idea! This PR now includes [#11751](https://github.com/OpenTTD/OpenTTD/pull/11751).

src/settings_gui.cpp
```diff
@@ -166,6 +167,178 @@ static const std::map<int, StringID> _volume_labels = {
...
+
+	std::string &GetLongestVersion() const
+	{
+		std::string *longest = &this->plugins[0]->version;
+		for (auto *plugin : this->plugins) {
+			if (plugin->version.length() > longest->length()) longest = &plugin->version;
+		}
+
+		return *longest;
+	}
```
PeterN: These need to use the longest pixel length, not the longest character length.

WWWWWWWW is wider than ..........

TrueBrain: It would really help me if you tell how, instead of just: this is wrong :P

PeterN: Normally I'd use `GetStringBoundingBox().width` in the window's `UpdateWidgetSize()` and return the width there, but pattern doesn't really work with your everything-in-a-custom-widget layout (which I really like btw).

So something like this, although maybe there's a std algorithm or some other pattern that can be used to reduce the duplication:
```
std::string *longest = &this->plugins[0]->version;
int longest_length = 0;
for (auto *plugin : this->plugins) {
    int length = GetStringBoundBox(plugin->version).width;
    if (length > longest_length) {
        longest_length = length;
        longest = &plugin->version;
    }
}

return *longest;
```

TrueBrain:
> Also a tiny potential suggestion, depending on whether bananas integration is going to happen any time soon: a short manual on a wiki page listing known plugins and how to install them. That is linked from the social page when there are no plugins available.

As a follow-up on this:

I created three PRs, one for each plugin:
- https://github.com/OpenTTD/discord-social/pull/2
- https://github.com/OpenTTD/steam-social/pull/2
- https://github.com/OpenTTD/gog-galaxy-social/pull/2

Once these are merged, they will be available for download from https://cdn.openttd.org/. After that, I can give them an URL similar to https://www.openttd.org/downloads/opengfx-releases/latest . After that, I can update the wiki to reference those URLs, and after 14.0 is released we can mention them on https://www.openttd.org/downloads/openttd-releases/latest , much like OpenGFX / OpenSFX and OpenMSX are mentioned. (and ofc, the nightly URL can also get this information whenever this is merged).

For Steam and GOG I will update the distribution code to include these plugins by default. Well, Steam + Discord for Steam, and GOG + Discord for GOG.

After all that, we can update the ingame markdown etc to include references to that, including the window if there are no plugins. So yeah, good suggestion, just needs tons of steps to get there, but I will get there :D

src/settings_gui.cpp
TrueBrain: Awesome, tnx! Templating power will take care of the rest :)

src/os/unix/unix.cpp
```diff
@@ -13,6 +13,7 @@
 #include "../../string_func.h"
 #include "../../fios.h"
 #include "../../thread.h"
+#include "../../debug.h"
```
rubidium42: Stray include.

src/settings_gui.cpp
```diff
@@ -166,6 +167,171 @@ static const std::map<int, StringID> _volume_labels = {
...
+
+					case SocialIntegrationPlugin::DUPLICATE:
+						SetDParam(0, STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_DUPLICATE);
+						break;
+
+					case SocialIntegrationPlugin::UNSUPPORTED_API:
+						SetDParam(0, STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_UNSUPPORTED_API);
+						break;
+				}
+			} break;
```
2TallTyler: `break` goes on its own line, no?

src/settings_gui.cpp
```diff
@@ -166,6 +167,171 @@ static const std::map<int, StringID> _volume_labels = {
...
+					break;
+				}
+
+				SetDParamStr(0, this->plugins[this->current_index]->social_platform);
+				break;
+
+			case WID_GO_SOCIAL_PLUGIN_STATE: {
+				/* For SetupSmallestSize, use the longest string we have. */
+				if (this->current_index < 0) {
+					SetDParam(0, STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_UNSUPPORTED_API);
```
rubidium42: This is not necessarily the longest string in all translations. Actually "Discord not running" might even be a bit longer that "Unsupported version".
Practically you'd have to run through all the possibilities, i.e. all the states and all the plugins to determine the actual longest string. I'm not quite sure whether that's worth it though.

If the strings and the state are both in the same order, I'd just use the following to replace the big switch below:
```
SetDParam(0, STR_GAME_OPTIONS_SOCIAL_PLUGIN_STATE_RUNNING + plugin->state);
SetDParamStr(1, plugin->social_platform);
```

Then for getting the longest string, you can just iterate through the strings (by state) and set the platform, and simply try all cases. Yes, you will be setting the string unnecessarily, but I wonder if that's a big problem since it's not called that often.

src/settings_gui.cpp
```diff
@@ -166,6 +167,171 @@ static const std::map<int, StringID> _volume_labels = {
...
+	EndContainer(),
+};
+
+class NWidgetSocialPlugins : public NWidgetVertical {
+public:
+	NWidgetSocialPlugins()
+	{
+		this->plugins = SocialIntegration::GetPlugins();
+
+		if (this->plugins.size() == 0) {
```
rubidium42:
```suggestion
		if (this->plugins.empty()) {
```

src/social_integration.cpp
```diff
@@ -0,0 +1,239 @@
...
+		this->FileScanner::Scan(extension.c_str(), SOCIAL_INTEGRATION_DIR, false);
+	}
+
+	bool AddFile(const std::string &filename, size_t basepath_length, const std::string &) override
+	{
+		std::string basepath = filename.substr(basepath_length);
+		Debug(misc, 1, "[Social Integration: {}] Loading ...", basepath);
+
+		auto &plugin = _plugins.emplace_back(std::make_unique<InternalSocialIntegrationPlugin>(filename, basepath));
+		UpdateExternalPlugins();
```
rubidium42: Can't this be called in `Scan()` after the `FileScanner::Scan`? Then we do not rebuild the list for each plugin. Now with 3 plugins, you need 6 inserts into the vector.

src/social_integration.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+			default:
+				NOT_REACHED();
+		}
+	}
+};
+
+std::vector<SocialIntegrationPlugin *> &SocialIntegration::GetPlugins()
+{
+	return _external_plugins;
```
rubidium42: Personally I would have just created a local vector here, let that be filled from `_plugins` and then let return value optimisation take care to not introduce an actual copy of the vector. 

The current implementation will also work just fine.

src/social_integration.cpp
```diff
@@ -0,0 +1,239 @@
...
+ */
+template <typename T, typename... Ts>
+static void PluginCall(std::unique_ptr<InternalSocialIntegrationPlugin> &plugin, T func, Ts... args)
+{
+	if (plugin->external.state != SocialIntegrationPlugin::RUNNING) {
+		return;
+	}
+
+	if (func != nullptr) {
+		func(args...);
```
rubidium42, line 180: Maybe slap some try-catch block around the call, so if it's written in C++ and throws an exception and doesn't catch it by itself, then that plugin doesn't take down the whole of OpenTTD with it.
```suggestion
	try {
		func(args...);
	} catch (...) {
		Debug(misc, 0, "Social plugin {} threw an exception", plugin->plugin_info.name);
	}
```

README.md
```diff
@@ -118,6 +118,15 @@ Most types of add-on content can be downloaded within OpenTTD via the 'Check Onl
 Add-on content can also be installed manually, but that's more complicated; the [OpenTTD wiki](https://wiki.openttd.org/) may offer help with that, or the [OpenTTD directory structure guide](./docs/directory_structure.md).
 
 
+### 1.5.1) Social Integration
```
2TallTyler, line 121: If I'm a privacy-conscious user downloading the game from Steam (cherry-picking when I care about my privacy, naturally 😛), I get the plugins installed automatically. It would be nice to have a sentence here telling me that if I delete the file, the plugin is gone. Just to sooth worried minds.

2TallTyler (commented):
Two thoughts here:

First, plugins (in their own repos) currently use hard-coded strings. I assume can we add them to eints?

Second, you talk about plugin safety for our official plugins distributed on BaNaNaS. But if I'm reading this correctly, it seems that OpenTTD will run any plugin dropped into the correct folder. I wonder if we should lock that down and allow only official plugins to run.

Let's say that we don't have an official plugin for the user's favorite social platform, say ~~Twitter~~ X. Or we someday discontinue support for a social platform because we disagree with a ToS update and the survey shows 0.1% of our players are using it anyway. Some enterprising person could create their own plugin that includes a crypto-miner or keylogger or whatever, advertise it through Discord or GitHub or TT-Forums, and get a bunch of users to download it. Those players are accustomed to OpenTTD mods being safe to download and use, because of the sandbox that GRFs and AI/GS scripts are contained within. Not the case here. We didn't distribute the malicious code, sure, but if that user runs it through OpenTTD... do we share any of that responsibility?

src/social_integration.cpp
TrueBrain: The API is exported as a pure C API, so we shouldn't be able to capture anything, I would think.

TrueBrain:
> First, plugins (in their own repos) currently use hard-coded strings. I assume can we add them to eints?

No, kinda.

For Steam, we can make it work, as Steam allows translations.
For the other platforms, they do not facilitate translations. And it would be very weird if I see a French string because glx' OpenTTD is in French.

> Second, you talk about plugin safety for our official plugins distributed on BaNaNaS. But if I'm reading this correctly, it seems that OpenTTD will run any plugin dropped into the correct folder. I wonder if we should lock that down and allow only official plugins to run.

I understand your concern, but that is difficult in an Open Source world. MacOS does counter this (by only allowing plugins that have the same certificate as the binary), but on the other platforms there is no such infrastructure from the OS.
But from another point-of-view, it is pretty much the same argument as saying: but someone could download an OpenTTD version from the forums, which contains a bitcoin miner. I do not think there is much difference between the two. When a user downloads something on their own, they should think twice about the source; and of course they don't, but the plugin system won't be the easiest way to get in. Just publishing an OpenTTD version with a miner in it would be much easier.

That said, I do agree that having something might be nice. So I will think about this for a bit. Possibly a simple private/public key validation might just do.

README.md
TrueBrain: I am not sure.

Expecting Steam integration to be shipped with a game on Steam is kinda a no-brainer. Nobody is going to expect it to not be there, as most of the games on Steam ship it. If you are privacy conscious .. you would not have used Steam to download the game :D

For the Discord plugin on Steam, there you might have a point. I could make it a Steam DLC, if that helps.

The problem with deleting is that Steam will just restore it for you. So there is little point in suggesting something that is not going to work :)

src/social_integration.cpp
TrueBrain: I would classify this as over-optimizing :) Pretty sure the list will never get long enough to show up on any performance benchmark :) I would rather keep it local to where it is important.

That said, I will see if I can rework it to be a bit nicer. But not for the performance :P More because it is a bit awkward how I did it :)

Edit: I made it pretty!

README.md
2TallTyler: Hmm, DLC could be a good way to distribute it. I agree that shipping Steam integration with a Steam build shouldn't bother anybody -- and it really does, they can download through openttd.org.

src/social_integration.cpp
TrueBrain: You are not wrong; overoptimization on my side!

src/social_integration.cpp
rubidium42: I took the steam-social plugin and replaced all functions content with `throw 1`. The result, this PR crashes hard:
`terminate called after throwing an instance of 'int'`

Now with a try-catch block around the functions (I counted at least 3) you call, I get the debug message instead. So, apparently if the C code in between is thin enough it works fine for at least my current compiler. That doesn't mean it'll always work fine, but then we tried our best to keep OpenTTD running and there's nothing more to do than to tell them to remove the plugin.

TrueBrain: But in the same argument, if the plugin was in C, and crashes, it would still take the game down.

If (and that is a question: if) we want to solve this, I rather set a trap like we do in crashlog code, and capture any issue that might arise in the plugin handling.

But I think the question is here: do we actually want that? The plugins will be (after suggestion by 2TallTyler) guaranteed made available by us; so isn't it the same as a crash in the game, that we need to fix? Honest question, to be clear.
I just know from experience it is nearly impossible to prevent all ways a library can crash your application; so should we even try? Especially as a "throw" is rather unlikely to happen in the plugins.

src/settings_gui.cpp
TrueBrain: I did a thing. Mostly because I really didn't feel like making this `State` enum iteratable with `_BEGIN` and `_END` stuff. It makes it so ugly. And also because I don't like `STR_... + i` stuff :P It always makes my code fail when I change things :D

So I went for a different approach ... but not sure what people think. The compiler likes it equally, as in, with `-O3` there is no difference between this and a switch-case. So that is something.

TrueBrain:
I gave the point by @2TallTyler some thought, and I think we should do the following:

In JGRPP there is already https://monocypher.org/. We should also include that in our source tree, and with that, we can do validation of the plugins. One of the functions Monocypher offers is Public Key Signatures.

- We generate a private/public key pair, of which the private part is kept, well, private (and only available to our CI), and the public key is stored in OpenTTD itself.
- On release of our plugins, we create a manifest, in which we store the hash of both our library and the SDK, encrypted with the private key. Read: we create a Public Key Signature of the file.
- On loading of a plugin, we first open the manifest, and validate the signature matches the hash we calculated. If they match, we know the libraries are the same as when they were released AND that they were released by us. (ofc, for development purposes, there will be a build-flag that says something like: `-DALLOW_INSECURE_PLUGINS`, which doesn't require a valid signature for a plugin to load)

We have most of this code already, as we do the hash-validation part with basesets, NewGRFs, etc. So we have all the code for it. We only extend that by instead of reading the hash plaintext from a manifest, to have that signed, and validate the signature too. So we know the author of the plugin. It does however create more code, so I will have to see how I keep this PR reviewable.

The one drawback of this method is, that if the private key leaks, security is lost, as the public key is built-in to OpenTTD, and we cannot revoke it. Building infrastructure for that feels like over-complicated the situation way too much, with very little added value.

PS: although some OSes support adding the signature to the library itself (Windows and MacOS mostly), this isn't universal. Linux for example has no method of adding a signature to a library. So by using a manifest and doing this part ourselves, we make sure it works for all platforms.

src/social_integration.cpp
TrueBrain: I did some further reading, and my initial instinct is correct: C++ doesn't define the ABI of anything, let alone exceptions. This is why we fall back here ofc to a C ABI, as that is well defined.

It means what you tried basically only works between a single compiler, and MIGHT work between different compilers, but also might hopefully and utterly fail in worse ways than not doing it. Even different versions of the same compiler might handle the ABI differently, making a throw from the library incompatible with the application. As it is very likely the plugins will be compiled (over time) by other compilers and versions, it is not advised to send exceptions over the barrier.

A nice write-up I found about it here:

https://wiki.sei.cmu.edu/confluence/display/cplusplus/ERR59-CPP.+Do+not+throw+an+exception+across+execution+boundaries

Here they give a nice example that OpenTTD might be build in MinGW and the plugin with MSVC. This would absolutely not work, and MSVC uses SEH for exceptions, and MinGW does not. On Linux we might be a bit more lucky, but it is just that: luck. It is UB.

That was a nice trip down a rabbit hole which I actually enjoyed :)

What is left, I think, are two options:

- Use, as with crashlog code, a `setjmp` / SEH solution to trap any issues, and recover from it.
- Or, as because of the private/public key validation I will add, consider plugins as part of our codebase that are just loaded in at a later time, and that they are no different than any other routine in our codebase. They just live in another repo.

I still vote for the last, but if there is a good argument for the first, that can be done :)

rubidium42:
On the chat:
> 08:38 < truebrain> for MacOS that currently already is true btw; MacOS refuses to load libraries that aren't signed by the same certificate as the application 🙂

That sounds like a release nightmare to happen, but I might be wrong. I have no clue about how it works, so this is based on assumptions that might be wrong.

I am going to assume that, as is common practice, certificates should not be very long lived. I'll assume that's going to be 1 year, and then we'll need/get a new certificate.

Wouldn't that mean that if we build a new release/nightly, the previously working MacOS plugins will not work any more and we need to rebuild them as well? That'd require the user updating them, but when going back to a previous version (e.g. 15.0-beta1 vs 14.2 or so) the plugin won't work again. Will that then require essentially downgrading?

Or do we need some infrastructure for MacOS where there can be multiple versions of the same plugin, but just signed differently? Though, if MacOS refuses to load them at all, having many in the folder might not be such a big problem if only those with the right certificate are shown in the UI.

src/settings_gui.cpp
PeterN: IMHO if you don't know what the longest string might be, and you don't want to test all possible combinations, it's better to design it around a wrapping mutli-line string instead. Especially if some translations might be quite long.

I guess the video driver description in the game options window is one example of where this is done.

TrueBrain: For context, the string in question is `Steam is not running`. And I mostly got myself in this position as it is not a regular window, but a composed one :P
Wrapping would be a bit odd here, even for translations, in how it would look. So currently it makes the window wider. But I am not really sure this is the right approach.

Maybe I should just not try to put things this neatly, and just put it below each other, give it a width, and let wrapping happen when it needs to happen.

Meh :P

src/signature.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+#include "3rdparty/monocypher/monocypher.h"
+#include "3rdparty/nlohmann/json.hpp"
+
+#include "safeguards.h"
+
+/** The public keys used for signature validation. */
+static std::vector<std::string> _public_keys_v1 = {
+	"1234567890123456789012345678901234567890123456789012345678901234",
+	"1234567890123456789012345678901234567890123456789012345678901234",
```
TrueBrain: Ofc these are not valid yet; I need to generate a private/public pair and add the private part to the CI where needed. And the public pair here.

src/social_integration.cpp
```diff
@@ -25,16 +26,23 @@
  */
 class InternalSocialIntegrationPlugin {
 public:
-	InternalSocialIntegrationPlugin(const std::string &filename, const std::string &basepath) : library(filename), external(basepath)
+	InternalSocialIntegrationPlugin(const std::string &filename, const std::string &basepath) : library(nullptr), external(basepath)
 	{
 		openttd_info.openttd_version = _openttd_revision;
+
+		if (!ValidateSignatureFile(fmt::format("{}.sig", filename))) {
```
TrueBrain, line 33: For development builds, it should be possible to bypass this check, and just make it continue.

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2023 OpenTTD project
```
glx22:
```suggestion
 * Copyright 2024 OpenTTD project
```

src/3rdparty/openttd_social_integration_api/openttd_social_integration_api_v1.h
```diff
@@ -0,0 +1,157 @@
+/*
+ * Copyright 2023 OpenTTD project
```
glx22:
```suggestion
 * Copyright 2024 OpenTTD project
```

glx22 (commented):
Just to match the standalone license file.

src/social_integration.cpp
```diff
@@ -40,6 +59,45 @@ class SocialIntegrationFileScanner : FileScanner {
...
+				Debug(misc, 1, "[Social Plugin: {}] Loaded for {}: {} ({})", basepath, plugin->plugin_info.social_platform, plugin->plugin_info.name, plugin->plugin_info.version);
+				return true;
+
+			case OTTD_SOCIAL_INTEGRATION_V1_INIT_FAILED:
+				Debug(misc, 0, "[Social Plugin: {}] Failed to initialize", basepath);
+				return false;
+
+			case OTTD_SOCIAL_INTEGRATION_V1_INIT_PLATFORM_NOT_RUNNING:
+				Debug(misc, 0, "[Social Plugin: {}] Failed to initialize: {} is not running", basepath, plugin->plugin_info.social_platform);
+				return false;
```
rubidium42: The begin of the  `Debug` messages feel a bit inconsistent. Or is it wanted that messages before attempting the initialisation describe the plugin differently than after?
The first two are "Social Integration: <name>", the last three are "Social Plugin: <name>" but that might be intended.
The third is like the first two but missing a colon; this is definitely triggering my OCD.

src/signature.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+		if (hi < 0 || lo < 0) {
+			return false;
+		}
+
+		bytes[i] = (hi << 4) | lo;
+	}
+
+	return true;
+}
```
rubidium42: settings.cpp has `DecodeHexNibble` and `DecodeHexText`. Maybe it should be moved to string.cpp/string_func.h, so it they can share the same function.
Found out about their existence when looking at PR 11855, because it changed the NewGRF configuration output.

src/signature.cpp
```diff
@@ -0,0 +1,328 @@
...
+	 *
+	 *   {
+	 *     "files": [
+	 *       {
+	 *         "checksum": "version$hash"
+	 *         "filename": "filename",
+	 *       },
+	 *       ...
+	 *     ],
+	 *     "signature": "version$signature"
```
rubidium42, line 232: What's the reasoning for putting version into the checksum/signature string, and not as a separate field? Now you need to manually split the string, instead of reading two fields.

src/signature.cpp
```diff
@@ -0,0 +1,328 @@
...
+{
+	_ValidateSignatureFile(filename);
+	return true;
+}
+#else
+bool ValidateSignatureFile(const std::string &filename)
+{
+	return _ValidateSignatureFile(filename);
+}
+#endif
```
rubidium42: I would put the `#if`, `#else` and `#endif` within the function, so there is only one function definition.

rubidium42 (commented):
Can't find anything major wrong, just some minutiae.

src/social_integration.cpp
TrueBrain: I am happy with your OCD, as they were all meant to be the same :D

src/signature.cpp
TrueBrain: Yeah, I copied the Nibble from it. The DecodeHexText has a very very odd return value. I might refactor that first, as it just hurts my eyes :)

src/signature.cpp
TrueBrain: Unix. I like their passwd which also includes the version. If you copy the string now, it is obvious what version it refers to, and there is no question. But it isn't a strong reason.

Will give it some thought if I want to keep it like this or not.

TrueBrain: Gave it some more thought, and I think this is the right way to go (what is currently done).

If I would do it in a separate field, you would get something like `checksum_version`, which feels really awkward and bulky. And I don't want to do a `schema` or something, as I suspect that it is more likely that one of the two (signature or checksum) changes, than both at the same time. This is more descriptive long-term, I think.

The other solution would be to nest the checksum, so you get:

```
{
  "checksum": {
    "version": 1,
    "checksum": "...",
  }
}
```

But I dunno .. lot of letters for basically the same thing.

Always tricky, designing these things with long-term in mind :D

rubidium42 (approved):

PR #11633, opened 2023-12-27
Fix: Changing default livery did not propagate to group liveries.

PeterN:
## Motivation / Problem

Setting a group livery to "default" set the group's colours the the default at-the-time. Changing the default company livery after would not change group liveries set to default, if they were partially default. This was never intended behaviour.

image
image

## Description

#11614 attempted to address this but did not handle 2CC properly, and changes to the default livery were not handled.

Instead, explicitly update a company's group liveries when the default livery scheme is changed. This correctly handles partially-default group liveries.

Because the incorrect colours are stored in savegames, this is also called on loading savegames.

image
image

## Checklist for review

### DIFF
--- a/src/group_cmd.cpp
+++ b/src/group_cmd.cpp
@@ -466,10 +466,11 @@ CommandCost CmdAlterGroup(DoCommandFlag flags, AlterGroupMode mode, GroupID grou
 			g->parent = (pg == nullptr) ? INVALID_GROUP : pg->index;
 			GroupStatistics::UpdateAutoreplace(g->owner);
 
-			if (g->livery.in_use == 0) {
+			if (!HasBit(g->livery.in_use, 0) || !HasBit(g->livery.in_use, 1)) {
+				/* Update livery with new parent's colours if either colour is default. */
 				const Livery *livery = GetParentLivery(g);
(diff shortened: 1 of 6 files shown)


src/group_cmd.cpp
```diff
@@ -278,14 +278,16 @@ const Livery *GetParentLivery(const Group *g)
...
  */
-void PropagateChildLivery(const Group *g)
-{
-	/* Company colour data is indirectly cached. */
-	for (Vehicle *v : Vehicle::Iterate()) {
-		if (v->group_id == g->index && (!v->IsGroundVehicle() || v->IsFrontEngine())) {
-			for (Vehicle *u = v; u != nullptr; u = u->Next()) {
-				u->colourmap = PAL_NONE;
-				u->InvalidateNewGRFCache();
+void PropagateChildLivery(const Group *g, bool reset_cache = true)
```
rubidium42: Shouldn't the call at line 297/299 of this file pass `reset_cache`?

PeterN: Good point. I think I'll make it non-optional to be safe :)

rubidium42 (approved):

PR #11634, opened 2023-12-28
Feature: Custom company colours.

PeterN:
## Motivation / Problem

Company colours only allow 16 preset colour choices from the 8bpp palette. While there have been limited ways to change those presets, it is not possible to freely choose any colour.

## Description

Implement support for custom company colours.

Custom company colours can be selected by hue, saturation, value (brightness) and an extra parameter contrast which controls the difference between the brightest and darkest colours.

There are 16 custom colour presets stored in the config, set by choosing a colour then Ctrl+Click on the preset button.

image

image

## Checklist for review


zephyris:
For the problem of not enough company colours, more 8bpp company colours and 32bpp company colours are not mutually exclusive solutions. Another, non-mutually exclusive solution is wider support for 2CC (eg. in station signs, base set infrastructure/vehicles, etc.).

This solution looks good to me, except:
* If 8bpp blitters are not being depreciated then some thought about 8bpp fallback might be needed?
* Presumably a minimum hue/sat/lum distance could be enforced between players? First come, first served, as per normal CCs.

_In the future_ saving a preferred colour would be nice.

2TallTyler (commented):
A few thoughts, without digging too deeply into the code:
* The first two commits could be done as separate PRs now, they seem like good improvements to make anyway.
  * Codechange: Remove direct access to _colour_gradient.
  * Codechange: Rename Colour to RgbaColour.
* The user interface could do with some labels. At the very least, the two columns of swatches -- I assume left is Presets and right is Custom/Saved?
* I really like the big slider handles paired with the fine indicator lines. I am slightly bothered by the mismatch between their positions, but I see why syncing them properly is a no-go: not only would the buttons need to extend beyond the rectangle of the current UI, the indicators would need to line up with the center of the handles, so the scrollbar tracks would need to extend past the color box too. At the bottom right of the 2D selector, these would collide with each other. So I vote to keep it the way you've designed it; I can learn to love it. 🙂

PeterN:
(Earlier commits are designed to be broken into smaller more easily digestible PRs, but may need reworking so best to keep them all together for now.)

PeterN:
Related to conversation in #15505, but technically this is off-topic there.

It is mainly a problem that we already have 4 different ways to specify colours at the moment.

* `Colour`, which is 32bpp RGBA data.
* `PixelColour`, which is just an index into the 8bpp palette.
* `TextColour`, which is an index into a lookup table, or it can be a 8bpp palette index. This also contains some string drawing flags, so it's 16 bits.
* `Colours`, which is the fixed 16 company colours.

The custom company colours branch doesn't remove any of these, but extends them:

* `Colour` stays the same.
* `PixelColour` is extended to 32 bits and contains "RGBM" data. A pixel colour can then be an arbitrary RGB value, or the closest matching palette index.
* `TextColour` is extended to 32 bits (from 16 bits), and the upper bits can be packed with 18-bit RGB values.
* `HsvColour` is new, and is a 32 bit value contains 16 bit hue, 8 bit saturation, and 8 bit value. This is used in the custom colour interface.
* `Colours` is extended to 32 bits, and packs in a 9 bit hue, 6 bit saturation, 6 bit value, and 4 bit contrast.

HSV is used because it still represents hue in situations where RGB completely loses it and is just better for humans to deal with when picking colours.

PR #11635, opened 2023-12-28
Fix: Base life under 8 years could underflow, leading to very long-lived engines.

PeterN:
## Motivation / Problem

Engine reliability duration phase 2 takes 96 months off the specified value (for reasons unknown to me, but changing that would break the default engines, and break all NewGRFs.)

If an engine exists with a specified base life of less than 96 months, then this could underflow, leading to a life in the order of 65560 months.

## Description

Ensure duration phase 2 does not go below zero.

While working here, I've also replaced some magic hex numbers with some more meaningful constants for base reliability values.

## Checklist for review

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -717,12 +717,23 @@ void StartupOneEngine(Engine *e, TimerGameCalendar::Date aging_date, uint32_t se
 	              e->type ^
 	              e->GetGRFID());
 
-	r = Random();
-	e->reliability_start = GB(r, 16, 14) + 0x7AE0;
-	e->reliability_max   = GB(r,  0, 14) + 0xBFFF;
+	/* Base reliability defined as a percentage of UINT16_MAX. */
+	const uint16_t RELIABILITY_START = UINT16_MAX * 48 / 100;
+	const uint16_t RELIABILITY_MAX   = UINT16_MAX * 75 / 100;
+	const uint16_t RELIABILITY_FINAL = UINT16_MAX * 25 / 100;
+
+	static_assert(RELIABILITY_START == 0x7AE0);
+	static_assert(RELIABILITY_MAX   == 0xBFFF);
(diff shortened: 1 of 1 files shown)


src/engine.cpp
```diff
@@ -717,14 +717,24 @@ void StartupOneEngine(Engine *e, TimerGameCalendar::Date aging_date, uint32_t se
...
+	const uint16_t RELIABILITY_FINAL = UINT16_MAX * 25 / 100;
+
+	static_assert(RELIABILITY_START == 0x7AE0);
+	static_assert(RELIABILITY_MAX   == 0xBFFF);
+	static_assert(RELIABILITY_FINAL == 0x3FFF);
+
 	r = Random();
-	e->reliability_start = GB(r, 16, 14) + 0x7AE0;
-	e->reliability_max   = GB(r,  0, 14) + 0xBFFF;
+	e->reliability_start = GB(r, 16, 14) + RELIABILITY_START;
```
nielsmh, line 731: Is it worth also noting that taking 14 bits gives a range of 25 %p to adjust by? It took me a couple minutes of staring to realize.

rubidium42 (approved):

PeterN:
Some NewGRFs ignore the spec and misuse 0 to mean never expires, so this fix breaks them.

Should we revise this to treat 0 as 255 to not change their behaviour?

reldred:
For reference, only grf I've directly observed this is NARS addon. I've got the sources on hand, if necessary I can update that grf and submit it to the author for upload as they're inactive now, but who knows how many other .grf's are (ab)using this behavior.

reldred:
Update: I submitted an update to the old maintainer of NARS Addon which is now on the content service as an update. Unless this behavior is rampant in other grf's I think this PR can be left merged as it is with no further action necessary.

PR #11641, opened 2023-12-29
Codechange: Replace C-style nested array with std::map for widget lookup.

PeterN:
## Motivation / Problem

For indexed widget access, an array is allocated and two stages, first to find the maximum index, then to fill in the array.

This uses C-style memory allocation, and is generally clumsy.

## Description

Rename nested_array to widget_lookup, to name the member by functionality rather than type, and replace the C-style memory allocation with a std::map, which handles its own allocation and can be filled in as needed.

## Checklist for review

### DIFF
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -855,9 +855,8 @@ GUIGameServerList::FilterFunction * const NetworkGameWindow::filter_funcs[] = {
 	&NGameSearchFilter
 };
 
-static NWidgetBase *MakeResizableHeader(int *biggest_index)
+static NWidgetBase *MakeResizableHeader()
 {
-	*biggest_index = std::max<int>(*biggest_index, WID_NG_INFO);
 	return new NWidgetServerListHeader();
 }
 
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -496,7 +496,7 @@ bool Window::SetFocusedWidget(int widget_index)
 	/* Do nothing if widget_index is already focused, or if it wasn't a valid widget. */
 	if ((uint)widget_index >= this->nested_array_size) return false;
 
-	assert(this->nested_array[widget_index] != nullptr); // Setting focus to a non-existing widget is a bad idea.
+	assert(this->widget_lookup[widget_index] != nullptr); // Setting focus to a non-existing widget is a bad idea.
 	if (this->nested_focus != nullptr) {
 		if (this->GetWidget<NWidgetCore>(widget_index) == this->nested_focus) return false;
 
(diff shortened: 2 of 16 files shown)


src/network/network_gui.cpp
```diff
@@ -855,9 +855,8 @@
 	&NGameSearchFilter
 };
 
-static NWidgetBase *MakeResizableHeader(int *biggest_index)
+static NWidgetBase *MakeResizableHeader()
```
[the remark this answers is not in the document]
PeterN: Not sure what this is about, it's used by `NWidgetFunction(MakeResizableHeader)` on line 882 :o

src/window.cpp
```diff
@@ -531,21 +525,23 @@ void Window::OnFocusLost(bool)
...
+	for (auto &pair : this->widget_lookup) {
+		WidgetType type = pair.second->type;
+		NWidgetCore *wid = dynamic_cast<NWidgetCore *>(pair.second);
 		if (((type & ~WWB_PUSHBUTTON) < WWT_LAST || type == NWID_PUSHBUTTON_DROPDOWN) &&
-				(!autoraise || (type & WWB_PUSHBUTTON) || type == WWT_EDITBOX) && this->IsWidgetLowered(i)) {
-			this->RaiseWidget(i);
-			this->SetWidgetDirty(i);
+				(!autoraise || (type & WWB_PUSHBUTTON) || type == WWT_EDITBOX) && wid->IsLowered()) {
+			wid->SetLowered(false);
+			wid->SetDirty(this);
```
TrueBrain, line 534: This change feels unexpected. Was it intentional? And if so, is it really part of this PR?

src/window_gui.h
```diff
@@ -260,8 +260,7 @@ struct Window : ZeroedMemoryAllocator {
 	const NWidgetCore *nested_focus; ///< Currently focused nested widget, or \c nullptr if no nested widget has focus.
 	std::map<int, QueryString*> querystrings; ///< QueryString associated to WWT_EDITBOX widgets.
 	NWidgetBase *nested_root;        ///< Root of the nested tree.
-	NWidgetBase **nested_array;      ///< Array of pointers into the tree. Do not access directly, use #Window::GetWidget() instead.
-	uint nested_array_size;          ///< Size of the nested array.
+	WidgetLookup widget_lookup; ///< Indexed access to the nested widget tree. Do not access directly, use #Window::GetWidget() instead.
```
TrueBrain, line 263: Could this now also be private (or protected)?

Seems all access has wrappers around it.

src/window.cpp
PeterN: Kinda. The methods on Windows basically just use GetWidget<>(i), and call the method on the widget... because we iterate widget_lookup, we already have the widget, so can just call them directly.

But I could undo it and replace i with pair.first instead.

TrueBrain: The raised vs lowered was making me go: euhm? I guess it does the same, but one is an action, and one changes the state. I can't see if there are side-effects :(

Edit: went up and down the pipe, seems fine this. So keep it :)

src/window_gui.h
PeterN: DecreaseWindowCounters() accesses it, and that's static, so not quite.

TrueBrain (approved):

PR #11648, opened 2023-12-30
Fix #11646: Non-thread safe shared buffer returned from GetLogPrefix().

### LINKED ISSUE #11646: [Bug]: GetLogPrefix function static state is not thread-safe
### Version of OpenTTD

master

### Expected result

No thread safety issues.

### Actual result

`GetLogPrefix` is called via the `Debug` macro, and other routes.
It returns a pointer to a static std::string buffer.
Other threads may concurrently modify/reallocate that state (particularly when _settings_client.gui.show_date_in_logs is set).

### Steps to reproduce

Do something which calls Debug in multiple threads.

PeterN:
## Motivation / Problem

GetLogPrefix reuses a static std::string, and returns its buffer after updating it. This is not thread safe.

This is probably a left-over from using C-style strings.

## Description

Return string from GetLogPrefix instead of shared string's buffer.

## Checklist for review

### DIFF
--- a/src/console.cpp
+++ b/src/console.cpp
@@ -51,10 +51,9 @@ static void IConsoleWriteToLogFile(const std::string &string)
 {
 	if (_iconsole_output_file != nullptr) {
 		/* if there is an console output file ... also print it there */
-		const char *header = GetLogPrefix();
-		if ((strlen(header) != 0 && fwrite(header, strlen(header), 1, _iconsole_output_file) != 1) ||
-				fwrite(string.c_str(), string.size(), 1, _iconsole_output_file) != 1 ||
-				fwrite("\n", 1, 1, _iconsole_output_file) != 1) {
+		try {
+			fmt::print(_iconsole_output_file, "{}{}\n", GetLogPrefix(), string);
+		} catch (const std::system_error &) {
 			fclose(_iconsole_output_file);
 			_iconsole_output_file = nullptr;
 			IConsolePrint(CC_ERROR, "Cannot write to console log file; closing the log file.");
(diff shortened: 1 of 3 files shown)


src/debug.cpp
```diff
@@ -235,15 +235,13 @@ std::string GetDebugString()
  * the date, otherwise it returns nothing.
  * @return the prefix for logs (do not free), never nullptr
```
rubidium42: This comment seems outdated now.

glx22 (dismissed):

src/console.cpp
```diff
@@ -51,8 +51,8 @@ static void IConsoleWriteToLogFile(const std::string &string)
 {
 	if (_iconsole_output_file != nullptr) {
 		/* if there is an console output file ... also print it there */
-		const char *header = GetLogPrefix();
-		if ((strlen(header) != 0 && fwrite(header, strlen(header), 1, _iconsole_output_file) != 1) ||
+		std::string header = GetLogPrefix();
+		if ((!header.empty() && fwrite(header.c_str(), header.size(), 1, _iconsole_output_file) != 1) ||
 				fwrite(string.c_str(), string.size(), 1, _iconsole_output_file) != 1 ||
 				fwrite("\n", 1, 1, _iconsole_output_file) != 1) {
```
rubidium42:
```suggestion
		try {
			fmt::print(_iconsole_output_file, "{}{}\n", GetLogPrefix(), string);
		} catch (const std::system_error &e) {
```
What about this solution to writing the output? I checked `format-inl.h`, and that seems to be throwing `std::system_error` when `fwrite` does not write everything.

rubidium42 (approved):

PR #11651, opened 2023-12-30
Fix: [CI] patch in SHF_COMPRESSED symbol for our Linux Generic binaries

### LINKED ISSUE #11650: [Bug]: Nightly builds are failing since December 27
### Version of OpenTTD

Since ce6a65d7b5174da6556057e10f05081fdfb7d8a5

### Expected result

Nightly builds should build and publish for all platforms at the expected time.

### Actual result

Since December 27, the nightly builds are failing in the Linux (Generic) category, causing no builds to be published.

### Steps to reproduce

Wait for the nightly build via GitHub Actions, then check the logs.
Most recent failing run: https://github.com/OpenTTD/OpenTTD/actions/runs/7359178501

TrueBrain:
## Motivation / Problem

Fixes #11650.

vcpkg upgraded to a breakpad that is somewhat recent (over a year old, but still). Sadly, in this new version breakpad started to use `SHF_COMPRESSED`, without actually checking of the OS knows that symbol. They just assume your OS is new enough ..

And guess what: our linux-generic builds on an OS a tiny bit older :P (to make sure we are as compatible as we can be).

In short: build fails because of it.

## Description

Solution is simply: patch in that symbol on system-level.

## Checklist for review

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -97,6 +97,14 @@ jobs:
           cmake --build . -j $(nproc)
           cmake --install .
         )
+
+        # The container we use is old enough, that it doesn't know SHF_COMPRESSED.
+        # But, breakpad needs this symbol to exist. So we patch it in our system
+        # libraries.
+        (
+          cd /
+          patch -p1 < ${GITHUB_WORKSPACE}/os/linux/shf-compressed.patch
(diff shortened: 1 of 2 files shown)


glx22 (approved):
Probably the only viable solution

PR #11653, opened 2023-12-30
Change: add vcpkg.json to instruct vcpkg what dependencies we require

TrueBrain:
## Motivation / Problem

MSVC now integrated vcpkg, meaning the developer experience to get going is much smoother if we supply a `vcpkg.json` which states the dependencies.

Not unexpected, it turns out that this makes listing our dependencies also a lot easier for the other targets, like Linux and MacOS. It also gives a nice overview of what dependencies we use for what OS.

## Description

So, introducing, `vcpkg.json`, which lists our dependencies to get going.

There is one trick .. for our Linux CI we depend on system libraries rather than vcpkg libraries, with the exception of breakpad. But when vcpkg sees a `vcpkg.json`, it absolutely refuses to install a single package. So we just rename the file.
Similar for the Release CI, we need to install Python3 first. For this we simply change folder.

This PR grew a bit bigger than just this change, and now also changes how vcpkg is being cached. We now use the integrated GitHub Actions method, instead of doing it ourselves. This greatly speeds up compiles in most situations, while always installing the latest dependencies.

Also, while at it, 3 commits to clean up some minor stuff with our GitHub Workflows.

Example run of a release: https://github.com/TrueBrain/OpenTTD/actions/runs/7365346246

## Limitations

I had to revert an older commit which made the Linux Release use the vcpkg of the host system. The problem with this is that it became difficult to locally test if the workflow was using. As the problems we experiences were unrelated to what vcpkg to use anyway, switching back to cloning vcpkg locally, and using that. That makes it far easier to test out the Linux Release.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -108,6 +108,14 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
 
+    - name: Setup vcpkg caching
+      uses: actions/github-script@v6
+      with:
+        script: |
+          core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+          core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+          core.exportVariable('VCPKG_BINARY_SOURCES', 'clear;x-gha,readwrite')
+
     - name: Install dependencies
       run: |
         echo "::group::Update apt"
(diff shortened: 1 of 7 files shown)


glx22:
Of course linux CI fails because https://github.com/OpenTTD/OpenTTD/pull/11543 added an extra check

TrueBrain:
Running an example release flow here:

https://github.com/TrueBrain/OpenTTD/actions/runs/7365346246

glx22 (approved):

PR #11654, opened 2023-12-30
Codechange: Use vector and unique_ptr for nested widget tree.

PeterN:
## Motivation / Problem

A nested widget tree uses linked lists to link sibling widgets, and manual pointer management to clear up.

## Description

Implement factory pattern to create widgets -- this prevents widgets being created as non-unique_ptrs, and allows the change to unique_ptr to be slightly more localised.

Replace linked lists with std::vectors on containers (instead of storing head/tail pointers) and use std::unique_ptr.

## Checklist for review

### DIFF
--- a/src/graph_gui.cpp
+++ b/src/graph_gui.cpp
@@ -109,19 +109,19 @@ struct GraphLegendWindow : Window {
  * Construct a vertical list of buttons, one for each company.
  * @return Panel with company buttons.
  */
-static NWidgetBase *MakeNWidgetCompanyLines()
+static std::unique_ptr<NWidgetBase> MakeNWidgetCompanyLines()
 {
-	NWidgetVertical *vert = new NWidgetVertical(NC_EQUALSIZE);
+	auto vert = std::make_unique<NWidgetVertical>(NC_EQUALSIZE);
 	vert->SetPadding(2, 2, 2, 2);
 	uint sprite_height = GetSpriteSize(SPR_COMPANY_ICON, nullptr, ZOOM_LVL_OUT_4X).height;
 
 	for (WidgetID widnum = WID_GL_FIRST_COMPANY; widnum <= WID_GL_LAST_COMPANY; widnum++) {
-		NWidgetBackground *panel = new NWidgetBackground(WWT_PANEL, COLOUR_BROWN, widnum);
+		auto panel = std::make_unique<NWidgetBackground>(WWT_PANEL, COLOUR_BROWN, widnum);
 		panel->SetMinimalSize(246, sprite_height + WidgetDimensions::unscaled.framerect.Vertical());
 		panel->SetMinimalTextLines(1, WidgetDimensions::unscaled.framerect.Vertical(), FS_NORMAL);
 		panel->SetFill(1, 1);
 		panel->SetDataTip(0x0, STR_GRAPH_KEY_COMPANY_SELECTION_TOOLTIP);
-		vert->Add(panel);
+		vert->Add(std::move(panel));
 	}
 	return vert;
 }
@@ -1337,7 +1337,7 @@ CompanyID PerformanceRatingDetailWindow::company = INVALID_COMPANY;
  * Make a vertical list of panels for outputting score details.
  * @return Panel with performance details.
  */
-static NWidgetBase *MakePerformanceDetailPanels()
+static std::unique_ptr<NWidgetBase> MakePerformanceDetailPanels()
 {
 	const StringID performance_tips[] = {
 		STR_PERFORMANCE_DETAIL_VEHICLES_TOOLTIP,
@@ -1354,18 +1354,18 @@ static NWidgetBase *MakePerformanceDetailPanels()
 
 	static_assert(lengthof(performance_tips) == SCORE_END - SCORE_BEGIN);
 
-	NWidgetVertical *vert = new NWidgetVertical(NC_EQUALSIZE);
+	auto vert = std::make_unique<NWidgetVertical>(NC_EQUALSIZE);
 	for (WidgetID widnum = WID_PRD_SCORE_FIRST; widnum <= WID_PRD_SCORE_LAST; widnum++) {
-		NWidgetBackground *panel = new NWidgetBackground(WWT_PANEL, COLOUR_BROWN, widnum);
+		auto panel = std::make_unique<NWidgetBackground>(WWT_PANEL, COLOUR_BROWN, widnum);
 		panel->SetFill(1, 1);
 		panel->SetDataTip(0x0, performance_tips[widnum - WID_PRD_SCORE_FIRST]);
-		vert->Add(panel);
+		vert->Add(std::move(panel));
 	}
 	return vert;
 }
 
 /** Make a number of rows with buttons for each company for the performance rating detail window. */
-NWidgetBase *MakeCompanyButtonRowsGraphGUI()
+std::unique_ptr<NWidgetBase> MakeCompanyButtonRowsGraphGUI()
 {
 	return MakeCompanyButtonRows(WID_PRD_COMPANY_FIRST, WID_PRD_COMPANY_LAST, COLOUR_BROWN, 8, STR_PERFORMANCE_DETAIL_SELECT_COMPANY_TOOLTIP);
 }
--- a/src/network/network_gui.cpp
+++ b/src/network/network_gui.cpp
@@ -115,18 +115,18 @@ public:
 		this->resize_y = 0; // We never resize in this direction
 
 		/* First initialise some variables... */
-		for (NWidgetBase *child_wid = this->head; child_wid != nullptr; child_wid = child_wid->next) {
+		for (const auto &child_wid : this->children) {
 			child_wid->SetupSmallestSize(w);
 			this->smallest_y = std::max(this->smallest_y, child_wid->smallest_y + child_wid->padding.Vertical());
 		}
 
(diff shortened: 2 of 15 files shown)


src/graph_gui.cpp
```diff
@@ -109,19 +109,19 @@ struct GraphLegendWindow : Window {
  * Construct a vertical list of buttons, one for each company.
  * @return Panel with company buttons.
  */
-static NWidgetBase *MakeNWidgetCompanyLines()
+static std::unique_ptr<NWidgetBase> MakeNWidgetCompanyLines()
 {
-	NWidgetVertical *vert = new NWidgetVertical(NC_EQUALSIZE);
+	std::unique_ptr<NWidgetVertical> vert = std::make_unique<NWidgetVertical>(NC_EQUALSIZE);
```
LordAro: This is one of those places where I'd use auto, rather than just repeating the type

src/network/network_gui.cpp
```diff
@@ -138,10 +138,12 @@ class NWidgetServerListHeader : public NWidgetContainer {
 		this->current_x = given_width;
 		this->current_y = given_height;
 
-		given_width -= this->tail->smallest_x;
+		given_width -= this->children.back()->smallest_x;
 		/* The first and last widget are always visible, determine which other should be visible */
-		for (NWidgetBase *child_wid = this->head->next; child_wid->next != nullptr; child_wid = child_wid->next) {
-			if (given_width > ScaleGUITrad(MINIMUM_NAME_WIDTH_BEFORE_NEW_HEADER) + child_wid->smallest_x && child_wid->prev->current_x != 0) {
+		auto end = std::prev(std::end(this->children));
+		for (auto it = std::next(std::begin(this->children)); it != end; ++it) {
```
LordAro: apart from being very ugly (though i'm not sure how that could be improved), this also has the potential issue of a one-length array never terminating (it = container.end, end = container.0th) . Is that worth worrying about?

src/network/network_gui.cpp
```diff
@@ -150,15 +152,21 @@ class NWidgetServerListHeader : public NWidgetContainer {
...
 				child_wid->AssignSizePosition(sizing, x + position, y, child_wid->current_x, this->current_y, rtl);
 				position += child_wid->current_x;
 			}
+		};
+
+		if (rtl) {
+			std::for_each(std::rbegin(this->children), std::rend(this->children), assign_position);
+		} else {
+			std::for_each(std::begin(this->children), std::end(this->children), assign_position);
 		}
```
LordAro, line 173: Not a huge fan of this. I suppose
```
if (rtl) {
    start = rbegin...
    end = rend...
} else {
    start = begin...
    end = end...
}
for (auto it = start; it != end; ++it) {
    ...
}
```
won't work due to type differences?

src/newgrf_gui.cpp
```diff
@@ -1600,27 +1600,22 @@ NewGRFWindow::GUIGRFConfigList::FilterFunction * const NewGRFWindow::filter_func
...
 public:
 	static const uint MAX_EXTRA_INFO_WIDTH;    ///< Maximal additional width given to the panel.
 	static const uint MIN_EXTRA_FOR_3_COLUMNS; ///< Minimal additional width needed before switching to 3 columns.
 
-	NWidgetBase *avs; ///< Widget with the available grfs list and buttons.
-	NWidgetBase *acs; ///< Widget with the active grfs list and buttons.
-	NWidgetBase *inf; ///< Info panel.
+	std::unique_ptr<NWidgetBase> avs; ///< Widget with the available grfs list and buttons.
+	std::unique_ptr<NWidgetBase> acs; ///< Widget with the active grfs list and buttons.
+	std::unique_ptr<NWidgetBase> inf; ///< Info panel.
```
LordAro, line 1610: worth extending the names?

src/toolbar_gui.cpp
```diff
@@ -1381,12 +1381,13 @@ class NWidgetToolbarContainer : public NWidgetContainer {
...
-			child_wid->current_x = 0; /* Hide widget, it will be revealed in the next step. */
-			if (child_wid->type == NWID_SPACER) continue;
-			widgets[((NWidgetCore*)child_wid)->index] = child_wid;
+		/* Create us ourselves a quick lookup table from WidgetID to slot. */
+		std::map<WidgetID, uint> lookup;
+		for (auto it = std::begin(this->children); it != std::end(this->children); ++it) {
+			NWidgetBase *nwid = it->get();
+			nwid->current_x = 0; /* Hide widget, it will be revealed in the next step. */
+			if (nwid->type == NWID_SPACER) continue;
+			lookup[dynamic_cast<NWidgetCore *>(nwid)->index] = std::distance(this->children.begin(), it);
```
LordAro, line 1390: null protection? the c-cast wouldn't have ever returned null (just garbage, presumably)

src/widget.cpp
```diff
@@ -1857,7 +1811,7 @@ void NWidgetVertical::AssignSizePosition(SizingType sizing, int x, int y, uint g
 
 		if (num_changing_childs == 0 && (flags & NC_BIGFIRST) && biggest_stepsize > 0) {
 			/* Second.5 loop: count how many children are of the updated biggest step size. */
-			for (NWidgetBase *child_wid = this->head; child_wid != nullptr; child_wid = child_wid->next) {
+			for (auto &child_wid : this->children) {
```
LordAro: feels like quite a lot of these for range loops could be const ?

src/widget.cpp
```diff
@@ -2168,18 +2121,16 @@ void NWidgetMatrix::GetScrollOffsets(int &start_x, int &start_y, int &base_offs_
...
+	this->child = std::move(child);
 	if (this->child != nullptr) this->child->parent = this;
 	this->SetAlignment(SA_TOP | SA_LEFT);
 }
 
 NWidgetBackground::~NWidgetBackground()
-{
-	if (this->child != nullptr) delete this->child;
-}
+{ }
```
LordAro: do we normally leave a space for empty functions? I think mostly they're just each on their own line as "normal"

LordAro (commented):
just some thoughts. feel free to ignore/reject as usual

src/widget.cpp
PeterN: Hmm, well, actually they can ALL be const, but that looks weird...

LordAro: weird how? To me all range-for should be `const auto &`, anything else is "strange" and suggests that the object is being modified

PeterN:
Pants, MSVC doesn't let you friend std::make_unique :o

src/widget.cpp
TrueBrain: `~NWidgetBackground() = defauilt` in definition?

Empty dtors are always a bit weird to see :)

PeterN: Oops, when I back-tracked to the pre-factory pattern version I lost the change that just removes this destructor.

src/network/network_gui.cpp
PeterN: Indeed, sadly forward and reverse iterators can be mixed like that, it would basically need two separate duplicate loops. And this way basically is, but done behind for_each.

src/newgrf_gui.cpp
PeterN: Probably, but they're used in enough places that that will have to be a separate PR.

src/network/network_gui.cpp
rubidium42: It's not like the original was safe for a one-length array: `child_wid` would be `nullptr` because `this->head->next` must be `nullptr` for a one-length array. Then the condition of the for loop would do a nullptr-dereference.

I'm seeing asserts below to check for a specific `size()` of children, might it be useful to add something like `assert(this->children.size() >= 2);`?

src/network/network_gui.cpp
rubidium42: Might it make sense to reverse the order of `this->children` when it's a horizontal container widget and rtl is active? Though that would probably be way out of scope for this PR.

src/widget.cpp
PeterN: To me it looks like I shouldn't be able to change child_wid's variables nor call non-const functions. But I can, because it's the unique_ptr that's const, not the object that the unique_ptr points to.

That's why it looks odd.

src/widget.cpp
```diff
@@ -1401,11 +1369,11 @@ void NWidgetStacked::SetupSmallestSize(Window *w)
 	/* First sweep, recurse down and compute minimal size and filling. */
 	this->smallest_x = 0;
 	this->smallest_y = 0;
-	this->fill_x = (this->head != nullptr) ? 1 : 0;
-	this->fill_y = (this->head != nullptr) ? 1 : 0;
-	this->resize_x = (this->head != nullptr) ? 1 : 0;
-	this->resize_y = (this->head != nullptr) ? 1 : 0;
-	for (NWidgetBase *child_wid = this->head; child_wid != nullptr; child_wid = child_wid->next) {
+	this->fill_x = (!this->children.empty()) ? 1 : 0;
```
rubidium42: Maybe use `this->IsEmpty()` instead?

src/network/network_gui.cpp
PeterN: Due to the nature of this widget, it's not possible, but I have added the check for size() > 2 anyway.

src/network/network_gui.cpp
PeterN: If I reverse the widgets then the loop that determines which header should be hidden needs to be reversed depending on RTL. Can't win!

rubidium42 (approved):

PR #11660, opened 2023-12-31
Fix: Calculation of initial engine age was inaccurate.

PeterN:
## Motivation / Problem

Engine age in months is calculated as the difference in days / 32, instead of the actual difference in months. This results in engines being artificially younger if a game is started at a later date than they are if started at an earlier date.

This not easily visible as engine age is not displayed anywhere.

## Description

Instead, convert introduction date and aging date to YearMonthDay and compare the Year and Month components, so that age in months is correct.

## Checklist for review

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -675,7 +675,7 @@ void SetYearEngineAgingStops()
  * @param aging_date The date used for age calculations.
  * @param seed Random seed.
  */
-void StartupOneEngine(Engine *e, TimerGameCalendar::Date aging_date, uint32_t seed)
+void StartupOneEngine(Engine *e, const TimerGameCalendar::YearMonthDay &aging_ymd, uint32_t seed)
 {
 	const EngineInfo *ei = &e->info;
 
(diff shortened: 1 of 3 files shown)


src/engine.cpp
```diff
@@ -699,7 +699,11 @@ void StartupOneEngine(Engine *e, TimerGameCalendar::Date aging_date, uint32_t se
 	 * Note: TTDP uses fixed 1922 */
 	e->intro_date = ei->base_intro <= TimerGameCalendar::ConvertYMDToDate(_settings_game.game_creation.starting_year + 2, 0, 1) ? ei->base_intro : (TimerGameCalendar::Date)GB(r, 0, 9) + ei->base_intro;
 	if (e->intro_date <= TimerGameCalendar::date) {
-		e->age = (aging_date - e->intro_date).base() >> 5;
+		TimerGameCalendar::YearMonthDay intro_ymd = TimerGameCalendar::ConvertDateToYMD(e->intro_date);
+		int aging_months = aging_ymd.year.base() * 12 + aging_ymd.month;
+		int intro_months = intro_ymd.year.base() * 12 + intro_ymd.month;
+		e->age = aging_months - intro_months;
+		if (aging_ymd.day < intro_ymd.day) e->age--; // Don't count partial months
```
rubidium42: I'm seeing that `e->age` is calculated differently here than it would be if the intro date were the same and the game started earlier.

When I start the game at 01-01-2020 and the vehicle introduction is on 16-12-2023, then at 01-01-2024 the vehicle is introduced and age is 0. Only on 01-02-2024 its age will become 1.
When I start the game at 17-01-2020, the vehicle's age will be set to 1, and on 01-02-2024 it'll be 2.

Though I also know that games can only start at the first of January, in which case this 'accidentally' works. But that's more because `aging_ymd.day` is always 1 and when `intro_ymd.day` is anything other than 1, it is actually introduced the next month and the age is decremented correctly.
So, `if (intro_ymd.day != 1) e->age--; // Engines are introduced at the first month start at/after the intro_date` might explain the actual logic better. And it would be consistent with the normal introductions.

PeterN: Ah, I forgot that vehicles are not introduced on their intro date, only on the next month...

rubidium42 (approved):

PR #11661, opened 2023-12-31
Change: Only remove orders from primary vehicles

SamuXarick:
## Motivation / Problem
When a station, waypoint and a depot is removed or when an airport is replaced with another without hangars, a remove orders is issued on all vehicles, but checking disaster, effect vehicles, free wagons or aircraft shadows and rotors isn't necessary.

## Description
When a station, waypoint and a depot is removed or when an airport is replaced with another without hangars, remove orders only from primary vehicles.
Checking IsPrimaryVehicle here does just that. It can also help with performance as well.

## Limitations
None.

## Checklist for review

### DIFF
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -1747,6 +1747,8 @@ void RemoveOrderFromAllVehicles(OrderType type, DestinationID destination, bool 
 
 	/* Go through all vehicles */
 	for (Vehicle *v : Vehicle::Iterate()) {
+		if (!v->IsPrimaryVehicle()) continue;
+
 		if ((v->type == VEH_AIRCRAFT && v->current_order.IsType(OT_GOTO_DEPOT) && !hangar ? OT_GOTO_STATION : v->current_order.GetType()) == type &&
 				(!hangar || v->type == VEH_AIRCRAFT) && v->current_order.GetDestination() == destination) {
 			v->current_order.MakeDummy();


TrueBrain:
So many things in here that makes my spidey-senses tickle ...

You say it is a "fix", but a fix suggest something was broken. What was broken before this PR? Non-primary-vehicles have an empty order list, so everything does ... exactly what it should do? The motivation/description do not mention what this is fixing.

To remind you, as by https://github.com/OpenTTD/OpenTTD/blob/master/CODINGSTYLE.md#commit-message, a "fix" should be fixing stuff.

Under limitations you have "possibly none" .. so .. you think there are any limitations? Are you sure there are none? I mean .. I am not sure now? Do I need to do your work again to know it is none? I am a bit confused.

Next, does this PR actually fix anything? For any vehicles without any orders the loop is very fast .. and performance isn't an issue either way (as it is a rare action), but one can debate that checking for all vehicles if it is primary is more expensive than just checking for any orders.

Lastly, your description says this is about removing of a station. But this function is used in more places than just those. So it feels like you only had a very narrow scope when creating this PR, which brings us back to feedback you receive a lot from us: does it still work out fine when you look at the total scope of the game?

So that brings us back to the motivation .. I am unsure what problem this is fixing? Maybe it is just a "codechange", but even then I somewhat doubt it is worth the PR? I am just a bit confused.

SamuXarick:
> Next, does this PR actually fix anything? For any vehicles without any orders the loop is very fast .. and performance isn't an issue either way (as it is a rare action), but one can debate that checking for all vehicles if it is primary is more expensive than just checking for any orders.

I will investigate if it's faster to iterate from the OrderListPool than the VehiclePool.

SamuXarick:
Closed. Reason, `IsPrimaryVehicle` is a slow check. :(
image

I tested 3 alternatives to master, and starting the iteration from the `OrderListPool` may be worthy of consideration. It's also a simple change.

```
    for (const OrderList *order_list : OrderList::Iterate()) {
        for (Vehicle *v = order_list->GetFirstSharedVehicle(); v != nullptr; v = v->NextShared()) {
```
Just that.

PR #11662, opened 2023-12-31
Change: Don't allow station/roadstop/object picker text to extend width.

PeterN:
## Motivation / Problem

Some NewGRF stations/roadstops/objects use very long strings for class or type names, which results in a very wide window.

image

## Description

Instead, truncate or wrap the text as appropriate.

image

## Limitations

Long class names are truncated, but I'm not sure if that's much of an issue. WWT_MATRIX can't have variable height entries.

## Checklist for review

### DIFF
--- a/src/object_gui.cpp
+++ b/src/object_gui.cpp
@@ -52,6 +52,7 @@ class BuildObjectWindow : public Window {
 	int object_margin;                     ///< The margin (in pixels) around an object.
 	int line_height;                       ///< The height of a single line.
 	int info_height;                       ///< The height of the info box.
+	Dimension name;                        ///< Dimensions of object type name widget.
 	Scrollbar *vscroll;                    ///< The scrollbar.
 
 	static Listing   last_sorting;         ///< Default sorting of #GUIObjectClassList.
@@ -187,13 +188,6 @@ public:
 	void SetStringParameters(WidgetID widget) const override
 	{
 		switch (widget) {
-			case WID_BO_OBJECT_NAME: {
-				ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
-				const ObjectSpec *spec = objclass->GetSpec(_selected_object_index);
-				SetDParam(0, spec != nullptr ? spec->name : STR_EMPTY);
-				break;
-			}
-
 			case WID_BO_OBJECT_SIZE: {
 				ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
 				const ObjectSpec *spec = objclass->GetSpec(_selected_object_index);
@@ -216,11 +210,6 @@ public:
 	{
 		switch (widget) {
 			case WID_BO_CLASS_LIST: {
-				for (auto object_class_id : this->object_classes) {
-					ObjectClass *objclass = ObjectClass::Get(object_class_id);
-					if (objclass->GetUISpecCount() == 0) continue;
-					size->width = std::max(size->width, GetStringBoundingBox(objclass->name).width + padding.width);
-				}
 				this->line_height = GetCharacterHeight(FS_NORMAL) + padding.height;
 				resize->height = this->line_height;
 				size->height = 5 * this->line_height;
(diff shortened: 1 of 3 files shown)


src/object_gui.cpp
```diff
@@ -321,6 +310,13 @@ class BuildObjectWindow : public Window {
 				break;
 			}
 
+			case WID_BO_OBJECT_NAME: {
+				const ObjectSpec *spec = ObjectClass::Get(_selected_object_class)->GetSpec(_selected_object_index);
+				StringID str = spec != nullptr ? spec->name : STR_EMPTY;
+				DrawStringMultiLine(r, str, TC_ORANGE, SA_CENTER);
```
rubidium42:
```suggestion
				if (spec != nullptr) DrawStringMultiLine(r, spec->name, TC_ORANGE, SA_CENTER);
```
Why go through the string drawing, when we know it's an empty string?

src/rail_gui.cpp
```diff
@@ -1200,21 +1196,9 @@ struct BuildRailStationWindow : public PickerWindowBase {
...
-				StringID str = this->GetWidget<NWidgetCore>(widget)->widget_data;
-				for (auto station_class : this->station_classes) {
-					StationClass *stclass = StationClass::Get(station_class);
-					for (uint j = 0; j < stclass->GetSpecCount(); j++) {
-						const StationSpec *statspec = stclass->GetSpec(j);
-						SetDParam(0, (statspec != nullptr && statspec->name != 0) ? statspec->name : STR_STATION_CLASS_DFLT_STATION);
-						d = maxdim(d, GetStringBoundingBox(str));
-					}
+				} else {
+					size->width = 0;
```
rubidium42: Don't both if-branches have `size->width = 0;` now? So, can't it be moved out of the if?

src/rail_gui.cpp
```diff
@@ -1308,18 +1299,37 @@ struct BuildRailStationWindow : public PickerWindowBase {
...
+	 */
+	Dimension GetStationTypeNameDimension(int width) const
+	{
+		int y = 0;
+		for (StationClassID cls_id = STAT_CLASS_BEGIN; cls_id < StationClass::GetClassCount(); cls_id++) {
+			if (cls_id == STAT_CLASS_WAYP) continue;
+			const StationClass *stclass = StationClass::Get(cls_id);
+			for (uint j = 0; j < stclass->GetSpecCount(); j++) {
+				const StationSpec *statspec = stclass->GetSpec(j);
+				StringID str = (statspec != nullptr && statspec->name != 0) ? statspec->name : STR_STATION_CLASS_DFLT_STATION;
```
rubidium42: Maybe move this logic into its own function, because it's used multiple times and it's not that trivial.

src/road_gui.cpp
```diff
@@ -1480,18 +1473,37 @@ struct BuildRoadStationWindow : public PickerWindowBase {
...
+	 */
+	Dimension GetRoadStopTypeNameDimension(int width) const
+	{
+		int y = 0;
+		for (RoadStopClassID cls_id = ROADSTOP_CLASS_BEGIN; cls_id < RoadStopClass::GetClassCount(); cls_id++) {
+			if (cls_id == ROADSTOP_CLASS_WAYP) continue;
+			RoadStopClass *rs_class = RoadStopClass::Get(cls_id);
+			for (uint j = 0; j < rs_class->GetSpecCount(); j++) {
+				const RoadStopSpec *roadstopspec = rs_class->GetSpec(j);
+				StringID str = (roadstopspec != nullptr && roadstopspec->name != 0) ? roadstopspec->name : STR_STATION_CLASS_DFLT_ROADSTOP;
```
rubidium42: Maybe move this to a separate function as well?

2TallTyler (commented):
This appears to truncate quite a few class names. If you drag the window wider, does it reveal more of the name?

PeterN:
> This appears to truncate quite a few class names. If you drag the window wider, does it reveal more of the name?

No. Making the window wider expands the picker preview area, and that would not work properly if the class name was allowed to expand.

And making sure the class name doesn't make this window fill the screen is the entire point.

PeterN:
Perhaps selected class name could go in the window title? Someone did suggest this, but I'm not sure.

2TallTyler:
Could the class selection widget span the entire width of the window? It wouldn't match other picker windows, but the object picker has the same problem already and could benefit from the same change.

PR #11665, opened 2023-12-31
Fix: replace the cedilla characters with comma ones for Romanian

georgeabr:
This is specifically for s-cedilla and t-cedilla to their comma variants.
These variants, especially in smaller font sizes, look almost identical but
they are different. Currently the translation uses a mix of the cedilla and
comma variants, where the cedilla ones are often in the older strings.

Replace reinnoi with innoi, as it is the correct form.

### DIFF
--- a/src/lang/romanian.txt
+++ b/src/lang/romanian.txt
@@ -308,7 +308,7 @@ STR_OSK_KEYBOARD_LAYOUT_CAPS                                    :~!@#$%^&*()_+|Q
 STR_MEASURE_LENGTH                                              :{BLACK}Lungime: {NUM}
 STR_MEASURE_AREA                                                :{BLACK}Suprafața: {NUM} x {NUM}
 STR_MEASURE_LENGTH_HEIGHTDIFF                                   :{BLACK}Lungime: {NUM}{}Diferența de înălțime: {HEIGHT}
-STR_MEASURE_AREA_HEIGHTDIFF                                     :{BLACK}Suprafaţa: {NUM} x {NUM}{}Diferenţa de înălţime: {HEIGHT}
+STR_MEASURE_AREA_HEIGHTDIFF                                     :{BLACK}Suprafața: {NUM} x {NUM}{}Diferența de înălțime: {HEIGHT}
 
 
 # These are used in buttons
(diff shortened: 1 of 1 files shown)


glx22:
Translations are handled via [web translator](https://translator.openttd.org/)

nielsmh:
Actually, I'm reopening this, because I see that the translation is already inconsistent in what character is used, and a (very) quick investigation points to it possibly only being really old translations, like 10-12 years old, that have the wrong letter forms.
Given how many lines are modified, I wonder if this can be accepted anyway.

rubidium42:
I think it's fine to merge this as it stands, as it's about a huge number of strings and hopefully the translator tool picks it up correctly. However, please do further improvements via the translator tool. You can ask for permissions at https://github.com/OpenTTD/team/issues/new/choose, and when that's done improve the translation at https://translator.openttd.org/.

Having said that, there are some pretty strict rules about the content of the commit message. That's done both for consistency and for making it easier to write tools to help collate changelogs for releases. I have tried correcting the commit message, but I'm not allowed to push to your repository, so you have to fix it yourself.
The first thing is that there must be a colon after the "Update", and the second thing is that the message should consist of only printable ASCII characters. So the actual characters can't be in the commit message, although I would agree that having them in the commit message helps with explaining why there are so many that are wrong in the translation.

I would suggest a message like:
```
Update: replace the cedilla characters with comma ones for Romanian

This is specifically for s-cedilla and t-cedilla to their comma variants.
These variants, especially in smaller font sizes, look almost identical but
they are different. Currently the translation uses a mix of the cedilla and
comma variants, where the cedilla ones are often in the older strings.

Replace reinnoi with innoi, as it is the correct form.
```

rubidium42:
I've got some bad news.

[Apparently](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md#i-want-to-revert-a-broken-change-some-translator-just-did-via-eints) merging translations is pointless because the translator tool will not be updated, and the tool will overwrite the translations on its next commit run. Essentially if we merge this, it will be automatically undone by the translator tool within a day.

There seems to be functionality to upload a translation file into the translator tool, but that requires someone with admin rights. The translator tool has been migrated to a different authentication system, but authorization for admins has never been implemented. So first someone with knowledge of the translator tool and the new authentication system needs to fix that. The really bad news is that there seems to be nobody able or willing to do that.

The alternative is that the translator tool does pick up changes from the repository, but also that needs to be implemented. That seems to be an even bigger endeavour.

frosch123 (approved):
Changing translations via git works as long as they do not revert strings to an older version in eints' translation memory.
The changes in this PR are new changes, so it works.
Reverting this PR later would not work :)

PR #11670, opened 2024-01-01
Change: Limit total script ops that can be consumed by a list valuate

JGRennison:
## Motivation / Problem

Script functions called by ScriptList::Valuate are never suspended.
A single script function called by ScriptList::Valuate could therefore consume an unlimited amount of CPU time, blocking the actual game.

## Description

Generate an error, killing the script, once a threshold number of ops is consumed by a (top level) valuate operation, even in the case where the function would never otherwise return.

## Limitations

Using the suspend parameter of sq_call complicates the accounting of the remaining ops and the handling of the stack in the case where the limit has been reached. An attempt was made using this, but was abandoned in favour of the approach in this PR.

The existing squirrel code is mostly very ugly. In order to maintain a consistent code style care has been taken to ensure that this PR is also aesthetically displeasing.

## Checklist for review

### DIFF
--- a/src/script/api/script_list.cpp
+++ b/src/script/api/script_list.cpp
@@ -11,7 +11,9 @@
 #include "script_list.hpp"
 #include "script_controller.hpp"
 #include "../../debug.h"
+#include "../../core/backup_type.hpp"
 #include "../../script/squirrel.hpp"
+#include <../squirrel/sqvm.h>
 
 #include "../../safeguards.h"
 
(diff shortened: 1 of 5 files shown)


src/script/api/script_list.cpp
```diff
@@ -12,9 +12,13 @@
 #include "script_controller.hpp"
 #include "../../debug.h"
 #include "../../script/squirrel.hpp"
+#include <../squirrel/sqvm.h>
```
PeterN, line 16: Using `../` with `#include <>` looks wrong to me.

JGRennison: That's how both of the existing squirrel includes in /src/script/ are.

TrueBrain: Most likely by error. A while ago we fixed a bunch of those `""` vs `<>` for squirrel. Guess those were missed :)

glx22:
`ScriptVehicleList` constructor will need the same kind of limitation for the filter.

src/script/api/script_list.cpp
```diff
@@ -906,6 +918,7 @@ SQInteger ScriptList::Valuate(HSQUIRRELVM vm)
 				/* See below for explanation. The extra pop is the return value. */
 				sq_pop(vm, nparam + 4);
 
+				vm->_ops_till_suspend_error_threshold = backup_ops_error_threshold;
```
TrueBrain: Maybe we should make a small Backup wrapper here, so it is automatically restored for us when leaving scope?

Dunno if that is actually an improvement tbh .. but missing a `return` is also an easy mistake made :(

PeterN: Make, or use `AutoRestoreBackup`?

src/script/api/script_list.cpp
JGRennison: It doesn't compile with the `../` removed.
In CMakeLists.txt there is `include_directories(${CMAKE_SOURCE_DIR}/src/3rdparty/squirrel/include)`, but 
`src/script/squirrel.cpp` wants includes in `src/3rdparty/squirrel/squirrel/`.

rubidium42 (approved):

PR #11672, opened 2024-01-03
Allow GS to add animated/static text to the map

npabisz:
## Motivation / Problem

Motivation to develop this feature is adding more features which could be used in multiplayer by Game Scripts. This will be very similar to how income from eg. trains shows on the map.

Screenshot_1586

## Description

This PR adds `SpawnAnimatedText` method to `GSTile` class. This could be useful in games which use some kind of scenarios or game scripts to indicate of changes, eg. in city builder mode it could indicate increase in population.

## Limitations

**Last time I coded in C++ was 12 years ago**. 

So any help and advices would be appreciated. I would like this to take full advantage of `GSText` with colors, but I'm stuck. I'm not sure how to work with `string`/`StringID`. As I understand GS will execute command with string which will be converted from `GSText`. However I don't know how to actually display it since method responsible for showing income `ShowCostOrIncomeAnimation` (which functionality I base on) uses `StringID`.

### DIFF
--- a/src/script/api/script_text_effect.cpp
+++ b/src/script/api/script_text_effect.cpp
@@ -1,0 +1,63 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file script_text_effect.cpp Implementation of ScriptTextEffect with multiplayer support. */
+
+#include "../../stdafx.h"
+#include "script_text_effect.hpp"
+#include "script_error.hpp"
+#include "script_map.hpp"
+#include "../script_instance.hpp"
+#include "../../texteff.hpp"
+#include "../../strings_func.h"
+#include "../../tile_map.h"
+#include "../../command_func.h"
+#include "../../texteff_cmd.h"
+
+#include "../../safeguards.h"
+
+/* static */ TextEffectID ScriptTextEffect::CreateAtPosition(SQInteger x, SQInteger y, Text *text, ScriptTextEffectMode mode)
+{
+    ScriptObjectRef counter(text);
+
+    EnforceDeityMode(false);
+    EnforcePrecondition(false, text != nullptr);
+    EnforcePrecondition(false, !text->GetEncodedText().empty());
+    EnforcePrecondition(false, mode == TE_RISING || mode == TE_STATIC);
+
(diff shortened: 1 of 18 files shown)


PeterN:
Not sure this would live in `GSTile` or in something new like `GSTextEffect`. I'm also not keen on Spawn instead of Create. Plus a GS might want to remove or update text as well.

npabisz:
@PeterN I've made changes in code as suggested by you to create new GS class `GSTextEffect` which implements creating, updating and removing `TextEffect`.

nielsmh:
Make sure you check the coding style guide. The automated checker is warning about some of your functions that use a wrong indentation style.

A thing to beware is that text effects with a timeout disappear and are deallocated on their own, meaning that text effects a GS created might end up disappearing before the next time the GS has time to visit it to update. Here the risk is that the GS might end up updating or removing an unrelated text effect, or referencing an invalid ID. The code should probably validate that the text effect ID is in bounds and valid before trying to update or remove one.

src/script/api/script_text_effect.cpp
```diff
@@ -0,0 +1,63 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
```
Rito13: The link to GPL 2 has been changed since you last visited to <https://www.gnu.org/licenses/old-licenses/gpl-2.0>.

src/texteff_cmd.cpp
```diff
@@ -0,0 +1,75 @@
...
+            te_id = AddTextEffect(std::move(encoded_text), pt.x, pt.y, 0, TE_STATIC);
+        }
+        
+        return { CommandCost(), te_id };
+    }
+
+    return { CommandCost(), INVALID_TE_ID };
+}
+
+CommandCost CmdUpdateTextEffect(DoCommandFlags flags, TextEffectID te_id, const EncodedString &text)
```
Rito13: It would be nice to have doxygen comment for this function. If you want, you can look into rail_cmd.cpp for inspirations.

src/texteff_cmd.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    if (flags.Test(DoCommandFlag::Execute)) {        
+        EncodedString encoded_text = text;
+        UpdateTextEffect(te_id, std::move(encoded_text));
+    }
+
+    return CommandCost();
+}
+
+CommandCost CmdRemoveTextEffect(DoCommandFlags flags, TextEffectID te_id)
```
Rito13: And also for this one.

npabisz:
Text effects now use dedicated pooled IDs, deterministic game-tick animation, save/load support, and `GSCompany::Colours`. I also tightened command/API validation and added `GSTextEffect` to the 16.0 API changelog.

PR #11673, opened 2024-01-03
Change: Don't handle 'missing' string parameters as 0.

PeterN:
## Motivation / Problem

If not enough parameters are supplied for a string, then a value of 0 was used, which could result in incorrect information being displayed.

Players may not notice there is an issue.

image

## Description

Instead, throw an exception and include an error in the string. This is non-translatable, like existing string error messages.

image

## Limitations

This is not intended as a fix for GSText compatibility, that is another issue :)

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -81,15 +81,13 @@ StringParameter *StringParameters::GetNextParameterPointer()
 {
 	assert(this->next_type == 0 || (SCC_CONTROL_START <= this->next_type && this->next_type <= SCC_CONTROL_END));
 	if (this->offset >= this->parameters.size()) {
-		Debug(misc, 0, "Trying to read invalid string parameter");
-		return nullptr;
+		throw std::out_of_range("Trying to read invalid string parameter");
 	}
 
 	auto &param = this->parameters[this->offset++];
 	if (param.type != 0 && param.type != this->next_type) {
-		Debug(misc, 0, "Trying to read string parameter with wrong type");
 		this->next_type = 0;
-		return nullptr;
+		throw std::out_of_range("Trying to read string parameter with wrong type");
 	}
 	param.type = this->next_type;
 	this->next_type = 0;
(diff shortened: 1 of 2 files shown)


src/strings.cpp
```diff
@@ -903,732 +903,736 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
...
+				}
 
-			default:
-				builder.Utf8Encode(b);
-				break;
+				default:
+					builder.Utf8Encode(b);
+					break;
+			}
+		} catch (...) {
```
LordAro: Just catch out_of_range rather than everything?

LordAro: Not sure how feasible an additional debug message might be, with additional information (exception text, at least?)

PeterN: I left the original Debug() calls in, but I guess I can remove them and do a Debug() from here.

But I don't see a way to easily get useful extra information—mostly because we're processing a sub-string, but it's the higher-level string that has the real issue.

src/strings.cpp
```diff
@@ -903,732 +901,737 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
...
-				/* raw_string can be(come) nullptr when the parameter is out of range and 0 is returned instead. */
-				if (raw_string == nullptr) {
-					builder += "(invalid RAW_STRING parameter)";
+				case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
+					const char *raw_string = args.GetNextParameterString();
+					/* raw_string can be(come) nullptr when the parameter is out of range and 0 is returned instead. */
+					if (raw_string == nullptr) {
+						builder += "(invalid RAW_STRING parameter)";
+						break;
+					}
```
glx22:
```suggestion
```
`nullptr` case is now handled by try/catch.

src/strings.cpp
```diff
@@ -903,732 +901,737 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
...
+				case SCC_STRING6:
+				case SCC_STRING7: { // {STRING1..7}
+					/* Strings that consume arguments */
+					StringID string_id = args.GetNextParameter<StringID>();
+					if (game_script && GetStringTab(string_id) != TEXT_TAB_GAMESCRIPT_START) break;
+					uint size = b - SCC_STRING1 + 1;
+					if (game_script && size > args.GetDataLeft()) {
+						builder += "(too many parameters)";
+					} else {
+						StringParameters sub_args(args, size);
```
glx22:
```suggestion
						StringParameters sub_args(args, game_script ? args.GetDataLeft() : size);
```
because here they may also use the wrong STRINGN.

PeterN: Hmm, I figured if they are using `STRINGn` they are probably doing it right... but I guess it won't hurt.

JGRennison: This mode of String parameters modifies the offset of the parent by the size given, so this isn't valid.
The size and offset adjustment need different values.

glx22: Oh right, don't touch `sub_args` but use `game_script ? args : sub_args` in the following call.
And the same will apply to SCC_STRING.

glx22: Hmm doesn't seem correct either.

JGRennison: If you do it that way you'll have to correct the offset afterwards.

glx22:
https://github.com/PeterN/OpenTTD/compare/invalid-parameters...glx22:OpenTTD:pr/11673 tested with
```
		local text = GSText(GSText.STR_TEST);
		text.SetParam(1, GSText(GSText.STR_FOO, "foo"));
		text.SetParam(2, GSWindow.TC_RED);
		text.SetParam(3, 42);
		GSTown.SetText(0, text);
```
and
```
STR_TEST :{STRING} - {COLOUR}{NUM} is supposed to be 42
STR_FOO  :'{RAW_STRING}'
```
seems to show what I (and ScriptText validation) expect
image
encoded string is 
image

PeterN:
I've removed the second commit now because it was obviously wrong, and I think it can be raised in a separate PR.

src/strings.cpp
```diff
@@ -903,733 +901,733 @@ static void FormatString(StringBuilder &builder, const char *str_arg, StringPara
...
-			case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
-				const char *raw_string = args.GetNextParameterString();
-				/* raw_string can be(come) nullptr when the parameter is out of range and 0 is returned instead. */
-				if (raw_string == nullptr) {
-					builder += "(invalid RAW_STRING parameter)";
+				case SCC_RAW_STRING_POINTER: { // {RAW_STRING}
+					const char *raw_string = args.GetNextParameterString();
+					FormatString(builder, raw_string, args);
 					break;
 				}
```
rubidium42: It could also be `nullptr` when you, for whatever reason, pass `nullptr` to `SetDParamStr`. You can currently do that, and it will also trigger this code.
The question is whether we want to game to crash here, or should we maybe add an `assert` in `SetDParamStr` where applicable?

PeterN: If it's still possible then I'll put it back to how it was I was misinformed :-)

rubidium42: Maybe replace the comment with `/* raw_string can be nullptr. */`, i.e. remove the bit about the parameter being out of range.

glx22: My bad, the comment was misleading :)

rubidium42 (approved):

PR #11675, opened 2024-01-03
Codechange: Use vehicle viewport hash to find clicked vehicle.

PeterN:
## Motivation / Problem

Clicking on a vehicle in a viewport needs to find the which vehicle was clicked on. This is done by iterating through all vehicles that exist to find the nearest one.

On a game with a lot of vehicles this can take some time (but only in the order of milliseconds.)

## Description

Use the existing vehicle viewport hash instead.

This avoids having to iterate all vehicles, which can be a minor performance improvement when there are many vehicles.

Performance measurements with the old stalwart wentbourne game:

| version | release      | debug         |
| --------| ------------ | ------------- |
| master  |   avg 1100µs (1.1ms) | avg 6500µs (6.5ms) |
| this pull request | 5-20µs | 10-30µs |

## Limitations

The code is more complex as it duplicates the viewport hash loop from `ViewportAddVehicles()`. While that could be abstracted out to allow reuse, it would then be slightly detrimental to viewport drawing performance.

This is a fairly unnecessary performance change, as a user cannot click on vehicles fast enough to notice an issue.

## Checklist for review

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -1222,21 +1222,40 @@ Vehicle *CheckClickOnVehicle(const Viewport *vp, int x, int y)
 	x = ScaleByZoom(x, vp->zoom) + vp->virtual_left;
 	y = ScaleByZoom(y, vp->zoom) + vp->virtual_top;
 
-	for (Vehicle *v : Vehicle::Iterate()) {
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):
I assume you tested it :)

PeterN:
> I assume you tested it :)

"It's so quick you don't even see the window appear..."

PR #11676, opened 2024-01-03
Codechange: Build station and depot vehicle lists from shared order lists.

PeterN:
## Motivation / Problem

When building lists of vehicles from stations and depots, we iterate all vehicles, filter to find suitable vehicles, and then check each other to see if it matches the station/depot we're listing.

This is a bit bruteforce, and means for shared orders, we are checking order lists multiple times.

## Description

Instead, build station and depot vehicle lists from shared order lists.

This brings some performance advantages:

* No need to iterate all vehicles and check for primary vehicle as only vehicles that can have orders are listed.
* Orders shared by vehicles only need to be tested once instead of for each vehicle sharing them.
* Vehicle tests only need to be performed on the first shared vehicle instead of all.

Stats from Xarick's corona test game:

| version | AI station |
| ------ | ------- |
| master | avg 3600µs | 
| this PR | avg 1200µs | 

Stats from Wentbourne:

| version | VLI station list | VLI depot list |
| -------- | -------- | ------- |
| master | avg 2100µs | avg 2000µs |
| this PR | avg 350µs | avg 350µs |

## Limitations

This is reasonable improvement, and only uses data that we already have.

## Checklist for review

### DIFF
--- a/src/script/api/script_vehiclelist.cpp
+++ b/src/script/api/script_vehiclelist.cpp
@@ -14,6 +14,7 @@
 #include "script_station.hpp"
 #include "../../depot_map.h"
 #include "../../vehicle_base.h"
+#include "../../vehiclelist_func.h"
 #include "../../train.h"
 #include "../../core/backup_type.hpp"
 #include <../squirrel/sqvm.h>
@@ -111,16 +112,12 @@ ScriptVehicleList_Station::ScriptVehicleList_Station(StationID station_id)
 
 	bool is_deity = ScriptCompanyMode::IsDeity();
 	CompanyID owner = ScriptObject::GetCompany();
-	for (const Vehicle *v : Vehicle::Iterate()) {
-		if ((v->owner == owner || is_deity) && v->IsPrimaryVehicle()) {
-			for (const Order *order : v->Orders()) {
-				if ((order->IsType(OT_GOTO_STATION) || order->IsType(OT_GOTO_WAYPOINT)) && order->GetDestination() == station_id) {
-					this->AddItem(v->index);
-					break;
-				}
-			}
-		}
-	}
+
+	FindVehiclesWithOrder(
+		[is_deity, owner](const Vehicle *v) { return is_deity || v->owner == owner; },
+		[station_id](const Order *order) { return (order->IsType(OT_GOTO_STATION) || order->IsType(OT_GOTO_WAYPOINT)) && order->GetDestination() == station_id; },
+		[this](const Vehicle *v) { this->AddItem(v->index); }
+	);
 }
 
 ScriptVehicleList_Depot::ScriptVehicleList_Depot(TileIndex tile)
@@ -162,16 +159,12 @@ ScriptVehicleList_Depot::ScriptVehicleList_Depot(TileIndex tile)
 
 	bool is_deity = ScriptCompanyMode::IsDeity();
 	CompanyID owner = ScriptObject::GetCompany();
-	for (const Vehicle *v : Vehicle::Iterate()) {
-		if ((v->owner == owner || is_deity) && v->IsPrimaryVehicle() && v->type == type) {
-			for (const Order *order : v->Orders()) {
-				if (order->IsType(OT_GOTO_DEPOT) && order->GetDestination() == dest) {
-					this->AddItem(v->index);
-					break;
-				}
-			}
-		}
-	}
+
+	FindVehiclesWithOrder(
+		[is_deity, owner, type](const Vehicle *v) { return (is_deity || v->owner == owner) && v->type == type; },
+		[dest](const Order *order) { return order->IsType(OT_GOTO_DEPOT) && order->GetDestination() == dest; },
+		[this](const Vehicle *v) { this->AddItem(v->index); }
+	);
 }
 
 ScriptVehicleList_SharedOrders::ScriptVehicleList_SharedOrders(VehicleID vehicle_id)
--- a/src/vehiclelist.cpp
+++ b/src/vehiclelist.cpp
@@ -10,6 +10,7 @@
 #include "stdafx.h"
 #include "train.h"
 #include "vehiclelist.h"
+#include "vehiclelist_func.h"
 #include "group.h"
 
 #include "safeguards.h"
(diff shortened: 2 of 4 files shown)


2TallTyler (dismissed):
Seems reasonable to me.

src/script/api/script_vehiclelist.cpp
```diff
@@ -107,11 +107,17 @@ ScriptVehicleList_Station::ScriptVehicleList_Station(StationID station_id)
...
 
-	for (const Vehicle *v : Vehicle::Iterate()) {
-		if ((v->owner == ScriptObject::GetCompany() || ScriptCompanyMode::IsDeity()) && v->IsPrimaryVehicle()) {
-			for (const Order *order : v->Orders()) {
+	for (const OrderList *orderlist : OrderList::Iterate()) {
+		const Vehicle *v = orderlist->GetFirstSharedVehicle();
+		/* We assume all vehicles sharing an order list have the same owner. */
+		if ((v->owner == ScriptObject::GetCompany() || ScriptCompanyMode::IsDeity())) {
+			/* Vehicle is a candidate, search for a matching order. */
+			for (const Order *order = orderlist->GetFirstOrder(); order != nullptr; order = order->next) {
```
SamuXarick:
```suggestion
			for (const Order *order : v->Orders()) {
```
Isn't this faster?

PeterN: Well, you tell me?

rubidium42:
I'm seeing four times essentially the same loop, with only small differences in conditionals at different places. It looks to me like this can be written in a single templated function, with some small functions for the differentiating logic. Like whether to filter on company or not, what type of order is allowed, and how to add to a collection.

That would split the internals of the (complicated) loops from the actual important bits for each of the instances. It's bit similar to something like [`std::copy_if`](https://en.cppreference.com/w/cpp/algorithm/copy), which has an `UnaryPredicate` to separate the `if` logic from the looping logic.

src/script/api/script_vehiclelist.cpp
SamuXarick: It isn't, nevermind, sorry...
image

PeterN:
> I'm seeing four times essentially the same loop, with only small differences in conditionals at different places. It looks to me like this can be written in a single templated function, with some small functions for the differentiating logic. Like whether to filter on company or not, what type of order is allowed, and how to add to a collection.
> 
> That would split the internals of the (complicated) loops from the actual important bits for each of the instances. It's bit similar to something like [`std::copy_if`](https://en.cppreference.com/w/cpp/algorithm/copy), which has an `UnaryPredicate` to separate the `if` logic from the looping logic.

I had a play with this and ended up with the function like this.

```
template <class VehiclePredicate, class OrderPredicate, class VehicleFunc>
void FindVehiclesWithOrder(VehiclePredicate vp, OrderPredicate op, VehicleFunc vf) {
...
}
```

This seems to be at least about 50% slower. Not sure if I'm doing it wrong, or if it's because it needs to capture variables.

glx22:
> And here's a good thing. Making locals out of `ScriptObject::GetCompany()` and `ScriptCompanyMode::IsDeity()` made another performance difference:

Maybe do the same for `ScriptVehicleList` too.

PeterN:
Ah, good point!

src/script/api/script_vehiclelist.cpp
```diff
@@ -156,14 +162,20 @@ ScriptVehicleList_Depot::ScriptVehicleList_Depot(TileIndex tile)
...
-				if (order->IsType(OT_GOTO_DEPOT) && order->GetDestination() == dest) {
-					this->AddItem(v->index);
-					break;
-				}
+	Owner owner = ScriptObject::GetCompany();
+	bool is_deity = ScriptCompanyMode::IsDeity();
+	for (const OrderList *orderlist : OrderList::Iterate()) {
+		const Vehicle *v = orderlist->GetFirstSharedVehicle();
+		/* We assume all vehicles sharing an order list have the same owner. */
+		if ((!is_deity && v->owner != owner) || v->type != type) continue;
```
TrueBrain: If you like, you can make this in two `continue` statements. Not sure it matters.

src/vehiclelist.cpp
```diff
@@ -116,15 +116,18 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
...
-					for (const Order *order : v->Orders()) {
-						if ((order->IsType(OT_GOTO_STATION) || order->IsType(OT_GOTO_WAYPOINT) || order->IsType(OT_IMPLICIT))
-								&& order->GetDestination() == vli.index) {
-							list->push_back(v);
-							break;
-						}
+			for (const OrderList *orderlist : OrderList::Iterate()) {
+				const Vehicle *v = orderlist->GetFirstSharedVehicle();
+				if (v->type != vli.vtype) continue;
+				/* Vehicle is a candidate, search for a matching order. */
```
TrueBrain: Bit of whitespace goes a long way here. Just a newline just before this comment makes already a visual break from what is going on, and makes it less blobby. Pretty sure with some well times newlines in other parts we increase readability :)

PeterN:
Performance difference since #11686:

dbg: [misc] [old] 207814 us [avg: 2078.1 us]
dbg: [misc] [new] 17636 us [avg: 176.4 us]
dbg: [misc] [old] 193357 us [avg: 1933.6 us]
dbg: [misc] [new] 16664 us [avg: 166.6 us]
dbg: [misc] [old] 203591 us [avg: 2035.9 us]
dbg: [misc] [new] 17441 us [avg: 174.4 us]
dbg: [misc] [old] 192353 us [avg: 1923.5 us]
dbg: [misc] [new] 16598 us [avg: 166.0 us]
dbg: [misc] [old] 202319 us [avg: 2023.2 us]
dbg: [misc] [new] 17576 us [avg: 175.8 us]

But please double check I haven't got the conditions wrong :)

src/vehiclelist.cpp
```diff
@@ -161,14 +166,19 @@ bool GenerateVehicleSortList(VehicleList *list, const VehicleListIdentifier &vli
...
-							list->push_back(v);
-							break;
-						}
+			for (const OrderList *orderlist : OrderList::Iterate()) {
+				const Vehicle *v = orderlist->GetFirstSharedVehicle();
+				if (v->type != vli.vtype) continue;
+
+				/* Vehicle is a candidate, search for a matching order. */
+				for (const Order *order = orderlist->GetFirstOrder(); order != nullptr; order = order->next) {
+					if (!(order->IsType(OT_GOTO_DEPOT) && !(order->GetDepotActionType() & ODATFB_NEAREST_DEPOT) && order->GetDestination() == vli.index)) continue;
```
TrueBrain: `if (!order->IsType(OT_GOTO_DEPOT) || (order->GetDepotActionType() & ODATFB_NEAREST_DEPOT) || order->GetDestination() != vli.index)) continue;` ? Could also be 3 if-statements. Not sure that matters.

PeterN: If makes it easier to read it's worth it.

rubidium42 (approved):
In the category of massive bike shedding... but if you're looking for optimal performance, some things that might help:
* move the `order->GetDestination` check to before the `order->IsType` checks. For an order it's generally way more likely that the destination ID differs from what you want, than whether it's station vs depot, as such the destination check drops way more and that saves the two or three IsType expressions. Technically `order->GetDestination` requires the order to be one of a few specific types, but that isn't checked and I think the documentation of `order->GetDestination` is not even listing all of them.
* when is_deity is true, you can make a separate call where the vehicle predicate just returns true, which ought to remove the whole predicate with minor optimisations; then for the other branch you only test for owner. It's more code, but fewer comparisons, so it might make a difference.

Of both I'm definitely not saying that we should do it, but rather that it could be done and could improve performance even more. Even though I do not think this code is called that often.

PeterN:
Splitting the deity check seems to save on average about 20µs (so about 10%) in the stationlist case.

PR #11682, opened 2024-01-04
Feature: Setting to disable loading speed penalty for trains at short platforms

EmperorJake:
## Motivation / Problem

Trains get a penalty to loading speed if the platform is shorter than the train. There is currently no way to get around this, however there are several use cases why a player might want to do so:

- Underground stations where the rest of the platform is "hidden" inside the tunnel or on a bridge. This applies to eyecandy players who might do creative things with overlapping station tiles and such.
- Saving space by loading trains in curved tracks rather than one long platform. Similar to "through loading" as found in JGRPP, but without the extra train movement.
- Saving infrastructure costs by not needing as many station tiles. This applies to min-maxer players.

There have been various patches in the past to the same effect. Some have added a cheat, others have straight up deleted the code responsible for the loading speed penalty. This feature has been in demand by a subsection of the playerbase for a long time now, it's time it was made a proper setting.

## Description

This PR adds a setting which allows the player to disable the loading speed penalty for trains at shorter platforms.

The setting is an "Expert" setting as it has relatively niche uses and new players don't really need to see/change it.

The setting is under the "Limitations" category as this is more in line with https://github.com/OpenTTD/OpenTTD/pull/11683

## Limitations

The setting does not affect the train pathfinder avoiding short platforms. But that's a good thing as it would mess up many rail layouts otherwise.

## Checklist for review


src/economy.cpp
```diff
@@ -1612,7 +1613,8 @@ static void ReserveConsist(Station *st, Vehicle *u, CargoArray *consist_capleft,
  */
 static void UpdateLoadUnloadTicks(Vehicle *front, const Station *st, int ticks)
 {
-	if (front->type == VEH_TRAIN) {
+	if (_cheats.loading_speed.value) {}		
+	else if (front->type == VEH_TRAIN) {
```
2TallTyler: Try this instead of the `if`/`else if` silliness. 😄 
```suggestion
	if (front->type == VEH_TRAIN && !_cheats.loading_speed.value) {
```

While I'm at it, a little C++ lesson for you:

We use `&&` (logical AND) and `||` (logical OR) a lot in OpenTTD. These, and using [early returns](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement) instead of `if`/`else` blocks, are two of the simplest and most common tools you should get used to reaching for in your toolbelt.

It's worth noting that these logical operators are short-circuiting: in my suggestion, if `front->type == VEH_TRAIN` is false, we skip past the block without even checking the cheat value. The opposite is true for `||`, where if the first condition is true we enter the block without checking the second condition. 

In this case, the order of the conditions doesn't matter. But sometimes they do, like if we're checking `IsThingValid() && ValueOfThing() == stuff` where trying to get `ValueOfThing()` of an invalid thing would cause a crash. The short-circuit makes it safe. The other use for choosing a specific order is if one check is computationally cheap and another is more expensive; we can avoid checking the expensive condition unnecessarily by placing it second.

src/saveload/saveload.h
```diff
@@ -367,6 +367,8 @@ enum SaveLoadVersion : uint16_t {
 	SLV_TIMETABLE_START_TICKS_FIX,          ///< 322  PR#11557 Fix for missing convert timetable start from a date to ticks.
 	SLV_TIMETABLE_TICKS_TYPE,               ///< 323  PR#11435 Convert timetable current order time to ticks.
 
+	SLV_LOADING_SPEED_CHEAT,                ///< 324   Add cheat to remove loading speed penalty for short platforms.
```
2TallTyler:
```suggestion
	SLV_LOADING_SPEED_CHEAT,                ///< 324  PR#11682 Add cheat to remove loading speed penalty for short platforms.
```

2TallTyler (commented):
I'm not sure how I feel about this. It feels like a very niche feature. That said, I agree that having it as a cheat would be the proper way to do it.

You'll notice that commit checker has failed for invalid tab usage. We only use tabs for the leftmost indentation. Any spacing/indentation besides that is created with spaces.

src/economy.cpp
EmperorJake: I almost had it like that, I just used a comma instead of && because I'm used to NML (it obviously didn't work) :)

EmperorJake:
I agree that it's a fairly niche feature, but I think there is a small but important part of the playerbase who would greatly appreciate this feature. If the developers feel that it's still too niche, perhaps it would be better ported to JGRPP.

ldpl:
I totally disagree with it being a cheat, it achieves nothing but making this setting inconvenient to use (can't change in mp etc.).

EmperorJake:
That's a valid point, it would work better as a setting. However, in the past devs have voiced concerns about "significant gameplay implications" (see https://github.com/JGRennison/OpenTTD-patches/issues/330) so I thought it would be more palatable as a cheat.

ldpl:
Cargodist, breakdowns and build on pause have much more significant "gameplay implications". This just allows you to build one station tile instead of TL, big deal. Also, you can already unload train on a small station quickly by reversing it a few times.

EmperorJake:
I just found a patch to the same end on a forum thread from over 10 years ago. This goes to show that this feature has been in demand for a very long time https://www.tt-forums.net/viewtopic.php?p=1074153

PeterN:
> That's a valid point, it would work better as a setting. However, in the past devs have voiced concerns about "significant gameplay implications" (see [JGRennison/OpenTTD-patches#330](https://github.com/JGRennison/OpenTTD-patches/issues/330)) so I thought it would be more palatable as a cheat.

I don't see any vanilla dev's opinions on that link...

I agree (shock horror) with @ldpl here.

src/settings_type.h
```diff
@@ -501,6 +501,7 @@ struct OrderSettings {
 	bool   selectgoods;                      ///< only send the goods to station if a train has been there
 	bool   no_servicing_if_no_breakdowns;    ///< don't send vehicles to depot when breakdowns are disabled
 	bool   serviceathelipad;                 ///< service helicopters at helipads automatically (no need to send to depot)
+	bool   loading_speed_penalty;            ///< Loading speed penalty for trains longer than the station
```
2TallTyler: `loading_speed_penalty` is the effect, not the cause. What about:

```suggestion
	bool   station_length_loading_penalty;            ///< make trains longer than the station load more slowly
```

src/lang/english.txt
```diff
@@ -1777,6 +1777,9 @@ STR_CONFIG_SETTING_SERVINT_DISABLED                             :Disabled
 STR_CONFIG_SETTING_NOSERVICE                                    :Disable servicing when breakdowns set to none: {STRING2}
 STR_CONFIG_SETTING_NOSERVICE_HELPTEXT                           :When enabled, vehicles do not get serviced if they cannot break down
 
+STR_CONFIG_SETTING_LOADING_SPEED_PENALTY                        :Loading speed penalty for trains that are longer than the station: {STRING2}
+STR_CONFIG_SETTING_LOADING_SPEED_PENALTY_HELPTEXT               :Disabling this setting allows trains to load/unload at maximum speed regardless of station length
```
```suggestion
STR_CONFIG_SETTING_LOADING_SPEED_PENALTY_HELPTEXT               :When enabled, trains which are too long for the station load more slowly than a train which fits the station. This setting does not affect pathfinding.
```

2TallTyler (commented):
Okay, your comments here and this morning's Discord discussion have changed my mind. I like this as a setting. 🙂 

Oh, and you'll want to squash your commits.

2TallTyler (approved):

rubidium42:
Is this setting properly saved in the save game? As far as I can see it's not; at least, I'm not seeing a save game version and such.

JGRennison:
The settings table doesn't require a savegame bump for new settings.
It isn't otherwise necessary to bump to savegame version because there isn't any new/changed state which prior versions couldn't understand.

EmperorJake:
I tested saving and loading a game, and the state of the setting got saved too. 

The original cheat version of the PR had some saveload code, but I assumed that was due to cheats being recorded by the save. I copied a different PR that added a setting and it didn't come with saveload code either.

PR #11639, opened 2023-12-29
Fix: race-conditions in GUI updates when downloading HTTP files

TrueBrain:
## Motivation / Problem

Fixes #11636.

HTTP downloads, both on WIndows and Linux, are processed in a thread. This thread calls `HTTPCallback` object with updates, but these are executed from that download-thread. However, that code assumes it is called from the OpenTTD Game thread, and updates things on the GUI. This turns out to be a bit of a mistake, and causes all kinds of (small) race-conditions.

## Description

Put a bit of glue between the `HTTPCallback` and the HTTP downloaders, called `HTTPThreadSafeCallback`. This Thread-Safe callback puts the ReceiveData and Failure information on a queue, which is dequeue from the OpenTTD Game thread. That way, thread-safety is restored again, and in result, the Thread Analyzer no longer finds any mistakes.

Sadly, due to how the current code is designed, this gives the need for a bit of mutex magic, to ensure things can be added and removed in moments that are safe to do so. The main issue here is that a callback can cause a new HTTP request to be created, which makes it a tiny bit more messy than I would like.

One fundamental change required here, is that `OnReceiveData` becomes the owner of `data`. As otherwise the HTTP downloader would already release the buffer before it could have been processed. On Windows, we already allocated a buffer ourselves. On Linux, this requires allocating another (temporary) buffer, and copying the content. A tiny bit of a slow-down, but as HTTP happens rarely, not something anyone should notice.

## Limitations

There is another problem here, that I didn't dare to address. `NetworkContent`'s `OnReceiveData` is .. special. It really became hard to read and hard to understand. In an ideal world, we would rewrite that piece of code, making that `OnReceiveData` thread-safe. That would be much more efficient. But, that also increases the chances of introduces new bugs. So I went for this approach to avoid that rabbit hole. The `NetworkContent`'s `OnReceiveData` is not broken; just .. hard to make thread-safe in its current state.

(for those who don't want to open up the code: it is built under the assumption the first request is always to the BaNaNaS mirror redirect service, which returns the URLs of all the files to download; the 2nd and next requests are then one by one an URL from that list. Instead that this is two HTTP handlers, one simply writing a file, the other handling the list, this is build in a single function with some nasty complexity).

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
[...]

PR #11687, opened 2024-01-05
Fix: don't unneededly block on transmitting survey on exit

TrueBrain:
## Motivation / Problem

The game always has a 2 second delay when exiting, because the survey is sending (and the result is not being polled).

## Description

On sending a blocking survey, keep running the network background thread. As since #11639 this isn't done in the curl-thread anymore, so we need to pump the async callbacks for new data.

## Checklist for review

### DIFF
--- a/src/network/network_survey.cpp
+++ b/src/network/network_survey.cpp
@@ -11,6 +11,7 @@
 #include "network_survey.h"
 #include "settings_table.h"
 #include "network.h"
+#include "network_func.h"
 #include "../debug.h"
 #include "../survey.h"
 #include "../3rdparty/fmt/chrono.h"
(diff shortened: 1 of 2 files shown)


PeterN:
It is doesn't block, what happens to it? Do we just fire it off and not care about a response?

TrueBrain:
> It is doesn't block, what happens to it? Do we just fire it off and not care about a response?

When you quit a game (so go back to the menu), it is fired in the background. The assumption is that you won't be able to close the game before it is done sending the survey.

Only when you exit the game, the blocked is set, and it waits a small amount of time to ensure the survey is transmitted.

So normally it is indeed fire&forget, except for when you exit the game; then we block :)

PeterN (approved):
Seems okay here.

PR #11691, opened 2024-01-05
Add: Latvian Lats currency

kputnins:
## Motivation

- There exists localisations for all 3 of the Baltic states - Estonia, Latvia and Lithuania
- There exists currencies for only 2 of the Baltic states - Estonia (Estonian Krooni) and Lithuania (Lithuanian Lits)
- The Latvian Lats is missing

The 3 Baltic countries are usually represented together in various geographical, historical and political contexts so it feels very strange seeing EEK, LTL, but no LVL.

Judging that the [Latvian localisation on translator](https://translator.openttd.org/language/lv_LV) is fairly complete, there exists a section of players who would enjoy this

## Description

Adds the Latvian Lats currency to the game with the value of 1 GBP (real life value before adopting euro was 1.22 GBP)

## Limitations

Translations for the currency will need to be provided in other localisations

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -959,7 +959,7 @@ STR_GAME_OPTIONS_CURRENCY_UNITS_DROPDOWN_TOOLTIP                :{BLACK}Currency
 
 STR_GAME_OPTIONS_CURRENCY_CODE                                  :{STRING} ({RAW_STRING})
 
-###length 42
+###length 43
 STR_GAME_OPTIONS_CURRENCY_GBP                                   :British Pound
 STR_GAME_OPTIONS_CURRENCY_USD                                   :American Dollar
 STR_GAME_OPTIONS_CURRENCY_EUR                                   :Euro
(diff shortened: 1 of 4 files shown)


rubidium42:
The currency needs to be added after the "Malaysian Ringgit", since the index into the array is stored in the save games. By adding it in the middle, you change the currency for many save games.

Furthermore in english.txt there is a counter just before STR_GAME_OPTIONS_CURRENCY_GBP that would need to be incremented.

kputnins:
@rubidium42 thanks for review - I updated the order and incremented the length counter

src/lang/english.txt
```diff
@@ -988,6 +988,7 @@ STR_GAME_OPTIONS_CURRENCY_TRY                                   :Turkish Lira
 STR_GAME_OPTIONS_CURRENCY_SKK                                   :Slovak Koruna
 STR_GAME_OPTIONS_CURRENCY_BRL                                   :Brazilian Real
 STR_GAME_OPTIONS_CURRENCY_EEK                                   :Estonian Krooni
+STR_GAME_OPTIONS_CURRENCY_LVL                                   :Latvian Lats
```
glx22: Order of the strings should also match :)

kputnins: fixed string order 🤦🏼‍♂️

TrueBrain (approved):

PR #11695, opened 2024-01-06
Codechange: replace static inline with static for non-class functions

rubidium42:
## Motivation / Problem

Looking at the region based pathfinder I wondered why it used `inline` for functions we would usually `static inline`.

It turns out that a lot of `static inline`, especially in headers, is to please the C linker not complaining about duplicate definitions. With C++ only `inline` in the headers is fine for non-struct/class functions; actually, it is kind of encouraged. Why? Because the `inline` is only a wish that the compiler may ignore. However, the linker at the end only have one non-inlined version of the code.
With `static inline` the static will make it local to the translation unit, and as such linking is not allowed to merge the same functions. The result is a larger binary.

On my Linux machine a debug build is roughly 500 kB smaller (on 300 MB) and a stripped release build 20 kB (on 15 MB). So fairly insignificant in the grand scheme of things, but it is a C-thing that transcended into our C++-code.

## Description

Remove `static` from `static inline` functions that are not part of a class and are in a header.

## Limitations

I left `static inline` in the non-header files, even though they could technically be removed. This is more as the static in there also denoted that it's local to the file and as such should not get external linkage.

## Checklist for review


TrueBrain (approved):
We are really going for the big numbers lately :D

PR #10121, opened 2022-11-01
Codechange: rename and move strings to make them more consistent

rubidium42:
## Motivation / Problem

We've got english.txt which defines all the strings. The usage of some strings has changed over time, so their label/StringID and/or location in the file is not in line with the rest of the code. See #9101.

## Description

Just the names and location in english.txt of some strings has changed. The renames have also been applied to the other languages and the code itself.

## Limitations

I did not implement TrueBrain's suggestion from https://github.com/OpenTTD/OpenTTD/discussions/9101#discussioncomment-1261331 as that is quite complicated to get right and not really in the scope of the discussion.

I also changed the translations in the hope that the translator does the right thing, i.e. understands that they are renames of the strings and they do not need to be translated again.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11699, opened 2024-01-07
Fix: Use consistent wording for lorry stations

2TallTyler:
## Motivation / Problem

In OpenTTD, a lorry station (truck station) is usually called just that, except for three strings which call it a "lorry loading bay" or a "lorry loading area."

This is inconsistent and confusing, as I witnessed this morning trying to teach a new player who actually reads the tooltips.

## Description

Let's use "lorry station" consistently.

String names are left alone to mark translations as outdated, but not delete them.

## Limitations

Does not address the inconsistency of every string name referring to lorries as trucks. 😄 

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -771,7 +771,7 @@ STR_SMALLMAP_LEGENDA_AIRCRAFT                                   :{TINY_FONT}{BLA
 STR_SMALLMAP_LEGENDA_TRANSPORT_ROUTES                           :{TINY_FONT}{BLACK}Transport Routes
 STR_SMALLMAP_LEGENDA_FOREST                                     :{TINY_FONT}{BLACK}Forest
 STR_SMALLMAP_LEGENDA_RAILROAD_STATION                           :{TINY_FONT}{BLACK}Railway Station
-STR_SMALLMAP_LEGENDA_TRUCK_LOADING_BAY                          :{TINY_FONT}{BLACK}Lorry Loading Bay
+STR_SMALLMAP_LEGENDA_TRUCK_LOADING_BAY                          :{TINY_FONT}{BLACK}Lorry Station
 STR_SMALLMAP_LEGENDA_BUS_STATION                                :{TINY_FONT}{BLACK}Bus Station
 STR_SMALLMAP_LEGENDA_AIRPORT_HELIPORT                           :{TINY_FONT}{BLACK}Airport/Heliport
 STR_SMALLMAP_LEGENDA_DOCK                                       :{TINY_FONT}{BLACK}Dock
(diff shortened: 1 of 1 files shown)


src/lang/english.txt
```diff
@@ -2835,7 +2835,7 @@ STR_BUILD_DEPOT_TRAM_ORIENTATION_SELECT_TOOLTIP                 :{BLACK}Select t
 STR_STATION_BUILD_BUS_ORIENTATION                               :{WHITE}Bus Station Orientation
 STR_STATION_BUILD_BUS_ORIENTATION_TOOLTIP                       :{BLACK}Select bus station orientation
 STR_STATION_BUILD_TRUCK_ORIENTATION                             :{WHITE}Lorry Station Orientation
-STR_STATION_BUILD_TRUCK_ORIENTATION_TOOLTIP                     :{BLACK}Select lorry loading bay orientation
+STR_STATION_BUILD_TRUCK_ORIENTATION_TOOLTIP                     :{BLACK}Select lorry station orientation
```
TrueBrain, line 2838: Your description says one thing, and your patch does another.

But do we actually want to remove translations? Is that actually worth it? I mean .. translations that won't get to updating it, will have an english text all of a sudden, and possibly in most translations it is already translated consistently. But with this rename, it gets entered as a new string, and not an updated, losing that translation.

I don't have an answer, just a wonder: do we actually want to do this? But most of all, if you claim you rename to invalidate translations, do it consistently, and not 3 out of 4 :) That makes it looks like you weren't actually after invalidating, but you just wanted the STR name to be "correct" :)

TrueBrain: Just to be clear: if you do not rename the STR, translators will see the english text was updated, and they need to reconsider their translation. BUT. The old translation remains there.

What you do with the other 3 strings will remove the translation completely, forcing it back to english, until they retranslated it again.

2TallTyler: Yes, this one doesn't get deleted but will be flagged as Outdated.

The ideal outcome is Outdated for all three, but the only way to do that is to leave the string names alone, which feels sloppy to leave for someone to clean up later...

So yes, I do want the STR name to be "correct." 🙂

TrueBrain: So for your own OCD, you will make the life of .. 30+ people harder? Seems like the wrong balance :)

Many STR strings get out-of-sync with their translation over time. For 13.0 we did this:
https://github.com/OpenTTD/OpenTTD/pull/10121

That might be the better approach. Not in this PR. But just a PR to rename everything we consider "wrong". Worked last time at least. And punishes our translators a lot less.

2TallTyler: Ah, so if we change string values and names in two steps with an eints run in the middle, we can avoid deleting them? Sounds ideal!

Also, you taught me to be OCD! 😛

TrueBrain: No, eints has little to do with that. With #10121, all translations were renamed too. The difference with the approach is:

Translators will have a few weeks / months to see the english text changed, and can make adjustments if needed. We then rename the string for them, and eints is like: owh new string, how lovely, and adds it to the list of "maybe look at this", without history. So translators wouldn't know why it is there. Which is fine, as we can tell them: yo, we renamed these N strings.

More the point is: in this PR, you just want to fix the english text. Not the STR. That is our internal bookkeeping. As I said on Discord:

Making user-facing things consistent: good idea
Making things for us consistent but giving translators a lot of work: bad idea (imo)

We can live with "LORRY_STATION"; we know what it means. And if we have a list long enough to do a mass renaming, sure, go for it :)

https://github.com/OpenTTD/OpenTTD/discussions/9101 was an excellent idea to track those strings btw. We might want to start one up again, for 15.0 or something.

2TallTyler: Ah, thanks.

If/when this gets merged, I will start a Discussion for 15.0 and add these strings.

TrueBrain (approved):

PeterN:
But would "Lorry loading bay" consistently, make more sense?

2TallTyler:
No, IMO, because anybody new enough to be reading tooltips is going to be trying to determine the difference between a Bus Station and a Lorry Station. Both should be Stations to allow them to figure it out. I've taught several new players and they are often confused by Depots, so I think it's important to have the toolbar show A/B1/B2 and not A/B/C.

PR #11700, opened 2024-01-07
Fix: Use consistent wording on waypoint button tooltip

2TallTyler:
## Motivation / Problem

The tooltip to open the waypoint picker says `Convert rail to waypoint.`

Rail does not get converted. You just build a waypoint on top of the track, just like building a signal.

## Description

@TrueBrain had concerns (see below) about not being able to build waypoints or signals without underlying track, so we found a new wording and I also updated the signal tooltip to match.

## Checklist for review


src/lang/english.txt
```diff
@@ -2720,7 +2720,7 @@ STR_RAIL_TOOLBAR_MAGLEV_CONSTRUCTION_CAPTION                    :Maglev Construc
 STR_RAIL_TOOLBAR_TOOLTIP_BUILD_RAILROAD_TRACK                   :{BLACK}Build railway track. Ctrl toggles build/remove for railway construction. Shift toggles building/showing cost estimate
 STR_RAIL_TOOLBAR_TOOLTIP_BUILD_AUTORAIL                         :{BLACK}Build railway track using the Autorail mode. Ctrl toggles build/remove for railway construction. Shift toggles building/showing cost estimate
 STR_RAIL_TOOLBAR_TOOLTIP_BUILD_TRAIN_DEPOT_FOR_BUILDING         :{BLACK}Build train depot (for buying and servicing trains). Shift toggles building/showing cost estimate
-STR_RAIL_TOOLBAR_TOOLTIP_CONVERT_RAIL_TO_WAYPOINT               :{BLACK}Convert rail to waypoint. Ctrl enables joining waypoints. Shift toggles building/showing cost estimate
+STR_RAIL_TOOLBAR_TOOLTIP_CONVERT_RAIL_TO_WAYPOINT               :{BLACK}Build railway waypoint. Ctrl enables joining waypoints. Shift toggles building/showing cost estimate
```
TrueBrain: It is a bit weird (but so are signals): you can't actually build them without first building a railway.

Now we can discuss if that actually make sense: why can't I just build a waypoint, but why can I build a station. It makes little sense, this inconsistency.

So the original text was more correct .. but in that sense, the signal text was wrong. So having it consistent makes sense. But it no longer suggests you first need a railway. But neither did signals.

It is a mess. Not your fault :)

I really don't care or mind either way btw; just an observation it is messy either way :)

2TallTyler: I don't think "build" infers that you can build it by itself. You can build a roof on a house, right?

"Add" might make sense, but I still prefer "Build."

TrueBrain: Ironically, you indeed say you build a roof ON a house. If you say: I am building a roof, it is a bit confusing .. you are building just a roof? Are you going to put it on the house later? So yeah, great example to highlight how weird this is :D

TrueBrain: Maybe it should be:

`Build waypoint on railway`

And also

`Build signal on railway`

Or, what I won't be opposed to, allow building a waypoint without having a railway first. Not for this PR. But again, it is just weird :)

PS: why is it `signals` but `station`, `waypoint`, etc? What made signals so special they have to be plural? Also weird :D

2TallTyler: Ask and ye shall receive. 🙂

TrueBrain (approved):

PR #11717, opened 2024-01-07
Change: Rewrite a few main toolbar tooltips

2TallTyler:
## Motivation / Problem

My OCD has a few problems with tooltips for the main toolbar:

1. Buttons never `Show` or `Display` a menu, they `Open` it
2. Tooltips should list the contents of its dropdown
3. Construction menus should name the transport category, not just the network type (so `Build railway infrastructure`, not just `Build railway track`)
4. Zoom in / Zoom out need not be so wordy 😉 

## Description

Rewrite strings accordingly.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -370,21 +370,21 @@ STR_TOOLBAR_TOOLTIP_PAUSE_GAME                                  :{BLACK}Pause ga
 STR_TOOLBAR_TOOLTIP_FORWARD                                     :{BLACK}Fast forward the game
 STR_TOOLBAR_TOOLTIP_OPTIONS                                     :{BLACK}Options and settings
 STR_TOOLBAR_TOOLTIP_SAVE_GAME_ABANDON_GAME                      :{BLACK}Save, load or abandon game, exit program
-STR_TOOLBAR_TOOLTIP_DISPLAY_MAP                                 :{BLACK}Display map, extra viewport, cargo flow or list of signs
-STR_TOOLBAR_TOOLTIP_DISPLAY_TOWN_DIRECTORY                      :{BLACK}Display town directory
-STR_TOOLBAR_TOOLTIP_DISPLAY_SUBSIDIES                           :{BLACK}Display subsidies
-STR_TOOLBAR_TOOLTIP_DISPLAY_LIST_OF_COMPANY_STATIONS            :{BLACK}Display list of company's stations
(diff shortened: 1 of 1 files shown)


src/lang/english.txt
```diff
@@ -380,21 +380,21 @@ STR_TOOLBAR_TOOLTIP_DISPLAY_STORY_BOOK                          :{BLACK}Display
 STR_TOOLBAR_TOOLTIP_DISPLAY_GOALS_LIST                          :{BLACK}Display goal list
 STR_TOOLBAR_TOOLTIP_DISPLAY_GRAPHS                              :{BLACK}Display company graphs and cargo payment rates
 STR_TOOLBAR_TOOLTIP_DISPLAY_COMPANY_LEAGUE                      :{BLACK}Display company league table
-STR_TOOLBAR_TOOLTIP_FUND_CONSTRUCTION_OF_NEW                    :{BLACK}Examine industries or fund construction of a new industry
+STR_TOOLBAR_TOOLTIP_FUND_CONSTRUCTION_OF_NEW                    :{BLACK}Display industry directory
```
rubidium42: `STR_TOOLBAR_TOOLTIP_DISPLAY_MAP`, `STR_TOOLBAR_TOOLTIP_SHOW_LAST_MESSAGE_NEWS` and `STR_TOOLBAR_TOOLTIP_LAND_BLOCK_INFORMATION` essentially list all options. Why should this tooltip just show one of the three options?

2TallTyler: Ah, maybe the right answer is to make that the standard here. 😄

rubidium42 (dismissed):

nielsmh:
On the other hand, having the most actionable verb first in the tooltip makes the difference between the construction window buttons and the vehicle list buttons more obvious. It's not a button for initiating the display of something, it's a button for initiating the building process.

nielsmh:
Another thing, my intuition is that it's more common to say that a button "opens a window" rather than "displays a window", even if the latter is technically more correct since it can also just bring an already-open window to front (ensuring it is visible).

2TallTyler:
I agree with "open" versus "display," although I wasn't sure if that was American English so I went with the common "display" instead of introducing a new word.

If we wanted to emphasize an action verb, we could say `Build railway infrastructure` and similar. That said, button tooltips tend to indicate what that button does and not why you'd click it. In that sense, these buttons really do just display the construction menus.

rubidium42 (approved):

PR #11719, opened 2024-01-08
Change: Cargo type is label

PeterN:
## Motivation / Problem

`CargoType` is landscape specific, so values change depending on which climate is selected, and it is used for default cargos of engines, industries and houses.

These default cargo types are not mapped at all, so if a NewGRF disables a cargo slot, or moves it from one slot to another, incorrect or invalid cargo can be produced.

There are 4 ways to identify cargo types:

- CargoType - climate specific cargo types, up to 12 per climate. Some climates have holes.
- CargoLabel - FourCC labels, defined for default cargo types, and used for setting up cargo for each climate, but otherwise only used by NewGRFs.
- bitnum - obsolete, nobody cares, except 0xFF is used to disable a cargo slot.
- CargoID - game-dependent cargo types, up to 64 cargo types, varies depending on newgrf configuration.

This is too many.

## Description

This PR completely removes CargoType. Default engines, industries and houses are defined in terms of their cargo label which, for convenience, are defined using the same names as the CargoType enum, e.g. instead of CT_PASSENGERS = 0, CT_PASSENGERS = 'PASS'.

This means that passengers no longer has to be defined as slot 0, nor mail as slot 2. Core cargo types can now be moved or undefined without causing invalid cargo to appear.

This is built on top of #11378, which already removes hardcoded town cargo types.

## Limitations

Possibly breaks some ancient cargo NewGRFs that don't define labels?

## Checklist for review


src/cargo_type.h
```diff
@@ -61,14 +61,15 @@ enum CargoType {
...
-	CT_AUTO_REFIT   = 0xFD, ///< Automatically choose cargo type when doing auto refitting.
-	CT_NO_REFIT     = 0xFE, ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
 	CT_INVALID      = 0xFF, ///< Invalid cargo type.
 };
 
+static const CargoID NUM_ORIGINAL_CARGO = 12; ///< Original number of cargo types.
+static const CargoID NUM_CARGO = 64; ///< Maximal number of cargo types in a game.
+
+static const CargoID CARGO_AUTO_REFIT = 0xFD; ///< Automatically choose cargo type when doing auto refitting.
+static const CargoID CARGO_NO_REFIT = 0xFE; ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
```
2TallTyler: What is `auto-new`? Typo of `auto-renew`?

src/cargo_type.h
```diff
@@ -61,14 +61,15 @@ enum CargoType {
...
-	NUM_ORIGINAL_CARGO = 12,
-	NUM_CARGO       = 64,   ///< Maximal number of cargo types in a game.
-
-	CT_AUTO_REFIT   = 0xFD, ///< Automatically choose cargo type when doing auto refitting.
-	CT_NO_REFIT     = 0xFE, ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
 	CT_INVALID      = 0xFF, ///< Invalid cargo type.
 };
 
+static const CargoID NUM_ORIGINAL_CARGO = 12; ///< Original number of cargo types.
+static const CargoID NUM_CARGO = 64; ///< Maximal number of cargo types in a game.
```
2TallTyler: Not your code, but I mostly see `maximum` instead of `maximal`. It's technically fine, but looks odd. 🙂

src/industry_gui.cpp
```diff
@@ -1247,8 +1247,8 @@ typedef GUIList<const Industry *, const CargoID &, const std::pair<CargoID, Carg
 
 /** Special cargo filter criteria */
 enum CargoFilterSpecialType {
-	CF_ANY  = CT_NO_REFIT, ///< Show all industries (i.e. no filtering)
-	CF_NONE = INVALID_CARGO, ///< Show only industries which do not produce/accept cargo
+	CF_ANY  = NUM_CARGO,     ///< Show all industries (i.e. no filtering)
+	CF_NONE = NUM_CARGO + 1, ///< Show only industries which do not produce/accept cargo
```
2TallTyler: Might be TMWFTLB, but this partially duplicates the filter types in `build_vehicle_gui.cpp` and has the same names. Is there a way to de-duplicate this, or even just take the values directly instead of using the same magic numbers?

Additionally, there's another CargoFilterSpecialType in `vehicle_gui_base.h` -- same story.

2TallTyler (commented):
Do NewGRF cargos inherit from base game cargos based on the chosen ID? (Similar to houses, industrytiles, etc.) And if so, does this break anything related to this?

PeterN:
> Do NewGRF cargos inherit from base game cargos based on the chosen ID? (Similar to houses, industrytiles, etc.) And if so, does this break anything related to this?

Cargoes directly edit the base game cargoes (after being copied from the original const definitions to the runtime definitions) by slot number, rather than making copies, so I don't think there will be any breakages there.

src/cargo_type.h
PeterN: Dunno, I just moved it!

PeterN:
Parts of this draft are now in #11720 in a revised form, and this draft should be considered as depending on that PR too—this breaks all the changes into more manageable logical pieces.

src/table/cargo_const.h
```diff
@@ -50,56 +50,56 @@
...
-static const CargoLabel _default_climate_cargo[NUM_LANDSCAPE][NUM_ORIGINAL_CARGO] = {
-	{ 'PASS', 'COAL', 'MAIL', 'OIL_', 'LVST', 'GOOD', 'GRAI', 'WOOD', 'IORE', 'STEL', 'VALU',     33, },
-	{ 'PASS', 'COAL', 'MAIL', 'OIL_', 'LVST', 'GOOD', 'WHEA', 'WOOD',     34, 'PAPR', 'GOLD', 'FOOD', },
-	{ 'PASS', 'RUBR', 'MAIL',      4, 'FRUT', 'GOOD', 'MAIZ',     11, 'CORE', 'WATR', 'DIAM', 'FOOD', },
-	{ 'PASS', 'SUGR', 'MAIL', 'TOYS', 'BATT', 'SWET', 'TOFF', 'COLA', 'CTCD', 'BUBL', 'PLST', 'FZDR', },
+static const std::variant<CargoLabel, int> _default_climate_cargo[NUM_LANDSCAPE][NUM_ORIGINAL_CARGO] = {
+	{ CT_PASSENGERS, CT_COAL,   CT_MAIL, CT_OIL,  CT_LIVESTOCK, CT_GOODS, CT_GRAIN,  CT_WOOD, CT_IRON_ORE,     CT_STEEL,   CT_VALUABLES, 33, },
+	{ CT_PASSENGERS, CT_COAL,   CT_MAIL, CT_OIL,  CT_LIVESTOCK, CT_GOODS, CT_WHEAT,  CT_WOOD, 34,              CT_PAPER,   CT_GOLD,      CT_FOOD, },
+	{ CT_PASSENGERS, CT_RUBBER, CT_MAIL, 4,       CT_FRUIT,     CT_GOODS, CT_MAIZE,  11,      CT_COPPER_ORE,   CT_WATER,   CT_DIAMONDS,  CT_FOOD, },
+	{ CT_PASSENGERS, CT_SUGAR,  CT_MAIL, CT_TOYS, CT_BATTERIES, CT_CANDY, CT_TOFFEE, CT_COLA, CT_COTTON_CANDY, CT_BUBBLES, CT_PLASTIC,   CT_FIZZY_DRINKS, },
```
rubidium42, line 103: I'm not quite sure about this. I do not think the 33 and 34 ever meant something, so maybe use CT_INVALID instead?
The numeric tropic data seems to be respectively `CT_OIL` and `CT_WOOD` based upon https://newgrf-specs.tt-wiki.net/wiki/CargoDefaultProps. Or is there a good reason not to have them mapped?

PeterN: For tropic, they are labels that are reused, but the stats are different from the first definition. 33 & 34 map to be unused slots with different data, that NewGRF could rely on being set. (Unlikely but possible)

This should be set up the same way as before, if not that's a bug.

rubidium42: Looking further into this I've got the feeling this is a hot mess, but can't think of any reasonable solution to make it any better. Making it a variant at least helps a bit in noticing there might be vastly different behaviour.

rubidium42 (dismissed):

PeterN:
Re-ordering of `CalculateRefitMasks()` and `FinaliseEngineArray()` broke automatic cargo types for default vehicles, so this has been put back to how it was before, and the mapping from default `CargoLabel` to `CargoID` is done in `CalculateRefitMasks()` instead.

rubidium42 (approved):

PR #11720, opened 2024-01-08
Codechange: Split CT_INVALID/INVALID_CARGO and relocate special cargo placeholders.

PeterN:
## Motivation / Problem

CT_INVALID is a CargoType but used like a CargoID. I previously unified these two as there were two names for the same thing, and this helped, however upon further inspection, I'm taking a different course.

There are also additional values used in places where CargoIDs are accepted, used for filtering UI lists, or used in spritegroup resolving tasks. These are (re-)defined in different source files and it's generally not clear how they relate.

## Description

Re-add INVALID_CARGO as a CargoID, and use it throughout the code where a CargoID is expected. Where a CargoType is expected (e.g. in defaults tables), continue using CT_INVALID. (This is not too well defined right now but will be addressed.)

Additionally, define all cargo IDs used by UI cargo filters in a single location, in a namespace to make their special use more distinct.

And do the same with CT_INVALID/CT_PURCHASE/CT_INVALID_NA.

I used namespaces instead of `enum class` as these do need to be compatible with CargoID.

(Some of this is part of #11719, but I think there is enough here to warrant a separate PR.)

## Checklist for review

### DIFF
--- a/src/cargo_type.h
+++ b/src/cargo_type.h
@@ -69,10 +69,12 @@ enum CargoType {
 	CT_INVALID      = 0xFF, ///< Invalid cargo type.
 };
 
+static const CargoID INVALID_CARGO = UINT8_MAX;
+
 /** Test whether cargo type is not CT_INVALID */
 inline bool IsValidCargoType(CargoType t) { return t != CT_INVALID; }
-/** Test whether cargo type is not CT_INVALID */
-inline bool IsValidCargoID(CargoID t) { return t != CT_INVALID; }
+/** Test whether cargo type is not INVALID_CARGO */
+inline bool IsValidCargoID(CargoID t) { return t != INVALID_CARGO; }
 
 typedef uint64_t CargoTypes;
 
--- a/src/cargo_type.h
+++ b/src/cargo_type.h
@@ -61,15 +61,28 @@ enum CargoType {
 	CT_PLASTIC      = 10,
 	CT_FIZZY_DRINKS = 11,
 
-	NUM_ORIGINAL_CARGO = 12,
-	NUM_CARGO       = 64,   ///< Maximal number of cargo types in a game.
-
-	CT_AUTO_REFIT   = 0xFD, ///< Automatically choose cargo type when doing auto refitting.
-	CT_NO_REFIT     = 0xFE, ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
 	CT_INVALID      = 0xFF, ///< Invalid cargo type.
 };
 
+static const CargoID NUM_ORIGINAL_CARGO = 12; ///< Original number of cargo types.
+static const CargoID NUM_CARGO = 64; ///< Maximum number of cargo types in a game.
+
+/* CARGO_AUTO_REFIT and CARGO_NO_REFIT are stored in save-games for refit-orders, so should not be changed. */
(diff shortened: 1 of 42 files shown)


src/cargo_type.h
```diff
@@ -61,16 +61,28 @@ enum CargoType {
...
+static const CargoID NUM_ORIGINAL_CARGO = 12; ///< Original number of cargo types.
+static const CargoID NUM_CARGO = 64; ///< Maximal number of cargo types in a game.
+
+/* CARGO_AUTO_REFIT and CARGO_NO_REFIT are stored in save-games for refit-orders, so should not be changed. */
+static const CargoID CARGO_AUTO_REFIT = 0xFD; ///< Automatically choose cargo type when doing auto refitting.
+static const CargoID CARGO_NO_REFIT = 0xFE; ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-renew).
+
 static const CargoID INVALID_CARGO = UINT8_MAX;
 
+/** Special cargo filter criteria.
```
2TallTyler: When `/**` is multiline, the first line of the comment usually goes on the next line, no?

src/cargo_type.h
```diff
@@ -61,16 +61,28 @@ enum CargoType {
...
-	NUM_ORIGINAL_CARGO = 12,
-	NUM_CARGO       = 64,   ///< Maximal number of cargo types in a game.
-
-	CT_AUTO_REFIT   = 0xFD, ///< Automatically choose cargo type when doing auto refitting.
-	CT_NO_REFIT     = 0xFE, ///< Do not refit cargo of a vehicle (used in vehicle orders and auto-replace/auto-new).
 	CT_INVALID      = 0xFF, ///< Invalid cargo type.
 };
 
+static const CargoID NUM_ORIGINAL_CARGO = 12; ///< Original number of cargo types.
+static const CargoID NUM_CARGO = 64; ///< Maximal number of cargo types in a game.
```
2TallTyler: Since you're changing this line anyway... 😛 

```suggestion
static const CargoID NUM_CARGO = 64; ///< Maximum number of cargo types in a game.
```

src/newgrf_cargo.h
```diff
@@ -14,9 +14,15 @@
 #include "cargo_type.h"
 #include "gfx_type.h"
 
-static const CargoID CT_DEFAULT      = NUM_CARGO + 0;
-static const CargoID CT_PURCHASE     = NUM_CARGO + 1;
-static const CargoID CT_DEFAULT_NA   = NUM_CARGO + 2;
+/** Sprite Group Cargo types.
```
2TallTyler: As elsewhere, this should be on its own line, I think.

2TallTyler (commented):
Looks good to me, just a few nitpicks.

The new CargoFilterCriteria namespace is a nice way to remove duplication. 😃 From a few minutes of Googling, I gather that it's a namespace because an `enum class` can't do `constexpr`?

PeterN:
> Looks good to me, just a few nitpicks.
> 
> The new CargoFilterCriteria namespace is a nice way to remove duplication. 😃 From a few minutes of Googling, I gather that it's a namespace because an `enum class` can't do `constexpr`?

Actually I used a namespace because an `enum class` would explicitly make it incompatible with CargoID, so casting would be needed. A `std::variant` is one possibility to handle this but would make things more complex.

2TallTyler (approved):

PR #11721, opened 2024-01-09
Fix: [Script] Improve ScriptText validation

glx22:
## Motivation / Problem
While running some GS to see how they are affected by ScriptText validation [awards v8](https://bananas.openttd.org/package/game-script/41574453) crashed with `Your script made an error: STR_RANK: Not enough parameters`.
It's because it uses constructs that used to work, producing a valid encoded string (`\SCC_ENCODED4:\SCC_ENCODED13:1:0:0`)
```
	res = GSText(GSText.STR_GREEN);
	res.AddParam(GSText(GSText.STR_RANK));
	res.AddParam(counter);
	res.AddParam(comp);
	res.AddParam(points);
```
with
```
STR_GREEN		: {GREEN}{STRING}
STR_RANK		: {NUM} - {COMPANY} with {NUM} points
```

But that doesn't pass the "strict" validation.

## Description
Current validation checks types only for the parameters of the current string, and disables the checks for extra parameters coming from substrings. So it needs to be reworked to actually validate all parameters, and also accept things that used to work (even if the usage is incorrect).

So I reworked the validation process to use two steps:
- First we get list of the parameters in the same order they used to be in encoded string. That way GSTexts that used to work before introduction of validation should still work.
- The second step encodes the string like it used to be, but each parameter is validated as FormatString would use them.

Holes in parameters are now allowed, IIRC some GS use paramindex to work around their improper STRING/STRINGN usage.

I rewrote the messages to be more helpful:
- param numbers are 1-indexed to match SetParam() syntax.
- the string command is shown in the message.
- the expected/consumed mismatch message is more explicit.
- when a GSText consumes parameters other than its own, this is explicit too.

And now `awards` GS works as it used to (with extra messages in script log because it doesn't conform to the expected usage).

Also my previous test
```
		local text = GSText(GSText.STR_TEST);
		text.SetParam(1, GSText(GSText.STR_FOO, "foo"));
		text.SetParam(2, GSWindow.TC_RED);
		text.SetParam(3, 42);
		GSTown.SetText(0, text);
```
and
```
STR_TEST :{STRING} - {COLOUR}{NUM} is supposed to be 42
STR_FOO  :'{RAW_STRING}'
```
Now fails with
image
instead of showing a "broken" string
image

## Checklist for review

### DIFF
--- a/src/script/api/script_text.cpp
+++ b/src/script/api/script_text.cpp
@@ -160,17 +160,28 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
 
 std::string ScriptText::GetEncodedText()
 {
-	static StringIDList seen_ids;
+	StringIDList seen_ids;
+	ParamList params;
 	int param_count = 0;
-	seen_ids.clear();
 	std::string result;
 	auto output = std::back_inserter(result);
-	this->_GetEncodedText(output, param_count, seen_ids);
+	this->_FillParamList(params);
+	this->_GetEncodedText(output, param_count, seen_ids, params);
 	if (param_count > SCRIPT_TEXT_MAX_PARAMETERS) throw Script_FatalError(fmt::format("{}: Too many parameters", GetGameStringName(this->string)));
 	return result;
 }
 
-void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output, int &param_count, StringIDList &seen_ids)
+void ScriptText::_FillParamList(ParamList &params)
+{
+	for (int i = 0; i < this->paramc; i++) {
+		Param *p = &this->param[i];
+		params.emplace_back(this->string, i, p);
+		if (!std::holds_alternative<ScriptTextRef>(*p)) continue;
+		std::get<ScriptTextRef>(*p)->_FillParamList(params);
+	}
+}
+
+void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output, int &param_count, StringIDList &seen_ids, ParamSpan args)
 {
 	const std::string &name = GetGameStringName(this->string);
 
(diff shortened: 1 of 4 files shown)


LordAro:
That static variable is suspect to me - what happens if you try to build multiple GSTexts at the same time, with interleaved SetParam calls? param_stack would be erroneously cleared/set, no?

glx22:
Encoding happens when a GSText is passed to openttd, and it's atomic.

PeterN:
As it is cleared every call, and passed as a parameter to child functions, why does it (and seen_ids) need to be static?

src/script/api/script_text.cpp
```diff
@@ -161,88 +161,107 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
...
-				}
+	int idx = 0;
+	int extra = 0;
+	ParamStack param_backup;
+
+	auto pop_param = [&]() {
+		if (param_stack.empty()) throw Script_FatalError(fmt::format("{}: Not enough parameters", name));
+		Param *p = param_stack.back();
+		if (extra > 0) {
+			ScriptLog::Warning(fmt::format("{}: Parameter {} consumes from a substring", name, param_count));
```
rubidium42:
```suggestion
			ScriptLog::Warning(fmt::format("{}: Parameter {} consumed from a substring", name, param_count));
```

src/script/api/script_text.cpp
```diff
@@ -161,88 +161,107 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
...
+			param_stack.pop_back();
+			extra--;
+		} else {
+			if (idx >= this->paramc) ScriptLog::Warning(fmt::format("{}: Parameter {} consumes from parent string", name, param_count));
```
rubidium42:
```suggestion
			if (idx >= this->paramc) ScriptLog::Warning(fmt::format("{}: Parameter {} consumed from parent string", name, param_count));
```

rubidium42 (commented):
This is method is getting awfully complex; would it be feasible to have some sort of tests for it? Not sure whether to use C++ unit test to set up the test cases, or maybe a GS regression test of sorts?

glx22:
I decided to use a different method.
It's simpler and should improve backward compatibility.
Updated the PR description.

src/script/api/script_text.cpp
```diff
@@ -161,16 +161,28 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
...
 	if (param_count > SCRIPT_TEXT_MAX_PARAMETERS) throw Script_FatalError(fmt::format("{}: Too many parameters", GetGameStringName(this->string)));
 	return result;
 }
 
-void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output, int &param_count, StringIDList &seen_ids)
+void ScriptText::_FillParamList(ParamList &params)
+{
+	for (int i = 0; i < this->paramc; i++) {
+		Param *p = &this->param[i];
+		params.push_back(ParamCheck{ this->string, i, p });
```
rubidium42: Why not `emplace_back` if the constructor is in the `push_back` call?

PeterN:
As per #11754, can we script-test or unit-test this?

src/script/api/script_text.cpp
```diff
@@ -160,17 +160,31 @@ SQInteger ScriptText::_set(HSQUIRRELVM vm)
...
 
-void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output, int &param_count, StringIDList &seen_ids)
+void ScriptText::_FillParamList(ParamList &params)
+{
+	for (int i = 0; i < this->paramc; i++) {
+		Param *p = &this->param[i];
+		params.emplace_back(this->string, i, p
+
+
+		);
```
rubidium42:
```suggestion
		params.emplace_back(this->string, i, p);
```

rubidium42 (approved):

PR #11743, opened 2024-01-09
Fix: Clear one-way state when removing road from road+tram tile

JGRennison:
## Motivation / Problem

When a road+tram tile with a one-way state set is changed to a tram only tile, the one way state is preserved, but invisible.
If road is re-added in either axis it will unexpectedly have a one-way state.

## Description

Clear the one-way state when removing road from a road+tram tile.

## Limitations

N/A

## Checklist for review


2TallTyler (approved):
I think the right prefix for this is `Fix:`, but we could change that when merging. 🙂

PR #11750, opened 2024-01-10
Change: simplified water region evaluation, removed savegame data

Kuhnovic:
## Motivation / Problem

This is closely related to the recently merged region-based ship pathfinder https://github.com/OpenTTD/OpenTTD/pull/10543

There was some discussion with @JGRennison and @SamuXarick on discord, there were worries about desyncs. In 10543 I've tried to prevent desyncs by saving the "is the region initialized "state of all regions in the savegame. The idea was to synchronize the "initialized-ness" in a bug-compatible way, meaning that if it wasn't initialized on the server then it wouldn't be initialized on the joining client either. But this is what can happen:

- server starts a game
- some action, say terraforming, doesn't properly invalidate water region X. Region X is now incorrectly marked as up-to-date.
- a client joins the game. The client needs to scan / label (i.e. update) all water regions that are labeled as up-to-date. This should put the client's water region cache in sync with the one on the server...
- ... but the cache on the server is corrupted, because the "up-to-dateness" of region X is incorrectly set to true. Region X on the server has a different state than region X on the client. This could leads to desyncs.

The current solution of syncing the up-to-date-state of the regions simply doesn't solve this.

Link to discord: https://discord.com/channels/142724111502802944/1008473233844097104/1194656849237127308

## Description

So I'm taking a slightly more simple and brute force approach: *invalidate a water region on any change*. This includes changes to tiles that might not be entirely relevant to the water regions. This by itself shouldn't really have big impact on performance, region invalidation is just a matter of setting a boolean.

## Limitations

- This approach will invalidate more regions than necessary, which in turn will cause more water regions having to update (re-do their connected component labeling). This will have some impact on performance, but I don't think it will be significant.
- I lowered the savegame version and removed the region data from the savegame. This will render some savegames incompatible, but only ones that were created with recent builds or the nightlies. I can change this if that's not acceptable.
- There's some debug code that's nice for testing, but ultimately has to be removed.

## Checklist for review

### DIFF
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -36,6 +36,7 @@
 #include "landscape_cmd.h"
 #include "terraform_cmd.h"
 #include "station_func.h"
+#include "pathfinder/water_regions.h"
 
 #include "table/strings.h"
 #include "table/sprites.h"
@@ -538,6 +539,8 @@ void DoClearSquare(TileIndex tile)
 	MakeClear(tile, CLEAR_GRASS, _generating_world ? 3 : 0);
 	MarkTileDirtyByTile(tile);
 	if (remove) RemoveDockingTile(tile);
+
+	InvalidateWaterRegion(tile);
 }
 
 /**
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -18,6 +18,7 @@
 #include "tunnelbridge_map.h"
 #include "follow_track.hpp"
 #include "ship.h"
+#include "debug.h"
 
 using TWaterRegionTraversabilityBits = uint16_t;
 constexpr TWaterRegionPatchLabel FIRST_REGION_LABEL = 1;
@@ -114,6 +115,7 @@ public:
 	 */
 	void ForceUpdate()
 	{
+		Debug(map, 3, "Updating water region ({},{})", GetWaterRegionX(this->tile_area.tile), GetWaterRegionY(this->tile_area.tile));
 		this->has_cross_region_aqueducts = false;
 
 		this->tile_patch_labels.fill(INVALID_WATER_REGION_PATCH);
@@ -267,8 +269,9 @@ WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile)
 void InvalidateWaterRegion(TileIndex tile)
 {
 	const int index = GetWaterRegionIndex(tile);
-	if (index > static_cast<int>(_water_regions.size())) return;
 	_water_regions[index].Invalidate();
+
+	Debug(map, 3, "Invalidated water region ({},{})", GetWaterRegionX(tile), GetWaterRegionY(tile));
 }
 
 /**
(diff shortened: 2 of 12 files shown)


src/landscape.cpp
```diff
@@ -41,6 +41,7 @@
 #include "table/sprites.h"
 
 #include "safeguards.h"
+#include <pathfinder/water_regions.h>
```
TrueBrain: This include shouldn't be here (should be before safeguards), and shouldnt be `<>`?

src/pathfinder/water_regions.cpp
```diff
@@ -267,8 +268,11 @@ WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile)
 void InvalidateWaterRegion(TileIndex tile)
 {
 	const int index = GetWaterRegionIndex(tile);
-	if (index > static_cast<int>(_water_regions.size())) return;
+	if (index >= static_cast<int>(_water_regions.size())) return;
 	_water_regions[index].Invalidate();
+
+	// TODO KB remove
+	Debug(map, 2, "Invalidated water region ({},{})", GetWaterRegionX(tile), GetWaterRegionY(tile));
```
TrueBrain: TODO, but not a draft?

Generally speaking, a PR can be a draft (with TODOs) or be ready-for-review (without TODOs). Makes it easier for us how to look at code :)

src/pathfinder/water_regions.cpp
```diff
@@ -342,38 +346,19 @@ void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_pat
...
- * Initializes all water regions. All water tiles will be scanned and interconnected water patches within regions will be identified.
+ * Allocates the appropriate amount of water regions for the current map size
  */
-void InitializeWaterRegions()
+void AllocateWaterRegions()
 {
 	_water_regions.clear();
 	_water_regions.reserve(static_cast<size_t>(GetWaterRegionMapSizeX()) * GetWaterRegionMapSizeY());
 
+	Debug(map, 1, "Allocating {} x {} water regions", GetWaterRegionMapSizeX(), GetWaterRegionMapSizeY());
```
TrueBrain: Is this actually meant to go in the code? If so, please increase the debugging level; doesn't sound worthy of 1 :)

src/water_cmd.cpp
```diff
@@ -945,6 +933,10 @@ static void DrawTile_Water(TileInfo *ti)
 			DrawWaterDepot(ti);
 			break;
 	}
+
+	// TODO KB remove: draw region boundaries
```
TrueBrain: TODOs

TrueBrain (commented):
I have no way of judging this code, but if you say it works, I will approve it. It just does need a bit of cleanup :)

PeterN:
This will definitely need to handle the saveload chunk being removed (unless that happens automagically.)

src/pathfinder/water_regions.cpp
Kuhnovic: Map::allocate had debug level 1 so I used that one, although you could argue that the map is quite a bit more important. I've changed this debug level to 2 and the others to 3.

src/pathfinder/water_regions.cpp
Kuhnovic: My bad, should be fixed now

src/landscape.cpp
Kuhnovic: Automatically added by VS, should have paid more attention ;)

src/saveload/saveload.h
```diff
@@ -365,7 +365,6 @@ enum SaveLoadVersion : uint16_t {
 	SLV_TIMETABLE_START_TICKS,              ///< 321  PR#11468 Convert timetable start from a date to ticks.
 	SLV_TIMETABLE_START_TICKS_FIX,          ///< 322  PR#11557 Fix for missing convert timetable start from a date to ticks.
 	SLV_TIMETABLE_TICKS_TYPE,               ///< 323  PR#11435 Convert timetable current order time to ticks.
-	SLV_WATER_REGIONS,                      ///< 324  PR#10543 Water Regions for ship pathfinder.
```
PeterN, line 368: I don't removing this is a good idea, it would mean saves which have the water region chunks wouldn't be loadable in future versions. I'm not sure they'd be loaded in the current version either.

src/pathfinder/water_regions.cpp
```diff
@@ -342,38 +346,37 @@ void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_pat
...
+	void Load() const override
+	{
+		SlTableHeader({});
+		SlSkipArray();
+	};
+};
+
+static const WaterRegionChunkHandler WRGN;
+static const ChunkHandlerRef water_region_chunk_handlers[] = { WRGN };
+extern const ChunkHandlerTable _water_region_chunk_handlers(water_region_chunk_handlers);
```
TrueBrain: Please keep this in `water_regions_sl.cpp`. No need to hide it.

TrueBrain:
One last thing, sorry about that: I think we should bump the savegame version. Otherwise you can create a savegame which you cannot load, with a very "corrupted savegame" error in the older version. Better to have it say it is not compatible, I think.

src/landscape.cpp
```diff
@@ -538,6 +539,8 @@ void DoClearSquare(TileIndex tile)
 	MakeClear(tile, CLEAR_GRASS, _generating_world ? 3 : 0);
 	MarkTileDirtyByTile(tile);
 	if (remove) RemoveDockingTile(tile);
+
+	InvalidateWaterRegion(tile);
```
PeterN, line 543: Looks like you now rely on DoClearSquare() being called when any tile is modified to refresh regions (except for objects). Is this always the case? Industries come to mind.

src/pathfinder/water_regions.cpp
```diff
@@ -267,8 +269,10 @@ WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile)
 void InvalidateWaterRegion(TileIndex tile)
 {
 	const int index = GetWaterRegionIndex(tile);
-	if (index > static_cast<int>(_water_regions.size())) return;
+	if (index >= static_cast<int>(_water_regions.size())) return;
```
PeterN: When does this condition occur? If never and it's a safety check then this should probably be an assert (or asserting that tile is valid)?

src/landscape.cpp
Kuhnovic: The reason I use DoClearSquare() is that everything that wants to build something on a tile first has to check if that tile can be cleared before it can "claim it for itself". Industry construction does this too. The only exception I've found so far is placing objects, where tile clearing can't be done. I forgot the exact reason but it's in the comments in the place object command, something to do with NewGRFs.

src/pathfinder/water_regions.cpp
Kuhnovic: I think this is a remnant of the time where the initialization code was poorly timed. I remember that abandoning a game with a map larger than the menu game issues, but they no longer seem to occur. I will remove this check, bounds checking is already compiled into the [] operator in debug.

TrueBrain (approved):
Up to @PeterN to merge this, if he is also okay with these changes.

Tnx for addressing these issues btw, it is appreciated!

PR #11751, opened 2024-01-10
Codechange: refactor the Windows-only DllLoader in a cross-platform LibraryLoader

TrueBrain:
## Motivation / Problem

For another PR (#11628) I had need to dynamically load a library on any OS. We already had some code to do something similar for Windows only (in `DllLoader`). So this PR extends its capabilities and makes it cross-platform.

## Description

`DllLoader` was very scoped around Windows (both in name and implementation). This PR recreates the API of this loader, making sure it can also be implemented for Linux and MacOS.

When-ever an instance of `LibraryLoader` gets out of scope, everything is cleaned up nicely. This means that any function pointer you retrieve from it only remains valid for as long as the object lives. It -might- live longer, but that is OS-dependent, and kinda depends if anything else still has the library loaded or not.

During implementation I found out that some headers were missing from other files, now `<windows.h>` wasn't included anymore in `os/windows/win32.h`.

`LibraryLoader` has functions to see if loading succeeded and/or what the error was (if any). This helps in producing nice error-messages to the user, instead of internal error codes.

PS: yes, the MacOS implementation is in there: it is the same as the Linux implementation, and both OSes share the implementation from `os/unix/..`, as is already the case with some other pieces of code in there.
PPS: no, Emscripten is not (and will not) be supported, as there is no concept of "library" in that sense. We could, in theory, make them WASM plugins, but that requires other infrastructure to load in.

## Checklist for review

### DIFF
--- a/src/library_loader.h
+++ b/src/library_loader.h
@@ -1,0 +1,112 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file library_loader.h Functions/types related to loading libraries dynamically. */
+
+#ifndef LIBRARY_LOADER_H
+#define LIBRARY_LOADER_H
+
+class LibraryLoader {
+public:
+	/**
+	 * A function loaded from a library.
+	 *
+	 * Will automatically cast to the correct function pointer type on retrieval.
+	 */
+	class Function {
+	public:
+		explicit Function(void *p) : p(p) {}
+
+		template <typename T, typename = std::enable_if_t<std::is_function_v<T>>>
+		operator T *() const
+		{
+			return reinterpret_cast<T *>(this->p);
+		}
+
+	private:
+		void *p;
+	};
+
+	/**
+	 * Load a library with the given filename.
+	 */
+	explicit LibraryLoader(const std::string &filename)
+	{
+		this->handle = this->OpenLibrary(filename);
+	}
+
+	/**
+	 * Close the library.
+	 */
+	~LibraryLoader()
+	{
+		if (this->handle != nullptr) {
+			this->CloseLibrary();
+		}
+	}
+
+	/**
+	 * Check whether an error occurred while loading the library or a function.
+	 *
+	 * @return Whether an error occurred.
+	 */
+	bool HasError()
+	{
+		return this->error.has_value();
+	}
+
+	/**
+	 * Get the last error that occurred while loading the library or a function.
+	 *
+	 * @return The error message.
+	 */
+	std::string GetLastError()
+	{
+		return this->error.value_or("No error");
+	}
+
+	/**
+	 * Get a function from a loaded library.
+	 *
+	 * @param symbol_name The name of the function to get.
+	 * @return The function. Check HasError() before using.
+	 */
+	Function GetFunction(const std::string &symbol_name)
+	{
+		if (this->error.has_value()) return Function(nullptr);
+		return Function(this->GetSymbol(symbol_name));
+	}
+
+private:
+	/**
+	 * Open the library with the given filename.
+	 *
+	 * Should set error if any error occurred.
+	 *
+	 * @param filename The filename of the library to open.
+	 */
+	void *OpenLibrary(const std::string &filename);
+
+	/**
+	 * Close the library.
+	 */
+	void CloseLibrary();
+
+	/**
+	 * Get a symbol from the library.
+	 *
+	 * Should set error if any error occurred.
+	 *
+	 * @param symbol_name The name of the symbol to get.
+	 */
+	void *GetSymbol(const std::string &symbol_name);
+
+	std::optional<std::string> error = {}; ///< The last error that occurred, if set.
+	void *handle = nullptr; ///< Handle to the library.
+};
+
+#endif /* LIBRARY_LOADER_H */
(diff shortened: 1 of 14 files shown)


src/library_loader.h
```diff
@@ -0,0 +1,115 @@
...
+	 * Get a symbol from the library.
+	 *
+	 * Should set has_error / error if any error occurred.
+	 *
+	 * @param symbol_name The name of the symbol to get.
+	 */
+	void *GetSymbol(const std::string &symbol_name);
+
+	bool has_error = false; ///< Whether an error occurred.
+	std::string error = {}; ///< The last error that occurred, if has_error is set.
```
rubidium42: This smells like `std::optional<std::string> error` should be used instead. That way, setting the string automatically sets the "has error" bit, resulting code where it's harder to miss setting`has_error` or `error`.

src/os/windows/library_loader_win.cpp
```diff
@@ -0,0 +1,61 @@
...
+	auto error = GetLastError();
+	std::string message;
+	char buffer[512];
+	if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, error,
+		MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), buffer, sizeof(buffer), nullptr) == 0) {
+		message.assign(fmt::format("Unknown error {}", error));
+	} else {
+		message.assign(buffer);
+	}
+	return message;
```
rubidium42:
```suggestion
	char buffer[512];
	if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, error,
		MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), buffer, sizeof(buffer), nullptr) == 0) {
		return fmt::format("Unknown error {}", error);
	}
	return buffer;
```
This saves at least the copy that is happening due to the `assign`. And if the compiler can do copy elision/return value optimization, it will write directly in the structure of the caller of this function.

src/library_loader.h
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Close the library.
+	 */
+	~LibraryLoader()
+	{
+		if (this->handle != nullptr) {
+			this->CloseLibrary();
+		}
```
rubidium42, line 49: Not quite sure about this destructor.

On one hand having the separate `CloseLibrary` function means this `nullptr` check can be left out of all the implementers, but it also triggers the question: shouldn't `CloseLibrary` set the handle to `nullptr`, just in case someone changes the internal logic of this at some later point.

On the other hand, the different OS specific implementations can also just implement the destructor and essentially rename `CloseLibrary` to `~LibraryLoader`. Though that'll make the constructor feel weird, but the `OpenLibrary` implementations could also become the constructor in the implementations.

src/os/unix/library_loader_unix.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void LibraryLoader::CloseLibrary()
+{
+}
+
+void *LibraryLoader::GetSymbol(const std::string &)
+{
+	this->has_error = true;
+	this->error = "Dynamic loading is not supported on this platform.";
```
rubidium42: This isn't needed, because `OpenLibrary` already set these in the constructor of the loader.

src/library_loader.h
```diff
@@ -0,0 +1,115 @@
...
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file library_loader.h Functions/types related to loading libraries dynamically. */
+
+#ifndef LIBRARY_LOADER_H
+#define LIBRARY_LOADER_H
+
+#include <string>
```
rubidium42: `string` is included in `stdafx.h`, so it's not needed here.

src/os/unix/unix.cpp
```diff
@@ -15,6 +15,7 @@
 #include "../../thread.h"
 
 
+#include <dlfcn.h>
```
rubidium42: Why is this include needed here?

src/os/windows/win32.h
```diff
@@ -10,50 +10,9 @@
...
-	}
-
-
-	~DllLoader()
-	{
-		::FreeLibrary(this->hmodule);
-	}
-
-	bool Success() { return this->success; }
+#include <string>
```
rubidium42: This shouldn't be needed.

rubidium42 (dismissed):

src/os/windows/win32.h
TrueBrain: Why not? It very much is :)

This header file uses `std::string_view`, defined in `<string>`. Realise, as I mentioned in the description, that there used to be a `#include <windows>` in this file. It was hiding all these dependencies.

TrueBrain: Ah, this is in reference to an earlier comment I guess. Sure.

src/os/unix/library_loader_unix.cpp
TrueBrain: Might not be needed, but does feel more correct. To set the error and the error message. And it doesn't hurt anyone, as you shouldn't be calling this function if OpenLibrary fails.

So a bit pedantic code from my side, but I am going to leave it in :)

src/library_loader.h
TrueBrain: Tomato tomato. Nothing wrong with the current choice, so going to leave it as it is :) I don't feel like retesting everything for just this, sorry :)

src/library_loader.h
LordAro: or just use error.empty() as the condition?

TrueBrain: > or just use error.empty() as the condition?

An empty error sadly doesn't mean there was no error :( I can't guarantee one of the OSes doesn't show any error.

src/library_loader.h
LordAro: unique_ptr with custom destructor?

rubidium42 (approved):

src/os/windows/win32.h
glx22: Yeah commonly used std headers are included in stdafx.h for precompiled headers

PR #11753, opened 2024-01-11
Fix #11485: new run on same line must not use last_space of previous run

### LINKED ISSUE #11485: [Bug]: Storybook duplicates some words when wrapping
### Version of OpenTTD

13.4

### Expected result

When using a Game Script like Renewed Village Growth, the storybook text wraps properly.

### Actual result

wrap_double

The word "faster" is repeated on the next line, when it is the last word in the row and is near the edge of the window. This requires resizing the window.

Oddly, no other words in this window seem to do this.

The [string in question](https://github.com/F1rrel/RenewedVillageGrowth/blob/master/lang/english.txt#L42):
```
STR_SB_WELCOME_CARGO     :Economy {SILVER}{STRING}{BLACK} consists of {SILVER}{NUM} categories{BLACK}. To fulfill a category, deliver one or more of the category specific cargos to a {ORANGE}station near the town which accepts that cargo type{BLACK}. Fulfill the requirements above {SILVER}{NUM}%{BLACK} to {ORANGE}make the towns grow faster{BLACK}. More information about cargo categories are shown in the {ORANGE}Cargo categories story book page{BLACK}.
```

### Steps to reproduce

1. Start a new game with the "Renewed Village Growth 9.2" game script.
2. Wait for the Story Book window to open.
3. Change the width of the window until the word "faster" is duplicated on two lines of text.

rubidium42:
## Motivation / Problem

Fixes #11485.

The fallback layout "next line" function maintains a `last_space` variable with the location of the last space. When that variable is not `nullptr`, the start point for the next line will be reset to the location of the last space.
When the font colour changes, there is a new so-called `run`. At that point the line up to the `run` is added to the flowed strings. Then it will continue in the new run at the first character. If that character causes the line to become too long, the start point for the next line is set to the location of the last space, which is before the run.

The result is that the last bit of the previous run is going to be output on the following line as well. Even worse, in theory when the window is just narrow enough (or the word without spaces long enough), this could go on infinitely.

## Description

Set `last_space` to fake that there is a space just before the new run started.

## Limitations

This will cause that punctuation after a new run starts might not be placed on the same line as the end of the previous run. In other words, the full stop might end up on the next line. Fixing that is a whole can of worms that I do not want to touch.

## Checklist for review

### DIFF
--- a/src/gfx_layout_fallback.cpp
+++ b/src/gfx_layout_fallback.cpp
@@ -256,6 +256,11 @@ std::unique_ptr<const ParagraphLayouter::Line> FallbackParagraphLayout::NextLine
 
 			next_run = this->buffer_begin + iter->first;
 			begin = this->buffer;
+			/* Since a next run is started, there is already some text that
+			 * will be shown for this line. However, we do not want to break
+			 * this line at the previous space, so pretend we passed a space
+			 * just before this next run. */
+			last_space = begin - 1;
 		}
 
 		if (IsWhitespace(c)) last_space = this->buffer;


rubidium42:
> Basically a revert of #10356 which removed the assignment of nullptr.

It used to be, because I didn't spot that problem. It should be solved now by not using `nullptr` but assuming that when a new run starts that's also the location of the last space.

TrueBrain (approved):
My head still spins when looking at layouters :) (and yes, I wrote one, but it still makes my head spin :P)

PR #11754, opened 2024-01-11
Revert a2edf52: SQOpsLimiter does a more precise job

glx22:
## Motivation / Problem
a2edf52 introduced a check to kill a script if Valuate() takes too much ops (but it can prevent infinite loops in the valuator itself)
#11670 (modified by #11698) introduced another way to limit the ops a Valuate() call can consume, but inside the VM so it can catch infinite loops.

The first check is now redundant.

## Description
Remove the redundant check.

## Checklist for review


TrueBrain (commented):
Would maybe be good to add something like a test to the regression or where ever, so we know this is all actually still working?

All I have with this PR is your word; not even a test AI to validate your claim :D And I am very tempted to just approve, but before, first a check if we can do anything to show Valuators are still actually cut of :)

glx22:
Added a test to regression.
This test would have caused a hang before #11670.

TrueBrain (approved):
Awesome, tnx!

PR #11766, opened 2024-01-12
Add: Support zero-width space {ZWSP} to indicate where strings may additionally break

PeterN:
## Motivation / Problem

Non-Latin translations may use characters after which a break could be made, which are not treated as breaking characters, e.g. `：` instead of `:`. As an example the string:

`STR_NEWGRF_SETTINGS_FILENAME                                    :{BLACK}文件名：{SILVER}{STRING}`

may be wrapped as

```
文件
名：filename
```

instead of

```
文件名：
filename
```

## Description

Add support for zero-width space as `{ZWSP}`. This allows translators to continue using non-ascii symbols like `：` instead of `: `.

## Limitations

Probably requires modifications to eints, and some how translators need to know this is an option.

I've tested this as far as I can, but ICU already seems to wrap after `：`.

## Checklist for review

### DIFF
--- a/src/gfx_layout.cpp
+++ b/src/gfx_layout.cpp
@@ -93,6 +93,8 @@ static inline void GetLayouter(Layouter::LineCacheItem &line, std::string_view s
 			state.PopColour();
 		} else if (c >= SCC_FIRST_FONT && c <= SCC_LAST_FONT) {
 			state.SetFontSize((FontSize)(c - SCC_FIRST_FONT));
+		} else if (c == 0x200B) {
+			/* Zero-width space. */
 		} else {
 			/* Filter out non printable characters */
 			if (!IsPrintable(c)) continue;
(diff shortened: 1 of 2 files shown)


PeterN:
Hmm, I guess it's not quite there. With `{ZWSP}` used in strings (instead of the dynamic embedded code I tested with)...

`dbg: [fontcache] Font is missing glyphs to display char 0x200B in medium font size`

rubidium42:
This implementation is not going to work as uniscribe does not itemize the string on style (font colour) changes, which is the trick that this is using.

If uniscribe where to itemize on style, then the example string would not show the problem in the UI because there is a colour change. So `{ZWSP}` would only be needed where there's non-ASCII punctuation and not a new string colour.

We might even want to consider automatically adding `{ZWSP}` in strgen whenever one of the non-ASCII punctuations comes up in the string and there's no colour command after it.

TrueBrain:
Just curious: any progress on this, or did we abandon the idea of ZWSP?

PeterN:
> Just curious: any progress on this, or did we abandon the idea of ZWSP?

No, but also no.

But will close while there's no movement.

PR #11768, opened 2024-01-13
Fix #5713: FindClosestShipDepot only considers depots that are actually reachable

### LINKED ISSUE #5713: Boats sent to unreachable depots

Kuhnovic:
## Motivation / Problem

When a ship needs to go to a depot, it simply finds the closest depot based on as-the-crow-flies distance. It does not check whether that depot is actually reachable or not. Checking this was simply not feasible as the YAPF pathfinder would reach it's tile limit very quickly and just give up.

## Description

Now that https://github.com/OpenTTD/OpenTTD/pull/10543 we can do the reachability check using water regions, which is orders of magnitude faster. First a "blob" of reachable water region patches up to a certain distance is determined. Only depots within this blob are considered.

You might ask "why aren't you using the BFS to find the closest depot that way". That was my initial solution too, but the 16x16 water regions make for a too coarse resolution to pick the closest depot. Here's an example with some debug drawing added to show the region boundaries. The BFS search would prefer the depot in the same region as the ship, simply because it's "fever regions away":
image 

... in the end I went for a K.I.S.S. solution.

## Limitations

Note that an attempt has already been made at this: https://github.com/OpenTTD/OpenTTD/pull/10544 . I wanted to offer a simpler solution.

I took some artistic license on this:
- I've limited the max depot search to 80 tiles when the user clicks the "go to depot" button on a ship. I feel that it doesn't make sense for a ship to sail all across the map if there happens to be a depot there, possibly taking years. It also prevents the water region search to expand the entire map, although that would still be fast. *Note that ships looking for depot to service in use a search distance of 20 by default, this depends on pathfinder penalty settings. *
- I've changed the distance check from manhattan to circular. This makes more sense IMO.

One can still create situations where the "wrong" depot will be selected. In the screenshot below the depot with the yellow circle will be selected, even though the other depot is closer. This is existing behavior, and not something that this PR is trying to address.
image

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -24,6 +24,7 @@ constexpr TWaterRegionPatchLabel FIRST_REGION_LABEL = 1;
 constexpr TWaterRegionPatchLabel INVALID_WATER_REGION_PATCH = 0;
 
 static_assert(sizeof(TWaterRegionTraversabilityBits) * 8 == WATER_REGION_EDGE_LENGTH);
+static_assert(sizeof(TWaterRegionPatchLabel) == sizeof(byte)); // Important for the hash calculation.
 
 static inline TrackBits GetWaterTracks(TileIndex tile) { return TrackStatusToTrackBits(GetTileTrackStatus(tile, TRANSPORT_WATER, 0)); }
 static inline bool IsAqueductTile(TileIndex tile) { return IsBridgeTile(tile) && GetTunnelBridgeTransportType(tile) == TRANSPORT_WATER; }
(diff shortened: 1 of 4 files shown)


TrueBrain (approved):
Only looked at coding-style. Functionality-wise, I am sure you know better what you are doing than I am at the moment.

Tnx for picking up these tickets .. it is really nice the shipPF finally performs :)

PR #11769, opened 2024-01-14
Fix #10118: Always cycle through all visible signal types

### LINKED ISSUE #10118: [Bug]: «Сycle trough signal types» set to «Path signals only» by default

2TallTyler:
## Motivation / Problem

We have two separate settings which control which signals are cycled through when Ctrl+Clicking a signal. The first is the signal menu mode: if all signals are shown or just path signals. The second is `Cycle through signal types` which since #8688 has two choices: `All visible` and `Path signals only`.

This second setting doesn't really make sense, as reported in #10118. In the comments of #10119, @frosch123 suggested we remove it entirely to always cycle through visible signals only, and I agree.

## Description

Remove the setting, so we always cycle through whatever signals are visible to the player.

Closes #10118.

## Checklist for review

### DIFF
--- a/src/rail_gui.cpp
+++ b/src/rail_gui.cpp
@@ -240,8 +240,8 @@ static void GenericPlaceSignals(TileIndex tile)
 	if (_remove_button_clicked) {
 		Command<CMD_REMOVE_SIGNALS>::Post(STR_ERROR_CAN_T_REMOVE_SIGNALS_FROM, CcPlaySound_CONSTRUCTION_RAIL, tile, track);
 	} else {
-		/* Which signals should we cycle through? */
-		SignalType cycle_start = _settings_client.gui.cycle_signal_types == SIGNAL_CYCLE_ALL && _settings_client.gui.signal_gui_mode == SIGNAL_GUI_ALL ? SIGTYPE_BLOCK : SIGTYPE_PBS;
+		/* Only cycle through signals that are visible in the menu. */
+		SignalType cycle_start = _settings_client.gui.signal_gui_mode == SIGNAL_GUI_ALL ? SIGTYPE_BLOCK : SIGTYPE_PBS;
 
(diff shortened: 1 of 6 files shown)


src/rail_gui.cpp
```diff
@@ -240,8 +240,8 @@ static void GenericPlaceSignals(TileIndex tile)
 	if (_remove_button_clicked) {
 		Command<CMD_REMOVE_SIGNALS>::Post(STR_ERROR_CAN_T_REMOVE_SIGNALS_FROM, CcPlaySound_CONSTRUCTION_RAIL, tile, track);
 	} else {
-		/* Which signals should we cycle through? */
-		SignalType cycle_start = _settings_client.gui.cycle_signal_types == SIGNAL_CYCLE_ALL && _settings_client.gui.signal_gui_mode == SIGNAL_GUI_ALL ? SIGTYPE_NORMAL : SIGTYPE_PBS;
+		/* Only cycle through signals that visible in the menu. */
```
michicc:
```suggestion
		/* Only cycle through signals that are visible in the menu. */
```

PeterN:
Hmm, I feel these two commits are basically separate things that happen to be about signals.

TrueBrain (approved):

ldpl:
Show all cycle path is my preferred config as I mostly use path signals I don't want it cycling through all but I'm not some newbie that doesn't know block signals to hide them.
And there is more in that comment from froshch than just removing the setting. Having it switch block to path is weird, that I agree. But having an option to only cycle path signals is useful. Whether it also cycles blocks signal among themselves doesn't matter to me but sounds logical.

2TallTyler:
Superseded by #11798.

PR #11776, opened 2024-01-14
Remove: Debug redirect over network

### LINKED ISSUE #11590: [Bug]: Debug() redirection to TCP does not work when running as dedicated server
### Version of OpenTTD

13.4, Windows 10

### Expected result

The `-l` option should redirect all log output to the specified ip and port regardless of whether the game is in client or dedicated server mode. Logging should also start much sooner, early logs are never sent to the given ip and port.

### Actual result

The `-l` option only redirects log output to the given ip and port if the game is running as a client. Logging also starts very late, after the main menu has loaded.

### Steps to reproduce

1. Start the game with parameters like `-D -d 9 -l 127.0.0.1`
2. Notice that no connection is made to a listener on the default/specified port

rubidium42:
## Motivation / Problem

It does not work for dedicated servers because upon starting the process to resolve the address to redirect to gets killed. Also with all the async going on in the network code, the debug redirection will start very late in the process.

Since the problem has been introduced in January of 2009 there is only one issue (Closes #11590) mentioning this problem, so it is a very rarely used feature.
The problem can be solved by special casing the redirecting, so everything happens in a blocking manner. However, that feels like a maintenance nightmare. So removing it seems to be the better solution.

## Description

Just remove it.

## Limitations

No more simple debug redirection available at all in the application.
Though, maybe just redirect OpenTTD's output to `nc` and let that connect to another server?

## Checklist for review

### DIFF
--- a/src/network/network.cpp
+++ b/src/network/network.cpp
@@ -1218,32 +1218,6 @@ static void NetworkGenerateServerId()
 	_settings_client.network.network_id = GenerateUid("OpenTTD Server ID");
 }
 
-class TCPNetworkDebugConnecter : TCPConnecter {
-private:
-	std::string connection_string;
-
-public:
-	TCPNetworkDebugConnecter(const std::string &connection_string) : TCPConnecter(connection_string, NETWORK_DEFAULT_DEBUGLOG_PORT), connection_string(connection_string) {}
-
-	void OnFailure() override
-	{
-		Debug(net, 0, "Failed to open connection to {} for redirecting Debug()", this->connection_string);
-	}
-
-	void OnConnect(SOCKET s) override
-	{
-		Debug(net, 3, "Redirecting Debug() to {}", this->connection_string);
-
-		extern SOCKET _debug_socket;
-		_debug_socket = s;
-	}
-};
-
-void NetworkStartDebugLog(const std::string &connection_string)
-{
-	new TCPNetworkDebugConnecter(connection_string);
-}
-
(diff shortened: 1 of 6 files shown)


src/network/network.cpp
```diff
@@ -1218,32 +1218,6 @@ static void NetworkGenerateServerId()
 	_settings_client.network.network_id = GenerateUid("OpenTTD Server ID");
 }
 
-class TCPNetworkDebugConnecter : TCPConnecter {
-private:
-	std::string connection_string;
-
-public:
-	TCPNetworkDebugConnecter(const std::string &connection_string) : TCPConnecter(connection_string, NETWORK_DEFAULT_DEBUGLOG_PORT), connection_string(connection_string) {}
```
TrueBrain, line 1226: `NETWORK_DEFAULT_DEBUGLOG_PORT` can also be removed :)

TrueBrain:
As I added this functionality back in 2007, I can just say it without filter: this functionality was extremely naive and stupid to implement.

A bit of a backstory, I needed this to be able to develop for the PSP port I was working on. The emulator had no way for a console, so I was going blind when debugging issues. And I used a lot of debug-statement to debug silly problems. This was my cheating way to get the information back, and I wasn't really thinking about any side-effects or long-term use. It really should never have been committed :P

So yes, I agree. Better to use any tooling available that simply redirects the log to stdout/stderr over the network; much more robust, and lot less buggy code for us to maintain.

TrueBrain (approved):

Issue #11353, opened 2023-10-06
[Bug]: incorrect resolution set with multiple monitors


Lamrr:
### Version of OpenTTD

13.4, nightly 20231004; Fedora 38; NVIDIA propriety drivers

### Expected result

I have a multi-monitor setup with two displays, one with a resolution of 1920 x 1080 and another with a resolution of 1680 x 1050.

When configuring OpenTTD I expect to be able to run the game full-screen in the 1080p monitor at its native resolution.

### Actual result

Resolution is consistently limited to the resolution of the other monitor (i.e. 1680 x 1050) when using wayland. Under X11, the resolution is forced to the combined size (i.e. 3600 x 1080) of the monitors. If the resolution is manually set to 1920 x 1080 using openttd.cfg it gets overridden when the game launches.

Using 13.4 build distributed by Fedora;
`openttd -ddriver2`

```
dbg: [driver] Successfully loaded blitter '32bpp-anim'
dbg: [driver] Resolution for display: 1920x1080 [1]
dbg: [driver] SDL2: Mouse is at (0, 0), use display 1 (0, 0, 1920, 1080) [2]
dbg: [driver] SDL2: using mode 1680x1050 [3]
dbg: [driver] SDL2: using driver 'wayland'
dbg: [driver] OpenGL driver: NVIDIA Corporation - NVIDIA GeForce GTX 1060 6GB/PCIe/SSE2 (4.6.0 NVIDIA 535.113.01)
dbg: [driver] OpenGL shading language version: 4.60 NVIDIA, texture units = 192
dbg: [driver] Successfully probed video driver 'sdl-opengl'
dbg: [driver] Successfully probed sound driver 'sdl'
dbg: [driver] Fluidsynth: sf (null)
dbg: [driver] Fluidsynth: samplerate 44100
dbg: [driver] Successfully probed music driver 'fluidsynth'
dbg: [driver] using thread for game-loop
dbg: [driver] Switching blitter from '32bpp-anim' to '40bpp-anim'...
dbg: [driver] Successfully loaded blitter '40bpp-anim'
dbg: [driver] Successfully switched to 40bpp-anim.
dbg: [driver] SDL2: using mode 1680x1050
dbg: [driver] SDL2: using mode 1680x1050
```
Resolution is initially set to 1920x1080 at [1], the right display is grabbed at [2] but the mode used is the maximum resolution of the other display [3].

[...]

PR #11779, opened 2024-01-14
Fix: [SDL2] fullscreen always used the resolution of the first display

TrueBrain:
## Motivation / Problem

When toggling fullscreen, OpenTTD always assumes it would be on display 0. But SDL assumes it is on the display the window is currently at. So on a setup that used different resolutions for different displays, the wrong resolution could be used.

This is the bug mentioned in #11353 classified by me as bug `#1.2`.

## Description

Instead of using display 0 to find the fullscreen resolution, ask SDL what display the window is at, and use that. This is also what SDL uses to build the fullscreen on.

## Checklist for review


glx22 (approved):
Makes sense

Issue #9680, opened 2021-11-09
[Crash]: Loading TTD save causes a crash


AnrDaemon:
### Version of OpenTTD

12.1.0.26020

### Steps to reproduce

Simply loading a provided save causes OpenTTD to crash.

### Upload crash files

[crash-2021-11-09.zip](https://github.com/OpenTTD/OpenTTD/files/7500954/crash-2021-11-09.zip)

Attached is a crash log, crash dump and emergency save.
Also attached a save of `Sept, 13 1953`, which OpenTTD loading just fine, and a save from a day later, which crashes the game.

glx22:
```
>	openttd.exe!GetObjectType(unsigned int t) Ligne 65	C++
 	[Cadre en ligne] openttd.exe!IsObjectType(unsigned int) Ligne 27	C++
 	openttd.exe!GetSlopePixelZ_Object(unsigned int tile, unsigned int x, unsigned int y) Ligne 425	C++
 	openttd.exe!SetAircraftPosition(Aircraft * v, int x, int y, int z) Ligne 539	C++
 	openttd.exe!AircraftLeaveHangar(Aircraft * v, Direction exit_dir) Ligne 1472	C++
 	openttd.exe!UpdateOldAircraft() Ligne 182	C++
 	openttd.exe!AfterLoadGame() Ligne 967	C++
 	openttd.exe!SaveOrLoad(const std::string & filename, SaveLoadOperation fop, DetailedFileType dft, Subdirectory sb, bool threaded) Ligne 3267	C++
 	openttd.exe!SafeLoad(const std::string & filename, SaveLoadOperation fop, DetailedFileType dft, GameMode newgm, Subdirectory lf, LoadFilter *) Ligne 941	C++
 	openttd.exe!SwitchToMode(SwitchMode new_mode) Ligne 1038	C++
 	openttd.exe!GameLoop() Ligne 1448	C++
 	openttd.exe!VideoDriver::GameLoop() Ligne 37	C++
 	openttd.exe!VideoDriver::GameThread() Ligne 46	C++
 	openttd.exe!StartNewThread::__l3::<lambda>(const char * name, void(*)(VideoDriver *) && F, VideoDriver * && <A_0>) Ligne 65	C++
 	[Code externe]	
 	openttd.exe!thread_start<unsigned int (__cdecl*)(void *),1>(void * const parameter) Ligne 97	C++
 	[Code externe]	
```

glx22:
[...]

PR #11786, opened 2024-01-15
Fix #11783: make TTO savegames load again

### LINKED ISSUE #11783: [Crash]: Loading any TTO savegame crashes OpenTTD with "Invalid map size"
### Version of OpenTTD

a6873ef7dd1da93f0d954817d7b68a1621b4a04c on Windows 10, 64-bit

### Steps to reproduce

1. Try to load any TTO savegame.
2. The game crashes with "Invalid map size"

(Is this a true regression, as I remember TTO savegames loading fine as recent as OpenTTD version 1.10, or are all of the provided savegames corrupt?)

All of the TTD savegames I found load fine (though some crash when unpausing for unrelated reasons).

### Upload crash files

The following zip file contains one of the TTO savegames I tried to load, and the resulting crash logs: [crash20240115152431.zip](https://github.com/OpenTTD/OpenTTD/files/13940506/crash20240115152431.zip)

I also have 7 TTO savegames, all of which crash the game on load with the same error: [tto_saves.zip](https://github.com/OpenTTD/OpenTTD/files/13940546/tto_saves.zip)

### LINKED ISSUE #11785: [Crash]: when loading a TTD savegame with vehicles and then unpausing
### Version of OpenTTD

a6873ef7dd1da93f0d954817d7b68a1621b4a04c on Windows 10, 64-bit

### Steps to reproduce

1. Load any TTD savegame with vehicles inside.
2. Unpause the game.
3. The game asserts here after a few ticks: https://github.com/OpenTTD/OpenTTD/blob/a6873ef7dd1da93f0d954817d7b68a1621b4a04c/src/economy.cpp#L1633

### Upload crash files

The following ZIP file contains the TTD savegame that was loaded when reproducing the issue, and the resulting crash logs: [crash20240115180153.zip](https://github.com/OpenTTD/OpenTTD/files/13941802/crash20240115180153.zip)

However, the issue is not specific to that TTD savegame; many other TTD savegames crash OpenTTD when unpaused with the same assertion error.

rubidium42:
## Motivation / Problem

Fixes #11783.
Fixes #11785.

## Description

To solve #11783, I ran into the issue that `OLD_MAP_SIZE` isn't clear in whether it's the number of tiles, or the number of tiles along an edge. Since it is mostly used to loop over the map array, I wondered whether the constant was actually needed that often. Well, it isn't... iterating using `Map::Iterate` works fine in most cases, although there's a bit of a mess with `old_map3` which we split into two values. However, why first copy the data to one array, to copy it into the already allocated map array?

Then there were crashes with resolving towns. It wasn't the issue in the end, but making since depots already had some resiliency baked in, I also used that for industries and stations.
The reason the loading crashed, was because it was reading 4 bytes that are only for TTD when getting the TTO save, and because validating the type did not set the position of the file to the expected location. Solving this made the saves loadable.

All but one would not run, and most were running into the same issue as #11785. The problem here is that for very old saves the aircraft, in some states, are bumped into the air as migrating apparently was not possible. With #9680 a crash was solved by moving this process to later in `AfterLoadGame`, specifically when the map would be valid. However, in between all the map fixing, vehicles with a loading state (before version 57, so including TTO and TTD) were added to the list of loading vehicles at a station. When aircraft are then bumped into the air, they are in the list of loading vehicles but not actually in the loading state. So, simply remove aircraft not in a loading state from those lists again.

That leaves one, `5965 - Trt03.sv1`, that crashes the game when unpausing. However, that also happens in 1.11 and 1.6, which is old enough for me to draw the proverbial line stating that this one likely never loaded correctly in OpenTTD.

## Limitations

Don't have a large corpus of old test games to test with.

## Checklist for review

### DIFF
--- a/src/saveload/oldloader_sl.cpp
+++ b/src/saveload/oldloader_sl.cpp
@@ -957,9 +957,9 @@ static const OldChunks _company_chunk[] = {
 	OCL_CNULL( OC_TTD, 1 ), // avail_railtypes
 	OCL_SVAR(   OC_TILE, Company, location_of_HQ ),
 
-	OCL_NULL( 4 ),           // Shares
+	OCL_CNULL( OC_TTD, 4 ), // Shares
 
-	OCL_CNULL( OC_TTD, 8 ), ///< junk at end of chunk
+	OCL_CNULL( OC_TTD, 8 ), // junk at end of chunk
 
 	OCL_END()
 };
--- a/src/saveload/oldloader_sl.cpp
+++ b/src/saveload/oldloader_sl.cpp
@@ -43,8 +43,6 @@ static bool _read_ttdpatch_flags;    ///< Have we (tried to) read TTDPatch extra
 static uint16_t _old_extra_chunk_nums; ///< Number of extra TTDPatch chunks
 static byte _old_vehicle_multiplier; ///< TTDPatch vehicle multiplier
 
-static uint8_t *_old_map3;
-
 void FixOldMapArray()
 {
 	/* TTO/TTD/TTDP savegames could have buoys at tile 0
@@ -54,13 +52,6 @@ void FixOldMapArray()
 
 static void FixTTDMapArray()
 {
-	/* _old_map3 is moved to _m::m3 and _m::m4 */
-	for (TileIndex t = 0; t < OLD_MAP_SIZE; t++) {
-		Tile tile(t);
-		tile.m3() = _old_map3[t.base() * 2];
-		tile.m4() = _old_map3[t.base() * 2 + 1];
-	}
-
 	for (TileIndex t = 0; t < OLD_MAP_SIZE; t++) {
 		Tile tile(t);
 		switch (GetTileType(tile)) {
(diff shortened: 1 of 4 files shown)


TrueBrain (dismissed):

src/saveload/oldloader_sl.cpp
```diff
@@ -528,18 +532,26 @@ static void ReadTTDPatchFlags()
 	 * 1 vehicle   == 128 bytes */
 	_bump_assert_value = (_old_vehicle_multiplier - 1) * 850 * 128;
 
-	for (uint i = 0; i < 17; i++) { // check tile 0, too
-		if (_old_map3[i] != 0) _savegame_type = SGT_TTDP1;
+	/* The first 17 bytes are used by TTDP1, which translates to the first 9 m3s and first 8 m2s. */
```
glx22:
```suggestion
	/* The first 17 bytes are used by TTDP1, which translates to the first 9 m3s and first 8 m4s. */
```

TrueBrain (approved):

PR #11797, opened 2024-01-16
Codechange: Make window widget lists build-time.

PeterN:
## Motivation / Problem

Window NWidgetPart lists are const, but are actually built on start up because NWidgetPart functions are not constexpr.

Making NWidgetParts constexpr means that runtime variables cannot be used inadvertently.

## Description

WIdgetDimensions::unscaled needs to be moved to the header file to also be constexpr, C++17 should support this without making multiple copies per file that includes it.

Add constexpr NWidgetPart constructors, and make NWidgetPart factory functions constexpr, and make NWIdgetPart lists constexpr too.

## Limitations

Whew

## Checklist for review

### DIFF
--- a/src/core/geometry_type.hpp
+++ b/src/core/geometry_type.hpp
@@ -57,13 +57,13 @@ struct RectPadding {
 	 * Get total horizontal padding of RectPadding.
 	 * @return total horizontal padding.
 	 */
-	inline uint Horizontal() const { return this->left + this->right; }
+	constexpr uint Horizontal() const { return this->left + this->right; }
 
 	/**
 	 * Get total vertical padding of RectPadding.
 	 * @return total vertical padding.
 	 */
-	inline uint Vertical() const { return this->top + this->bottom; }
+	constexpr uint Vertical() const { return this->top + this->bottom; }
 };
 
 inline const RectPadding RectPadding::zero{};
--- a/src/window_gui.h
+++ b/src/window_gui.h
@@ -31,7 +31,8 @@ enum FrameFlags {
 
 DECLARE_ENUM_AS_BIT_SET(FrameFlags)
 
-struct WidgetDimensions {
+class WidgetDimensions {
+public:
 	RectPadding imgbtn;
 	RectPadding inset;
 	RectPadding vscrollbar;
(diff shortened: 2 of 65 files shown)


src/window_gui.h
```diff
@@ -65,6 +66,151 @@ struct WidgetDimensions {
...
 	static WidgetDimensions scaled;         ///< Widget dimensions scaled for current zoom level.
+
+private:
+	/**
+	 * Distances used in drawing widgets.
+	 * These constants should not be used elsewhere, use scaled/unscaled WidgetDimensions instead.
+	 */
+	enum WidgetDrawDistances {
+		/* WWT_IMGBTN(_2) */
+		WD_IMGBTN_LEFT    = 1,      ///< Left offset of the image in the button.
```
TrueBrain, line 77: Now the `WD_` prefix is a bit unneeded. Maybe not for this PR :)

PeterN: To be honest most of these can be removed anyway and replaced with literal numbers in the unscaled definitions. Like I've been doing for new entries.

src/core/geometry_type.hpp
```diff
@@ -57,13 +57,13 @@ struct RectPadding {
 	 * Get total horizontal padding of RectPadding.
 	 * @return total horizontal padding.
 	 */
-	inline uint Horizontal() const { return this->left + this->right; }
+	inline constexpr uint Horizontal() const { return this->left + this->right; }
```
TrueBrain: `[dcl.constexpr], §7.1.5/2 in the C++11 standard): "constexpr functions and constexpr constructors are implicitly inline (7.1.2)`.

Could drop the `inline`. Nitpicking ftw :)

src/window_gui.h
PeterN: Hmm... looks kinda random... https://gist.github.com/PeterN/d083990f8445f71b4828974b48f80fd0

rubidium42: I do like https://gist.github.com/PeterN/d083990f8445f71b4828974b48f80fd0?permalink_comment_id=4834397#gistcomment-4834397. It's quite self explanatory.
Except that left/right/top/bottom is not directly clear. Maybe those could be added for completeness too? But that might clutter it a lot too, so... don't really know.

src/core/geometry_type.hpp
PeterN: Done, there was quite a lot added, so worth it :)

TrueBrain (approved):
The open comment can be picked up in another PR. Your commit just moves code. So I see no reason for that to be a blocker for merging this.

PR #11798, opened 2024-01-16
Fix #10118: Cycle through current signal group, not just path signals

### LINKED ISSUE #10118: [Bug]: «Сycle trough signal types» set to «Path signals only» by default

2TallTyler:
## Motivation / Problem

First, a bit of backstory: Once upon a time, players could chose which signal group to cycle through when Ctrl+Clicking signals: block, path, or all. #8688 changed that to `All visible` or `Path signals only`. My reasoning was that if this was set to block signals but block signals were hidden in the GUI, it would be confusing and hard to fix. Players would be able to build path signals and then convert them into a signal they can't see in the GUI -- a terrible user experience design. Thus the change.

Okay, now to the present. Players occasionally ask for this old setting back, citing the desire to cycle through block signals only (#10118) or swap between one-way and two-way path signals while still showing all types (https://github.com/OpenTTD/OpenTTD/pull/11769#issuecomment-1892751460).

## Description

This PR fulfils both feature requests by implementing @frosch123's idea in https://github.com/OpenTTD/OpenTTD/pull/10119#issuecomment-1299223877. The `Path signals only` option is changed to `Current group only`, consisting of two groups: block/entry/exit/combo and path/one-way path. 

Because of the interplay with the signal GUI mode, there are a couple edge cases which I think are handled in an intuitive way:

* If a block signal exists and the setting is `Current group only`, the signal is converted within its group even if the player cannot see this type of signal in the GUI. To me, this is different from the original problem because the player has already built a block signal -- and it's easily resolved by removing and rebuilding the signal with the correct type, rather than being stuck with the behavior due to a dropdown in the Settings menu.
* If a block signal exists and the setting is `All visible`, it is converted to a path signal because that's what the player can see.

Closes #10118.
Supersedes #11769.

## Checklist for review

### DIFF
--- a/src/table/settings/gui_settings.ini
+++ b/src/table/settings/gui_settings.ini
@@ -519,13 +519,13 @@ cat      = SC_EXPERT
 var      = gui.cycle_signal_types
 type     = SLE_UINT8
 flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_DROPDOWN
-def      = 0
-min      = 0
-max      = 1
+def      = SIGNAL_CYCLE_GROUP
+min      = SIGNAL_CYCLE_GROUP
+max      = SIGNAL_CYCLE_ALL
 interval = 1
 str      = STR_CONFIG_SETTING_CYCLE_SIGNAL_TYPES
 strhelp  = STR_CONFIG_SETTING_CYCLE_SIGNAL_TYPES_HELPTEXT
-strval   = STR_CONFIG_SETTING_CYCLE_SIGNAL_PBS
+strval   = STR_CONFIG_SETTING_CYCLE_SIGNAL_GROUP
 cat      = SC_ADVANCED
 
 [SDTC_VAR]
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1946,7 +1946,7 @@ STR_CONFIG_SETTING_SEMAPHORE_BUILD_BEFORE_DATE_HELPTEXT         :Set the year wh
 STR_CONFIG_SETTING_CYCLE_SIGNAL_TYPES                           :Cycle through signal types: {STRING2}
 STR_CONFIG_SETTING_CYCLE_SIGNAL_TYPES_HELPTEXT                  :Select which signal types to cycle through when Ctrl+Clicking on a built signal with the signal tool
 ###length 2
-STR_CONFIG_SETTING_CYCLE_SIGNAL_PBS                             :Path signals only
+STR_CONFIG_SETTING_CYCLE_SIGNAL_GROUP                           :Current group only
 STR_CONFIG_SETTING_CYCLE_SIGNAL_ALL                             :All visible
 
 STR_CONFIG_SETTING_SIGNAL_GUI_MODE                              :Show signal types: {STRING2}
--- a/src/rail_gui.cpp
+++ b/src/rail_gui.cpp
@@ -241,16 +241,36 @@ static void GenericPlaceSignals(TileIndex tile)
 		Command<CMD_REMOVE_SINGLE_SIGNAL>::Post(STR_ERROR_CAN_T_REMOVE_SIGNALS_FROM, CcPlaySound_CONSTRUCTION_RAIL, tile, track);
 	} else {
 		/* Which signals should we cycle through? */
-		SignalType cycle_start = _settings_client.gui.cycle_signal_types == SIGNAL_CYCLE_ALL && _settings_client.gui.signal_gui_mode == SIGNAL_GUI_ALL ? SIGTYPE_BLOCK : SIGTYPE_PBS;
+		bool tile_has_signal = IsValidTrack(track) && HasSignalOnTrack(tile, track);
+		SignalType cur_signal_on_tile = tile_has_signal ? GetSignalType(tile, track) : _cur_signal_type;
+		SignalType cycle_start;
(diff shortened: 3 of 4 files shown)


ldpl:
Tbh, I don't see why it should even be related to what signal types are visible. It's an action performed in the viewport, not signal toolbar. It's not like block signals disappear from the game if you hide them. If the goal is to not scroll path to block on default settings using "current ground only" as default would suffice.

2TallTyler:
Cycling a signal from a type we can see, to a type we cannot, is not an acceptable action to me. We shouldn't hide footguns in the Settings menu. 🙂

src/table/settings/gui_settings.ini
```diff
@@ -512,13 +512,13 @@ cat      = SC_EXPERT
 var      = gui.cycle_signal_types
 type     = SLE_UINT8
 flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_GUI_DROPDOWN
-def      = 0
-min      = 0
-max      = 1
+def      = SIGNAL_CYCLE_ALL
```
TrueBrain: This default seems odd to me. Wouldn't we want to restrict new players to only cycle through the current signals?

(well, they don't have block signals active, but still, say they do activate that, because of some rant on some forum they read).

TrueBrain:
Unrelated to this PR, but we might also need some way to differentiate the types in the Signal GUI. It is now a bit unclear what is considered a "group".

src/table/settings/gui_settings.ini
2TallTyler: Sure, I can make the other option default. 🙂

DefinitelyNotRau117:
The division into block / PBS groups makes only conceptual sense, but there will be no real usefulness from the ability to switch within the same group.

Using 100% block signals is theoretically possible, but makes no practical sense. The player will still have to switch to the PBS group occasionally.

The ability to switch within a PBS group only makes sense when PBS is the **only** preferred choice. When used together with block signals, one-way pbs loses its meaning - it simply has fewer capabilities than two-way pbs.

My suggestion is to restructure the groups from `block/pbs` to `pre/normal (block + one and two-way PBS)`, and leave the ability to hide all block signals and only show/scroll through PBS.
изображение

In addition – re-implement ctrl+click on a empty track. Simple click to build choosen signal, ctrl+click to place opposite group 1st signal. _If the “PBS only” group is selected, then place regular block-signal._

Now this combination, when using a light signal, places a semaphore, and its usefulness is extremely doubtful.
изображение

Then it will be possible to use the “normal signals” subgroup in a “regular casual game” - block signals on lines, PBS on junctions. And in a more advanced, hardcore-coop-style just use pre-signals via ctrl+click.

2TallTyler:
That suggestion is out of scope for this PR. But I'll explain my opinion anyway:

We hid block signals for a reason. Your frequent request for block signals for aesthetic reasons makes sense -- they are smaller and always green. Fulfilling those feature requests will not involve block signals, however. It would be nice for one-way path signals, when placed on a plain block with no junctions ahead, to default to showing green like block signals. Players who want a smaller signal without the one-way plate could use a GRF to replace semaphore signal sprites with a smaller variant of the electric-light signals. Or we could consider adding a third type of path signal -- if we can make it visually distinct from a real block signal, to avoid confusing players. But in any case, actual block signals will stay hidden.

TrueBrain (approved):
There can be all kinds of arguments here, but in the end, having it cycle only through PBS when there is a block signal is just a bit weird. So this should improve the setting for most people.

Of course there will be the individual space heater. There isn't a golden bullet here.

2TallTyler:
To sum up my understanding of why @DefinitelyNotRau117 is advocating for block signals:
* Path signals are ugly, because
  * They are red by default
  * They are tall and have a big white/red plate on the back

I understand that argument. But in my mind, the solution is to improve path signals, not make it easier to use block signals. Making path signals act like default-green block signals when there's a simple clear block ahead is totally doable (in JGRPP realistic braking mode, I think this is how block signals work -- they are secretly path signals). This is the route we should go, in my mind.

Of course, if someone else makes a PR to propose block signals, I'll consider it, but I won't be putting in the effort to design or code such a thing.

DefinitelyNotRau117:
> To sum up my understanding of why @DefinitelyNotRau117 is advocating for block signals:
> 
>     * Path signals are ugly, because
>    . . .

This is absolutely not what I meant. Appearance is an important, but not the most important thing. What I'm saying is that no matter how much you want to deny it, all signals have their place, and using a combination of all signals can greatly improve the efficiency of a train system. Of course, if this is not important to someone, there is no question, let them use the “only PBS / only current group” option.

> not make it easier to use block signals

What’s the problem with doing good for everyone? 

There is an even simpler and clearer method that will suit both fans of the “versatility” of PBS and... everyone else. Just add 2 more options “scroll only through block signal” and “scroll only through PBS”, for a total of 4 options: “all visible, only current group, only block, only PBS”.

This will not reduce the “convenience” of this PR, but will add it to those who still want to combine all signals.

In addition, I am absolutely sure that adding 2 options does not have any negative consequences. _Except for the slight illogicality of “I can change the signals to those that are not in the interface,” but if the player has changed this "not basic category" setting, most likely they knows what they is doing._

PR #11808, opened 2024-01-16
Codechange: allow certain enumeration to be added

rubidium42:
## Motivation / Problem

In C++20 mode GCC starts complaining a lot that adding enumerations is deprecated, but we are doing it quite often.
For some cases it might be worth it to add casts, but other enumerations such as `Owner`/`CompanyID` and `TextfileType` get added to (widget) enumerations all over the place.

## Description

Create an optional templated `operator+` that allows adding a given enum to any other enum. The left hand side can be any enumeration and the right hand side is the given enum.

This way `WIDGET_X + (CompanyID)i` works just fine, whereas you cannot add random enumerations without warning.
I have chosen for the right hand side to be the given enumeration, as that is the "variable" in most of the cases where it is used. And because it's more logical that you add the "company's index" to a base widget number, than the other way around.

I have added the `operator+` for `Owner` (and by proxy `CompanyID`) as well as `TextfileType`. The other cases do not seem as prevalent.

## Checklist for review

### DIFF
--- a/src/core/enum_type.hpp
+++ b/src/core/enum_type.hpp
@@ -37,5 +37,11 @@
 	inline constexpr enum_type& operator ^= (enum_type& m1, enum_type m2) {m1 = m1 ^ m2; return m1;} \
 	inline constexpr enum_type operator ~(enum_type m) {return (enum_type)(~(std::underlying_type<enum_type>::type)m);}
 
+/** Operator that allows this enumeration to be added to any other enumeration. */
+#define DECLARE_ENUM_AS_ADDABLE(EnumType) \
+	template <typename OtherEnumType, typename = typename std::enable_if<std::is_enum_v<OtherEnumType>, OtherEnumType>::type> \
+	constexpr OtherEnumType operator + (OtherEnumType m1, EnumType m2) { \
+		return static_cast<OtherEnumType>(static_cast<typename std::underlying_type<OtherEnumType>::type>(m1) + static_cast<typename std::underlying_type<EnumType>::type>(m2)); \
+	}
 
 #endif /* ENUM_TYPE_HPP */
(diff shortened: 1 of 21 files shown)


src/core/enum_type.hpp
```diff
@@ -37,5 +37,11 @@
 	inline constexpr enum_type& operator ^= (enum_type& m1, enum_type m2) {m1 = m1 ^ m2; return m1;} \
 	inline constexpr enum_type operator ~(enum_type m) {return (enum_type)(~(std::underlying_type<enum_type>::type)m);}
 
+/** Operator that allows this enumeration to be added to any other enumeration. */
+#define DECLARE_ENUM_AS_ADDABLE(EnumType) \
+	template <typename OtherEnumType, typename = typename std::enable_if<std::is_enum_v<OtherEnumType>, OtherEnumType>::type> \
+	constexpr OtherEnumType operator + (OtherEnumType m1, EnumType m2) { \
+		return (OtherEnumType)(static_cast<typename std::underlying_type<OtherEnumType>::type>(m1) + static_cast<typename std::underlying_type<EnumType>::type>(m2)); \
```
TrueBrain: C-style cast? Should we make it a C++-style while at it? (the `(OtherEnumType)` part)

TrueBrain:
I went a bit further and just did it for all enums we needed to have this behaviour. The alternative is a bunch of casting, which also doesn't actually improve readability. But I am curious what others think about that.

src/linkgraph/linkgraph_gui.cpp
```diff
@@ -728,6 +728,6 @@ void LinkGraphLegendWindow::OnInvalidateData([[maybe_unused]] int data, [[maybe_
 
 	/* Disable the companies who are not active */
 	for (CompanyID i = COMPANY_FIRST; i < MAX_COMPANIES; i++) {
-		this->SetWidgetDisabledState(i + WID_LGL_COMPANY_FIRST, !Company::IsValidID(i));
+		this->SetWidgetDisabledState((uint)i + WID_LGL_COMPANY_FIRST, !Company::IsValidID(i));
```
glx22:
```suggestion
		this->SetWidgetDisabledState(WID_LGL_COMPANY_FIRST + i, !Company::IsValidID(i));
```

glx22 (approved):

PR #11818, opened 2024-01-17
Codechange: use std::source_location over __FILE__/__LINE__ all over the code

rubidium42:
## Motivation / Problem

There is quite a bit of `__FILE__, __LINE__` (or `FILE_LINE`) spread throughout the code, for which C++20 has a nice solution using `std::source_location` and a default argument.

## Description

Replace C-style method with C++-style method for `Backup`, `WindowDesc`, `NOT_REACHED` and `AssertionFailedError`.

## Limitations

There's also `TILE_ADD` and all of `Random`, but those are more involved. So they are left out, so this can remain a relatively simple change.

## Checklist for review

### DIFF
--- a/src/ai/ai_core.cpp
+++ b/src/ai/ai_core.cpp
@@ -40,7 +40,7 @@
 	/* Clients shouldn't start AIs */
 	if (_networking && !_network_server) return;
 
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	Company *c = Company::Get(company);
 
 	AIConfig *config = c->ai_config.get();
@@ -81,7 +81,7 @@
 	assert(_settings_game.difficulty.competitor_speed <= 4);
 	if ((AI::frame_counter & ((1 << (4 - _settings_game.difficulty.competitor_speed)) - 1)) != 0) return;
 
-	Backup<CompanyID> cur_company(_current_company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company);
 	for (const Company *c : Company::Iterate()) {
 		if (c->is_ai) {
 			PerformanceMeasurer framerate((PerformanceElement)(PFE_AI0 + c->index));
@@ -109,7 +109,7 @@
 	if (_networking && !_network_server) return;
 	PerformanceMeasurer::SetInactive((PerformanceElement)(PFE_AI0 + company));
 
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	Company *c = Company::Get(company);
 
 	delete c->ai_instance;
@@ -129,7 +129,7 @@
 	 * for the server owner to unpause the script again. */
 	if (_network_dedicated) return;
 
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	Company::Get(company)->ai_instance->Pause();
 
 	cur_company.Restore();
@@ -137,7 +137,7 @@
 
 /* static */ void AI::Unpause(CompanyID company)
 {
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	Company::Get(company)->ai_instance->Unpause();
 
 	cur_company.Restore();
@@ -145,7 +145,7 @@
 
 /* static */ bool AI::IsPaused(CompanyID company)
 {
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	bool paused = Company::Get(company)->ai_instance->IsPaused();
 
 	cur_company.Restore();
@@ -259,7 +259,7 @@
 	}
 
 	/* Queue the event */
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Backup<CompanyID> cur_company(_current_company, company);
 	Company::Get(_current_company)->ai_instance->InsertEvent(event);
 	cur_company.Restore();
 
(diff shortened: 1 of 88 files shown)


.github/workflows/ci-build.yml
```diff
@@ -203,9 +203,10 @@ jobs:
 
     name: Mac OS (${{ matrix.arch }})
 
-    runs-on: macos-latest
+    runs-on: macos-13
     env:
       MACOSX_DEPLOYMENT_TARGET: 10.13
+      DEVELOPER_DIR: /Applications/Xcode_15.1.app/Contents/Developer
```
TrueBrain: Meh; this will be annoying to remember to remove at some point in time.

You now select 15.1, but 15 would also be sufficient already. 15.2 is released, but not in the image yet. In fact, it was scheduled for the 14th this month to be deployed.

Easy to forget about this, making less-optimal binaries.

On the other hand, `macos-latest` runs an even older XCode, so this is already an improvement :P So I guess I just have to get over the feeling of "meh". Only real concern is security-updates.

rubidium42: I used the path from https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md, and there I'm not seeing that /Applications/Xcode_15.app is a valid path.
Would that be possible? I got no clue. Lets try...

TrueBrain: No, Apple calls 15.0 just 15. And it got a security update, so it is called 15.0.1. I was reading Apple's docs when writing this comment :) There is no "15" to mean "latest in 15 series". There is just a default, which is 14.3.something.

rubidium42: So lets conclude that `std::source_location` support is still "too new" in MacOS/Xcode to include it right now.
We'll have to wait until there's a `macos-latest` that defaults to Xcode 15 or higher. That might be with `macos-14` in about two years (13 is almost a year old, and still not the default for latest), but it might also be that in a few months 13 becomes latest, and the default Xcode becomes 15. It's hard to tell at this moment.

The only remaining question is, what to do with this PR. Currently it won't be merged, but having it open will likely prevent others from attempting the conversion and running into the same issue.

TrueBrain: For sure leave it open.

@glx22 what do you think? Do we merge this now, knowing we have to keep an eye on the XCode version? Or do we postpone till GitHub runners change the default?

PS: bit weird that they don't change the default for a beta-runner, but okay :)

glx22: I think it's probably better to keep it open. Last time we hardcoded image version (it was macos-11 but I may be wrong) it was not changed until the warnings about close drop of the image started to be annoying.

glx22:
[https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/ and ](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/)
So `macos-13` will never be `macos-latest`, seems waiting was a good idea.

rubidium42:
> [https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/ and ](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/) So `macos-13` will never be `macos-latest`, seems waiting was a good idea.

The `macos-14` runner is failing in CMake trying to find a compiler. I hope that'll resolve itself, otherwise we'll have a fun surprise "in Q2 FY24 (April – June 2024)."

TrueBrain:
> > [https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/ and ](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/) So `macos-13` will never be `macos-latest`, seems waiting was a good idea.
> 
> The `macos-14` runner is failing in CMake trying to find a compiler. I hope that'll resolve itself, otherwise we'll have a fun surprise "in Q2 FY24 (April – June 2024)."

Vcpkg isn't installed in the GitHub Runner. Bit odd they didn't, but as it is just new, most likely something they will fix. Otherwise we need to install it.

Also interesting that this will be an arm64 runner. Fun times ahead :)

glx22:
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md mentions vcpkg.
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md doesn't.
(https://github.com/actions/runner-images/blob/main/images/macos/macos-13-arm64-Readme.md doesn't mention vcpkg either)

TrueBrain:
It is intentional that vcpkg is no longer installed in the image, and reading by the reply I suspect it also won't be in other images when they are added. So, we have to install ot ourself again. Not a biggy, and gives a way forward for this PR :)

rubidium42:
> It is intentional that vcpkg is no longer installed in the image, and reading by the reply I suspect it also won't be in other images when they are added. So, we have to install ot ourself again. Not a biggy, and gives a way forward for this PR :)

That's a pity. I hope if adding installing vcpkg can be done before macos-latest gets switched over to macos-14, as otherwise all builds will start failing at some "random" point in time (in Q2). For this PR we still have to wait for macos-14 to become latest.

I'm not sure from where vcpkg would need to be installed, so for me it would be mostly generating failed builds until something sticks. I doubt that's the way to go, so I hope someone else has an idea how to install vcpkg there. With the Ubuntu one it's probably just adding it to the apt call.

TrueBrain:
No worries, after branching 14.0, either me or @glx22 will take care of this, I am sure :)

Not before .. please don't touch the CI before that time :P

TrueBrain:
A rebase should fix the MacOS issue now!

TrueBrain (approved):

PeterN:
Turns out that, for me at least, `std::source_location::current()` as a default argument only gives the location of the call in the constructor, instead of where the object is created.

So this makes the unit tests that output the location of a `WindowDesc` not very useful, as they always say the error is in `window_gui.h`

PeterN:
Supported in clang 16.0:

https://github.com/llvm/llvm-project/commit/ca619613801233ef2def8c3cc7d311d5ed0033cb

But that is not able to build OpenTTD currently.

PR #11824, opened 2024-01-17
Fix #11801, 51f1e93: CalcClosestTownFromTile needs the kd-tree to be valid

### LINKED ISSUE #11801: [Crash]: when loading save from #2126 due to a regression
### Version of OpenTTD

2efd8976914d0a669656143a2cd20426df0b9901 on Windows 10, 64-bit

### Steps to reproduce

1. Load the attached savegame. It should appear as `(TTD) 1town` in the savegame listing.
2. The game asserts here: https://github.com/OpenTTD/OpenTTD/blob/2efd8976914d0a669656143a2cd20426df0b9901/src/core/kdtree.hpp#L443

This is a regression, starting no earlier than the previous nightly (a1690d5b7b).

### Upload crash files

[crash20240116195857.zip](https://github.com/OpenTTD/OpenTTD/files/13955182/crash20240116195857.zip)

rubidium42:
## Motivation / Problem

Fixes #11801.

## Description

Initialize the kd-tree before `CalcClosestTownFromTile`, otherwise it asserts because its empty.
Before 51f1e93 it used `Town::GetRandom`, but since the tile is known the closest is arguably the better solution.

## Limitations

None.

## Checklist for review

### DIFF
--- a/src/saveload/oldloader_sl.cpp
+++ b/src/saveload/oldloader_sl.cpp
@@ -499,7 +499,12 @@ static Town *RemapTown(TileIndex fallback)
 {
 	/* In some cases depots, industries and stations could refer to a missing town. */
 	Town *t = Town::GetIfValid(RemapTownIndex(_old_town_index));
-	if (t == nullptr) t = CalcClosestTownFromTile(fallback);
+	if (t == nullptr) {
+		/* In case the town that was refered to does not exist, find the closest.
+		 * However, this needs the kd-tree to be present. */
+		RebuildTownKdtree();
+		t = CalcClosestTownFromTile(fallback);
+	}
 	return t;
 }
 


TrueBrain (approved):

src/saveload/oldloader_sl.cpp
```diff
@@ -499,7 +499,12 @@ static Town *RemapTown(TileIndex fallback)
 {
 	/* In some cases depots, industries and stations could refer to a missing town. */
 	Town *t = Town::GetIfValid(RemapTownIndex(_old_town_index));
-	if (t == nullptr) t = CalcClosestTownFromTile(fallback);
+	if (t == nullptr) {
+		/* In case the town that was refered to does not exist, find the closest.
+		 * However, this needs the kd-tree to be present. */
+		RebuildTownKdtree();
```
TrueBrain, line 505: One can argue if we should load it every time, but .. it is oldloader .. I don't actually care about performance :D

rubidium42: Exactly my thoughts...

PR #11826, opened 2024-01-18
Fix: [HarfBuzz] make HarfBuzz use the same glyphs as we render

### LINKED ISSUE #11765: [Bug]: Font rendering is ugly on linux
### Version of OpenTTD

Latest git, Arch Linux

### Expected result

I like to use OpenTTD with these settings:
```
small_font = Tahoma
medium_font = Tahoma, Bold
large_font = Times New Roman
mono_font = 
small_size = 0
medium_size = 13
large_size = 0
mono_size = 0
```
On stable 13.4 it looks nice:
image

### Actual result

But somewhere between then and now the spacing between glyphs has been broken:
image

### Steps to reproduce

1. Install Microsoft Tahoma font (Tahomabd.ttf) (using Windows 11 version here but the Windows 7 one should look the same)
2. Set openttd.cfg as above

TrueBrain:
## Motivation / Problem

HarfBuzz loads FreeType font glyphs on its own. This is done with loadflags that are initialized by HarfBuzz. Normally this would be fine, but we also have our own loadflags. And they do not match.

For example, with AA, we tell FreeType we want a monochrome experience. But we never told HarfBuzz. In result, HarfBuzz uses other "advance" values as the font really is, making a total mess of where we render glyphs.

Fixes #11765.

## Description

The above is fixed in the last commit. Also found out that us setting the scale isn't all that useful, and HarfBuzz already takes care of that.

## Checklist for review

### DIFF
--- a/src/gfx_layout_icu.cpp
+++ b/src/gfx_layout_icu.cpp
@@ -157,6 +157,8 @@ ICUParagraphLayout::ICUVisualRun::ICUVisualRun(const ICURun &run, int x) :
 void ICURun::Shape(UChar *buff, size_t buff_length)
 {
 	auto hbfont = hb_ft_font_create_referenced(*(static_cast<const FT_Face *>(font->fc->GetOSHandle())));
+	/* Match the flags with how we render the glyphs. */
+	hb_ft_font_set_load_flags(hbfont, GetFontAAState(this->font->fc->GetSize()) ? FT_LOAD_TARGET_NORMAL : FT_LOAD_TARGET_MONO);
 
 	/* ICU buffer is in UTF-16. */
 	auto hbbuf = hb_buffer_create();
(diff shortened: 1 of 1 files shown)


src/gfx_layout_icu.cpp
```diff
@@ -243,7 +244,7 @@ int ICUParagraphLayout::ICULine::GetLeading() const
  */
 int ICUParagraphLayout::ICULine::GetWidth() const
 {
-	int length = 0;
+	float length = 0;
 	for (const auto &run : *this) {
 		length += run.GetAdvance();
 	}
```
PeterN: `return length;` needs to be `return ceil(length);` because our string rendering system only deals with whole-units. You can verify this by looking at the effect on drop-down lists.

PeterN (approved):
A new day, a fresh (true)brain.

PR #11831, opened 2024-01-18
Fix #11827: Make Layouter::GetCharPosition() aware of ligatures.

### LINKED ISSUE #11827: [Bug]: GetCharPosition does not work with ligatures
### Version of OpenTTD

master

### Expected result

Editing text and moving the text cursor should not crash ;)

### Actual result

Layouter::GetCharPosition() does not consider ligatures, which replaces (at least) a pair of characters with a single glyph.

These are used mostly by non-Latin script, but some Latin fonts also use them (e.g. Tahoma Bold has an 'fi' ligature)

### Steps to reproduce

1. Set medium font to "Tahoma, Bold" 12.
2. Open the console and enter "refit".
3. Use arrow keys to move cursor.

PeterN:
## Motivation / Problem

When ligatures happen the precise individual character position is not known. This currently causes GetCharPosition() to crash.

## Description

Instead, return the previous position (which is that of the ligature). The cursor will appear to not move for each extra character combined into a ligature.

## Limitations

For ligatures of Latin characters this may look odd when editing, but that's better than crashing.

## Checklist for review

### DIFF
--- a/src/gfx_layout.cpp
+++ b/src/gfx_layout.cpp
@@ -244,10 +244,12 @@ Point Layouter::GetCharPosition(std::string_view::const_iterator ch) const
 		if (!IsConsumedFormattingCode(c)) index += line->GetInternalCharLength(c);
 	}
 
+	/* Initial position, returned if character not found. */
+	static const std::vector<Point> zero = { {0, 0} };
+	auto position = zero.begin();
+
 	/* We couldn't find the code point index. */
-	if (str != ch) {
-		return { 0, 0 };
-	}
+	if (str != ch) return *position;
 
 	/* Valid character. */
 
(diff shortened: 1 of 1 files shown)


src/gfx_layout.cpp
```diff
@@ -259,7 +259,9 @@ Point Layouter::GetCharPosition(std::string_view::const_iterator ch) const
 
 		for (int i = 0; i < run.GetGlyphCount(); i++) {
 			/* Matching glyph? Return position. */
-			if ((size_t)charmap[i] == index) {
+			if ((size_t)charmap[i] >= index) {
+				/* If the index is passed then it's part of a ligature, so return the position of the combined glyph. */
+				if (i > 0 && (size_t)charmap[i] > index) i--;
```
rubidium42: Might this need to be a loop when there are more than two character leading to a combined glyph/ligature? Like the ﬃ ligature.

PeterN: No, I don't think so. With a two character ligature, `charmap` looks like this: `[0, 1, 2, 4]`
We're looking for index 3, but got to 4 which is higher, so we go back one and get the position of 2.
With three characters it would look like: `[0, 1, 2, 5]`
We'd be looking for 3 (or 4), but get to 5. So again we go back one and get the position of 2.

PeterN: Hmm, it will fail if there's no character after the ligature though. I'll revise!

rubidium42 (approved):

PR #11832, opened 2024-01-18
Codechange: use C++ functions to find bits

rubidium42:
## Motivation / Problem

C++20 introduced functions to count zeroed bits. With x86/x86-64 architectures, these map to actual CPU instructions.
Why should we then roll our own logic to find the first or last bit, if that count can be used?

## Description

Reimplement `FindFirstBit` and `FindLastBit` making use of `std::countl_zero` and `std::countr_zero`.
Replace other variants of `FindFirstBit` with just `FindFirstBit`.

For x86 architectures the C++ variant uses significantly fewer operations and jumps: https://godbolt.org/z/hKonh8Po8

## Limitations

I'm not sure whether `FIND_FIRST_BIT` is faster than the new `FindFirstBit`. In a tight loop it might have been faster, but it does require a memory lookup (or one of the caches if you're lucky) making it potentially quite slow. However, I do not know how to reliably prove this, since the calls to `FindFirstBit` aren't common enough to not be overwhelmed by the standard run-to-run variability.

## Checklist for review


src/core/bitmath_func.hpp
```diff
@@ -183,47 +183,39 @@ inline T ToggleBit(T &x, const uint8_t y)
...
 
 /**
- * Finds the position of the first non-zero bit in an integer.
- *
- * This function returns the position of the first bit set in the
- * integer. It does only check the bits of the bitmask
- * 0x3F3F (0011111100111111) and checks only the
- * bits of the bitmask 0x3F00 if and only if the
- * lower part 0x00FF is 0. This results the bits at 0x00C0 must
- * be also zero to check the bits at 0x3F00.
```
PeterN, line 210: What happens to these rules of FindFirstBit2x64?

Eddi-z: do they even matter? as far as i can tell that's meant to mask out unused bits from TrackdirBits, but when are those ever set?

PeterN: Hmm, I see, the rule is basically only in place because it's using a limited size lookup table.

PeterN (approved):
Less code is better for sure.

PR #11834, opened 2024-01-18
Cleanup: alignment related endian code

rubidium42:
## Motivation / Problem

There is code worrying specially about the alignment of data. That feels worrisome.
Good news, it's only two functions and a define, and only one functions is actually used (just once).

## Description

Replace the single used of `ReadLE16Aligned` with just `FROM_LE16`. With the current user the function makes no sense at all, as it's essentially getting the pointer of a value, casting it to void, then casting it back to its original type, to dereference it and pass that through `FROM_LE16`. In other words, just replace it with `FROM_LE16` and do not get the pointer.

Furthermore this code updated the offsets table in a loop (guarded by `#ifdef`s), after which it would loop over the same table to get the offset and do some work. Since this is the only place where the offset is accessed, we can just `FROM_LE16` the data there. Simple and effective.

Finally remove `ReadLE16Aligned`, `ReadLE16Unaligned` and `OTTD_ALIGNMENT`.

## Checklist for review

### DIFF
--- a/src/core/endian_func.hpp
+++ b/src/core/endian_func.hpp
@@ -38,18 +38,4 @@
 #	define TO_LE32X(x)  (x)
 #endif /* TTD_ENDIAN == TTD_BIG_ENDIAN */
 
-inline uint16_t ReadLE16Aligned(const void *x)
-{
-	return FROM_LE16(*(const uint16_t*)x);
-}
-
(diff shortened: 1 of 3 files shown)


TrueBrain (approved):
"Looks fine".

Nobody is going to notice anyway if it isn't. So .. SURE! :)

PR #11835, opened 2024-01-18
Codechange: Use vector of points instead of interlaced x/y values for layouter positions.

PeterN:
## Motivation / Problem

Dealing with layouter glyph positions is confusing because it's an interlaced vector of x/y coordinates, requiring multiplications and additions to get either x or y position.

## Description

Replace the vectors of floats with vectors of Points, which stores both x & y coordinate together in a way that accessing them is much clearer.

This also uses emplace_back() to add each point in one go, instead of using array indices and multipliers everywhere.

## Limitations

Tested with fallback and ICU/harfbuzz layouters on Linux, needs testing with Windows and MacOS.

## Checklist for review

### DIFF
--- a/src/gfx_layout.cpp
+++ b/src/gfx_layout.cpp
@@ -260,8 +260,7 @@ Point Layouter::GetCharPosition(std::string_view::const_iterator ch) const
 		for (int i = 0; i < run.GetGlyphCount(); i++) {
 			/* Matching glyph? Return position. */
 			if ((size_t)charmap[i] == index) {
-				Point p = { (int)positions[i * 2], (int)positions[i * 2 + 1] };
-				return p;
+				return positions[i];
 			}
 		}
 	}
@@ -291,8 +290,8 @@ ptrdiff_t Layouter::GetCharAtPosition(int x, size_t line_index) const
 			/* Not a valid glyph (empty). */
 			if (glyphs[i] == 0xFFFF) continue;
 
-			int begin_x = (int)positions[i * 2];
-			int end_x   = (int)positions[i * 2 + 2];
+			int begin_x = positions[i].x;
+			int end_x   = positions[i + 1].x;
 
 			if (IsInsideMM(x, begin_x, end_x)) {
 				/* Found our glyph, now convert to UTF-8 string index. */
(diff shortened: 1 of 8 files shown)


src/gfx_layout.cpp
```diff
@@ -291,8 +290,8 @@ ptrdiff_t Layouter::GetCharAtPosition(int x, size_t line_index) const
 			/* Not a valid glyph (empty). */
 			if (glyphs[i] == 0xFFFF) continue;
 
-			int begin_x = (int)positions[i * 2];
-			int end_x   = (int)positions[i * 2 + 2];
+			int begin_x = (int)positions[i].x;
+			int end_x   = (int)positions[i + 1].x;
```
TrueBrain: You can do away with the casts now, not?

PeterN: Good ... point ... hah. Coat.

TrueBrain (dismissed):
I am sure this breaks the layouter for some OS somewhere for some reason .. but I can't find it. So I guess it is fine :)

src/os/windows/string_uniscribe.cpp
```diff
@@ -478,30 +474,29 @@ int UniscribeParagraphLayout::UniscribeLine::GetWidth() const
...
 								: glyphs(std::move(other.glyphs)), positions(std::move(other.positions)), char_to_glyph(std::move(other.char_to_glyph)),
-								  start_pos(other.start_pos), total_advance(other.total_advance), num_glyphs(other.num_glyphs), font(other.font)
+								  start_pos(other.start_pos), total_advance(other.total_advance), num_glyphs(other.num_glyphs), font(other.font),
+								  glyph_to_char(std::move(other.glyph_to_char))
 {
-	this->glyph_to_char = other.glyph_to_char;
-	other.glyph_to_char = nullptr;
 }
 
 const int *UniscribeParagraphLayout::UniscribeVisualRun::GetGlyphToCharMap() const
```
glx22: The signature doesn't match

glx22 (approved):
Tested on windows, no visual diff spotted.

reldred:
for future reference, this was checked with a mac as well

PR #11837, opened 2024-01-18
Change: use a stronger hash and actual random information to generate Uids

TrueBrain:
## Motivation / Problem

Slowly it is time that we add actual encryption to the game, for various of things. For example, and what this PR is about: using it to generate UIDs.

Currently this is still done with MD5, which is known to be weak, and it builds on not-so-random information.

Later PRs could start using monocypher for things like:
- Encryption of the rcon communication (as done in JGRPP)
- Signature validation of social plugins (the actual reason I am starting with this), to ensure the plugins are written by a trusted partner (or us, ofc).

## Description

I took a look at JGRPP, and cherrypicked what I liked:

- JGRPP uses `monocypher`, which seems like an excellent fit. I did update to the latest version, but that is a minor difference.
- JGRPP uses `randombytes`, but this doesn't use `BCryptGenRandom` for Windows yet, and instead a deprecated API. As `BCryptGenRandom` exists since Vista, I rather avoid using already deprecated things (JGRPP added it 3 years ago, when it was not deprecated yet). So instead, I went for a minimalist function, [as advised by monocypher](https://monocypher.org/manual/#Random_number_generation) (I assume they know what they are talking about).
- Stole the `GenerateUid` from JGRPP, but updated it to C++-style, and added time in the hashed value. Just because our version already had that by now, and .. why not?

Windows and glibc variant of `RandomBytesWithFallback` can fail, but `arc4random_buf` cannot. I also added a fallback, that is really terrible in terms of security; but at least it is better than nothing, I guess. For visibility a compile-warning is thrown when this is likely going to be used.

Tnx @JGRennison for already paving the road, it is really appreciated.

## Checklist for review


src/misc.cpp
```diff
@@ -58,28 +65,47 @@ void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
 
+void RandomBytesWithFallback(void *buf, size_t bytes)
```
TrueBrain: Can be a `std::span`!

src/misc.cpp
```diff
@@ -58,28 +65,47 @@ void InitializeCheats();
...
+#if defined(_WIN32)
+	auto res = BCryptGenRandom(nullptr, static_cast<PUCHAR>(buf), static_cast<ULONG>(bytes), BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+	if (res >= 0) return;
+#elif defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+	 arc4random_buf(buf, bytes);
+#elif defined(__GLIBC__) && ((__GLIBC__ > 2) || ((__GLIBC__ == 2) && (__GLIBC_MINOR__ >= 25)))
+	auto res = getrandom(buf, bytes, 0);
+	if (res > 0 && static_cast<size_t>(res) == bytes) return;
+#else
+#	warning "No high quality random source available; using a fallback instead"
```
TrueBrain: Needs a call to Javascript to the `crypto` part, to generate the bytes for us.

src/3rdparty/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_subdirectory(catch2)
 add_subdirectory(fmt)
 add_subdirectory(icu)
 add_subdirectory(md5)
+add_subdirectory(monocypher)
```
rubidium42, line 5: The library needs to be mentioned in README.md 3.0 Licencing.

Since the library is a more security sensitive, maybe we should arrange something like the depend bot for especially this library. That might then be applied to some other libraries.

I do like that there has been an external audit of the code and no really shocking things have come out of that.

src/3rdparty/monocypher/monocypher.cpp
```diff
@@ -66,8 +66,8 @@ namespace MONOCYPHER_CPP_NAMESPACE {
 #define ZERO(buf, size)            FOR(_i_, 0, size) (buf)[_i_] = 0
 #define WIPE_CTX(ctx)              crypto_wipe(ctx   , sizeof(*(ctx)))
 #define WIPE_BUFFER(buffer)        crypto_wipe(buffer, sizeof(buffer))
-#define MIN(a, b)                  ((a) <= (b) ? (a) : (b))
-#define MAX(a, b)                  ((a) >= (b) ? (a) : (b))
+#define MC_MIN(a, b)               ((a) <= (b) ? (a) : (b))
+#define MC_MAX(a, b)               ((a) >= (b) ? (a) : (b))
```
rubidium42: Maybe upstream this issue?

src/misc.cpp
```diff
@@ -58,28 +67,61 @@ void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
 
+void RandomBytesWithFallback(std::span<uint8_t> buf)
+{
+#if defined(_WIN32)
+	auto res = BCryptGenRandom(nullptr, static_cast<PUCHAR>(buf.data()), static_cast<ULONG>(buf.size()), BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+	if (res >= 0) return;
+#elif defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+	 arc4random_buf(buf.data(), buf.size());
```
rubidium42: There's a space before `arc4random`, but more importantly: isn't this call as it stands now pointless? You will `buf` with random, and they you'll go into the fallback filling it with `InteractiveRandom`.

src/misc.cpp
```diff
@@ -58,28 +67,61 @@ void InitializeCheats();
 void InitializeNPF();
 void InitializeOldNames();
 
+void RandomBytesWithFallback(std::span<uint8_t> buf)
```
rubidium42: Missing doxygen.

src/misc.cpp
```diff
@@ -58,28 +67,61 @@ void InitializeCheats();
...
- * subjects. But as this is not an UUID, so it can't be ruled out either.
+ * It isn't as much of an unique ID but more a hashed digest of a random
+ * string and a time. It is very likely to be unique, but it does not follow
+ * any UUID standard.
  */
 std::string GenerateUid(std::string_view subject)
 {
-	auto current_time = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now().time_since_epoch()).count();
-	std::string coding_string = fmt::format("{}{}{}", InteractiveRandom(), current_time, subject);
+	uint8_t random_bytes[32] = {};
```
rubidium42: What about the more C++ way of doing this:
```suggestion
	std::array<uint8_t, 32> random_bytes = {};
```
and `crypto_blake2b_update(&ctx, random_bytes.data(), random_bytes.size());` later on?

src/misc.cpp
TrueBrain: Nice catch! Happens if you can't test a line of code :D

src/misc.cpp
TrueBrain: Does that actually matter? All I see is more letters to type, and both are used in the C++ world. Is there a benefit from one over the other?

rubidium42: It's mostly that raw arrays can lead to nasty effects when being passed as parameters; see https://wiki.sei.cmu.edu/confluence/display/c/ARR01-C.+Do+not+apply+the+sizeof+operator+to+a+pointer+when+taking+the+size+of+an+array
https://stackoverflow.com/a/77106115 also lists quite a number of benefits of `std::array` over raw array.

I'd be in favour of using `std::array`/`std::span` over raw arrays, but it'll definitely take a while before the code base has migrated to that. Though, it's something to strive for I'd say.

TrueBrain: You mentioned it lists "quite a number", but I actually can't find any? It only mentions two, and those things we already mitigate in our codebase (for example via `lengthof` or are not applicable in this case?

To be clear, I already changed the things into `std::array`, but I am mostly curious to why it would be better, if there is any fundamental reason. For things like `static_cast` I know there are, but here I have a hard time finding the benefit of one over the other.

For `std::span` it btw is clear: keeping ptr + len combined together, that is an obvious improvement for a wide set of reasons. But for `std::array` I have a hard time seeing them; and I like to understand, not blindly follow :)

TrueBrain: I think https://stackoverflow.com/a/77106115 is the first place I managed to find which actually lists a set of benefits. Which are interesting. Most do not apply to this situation ofc, and mostly have to do with passing an array around (which we already do in spans), and how to initialize them (which is irrelevant here).

The one that stands out is:

```
int data[] = {0, 1};
int copy[] = data;
```

Which is invalid. But works with `std::array`. And that it doesn't allow VLA, which is a positive thing.

michicc:
No idea how good or bad `arc4random_buf` is, but for macOS you maybe want [SecRandomCopyBytes](https://developer.apple.com/documentation/security/1399291-secrandomcopybytes) instead.

TrueBrain:
> No idea how good or bad `arc4random_buf` is, but for macOS you maybe want [SecRandomCopyBytes](https://developer.apple.com/documentation/security/1399291-secrandomcopybytes) instead.

As I am really not an expert in this area, I went with the opinion of someone who is:
https://monocypher.org/manual/#Random_number_generation

And I just followed that. I think this is the safest approach for us :)

src/3rdparty/monocypher/monocypher.cpp
TrueBrain: Turns out, this is only because of our precompiled headers. Normally this wouldn't be an issue. So yeah, we have to fix it here, and it doesn't belong upstream :)

rubidium42 (approved):

PR #11841, opened 2024-01-19
Fix #11840: Return a valid trackdir even when path is not found

### LINKED ISSUE #11840: [Bug]: Ship is stuck
### Version of OpenTTD

20240119-master-g6a2c37b4ed

### Expected result

Ship 1763 finds the way. The path is valid, unobstructed.

### Actual result

Ship 1763 is stuck. The path is valid, unobstructed.

https://github.com/OpenTTD/OpenTTD/assets/43006711/4638bd43-15d2-45ae-8252-35390dc94031

### Steps to reproduce

[Unnamed, 1941-04-02.zip](https://github.com/OpenTTD/OpenTTD/files/13989947/Unnamed.1941-04-02.zip)

SamuXarick:
## Motivation / Problem
Fix #11840 
Incorrect usage of `||` caused ship to get stuck.

## Description
Return a valid trackdir even when path is not found.

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -244,14 +244,14 @@ public:
 			path_found = pf.FindPath(v);
 			Node *node = pf.GetBestNode();
 			if (attempt == 0 && !path_found) continue; // Try again with restricted search area.
-			if (!path_found || !node) return INVALID_TRACKDIR;
+			if (!path_found && node == nullptr) break;
 
 			/* Return only the path within the current water region if an intermediate destination was returned. If not, cache the entire path
 			 * to the final destination tile. The low-level pathfinder might actually prefer a different docking tile in a nearby region. Without
 			 * caching the full path the ship can get stuck in a loop. */
 			const WaterRegionPatchDesc end_water_patch = GetWaterRegionPatchInfo(node->GetTile());
 			const WaterRegionPatchDesc start_water_patch = GetWaterRegionPatchInfo(tile);
-			while (node->m_parent) {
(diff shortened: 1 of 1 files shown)


src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -244,7 +244,7 @@ class CYapfFollowShipT
 			path_found = pf.FindPath(v);
 			Node *node = pf.GetBestNode();
 			if (attempt == 0 && !path_found) continue; // Try again with restricted search area.
-			if (!path_found || !node) return INVALID_TRACKDIR;
+			if (!path_found && !node) return INVALID_TRACKDIR;
```
PeterN: While you are here, that should be `node == nullptr`

Kuhnovic:
I don't see how this is an improvement. No path was found, but you will get a direction to the best node instead. This will get the ship moving, but the ship is still lost.

Judging from your recent messages on Discord (around this post https://discord.com/channels/142724111502802944/1008473233844097104/1197902413185691648 ), I think you ended up with a situation where a ship got stuck, and this will "dislodge" the ship so it ends up in a place where it can successfully find a path again. That's fixing symptoms and not causes.

SamuXarick:
It's what yapf used to do, it's what npf does. If that was wrong before, it at least didn't cause the ship to be stuck. 😦I checked it.

What may be of concern is CPU usage: lowering `max_search_nodes` for the low-level pf should alleviate it, considering the high-level pf returned no path.

Another issue is the random path choice. I'm of the opinion that ships shouldn't wander out there if the ship finds no path but still have a best node to get the closest they can to the destination.

PR #11846, opened 2024-01-20
Codechange: use std::shared_ptr for vector of TCPConnecters

rubidium42:
## Motivation / Problem

Multiple CodeQL warnings "Resource not released in destructor". Especially in the network coordinator.

My main problem is that there `new` calls all over the place, but the `delete` is hidden somewhere far away in the internals. This makes it difficult to reason, and the straight forward way of putting them in `std::unique_ptr` at the point of construction is wrong.

## Description

Make the collection of `TCPConnecter`'s store `std::unique_ptr` instead of raw pointers, thus having a clear location with the ownership of the `TCPConnecter`'s memory.
Introduce a static templated method to create an instance of a sub class of `TCPConnecter`, which puts it into the collector of connecters (instead of the constructors) and then returns a non-owning pointer to those that are interested in it.

Use vector's `clear` to kill all connection attempts, since the `unique_ptr` handles all the further clean up.

Use `std::remove_if` to simplify the `CheckCallbacks` function, i.e. not needing the manual iterator updates depending on whether it got removed or not.

## Checklist for review

### DIFF
--- a/src/network/core/tcp.h
+++ b/src/network/core/tcp.h
@@ -100,6 +100,8 @@ private:
 	NetworkAddress bind_address;                        ///< Address we're binding to, if any.
 	int family = AF_UNSPEC;                             ///< Family we are using to connect with.
 
+	static std::vector<std::shared_ptr<TCPConnecter>> connecters; ///< List of connections that are currently being created.
+
 	void Resolve();
 	void OnResolved(addrinfo *ai);
 	bool TryNextAddress();
@@ -132,6 +134,18 @@ public:
 
 	static void CheckCallbacks();
 	static void KillAll();
+
+	/**
+	 * Create the connecter, and initiate connecting by putting it in the collection of TCP connections to make.
+	 * @tparam T The type of connecter to create.
+	 * @param args The arguments to the constructor of T.
+	 * @return Shared pointer to the connecter.
+	 */
+	template <class T, typename... Args>
+	static std::shared_ptr<TCPConnecter> Create(Args&& ... args)
+	{
+		return TCPConnecter::connecters.emplace_back(std::make_shared<T>(std::forward<Args>(args)...));
+	}
 };
 
 class TCPServerConnecter : public TCPConnecter {
--- a/src/network/core/tcp_connect.cpp
+++ b/src/network/core/tcp_connect.cpp
@@ -18,8 +18,7 @@
 
 #include "../../safeguards.h"
 
-/** List of connections that are currently being created */
-static std::vector<TCPConnecter *> _tcp_connecters;
+/* static */ std::vector<std::shared_ptr<TCPConnecter>> TCPConnecter::connecters;
 
 /**
  * Create a new connecter for the given address.
@@ -32,8 +31,6 @@ TCPConnecter::TCPConnecter(const std::string &connection_string, uint16_t defaul
 	family(family)
 {
 	this->connection_string = NormalizeConnectionString(connection_string, default_port);
-
-	_tcp_connecters.push_back(this);
 }
 
 /**
@@ -57,8 +54,6 @@ TCPServerConnecter::TCPServerConnecter(const std::string &connection_string, uin
 		default:
 			NOT_REACHED();
 	}
-
-	_tcp_connecters.push_back(this);
 }
 
 TCPConnecter::~TCPConnecter()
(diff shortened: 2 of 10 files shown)


src/network/core/tcp.h
```diff
@@ -132,6 +134,18 @@ class TCPConnecter {
...
+	/**
+	 * Create the connecter, and initiate connecting by putting it in the collection of TCP connections to make.
+	 * @tparam T The type of connecter to create.
+	 * @param args The arguments to the constructor of T.
+	 * @return Non-owning pointer to T.
+	 */
+	template <class T, typename... Args>
+	static T *Create(Args&& ... args)
+	{
+		return static_cast<T *>(TCPConnecter::connecters.emplace_back(std::make_unique<T>(std::forward<Args>(args)...)).get());
```
TrueBrain: This feels so wrong to do ... this is heavily abusing what an unique-ptr stands for. And I am rather unsure if this is actually an improvement.

I would say, we could better make it a shared-ptr, and also change things like `this->game_connecter` to be a shared-ptr.

But to have an unique-ptr, and steal its pointer to give to other to do what-ever they want with it .. I am not sure we actually solving any problems here :) Mostly, and this has been a problem, it doesn't solve that `this->game_connecter` still has a pointer while the object is already deleted.

rubidium42: There once was the idea of [`std::observer_ptr`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1408r0.pdf), which is what this function would report. But that has been killed off, and `T*` is the de-facto standard for non-owning pointers. That's exactly what it's returning here.
So, I'm not making a `unique_ptr` to steal it's pointer, it's giving a pointer to observe the object.

What used to happen was creating the connecter and secretly (in the constructor) the ownership of the object was hijacked. The network coordinator was not allowed to delete the `game_connecter`, even though it created the object itself. So in that sense, the new code is better as the owner gives an observer pointer to the network coordinator.

In all code paths where the `unique_ptr` gets deleted, there is first a call of some sort that sets `game_connecter` to `nullptr`. So, there is no problem.

Having said that, I do agree that `std::shared_ptr` is the safer option in the long run.

src/network/core/tcp.h
```diff
@@ -132,6 +134,18 @@ class TCPConnecter {
 
 	static void CheckCallbacks();
 	static void KillAll();
+
+	/**
+	 * Create the connecter, and initiate connecting by putting it in the collection of TCP connections to make.
+	 * @tparam T The type of connecter to create.
+	 * @param args The arguments to the constructor of T.
+	 * @return Non-owning pointer to T.
```
TrueBrain: Guess this is no longer "non-owning"? Shared ownership :D

src/network/core/tcp_connect.cpp
```diff
@@ -467,24 +462,14 @@ void TCPServerConnecter::SetFailure()
...
 
 /** Kill all connection attempts. */
 /* static */ void TCPConnecter::KillAll()
 {
-	for (auto iter = _tcp_connecters.begin(); iter < _tcp_connecters.end(); /* nothing */) {
-		TCPConnecter *cur = *iter;
-		iter = _tcp_connecters.erase(iter);
-		delete cur;
-	}
+	TCPConnecter::connecters.clear();
```
TrueBrain, line 474: The stupid issue we have now, and this is more what I mentioned earlier, that how we handle connections is gotten so weird in our code, it is hard to untangle. But, the stupid issue: in the old code, the dtor was called when KillAll was called, which causes the connection to be closed. Which could leave `this->game_connecter` in an invalid state.

But now, it will remain open, as there is still a shared-ptr active. An improvement for sure, as there are no dangling things anymore, but it also means `KillAll` isn't actually what it sounds like.

So I wonder if we should just rename this function, or actually address the underlying issue, which requires a bit more rework.

Ugh, every time I look at this part of the code ... can we just CTRL+A DELETE it, and do it again? :D

rubidium42: Well, `KillAll` definitely kills all attempts. It just doesn't kill those that were doing the attempts.

I'm not going to worry about the `this->game_connecter`s being in a weird state in this PR, as they cannot be. `KillAll` is only called after the network coordinator client is told to close all connections. That calls `Kill` of the `game_connecter` and closes the turn and stun handler, which were all the classes that had a reference to the `TCPConnecter`.

I'd propose small steps, and leaving the not so desirable situation as-is for this PR. Especially since it's currently not broken due to the way everything is called.

TrueBrain (approved):

PR #11849, opened 2024-01-20
Codechange: Replace GroupStatistics' num_engines with map.

PeterN:
## Motivation / Problem

GroupStatistics has a manually memory managed array to count the number of engines of each type.

## Description

Replace with a map. This removes manual memory management with calloc/free calls, and allows only the needed entries to be populated, instead of an entry for every possible engine ID.

## Checklist for review

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -24,14 +24,11 @@ extern GroupPool _group_pool; ///< Pool of groups.
 struct GroupStatistics {
 	Money profit_last_year;                 ///< Sum of profits for all vehicles.
 	Money profit_last_year_min_age;         ///< Sum of profits for vehicles considered for profit statistics.
-	uint16_t *num_engines;                    ///< Caches the number of engines of each type the company owns.
+	std::map<EngineID, uint16_t> num_engines; ///< Caches the number of engines of each type the company owns.
 	uint16_t num_vehicle;                     ///< Number of vehicles.
 	uint16_t num_vehicle_min_age;             ///< Number of vehicles considered for profit statistics;
 	bool autoreplace_defined;               ///< Are any autoreplace rules set?
 	bool autoreplace_finished;              ///< Have all autoreplacement finished?
-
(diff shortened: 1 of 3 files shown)


TrueBrain (dismissed):

src/group.h
```diff
@@ -24,14 +24,13 @@ extern GroupPool _group_pool; ///< Pool of groups.
 struct GroupStatistics {
 	Money profit_last_year;                 ///< Sum of profits for all vehicles.
 	Money profit_last_year_min_age;         ///< Sum of profits for vehicles considered for profit statistics.
-	uint16_t *num_engines;                    ///< Caches the number of engines of each type the company owns.
+	std::vector<uint16_t> num_engines; ///< Caches the number of engines of each type the company owns.
```
glx22: All these comment not aligned :)

TrueBrain: Good! Let's stop trying to align them! :D

LordAro:
A vector that is just immediately resized to the maximum doesn't feel like good usage to me

PeterN:
> A vector that is just immediately resized to the maximum doesn't feel like good usage to me

It's possible a std::map would be a better fit here, presumably most groups have only a few types of engine. They'll never have all types...

PeterN:
Now using a std::map instead, so nothing is allocated initially, and only engines used in the group are counted.

`GetNumEngines()` exists to prevent making a new entry when we only need to find a value.

src/group_cmd.cpp
```diff
@@ -48,8 +38,19 @@ void GroupStatistics::Clear()
 	this->profit_last_year_min_age = 0;
 
 	/* This is also called when NewGRF change. So the number of engines might have changed. Reallocate. */
```
TrueBrain: Comment is now out-of-date.

TrueBrain (approved):

PR #11850, opened 2024-01-20
Codechange: Use per-vehicle-type list when ticking vehicles.

PeterN:
## Motivation / Problem

As per #7247, using PerformanceAccumulator when ticking vehicles can be a performance hit on some systems, as start/stop time needs to be determined fore each vehicle. PerformanceMeasurer is more effecient as only one start/stop time is needed, but then it is not possible to measure per vehicle type.

## Description

This PR implements a per-vehicle type std::map (ordered by vehicle index) of vehicles, each of which can be ticked in one go. This may improve performance.

it will also likely increase the reported value for trains, as due to #10055, not all parts of Train::Tick are counted.

## Limitations

Dodgy iterators :)

## Checklist for review

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -9,6 +9,7 @@
 
 #include "stdafx.h"
 #include "error.h"
+#include "disaster_vehicle.h"
 #include "roadveh.h"
 #include "ship.h"
 #include "spritecache.h"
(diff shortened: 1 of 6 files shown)


src/vehicle.cpp
```diff
@@ -950,92 +951,115 @@ static void RunVehicleDayProc()
...
-						if (!Aircraft::From(v)->IsNormalAircraft()) continue;
-						break;
+template <class T, typename Pred>
+static void IterateTickable(Pred pred)
+{
+	for (auto it = std::begin(T::tickable_vehicles), next_it = it; it != std::end(T::tickable_vehicles); it = next_it) {
+		++next_it;
+		pred(it->second);
+	}
+}
```
JGRennison, line 1035: This does not guarantee that the iterator is not invalidated.
When pred is something like CommonVehicleTick and Vehicle::Tick deletes the vehicle, any of the articulated/following parts via Vehicle::next will also be deleted, likely deleting the map item pointed to by next_it.

PeterN: Ah yes, multiple deletions...

PeterN: I guess deleting a vehicle could add it to a list of vehicles to be removed, but then the iteration would iterate deleted vehicles :/

JGRennison: Only iterating head vehicles would likely make handling deletions easier.

PeterN:
Flawed

PR #11855, opened 2024-01-21
Fix #6377: two tarballs with the same folder in them were considered as one

### LINKED ISSUE #6377: a bug in path handling of tar files
**_telanus_ opened the ticket and wrote:**

> The path handling of tar files seems not to be working.
> 
> When two of the same Newgrf's are in the same dir IE ogfx-airports.tar and ogfx-airports-v5748-80.tar, OpenTTD only sees ogfx-airports.tar and any game that is started with ogfx-airports-v5748-80.tar will "update" to ogfx-airports.tar. Unpacking the files into their own Directory "solves" this problem but isn't ideal.
> 
> Thread where it was "first" discussed, http://www.tt-forums.net/viewtopic.php?p=1150933# p1150933

**Reported version**: _trunk_
**Operating system**: _Windows_

---

TrueBrain:
## Motivation / Problem

When you have two tarballs, both with the same foldername in them, they are considered to be part of the same folder. This means they merged together.

This is especially tricky if they also contain the same files, but with different content. Which file is actually being seen by OpenTTD heavily depends on in which order the game loads the tarballs.

This was identified in #6377, but never resolved. Yes, I am solving 9 year old bugs that were marked "good first issues". Also, given nobody replied to the ticket in 7 years, it seems to not be that big of a problem. But okay.

Fixes #6377.

## Description

Although #6377 makes it sound like it is trivial to fix, and that "It is now no longer needed, but noone removed it.", it doesn't actually say what to remove. Turns out, it is not removing anything, it is about adding stuff.

Prefix all files added to the tar-list with the tarname they came from. This means that even if two tarballs have the same folder in them, they are different from one another.

## Limitations

This is really hard to test. Although this "should be fine", and initial testing shows that base graphics, NewGRFs and AIs work fine, it is very hard to say if not something made an assumption somewhere in how deep an `info.nut` or whatever has to be.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -529,6 +529,9 @@ bool TarScanner::AddFile(const std::string &filename, size_t, [[maybe_unused]] c
 
 	_tar_list[this->subdir][filename] = std::string{};
 
+	std::string filename_base = std::filesystem::path(filename).filename().string();
+	SimplifyFileName(filename_base);
+
 	TarLinkList links; ///< Temporary list to collect links
 
 	TarHeader th;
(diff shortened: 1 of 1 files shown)


LordAro:
What happens if there are two tarballs with the same name in different folders? :)

TrueBrain:
> What happens if there are two tarballs with the same name in different folders? :)

Different folders were already handled, as tarballs are indexed by the folder they are found int.

rubidium42 (approved):
There is a visible effect on the configuration file: the filename of the tar gets included in the NewGRF configuration/presets.
As such, this change will affect the functionality of lines 1055-1059 of settings.cpp. So I guess this breaks loading NewGRF configurations from before 1.5.0.
I would say: tough luck on that.

PR #11857, opened 2024-01-21
Codechange: Replace C-style TICC/TOCC macros with C++ RAII implementation.

PeterN:
## Motivation / Problem

The TICC/TOCC debug macros have some issues. They are:

1) Macros.
2) They hide a block.
3) If you return from inside the block the timing is skipped.

## Description

Replace with a class that performs the timing when its destructor is called. An extra class is used for storing static state.

Usage is more robust, and will always include timing when the TicToc object goes out of scope.

## Checklist for review

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -63,44 +63,38 @@ std::string GetDebugString();
 /* Shorter form for passing filename and linenumber */
 #define FILE_LINE __FILE__, __LINE__
 
-/**
- * Used for profiling.
- *
+/** TicToc profiling.
  * Usage:
- * TIC();
- *   --Do your code--
- * TOC("A name", 1);
- *
- * When you run the TIC() / TOC() multiple times, you can increase the '1'
- *  to only display average stats every N values. Some things to know:
- *
- * for (int i = 0; i < 5; i++) {
- *   TIC();
- *     --Do your code--
- *   TOC("A name", 5);
- * }
- *
- * Is the correct usage for multiple TIC() / TOC() calls.
- *
- * TIC() / TOC() creates its own block, so make sure not the mangle
- *  it with another block.
- *
- * The output is counted in microseconds. Mainly useful for local optimisations.
- **/
-#define TIC() {\
-	auto _start_ = std::chrono::high_resolution_clock::now();\
-	static uint64_t _sum_ = 0;\
-	static uint32_t _i_ = 0;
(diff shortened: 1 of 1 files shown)


src/debug.h
```diff
@@ -63,44 +63,38 @@ std::string GetDebugString();
...
-#define TOC(str, _count_)\
-	_sum_ += (std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now() - _start_)).count();\
-	if (++_i_ == _count_) {\
-		Debug(misc, 0, "[{}] {} us [avg: {:.1f} us]", str, _sum_, _sum_/(double)_i_);\
-		_i_ = 0;\
-		_sum_ = 0;\
-	}\
-}
+ * static TicToc::State state("A name", 1);
+ * TicToc tt(state);
```
TrueBrain, line 69: Hmm, did not expect a RAII with two parts. Why the state? Why not just `TicToc tt("a name", 1)`? (honest question, ofc)

PeterN: If there's no state you can't persist the state between runs when count is not 1.

TrueBrain: Lol, doh. Fair. You could have mentioned that somewhere, but okay :P :P Compensate for my lack of understanding! :P

TrueBrain (dismissed):

PeterN:
Well, looks like I don't know how to use chrono & time points...

src/debug.h
```diff
@@ -63,44 +63,38 @@ std::string GetDebugString();
...
+		uint32_t count = 0;
+		uint64_t chrono_sum = 0;
+
+		constexpr State(std::string_view name, uint32_t max_count) : name(name), max_count(max_count) { }
+	};
+
+	State &state;
+	std::chrono::system_clock::time_point chrono_start; ///< real time count.
+
+	inline TicToc(State &state) : state(state), chrono_start(std::chrono::high_resolution_clock::now()) { }
```
rubidium42: I guess `system_clock` and `high_resolution_clock` are not the same on some platforms.
```suggestion
	std::chrono::high_resolution_clock::time_point chrono_start; ///< real time count.

	inline TicToc(State &state) : state(state), chrono_start(std::chrono::high_resolution_clock::now()) { }
```

rubidium42 (approved):

PR #11858, opened 2024-01-21
Fix #9722: create vital windows as soon as local_company is set

### LINKED ISSUE #9722: [Crash]: Crashed while creating small map
### Version of OpenTTD

12.1, Windows 10

### Steps to reproduce

1. Create game, select temperate, industries very low, towns very low, map size 64x64.

### Upload crash files

[OpenTTD.zip](https://github.com/OpenTTD/OpenTTD/files/7615698/OpenTTD.zip)

TrueBrain:
## Motivation / Problem

Fixes #9722.

There is a very small window during world generation where `local_company` is valid but the vital windows don't exist yet. However, the rest of the code assumes that if `local_company` is valid, the vital windows exist.

## Description

Resolve this by assuring that if `local_company` is valid, the vital windows also exist.

That said, there are a gazillion other ways to resolve this. For example, by having a boolean to indicate whether terrain generation is done (and no `_generate_world` is not that boolean. It means something else). Or to make the code check if vital windows exists, before assuming they are. But all those had a much higher impact on the code base, and it is hard to validate if all possible paths are covered that way. This one was just easier.

PS: this is mostly a problem for hotkeys, as they bypass all kind of other validations.

## Checklist for review

### DIFF
--- a/src/genworld.cpp
+++ b/src/genworld.cpp
@@ -72,8 +72,6 @@ static void CleanupGeneration()
 	_generating_world = false;
 
 	SetMouseCursorBusy(false);
-	/* Show all vital windows again, because we have hidden them */
-	if (_game_mode != GM_MENU) ShowVitalWindows();
 	SetModalProgress(false);
 	_gw.proc     = nullptr;
 	_gw.abortp   = nullptr;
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
Makes sense to me. Good find!

Issue #8968, opened 2021-04-07
Arrow key camera panning not consistent with framerate


jasperfirecai2:
## Version of OpenTTD
1.11.0
## Expected result
Screen scrolling with the arrow keys is smooth on all framerates
## Actual result
Screen scrolling is a fixed rate independent of framerate, causing screen tearing on higher framerate settings, and inconsistency between right mouse panning and arrow key panning.
I'm afraid I don't have a good way to show this with an image or video
## Steps to reproduce
1. Load up the game
2. Change the framerate to something higher than the default, in my case: 144Hz
3. Use the arrow keys to pan the camera
4. Compare with right mouse button panning

TrueBrain:
I think this is not so much related to the refresh-rate, even on 60Hz it feels terrible.

Enabling `smooth viewport scrolling` resolves the issue mostly, which means it is most likely caused that arrow key movements are not per pixels, but by a bunch of pixels at once. There are more places in OpenTTD that have similar issues, like news popping up.

Will need to look further into how arrow key movement works to see if this can be easily fixed, but possibly a simple fix is always applying smooth viewport scrolling for arrow key movement.

TrueBrain:
After looking into this in more detail, indeed, the best fix is to enable smooth-scrolling. I am currently applying some fixes to make smooth-scrolling a bit more smooth, but that is really the solution to this ticket. If you use arrow-keys, enable smooth-scrolling.

Initially I wanted to force enable it when using arrow-keys, but decided that would be weird; some people might not want it. So instead, I have made a suggested change to make smooth-scrolling the default.

PS: did you know that when you press SHIFT the arrow movement goes 5 times as fast? Easier to get to places :)

PR #11860, opened 2024-01-21
Change: set smooth-scrolling on by default

TrueBrain:
## Motivation / Problem

Without smooth-scrolling it is very hard to get a feel towards which direction you "teleported" to. Additionally, moving the viewport with the arrows is a terrible experience without smooth-scrolling. For example, #8968 would be resolved by making this the default.

On the other hand, with smooth-scrolling you get a much better sense if you move the viewport to an industry or train or whatever, where abouts on the map it is from where you came from.

I think it is one of those settings that is a bit hidden from view, and that most people don't know how much better your map awareness gets when you enable it.
On JGRPP survey results we can see that 40% already have it enabled. So I am not alone in thinking that smooth-scrolling is just better for .. well .. everything.

## Description

Flip the switch, and learn everyone to be more aware where on the map they are going to.

(As always, this changes the default; it will not change your game if you played before. This only changes the new user experience).

NOTE: the smooth-scrolling experience has been improved (by a lot, if you ask me) in recent master. Please check that out before you create an opinion :)

## Checklist for review

### DIFF
--- a/src/table/settings/gui_settings.ini
+++ b/src/table/settings/gui_settings.ini
@@ -123,7 +123,7 @@ cat      = SC_BASIC
 [SDTC_BOOL]
 var      = gui.smooth_scroll
 flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
-def      = false
+def      = true
 str      = STR_CONFIG_SETTING_SMOOTH_SCROLLING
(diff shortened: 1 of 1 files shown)


frosch123:
Earlier discussion for reference: https://github.com/OpenTTD/OpenTTD/discussions/8393#discussioncomment-220229

PeterN (approved):

PR #11861, opened 2024-01-21
Fix: smooth-scrolling large distances didn't got smoothly in one direction

TrueBrain:
## Motivation / Problem

Smooth-scrolling only does N tiles per tick, but this is done individually of X and Y.

In result, if you create a 64x4096 map, and scroll from one end of the other, you get something funky:

One of the axis is clamped to N, but the other doesn't, as it can reach it easily. But that means that it starts off by going in one direction, changes to another a bit later, and just before landing at the destination, it changes again. Just try it out, it is really funny to observe.

## Description

Keep the ratio between x/y the same during scrolling, so we always go in one direction.

## Checklist for review


frosch123 (approved):
Looks like the various "if" ensure that `delta_x` and `delta_y` are non-zero when dividing. Quite magical, but ok :)

PR #11864, opened 2024-01-21
Codechange: rename .hpp to .h

rubidium42:
## Motivation / Problem

Someone was wondering why there are .hpp and .h files.

## Description

The .h files seem to be in the majority, so I renamed .hpp to .h.

I excluded the 3rdparty files (json.hpp and catch.hpp). I we were going with .hpp instead, then the other 3rdparty files should be excluded (format, monocypher, etc).

## Limitations

This is might be breaking a lot of patches. The third commit is completely scripted and can therefore easily be executed when you'd run into conflicts.
We might want to wait with merging this until just before branching for release 14.

## Checklist for review


LordAro:
-1. Change for change's sake

TrueBrain:
> -1. Change for change's sake

Given we had so many patch-breakers in 14.0, I would argue there isn't a better time. C-char -> std::string, formatting, uint8 -> uint8_t, etc etc. We dumped a lot in 14.0. So if we want to make extensions make sense, this is the moment, I would argue.

LordAro:
I'm not concerned about patch breakers. I fail to see the justification for making this change. A mild bit of consistency isn't enough, personally

glx22:
Wasn't it .hpp for headers with templates and .h for normal headers ?

TrueBrain:
> Wasn't it .hpp for headers with templates and .h for normal headers ?

`If you are writing one or more template class in the dedicated header file, use file.hpp for its name instead of file.h. This will let others know that it is template library (includes also implementation), not just header with declarations.`

PeterN:
I think we should 'fix' that rule, because it's pretty impossible to follow now.

PR #11867, opened 2024-01-22
Add: list_[scenario|heightmap] and load_[scenario|height] console commands

### LINKED ISSUE #10567: [Bug]: Cannot open scenario from console
### Version of OpenTTD

13.0, Debian 11.6

### Expected result

When downloading any content from the Dedicated Server's perspective (with `content` command), I should be able to use the content in question right away, without having to manually un-tar the archive, and put it into the respective folders, or at least be able to `load` the tar-ed scenario (in this case). `docs/directory_structure.md` states that it should be possible. On my Windows OpenTTD running as a client I can use the content just fine, even tho it's tar-ed.

### Actual result

1. The `content` (scenario in this case) is downloaded
1.1. The content is showing as "Installed"
1.2. The content is visible under `~/.local/share/openttd/content_download/scenario/`, as a tar archive
2. I can't load the content, either via `./openttd -D -g ~/.local/share/openttd/content_download/scenario/xyz.tar`
2.1. I can't load the content via traversing the directory layout, with the dedicated server console (either tho it's there, it just doesn't "want" to see the .tar archive)
```
pwd
/home/openttd/.local/share/openttd/content_download/scenario/
ls
0) .. (folder nadrzędny)
1) heightmap/ (folder)
```

### Steps to reproduce

1. Start the dedicated server via `./openttd -D`
2. Update the content cache (`content update`)
3. Select a scenario (`content select 3927155`)
4. Download the scenario (`content download`)
(you can even restart the game here, won't help)
You can then check the `content_download/scenario` folder, and you'll see the tar-ed archive.
You can try to load it via `./openttd -D -g (path)`, or just via load command, it won't work.

TrueBrain:
## Motivation / Problem

Fixes #10567

Via console there was no way to load (or list) a scenario/heightmap.

## Description

Add the ability to do so. There are now four new commands:

- `load_scenario <index|file>`
- `load_heightmap <index|file>`
- `list_scenario`
- `list_heightmap`

For consistency, you can also use `load_save` and `list_saves` instead of `load` and `ls`.

They had to be different commands, as initially I tried combining them al, so that `ls` lists them all. But as paths are tracked in different variables, and things behave differently depending if you are in the scenario folder (read tar-files) or not (do not read tar-files), the interface became very vague.

To avoid needing a `cd_scenario` or something, no directory listing is allowed for scenario / heightmap. I would even argue we should do the same for savegames, and only list those savegames in a `save` folder, instead of allowing for this directory walking as we currently have it. But okay, that is not for this PR.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -56,9 +56,8 @@ static uint _script_current_depth; ///< Depth of scripts running (used to abort 
 /** File list storage for the console, for caching the last 'ls' command. */
 class ConsoleFileList : public FileList {
 public:
-	ConsoleFileList() : FileList()
+	ConsoleFileList(AbstractFileType abstract_filetype, bool show_dirs) : FileList(), abstract_filetype(abstract_filetype), show_dirs(show_dirs)
 	{
-		this->file_list_valid = false;
 	}
 
 	/** Declare the file storage cache as being invalid, also clears all stored files. */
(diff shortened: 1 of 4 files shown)


src/console_cmds.cpp
```diff
@@ -2531,10 +2617,16 @@ void IConsoleStdLibRegister()
 	IConsole::CmdRegister("scrollto",                ConScrollToTile);
 	IConsole::CmdRegister("alias",                   ConAlias);
 	IConsole::CmdRegister("load",                    ConLoad);
+	IConsole::CmdRegister("load_save",               ConLoad);
```
glx22, line 2620: This could be an alias.

src/console_cmds.cpp
```diff
@@ -2531,10 +2617,16 @@ void IConsoleStdLibRegister()
...
+	IConsole::CmdRegister("load_scenario",           ConLoadScenario);
+	IConsole::CmdRegister("load_heightmap",          ConLoadHeightmap);
 	IConsole::CmdRegister("rm",                      ConRemove);
 	IConsole::CmdRegister("save",                    ConSave);
 	IConsole::CmdRegister("saveconfig",              ConSaveConfig);
 	IConsole::CmdRegister("ls",                      ConListFiles);
+	IConsole::CmdRegister("list_saves",              ConListFiles);
```
glx22, line 2627: Same here.

glx22 (approved):

PR #11872, opened 2024-01-22
Change: Don't send duplicate neighbours to the pathfinder

SamuXarick:
## Motivation / Problem
Water region pathfinder can receive duplicated water region patches from `VisitWaterRegionPatchNeighbors` as neighbours when adding nodes. It can happen when multiple aqueducts cross into the same neighbour region.

I'm not exactly sure of the internals of Yapf, what it does in a situation like this. I see them nodes being pushed into the node lists and I thought that maybe fixing this, it could theoretically reduce the number of nodes to search. Less nodes, faster search.

## Description
Combines `VisitWaterRegionPatchNeighbors` and `VisitAdjacentWaterRegionPatchNeighbors` into one function and restructured them to prevent sending duplicate water region patch neighbour entries to the high level pathfinder.

## Limitations
I'm under the impression the changes to how it processes aqueduct checks are faster than iterating all the tiles in the region.

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -275,50 +275,32 @@ void InvalidateWaterRegion(TileIndex tile)
 }
 
 /**
- * Calls the provided callback function for all water region patches
- * accessible from one particular side of the starting patch.
- * @param water_region_patch Water patch within the water region to start searching from
- * @param side Side of the water region to look for neigboring patches of water
- * @param callback The function that will be called for each neighbor that is found
+ * Find the tile of a cross-region aqueduct based on the given parameters.
+ * @param region_x The x-coordinate of the water region.
+ * @param region_y The y-coordinate of the water region.
+ * @param side The side the aqueduct crosses.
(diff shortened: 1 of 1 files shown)


Kuhnovic:
The pathfinder adds a node for each water region patch. If a node for a particular water region patch already exists (checked by the hash), then it checks if the path via the current parent is better than the one stored already. If it is it removes the old node and stores the new one. This is a fundamental aspect of A* and essential for finding optimal paths. There's a bit more to it, this is a great article if you're interested https://csis.pace.edu/~benjamin/teaching/cs627/webfiles/Astar.pdf. But adding a few extra nodes when they already exists is not going to hurt the performance, since A* pathfinders are designed to deal with this situation often (and therefore this operation should be fast).

I'm generally against this change because:
- This doesn't solve any real problem that we currently have.
- It adds complexity.
- It turns two small somewhat-easy-to-understand functions into a monolith which is harder to understand and easier to create bugs in.
- It "improves" a very niche use case.
- The performance gain will be negligible, if there is one at all.

PR #11874, opened 2024-01-23
Change: rebrand Cheats as Sandbox Options

TrueBrain:
## Motivation / Problem

People have an emotion when they read "cheats", and are like: NOOOO, I DO NOT WANT TO USE CHEATS, but they do want to functionality it offers. Like "infinite money". Or "magic bulldozer".

I read this one too many times now to know it actually holds people back from using them, as "it is a cheat". But if you want to get more money to play the way you want to play, go nuts. Don't feel constrained because we call it "cheats".

This includes the warning at the bottom, and the "this has been used" marker on the left. Like we are punishing people for playing the way they want.

## Description

Remove the sentiment of "cheats" by rebranding it as "sandbox options".

Also remove the checkbox of "it is used" and the text that comes with it.

We do still store it in the savegame, but that has more to do for analysing bug reports; a savegame where the year has been changed has to be handled a bit differently.

image

As bonus, and on popular demand, also added it to the menu (only in-game; not in SE)

image

## Limitations

- When you use one of these options in multiplayer, which you can't without modifying the client, you are still kicked for being a cheater. I am fine with that.
- I did not rename the internal references to "cheat", as .. we don't actually care how it is called :)
- You can have huge debates if certain options shouldn't be under settings. And some most likely do. Where others clearly don't. So the window will stay in all those scenarios, and I leave the debate about what should go where for another day. Not this PR at least.

## Checklist for review

### DIFF
--- a/src/cheat_gui.cpp
+++ b/src/cheat_gui.cpp
@@ -217,10 +217,7 @@ static constexpr NWidgetPart _nested_cheat_widgets[] = {
 		NWidget(WWT_SHADEBOX, COLOUR_GREY),
 		NWidget(WWT_STICKYBOX, COLOUR_GREY),
 	EndContainer(),
-	NWidget(WWT_PANEL, COLOUR_GREY, WID_C_PANEL), SetDataTip(0x0, STR_CHEATS_TOOLTIP), EndContainer(),
-	NWidget(WWT_PANEL, COLOUR_GREY),
-		NWidget(WWT_LABEL, COLOUR_GREY, WID_C_NOTE), SetFill(1, 1), SetDataTip(STR_CHEATS_NOTE, STR_NULL), SetPadding(WidgetDimensions::unscaled.frametext),
-	EndContainer(),
+	NWidget(WWT_PANEL, COLOUR_GREY, WID_C_PANEL), EndContainer(),
 };
 
 /** GUI for the cheats. */
@@ -228,7 +225,6 @@ struct CheatWindow : Window {
 	int clicked;
 	int clicked_widget;
 	uint line_height;
-	Dimension box;      ///< Dimension of box sprite
 	Dimension icon;     ///< Dimension of company icon sprite
 
 	CheatWindow(WindowDesc *desc) : Window(desc)
@@ -238,7 +234,6 @@ struct CheatWindow : Window {
 
 	void OnInit() override
 	{
-		this->box = maxdim(GetSpriteSize(SPR_BOX_EMPTY), GetSpriteSize(SPR_BOX_CHECKED));
 		this->icon = GetSpriteSize(SPR_COMPANY_ICON);
 	}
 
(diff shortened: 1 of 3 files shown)


TrueBrain:
> While this is a step in a direction, I think more could to be done to allow sandbox functions to be used in multiplayer. It is a separate concern but when they're called 'cheat' it's clear why they are not allowed in multiplayer, as 'sandbox' not so much.

I agree. Not for this PR btw, but I do agree.

TrueBrain:
Of course those checkboxes are a rather weak signal that someone "cheated", as there are plenty of "cheats" in the settings. Station-walking etc etc can earn you a lot of money, and you can undo that just before handing over your save.

That said, as I wrote in the PR description, we still track if cheats are used yes/no, so with a savegame analyzer like https://truebrain.github.io/OpenTTD-savegame-reader/ they can easily discover if someone did. That part was not modified. So I guess nothing changed there, and they can still have their challenges in the same way :)

You agree?

2TallTyler:
I would rather move these features and eliminate the cheat menu altogether. Not that I am suggesting you do this, I am willing to make a PR (or maybe several) for it myself. 😉 

Most of these could fit in Settings, with a few possible exceptions:
* Increase/decrease money sends commands instead of changing a setting. Maybe it could go in the finance window somehow?
* Company selection could go in the AI configuration window, perhaps?
* Change date is also not a setting. I'm not sure where to put this. Maybe another hidden feature by Ctrl+Clicking on the date in the status bar, because hidden features aren't a terrible idea or anything. 😛

TrueBrain:
> I would rather move these features and eliminate the cheat menu altogether.

I don't know how to read your reply, whether you are opposed to this PR or not :D For simplicity, I am going to assume your reply is to not do this PR. On that:

All good ideas (in fact, I wrote about it in my description!); but for now, I think this PR is at least a good (and small) step forward. It doesn't hurt anyone, and those people that feel hold back by the naming of "cheat" are helped.

My main concern with your idea: it is not new. It has been talked about for months / years now, yet nobody actually does it. And I am sure when someone does, it will take weeks / months of discussion, whether that one place is the right place, and bladiebladiebla. You know how it goes by now :) So I do appreciate your enthusiasm, here, and please keep it going. And for now .. why not do this PR, so at least the biggest problem is resolved? :)

The best thing about this PR, especially when approved/merged, it shows how silly some of these options are to put them in this window; and not just in settings. It just .. highlights that fact a bit more. Which is one of the discussion that has been held over and over: "should Magic Bulldozer not be in the settings?!" :D

Anyway, let's please not not make a small step because we see a better big step; if it is forwards, it is at least in the right direction (if you ask me).

2TallTyler (approved):
It's hard to distinguish incremental progress from repeatedly changing things and confusing users, but you are right and I'm fine with this as it stands.

DefinitelyNotRau117:
> That said, as I wrote in the PR description, we still track if cheats are used yes/no, so with a savegame analyzer like https://truebrain.github.io/OpenTTD-savegame-reader/ they can easily discover if someone did. That part was not modified. So I guess nothing changed there, and they can still have their challenges in the same way :)
> 
> You agree?

Not really. To use savegame reader you need to at least know that it exists. And to know how to use it, the interface there is far from the user-friendly. And also have access to the site, since it does not come with the game. It's... _a little_ more complicated than just checking for checkmarks.

Maybe just make a console command? This will be much easier than loading a save to reader and scrolling through a ton of strange nodes.

PeterN:
This change brings me a dilemma now.

If I want to add an option which is considered a sandbox feature, should it go into the regular game settings, or the sandbox options window?

I don't want to add anything new into the 'cheats' system, but equally having a sandbox options window which doesn't contain all the sandbox settings is not good for user experience.

PR #11878, opened 2024-01-24
Add: public key authentication to the admin connection

rubidium42:
## Motivation / Problem

Sending passwords, especially plain text ones, over unencrypted lines is bad. Why not allow configuring public key authentication for the admin connection?

After all, we already have this for the game connection and securing companies.

## Description

Use the new crypto helpers from #12320 to implement basically the same authentication principles as implemented for the game socket in #12323. Furthermore apply the packet encryption like done in #12300.

Unlike with the game socket, the admin socket requires some sort of backward compatibility and to achieve that a new join packet is introduced. This join packet kicks off the new/secure authentication and finally enablement of an encrypted network connection.

The old join packet is still available and works, however there is a setting that disables this by default as using passwords with the secure authentication is much safer as the password isn't actually sent over the wire anymore.

I have cobbled together some changes to libottdadmin2 to test this: Xaroth/libottdadmin2#17. This includes a key generator to generate your own private and public key.

## Limitations

None I can think of.

## Checklist for review


glx22:
Maybe a new ADMIN_JOIN packet would be a better option.

PeterN:
Is this based on some standard for pub-key authentication? What about trust between client and server? Does TLS fit in anywhere?

TrueBrain:
Awesome work, really cool :D

The way you went, is absolutely a valid one. Install a public key on the server, keep the private key on the client, and clients can proof their identity. Basically, client-certificates (if I understand your implementation right). Might also be worth branding it like that, as that is how people are used to call it. Makes it a bit more clear that it is a method to proof the identity of the client (instead of a password).

There is another approach to this, and that is what JGRPP is using [source](https://github.com/JGRennison/OpenTTD-patches/blob/jgrpp/src/network/network_server.cpp#L239), where no keys are stored on the server, but everything is generated on-the-fly to setup a secure channel and exchange an old-fashioned password. [Authenticated Encrpyption](https://monocypher.org/manual/aead) and [X25519 Key Exchange](https://monocypher.org/manual/x25519) are involved there. But, this would still require a password; it just isn't send plaintext anymore.

I am not saying one is better than the other btw. Just mentioning we have alternatives here, if we like.

What is nice about your solution, is that it could be the beginning of another trust-model in OpenTTD, where we use client-certificate to identify people joining games. So that is how it would fit in to multiplayer.

Important to mention, this has little to do with encrypting the connections; this is just a way to proof identity.

rubidium42:
> Maybe a new ADMIN_JOIN packet would be a better option.

Could be done; a new packet type is cheap for now. I did it this way so the client could be somewhat agnostic about whether you pass a password or public key. It would then only need to look for its private key when the server asks for the challenge.


> Is this based on some standard for pub-key authentication? What about trust between client and server? Does TLS fit in anywhere?

With TLS you could practically achieve a form of public key authentication, but then the server needs to validate the certificate of the client against a list of authorized certificates.

What is implemented here is practically the public key authentication of SSH. When you connect you send your public key, and then the SSH server sends a challenge to the client, which the client has to sign using the private key associated with the public key that was just sent. The SSH server then validates the result of the challenge.
I do not know at which stage it actually checks whether the public key is in the SSH server's configuration.

What this system does not do but what SSH does do, is return the identity of the server. I do not know whether that's really necessary at this stage, since there is nothing proving the identity now either. Though the protocol could be extended so the client also sends a challenge to the server, which it can answer together with its public key. Then the client could also check whether the server's public key is known.
Though, would that need to be done "authorized_keys" style (the client manually add the public key of the server to the client) or "known_hosts" style (if the client doesn't know the server, it gets added to a list of known hosts otherwise it checks against earlier states). However, known hosts also does things with linking to IP addresses making it quite a hassle when there server's IP address is not stable.


The biggest problem I have now, is figuring out how this might be used for connecting to the server and joining companies. Primarily as you have to consider practically security cryptography illiterate people, who you cannot ask to generate a key pair with some "standard" tool.
With the admin connection it's slightly easier, as the problem of getting the public key in the server is not part of OpenTTD. That's why I started with this, so I could get some insights in how things could work... and hopefully get some more brains working on the problem.

TrueBrain:
> The biggest problem I have now, is figuring out how this might be used for connecting to the server and joining companies. Primarily as you have to consider practically security cryptography illiterate people, who you cannot ask to generate a key pair with some "standard" tool.

This is kinda easier than you think; we already generate a network-id on first start. You can also generate a private-key on start, and store that in the private-cfg (on the client, ofc). That way, the client never needs to know this was done for them.

Ofc someone can wipe the configuration and get a new ID. But then the server won't have it whitelisted anymore.

TrueBrain:
> With TLS you could practically achieve a form of public key authentication, but then the server needs to validate the certificate of the client against a list of authorized certificates.

What you describe here is [X25519 Key Exchange](https://monocypher.org/manual/x25519). This is an exchange of public keys between server and client, of which a shared secret is derived that you can only get to if you have the private key of one of the two public keys. With a challenge to both sides you proof you both say who you are.

If you would do this, instead of the current Public Key Signature (which is one-way), you could after that apply [Authenticated Encryption](https://monocypher.org/manual/aead), which supports streaming encryption.

That would mean we end up with mTLS (Mutual TLS, or Server + Client Certification Authentication + Encryption).

Doing X25519 now might be worth it; the encryption we can do after :)

src/network/network_admin.cpp
```diff
@@ -647,10 +678,48 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::Receive_ADMIN_JOIN(Packet *p)
...
+		std::array<uint8_t, NETWORK_PUBLIC_KEY_SIGNATURE_LENGTH> signature;
+		for (uint8_t &s : signature) s = p->Recv_uint8();
+
+		valid_response = AdminCheckSignatureV1(signature, this->public_key, this->challenge);
+	} else {
+		/* In the ADMIN_JOIN the version has already been checked. */
+		NOT_REACHED();
+	}
+
+	const auto &allowed_public_keys = _settings_client.network.admin_public_keys;
```
TrueBrain: I really like you only validate the public key AFTER the client has verified it has access to the private key. That is a rather clever approach :)

JGRennison:
> There is another approach to this, and that is what JGRPP is using [[source](https://github.com/JGRennison/OpenTTD-patches/blob/jgrpp/src/network/network_server.cpp#L239)], where no keys are stored on the server or client, but everything is generated on-the-fly to setup a secure channel and exchange an old-fashioned password. [Authenticated Encrpyption](https://monocypher.org/manual/aead) and [X25519 Key Exchange](https://monocypher.org/manual/x25519) are involved there. But, this would still require a password; it just isn't sent plaintext anymore.

What I'm doing is a slightly different from that, in that the password is not sent in the encrypted channel set up from the keys from the X25519 key exchange, instead the password is part of the key derivation (i.e. a PAKE scheme), and is not itself part of the payload. That way the server can't learn the password that the client sent if it didn't know it already, and a MITM attacker can't trivially do X25519 key exchanges with the server and client and read the password as it goes past.

TrueBrain:
To put Discord here too, what might be nice is to have a total of three ways to login to the admin port (and I know, we might be over asking for this PR; but we can help out, etc etc).

- Old unsafe method of sending the password plaintext; I suggest we disable this ASAP.
- X25519 key exchange (to prepare for encryption), where the client pub-key is used for access. Much like this PR currently is, just with X25519 instead of eddsa.
- X25519 key exchange (to prepare for encryption), where the password is part of the shared key derivation.

This way, the client has (ignore the first option) two ways of logging in:

- Either via their whitelisted pub-key
- Or via the password

In all cases, the password is never sent to the server. This most likely does require two packets (or a packet with a boolean indicating which form you are using). As you need to know if the password is part of the key derivation or not :)

This also helps in multiplayer, as now a client can join because they are whitelisted, without password. Or with a password.

TrueBrain:
You don't like spinning? :D To think along with you a bit:

Indeed, for the game-protocol it is a lot easier. And I agree, for the admin-protocol it needs a tiny bit more thinking. I believe it is also the only protocol we never added a "version" indicator, where we did with the rest. So that is a bit problematic :) Maybe we should fix that while we are at it?

Looking at TLS, you see that the first packet from client -> server is telling which ciphers etc it supports. So that approach is very common.

As for retaining the private/public key on the server .. there are two sides to that coin. If we cycle them every connection / startup, nobody ever has to worry about keys leaking. But you also can't validate the pubkey being the server you expect. As we are not using pub-keys that can be verified via a chain, that is a bit annoying. Personally, I am not all that worried a private key leaks; as in, the only thing it means is that someone can impersonate an OpenTTD server. Doesn't sound like a huge impact. And the main goal is to verify things like passwords without exchanging them. Or to encrypt a channel. So the impact is rather low.
On the other hand, I truly doubt if there ever will be an admin client that actually validates the public key of a server. I doubt anyone is going to do that. So we can just make it ourselves a bit easier, and use a new key every connection / startup. Much like JGRPP does for password exchanges.

For password validation, you only need the server to send a challenge to the client. The server doesn't need to proof that it has the private key, as the client is not going to do anything with it. And that challenge can be in the initial message with Authenticated Encryption. (shared secret to the rescue!)
If we do want to prepare for encryption (might be wise), if the server doesn't have the private key, it will deduce another shared key, and as such, the first packet will not pass verification. So you can use that too to implicitly validate both sides know the private key. I am not sure what the common behaviour is here; bit hard to find, honestly. But even in the TLS handshake I cannot find a mention of a challenge. Will investigate a bit further :)

src/network/network_admin.cpp
```diff
@@ -791,6 +796,61 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::Receive_ADMIN_EXTERNAL_CHAT(P
...
+		return this->SendError(NETWORK_ERROR_ILLEGAL_PACKET);
+	}
+
+	/* Mask out any protocols that are too new for us. */
+	protocol_mask &= (1 << ADMIN_AUTH_METHOD_END) - 1;
+	if (_settings_client.network.admin_public_keys.empty()) ClrBit(protocol_mask, ADMIN_AUTH_METHOD_X25519_AUTHORIZED_KEY);
+
+	if (protocol_mask == 0) {
+		/* There is no way to authorize this admin, with the protocols the admin and we both support. */
+		return this->SendError(NETWORK_ERROR_NOT_AUTHORIZED);
```
TrueBrain: It wouldn't hurt to return here to the client the ciphers didn't align. That is valuable information during debugging. `NETWORK_ERROR_NO_CIPHER` or something.

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,112 @@
...
+	bool valid_unlock = crypto_aead_unlock(message.data(), mac.data(), shared_keys.ClientToServer().data(),
+			nonce.data(), their_public_key.data(), their_public_key.size(), message.data(), message.size()) == 0;
+	bool valid_message = crypto_verify32(message.data(), this->my_public_key.data());
+
+	const auto &allowed_public_keys = _settings_client.network.admin_public_keys;
+	std::string their_public_key_string = FormatArrayAsHex(their_public_key);
+	bool allowed_public_key = std::find(allowed_public_keys.begin(), allowed_public_keys.end(), their_public_key_string) != allowed_public_keys.end();
+
+	/* Check everything at the end, as to attempt to not leak much information about what went wrong via timing. */
+	return valid_exchange && valid_unlock && valid_message && allowed_public_key;
```
TrueBrain: I am not sure about this. I get where you are coming from, but it is rather unusual. And I wouldn't worry about timing attacks on this level as much.

The issue you get now, is that `shared_keys` might be zeros, which you feed to `crypto_aead_unlock`, which will have different behaviour. So your timing is influenced anyway.

I am more tempted to follow monocypher's example, and just return when things return an error. It just makes things less complicated.

src/network/network_admin.cpp
```diff
@@ -627,14 +636,12 @@ NetworkRecvStatus ServerNetworkAdminSocketHandler::SendCmdLogging(ClientID clien
...
-NetworkRecvStatus ServerNetworkAdminSocketHandler::Receive_ADMIN_JOIN(Packet *p)
+NetworkRecvStatus ServerNetworkAdminSocketHandler::Receive_ADMIN_JOIN_UNSAFE(Packet *p)
 {
 	if (this->status != ADMIN_STATUS_INACTIVE) return this->SendError(NETWORK_ERROR_NOT_EXPECTED);
 
 	std::string password = p->Recv_string(NETWORK_PASSWORD_LENGTH);
-
-	if (_settings_client.network.admin_password.empty() ||
-			_settings_client.network.admin_password.compare(password) != 0) {
+	if (_settings_client.network.admin_password.empty()) {
```
TrueBrain: Was it intentional to remove the actual password check?

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+X25519Key::~X25519Key()
+{
+	crypto_wipe(*this);
+}
+
+static bool X25519KeyExchange(SharedKeys<32> &shared_keys, const X25519Key &my_private_key, const X25519Key &their_public_key)
+{
+	X25519Key shared_secret;
+	crypto_x25519(shared_secret.data(), my_private_key.data(), their_public_key.data());
```
TrueBrain: Btw, for your own sanity, I would use a bit of different naming here:

Monocypher calls it "secret key" and "public key". I would adapt that. The reason is, otherwise they are both `pk`, which often confuses people. But using `sk` vs `pk`, you resolve that issue.

And not doing a code-review, just functional looking into what you are up to, but `my_` and `their_` .. we need to find a better name :D The static `my_client` we have is also weird :P But okay, nitpicks is for another day.

TrueBrain (commented):
Not sure what your plans are, or if you have any (which is fine both ways), but some feedback when reading through your code. Functional, I left all the code review comments out :)

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+	p.TransferOut(&TransferOutSpan, their_public_key);
+	p.TransferOut(&TransferOutSpan, message);
+	p.TransferOut(&TransferOutSpan, mac);
+
+	SharedKeys<32> shared_keys;
+	bool valid_exchange = X25519KeyExchange(shared_keys, this->my_private_key, their_public_key);
+	bool valid_unlock = crypto_aead_unlock(message.data(), mac.data(), shared_keys.ClientToServer().data(),
+			nonce.data(), their_public_key.data(), their_public_key.size(), message.data(), message.size()) == 0;
+	bool valid_message = crypto_verify32(message.data(), this->my_public_key.data());
```
TrueBrain: The message is already valid; `crypto_aead_unlock` promises that. So this is kinda redundant. There isn't really a payload to put in here that is needed.

Of course we can't send nothing, so something has to go in. And for that, I can see something like this. But it really isn't needed from a security point of view. Just making sure you know :D

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+X25519AuthorizedKeyHandler::~X25519AuthorizedKeyHandler()
+{
+	crypto_wipe(this->nonce);
+}
+
+/* virtual */ void X25519AuthorizedKeyHandler::SendRequest(Packet &p)
+{
+	p.Send_bytes(this->my_public_key);
+	p.Send_bytes(this->nonce);
```
TrueBrain: From a lot of examples I have seen, normally the nonce is send with the mac. I don't think it hurts to do it this way; but it is a bit unusual.

The main reason I can imagine this being better, is that we now control the nonce; otherwise the client might be reusing theirs, which is an absolute no-no.
So I can imagine you did it for that reason. Just checking in, basically.

Crypto is hard. My head hurts. :P

src/network/network_crypto.cpp
rubidium42: The message is valid, but is it what we expect the message to be?
At the client, I just inject some random bytes to encrypt and leave this check out. However, what are they going to put in there. If it works with all zero, then why not? Is that a situation we would want? I don't know.

JGRPP has an increasing counter which may not go backward in the payload, just to check it's not doing something fishy and the content is not always the same. Maybe to prevent replay attacks or something?

Also bear in mind that the primary reason I pushed was that I was at a road block for not being able to find the right "encrypt" function in the Python library. So I pushed what I had to ask around, but apparently doing something different for a while made me come up with a (nasty) solution in Python land (using two crypto libraries).

src/network/network_crypto.cpp
rubidium42: Yes, I sent the nonce here to ensure it to have at least some random-ness, instead of hoping the authors of the ottd-admin libraries are going to do the right thing.

rubidium42:
It now seems to work with the x25519 key exchange trick for the authorized keys. I haven't really looked at the diff myself yet, so this is the draftiest of drafts. Please don't read too much into the existing code already.

src/network/network_crypto.cpp
JGRennison: Replay attacks was the main motivation for that, yes.

TrueBrain: We could make the step to encryption, as that would kinda make sense at this point. Then the payload can be the next packet the client sends to the server.

Just a random thought that came to mind.

src/network/network_crypto.cpp
rubidium42: Well, sanity... this function has:
- my private key (or monocypher: your secret key)
- my public key (monocypher: your public key)
- their public key
- our shared secret
- our (two) shared keys

In Python [cryptography land](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519/) they are called, respectively:
- private key
- (the public key is not a separate variable from the private key)
- peer public key
- shared key
- derived key


So, whatever we choose, either `pk` or `sk` abbreviations would have duplicate meanings.
Note that "your" makes sense in monocypher's API, but not that much in OpenTTD's realm. If we were talking to each other, and I would pass "my key" to you, you would refer to that key as "your key".
Similarly, "our" could have worked instead of "my", except that due to the key exchange we are joining "my" and "their" to derive a shared key between us, which would be "our" key.

I want this code to be general enough that it's trivial to be used in both the client and the server. The only difference between the client and server would be the order of passing "my" and "their" public key to blake2b the hash.

src/network/network_admin.cpp
rubidium42: No, small oversight when removing code.

src/network/network_admin.h
```diff
@@ -24,7 +24,7 @@ extern NetworkAdminSocketPool _networkadminsocket_pool;
 /** Class for handling the server side of the game connection. */
 class ServerNetworkAdminSocketHandler : public NetworkAdminSocketPool::PoolItem<&_networkadminsocket_pool>, public NetworkAdminSocketHandler, public TCPListenHandler<ServerNetworkAdminSocketHandler, ADMIN_PACKET_SERVER_FULL, ADMIN_PACKET_SERVER_BANNED> {
 protected:
-	NetworkRecvStatus Receive_ADMIN_JOIN(Packet *p) override;
+	NetworkRecvStatus Receive_ADMIN_JOIN_UNSAFE(Packet *p) override;
```
rubidium42: Also a contentious point I have not fully resolved. For the APIs it would be better if the packet names remain the same; you can still use the "old" version. In the messing with libottdadmin2 I kept this name the same, and named the new one 'Safe' but that might not be the best name either.

Does anyone have a good suggestion with naming the two joins?

src/network/network_admin.cpp
rubidium42: My thought was that there's no authorized method that can be used. The nasty thing is that we are reusing the "game" error codes enumeration here, and it would generally be an unused error with a translation that no-one is ever going to see in-game unless they look at the console with debug net>=1.

Sure, I can add an error, but they're not as cheap as save game versions.

src/network/network_admin.h
michicc: Random idea: `ADMIN_JOIN_SECURE`?

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,110 @@
...
+	p.Recv_bytes(message);
+	p.Recv_bytes(mac);
+
+	SharedKeys<32> shared_keys;
+	if (!X25519KeyExchange(shared_keys, their_public_key, this->my_private_key, this->my_public_key, shared_key_extra_payload)) return {};
+
+	if (crypto_aead_unlock(message.data(), mac.data(), shared_keys.ClientToServer().data(),
+			nonce.data(), their_public_key.data(), their_public_key.size(), message.data(), message.size()) != 0) return {};
+
+	if (crypto_verify32(message.data(), shared_keys.ServerToClient().data()) != 0) return {};
```
TrueBrain: Does this means the client is sending his shared key over the wire?

The idea of those shared keys is that they are never (ever) exchanged over the wire, encrypted or not. Similar as a private key. They need to remain local.
So even if we know this packet is just to have something in there, the public key you had before is better than this :)

Maybe for now just send a single byte, a `1` or something, to not give the impression it does anything else?

As I mentioned in another comment, I would suspect we use this as a start to encrypt the traffic (as we might as well, we are all the way there after all). So it might change later anyway.

rubidium42: Yeah, I was already rethinking what to do instead. JGRPP has some incrementing counter, which could work although for now there would be very little to test content wise. Ah well, I'll keep thinking ahead for a while to find a nice solution.

TrueBrain: Don't overthink this too much. The payload is not relevant for what you try to achieve here :)

A counter only works for a connection that remains open, as how JGRPP uses it. So you can't try a company password twice with the same payload. But as this comes from a new connection, that doesn't really solve anything.

rubidium42: It's definitely true that the payload is not relevant here.

However, it would be nice if the decrypt that we are intending to do at some later point, were to be able to use the exact same code in OpenTTD (and as such the libraries). And since it's not the game connection, but the admin connection, it's (much) harder to make changes at a later stage without breaking things.

TrueBrain: It is also why I would suggest we do this in 15.0, to give ourselves a bit of time figuring out what to do here exactly.

But yeah, creating something that survives the future is the tricky part. Maybe give a run out multiplayer authentication and encryption, and see how that fits in this work?

rubidium42: I'm toying with the idea of a quite surgical approach to introducing encryption of the data. I'm not sure whether it's a good idea yet, but why not encrypt the payload of "all" the packets that we send over the line?
Or at least those packets when we have a shared secret. This would put the nonce and mac (signature) between the size of the packet and the type byte of the packet. That will increase each packet by 40 bytes though.

I was wondering whether sending the packet size unencrypted was useful, but I doubt it makes any sense not to send it unencrypted because the decrypt needs to know the length. And that we can derive from the packet size.

The major question is whether adding 40 bytes to the packet is a big problem. Also, do we want packet sequence numbers to prevent replay attacks? That would add another 4 or 8 bytes.
Although we could be sneaky, as there is no real reason to make a whole new random nonce each time. It is, as far as I can see, fine to just increment the value by one each iteration. The monocypher documentation even mentions that for shorter nonces random is not safe enough, and you should use a counter. So, in theory we need to agree on a nonce once (say 16 bytes), and just append a 8 byte counter to that, that is incremented with each packet. That would implicitly prevent replay attacks, as the nonce would be wrong. In other words, with just 16 bytes per packet, less than the size bump for going from IPv4 to IPv6, we could encrypt every packet after the key exchange.

The only question is then, at what point do we want to perform the jump to encryption.  For game I'd say after the NewGRF and client version have been checked, but before the name is sent. That way old clients attempting to join just get the "right" you've got the wrong version error, instead of some invalid/illegal packet message. Since we already know the version is "right", the client can also put their public key in the join packet. If the server then sends its public key and starter nonce back, both sides have enough information to encrypt the second packet that either client and server sends.

For admin we got the method handshake, so the client sends the methods it supports. The server chooses and and sends the nonce and public key, and the the client replies with its public key and everything can be encrypted from then on.

There seems to be a limit of [256 GiB](https://pycryptodome.readthedocs.io/en/latest/src/cipher/chacha20_poly1305.html) for the maximum amount of plaintext before a new key must be agreed upon. That is a heck of a lot, but with maximum packet size that's "only" 8 million packets (2**23). On the other hand... that's an insane amount, so just "oops"-killing the connection after 256 GiB of packets per socket might not even be the worst solution.

TrueBrain: 40 bytes per packet sounds fine by me. About the same overhead as the IP header itself.

Otherwise, sounds like a plan to at least try out! We won't get this right first drafts, and there is nothing obvious weird in your story, so yeah.

About the nonce, replay attacks aren't actually possible. Monocypher cycles the key every packet send, and the nonce is random at start. So this is already sequenced. At least, as far as I can tell, there isn't a packet you can replay and would be accepted at any stage, assuming the server generates the initial nonce. That is the lovely thing about these methods, they are actually secure with little effort :)

src/network/network_crypto.cpp
rubidium42: Given the server currently is always generating a new keypair, it's not feasible to do a replay attack. So I changed it to contain, for this authentication just some random data.

src/network/network_crypto.cpp
rubidium42: I think I'm going with:
- our secret key
- our public key
- peer public key
- shared secret
- derived keys

It's a bit of a mix-match from all, but probably the most different names that I could come up with.

rubidium42:
Had some more "fun" to implement automatic fallbacks, so if the authorized key fails it will try the password when it's available.
```
[2024-02-01 18:18:44] dbg: [net:3] [admin] 'libottdadmin2' (0.0.3a1) has connected
[2024-02-01 18:18:44] dbg: [net:3] [admin] 'libottdadmin2' (0.0.3a1) authenticating using X25519-AuthorizedKey-server
[2024-02-01 18:18:44] dbg: [net:3] [admin] 'libottdadmin2' (0.0.3a1) authentication failed, trying next method
[2024-02-01 18:18:44] dbg: [net:3] [admin] 'libottdadmin2' (0.0.3a1) authenticating using X25519-PAKE-server
[2024-02-01 18:18:44] dbg: [net:3] [admin] 'libottdadmin2' (0.0.3a1) authenticated
```
Next: trying to do logging in to the game using authorized keys and PAKE.

src/settings_type.h
```diff
@@ -316,7 +316,9 @@ struct NetworkSettings {
 	std::vector<std::string> server_authorized_keys; ///< Public keys of clients that are authorized to connect to the game.
 	std::string rcon_password;                            ///< password for rconsole (server side)
 	std::vector<std::string> rcon_authorized_keys; ///< Public keys of clients that are authorized to use the rconsole (server side).
+	bool allow_unsecure_admin_login; ///< Whether to allow logging in as admin using the unsecured old JOIN packet.
```
TrueBrain: The famous unsecured vs insecured :D

The problem with unsecured, `unsecure` is not a word in the English language. Only `unsecured` is. "Your bike is unsecured". Something like that. Which makes it a really annoying word to deal with.

In general, it is common to call a setting like this `allow_insecure_admin_login`, as you know it is insecure, as the password is unsecured :D Still with me? :P

```suggestion
	bool allow_insecure_admin_login; ///< Whether to allow logging in as admin using the unsecured old JOIN packet.
```

https://grammarhow.com/unsecure-vs-insecure/ has a nice write-up about it.

rubidium42:
Still need to fix/clean up the libottdadmin2 changes to make sure everything still works as I expect/hope, so this should not be merged yet.

Most of the underlying code has reached master, so this PR is now only making the required changes to the admin code and some minor sprinklings around adding the setting and some errors.

rubidium42:
I have also implemented encrypting the admin network connection, like is done for the game protocol. However, only for the secure join; with the insecure one everything will remain unencrypted.

I have hacked the encryption functionality into libottdadmin2, but that needs a lot of clean up work as well as pymonocypher's upstream merging some PRs to implement monocypher 4.0.2 and the incremental authenticated encryption interface. So there is a bit of time remaining to attempt to clean up the mess I made of libottdadmin2.

rubidium42:
I've cleaned up the libottdadmin2 code, and made a pull request for that (https://github.com/Xaroth/libottdadmin2/pull/17).

This PR has been tested with the changed libottdadmin2 code. Only caveat is that it adds a dependency on pymonocypher which is not up-to-date yet, but a PR has been made for that as well to introduce everything needed for libottdadmin2. We could wait on that before merging, but could also merge it sooner so other admin libraries have more time to amend their code, for those that are still active.

rubidium42:
The dependency that needed to be updated for libottdadmin2 to function has been released. So as far as I am aware there's nothing limiting this to be merged.

TrueBrain (approved):

PR #11884, opened 2024-01-26
Fix ea8c1d85: improve smoothness of smooth-scrolling for 60+Hz screens

### LINKED ISSUE #11883: [Bug(ish)]: Poor edge scrolling on high refresh rate screens
### Version of OpenTTD

20240125 nightly

### Expected result

Smooth edge scrolling and arrow key movements.

### Actual result

Since it is difficult to describe the situation in plain text, here is a video explaining the problem.
Try playing it at 0.5x speed.

[2024-01-25 22-02-26.zip](https://github.com/OpenTTD/OpenTTD/files/14060827/2024-01-25.22-02-26.zip) (20240125)

[2024-01-25 22-11-23.zip](https://github.com/OpenTTD/OpenTTD/files/14060916/2024-01-25.22-11-23.zip) (20240123)

Basically, when using a display with a high refresh rate (in my example, 240hz, but the video is at 120hz), the quality of edge scrolling and arrow key movements is quite bad. I am not sure if this is how the game is supposed to work.

I've also tested on earlier versions (20240123) and it worked fine.

### Steps to reproduce

1. Start a game
2. Use edge scrolling or arrow keys to move the screen

TrueBrain:
## Motivation / Problem

In ea8c1d85 I introduced a step-size of 30ms, which means that on a 30Hz monitor it looks fine. On a 60Hz it is hard to notice, but already less fine. On a 144Hz it is just wrong.

Fixes #11883.

## Description

Use some floats to make it a continuous function, instead of a discrete one.

Use of float is safe here, as it is client-side only.

## Limitations

Around the border, and mostly the NE and NW border when moving E or W respectively, there is still an issue that sometimes you can get stuck. That is to say: the smoothing says: move (3, 1) pixels, and the game moves (0, 0) after clamping.
To overcome this problem, which only happens when closeby the destination and only when clicking outside the border: teleport to the destination. This is observable, but hardly.
It is not ideal, but bouncing between a few pixels is neither. The highest your refresh-rate, the more apparent this issue becomes (as you are more likely to make small steps).

## Checklist for review

### DIFF
--- a/src/viewport.cpp
+++ b/src/viewport.cpp
@@ -1884,24 +1884,13 @@ static void ClampSmoothScroll(uint32_t delta_ms, int64_t delta_hi, int64_t delta
 {
 	/** A tile is 64 pixels in width at 1x zoom; viewport coordinates are in 4x zoom. */
 	constexpr int PIXELS_PER_TILE = TILE_PIXELS * 2 * ZOOM_LVL_BASE;
-	constexpr int MS_PER_STEP = 30; ///< Time between each step in the smooth scroll.
-
-	static uint32_t remainder_time = 0;
 
 	assert(delta_hi != 0);
 
-	int64_t delta_left = delta_hi;
-	int max_scroll = 0;
-
-	for (uint count = 0; count < (delta_ms + remainder_time) / MS_PER_STEP; count++) {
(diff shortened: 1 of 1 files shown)


PeterN:
There's no reason floats can't be used here, it does not affect game state.

PeterN (approved):

PR #11886, opened 2024-01-26
Add: show in multiplayer the amount of hours a game has been unpaused

TrueBrain:
## Motivation / Problem

@2TallTyler made it so the calendar can freeze in time. Problematic. As we show that date on our website and in-game when you look at a server.

So time to solve that!

## Description

Send to all clients the amount of ticks a game has been played. From this, real-time can be deduced. And this is shown, in hours, in the in-game multiplayer list (and soon on the website; but that is not for this PR to fix :P ).

Also, renamed `game_date` to `calendar_date`, to avoid confusion later on.

image
image

## Limitations

- "recently" we changed how long a tick takes, so calculated the real-time can be off for older servers.
- This only counts the time a server was not paused. Which can be a good or bad thing, depending whether "build-when-paused" is enabled or not ;)

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -2362,8 +2362,9 @@ STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION                        :{BLACK}Map size
 STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION_TOOLTIP                :{BLACK}Map size of the game{}Click to sort by area
 STR_NETWORK_SERVER_LIST_DATE_CAPTION                            :{BLACK}Date
 STR_NETWORK_SERVER_LIST_DATE_CAPTION_TOOLTIP                    :{BLACK}Current date
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION                           :{BLACK}Years
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION_TOOLTIP                   :{BLACK}Number of years{}the game is running
+STR_NETWORK_SERVER_LIST_PLAY_TIME_SHORT                         :{BLACK}{NUM}h {NUM}m
+STR_NETWORK_SERVER_LIST_PLAY_TIME_CAPTION                       :{BLACK}Play time
+STR_NETWORK_SERVER_LIST_PLAY_TIME_CAPTION_TOOLTIP               :{BLACK}Time played while{}game was not paused
 STR_NETWORK_SERVER_LIST_INFO_ICONS_TOOLTIP                      :{BLACK}Language, server version, etc.
 
 STR_NETWORK_SERVER_LIST_CLICK_GAME_TO_SELECT                    :{BLACK}Click a game from the list to select it
@@ -2379,6 +2380,7 @@ STR_NETWORK_SERVER_LIST_SERVER_ADDRESS                          :{SILVER}Server 
 STR_NETWORK_SERVER_LIST_INVITE_CODE                             :{SILVER}Invite code: {WHITE}{RAW_STRING}
 STR_NETWORK_SERVER_LIST_START_DATE                              :{SILVER}Start date: {WHITE}{DATE_SHORT}
 STR_NETWORK_SERVER_LIST_CURRENT_DATE                            :{SILVER}Current date: {WHITE}{DATE_SHORT}
+STR_NETWORK_SERVER_LIST_PLAY_TIME                               :{SILVER}Play time: {WHITE}{NUM}h {NUM}m
 STR_NETWORK_SERVER_LIST_GAMESCRIPT                              :{SILVER}Game Script: {WHITE}{RAW_STRING} (v{NUM})
 STR_NETWORK_SERVER_LIST_PASSWORD                                :{SILVER}Password protected!
 STR_NETWORK_SERVER_LIST_SERVER_OFFLINE                          :{SILVER}SERVER OFFLINE
--- a/src/network/core/network_game_info.cpp
+++ b/src/network/core/network_game_info.cpp
@@ -14,6 +14,7 @@
 #include "../../core/bitmath_func.hpp"
 #include "../../company_base.h"
 #include "../../timer/timer_game_calendar.h"
+#include "../../timer/timer_game_tick.h"
 #include "../../debug.h"
 #include "../../map_func.h"
 #include "../../game/game.hpp"
@@ -123,7 +124,7 @@ void CheckGameCompatibility(NetworkGameInfo &ngi)
 void FillStaticNetworkServerGameInfo()
 {
 	_network_game_info.use_password   = !_settings_client.network.server_password.empty();
-	_network_game_info.start_date     = TimerGameCalendar::ConvertYMDToDate(_settings_game.game_creation.starting_year, 0, 1);
+	_network_game_info.calendar_start = TimerGameCalendar::ConvertYMDToDate(_settings_game.game_creation.starting_year, 0, 1);
 	_network_game_info.clients_max    = _settings_client.network.max_clients;
 	_network_game_info.companies_max  = _settings_client.network.max_companies;
 	_network_game_info.map_width      = Map::SizeX();
@@ -149,7 +150,8 @@ const NetworkServerGameInfo *GetCurrentNetworkServerGameInfo()
 	 */
 	_network_game_info.companies_on  = (byte)Company::GetNumItems();
 	_network_game_info.spectators_on = NetworkSpectatorCount();
-	_network_game_info.game_date     = TimerGameCalendar::date;
+	_network_game_info.calendar_date = TimerGameCalendar::date;
+	_network_game_info.ticks_playing = TimerGameTick::counter;
 	return &_network_game_info;
 }
 
@@ -194,6 +196,9 @@ void SerializeNetworkGameInfo(Packet *p, const NetworkServerGameInfo *info, bool
 	/* Update the documentation in game_info.h on changes
 	 * to the NetworkGameInfo wire-protocol! */
 
+	/* NETWORK_GAME_INFO_VERSION = 7 */
+	p->Send_uint64(info->ticks_playing);
+
 	/* NETWORK_GAME_INFO_VERSION = 6 */
 	p->Send_uint8(send_newgrf_names ? NST_GRFID_MD5_NAME : NST_GRFID_MD5);
 
@@ -227,8 +232,8 @@ void SerializeNetworkGameInfo(Packet *p, const NetworkServerGameInfo *info, bool
 	}
 
 	/* NETWORK_GAME_INFO_VERSION = 3 */
-	p->Send_uint32(info->game_date.base());
-	p->Send_uint32(info->start_date.base());
+	p->Send_uint32(info->calendar_date.base());
+	p->Send_uint32(info->calendar_start.base());
 
 	/* NETWORK_GAME_INFO_VERSION = 2 */
 	p->Send_uint8 (info->companies_max);
@@ -267,6 +272,10 @@ void DeserializeNetworkGameInfo(Packet *p, NetworkGameInfo *info, const GameInfo
 	 * to the NetworkGameInfo wire-protocol! */
 
 	switch (game_info_version) {
+		case 7:
+			info->ticks_playing = p->Recv_uint64();
+			FALLTHROUGH;
+
 		case 6:
 			newgrf_serialisation = (NewGRFSerializationType)p->Recv_uint8();
 			if (newgrf_serialisation >= NST_END) return;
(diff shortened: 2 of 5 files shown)


PeterN:
"older servers" probably includes 13.x?

Is it reasonable to adjust the time based on server major version? Not that it really matters, given you won't be able to play on them anyway.

TrueBrain:
> "older servers" probably includes 13.x?
> 
> Is it reasonable to adjust the time based on server major version? Not that it really matters, given you won't be able to play on them anyway.

The issue is that the version is sent by string. So that would be an annoying fuzzy match and stuff. So I was like .. what is going to happen .. we release 14.0, within a few days everyone upgraded, and nobody will actually notice :P

And even so, the difference is so little, that it would require someone with either a very long running server, or a stopwatch, to spot the difference. I mean, it is just 10%, no, 11%, no, 12%, aaarrrrrgggghhhhh.

But yeah, in theory, we could. I could also multiply with the old value; that way the nightlies are the only ones who are "wrong". But even that .. too-much-effort-too-little-benefit, if you ask me :)

If you disagree, let me know :) I am just being lazy here :)

PeterN:
No disagreement. If it was a number it would be simpler but as it's not let's not care :D

src/network/core/network_game_info.cpp
```diff
@@ -149,7 +150,8 @@ const NetworkServerGameInfo *GetCurrentNetworkServerGameInfo()
 	 */
 	_network_game_info.companies_on  = (byte)Company::GetNumItems();
 	_network_game_info.spectators_on = NetworkSpectatorCount();
-	_network_game_info.game_date     = TimerGameCalendar::date;
+	_network_game_info.calendar_date = TimerGameCalendar::date;
+	_network_game_info.ticks_running  = TimerGameTick::counter;
```
rubidium42:
```suggestion
	_network_game_info.ticks_running = TimerGameTick::counter;
```

src/network/core/network_game_info.h
```diff
@@ -92,8 +95,9 @@ enum NewGRFSerializationType {
  */
 struct NetworkServerGameInfo {
 	GRFConfig *grfconfig;        ///< List of NewGRF files used
-	TimerGameCalendar::Date start_date; ///< When the game started
-	TimerGameCalendar::Date game_date;  ///< Current date
+	TimerGameCalendar::Date calendar_start; ///< When the game started.
+	TimerGameCalendar::Date calendar_date;  ///< Current calendar date.
+	TimerGameTick::TickCounter ticks_running;  ///< Amount of ticks the game has been running.
```
rubidium42: Not sure whether the comment of `calendar_date` should align with `calendar_start`. Though the two spaces for `ticks_running` seems too much.
```suggestion
	TimerGameCalendar::Date calendar_date; ///< Current calendar date.
	TimerGameTick::TickCounter ticks_running; ///< Amount of ticks the game has been running.
```

src/lang/english.txt
```diff
@@ -2363,8 +2363,8 @@ STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION                        :{BLACK}Map size
 STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION_TOOLTIP                :{BLACK}Map size of the game{}Click to sort by area
 STR_NETWORK_SERVER_LIST_DATE_CAPTION                            :{BLACK}Date
 STR_NETWORK_SERVER_LIST_DATE_CAPTION_TOOLTIP                    :{BLACK}Current date
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION                           :{BLACK}Years
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION_TOOLTIP                   :{BLACK}Number of years{}the game is running
+STR_NETWORK_SERVER_LIST_HOURS_CAPTION                           :{BLACK}Hours
+STR_NETWORK_SERVER_LIST_HOURS_CAPTION_TOOLTIP                   :{BLACK}Number of hours{}the game is running
```
2TallTyler: No idea what the line break is doing here, but maybe:

```suggestion
STR_NETWORK_SERVER_LIST_HOURS_CAPTION_TOOLTIP                   :{BLACK}Number of hours{}that the game has been played
```

src/network/core/network_game_info.cpp
2TallTyler: I know you don't care about nice alignment, but using two spaces doesn't align with anything 😛 

```suggestion
	_network_game_info.ticks_running = TimerGameTick::counter;
```

src/network/network_gui.cpp
```diff
@@ -283,18 +283,20 @@ class NetworkGameWindow : public Window {
...
 	{
-		auto r = a->info.game_date - b->info.game_date;
+		auto r = a->info.calendar_date - b->info.calendar_date;
 		return (r != 0) ? r < 0 : NGameClientSorter(a, b);
 	}
 
-	/** Sort servers by the number of days the game is running */
-	static bool NGameYearsSorter(NetworkGameList * const &a, NetworkGameList * const &b)
+	/** Sort servers by the number of ticks the game is running. */
+	static bool NGameRunningDateSorter(NetworkGameList * const &a, NetworkGameList * const &b)
```
rubidium42: Maybe duration is a better name than date? Maybe even `NGameDurationSorter`?
```suggestion
	static bool NGameRunningDurationSorter(NetworkGameList * const &a, NetworkGameList * const &b)
```

src/network/network_gui.cpp
```diff
@@ -392,17 +394,15 @@ class NetworkGameWindow : public Window {
...
 				/* current date */
 				Rect date = nwid->GetCurrentRect();
-				TimerGameCalendar::YearMonthDay ymd = TimerGameCalendar::ConvertDateToYMD(cur_item->info.game_date);
+				TimerGameCalendar::YearMonthDay ymd = TimerGameCalendar::ConvertDateToYMD(cur_item->info.calendar_date);
 				SetDParam(0, ymd.year);
 				DrawString(date.left, date.right, y + text_y_offset, STR_JUST_INT, TC_BLACK, SA_HOR_CENTER);
 			}
 
 			if (const NWidgetBase *nwid = this->GetWidget<NWidgetBase>(WID_NG_YEARS); nwid->current_x != 0) {
 				/* number of years the game is running */
```
rubidium42:
```suggestion
				/* number of hours the game is running. */
```

2TallTyler (commented):
No screenshot? 😛 

I am not bothered in the slightest about pre-13.x servers showing a slightly wrong total based on a different ms-per-tick speed. As Peter said, you won't be able to play on them. My understanding of this value is that it's used to determine if you can join a game and still have room to build, or if the map is filled up and nearing a reset.

src/lang/english.txt
TrueBrain: This `{}` was already in there. It is there for all tooltips. I don't feel like making that change in this PR, sorry :(

2TallTyler: It doesn't matter, suggestion is about wording. Edited. 🙂

src/lang/english.txt
```diff
@@ -2380,6 +2380,7 @@ STR_NETWORK_SERVER_LIST_SERVER_ADDRESS                          :{SILVER}Server
 STR_NETWORK_SERVER_LIST_INVITE_CODE                             :{SILVER}Invite code: {WHITE}{RAW_STRING}
 STR_NETWORK_SERVER_LIST_START_DATE                              :{SILVER}Start date: {WHITE}{DATE_SHORT}
 STR_NETWORK_SERVER_LIST_CURRENT_DATE                            :{SILVER}Current date: {WHITE}{DATE_SHORT}
+STR_NETWORK_SERVER_LIST_PLAY_TIME                               :{SILVER}Play time: {WHITE}{NUM} hours
```
rubidium42: Would it make sense to use `{DECIMAL}`, so you can differentiate between "running a few minutes" vs "running for almost an hour"? In some competitive games that might be quite important.

I'm myself still really wondering whether it would make sense, so maybe some others can give their opinions about it before making any changes.

src/lang/english.txt
TrueBrain: Counter suggestion: `Hours of play time`

Not sure why it has to be that long (it was already weirdly long)

src/lang/english.txt
TrueBrain: Ideally, I would have it say `1d12h3m` .. but that went a bit far for this PR :D

Decimals are a bit annoying, as 1.25 is 1h15m .. and I find that confusing to read. So I rather avoid that.

src/lang/english.txt
2TallTyler: Even better! 😄

src/lang/english.txt
TrueBrain: Going with: `Play time: 12 hours, 43 minutes`

That should work (for now) :)

src/network/network_coordinator.cpp
```diff
@@ -500,7 +500,7 @@ void ClientNetworkCoordinatorSocketHandler::GetListing()
 
 	Packet *p = new Packet(PACKET_COORDINATOR_CLIENT_LISTING);
 	p->Send_uint8(NETWORK_COORDINATOR_VERSION);
-	p->Send_uint8(NETWORK_GAME_INFO_VERSION);
+	p->Send_uint8(6); //NETWORK_GAME_INFO_VERSION);
```
TrueBrain: Will be fixed before going un-draft :)

TrueBrain:
Ready for review.

New Game Coordinator that supports game-info v7 is deployed. Tested it in several ways:

- Old servers being queried from new clients: show an estimation of "play time" (based on calendar)
- Old servers being queried from old clients: still works
- New servers being queried from new clients: shows "play time" correctly via new info
- New servers being queried from old clients: shows all the info the old client shows, correctly

So that seems to wrap it up as far as testing goes :)

src/lang/english.txt
```diff
@@ -2362,8 +2362,9 @@ STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION                        :{BLACK}Map size
 STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION_TOOLTIP                :{BLACK}Map size of the game{}Click to sort by area
 STR_NETWORK_SERVER_LIST_DATE_CAPTION                            :{BLACK}Date
 STR_NETWORK_SERVER_LIST_DATE_CAPTION_TOOLTIP                    :{BLACK}Current date
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION                           :{BLACK}Years
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION_TOOLTIP                   :{BLACK}Number of years{}the game is running
+STR_NETWORK_SERVER_LIST_PLAY_TIME_SHORT                         :{BLACK}{NUM}h {NUM}m
+STR_NETWORK_SERVER_LIST_PLAY_TIME_CAPTION                       :{BLACK}Play time
+STR_NETWORK_SERVER_LIST_PLAY_TIME_CAPTION_TOOLTIP               :{BLACK}Play time
```
rubidium42: A tooltip with the exact same text seems pointless to me. Maybe explain that it is about the time the game was not paused?
```suggestion
STR_NETWORK_SERVER_LIST_PLAY_TIME_CAPTION_TOOLTIP               :{BLACK}Time game was not paused
```

src/lang/english.txt
```diff
@@ -2362,8 +2362,9 @@ STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION                        :{BLACK}Map size
 STR_NETWORK_SERVER_LIST_MAP_SIZE_CAPTION_TOOLTIP                :{BLACK}Map size of the game{}Click to sort by area
 STR_NETWORK_SERVER_LIST_DATE_CAPTION                            :{BLACK}Date
 STR_NETWORK_SERVER_LIST_DATE_CAPTION_TOOLTIP                    :{BLACK}Current date
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION                           :{BLACK}Years
-STR_NETWORK_SERVER_LIST_YEARS_CAPTION_TOOLTIP                   :{BLACK}Number of years{}the game is running
+STR_NETWORK_SERVER_LIST_PLAY_TIME_SHORT                         :{BLACK}{NUM}h {NUM}m
```
rubidium42, line 2365: Given I have the feeling the web translator shows the strings in a more-or-less random order,  will the translators have enough context to determine what `h` and `m` mean? Maybe adding a bit of information to the string name will go a long way. Also holds for the non-`_SHORT` version.
```suggestion
STR_NETWORK_SERVER_LIST_PLAY_TIME_SHORT_HOURS_MINUTES           :{BLACK}{NUM}h {NUM}m
```

TrueBrain: I was going for consistency, where a bit higher we have `MAP_SIZE_SHORT` doing the same thing. I was thinking that would give translators a good idea that these are around the same place.

In general, we don't actually postfix string-names with the content of the value.

TrueBrain: To put Discord here:

Ideally, we have something like a `{DURATION}` to resolve this, as that would both de-duplicate the entries and make it much more clear to translators what is going on.

However, the exact syntax is something we need to work out, and possibly it can be used in other places too. For that reason, not for this PR.

With that knowledge, I rather leave the strings as they are, as if we would be to add `{DURATION}`, having a postfix of `_HOURS_MINUTES` would be weird (and renaming is expensive).

rubidium42 (approved):

PR #11895, opened 2024-01-27
Fix: Memory leak in ICUParagraphLayout::NextLine()

PeterN:
## Motivation / Problem

The function `ICUParagraphLayout::NextLine()` calls `icu::BreakIterator::createLineInstance()` but does not clean up after it, resulting in a memory leak whenever lines have to be wrapped.

## Description

While the BreakIterator instance could be deleted every time after use, and doing so would mostly match current behaviour and be simpler, that has potential performance considerations as the line break rules need to be loaded each time.

Instead implement a static instance that is `clone()`d (for thread-safety) and then auto-deleted as necessary.

This uses `std::unique_ptr`s so that we don't need to manage clearing up afterwards.

## Checklist for review

### DIFF
--- a/src/gfx_layout_icu.cpp
+++ b/src/gfx_layout_icu.cpp
@@ -381,6 +381,30 @@ std::vector<ICURun> ItemizeStyle(std::vector<ICURun> &runs_current, FontMap &fon
 	return new ICUParagraphLayout(runs, buff, length);
 }
 
+/* static */ std::unique_ptr<icu::BreakIterator> ICUParagraphLayoutFactory::break_iterator;
+
+/**
+ * Initialize data needed for the ICU layouter.
+ */
+/* static */ void ICUParagraphLayoutFactory::InitializeLayouter()
+{
+	auto locale = icu::Locale(_current_language->isocode);
+	UErrorCode status = U_ZERO_ERROR;
+	ICUParagraphLayoutFactory::break_iterator.reset(icu::BreakIterator::createLineInstance(locale, status));
+	assert(U_SUCCESS(status));
+}
+
+/**
+ * Get a thread-safe line break iterator.
+ * @returns unique_ptr managed BreakIterator instance.
+ */
+/* static */ std::unique_ptr<icu::BreakIterator> ICUParagraphLayoutFactory::GetBreakIterator()
+{
+	assert(ICUParagraphLayoutFactory::break_iterator != nullptr);
+
+	return std::unique_ptr<icu::BreakIterator>(ICUParagraphLayoutFactory::break_iterator->clone());
+}
+
 std::unique_ptr<const ICUParagraphLayout::Line> ICUParagraphLayout::NextLine(int max_width)
 {
 	std::vector<ICURun>::iterator start_run = this->current_run;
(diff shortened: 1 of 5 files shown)


src/gfx_layout_icu.cpp
```diff
@@ -381,6 +381,32 @@ std::vector<ICURun> ItemizeStyle(std::vector<ICURun> &runs_current, FontMap &fon
 	return new ICUParagraphLayout(runs, buff, length);
 }
 
+struct BreakIteratorInstance
+{
+	std::unique_ptr<icu::BreakIterator> break_iterator;
+	std::string isocode;
+
+	static BreakIteratorInstance instance;
```
rubidium42: Won't making this instance thread local, make it even thread safer?
```suggestion
	thread_local static BreakIteratorInstance instance;
```
Not sure whether this yields the result you're looking for though. Also not sure what the performance impact of `thread_local` is. It's just something to consider.

PeterN: I did actually try this but it made it really hard to debug it because the debugger is never in the right thread, or something :D

PeterN:
This is now rewritten to use explicit initialization, avoiding the faff of testing if the iterator exists and is for the right language every call, and avoiding the non-thread-safe initialization.

While doing this I noticed per-layouter specific stuff all over the place, but I think tidying that up is the job of another PR...

Also a quick benchmark of this applied shows an improvement on wrapping the textfile window showing the OpenTTD Changelog of 50ms down to 30ms, so it genuinely does help a little bit :)

src/gfx_layout_icu.cpp
```diff
@@ -381,6 +381,30 @@ std::vector<ICURun> ItemizeStyle(std::vector<ICURun> &runs_current, FontMap &fon
...
+
+/**
+ * Get a thread-safe line break iterator.
+ * @returns unique_ptr managed BreakIterator instance.
+ */
+/* static */ std::unique_ptr<icu::BreakIterator> ICUParagraphLayoutFactory::GetBreakIterator()
+{
+	assert(ICUParagraphLayoutFactory::break_iterator != nullptr);
+
+	return std::unique_ptr<icu::BreakIterator>(ICUParagraphLayoutFactory::break_iterator->clone());
```
TrueBrain, line 405: Why not a `shared_ptr` instead?

Or better worded: if we would use a `shared_ptr` , can we avoid the `clone()` call?

PeterN: Because there's no need for a shared_ptr. The main iterator instance is not thread-safe, using a `clone()` gets around that without having to recreate the iterator from scratch. unique_ptr is used so that we know it's deleted properly.

Alternatively, like the previous version, the main iterator can be thread_local, but then this method of initialization doesn't work.

(I'm not sure if the break iterator even needs to be thread-safe, but I think defaulting to assuming it should be is wise.)

src/strings.cpp
```diff
@@ -37,6 +37,7 @@
 #include "network/network_content_gui.h"
 #include "newgrf_engine.h"
 #include "core/backup_type.hpp"
+#include "gfx_layout_icu.h"
```
TrueBrain: This is missing ifdef guards :D

src/strings.cpp
```diff
@@ -1974,6 +1975,7 @@ bool ReadLanguagePack(const LanguageMetadata *lang)
 	if (U_FAILURE(status)) {
 		_current_collator.reset();
 	}
+	ICUParagraphLayoutFactory::InitializeLayouter();
```
TrueBrain: This is missing HARFBUZZ ifdef-guard

PeterN: Ah okay, I incorrectly assumed the WITH_ICU_I18N was all part of the same stuff.

TrueBrain (approved):

PR #11896, opened 2024-01-27
Change: Limit scheduled window invalidation events to just one.

PeterN:
## Motivation / Problem

Stacking up window invalidation events does not serve much use as in all cases the data passed is the same.

When processing the events, it would then invalidate windows multiple times for no reason.

## Description

Instead, store the intended invalidation data in a `std::optional`. Windows are now only invalidated once even if multiple scheduled invalidations occur. The last data value passed is the value used for invalidating.

## Limitations

The `std::vector` queue probably had an intended purpose at some point but given all invalidations using it use the same data it seemed pointless to keep it around.

## Checklist for review

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -3121,7 +3121,9 @@ void Window::InvalidateData(int data, bool gui_scope)
 	this->SetDirty();
 	if (!gui_scope) {
 		/* Schedule GUI-scope invalidation for next redraw. */
-		this->scheduled_invalidation_data.push_back(data);
+		if (std::find(std::begin(this->scheduled_invalidation_data), std::end(this->scheduled_invalidation_data), data) == std::end(this->scheduled_invalidation_data)) {
+			this->scheduled_invalidation_data.push_back(data);
+		}
 	}
 	this->OnInvalidateData(data, gui_scope);
 }


TrueBrain:
I guess the vector has a reason:

There are two windows open about two different road vehicles. They both get invalidated with their `data` (as behind the RV index).

Now only one of the two gets updated? Or am I missing something?

PeterN:
There are no deferred `InvalidateData()` that do this. The second parameter of `InvaldateData()` must be false to be a deferred invalidation, otherwise it is processed immediately.

There are only two such calls in the whole code, and both of them only pass a single constant value.

We could keep the vector (or use an unordered_map) and ensure only one instance of each value is provided. But no callers need this at the moment.

src/window.cpp
```diff
@@ -3121,21 +3121,20 @@ void Window::InvalidateData(int data, bool gui_scope)
 	this->SetDirty();
 	if (!gui_scope) {
 		/* Schedule GUI-scope invalidation for next redraw. */
-		this->scheduled_invalidation_data.push_back(data);
+		this->scheduled_invalidation_data = data;
```
TrueBrain: Given your explanation, mind adding an `assert` here to ensure `scheduled_invalidation_data` is either equal to `data` or not set? Just to avoid future-us assuming it is possible, and breaking weird shit :D

TrueBrain (dismissed):

TrueBrain (dismissed):

TrueBrain (dismissed):

PeterN:
Rewritten to just limit insertion to one-data value per event. There is the possibility of a window requiring a specific order of data values which would now not necessarily be correct if multiple same-data-value events are queued...

I might just put this on hold because it clearly needs more thought that I have brain right now.

TrueBrain:
PS: if you ever going to merge it, I think it is a `Codechange`, not an actual `Change` :)

PR #11897, opened 2024-01-27
Change: Add dividers between signal groups

2TallTyler:
## Motivation / Problem

#11798 proposes to cycle signals by group. But as @TrueBrain pointed out, it's not apparent which groups this refers to.

Also, new players who find the button to show all signal types, might not realize that block and path signals behave in different ways so it would be nice to have a hint that we have two categories of signals.

## Description

Add a divider between block and path signal groups.

Also refactor the widget structure to make it simpler and better documented.

full

When block signals are hidden, the divider is hidden too.

reduced

## Limitations

I originally added a divider between path signals and the convert/autofill buttons, but it had width issues.

## Checklist for review


src/rail_gui.cpp
```diff
@@ -1856,44 +1851,62 @@ struct BuildSignalWindow : public PickerWindowBase {
...
-			EndContainer(),
-			NWidget(NWID_SELECTION, INVALID_COLOUR, WID_BS_SEMAPHORE_EXIT_SEL),
-				NWidget(WWT_PANEL, COLOUR_DARK_GREEN, WID_BS_SEMAPHORE_EXIT), SetDataTip(STR_NULL, STR_BUILD_SIGNAL_SEMAPHORE_EXIT_TOOLTIP), EndContainer(),
+
+	/* Container for both signal groups, spacers, and convert/autofill buttons. */
+	NWidget(NWID_HORIZONTAL),
+		/* Block signals. */
+		NWidget(NWID_VERTICAL, NC_EQUALSIZE),
+			/* Semaphore block signals. */
+			NWidget(NWID_SELECTION, INVALID_COLOUR, WID_BS_SEMAPHORE_BLOCK_SEL),
```
PeterN: You could put this before the `NWID_VERTICAL` and have one selection instead of two?

src/rail_gui.cpp
```diff
@@ -1856,44 +1851,62 @@ struct BuildSignalWindow : public PickerWindowBase {
...
-			EndContainer(),
-			NWidget(NWID_SELECTION, INVALID_COLOUR, WID_BS_ELECTRIC_ENTRY_SEL),
-				NWidget(WWT_PANEL, COLOUR_DARK_GREEN, WID_BS_ELECTRIC_ENTRY), SetDataTip(STR_NULL, STR_BUILD_SIGNAL_ELECTRIC_ENTRY_TOOLTIP), EndContainer(),
-			EndContainer(),
-			NWidget(NWID_SELECTION, INVALID_COLOUR, WID_BS_ELECTRIC_EXIT_SEL),
-				NWidget(WWT_PANEL, COLOUR_DARK_GREEN, WID_BS_ELECTRIC_EXIT), SetDataTip(STR_NULL, STR_BUILD_SIGNAL_ELECTRIC_EXIT_TOOLTIP), EndContainer(),
+
+		/* Divider (only shown if block signals visible). */
+		NWidget(NWID_SELECTION, INVALID_COLOUR, WID_BS_BLOCK_SPACER_SEL),
+			NWidget(WWT_PANEL, COLOUR_DARK_GREEN), SetMinimalSize(4, 0), EndContainer(),
```
PeterN, line 1884: While I used 4 in my example, it's preferable to use a WidgetDimension here. But I don't think there is one for this purpose yet, as the other toolbars that have these spacers also use random pixel values, so we can 'forgive' this for the moment :)

2TallTyler: As discussed on Discord, this should be a separate PR since the dimensions in other toolbars is wildly inconsistent.

DefinitelyNotRau117:
2 thin buttons and one thick separator (standard path signal variant) looks pretty unaestetic. Unusable filler areas is bad design (in my opinion). Stretched buttons were less noisy and slightly more convinient, like you need to aim your mouse a little less at the thick button.

In other languages, with longer names, it will look even more ridiculous and pointless. Maybe just shortened “signal selection” to “signals”

2TallTyler:
I removed the divider between path signals and the convert/autofill button, and updated the screenshots in the PR description.

TrueBrain (approved):

PR #11902, opened 2024-01-27
Feature: Infinite money mode

merni-ns:
## Motivation / Problem

Many players want to play in a "sandbox" style where they don't care about the money-making mechanics and just want to build cool networks. At present, this has to be achieved by cheating arbitrary amounts of money, which is an unsatisfactory and clunky solution, apart from messing up the company financial statements.

## Description
This PR adds a setting called "Infinite money mode". When enabled, it allows any command to be performed even if the company's money balance is less than the amount required (or even negative). It also disables the company bankruptcy mechanism, and replaces the balance in the status bar with "infinite money". The actual balance is still shown in the finance window. 

This approach has the advantage that it maintains the integrity of the company financial statements. It charges the company all costs, and does not add any extra arbitrary income (like the money cheat does). It also removes the need for occasionally cheating more money or otherwise infusing money -- players can just forget about the existence of the money mechanism for the duration of the game. Furthermore, it works perfectly fine with AI and multiplayer.

Screenshot:
image

## Limitations
- It takes effect on all companies, even AI. At present no check is done whether the company is a human or AI one.
- Looking at vehicle, company, station, etc income statistics, still shows the "real" negative values which (says @TrueBrain) might put off players. (But it is anyway the same at present except for companies)
- Not tested in multiplayer.
- There may be some other edge case where some command/function directly checks how much money a company has instead of going through functions such as `GetAvailableMoneyForCommand`, `CheckCompanyHasMoney`, etc. I have done my best to find and fix all such instances, but there is no guarantee that one more is not lurking somewhere.

## Checklist for review

### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -161,21 +161,6 @@ bool IsCommandAllowedWhilePaused(Commands cmd)
 	return _game_mode == GM_EDITOR || command_type_lookup[_command_proc_table[cmd].type] <= _settings_game.construction.command_pause_level;
 }
 
-/*!
- * This functions returns the money which can be used to execute a command.
- * This is either the money of the current company or INT64_MAX if there
- * is no such a company "at the moment" like the server itself.
- *
- * @return The available money of a company or INT64_MAX
- */
-Money GetAvailableMoneyForCommand()
-{
-	CompanyID company = _current_company;
-	if (!Company::IsValidID(company)) return INT64_MAX;
-	return Company::Get(company)->money;
-}
-
-
 /**
  * Prepare for calling a command proc.
  * @param top_level Top level of command execution, i.e. command from a command.
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -202,14 +202,15 @@ void InvalidateCompanyWindows(const Company *company)
  */
 Money GetAvailableMoney(CompanyID company)
 {
+	if (_settings_game.difficulty.infinite_money) return INT64_MAX;
 	if (!Company::IsValidID(company)) return INT64_MAX;
 	return Company::Get(company)->money;
 }
 
 /**
  * This functions returns the money which can be used to execute a command.
- * This is either the money of the current company or INT64_MAX if there
- * is no such a company "at the moment" like the server itself.
+ * This is either the money of the current company, or INT64_MAX if infinite money
+ * is enabled or there is no such a company "at the moment" like the server itself.
  *
  * @return The available money of the current company or INT64_MAX
  */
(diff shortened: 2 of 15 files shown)


src/saveload/saveload.h
```diff
@@ -371,6 +371,7 @@ enum SaveLoadVersion : uint16_t {
 	SLV_ECONOMY_DATE,                       ///< 326  PR#10700 Split calendar and economy timers and dates.
 	SLV_ECONOMY_MODE_TIMEKEEPING_UNITS,     ///< 327  PR#11341 Mode to display economy measurements in wallclock units.
 	SLV_CALENDAR_SUB_DATE_FRACT,            ///< 328  PR#11428 Add sub_date_fract to measure calendar days.
+	SLV_IGNORE_BANKRUPTCY,                  ///< 329  PR#..... Add setting to ignore bankruptcy
```
```suggestion
	SLV_IGNORE_BANKRUPTCY,                  ///< 329  PR#11902 Add setting to ignore bankruptcy
```

merni-ns: Ah yes. :facepalm:

PeterN:
I don't think the last three bullet points are needed. 3 and 4 are already existing options independently toggeable, and 5 is just... unnecessary?

PeterN:
IMHO sandbox money just means you don't want the game to limit what you do because of money. A player might still want to make an cost-effective network at the same time.

src/saveload/saveload.h
```diff
@@ -371,6 +371,7 @@ enum SaveLoadVersion : uint16_t {
 	SLV_ECONOMY_DATE,                       ///< 326  PR#10700 Split calendar and economy timers and dates.
 	SLV_ECONOMY_MODE_TIMEKEEPING_UNITS,     ///< 327  PR#11341 Mode to display economy measurements in wallclock units.
 	SLV_CALENDAR_SUB_DATE_FRACT,            ///< 328  PR#11428 Add sub_date_fract to measure calendar days.
+	SLV_INFINITE_MONEY,                     ///< 329  PR#11902 Add infinite money mode
```
frosch123: I don't think adding a new setting requires a savegame bump. There are no conversions.

TrueBrain: And adding of settings no longer needs a savegame bump either. It used to, but we solved that a few version ago.

merni-ns: Removed the savegame bump that I added thanks to [outdated docs on the wiki](https://wiki.openttd.org/en/Archive/Source/OpenTTDDevBlackBook/Settings/AddSetting).

src/company_cmd.cpp
```diff
@@ -196,11 +196,12 @@ void InvalidateCompanyWindows(const Company *company)
...
  * Verify whether the company can pay the bill.
  * @param[in,out] cost Money to pay, is changed to an error if the company does not have enough money.
- * @return Function returns \c true if the company has enough money, else it returns \c false.
+ * @return Function returns \c true if the company has enough money or infinite money is enabled,
+ * else it returns \c false.
  */
 bool CheckCompanyHasMoney(CommandCost &cost)
 {
-	if (cost.GetCost() > 0) {
+	if (cost.GetCost() > 0 && !(_settings_game.difficulty.infinite_money)) {
```
2TallTyler: This would be nicer as an early return, as you did in `CompanyCheckBankrupt()`.

2TallTyler (commented):
The commit prefix for this one should definitely be `Feature:`. I suspect there will be a fair amount of excitement for this one. 😄 Other than that and my one code nitpick, it looks good!

src/company_cmd.cpp
merni-ns: I did this and moved the cost condition also to an early return.

2TallTyler (commented):
Sorry, just noticed... All of our .ini files end with an empty line. Add that back and I'll approve this! 😄

2TallTyler (dismissed):

rubidium42:
Sadly the infinite money isn't as infinite as I would expect. When I try to clear/flatten/raise/lower the whole world, I'm still getting error messages about insufficient funds (start a game with no to little money). This is due to those functions checking for money in a different manner, because they are not all-or-nothing commands, but do-as-much-as-possible for the money you have commands.

There might be more commands than just clear/flatten/raise/lower that exhibit this behaviour, but I haven't looked into that.

src/command.cpp
```diff
@@ -97,7 +97,7 @@ inline constexpr auto MakeCommandsFromTraits(std::integer_sequence<T, i...>) noe
 static constexpr auto _command_proc_table = MakeCommandsFromTraits(std::make_integer_sequence<std::underlying_type_t<Commands>, CMD_END>{});
 
 
-/*!
```
2TallTyler: Changing comments of unrelated functions should be in a separate PR. 🙂

2TallTyler (approved):

Ufiby:
In a slightly different way, let's call it "free money". It gives you the opportunity to build the way you want, buy transport, service and interest rate, everything costs 0 money. An idea?

PeterN:
Should be squashed on merge as all the commits (except #2) are part of the feature.

PR #11904, opened 2024-01-27
Fix: Extra refit button when train/RV is in a depot

2TallTyler:
## Motivation / Problem

refit

When a train or road vehicle is stopped in a depot, there are two refit buttons. Only one works. Why?

## Description

The disabled button is a placeholder for the dropdown that appears when a station is highlighted. It's called `WID_O_EMPTY` and is a wonderful exploration into vestigial code. It's documented as a placeholder to be used when looking at other companies' orders -- but if you try that, you'll see that none of the buttons are shown in this case.

The widget selection code indicates what is supposed to be here:

```
DP_RIGHT_EMPTY     = 0, ///< Display an empty panel in the right button of the top row of the train/rv order window.
```

Let's do that instead.

empty_box

## Limitations

It would be cleaner to show a three-column layout instead. But the widget code for the orders window is a nightmare and refactoring it is a much bigger project.

## Checklist for review

### DIFF
--- a/src/order_gui.cpp
+++ b/src/order_gui.cpp
@@ -778,9 +778,6 @@ public:
 		this->CreateNestedTree();
 		this->vscroll = this->GetScrollbar(WID_O_SCROLLBAR);
 		this->FinishInitNested(v->index);
-		if (v->owner == _local_company) {
-			this->DisableWidget(WID_O_EMPTY);
-		}
 
 		this->selected_order = -1;
 		this->order_over = INVALID_VEH_ORDER_ID;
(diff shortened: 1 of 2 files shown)


src/order_gui.cpp
```diff
@@ -1600,8 +1597,7 @@ static constexpr NWidgetPart _nested_orders_train_widgets[] = {
 															SetDataTip(STR_ORDER_SERVICE, STR_ORDER_SERVICE_TOOLTIP), SetResize(1, 0),
 				EndContainer(),
 				NWidget(NWID_SELECTION, INVALID_COLOUR, WID_O_SEL_TOP_RIGHT),
-					NWidget(WWT_TEXTBTN, COLOUR_GREY, WID_O_EMPTY), SetMinimalSize(93, 12), SetFill(1, 0),
-															SetDataTip(STR_ORDER_REFIT, STR_ORDER_REFIT_TOOLTIP), SetResize(1, 0),
+					NWidget(WWT_PANEL, COLOUR_GREY), SetMinimalSize(93, 12), EndContainer(),
```
frosch123: Don't the `SetFill`  and `SetResize` have to stay?

2TallTyler: Yes, I will add those back tomorrow. 🙂

frosch123 (approved):

PR #11916, opened 2024-01-28
Change: only open story-book in center when a GS does it

### LINKED ISSUE #9030: Global story book opens on top of company story book
## Version of OpenTTD
1.11.0

## Expected result
Not open two near-identical windows on top of each other. Be more like goal-window, I would say.

## Actual result
Two windows of identical size open on top of each other, with only minor difference in colouring. Very confusing.

## Steps to reproduce
- Open company story book
- Open global story book
- Be: huh? Did I open global?
- Close global
- Notice company is below that
- Be: huh?

No other window I know off opens on top of another. Seems like a simple boo-boo somehwere.

Additionally, Goal Window has a nice "Global" and "Company" button, but Story Book doesn't. This feels inconsistent. Maybe both can be tackled at the same time. But the overlapping windows was by far the more frustrating part :)

TrueBrain:
## Motivation / Problem

Fixes #9030.

As mentioned in that ticket, it really annoys me that the story book opens in the center of the screen. Like it is feeling really important, more so than other windows (the audacity!). But worst of all: if you open two of them, they are flat on top of each other, and no way to see that is the case. I consider this bad UX.

There are other windows that open in the center, but they are all windows that require attention. Editbox, engine preview, ... Opening the story book from the menu doesn't require more attention than opening the finance window. Or the Company window. I mean .. it is just a window! /me ends rant here.

The only reason I can understand that opening in the center would be useful, is when it is done from a GS, as it requires attention. 

## Description

By default, open the story book like any other window, somewhere where there is space.

The only exception: when it is opened via the `CmdShowStoryPage`, which GS uses to open a story page.

## Checklist for review

### DIFF
--- a/src/story_gui.cpp
+++ b/src/story_gui.cpp
@@ -969,7 +969,14 @@ static constexpr NWidgetPart _nested_story_book_widgets[] = {
 };
 
 static WindowDesc _story_book_desc(__FILE__, __LINE__,
-	WDP_CENTER, "view_story", 400, 300,
+	WDP_AUTO, "view_story", 400, 300,
+	WC_STORY_BOOK, WC_NONE,
+	0,
+	std::begin(_nested_story_book_widgets), std::end(_nested_story_book_widgets)
+);
+
+static WindowDesc _story_book_gs_desc(__FILE__, __LINE__,
+	WDP_CENTER, "view_story_gs", 400, 300,
 	WC_STORY_BOOK, WC_NONE,
 	0,
 	std::begin(_nested_story_book_widgets), std::end(_nested_story_book_widgets)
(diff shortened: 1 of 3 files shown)


src/story_gui.cpp
```diff
@@ -617,6 +619,23 @@ struct StoryBookWindow : Window {
...
+			int16_t default_width  = std::max(this->window_desc->GetDefaultWidth(),  sm_width);
+			int16_t default_height = std::max(this->window_desc->GetDefaultHeight(), sm_height);
+
+			pt.x = (_screen.width - default_width) / 2;
+			pt.y = (_screen.height - default_height) / 2;
+
+			return pt;
+		}
+
+		return Window::OnInitialPosition(sm_width, sm_height, window_number);
```
rubidium42: I'd rather make `LocalGetWindowPlacement` non-static and add a parameter for the position which the current only caller would fill with `desc->default_pos`. Then this function can simply call `LocalGetWindowPlacement` with the appropriate `WDP` value.
That way someone looking for implementing the same functionality in another window won't be looking at duplicating this centering code.

TrueBrain: I tried that before this attempt, but it looks really weird. You give it a description of the window, but also a field from the description, as someone might change it. It just didn't sit right with me, and felt hackish.

But, this solution still feels hackish how ever I slice it .. so I am going back to an other way of doing this, which is just way simpler and more descriptive :D

Back&forth, back and forth .... :)

rubidium42 (approved):

PR #11917, opened 2024-01-29
Change: Make the unloading button toggle Transfer, not Unload All

2TallTyler:
## Motivation / Problem

The order window has a dropdown to select the unloading type of the selected order. Clicking the button without opening the dropdown sets the selected order to `Unload all`.

This action forcibly unloads the vehicle's cargo. If it is accepted by the station, it is consumed by whatever industry/houses are in the station's coverage, while if it is not accepted, it is placed on the station platform, just like `Transfer`.

I never use `Unload all` but I do use `Transfer` frequently.

## Description

Change the default action of the button to `Transfer`.

## Limitations

Some people might prefer `Unload all`. 🙂

The real question is: What is a more common action, for everybody besides me?

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -4553,7 +4553,7 @@ STR_ORDER_DROP_FULL_LOAD_ANY                                    :Full load any c
 STR_ORDER_DROP_NO_LOADING                                       :No loading
 STR_ORDER_TOOLTIP_FULL_LOAD                                     :{BLACK}Change the loading behaviour of the highlighted order
 
-STR_ORDER_TOGGLE_UNLOAD                                         :{BLACK}Unload all
+STR_ORDER_TOGGLE_TRANSFER                                       :{BLACK}Transfer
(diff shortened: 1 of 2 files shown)


M3Henry:
I'm in support of this, I too use `Transfer` and never `Unload all`

TBH it would also be nice to do `Ctrl-Shift-Click` to when creating a `Transfer` order in the same way `Ctrl-Click` creates a `Full load any` order.

Does anyone use `Full load all` regularly? If so this might be a calling for a pair of settings "Default load order type" and "Default unload order type"

PeterN:
Shift is estimate the cost of Ctrl-Click :-)

TrueBrain:
If I am being honest, I think we are pushing it with this change. People are habit animals, and this change is really "I have an itch", more than a broader "what benefits people". And as it is not a default of some setting, but something a user cannot actually change back, I am just not a fan or this, especially this late in the release process.

You marked this as 14.0, as otherwise I would have said: let this go to 15.0, and see how it lands in the nightly. But for now, for me this is a "no".

PR #11921, opened 2024-01-29
Add: {CURRENCY_SHORT} only did k / m suffix. Add bn / tn and make translatable.

TrueBrain:
## Motivation / Problem

While creating #11915, I noticed this:

300295128-07e07641-f157-429c-b4ee-675a5d17743a

And it really really annoyed me that it does read `M`, but ... still a HUGE number. Clearly that must be broken.

## Description

Turn out, the "compact" functionality was written for 32bit, and never extended to 64bit. So, here we go! Now it looks like this.

image

Further more, turns out that the old code uses `k` and `M`, which is neither UK nor US method of presenting money. And as some comments below show, some people really would like that. As such, they are now part of the string system, and anyone can translate them how ever they like for their language to fit :)

PS: our `Money` class doesn't want to add/divide by `int64_t`, and only with `Money`. This seems a deliberate action in the `Money` class, to avoid people doing math with containers that are larger than `Money` (and `int64_t` is larger than `Money`, by 1 value on each side).

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -5799,6 +5799,11 @@ STR_TOWN_NAME                                                   :{TOWN}
 STR_VEHICLE_NAME                                                :{VEHICLE}
 STR_WAYPOINT_NAME                                               :{WAYPOINT}
 
+STR_CURRENCY_SHORT_KILO                                         :{NBSP}k
+STR_CURRENCY_SHORT_MEGA                                         :{NBSP}m
+STR_CURRENCY_SHORT_GIGA                                         :{NBSP}bn
+STR_CURRENCY_SHORT_TERA                                         :{NBSP}tn
+
 STR_JUST_CARGO                                                  :{CARGO_LONG}
 STR_JUST_RIGHT_ARROW                                            :{RIGHT_ARROW}
 STR_JUST_CHECKMARK                                              :{CHECKMARK}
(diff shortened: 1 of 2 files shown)


PeterN:
I don't know where you got "G" from. Money is in Millions, Billions and Trillions...

TrueBrain:
> I don't know where you got "G" from. Money is in Millions, Billions and Trillions...

Don't know in what system you live, but I live in: kilo, mega, giga, tera, ...

SI BABY!

LordAro:
Nah, thousands, millions, billions, trillions.

Unless of course you want to make it translatable and configurable per-language ;)

TrueBrain:
> Nah, thousands, millions, billions, trillions.
> 
> Unless of course you want to make it translatable and configurable per-language ;)

Which is the issue and also why I assumed it was kilo, mega. In Dutch it is: "duizend", "miljoen", "miljard", "biljoen" .. so that would make using k / M / B / T rather difficult, as it would confuse Dutch people for one, and I am sure many other languages. And doing an "english only thing" is a bad move imo.

TrueBrain:
As @frosch123 figured out, the UK system uses `k`, `m`, `bn` , `tn`, and the US system uses `K`, `M`, `B`, `T`.

So indeed, the current system appears to use SI units, rather than either of the two. But those are easily confused. To not play childish games, let's fix this properly, and I just made them language configurable. Means we can actually use the letter that work with currencies, and not bother people with SI units :)

frosch123 (approved):

PR #11923, opened 2024-01-29
Change: position error window closer to cursor on large screens

### LINKED ISSUE #9029: Error message opening nowhere near my mouse
## Version of OpenTTD
1.11.0

## Expected result
Error message near my mouse. The reason for this is simple:

You are focused close to your mouse, as that is where you perform the action. So instead of the error appearing .. "somewhere", do it relative close to the mouse so it is seen by the user.
Focus should be localized, and while this wasn't an issue when screens were 640x480, with modern resolutions this is more of an issue.

Alternatively, another form of feedback that the action failed, so I know to look for the error message, works fine for me to. But I was just clicking and clicking being annoyed nothing happened.

## Actual result
Error message somewhere on my screen, maybe in this corner? or this? No there? Owh, wait, here. Mostly caused by a 2k monitor and not scaling the UI (which works fine for most things except this).

It seems they mostly want to be in some corner, far away from you. Like it is scared of being seen.

## Steps to reproduce
- Have a sufficient large screen.
- Open some windows.
- Try to build something that fails.
- Be confused, try again.
- And again.
- Find out that somewhere far away from your mouse is an error telling you it cannot be done.

I have to admit this might be a me-problem, but I think/hope I am not alone in the frustration here.

TrueBrain:
## Motivation / Problem

Fixes #9029.

One of my personal annoyances with the game, is that error windows appear in one of the four corners of the screen.

I don't know about you, but my screens are HUGE. So .. it feels like playing a tennis-match trying to find the error message. Even worse, often it takes for-ever for me to realise that my action isn't actually happening as there is a window popping up outside my field of vision / focus. So that frustrates me, I press some more, and only then: AH! There is an error ..

The reason the error window is opened at a corner is to make the chance it overlaps something you are working on as slim as possible. And on a 640x480 screen that made sense. On a 2k or 4k screen, it does not.

I wanted to fix that. This PR is an attempt in doing so.

## Description

Open the error window always relatively nearby the mouse: 200px above the cursor when possible. When too close to the top, put it 200px below the cursor. This is almost always within the field of vision / focus, and rarely obstructing anything I am working on. Even big stations are still fully visible.

## Limitations

This is a very opinionated way of resolving this. This works for me, but it might not work for you. Comments are open.

## Checklist for review

### DIFF
--- a/src/error_gui.cpp
+++ b/src/error_gui.cpp
@@ -223,26 +223,20 @@ public:
 			return pt;
 		}
 
-		/* Find the free screen space between the main toolbar at the top, and the statusbar at the bottom.
-		 * Add a fixed distance 20 to make it less cluttered.
-		 */
-		int scr_top = GetMainViewTop() + 20;
-		int scr_bot = GetMainViewBottom() - 20;
+		constexpr int distance_to_cursor = 200;
 
-		Point pt = RemapCoords(this->position.x, this->position.y, GetSlopePixelZOutsideMap(this->position.x, this->position.y));
-		const Viewport *vp = GetMainWindow()->viewport;
-		if (this->face == INVALID_COMPANY) {
-			/* move x pos to opposite corner */
-			pt.x = UnScaleByZoom(pt.x - vp->virtual_left, vp->zoom) + vp->left;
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
Works for me! Nice to have a real solution for this instead of an overzealous developer going "works as intended, won't fix" just to make the issue count go down. 😛

PR #11926, opened 2024-01-30
Add: Setting for scaling time in cargo payment calculations

ldpl:
## Motivation / Problem

In the aftermath of #10596, it has become apparent that certain players favor very lengthy routes. As a result, cargo profits, originally optimized for specific distances, do not align with every playstyle. Compounding the issue is that, in some cases (such as maglev), peak distances extend beyond the boundaries, even on a 4k*4k map. Before #10596, this situation was somewhat mitigated by the payment formula not being tailored for large distances, yielding some illogical income figures. However, a more comprehensive solution is now imperative.

While a similar effect can be achieved through cargo aging in a newgrf, this property is associated with the vehicle, not the cargo. Expecting every vehicle newgrf to introduce a scaling parameter is unreasonable, and it doesn't address pure vanilla gameplay.

## Description

Given that the desired optimal profit distance is a characteristic of playstyle and not directly correlated with map size, it cannot be automatically inferred. Therefore, this PR introduces a separate setting that scales time for payment calculations. Players who prefer longer distances can set the scaling below 100%, while those aiming to optimize for shorter distances can set it above 100%. There is also the option to set it to 0 for achieving "flat" payment rates, although the practicality of this is uncertain.

[Screencast from 30-01-24 17:37:10.webm](https://github.com/OpenTTD/OpenTTD/assets/413570/eb2620f9-4c23-4382-8bd1-576ea2d2e019)

## Limitations

More settings for the God of Settings.

## Checklist for review

### DIFF
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -352,4 +352,29 @@ constexpr int RoundDivSU(int a, uint b)
 
 uint32_t IntSqrt(uint32_t num);
 
+/**
+ * Scale a number by the required percentage.
+ *
+ * Calculation is performed in the type U of the num parameter.
+ * The result is clamped to the limits of type T if needed.
+ *
+ * @param num The number to scale.
+ * @param percentage The percentage value. 100% = don't scale.
+ * @return The number scaled by the percentage value.
+ */
+template <typename T, typename U>
+constexpr T ScaleByPercentage(U num, uint16_t percentage)
+{
+	U scaled;
+	/* We might not need to do anything. */
+	if (percentage == 100) {
+		scaled = num;
+	} else {
+		scaled = (num * static_cast<U>(percentage)) / 100;
+	}
+
+	/* Make sure the value fits resulting type T. */
+	return ClampTo<T>(scaled);
+}
+
 #endif /* MATH_FUNC_HPP */
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -984,6 +984,9 @@ Money GetTransportedGoodsIncome(uint num_pieces, uint dist, uint16_t transit_per
 		return 0;
 	}
 
+	/* Scale transit periods according to the game setting. NewGRFs callbacks operate on scaled value for compatibility. */
+	transit_periods = ScaleByPercentage<uint16_t, uint32_t>(transit_periods, _settings_game.economy.payment_time_scale);
+
 	/* Use callback to calculate cargo profit, if available */
 	if (HasBit(cs->callback_mask, CBM_CARGO_PROFIT_CALC)) {
 		uint32_t var18 = ClampTo<uint16_t>(dist) | (ClampTo<uint8_t>(num_pieces) << 16) | (ClampTo<uint8_t>(transit_periods) << 24);
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1520,6 +1520,9 @@ STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE                       :{NUM}
 ###setting-zero-is-special
 STR_CONFIG_SETTING_MINUTES_PER_YEAR_FROZEN                      :0 (calendar time frozen)
 
+STR_CONFIG_SETTING_PAYMENT_TIME_SCALE                           :Time scale for cargo payments: {STRING2}
+STR_CONFIG_SETTING_PAYMENT_TIME_SCALE_HELPTEXT                  :Adjusts the time scale factor (in percentage) in the cargo payment calculations. A lower value means faster deliveries, resulting in higher payments.
+
 STR_CONFIG_SETTING_AUTORENEW_VEHICLE                            :Autorenew vehicle when it gets old: {STRING2}
 STR_CONFIG_SETTING_AUTORENEW_VEHICLE_HELPTEXT                   :When enabled, a vehicle nearing its end of life gets automatically replaced when the renew conditions are fulfilled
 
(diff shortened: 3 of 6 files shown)


src/core/math_func.hpp
```diff
@@ -352,4 +352,34 @@ constexpr int RoundDivSU(int a, uint b)
...
+		scaled = (num * static_cast<U>(percentage)) / 100;
+	}
+
+	/* If there is no type conversion we're done. */
+	if constexpr (std::is_same_v<T, U>) {
+		return scaled;
+	}
+
+	/* Make sure the value fits resulting type T (clapmping the ends). */
+	return static_cast<T>(Clamp<U>(scaled, std::numeric_limits<T>::min(), std::numeric_limits<T>::max()));
```
PeterN: We already have ClampTo which handles edge cases, is that usable here instead?

src/table/settings/economy_settings.ini
```diff
@@ -315,3 +315,16 @@ strval   = STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE
 pre_cb   = [](auto) { return _game_mode == GM_MENU || _settings_game.economy.timekeeping_units == 1; }
 post_cb  = ChangeMinutesPerYear
 cat      = SC_BASIC
+
+[SDT_VAR]
+var      = economy.payment_time_scale
+type     = SLE_UINT16
+def      = 100
+min      = 0
+max      = 50000
```
PeterN, line 324: 50000%? That's a challenge :)

src/core/math_func.hpp
ldpl: oh, yeah, missed that

src/table/settings/economy_settings.ini
ldpl: Well, why not, also default maglev sometimes peaks at like 10k tiles and smth like vactube peaks who knows where.

src/lang/english.txt
```diff
@@ -1520,6 +1520,9 @@ STR_CONFIG_SETTING_MINUTES_PER_YEAR_VALUE                       :{NUM}
 ###setting-zero-is-special
 STR_CONFIG_SETTING_MINUTES_PER_YEAR_FROZEN                      :0 (calendar time frozen)
 
+STR_CONFIG_SETTING_PAYMENT_TIME_SCALE                           :Time scale for cargo payments: {STRING2}
+STR_CONFIG_SETTING_PAYMENT_TIME_SCALE_HELPTEXT                  :Adjusts the time scale factor (in percentage) in the cargo payment calculations. A lower value means faster deliveries, resulting in higher payments.
```
2TallTyler, line 1524:
```suggestion
STR_CONFIG_SETTING_PAYMENT_TIME_SCALE_HELPTEXT                  :Adjust how quickly cargo loaded on a vehicle decreases in value over time. A higher value means it loses value more quickly, making the game more difficult
```

src/economy.cpp
```diff
@@ -984,6 +984,9 @@ Money GetTransportedGoodsIncome(uint num_pieces, uint dist, uint16_t transit_per
 		return 0;
 	}
 
+	/* Scale transit periods according to the game setting. NewGRFs callbacks operate on scaled value for compatibility. */
```
2TallTyler, line 987: Why mention NewGRFs when they don't seem to be handled any differently? If that is done here, don't mention it. If it's done elsewhere... I can't see it in the changed files. 😉

2TallTyler (commented):
I am wary of referring to time if we don't absolutely have to. I suspect players will be somewhat confused by the new timekeeping mode selection and will wonder if this interacts with it somehow. 

The graph in your screencast doesn't show the time changing (it's still graphed over the same period), but instead how quickly the cargo decays, so I wonder if it could be branded "Cargo aging scale" or something like that.

src/economy.cpp
ldpl: It was supposed to explain why is scaling done before the callback and not after

2TallTyler: Ah, I see now. My mistake. 🙂

2TallTyler:
```suggestion
	/* Scale transit periods according to the game setting. We pass this scaled value to the NewGRF callback. */
```

PeterN: "We also pass this"

src/lang/english.txt
ldpl: While I somewhat understand the desire for this not to be confused with timekeeping mode you can't just monopolize every time-related word :p If anything, what you did is traditionnally referred to as "day length", and the main setting is "minutes per year". And payment is a function of distance and time, so some reference to time is unavoidable here. At the very least it should be searchable by "time" IMO. As "aging" is just a term specific to newgrf development and is not used in the game iterface at all. And in OpenTTD context in general seems to more often refer to vehicle age than cargo age. It's also not technically correct as aging period has inverse relation with transit time. And I would definitely avoid any reference to difficulty here as whilst somewhat true this is not a setting that should be used to make the game "harder" or "easier".

The graph looks are also just a matter of representation. It could be done by having lines stay constant and changing time axis instead.

On the other hand this should also be searchable by "distance". But that relation is via speed so calling it distance scaling would definitely be misleading. And I'd also like to avoid it being called "time scale" directly even though that's exactly what it is. So far the best I came up with is "Transfer time scale" to at least divert some attention from the "time" here.

PeterN:
Needs an update (and some resolutions)

PR #11928, opened 2024-01-30
Fix 48b6b18: Increase MAX_VALUATE_OPS to match the previous limit

glx22:
## Motivation / Problem
48b6b18 introduced an in-VM limit using `MAX_VALUATE_OPS = 500000`.
Before this commit the limit check was doing `if (ScriptController::GetOpsTillSuspend() < -1000000)`, so the limit was between 1000000 and 1000000+allowed_ops_per_tick, depending on when `Valuate()` call happened during the tick.

This change affects some AIs which used to run fine and are now killed due to excessive CPU usage.

## Description
Increase `MAX_VALUATE_OPS` to 1000000, so it's close to the previous limit.

## Checklist for review

### DIFF
--- a/src/script/api/script_list.hpp
+++ b/src/script/api/script_list.hpp
@@ -14,7 +14,7 @@
 #include "script_object.hpp"
 
 /** Maximum number of operations allowed for valuating a list. */
-static const int MAX_VALUATE_OPS = 500000;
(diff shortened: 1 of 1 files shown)


2TallTyler (approved):
I don't know anything about scripts, but matching previous behavior to avoid breaking things seems sensible.

Issue #9742, opened 2021-12-10
[Bug]: "Your PC is too slow to keep up with this server" (2/6 Authorizing)


Azophyte:
### Version of OpenTTD

Steam 12.1

### Expected result

Connects to my game

### Actual result

Gets stuck loading server at 2/6, then gives error saying "Your PC is too slow t o keep up with this server"

### Steps to reproduce

1. Have friend start an OpenTTD server, set to public with no password (Size of world and other settings don't seem to matter)
2. Find server in the server list, try to join.

This only happens on my friend's computer - we have no idea what's wrong. His computer is definitely more powerful than mine, and his internet is better too, so that's not the issue.

I'll update this with a link to my bug report when it gets approved by admins

glx22:
Single core speed is what matters for OpenTTD.
And if possible try to put the server on the less powerful machine.
Enabling "pause on join" may help too.

nielsmh:
It may seem counter-intuitive, but you're generally better off running an OpenTTD server on the *slowest* computer participating.
If the server simulates the world faster than clients are able to (which can happen when the game gets big/complex with lots of vehicles, or on large maps), the server will not wait for the clients that are falling behind, but instead kick them for being too slow.
Clients can wait for the server, but the server can't wait for clients. (If the server was able to wait for clients to catch up, that would be a way to grief games on public servers.)

What you can also try doing is putting the game on pause on the server, while clients are joining the game. That way the server won't be running the simulation while the clients are connecting, and the clients won't need to catch up with the server after downloading the game.

James103:
What CPU do you have and what CPU does your friend have? Please list the exact make and model if possible.

Azophyte:
Oh wow you guys responded fast! The forum post I've made hasn't been approved yet, so I'll post the whole thing here:

----
Hi, can't find any similar posts online relating to this issue - seems to be pretty rare.
Specs:

x64-based PC
Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz, 3601 Mhz, 4 Cores, 8 Logical processors
Installed Physical Memory (RAM): 16GB
His upload and download speeds are both higher than mine, at 85MB download and 265MB upload.

I'm going to be running a game this Sunday for my friends to play OpenTTD on one map, but he can't join for some reason. I've searched a while for a solution but none have worked. I'm hosting on steam.

Here are the things we have tried (that haven't worked):
-Setting affinity of OpenTTD to just one core (taskmanager)
[...]

PR #11931, opened 2024-01-30
Change: make -dnet=9 give traces of the details of the network protocol

TrueBrain:
## Motivation / Problem

In many cases debug logs aren't as useful as I would like, when it is about network issues. Look for example at #9742, the lack of detail makes it very difficult to estimate where abouts things go wrong.

## Description

Introduce `-dnet=9` trace logs, which trace in much more detail what function is being called, and if possible, with what parameters.

I went for all Connecter callbacks and all server/client packets being send/received. We could go much further, but adding these things is a really annoying task, so I am already happy I survived adding these :)

There are two packets not being logged: map-content packets, and server-frames. They are really spammy, and most likely means the important stuff runs out of console really quick. So by removing those, that is avoided. There isn't all that much info in both packets: we see when downloads start and finish, and we do see server-sync frames. That will have to do :)

## Limitations

Creating 100+ debug statements is a very annoying task; hopefully I didn't miss anything important :)

## Checklist for review

### DIFF
--- a/src/network/network.cpp
+++ b/src/network/network.cpp
@@ -633,6 +633,8 @@ public:
 
 	void OnFailure() override
 	{
+		Debug(net, 9, "Query::OnFailure(): connection_string={}", this->connection_string);
+
 		NetworkGameList *item = NetworkGameListAddItem(connection_string);
 		item->status = NGLS_OFFLINE;
 		item->refreshing = false;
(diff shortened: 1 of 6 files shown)


rubidium42 (approved):
It's tough getting through the PR as well.

PR #11933, opened 2024-01-30
Remove macros with unguarded "if"

frosch123:
## Motivation / Problem

Macros starting with `if` are dangerous, since they break if macro usage is followed with `else`.

## Description

* Wrap affected macros into `do {  } while (false)`. This is the standard solution for macros, which should be used with a following `;`
* Even better: replace macros with lambdas.

## Limitations

There is another weird macro left in `ClientNetworkContentSocketHandler::OnReceiveData`, but the whole method should probably get rewritten using `std::string` or CTRE (if we ever get it).

## Checklist for review

### DIFF
--- a/src/rail_cmd.cpp
+++ b/src/rail_cmd.cpp
@@ -2387,7 +2387,9 @@ static void DrawTrackBits(TileInfo *ti, TrackBits track)
 
 static void DrawSignals(TileIndex tile, TrackBits rails, const RailTypeInfo *rti)
 {
-#define MAYBE_DRAW_SIGNAL(x, y, z, t) if (IsSignalPresent(tile, x)) DrawSingleSignal(tile, rti, t, GetSingleSignalState(tile, x), y, z)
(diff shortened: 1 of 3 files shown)


TrueBrain (approved):

src/rail_cmd.cpp
```diff
@@ -2387,7 +2387,9 @@ static void DrawTrackBits(TileInfo *ti, TrackBits track)
 
 static void DrawSignals(TileIndex tile, TrackBits rails, const RailTypeInfo *rti)
 {
-#define MAYBE_DRAW_SIGNAL(x, y, z, t) if (IsSignalPresent(tile, x)) DrawSingleSignal(tile, rti, t, GetSingleSignalState(tile, x), y, z)
+	auto MAYBE_DRAW_SIGNAL = [&](byte signalbit, SignalOffsets image, uint pos, Track track) {
```
LordAro, line 2390: I know it's a bit more of a pain, but should probably get rid of the all caps name

Also it only needs to capture tile & rti, I think?

PR #11934, opened 2024-01-30
Fix #9655: [Emscripten] install the correct files on install

### LINKED ISSUE #9655: [Bug]: Install target for emscripten is incorrect
### Version of OpenTTD

12.0-RC1

### Expected result

As an installaction result I expect to see 4 files gather in same directory:
* *openttd.html*
* *openttd.data*
* *openttd.js*
* *openttd.wasm*

### Actual result

Only *openttd.html* appears in final installation directory (`install/games` dir).

Also *share* directory is created as if would install it on Linux.

### Steps to reproduce

1. Initialize build directory with cmake. Use custom installation path by means of `-DCMAKE_INSTALL_PREFIX=<>`
2. Build openttd.
3. Install it.
4. Observe installation directory.

TrueBrain:
## Motivation / Problem

Fixes #9655.

The reported already made a commit. It was mostly right.

## Description

Emscripten doesn't need datafiles, and can't load scripts. But it does need `wasm` and other files to work. So copy the right files for that target to the same folder as the binaries, so it is all in one tiny lovely place :)

## Checklist for review

### DIFF
--- a/cmake/InstallAndPackage.cmake
+++ b/cmake/InstallAndPackage.cmake
@@ -23,16 +23,28 @@ install(TARGETS openttd
             COMPONENT Runtime
         )
 
-install(DIRECTORY
-                ${CMAKE_BINARY_DIR}/lang
-                ${CMAKE_BINARY_DIR}/baseset
-                ${CMAKE_BINARY_DIR}/ai
-                ${CMAKE_BINARY_DIR}/game
-                ${CMAKE_SOURCE_DIR}/bin/scripts
-        DESTINATION ${DATA_DESTINATION_DIR}
-        COMPONENT language_files
-        REGEX "ai/[^\.]+$" EXCLUDE # Ignore subdirs in ai dir
-)
+if (NOT EMSCRIPTEN)
+    # Emscripten embeds these files in openttd.data.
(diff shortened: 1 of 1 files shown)


cmake/InstallAndPackage.cmake
```diff
@@ -23,16 +23,26 @@ install(TARGETS openttd
...
-        DESTINATION ${DATA_DESTINATION_DIR}
-        COMPONENT language_files
-        REGEX "ai/[^\.]+$" EXCLUDE # Ignore subdirs in ai dir
-)
+if (NOT EMSCRIPTEN)
+    install(DIRECTORY
+                    ${CMAKE_BINARY_DIR}/lang
+                    ${CMAKE_BINARY_DIR}/baseset
+                    ${CMAKE_BINARY_DIR}/ai
+                    ${CMAKE_BINARY_DIR}/game
```
glx22, line 33: It can use AI and GS (from content download), and they might need compatibility scripts.

TrueBrain: No, those files are already embedded in the `openttd.data` file.

glx22 (approved):

PR #11940, opened 2024-01-31
Fix: Display rank correctly with more than 15 companies in a league table

ldpl:
## Motivation / Problem

There can be more than 15 ranks in GameScript league tables so numeration got broken.

## Description

Replace ordinal strings with "#N" allowing arbitrary number of ranks. While there also moved colour into the string like many other strings do. 

Before (13.4)/After:
Screenshot from 2024-01-31 23-29-53

## Checklist for review


2TallTyler (commented):
Why remove the highlighting of the top company's number?

(Not arguing one way or the other, just... Chesterton's Fence. 😉 )

ldpl:
@2TallTyler Well, higlighting was there because that's how it was in the original game. Don't think we should be too concerned with that considering whole background is already different ;)

I removed it as it has no functional value and only adds color clutter. First row is already visible enough simply by the virtue of it being first. And it doesn't work for split first place in the non-GS table anyway. Also allows to move colour into the string like it's done for company name and many other strings.

2TallTyler (approved):
That's a good enough reason for me. 🙂

PR #11944, opened 2024-02-01
Fix: [Script] Apply random deviation only at script start

glx22:
## Motivation / Problem
Random deviation is applied inconsistently:
* It's applied right before game start (in `MakeNewgameSettingsLive()`) so after the user decided the settings for the script.
* It's applied when changing script (`ScriptConfig::Change()`) for a slot during game, but before the user has access to the settings.
* It's applied when loading a savegame (`ScriptConfig::Change()`).
* It's applied when a random AI starts (`ScriptConfig::Change()`).
* It's applied during `start_ai` command (`ScriptConfig::Change()`) but only if a script name is specified and before applying the provided settings, else it uses the current config for the slot without deviation.

## Description
The random deviation is now applied only when a script starts (random or not), so always after user decided the settings for the script. And it's never applied when loading a savegame.

Documentation was not very clear about when the random deviation would be applied, so I fixed it.

## Checklist for review

### DIFF
--- a/src/ai/ai_core.cpp
+++ b/src/ai/ai_core.cpp
@@ -48,6 +48,7 @@
 		/* Load default data and store the name in the settings */
 		config->Change(info->GetName(), -1, false, true);
 	}
+	if (rerandomise_ai) config->AddRandomDeviation();
 	config->AnchorUnchangeableSettings();
 
 	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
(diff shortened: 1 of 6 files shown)


TrueBrain (approved):
Although I don't like that the UI doesn't show all this, at least this is much more consistent than it currently is. And we can fix the UI after this PR :)

SamuXarick:
This broke "restart" functionality.
- Setup an AI that uses random deviation.
- Start game.
- Wait for the AI to start - random deviation is applied.
- type "restart" in console - values of the random deviation remain.
- Wait for the AI to start again - random deviation is applied again, different than the values before.

PR #11947, opened 2024-02-01
Feature: NewGRF town production effect and multiplier

PeterN:
## Motivation / Problem

Typically industry sets define the cargo types used in a game, and house sets set up which cargos are accepted or produced by houses.

House sets can control production by using callbacks, but they have to know which industry/cargo set is in use.

This means that either an industry set and house set have to be very tightly coupled, or (more likely) cargo sets don't pay any attention to towns.

## Description

This PR is built on top of #11378 (all of this PR except the last commit) and adds two NewGRF properties for cargo.

These allow the cargo set to specific a town production behaviour, along with a multiplier to control how much cargo is produced relative to normal production.

This gives cargo sets more flexibility in use without having to redefine every single house, which would break compatibility with every house set...

## Checklist for review

### DIFF
--- a/src/cargotype.h
+++ b/src/cargotype.h
@@ -65,6 +65,8 @@ enum CargoClass {
 
 static const byte INVALID_CARGO_BITNUM = 0xFF; ///< Constant representing invalid cargo
 
+static const uint TOWN_PRODUCTION_DIVISOR = 256;
+
 /** Specification of a cargo type. */
 struct CargoSpec {
 	CargoLabel label;                ///< Unique label of the cargo type.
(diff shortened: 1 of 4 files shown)


src/cargotype.h
```diff
@@ -65,6 +65,8 @@ enum CargoClass {
 
 static const byte INVALID_CARGO_BITNUM = 0xFF; ///< Constant representing invalid cargo
 
+static const uint TOWN_PRODUCTION_DIVISOR = 0x100;
```
2TallTyler: Why is this a hexadecimal?

src/town_cmd.cpp
```diff
@@ -558,7 +558,7 @@ static void TownGenerateCargoOriginal(Town *t, TownProductionEffect tpe, uint8_t
 		uint32_t r = Random();
 		if (GB(r, 0, 8) < rate) {
 			CargoID cid = cs->Index();
-			uint amt = GB(r, 0, 8) / 8 + 1;
+			uint amt = (GB(r, 0, 8) * cs->town_production_multiplier / TOWN_PRODUCTION_DIVISOR) / 8 + 1;
```
frosch123, line 561: Should this use `CeilDiv` instead of rounding down?

frosch123: nevermind, there is already a `+ 1` added.

src/cargotype.h
frosch123: NewGRF often use fixed float precision with 8 fractional bits. 8 bits is easier to write as `0x100`.

PeterN: Habit, as the existing cargo quantity multiplier is also 0x100, although there's no constant for that one.

rubidium42 (approved):
It still needs to be added to the NewGRF tracker.

PeterN:
Tracker is updated.

PR #11950, opened 2024-02-02
Fix: Restore Random AI slot upon stopping a randomly started AI

SamuXarick:
## Motivation / Problem
Slight improvement to "restart" console command to be better at replicating the AI/GS config.

The issue is that when a Random AI starts and becomes an AI, the config slot is modified to the config of the AI it randomized to. Doing a "restart" would then use the config of the AI it randomized to instead or rolling back to Random AI.

## Description

With the help of a customized AI which uses random deviation settings, I checked the behaviours of "newgame", "restart" and savegame load, and I believe restoring the Random AI slot is a step in the right direction to reach reproduceability. It's still not totally there, the other issue is random deviation on non-random AIs, which is not addressed by this PR.

Notes summarised: https://gist.github.com/SamuXarick/5dec2054a0f15122268038f6ce120715 

Additionally, when restarting after a new game, use the seed of the current game and use the new game settings configuration.

## Limitations
Someone would prefer to have the config not revert back to Random AI upon stopping the AI of that slot. It's a compromise.

## Checklist for review

### DIFF
--- a/src/ai/ai_core.cpp
+++ b/src/ai/ai_core.cpp
@@ -115,6 +115,10 @@
 	cur_company.Restore();
 
 	InvalidateWindowClassesData(WC_SCRIPT_DEBUG, -1);
+
+	if (AIConfig::GetConfig(company)->IsRandom()) {
+		AIConfig::GetConfig(company)->Change(std::nullopt);
+	}
 	CloseWindowById(WC_SCRIPT_SETTINGS, company);
 }
 
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -1227,7 +1227,12 @@ DEF_CONSOLE_CMD(ConRestart)
 	/* Don't copy the _newgame pointers to the real pointers, so call SwitchToMode directly */
 	_settings_game.game_creation.map_x = Map::LogX();
 	_settings_game.game_creation.map_y = Map::LogY();
-	_switch_mode = SM_RESTARTGAME;
+	if (_file_to_saveload.abstract_ftype == FT_SAVEGAME || _file_to_saveload.abstract_ftype == FT_SCENARIO || _file_to_saveload.abstract_ftype == FT_HEIGHTMAP) {
+		_switch_mode = SM_RESTARTGAME;
+	} else {
+		_settings_newgame.game_creation.generation_seed = _settings_game.game_creation.generation_seed;
+		_switch_mode = SM_NEWGAME;
+	}
 	return true;
 }
 


src/ai/ai_core.cpp
```diff
@@ -115,6 +115,10 @@
 	cur_company.Restore();
 
 	InvalidateWindowClassesData(WC_SCRIPT_DEBUG, -1);
+
+	if (AIConfig::GetConfig(company)->IsRandom()) {
+		AIConfig::GetConfig(company)->Change(std::nullopt);
+	}
```
TrueBrain, line 121: Although it does solve part of the issue, several other issues with similar situation remains.

In the core of the issue is that restart doesn't revert back the settings to the state they were when starting the game. This is including things like construction settings, loan, etc, but also AI.

So although this would fix that specific issue, it wouldn't deal with the others. We might be better of just reverting the settings on restart, which would also address this issue; but also ensures random deviation is applied the same very restart, as example.

TrueBrain: What might work better, but I did not test this, is change the restart console command a bit.

If _file_to_saveload.abstract_ftype indicates the game was loaded from a savegame, scenario or heightmap, call the current restart code. Otherwise call the newgame command with the seed of the settings.

That most likely gives the best of both worlds. If you press restart from a new game, all AI settings revert. If you do it from a savegame, they go back to how they were in the savegame.

Solves all your restart issues, I think?

SamuXarick: > If _file_to_saveload.abstract_ftype indicates the game was loaded from a savegame, scenario or heightmap, call the current restart code.

It won't, unfortunately. Unless you mean doing that in addition to this PR's changes?

TrueBrain: A bit more detail is needed. It won't what?

SamuXarick: If loading the savegame where the random AI already started, and then typing "restart", the current restart code will not revert the config slot of that company back to Random AI. And in the case of a non random AI which is already started, the random deviation values won't go back to how they were originally.

It however solves this other situation, when the game was started via newgame and a restart is issued:

> In the core of the issue is that restart doesn't revert back the settings to the state they were when starting the game. This is including things like construction settings, loan, etc, but also AI.

TrueBrain: Ah, yes, but that is fully expected. It is even noted with `restart` that doing it on a loaded savegame is expected to not revert it back to how it was when starting a new game.

Basically, `restart` is a developer command, and is mostly used on new games to run them over and over and over and over again, getting a certain result, validating an AI, a change in algorithm, etc. For this it works excellent.

As the comment in `restart` already indicates, doing the same on savegames just doesn't have the same effect, as we do not know the original information. So it is restarted with the information known at that moment in time.

So from my perspective, I absolutely expect a `restart` on a savegame to have "double applied" random deviations etc. It is just something we have to life with, but giving nobody complained in 15+ years about that behaviour (or any of the other quirks with running `restart` on a savegame), I think we are good.

And think about it like this, at least the issue you ran into is solved this way, as from what I understood, you mostly use `restart` on newgames, not?

SamuXarick: The random deviation issue was better managed in 13.4. The very recent change that makes deviation run upon company start is a regression in this regard. It is preferable to have fixed-AIs to restart with the same deviated values they started with.

I use restart quite often, sometimes after new game, sometimes after savegame load.  For AI bug reports (from the forum), a savegame is usually provided. Sometimes "restart" helps reproduce the issues at hand. Sure, Random AIs reproduceability here is not really possible (difference in AIs installed), but it's most often reports with a configured non-random AI, where it's important to have AIs replicate the same play.

glx22: For savegames, use reload.

src/console_cmds.cpp
```diff
@@ -1227,7 +1227,12 @@ DEF_CONSOLE_CMD(ConRestart)
 	/* Don't copy the _newgame pointers to the real pointers, so call SwitchToMode directly */
 	_settings_game.game_creation.map_x = Map::LogX();
 	_settings_game.game_creation.map_y = Map::LogY();
-	_switch_mode = SM_RESTARTGAME;
+	if (_file_to_saveload.abstract_ftype == FT_SAVEGAME || _file_to_saveload.abstract_ftype == FT_SCENARIO || _file_to_saveload.abstract_ftype == FT_HEIGHTMAP) {
```
TrueBrain, line 1230: please use `abstract_ftype != FT_NONE`. More future-proof.

src/console_cmds.cpp
```diff
@@ -1227,7 +1227,12 @@ DEF_CONSOLE_CMD(ConRestart)
...
+		_switch_mode = SM_RESTARTGAME;
+	} else {
+		_settings_newgame.game_creation.generation_seed = _settings_game.game_creation.generation_seed;
+		_switch_mode = SM_NEWGAME;
```
TrueBrain, line 1234: Please use `StartNewGameWithoutGUI(_settings_game.game_creation.generation_seed)`, similar as `newgame` does it.

src/ai/ai_core.cpp
TrueBrain: Please remove this change from the PR-set. It is either one, not both, that are needed to address this issue.

src/console_cmds.cpp
SamuXarick: What about `FT_INVALID` which pop ups sometimes.

TrueBrain: Fair point. Should never make it in here, but better be safe than sorry. So yeah, != NONE && != INVALID sounds good!

src/ai/ai_core.cpp
SamuXarick: No. It wouldn't restore Random AI when doing restart after loading from a savegame with a random AI already started.

It's either both, or just these that are quoted. I'd prefer both, but if you're forcing me to chose :(... I prefer these quoted.

TrueBrain: We talked about this at length in here, and you also talked about this at length on Discord with @glx22. We can't keep having the same conversation over and over, and you hoping all of a sudden the result is different. It doesn't work like that.

When you "restart" a savegame, the expected behaviour is that the Random AI is not being re-randomized. Simply because there isn't enough information available to be correct or useful in the generic case.
It only works, and ONLY works, in the case the person doing the restart has the exact same AI library as the one that started the game. And if that is the case, starting a new game with the same seed is the correct way forward.

Besides, this change has too many side-effects for it to make sense. So when looking at it from a top view, we see that in 99% of the cases, we cannot recover the AI correctly, and it is better to restart the same AI. We can't tailor to the 1%.

TrueBrain: I had another look at restart/reload, and realised restart was even weirder than I expected. So I made a suggestion (#11962) that is a bit more radical than this approach. And it might also address the usecases you had for the "restart" command.

Very basically, if with my PR you run "restart" on any game, newgame or savegame alike, and your newgame settings have a Random AI, it will get a Random AI again.

Additionally, when you didn't change your settings since the savegame you made, it will reproduce the exact same game, down to the last pixel, when you hit "restart". As such, it might so solve the cases you brought up.

PR #11955, opened 2024-02-02
Better procedural trees

ldpl:
## Motivation / Problem

While #11263 is a solid PR on its own it still left that anoying bit of randomness in the map array and so this PR sets to address that. And while it includes #11263 as a first commit it's designed to be squashed or at least merged together as otherwise second commit would technically require a savegame version bump to zero that damn bit in tree counter.

## Description

To get rid of the counter parity bit tree drawing code needs to know whether the tile was already iterated in the current tile loop 256-tick cycle. To achieve that I had to make tile loop a bit less random, so instead of the whole map permutation it now only randomizes a 64x64 pattern and then repeats it over the map. As a side effect of that in my tests I observed about 20% improvement in the performance of a world tick and that can probably be improved further by rearranging the loops. It likely makes this change worthwhile just for the performance alone but that wasn't my goal here so I didn't explore further.

I ran some tests comparing the savegame size (in  % to master) the best I could (temperate 2k*2k, same seed):

Growth, no spread new game:
PR #11263: 77%
This PR: 74%

Growth, no spread 1 year:
PR #11263: 80%
This PR: 77%

Growth and spread new game:
PR #11263: 79%
This PR: 76%
No trees: 28%

Growth and spread 2 years:
PR #11263: 93%
This PR: 93%

Growth and spread 10 years:
PR #11263: 98%
This PR: 98%

Results aren't particularly scientific since random influences it to like 5% or even more. But still, around 25% improvement for no spread case is nice as it gives servers an option to keep trees with less cost in the save size. Still long ways to go though as without trees it's about 70% less size. And unfortunately when enabling spread advantages of having procedural trees are barely noticeable as this PR only affects fully grown trees. Also worth noting that this PR only lays the groundwork and doesn't fine-tune compression so there may be some easy gains left on the table.

## Limitations

I've renamed _cur_tileloop_tile to _cur_tileloop_seed to better match it's new role but I'm not sure what do with the savegame key. It's still savegame-compatible both ways but changing the key would basically mean removing and re-adding the field breaking that compatibility. 

## Checklist for review

### DIFF
--- a/src/core/math_func.hpp
+++ b/src/core/math_func.hpp
@@ -365,4 +365,22 @@ constexpr uint64_t PowerOfTen(int power)
 
 uint32_t IntSqrt(uint32_t num);
 
+
+/**
+ * Simple 32 bit to 32 bit hash.
+ * From MurmurHash3.
+ * @param h The value to take the hash of.
+ * @return The hash of the value.
+ */
+inline uint32_t SimpleHash32(uint32_t h)
+{
+	h ^= h >> 16;
+	h *= 0x85ebca6b;
+	h ^= h >> 13;
+	h *= 0xc2b2ae35;
+	h ^= h >> 16;
+
+	return h;
+}
+
 #endif /* MATH_FUNC_HPP */
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -759,49 +759,78 @@ std::tuple<CommandCost, Money> CmdClearArea(DoCommandFlag flags, TileIndex tile,
 }
 
 
-TileIndex _cur_tileloop_tile;
+/* Tile loop repeats pattern every MIN_MAP_SIZE tiles. Having it too small might
+ * be noticeable, And too big is a waste of resources. */
+static_assert(MIN_MAP_SIZE == 64);
 
-/**
- * Gradually iterate over all tiles on the map, calling their TileLoopProcs once every TILE_UPDATE_FREQUENCY ticks.
+static const uint TILE_LOOP_PATTERN_SIZE = MIN_MAP_SIZE * MIN_MAP_SIZE; // Size of the pattern
+static uint8_t _cur_tile_loop_pattern[TILE_LOOP_PATTERN_SIZE]; // Tile loop pattern, changes every TILE_UPDATE_FREQUENCY ticks
+
+uint32_t _cur_tileloop_seed;  // Seed value that was used to generate current tile loop pattern
+
+/*
+ * Determines whether the tile loop was already called for the tile in the
+ * current cycle (TILE_UPDATE_FREQUENCY ticks).
+ * @param tile TileIndex of the tile
+ * @returns true iff tile was already iterated
  */
-void RunTileLoop()
+bool WasTileIteratedThisCycle(TileIndex tile)
 {
-	PerformanceAccumulator framerate(PFE_GL_LANDSCAPE);
+	uint pattern_index = (TileY(tile) % MIN_MAP_SIZE) * MIN_MAP_SIZE + TileX(tile) % MIN_MAP_SIZE;
+	return (_cur_tile_loop_pattern[pattern_index] <= TimerGameTick::counter % TILE_UPDATE_FREQUENCY);
+}
 
+/*
+ * Generates a pattern to use for the tile loop seeded on the value of _cur_tileloop_seed
+ */
+void GenerateTileLoopPattern() {
 	/* The pseudorandom sequence of tiles is generated using a Galois linear feedback
 	 * shift register (LFSR). This allows a deterministic pseudorandom ordering, but
 	 * still with minimal state and fast iteration. */
 
-	/* Maximal length LFSR feedback terms, from 12-bit (for 64x64 maps) to 24-bit (for 4096x4096 maps).
-	 * Extracted from http://www.ece.cmu.edu/~koopman/lfsr/ */
-	static const uint32_t feedbacks[] = {
-		0xD8F, 0x1296, 0x2496, 0x4357, 0x8679, 0x1030E, 0x206CD, 0x403FE, 0x807B8, 0x1004B2, 0x2006A8, 0x4004B2, 0x800B87
-	};
-	static_assert(lengthof(feedbacks) == 2 * MAX_MAP_SIZE_BITS - 2 * MIN_MAP_SIZE_BITS + 1);
-	const uint32_t feedback = feedbacks[Map::LogX() + Map::LogY() - 2 * MIN_MAP_SIZE_BITS];
+	static int16_t lfsr_current = _cur_tileloop_seed % TILE_LOOP_PATTERN_SIZE;
+	const uint32_t LFSR_FEEDBACK = 0xD8F;  // 12-bit feedback term extracted from http://www.ece.cmu.edu/~koopman/lfsr/
+	static_assert(TILE_LOOP_PATTERN_SIZE == 4096);  // If we ever change pattern size, we need a different feedback term.
 
-	/* We update every tile every TILE_UPDATE_FREQUENCY ticks, so divide the map size by 2^TILE_UPDATE_FREQUENCY_LOG = TILE_UPDATE_FREQUENCY */
-	static_assert(2 * MIN_MAP_SIZE_BITS >= TILE_UPDATE_FREQUENCY_LOG);
-	uint count = 1 << (Map::LogX() + Map::LogY() - TILE_UPDATE_FREQUENCY_LOG);
+	_cur_tile_loop_pattern[0] = 0; // LFSR doesn't iterate 0 so add it separately
+	for (uint i = 1; i < TILE_LOOP_PATTERN_SIZE; i++) {
+		lfsr_current = (lfsr_current >> 1) ^ (-(lfsr_current & 1) & LFSR_FEEDBACK);
+		_cur_tile_loop_pattern[lfsr_current] = i / (TILE_LOOP_PATTERN_SIZE >> TILE_UPDATE_FREQUENCY_LOG);
+	}
+}
 
-	TileIndex tile = _cur_tileloop_tile;
-	/* The LFSR cannot have a zeroed state. */
-	assert(tile != 0);
+void RunTileLoop()
+{
+	PerformanceAccumulator framerate(PFE_GL_LANDSCAPE);
 
-	/* Manually update tile 0 every TILE_UPDATE_FREQUENCY ticks - the LFSR never iterates over it itself.  */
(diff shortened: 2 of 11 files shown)


src/core/math_func.hpp
```diff
@@ -352,4 +352,20 @@ constexpr int RoundDivSU(int a, uint b)
 
 uint32_t IntSqrt(uint32_t num);
 
+
+/**
+ * Simple 32 bit to 32 bit hash.
+ * From MurmurHash3.
+ */
+inline uint32_t SimpleHash32(uint32_t h)
```
2TallTyler, line 375: Needs doxygen for parameter and return value.

src/landscape.cpp
```diff
@@ -751,48 +751,78 @@ std::tuple<CommandCost, Money> CmdClearArea(DoCommandFlag flags, TileIndex tile,
...
-	/* Maximal length LFSR feedback terms, from 12-bit (for 64x64 maps) to 24-bit (for 4096x4096 maps).
-	 * Extracted from http://www.ece.cmu.edu/~koopman/lfsr/ */
-	static const uint32_t feedbacks[] = {
-		0xD8F, 0x1296, 0x2496, 0x4357, 0x8679, 0x1030E, 0x206CD, 0x403FE, 0x807B8, 0x1004B2, 0x2006A8, 0x4004B2, 0x800B87
-	};
-	static_assert(lengthof(feedbacks) == 2 * MAX_MAP_SIZE_BITS - 2 * MIN_MAP_SIZE_BITS + 1);
-	const uint32_t feedback = feedbacks[Map::LogX() + Map::LogY() - 2 * MIN_MAP_SIZE_BITS];
+	static int16_t lfsr_current = _cur_tileloop_seed % TILE_LOOP_PATTERN_SIZE;
+	const uint32_t LFSR_FEEDBACK = 0xD8F;  // 12-bit feedback term extracted from http://www.ece.cmu.edu/~koopman/lfsr/
+	static_assert(TILE_LOOP_PATTERN_SIZE == 4096);  // different size requires different feedback
```
2TallTyler: Unclear comment, is this what you mean?
```suggestion
	static_assert(TILE_LOOP_PATTERN_SIZE == 4096);  // If we ever allow larger maps, we need a different feedback term.
```

src/landscape.cpp
```diff
@@ -751,48 +751,78 @@ std::tuple<CommandCost, Money> CmdClearArea(DoCommandFlag flags, TileIndex tile,
...
 
-	/* We update every tile every 256 ticks, so divide the map size by 2^8 = 256 */
-	uint count = 1 << (Map::LogX() + Map::LogY() - 8);
+	_cur_tile_loop_pattern[0] = 0; // LFSR doesn't iterate 0 so add it separately
+	for (uint i = 1; i < TILE_LOOP_PATTERN_SIZE; i++) {
+		lfsr_current = (lfsr_current >> 1) ^ (-(lfsr_current & 1) & LFSR_FEEDBACK);
+		_cur_tile_loop_pattern[lfsr_current] = i / (TILE_LOOP_PATTERN_SIZE / 256);
```
2TallTyler: Is there a constant for this 256? Maybe a `constexpr` calculation using constants, if we don't already have a suitable constant?

src/landscape.cpp
```diff
@@ -751,48 +751,78 @@ std::tuple<CommandCost, Money> CmdClearArea(DoCommandFlag flags, TileIndex tile,
...
-	TileIndex tile = _cur_tileloop_tile;
-	/* The LFSR cannot have a zeroed state. */
-	assert(tile != 0);
+void RunTileLoop()
+{
+	PerformanceAccumulator framerate(PFE_GL_LANDSCAPE);
 
-	/* Manually update tile 0 every 256 ticks - the LFSR never iterates over it itself.  */
+	/* Every 256 ticks generate a new tile pattern just to make it a bit more random. */
 	if (TimerGameTick::counter % 256 == 0) {
```
2TallTyler: Maybe this could be the same const as I mentioned above?

src/landscape.cpp
```diff
@@ -751,48 +751,78 @@ std::tuple<CommandCost, Money> CmdClearArea(DoCommandFlag flags, TileIndex tile,
...
+	assert(Map::SizeY() % MIN_MAP_SIZE == 0);
 
-		/* Get the next tile in sequence using a Galois LFSR. */
-		tile = (tile.base() >> 1) ^ (-(int32_t)(tile.base() & 1) & feedback);
-	}
+	/* We update every tile every 256 ticks, */
+	for (uint i = 0; i < TILE_LOOP_PATTERN_SIZE; i++) {
+
+		/* Find tiles in a pattern that need to be updated on a current iteration, */
+		if (_cur_tile_loop_pattern[i] != TimerGameTick::counter % 256) continue;
```
2TallTyler: Same const? 😃

src/tree_cmd.cpp
```diff
@@ -134,24 +140,46 @@ static void PlantTreesOnTile(TileIndex tile, TreeType treetype, uint count, uint
...
+ * @param tile The tile to get a TreeType from
+ * @param tile_hash The hash of this tile, used as source of randomness
+ * @return The tree type
+ */
+static TreeType GetProceduralTreeType(TileIndex tile, uint32_t tile_hash)
+{
+	TreeType type = GetTreeType(tile);
+	switch (type) {
+		case TREE_RANDOM_TEMPERATE:
+			return (TreeType)(tile_hash % TREE_COUNT_TEMPERATE + TREE_TEMPERATE);
```
2TallTyler: We are moving to `static_cast` instead of c-style casts.

src/tree_cmd.cpp
```diff
@@ -515,6 +543,34 @@ struct TreeListEnt : PalSpriteID {
 	byte x, y;
 };
 
+/**
+ * Determines whethere new trees can be planted on the tile or around it according to game settings.
+ * @param tile The tile to check.
+ * @return True iff ne trees can be planted.
```
```suggestion
 * @return True iff new trees can be planted.
```

src/tree_cmd.cpp
```diff
@@ -134,24 +140,46 @@ static void PlantTreesOnTile(TileIndex tile, TreeType treetype, uint count, uint
...
+		case TREE_RANDOM_TROPIC_RAINFOREST:
+			return (TreeType)(tile_hash % TREE_COUNT_RAINFOREST + TREE_RAINFOREST);
+		case TREE_RANDOM_TOYLAND:
+			return (TreeType)(tile_hash % TREE_COUNT_TOYLAND + TREE_TOYLAND);
 		default:
-			return (TreeType)(seed * TREE_COUNT_TOYLAND / 256 + TREE_TOYLAND);
+			return type;
 	}
 }
-
```
2TallTyler, line 154: I think you still want this newline between functions. 😉

src/tree_cmd.cpp
```diff
@@ -515,6 +543,34 @@ struct TreeListEnt : PalSpriteID {
 	byte x, y;
 };
 
+/**
+ * Determines whethere new trees can be planted on the tile or around it according to game settings.
```
```suggestion
 * Determines whether new trees can be planted on the tile or around it according to game settings.
```

src/tree_cmd.cpp
```diff
@@ -549,9 +608,28 @@ static void DrawTile_Trees(TileInfo *ti)
...
+
+		/* Use a fixed value for the counter (0) when trees aren't growing. */
+		if (_settings_game.construction.extra_tree_placement != ETP_NO_GROWTH_NO_SPREAD) {
+			uint64_t cycle = tile_hash + (TimerGameTick::counter >> 8);
+			counter = cycle >> 4;
+
+			/* It may happen that counter has not increased yet but the tile was already
+			 * iterated with the new value so we need to adjust.
+			 */
+			if ((cycle & 15) == 15 && WasTileIteratedThisCycle(ti->tile)) counter++;
```
2TallTyler, line 622: What's the 15? Could it be a named constant? (even if just locally)

src/tree_cmd.cpp
```diff
@@ -726,86 +798,121 @@ static void TileLoop_Trees(TileIndex tile)
...
+	}
 
-		default:
-			AddTreeGrowth(tile, 1);
-			break;
+	if (spread) {
+		TreeType treetype = GetProceduralTreeType(tile, tile_hash);
+
+		/* We never spread while changing the current tile so it's ok to overwrite and only mark the new tile dirty later. */
+		tile += TileOffsByDir((Direction)(Random() % DIR_END));
```
2TallTyler, line 906:
```suggestion
		tile += TileOffsByDir(static_cast<Direction>(Random() % DIR_END));
```

src/tree_cmd.cpp
```diff
@@ -726,86 +798,121 @@ static void TileLoop_Trees(TileIndex tile)
...
-						}
-						break;
+			case 6: // final stage of tree destruction
+				if (!can_plant_extra) {
+					/* if trees can't spread just plant a new one to prevent deforestation */
+					SetTreeGrowth(tile, 0);
+				} else if (count > 1) {
+					/* more than one tree, delete it */
+					AddTreeCount(tile, -1);
+					SetTreeGrowth(tile, 3);
```
2TallTyler: In https://github.com/OpenTTD/OpenTTD/pull/11263#discussion_r1434663842 we discussed growth stages and how a lot of these magic numbers would be an enum like `GROWING_1`, `GROWING_2`, etc. I actually think that's exactly what we need here. 🙂

2TallTyler (commented):
I never reviewed this because I thought it was an add-on to #11263, and that was waiting on review comments to be processed. Now I see it's an alternative and was ready for review this whole time. Sorry about that. 😅 

I think it makes sense to do this all in one go, instead of doing #11263 first and having a weird in-process state.

In general, this looks conceptually sound to me, but needs a lot of magic numbers replaced with constants. Most of the mess is pre-existing and not your work, but I have no chance of understanding the original code much less your modifications without some clean-up. 😛

src/tree_cmd.cpp
nielsmh: To me this immediately pops out as "test whether all four lower bits are set", maybe it would be easier to recognize with the constants written in hex or binary?
`(cycle & 0b1111) == 0b1111`

michicc:
Does this (or the other PR) leave any option for NewGRF trees?

ldpl:
This makes it a bit trickier for implementing newgrf trees but nothing too crazy. They just need to be aware of procedural trees and provide custom random groups instead of hardcoded ones. In fact, this likely needs to be added before any newgrf trees for any hope of procedural trees to ever exist. As newgrf trees that are developed without regard to procedural trees are likely to make it impossible to implement later.

PR #11956, opened 2024-02-02
Codechange: only copy those parameters actually used with CopyOutDParam()

TrueBrain:
## Motivation / Problem

Alternative to #11952.
Closes #11952.

ErrorWindow, as example, copies all 20 global parameters to an internal buffer. The issue is, there can be references on that buffer that are no longer valid, but also no longer used.
This can cause out-of-bound reading.

## Description

Of all the solutions possible, I went with the most painful one. Instead of using the global parameters, force all such functions that copy, to use a local set.

## Limitations

Although I did my best to validate whether strings needed params, and found where they were, it is very likely I made a mistake. This PR will cause bug-reports.

## Checklist for review


TrueBrain:
Although I am rather happy with the result of this PR, it is also very likely this causes a bug or two. We hide our SetDParams from time to time .. I went to most (all?) strings, and check if they needed a parameter. Not sure I discovered them all.

For 14.0 I suggest we do not merge this, but the other PR. Maybe revert that after, and merge this. But ... yeah ... big PR, high chance of breaking things.

PR #11957, opened 2024-02-02
Codechange: use std managed pointers in saveload

rubidium42:
## Motivation / Problem

First of all replacing manual memory management with automatic management.
Secondly there are at least three places where special actions are taken to prevent double frees.

## Description

Replace the saveload filter pointers with `std::shared_ptr`. This due to some interactions with the networking code, for example the server wanting to retain the savegame after it's done saving. So, since network and saveload can run async from each other having shared custody over the memory seems best.

Replace two other pointers in `SaveloadParams` with `std::unique_ptr`.

## Limitations

There are still `CallocT`-ed bits of memory in the saveload code, but I'll consider that out of scope for now. 

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -88,14 +88,14 @@ struct ReadBuffer {
 	byte buf[MEMORY_CHUNK_SIZE]; ///< Buffer we're going to read from.
 	byte *bufp;                  ///< Location we're at reading the buffer.
 	byte *bufe;                  ///< End of the buffer we can read from.
-	LoadFilter *reader;          ///< The filter used to actually read.
+	std::shared_ptr<LoadFilter> reader; ///< The filter used to actually read.
 	size_t read;                 ///< The amount of read bytes so far from the filter.
 
 	/**
 	 * Initialise our variables.
 	 * @param reader The filter to actually read data.
 	 */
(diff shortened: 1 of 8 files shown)


src/saveload/saveload.cpp
```diff
@@ -199,10 +199,10 @@ struct SaveLoadParams {
 	bool expect_table_header;            ///< In the case of a table, if the header is saved/loaded.
 
 	MemoryDumper *dumper;                ///< Memory dumper to write the savegame to.
-	SaveFilter *sf;                      ///< Filter to write the savegame to.
+	std::shared_ptr<SaveFilter> sf; ///< Filter to write the savegame to.
 
 	ReadBuffer *reader;                  ///< Savegame reading buffer.
-	LoadFilter *lf;                      ///< Filter to read the savegame from.
+	std::shared_ptr<LoadFilter> lf; ///< Filter to read the savegame from.
```
michicc: Comment alignment?

rubidium42: Keeping these trailing comments aligned over time is quite a hassle, so some of us have stopped bothering with it.
Essentially some of us have just been removing the spaces when we were manually changing the lines, without touching the other lines.

michicc (approved):

Issue #11943, opened 2024-01-31
[Bug]: when AI settings are applied is unclear to the user


TrueBrain:
### Version of OpenTTD

ccaa383e85c16786fea3567e46559840dfb450d0

### Expected result

A clear system where I can see what settings will be used in-game

### Actual result

It is a bit of a story, but let me give a short summary:

- When you select an AI from the main menu, you can configure it.
- This configuration uses the default values for the selected `settings_profile` (easy / medium / hard).
- Internally there is a `custom` category, but you cannot select this (anymore)

Now there are three things going on:

- when you do not make any changes to the settings, the value of the settings will change if you update `settings_profile` to the value the AI assigned to easy / medium / hard
- when you do make a change, it no longer updates
- when you start the game, `random_deviation` is applied, possibly changing the value.

But all this information is poorly / not communicated to the user. More specifically:

1) You cannot see when you are not using the default value.
2) You cannot reset to the default value.
3) You are not informed any `random_deviation` might be applied to the setting. So once in-game you can find the AI having another setting than you configured.

Personally, I actually like how the interaction works; I just don't like how it is communicated with the user.

### Steps to reproduce

See above.

TrueBrain:
A suggestion for the `random_deviation` might be to add an icon (like a `(i)`) at the end of the setting, saying: on the start of the game this value will be get a random deviation.

Additionally to that, it might be good to just completely remove `settings_profile`, as it is very hidden, and the interaction is unclear.

And it might be good to have a "return to default" option, like the settings have.

TrueBrain:
#11958 addresses the 3rd point.

TrueBrain:
With the removal of `settings_profile`, I am pretty much okay with how it is now. Tnx to @glx22 for the help etc :) Closing this up!

PR #11959, opened 2024-02-02
Change: [Script] replace easy/medium/hard values with default value

TrueBrain:
## Motivation / Problem

As mentioned in #11943, the `settings_profile` is a rather hidden setting. And its interaction with AI Configuration is vague at best (see ticket for more details).

Difficulty was removed from OpenTTD years ago, but it was left behind for AIs. Just a bit hidden.

Time to make the next step, and remove the difficulty completely from the game.

## Description

Remove `settings_profile` from game settings and `easy_value` / `medium_value` / `hard_value` / `custom_value` from `AddSetting`. Add `default_value` in `AddSetting`. Together with `random_deviation` it should still allow for variation. Just not based on difficulty.

As most AI authors do actually use easy/medium/hard, I didn't want to just waste the information and pick one of the three. So instead, medium is picked as median, and easy/hard are converted into `random_deviation`. That way, games are always somewhere between easy and hard. Best we can do, I would say.

`custom_value` was never actually used. It was impossible to set `settings_profile` to custom, so it was just being there. And not being useful.

An alternative to this PR could be to move the `settings_profile` into the AI configuration window. But given we removed it from OpenTTD, I think this approach is better.

## Limitations

Did basic testing, but nothing in-depth.

## Checklist for review

### DIFF
--- a/src/script/script_info.cpp
+++ b/src/script/script_info.cpp
@@ -87,6 +87,10 @@ SQInteger ScriptInfo::AddSetting(HSQUIRRELVM vm)
 	ScriptConfigItem config;
 	uint items = 0;
 
+	int easy_value = INT32_MIN;
+	int medium_value = INT32_MIN;
+	int hard_value = INT32_MIN;
+
 	/* Read the table, and find all properties we care about */
 	sq_pushnull(vm);
 	while (SQ_SUCCEEDED(sq_next(vm, -2))) {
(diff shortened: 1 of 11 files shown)


src/script/script_info.cpp
```diff
@@ -162,6 +168,28 @@ SQInteger ScriptInfo::AddSetting(HSQUIRRELVM vm)
 	}
 	sq_pop(vm, 1);
 
+	/* Check if default_value is set. Although required, this was changed with
+	 * 15.0, and as such, older AIs don't use it yet. So we convert the older
```
rubidium42: Is this really intended to be merged after 14 has been branched?

TrueBrain: lolz; nice catch :D

rubidium42 (approved):

PR #11962, opened 2024-02-03
Change: "restart" now uses your newgame settings, where "reload" uses the current settings

TrueBrain:
## Motivation / Problem

After reading #11950, and looking over our options, I realised there is just something weird about our "reload" and "restart" commands. They behave differently depending on what is loaded (newgame, savegame, ..), and in a way that is hard to explain to users.

So I thought, maybe we should be a bit more extreme than what I suggested in #11950, and just draw a clear line what settings become active with "reload", and what settings with "restart". Also why I made a new PR, I didn't want to do it to the author of #11950 to get blamed for being a bit more drastic, as I am with this PR.

The issue that brought this to life is that when an Random AI is assigned what to run exactly, "restart" was not reverting this, but kept that AI assigned. Although #11950 set out to fix that, the problem is much bigger: any setting changed since the start was not reverted, which means that a "restart" could result in completely different games. And this was both against how "restart" was documentated and also not the functionality most of us envision with "restart".
An other example that came to light: if you change NewGRFs during a game (you shouldn't!), a "restart" used the new NewGRFs, giving you a totally different experience. And there are many more examples where this just acts weird / unexpected / unpredictable.

Closes #11950

## Description

"restart" now uses the seed of the current map and your newgame settings to rebuild the map.
- For newgames this now means that you start the exact (identical, in every way) same game, as people expect. If it doesn't, I would consider that a bug to be solved.
- For savegames it means it will create the same map IF you have the exact same settings. Which, for AI and NewGRF testing is not unlikely (and these commands aim those kind of players). But for any game you receive from someone else, you might not be so lucky.

"reload" stays the same but becomes easier to explain:
- For newgames this restarts the map with the current settings. So if you made any changes, if a Random AI was being assigned, etc, those all stay exactly the same. In result, this might produce a different map all together (if you changed world generation settings, for example).
- For savegames it simply reloads the game.

Hopefully this makes it a bit easier to understand for users what command does what.

## Limitations

There might be spacebar heaters.

There now isn't a way to restart a savegame with the current settings of the game. But this also gives very surprising results, and what caused a bit of the back & forth about what is actually going on. As we don't store the newgame settings as they were at the start of the map, it doesn't reproduce the state of the start of the savegame, but some interpretation of that. So I think we are better off just not supporting that.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -1217,17 +1217,13 @@ DEF_CONSOLE_CMD(ConRestart)
 {
 	if (argc == 0) {
 		IConsolePrint(CC_HELP, "Restart game. Usage: 'restart'.");
-		IConsolePrint(CC_HELP, "Restarts a game. It tries to reproduce the exact same map as the game started with.");
-		IConsolePrint(CC_HELP, "However:");
-		IConsolePrint(CC_HELP, " * restarting games started in another version might create another map due to difference in map generation.");
-		IConsolePrint(CC_HELP, " * restarting games based on scenarios, loaded games or heightmaps will start a new game based on the settings stored in the scenario/savegame.");
+		IConsolePrint(CC_HELP, "Restarts a game, using the newgame settings.");
+		IConsolePrint(CC_HELP, " * if you started from a new game, and your newgame settings haven't changed, the game will be identical to when you started it.");
+		IConsolePrint(CC_HELP, " * if you started from a savegame / scenario / heightmap, the game might be different, because your settings might differ.");
 		return true;
 	}
 
-	/* Don't copy the _newgame pointers to the real pointers, so call SwitchToMode directly */
-	_settings_game.game_creation.map_x = Map::LogX();
-	_settings_game.game_creation.map_y = Map::LogY();
-	_switch_mode = SM_RESTARTGAME;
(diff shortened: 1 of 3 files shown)


src/console_cmds.cpp
```diff
@@ -1217,31 +1217,27 @@ DEF_CONSOLE_CMD(ConRestart)
 {
 	if (argc == 0) {
 		IConsolePrint(CC_HELP, "Restart game. Usage: 'restart'.");
-		IConsolePrint(CC_HELP, "Restarts a game. It tries to reproduce the exact same map as the game started with.");
-		IConsolePrint(CC_HELP, "However:");
-		IConsolePrint(CC_HELP, " * restarting games started in another version might create another map due to difference in map generation.");
-		IConsolePrint(CC_HELP, " * restarting games based on scenarios, loaded games or heightmaps will start a new game based on the settings stored in the scenario/savegame.");
+		IConsolePrint(CC_HELP, "Restarts a game, using the newgame settings");
+		IConsolePrint(CC_HELP, " * if you started from a new game, the game will be identical to when you started it.");
```
rubidium42: Pedantic: unless you used `setting_newgame` :(

rubidium42 (approved):

PR #11965, opened 2024-02-03
Codechange: use C++ idioms to manage (network) packets

rubidium42:
## Motivation / Problem

There is a custom linked list, with custom methods to clear the memory.
Manual memory management.

Based on the function's signature it is not clear whether you are handing over ownership of the `Packet`, or not.

## Description

Use a `std::deque` instead of the custom linked list.
Use `std::unique_ptr<Packet>` for all functions/locations where ownership of the `Packet` is transferred.
Use `Packet &` anywhere where the ownership of the `Packet` is not transferred.

In the server code there are a few functions that check whether to ask for the game/company password. Instead of duplicating the code, I changed the respective functions to call the new `SendXXX` in the chain when they do not need to ask for the password. So now practically the send XXX and receive XXX functions need to know about the next step in the process, instead of multiple receive XXX functions having to know the next step in the process (`SendWelcome` was known by three receive XXX functions).

## Checklist for review


michicc:
Bikeshedding here: To be really explicit, one could use the `std::queue` container adapter here.

rubidium42:
> Bikeshedding here: To be really explicit, one could use the `std::queue` container adapter here. It's even defaulting to `dequeue` for storage.

I could do that for half of the cases. The one in the save game writing is pushing the final size of the compressed save game in front of all the still to be sent data chunks.
For the actual queue in the socket handler there's no such requirement yet, so I could change it there.

LordAro:
Perhaps a comment with the declaration explaining why std::queue can't be used?

LordAro (approved):
Seems like the sort of change where the compiler would flag up any issues

PR #11973, opened 2024-02-03
Fix #11968: Always set string parameters to avoid undefined behavior

### LINKED ISSUE #11968: [Crash]: Crash when opening orders of new vehicle
### Version of OpenTTD

OpenTTD 14.0-beta1, Windows 10

### Steps to reproduce

1. Bought a new road vehicle. (Road vehicle 23)
2. I clicked the orders button on the window for the road vehicle.
3. The game crashed.

This also happens when the attached savefile is loaded, and the orders button for road vehicle 23 is pushed.

### Upload crash files

[crash20240203220310.zip](https://github.com/OpenTTD/OpenTTD/files/14154430/crash20240203220310.zip)

2TallTyler:
## Motivation / Problem

#11968 crashes because we break out of setting a string parameter, resulting in undefined behavior.

## Description

Define the behavior by setting the parameter to a safe value.

Closes #11968.

## Checklist for review

### DIFF
--- a/src/order_gui.cpp
+++ b/src/order_gui.cpp
@@ -1166,7 +1166,11 @@ public:
 			case WID_O_DEPOT_ACTION: {
 				VehicleOrderID sel = this->OrderGetSel();
 				const Order *order = this->vehicle->GetOrder(sel);
-				if (order == nullptr || !order->IsType(OT_GOTO_DEPOT)) break;
+				if (order == nullptr || !order->IsType(OT_GOTO_DEPOT)) {
+					/* We can't leave this param unset or the undefined behavior can cause a crash. */
+					SetDParam(0, STR_EMPTY);
+					break;
+				};
 
 				/* Select the current action selected in the dropdown. The flags don't match the dropdown so we can't just use an index. */
 				if (order->GetDepotOrderType() & ODTFB_SERVICE) {


src/order_gui.cpp
```diff
@@ -1166,7 +1166,11 @@ struct OrdersWindow : public Window {
 			case WID_O_DEPOT_ACTION: {
 				VehicleOrderID sel = this->OrderGetSel();
 				const Order *order = this->vehicle->GetOrder(sel);
-				if (order == nullptr || !order->IsType(OT_GOTO_DEPOT)) break;
+				if (order == nullptr || !order->IsType(OT_GOTO_DEPOT)) {
+					/* We can't leave this param empty or the undefined behavior can cause a crash. */
+					SetDParam(0, STR_EMPTY);
```
michicc: Maybe say "We can't leave this param unset", it reads a bit strange to that you may not leave it empty and yet it is set to STR_EMPTY :)

2TallTyler: Indeed. 🙂

glx22:
A failsafe solution could be use `SetDParam()` unconditionally as first thing for the case.
Possibly the same for `WID_O_COND_VALUE` as its STR_JUST_COMMA could receive random garbage too.

rubidium42 (approved):

PR #11979, opened 2024-02-04
Fix: [CMake] Link test executable with log library in Android

Biswa96:
This fixes the following compiler error.
    
    ld.lld: error: undefined symbol: __android_log_write
    referenced by test_main.cpp
    CMakeFiles/openttd_test.dir/src/tests/test_main.cpp.o:
    (Catch::writeToDebugConsole(std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>> const&))

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,6 +288,10 @@ target_link_libraries(openttd
 )
 
 target_link_libraries(openttd_test PRIVATE openttd_lib)
+if(ANDROID)
+    target_link_libraries(openttd_test PRIVATE log)
+endif()
+
 include(Catch)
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):

TrueBrain:
Do know we don't officially support OpenTTD; there is effort by the community to make a usable Android port, with the latest source at https://github.com/n-ice-community/openttd-android .

Biswa96:
It is possible to use openttd with termux, see https://github.com/termux/termux-packages/tree/master/x11-packages/openttd

It's not perfect in the termux environment but it works and usable.

PR #11817, opened 2024-01-17
Fix #11802: Don't set edge traversability bit on aqueducts crossing that same edge side

SamuXarick:
## Motivation / Problem
#11802 - Neighbouring regions on the same x_or_y may indeed have the traversability bits sets, but that doesn't necessarily mean ships can traverse their edges.

## Description
Actually check if water regions are reachable.

Don't set edge traversability bit on aqueducts crossing that same edge side.

Closes #11802.

## Limitations

Almost same performance. What used to be 3000 us on a 4096x4096 map is now 3100 us, even without aqueducts.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, game_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #11986, opened 2024-02-04
Fix #11802: Made determining water region edge traversability more robust

### LINKED ISSUE #11802: [Bug]: Incorrect water region connectivity check with aqueduct on an edge
### Version of OpenTTD

20240116-master-g2efd897691

### Expected result

No connectivity!

Something like this should do the check the right way:
https://github.com/OpenTTD/OpenTTD/blob/2efd8976914d0a669656143a2cd20426df0b9901/src/script/api/script_marine.cpp#L59-L77

### Actual result

Captura de ecrã 2024-01-16 173937

### Steps to reproduce

https://github.com/OpenTTD/OpenTTD/blob/2efd8976914d0a669656143a2cd20426df0b9901/src/pathfinder/water_regions.cpp#L294-L297

Kuhnovic:
## Motivation / Problem

Determining water region edge traversability (i.e. whether you can move from one region into an adjacent one) wasn't done in the most robust way. At least one edge case was already found and a fix is in this PR https://github.com/OpenTTD/OpenTTD/pull/11817 . While the fix is good, there might be other edge cases that we don't know of. I wanted to come up with a more robust method.

## Description

I've found that the best way to fix these kinds of issues are to "play by the same rules as ships". This means using the Track Follower for ships. This trick was already used for the connected component labeling (CCL) inside the water regions, but not yet to determine the inter-region traversability.

After this change the CCL is allowed to "hop over the edges" of the water region. When it does it sets the traversability of that particular tile. This actually made cross-region-aqueduct-detection easier too.

## Limitations

- This introduces a dependency on the water tiles of the edge of any adjacent water region. If any of those external edge tiles get changed it could affect the water region traversability. The InvalidateWaterRegion function has been modified to take this into account and invalidates any adjacent water regions in such a case.
- This stuff is hard to inspect / debug, so I added some debug printing. The land info tool calls it. This is the printout for the region marked in white (it includes the void edge tiles):
image

## Checklist for review

### DIFF
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -29,6 +29,7 @@
 #include "rev.h"
 #include "timer/timer.h"
 #include "timer/timer_window.h"
+#include "pathfinder/water_regions.h"
 
 #include "widgets/misc_widget.h"
 
@@ -128,6 +129,8 @@ public:
 		Debug(misc, LANDINFOD_LEVEL, "m6     = 0x{:x}", tile.m6());
 		Debug(misc, LANDINFOD_LEVEL, "m7     = 0x{:x}", tile.m7());
 		Debug(misc, LANDINFOD_LEVEL, "m8     = 0x{:x}", tile.m8());
+
+		PrintWaterRegionDebugInfo(tile);
 #undef LANDINFOD_LEVEL
 	}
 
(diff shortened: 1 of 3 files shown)


TrueBrain:
Mind splitting the fix from the debugging stuff, in 2 commits or something? Makes `git blame` in the future easier to digest :)

LordAro (approved):

PR #11993, opened 2024-02-04
Feature: Support more number formats than just the 'Western' one

### LINKED ISSUE #11953: [Bug(?)]: Regional Currency/Number Format (Localization)
### Version of OpenTTD

Master

### Expected result

PR #11921 changed the currency format for {CURRENCY_SHORT}, allowing translators to translate the suffexes into their language.  The code works for SI and most other formats, but in CJK areas there is another commonly used format that doesn't even follow (?) the SI three-digit format, but is a four-digit format. Although I could translate the suffexes, according to my own experience, numbers like "10千" (10 k) are very awkward and unintelligible, and only appear in science-related subjects, while financial-related subjects rarely use the SI format. This makes translation of these suffixes virtually impossible.

The SI way separates large numbers like this:
```
000,000,000,000
000.000.000.000
000 b 000 m 000 k 000
00,000 k
```
Separators appear every three digits, and suffixes follow the same pattern, 1k = 1\*10^3, 1m = 1\*10^6, 1b = 1\*10^9, etc.
The format widely used in CJK areas looks like this
```
0000 兆 0000 億 0000 萬 0000
```
image
*A screenshot from A-Tourism for illustration. "9兆7566亿701万日元" means 9,756,607,010,000 Yen*.

There are no separators, instead the suffixes are the separators. 1 兆 = 1\*10^12, 1 億 = 1\*10^8, 1 萬 = 1\*10^4.

However, only about 20% [(survey wk4 jgrpp)](https://survey.openttd.org/summaries/2024/wk04#jgrpp-0.56.2-info.configuration.language.filename) or 45% [(survey wk4 master)](https://survey.openttd.org/summaries/2024/wk04#vanilla-master-info.configuration.language.filename) of all players would use such a format. Most importantly, most players know how to convert numbers automatically, making this implementation almost useless and unnecessary. It could even be misleading to other non-CJK players: 
- What would the English text of the new suffexes be?
- How would the description be?
- How could translators translate the new strings?

Solutions might be:
- Make the suffexes things like 1*10^n;
- Write "not recommended for non-CJK players";
- For non-CJK translators, keep the original format, and for CJK translators, translate to their own language (as the characters used are different, "亿" in zh_CN and "億" in zh_TW and ja_JP, etc)

There might be a way to fix the misleading problem, which would be to make the setting available only for a few languages, i.e. zh_CN, ja_JP, zh_TW, and ko_KR, but that might lead to a bigger problem, which is shoudn't we make all settings in the settings menu availble for all players?

### Actual result

/

### Steps to reproduce

/

rubidium42:
## Motivation / Problem

Fixes #11953. It has been known that the number formats in OpenTTD are not catered for many non-Western languages due to different grouping of the digits.

In the Western countries we group our numbers every three digits. In CJK that's every four digits, and the separators differ for each of the locations. In Hindi it's first three digits and then every two digits. And I probably forgot some other language (group) that does it in an even different manner.

## Description

First a "number format" gets introduced, which is a simple string with twenty zeros (number of digits in uint64_t) where the translators can put characters between at the appropriate place. This allows separators at any location.

Next a "number abbreviation" gets introduced, which defines a mapping between a power and a "number format" for the remaining digits. With these mappings an appropriate version can be chosen for `{CURRENCY_SHORT}`.

The format and abbreviation are added as pragmas/attributes to the language files, so any updates need to be done via a PR and not via eints (at the moment). The format and abbreviation are also configurable in the configuration file, when the user does not like the default choice for the language.

With this change we lose the digit separator and digit separator currency. These two have been the same since their introduction, except if/when someone changed their configuration. It also removes the decimal separator from the saves, so it only remains in the configuration file (moved to the gui section).

English:
Unnamed, 2035-10-13
Hindi:
Unnamed, 2035-10-13#1
CJK:
未命名, 2035-10-13#3

Configuring:
未命名, 2035-10-13#4

## Limitations

The number formats have been set for each of the languages based on their digit group separator. These might not be right for all languages, e.g. Hindi.

The number abbreviations have been set for each of the languages based on the defaults for English, where the `STR_CURRENCY_SHORT_*` values have been replaced. The higher powers are still untranslated.

The highest power will never actually be shown in the UI, since the `{CURRENCY_SHORT}` code is written to show a least two digit groups. That means it needs to know the size of the next group, which is why those are added but never actually used. However, if we ever decide we need a `{CURRENCY_TINY}`, then it would be trivial to use it.

There is a validation of the settings for the format and abbreviations, but it's quite rude as it kills the application with a `UserError`. I have done this as I need to load these settings before the language gets loaded, and using the normal mechanism will cause the error messages to be lost when the UI is finally fully initialised. Given "Survey says" that the current digit grouping and decimal are always empty, i.e. unused, it's not likely to be a huge base of users that are going to change it.

So technically, for most languages the mapping table is not complete. Since there is no support in eints yet, this might cause quite a few bug reports/PRs in the rear future.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -5,8 +5,6 @@
 ##textdir ltr
 ##numberformat 00,000,000,000,000,000,000
 ##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000{NBSP}bn|12=00,000,000{NBSP}tn|15=00,000{NBSP}Qa|18=00{NBSP}Qi
-##digitsep ,
-##digitsepcur ,
 ##decimalsep .
 ##winlangid 0x0809
 ##grflangid 0x01
@@ -5811,11 +5809,6 @@ STR_STATION_NAME                                                :{STATION}
 STR_TOWN_NAME                                                   :{TOWN}
 STR_VEHICLE_NAME                                                :{VEHICLE}
 STR_WAYPOINT_NAME                                               :{WAYPOINT}
-
-STR_CURRENCY_SHORT_KILO                                         :{NBSP}k
-STR_CURRENCY_SHORT_MEGA                                         :{NBSP}m
-STR_CURRENCY_SHORT_GIGA                                         :{NBSP}bn
-STR_CURRENCY_SHORT_TERA                                         :{NBSP}tn
 
 STR_JUST_CARGO                                                  :{CARGO_LONG}
 STR_JUST_RIGHT_ARROW                                            :{RIGHT_ARROW}
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3,6 +3,8 @@
 ##isocode en_GB
 ##plural 0
 ##textdir ltr
+##numberformat 00,000,000,000,000,000,000
+##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000{NBSP}bn|12=00,000,000{NBSP}tn|15=00,000{NBSP}Qa|18=00{NBSP}Qi
 ##digitsep ,
 ##digitsepcur ,
 ##decimalsep .
--- a/src/lang/german.txt
+++ b/src/lang/german.txt
@@ -3,8 +3,6 @@
 ##isocode de_DE
 ##plural 0
 ##textdir ltr
-##digitsep .
-##digitsepcur .
 ##decimalsep ,
 ##winlangid 0x0407
 ##grflangid 0x02
--- a/src/lang/german.txt
+++ b/src/lang/german.txt
@@ -3,6 +3,8 @@
 ##isocode de_DE
 ##plural 0
 ##textdir ltr
+##numberformat 00.000.000.000.000.000.000
+##numberabbreviations 3=00.000.000.000.000.000{NBSP}t|6=00.000.000.000.000{NBSP}Mio|9=00.000.000.000{NBSP}Mrd|12=00.000.000{NBSP}Bio|15=00.000{NBSP}Qa|18=00{NBSP}Qi
 ##decimalsep ,
 ##winlangid 0x0407
 ##grflangid 0x02
--- a/src/lang/german.txt
+++ b/src/lang/german.txt
@@ -4,7 +4,7 @@
 ##plural 0
 ##textdir ltr
 ##numberformat 00.000.000.000.000.000.000
-##numberabbreviations 3=00.000.000.000.000.000{NBSP}t|6=00.000.000.000.000{NBSP}Mio|9=00.000.000.000{NBSP}Mrd|12=00.000.000{NBSP}Bio|15=00.000{NBSP}Qa|18=00{NBSP}Qi
+##numberabbreviations 3=00.000.000.000.000.000{NBSP}Tsd.|6=00.000.000.000.000{NBSP}Mio.|9=00.000.000.000{NBSP}Mrd.|12=00.000.000{NBSP}Bio.|15=00.000{NBSP}Brd.|18=00{NBSP}Trill.
 ##decimalsep ,
 ##winlangid 0x0407
 ##grflangid 0x02
--- a/src/lang/hindi.txt
+++ b/src/lang/hindi.txt
@@ -3,8 +3,6 @@
 ##isocode hi_IN
 ##plural 0
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
 ##decimalsep .
 ##winlangid 0x0439
 ##grflangid 0x17
--- a/src/lang/hindi.txt
+++ b/src/lang/hindi.txt
@@ -3,6 +3,8 @@
 ##isocode hi_IN
 ##plural 0
 ##textdir ltr
+##numberformat 00,000,000,000,000,000,000
+##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000{NBSP}bn|12=00,000,000{NBSP}tn|15=00,000{NBSP}Qa|18=00{NBSP}Qi
 ##decimalsep .
 ##winlangid 0x0439
 ##grflangid 0x17
--- a/src/lang/simplified_chinese.txt
+++ b/src/lang/simplified_chinese.txt
@@ -3,8 +3,6 @@
 ##isocode zh_CN
 ##plural 1
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
 ##decimalsep .
 ##winlangid 0x0804
 ##grflangid 0x56
--- a/src/lang/simplified_chinese.txt
+++ b/src/lang/simplified_chinese.txt
@@ -3,8 +3,8 @@
 ##isocode zh_CN
 ##plural 1
 ##textdir ltr
-##numberformat 00,000,000,000,000,000,000
-##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000b|12=00,000,000t|15=00,000{NBSP}Qa|18=00{NBSP}Qi
+##numberformat 0000京0000兆0000亿0000万0000
+##numberabbreviations 4=0000京0000兆0000亿0000万|8=0000京0000兆0000亿|12=0000京0000兆|16=0000京
 ##decimalsep .
 ##winlangid 0x0804
 ##grflangid 0x56
--- a/src/lang/simplified_chinese.txt
+++ b/src/lang/simplified_chinese.txt
@@ -3,6 +3,8 @@
 ##isocode zh_CN
 ##plural 1
 ##textdir ltr
+##numberformat 00,000,000,000,000,000,000
+##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000b|12=00,000,000t|15=00,000{NBSP}Qa|18=00{NBSP}Qi
 ##decimalsep .
 ##winlangid 0x0804
 ##grflangid 0x56
--- a/src/lang/traditional_chinese.txt
+++ b/src/lang/traditional_chinese.txt
@@ -3,6 +3,8 @@
 ##isocode zh_TW
 ##plural 1
 ##textdir ltr
+##numberformat 00,000,000,000,000,000,000
+##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000{NBSP}bn|12=00,000,000{NBSP}tn|15=00,000{NBSP}Qa|18=00{NBSP}Qi
 ##decimalsep .
 ##winlangid 0x0404
 ##grflangid 0x0c
--- a/src/lang/traditional_chinese.txt
+++ b/src/lang/traditional_chinese.txt
@@ -3,8 +3,6 @@
 ##isocode zh_TW
 ##plural 1
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
 ##decimalsep .
 ##winlangid 0x0404
 ##grflangid 0x0c
--- a/src/lang/traditional_chinese.txt
+++ b/src/lang/traditional_chinese.txt
@@ -3,8 +3,8 @@
 ##isocode zh_TW
 ##plural 1
 ##textdir ltr
-##numberformat 00,000,000,000,000,000,000
-##numberabbreviations 3=00,000,000,000,000,000{NBSP}k|6=00,000,000,000,000{NBSP}m|9=00,000,000,000{NBSP}bn|12=00,000,000{NBSP}tn|15=00,000{NBSP}Qa|18=00{NBSP}Qi
+##numberformat 0000京0000兆0000億0000万0000
+##numberabbreviations 4=0000京0000兆0000億0000万|8=0000京0000兆0000億|12=0000京0000兆|16=0000京
 ##decimalsep .
 ##winlangid 0x0404
 ##grflangid 0x0c
--- a/src/language.h
+++ b/src/language.h
@@ -31,6 +31,10 @@ struct LanguagePackHeader {
 	char isocode[16];   ///< the ISO code for the language (not country code)
 	uint16_t offsets[TEXT_TAB_END]; ///< the offsets
 
+	/** The raw formatting string for numbers. */
+	char number_format[64];
+	/** The raw formatting string for number abbreviations. */
+	char number_abbreviations[256];
 	/** Thousand separator used for anything not currencies */
 	char digit_group_separator[8];
 	/** Thousand separator used for currencies */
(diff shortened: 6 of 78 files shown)


src/main_gui.cpp
```diff
@@ -340,7 +340,7 @@ struct MainWindow : Window
 
 			case GHK_MONEY: // Gimme money
 				/* You can only cheat for money in singleplayer mode. */
-				if (!_networking) Command<CMD_MONEY_CHEAT>::Post(10000000);
+				if (!_networking) Command<CMD_MONEY_CHEAT>::Post(1'000'000'000);
```
TrueBrain: Also increasing the value? :D was that intended?

src/lang/english.txt
```diff
@@ -5806,10 +5806,10 @@ STR_TOWN_NAME                                                   :{TOWN}
 STR_VEHICLE_NAME                                                :{VEHICLE}
 STR_WAYPOINT_NAME                                               :{WAYPOINT}
 
-STR_CURRENCY_SHORT_KILO                                         :{NBSP}k
-STR_CURRENCY_SHORT_MEGA                                         :{NBSP}m
-STR_CURRENCY_SHORT_GIGA                                         :{NBSP}bn
-STR_CURRENCY_SHORT_TERA                                         :{NBSP}tn
+STR_CURRENCY_SHORT_ABBREVIATIONS_PER_POWER_OF_TEN               :3={NBSP}k|6={NBSP}m|9={NBSP}bn|12={NBSP}tn|15={NBSP}Qa|18={NBSP}Qi
```
LordAro: Should this be a language attribute rather than trying to force it into a string?

rubidium42: Is that attribute going to have a significantly different format? It would likely mean parsing the string in strgen, then encoding it somehow and decoding is in OpenTTD. That seems like way more work overall.

Does the translator support changing attributes? If not, who's going to add that, or figure out the separators for all languages?

TrueBrain: We also don't have a way to explain to the translators what this format means. So that is something to consider. Having it as language properties might help, as they have to change that via us. Or we could look at a way to add hints in eints based on a string.

src/lang/simplified_chinese.txt
```diff
@@ -5831,3 +5831,8 @@ STR_PLANE                                                       :{BLACK}{PLANE}
 STR_SHIP                                                        :{BLACK}{SHIP}
 
 STR_TOOLBAR_RAILTYPE_VELOCITY                                   :{STRING} ({VELOCITY})
+
+STR_CURRENCY_SHORT_ABBREVIATIONS_PER_POWER_OF_TEN               :4=萬|8=億|12=兆|16=京
+STR_CURRENCY_NUMBER_FORMAT_TAKES_ONLY_NEEDED_ZEROS_FROM_END     :000京0000兆0000億0000萬0000
+STR_NUMBER_FORMAT_TAKES_ONLY_NEEDED_ZEROS_FROM_END              :000京0000兆0000億0000萬0000
+STR_DECIMAL_NUMBER_FORMAT_TAKES_ONLY_NEEDED_ZEROS_FROM_END      :000京0000兆0000億0000萬0000.0
```
wensimehrp: it's 京-兆-亿-万 in Simplifid Chinese,
and 京-兆-億-萬 in Traditional Chinese
Japanese uses 京-兆-億-万,
maybe we should leave this to the translators

telk5093: Note: Korean uses ``경-조-억-만``

PeterN:
If a parsing a single string is considered too complicated for translators, a string per magntitude might work? Still parsed on language change.

```
STR_CURRENCY_SHORT_MAGNITUDE_1   :
STR_CURRENCY_SHORT_MAGNITUDE_2   :
STR_CURRENCY_SHORT_MAGNITUDE_3   :{NBSP}k
STR_CURRENCY_SHORT_MAGNITUDE_4   :
STR_CURRENCY_SHORT_MAGNITUDE_5   :
STR_CURRENCY_SHORT_MAGNITUDE_6   :{NBSP}m
STR_CURRENCY_SHORT_MAGNITUDE_7   :
STR_CURRENCY_SHORT_MAGNITUDE_8   :
STR_CURRENCY_SHORT_MAGNITUDE_9   :{NBSP}bn
STR_CURRENCY_SHORT_MAGNITUDE_10  :
STR_CURRENCY_SHORT_MAGNITUDE_11  :
STR_CURRENCY_SHORT_MAGNITUDE_12  :{NBSP}tn
```

src/lang/simplified_chinese.txt
rubidium42: > maybe we should leave this to the translators

That was my whole point of this endeavour. However, for testing and "pretty pictures" I needed to "do" some translations myself. It's much easier to change a language in OpenTTD to see the effect, than changing the string in English and restarting OpenTTD to test things work with different configurations.

Before this PR is done all translations that were created for testing purposes will be removed.

src/lang/english.txt
rubidium42: Hints to eints might be a good idea,  even generically for some strings which could use so more information when translating. However, eints is quite far out of my comfort zone and honestly I do not want this feature that bad that I'm going to spend lots of time learning eints and its intricacies.

I think the number formats could fairly easily be attributes, as I doubt there are significantly many more translated languages than the CJK and Hindi. And the separators for those are already mentioned in the comments of this PR.

The abbreviations are way trickier, as those will essentially be different for each language and that means likely over 50 requests for changing the properties.

TrueBrain:
> I see few caveats with this.
>
> (..)

All I read was: "I actually never used eints, (..)", so I volunteered you to a language ;) Not for you to make any actual changes, but so you can click around and see what it actually does :) Might help in creating an opinion and approach for this issue :)

For the rest of us: extrapolating what you see happening with Hindi doesn't define how eints works :) eints is actually pretty good in giving context, and will never go from 5 to 7, skipping 6. That would make a shitty system :)

src/lang/english.txt
TrueBrain: > that I'm going to spend lots of time learning eints and its intricacies.

Please don't. But good news for you: you have a bunch of people around you that can help out! So if we think that is the best way to solve this problem, we can make that happen :) But it first requires knowing that is actually the best way to approach it ;) But please don't feel bad you don't know how to modify eints, or spend your time learning it if you don't want to. It really isn't a problem :)

(I still think attributes is the better solution, so at the moment I just await what form will eventually be chosen by this PR :) But if I turn out to be placing my eggs in the wrong basket, just let us know!)

Edit: to be clear, I meant your suggestion of putting the formats etc in attributes, the things that we can easily do, and keep the rest as translation. We already do the first for decimal separator etc, so why not continue that line :) But don't put in there what needs actual translation, like abbreviations, ofc :)

TrueBrain:
And because I didn't say it here yet: I am really a supporter of this approach; I like that it also fixes one of the last non-translatable parts of the game. So I really appreciate the effort here :) Let's bring it homeeeeeee :D

src/lang/english.txt
frosch123: Eints will preserve `##attributes`, which it does not know (like `##decimalsep`).
No problem in adding more of those.

TrueBrain:
> 
> For the rest of us: extrapolating what you see happening with Hindi doesn't define how eints works :) eints is actually pretty good in giving context, and will never go from 5 to 7, skipping 6. That would make a shitty system :)

I take it back; only in certain situations it is predictable, in others it is random. I will correct that issue soon ;)

rubidium42:
Added the information to pragmas. The first thing that become apparent is that there's no mechanism for fallback, so without setting it you'll just get no separators at all.

However, the whole routine with the abbreviations I've made is not going to fully work because it's not just a suffix. If the number before the suffix is more than 10.000, the next higher one needs to be injected between the 1 and 0. So abbreviating '1亿0000万0000' should become '1亿0000万' and not '1万0000万'. That'll require some more thought on how to solve that relatively efficiently.

rubidium42:
I did some more fiddling.
- The abbreviations now have their own number format, instead of just the trailing abbreviation, so with CJK languages the right characters can be injected. Previously it would use the "normal" format and put the abbreviation after that.
- Removed distinction between normal numbers, numbers with decimals and currency numbers. They now all use the same format. For decimals, there is a separator character that gets added and then the remaining digits.
- strgen will read it using pragmas and validate the format/abbreviation. In OpenTTD itself it will parse it again, so the number format and abbreviations can become user configurable. With the same caveats of the format not being very self-explanatory without nearby documentation.
- the number/decimal format locale settings have been removed from the savegame; the vast majority does not seem to have it set base on the surveys.

rubidium42:
I think this is ready for review now.

The format/abbreviation configuration are attributes/pragmas in the language file, which cannot be translated at the moment via eints. I'm not sure whether support for them will be added, but until then PRs have to do.
They are also settings that the users can change themselves if they know what they need to enter exactly.

@WenSimEHRP, @telk5093: I have updated the format/abbreviation tables for Chinese, Japanese and Korean. Could you check that they are correct?

src/lang/traditional_chinese.txt
```diff
@@ -3,8 +3,8 @@
 ##isocode zh_TW
 ##plural 1
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
+##numberformat 0000京0000兆0000億0000万0000
+##numberabbreviations 4=0000京0000兆0000億0000万|8=0000京0000兆0000億|12=0000京0000兆|16=0000京
```
wensimehrp, line 7:
```suggestion
##numberabbreviations 4=0000京0000兆0000億0000萬|8=0000京0000兆0000億|12=0000京0000兆|16=0000京
```

src/lang/traditional_chinese.txt
```diff
@@ -3,8 +3,8 @@
 ##isocode zh_TW
 ##plural 1
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
+##numberformat 0000京0000兆0000億0000万0000
```
wensimehrp, line 6:
```suggestion
##numberformat 0000京0000兆0000億0000萬0000
```

wensimehrp (changes requested):

src/lang/hindi.txt
```diff
@@ -3,8 +3,8 @@
 ##isocode hi_IN
 ##plural 0
 ##textdir ltr
-##digitsep ,
-##digitsepcur ,
+##numberformat 00,000,000,000,000,000,000
```
James103, line 6: Looks like one of the times when you updated the PR, the translated number format and abbreviations were somehow lost.
```suggestion
##numberformat 0,00,00,00,00,00,00,00,00,000
```

rubidium42: I have deliberately reverted that to the original format, because the format and abbreviations should be in-sync with eachother and I do not know how the digit grouping with the abbreviations is. Is it 10,000 crore or 100,00 crore? So I will leave it to the translators.

For CJK it is much clearer and since their translators are active, I have a clearer image what needed to happen there, and when I was wrong or copied the wrong character they quickly notified me.

src/language.h
```diff
@@ -107,7 +111,23 @@ extern const LanguageMetadata *_current_language;
 extern std::unique_ptr<icu::Collator> _current_collator;
 #endif /* WITH_ICU_I18N */
 
+/** The number digits available in a uint64_t. */
+constexpr int DIGITS_IN_UINT64_T = 20;
+/** Table with the text to place between each of the digits of a number. */
```
frosch123: This comment could be more descriptive:
* Are the digits counted from left to right, or from right to left?
* Are the strings inserted before or after the dight?

Something like ``Text at index "20 - i" is inserted after the digit with value "10**i".`` Though I have no idea whether this is correct, or the other way around. Hence why I am asking for a better comment :)

src/strgen/strgen_base.cpp
```diff
@@ -980,3 +982,113 @@ void LanguageWriter::WriteLang(const StringData &data)
...
+	}
+	return string;
+}
+
+/**
+ * Parse the \c NumberFormatSeparators out of the given format string, with the expected number of digits.
+ *
+ * Different cultures have different ways to separate their numbers when they get really big. In the Western world
+ * these are often called thousands separators which come every three digits counted from the back. The actual
+ * separator differs per country. In Chinese, Japanese and Korean they add a character every four digits counted
```
frosch123:
```suggestion
 * separator differs per language. In Chinese, Japanese and Korean they add a character every four digits counted
```

src/saveload/saveload.h
```diff
@@ -377,6 +377,7 @@ enum SaveLoadVersion : uint16_t {
 	SLV_DEPOT_UNBUNCHING,                   ///< 331  PR#11945 Allow unbunching shared order vehicles at a depot.
 	SLV_AI_LOCAL_CONFIG,                    ///< 332  PR#12003 Config of running AI is stored inside Company.
 	SLV_SCRIPT_RANDOMIZER,                  ///< 333  PR#12063 Save script randomizers.
+	SLV_NO_NUMBER_FORMATS,                  ///< 334  PR#11993 Number format rewriting; removal of saving of number formats in the savegame.
```
frosch123: Adding or removing settings requires no savegame bump. Only data migrations do.

src/settings_type.h
```diff
@@ -222,6 +222,8 @@ struct GUISettings {
 
 	bool   scale_bevels;                     ///< bevels are scaled with GUI scale.
 
+	std::string digit_decimal_separator; ///< decimal separator
```
frosch123, line 227: I have no idea why this is or was a setting. I would remove it :)

Is this for debugging?

src/lang/german.txt
```diff
@@ -3,6 +3,8 @@
 ##isocode de_DE
 ##plural 0
 ##textdir ltr
+##numberformat 00.000.000.000.000.000.000
+##numberabbreviations 3=00.000.000.000.000.000{NBSP}t|6=00.000.000.000.000{NBSP}Mio|9=00.000.000.000{NBSP}Mrd|12=00.000.000{NBSP}Bio|15=00.000{NBSP}Qa|18=00{NBSP}Qi
```
frosch123: This is a weird mixture of default and adjusted values :)
Correct would be: Tsd., Mio., Mrd., Bio., Brd., Trill.

src/settings_type.h
rubidium42: It's so someone can override whatever their chosen language dictates. Just like the format and abbreviation configuration.

src/strgen/strgen_base.cpp
rubidium42: In South Africa the decimal separators are different than in the UK, yet both use English as language. See e.g. https://en.wikipedia.org/wiki/Decimal_separator. So it's quite debatable whether it's language or country.

src/lang/german.txt
rubidium42: The first four are coming from the current translation, the others are just the English ones as they do not exist in the translations yet. My thoughts were to at least copy the existing ones, and leave the remaining untranslated. That would be less of a problem than just throwing away the translations.

I'll add these to the PR.

src/saveload/saveload.h
rubidium42: Hmm... how is this going to work?

I removed the settings completely and now some save game versions fail to load (I reckon from version 118).
Or should I have placed `SLV_SCRIPT_RANDOMIZER + 1` as last version in the settings ini file? In that case I would have gotten the comment that I just should've made a saveload version.
Or should I just have placed `SLV_SCRIPT_RANDOMIZER` as last version in the settings ini file? That would like have gotten a comment about misusing the saveload version as well.

So, what am I missing here?

frosch123: Probably https://github.com/OpenTTD/OpenTTD/blob/master/src/saveload/compat/settings_sl_compat.h#L261 needs some SLC_NULL. Though acually unsure whether we have that for strings.

frosch123 (approved):

Issue #11995, opened 2024-02-05
[Bug]: Some industries in tropical and arctic climates not producing cargo


Bonehead-101:
### Version of OpenTTD

14.0-beta2

### Expected result

Expected the Bank to accept Diamonds since that's it's good. 

### Actual result

I spent a good chunk of change to build a maglev path, to find out there were no diamond cars, I then downgraded to monorail to use the one engine with cargo capacity to find out the bank wasn't accepting the goods.

### Steps to reproduce

![image](https://github.com/OpenTTD/OpenTTD/assets/158988224/b3e7113d-5b8c-4260-9b8e-0ce96b9aa7cd)

PeterN:
Fixed by #12000

PR #12000, opened 2024-02-05
Fix 2fd90960: Missing default vehicles and industry acceptance/production.

PeterN:
## Motivation / Problem

Some default definitions are used across multiple climate types and rely on climate-independent cargo slot even though they specified a climate-dependent cargo type.

This no longer works since #11719 since cargo types are now cargo labels, which don't exist in some climates.

image

## Description

Add MixedCargoType that indirectly allows multiple labels to be specified for these.

Default definitions for vehicles, industries, and  industrytiles now use either a CargoLabel or a MixedCargoType, which is resolved during finalisation.

This is not applied to town houses as they don't have any cargo overlap.

image

## Limitations

Still a chance of unintended behaviour with old pre-label NewGRFs.

(A revert would also fix this issue, and all the limitations with that are known...)

## Checklist for review


LordAro:
Does every `GetCargoIDByLabel` call need to first call `GetActiveCargoLabel` ? If so, maybe merge the two?

PeterN:
> Does every `GetCargoIDByLabel` call need to first call `GetActiveCargoLabel` ? If so, maybe merge the two?

No, only those that access a `std::variant<>` cargo_label of default definitions during NewGRF finalisation need the extra call.

rubidium42 (dismissed):

PeterN:
This now caters for default industries that aren't cross-climate, but could potentially be substituted and set up that way by NewGRFs—this means it uses the three mixed cargo types even if the industry is only single-climate.

michicc (approved):

PeterN:
Fixes #11995

PR #12003, opened 2024-02-05
Change: Store running AI config inside Company

glx22:
## Motivation / Problem
A starting AI can change the config for its slot, like starting a "Random AI" will set its slot to the randomly chosen AI.
Also random deviation is applied to the config, and will be stored before being applied again on next start.

## Description
A starting AI copies the config and store it inside Company. All modifications are applied to this copy.
Settings window also use this copy while the AI is running.
When the AI is killed, settings window will use the original unmodified config.

https://github.com/OpenTTD/OpenTTD/assets/2952192/2532644f-1e27-42ff-9265-66d9c498337f

`ScriptConfig::is_random` is now unneeded, and removed.
Savegame stores `setting_game` config, then the running config if needed.
On load, the company is preloaded with the saved running config.
And `setting_game` config is reset to "Random AI" for old saves with running AI which were randomly chosen, else it just uses the running config, as there's no way to know what it was before starting.

`start_ai <AI>` behaviour is unchanged, it will modify config before starting the AI. The effect is similar to changing AI in GUI before using `start_ai`.

## Checklist for review

### DIFF
--- a/src/ai/ai_core.cpp
+++ b/src/ai/ai_core.cpp
@@ -33,26 +33,31 @@
 	return !_networking || (_network_server && _settings_game.ai.ai_in_multiplayer);
 }
 
-/* static */ void AI::StartNew(CompanyID company, bool rerandomise_ai)
+/* static */ void AI::StartNew(CompanyID company, bool deviate)
 {
 	assert(Company::IsValidID(company));
 
 	/* Clients shouldn't start AIs */
 	if (_networking && !_network_server) return;
 
-	AIConfig *config = AIConfig::GetConfig(company, AIConfig::SSS_FORCE_GAME);
+	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
+	Company *c = Company::Get(company);
+
+	AIConfig *config = c->ai_config.get();
+	if (config == nullptr) {
+		c->ai_config = std::make_unique<AIConfig>(AIConfig::GetConfig(company, AIConfig::SSS_FORCE_GAME));
+		config = c->ai_config.get();
+	}
+
 	AIInfo *info = config->GetInfo();
-	if (info == nullptr || (rerandomise_ai && config->IsRandom())) {
+	if (info == nullptr) {
 		info = AI::scanner_info->SelectRandomAI();
 		assert(info != nullptr);
 		/* Load default data and store the name in the settings */
-		config->Change(info->GetName(), -1, false, true);
+		config->Change(info->GetName(), -1, false);
 	}
-	if (rerandomise_ai) config->AddRandomDeviation();
+	if (deviate) config->AddRandomDeviation();
 	config->AnchorUnchangeableSettings();
-
-	Backup<CompanyID> cur_company(_current_company, company, FILE_LINE);
-	Company *c = Company::Get(company);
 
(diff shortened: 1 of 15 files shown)


src/ai/ai_core.cpp
```diff
@@ -111,11 +116,12 @@
 	delete c->ai_instance;
 	c->ai_instance = nullptr;
 	c->ai_info = nullptr;
+	delete c->ai_config;
+	c->ai_config = nullptr;
```
TrueBrain: As I hinted earlier on Discord, maybe a unique_ptr makes this all a bit more pretty, as then you just have to reset the unique_ptr and not manage anything else :)

glx22: Yeah, I tried, but incomplete type is a pain.

glx22: Found how to fix my unique_ptr issue :)

src/saveload/ai_sl.cpp
```diff
@@ -49,12 +55,21 @@ static void SaveReal_AIPL(int *index_ptr)
 		_ai_saveload_version = -1;
 	}
 
-	_ai_saveload_is_random = config->IsRandom();
```
TrueBrain, line 52: Doesn't `_ai_saveload_is_random` now remain unset? As you do read it a bit later on as if it is always set. Guess you need to force it to false if the new description exists?

src/saveload/saveload.h
```diff
@@ -374,7 +374,8 @@ enum SaveLoadVersion : uint16_t {
 	SLV_SHIP_ACCELERATION,                  ///< 329  PR#10734 Start using Vehicle's acceleration field for ships too.
 
 	SLV_MAX_LOAN_FOR_COMPANY,               ///< 330  PR#11224 Separate max loan for each company.
-	SLV_DEPOT_UNBUNCHING,                   ///< 330  PR#11945 Allow unbunching shared order vehicles at a depot.
+	SLV_DEPOT_UNBUNCHING,                   ///< 331  PR#11945 Allow unbunching shared order vehicles at a depot.
+	SLV_AI_LOCAL_CONFIG,                    ///< 332  PR#XXXXX Config of running AI is stored inside Company.
```
TrueBrain: As reminder :)

TrueBrain (commented):
What you also fix with this, if I am getting this right, is that when you make an AI Config change during runtime, for the settings that allow that, that those too are reverted when you cycle an AI?
As that is one of these things that annoyed me, that it didn't. So if this PR does that too, that is nice :)

That said, it does feel like this is tailored to developer-only situations, which normal players don't run into. But as it actually cleans up some of the weirdness of the code, I don't think that is an actual problem :)

src/saveload/ai_sl.cpp
glx22: It's not saved, and it's set to false before loading. But the variable will be read from old saves.

glx22:
I could add a new chunk to save the randomizers.
I'm also wondering about using the company/deity randomiser instead of OWNER_NONE for AddRandomDeviation().
But keeping OWNER_NONE for random AI selection.

rubidium42 (approved):

PR #12011, opened 2024-02-06
Fix: "restart current" reproduces how the script config of the current game was when it started

SamuXarick:
## Motivation / Problem
Alternative approach to #12003, without requiring a _saveload_ version bump.

In #11944, it was determined `random_deviation` was _inconsistent_, and an attempt to fix it was made so it would only deviate during script start. But this caused a few issues regarding reproducibility when using the recently introduced "restart current" command in #11963.
- `Reload AI` on a non-random AI makes the deviated values based on the current value, meaning they would walk outside the min-max range it had when the AI initially started.
- `restart current` does not roll back AIs that started as `Random AI`, does not roll back to the original values for parameters with random_deviation, similar to `Reload AI`, they could walk outside the min-max range.
- AIs that have not yet started at the time of saving game deviate to a different value once the company starts after loading from the savegame.
- `startai` when provided with the name of a script and parameters, applies random deviation on top of the user provided values.

## Description
- Restore Random AI slot upon stopping a randomly started AI.
- On "restart current" console command, use the seed of the current game to set the seed of scripts.

- Apply random deviation to non-random script settings upon starting a new game, except when using "restart current" via console, where the randomization counter is simulated but the values remain not deviated.

- Show random deviation range for scripts only outside normal game mode.

- Explicitly control whether to add random deviation
- Reloading AIs will only add random deviation if they were a Random AI.

- Random deviation is no longer applied after selecting a script during a running game.

- "startai" console command applies random deviation depending on the number of arguments:
  - 0 arguments - Only if the slot config is Random AI.
  - 1 argument - Always, no parameters were provided.
  - 2 arguments - Never, user provided parameters.

## Limitations
There's a partial revert of #11958 regarding GUI and #11944 regarding when random deviation is applied. This is done this way to make "restart current" reproducibility without necessitating a backup.

## Checklist for review

### DIFF
--- a/src/ai/ai.hpp
+++ b/src/ai/ai.hpp
@@ -28,8 +28,9 @@ public:
 	 * Start a new AI company.
 	 * @param company At which slot the AI company should start.
 	 * @param rerandomise_ai Whether to rerandomise the configured AI.
+	 * @param deviate Whether to add random deviation to settings that allow it.
 	 */
-	static void StartNew(CompanyID company, bool rerandomise_ai = true);
+	static void StartNew(CompanyID company, bool rerandomise_ai = true, bool deviate = true);
 
 	/**
 	 * Called every game-tick to let AIs do something.
--- a/src/ai/ai_config.hpp
+++ b/src/ai/ai_config.hpp
@@ -24,8 +24,8 @@ public:
 		ScriptConfig()
 	{}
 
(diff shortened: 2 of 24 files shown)


glx22:
#12003 solves the 2 first points, I have some ideas for 3rd point, and 4th works as expected to me.

PR #12016, opened 2024-02-06
Fix 5a88027: [Script] Avoid overflow in scripts when infinite money is enabled

merni-ns:
## Motivation / Problem
Infinite money mode as implemented in #11902 causes the script API's bank balance function to return `INT64_MAX`. When scripts do calculations with this value, overflows are likely to occur, since they do not use an overflow-proof data type to handle money.

## Description
Return `INT32_MAX` instead.

## Limitations
Scripts may think they don't have sufficient money to spend more than 2147483647 units of currency at a time.

## Checklist for review

### DIFF
--- a/src/script/api/script_company.cpp
+++ b/src/script/api/script_company.cpp
@@ -178,6 +178,8 @@
 {
 	company = ResolveCompanyID(company);
 	if (company == COMPANY_INVALID) return -1;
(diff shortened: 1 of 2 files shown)


michicc (approved):
Well, that is number is a good or bad as any other.

PR #12017, opened 2024-02-06
Codechange: also store the sending socket in Packets

rubidium42:
## Motivation / Problem

When adding encryption to our network packets, we need to add a "message authentication code" (MAC) data to the packet. This will take up some space if/when the packet gets encrypted.
Since some packets are filled up to the maximum size, before continuing to the next packet, the allocation for the MAC needs to be made when the packet is created. For this it needs to know whether the encryption is enabled on the socket.

There are multiple ways of achieving this:
* adding a `boolean` to the constructor, and then add checks to the socket handler at each of the call sites.
* adding the socket handler to the constructor.

When adding the `boolean` to the constructor you still need to pass the socket handler in `PrepareToSend` and you need to add some state whether the encryption has been added to the packet as well.
When passing the socket handler, it can be put in the already existing field for the socket handler and it can be set to `nullptr` in case encryption is not needed any more, saving a bit of space.

Note that adding encryption will be a post-14-branch endeavour. However, having this infrastructure available in the 14 branch makes it easier to backport any network changes that are not related with the encryption/key exchange.

## Description

`TransferToNetworkQueue` has a socket as parameter, but that is always equals to `this->cs` except when the server already closed the connection. Though then you would not be copying packets to the network queue.

The `PacketWriter`, that writes the network "map" packets during saving the game, reads and writes some instance variables that are accessed in other places within a lock. Either it is for allocating memory based on the variable being `nullptr`, or it is for checking the connection got closed. In case of the former that could technically be a problem when doing multiple simultaneous calls. In the latter case you could check for `nullptr`, then wait on the lock that `Destroy` has done and still do the expensive work. Moving it here just avoids that problem, for a tiny penalty when the socket got closed during download, though arguably that should be happening that often.

The last commit finally adds the socket to the packet's constructor. Except for some error messages that are sent before we even create the socket handler.

## Checklist for review

### DIFF
--- a/src/network/core/packet.cpp
+++ b/src/network/core/packet.cpp
@@ -38,13 +38,14 @@ Packet::Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size)
 
 /**
  * Creates a packet to send
+ * @param cs    The socket handler associated with the socket we are writing to; could be \c nullptr.
  * @param type  The type of the packet to send
  * @param limit The maximum number of bytes the packet may have. Default is COMPAT_MTU.
  *              Be careful of compatibility with older clients/servers when changing
  *              the limit as it might break things if the other side is not expecting
  *              much larger packets than what they support.
  */
-Packet::Packet(PacketType type, size_t limit) : pos(0), limit(limit), cs(nullptr)
+Packet::Packet(NetworkSocketHandler *cs, PacketType type, size_t limit) : pos(0), limit(limit), cs(cs)
 {
 	/* Allocate space for the the size so we can write that in just before sending the packet. */
 	this->Send_uint16(0);
@@ -57,7 +58,8 @@ Packet::Packet(PacketType type, size_t limit) : pos(0), limit(limit), cs(nullptr
  */
 void Packet::PrepareToSend()
 {
-	assert(this->cs == nullptr);
+	/* Prevent this to be called twice and for packets that have been received. */
+	assert(this->buffer[0] == 0 && this->buffer[1] == 0);
 
 	this->buffer[0] = GB(this->Size(), 0, 8);
 	this->buffer[1] = GB(this->Size(), 8, 8);
@@ -243,7 +245,6 @@ size_t Packet::Size() const
  */
 bool Packet::ParsePacketSize()
 {
-	assert(this->cs != nullptr);
 	size_t size = (size_t)this->buffer[0];
 	size       += (size_t)this->buffer[1] << 8;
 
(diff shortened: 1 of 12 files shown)


src/network/core/packet.cpp
```diff
@@ -57,7 +58,7 @@ Packet::Packet(PacketType type, size_t limit) : pos(0), limit(limit), cs(nullptr
  */
 void Packet::PrepareToSend()
 {
-	assert(this->cs == nullptr);
+	assert(this->buffer[0] == 0 && this->buffer[1] == 0);
```
TrueBrain, line 62: This change feels unexpected? Should it be part of this commit? And what is the reason behind it?

I do agree buffer[0] and [1] should be zero. But why no longer check if cs isn't nullptr? The Packet could still be made with a nullptr, so it is good check, not?

TrueBrain: On the other hand, `cs` is not used in this function, so that is a bit odd. Guess mostly it felt random in this commit that made me wonder. But I guess it also doesn't hurt. So if this was intended, I am fine with this.

Edit: on line 247 is a similar assert, while there too `cs` is never used. Maybe we should just remove that one as well.
Edit2: ironically, `CanReadFromPacket` does use `cs`, but does not have an `assert`. Lol.

rubidium42: The `assert(this->cs == nullptr)` was to prevent this function from being called on a received packet, because `cs` was `nullptr` when sending the packet.
Since `cs` can now also be `nullptr`, I needed another method. Upon creation of the packet, two zero bytes are reserved to later save the size of the packet. That's those two bytes. When you receive a packet,  they will not be both zero as you interpret the size in the code reading the bytes from the socket.

So the new assertion is slightly stronger, as it will fail upon calling `PrepareToSend` when the packet was received from the network, but it will also prevent `PrepareToSend` from being called twice.

TrueBrain: So in that logic, line 247 was to ensure it was done on a received packet, and not on a write packet. That was already unclear (lack of comments anyone? :D ), but maybe it would be good to also address that in this PR, by the same logic as this change?

I can imagine us just removing the line, as it is rather pedantic to have it as this stage. But maybe you have something clever to know if it is actually reading or writing?

rubidium42: Removing the line 247 one would be the best way going forward, as I can't do the all zero's trick as then some lovely person is going to send two zeros as size and crash the server/client when assertions are enabled.

TrueBrain (approved):

PR #12023, opened 2024-02-06
Change: When adding orders, Ctrl+Click on a depot to unbunch, instead of service if required

2TallTyler:
## Motivation / Problem

In https://github.com/OpenTTD/OpenTTD/pull/8342#issuecomment-1928615186, the author points out that adding an unbunching depot order is more clicks than the alternate depot-less solution proposed (but not finished in time for beta1).

## Description

Currently, when you're adding orders, Ctrl+Click on a depot makes the order "Service if required."

This PR changes this shortcut to "Unbunch," saving several seconds per route since you no longer have to select the order and use the dropdown to change the action. This adds up.

## Limitations

Workflow.png

I guess we could add a setting to select what happens when you Ctrl+Click on a depot.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -4613,7 +4613,7 @@ STR_ORDER_GO_TO_NEAREST_DEPOT                                   :Go to nearest d
 STR_ORDER_GO_TO_NEAREST_HANGAR                                  :Go to nearest hangar
 STR_ORDER_CONDITIONAL                                           :Conditional order jump
 STR_ORDER_SHARE                                                 :Share orders
-STR_ORDERS_GO_TO_TOOLTIP                                        :{BLACK}Insert a new order before the highlighted order, or add to end of list. Ctrl+Click on a station for 'full load any cargo', on a waypoint to invert the 'non-stop by default' setting, or on a depot for 'service'. Click on another vehicle to copy its orders or Ctrl+Click to share orders. A depot order disables automatic servicing of the vehicle
+STR_ORDERS_GO_TO_TOOLTIP                                        :{BLACK}Insert a new order before the highlighted order, or add to end of list. Ctrl+Click on a station for 'full load any cargo', on a waypoint to invert the 'non-stop by default' setting, or on a depot for 'unbunch'. Click on another vehicle to copy its orders or Ctrl+Click to share orders. A depot order disables automatic servicing of the vehicle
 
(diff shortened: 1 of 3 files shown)


src/order_gui.cpp
```diff
@@ -396,8 +396,16 @@ static Order GetOrderCmdFromTile(const Vehicle *v, TileIndex tile)
 				ODTFB_PART_OF_ORDERS,
 				(_settings_client.gui.new_nonstop && v->IsGroundVehicle()) ? ONSF_NO_STOP_AT_INTERMEDIATE_STATIONS : ONSF_STOP_EVERYWHERE);
 
-		if (_ctrl_pressed) order.SetDepotOrderType((OrderDepotTypeFlags)(order.GetDepotOrderType() ^ ODTFB_SERVICE));
```
glx22, line 401: The XOR makes me think there's a setting for "service by default", but I didn't check.

Edit: After checking it seems it's just used like an normal OR.

src/lang/english.txt
```diff
@@ -4609,7 +4609,7 @@ STR_ORDER_GO_TO_NEAREST_DEPOT                                   :Go to nearest d
 STR_ORDER_GO_TO_NEAREST_HANGAR                                  :Go to nearest hangar
 STR_ORDER_CONDITIONAL                                           :Conditional order jump
 STR_ORDER_SHARE                                                 :Share orders
-STR_ORDERS_GO_TO_TOOLTIP                                        :{BLACK}Insert a new order before the highlighted order, or add to end of list. Ctrl+Click on a station for 'full load any cargo', on a waypoint for 'non-stop', or on a depot for 'service'. Click on another vehicle to copy its orders or Ctrl+Click to share orders. A depot order disables automatic servicing of the vehicle
+STR_ORDERS_GO_TO_TOOLTIP                                        :{BLACK}Insert a new order before the highlighted order, or add to end of list. Ctrl+Click on a station for 'full load any cargo', on a waypoint for 'non-stop', or on a depot for 'unbunch'. Click on another vehicle to copy its orders or Ctrl+Click to share orders. A depot order disables automatic servicing of the vehicle
```
glx22: `on a waypoint for 'non-stop'` actually when looking at the code, it's more like "invert the effect of non-stop by default setting"

2TallTyler: That's a separate bug, no? 😄

glx22: I guess nobody reads the tooltips :)

2TallTyler: Fixed in #12081.

rubidium42 (approved):

PR #12034, opened 2024-02-07
Revert: Build signals to the next junction when dragging regardless of the Ctrl state

### LINKED ISSUE #12033: [Bug]: Impossible to build/remove signals for a given length
### Version of OpenTTD

Master-2024-02-07; Windows 10

### Expected result

Considering the recent reverse of ctrl+drag on signals, simple drag should build signals on full sectin; ctrl+drag should build signals to a given length:
изображение

### Actual result

Ctrl+drag acts in the same way as a regular drag; signals are built until the end of the section in any case.

https://github.com/OpenTTD/OpenTTD/assets/115672237/a84bd7cc-7533-4425-982f-4e8bf02fe469

Same problem with signal removal.

https://github.com/OpenTTD/OpenTTD/assets/115672237/b711a3f6-bd50-478e-a414-97ef6c17fe66

### Steps to reproduce

1. Build a relatively long section of tracks
2. Try to build signals on only half of this section

TrueBrain:
This reverts commit b370ae1212abdc97034a6b4b8a2dffcffa1911f0 and #9637.

## Motivation / Problem

Honestly, the idea of the PR is solid. Just the execution is confusing to any existing player, and basically alienates them. I think what has been missed in #9637 what this does to the current players. Let me explain a bit more.

When dragging signals it normally placed signals in the block that becomes white. Visually very clear what is going on. When holding CTRL it would continue to the next block. Nothing visually telling you that, but if you knew, you knew.

Now with #9637 we completely removed the non-CTRL and made the CTRL the default. Not an option, not a way to invert this behaviour, nothing. Which I am not against, except ....

There is no visual indicator that this is going on. So from a player perspective, I drag my signal for N tiles, and it does this all the way to the next junction. This is mostly confusing if you select OVER a junction: signals are not appearing now passed the junction.
Even worse, the "Length: " text still gives the suggestion it will do it for your selection. It is a lie :)

For new players, this most likely isn't the biggest problem, but for existing players this is just hostile. You don't understand what is going on, and no feedback is given what actually happens. This is because visually NOTHING changed. But the behaviour changed drastically.

Video to illustrate the frustration:

https://github.com/OpenTTD/OpenTTD/assets/1663690/43ad8290-80ab-40a3-8966-627ba281ae93

## Description

I think we need to take #9637 back to the drawing board. For example, a good solution might be to highlight all the track the signals will be placed on. Nothing more, nothing less. This makes it very clear what is going to happen. People might still disagree with the change, but at least visually there is a very clear indication that the behaviour changed, and how.

Another option might be to stop the highlighting after 3 tiles, to indicate it is going to auto-fill. Or show an icon to indicate auto-fill. Or basically ... ANYTHING, to show the behaviour is different.

Cheesing our way out, it could help to change the label "Length: NN" to "Till next junction", or something.

Yet another suggestion from Discord: add a button in the Signal GUI to enable/disable auto-fill.

But, as we are closing in on 14.0, I suggest we just revert the PR for now, so we can think about this in less of a rush, to find a better solution.

Again, I have nothing against the concept of #9637; I do about making a change we force upon everyone without having any visual indicators we did so :) Which makes this very different from changing a default setting, hiding signals by default, etc etc.

#12033 should be considered in the new solution too. For now, this closes #12033.

PS: what mostly tripped me up was trying to remove 2 signals on a rail; it removes all to the next junction, without telling me. I was confused. A lot :) Muscle memory is a tricky thing, especially when no visual conflict is created :)

https://github.com/OpenTTD/OpenTTD/assets/1663690/45cb1e10-ae4a-4807-b027-31c10a52f227

Again, visual indicator is the issue. Not the actual change :)

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1923,7 +1923,7 @@ STR_CONFIG_SETTING_DRAG_SIGNALS_DENSITY                         :When dragging, 
 STR_CONFIG_SETTING_DRAG_SIGNALS_DENSITY_HELPTEXT                :Set the distance at which signals will be built on a track up to the next obstacle (signal, junction), if signals are dragged
 STR_CONFIG_SETTING_DRAG_SIGNALS_DENSITY_VALUE                   :{COMMA} tile{P 0 "" s}
 STR_CONFIG_SETTING_DRAG_SIGNALS_FIXED_DISTANCE                  :When dragging, keep fixed distance between signals: {STRING2}
-STR_CONFIG_SETTING_DRAG_SIGNALS_FIXED_DISTANCE_HELPTEXT         :Select the behaviour of signal placement when dragging signals. If disabled, signals are placed around tunnels or bridges to avoid long stretches without signals. If enabled, signals are placed every n tiles, making alignment of signals at parallel tracks easier
+STR_CONFIG_SETTING_DRAG_SIGNALS_FIXED_DISTANCE_HELPTEXT         :Select the behaviour of signal placement when Ctrl+dragging signals. If disabled, signals are placed around tunnels or bridges to avoid long stretches without signals. If enabled, signals are placed every n tiles, making alignment of signals at parallel tracks easier
 
 STR_CONFIG_SETTING_SEMAPHORE_BUILD_BEFORE_DATE                  :Automatically build semaphores before: {STRING2}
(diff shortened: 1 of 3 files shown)


PeterN (approved):

Bouke:
> For example, a good solution might be to highlight all the track the signals will be placed on.

My 2 cents: Having to press ctrl to build until the next junction, combined with the highlighting would be a nice way of 'progressive disclosure'. Drag signals and then press ctrl and immediately see the highlight running until the next junction makes it much clearer what ctrl does.

PR #12035, opened 2024-02-07
Change: Don't treat Humans and AIs differently when it comes to demolishing houses.

PeterN:
## Motivation / Problem

`CanDeleteHouse()` allows human players to demolish any house, but prevents AI players doing the same.

This is an old left-over from before NoAI existed.

I think the general principle these days is that AIs should be able to do the same things as normal players.

## Description

Resolved by changing the test in `CanDeleteHouse()` from a valid human-player company to any valid company.

## Limitations

Untested: Xarick? :)

## Checklist for review

### DIFF
--- a/src/newgrf_house.cpp
+++ b/src/newgrf_house.cpp
@@ -513,9 +513,9 @@ bool CanDeleteHouse(TileIndex tile)
 {
 	const HouseSpec *hs = HouseSpec::Get(GetHouseType(tile));
 
-	/* Humans are always allowed to remove buildings, as is water and disasters and
+	/* Companies are always allowed to remove buildings, as is water and disasters and
 	 * anyone using the scenario editor. */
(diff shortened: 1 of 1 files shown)


TrueBrain (dismissed):
Sold!

frosch123:
I don't think this is about old/new AIs. This is a pure NewGRF feature.

>  This building is protected, i.e. towns and AI players won't remove it. Human players can still remove it, so you may need to set a high remove cost/rating to make them think twice. 

I think this is used for historical buildings, which the NewGRF does not want to get demolished.

As such, I propose the opposite: 
* Also deny human players to remove these buildings, 
* except when the recently promoted "sandbox" mode is enabled.

rubidium42 (dismissed):

SamuXarick:
> Untested: Xarick? :)

Game Scripts should perhaps be able to delete these structures? But they could be in CompanyMode, so... I'm not sure.

PeterN:
Frosch said no.

PR #12039, opened 2024-02-09
Change: Restore original width of spacers in main toolbars.

PeterN:
## Motivation / Problem

Over time we've added more buttons to the toolbar, which has reduced the width of the spacers. However, the toolbar itself is not actually fixed width, so I looked into how the width of the spacers are determined.

And I found that they only have space due to a bug.

Width for all spacers is included only due to an off-by-one from counting buttons and not excluding the normally hidden switcher buttons—so all spacers fit into the size of a single button, and they're barely visible now.

image

## Description

Instead, exclude the switcher button from the button count, and include spacer width of half the button width is now included explicitly. This width was chosen to match original TTD, although it's not exact as that varies between 11 or 12 pixels.

image

## Limitations

We're all probably used to the narrow spacers by now...

## Checklist for review

### DIFF
--- a/src/toolbar_gui.cpp
+++ b/src/toolbar_gui.cpp
@@ -1414,7 +1414,13 @@ public:
 				child_wid->current_x = child_wid->smallest_x;
 			}
 		}
-		_toolbar_width = nbuttons * this->smallest_x;
+
+		/* Exclude the switcher button which is not displayed when the toolbar fits the screen. When the switch is
+		 * displayed there will be no spacers anyway. */
+		--nbuttons;
+
+		/* Allow space for all buttons, and include spacers at quarter the width of buttons. */
+		_toolbar_width = nbuttons * this->smallest_x + this->spacers * this->smallest_x / 4;
 	}
 
 	void AssignSizePosition(SizingType sizing, int x, int y, uint given_width, uint given_height, bool rtl) override


src/toolbar_gui.cpp
```diff
@@ -1378,7 +1378,8 @@ class NWidgetToolbarContainer : public NWidgetContainer {
 				child_wid->current_x = child_wid->smallest_x;
 			}
 		}
-		_toolbar_width = nbuttons * this->smallest_x;
+		nbuttons--; // Exclude the switcher button whch is not displayed when the toolbar fits the screen.
```
glx22:
```suggestion
		nbuttons--; // Exclude the switcher button which is not displayed when the toolbar fits the screen.
```

glx22 (dismissed):

EmperorJake:
They certainly weren't that wide back in the olden days, I agree somewhere around 1/4 would look better
image

PeterN:
This is 1/3 width:
image

This is 1/4 width:
image

TrueBrain:
Haha, this one is funny. I too am so used to the smaller ones, that the bigger ones look weirdddddd :D

Shall we just do 1/4th, and go with it? As the code in this PR does make sense to add :)

2TallTyler:
I'd support going with 1/4 width.

PeterN:
Code is now commented better and changed to be quarter width.

TrueBrain (approved):

PR #12048, opened 2024-02-09
Change: Draw north-edge fences at tile edge instead of 1/16th in.

PeterN:
## Motivation / Problem

(Fences here means fences, hedges and walls.)

Drawing fences at 1/16th in and at 16/16ths means that is it not possible to make fences align nicely. This is more problematic with high resolution 4x extra zoom graphics.

These sprites were also originally (in OpenTTD before r23168 and in TTD) only drawn on tiles edges, never 1/16th in.

r23168 also allowed adjacent fences to be created, which can look a bit odd.

image
image

## Description

Draw north-edge fences at the tile edge instead. This allows high resolution sprites to align better around fields, and matches the older behaviour. When farms place fields we also now avoid placing fields along an edge that already has fences.

This is listed as a change not a fix, as the old behaviour was not broken as such, but it can be awkward for artists to work with.

image
image

## Limitations

Pitchforks somewhere.

## Checklist for review

### DIFF
--- a/src/clear_cmd.cpp
+++ b/src/clear_cmd.cpp
@@ -69,14 +69,14 @@ static void DrawClearLandFence(const TileInfo *ti)
 	if (fence_nw != 0) {
 		int z = GetSlopePixelZInCorner(ti->tileh, CORNER_W);
 		SpriteID sprite = _clear_land_fence_sprites[fence_nw - 1] + _fence_mod_by_tileh_nw[ti->tileh];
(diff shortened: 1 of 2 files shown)


2TallTyler (approved):
It took me an embarrassingly long time to realize that you mean 1/16th of a tile. Maybe I'm just dense, or perhaps `Change: Don't inset fences into farm tiles` would be more readable in the commit log. 🙂

Issue #8987, opened 2021-04-10
Loading JGRPP savegames in official OpenTTD gives "too new" instead of "modified version" error


James103:
## Version of OpenTTD
1.11.0

## Expected result
Attempting to load a JGRPP savegame (with version `>= 0x8000`) in the official master or release branch results in the error
`Game Load Failed - Savegame is made with a modified version`.

## Actual result
Attempting to load a JGRPP savegame (with version `>= 0x8000`) in the official master or release branch results in the error
`Game Load Failed - Savegame is made with newer version`.

## Steps to reproduce
1. Download and run any JGRPP version of OpenTTD.
2. Using JGRPP, save a game.
3. Download and run the latest official release or nightly version of OpenTTD.
4. Using the official release or nightly, attempt to load the savegame that was saved using JGRPP.
5. The loading fails with `Game Load Failed - Savegame is made with newer version` as it attempts to load a savegame whose version is approx. `0x8000` greater than the current latest savegame version.

## Justification
The following text explains why using `STR_GAME_SAVELOAD_ERROR_PATCHPACK` would be more helpful instead of `STR_GAME_SAVELOAD_ERROR_TOO_NEW_SAVEGAME`:
https://github.com/OpenTTD/OpenTTD/blob/e98aed8b485caa63145ae1d8f48b177da3b54006/src/saveload/saveload.h#L308-L320

TrueBrain:
This ticket presents a solution rather than a problem, so it is always a bit tricky to look through that.

I am not a fan of using `0x8000` marker to detect patch-packs. For JGRPP it holds true, but JGRPP is not the only patchpack out there. There are several others, which use other ways to fiddle with their savegame versions. The block you reference shows this clearly. So if we would be to use `0x8000` it is specially for JGRPP, not for any patchpack.

That said, under this ticket is of course the question: when you load a savegame from a patchpack (any patchpack), it says it is made with a `newer` version. Which strictly seen is true, and as such, not a bug. But we could maybe be nicer about these things.

What I have been thinking for a while, that maybe we should include a string in the savegame to indicate what "version" it is. We set it to "vanilla", JGRPP can set it to "JGRPP", and any other patchpack to what-ever the want to be called. That means that if we spot a mismatch, we can tell where the savegame came from.

The other approach is to upstream some parts of the savegame system JGRPP uses, which allows us to detect if there are blocks we don't understand. That also makes it possible to give a more clear error.

[...]

PR #12054, opened 2024-02-11
Feature: Show revision of newer savegames that have this information

rubidium42:
## Motivation / Problem

When looking at #8987 I thought... wouldn't is be even better to just show the revision of the version that is deemed too new? That will be way more useful for the user, as they know what version they need to load said save.
And for developers it's useful to see whether the save that can't be loaded is from one of our own development branches, and as such it's expected that it might not load.

## Description

Add the revision-string right after the eight byte header, and misuse an unused byte to store the length. This also means that the string will be saved uncompressed, so even when unknown compressions are used the revision can still be determined.

I have chosen for this solution because a save-load version is definitely not going to work, as there will be plenty of JGRPP saves that won't have this information with a higher save-load version. The byte has been saved as zero since at least the 0.4 era.
Also trying to read the game log chunks will eventually fail when there are more types are added, or when types change. So, this seems like the least intrusive way to add the revision.

## Limitations

This will break any save reading libraries/tools.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -4902,7 +4902,7 @@ STR_ERROR_UNABLE_TO_DELETE_FILE                                 :{WHITE}Unable t
 STR_ERROR_GAME_LOAD_FAILED                                      :{WHITE}Game Load Failed{}{STRING1}
 STR_GAME_SAVELOAD_ERROR_BROKEN_INTERNAL_ERROR                   :Internal error: {RAW_STRING}
 STR_GAME_SAVELOAD_ERROR_BROKEN_SAVEGAME                         :Broken savegame - {RAW_STRING}
-STR_GAME_SAVELOAD_ERROR_TOO_NEW_SAVEGAME                        :Savegame is made with newer version
+STR_GAME_SAVELOAD_ERROR_TOO_NEW_SAVEGAME_WITH_REVISION          :Savegame is made with newer/unsupported version {RAW_STRING}
 STR_GAME_SAVELOAD_ERROR_FILE_NOT_READABLE                       :File not readable
 STR_GAME_SAVELOAD_ERROR_FILE_NOT_WRITEABLE                      :File not writeable
 STR_GAME_SAVELOAD_ERROR_DATA_INTEGRITY_CHECK_FAILED             :Data integrity check failed
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -138,7 +138,7 @@ struct PacketWriter : SaveFilter {
 		return false;
 	}
 
-	void Write(byte *buf, size_t size) override
+	void Write(const byte *buf, size_t size) override
 	{
 		/* We want to abort the saving when the socket is closed. */
 		if (this->cs == nullptr) SlError(STR_NETWORK_ERROR_LOSTCONNECTION);
@@ -147,7 +147,7 @@ struct PacketWriter : SaveFilter {
 
 		std::lock_guard<std::mutex> lock(this->mutex);
 
-		byte *bufe = buf + size;
+		const byte *bufe = buf + size;
 		while (buf != bufe) {
 			size_t written = this->current->Send_bytes(buf, bufe);
 			buf += written;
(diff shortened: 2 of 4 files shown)


TrueBrain:
"Feature" .. no offense, but aren't you overselling this?

JGRennison:
Breaking all of the existing save/load tooling is not something to take lightly.

Not being able to understand the compression format involved is rarely an issue.
The gamelog chunk is quite complicated, so parsing it is likely not worth the bother/too complicated to safely do in a forwards compatible way.
However you could fairly easily add a new chunk which just contains the version string and nothing else, and which isn't going to change in the future. A RIFF is presumably more future-proof but a table is most likely also be fine, if the table format isn't likely to be radically changed in future. Existing tooling can then quite happily skip over it the same as any other unknown chunk type.

frosch123:
I agree, showing the revision is not worth breaking all savegame readers and tools we have.
The gamelog is already part `LoadCheckData` for [13 years](https://github.com/OpenTTD/OpenTTD/commit/433f74edd9fd3d9f3fed5cf09c7d6c510600b0aa#diff-ff928d001527c3a2684775db76c6765da0ae5f2d9b2a68b037a061661c56fb74). So it seems pretty stable.

rubidium42:
> I agree, showing the revision is not worth breaking all savegame readers and tools we have. The gamelog is already part of `LoadCheckData` for [13 years](https://github.com/OpenTTD/OpenTTD/commit/433f74edd9fd3d9f3fed5cf09c7d6c510600b0aa#diff-ff928d001527c3a2684775db76c6765da0ae5f2d9b2a68b037a061661c56fb74). So it seems pretty stable. Since it is part of `LoadCheckData` you could even show this in the file browser, before loading the game.

The saveload code won't load the gamelog, because the savegame version is newer/unsupported. And that gamelog could contain things that the older OpenTTD does not understand. After all, the content of the savegame was never intended to be forward compatible. Yes, we might get away with loading the gamelog for a while... but at some point it will fail to be forward compatible.
When the savegame version is newer, then that does not guarantee the table headers to be present. For example the JGRPP versions before we added this feature. You could try to read the data as table headers, but how would you know that the headers are valid? So attempting to read the table headers feels quite dangerous.

> Not being able to understand the compression format involved is rarely an issue.

Until there is some new compression algorithm that starts to get used. Or we decide to encrypt the content of the savegames for some reason. Just any new format that gets added is unsupported, and makes this feature fail.


Would adding something after all the RIFFs work for the tools, i.e. after the terminator? In other words instead of injecting the string before the RIFFs, appending it after the RIFFs? Or is using the lower 8 bits of the `uint` that saves the version information already too much of a change for the tools?

frosch123:
Here are two example readers: they probably copied a fair share from each other
* https://github.com/TrueBrain/OpenTTD-savegame-reader/blob/main/savegame_reader/savegame.py
* https://github.com/OpenTTD/bananas-api/blob/main/bananas_api/new_upload/readers/scenario.py

I guess the bananas reader in particular is supposed to detect any crap injected into the savegame. So, if you find a place to exploit it, that would probably count as a bug in the reader :)

JGRennison:
Table chunks have their own chunk type, you don't need to examine the savegame version field to detect whether they're using table headers or not.

I don't see that it's necessary to try to avoid putting the version string in a chunk.
The point of the chunk format is that chunks are separate, unknown chunks can be just skipped over, and chunks can be added/removed as required.

That said, if it really must be outside of the compressed section, perhaps consider adding a proper secondary header format?
Using the last byte of the primary header in this way means that if you want to go through this process again with some other thing you'll be a bit stuck as it's now even less extensible.

frosch123:
I don't consider new compression algorithms that likely. It's more likely that OpenTTD is compiled for some system without the required libraries, so some compression algorithm is missing. In that case showing "you need OpenTTD version X to load" is wrong, you would rather have to show "you need to compile OpenTTD with support for compression X".

So I still prefer just using the gamelog. And in the rare case that it is not readable, show some generic message.
It's rather unlikely that a the savegame format will be changed incompatibly. In fact this PR is the first one to do so in a decade.

rubidium42:
Seems like this implementation is too much trouble for too little benefit.

PR #7962, opened 2020-01-25
Improve rendering of large viewport areas

ldpl:
My take on sprite sorter optimization, I estimate it to be about 10x faster than the current one. And while it's still not ideal it's decently fast and I'm tired of this shit anyway :P

It's made to perfectly replicate the current sorting algorithm (including weird reversing of moved sprites) so there shouldn't be any (new) glitches. 

Also, there is definitely some room for further optimizations. At the very least map of sets can probably be replaced with some more specialized structure. For example, kd tree should be faster but its current implementation of node removal is very inefficient and needs to be optimized. It's also possible that deviating from the original algorithm can be beneficial to performance (std::sort does it instantly but I couldn't find a good ordering relation). 

If someone is interested here is my testing code and over a dozen sorter implementations:
[sprite-sorter.zip](https://github.com/OpenTTD/OpenTTD/files/4112927/sprite-sorter.zip)

PR #12055, opened 2024-02-11
Change: Improve performance of finding free pool slots.

PeterN:
## Motivation / Problem

Finding a free pool slot often needs to iterate the pool. The first free index is stored, but after that is used, it needs to individually check if each item is empty.

Some vehicles produce lots of effect vehicles, so in a game with lots of vehicles finding a free slot can be called very often.

## Description

Add a bitmap of used pool slots which allows finding a free pool slot without having to check if each index is already used or not. The size of the bitmap grows along with the size of the pool, although it is aligned to 64 items (one uint64_t) at a time.

This allows checking a block of 64 items in one go, and we also know which items are free in that block without individually checking.

The bitmap stores a bit per pool item, which means 8 bytes can store the state of 64 items. Conversely, each pool item pointer itself uses 8 bytes. This means that the bitmap uses only an additional 1/64th the memory  that is already used by the pool item pointers.

Loosely based on a JGRPP patch.

Performance runs with a heavy (4kx4k, 51MB) test save:

| Pool | master 10000 calls | master avg | this pr 10000 calls | this pr avg |
| ------- | ------------------ | ------------ | ------------------- | ----------- |
| Vehicle | 115050 us | avg: 11.5 us | 1837 us | avg: 0.2 us |
| Vehicle | 124895 us | avg: 12.5 us | 1550 us | avg: 0.2 us |
| Vehicle | 109787 us | avg: 11.0 us | 1402 us | avg: 0.1 us |
| Vehicle | 104227 us | avg: 10.4 us | 1429 us | avg: 0.1 us |
| Vehicle | 124463 us | avg: 12.4 us | 1659 us | avg: 0.2 us |
| Vehicle | 123507 us | avg: 12.4 us | 1477 us | avg: 0.1 us |
| Vehicle | 128174 us | avg: 12.8 us | 1525 us | avg: 0.2 us |
| Vehicle | 119987 us | avg: 12.0 us | 1667 us | avg: 0.2 us |
| Vehicle | 115823 us | avg: 11.6 us | 1639 us | avg: 0.2 us |
| Vehicle | 98529 us | avg: 9.9 us | 1492 us | avg: 0.1 us |
| Vehicle | 96376 us | avg: 9.6 us | 1452 us | avg: 0.1 us |
| Vehicle | 126777 us | avg: 12.7 us | 1455 us | avg: 0.1 us |
| Vehicle | 107639 us | avg: 10.8 us | 1387 us | avg: 0.1 us |
| Vehicle | 116721 us | avg: 11.7 us | 1481 us | avg: 0.1 us |
| Vehicle | 114456 us | avg: 11.4 us | 1738 us | avg: 0.2 us |

## Checklist for review

### DIFF
--- a/src/core/pool_type.hpp
+++ b/src/core/pool_type.hpp
@@ -83,6 +83,9 @@ struct Pool : PoolBase {
 
 	static constexpr size_t MAX_SIZE = Tmax_size; ///< Make template parameter accessible from outside
 
+	using BitmapStorage = size_t;
+	static constexpr size_t BITMAP_SIZE = std::numeric_limits<BitmapStorage>::digits;
+
 	const char * const name; ///< Name of this pool
 
 	size_t size;         ///< Current allocated size
@@ -95,6 +98,7 @@ struct Pool : PoolBase {
 	bool cleaning;       ///< True if cleaning pool (deleting all items)
 
 	Titem **data;        ///< Pointer to array of pointers to Titem
+	std::vector<BitmapStorage> used_bitmap; ///< Bitmap of used indices.
 
 	Pool(const char *name);
 	void CleanPool() override;
--- a/src/core/pool_func.hpp
+++ b/src/core/pool_func.hpp
@@ -11,6 +11,7 @@
 #define POOL_FUNC_HPP
 
 #include "alloc_func.hpp"
+#include "bitmath_func.hpp"
 #include "mem_func.hpp"
 #include "pool_type.hpp"
 #include "../error_func.h"
@@ -60,6 +61,16 @@ DEFINE_POOL_METHOD(inline void)::ResizeFor(size_t index)
 	this->data = ReallocT(this->data, new_size);
 	MemSetT(this->data + this->size, 0, new_size - this->size);
 
+	this->used_bitmap.resize(Align(new_size, BITMAP_SIZE) / BITMAP_SIZE);
+	if (this->size % BITMAP_SIZE != 0) {
+		/* Already-allocated bits above old size are now unused. */
+		this->used_bitmap[this->size / BITMAP_SIZE] &= ~((~static_cast<BitmapStorage>(0)) << (this->size % BITMAP_SIZE));
+	}
+	if (new_size % BITMAP_SIZE != 0) {
+		/* Bits above new size are considered used. */
+		this->used_bitmap[new_size / BITMAP_SIZE] |= (~static_cast<BitmapStorage>(0)) << (new_size % BITMAP_SIZE);
+	}
+
 	this->size = new_size;
 }
 
@@ -69,25 +80,20 @@ DEFINE_POOL_METHOD(inline void)::ResizeFor(size_t index)
  */
 DEFINE_POOL_METHOD(inline size_t)::FindFirstFree()
 {
-	size_t index = this->first_free;
-
-	for (; index < this->first_unused; index++) {
-		if (this->data[index] == nullptr) return index;
+	for (auto it = std::next(std::begin(this->used_bitmap), this->first_free / BITMAP_SIZE); it != std::end(this->used_bitmap); ++it) {
+		BitmapStorage available = ~(*it);
+		if (available == 0) continue;
+		return std::distance(std::begin(this->used_bitmap), it) * BITMAP_SIZE + FindFirstBit(available);
 	}
 
-	if (index < this->size) {
-		return index;
-	}
-
-	assert(index == this->size);
 	assert(this->first_unused == this->size);
 
-	if (index < Tmax_size) {
-		this->ResizeFor(index);
-		return index;
+	if (this->first_unused < Tmax_size) {
+		this->ResizeFor(this->first_unused);
+		return this->first_unused;
 	}
 
-	assert(this->items == Tmax_size);
+	assert(this->first_unused == Tmax_size);
 
 	return NO_FREE_ITEM;
 }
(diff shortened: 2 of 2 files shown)


src/core/pool_type.hpp
```diff
@@ -83,6 +84,9 @@ struct Pool : PoolBase {
 
 	static constexpr size_t MAX_SIZE = Tmax_size; ///< Make template parameter accessible from outside
 
+	using BitmapStorage = uint64_t;
+	static constexpr size_t BITMAP_SIZE = std::numeric_limits<BitmapStorage>::digits;
```
nielsmh, line 87: Maybe change this to a plain `uint`, in part to silence the warning produced for MSVC builds, and in part just because it isn't actually used an index into an STL container.

Edit: Oops that's not what it's complaining about now that I read it one more time.

ldpl:
Have you tried simply doing list of free pool slots? With OpenTTD pools never really going down much I'd expect that to be more efficient solution. Maybe a sorted list though.

PeterN:
> Have you tried simply doing list of free pool slots? With OpenTTD pools never really going down much I'd expect that to be more efficient solution.

I have not.

JGRennison:
Allocation of free pool IDs needs to be the same for all instances when in a multiplayer, maintaining a last freed structure complicates that.
Pool IDs can also be allocated directly, which also somewhat complicates maintenance of a such a structure.

ldpl:
You can always take the min index from the list to avoid having to sync it in mp. And pre-fill the list up to max index on afterload.

src/core/pool_type.hpp
```diff
@@ -83,6 +84,9 @@ struct Pool : PoolBase {
 
 	static constexpr size_t MAX_SIZE = Tmax_size; ///< Make template parameter accessible from outside
 
+	using BitmapStorage = uint64_t;
```
rubidium42: Maybe the following is a faster solution on builds with a 32 bits architecture. After all, if the compiler can't access the 64 bit registers, then it has to do this with two registers which might worse for performance than just splitting the calculations.
Note that I have not run any benchmarks, and that 32 bits builds are generally < 0.5% in the survey.
```suggestion
	using BitmapStorage = size_t;
```

PeterN: Seems fair, and I designed the rest of the code for this occurrence... clearly ;)

nielsmh: More appropriate might be `uint_fast32_t`, which will typically be 64 bit on 64 bit native platforms, and 32 bit on those where 64 bit is not efficient. (https://stackoverflow.com/a/59593614)

Edit: Or `uintptr_t` of course.

rubidium42: I looked into `uint_fast32_t`, but... MSVC x64's version is 32 bits and not 64 bits: https://godbolt.org/z/fY31o6v53
If that would've worked, I would've suggested that instead of `size_t`.

rubidium42 (dismissed):
I doubt other solutions are significantly better. If they are, that could be a different PR.

PeterN:
Here's some stats from testing with priority_queue. While it's faster in extreme cases, the performance of bitmap is sometimes faster. Wentbourne doesn't have effect vehicles due to maglev, so the primary benefit is CargoPackets.

| Action                       | Vanilla  | Bitmap | Priority_Queue |
| ---------------------------- | -------- | ------ | -------------- |
| Wentbourne CargoPacket GetNew x 10,000   | 6608us   | 38us   | 108us          |
| Wentbourne CargoPacket FreeItem x 10,000 | 7.6us    | 1.75us | 8us            |
| Xarick Vehicle GetNew x 10,000       | 113994us | 1915us | 133us          |
| Xarick Vehicle FreeItem x 10,000     | 9.3us    | 6.4us  | 13.1us         |

LordAro:
Any testing in non pathological cases as well? Don't exactly want to optimise for the worst case at the expense of everything else

PeterN:
Minor tweak, removed an extraneous header addition (clangd strikes again!) and moved the ClrBit() so that it's not called during pool cleaning. Not going to be noticeable but the test is already there so it's not adding anything.

ldpl:
As these timings are way below TicToc precision I did some measurements with the same code I used in #7962. Afaict error is about 10% so the diffirence between bitmap and priority queue is not somethnig that can be reliably measured within the game as individual timings start to approach a single tick territory.
| Action | Vanilla |	Bitmap	| priority_queue |
| --- | -- | -- | -- |
| CargoPacket GetNew |	1986 |	238 |	266 |
| CargoPacket FreeItem |	575 |	605	| 607 |

PeterN:
Performance analysis from loading a Reddit server 1 game.

image

"It's not worse"

rubidium42 (approved):
Given `GetNew` and `FreeItem` are going to be called roughly the same amount of times for vehicle and cargo packets per tick, I looked at the different statistics for those pairs combined and it seems like the bitmap variant is always on top in the timings over master and priority queue.
Priority queue and bitmap are quite close in the less pathological cases.
In the worst cases of the bitmap (one free slot at the start and the next at the end of the pool, and allocating both and then freeing them again), it would still be about two orders faster than the current implementation.
The worse case for the priority queue would be freeing huge amounts of vehicles (company going bankrupt), and then getting stuck with a massive priority queue. Furthermore, the more elements in the queue the slower it gets for each push/pop (O(log(n)) behaviour with worse cache locality). Even when releasing and allocating vehicles from a small section  of the priority queue.

So for me the constrained memory usage with companies going bankrupt, the slightly better performance in the normal case over priority queues and the basically guaranteed at least order of magnitude better performance over the current solution at the cost of one bit per pool element seems like a good trade off.

PR #12060, opened 2024-02-11
Fix #12057: Prevent NewGRF industries and houses producing/accepting a cargo type multiple times.

### LINKED ISSUE #12057: [Crash]: Assert when clicking on Industry Chain button
### Version of OpenTTD

14.0-beta3

### Steps to reproduce

industry_chain_crash

This is a tricky crash to reproduce. Even in the attached save, it only crashes for some industries and not others of the same type.

It does not seem to be connected to the industry ID, because in the crash save I get the assertion for one Steel Mill (but not the other) as well as a Building Materials Factory.

It also happens when selecting the industry chain from the Industry Chain GUI and the Select Industry dropdown.

1. Play a game with Improved Town Industries 2.9
2. Find a Steel Mill or Building Materials Factory
3. Click on the industry, then on the Industry Chain button
4. Maybe get an assertion failure.

Here is the Steel Mill [NML source](https://github.com/2TallTyler/improved_town_industries/blob/master/src/steel_mill.nml).

### Upload crash files

[crash20240211131433.zip](https://github.com/OpenTTD/OpenTTD/files/14232483/crash20240211131433.zip)

PeterN:
## Motivation / Problem

It's possible via NewGRF for an industry or house to specify in its produce or accept slots the same cargo multiple times. This breaks some game assumptions, at least the Industry Chain window.

## Description

Detect when multiple slots contain the same cargo type. Any extra occurrences will be set to INVALID_CARGO.

## Limitations

Not sure if the NewGRF spec mentions this case at all, but it doesn't seem to make sense to allow it.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -9399,11 +9399,16 @@ static void FinaliseHouseArray()
 			hs->building_flags = TILE_NO_FLAG;
 		}
 
+		if (!hs->enabled) continue;
+
 		/* Apply default cargo translation map for unset cargo slots */
-		for (uint i = 0; i < lengthof(hs->accepts_cargo); ++i) {
-			if (!IsValidCargoID(hs->accepts_cargo[i])) hs->accepts_cargo[i] = GetCargoIDByLabel(hs->accepts_cargo_label[i]);
+		for (auto it = std::begin(hs->accepts_cargo); it != std::end(hs->accepts_cargo); ++it) {
+			auto slot = std::distance(std::begin(hs->accepts_cargo), it);
+			if (!IsValidCargoID(*it)) *it = GetCargoIDByLabel(hs->accepts_cargo_label[slot]);
+			/* If cargo is accepted already, disable this slot. */
+			if (IsValidCargoID(*it) && std::find(std::begin(hs->accepts_cargo), it, *it) != it) *it = INVALID_CARGO;
 			/* Disable acceptance if cargo type is invalid. */
-			if (!IsValidCargoID(hs->accepts_cargo[i])) hs->cargo_acceptance[i] = 0;
+			if (!IsValidCargoID(*it)) hs->cargo_acceptance[slot] = 0;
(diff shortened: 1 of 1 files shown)


src/newgrf.cpp
```diff
@@ -9478,21 +9483,30 @@ static void FinaliseIndustriesArray()
...
+			continue;
 		}
 
 		/* Apply default cargo translation map for unset cargo slots */
-		for (uint i = 0; i < lengthof(indsp.produced_cargo); ++i) {
-			if (!IsValidCargoID(indsp.produced_cargo[i])) indsp.produced_cargo[i] = GetCargoIDByLabel(GetActiveCargoLabel(indsp.produced_cargo_label[i]));
+		for (auto it = std::begin(indsp.produced_cargo); it != std::end(indsp.produced_cargo); ++it) {
+			if (!IsValidCargoID(*it)) *it = GetCargoIDByLabel(GetActiveCargoLabel(indsp.produced_cargo_label[std::distance(std::begin(indsp.produced_cargo), it)]));
+			/* If cargo is produced already, disable this slot. */
+			if (IsValidCargoID(*it) && std::find(std::begin(indsp.produced_cargo), it, *it) != it) *it = INVALID_CARGO;
```
TrueBrain, line 9493: Should it throw a warning to the user? On `newgrf, 1` or something? No clue if that is useful at all.

PeterN: I thought about it but `GrfMsg()` doesn't show the correct NewGRF or NFO line number here as it's after loading. Not sure.

glx22:
Looks like many cargo acceptance are modified even for vanilla.

src/newgrf.cpp
2TallTyler: I don't think the line number matters, if you specify which industry has the problem. I think a pop-up warning is a good idea to help GRF authors catch their mistakes.

PR #12067, opened 2024-02-11
Fix #10079: don't render at 1000fps if HW acceleration + vsync is requested but not active

### LINKED ISSUE #10079: [Bug]: High CPU usage when Vsync is enabled on Win32 videodriver
### Version of OpenTTD

Windows 12.2

### Expected result

With vsync enabled, FPS is automatically tied to the screen refresh rate, the processor is used in normal mode

### Actual result

изображение
image
FPS limit is removed completely, the game runs at 999+ fps, OpenTTD creates a high CPU load (one core).
Re-tested on a weaker laptop without a discrete GPU, the result is similar, ~999 fps and high load

### Steps to reproduce
1. Change videodriver in your openttd.cfg to Win32 (or run openttd with -v win32 parameter)
2. Open Game settings
3. Enable Vsync
4. Check task manager

TrueBrain:
## Motivation / Problem

Fixes #10079.

When you enable HW acceleration in-game, you get told to restart. But if you don't ....... you can also enable vsync, and the game actually tries to align with vsync. But there is no vsync in non-HW accelerated backends. So it just goes to 1000fps.

## Description

Query the driver to know whether it is hw-accelerated.

Sadly, this information was currently only available in the `DriverFactoryBase`, so it had to be duplicated into the `VideoDriver` to be available (in quick way) for `GetDrawInterval`.

## Checklist for review

### DIFF
--- a/src/video/cocoa/cocoa_v.mm
+++ b/src/video/cocoa/cocoa_v.mm
@@ -87,7 +87,8 @@ static const Dimension _default_resolutions[] = {
 };
 
 
-VideoDriver_Cocoa::VideoDriver_Cocoa()
+VideoDriver_Cocoa::VideoDriver_Cocoa(bool uses_hardware_acceleration)
+	: VideoDriver(uses_hardware_acceleration)
 {
 	this->setup         = false;
 	this->buffer_locked = false;
--- a/src/video/cocoa/cocoa_ogl.h
+++ b/src/video/cocoa/cocoa_ogl.h
@@ -23,7 +23,7 @@ class VideoDriver_CocoaOpenGL : public VideoDriver_Cocoa {
 	const char *AllocateContext(bool allow_software);
 
 public:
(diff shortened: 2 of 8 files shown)


rubidium42 (commented):
When the vtable is too slow, one can always consider making it an instance variable of `VideoDriver` and setting that in the constructor.

TrueBrain:
> When the vtable is too slow, one can always consider making it an instance variable of `VideoDriver` and setting that in the constructor.

Not sure that is actually an improvement, but for visibility, pushed it nevertheless.

src/video/cocoa/cocoa_v.mm
```diff
@@ -87,7 +87,8 @@
 };
 
 
-VideoDriver_Cocoa::VideoDriver_Cocoa()
+VideoDriver_Cocoa::VideoDriver_Cocoa(uses_hardware_acceleration)
```
rubidium42:
```suggestion
VideoDriver_Cocoa::VideoDriver_Cocoa(bool uses_hardware_acceleration)
```

rubidium42 (approved):

PR #12070, opened 2024-02-11
Fix #10500: as client, try to reconnect more than once on server restart

### LINKED ISSUE #10500: [Bug]: Restarting server does not reconnect clients
### Version of OpenTTD

13.0

### Expected result

All clients connected to server should automatically reconnect after server is restarted or made new game via rcon.

### Actual result

Error message saying that server is restarting and please wait, but there is no retry after first fail to reconnect (which is made during map creation).

Screenshot_1088

Screenshot_1089

### Steps to reproduce

1. Start multiplayer server
2. Connect some client to it
3. Send rcon `restart`

TrueBrain:
## Motivation / Problem

Fixes #10500 

In the old days, reconnecting was as simple as creating a connection to the IP:port you just came from. As the server was quick enough to have it ready to accept connections again, you just got queued till the server was ready to process your connection.

But ever since 12.0, we also have invite-codes. When a server restarts, its invite-code is invalid till the server is back up and running (including map generation). This means that clients that connected by invite-code (read: almost all of them) couldn't actually reconnect to a server that is restarting.

## Description

Make the reconnect try more than once, and make it depends on time, not ticks.

It uses a simple backoff algorithm to not retry constantly. First attempt is after 1 second, second after 2, third after 4, with a max of 30s delay.

If after ~2.5 minutes the connection is not established, show a nice error that reconnecting failed. Here I have the assumption that servers will be back in 2.5 minutes, but ofc on big maps with a lot of NewGRFs this might not be true. So we could increase that value. Not sure that is actually practical .. pretty sure most players already gave up by then.

Also moved the errors to an `WL_ERROR` instead of a `WL_CRITICAL`, as a critical stays on top, there can only be one critical (they are queued), and it is just annoying to deal with. With `WL_ERROR` it just replaces what ever `WL_ERROR` was active.

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -1598,11 +1598,30 @@ void GameLoop()
 		/* Multiplayer */
 		NetworkGameLoop();
 	} else {
-		if (_network_reconnect > 0 && --_network_reconnect == 0) {
-			/* This means that we want to reconnect to the last host
-			 * We do this here, because it means that the network is really closed */
-			NetworkClientConnectGame(_settings_client.network.last_joined, COMPANY_SPECTATOR);
+		if (_network_reconnect_attempts > 0 && std::chrono::steady_clock::now() > _network_reconnect) {
+			Debug(net, 3, "Reconnect attempt #{}", _network_reconnect_attempts);
+
+			_network_reconnect_attempts--;
+			_network_reconnect_backoff = std::min(_network_reconnect_backoff * 2, std::chrono::seconds(30));
+			_network_reconnect = std::chrono::steady_clock::now() + _network_reconnect_backoff;
+
+			if (_network_reconnect_attempts == 0) {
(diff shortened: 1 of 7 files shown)


src/openttd.cpp
```diff
@@ -1598,11 +1598,30 @@ void GameLoop()
 		/* Multiplayer */
 		NetworkGameLoop();
 	} else {
-		if (_network_reconnect > 0 && --_network_reconnect == 0) {
-			/* This means that we want to reconnect to the last host
-			 * We do this here, because it means that the network is really closed */
-			NetworkClientConnectGame(_settings_client.network.last_joined, COMPANY_SPECTATOR);
+		if (_network_reconnect_attempts > 0 && std::chrono::steady_clock::now() > _network_reconnect) {
+			Debug(net, 3, "Reconnect attempt #{}", _network_reconnect_attempts);
```
glx22, line 1602: The ouput will be
```
Reconnect attempt #8
Reconnect attempt #7
...
```
which feels weird to me.

TrueBrain:
You can't abort the reconnecting. That is just wrong. This needs a bit more work and attention :)

PR #12077, opened 2024-02-12
Remove: setting_newgame console command

### LINKED ISSUE #12059: [Bug]: setting_newgame console commands trigger post_cb callbacks which implicitly assume settings
### Version of OpenTTD

master

### Expected result

Use of the setting_newgame console command shouldn't effect changes as if the active settings were changed.

### Actual result

Many of the post_cb implementations use a comparison between _game_mode and GM_MENU to decide whether the change should be effected on the running game and/or on _settings.

Use of setting_newgame as a multiplayer client can cause multiplayer desyncs or other state divergences.

### Steps to reproduce

Use the `setting_newgame` console command.
Ones with interesting effects include: economy.minutes_per_calendar_year, service interval settings and freeform_edges.

JGRennison:
## Motivation / Problem

#12059

None of the pre and post callbacks which do different things for the active game and game menu cases correctly handle changing settings in this way (the interface offers no means to do so).

## Description

Remove setting_newgame console command
Remove now unused force_newgame parameters

Fixes: #12059

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -2163,25 +2163,6 @@ DEF_CONSOLE_CMD(ConSetting)
 	return true;
 }
 
-DEF_CONSOLE_CMD(ConSettingNewgame)
-{
-	if (argc == 0) {
-		IConsolePrint(CC_HELP, "Change setting for the next game. Usage: 'setting_newgame <name> [<value>]'.");
-		IConsolePrint(CC_HELP, "Omitting <value> will print out the current value of the setting.");
-		return true;
-	}
-
-	if (argc == 1 || argc > 3) return false;
-
-	if (argc == 2) {
-		IConsoleGetSetting(argv[1], true);
-	} else {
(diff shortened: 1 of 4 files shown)


frosch123:
I don't run dedicated server, but I assumed you need this command for them?

If server is running, and you want to start a new game with different settings via rcon:
* Change settings via `setting_newgame`
* Start `newgame`

Is there a different way to do this via rcon?

agentw4b:
Please don't break that command, I'd like to use it on the server.
If it only overrides the parameters in openttd.cfg but doesn't interfere with the active game, there shouldn't be a problem. So please fix it so it doesn't interfere with the active game, thanks in advance...

JGRennison:
Yes, probably better to just make it work.

agentw4b:
Maybe it would be good to introduce another openttdnew.cfg to the openttd.cfg and it could write the settings_newgame parameters there and the file would replace the active one on restart.

PR #12087, opened 2024-02-13
Add: [Script] GetAirportNumTerminals

SamuXarick:
## Motivation / Problem
There's no method to get the number of terminals of an airport type.

## Description
First, move `GetNumTerminals` away from aircraft_cmd.cpp into `airport.cpp` and declare it in `airport.h`.
Then add `ScriptAirport::GetAirportNumTerminals` to AI/GS to get the number of terminals of a given `AirportType`.

## Limitations
I assume I didn't break anything with the move process.

## Checklist for review

### DIFF
--- a/src/airport.h
+++ b/src/airport.h
@@ -198,6 +198,7 @@ public:
 
 
 const AirportFTAClass *GetAirport(const uint8_t airport_type);
+uint GetNumTerminals(const AirportFTAClass *apc);
 uint8_t GetVehiclePosOnBuild(TileIndex hangar_tile);
 
 #endif /* AIRPORT_H */
--- a/regression/regression/main.nut
+++ b/regression/regression/main.nut
@@ -225,6 +225,7 @@ function Regression::Airport()
 		print("  GetAirportHeight(" + i + "):              " + AIAirport.GetAirportHeight(i));
 		print("  GetAirportCoverageRadius(" + i + "):      " + AIAirport.GetAirportCoverageRadius(i));
 		print("  GetAirportNumHelipads(" + i + "):         " + AIAirport.GetAirportNumHelipads(i));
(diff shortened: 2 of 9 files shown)


src/airport.h
```diff
@@ -181,6 +181,22 @@ struct AirportFTAClass {
 	byte nofelements;                     ///< number of positions the airport consists of
 	const byte *entry_points;             ///< when an airplane arrives at this airport, enter it at position entry_point, index depends on direction
 	byte delta_z;                         ///< Z adjustment for helicopter pads
+
+	/**
+	 * Get the number of terminals at the airport.
+	 * @param apc Airport description.
```
2TallTyler: This parameter no longer exists.
```suggestion
```

SamuXarick: oops. Bad copy paste

src/airport.h
```diff
@@ -181,6 +181,22 @@ struct AirportFTAClass {
...
+	uint GetNumTerminals() const
+	{
+		uint num = 0;
+
+		if (terminals != nullptr) {
+			for (uint i = terminals[0]; i > 0; i--) num += terminals[i];
+		}
+
+		return num;
+	}
```
PeterN: What purpose does making this a member function serve?

SamuXarick: Seemed to be the best way to go about it, now that it is called from two places.

SamuXarick: I just moved it away from the FTA to somewhere else.

frosch123:
The API should not expose any more internals of the airport state machines.
* If we ever get modular/tile-based airports, all these functions are a compatibility hassle.
* None of these functions appear useful to an AI for me. If an AI wants to rank the efficiency of the different airports, it should rather hard-code a list of types.

PR #12088, opened 2024-02-14
Add: [Script] GetAirportPlaneCompatibility

SamuXarick:
## Motivation / Problem
There's no way to know for sure if an aircraft can use an airport before building them.

## Description
Add `ScriptAirport::GetAirportPlaneCompatibility` to check whether a type of plane can use a type of airport.
The return value is from an enum:
 `ScriptAirport::APC_INVALID` for "Preconditions not met.",
 `ScriptAirport::APC_INCOMPATIBLE` for "The aircraft can't use the airport at all.",
 `ScriptAirport::APC_COMPATIBLE` for "The aircraft can use the airport in a safe manner.",
 `ScriptAirport::APC_SHORT_RUNWAY` for "The aircraft can use the airport but the runway is too short."

## Limitations
I decided to not use a boolean just so one function would be sufficient to get the answers. Maybe it's the wrong approach, the alternative was multiple boolean functions, one to check helicopters, one other to check planes and another to check runway.

## Checklist for review

### DIFF
--- a/src/script/api/script_airport.cpp
+++ b/src/script/api/script_airport.cpp
@@ -179,3 +179,27 @@
 
 	return ::AirportSpec::Get(type)->fsm->num_helipads;
 }
+
+/* static */ ScriptAirport::AirportPlaneCompatibility ScriptAirport::GetAirportPlaneCompatibility(AirportType airport_type, PlaneType plane_type)
+{
+	if (!IsAirportInformationAvailable(airport_type)) return APC_INVALID;
+
+	auto flags = ::AirportSpec::Get(airport_type)->fsm->flags;
+	switch (plane_type) {
+		case PT_HELICOPTER:
+			if (flags.Test(::AirportFTAClass::Flag::Helicopters)) return APC_COMPATIBLE;
+			break;
+		case PT_SMALL_PLANE:
+			if (flags.Test(::AirportFTAClass::Flag::Airplanes)) return APC_COMPATIBLE;
+			break;
+		case PT_BIG_PLANE:
+			if (flags.Test(::AirportFTAClass::Flag::Airplanes)) {
+				if (flags.Test(::AirportFTAClass::Flag::ShortStrip)) return APC_SHORT_RUNWAY;
+				return APC_COMPATIBLE;
+			}
+			break;
+		default:
+			return APC_INVALID;
+	}
+	return APC_INCOMPATIBLE;
+}
--- a/src/script/api/script_airport.hpp
+++ b/src/script/api/script_airport.hpp
@@ -50,6 +50,18 @@ public:
 	};
 
 	/**
+	 * Airport plane compatibility return values.
+	 */
+	enum AirportPlaneCompatibility {
+		/* Note: only one of these values is returned by ScriptAirport::GetAirportPlaneCompatibility */
+		APC_INCOMPATIBLE, ///< The aircraft can't use the airport at all.
+		APC_COMPATIBLE, ///< The aircraft can use the airport in a safe manner.
+		APC_SHORT_RUNWAY, ///< The aircraft can use the airport but the runway is too short.
+
+		APC_INVALID = -1, ///< Preconditions not met.
+	};
+
+	/**
 	 * Checks whether the given AirportType is valid and available.
 	 * @param type The AirportType to check.
 	 * @return True if and only if the AirportType is valid and available.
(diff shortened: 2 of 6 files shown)


src/script/api/script_airport.cpp
```diff
@@ -170,3 +170,28 @@
...
+			if ((flags & ::AirportFTAClass::AIRPLANES) != 0) return 1;
+			break;
+		case PT_BIG_PLANE:
+			if ((flags & ::AirportFTAClass::AIRPLANES) != 0) {
+				if ((flags & ::AirportFTAClass::SHORT_STRIP) != 0) return 2;
+				return 1;
+			}
+			break;
+		default:
+			NOT_REACHED();
```
2TallTyler:
```suggestion
			return -1;
```

src/script/api/script_airport.cpp
```diff
@@ -170,3 +170,28 @@
 
 	return (int64_t)GetMaintenanceCostFactor(type) * _price[PR_INFRASTRUCTURE_AIRPORT] >> 3;
 }
+
+/* static */ SQInteger ScriptAirport::CanPlaneUseAirport(AirportType airport_type, PlaneType plane_type)
+{
+	if (!IsAirportInformationAvailable(airport_type)) return -1;
+	if (plane_type != PT_HELICOPTER && plane_type != PT_SMALL_PLANE && plane_type != PT_BIG_PLANE) return -1;
```
2TallTyler: There's no reason to check the `plane_type` before the switch, IMHO. Early returns are useful to avoid unnecessary computing or an unsafe check, but neither of those happen here and this just adds complexity. You need to deal with the "none of the above" case in the switch already, so do it all there instead of splitting it.
```suggestion

```

2TallTyler (commented):
Code looks fine, but returning a magic number is never really ideal. I wonder if creating a new enum to return would be better, something like `AirportCompatibility`.

glx22:
Yeah not a fan of the magic return values.

2TallTyler (dismissed):
Sure.

src/script/api/script_airport.hpp
```diff
@@ -49,6 +49,18 @@ class ScriptAirport : public ScriptObject {
...
+	 * Airport plane compatibility return values.
+	 */
+	enum AirportPlaneCompatibility {
+		/* Note: only one of these values is returned by ScriptAirport::GetAirportPlaneCompatibility */
+		APC_INCOMPATIBLE  =   0, ///< The aircraft can't use the airport at all.
+		APC_COMPATIBLE    =   1, ///< The aircraft can use the airport in a safe manner.
+		APC_SHORT_RUNWAY  =   2, ///< The aircraft can use the airport but the runway is too short.
+
+		APC_INVALID       =  -1, ///< Preconditions not met.
+	};
```
PeterN, line 62: We don't align like this any more, and the values are probably not needed either.

src/script/api/script_airport.cpp
```diff
@@ -179,3 +179,27 @@
 
 	return ::AirportSpec::Get(type)->fsm->num_helipads;
 }
+
+/* static */ ScriptAirport::AirportPlaneCompatibility ScriptAirport::GetAirportPlaneCompatibility(AirportType airport_type, PlaneType plane_type)
+{
+	if (!IsAirportInformationAvailable(airport_type)) return APC_INVALID;
+
+	const auto flags = ::AirportSpec::Get(airport_type)->fsm->flags;
```
PeterN: No need for`const`

src/script/api/script_airport.hpp
2TallTyler: The raw values are used in the regression test results, although I guess if anybody changes them regression will fail. 🙂

2TallTyler (dismissed):

src/script/api/script_airport.hpp
```diff
@@ -49,6 +49,18 @@ class ScriptAirport : public ScriptObject {
 		PT_INVALID       =  -1, ///< An invalid PlaneType
 	};
 
+	/**
+	 * Airport plane compatibility return values.
+	 */
+	enum AirportPlaneCompatibility {
+		/* Note: only one of these values is returned by ScriptAirport::GetAirportPlaneCompatibility */
```
Rito13, line 56: Why not include that message in the doxygen comment above with `@note` instead of `Note:`?

PR #12090, opened 2024-02-14
Change: Switch road stop window between 6 or 2 orientations depending on road stop type.

PeterN:
## Motivation / Problem

Custom road stops can have 6 (4 drive-in bay and 2 drive-through) or just 2 (drive-through only) orientations, and the window currently masks the drive-in bay buttons if a road stop only provides drive-through layouts.

Road stop with 6 orientations:

image

Road stop with 2 orientations:

image

## Description

Change the window to switch between a 6-button layout and a 2-button layout as appropriate. The window itself is not resized as that could make selecting different types fiddly, so the 2-button layout is padded instead.

Road stop with 2 orientations:

image

## Checklist for review

### DIFF
--- a/src/road_gui.cpp
+++ b/src/road_gui.cpp
@@ -1129,8 +1129,15 @@ private:
 
 	void CheckOrientationValid()
 	{
-		if (_roadstop_gui_settings.orientation >= DIAGDIR_END) return;
 		const RoadStopSpec *spec = RoadStopClass::Get(_roadstop_gui_settings.roadstop_class)->GetSpec(_roadstop_gui_settings.roadstop_type);
+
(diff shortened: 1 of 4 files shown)


src/road_gui.cpp
```diff
@@ -1129,8 +1129,13 @@ struct BuildRoadStationWindow : public PickerWindowBase {
 
 	void CheckOrientationValid()
 	{
-		if (_roadstop_gui_settings.orientation >= DIAGDIR_END) return;
+		this->RaiseWidget(WID_BROS_STATION_NE + _roadstop_gui_settings.orientation);
```
2TallTyler: I assume raising and then lowering the widget has to do with setting the displayed plane...but this could use a comment explaining what's going on for future readers. 🙂

2TallTyler (approved):

PR #12095, opened 2024-02-15
Codechange: use {HEX} instead of custom hex format and {(RAW_)STRING}

rubidium42:
## Motivation / Problem

We have a nice `{HEX}` string command, but somehow we insist in formatting the string ourselves and then using `SetDParamStr`.

## Description

Use `{HEX}` over `{RAW_STRING}`.
This does remove forcing the number to be at least 4 digits, which has been insufficient for the bigger map sizes since they have been introduced.

## Limitations

There is some discussion whether it should be just decimal or exist at all...

## Checklist for review


TrueBrain:
Fixed by #12104.

PR #12097, opened 2024-02-16
Change: [Script] Improve some ScriptAirport return values for AT_OILRIG airport type

SamuXarick:
## Motivation / Problem
Not really a problem, but I noticed some ScriptAirport functions could retrieve valid values for oil rigs without much effort.

Before:
image

## Description
Improve some `ScriptAirport `return values for `AT_OILRIG` `AirportType`

Affected functions are:
- `IsAirportInformationAvailable`
- `GetAirportWidth`
- `GetAirportHeight`
- `GetAirportCoverageRadius`
- `GetNumHangars`
- `GetHangarOfAirport`
- `GetNoiseLevelIncrease`
- `GetNearestTown`
- `GetMaintenanceCostFactor`
- `GetMonthlyMaintenanceCost`
- `GetAirportNumHelipads`

Add `AT_OILRIG` to the `AirportType` enum even though it can't be built.

Add an `AirportTileTable` for oilrig in `AirportSpec` definitions.

After:
image

## Checklist for review

### DIFF
--- a/src/script/api/script_airport.cpp
+++ b/src/script/api/script_airport.cpp
@@ -24,7 +24,7 @@
 
 /* static */ bool ScriptAirport::IsAirportInformationAvailable(AirportType type)
 {
-	return type >= 0 && type < (AirportType)NUM_AIRPORTS && AirportSpec::Get(type)->enabled;
+	return type == AT_OILRIG || (type >= 0 && type < (AirportType)NUM_AIRPORTS && AirportSpec::Get(type)->enabled);
 }
 
 /* static */ Money ScriptAirport::GetPrice(AirportType type)
@@ -67,6 +67,7 @@
 {
 	if (!IsAirportInformationAvailable(type)) return -1;
 
+	if (type == AT_OILRIG && !_settings_game.station.serve_neutral_industries) return (uint)CA_NONE;
 	return _settings_game.station.modified_catchment ? ::AirportSpec::Get(type)->catchment : (uint)CA_UNMODIFIED;
 }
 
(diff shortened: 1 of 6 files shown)


rubidium42:
Won't it be much simpler to just make `IsAirportInformationAvailable` also return true for `AT_OILRIG` and define the actual `AirportTileTable` in `_origin_airport_specs`? Maybe "abuse" the heliport for the table like is already done for many of the other tables of the oilrig.
That would remove a lot of special casing in this diff as much of the data you special case is just available in the specs table. Since the oil rig is not enabled at any time it should not harm NewGRFs either.

src/script/api/script_airport.cpp
```diff
@@ -136,7 +137,7 @@
 	const AirportSpec *as = ::AirportSpec::Get(type);
 	if (!as->IsWithinMapBounds(0, tile)) return -1;
 
-	if (_settings_game.economy.station_noise_level) {
+	if (_settings_game.economy.station_noise_level || type == AT_OILRIG) {
```
rubidium42: What's special about oil rigs that they always need to go into this branch?

src/script/api/script_airport.cpp
```diff
@@ -67,7 +67,8 @@
 {
 	if (!IsAirportInformationAvailable(type)) return -1;
 
-	return _settings_game.station.modified_catchment ? ::AirportSpec::Get(type)->catchment : (uint)CA_UNMODIFIED;
+	const auto coverage = _settings_game.station.modified_catchment ? ::AirportSpec::Get(type)->catchment : (uint)CA_UNMODIFIED;
+	return (type != AT_OILRIG || _settings_game.station.serve_neutral_industries) ? coverage : 0;
```
rubidium42:
```suggestion
	if (type == AT_OILRIG && !_settings_game.station.serve_neutral_industries) return 0;
	return _settings_game.station.modified_catchment ? ::AirportSpec::Get(type)->catchment : (uint)CA_UNMODIFIED;
```
Why calculate the coverage when you know it's going to be zero anyway?

src/script/api/script_airport.cpp
SamuXarick: They don't produce noise.

SamuXarick: https://gist.github.com/SamuXarick/db932246d86aecd117affca3c023d93e - which method do you want me to use?

image

PR #12103, opened 2024-02-17
Codechange: Skip non-water water region patches in neigbor search

### LINKED ISSUE #11862: [Bug]: High level pathfinder passing patches with no water as the origin.
### Version of OpenTTD

20240121-master-g0841978304

### Expected result

High level pathfinder always passes patches with water.

### Actual result

High level pathfinder passes patches with no water as the origin. When getting neighbours with a patch like that, it eithers gets those where both regions have 1 patch, or none.

### Steps to reproduce

https://github.com/OpenTTD/OpenTTD/blob/c7d5cedc7b094a698b65a6d6c600462d559b0eb5/src/pathfinder/yapf/yapf_ship_regions.cpp#L212-L213

Load savegame, when the ship departs from the dock, it creates a scenario where the order type is changed to OT_LEAVING, but v->dest_tile points to the slope of the dock, which has no water.
[ships crashing water regions.zip](https://github.com/OpenTTD/OpenTTD/files/14002616/ships.crashing.water.regions.zip)

Kuhnovic:
## Motivation / Problem / Description

This address some of the points mentioned in #11862 but doesn't fully solve it. But it generally makes the water region neighbor search more robust. If an invalid water region patch is provided (in English: a patch of not-water is provided) then the functions simply return and do nothing. This makes them more resilient against edge cases such as what happens in #11862.

## Limitations

None

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -22,7 +22,6 @@
 
 using TWaterRegionTraversabilityBits = uint16_t;
 constexpr TWaterRegionPatchLabel FIRST_REGION_LABEL = 1;
-constexpr TWaterRegionPatchLabel INVALID_WATER_REGION_PATCH = 0;
 
 static_assert(sizeof(TWaterRegionTraversabilityBits) * 8 == WATER_REGION_EDGE_LENGTH);
 static_assert(sizeof(TWaterRegionPatchLabel) == sizeof(byte)); // Important for the hash calculation.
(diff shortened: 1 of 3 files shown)


2TallTyler (dismissed):
Seems sensible to me. 

Also, the way you wrote the description using the `fixes` keyword means that merging this PR will close the linked issue, that you say is "not really" fixed. If you don't want that, please change your description slightly. 😉

2TallTyler (approved):

PR #12104, opened 2024-02-17
Change: Show tile index as decimal number in land info window

Kuhnovic:
## Motivation / Problem

When debugging I often inspect tile indices. Visual Studio displays these as decimal numbers, but the land area information tool shows this number in hex. And my brain's hex-to-decimal-converter is broken.

## Description

I changed the hex number to a decimal number:

image

## Limitations

Pitchforks, [spacebar heating](https://xkcd.com/1172/), and running the risk of being trolled by @TrueBrain on Discord.

All kidding aside, there might be legitimate reasons to have hex that I'm not aware of.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -3078,7 +3078,8 @@ STR_LAND_AREA_INFORMATION_TRAM_OWNER                            :{BLACK}Tramway 
 STR_LAND_AREA_INFORMATION_RAIL_OWNER                            :{BLACK}Railway owner: {LTBLUE}{STRING1}
 STR_LAND_AREA_INFORMATION_LOCAL_AUTHORITY                       :{BLACK}Local authority: {LTBLUE}{STRING1}
 STR_LAND_AREA_INFORMATION_LOCAL_AUTHORITY_NONE                  :None
-STR_LAND_AREA_INFORMATION_LANDINFO_COORDS                       :{BLACK}Coordinates: {LTBLUE}{NUM} x {NUM} x {NUM} ({RAW_STRING})
+STR_LAND_AREA_INFORMATION_LANDINFO_COORDS                       :{BLACK}Coordinates: {LTBLUE}{NUM} x {NUM} x {NUM}
+STR_LAND_AREA_INFORMATION_LANDINFO_INDEX                        :{BLACK}Tile index: {LTBLUE}{NUM} ({HEX})
 STR_LAND_AREA_INFORMATION_BUILD_DATE                            :{BLACK}Built/renovated: {LTBLUE}{DATE_LONG}
 STR_LAND_AREA_INFORMATION_STATION_CLASS                         :{BLACK}Station class: {LTBLUE}{STRING}
 STR_LAND_AREA_INFORMATION_STATION_TYPE                          :{BLACK}Station type: {LTBLUE}{STRING}
--- a/src/misc_gui.cpp
+++ b/src/misc_gui.cpp
@@ -118,7 +118,7 @@ public:
 #else
 #	define LANDINFOD_LEVEL 1
 #endif
-		Debug(misc, LANDINFOD_LEVEL, "TILE: 0x{:x} ({},{})", (TileIndex)tile, TileX(tile), TileY(tile));
+		Debug(misc, LANDINFOD_LEVEL, "TILE: {0} (0x{0:x}) ({1},{2})", (TileIndex)tile, TileX(tile), TileY(tile));
 		Debug(misc, LANDINFOD_LEVEL, "type   = 0x{:x}", tile.type());
 		Debug(misc, LANDINFOD_LEVEL, "height = 0x{:x}", tile.height());
 		Debug(misc, LANDINFOD_LEVEL, "m1     = 0x{:x}", tile.m1());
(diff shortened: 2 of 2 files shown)


src/misc_gui.cpp
```diff
@@ -118,17 +118,17 @@ class LandInfoWindow : public Window {
...
-		Debug(misc, LANDINFOD_LEVEL, "height = 0x{:x}", tile.height());
-		Debug(misc, LANDINFOD_LEVEL, "m1     = 0x{:x}", tile.m1());
-		Debug(misc, LANDINFOD_LEVEL, "m2     = 0x{:x}", tile.m2());
-		Debug(misc, LANDINFOD_LEVEL, "m3     = 0x{:x}", tile.m3());
-		Debug(misc, LANDINFOD_LEVEL, "m4     = 0x{:x}", tile.m4());
-		Debug(misc, LANDINFOD_LEVEL, "m5     = 0x{:x}", tile.m5());
-		Debug(misc, LANDINFOD_LEVEL, "m6     = 0x{:x}", tile.m6());
-		Debug(misc, LANDINFOD_LEVEL, "m7     = 0x{:x}", tile.m7());
-		Debug(misc, LANDINFOD_LEVEL, "m8     = 0x{:x}", tile.m8());
+		Debug(misc, LANDINFOD_LEVEL, "TILE: {:#x} ({},{})", (TileIndex)tile, TileX(tile), TileY(tile));
```
Kuhnovic: This the fmt way to achieve 0x... formatting, search for "hex" on this page https://hackingcpp.com/cpp/libs/fmt.html

TrueBrain:
As mentioned on Discord, with hex you can see on the nibbles what is x and y, which can help greatly. Other than that, one is as arbitrary as the other.

Changing it because MSVC visualizes it in one way might be a very thin motivation, as the next person can say: my MSVC shows it in hex! You also have to get a bit used to hex when you are working with computers, and especially with tiles.

That said, I don't give a single fuck what we do here, and I am fine with whatever. Just wanted to add some arguments that passed by on Discord.

src/misc_gui.cpp
rubidium42: I know, but did you know that `{:#04x}` with 0 yields `0x00`? So you're not getting 4 digits, you're getting 4 characters.
That's why I reverted to `0x{:x}`, even though I know there's a flag adding the `0x`. Just because with the flag it will mess with you when setting the number of digits.

rubidium42:
> As mentioned on Discord, with hex you can see on the nibbles what is x and y, which can help greatly. Other than that, one is as arbitrary as the other.

But only when your log2 map size X is a multiple of 8 (e.g. 256 or 4096). In all other cases bits of the X and Y axis will be in the same nibble. So the argument made sense back when 256x256 was the only choice in map size.

I do agree that MSVC visualising something in a particular manner with given configuration isn't a really strong argument.

Showing hexadecimal to our end users is very uncommon though. The only other place is with NewGRF IDs and MD5 checksums for which the hexadecimal representation is the de facto default even outside OpenTTD. Not to forget the NewGRF debugging, but that's definitely not a standard user and in NewGRF hexadecimal is basically standard.

The only reason to show the number in the UI at all, for an end user's point of view, is to provide the parameter for the `scrollto` command. Since clicking on the tile will also generate `Debug` output with hexadecimal, and that stays at it is, the developers are actually catered for in that output if they really need a hexadecimal output.


Only question that remains is whether we want to throw all translations away for changing `{STRING}` with `{NUM}`.
`for i in src/lang/*.txt; do sed '/^STR_LAND_AREA_INFORMATION_LANDINFO_COORDS/s/STRING}/NUM}/' -i $i; done`

stormcone:
What about breaking it into 2 lines?
```
Coordinates: 25x22x1
Tile index: 5657 (0x2C19)
```

Kuhnovic:
I've implemented Snowcone's suggestion. The window looks looks this now:
image

And the log:
image

TrueBrain (dismissed):
Kinda surprised nobody approved this yet, tbh :P

But this for sure is much more clear what that number at the end means; nice :)

src/lang/english.txt
```diff
@@ -3078,7 +3078,8 @@ STR_LAND_AREA_INFORMATION_TRAM_OWNER                            :{BLACK}Tramway
 STR_LAND_AREA_INFORMATION_RAIL_OWNER                            :{BLACK}Railway owner: {LTBLUE}{STRING1}
 STR_LAND_AREA_INFORMATION_LOCAL_AUTHORITY                       :{BLACK}Local authority: {LTBLUE}{STRING1}
 STR_LAND_AREA_INFORMATION_LOCAL_AUTHORITY_NONE                  :None
-STR_LAND_AREA_INFORMATION_LANDINFO_COORDS                       :{BLACK}Coordinates: {LTBLUE}{NUM} x {NUM} x {NUM} ({RAW_STRING})
+STR_LAND_AREA_INFORMATION_LANDINFO_COORDS                       :{BLACK}Coordinates: {LTBLUE}{NUM} x {NUM} x {NUM}
+STR_LAND_AREA_INFORMATION_LANDINFO_INDEX                        :{BLACK}Tile index: {LTBLUE}{NUM} ({RAW_STRING})
```
rubidium42: What about:
```suggestion
STR_LAND_AREA_INFORMATION_LANDINFO_INDEX                        :{BLACK}Tile index: {LTBLUE}{NUM} ({HEX})
```
and then just `SetDParam(1, tile);` later on?

glx22: ~~`{0:HEX}` and no need for `SetDParam(1, tile)`~~
That is not possible actually.

Kuhnovic: That's an even cleaner solution indeed, applied it

TrueBrain (approved):

PR #12108, opened 2024-02-17
Add: Portuguese Escudo currency

jcteotonio:
## Motivation

- A good chunk of historical pre-Euro currencies are available
- Some European currencies were added during OpenTTD development that weren't originally available in TTD
- Portuguese (European / Portugal) is available as a selectable language
- Portuguese Escudo is not available

Adding to this points, there are also some Portugal scenarios / heightmaps available for download in-game, where this addition is welcome.

## Description

- Add Portuguese Escudo (PTE) as one of the game's currencies
- The exchange rate is 400$00 = £1, this value is based on the implemented rate for the Euro in-game (€2 = £1) and rate for the Euro in 2002 when the Escudo was discontinued (200$00 = €1) 
- $ is both in place of the currency symbol and as the decimal separator. Two decimal places are displayed even if the value is an integer ($00).
- If the value is written by hand the currency symbol use two strokes instead of one. This practice is not standardized. Unicode considers the one stroked $ and the two stroked version a typeface design choice. So in computer software $ was used in it's place.

## Limitations

A new string was added to the English translation, new translations are needed in other available languages.

## Checklist for review


jcteotonio:
Currency separator was removed on master, solved conflicts taking the change into acount

2TallTyler (approved):
Thank you for this, @jcteotonio. 😃 

I've tested this in the preview and it seems to work fine.

It makes sense to me that if we have a complete language translation (and Portuguese is 100% as I write this), we should also include a currency to match -- if nothing else, as a "thank you" to the translators.

PR #12131, opened 2024-02-19
Fix: Scale graph gridlines and axes with GUI scale

zephyris:
## Motivation / Problem

Gridlines and axes on the graphs do not scale with GUI scaling, making them vanishingly thin when you get to larger GUI scales.

## Description

This is a simple update to scale the gridlines and axes width with GUI scale. Should work pixel-perfectly, tested at 1x, 2x and 3x (including checking for the annoying missing pixel at corners).

graphlinescale

It scales the gridline widths with the size of the bevels rather than GUI scale - IMO this is a little more visually consistent.

Also adds scaling of graph line and point wth GUI scale. These are scaled with the GUI scale, taking into account the graph line width user setting.

## Limitations

Scales the lines down-rightwards, rather than trying to centre them.

## Checklist for review

### DIFF
--- a/src/graph_gui.cpp
+++ b/src/graph_gui.cpp
@@ -378,7 +378,9 @@ protected:
 		/* Draw the background of the graph itself. */
 		GfxFillRect(r.left, r.top, r.right, r.bottom, GRAPH_BASE_COLOUR);
 
-		/* Draw the vertical grid lines. */
+		/* Draw the grid lines. */
+		int gridline_width = WidgetDimensions::scaled.bevel.top;
+		int grid_colour = GRAPH_GRID_COLOUR;
 
 		/* Don't draw the first line, as that's where the axis will be. */
 		if (rtl) {
@@ -388,13 +390,12 @@ protected:
 			x = r.left + x_sep;
 		}
 
-		int grid_colour = GRAPH_GRID_COLOUR;
 		for (int i = 1; i < this->num_vert_lines + 1; i++) {
 			/* If using wallclock units, we separate periods with a lighter line. */
 			if (TimerGameEconomy::UsingWallclockUnits()) {
 				grid_colour = (i % 4 == 0) ? GRAPH_YEAR_LINE_COLOUR : GRAPH_GRID_COLOUR;
 			}
-			GfxFillRect(x, r.top, x, r.bottom, grid_colour);
+			GfxFillRect(x, r.top, x + gridline_width - 1, r.bottom, grid_colour);
 			x += x_sep;
 		}
 
@@ -403,20 +404,20 @@ protected:
 
 		for (int i = 0; i < (num_hori_lines + 1); i++) {
 			if (rtl) {
-				GfxFillRect(r.right + 1, y, r.right + ScaleGUITrad(3), y, GRAPH_AXIS_LINE_COLOUR);
+				GfxFillRect(r.right + 1, y, r.right + ScaleGUITrad(3), y + gridline_width - 1, GRAPH_AXIS_LINE_COLOUR);
 			} else {
-				GfxFillRect(r.left - ScaleGUITrad(3), y, r.left - 1, y, GRAPH_AXIS_LINE_COLOUR);
+				GfxFillRect(r.left - ScaleGUITrad(3), y, r.left - 1, y + gridline_width - 1, GRAPH_AXIS_LINE_COLOUR);
 			}
-			GfxFillRect(r.left, y, r.right, y, GRAPH_GRID_COLOUR);
+			GfxFillRect(r.left, y, r.right + gridline_width - 1, y + gridline_width - 1, GRAPH_GRID_COLOUR);
 			y -= y_sep;
 		}
 
 		/* Draw the y axis. */
-		GfxFillRect(r.left, r.top, r.left, r.bottom, GRAPH_AXIS_LINE_COLOUR);
+		GfxFillRect(r.left, r.top, r.left + gridline_width - 1, r.bottom + gridline_width - 1, GRAPH_AXIS_LINE_COLOUR);
 
 		/* Draw the x axis. */
 		y = x_axis_offset + r.top;
-		GfxFillRect(r.left, y, r.right, y, GRAPH_ZERO_LINE_COLOUR);
+		GfxFillRect(r.left, y, r.right + gridline_width - 1, y + gridline_width - 1, GRAPH_ZERO_LINE_COLOUR);
 
 		/* Find the largest value that will be drawn. */
 		if (this->num_on_x_axis == 0) return;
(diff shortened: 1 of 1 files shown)


frosch123:
There is a distinct setting `Thickness of lines in graph`. Should we replace that with the general GUI scale as well?

zephyris:
Fair point. The vast majority of players use the default 3px, but some have adjusted it: https://survey.openttd.org/summaries/2024/wk07#14.0-beta3-game.settings.gui.graph_line_thickness

Retain the setting, but graph line width with GUI scale too?

src/graph_gui.cpp
```diff
@@ -358,25 +359,25 @@ struct BaseGraphWindow : Window {
...
 		}
 
 		/* Draw the horizontal grid lines. */
 		y = r.bottom;
 
 		for (int i = 0; i < (num_hori_lines + 1); i++) {
-			GfxFillRect(r.left - ScaleGUITrad(3), y, r.left - 1, y, GRAPH_AXIS_LINE_COLOUR);
-			GfxFillRect(r.left, y, r.right, y, GRAPH_GRID_COLOUR);
+			GfxFillRect(r.left - ScaleGUITrad(3), y, r.left - 1, y + gridline_width - 1, GRAPH_AXIS_LINE_COLOUR);
+			GfxFillRect(r.left, y, r.right + gridline_width - 1, y + gridline_width - 1, GRAPH_GRID_COLOUR);
```
TrueBrain, line 411: If you are going to `gridline_width - 1` everything, why not just do it once at the start, and not have to do it everywhere else? :)

zephyris: I'd argue it's better for readability, as the drawn line width is `gridline_width` pixels wide. The `- 1` is only needed because of `GfxFillRect` behaviour.

But, not a strong opinion!

src/graph_gui.cpp
```diff
@@ -348,6 +348,7 @@ struct BaseGraphWindow : Window {
 		GfxFillRect(r.left, r.top, r.right, r.bottom, GRAPH_BASE_COLOUR);
 
 		/* Draw the vertical grid lines. */
```
TrueBrain: Nitpick: this comment was already weird, and it is now even weirder. The next statement does not draw the vertical grid lines :D Maybe another newline? Dunno .. it is just a weird comment .. the rest should have been in a function looking at the comment :P

src/graph_gui.cpp
TrueBrain: The longer I look at the original code, the more weirded out I am. It feels very odd, in the way it is structured. And that makes me even more confused about the "sometimes - 1 but not always" of the old code. But that is not your problem, so I am not going to bother you with it :)

I am going to believe you on your word, that this is easier to understand, and let's leave it as it is :)

PeterN: "sometimes - 1 but not always" is because sometimes we have a left and right, and sometimes we have a left and width. `right` is always `left + width - 1`.

src/graph_gui.cpp
```diff
@@ -434,8 +435,8 @@ struct BaseGraphWindow : Window {
 			}
 		}
 
-		/* draw lines and dots */
-		uint linewidth = _settings_client.gui.graph_line_thickness;
+		/* Draw lines and dots. */
+		uint linewidth = ScaleGUITrad(_settings_client.gui.graph_line_thickness);
 		uint pointoffs1 = (linewidth + 1) / 2;
 		uint pointoffs2 = linewidth + 1 - pointoffs1;
```
PeterN: Should the size of the point be scaled too, instead just 1 pixel extra?

src/graph_gui.cpp
zephyris: In my experience this type of code always gets a bit messy. It's the combination of line lengths becoming dependent on line width, and the behaviour of `GfxFillRect` meaning when `left == right` the width is one rather than zero.

src/graph_gui.cpp
```diff
@@ -347,36 +347,37 @@ struct BaseGraphWindow : Window {
...
 		}
 
 		/* Draw the y axis. */
-		GfxFillRect(r.left, r.top, r.left, r.bottom, GRAPH_AXIS_LINE_COLOUR);
+		GfxFillRect(r.left, r.top, r.left + gridline_width - 1, r.bottom + gridline_width - 1, GRAPH_AXIS_LINE_COLOUR);
 
 		/* Draw the x axis. */
 		y = x_axis_offset + r.top;
-		GfxFillRect(r.left, y, r.right, y, GRAPH_ZERO_LINE_COLOUR);
+		GfxFillRect(r.left, y, r.right + gridline_width - 1, y + gridline_width - 1, GRAPH_ZERO_LINE_COLOUR);
```
PeterN, line 420: Not sure the right extent of the horizontal grid line should be extended. It's only the thickness of the line that needs to be adjusted.

Same for the y-axis, although there might be an exception if the gridlines meet at the bottom left corner.

Similar situations for the other lines, depending on horizontal vs vertical.

zephyris: I have thought about this and I think it is needed. It took me quite a few iterations to get all of these line extents pixel-perfect, and stems from the behaviour that `GfxFillRect(x, y, x, y)` draws 1x1px rather than nothing.

The horizontal lines are drawn over the vertical, and to draw the horizontal lines over the _final_ vertical line currently uses `r.right`. This gives a line 1px longer than you might expect (`length = r.right - r.left + 1`), and that extra 1px covers a 1px-wide final vertical line.

Based on that logic, to preserve the current behaviour with `gridline_width = 1`, `r.right + gridline_width - 1 = r.right + 1 - 1 = r.right`. So, mine is the naive correct solution for larger line widths. And, similar logic applies to the vertical extent of vertical lines.

(In this specific case, leaving it as `r.right` would partially cover the vertical line for wider lines. At, say, 4x GUI scale a 1px cap poking into the 4px wide line. The alternative is `r.right - 1`, which would remove the overlap of the final vertical line.)

src/graph_gui.cpp
```diff
@@ -376,7 +376,9 @@ struct BaseGraphWindow : Window {
 		/* Draw the background of the graph itself. */
 		GfxFillRect(r.left, r.top, r.right, r.bottom, GRAPH_BASE_COLOUR);
 
-		/* Draw the vertical grid lines. */
+		/* Draw the grid lines. */
+		int gridline_width = WidgetDimensions::scaled.bevel.top;
```
PeterN, line 382: This might've been mentioned before, apologies if so. Should we pick `bevel` or `fullbevel` here? `bevel` is dependent on the "Scale bevels" option, whereas `fullbevel` is always scaled by interface scale.

zephyris: I intentionally chose `bevel`, because I thought it looked nice to have skinny gridlines to match the skinny bevels.

2TallTyler (approved):
#YOLO, also @zephyris is good about fixing any bugs they make. 🙂

PR #12137, opened 2024-02-20
Fix #12098: Don't allow unbunching at "nearest depot"

### LINKED ISSUE #12098: [Bug]: Depot unbunching should be disallowed for "nearest depot"
### Version of OpenTTD

14.0-beta3

### Expected result

Trying to set a "go to nearest depot" order as an unbunching point should not be allowed, and cause an error message.

### Actual result

The action is allowed, but unbunching may not work properly.

### Steps to reproduce

1. Give a vehicle a "go to nearest depot" order
2. Highlight that order and set the behavior as "Unbunch"
3. See that it is allowed

2TallTyler:
## Motivation / Problem

Vehicles can be told to go to the nearest depot. We should not allow setting this as an unbunching depot, because odd things could happen. Players should choose a specific depot to set for unbunching.

## Description

Block this action and give a descriptive error.

Closes #12098.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -5113,6 +5113,7 @@ STR_ERROR_UNBUNCHING_NO_FULL_LOAD                               :{WHITE}... cann
 STR_ERROR_UNBUNCHING_NO_UNBUNCHING_FULL_LOAD                    :{WHITE}... cannot unbunch a vehicle with a full load order
 STR_ERROR_UNBUNCHING_NO_CONDITIONAL                             :{WHITE}... cannot use conditional orders when vehicle has an unbunching order
 STR_ERROR_UNBUNCHING_NO_UNBUNCHING_CONDITIONAL                  :{WHITE}... cannot unbunch a vehicle with a conditional order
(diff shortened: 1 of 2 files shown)


SamuXarick:
What about when the order list has another non-unbunch depot order with go to nearest, or go to depot and stop?

2TallTyler:
Those are not a problem. "Go to nearest depot" is no more problematic than a player modifying their roads/rails/canals to change the average trip duration, and vehicles have their unbunching data cleared when they are sent to a depot, or started/stopped.

SamuXarick:
Are you saying it can't use go to nearest depot on an unbunch order, but it can on a non-unbunch order?
image

2TallTyler:
Correct. 

The case I'm trying to avoid is having vehicles in two separate depots waiting on the same unbunching departure time, which can easily happen with "nearest depot" and multi-tile stations. This shouldn't break unbunching (feel free to test and prove me wrong) but I can foresee players inventing some really obscure corner cases and complaining that unbunching doesn't handle it properly.

2TallTyler:
I can't think of a way this could be a problem, so let's allow it until someone finds a reason not to.

PR #12144, opened 2024-02-20
Codechange: Use FindVehiclesWithOrder when removing a road stop.

PeterN:
## Motivation / Problem

When removing a road stop tile, the road vehicle pool is iterated to find any road vehicles that are currently heading to the tile, so they can be rerouted. This can take time if there are a lot of vehicles.

## Description

Road vehicles can only have the road stop tile as their destination if it is in their order lists, and searching shared order lists can be faster than searching the vehicle pool.

Search for vehicles with the road stop in their orders, and then reroute if their destination is the removed tile.

The change is inconsequential with most saves, however when road stops are removed in bulk (company bankrupt/reset) it can add up.

| Save | Master | This PR |
| ---- | ------ | ------- |
| Wentbourne remove 1 road stop | 28µs | 6µs |
| Wentbourne `reset_company 1` | 39,422µs | 7,981µs |
| Xarick's 4kx4k remove 1 road stop | 5,380µs | 299µs |
| Xarick's 4kx4k `stop_ai 2` | 7,531,394µs | 418,717µs |

## Checklist for review

### DIFF
--- a/src/vehiclelist_func.h
+++ b/src/vehiclelist_func.h
@@ -26,7 +26,7 @@ void FindVehiclesWithOrder(VehiclePredicate veh_pred, OrderPredicate ord_pred, V
 	for (const OrderList *orderlist : OrderList::Iterate()) {
 
 		/* We assume all vehicles sharing an order list match the condition. */
-		const Vehicle *v = orderlist->GetFirstSharedVehicle();
+		Vehicle *v = orderlist->GetFirstSharedVehicle();
 		if (!veh_pred(v)) continue;
 
 		/* Vehicle is a candidate, search for a matching order. */
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -10,6 +10,7 @@
 #include "stdafx.h"
 #include "aircraft.h"
 #include "bridge_map.h"
+#include "vehiclelist_func.h"
 #include "viewport_func.h"
 #include "viewport_kdtree.h"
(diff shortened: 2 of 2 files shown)


src/vehiclelist_func.h
```diff
@@ -26,7 +26,7 @@ void FindVehiclesWithOrder(VehiclePredicate veh_pred, OrderPredicate ord_pred, V
 	for (const OrderList *orderlist : OrderList::Iterate()) {
 
 		/* We assume all vehicles sharing an order list match the condition. */
-		const Vehicle *v = orderlist->GetFirstSharedVehicle();
+		Vehicle *v = orderlist->GetFirstSharedVehicle();
```
LordAro, line 29: What's this change doing?

LordAro:
What happens when the number of orders is significantly larger than the list of vehicles? :)

src/vehiclelist_func.h
PeterN: Alowing non-const use of `v` in `veh_func`, which is a bit hidden by the templating.

PeterN:
Yup, 5000 road vehicles with 255 unshared orders each, is considerably slower this way.

Master: 80µs
This PR: 3,675µs

TrueBrain:
@LordAro let's not backport these kind of changes moments before release. Feels like asking for trouble, don't you agree? :)

rubidium42 (approved):

PR #12152, opened 2024-02-21
Fix #12123: Allow not emitting zero digits in certain cases in number formats

### LINKED ISSUE #12123: [Bug]: PR#11993 is not properly display CJK currency
### Version of OpenTTD

14.0-RC1

### Expected result

1. For £10,002,003,405,670,089, it should be ``(₩) 1경 2조 34억 567만 89`` in Korean
2. For £185,000,000, it should be ``(₩) 1억 8500만`` in Korean (for ``{CURRENCY_LONG}`` maybe)

### Actual result

1. It is displayed as ``(₩) 1경 0002조 0034억 0567만 0089``
2. ``(₩) 1억 8500만 0000`` (for ``{CURRENCY_LONG}`` maybe)

### Steps to reproduce

PR #11993 have introduced the non-western number formats which is very thankful for CJK languages.
As I wrote above, however, the implemented shown number format (maybe for ``{CURRENCY_LONG}``) in my language, Korean, is somehow akward.

1. It would be great to bind something like ``0경{NBSP}0조{NBSP}0억{NBSP}0만`` for ``##numberabbreviations`` to omit the leading zeros for each units.
2. It would be great to be able to omit last zeros (as for example 2) for ``{CURRENCY_LONG}``

For now, Korean translation will be going to use western number format until these requested things would be improved.

rubidium42:
## Motivation / Problem

See #12123.
It looks like you are allowed to just leave out whole ranges of zero digits in Korean (and possible also Chinese/Japanese).

## Description

Introduce a flag in the number format that essentially allows you to reset whether a higher digit has been seen. If you use this at the start of each of the digit groups in Korean, it's handling it as if it were a much smaller number and omit the zeros and separators that are not needed.

However, this brings a problem at the end of the number. If one of the higher digit grouping already had a number, then the last digit grouping should be omitted if everything is zero. So, introduce yet another flag that's automatically set for the last digit which will cause the digit only to be omitted if nothing was omitted yet (or any of the other reasons why the default case would omit a digit).

이름 없음의 게임, 2035-01-01

## Limitations

There's probably something else I'm missing in the CJK number formats.

Since Korean wants a space after the character (see #12124), but it also wants to omit whole groups the `{NBSP}` of a higher grouping could end up to be the last character. This could cause a visual jump in tables such as in the finance window. I do not know an elegant solution for it, but a nasty work-around would be to always emit a trailing `{NBSP}` so they are aligned. Or maybe yet another flag to influence `{NBSP}` placement or something. Who has a great idea?

## Checklist for review

### DIFF
--- a/src/lang/korean.txt
+++ b/src/lang/korean.txt
@@ -3,8 +3,8 @@
 ##isocode ko_KR
 ##plural 11
 ##textdir ltr
-##numberformat 0000경0000조0000억0000만0000
-##numberabbreviations 4=0000경0000조0000억0000만|8=0000경0000조0000억|12=0000경0000조|16=0000경
+##numberformat 0000경{NBSP}1000조{NBSP}1000억{NBSP}1000만1000
+##numberabbreviations 4=0000경{NBSP}1000조{NBSP}1000억{NBSP}1000만|8=0000경{NBSP}1000조{NBSP}1000억|12=0000경{NBSP}1000조|16=0000경
 ##decimalsep .
 ##winlangid 0x0412
 ##grflangid 0x3a
--- a/src/language.h
+++ b/src/language.h
@@ -109,13 +109,23 @@ extern std::unique_ptr<icu::Collator> _current_collator;
 
 /** The number digits available in a uint64_t. */
 constexpr int DIGITS_IN_UINT64_T = 20;
+/** Container for the formatting of a single digit. */
+struct DigitFormat {
+	enum EmitBehaviour {
+		DEFAULT, ///< Only emit the digit when it's not zero, or there is a higher digit not zero.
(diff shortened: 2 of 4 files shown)


frosch123:
Yeah, this is nasty... It may need a special flag for numbers in tables / lists.

Option 1: unaligned
```
                 1조
-                1억
———————————————————
              9999억
```

Option 2: keep all zeros for aligned numbers
```
   1조 0000억 0000만
-         1억 0000만
———————————————————
       9999억 0000만
```

Option 3: replace some zeros with NBSP
```
   1조    0억    0만
-         1억    0만
———————————————————
       9999억    0만
```

Option 4: replace empty groups with NBSP:
```
   1조             
-         1억      
———————————————————
       9999억      
```

@OpenTTD/ko_kr Which option appears most natural to you?

PR #12155, opened 2024-02-22
Fix f6dd505: Only reset unbunching departure data in the correct depot

2TallTyler:
## Motivation / Problem

Right before a vehicle leaves a depot, it calculates its next departure time. This is intended to happen only when leaving its unbunching depot, but it happens when leaving _any_ depot.

This has the effect of not unbunching vehicles at the correct depot, unless the route is very short, so we might get some bug reports about unbunching not working that are caused by this. 

## Description

Only calculate the departure time after the correct depot order.

## Checklist for review

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -2433,10 +2433,27 @@ bool Vehicle::HasUnbunchingOrder() const
 }
 
 /**
+ * Check if the previous order is a depot unbunching order.
+ * @return true Iff the previous order is a depot order with the unbunch flag.
+ */
+static bool PreviousOrderIsUnbunching(const Vehicle *v)
+{
+	/* If we are headed for the first order, we must wrap around back to the last order. */
+	bool is_first_order = (v->GetOrder(v->cur_real_order_index) == v->GetFirstOrder());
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):

src/vehicle.cpp
```diff
@@ -2432,11 +2432,28 @@ bool Vehicle::HasUnbunchingOrder() const
...
+ * @return true Iff the previous order is a depot order with the unbunch flag.
+ */
+static bool PreviousOrderIsUnbunching(const Vehicle *v)
+{
+	/* If we are headed for the first order, we must wrap around back to the last order. */
+	bool is_first_order = (v->GetOrder(v->cur_real_order_index) == v->GetFirstOrder());
+	Order *previous_order = (is_first_order) ? v->GetLastOrder() : v->GetOrder(v->cur_real_order_index - 1);
+
+	if (previous_order == nullptr || !previous_order->IsType(OT_GOTO_DEPOT)) return false;
+	return (previous_order->GetDepotActionType() & ODATFB_UNBUNCH) != 0;
```
TrueBrain, line 2446: I applied our coding style, and added `!= 0`.

2TallTyler: Thank you!

PR #12156, opened 2024-02-22
Fix #11862: Ships leaving docks could be momentarily lost

### LINKED ISSUE #11862: [Bug]: High level pathfinder passing patches with no water as the origin.

SamuXarick:
## Motivation / Problem
When a ship is departing from a dock, and wants to go back to it again, the destination tile is set as the origin, making the ship have a `OT_LEAVESTATION` `OrderType`. With this order type, the docking tiles aren't added as origins even though the tile is still that of a dock.

On YAPF, the ship becomes momentarily flagged as lost when this happens.

## Description
`OT_LEAVESTATION` `OrderType` still has the same destination as that of `OT_GOTO_STATION`, so assume they still belong to the same dock.
Fixes #11862 

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -44,8 +44,9 @@ protected:
 public:
 	void SetDestination(const Ship *v)
 	{
-		if (v->current_order.IsType(OT_GOTO_STATION)) {
+		if (v->current_order.IsType(OT_GOTO_STATION) || v->current_order.IsType(OT_LEAVESTATION)) {
 			this->dest_station = v->current_order.GetDestination();
+			assert(Station::IsValidID(this->dest_station));
 			this->dest_tile = CalcClosestStationTile(this->dest_station, v->tile, StationType::Dock);
 			this->dest_trackdirs = INVALID_TRACKDIR_BIT;
 		} else {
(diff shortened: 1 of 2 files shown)


Kuhnovic:
Fix looks good to me. I just hope that we don't end up with more order-specific handling over time. It would be nice if we could take this order-specific behavior out of the pathfinder altogether, but that's beyond the scope of this PR if you ask me.

Kuhnovic:
When I run the savegame provided in #11862 I see no lost ships, and that's without this fix. I added if cases specific for OT_LEAVESTATION in both the ship and region pathfinder, and added breakpoints which never got triggered. I tested using the savegame from 11862 and Wentbourne. It seems that something else has been changed that prevents this case from happening.

... which is good, because it actually makes little sense to call the pathfinder when you are leaving a station IMO. If you know your destination then you are heading somewhere, and the order should be e.g. OT_GOTO_STATION. And this seems to have been fixed.

PR #12159, opened 2024-02-22
Codechange: Store industry cargo data in vector instead of array.

PeterN:
## Motivation / Problem

Industries store data for 16 incoming and 16 outgoing cargos. Most of the time these are empty as original specs only permitted 2 or 3 of each.

#10541 wants to extend the industry cargo history from 2 periods to 25 periods, which would make the industry data size increase quite a bit.

## Description

Replace fixed accepts and produced arrays with vectors. This allows the cargo data to be sized more appropriately.

## Limitations

This does not have much memory impact by itself, but is useful for #10541.

Does not rename types, so e.g. `ProducedCargoArray` is now a vector...

## Checklist for review

### DIFF
--- a/src/saveload/afterload.cpp
+++ b/src/saveload/afterload.cpp
@@ -3044,16 +3044,6 @@ bool AfterLoadGame()
 	if (IsSavegameVersionBefore(SLV_EXTEND_INDUSTRY_CARGO_SLOTS)) {
 		/* Make sure added industry cargo slots are cleared */
 		for (Industry *i : Industry::Iterate()) {
-			for (auto it = std::begin(i->produced) + 2; it != std::end(i->produced); ++it) {
-				it->cargo = INVALID_CARGO;
-				it->waiting = 0;
-				it->rate = 0;
-				it->history = {};
-			}
-			for (auto it = std::begin(i->accepted) + 3; it != std::end(i->accepted); ++it) {
-				it->cargo = INVALID_CARGO;
-				it->waiting = 0;
-			}
 			/* Make sure last_cargo_accepted_at is copied to elements for every valid input cargo.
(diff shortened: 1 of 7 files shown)


src/saveload/afterload.cpp
```diff
@@ -3038,21 +3039,19 @@ bool AfterLoadGame()
...
-				it->cargo = INVALID_CARGO;
-				it->waiting = 0;
-				it->rate = 0;
-				it->history = {};
-			}
-			for (auto it = std::begin(i->accepted) + 3; it != std::end(i->accepted); ++it) {
-				it->cargo = INVALID_CARGO;
-				it->waiting = 0;
-			}
+			/* Remove all entries from the first with invalid cargo */
```
2TallTyler: This comment (repeated several times) seems to be missing a word for me to understand it. Remove all entries from the first...what? First slot? First industry? 

(The code is beyond me or I'd answer my own question.) 🙂

2TallTyler (commented):
If you're going to go to all this trouble of making this very nice, wide-ranging PR, why not rename types while you're at it? (Honest question, not a suggestion.) 😄

PeterN:
> If you're going to go to all this trouble of making this very nice, wide-ranging PR, why not rename types while you're at it? (Honest question, not a suggestion.) 😄

Mainly because it makes a PR bigger and changes lines that otherwise aren't changed. But I've done it now.

src/saveload/afterload.cpp
PeterN: From the first slot onwards. But looking at this, along with the recent cargo type changes, I think the code needs to be revised anyway.

rubidium42 (approved):

PR #12163, opened 2024-02-22
Add: Basic autocompletion on tab for console commands

Fefer-Ivan:
## Motivation / Problem
When typing console commands, it is very common to have some kind of autocompletion when pressing tab.
It is also a great way to discover console commands.

## Description
When in console, the user pressed tab and have at least one character entered, simple autocompletion will happen:
1. If the entered text is a prefix to no commands or aliases, nothing will happen.
```
] abacab
<Tab>
] abacab
```
2. If the entered text is a prefix to a single command or alias, it will be replaced with this command/alias.
```
] he
<Tab>
] help
```
4. If there are multiple options, the first one will be auto-competed.
```
] st
<Tab>
] start_ai
```
5. If tab is pressed again, it will change to the next available option.
```
] start_ai  (from previous example)
<Tab>
] status
```
6. If we run out of options, we will restore original text.
```
] stop_ai
<Tab>
] st
```

This is the same logic that is used in network_chat_gui. It is extracted into autocomplete.h/.cpp and is used in both places.

## Limitations

This is very basic algorithm.
It can be improved by using more context-dependent hints.

For example, when trying to autocomplete `start_ai A` we can look at the list of installed AIs and autocomplete from it. For example `start_ai AdmiralAI`.

However, this will require more complex code modifications, so I want to test my waters with more simple PR, before submitting something bigger.

## Checklist for review

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,6 +52,8 @@ add_files(
     animated_tile_func.h
     articulated_vehicles.cpp
     articulated_vehicles.h
+    autocompletion.cpp
+    autocompletion.h
     autoreplace.cpp
     autoreplace_base.h
     autoreplace_cmd.cpp
--- a/src/autocompletion.cpp
+++ b/src/autocompletion.cpp
@@ -1,0 +1,66 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file autocompletion.cpp Generic auto-completion engine. */
+
+#include "stdafx.h"
+
+#include "autocompletion.h"
+
+#include "console_internal.h"
+#include "town.h"
+#include "network/network_base.h"
+
+#include "safeguards.h"
+
+bool AutoCompletion::AutoComplete()
+{
+	// We are pressing TAB for the first time after reset.
+	if (this->suggestions.empty()) {
+		this->InitSuggestions(this->textbuf->buf);
+		if (this->suggestions.empty()) {
+			return false;
+		}
+		this->ApplySuggestion(prefix, suggestions[0]);
+		return true;
+	}
+
+	// We are pressing TAB again on the same text.
+	if (this->current_suggestion_index + 1 < this->suggestions.size()) {
+		this->ApplySuggestion(prefix, this->suggestions[++this->current_suggestion_index]);
+	} else {
+		// We are out of options, restore original text.
+		this->textbuf->Assign(initial_buf);
+		this->Reset();
+	}
+	return true;
+}
+
+void AutoCompletion::Reset()
+{
+	this->prefix = "";
+	this->query = "";
+	this->initial_buf.clear();
+	this->suggestions.clear();
+	this->current_suggestion_index = 0;
+}
+
+void AutoCompletion::InitSuggestions(std::string_view text)
+{
+	this->initial_buf = text;
+	size_t space_pos = this->initial_buf.find_last_of(' ');
+	this->query = this->initial_buf;
+	if (space_pos == std::string::npos) {
+		this->prefix = "";
+	} else {
+		this->prefix = this->query.substr(0, space_pos + 1);
+		this->query.remove_prefix(space_pos + 1);
+	}
+
+	this->suggestions = this->GetSuggestions(prefix, query);
+	this->current_suggestion_index = 0;
+}
--- a/src/autocompletion.h
+++ b/src/autocompletion.h
@@ -1,0 +1,46 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file autocompletion.h Generic auto-completion engine. */
+
+#ifndef AUTOCOMPLETION_H
+#define AUTOCOMPLETION_H
+
+#include "textbuf_type.h"
+
+class AutoCompletion {
+protected:
+	Textbuf *textbuf;
+
+private:
+	std::string              initial_buf; ///< Value of text buffer when we started current suggestion session.
+
+	std::string_view         prefix;      ///< Prefix of the text before the last space.
+	std::string_view         query;       ///< Last token of the text. This is used to based the suggestions on.
+
+	std::vector<std::string> suggestions;
+	size_t                   current_suggestion_index;
+
+public:
+	AutoCompletion(Textbuf *textbuf) : textbuf(textbuf)
+	{
+		this->Reset();
+	}
+	virtual ~AutoCompletion() = default;
+
+	// Returns true the textbuf was updated.
+	bool AutoComplete();
+	void Reset();
+
+private:
+	void InitSuggestions(std::string_view text);
+
+	virtual std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) = 0;
+	virtual void ApplySuggestion(std::string_view prefix, std::string_view suggestion) = 0;
+};
+
+#endif /* AUTOCOMPLETION_H */
--- a/src/console_gui.cpp
+++ b/src/console_gui.cpp
@@ -10,6 +10,7 @@
 #include "stdafx.h"
 #include "textbuf_type.h"
 #include "window_gui.h"
+#include "autocompletion.h"
 #include "console_gui.h"
 #include "console_internal.h"
 #include "window_func.h"
@@ -68,9 +69,44 @@ static std::deque<IConsoleLine> _iconsole_buffer;
 
 static bool TruncateBuffer();
 
+class ConsoleAutoCompletion final : public AutoCompletion {
+public:
+	using AutoCompletion::AutoCompletion;
+
+private:
+	std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) override
+	{
+		prefix = StrTrimView(prefix);
+		std::vector<std::string> suggestions;
+
+		/* We only suggest commands or aliases, so we only do it for the first token or an argument to help command. */
+		if (!prefix.empty() && prefix != "help") {
+			return suggestions;
+		}
+
+		for (const auto &[_, command] : IConsole::Commands()) {
+			if (command.name.starts_with(query)) {
+				suggestions.push_back(command.name);
+			}
+		}
+		for (const auto &[_, alias] : IConsole::Aliases()) {
+			if (alias.name.starts_with(query)) {
+				suggestions.push_back(alias.name);
+			}
+		}
+
+		return suggestions;
+	}
+
+	void ApplySuggestion(std::string_view prefix, std::string_view suggestion) override
+	{
+		this->textbuf->Assign(fmt::format("{}{} ", prefix, suggestion));
+	}
+};
 
 /* ** main console cmd buffer ** */
 static Textbuf _iconsole_cmdline(ICON_CMDLN_SIZE);
+static ConsoleAutoCompletion _iconsole_tab_completion(&_iconsole_cmdline);
 static std::deque<std::string> _iconsole_history;
 static ptrdiff_t _iconsole_historypos;
 IConsoleModes _iconsole_mode;
@@ -86,6 +122,7 @@ static void IConsoleClearCommand()
 	_iconsole_cmdline.pixels = 0;
 	_iconsole_cmdline.caretpos = 0;
 	_iconsole_cmdline.caretxoffs = 0;
+	_iconsole_tab_completion.Reset();
 	SetWindowDirty(WC_CONSOLE, 0);
 }
 
@@ -258,8 +295,18 @@ struct IConsoleWindow : Window
 				IConsoleCmdExec("clear");
 				break;
 
-			default:
-				if (_iconsole_cmdline.HandleKeyPress(key, keycode) != HKPR_NOT_HANDLED) {
+			case WKC_TAB:
+				if (_iconsole_tab_completion.AutoComplete()) {
+					this->SetDirty();
+				}
+				break;
+
+			default: {
+				HandleKeyPressResult handle_result = _iconsole_cmdline.HandleKeyPress(key, keycode);
+				if (handle_result != HKPR_NOT_HANDLED) {
+					if (handle_result == HKPR_EDITING) {
+						_iconsole_tab_completion.Reset();
+					}
 					IConsoleWindow::scroll = 0;
 					IConsoleResetHistoryPos();
 					this->SetDirty();
@@ -267,6 +314,7 @@ struct IConsoleWindow : Window
 					return ES_NOT_HANDLED;
 				}
 				break;
+			}
 		}
 		return ES_HANDLED;
 	}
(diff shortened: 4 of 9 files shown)


michicc:
I like the idea.

Without having looked at the actual code or tried it, a few formal things:
We have a strict commit message check (see [CODINGSTYLE.MD](https://github.com/OpenTTD/OpenTTD/blob/master/CODINGSTYLE.md#commit-message)). We also usually like to squash/rebase down to a sane final state, so you can just squash the commits and give a new message.

Also, indentation is with tabs, not spaces.

glx22:
Handling of <TAB> needs to be very careful as it activates fast forward.

Fefer-Ivan:
Thanks for the feedback.
I have rebased the branch. I don't see any spaces.

Yes, when in game, I can even see Fast-Forward button flash pressed when I press tab. 
Any advice how to handle that? Prohibit fast-forward when console is open? 

I will take a look at network_gui auto-completion implementation and refactoring into a separate class for reusing.
Thanks.

ldpl:
IMO fast-forwarding Tab should be 
a) configurable in hotkeys.cfg
b) mapped to something else by default
also, iirc Tab only fast-fowards in debug build.
Not sure how that relates to this PR though.

2TallTyler:
Most hotkeys are disabled when the cursor is captured in a text box, because they are being used to type in the box — try pressing X for transparency when editing the name of a station, for example. I see no reason why <TAB> should be any different. But it is, both in the console and in text boxes. Changing this might be a separate PR.

LordAro:
Yeah, we're not talking large amounts of memory here, I can't imagine that's an issue. Especially with the frequency that it'd be used

I will say that I'm not sure it requires 6 new files though - we don't have any sort of requirement for one class per file, so feel free to shove them wherever is most appropriate (which might be 1 new cpp/h files, might be none at all, idk)

src/autocompletion.h
```diff
@@ -0,0 +1,64 @@
...
+};
+
+class NetworkChatAutoCompletion final : public AutoCompletion {
+public:
+	using AutoCompletion::AutoCompletion;
+
+private:
+	std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) override;
+	void ApplySuggestion(std::string_view prefix, std::string_view suggestion) override;
+};
```
glx22: I would move this declaration and the override implementations to `network_chat_gui.cpp`, like we do for `Window` subclasses.

src/autocompletion.h
```diff
@@ -0,0 +1,64 @@
...
+
+class ConsoleAutoCompletion final : public AutoCompletion {
+public:
+	using AutoCompletion::AutoCompletion;
+
+private:
+	std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) override;
+	void ApplySuggestion(std::string_view prefix, std::string_view suggestion) override;
+};
```
glx22: Same here but in `console_gui.cpp`.

src/autocompletion.cpp
```diff
@@ -0,0 +1,130 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file autocompletion.cpp Generic auto-completion engine. */
+
+#include "autocompletion.h"
```
glx22, line 12:
```suggestion
#include "stdafx.h"

#include "autocompletion.h"
```

src/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ add_files(
     animated_tile_func.h
     articulated_vehicles.cpp
     articulated_vehicles.h
+    autocompletion.cpp
```
glx22, line 55:
```suggestion
    autocompletion.cpp
    autocompletion.h
```

src/autocompletion.h
Fefer-Ivan: Done

src/autocompletion.h
Fefer-Ivan: Done

Fefer-Ivan:
@glx22 Thanks for the review. I have moved the code as suggested.
I am aware of squash and commit naming requiremenets. I will do it before merging.

src/autocompletion.h
```diff
@@ -0,0 +1,46 @@
...
+	std::string_view         prefix;      // Prefix of the text before the last space.
+	std::string_view         query;       // Last token of the text. This is used to based the suggestions on.
+
+	std::vector<std::string> suggestions;
+	size_t                   current_suggestion_index;
+
+public:
+	AutoCompletion(Textbuf* textbuf) : textbuf(textbuf)
+	{
+		Reset();
```
LordAro: missing `this->`

src/autocompletion.h
```diff
@@ -0,0 +1,46 @@
...
+/** @file autocompletion.h Generic auto-completion engine. */
+
+#ifndef AUTOCOMPLETION_H
+#define AUTOCOMPLETION_H
+
+#include "textbuf_type.h"
+
+class AutoCompletion {
+protected:
+	Textbuf*                 textbuf;
```
LordAro: we no longer generally bother with the alignment stuff, especially not multiple "sections"

Also, `*` goes on the variable

src/autocompletion.h
```diff
@@ -0,0 +1,46 @@
...
+
+private:
+	std::string              initial_buf; // Value of text buffer when we started current suggestion session.
```
LordAro: doxygen comments are `///<`

src/console_gui.cpp
```diff
@@ -68,9 +69,52 @@ static std::deque<IConsoleLine> _iconsole_buffer;
...
+class ConsoleAutoCompletion final : public AutoCompletion {
+public:
+	using AutoCompletion::AutoCompletion;
+
+private:
+	std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) override
+	{
+		while (prefix.ends_with(" ")) {
+			prefix.remove_suffix(1);
+		}
```
LordAro: Do we have a "trim" string function already? Seems like it would be more efficient than the explicit loop like that

src/console_gui.cpp
```diff
@@ -68,9 +69,52 @@ static std::deque<IConsoleLine> _iconsole_buffer;
...
+
+		std::vector<std::string> suggestions;
+
+		// We only suggest commands or aliases, so we only do it for the first token or
+		// an argument to help command.
+		if (!prefix.empty() && prefix != "help") {
+			return {};
```
LordAro: Might as well just return `suggestions` ? (or move the declaration of the variable down)

src/console_gui.cpp
```diff
@@ -68,9 +69,52 @@ static std::deque<IConsoleLine> _iconsole_buffer;
...
+	std::vector<std::string> GetSuggestions(std::string_view prefix, std::string_view query) override
+	{
+		while (prefix.ends_with(" ")) {
+			prefix.remove_suffix(1);
+		}
+
+		std::vector<std::string> suggestions;
+
+		// We only suggest commands or aliases, so we only do it for the first token or
+		// an argument to help command.
```
LordAro: comments on their own should use `/* */` (and we don't have hard line length limits, no need for a line break here)

src/console_gui.cpp
```diff
@@ -68,9 +69,52 @@ static std::deque<IConsoleLine> _iconsole_buffer;
...
+		return suggestions;
+	}
+
+	void ApplySuggestion(std::string_view prefix, std::string_view suggestion) override
+	{
+		if (prefix.empty()) {
+			this->textbuf->Assign(fmt::format("{} ", suggestion));
+		} else {
+			this->textbuf->Assign(fmt::format("{}{} ", prefix, suggestion));
+		}
```
LordAro: Seems a redundant distinction, as both branches do the same thing. Keep it simple, I say

src/console_gui.cpp
```diff
@@ -258,8 +303,18 @@ struct IConsoleWindow : Window
...
 
+			case WKC_TAB:
+				if (_iconsole_tab_completion.AutoComplete()) {
+					this->SetDirty();
+				}
+				break;
+
 			default:
-				if (_iconsole_cmdline.HandleKeyPress(key, keycode) != HKPR_NOT_HANDLED) {
+				HandleKeyPressResult handle_result = _iconsole_cmdline.HandleKeyPress(key, keycode);
```
LordAro: It mostly works, but it's good practice to use braces when introducing variable declarations into a case block

src/network/network_chat_gui.cpp
```diff
@@ -261,20 +261,55 @@ static void SendChat(const std::string &buf, DestType type, int dest)
...
+		return suggestions;
+	}
+
+	void ApplySuggestion(std::string_view prefix, std::string_view suggestion) override
+	{
+		if (prefix.empty()) {
+			this->textbuf->Assign(fmt::format("{}: ", suggestion));
+		} else {
+			this->textbuf->Assign(fmt::format("{}{} ", prefix, suggestion));
+		}
```
LordAro, line 296: Seems a redundant distinction, as both branches do the same thing. Keep it simple, I say

src/network/network_chat_gui.cpp
```diff
@@ -261,20 +261,55 @@ static void SendChat(const std::string &buf, DestType type, int dest)
...
+	}
+};
 
 /** Window to enter the chat message in. */
 struct NetworkChatWindow : public Window {
 	DestType dtype;       ///< The type of destination.
 	int dest;             ///< The identifier of the destination.
 	QueryString message_editbox; ///< Message editbox.
+	NetworkChatAutoCompletion chat_tab_completion;
```
LordAro: missing doxygen comment

src/network/network_chat_gui.cpp
```diff
@@ -261,20 +261,55 @@ static void SendChat(const std::string &buf, DestType type, int dest)
 		NetworkServerSendChat((NetworkAction)(NETWORK_ACTION_CHAT + type), type, dest, buf, CLIENT_ID_SERVER);
 	}
 }
+class NetworkChatAutoCompletion final : public AutoCompletion {
```
LordAro, line 265: missing empty line prior to class declaration

src/console_gui.cpp
```diff
@@ -68,9 +69,52 @@ static std::deque<IConsoleLine> _iconsole_buffer;
...
+
+		std::vector<std::string> suggestions;
+
+		// We only suggest commands or aliases, so we only do it for the first token or
+		// an argument to help command.
+		if (!prefix.empty() && prefix != "help") {
+			return {};
+		}
+
+		for (const auto& [_, command] : IConsole::Commands()) {
```
LordAro: similar to pointer, reference goes with the variable (also below)

LordAro (commented):
Largely codestyle nitpicks

src/autocompletion.h
Fefer-Ivan: Done

src/autocompletion.h
Fefer-Ivan: Done

src/autocompletion.h
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: I looked at string_func.h and there are two trim functions:
One for std::string and one for a null-terminated utf-8 strings.
Both of them doesn't combine well with std::string_view (as they can be not null terminated).

I can add one more for std::string_view.
Something like std::string_view StrTrimView(std::string_view s).
It also should be possible to implement StrTrimInPlace as (s = StrTrimView(s)).

src/console_gui.cpp
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: Done

src/network/network_chat_gui.cpp
Fefer-Ivan: So here it is actually important, as we add ": " if there is no prefix.
This is consistent with existing behaviour:  https://github.com/OpenTTD/OpenTTD/blob/626ee4ab4184068a9779813295a9dace14c3cc82/src/network/network_chat_gui.cpp#L416

src/network/network_chat_gui.cpp
Fefer-Ivan: Done

src/network/network_chat_gui.cpp
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: Done

src/console_gui.cpp
Fefer-Ivan: Implemented my suggestion and added unit tests.
Please let me know if you want me to extract it into a separate Pr.

src/network/network_chat_gui.cpp
glx22: Maybe copy the old comment in this new location then.
And old code was `{} {}` for the other case.

Fefer-Ivan: In new code prefix includes the separator space, so there is no need to add it in format string.
Added the comment.

src/console_gui.cpp
```diff
@@ -258,22 +295,34 @@ struct IConsoleWindow : Window
...
+				if (_iconsole_tab_completion.AutoComplete()) {
+					this->SetDirty();
+				}
+				break;
+
+			default: {
+				HandleKeyPressResult handle_result = _iconsole_cmdline.HandleKeyPress(key, keycode);
+				if (handle_result != HKPR_NOT_HANDLED) {
+					if (handle_result == HKPR_EDITING) {
+						_iconsole_tab_completion.Reset();
```
Fefer-Ivan, line 308: The current autocompletion is stateful - it cycles through suggestions in a loop on Tab.

Every time the user edits _iconsole_cmdline text buffer, we need to reset the state of autocompletion.

I have added a Reset every time we change _iconsole_cmdline in this file.

However, it is also exposed through GetFocusedTextbuf() method of Window and I don't have any way to "catch" edit events if textbuf is updated via this method.

In network_chat_gui.cpp it is fine, because we have OnEditboxChanged, but this event is not called when we type text in the console.

However, when. I tried adding OnEditboxChanged to IConsoleWindow, it was not called when typing in console and I am not sure why.

So I am not sure that this change is actually fully correct.
But manual testing didn't show any problems.

glx22: IConsoleWindow acts like an editbox, but it's not an editbox and it doesn't have an editbox widget either. This is why OnEditboxChanged() is never called for it.

Fefer-Ivan: How is GetFocusedTextbuf() works? Maybe we can remove this method, so _iconsole_cmdline will become fully private?

glx22: `GetFocusedTextbuf()` is used only by macos, in ObjC++ functions related to text handling (selection, insertion, ...).
Console being a kind of editbox, this function can't be removed.

Fefer-Ivan: I checked that actually GetFocusedTextbuf() returns a const pointer to Textbuf, so it is only used to read the current text, not modify it, so it is safe.

With this I am resolving this thread

src/console_gui.cpp
```diff
@@ -258,29 +295,41 @@ struct IConsoleWindow : Window
...
 		if (_iconsole_cmdline.InsertString(str, marked, caret, insert_location, replacement_end)) {
+			_iconsole_tab_completion.Reset();
 			IConsoleWindow::scroll = 0;
 			IConsoleResetHistoryPos();
 			this->SetDirty();
 		}
 	}
 
-	Textbuf *GetFocusedTextbuf() const override
+	const Textbuf *GetFocusedTextbuf() const override
```
Fefer-Ivan, line 332: Why on Earth did it compile without it?

Fefer-Ivan: The return type of an overriding function shall be either identical to the return type of the overridden function or covariant with the classes of the functions. If a function D::f overrides a function B::f, the return type of the functions are covariant if the satisfy the following criteria:

both are pointers to classes or references to classes98)
the class in the return type of B::f is the same class as the class in the return type of D::f or, is an unambiguous direct or indirect base class of the class in the return type of D::f and is accessible in D
both pointers or references have the same cv-qualification and the class type in the return type of D::f has the same cv-qualification as or less cv-qualification than the class type in the return type of B::f.

Fefer-Ivan: TIL.

But it is confusing, so I would like to fix it.

src/autocompletion.h
LordAro: Not everywhere :)

Fefer-Ivan: 🤦

Fefer-Ivan: Looked through the code one more time and fix the only place I noticed.

LordAro (approved):
Code LGTM

PR #12169, opened 2024-02-23
Codechange: Use functions to access colour gradients and replace magic numbers.

PeterN:
## Motivation / Problem

Mapping from `Colours` to pixel colour uses direct access to the `_colour_gradient` array, along with an additional index for the shade.

There is limited bounds checking for this access, and due to an off-by-one in TTD, the shades range from 1 to 7 instead of 0 to 7.

## Description

Instead, implement and use functions `GetColourGradient()` and `SetColourGradient()`. These ensure all access is within bounds.

A new enum is added to allow the shade to be specified by name, and the off-by-one is removed (technically a fix but it did not affect anything so not marked as a fix) which allows access to the light(er)est shade.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -146,8 +146,8 @@ void SetLocalCompany(CompanyID new_company)
  */
 TextColour GetDrawStringCompanyColour(CompanyID company)
 {
-	if (!Company::IsValidID(company)) return (TextColour)GetColourGradient(COLOUR_WHITE, 4) | TC_IS_PALETTE_COLOUR;
-	return (TextColour)GetColourGradient(_company_colours[company], 4) | TC_IS_PALETTE_COLOUR;
+	if (!Company::IsValidID(company)) return (TextColour)GetColourGradient(COLOUR_WHITE, SHADE_4) | TC_IS_PALETTE_COLOUR;
+	return (TextColour)GetColourGradient(_company_colours[company], SHADE_4) | TC_IS_PALETTE_COLOUR;
 }
 
 /**
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -146,8 +146,8 @@ void SetLocalCompany(CompanyID new_company)
  */
 TextColour GetDrawStringCompanyColour(CompanyID company)
 {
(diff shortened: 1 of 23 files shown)


src/misc_gui.cpp
```diff
@@ -794,7 +794,7 @@ void QueryString::DrawEditBox(const Window *w, WidgetID wid) const
 
 	DrawFrameRect(cr, wi->colour, wi->IsLowered() ? FR_LOWERED : FR_NONE);
 	DrawSpriteIgnorePadding(rtl ? SPR_IMG_DELETE_RIGHT : SPR_IMG_DELETE_LEFT, PAL_NONE, cr, SA_CENTER);
-	if (this->text.bytes == 1) GfxFillRect(cr.Shrink(WidgetDimensions::scaled.bevel), _colour_gradient[wi->colour & 0xF][2], FILLRECT_CHECKER);
```
2TallTyler, line 797: What was this `colour & 0xF` doing? (here and elsewhere)

src/company_cmd.cpp
```diff
@@ -146,8 +146,8 @@ void SetLocalCompany(CompanyID new_company)
  */
 TextColour GetDrawStringCompanyColour(CompanyID company)
 {
-	if (!Company::IsValidID(company)) return (TextColour)_colour_gradient[COLOUR_WHITE][4] | TC_IS_PALETTE_COLOUR;
-	return (TextColour)_colour_gradient[_company_colours[company]][4] | TC_IS_PALETTE_COLOUR;
+	if (!Company::IsValidID(company)) return (TextColour)GetColourGradient(COLOUR_WHITE, SHADE_NORMAL) | TC_IS_PALETTE_COLOUR;
```
2TallTyler, line 149: Can these be static_casts?

src/misc_gui.cpp
PeterN: It was making sure the value of `wi->colour` is within bounds (not always done) which is now done by `GetColourGradient()`

2TallTyler: Ah, thanks.

rubidium42 (approved):

PR #12170, opened 2024-02-24
Fix #8022: Ship automatic service causes them to be stuck in a loop

### LINKED ISSUE #8022: Ship automatic service causes them to be stuck in a loop

Kuhnovic:
## Motivation / Problem

Fixes #8022

## Description

Before this change, CheckIfShipNeedsService would try to find a new depot each time it's called (which is every new economy day). Depending on the position of the ship, it might not find a depot. In some edge cases a ship could get stuck, this is what happens:
1. Ship needs service. CheckIfShipNeedsService  is called and it finds a depot.
2. In order to get to the depot, the ship needs to move away from the depot first.
3. CheckIfShipNeedsService is called again. This time no depot is found.
4. Order is set to a dummy order.
5. The ships continues to it's regular order, which is to go to the dock.
6. The whole thing repeats.

With my changes, the ship will do the following:
- When looking for automatic service, it will search for a depot, but only if it isn't already heading for one
- If the pathfinder is called an the depot can no longer be reached, then abandon the automatic service order
- If the depot can be reached but suddenly is too far away due to terraforming, abandon the automatic service order

This should only affect automatic service orders. Regular orders are scheduled by the user and should trigger "ship is lost" if a depot can't be reached.

## Limitations

- Continuously looking for an even closer depot would be nice but has the potential to cause loops where a ship goes back and forth between two depots. So I decided not to do this.
- Orders seem to be quite complex and I don't have much experience with them. I might have inadvertently broken something. 

## Checklist for review

### DIFF
--- a/src/ship_cmd.cpp
+++ b/src/ship_cmd.cpp
@@ -45,6 +45,12 @@
 /** Max distance in tiles (as the crow flies) to search for depots when user clicks "go to depot". */
 constexpr int MAX_SHIP_DEPOT_SEARCH_DISTANCE = 80;
 
+/** Returns the maximum distance to use when searching for a ship depot. */
+uint GetMaxShipDepotServiceSearchDistance()
+{
+	return _settings_game.pf.yapf.maximum_go_to_depot_penalty / YAPF_TILE_LENGTH;
+}
+
 /**
  * Determine the effective #WaterClass for a ship travelling on a tile.
  * @param tile Tile of interest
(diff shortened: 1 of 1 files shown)


src/ship_cmd.cpp
```diff
@@ -217,19 +217,24 @@ static void CheckIfShipNeedsService(Vehicle *v)
...
 
-	const Depot *depot = FindClosestShipDepot(v, max_distance);
-
-	if (depot == nullptr) {
-		if (v->current_order.IsType(OT_GOTO_DEPOT)) {
-			v->current_order.MakeDummy();
+	if (v->current_order.IsType(OT_GOTO_DEPOT)) {
+		const Depot *depot = Depot::GetByTile(v->dest_tile);
+		if (depot) {
+			v->current_order.MakeGoToDepot(depot->index, ODTFB_SERVICE);
```
Kuhnovic: I'm not sure this is really needed. It seems like a ship will always service at a depot if it has the chance. I'm hoping someone with a little more experience in this area reads this :)

src/ship_cmd.cpp
```diff
@@ -217,19 +217,24 @@ static void CheckIfShipNeedsService(Vehicle *v)
...
 	}
 
-	const Depot *depot = FindClosestShipDepot(v, max_distance);
-
-	if (depot == nullptr) {
-		if (v->current_order.IsType(OT_GOTO_DEPOT)) {
-			v->current_order.MakeDummy();
+	if (v->current_order.IsType(OT_GOTO_DEPOT)) {
+		const Depot *depot = Depot::GetByTile(v->dest_tile);
+		if (depot) {
```
2TallTyler:
```suggestion
		if (depot != nullptr) {
```

src/ship_cmd.cpp
```diff
@@ -217,19 +217,24 @@ static void CheckIfShipNeedsService(Vehicle *v)
...
+			v->current_order.MakeGoToDepot(depot->index, ODTFB_SERVICE);
 			SetWindowWidgetDirty(WC_VEHICLE_VIEW, v->index, WID_VV_START_STOP);
+			return;
 		}
+	}
+
+	const Depot *depot = FindClosestShipDepot(v, max_distance);
+	if (depot) {
```
2TallTyler:
```suggestion
	if (depot != nullptr) {
```

2TallTyler (commented):
I'm not knowledgeable enough about orders to answer your question, but here are a couple code style things. 😉

Kuhnovic:
I've converted this back to a draft. This solves #8022 but introduces a new problem: the ship will never be able to detect whether the depot has suddenly become unreachable due to terraforming. Or it might still be reachable, but the ship has to take a massive detour.

TBC...

Kuhnovic:
Alright, made some more fixes, ready for review!

src/ship_cmd.cpp
```diff
@@ -148,50 +153,60 @@ void Ship::GetImage(Direction direction, EngineImageType image_type, VehicleSpri
...
 	static std::unordered_set<int> visited_patch_hashes;
-	static std::deque<WaterRegionPatchDesc> patches_to_search;
+	static std::deque<SearchNode> patches_to_search;
 	visited_patch_hashes.clear();
 	patches_to_search.clear();
 
-	/* Step 1: find a set of reachable Water Region Patches using BFS. */
 	const WaterRegionPatchDesc start_patch = GetWaterRegionPatchInfo(v->tile);
-	patches_to_search.push_back(start_patch);
+	patches_to_search.push_back({ start_patch, 0 });
```
PeterN: `....emplace_back(start_path, 0)`

src/ship_cmd.cpp
```diff
@@ -148,50 +153,60 @@ void Ship::GetImage(Direction direction, EngineImageType image_type, VehicleSpri
...
-					std::abs(water_region_patch.y - start_patch.y) > max_region_distance) return;
+			const uint distance = current_node.distance + std::abs(water_region_patch.x - current_node.patch.x) + std::abs(water_region_patch.y - current_node.patch.y);
+			if (distance > max_region_distance) return;
 
 			const int hash = CalculateWaterRegionPatchHash(water_region_patch);
-			if (visited_patch_hashes.count(hash) == 0) {
+			if (!visited_patch_hashes.contains(hash)) {
 				visited_patch_hashes.insert(hash);
-				patches_to_search.push_back(water_region_patch);
+				patches_to_search.push_back({ water_region_patch, distance });
```
PeterN: `....emplace_back(water_region_patch, distance);`

src/ship_cmd.cpp
```diff
@@ -45,6 +45,12 @@
 /** Max distance in tiles (as the crow flies) to search for depots when user clicks "go to depot". */
 constexpr int MAX_SHIP_DEPOT_SEARCH_DISTANCE = 80;
 
+/** Returns the maximum distance to use when searching for a ship depot. */
```
Rito13, line 48: Just a couple of suggestions to make doxygen happy.

```suggestion
/** Called when searching for a ship depot to figure out @return the maximum distance to use. */
```

```suggestion
/**
 * Returns the maximum distance to use when searching for a ship depot.
 * @return The maximum distance.
 */
```

PR #8006, opened 2020-02-16
Codechange: Increase scrollbar length limit to 32bits

LordAro:
Also makes scrollbar internals properly use unsigned numbers. Bumped scrollbar limit to uint(32) as well, because why not

Fixes #7613

PR #12151, opened 2024-02-21
Fix #12086: Crash if NWidgetMatrix holds content taller than 64K pixels.

PeterN:
## Motivation / Problem

NWidgetMatrix can crash if the height of all items exceeds 65535, as it sets its scrollbar to the number of pixels to scroll. This can happen when lots of NewGRF objects are loaded, and is exacerbated by interface scaling.

## Description

Work around by scaling the count in NWidgetMatrix to ensure it is within the scrollbar limits.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR touches english.txt or translations? Check the [guidelines](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md)
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, game_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #12178, opened 2024-02-25
Codechange: use int32_t instead of uint16_t for scroll bar position/size/capacity

### LINKED ISSUE #12086: [Crash]: game suddenly crashes when opening Object List
### Version of OpenTTD

openttd-20240211-master-g8d9fa0ea89-linux-generic-amd64

### Steps to reproduce

Create new game with the NewGRF list attached.
When opening the Object List (from Terrain Menu), game crashes.
(Maybe caused by the high number of objects (or subcategories) because of many NewGRF's used.)
Message of json.log: 
    "crash": {
        "reason": "Assertion failed at line 762 of /__w/OpenTTD/OpenTTD/src/widget_type.h: num <= MAX_UVALUE(uint16_t)"
    },
Note: does not happen in 13.4 (screenshot)
### Upload crash files

[crash20240213231158.dmp](https://github.com/OpenTTD/OpenTTD/files/14273173/crash20240213231158.dmp)
[crash20240213231158.json.log](https://github.com/OpenTTD/OpenTTD/files/14273175/crash20240213231158.json.log)
[newgrfs.txt](https://github.com/OpenTTD/OpenTTD/files/14273179/newgrfs.txt)
image

rubidium42:
## Motivation / Problem

Seeing the hack in #12151 for #12086 and wondering whether just changing the type is better now, like #8006 attempted.

So, this closes #12086 and closes #8006, and "fixes" #12086 for about four orders of magnitude more elements in the list.

## Description

Change `uint16_t` to just `int32_t`, and amend the code using it for overflow and signedness related issues.

Also change `INT_MAX` to `INT32_MAX` when handling `GetScrolledRowFromWidget` because that now returns `int32_t` instead of `int`.

Furthermore hide the type of the size of a scrollbar away behind `using`/`constexpr` constants, but only in `Scrollbar` to aid with backporting.

## Limitations

I have not tested this much, especially not the #12086 crash situation.

I think the requests for changes in this PR have ballooned it too much already to really consider backporting this, and it leaves it in a messier state than what it used to be. So I'm not really standing behind this PR any more.

## Checklist for review

### DIFF
--- a/src/airport_gui.cpp
+++ b/src/airport_gui.cpp
@@ -495,8 +495,8 @@ public:
 				break;
 
 			case WID_AP_AIRPORT_LIST: {
-				int num_clicked = this->vscroll->GetScrolledRowFromWidget(pt.y, this, widget, 0, this->line_height);
-				if (num_clicked == INT_MAX) break;
+				int32_t num_clicked = this->vscroll->GetScrolledRowFromWidget(pt.y, this, widget, 0, this->line_height);
+				if (num_clicked == INT32_MAX) break;
 				const AirportSpec *as = AirportClass::Get(_selected_airport_class)->GetSpec(num_clicked);
 				if (as->IsAvailable()) this->SelectOtherAirport(num_clicked);
 				break;
--- a/src/widget_type.h
+++ b/src/widget_type.h
@@ -678,12 +678,16 @@ public:
  * Scrollbar data structure
  */
 class Scrollbar {
+public:
+	using size_type = int32_t;
+	static constexpr size_type max_size_type = std::numeric_limits<size_type>::max();
+	static constexpr size_type npos = max_size_type;
 private:
 	const bool is_vertical; ///< Scrollbar has vertical orientation.
-	uint16_t count;           ///< Number of elements in the list.
-	uint16_t cap;             ///< Number of visible elements of the scroll bar.
-	uint16_t pos;             ///< Index of first visible item of the list.
-	uint16_t stepsize;        ///< Distance to scroll, when pressing the buttons or using the wheel.
+	size_type count; ///< Number of elements in the list.
+	size_type cap; ///< Number of visible elements of the scroll bar.
+	size_type pos; ///< Index of first visible item of the list.
+	size_type stepsize; ///< Distance to scroll, when pressing the buttons or using the wheel.
 
 public:
 	/** Stepping sizes when scrolling */
@@ -701,7 +705,7 @@ public:
 	 * Gets the number of elements in the list
 	 * @return the number of elements
 	 */
-	inline uint16_t GetCount() const
+	inline size_type GetCount() const
 	{
 		return this->count;
 	}
@@ -710,7 +714,7 @@ public:
 	 * Gets the number of visible elements of the scrollbar
 	 * @return the number of visible elements
 	 */
-	inline uint16_t GetCapacity() const
+	inline size_type GetCapacity() const
 	{
 		return this->cap;
 	}
@@ -719,7 +723,7 @@ public:
 	 * Gets the position of the first visible element in the list
 	 * @return the position of the element
 	 */
-	inline uint16_t GetPosition() const
+	inline size_type GetPosition() const
 	{
 		return this->pos;
 	}
@@ -729,7 +733,7 @@ public:
 	 * @param item to check
 	 * @return true iff the item is visible
 	 */
-	inline bool IsVisible(uint16_t item) const
+	inline bool IsVisible(size_type item) const
 	{
 		return IsInsideBS(item, this->GetPosition(), this->GetCapacity());
 	}
@@ -751,7 +755,7 @@ public:
 	{
 		assert(stepsize > 0);
 
-		this->stepsize = ClampTo<uint16_t>(stepsize);
+		this->stepsize = ClampTo<size_type>(stepsize);
 	}
 
 	/**
@@ -761,9 +765,9 @@ public:
 	 */
 	void SetCount(size_t num)
 	{
-		assert(num <= MAX_UVALUE(uint16_t));
+		assert(num < Scrollbar::max_size_type);
 
-		this->count = ClampTo<uint16_t>(num);
+		this->count = ClampTo<size_type>(num);
 		/* Ensure position is within bounds */
 		this->SetPosition(this->pos);
 	}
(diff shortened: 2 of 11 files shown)


src/widget_type.h
```diff
@@ -701,7 +701,7 @@ class Scrollbar {
 	 * Gets the number of elements in the list
 	 * @return the number of elements
 	 */
-	inline uint16_t GetCount() const
+	inline int GetCount() const
```
TrueBrain: random question I don't know the answer to:

do we want to use `int32_t` and `INT32_MAX`, instead of `int` and `INT_MAX`, just to have it more clear to everyone what type we talk about exactly?

An `int` for a for-loop is ofc perfectly fine, but for things like this, I wonder if we should be more explicit. Open question, don't know the answer myself :)

rubidium42: That it is `int` is because of one of the properties of `int` that `int32_t` does not necessarily have. It's basically the default type things get promoted to. 

Going with `int` in the for-loops will eventually get you into problems if you need to convert it back to `int32_t` when some compiler makes `int` 64 bits. So, in my opinion, it's basically an all-or-nothing approach that should be taken to just disallow `int`/`unsigned int`/`uint`. That's definitely not in scope for this PR.

PeterN: This is one of the reasons I've slowly been working on things like `GetScrolledItemFromWidget()` where the caller doesn't have to do anything with different integer types.

TrueBrain: @rubidium42 sometimes it feels we talk about different dimensions :) Let me retry this ..

My for-loop remark was to avoid the common response: OMG HE WANTS TO CHANGE ALL `int` TO `int32_t`. I am not. A `for (int i = 0; i < 10; i++)` is perfectly fine in my world.

What I am questioning however, if we shouldn't be more explicit in our datatypes in structs and these kind of functions. `int32_t` is clear to everyone what is meant, where `int` is not always all that clear. And when someone would compile it on a 16bit compiler, would give different results ;)

As `int` is an `int32_t` on all compilers we care about, "promotion" isn't an issue, as they are identical in type. So the code would compile exactly the same with both types as far as I know.

So my only question is: for this PR, for the changes to `int`: shouldn't we be doing `int32_t` instead, to just make people's life a bit easier that we actually mean a 32bit signed value?

And this is just an honest question, mostly to avoid people starting to use `int` and `uint` in all sorts of data containers we have. Do we actually like / want that?

TrueBrain: And maybe to be more verbose in what I mean:

https://github.com/TrueBrain/OpenTTD/pull/28

Same result, just more explicit that we want a 32bit signed integer for our scroll location.

rubidium42: The code interfacing with this does use `int` quite often. For example `Window::GetRowFromWidget` which returns `int` and `INT_MAX`.
In my mind it would be conceptually weird to let that generate `INT_MAX` and then later check against `INT32_MAX`, so that should be changed to keep the types and constant the same... even though they happen to be the same type/value with our current compilers.

How I understood your comment about `int` in for-loops being fine was that you see no problem in:
```
int32_t start_value = 32;
int32_t end_value = 64;
for (int i = start_value; i < end_value; ++i) something(i);
```
That, in my mind, is just messy.

Finally, I've got the feeling such debates have happened in the past and just removing `int` and `uint` from the codebase completely would remove such debates in the future and just set a clear standard that everyone can abide by instead of asking open questions every now and then whether something should be `int` or `int32_t`.

TrueBrain: The reason we keep having certain debates, is because we are getting cleaner and cleaner code in terms of typing etc. But it is a slow process. But we keep pushing it forward, which has made the code a lot more readable over the last few years. These conversations are, how ever annoying, also healthy. Just so we deliberately make choices :)

I do not remember we ever talked about flat-out removing `int`, but I do remember we have had talks about making storage-types more explicit, and not depend on the "at least NNN" of `int` and friends.

Which brought me to my original question :) If we deliberately say: no, we use `int` here, I am fine with it. As long as we have a good reason for it, and not "because the old us did it, so I am going to do it". That way we never grow better, in our code base :) But you have been pushing most of these "moving forward", so I am preaching to the choir here :)

rubidium42: Yes, it should eventually probably be `int32_t`. However, this is likely going to be backported so I'm also trying to keep it relatively small while not mix-and-matching types.

Changing the types to `int32_t` would, in my opinion, also mean changing `GetScrolledRowFromWidget` to return `int32_t` and `INT32_MAX` in 'wrong position' situations. By proxy that should probably also affect `GetRowFromWidget` in the same way to keep them consistent, or keep the `pos` in `GetScrolledRowFromWidget` as `int` and check that against `INT_MAX`. Next to that all users of the scrollbar (and potentially) `GetRowFromWidget` should be checked and amended.

Or are we fine with only changing `int` to `int32_t` when we (have to) change the line?

PeterN: May I suggest a typedef (in `using x = y;` format) to define the type, and a constexpr of `std::numeric_limits<x>::max()` for the max value?

Keep it only to scrollbar for now for this PR for back-portability, and then later it we can update other code that uses it if we decide it needs it?

src/airport_gui.cpp
```diff
@@ -495,8 +495,8 @@ class BuildAirportWindow : public PickerWindowBase {
 				break;
 
 			case WID_AP_AIRPORT_LIST: {
-				int num_clicked = this->vscroll->GetScrolledRowFromWidget(pt.y, this, widget, 0, this->line_height);
-				if (num_clicked == INT_MAX) break;
+				int32_t num_clicked = this->vscroll->GetScrolledRowFromWidget(pt.y, this, widget, 0, this->line_height);
+				if (num_clicked == INT32_MAX) break;
```
PeterN, line 499: Should these be `Scrollbar::npos`? (Actual question, not sure myself...)

rubidium42: Yes they should, but also...

> Keep it only to scrollbar for now for this PR for back-portability, and then later it we can update other code that uses it if we decide it needs it?

Technically also replacing `int32_t` with `Scrollbar::size_type`, and everything that depends on that...
Practically making this PR not back-portable any more as it's just changing way too much.

PeterN: Ah yes, fair enough :)

PeterN (approved):

TrueBrain:
Should this be backported? (if so, please add label :) ).

MontyMontana:
I have tested this with the Object list of https://github.com/OpenTTD/OpenTTD/issues/12086 in nightly and the crash seems solved. Not tested in 14.0-RC3 since the merge was performed after the release.

PR #12180, opened 2024-02-25
Codefix: DrawEngineList does not accept EngineID.

PeterN:
## Motivation / Problem

`DrawEngineList()` accepts two `uint16_t` parameters, `min` and `max`, but the caller in autoreplace_gui thinks they are `EngineID`s and casts to that instead first.

## Description

Instead of just correcting the type from `EngineID` to `uint16_t`:

* Add `GetVisibleRangeIterators()` to `Scrollbar` which returns iterators to be first and last (+ 1) visible items. This has plenty of uses beyond this change.
* Replace the min/max parameters of `DrawEngineList()` with scrollbar reference, and use `GetVisibleRangeIterators()` instead of indices.

## Limitations

Ideally `GetVisibleRangeIterators()` would return a strong `std::pair<Tcontainer::const_iterator, Tcontainer::const_iterator>`  but it returns `auto` due to `std::span` (not used here, but intended to be used elsewhere later) not supporting `const_iterator` in C++20.

## Checklist for review

### DIFF
--- a/src/autoreplace_gui.cpp
+++ b/src/autoreplace_gui.cpp
@@ -33,7 +33,7 @@
 
 #include "safeguards.h"
 
-void DrawEngineList(VehicleType type, const Rect &r, const GUIEngineList &eng_list, uint16_t min, uint16_t max, EngineID selected_id, bool show_count, GroupID selected_group);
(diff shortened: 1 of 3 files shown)


2TallTyler (dismissed):
Makes sense to me.

rubidium42 (approved):

PR #12186, opened 2024-02-27
Codechange: Use vehicle tile hash to build depot vehicle list.

PeterN:
## Motivation / Problem

`BuildDepotVehicleList()` iterates the vehicle pool, and the first things it checks are vehicle type and tile. This means it can easily be improved (performance-wise) by using the vehicle tile hash instead.

Depot vehicle lists are used when performing bulk depot commands, as well as for displaying the vehicles in a depot.

## Description

Use `FindVehicleOnPos()` to build the depot vehicle list instead of iterating the entire vehicle pool.

Observed performance measurements:

| Save           | Depot | Master   | This PR |
| -------------- | ----- | -------- | ------- |
| Wentbourne     | Train | 1,073us  | 7us+    |
| Wentbourne     | Ship  | 714us    | 7us+    |
| Xarick's 4kx4k | Train | 13,386us | 25us+   |
| Xarick's 4kx4k | Ship  | 8,734us  | 25us+   |
| Xarick's 4kx4k | Road  | 9,248us  | 25us+   |

Due to the nature of this change, performance changes from relatively constant to being proportional to the number of vehicles in the depot, simply because the number of vehicles to check changes.

## Checklist for review

### DIFF
--- a/src/vehiclelist.cpp
+++ b/src/vehiclelist.cpp
@@ -9,6 +9,7 @@
 
 #include "stdafx.h"
 #include "train.h"
+#include "vehicle_func.h"
 #include "vehiclelist.h"
 #include "vehiclelist_func.h"
 #include "group.h"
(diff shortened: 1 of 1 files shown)


src/vehiclelist.cpp
```diff
@@ -72,32 +106,8 @@ void BuildDepotVehicleList(VehicleType type, TileIndex tile, VehicleList *engine
 	engines->clear();
 	if (wagons != nullptr && wagons != engines) wagons->clear();
 
-	for (const Vehicle *v : Vehicle::Iterate()) {
-		/* General tests for all vehicle types */
-		if (v->type != type) continue;
```
J0anJosep, line 77: You are reordering the checks to v->IsInDepot().
I wonder which part of the performance improvement is due to using the vehicle tile hash and which part is related to checking this earlier.
Probably, it is mostly due to vehicle tile hash, but it would be great to compare them.

Can you provide the performance measurements against something like this?

```
for (const Vehicle *v : Vehicle::Iterate()) {
	/* General tests for all vehicle types */
	if (v->type != type || !v->IsInDepot()) continue;
	if (v->tile != tile) continue;

	if (type ==VEH_TRAIN) {
		const Train *t = Train::From(v);
		if (t->IsArticulatedPart() || t->IsRearDualheaded()) continue;
		if (wagons != nullptr && t->First()->IsFreeWagon()) {
			if (individual_wagons || t->IsFreeWagon()) wagons->push_back(t);
			continue;
		}
	}
	if (v->IsPrimaryVehicle()) engines->push_back(v);
}
```

PeterN: With the `v->tile != tile` check moved first to make it fairer (and more efficient) as well:

Wentbourne Train: ~620us
Wentbourne Ship: ~700us
Xarick's Train: ~7,500us
Xarick's Ship: ~7,500us
Xarick's Road: ~7,500us

Basically, checking `IsInDepot()` should always have been done earlier, as it always needs to be done anyway and means it was unnecessarily checking train properties. But it's no where near like using the tile hash.

rubidium42 (approved):

src/vehiclelist.cpp
J0anJosep: `Basically, checking IsInDepot() should always have been done earlier, as it always needs to be done anyway and means it was unnecessarily checking train properties. But it's no where near like using the tile hash.`

Yes, I agree with you. But separating those changes in two different commits or adding a comment for this change in the PR would have been better.

PR #12192, opened 2024-02-28
Codefix: Remove no-longer used ship special-case.

PeterN:
## Motivation / Problem

`ChooseShipTrack()` has some special casing for when a ship reaches its destination tile, however this special-case has not been triggered since multi-tile docks were introduced, as v->dest_tile points to the station sign tile instead.

## Description

The special case appears to be designed to keep a ship travelling in the same direction instead of turning when it reaches the dock. This seems to assume the ship is approaching the dock head on, which was never always the case, and ships are since allowed to stop either side of the dock as well now. As this code path hasn't been triggered for a few years, simply remove the special case,

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf.h
+++ b/src/pathfinder/yapf/yapf.h
@@ -22,11 +22,10 @@
  * @param v        the ship that needs to find a path
  * @param tile     the tile to find the path from (should be next tile the ship is about to enter)
  * @param enterdir diagonal direction which the ship will enter this new tile from
- * @param tracks   available tracks on the new tile (to choose from)
  * @param path_found [out] Whether a path has been found (true) or has been guessed (false)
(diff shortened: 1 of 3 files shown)


Kuhnovic:
Less special casing, I like it!

TrueBrain (approved):
I am curious how you know it hasn't been triggered in a few years :P :P

If @Kuhnovic likes it, it is good enough for me!

PeterN:
Since multi-tile docks, when ships head to a station v->dest_tile points to a land tile instead of the docking tile, so ships can never reach it.

I was not quite right though, this condition is still reached when the destination is a depot, or can be reached if the destination is a buoy, but unless the pathfinder is broken (which it may be), then normal pathfinding should be fine there.

PR #12194, opened 2024-02-29
Fix: Changing NPF max search nodes while in-game had no effect

SamuXarick:
## Motivation / Problem
Changing NPF max search nodes while in-game via console has no effect.

## Description
Fix it so that it takes effect immediately.

## Checklist for review

### DIFF
--- a/src/pathfinder/npf/npf.cpp
+++ b/src/pathfinder/npf/npf.cpp
@@ -1040,6 +1040,10 @@ static NPFFoundTargetData NPFRouteInternal(AyStarNode *start1, bool ignore_start
 	/* Initialize user_data */
 	_npf_aystar.user_data = user;
 
+	/* We will limit the number of nodes for now, until we have a better
+	 * solution to really fix performance */
+	_npf_aystar.max_search_nodes = _settings_game.pf.npf.npf_max_search_nodes;
+
 	/* GO! */
 	[[maybe_unused]] int r = _npf_aystar.Main();
 	assert(r != AYSTAR_STILL_BUSY);
@@ -1113,9 +1117,6 @@ void InitializeNPF()
 	}
 	_npf_aystar.loops_per_tick = 0;
 	_npf_aystar.max_path_cost = 0;
-	/* We will limit the number of nodes for now, until we have a better
-	 * solution to really fix performance */
-	_npf_aystar.max_search_nodes = _settings_game.pf.npf.npf_max_search_nodes;
 }
 
 static void NPFFillWithOrderData(NPFFindStationOrTileData *fstd, const Vehicle *v, bool reserve_path = false)


src/pathfinder/npf/npf.cpp
```diff
@@ -1040,10 +1040,18 @@ static NPFFoundTargetData NPFRouteInternal(AyStarNode *start1, bool ignore_start
 	/* Initialize user_data */
 	_npf_aystar.user_data = user;
 
+	/* We will limit the number of nodes for now, until we have a better
+	 * solution to really fix performance */
+	uint32_t max_search_nodes_before = _npf_aystar.max_search_nodes;
+	_npf_aystar.max_search_nodes = _settings_game.pf.npf.npf_max_search_nodes;
```
rubidium42: This looks like it's doing what `AutoRestoreBackup` was designed/created for; temporarily setting a value to something different and then (automatically) reverting to the original state.

However, I doubt whether there's any use in restoring the value, as you'd be restoring it to the value when the game was loaded and this is the only function using it.

src/pathfinder/npf/npf.cpp
```diff
@@ -1040,6 +1041,10 @@ static NPFFoundTargetData NPFRouteInternal(AyStarNode *start1, bool ignore_start
 	/* Initialize user_data */
 	_npf_aystar.user_data = user;
 
+	/* We will limit the number of nodes for now, until we have a better
+	 * solution to really fix performance */
+	AutoRestoreBackup max_search_nodes(_npf_aystar.max_search_nodes, _settings_game.pf.npf.npf_max_search_nodes);
```
rubidium42: What is the purpose of restoring the value? Where/when is the restored value used?

SamuXarick: I'm unsure if on new game the river generator sets a new value from fresh, or re-uses the existing one. Just wanted to be safe than sorry.

rubidium42: That comment makes the reasoning of this PR even weirder.

You don't know whether the river generator sets the value itself, so you remove the only thing that initialises it before the river generator runs.
The PR is about changing max search nodes not having an effect, well if river generation would be using it, it would now be using a max search nodes that was never initialised.

Even then, `_npf_aystar` is `static` in this file, so it should not be "leaking" to any other place unless that actually uses NPF. The river generator uses only `AyStar` and has its own instance allocated on the stack with infinite search nodes. So the river generator does not use the NPF one.

SamuXarick: `AutoRestoreBackup` is not needed for River generation, just tested 🙂.
`AyStar` has a 10000 `max search nodes` when it `Init` by default.
`NPF` uses `AyStar`, so inits with 10k.
RiverGen uses `AyStar`, so inits with 10k.
Each `NPFRouteInternal` call still sets the value of `max search nodes` to what the user has in the settings.
I guess that's safe enough for between game starts, load saves uses the saved setting.

rubidium42 (approved):

PR #12199, opened 2024-03-01
Codechange: Don't use NOT_REACHED() when catching unhandled thread exceptions

glx22:
## Motivation / Problem
When an unhandled thread exception is caught, a "NOT_REACHED" error is issued. 
image

This is not very helpful.

## Description
Output a little more useful error.
image

This still won't help to know where it was thrown, but at least it gives the reason.

## Checklist for review

### DIFF
--- a/src/thread.h
+++ b/src/thread.h
@@ -12,6 +12,7 @@
 
 #include "debug.h"
 #include "crashlog.h"
+#include "error_func.h"
 #include <system_error>
 #include <thread>
 #include <mutex>
@@ -62,6 +63,8 @@ inline bool StartNewThread(std::thread *thr, const char *name, TFn&& _Fx, TArgs&
 				try {
 					/* Call user function with the given arguments. */
 					F(A...);
+				} catch (std::exception &e) {
+					FatalError("Unhandled exception in {} thread: {}", name, e.what());
 				} catch (...) {
 					NOT_REACHED();
 				}


src/thread.h
```diff
@@ -62,8 +63,8 @@ inline bool StartNewThread(std::thread *thr, const char *name, TFn&& _Fx, TArgs&
 				try {
 					/* Call user function with the given arguments. */
 					F(A...);
-				} catch (...) {
-					NOT_REACHED();
+				} catch (std::exception &e) {
+					FatalError("Unhandled exception in {} thread: {}", name, e.what());
```
rubidium42, line 67: Not everything that's thrown is a (sub class of) `std::exception`. For example win32_s.cpp throwing C-string pointers, or `OTTDByteReaderSignal` and `AbortGenerateWorldSignal` that are not a sub-class of `std::exception`.

Maybe retaining the `catch (...)` with `NOT_REACHED()` is better over the default `abort()` behaviour? So first catching `std::exception` and if that doesn't match falling back to catching `...`.

glx22: `OTTDByteReaderSignal` should never reach this point (newgrf code will catch it).
`AbortGenerateWorldSignal` is handle in `_GenerateWorld()` so should not reach this point either.
win32_s.cpp ones are also handled (in `SoundDriver_Win32::Start()`).
But I agree, I can keep the `...` just in case.

rubidium42 (approved):

PR #12200, opened 2024-03-01
Codechange: Replace custom rounded square root logic with standard library functions

1f5:
## Motivation / Problem
The old function is complicated and slow, because it calculates the square root using only integers. This might be useful for code running on systems where no floating point numbers are available, but we don't have such limitations. Using standard library functions is faster and need less code.

## Description
Replaces the whole function logic by simply getting the square root of the input and rounding it. The output is cast to uint32_t.

## Limitations
Performance increase should not be noticeable during normal gameplay.
Does not remove any tests.

## Checklist for review

### DIFF
--- a/src/core/math_func.cpp
+++ b/src/core/math_func.cpp
@@ -37,28 +37,8 @@ int DivideApprox(int a, int b)
  * Compute the integer square root.
  * @param num Radicand.
  * @return Rounded integer square root.
- * @note Algorithm taken from http://en.wikipedia.org/wiki/Methods_of_computing_square_roots
  */
 uint32_t IntSqrt(uint32_t num)
 {
-	uint32_t res = 0;
-	uint32_t bit = 1UL << 30; // Second to top bit number.
-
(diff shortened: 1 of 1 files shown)


LordAro:
Definitely going to need some numbers on "it's faster"

LordAro:
Probably on a variety of platforms & architectures (we support many, remember)

frosch123:
OpenTTD multi-player relies on a deterministic progression of the game state.
That's why all game logic only uses integers, and no floats.
I am not aware of `IntSqrt` being part of some performance critical code-path.

As such I value potential non-determinism/desyncs worse than some CPU cycles.

rubidium42:
OpenTTD's logic for the state loops does not use floating points numbers/calculations. This rule was introduced to prevent desyncs due to implementations of floating point logic differing between platforms.

So, the bar for adding floating point calculations in the state loop is very high as you need to essentially prove that there is no platform where the floating point square root behaves different.

1f5:
Thanks for clarifying this.

PR #12205, opened 2024-03-01
Change: [Script] Match FormatString behaviour more closely

glx22:
## Motivation / Problem
GS string validation is still too strict compared to `FormatString()`.
image
Using the save from 13.4 (so town extra string is stored)
image
Master after #12187 (better but GS still crash)
image

## Description
`FormatString()` uses an array of 20 (SCRIPT_TEXT_MAX_PARAMETERS) parameters initialised with `0` to format a string from GS.
So do the same for the validation by filling extra `0` in parameter list.
`FormatString()` checks a parameter is used only for one command type (for instance, a parameter used in {STRING} can't be used later in {STRING2}), and don't consume the expected number of args in this case (they will be available to the next command).

So we do the same for the validation by filling extra `0` in parameter list, and by implementing a type check mechanism to match `FormatString()` behaviour.

I also noticed GSText parameters could be added more than once in the encoded string, so I also fixed that.

And I reduced the string consumption check from error to warning.

image

In the end the encoded string is the same as in 13.4, but with some extra `:0` at the end depending on validation process (`FormatString()` would have these extra `0` anyway).

## Checklist for review

### DIFF
--- a/src/script/api/script_text.cpp
+++ b/src/script/api/script_text.cpp
@@ -184,10 +184,19 @@ void ScriptText::_FillParamList(ParamList &params, ScriptTextList &seen_texts)
 	}
 
 	seen_texts.pop_back();
+
+	/* Fill with dummy parameters to match FormatString() behaviour. */
+	if (seen_texts.empty()) {
+		static Param dummy = 0;
+		int nb_extra = SCRIPT_TEXT_MAX_PARAMETERS - (int)params.size();
+		for (int i = 0; i < nb_extra; i++)
+			params.emplace_back(-1, i, &dummy);
+	}
 }
 
-void ScriptText::ParamCheck::Encode(std::back_insert_iterator<std::string> &output)
+void ScriptText::ParamCheck::Encode(std::back_insert_iterator<std::string> &output, const char *cmd)
 {
+	if (this->cmd == nullptr) this->cmd = cmd;
 	if (this->used) return;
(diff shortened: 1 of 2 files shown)


src/script/api/script_text.cpp
```diff
@@ -220,46 +229,54 @@ void ScriptText::_GetEncodedText(std::back_insert_iterator<std::string> &output,
...
+				case StringParam::STRING:
+				{
 					ParamCheck &p = *get_next_arg();
-					p.Encode(output);
-					if (!std::holds_alternative<SQInteger>(*p.param)) ScriptLog::Error(fmt::format("{}({}): {{{}}} expects an integer", name, param_count + i + 1, cur_param.cmd));
+					p.Encode(output, cur_param.cmd);
+					if (p.cmd != cur_param.cmd) throw 1;
+					if (!std::holds_alternative<ScriptTextRef>(*p.param)) {
+						ScriptLog::Error(fmt::format("{}({}): {{{}}} expects a GSText", name, param_count + 1, cur_param.cmd));
+						throw 1;
```
rubidium42: Why not something like
```
param_count++;
continue;
```
instead?

I'm seeing only one throw, and in this case you just want to add something to `param_count` and then continue with the next parameter, right?

glx22: There are more than one throw, but changing this one allows me to put a warning in the `catch`.

TrueBrain (approved):
I do not understand any of this, but I am sure you tested it :) And assuming the open conversation is actually resolved, let's go!

PR #12206, opened 2024-03-01
Fix #12193: Don't use the entire docking area when computing closest station tile

### LINKED ISSUE #12193: [Bug]: CalcClosestStationTile on a huge docking area makes Yapf fail pathfinding on first attempt
### Version of OpenTTD

20240228-master-gc4f9479507

### Expected result

1. `CalcClosestStationTile` (or some other) should return all of the 8 docking tiles.
2. High level pathfinder returns a path towards the east side.
3. Low level pathfinder estimate cost is computed for each of the 8 destinations.
4. Low level pathfinder finds the optimal path on attempt 0, a path towards the southwest side.

### Actual result

1. `CalcClosestStationTile` returns a single tile and it happens to match ship's current position.
2. High level pathfinder returns a path towards the east side.
3. Low level pathfinder fails to return a path on attempt 0. It is trying to pathfind to the ship's current position.
4. Low level pathfinder finds an unoptimal path on attempt 1, restricted to a corridor. The path towards southwest would be the shortest should the low level pathfinder not fail on attempt 0.

### Steps to reproduce

[dock station spread.zip](https://github.com/OpenTTD/OpenTTD/files/14440710/dock.station.spread.zip)
The ship leaving the depot is inside the huge docking tile area of the dock station and wants to find the shortest path to it. Only 8 tiles are actual docking points for the ship, and they are in 3 separate water regions. The 3 docks belong to the same station.

SamuXarick:
## Motivation / Problem

Docks have a `TileArea` of tiles designated for ships to dock at, and pathfinders require one of those tiles as the target to compute a path. `CalcClosestStationTile` is used for that effect, however it is flawed. The area contains more tiles than necessary. In #12193, 3 docks are placed far apart from each other, and all 3 belong to the same station. This generates a very large area covering nearly the entire map. If the ship is inside this area, `CalcClosestStationTile` returns the ship's current position as the target for pathfinders, essentially saying the ship is already there. This is not handled properly and the result is sub-optimal paths at best.

For YAPF, the search is compounded by a high level pf and a low level pf with two attempts. The high level pf finds a path without issues to the region which requires the least number of regions to traverse, but in this example, that path ends up being sub-optimal. The low level pf is affected, as it's here that `CalcClosestStationTile` is used, and that uses the ship's current position for the first attempt. Though it correctly paths to itself, it fails `IsShipDestinationTile` test.  On the second attempt, the search is restricted to a corridor of regions and a path is found, but it makes the ship take a big detour to the back side due to the restrictions. This path is doubly sub-optimal.

## Description
Fix #12193
Don't use `CalcClosestStationTile` for ships. Instead, use a specialized `GetShipDestinationTiles` which creates a list of possible destination tiles. For docks, it only takes into consideration the tiles that pass both `IsDockingTile` and `IsShipDestinationTile` tests. For the other cases it just takes the ship's current `dest_tile`.

Modified `CYapfDestinationTileWaterT` class to accept multiple destinations. In this manner, the estimate cost is calculated for each of the destination tiles and the shortest estimate is returned. Some adaptation was necessary to make this possible to work for both the high- and low-level pathfinders and the existing functions. `ChooseShipTrack` and its `FindWaterRegionPath` brother both will take the same destination tiles which are calculated at either `YapfShipChooseTrack` or `YapfShipCheckReverse`.

## Limitations
None, now that NPF was removed.

## Checklist for review

### DIFF
--- a/src/pathfinder/pathfinder_func.h
+++ b/src/pathfinder/pathfinder_func.h
@@ -12,6 +12,40 @@
 
 #include "../tile_cmd.h"
 #include "../waypoint_base.h"
+#include "../ship.h"
+
+/**
+ * Creates a list containing possible destination tiles for a ship.
+ * @param v The ship
+ * return Vector of tiles filled with all possible destinations.
+ */
+inline std::vector<TileIndex> GetShipDestinationTiles(const Ship *v)
+{
+	std::vector<TileIndex> dest_tiles;
+
+	if (v->current_order.IsType(OT_GOTO_STATION)) {
+		const StationID station = v->current_order.GetDestination().ToStationID();
+
+		const BaseStation *st = BaseStation::Get(station);
+		TileArea ta = st->GetTileArea(StationType::Dock);
+
+		/* If the dock station is (temporarily) not present, use the station sign to drive near the station. */
+		if (ta.tile == INVALID_TILE) {
+			dest_tiles.push_back(st->xy);
+		} else {
+			for (const TileIndex &docking_tile : ta) {
+				if (!IsDockingTile(docking_tile) || !IsShipDestinationTile(docking_tile, station)) continue;
+				dest_tiles.push_back(docking_tile);
+			}
+		}
+	} else {
+		dest_tiles.push_back(v->dest_tile == INVALID_TILE ? TileIndex{} : v->dest_tile);
+	}
+
+	assert(!dest_tiles.empty());
+
+	return dest_tiles;
+}
 
 /**
  * Calculates the tile of given station that is closest to a given tile
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -32,8 +32,7 @@ public:
 	typedef typename Node::Key Key; ///< key to hash tables.
 
 protected:
-	TileIndex dest_tile;
-	TrackdirBits dest_trackdirs;
+	std::span<TileIndex> dest_tiles;
 	StationID dest_station;
 
 	bool has_intermediate_dest = false;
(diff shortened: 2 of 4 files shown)


Kuhnovic:
First of all, I think you should focus on YAPF. NPF can be considered a legacy PF and is most likely going to be removed at some point.

You mention suboptimality being a problem, but the high level / low level ship pathfinder never claims to be optimal. It claims to be very fast, and will return near optimal paths in most cases. Near optimal, definitely not optimal.

To me, making the pathfinder "a bit slower" (did you measure this? How much slower?) to support an edge case seems wrong. Furthermore, the bug you are fixing is not really a bug. Taking two attempts to find the path, the second one restricted, is a feature of the ship pathfinder. It might not yield the shortest path, and you might not like that. But it is finding the path when there is one, and it is a correct path.

All of that said, I think there are some good elements to this change. Considering only docking tiles as viable destination tiles makes sense.

SamuXarick:
image

SamuXarick:
image


> To me, making the pathfinder "a bit slower" (did you measure this? How much slower?) to support an edge case seems wrong. 
Just solved this. It's as fast as master now, pretty much.

src/pathfinder/pathfinder_func.h
```diff
@@ -31,6 +65,19 @@ inline TileIndex CalcClosestStationTile(StationID station, TileIndex tile, Stati
...
+		uint closest_dist = std::numeric_limits<uint>::max();
+		TileIndex closest_tile;
+		for (const TileIndex &docking_tile : ta) {
+			if (!IsDockingTile(docking_tile) || !IsShipDestinationTile(docking_tile, station)) continue;
+			const uint dist = DistanceManhattan(docking_tile, tile);
+			if (dist >= closest_dist) continue;
+			closest_dist = dist;
+			closest_tile = docking_tile;
+		}
+		return closest_tile;
```
PeterN: I've not read the whole PR to know how this works, but by picking the "closest" docking location by direct distance means you can end up picking a location that cannot be reached, or the path to it could be much further than the path to another docking location.

EDIT: Probably the same could be said of the existing tile area approach, so I suppose it would not be a regression as such. But finding the shortest path is the pathfinder's job, and I think this detracts from that.

PR #12208, opened 2024-03-02
Fix #12076: Do not allow 'join' command on dedicated servers

### LINKED ISSUE #12076: [Bug]: As a dedicated server, `join` fails but does not print anything.
### Version of OpenTTD

8d9fa0ea89 on Windows 10, 64-bit (likely present since 04709f279809125e24c3652f182eac0e56c1d90f)

### Expected result

When running `join` as a dedicated server, the command should output an error saying it does not work there.

### Actual result

When running `join` as a dedicated server, the command does not output any error message despite it failing.

### Steps to reproduce

1. Start a dedicated server: `openttd -D`.
2. Load a savegame containing a human-player company, or have somebody join the server and create a company.
3. Run `join` on the newly created company.
4. The command does not do anything, nor does it output an error.
5. This can be verified by doing `join 255` and receiving the message `You are already there!`.

rubidium42:
## Motivation / Problem

Deep down `NetworkServerDoMove` silently ignores calls for the dedicated server, but the user typing the `join` command does not get notified of the failure.

## Description

Add a new hook that prevents commands from being executed on the dedicated server, and use it for the `join` command.

## Limitations

None I can think of.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -156,6 +156,21 @@ DEF_CONSOLE_HOOK(ConHookNeedNetwork)
 }
 
 /**
+ * Check whether we are in a multiplayer game and are playing, i.e. we are not the dedicated server.
+ * @return Are we a client or non-dedicated server in a network game? True when yes, false otherwise.
+ */
+DEF_CONSOLE_HOOK(ConHookNeedNonDedicatedNetwork)
+{
+	if (!NetworkAvailable(echo)) return CHR_DISALLOW;
+
+	if (_network_dedicated) {
+		if (echo) IConsolePrint(CC_ERROR, "This command is not available to a dedicated network server.");
+		return CHR_DISALLOW;
+	}
+	return CHR_ALLOW;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


glx22:
Other commands could use it.
I think `ConPart` does this test internally.
And I'm quite sure some more commands use internal checks and not hooks where it's possible

rubidium42:
> I think `ConPart` does this test internally. 

This command also works in non-network, so it's a different requirement/hook that would be needed. That hook could also be used for `ConFramerateWindow`. But, that's beyond the scope of fixing the bug that a similar check is missing in `join`.


> I'm quite sure some more commands use internal checks and not hooks where it's possible

I would place those out of scope for this PR that fixes a bug.


> Other commands could use it.

I am not aware of any that could. Yes besides `part` and `fps_wnd`, `move` uses `_network_dedicated` but only to guard against moving the dedicated server client. As dedicated server you are still allowed to move other clients.

glx22 (approved):

PR #12209, opened 2024-03-02
Remove: NPF

merni-ns:
## Motivation / Problem
- YAPF exists since [r4897](https://github.com/OpenTTD/OpenTTD/commit/5e73dce0e71791b87e5b096a890578eefcc26639) in 2006, that is 18 years ago.
- It has better performance and functionality than NPF.
- I have not found any example either personally or by searching internet/forums where NPF is better than YAPF. (If there is any, please provide it here.) 
- Surveys show consistently more than 99% of users (of versions surveyed so far, i.e. recent JGRPP and 14.0 betas/RCs) use YAPF, for example see [here](https://survey.openttd.org/summaries/2024/wk08/14.0-RC1#game.settings.pf.pathfinder_for_roadvehs). That is true for every survey in every version in the last seven weeks. More accurate data will be available after 14.0 releases.
- Removing NPF removes the maintenance burden of supporting two pathfinders.

## Description
- Remove NPF.
- Remove the settings to switch pathfinders.
- AyStar is retained as it is used for other purposes (river generation) as well.

## Limitations
- Some small proportion of users might actually want to use NPF.
- If another pathfinder is ever added, the settings will have to be added again.

## Checklist for review


TrueBrain:
> * Surveys show consistently more than 99% of users (of versions surveyed so far, i.e. recent JGRPP and 14.0 betas/RCs) use YAPF, for example see [here](https://survey.openttd.org/summaries/2024/wk08/14.0-RC1#game.settings.pf.pathfinder_for_roadvehs). That is true for every survey in every version in the last seven weeks.

For now, this feels like a premature conclusion. I get where you are coming from, and you are likely even right. But let's wait for the 14.0 release, and what those surveys say. It might be that RCs are only done by users with a certain taste :) Similar with JGRPP.

As I have mentioned a few times already, we have to be careful with the survey to not over-conclude on a certain group of people. JGRPP is one group. People who are willing and capable using a testing version are another. Both might not represent the common folks.

That said, we can all predict the numbers of 14.0 will be the same, but still .. if we use this as argument, let's use it the right way :D

src/vehicle_type.h
```diff
@@ -56,7 +56,7 @@ static const VehicleID INVALID_VEHICLE = 0xFFFFF; ///< Constant representing a n
 /** Pathfinding option states */
 enum VehiclePathFinders {
 	// Original PathFinder (OPF) used to be 0
-	VPF_NPF  = 1, ///< New PathFinder
+	// New PathFinder (NPF) used to be 1
 	VPF_YAPF = 2, ///< Yet Another PathFinder
```
2TallTyler: Is this still referenced anywhere now that it's the only option?

src/saveload/compat/settings_sl_compat.h
```diff
@@ -95,13 +95,9 @@ const SaveLoadCompat _settings_sl_compat[] = {
...
-	SLC_VAR("pf.new_pathfinding_all"),
-	SLC_VAR("pf.yapf.ship_use_yapf"),
-	SLC_VAR("pf.yapf.road_use_yapf"),
-	SLC_VAR("pf.yapf.rail_use_yapf"),
-	SLC_VAR("pf.pathfinder_for_trains"),
-	SLC_VAR("pf.pathfinder_for_roadvehs"),
-	SLC_VAR("pf.pathfinder_for_ships"),
+	SLC_NULL(1, SL_MIN_VERSION, SLV_87),
+	SLC_NULL(3, SLV_28, SLV_87),
+	SLC_NULL(3, SLV_87, SLV_TABLE_CHUNKS),
```
2TallTyler, line 100: Why did you choose this old version instead of creating `SLV_REMOVE_NPF`?

src/settings_type.h
```diff
@@ -412,42 +412,11 @@ struct ScriptSettings {
...
-};
-
 /** Settings related to the yet another pathfinder. */
 struct YAPFSettings {
 	bool   disable_node_optimization;        ///< whether to use exit-dir instead of trackdir in node key
 	uint32_t max_search_nodes;                 ///< stop path-finding when this number of nodes visited
 	uint32_t maximum_go_to_depot_penalty;      ///< What is the maximum penalty that may be endured for going to a depot
-	bool   ship_use_yapf;                    ///< use YAPF for ships
-	bool   road_use_yapf;                    ///< use YAPF for road
-	bool   rail_use_yapf;                    ///< use YAPF for rail
```
2TallTyler, line 447: Are these not referenced anywhere? I don't see any other change to them in this PR.

2TallTyler (commented):
The latest summaries for [14.0](https://survey.openttd.org/summaries/2024/wk18/14.0) and [14.1](https://survey.openttd.org/summaries/2024/wk18/14.1#game.settings.pf.pathfinder_for_roadvehs) continue to show that NPF is effectively unused. Let's blow this popsicle stand. 😃

src/settings_type.h
glx22: They were used in afterload.

src/saveload/compat/settings_sl_compat.h
rubidium42: Probably because the old way of storing went away with the introduction of "table chunks" and as such the more descriptive storing of the settings (with name) in the saves. Normally the versioning is configured to the settings ini files, and global variables are allocated to temporarily save the information in when it's needed for afterload purposes. In these cases they can completely be removed from the settings ini file, but for loading old games we still need to know which bytes of the pre-table-chunks saves we need to skip.

This has been done this way for other settings that got removed.

src/vehicle_type.h
rubidium42: Given the whole choice for pathfinders has been completely ripped out instead of just the NPF case, this enumeration won't be used so it ought to be removed as well I'd say.

Alternatively the setting for pathfinder is not removed, but forced to YAPF, and the original switches are returned without the NPF case. That would make introducing a new pathfinder easier, but how likely is that given YAPF is almost 18 years old and NPF is just over 19 years old. 

Given OpenTTD is just over 20 years old, the pathfinders were all introduced in the first ~10% of its lifetime and after that it's been silent. So a new pathfinder seems to be quite unlikely, and as such I'd vote for just removing the enumeration.

src/saveload/compat/settings_sl_compat.h
merni-ns: Further context: https://weblogs.openttd.org/openttd/2024/03/02.html#132949-195

src/vehicle_type.h
merni-ns: It's a nice bit of history, but I agree it's better to remove it.

2TallTyler (approved):

LordAro (approved):
SHIP IT

PR #12210, opened 2024-03-02
Fix #12147: reset all saved settings to their default before loading a game

### LINKED ISSUE #12147: [Bug]: Infinite money mode enabled for old games
### Version of OpenTTD

14.0rc1

### Expected result

Infinite money mode should be disabled when loading old games before infinite money mode was implemented.

### Actual result

Infinite money mode is enabled when loading old games, if infinite money mode is enabled in new game settings.

### Steps to reproduce

1. Enable infinite money mode in settings.
2. Load an older save game.
3. Infinite money mode is active.

TrueBrain:
## Motivation / Problem

Fixes #12147.

When adding new settings, we no longer have to bump the savegame and write afterload code to ensure a good value for the new setting. At least, that was the idea. Sadly, we never actually implemented that part, so the value of new settings was what-ever the value was of the last game played.

## Description

Always first reset the settings to their default (for settings that are suppose to be in the savegame). After that, load the savegame. This gives a predictable set of settings before going into the loading routine, instead of using what-ever was used in the last game.

## Limitations

ListSettingDesc isn't used for any setting that is saved in the savegame. Although the implementation should be trivial (it can be taken from `ParseValue`), there is no way to validate it is. As such, I left it as a `NOT_REACHED()`, to prevent it having an implementation that is broken.

## Checklist for review

### DIFF
--- a/src/saveload/settings_sl.cpp
+++ b/src/saveload/settings_sl.cpp
@@ -187,10 +187,20 @@ struct PATSChunkHandler : ChunkHandler {
 
 	void Load() const override
 	{
-		/* Copy over default setting since some might not get loaded in
-		 * a networking environment. This ensures for example that the local
-		 * currency setting stays when joining a network-server */
-		LoadSettings(this->GetSettingTable(), &_settings_game, _settings_sl_compat);
+		const auto settings_table = this->GetSettingTable();
+
+		/* Reset all settings to their default, so any settings missing in the savegame
+		 * are their default, and not "value of last game". AfterLoad might still fix
+		 * up values to become non-default, depending on the saveload version. */
+		for (auto &desc : settings_table) {
+			const SettingDesc *sd = GetSettingDesc(desc);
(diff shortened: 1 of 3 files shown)


src/saveload/settings_sl.cpp
```diff
@@ -187,10 +187,22 @@ struct PATSChunkHandler : ChunkHandler {
 
 	void Load() const override
 	{
-		/* Copy over default setting since some might not get loaded in
-		 * a networking environment. This ensures for example that the local
-		 * currency setting stays when joining a network-server */
-		LoadSettings(this->GetSettingTable(), &_settings_game, _settings_sl_compat);
+		const auto settings_table = this->GetSettingTable();
+
+		GameSettings copy = _settings_game;
```
rubidium42: This copy does not seem to be used.

TrueBrain: oops, left-over from testing :D

rubidium42 (approved):

PR #12211, opened 2024-03-02
Codechange: Use templates to handle different station speclist types.

PeterN:
## Motivation / Problem

Custom NewGRF stations and roadstops have a per-station list of index to custom spec mappings, which are almost the same other than the spec type.

Code is duplicated between the two for each type.

And in true fashion "I have for patch that" that adds another custom spec type, which would again duplicate all the code.

## Description

Use templates to define the custom spec mapping list. This allows reusing code for each spec type.

`StationSpecList` becomes `SpecMapping<StationSpec>`
`RoadStopSpecList` becomes `SpecMapping<RoadStopSpec>`

AllocateStationSpec/DeallocateStationSpec is not yet templated, as these are slightly different, so that is a separate WIP.

## Limitations

Saveload code for station spec and roadstop spec isn't quite the same so there's a type check to handle this slight difference, though perhaps it isn't needed.

Some saveload version changes don't actually appliy to roadstops as they didn't exist in older versions, but on the other hand they didn't exist in older versions so it won't cause any harm.

## Checklist for review

### DIFF
--- a/src/saveload/station_sl.cpp
+++ b/src/saveload/station_sl.cpp
@@ -205,9 +205,9 @@ static void SwapPackets(GoodsEntry *ge)
 class SlStationSpecList : public DefaultSaveLoadHandler<SlStationSpecList, BaseStation> {
 public:
 	inline static const SaveLoad description[] = {
-		SLE_CONDVAR(StationSpecList, grfid,    SLE_UINT32,                SLV_27,                    SL_MAX_VERSION),
-		SLE_CONDVAR(StationSpecList, localidx, SLE_FILE_U8 | SLE_VAR_U16, SLV_27,                    SLV_EXTEND_ENTITY_MAPPING),
-		SLE_CONDVAR(StationSpecList, localidx, SLE_UINT16,                SLV_EXTEND_ENTITY_MAPPING, SL_MAX_VERSION),
+		SLE_CONDVAR(SpecMapping<StationSpec>, grfid,    SLE_UINT32,                SLV_27,                    SL_MAX_VERSION),
+		SLE_CONDVAR(SpecMapping<StationSpec>, localidx, SLE_FILE_U8 | SLE_VAR_U16, SLV_27,                    SLV_EXTEND_ENTITY_MAPPING),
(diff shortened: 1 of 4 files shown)


src/saveload/station_sl.cpp
```diff
@@ -202,65 +202,43 @@ static void SwapPackets(GoodsEntry *ge)
...
+		SlSetStructListLength(speclist.size());
+		for (auto &sm : speclist) {
+			SlObject(&sm, this->GetDescription());
 		}
 	}
 
 	void Load(BaseStation *bst) const override
 	{
-		uint8_t num_specs = IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH) ? last_num_specs : (uint8_t)SlGetStructListLength(UINT8_MAX);
+		size_t num_specs = std::is_same_v<T, StationSpec> && IsSavegameVersionBefore(SLV_SAVELOAD_LIST_LENGTH) ? last_num_specs : SlGetStructListLength(UINT8_MAX);
```
rubidium42: Do we really need to check the class here? Isn't it the case that road stop specs were added after `SLV_SAVELOAD_LIST_LENGTH`?

src/saveload/station_sl.cpp
```diff
@@ -514,12 +492,12 @@ static const SaveLoad _old_station_desc[] = {
 	/* Used by newstations for graphic variations */
 	SLE_CONDVAR(Station, random_bits,                SLE_UINT16,                 SLV_27, SL_MAX_VERSION),
 	SLE_CONDVAR(Station, waiting_triggers,           SLE_UINT8,                  SLV_27, SL_MAX_VERSION),
-	SLEG_CONDVAR("num_specs", SlStationSpecList::last_num_specs, SLE_UINT8,      SLV_27, SL_MAX_VERSION),
+	SLEG_CONDVAR("num_specs", SlStationSpecList<StationSpec>::last_num_specs, SLE_UINT8,      SLV_27, SL_MAX_VERSION),
```
rubidium42: Maybe remove the extra spaces if they aren't aligned any more?

src/saveload/station_sl.cpp
PeterN: No we don't, this is the "type check" I mentioned in limitations :-)

rubidium42 (approved):

PR #12213, opened 2024-03-03
Codechange: Where the ship comes from is already known

SamuXarick:
This simplifies the handling of variables.

## Motivation / Problem
#12192 could be further cleaned! I am 99.99% sure that calculation for `src_tile` isn't required. `ChooseShipTrack` is called upon entering `tile`, and looking further back to the caller, it can be deduced that `v->tile` matches `src_tile`. With that said, and combined with your changes, `enterdir` can also be removed, as it's not used anywhere else!

Further cleanings! `CreateRandomPath` and `GetRandomFollowUpTrackdir` is being fed `src_tile` as it's 2nd parameter. This could be eliminated, as `v` is also being passed to it. Just use `v->tile` in those functions.

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf.h
+++ b/src/pathfinder/yapf/yapf.h
@@ -21,11 +21,10 @@
  * Finds the best path for given ship using YAPF.
  * @param v        the ship that needs to find a path
  * @param tile     the tile to find the path from (should be next tile the ship is about to enter)
- * @param enterdir diagonal direction which the ship will enter this new tile from
  * @param path_found [out] Whether a path has been found (true) or has been guessed (false)
  * @return         the best trackdir for next turn or INVALID_TRACK if the path could not be found
  */
-Track YapfShipChooseTrack(const Ship *v, TileIndex tile, DiagDirection enterdir, bool &path_found, ShipPathCache &path_cache);
+Track YapfShipChooseTrack(const Ship *v, TileIndex tile, bool &path_found, ShipPathCache &path_cache);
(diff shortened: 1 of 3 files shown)


src/pathfinder/yapf/yapf.h
```diff
@@ -21,11 +21,10 @@
  * Finds the best path for given ship using YAPF.
  * @param v        the ship that needs to find a path
  * @param tile     the tile to find the path from (should be next tile the ship is about to enter)
- * @param enterdir diagonal direction which the ship will enter this new tile from
  * @param path_found [out] Whether a path has been found (true) or has been guessed (false)
  * @return         the best trackdir for next turn or INVALID_TRACK if the path could not be found
  */
-Track YapfShipChooseTrack(const Ship *v, TileIndex tile, DiagDirection enterdir, bool &path_found, ShipPathCache &path_cache);
+Track YapfShipChooseTrack(const Ship *v, TileIndex tile, bool &path_found, ShipPathCache &path_cache);
```
Kuhnovic, line 27: I think we should also change `const Ship *v` to `const Ship &v` everywhere. It doesn't really make sense to provide a pointer, as `v == nullptr` would never work.

PeterN: Separate change though, don't confuse Xarick more ;-)

Kuhnovic (dismissed):

Kuhnovic (approved):

PR #12217, opened 2024-03-03
Fix: YAPF doesn't check destination for start nodes

Kuhnovic:
## Motivation / Problem

YAPF doesn't properly check if start tiles are destination tiles. This can cause YAPF to return a longer path than necessary if the vehicle just happens to be at the destination already.

I managed to hit such a situation while working on some changes for the water region pathfinder.

I also wonder if the code removed in #12192 could have been a workaround for this problem, but I'm not sure about that. 

## Description

Adding starting node is done via AddStartupNode. Subsequent iterations are then done using AddNewNode. AddNewNode, in addition to a lot of other stuff, does the PfDetectDestination check. But this check is never done for any of the starting nodes.

The solution I chose is how it's typically done for most A* implementations: just do the PfDetectDestination for the best node as it's popped of the open queue.

## Limitations

Pathfinders being pathfinders, there could be some weird edge case that I didn't think about. 

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_base.hpp
+++ b/src/pathfinder/yapf/yapf_base.hpp
@@ -113,45 +113,39 @@ public:
 		m_veh = v;
 
 		Yapf().PfSetStartupNodes();
-		bool bDestFound = true;
 
 		for (;;) {
 			m_num_steps++;
-			Node *n = m_nodes.GetBestOpenNode();
-			if (n == nullptr) {
+			Node *best_open_node = m_nodes.GetBestOpenNode();
+			if (best_open_node == nullptr) break;
+
+			if (Yapf().PfDetectDestination(*best_open_node)) {
+				m_pBestDestNode = best_open_node;
 				break;
 			}
 
-			/* if the best open node was worse than the best path found, we can finish */
-			if (m_pBestDestNode != nullptr && m_pBestDestNode->GetCost() < n->GetCostEstimate()) {
-				break;
-			}
+			Yapf().PfFollowNode(*best_open_node);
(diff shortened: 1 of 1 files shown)


src/pathfinder/yapf/yapf_base.hpp
```diff
@@ -113,45 +113,39 @@ class CYapfBaseT {
...
 			m_num_steps++;
-			Node *n = m_nodes.GetBestOpenNode();
-			if (n == nullptr) {
-				break;
-			}
+			Node *best_open_node = m_nodes.GetBestOpenNode();
+			if (best_open_node == nullptr) break;
 
-			/* if the best open node was worse than the best path found, we can finish */
-			if (m_pBestDestNode != nullptr && m_pBestDestNode->GetCost() < n->GetCostEstimate()) {
```
rubidium42, line 126: Isn't this check somewhat important?
Lets imagine two routes, one shorter with a (long) low speed bridge ending at the destination and one slightly longer which also ends at the destination.
At some point it will go over the slow bridge and reach its destination while following the path. Now it won't consider the longer but faster route any more, as it has found a solution.
Or am I missing something important in this change?

Kuhnovic: The PF always chooses the best node on the open list. In case of the long slow bridge you'll get a node with a pretty high cost. This means that there will likely be other nodes on the list with a lower cost, and those will be processed first. Those might be destination nodes as well. The slow bridge node will only stop the search if it's truly the best node on the list.

The check removed here was pretty important, but my code essentially does the same, just in a slightly different way. The destination nodes don't need any special treatment, they will eventually pop out of the open list just like any other node. This is how most A* implementations do this, at least the ones I've seen.

rubidium42 (approved):

PR #12219, opened 2024-03-04
Change: Allow rail and road depot overbuilding in current orientation in order to connect to rail or road

Kuhnovic:
## Motivation / Problem
openttd_iTAPbr1RRu

Sometimes you happen to build a depot right when a train is passing by. As a result no connecting rails are created. It would be nice if you could simply overbuild the depot in the same orientation as soon as the train is gone. But that currently results in an "Already built" error.

This also applies to road vehicles and trams.

## Description
openttd_Vf7jkFWwzs

We no longer throw a error when the depot is already in the desired orientation. We don't charge any money for it either, we just silently ignore it. This behavior is similar to overbuilding station tiles.

Money is charged for any pieces of connecting rail / road that get constructed as a result of this user action.

## Limitations

Not sure if this is something that could mess with AI logic.

## Checklist for review

### DIFF
--- a/regression/regression/result.txt
+++ b/regression/regression/result.txt
@@ -7510,7 +7510,7 @@ ERROR: IsEnd() is invalid as Begin() is never called
     BuildRailDepot():              false
     BuildRailDepot():              true
     BuildRailDepot():              true
-    BuildRailDepot():              false
+    BuildRailDepot():              true
     GetRailDepotFrontTile():       33412
     IsBuildable():                 false
     DepotList
@@ -7604,12 +7604,12 @@ ERROR: IsEnd() is invalid as Begin() is never called
     BuildRoadDepot():              false
     BuildRoadDepot():              true
     BuildRoadDepot():              true
-    BuildRoadDepot():              false
+    BuildRoadDepot():              true
     HasRoadType(Road):             true
     HasRoadType(Tram):             false
-    GetLastError():                259
-    GetLastErrorString():          ERR_ALREADY_BUILT
-    GetErrorCategory():            1
+    GetLastError():                0
+    GetLastErrorString():          ERR_NONE
+    GetErrorCategory():            0
     IsRoadTile():                  false
     GetRoadDepotFrontTile():       33412
     IsRoadDepotTile():             true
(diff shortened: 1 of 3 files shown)


2TallTyler:
Regression failed. 😉

rubidium42:
I'm wondering whether it's needed/wanted to change the command, or just check the orientation in the UI and then call the appropriate function in the UI realm that (re)considers whether to add the road/track (`ConnectRoadToStructure` and some bits of `CcRailDepot`).

After all, the building of the depot by scripts does explicitly not build these bits of track/road.

I'm also wondering more generally... should we maybe just return an empty `CommandCost` instead of returning a `STR_ERROR_ALREADY_BUILT` error to the end user in most of the other function? That would make this diff also a lot smaller by the way.
Though, what are the repercussions of this for scripts? After all, it will cost them extra time as failed command will return immediately, whereas 'just' returning an empty `CommandCost` will require at least one tick.

Kuhnovic:
> I'm wondering whether it's needed/wanted to change the command, or just check the orientation in the UI and then call the appropriate function in the UI realm that (re)considers whether to add the road/track (`ConnectRoadToStructure` and some bits of `CcRailDepot`).

I gave this a try for the rail gui. It works, but we end up spreading the logic to multiple places. It also means we have to duplicate quite a bit of the (edge case) checking, such as whether the rail types are compatible. IMO this is not the way to go, as it will make it harder to keep the behavior consistent.

> I'm also wondering more generally... should we maybe just return an empty CommandCost instead of returning a STR_ERROR_ALREADY_BUILT error to the end user in most of the other function? That would make this diff also a lot smaller by the way.
Though, what are the repercussions of this for scripts? After all, it will cost them extra time as failed command will return immediately, whereas 'just' returning an empty CommandCost will require at least one tick.

I'm not familiar enough with the details of scripts / command logic to have an educated opinion on this. But I'm going to call it "beyond the scope of this PR" ;)

Kuhnovic:
Changed my approach. I went for simply returning an empty command cost, which really simplifies things. 

This might have an impact on AIs, but overbuilding to not-the-current-direction is already allowed and doesn't result in an error either. If anything this makes it more consistent.

rubidium42 (approved):

PeterN:
How might this affect AIs? @SamuXarick

PR #12227, opened 2024-03-07
Change: Use same audio buffer size and config parameter for all sound drivers.

PeterN:
## Motivation / Problem

For Allegro, SDL and SDLv2 sound drivers, the parameter to change audio frame size is `samples`, however for the two Windows sound drivers (win32 and xaudio2) the parameter is named `bufsize`.

The default value for all drivers that use `samples` is `1024`, but the default for drivers that use `bufsize` is `8192`.

At 44100 Hz, these sizes provide a latency per buffer (usually there are at least 2 buffers) of 23 ms and 186 ms respectively.

Our audio mixer is limited to only playing 8 sound effects at once, and that limit applies to the whole buffer size. Therefore with a larger buffer size, it is more likely that mixer channels will be used up and new sounds get blocked from playing.

## Description

Bring the two Windows driver in line by changing the parameter name to `samples` for all drivers, and change the default buffer size to `1024`.

## Limitations

This does not touch the Cocoa driver, with how the macOS API works we don't actually set the buffer size and just use what we're given.

## Checklist for review

### DIFF
--- a/src/sound/win32_s.cpp
+++ b/src/sound/win32_s.cpp
@@ -70,7 +70,7 @@ const char *SoundDriver_Win32::Start(const StringList &parm)
 	wfex.nAvgBytesPerSec = wfex.nSamplesPerSec * wfex.nBlockAlign;
 
 	/* Limit buffer size to prevent overflows. */
-	_bufsize = GetDriverParamInt(parm, "bufsize", 8192);
+	_bufsize = GetDriverParamInt(parm, "samples", 1024);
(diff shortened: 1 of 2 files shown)


TrueBrain (approved):

PeterN:
Forgot to mention. Due to how our audio mixer works, all sounds are started at the start of the audio buffer. Therefore, if the game tick (27ms) is shorter than the 'audio tick' (186ms), sounds started on different game ticks will be started at the same time in the audio mixer.

This is mitigated (but not fully resolved) by this change.

PR #12234, opened 2024-03-08
Codechange: Remove pointer parameters in foundation/slope code.

michicc:
## Motivation / Problem

Multiple foundation/slope functions use pointer parameters for either non-optional values that are modified or to return a second value. Some functions are even confused in their docs if the argument is optional or not (e.g. `GetFoundationPixelSlope`).

## Description

Change secondary return values to use `std::tuple`. While the intention might have been to only calculate this secondary value if the caller wants it, all functions changed here calculate the value unconditionally anyway.

Functions with non-optional modified parameters are changed to use references to clearly indicate that the parameter is required.

## Limitations

References are not directly visible at the call site and thus it is not always obvious that the argument might be modified. Whether this is considered better or worse than passing invalid pointers is debatable.

## Checklist for review

### DIFF
--- a/src/bridge_map.cpp
+++ b/src/bridge_map.cpp
@@ -69,8 +69,7 @@ TileIndex GetOtherBridgeEnd(TileIndex tile)
  */
 int GetBridgeHeight(TileIndex t)
 {
-	int h;
-	Slope tileh = GetTileSlope(t, &h);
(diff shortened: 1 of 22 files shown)


rubidium42 (approved):

michicc:
Sorry, but the first commit didn't actually compile on its own :)

PR #12237, opened 2024-03-08
Codechange: [CI] switch MacOS to the macos-14 runner

TrueBrain:
## Motivation / Problem

We have a few commits that cannot be merged because they fail on MacOS. Not because the code is wrong, but because the current MacOS version uses an older XCode, which don't have the proper fixes to make those PRs work.

## Description

Switch to `macos-14`. All issues that initially came from this, are resolved in other PRs.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -208,7 +208,7 @@ jobs:
 
     name: Mac OS (${{ matrix.arch }})
 
-    runs-on: macos-latest
+    runs-on: macos-14
     env:
       MACOSX_DEPLOYMENT_TARGET: 10.13
 
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -12,7 +12,7 @@ jobs:
   macos:
     name: MacOS
 
-    runs-on: macos-latest
+    runs-on: macos-14
     env:
       MACOSX_DEPLOYMENT_TARGET: 10.13
 


TrueBrain:
https://github.com/TrueBrain/OpenTTD/actions/runs/8208446751/job/22451933562 to show the release build is also working fine.

TrueBrain:
(And yes, merging this requires changing which MacOS build is mandatory, and requires all PRs pending to rebase).

rubidium42 (dismissed):
Do we want to force macos-14, or should we wait until macos-latest points to macos-14?

If we force it, we need to change it back to macos-latest at some point, or we'll run into trouble when macos-14 gets deprecated.
If we don't force it, builds will start failing at some random point in Q2 of 2024 and then we need to merge this PR at that unfortunate moment.

I personally think forcing macos-14 will have less impact on development as we can plan for it, instead of the migration happening at a random time.

Having said that, should we force macos-12 for the 14 branch, so that won't start failing when macos-latest becomes macos-14 and this PR has not been merged in there?

TrueBrain:
I have been thinking about this for a bit, and I have approached this the wrong way.

In general, we should always backport CI commits, as otherwise it might be that something works on master, and not on the release. And fixing those problems is very annoying. Also, that way it is more likely that releases will keep working.

This close for a new release, that is the only moment one can debate about this. But as we plan to do another RC, there is a much lower risk.

That said, I am going to do this PR a bit different, and split it in three. One which adds vcpkg to all targets, manually. So no matter what GitHub decides to do, our workflows work. The other to change macos to 14. And the last to change macos CI to arm64. And we should backport them all, if you ask me, before next RC :) Or otherwise wait with the latest -> 14 till just after the release, and backport it then.

TrueBrain:
https://github.com/TrueBrain/OpenTTD/actions/runs/8215061859/job/22468166938 to show it still works fine :)

rubidium42 (approved):

PR #12239, opened 2024-03-08
Fix #10370: increase maximum user-set engine name length to 256

### LINKED ISSUE #10370: Player-input vehicle name limit shorter than NewGRF name limit

TheMowgliMan:
## Motivation / Problem

#10370 exists. @merni-ns fixed it in his local fork but never merged it.

## Description

Increase `AX_LENGTH_ENGINE_NAME_CHARS` to 256. Surely enough?

## Limitations

1. I have a commit (which I reverted) from something else which I couldn't delete because I'm a git noob.
2. Don't have the correct commit messages for most of the commits because I didn't catch them and can't fix them for the reason in 1.

## Checklist for review

### DIFF
--- a/src/currency.h
+++ b/src/currency.h
@@ -72,7 +72,7 @@ enum Currencies {
 
 /** Specification of a currency. */
 struct CurrencySpec {
-	float rate;           ///< The conversion rate compared to the base currency.
+	uint16_t rate;           ///< The conversion rate compared to the base currency.
 	std::string separator; ///< The thousands separator for this currency.
 	TimerGameCalendar::Year to_euro; ///< Year of switching to the Euro. May also be #CF_NOEURO or #CF_ISEURO.
 	std::string prefix;    ///< Prefix to apply when formatting money in this currency.
--- a/src/currency.h
+++ b/src/currency.h
@@ -72,7 +72,7 @@ enum Currencies {
 
 /** Specification of a currency. */
 struct CurrencySpec {
-	uint16_t rate;           ///< The conversion rate compared to the base currency.
+	float rate;           ///< The conversion rate compared to the base currency.
 	std::string separator; ///< The thousands separator for this currency.
 	TimerGameCalendar::Year to_euro; ///< Year of switching to the Euro. May also be #CF_NOEURO or #CF_ISEURO.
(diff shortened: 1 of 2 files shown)


TrueBrain:
Our network protocol already supports packets up to 32k in size (done by Rubidium in 2021); I thought the command also already uses that, but otherwise that is fixable. So if you guys focus on the rest, we will pick up the networking stuff :)

TrueBrain:
Our network protocol already supports packets up to 32k in size (done by Rubidium in 2021); I thought the command also already uses that, but otherwise that is fixable. So if you guys focus on the rest, we will pick up the networking stuff :)

TheMowgliMan:
Alright, thanks! With that, I have a plan to implement this (based upon your suggestions):

1. Close this PR---there are some things I broke (Namely [78d6628](https://github.com/OpenTTD/OpenTTD/pull/12239/commits/78d6628ea8aa2490bf0e18a7a5ae5e8452f69e9c) and [abdf80f](https://github.com/OpenTTD/OpenTTD/pull/12239/commits/abdf80ffa0425b30a7c26df67aaa172076f1a909), as well as the commit messages) that are easier to fix with a fresh branch.
2. Increase the limit to 100 characters (At least until multiplayer command packet size is increased), or count bytes instead? It's more reliable but not as user-friendly.
3. Do as @merni-ns suggested in #10384 and truncate after the 32nd character and put the whole name in the description in some manner, perhaps as a heading or a bit at the end going: "Full name: [...]" or something similar.

I'm not experienced with the codebase, but I should be able to figure this out. If I'm really clever, I'll do the same for NewGRF names.

PR #12242, opened 2024-03-09
Codechange: [CI] switch to arm64 for MacOS builds

TrueBrain:
## Motivation / Problem

This creates more variety in our CI builds, increasing the odds of us finding issues.

Also, arm64 is the main target for MacOS. So just better to use that in our CI.

## Description

Switch MacOS CI to arm64.

This changes nothing for releases.
This does mean that after merging this PR, all other open PRs need to rebase, as one of the checks will not work.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -203,8 +203,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - arch: x64
-          full_arch: x86_64
+        - arch: arm64
+          full_arch: arm64
 
     name: Mac OS (${{ matrix.arch }})
 


glx22:
Hmm we could even run both arm64 and x64 (they are super fast to build now)

glx22 (approved):

TrueBrain:
> Hmm we could even run both arm64 and x64 (they are super fast to build now)

I gave this some thought, and I think we should minimize the CPU-minutes used when possible. And as we so far never had a case where x64 succeeded, but arm64 failed, I think it is safe to say the reverse is also true. So from that logic, it currently seems unlikely that running both gives more information.

I suggest that if we ever have a situation where this no longer holds true, we add x64 then. (this to just write down the current reasoning, for the history books :) )

PR #12247, opened 2024-03-09
Fix #11055: Make saveload failure error messages consistent with others

### LINKED ISSUE #11055: [Bug]: "Game Load Failed" error message is shorter than other two-line messages
### Version of OpenTTD

Since c600158c1df336153eb6705eded718f0434fee75, tested on c316d9ecb7

### Expected result

When saving or loading the game to/from a file fails, the resulting error message should have some space between "Game Load Failed" (or "Game Save Failed") and the reason for the save/load failure.

Example: 
image

### Actual result

When saving or loading the game to/from a file fails, the resulting error message has no additional space between "Game Load Failed" (or "Game Save Failed") and the reason for the save/load failure, unlike other message where there is space.

Example:
image

### Steps to reproduce

1. Try to save (or load) the game, such that it fails.
2. Notice the resulting message has no extra space between "Game Save Failed" (or "Game Load Failed") and the reason for failure.
3. Perform any other action (in game) such that it fails.
4. Notice the resulting message has space between the two lines.

merni-ns:
## Motivation / Problem
Fixes #11055.

The "save/load failed" error messages were displayed by substituting the reason (`_sl.error_str`) with a parameter into `STR_ERROR_GAME_SAVE_FAILED`/`STR_ERROR_GAME_LOAD_FAILED`. This means the line separation is not the same as those formed by using the two parameters of `ShowErrorMessage`.

## Description
- Use the first parameter of `ShowErrorMessage` to show the initial string "Game save failed"/"Game load failed"
- Use the second parameter to show the reason that was originally the first parameter of `STR_ERROR_GAME_(SAVE|LOAD)_FAILED`.
- The `_sl.extra_msg`, an untranslated string returned by internal APIs, is untouched and used as the first parameter of reasons that require it (and thus the first parameter overall).
- Make the "Game save failed"/"Game load failed" more consistent by not Capitalising Each Word and adding an ellipsis at the end.

image

## Limitations

- Not tested, except for the case of a save file that the game has no permission to read. In particular the case of "internal error" or "broken savegame", which require one more parameter, are not tested.

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -2760,22 +2760,23 @@ void SetSaveLoadError(StringID str)
 	_sl.error_str = str;
 }
 
-/** Get the string representation of the error message */
-const char *GetSaveLoadErrorString()
+/** Return the appropriate initial string for an error depending on whether we are saving or loading. */
+StringID GetSaveLoadErrorType()
 {
-	SetDParam(0, _sl.error_str);
-	SetDParamStr(1, _sl.extra_msg);
+	return _sl.action == SLA_SAVE ? STR_ERROR_GAME_SAVE_FAILED : STR_ERROR_GAME_LOAD_FAILED;
+}
 
-	static std::string err_str;
-	err_str = GetString(_sl.action == SLA_SAVE ? STR_ERROR_GAME_SAVE_FAILED : STR_ERROR_GAME_LOAD_FAILED);
-	return err_str.c_str();
+/** Return the description of the error. **/
+StringID GetSaveLoadErrorMessage()
+{
+	SetDParamStr(0, _sl.extra_msg);
+	return _sl.error_str;
 }
 
 /** Show a gui message when saving has failed */
 static void SaveFileError()
 {
-	SetDParamStr(0, GetSaveLoadErrorString());
-	ShowErrorMessage(STR_JUST_RAW_STRING, INVALID_STRING_ID, WL_ERROR);
+	ShowErrorMessage(GetSaveLoadErrorType(), GetSaveLoadErrorMessage(), WL_ERROR);
 	SaveFileDone();
 }
 
@@ -2810,7 +2811,7 @@ static SaveOrLoadResult SaveFileToDisk(bool threaded)
 		 * cancelled due to a client disconnecting. */
 		if (_sl.error_str != STR_NETWORK_ERROR_LOSTCONNECTION) {
 			/* Skip the "colour" character */
-			Debug(sl, 0, "{}", GetSaveLoadErrorString() + 3);
+			Debug(sl, 0, "{}", GetString(GetSaveLoadErrorType()).substr(3) + GetString(GetSaveLoadErrorMessage()));
 			asfp = SaveFileError;
 		}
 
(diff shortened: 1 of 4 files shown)


src/saveload/saveload.cpp
```diff
@@ -2754,22 +2754,21 @@ void SetSaveLoadError(StringID str)
 	_sl.error_str = str;
 }
 
-/** Get the string representation of the error message */
-const char *GetSaveLoadErrorString()
-{
-	SetDParam(0, _sl.error_str);
-	SetDParamStr(1, _sl.extra_msg);
+/** Return the appropriate initial string for an error depending on whether we are saving or loading. */
+StringID GetSaveLoadErrorString() {
```
2TallTyler: Given that this and `GetSaveLoadErrorMessage()` are both dealing with StringIDs, I wonder if this might be a better name.

```suggestion
StringID GetSaveLoadErrorType() {
```

merni-ns: Sure.

src/saveload/saveload.cpp
```diff
@@ -2754,22 +2754,21 @@ void SetSaveLoadError(StringID str)
 	_sl.error_str = str;
 }
 
-/** Get the string representation of the error message */
-const char *GetSaveLoadErrorString()
-{
-	SetDParam(0, _sl.error_str);
-	SetDParamStr(1, _sl.extra_msg);
+/** Return the appropriate initial string for an error depending on whether we are saving or loading. */
+StringID GetSaveLoadErrorType() {
```
rubidium42:
```suggestion
StringID GetSaveLoadErrorType()
{
```

src/saveload/saveload.cpp
```diff
@@ -2754,22 +2754,21 @@ void SetSaveLoadError(StringID str)
...
+	return _sl.action == SLA_SAVE ? STR_ERROR_GAME_SAVE_FAILED : STR_ERROR_GAME_LOAD_FAILED;
+}
 
-	static std::string err_str;
-	err_str = GetString(_sl.action == SLA_SAVE ? STR_ERROR_GAME_SAVE_FAILED : STR_ERROR_GAME_LOAD_FAILED);
-	return err_str.c_str();
+/** Return the description of the error. **/
+StringID GetSaveLoadErrorMessage() {
```
rubidium42:
```suggestion
StringID GetSaveLoadErrorMessage()
{
```

src/saveload/saveload.cpp
```diff
@@ -2803,8 +2802,7 @@ static SaveOrLoadResult SaveFileToDisk(bool threaded)
 		/* We don't want to shout when saving is just
 		 * cancelled due to a client disconnecting. */
 		if (_sl.error_str != STR_NETWORK_ERROR_LOSTCONNECTION) {
-			/* Skip the "colour" character */
-			Debug(sl, 0, "{}", GetSaveLoadErrorString() + 3);
+			Debug(sl, 0, "{}", GetString(GetSaveLoadErrorType()));
```
rubidium42: Why is the colour not stripped any more? Also in the next bit. For example using `.substr(3)`.

merni-ns: Hopefully I have fixed it in the current force-push, but I have no idea about how colour codes work.

rubidium42 (approved):

PR #12249, opened 2024-03-09
Add: [Script] Unbunch order flag

SamuXarick:
## Motivation / Problem
Missing unbunch feature for AIs.

## Description
Allow AIs to unbunch vehicles.
Proper handle the new `OF_UNBUNCH_IN_DEPOT` order flag when combined with other existing order flags. Add regression tests to test this flag and the returned error strings associated with it.

## Limitations
Only for API version 15?

## Checklist for review


rubidium42:
Are there many script developers that are interested in checking for five extra error conditions in their code?

That there are many strings is just to provide the user some more user friendly explanation what's wrong. With scripts you should be more concise as returning too many errors is only making handling them correctly more cumbersome.

glx22:
Would probably be better to keep error as it was, but maybe add way for scripts to query extra message if they need to.

SamuXarick:
> Are there many script developers that are interested in checking for five extra error conditions in their code?
> 
> That there are many strings is just to provide the user some more user friendly explanation what's wrong. With scripts you should be more concise as returning too many errors is only making handling them correctly more cumbersome.

I can group the unbunch errors into 1 single message. But then it's going to take a bit more effort to identify the exact problem to solve it. Possible causes for the error would have to be properly documented.

SamuXarick:
> Would probably be better to keep error as it was, but maybe add way for scripts to query extra message if they need to.

Those errors in the form of two strings can't be reached from the GUI as far as I see. They're being handled in `order_gui.cpp`, `CmdInsertOrder` isn't called. Also, there's already inconsistency about the way the error is returned for a conditional order or full load order. Those are handled with `return_cmd_error`.

glx22:
But `ScriptInstance::DoCommandCallback` could access the extra message.

src/script/api/script_order.cpp
```diff
@@ -515,18 +519,18 @@ static ScriptOrder::OrderPosition RealOrderPositionToScriptOrderPosition(Vehicle
 			order.SetLoadType((OrderLoadFlags)GB(order_flags, 5, 3));
 			order.SetUnloadType((OrderUnloadFlags)GB(order_flags, 2, 3));
 			order.SetStopLocation(OSL_PLATFORM_FAR_END);
+			if (v->IsGroundVehicle()) order.SetNonStopType((OrderNonStopFlags)GB(order_flags, 0, 2));
 			break;
 
 		case OT_GOTO_WAYPOINT:
 			order.MakeGoToWaypoint(::GetStationIndex(destination));
+			if (v->type == VEH_TRAIN) order.SetNonStopType((OrderNonStopFlags)GB(order_flags, 0, 2));
```
SamuXarick:
```suggestion
			if (v->IsGroundVehicle()) order.SetNonStopType((OrderNonStopFlags)GB(order_flags, 0, 2));
```
This probably should be changed in the future when road waypoints are properly supported (#13398, #13419 and maybe some others...)

src/order_cmd.cpp
```diff
@@ -799,10 +799,11 @@ CommandCost CmdInsertOrder(DoCommandFlags flags, VehicleID veh, VehicleOrderID s
...
 			/* Check if we're allowed to have a new unbunching order. */
-			if ((new_order.GetDepotActionType() & ODATFB_UNBUNCH)) {
-				if (v->HasFullLoadOrder()) return CommandCost(STR_ERROR_CAN_T_ADD_ORDER, STR_ERROR_UNBUNCHING_NO_UNBUNCHING_FULL_LOAD);
-				if (v->HasUnbunchingOrder()) return CommandCost(STR_ERROR_CAN_T_ADD_ORDER, STR_ERROR_UNBUNCHING_ONLY_ONE_ALLOWED);
-				if (v->HasConditionalOrder()) return CommandCost(STR_ERROR_CAN_T_ADD_ORDER, STR_ERROR_UNBUNCHING_NO_UNBUNCHING_CONDITIONAL);
+			if ((new_order.GetDepotOrderType() & ODTFB_PART_OF_ORDERS) && (new_order.GetDepotActionType() & ODATFB_UNBUNCH)) {
+				if (new_order.GetDepotActionType() & ODATFB_HALT) return CMD_ERROR;
+				if (v->HasFullLoadOrder()) return CommandCost(STR_ERROR_UNBUNCHING_NO_UNBUNCHING_FULL_LOAD);
+				if (v->HasUnbunchingOrder()) return CommandCost(STR_ERROR_UNBUNCHING_ONLY_ONE_ALLOWED);
+				if (v->HasConditionalOrder()) return CommandCost(STR_ERROR_UNBUNCHING_NO_UNBUNCHING_CONDITIONAL);
```
LordAro, line 751: i'm not clear what this change is doing. adding a script function shouldn't be making any changes to the "actual" code

SamuXarick: These errors are only reachable via AIs, human players won't see it.
AIs however don't have the extra error message implemented in them, and the error translation from the original code to AI would only be taking the first message into account, `STR_ERROR_CAN_T_ADD_ORDER`, which isn't clear enough it's related to unbunch.

LordAro: I don't see any evidence this is AI only code, and you're removing the first part of the error message that would be displayed, no?

SamuXarick: Oh, and the other one is an invalid state which only AIs have the ability to set, but shouldn't. Setting both unbunch + stop at depot is supposed to be not possible.

SamuXarick: With 2 error messages:
```
GetLastErrorString(): ERR_UNKNOWN
GetErrorCategory(): 0 // ERR_CAT_NONE
```

with 1 error message:
```
GetLastErrorString(): ERR_ORDER_UNBUNCH
GetErrorCategory(): 10 // ERR_CAT_ORDER
```

If the 2 error messages are to be kept, I can map `STR_ERROR_CAN_T_ADD_ORDER` to script `ERR_CAN_T_ADD_ORDER`, and the result would be:
```
GetLastErrorString(): ERR_CAN_T_ADD_ORDER
GetErrorCategory(): 10 // ERR_CAT_ORDER
```
But this wouldn't exactly tell it's related to unbunch. It could be related to something else, like... incorrect station type for vehicle.

SamuXarick:
I decided to split the changes into multiple commits on this last rebase because I realized the PR does a bit more than just adding the unbunch feature to scripts:

~~X - Codechange: Return proper unbunch error strings for AIs:
I have a bit of an issue/dilema about the way the error message is generated for the AIs. Ideally, the AI could read the two tiered error messages, but this is not yet possible. A PR that attempts to expose these messages is also proposed here: https://github.com/OpenTTD/OpenTTD/pull/12269.~~ **EDIT:** this step has been resolved, #15107 reverted the error back to a single message, making #12269 no longer required.
1 - Codechange: [Script] Reorder variable declarations for clarity:
Just fancy reordering of declarations for better readability. Nothing special.
2 - Change: [Script] Silently ignore OrderNonStopFlags in ScriptOrder::InsertOrder for ships and aircraft:
The problem being solved here is to avoid passing invalid non-stop flag states to DoCommand, which could ultimately reject the command or even accept them and store it. I went with the solution to simply discard the invalid flags before calling DoCommand.
3 - Codechange: [Script] Set OrderDepotActionFlag::NearestDepot later to avoid repeated check:
Minor optimization tweak. Nothing special.
4 - Add: [Script] Unbunch order flag:
The real implementation of the unbunch flag for script. It has perhaps too many error messages related to unbunch rules. 
5 - Change: [Script] Use only one unbunch error string instead of five:
This reduces that number to a single generalized unbunch error message.

src/order_cmd.cpp
Rito13: These strings are used in 3 places:
1. https://github.com/OpenTTD/OpenTTD/blob/2d8cb5314c34af6c8f5be57868b3c8b6a2192110/src/order_gui.cpp#L382
Where user has to have controll pressed.
2. https://github.com/OpenTTD/OpenTTD/blob/2d8cb5314c34af6c8f5be57868b3c8b6a2192110/src/order_cmd.cpp#L1278
Which is shown when user modifies the orders.
3. The one you have changed, it seems like when new order will be added without use of ctrl you change will be visible to the player.

SamuXarick: In order_gui.cpp, `GetOrderCmdFromTile` has its own checks for adding an unbunch order, which protect it from the invalid condition checks in order_cmd.cpp's `CmdInsertOrder`.

If Ctrl isn't pressed, it’s not considered an unbunch order, and `new_order.GetDepotActionType().Test(OrderDepotActionFlag::Unbunch)` in `CmdInsertOrder` prevents it from ever reaching the error message checks.

From a player's perspective, `GetOrderCmdFromTile` makes it unlikely to reach `CmdInsertOrder` in a way that would trigger its errors to be shown. Instead, those errors would originate from `order_gui.cpp` in the `OnPlaceObject` function:
https://github.com/OpenTTD/OpenTTD/blob/2d8cb5314c34af6c8f5be57868b3c8b6a2192110/src/order_gui.cpp#L1471
The error it produces is only 2 lines, not 3.
image

Hypothetically, if errors in the order_cmd `CmdInsertOrder` were to occur, they could look something like this:
image

Rito13: You are right in terms of singleplayer, the only way to insert new unbunch order is to select the depot with go to tool while pressing ctrl, then when the order can't be inserted the error message will occur from `GetOrderCmdFromTile` called by `OnPlaceObject` as you have said, but on multiplayer there is slight delay between the server and client. In very specific situation, when two players change orders of one vehicle at the same time, it can end up that when client adds an order everything will be fine, and error in `GetOrderCmdFromTile` will not occur, but cmd from another user could get to the server first, then on the server `CmdInsertOrder` will be called directly and will fail as the orders have changed after the last check on the client side, then server will send error message back to the client and it will be show.
I was able to reproduce that with my PC and mobile (on mobile I have only OTTD 14.1, so thats why I have used that version), the server was hasted from the phone, and screenshot was taken on the PC.

Zrzut ekranu z 2026-01-13 14-03-33

SamuXarick: Oops, so it's possible to have the 3 lines message to show.

I'd still prefer to reduce it to 2 lines, to match the way the error comes from the gui side. The middle line infers incorrectly that it's a station order when we're adding a depot order. It would be more consistent and at the same time it allow the AI side of things to get the single error message, avoiding #12269 altogether. It's a simpler fix/implementation.

Rito13: I agree, the `STR_ERROR_CAN_T_ADD_ORDER` is just a wrong string to be used here, it should not be used with depots.
> STR_ERROR_CAN_T_ADD_ORDER                                       :{WHITE}... vehicle can't go to that station

PR #12253, opened 2024-03-10
Codechange: [CI] build some targets every night, instead of every PR

TrueBrain:
## Motivation / Problem

Currently we run a lot of CI jobs on every PR. This didn't used to be a problem, but with more people now active on OpenTTD, there often is a (long) queue of CI jobs. We recently had an "incident" where an author rebased 15 open PRs, causing a long queue on the CI. Although those rebases were mostly unneeded, it is not behaviour we can control.

This PR tries to address these queue lengths; not the individual CI time.

It is basic math: GitHub offers us 20 runners for free (and we are very grateful for that). This means that every hour we have 1200 CPU-hours available to us.

Before this PR, every PR consumes: 5x 10m (Linux) + 4m (MacOS) + 2x 25m (MinGW) + 4x 10m (Windows) + 5m (Emscripten) + 40m (codeQL), a total of ~200 CPU minutes per PR. Means it can process ~6 PRs per hour.

Most of these jobs that run actually don't give us much new information, or very very very very very rarely point out a problem. MinGW is the best example here. We have it running as a single developer of us still uses it, and as such we want to make sure he can continue developing for OpenTTD. But it consumes a wopping 50 CPU-minutes, 1/3rd of the total. And how often does it fail while other jobs fail? Well ... I can't remember any of my PRs ever failing on it, not in the last months at least. So the actual added value here is very low.

~~Similar, we still compile for MSVC on Windows 2019. This has an older MSVC installed, and as we can still compile OpenTTD with it, we don't want to let go. But it hogs up another 20 CPU minutes every PR. While I have haven't seen 2019 fail where the Windows 2022 variant didn't fail, in the many PRs I saw/made in the last few months. So here too, very little added value, while consuming a lot of CPU-minutes.~~ Edit: we removed Windows 2019 completely.

Lastly, we see the same with SDL2 vs SDL1.2. I think SDL1.2 never failed, when SDL2 didn't. But still we spend 10 CPU-minutes on it, "just to be sure". As that is how we roll ;)

All other targets have a more well-defined role:
- Linux: Debug vs Release, Clang vs GCC, no PCH
- MacOS: as .. it is MacOS
- Windows: x64 vs x86 .. one can debate if x86 is still relevant, but for now, this is fine.
- CodeQL: it shows from time to time actual issues with PRs

## Description

Split the "build on PR" into two groups: "build on PR" and "build every night". And push the above mentioned jobs to the latter. That way we fee up 80 CPU-minutes per PR, 2/5th of the time a PR needs in terms of CPU-minutes. Means we can now process 10 PRs per hour: almost double the amount.

And that for a very low loss in value, as these targets, as mentioned above, already didn't actually add value.

PS: this also adds back x86_64 to MacOS, in the nightly, to ensure we don't run into any weird issues during release.

Example of a nightly run: https://github.com/TrueBrain/OpenTTD/actions/runs/8221285336

## Limitations

- Failures to this nightly job will be invisible; but I will add later today some code to DorpsGek to notify us when these kind of jobs fail, so we can react to it.
- This change will make us reactive for some platforms. So it is possible that someone merges a PR that makes one of these targets go BOOM, and we notice it only the next night. Pretty much like the old subversion days. But just much less likely to happen :)
- Every PR after merging this needs to be rebased, before the CI will approve it. Because of a deduplication action, the path to the CI result has changed.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI - Build
 
 on:
   pull_request:
@@ -38,10 +38,6 @@ jobs:
           compiler: gcc
           cxxcompiler: g++
           libraries: libsdl2-dev
-        - name: GCC - SDL1.2
-          compiler: gcc
-          cxxcompiler: g++
-          libraries: libsdl1.2-dev
         - name: GCC - Dedicated
           compiler: gcc
           cxxcompiler: g++
@@ -83,38 +79,17 @@ jobs:
       matrix:
         include:
         - os: windows-latest
-          name: Windows
           arch: x86
         - os: windows-latest
-          name: Windows
           arch: x64
 
-    name: ${{ matrix.name }} (${{ matrix.arch }})
+    name: Windows (${{ matrix.arch }})
 
     uses: ./.github/workflows/ci-windows.yml
     secrets: inherit
 
     with:
       os: ${{ matrix.os }}
-      arch: ${{ matrix.arch }}
-
-  mingw:
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - msystem: MINGW64
-            arch: x86_64
-          - msystem: MINGW32
-            arch: i686
-
-    name: MinGW (${{ matrix.arch }})
-
-    uses: ./.github/workflows/ci-mingw.yml
-    secrets: inherit
-
-    with:
-      msystem: ${{ matrix.msystem }}
       arch: ${{ matrix.arch }}
 
   check_annotations:
@@ -124,7 +99,6 @@ jobs:
     - linux
     - macos
     - windows
-    - mingw
 
     if: always() && github.event_name == 'pull_request'
 
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -17,56 +17,8 @@ jobs:
   emscripten:
     name: Emscripten
 
-    runs-on: ubuntu-20.04
-    container:
-      # If you change this version, change the number in the cache step too.
-      image: emscripten/emsdk:3.1.42
-
-    steps:
-    - name: Checkout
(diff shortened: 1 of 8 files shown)


TrueBrain:
> Similar, we still compile for MSVC on Windows 2019. This has an older MSVC installed, and as we can still compile OpenTTD with it, we don't want to let go. But it hogs up another 20 CPU minutes every PR. While I have haven't seen 2019 fail where the Windows 2022 variant didn't fail, in the many PRs I saw/made in the last few months. So here too, very little added value, while consuming a lot of CPU-minutes.

And just when those words were spoken, we have a PR that only failed on Windows 2019. Grrr. Mostly I am in doubt if we still want to support that MSVC at all, or whether we should just remove it. Otherwise, maybe only run the x64, and not also x86. Or just take the exception, and just take it as example that this would be fixed the next day, and not a the moment of PR.

michicc:
Seeing that we are (almost) cutting edge on C++ now, I think we can just let MSVC 2019 kick the bucket.

.github/workflows/ci-build.yml
```diff
@@ -208,223 +66,33 @@ jobs:
...
+          name: Windows
+          arch: x64
 
-        echo "::group::CMake"
-        cmake .. \
-          -GNinja \
-          -DCMAKE_CXX_FLAGS="-fuse-ld=lld" \
-          # EOF
-        echo "::endgroup::"
+    name: ${{ matrix.name }} (${{ matrix.arch }})
```
glx22: Why not doing it like other targets?
```suggestion
    name: Windows (${{ matrix.arch }})
```

Of course also needs removal of `name` in the matrix.

TrueBrain: Was when 2019 was still relevant :P can be removed now

glx22 (approved):

PR #12259, opened 2024-03-10
Codechange: simplify TileAdd and replace macro's with functions

rubidium42:
## Motivation / Problem

`#define` magic over `C++` functionality.

The expression string describing expression passed into `TILE_ADD` is the same for all cases where `TILE_ADD` is called twice on the same line of code, and as such it doesn't distinguish between the two. Not a real problem, as if it fails it should also fail on the first call, but still inconsistent.
In all other cases the expression does not add a lot to the assertion message.

## Description

Remove the expression string in lieu of the column number from `std::source_location`.
Use `std::source_location` in the function with default value, to pass the file and line to `TILE_ADD` instead of going via a preprocessor macro.
Always passing the `std::source_location` from `TILE_ADDXY` to `TILE_ADD`; the compiler ought to optimise this away with even minimal optimisations, which are likely when debugging is disabled.

## Limitations

Deliberately adding [[maybe_unused]] to a variable that will never be used in that context, just to have the same signature and simplifying the other accessor of the functionality.

The `TileAdd` function got renamed to the name of the old macro `TILE_ADD`, just to prevent massive changes all over the code. This can be trivially changed in a separate PR. Something similar holds for `TILE_ADDXY`, which was a macro and has become a function.

## Checklist for review

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -64,7 +64,7 @@
 
 
 #ifdef _DEBUG
-TileIndex TILE_ADD(TileIndex tile, TileIndexDiff offset, std::source_location location)
+TileIndex TILE_ADD(TileIndex tile, TileIndexDiff offset)
 {
 	int dx = offset & Map::MaxX();
 	if (dx >= (int)Map::SizeX() / 2) dx -= Map::SizeX();
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -69,32 +69,26 @@ extern "C" _CRTIMP void __cdecl _assert(void *, void *, unsigned);
 
 
 #ifdef _DEBUG
-TileIndex TileAdd(TileIndex tile, TileIndexDiff add,
-	const char *exp, const char *file, int line)
+TileIndex TILE_ADD(TileIndex tile, TileIndexDiff offset, std::source_location location)
 {
-	int dx;
-	int dy;
-	uint x;
-	uint y;
-
-	dx = add & Map::MaxX();
+	int dx = offset & Map::MaxX();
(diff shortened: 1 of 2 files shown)


src/map.cpp
```diff
@@ -69,32 +64,18 @@ extern "C" _CRTIMP void __cdecl _assert(void *, void *, unsigned);
...
-	const char *exp, const char *file, int line)
+TileIndex TILE_ADD(TileIndex tile, TileIndexDiff offset, std::source_location location)
 {
-	int dx;
-	int dy;
-	uint x;
-	uint y;
-
-	dx = add & Map::MaxX();
+	int dx = offset & Map::MaxX();
```
TrueBrain, line 69: While at it, can't we use `TileX(offset)` and `TileY(offset)` here?

I was mostly triggered by `dy`, which looks a bit more complicated than it has to. `TileY` does it less complicated :D

(not your change ofc, but as we are here anyway)

rubidium42: Though `offset` isn't of type `TileIndex`, so technically `TileX` and friends are not the right function to call. Also `TileX` returns an unsigned int, whereas this needs a signed int.

TrueBrain: Owh, right, that weird `TileIndexDiff` that is only used in, what, 50 instances? I forgot about that :P We also have `TileIndexDiffC`, solving the same problem, but differently. I forgot about those weird types. Nothing for this PR.

rubidium42:
#12260 is the better solution to simplify `TileAdd`.

PR #12261, opened 2024-03-10
Fix: Off-by-one in EnsureVisibleCaption.

PeterN:
## Motivation / Problem

An off-by-one in EnsureVisibleCaption causes the minimum visible caption height to be 13 scaled pixels and 1 unscaled pixel. At 1x interface scale, this 'happens' to be the complete height of a caption, but at other interface scales it is not.

image

## Description

Instead of using a scaled fixed value, correct the off-by-one and just use the window's actual caption height instead.

image

## Checklist for review

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -1927,9 +1927,6 @@ static void HandleMouseOver()
 	}
 }
 
-/** The minimum number of pixels of the title bar must be visible in both the X or Y direction */
-static const int MIN_VISIBLE_TITLE_BAR = 13;
-
 /** Direction for moving the window. */
 enum PreventHideDirection {
 	PHD_UP,   ///< Above v is a safe position.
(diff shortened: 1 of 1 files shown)


src/window.cpp
```diff
@@ -1941,7 +1938,7 @@ enum PreventHideDirection {
  * If needed, move the window base coordinates to keep it visible.
  * @param nx   Base horizontal coordinate of the rectangle.
  * @param ny   Base vertical coordinate of the rectangle.
- * @param rect Rectangle that must stay visible for #MIN_VISIBLE_TITLE_BAR pixels (horizontally, vertically, or both)
+ * @param rect Rectangle that must stay visible (horizontall, vertically, or both)
```
LordAro: quick, comment on the missing 'y' before TB does

LordAro (approved):

PR #12263, opened 2024-03-10
Codechange: Use std::make_unique instead of passing new into a unique_ptr.

PeterN:
## Motivation / Problem

Mixing `std::unique_ptr` with `new`.

## Description

Replace assignment from `new` with `std::make_unique` instead.

## Checklist for review


rubidium42 (dismissed):

LordAro:
There was a reason (that I can't remember) why we couldn't use this before. Have those reasons gone with C++20?

PeterN:
Long ago if they were. `std::make_shared<byte[]>` is the one that is/was missing.

Some of these could perhaps just be using `std::vector` instead, but that's a bit more of a rabbit hole.

michicc:
It looks like the buffer is always overwritten or memset'ed anyway to me, so one could use `std::make_unique_for_overwrite` for a small performance speed-up.

PeterN:
Given that `std::make_unique` will actually initialise the allocated array, it seems better to just remove the `MemSetT`s after.

I've removed the change to grf.cpp as that one is better done with `std::vector`.

michicc (approved):

PR #12273, opened 2024-03-11
Change: [Script] Return different time values based on TimeMode

SamuXarick:
## Motivation / Problem
I'm not sure this is the right approach, but the problem I'm trying to address here is due to the Wallclock feature.
Some AIs use functions which use the Economy timer, mixed with functions that use Calendar timer. This results in broken behaviour. As an example, my AI tries to get vehicle current age to determine whether a running route is deemed profitable, but it bases that on values given by AIDate.GetCurrentDate and AIVehicle.GetAge. The values can't be used together like this. It breaks the logic in decision making. I also use some other logic to decide when to replace vehicles with old age. This one is tricky now, because AIVehicle.GetAgeLeft, albeit a small number, is now being used in the context of a dilated amount of time. It's going to depend on the use case.

## Description
Changes `ScriptEngine::GetMaxAge`, `ScriptVehicle::GetAge`, `ScriptVehicle::GetWagonAge`, `ScriptVehicle::GetMaxAge` and `ScriptVehicle::GetAgeLeft` return values depending on `TimeMode`.

## Limitations
This PR is trying to do something so that AIs can chose which value they expect to get based on the timer mode in use, though I don't really have much faith for old unmaintained AIs. There was only one timer before, and it's not possible to assume they always meant the returned values were of Economy timer or Calendar timer.

## Checklist for review

### DIFF
--- a/src/script/api/script_engine.cpp
+++ b/src/script/api/script_engine.cpp
@@ -10,6 +10,7 @@
 #include "../../stdafx.h"
 #include "script_engine.hpp"
 #include "script_cargo.hpp"
+#include "script_timemode.hpp"
 #include "../../company_base.h"
 #include "../../strings_func.h"
 #include "../../rail.h"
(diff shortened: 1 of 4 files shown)


rubidium42:
This is unneeded since #12362, right?

SamuXarick:
> This is unneeded since #12362, right?

I guess so. May be closed.

PR #12274, opened 2024-03-11
Codefix f1e999ec: use RandomRange(l) instead of _random.Next(l)

TrueBrain:
## Motivation / Problem

During review of #12252 I was wondering if `.Next(limit)` was actually used (answer: yes, by AI-specific randomizers). But there was a single instance where `_random.Next(limit)` was used instead of `RandomRange(limit)`. What is their difference? The latter can be debugged via `RANDOM_DEBUG`. But having a single randomizer bypass the `RANDOM_DEBUG` sounds like an interesting "WTF WTF WTF" sessions during desync analysis.

## Description

Don't bypass `RANDOM_DEBUG`, and use `RandomRange` instead.

## Checklist for review


rubidium42 (dismissed):

rubidium42:
In the same realm there's also `RandomCompanyManagerFaceBits` and `GenerateTownName` which take a `Randomizer` which is `_random` in some cases.
Those do, however, require more work than this simple change. Also, those cases are probably less likely to occur in games, making them less problematic.

michicc (approved):

PR #12275, opened 2024-03-11
Codechange: Make the road vehicle path cache behave more like a std::deque

Kuhnovic:
## Motivation / Problem

The Road Vehicle Path Cache is actually two std::deque's in one. That's a bit silly, but it is done to make saveload work. But the silliness has propaged into other parts of the code, where you basically have to push and pop items to both queues,

## Description

I tried to keep the two-queues-in-one an implementation detail. From the caller's perspective it works just like a single std::deque. Sort of, because I only implemented the functions that were needed ;)

## Checklist for review


PeterN:
It would actually be better to replace the two deque system. That was an oversight on my part that needlessly uses quite a lot of memory for each RoadVehicle object.

Kuhnovic:
#12345 Provides a better solution, closing this one

PR #12278, opened 2024-03-11
Codechange: Reduce size of class WaterRegion

JGRennison:
## Motivation / Problem

`class WaterRegion` is quite large, mainly due to the 256 byte tile patch array.
The number of instances of `class WaterRegion` is proportional to the map size and so is a measurable memory overhead.

In the majority of water regions this patch array is not needed due to one of the following reasons (in approximately descending order of frequency):
* The ship pathfinder never looks at the region, so the tile patch array never needs to be populated.
* The region is entirely land (0 patch labels) so storage of 256 of the same value is redundant.
* The region is entirely water (1 patch label) so storage of 256 of the same value is redundant.

## Description

The tile patch array is 256 bytes and is not needed for the majority of water regions, change it to be optional via `std::unique_ptr`.
This reduces the size of `class WaterRegion` from 280 to 32 bytes (64 bit builds).

## Limitations

`class WaterRegion` can be further reduced to 24 bytes by removing the `tile_area` member, which is redundant. This requires a relatively large re-arrangement diff though and so is not done in the PR.

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -38,6 +38,8 @@ static inline int GetWaterRegionMapSizeY() { return Map::SizeY() / WATER_REGION_
 static inline TWaterRegionIndex GetWaterRegionIndex(int region_x, int region_y) { return GetWaterRegionMapSizeX() * region_y + region_x; }
 static inline TWaterRegionIndex GetWaterRegionIndex(TileIndex tile) { return GetWaterRegionIndex(GetWaterRegionX(tile), GetWaterRegionY(tile)); }
 
+using TWaterRegionPatchLabelArray = std::array<TWaterRegionPatchLabel, WATER_REGION_NUMBER_OF_TILES>;
+
 /**
  * Represents a square section of the map of a fixed size. Within this square individual unconnected patches of water are
  * identified using a Connected Component Labeling (CCL) algorithm. Note that all information stored in this class applies
@@ -49,10 +51,10 @@ class WaterRegion
 private:
 	std::array<TWaterRegionTraversabilityBits, DIAGDIR_END> edge_traversability_bits{};
 	bool has_cross_region_aqueducts = false;
+	bool initialized = false;
 	TWaterRegionPatchLabel number_of_patches = 0; // 0 = no water, 1 = one single patch of water, etc...
 	const OrthogonalTileArea tile_area;
-	std::array<TWaterRegionPatchLabel, WATER_REGION_NUMBER_OF_TILES> tile_patch_labels{};
-	bool initialized = false;
+	std::unique_ptr<TWaterRegionPatchLabelArray> tile_patch_labels; ///< Tile patch labels, this may be nullptr in the following trivial cases: region is invalid, region is only land (0 patches), region is only water (1 patch)
 
 	/**
 	 * Returns the local index of the tile within the region. The N corner represents 0,
@@ -110,7 +112,10 @@ public:
 	TWaterRegionPatchLabel GetLabel(TileIndex tile) const
 	{
 		assert(this->tile_area.Contains(tile));
-		return this->tile_patch_labels[GetLocalIndex(tile)];
+		if (this->tile_patch_labels == nullptr) {
+			return this->NumberOfPatches() == 0 ? INVALID_WATER_REGION_PATCH : 1;
+		}
+		return (*this->tile_patch_labels)[GetLocalIndex(tile)];
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/pathfinder/water_regions.cpp
```diff
@@ -38,6 +38,10 @@ static inline int GetWaterRegionMapSizeY() { return Map::SizeY() / WATER_REGION_
 static inline TWaterRegionIndex GetWaterRegionIndex(int region_x, int region_y) { return GetWaterRegionMapSizeX() * region_y + region_x; }
 static inline TWaterRegionIndex GetWaterRegionIndex(TileIndex tile) { return GetWaterRegionIndex(GetWaterRegionX(tile), GetWaterRegionY(tile)); }
 
+using TWaterRegionPatchLabelArray = std::array<TWaterRegionPatchLabel, WATER_REGION_NUMBER_OF_TILES>;
+
+static std::unique_ptr<TWaterRegionPatchLabelArray> _spare_labels;
```
Kuhnovic: Can be made a static local variable of ForceUpdate, for a little bit of extra encapsulation

src/pathfinder/water_regions.cpp
```diff
@@ -66,6 +70,14 @@ class WaterRegion
 		return (TileX(tile) - TileX(this->tile_area.tile)) + WATER_REGION_EDGE_LENGTH * (TileY(tile) - TileY(this->tile_area.tile));
 	}
 
+	inline bool HasNonMatchingPatchLabel(TWaterRegionPatchLabel expected_label) const
+	{
+		for (TWaterRegionPatchLabel label : *this->tile_patch_labels) {
+			if (label != expected_label) return true;
+		}
+		return false;
+	}
```
Kuhnovic: Alternatively you could keep a counter that keeps track of how many times label 1 is assigned during ForceUpdate. Then you can simply check if that counter is equal to the number of region tiles. That would save another iteration over the entire label array, and the need for this function :)

Oh and if you decide to keep it the way it is, document this function.

src/pathfinder/water_regions.cpp
```diff
@@ -196,7 +226,7 @@ class WaterRegion
 		for (int y = 0; y < WATER_REGION_EDGE_LENGTH; ++y) {
 			std::string line{};
 			for (int x = 0; x < WATER_REGION_EDGE_LENGTH; ++x) {
-				const auto label = this->tile_patch_labels[x + y * WATER_REGION_EDGE_LENGTH];
+				const auto label = this->GetLabel(TileAddXY(tile_area.tile, x, y));
```
Kuhnovic, line 216: TileAddXY would increment the index with the map's X-size for each increment in the Y direction, resulting in out of bounds.

```suggestion
				const auto label = this->GetLabel(GetLocalIndex(TileAddXY(tile_area.tile, x, y)));
```

JGRennison: GetLabel already calls GetLocalIndex if required.

src/pathfinder/water_regions.cpp
JGRennison: Iterating the array that has just been written is very cheap. This seems simpler/more fool-proof than special-casing label 1 in the fill process. I'll add some suitable documentation later.

Edit: I might just get rid of said function in favour of a call to std::find_if or std::all_of instead.

src/pathfinder/water_regions.cpp
PeterN: What is the purpose of `_spare_labels`? Seems to be to avoid a memory deallocation/allocation, rather than actually necessary.

Is the added complexity measurably necessary? A comment about its purpose would not go amiss.

JGRennison: It's not going to make any enormous difference on any benchmark. I'm happy to just remove it for simplicity's sake.

src/pathfinder/water_regions.cpp
Kuhnovic: Oh you're right. Ignore my comment then :)

Kuhnovic:
I quickly whipped up some measuring code, here are some stats. Note that these numbers include all regions on the map. There typically are many regions that aren't touched by the pathfinder, but that depends on how the game is played.

- Southern Norway, available on BaNaNaS (shameless plug). It has lots of plain sea without any land.
image
- Generated 1024x1024 temperate map with lots of water:
image
[lots_of_water.zip](https://github.com/OpenTTD/OpenTTD/files/14577667/lots_of_water.zip)

Looking at these numbers, it might be worth considering a std::bitset of 256 bits to cover all the single-patch-regions, instead of just the single-patch-all-water-regions. That way most of the regions would never need any label array. But that comes at the cost of 32 bytes for each region. It's always a trade-off...

Kuhnovic (approved):
I think the static temporary label array was nice, but if the benchmarks show no difference then we might as well drop it. 

We do have situations where a new player joins in multiplayer and suddenly has to update a lot of regions because it has no local region cache yet. That could mean a lot of allocation if there are lots of ships on a large map. I can imagine that cutting down on allocations could save some time there, but we can always investigate and add that later. I'd rather get this into the release a.s.a.p.

Issue #12283, opened 2024-03-12
[Bug]: Inconsistent default action in vehicle group dropdown


2TallTyler:
### Version of OpenTTD

14.0-RC1

### Expected result

Some dropdowns in OpenTTD have a selected value which is selected when the dropdown is opened. This contains the current value of the selection.

### Actual result

The group action dropdown has `Replace vehicles` highlighted. This is not a value, but an action, and as @PeterN points out in the comments below, should not be highlighted at all.

![highlight-real](https://github.com/OpenTTD/OpenTTD/assets/55058389/fa5433cf-b34d-4108-bf2b-f6e85f397dd7)

### Steps to reproduce

1. Create a group of vehicles.
2. Open the group's action dropdown and notice that an action is selected as the current value.

PeterN:
A default value should only be set if

1) that's the current value
or
2) that's the action that will be performed if you don't select an option.

So I think you have this backwards, and the dropdown in the first screenshot is wrong.

DefinitelyNotRau117:
Hmm, on the other hand, why not add the ability to replace vehicles for a specific group of shared orders? This can help with "quickly changing a couple of trains" without having to create a new group.

2TallTyler:
Shared order groups are not real groups. I'm not sure how much work is required to make them act like full groups, but that suggestion is definitely not in-scope for this bug report which relates only to dropdown menus.

PR #12284, opened 2024-03-12
Change: Add dividers in vehicle group action dropdown

2TallTyler:
## Motivation / Problem

The vehicle group action dropdown makes it easy to accidentally send all vehicles to a depot, as described in #11132.

In https://github.com/OpenTTD/OpenTTD/discussions/11132#discussioncomment-8635641 @DefinitelyNotRau117 suggests adding a divider to separate the safe actions from the depot actions and make it a bit harder to misclick.

I like the divider idea to better organize the actions, and if it makes it protects against misclicks, even better!

## Description

There are three types of actions: Autoreplace, group, and depot. I moved the "unsafe" depot actions to the bottom where they're furthest from the dropdown itself (and misclicks), and added dividers between each type of action.

replace

Implicit groups, like vehicles that share orders or serve a station, don't have the `Replace vehicles` action.

shared-order-implicit-group

## Limitations

Dropdown highlights are currently broken, reported in #12283. I tried to choose the correct default selection but it didn't work, so I think it's a separate bug which is out of scope here.

## Checklist for review

### DIFF
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -454,16 +454,25 @@ DropDownList BaseVehicleListWindow::BuildActionDropdownList(bool show_autoreplac
 {
 	DropDownList list;
 
-	if (show_autoreplace) list.push_back(std::make_unique<DropDownListStringItem>(STR_VEHICLE_LIST_REPLACE_VEHICLES, ADI_REPLACE, false));
-	list.push_back(std::make_unique<DropDownListStringItem>(STR_VEHICLE_LIST_SEND_FOR_SERVICING, ADI_SERVICE, false));
-	list.push_back(std::make_unique<DropDownListStringItem>(this->vehicle_depot_name[this->vli.vtype], ADI_DEPOT, false));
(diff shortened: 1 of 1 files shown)


rubidium42:
Shouldn't 'Replace vehicles' be the top choice in the group, because that isn't going to perform any actions besides opening a window? And only then the ones that could mess with the content of the group?

DefinitelyNotRau117:
> Shouldn't 'Replace vehicles' be the top choice in the group, because that isn't going to perform any actions besides opening a window? And only then the ones that could mess with the content of the group?

By the way, yeah, “remove all vehicles” looks too dangerous to be at the top of the actions list.

In 2024-03-11 this is exactly the case - the item for removal at the «safier» very bottom, and replace at the «convinient» top.
However, the items in the “Shared orders” menu have also changed their order. For the better, but still.

2TallTyler:
replace

Good point. Replace is now at the top. 🙂

rubidium42 (approved):

PR #12286, opened 2024-03-12
Fix #12279: Update random path each step

### LINKED ISSUE #12279: [Bug]: CreateRandomPath not updating tile nor dir for each step it advances
### Version of OpenTTD

OpenTTD 20240312-master-g1addeddc07

### Expected result

Move to the next tile and retrieve the next trackdirs available each step.

### Actual result

`CreateRandomPath` `tile` and `trackdir` aren't being updated for each step it advances, generating a bad `path_cache` as a result.

### Steps to reproduce

it's not updating `tile` nor `dir` for each time it advances, also, going north should alternate between `TRACKDIR_LEFT_N` and `TRACKDIR_RIGHT_N`:
image

[look at ship 3.zip](https://github.com/OpenTTD/OpenTTD/files/14571410/look.at.ship.3.zip)

SamuXarick:
## Motivation / Problem
CreateRandomPath tile and trackdir aren't being updated for each step it advances, generating a bad path_cache as a result.

## Description
Move to the next tile and retrieve the next trackdirs available each step. Fix #12279 

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -172,28 +172,28 @@ public:
 		return 'w';
 	}
 
-	/** Returns a random trackdir that can be reached from the current tile/trackdir, or INVALID_TRACK if none is available. */
-	static Trackdir GetRandomFollowUpTrackdir(const Ship *v, TileIndex tile, Trackdir dir, bool include_90_degree_turns)
+	/** Returns a random tile/trackdir that can be reached from the current tile/trackdir, or tile/INVALID_TRACK if none is available. */
+	static std::pair<TileIndex, Trackdir> GetRandomFollowUpTileTrackdir(const Ship *v, TileIndex tile, Trackdir dir, bool include_90_degree_turns)
 	{
 		TrackFollower follower(v);
 		if (follower.Follow(tile, dir)) {
-			tile = follower.m_new_tile;
 			TrackdirBits dirs = follower.m_new_td_bits;
 			if (!include_90_degree_turns) dirs &= ~TrackdirCrossesTrackdirs(dir);
 			const int strip_amount = _random.Next(CountBits(dirs));
 			for (int s = 0; s < strip_amount; ++s) RemoveFirstTrackdir(&dirs);
-			return FindFirstTrackdir(dirs);
+			return { follower.m_new_tile, FindFirstTrackdir(dirs) };
 		}
-		return INVALID_TRACKDIR;
+		return { follower.m_new_tile, INVALID_TRACKDIR };
 	}
 
 	/** Creates a random path, avoids 90 degree turns. */
 	static Trackdir CreateRandomPath(const Ship *v, TileIndex tile, Trackdir dir, ShipPathCache &path_cache, int path_length)
 	{
+		std::pair<TileIndex, Trackdir> tile_dir = { tile, dir };
 		for (int i = 0; i < path_length; ++i) {
-			const Trackdir random_dir = GetRandomFollowUpTrackdir(v, tile, dir, false);
-			if (random_dir == INVALID_TRACKDIR) break;
-			path_cache.push_back(random_dir);
+			tile_dir = GetRandomFollowUpTileTrackdir(v, tile_dir.first, tile_dir.second, false);
+			if (tile_dir.second == INVALID_TRACKDIR) break;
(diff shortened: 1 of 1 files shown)


src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -173,7 +173,7 @@ class CYapfFollowShipT
 	}
 
 	/** Returns a random trackdir that can be reached from the current tile/trackdir, or INVALID_TRACK if none is available. */
-	static Trackdir GetRandomFollowUpTrackdir(const Ship *v, TileIndex tile, Trackdir dir, bool include_90_degree_turns)
+	static Trackdir GetRandomFollowUpTrackdir(const Ship *v, TileIndex &tile, Trackdir dir, bool include_90_degree_turns)
```
Kuhnovic: TileIndex is a value type, but calling this function changes the provided tile. This is not visible from the calling site (in CreateRandomPath), which makes is a bit counter intuitive.

 Some alternative options:
- return a std::pair<TileIndex, Trackdir>
- Add a parameter, for example `TileIndex &new_tile`. Fill that tile with the new tile from the Trackfollower.

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -206,7 +207,7 @@ class CYapfFollowShipT
 	static Trackdir ChooseShipTrack(const Ship *v, TileIndex tile, DiagDirection enterdir, bool &path_found, ShipPathCache &path_cache)
 	{
 		/* Move back to the old tile/trackdir (where ship is coming from). */
-		const TileIndex src_tile = TileAddByDiagDir(tile, ReverseDiagDir(enterdir));
+		TileIndex src_tile = TileAddByDiagDir(tile, ReverseDiagDir(enterdir));
```
Kuhnovic: Leave it const. The src_tile will always be the source tile.

Kuhnovic (changes requested):

src/pathfinder/yapf/yapf_ship.cpp
SamuXarick: I went with std::pair. I tried the 2nd alternative but I still fail at dealing with & and * stuff.

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -172,28 +172,28 @@ class CYapfFollowShipT
...
 	{
 		TrackFollower follower(v);
 		if (follower.Follow(tile, dir)) {
-			tile = follower.m_new_tile;
 			TrackdirBits dirs = follower.m_new_td_bits;
 			if (!include_90_degree_turns) dirs &= ~TrackdirCrossesTrackdirs(dir);
 			const int strip_amount = _random.Next(CountBits(dirs));
 			for (int s = 0; s < strip_amount; ++s) RemoveFirstTrackdir(&dirs);
-			return FindFirstTrackdir(dirs);
+			return std::make_pair(follower.m_new_tile, FindFirstTrackdir(dirs));
```
Kuhnovic: It's not wrong to use std::make_pair, but you can also initialize a pair using braced initialization: `{follower.m_new_tile, FindFirstTrackdir(dirs)}`

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -172,28 +172,28 @@ class CYapfFollowShipT
...
 		}
-		return INVALID_TRACKDIR;
+		return std::make_pair(follower.m_new_tile, INVALID_TRACKDIR);
```
Kuhnovic: All calls to GetRandomFollowUpTrackdir end up only using the trackdir (second) of the pair. And the function is called GetRandomFollowUpTrackdir, so you might as well just return the trackdir instead of the whole pair.

Kuhnovic (changes requested):

src/pathfinder/yapf/yapf_ship.cpp
SamuXarick: No, the `m_new_tile` is used in `CreateRandomPath` `for` loop to follow up to the next tile.

Kuhnovic: Oh right, I didn't see that one. Then rename the function to GetRandomFollowUpTileTrackdir to reflect what it does :)

Kuhnovic (approved):

PR #12288, opened 2024-03-12
Change: Increase house type limit to 4096.

PeterN:
## Motivation / Problem

The house limit of 512 is quite arbitrary and can be reached quite easily. There is plenty of free bits available to houses in the map array to increase this limit.

## Description

Increase the house type limit to 4096, moving map storing from 9 bits of m4+m3[6] to 12 bits of m8.

This requires a savegame bump for map conversion.

Built on top of #12287 which makes HouseSpec allocation dynamic.

## Limitations

Variable 0x66 stores the local id in 8 bits, so any local id higher than 0xFF will be indistinguishable from 0xFF.

Why limit to 4096? No particular reason, 65535 would also be fine, using all of m8.

## Checklist for review

### DIFF
--- a/src/house.h
+++ b/src/house.h
@@ -22,12 +22,12 @@
  */
 static const uint8_t TOWN_HOUSE_COMPLETED = 3;
 
-static const HouseID NUM_HOUSES_PER_GRF = 255;    ///< Number of supported houses per NewGRF; limited to 255 to allow extending Action3 with an extended byte later on.
-
 static const uint HOUSE_NO_CLASS      = 0;
-static const HouseID NEW_HOUSE_OFFSET = 110;    ///< Offset for new houses.
-static const HouseID NUM_HOUSES       = 512;    ///< Total number of houses.
-static const HouseID INVALID_HOUSE_ID = 0xFFFF;
+static const HouseID NEW_HOUSE_OFFSET = 110; ///< Offset for new houses.
+static const HouseID NUM_HOUSES       = 4096; ///< Total number of houses.
+static const HouseID INVALID_HOUSE_ID = UINT16_MAX;
+
+static const HouseID NUM_HOUSES_PER_GRF = NUM_HOUSES; ///< Number of supported houses per NewGRF.
 
 static const uint HOUSE_NUM_ACCEPTS = 16; ///< Max number of cargoes accepted by a tile
 static const uint HOUSE_ORIGINAL_NUM_ACCEPTS = 3; ///< Original number of accepted cargo types.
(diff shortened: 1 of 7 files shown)


rubidium42 (dismissed):

PeterN:
Looks like HouseScopeResolver variable 0x66 was written with the limit of 255 houses per NewGRF in mind, so that'll be broken :/

rubidium42 (dismissed):

src/house.h
```diff
@@ -22,12 +22,12 @@
...
-
 static const uint HOUSE_NO_CLASS      = 0;
-static const HouseID NEW_HOUSE_OFFSET = 110;    ///< Offset for new houses.
-static const HouseID NUM_HOUSES       = 512;    ///< Total number of houses.
-static const HouseID INVALID_HOUSE_ID = 0xFFFF;
+static const HouseID NEW_HOUSE_OFFSET = 110; ///< Offset for new houses.
+static const HouseID NUM_HOUSES       = 4096; ///< Total number of houses.
+static const HouseID INVALID_HOUSE_ID = UINT16_MAX;
+
+static const HouseID NUM_HOUSES_PER_GRF = NUM_HOUSES; ///< Number of supported houses per NewGRF.
```
2TallTyler, line 30: Why keep this, instead of changing its one use to `NUM_HOUSES`?

src/newgrf_house.cpp
```diff
@@ -492,7 +492,7 @@ static uint32_t GetDistanceFromNearbyHouse(uint8_t parameter, TileIndex tile, Ho
 				local_houseid = nearby_house_id;
 			} else {
 				local_houseid = (hs->grf_prop.grffile == this->ro.grffile ? 1 : 2) << 8;
-				local_houseid |= hs->grf_prop.local_id;
+				local_houseid |= ClampTo<uint8_t>(hs->grf_prop.local_id);
```
2TallTyler: What's the most elegant way to break this, I wonder? Looking for a specific HouseID at a specific tile probably isn't a common usecase (I've done some silly stuff in ITL Houses, and not touched this one), but I imagine a false negative is a better failure state than a false positive. In other words, instead of returning an incorrect ID that might mean something to the author (as it's in their house GRF), would it better to return the fallback base game house ID, that they are not looking for at all?

At the very least, I think a comment for future-us would be helpful.

PeterN: Probably just adding a new variable, like what happened for stations.

Ah, 0xFF is actually an impossible value for this variable right now, as the limit is 255 (0-254). So we need to add a new variable to allow testing for higher local ids.

src/house.h
PeterN: Basically for consistency with the other `MAX_..._PER_GRF` constants.

src/newgrf_house.cpp
2TallTyler: Do you want to do that here or as a later PR?

If later, I think a comment here noting that this ID can't read IDs above 255, and a similar note in the GRF specs, would be enough and I'd be happy to approve. 🙂

PeterN: Later PR I think. Comment added here for now.

2TallTyler (approved):
Please add a note to GRF specs after merging (I assume they'll need updating to reflect the new maximum anyway). 🙂

PR #12290, opened 2024-03-13
Change: Calculate occupancy of drive through road stops from most recent entered vehicle

zacc:
## Motivation / Problem
Road vehicles sometimes queue behind each other outside of drive through stops instead of entering the stop from a different direction that is unoccupied. Unloading is delayed and profits down.

When there are two vehicles loading in a road stop and the first one in the queue leaves, the `occupied` level is reduced. However, although there is now free tile space in front of the second vehicle, it is inaccessible by vehicles arriving at the stop. The YAPF penalty is still calculated from the occupied level, even though the road stop tiles are inaccessible.

## Description
This change causes the occupied level of a drive through road stop to be calculated from the vehicle that most recently entered it.

This causes the YAPF to calculate a higher cost of using the stop, down-rating these routes. The result is that road vehicles will now sometimes favor a slightly longer route to enter the drive through stop from another direction when the occupancy is lower, however they will be able to unload sooner.

The occupancy level of the stop is reduced to 0 only when there are 0 vehicles in the stop.

## Limitations

In terms of gameplay limitations, this does not entirely solve vehicle queueing. Two or more vehicles can still queue if they have chosen the same route and arrive at the same time. However that scenario is a pre-existing behaviour. The idea solution would be for vehicles to be able to reserve bays/slots in road stops like trains can do, although that would be a much bigger change (beyond the scope of my ability). 

The player should still see less queueing with this patch.

## Checklist for review

### DIFF
--- a/src/roadstop.cpp
+++ b/src/roadstop.cpp
@@ -221,7 +221,7 @@ void RoadStop::Leave(RoadVehicle *rv)
 		this->SetEntranceBusy(false);
 	} else {
 		/* Otherwise just leave the drive through's entry cache. */
-		this->GetEntry(DirToDiagDir(rv->direction))->Leave(rv);
+		this->GetEntry(DirToDiagDir(rv->direction))->Leave();
 	}
 }
 
@@ -275,11 +275,17 @@ bool RoadStop::Enter(RoadVehicle *rv)
 
 /**
  * Leave the road stop
- * @param rv the vehicle that leaves the stop
  */
-void RoadStop::Entry::Leave(const RoadVehicle *rv)
+void RoadStop::Entry::Leave()
 {
-	this->occupied -= rv->gcache.cached_total_length;
+	this->vehicles -= 1;
+
+	/* Only re-set occupied when the stop is fully clear of RVs.
+	 * occupied should match the last-entered vehicle so how many free spaces
+	 * at the entry to the stop can be calculated by pathfinders. */
+	if (this->vehicles == 0) this->occupied = 0;
+
+	assert(this->vehicles >= 0);
 	assert(this->occupied >= 0);
 }
 
@@ -292,7 +298,9 @@ void RoadStop::Entry::Enter(const RoadVehicle *rv)
 	/* we cannot assert on this->occupied < this->length because of the
 	 * remote possibility that RVs are running through each other when
 	 * trying to prevention an infinite jam. */
+	this->vehicles += 1;
 	this->occupied += rv->gcache.cached_total_length;
+	assert(this->vehicles >= 0);
 }
 
 /**
@@ -367,8 +375,10 @@ void RoadStop::Entry::Rebuild(const RoadStop *rs, int side)
 	}
 
 	this->occupied = 0;
+	this->vehicles = 0;
 	for (const auto &it : rserh.vehicles) {
 		this->occupied += it->gcache.cached_total_length;
+		this->vehicles += 1;
 	}
 }
 
@@ -386,5 +396,5 @@ void RoadStop::Entry::CheckIntegrity(const RoadStop *rs) const
 
 	Entry temp;
 	temp.Rebuild(rs, rs->east == this);
-	if (temp.length != this->length || temp.occupied != this->occupied) NOT_REACHED();
+	if (temp.length != this->length || temp.occupied != this->occupied || temp.vehicles != this->vehicles) NOT_REACHED();
 }
--- a/src/roadstop_base.h
+++ b/src/roadstop_base.h
@@ -33,12 +33,13 @@ struct RoadStop : RoadStopPool::PoolItem<&_roadstop_pool> {
 	private:
 		int length;      ///< The length of the stop in tile 'units'
 		int occupied;    ///< The amount of occupied stop in tile 'units'
+		int vehicles;    ///< The number of vehicles in the stop
 
 	public:
 		friend struct RoadStop; ///< Oh yeah, the road stop may play with me.
 
 		/** Create an entry */
-		Entry() : length(0), occupied(0) {}
+		Entry() : length(0), occupied(0), vehicles(0) {}
 
 		/**
 		 * Get the length of this drive through stop.
@@ -58,7 +59,7 @@ struct RoadStop : RoadStopPool::PoolItem<&_roadstop_pool> {
 			return this->occupied;
 		}
 
-		void Leave(const RoadVehicle *rv);
+		void Leave();
 		void Enter(const RoadVehicle *rv);
 		void CheckIntegrity(const RoadStop *rs) const;
 		void Rebuild(const RoadStop *rs, int side = -1);


src/roadstop.cpp
```diff
@@ -291,8 +298,13 @@ void RoadStop::Entry::Enter(const RoadVehicle *rv)
 {
 	/* we cannot assert on this->occupied < this->length because of the
 	 * remote possibility that RVs are running through each other when
-	 * trying to prevention an infinite jam. */
+	 * trying to prevention an infinite jam,
+	 * or if an articulated vehicle, has partially entered the stop.
```
2TallTyler:
```suggestion
	 * or if an articulated vehicle has partially entered the stop.
```

src/roadstop.cpp
```diff
@@ -277,9 +277,16 @@ bool RoadStop::Enter(RoadVehicle *rv)
  * Leave the road stop
  * @param rv the vehicle that leaves the stop
```
2TallTyler: This parameter no longer exists. 🙂 
```suggestion

```

src/roadstop.cpp
```diff
@@ -369,6 +381,7 @@ void RoadStop::Entry::Rebuild(const RoadStop *rs, int side)
 	this->occupied = 0;
 	for (const auto &it : rserh.vehicles) {
 		this->occupied += it->gcache.cached_total_length;
+		this->vehicles += 1;
```
2TallTyler, line 381: Shouldn't we set `this->vehicles = 0;` before the loop, to match `this->occupied = 0;`?

src/roadstop.cpp
```diff
@@ -291,8 +298,13 @@ void RoadStop::Entry::Enter(const RoadVehicle *rv)
 {
 	/* we cannot assert on this->occupied < this->length because of the
 	 * remote possibility that RVs are running through each other when
-	 * trying to prevention an infinite jam. */
+	 * trying to prevention an infinite jam,
+	 * or if an articulated vehicle, has partially entered the stop.
+	 */
+	this->vehicles += 1;
 	this->occupied += rv->gcache.cached_total_length;
+	this->occupied = std::min(this->occupied, this->length);
```
2TallTyler: This, along with the modified comment about articulated vehicles, feels like a separate bugfix to clamp the occupied length to the roadstop length. I don't see how it relates to this PR. If so, it should be a separate PR.

src/roadstop.cpp
```diff
@@ -277,9 +277,16 @@ bool RoadStop::Enter(RoadVehicle *rv)
...
  * @param rv the vehicle that leaves the stop
  */
-void RoadStop::Entry::Leave(const RoadVehicle *rv)
+void RoadStop::Entry::Leave()
 {
-	this->occupied -= rv->gcache.cached_total_length;
+	this->vehicles -= 1;
+	this->vehicles = std::max(this->vehicles, 0);
+
+	if(this->vehicles == 0){
```
2TallTyler: Missing spaces, although this statement could be inline.
```suggestion
	if (this->vehicles == 0) {
```

Or:
```suggestion
	if (this->vehicles == 0) this->occupied = 0;
```

It would also be nice to have a comment explaining why we don't clear occupancy until the stop is empty, because future us looking at the code won't know to find this PR. 😃

src/roadstop_base.h
```diff
@@ -33,12 +33,13 @@ struct RoadStop : RoadStopPool::PoolItem<&_roadstop_pool> {
 	private:
 		int length;      ///< The length of the stop in tile 'units'
 		int occupied;    ///< The amount of occupied stop in tile 'units'
+		int vehicles;    ///< The amount of vehicles in the stop
```
2TallTyler:
```suggestion
		int vehicles;    ///< The number of vehicles in the stop
```

2TallTyler (commented):
Thanks for the contribution! It's nice to see this issue getting some attention. 😄 

How does this work with stops can be accessed from either side? Will a stop that's full in one lane be undesirable even if the other lane is empty? I can see some potential for edge cases, although deadlocks shouldn't be possible so it may not be a big deal.

I also wonder about an alternate solution: When a vehicle departs a road stop, the vehicles queued up behind it within the stop move forward to take its place. That way, we don't waste the space at the front of the queue. Curious to hear your thoughts on this. 😃

src/roadstop.cpp
zacc: The relation to this PR is if there's an articulated vehicle half-entered the stop, `occupied` will be greater than the `length`. The only negative effect is that this route would be even more highly penalized in ypaf than if the stop was just occupied to exactly its length with vehicles, even though in both cases the stop is blocked and no vehicles could enter.

Having said that, this edge case would still have effect in the current code - a half-entered articulated vehicle would have its full length applied to the `occupied` despite not being within the stop's tile length.

I've convinced myself to remove this to keep it simple and we can resolve it separately as you suggest.

2TallTyler: Right — because it affects the current code, I think changing it should be a separate PR.

2TallTyler:
Ah, thanks for the clarification.

I'm not advocating for moving partially-loaded vehicles, just mentioning it as a possibly simpler (but also maybe undesired) alternative. 🙂

Oh, and once you open a PR you want a clean commit history, or the commit checker CI will fail. Instead of making a new commit for feedback changes, you can amend your original commit, or at this point, squash the commits with an interactive rebase. 😃

JGRennison:
A more general comment, but it may be worth considering how this feature will work in a multiplayer-context.
In particular `RoadStop::Entry::Rebuild` does not seem to be idempotent any more. A multiplayer client joining when a stop contains a vehicle which isn't at the front due to another vehicle previously being there is now going to end up with a different occupancy value than the server. This is likely to cause problems.

zacc:
> Oh, and once you open a PR you want a clean commit history, or the commit checker CI will fail. Instead of making a new commit for feedback changes, you can amend your original commit, or at this point, squash the commits with an interactive rebase. 😃

I'll squash with the next feedback change then. I'm used to BitBucket where we can squash at merge time.

Thanks for your help!

2TallTyler:
We also squash when merging, but the Commit checker workflow will fail and prevent merging if any commits don't match the standard wording. 😉

zacc:
> A more general comment, but it may be worth considering how this feature will work in a multiplayer-context. In particular `RoadStop::Entry::Rebuild` does not seem to be idempotent with respect to the road stop state any more. A multiplayer client joining when a stop contains a vehicle which isn't at the front due to another vehicle previously being there is now going to end up with a different occupancy value than the server. This is likely to cause problems.

Possibly it could call something like `FindVehicleOnPos` for each stop tile and reconstruct the correct occupancy value? This might be too expensive for every pathfinder query but might be OK for each Rebuild call.

zacc:
@JGRennison I have worked out what I need to do to reconstruct the occupied value correctly when `RoadStop::Entry::Rebuild` is called and sure it can be done without too much trouble. I'm just working on some code polish and testing now. I've never used multiplayer before but are there mechanisms to test it locally?
In the meantime I'll move this PR back to Draft status.

JGRennison:
You can run the server and any number of clients on the same machine.

zacc:
@JGRennison , @2TallTyler ,
I have spent quite a bit of time trying to make the Rebuild function match the same `occupied` calculations of `RoadStop::Entry::Enter` but have run into some issues dealing with vehicles that are already moving inside the stop when `RoadStop::Entry::Rebuild` is called.

One of the main differences is that `RoadStop::Entry::Enter` is called and `occupied` incremented when any road vehicle enters the first tile of the stop, whether it intends to stop there or not (this is almost a bug on its own). But when Rebuild is called, the vehicle may already be within the stop tiles and may or may not intend to stop and on which tile, or be leaving- none can be assumed! To get the exact position of the vehicle might involve checking the destination, speed and status of every vehicle within the stop and would seem to be quite complicated relative to the goal of the fix and I don't really want to do it like that.

A solution could be to switch `RoadStop::Entry::Enter` to be called when the vehicle actually reaches 0 speed within the stop (whether loading or queueing) and Leave when it starts moving again. Then Rebuild only needs to consider vehicles stopped within the roadstop tiles and much easier to make calculations against them.

Another much easier solution could simply be to retain the current logic, but penalize the occupied level harder.

I'm going to experiment with the latter option a bit, but I'm interested to hear your thoughts (and from anybody else who cares to weigh in with an opinion!)

zacc:
Closed due to this not being an appropriate fix. But I'm still keen to resolve the game behaviour so reach out if you have any ideas.

PR #12292, opened 2024-03-13
Codechange: [CI] Use Azure Code Signing for Windows build

orudge:
## Motivation / Problem

Our current code signing certificate is due for renewal in a few months, and due to recent policy changes by the Certificate Authority cabal, you now need to store a certificate on a pricey hardware dongle or a certified HSM vault. For something compatible with GitHub Actions, we'd probably need to use an Azure HSM vault, but would still need to pay several hundred GBP per year for a certificate, a considerable increase compared with what we've got just now.

However, the good folks at [ImageMagick](https://github.com/dlemstra/github-stories/blob/main/2023/ImageMagick%20now%20uses%20Azure%20Code%20Signing/Readme.md) alerted me to a preview programme Microsoft is running, called Azure Code Signing. This offers a more Mac-like solution, with Microsoft providing a code signing certificate. I applied for us to join their preview programme, and was accepted. Currently, the preview is free; it's not yet clear when this will become generally available or what pricing may be involved, but one would hope it'll be considerably cheaper than the alternatives.

## Description

This PR adds support for Azure Code Signing to the build process. To keep things fairly simple, I've created a `sign_azure.bat` script that acts as an alternative for `sign.bat` (which is called within the CMake process). Microsoft does offer a GitHub Action for Azure Code Signing, but that would involve a bit more surgery. (If we'd rather do things that way, however, it's certainly possible.) CMake will choose which code signing method to use based upon which environment variables are defined, as somebody might want to fork OpenTTD and sign it with a traditional certificate perhaps.

The actual code signing is performed via a PowerShell script. Microsoft takes a checksum of the binary, uploads it to Azure, where a signature is generated and returned (so the binaries themselves are not transferred). We don't get access to the actual certificate (which is valid for only 3 days before being regenerated automatically).

## Limitations

If a user happens to be using an older version of Windows that hasn't been kept up-to-date, they may receive a certificate warning as the certification authority root certificate dates back to 2020 and has only been included in Windows for a couple of years. It's quite possible this would happen with other recent code signing certificates anyway, though.

The Azure client secret is valid for a maximum of 2 years, so will need to be regenerated and updated in the GitHub secrets store periodically.

## Checklist for review

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -100,21 +100,6 @@ jobs:
       with:
         arch: ${{ matrix.host }}
 
-    - name: Import code signing certificate
-      shell: powershell
-      # If this is run on a fork, there may not be a certificate set up - continue in this case
-      continue-on-error: true
-      run: |
-        $tempFile = [System.IO.Path]::GetTempFileName()
-        $bytes = [System.Convert]::FromBase64String($env:WINDOWS_CERTIFICATE_P12)
-        [IO.File]::WriteAllBytes($tempFile, $bytes)
-        $pwd = ConvertTo-SecureString $env:WINDOWS_CERTIFICATE_PASSWORD -AsPlainText -Force
-        Import-PfxCertificate -FilePath $tempFile -CertStoreLocation Cert:\CurrentUser\My -Password $pwd
-        Remove-Item $tempFile
-      env:
-        WINDOWS_CERTIFICATE_P12: ${{ secrets.WINDOWS_CERTIFICATE_P12 }}
-        WINDOWS_CERTIFICATE_PASSWORD: ${{ secrets.WINDOWS_CERTIFICATE_PASSWORD }}
-
     - name: Build (with installer)
       if: inputs.is_tag == 'true'
       shell: bash
@@ -131,7 +116,6 @@ jobs:
           -DHOST_BINARY_DIR=${GITHUB_WORKSPACE}/build-host \
           -DCMAKE_BUILD_TYPE=RelWithDebInfo \
           -DOPTION_SURVEY_KEY=${{ inputs.survey_key }} \
-          -DWINDOWS_CERTIFICATE_COMMON_NAME="${WINDOWS_CERTIFICATE_COMMON_NAME}" \
           # EOF
         echo "::endgroup::"
 
@@ -139,7 +123,12 @@ jobs:
         cmake --build . --target openttd
         echo "::endgroup::"
       env:
-        WINDOWS_CERTIFICATE_COMMON_NAME: ${{ secrets.WINDOWS_CERTIFICATE_COMMON_NAME }}
+        AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+        AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+        AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        AZURE_CODESIGN_ACCOUNT_NAME: ${{ secrets.AZURE_CODESIGN_ACCOUNT_NAME }}
+        AZURE_CODESIGN_ENDPOINT: ${{ secrets.AZURE_CODESIGN_ENDPOINT }}
+        AZURE_CODESIGN_PROFILE_NAME: ${{ secrets.AZURE_CODESIGN_PROFILE_NAME }}
 
     - name: Build (without installer)
       if: inputs.is_tag != 'true'
@@ -156,7 +145,6 @@ jobs:
           -DHOST_BINARY_DIR=${GITHUB_WORKSPACE}/build-host \
           -DCMAKE_BUILD_TYPE=RelWithDebInfo \
           -DOPTION_SURVEY_KEY=${{ inputs.survey_key }} \
-          -DWINDOWS_CERTIFICATE_COMMON_NAME="${WINDOWS_CERTIFICATE_COMMON_NAME}" \
           # EOF
         echo "::endgroup::"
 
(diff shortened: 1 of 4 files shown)


orudge:
> Currently, the preview is free; it's not yet clear when this will become generally available or what pricing may be involved, but one would hope it'll be considerably cheaper than the alternatives.

As if by magic, MS have just sent an invitation to a call next week where they are announcing a Public Preview, plus pricing details and so on. I'll hopefully be able to attend and find out more then.

.github/workflows/release-windows.yml
```diff
@@ -201,10 +194,15 @@ jobs:
       # If this is run on a fork, there may not be a certificate set up - continue in this case
       continue-on-error: true
       run: |
-        cd ${GITHUB_WORKSPACE}/build/bundles
-        ../../os/windows/sign.bat *.exe "${WINDOWS_CERTIFICATE_COMMON_NAME}"
+        cd ${GITHUB_WORKSPACE}/os/windows
+        ./sign_azure.bat "${GITHUB_WORKSPACE}/build/bundles"
```
TrueBrain: Why not just call `pwsh -File ..` directly here? This level of indirection doesn't seem to add much value?

.github/workflows/release-windows.yml
```diff
@@ -156,15 +145,19 @@ jobs:
...
         cmake --build . --target openttd
         echo "::endgroup::"
       env:
-        WINDOWS_CERTIFICATE_COMMON_NAME: ${{ secrets.WINDOWS_CERTIFICATE_COMMON_NAME }}
+        AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+        AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+        AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        AZURE_CODESIGN_ACCOUNT_NAME: ${{ secrets.AZURE_CODESIGN_ACCOUNT_NAME }}
+        AZURE_CODESIGN_ENDPOINT: ${{ secrets.AZURE_CODESIGN_ENDPOINT }}
+        AZURE_CODESIGN_PROFILE_NAME: ${{ secrets.AZURE_CODESIGN_PROFILE_NAME }}
```
TrueBrain, line 160: Only `AZURE_CODESIGN_PROFILE_NAME` is actually used by this step; please remove the rest: less is more :D

cmake/InstallAndPackage.cmake
```diff
@@ -177,6 +177,11 @@ elseif(WIN32)
         POST_BUILD
         COMMAND "${CMAKE_SOURCE_DIR}/os/windows/sign.bat" "$<TARGET_FILE:openttd>" "${WINDOWS_CERTIFICATE_COMMON_NAME}"
       )
+    elseif(DEFINED ENV{AZURE_CODESIGN_PROFILE_NAME})
```
TrueBrain: I suggest we just remove the old method. We will not be using it again, and it might break without us noticing. If someone wants it, it will still be in our git history. That is a much better place to archive old code, in my opinion :)

This includes the old `sign.bat`, and everything related.

os/windows/sign_azure.ps1
```diff
@@ -0,0 +1,33 @@
...
+# Arguments: sign_azure.ps1 path_to_sign
+#
+# Environment variables must be set up before use:
+#
+# AZURE_TENANT_ID
+# AZURE_CLIENT_ID
+# AZURE_CLIENT_SECRET
+# AZURE_CODESIGN_ACCOUNT_NAME
+# AZURE_CODESIGN_ENDPOINT
+# AZURE_CODESIGN_PROFILE_NAME
```
TrueBrain, line 11: In other scripts we check if something like `AZURE_CLIENT_ID` is set, and if not, print a warning about this being a fork, and silently quit. That way the `continue-on-error` can also be removed, meaning the workflow actually fails if signing would fail. Has worked out for other repos/places at least :)

TrueBrain (commented):
Looking good! Some minor stuff to tidy things up a bit :)

Owh, and we should definitely wait for the pricing announcement, just so we know what we are getting ourselves into :)

.github/workflows/release-windows.yml
orudge: I was basically just copying what we already had for the other PowerShell scripts in the repository!

.github/workflows/release-windows.yml
orudge: They're all required by the PowerShell script to perform the signing.

cmake/InstallAndPackage.cmake
orudge: OK, can do that.

os/windows/sign_azure.ps1
orudge: Will check what is used elsewhere and do something similar here.

.github/workflows/release-windows.yml
TrueBrain: But the powerscript is not called in this step. That is my point :)

orudge: > But the powerscript is not called in this step. That is my point :)

It should be - the current sign.bat is called from the nightlies [at the moment](https://github.com/OpenTTD/OpenTTD/actions/runs/8275761762/job/22643263613#step:15:1042) in the "Build (without installer)" step.

TrueBrain: Ah, I see what you mean; meh. It is so bulky now, which makes it hard to read / maintain, so I was hoping we could slim it down. But no, you are right.

os/windows/sign_azure.ps1
```diff
@@ -0,0 +1,41 @@
...
+    # Files folder
+    [Parameter(Mandatory=$true, Position=0)]
+    $FilesFolder
+)
+
+if (!$Env:AZURE_CODESIGN_ENDPOINT -or !$Env:AZURE_CODESIGN_ACCOUNT_NAME -or !$Env:AZURE_CODESIGN_PROFILE_NAME -or
+	!$Env:AZURE_TENANT_ID -or !$Env:AZURE_CLIENT_ID -or !$Env:AZURE_CLIENT_SECRET)
+{
+	"The required environment variables have not been set up. Please check the script source for details."
+	"Skipping signing."
```
TrueBrain:
```suggestion
	"Codesigning variables not found; most likely running in a fork. Skipping signing."
```

TrueBrain (commented):
Looking good! One minor thing :)

TrueBrain (approved):

orudge:
For the historical record, I attended a meeting with Microsoft today regarding the public preview and eventual production release of the Azure Code Signing service. This included information on pricing, and while I don't know that I'm at liberty to share that information at this time, I can confirm that it should be quite affordable for us (and certainly cheaper than an EV code signing certificate and so on).

PR #12294, opened 2024-03-13
Codechange: streamline/harden join process

rubidium42:
## Motivation / Problem

The check for NewGRFs is done in the handling of `CLIENT_JOIN`, instead of `SendNewGRFCheck`. This causes that when checks are reordered, the logic whether the NewGRF check has to be done needs to be moved around. The game and company password checking already works in this manner, so make it more uniform.

Next the routines for sending NewGRF check and password requests bailed out to the next function without any validation. Technically this meant that you were free to call `SendNeedCompanyPassword` whenever the client had not checked NewGRFs or the game password, and it would just skip those steps.

The underlying motivation for these changes is that I want to move the `SendNeedGamePassword` to earlier in the process when introducing an encrypted network connection based on monocypher. With this the password can be used to derive the encryption key via PAKE (Password Authenticated Key Exchange). As a result these states need to be reordered.

## Description

Move the check for NewGRFs into `SendNewGRFCheck`.

For `SendNewGRFCheck`, `SendNeedGamePassword` and `SendNeedCompanyPassword`, first check the status to be *only* the preceding status and update the status before skipping to the next step.
This means that we will step through each of the statuses, and not allow skipping any statuses.

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -401,8 +401,8 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::SendNewGRFCheck()
 {
 	Debug(net, 9, "client[{}] SendNewGRFCheck()", this->client_id);
 
-	/* Invalid packet when status is STATUS_NEWGRFS_CHECK or higher */
-	if (this->status >= STATUS_NEWGRFS_CHECK) return this->CloseConnection(NETWORK_RECV_STATUS_MALFORMED_PACKET);
+	/* Invalid packet when status is anything but STATUS_INACTIVE. */
+	if (this->status != STATUS_INACTIVE) return this->CloseConnection(NETWORK_RECV_STATUS_MALFORMED_PACKET);
 
 	Debug(net, 9, "client[{}] status = NEWGRFS_CHECK", this->client_id);
 	this->status = STATUS_NEWGRFS_CHECK;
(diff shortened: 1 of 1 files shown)


TrueBrain:
You already kinda mention it in your description, and maybe you change this in a later PR (which would be fine). But one thing always bothered me: we first check NewGRF, then game password, then company password.

I would expect the order to be: game password, NewGRF, company password. First check if you can get into the gate, then check if you comply with the regulation, then check if you can get into your office.

What do you think?

rubidium42:
> You already kinda mention it in your description, and maybe you change this in a later PR (which would be fine). But one thing always bothered me: we first check NewGRF, then game password, then company password.
> 
> I would expect the order to be: game password, NewGRF, company password. First check if you can get into the gate, then check if you comply with the regulation, then check if you can get into your office.
> 
> What do you think?

I think it's a good idea. Let me cook up a PR... #12300 ;)

src/network/network_server.cpp
```diff
@@ -399,8 +399,19 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::SendError(NetworkErrorCode err
 /** Send the check for the NewGRFs. */
 NetworkRecvStatus ServerNetworkGameSocketHandler::SendNewGRFCheck()
 {
+	if (_grfconfig == nullptr) {
+		/* There are no NewGRFs, continue with the game password. */
+		return this->SendNeedGamePassword();
+	}
```
TrueBrain: If you don't plan to squash this PR, maybe already move this blob to where the next commit will move it to?

TrueBrain (approved):

PR #12296, opened 2024-03-14
Codechange: split initiating of joining and identification of the client

rubidium42:
## Motivation / Problem

When encrypting the communication between the server and the client, it would be bad to send the identity of the client over the wire before encryption got initiated.

## Description

At the moment you start the joining process by sending a "CLIENT_JOIN" packet with your (NewGRF) version, client name and which company you want to join. 
At the moment the server first checks the (NewGRF) version. If they do not match, you are kicked out. When they do match, it will create a `NetworkClientInfo` with your name attached to the socket.

This PR just adds an intermediate state `IDENTIFY` and splits the "CLIENT_JOIN" into "CLIENT_JOIN" with just the (NewGRF) version, and "CLIENT_IDENTIFY" with the client name and company to play as.

This will make it possible to do the key exchange between `CLIENT_JOIN` and the identification.

## Limitations

Built upon #12294, so that needs to be merged first or the combined PRs gets merged.

Technically the "CLIENT_JOIN" packet is not fully backward compatible. When trying to join a server with a version before 15, we won't be sending the client's name and company to join any more; however, since the version check happens and the NewGRF version is incremented with each major release, this should not be a big issue as the client would get disconnected before that. Except when server/client are built without version string between branching of 14 and merging this.
When trying to join a server with a version before 1.2, there is no NewGRF version. So technically we should not be sending that to those versions, but the ship for fixing that has long sailed. Even then, there are only problems when server/client is built without version string and then the version check might not fail but things will fail at a later point. Since 14 already had that problem, it's not something to be solved by this PR.

## Checklist for review

### DIFF
--- a/src/network/core/tcp_game.h
+++ b/src/network/core/tcp_game.h
@@ -56,7 +56,10 @@ enum PacketGameType : uint8_t {
 	 * the map and other important data.
 	 */
 
-	/* After the join step, the first is checking NewGRFs. */
+	/* After the initial join, the next step is identification. */
+	PACKET_CLIENT_IDENTIFY,              ///< Client telling the server the client's name and requested company.
+
+	/* After the identify step, the next is checking NewGRFs. */
 	PACKET_SERVER_CHECK_NEWGRFS,         ///< Server sends NewGRF IDs and MD5 checksums for the client to check.
 	PACKET_CLIENT_NEWGRFS_CHECKED,       ///< Client acknowledges that it has all required NewGRFs.
 
@@ -162,10 +165,13 @@ protected:
 
 	/**
 	 * Try to join the server:
-	 * string  OpenTTD revision (norev000 if no revision).
-	 * string  Name of the client (max NETWORK_NAME_LENGTH).
-	 * uint8_t   ID of the company to play as (1..MAX_COMPANIES).
-	 * uint8_t   ID of the clients Language.
+	 * string   OpenTTD revision (norev000 if no revision).
+	 * uint32_t NewGRF version (added in 1.2).
+	 * string   Name of the client (max NETWORK_NAME_LENGTH) (removed in 15).
+	 * uint8_t  ID of the company to play as (1..MAX_COMPANIES) (removed in 15).
+	 * uint8_t  ID of the clients Language (removed in 15).
+	 * string   Client's unique identifier (removed in 1.0).
+	 *
 	 * @param p The packet that was just received.
 	 */
 	virtual NetworkRecvStatus Receive_CLIENT_JOIN(Packet &p);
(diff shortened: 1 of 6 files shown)


LordAro:
Is this a codechange as such? It could conceivably be user visible, if anyone is making their own servers...

rubidium42:
> Is this a codechange as such? It could conceivably be user visible, if anyone is making their own servers...

I'd tackle the question in reverse: is this something that should end up in the changelog for 15? Will that user notice anything?
I do not think we should think about the hypothetical effect this change could have on the users of someone developing their own server. But yes, if you have made your own server that ignores the version check then a client with this PR will fail to join. However, many other codechanges could conceivably be user visible via for example desyncs.

TrueBrain:
We always states the game network protocol is internal, and we do not support compatibility or documentation here. This addresses both @LordAro's comment and the talk in the limitation.

This is different from things like admin ports etc, where we do explicitly want others to write software against. But if someone writes a custom server, they will have to manually monitor these files and update where needed. So I am not worried about "backwards compatibility" etc here.

src/network/core/tcp_game.h
```diff
@@ -162,10 +165,17 @@ class NetworkGameSocketHandler : public NetworkTCPSocketHandler {
...
+	 * string   OpenTTD revision (norev000 if no revision).
+	 * uint32_t NewGRF version.
+	 *
+	 * Before 15 there used to be more fields:
+	 * string   OpenTTD revision (norev000 if no revision).
+	 * uint32_t NewGRF version (added in 1.2).
+	 * string   Name of the client (max NETWORK_NAME_LENGTH).
+	 * uint8_t  ID of the company to play as (1..MAX_COMPANIES).
+	 * uint8_t  ID of the clients Language (functionally always set to 0).
+	 * string   Client's unique identifier (removed in 1.0).
```
TrueBrain, line 173: Not sure if documenting that it used to be different before, is useful. We have a git history for that, I guess.

rubidium42: I added this documentation because for all intents and purposes `CLIENT_JOIN`, `SERVER_FULL`, `SERVER_BANNED`, `SERVER_ERROR` are supposed to be stable. Adding this documentation shows that `CLIENT_JOIN` is not fully stable, even though earlier documentation (PacketType) suggests that it must be.

There is also `CLIENT_GAME_INFO` and `SERVER_GAME_INFO`, though they have internal versioning and all versions written out.

TrueBrain: GameInfo is different, as in they have all versions implemented in the code too. This is just a comment that can also be found back in git history. So I am still trying to find a reason to keep this here. As in, what purpose it serves to keep it around, other than creating confusion as there is no code dealing with the pre-15 information.

Compare it with `PACKET_CLIENT_UNUSED`, which also no longer contains any details how the packet looked.

If you want to keep it, maybe do it a bit in reverse:

```
	 * string   OpenTTD revision (norev000 if no revision).
	 * uint32_t NewGRF version (added in 1.2).
	 * string   Name of the client (max NETWORK_NAME_LENGTH) (removed in 15).
	 * uint8_t  ID of the company to play as (1..MAX_COMPANIES) (removed in 15).
	 * uint8_t  ID of the clients Language (removed in 15).
	 * string   Client's unique identifier (removed in 15).
```

That also highlights better that the first two fields are kept around.

TrueBrain (approved):

PR #12297, opened 2024-03-14
Change: Add per-company group numbers.

PeterN:
## Motivation / Problem

When creating groups, the default group name is `Group <index>`, where index is a global number starting at zero, and shared by all companies.

This results in arbitrary looking numbers when creating groups and is affected by other companies, and the first group created by a company is "Group 0",

image

## Description

Instead, assign a per-company unique number to each group. The first group for each company will be named "Group 1". Other companies creating/deleting groups no longer affects another company's groups.

image

This uses the existing FreeUnitIDGenerator system (which is now slightly misnamed) to get the next available free number. Similar to unit numbers, when a company is bought out, new group numbers are allocated.

Requires a savegame bump to store the group number.

## Limitations

Regression test results change as it expected to see "Group 0", but the first group is now "Group 1".

## Checklist for review

### DIFF
--- a/src/group_cmd.cpp
+++ b/src/group_cmd.cpp
@@ -350,8 +350,9 @@ std::tuple<CommandCost, GroupID> CmdCreateGroup(DoCommandFlag flags, VehicleType
 		g->vehicle_type = vt;
 		g->parent = INVALID_GROUP;
 
+		Company *c = Company::Get(g->owner);
+		g->number = c->freegroups.UseID(c->freegroups.NextID());
 		if (pg == nullptr) {
-			const Company *c = Company::Get(_current_company);
 			g->livery.colour1 = c->livery[LS_DEFAULT].colour1;
 			g->livery.colour2 = c->livery[LS_DEFAULT].colour2;
 			if (c->settings.renew_keep_length) SetBit(g->flags, GroupFlags::GF_REPLACE_WAGON_REMOVAL);
@@ -397,14 +398,15 @@ CommandCost CmdDeleteGroup(DoCommandFlag flags, GroupID group_id)
 		/* Update backupped orders if needed */
 		OrderBackup::ClearGroup(g->index);
 
-		/* If we set an autoreplace for the group we delete, remove it. */
-		if (_current_company < MAX_COMPANIES) {
-			Company *c;
+		if (g->owner < MAX_COMPANIES) {
+			Company *c = Company::Get(g->owner);
 
-			c = Company::Get(_current_company);
+			/* If we set an autoreplace for the group we delete, remove it. */
 			for (EngineRenew *er : EngineRenew::Iterate()) {
 				if (er->group_id == g->index) RemoveEngineReplacementForCompany(c, er->from, g->index, flags);
 			}
+
+			c->freegroups.ReleaseID(g->number);
 		}
 
 		VehicleType vt = g->vehicle_type;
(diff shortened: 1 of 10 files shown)


src/saveload/group_sl.cpp
```diff
@@ -59,6 +60,13 @@ struct GRPSChunkHandler : ChunkHandler {
 				g->livery.colour1 = c->livery[LS_DEFAULT].colour1;
 				g->livery.colour2 = c->livery[LS_DEFAULT].colour2;
 			}
+
+			Company *c = Company::Get(g->owner);
+			if (IsSavegameVersionBefore(SLV_GROUP_NUMBERS)) {
+				g->number = c->freegroups.UseID(c->freegroups.NextID());
```
2TallTyler: I'm not sure renumbering existing groups is the right choice. I can't think of a reason that players would use automatically-numbered groups, but I've learned not to assume I would know. 🙂 

Maybe it would be better to keep the existing number?

```suggestion
				g->number = c->freegroups.UseID(g->index);
```

src/group_cmd.cpp
```diff
@@ -398,13 +399,13 @@ CommandCost CmdDeleteGroup(DoCommandFlag flags, GroupID group_id)
 		OrderBackup::ClearGroup(g->index);
 
 		/* If we set an autoreplace for the group we delete, remove it. */
```
2TallTyler: This comment no longer applies to the entire `if` block and should move down a couple lines, I think. 🙂

2TallTyler (commented):
Code-wise this is an addition, but to the player I wonder if `Change: Use per-company group numbers` is a better commit title for next year's changelog.

src/saveload/group_sl.cpp
PeterN: If you want to keep the group number then `g->index` is the wrong thing to use.

But also, while groups themselves don't care about unique names, the ID generator doesn't have any idea what to do with duplicated numbers and will not do the right thing (and would be inconsistent in multiplayer.)

2TallTyler: Ah, well then, changing numbers is probably fine.

JGRennison: If players didn't use automatically-numbered groups, there would be no need for this PR, so that seems a bit odd.
n = 1, but I use both automatically and manually named groups.
Having existing groups renumbered on updating the game would seem to me a bit surprising.
Given that g->index is by definition unique, duplicated numbers at savegame after-load time does not seem like it should be a problem. This after-load process will not be run in a multi-player context.

PeterN: You are correct, I was thinking about completely the wrong thing -- taking over a company, which deliberately renumbers just it does for unit numbers.

Sorry!

So, yes, it is feasible to use `g->index` during savegame conversion.

2TallTyler (dismissed):

PeterN:
Changed the list sort fallback to use `number` instead of `index` as well. Although not sure it makes a lot of difference.

2TallTyler (dismissed):
Number and index might not always be in the same order because of ID reuse, but I don't see a problem with that -- and if the user has automatically-numbered groups ordering by the visible number instead of the invisible index is better. 🙂

PeterN:
I've moved the part that converts index to number and/or flags each group number in use to `AfterLoadGame()`.

2TallTyler (dismissed):

PeterN:
Just an update due to savegame bump.

2TallTyler (approved):

PR #12300, opened 2024-03-14
Feature: encrypted game sockets

rubidium42:
## Motivation / Problem

The game socket is not encrypted, making passwords that are sent over it trivially extractable.

## Description

Use Monocypher to perform a key exchange, and use monocypher to encrypt the content of `Packet`s.

The key exchange is done directly after the join, where the server sends its public key, a nonce and whether or not the server's password is used in the key exchange. When the client receives the message, it has to check whether it needs to ask for the password, and once it has the password it can start the key exchange where it creates the keys for both directions of the connection locally using its own public and private key, the servers public key and nonce, and the password.

The client will then generate a message and sign that, and send that together with its public key to the server. The server then performs the same key exchange now it knows the public key of the client, and decrypts and checks the signature of the message from the client. When that is wrong, the client is kicked out for not knowing the password .

The key exchange will also yield a set of keys for client-to-server and server-to-client encryption. From the moment those are known, they will be used for encrypting and decrypting the packets. For this the first two bytes with the packet size remain untouched, but 16 bytes are injected for the message authentication code (MAC), and after that the packet is filled as normal. When sending, the content is encrypted and the MAC is filled. For receiving messages the same steps happen, but in reverse. 

Since we could be sending an error message back for kicking the user because the password was incorrect, a new packet has been introduced to instruct the client to enable encryption. After this all packets are considered encrypted, and when a packet is too small to be encrypted, the connection is closed due to receiving invalid data.

This PR also contains basic authorized keys for the game connection, but that can be split to a different PR.

## Limitations

To keep the scope small, this does not:
* handle company passwords
* handle other protocols, such as the admin connection like #11878 does

This is built upon #12017, #12294 and #12296.

## Checklist for review

### DIFF
--- a/src/network/core/packet.cpp
+++ b/src/network/core/packet.cpp
@@ -48,7 +48,14 @@ Packet::Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size)
 Packet::Packet(NetworkSocketHandler *cs, PacketType type, size_t limit) : pos(0), limit(limit), cs(cs)
 {
 	/* Allocate space for the the size so we can write that in just before sending the packet. */
-	this->Send_uint16(0);
+	size_t size = EncodedLengthOfPacketSize();
+	if (cs != nullptr && cs->send_encryption_handler != nullptr) {
+		/* Allocate some space for the message authentication code of the encryption. */
+		size += cs->send_encryption_handler->MACSize();
+	}
+	assert(this->CanWriteToPacket(size));
+	this->buffer.resize(size, 0);
+
 	this->Send_uint8(type);
 }
 
@@ -63,6 +70,13 @@ void Packet::PrepareToSend()
 
 	this->buffer[0] = GB(this->Size(), 0, 8);
 	this->buffer[1] = GB(this->Size(), 8, 8);
+
+	if (cs != nullptr && cs->send_encryption_handler != nullptr) {
+		size_t offset = EncodedLengthOfPacketSize();
+		size_t mac_size = cs->send_encryption_handler->MACSize();
+		size_t message_offset = offset + mac_size;
+		cs->send_encryption_handler->Encrypt(std::span(&this->buffer[offset], mac_size), std::span(&this->buffer[message_offset], this->buffer.size() - message_offset));
+	}
 
 	this->pos  = 0; // We start reading from here
 	this->buffer.shrink_to_fit();
@@ -259,11 +273,21 @@ bool Packet::ParsePacketSize()
 
 /**
  * Prepares the packet so it can be read
+ * @return True when the packet was valid, otherwise false.
  */
-void Packet::PrepareToRead()
+bool Packet::PrepareToRead()
 {
 	/* Put the position on the right place */
 	this->pos = static_cast<PacketSize>(EncodedLengthOfPacketSize());
+
+	if (cs == nullptr || cs->receive_encryption_handler == nullptr) return true;
+
+	size_t mac_size = cs->receive_encryption_handler->MACSize();
+	if (this->buffer.size() <= pos + mac_size) return false;
+
+	bool valid = cs->receive_encryption_handler->Decrypt(std::span(&this->buffer[pos], mac_size), std::span(&this->buffer[pos + mac_size], this->buffer.size() - pos - mac_size));
+	this->pos += static_cast<PacketSize>(mac_size);
+	return valid;
 }
 
 /**
@@ -273,7 +297,9 @@ void Packet::PrepareToRead()
 PacketType Packet::GetPacketType() const
 {
 	assert(this->Size() >= EncodedLengthOfPacketSize() + EncodedLengthOfPacketType());
-	return static_cast<PacketType>(buffer[EncodedLengthOfPacketSize()]);
+	size_t offset = EncodedLengthOfPacketSize();
+	if (cs != nullptr && cs->send_encryption_handler != nullptr) offset += cs->send_encryption_handler->MACSize();
+	return static_cast<PacketType>(buffer[offset]);
 }
 
 /**
--- a/src/network/core/packet.cpp
+++ b/src/network/core/packet.cpp
@@ -223,7 +223,7 @@ bool Packet::CanReadFromPacket(size_t bytes_to_read, bool close_connection)
  */
 bool Packet::HasPacketSizeData() const
 {
-	return this->pos >= sizeof(PacketSize);
+	return this->pos >= EncodedLengthOfPacketSize();
 }
 
 /**
@@ -250,10 +250,10 @@ bool Packet::ParsePacketSize()
 	/* If the size of the packet is less than the bytes required for the size and type of
 	 * the packet, or more than the allowed limit, then something is wrong with the packet.
 	 * In those cases the packet can generally be regarded as containing garbage data. */
-	if (size < sizeof(PacketSize) + sizeof(PacketType) || size > this->limit) return false;
+	if (size < EncodedLengthOfPacketSize() + EncodedLengthOfPacketType() || size > this->limit) return false;
 
 	this->buffer.resize(size);
-	this->pos = sizeof(PacketSize);
+	this->pos = static_cast<PacketSize>(EncodedLengthOfPacketSize());
 	return true;
 }
 
@@ -263,7 +263,7 @@ bool Packet::ParsePacketSize()
 void Packet::PrepareToRead()
 {
 	/* Put the position on the right place */
-	this->pos = sizeof(PacketSize);
+	this->pos = static_cast<PacketSize>(EncodedLengthOfPacketSize());
 }
 
 /**
@@ -272,8 +272,8 @@ void Packet::PrepareToRead()
  */
 PacketType Packet::GetPacketType() const
 {
-	assert(this->Size() >= sizeof(PacketSize) + sizeof(PacketType));
-	return static_cast<PacketType>(buffer[sizeof(PacketSize)]);
+	assert(this->Size() >= EncodedLengthOfPacketSize() + EncodedLengthOfPacketType());
+	return static_cast<PacketType>(buffer[EncodedLengthOfPacketSize()]);
 }
 
 /**
--- a/src/network/core/packet.h
+++ b/src/network/core/packet.h
@@ -40,6 +40,8 @@ typedef uint8_t  PacketType; ///< Identifier for the packet
  *     (year % 4 == 0) and ((year % 100 != 0) or (year % 400 == 0))
  */
 struct Packet {
+	static constexpr size_t EncodedLengthOfPacketSize() { return sizeof(PacketSize); }
+	static constexpr size_t EncodedLengthOfPacketType() { return sizeof(PacketType); }
 private:
 	/** The current read/write position in the packet */
 	PacketSize pos;
@@ -52,7 +54,7 @@ private:
 	NetworkSocketHandler *cs;
 
 public:
-	Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size = sizeof(PacketSize));
+	Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size = EncodedLengthOfPacketSize());
 	Packet(NetworkSocketHandler *cs, PacketType type, size_t limit = COMPAT_MTU);
 
 	/* Sending/writing of packets */
--- a/src/network/core/packet.h
+++ b/src/network/core/packet.h
@@ -74,7 +74,7 @@ public:
 	bool HasPacketSizeData() const;
 	bool ParsePacketSize();
 	size_t Size() const;
-	void PrepareToRead();
+	[[nodiscard]] bool PrepareToRead();
 	PacketType GetPacketType() const;
 
 	bool   CanReadFromPacket(size_t bytes_to_read, bool close_connection = false);
--- a/src/network/core/tcp_game.h
+++ b/src/network/core/tcp_game.h
@@ -15,7 +15,6 @@
 #include "os_abstraction.h"
 #include "tcp.h"
 #include "../network_type.h"
-#include "../network_crypto.h"
 #include "../../core/pool_type.hpp"
 #include <chrono>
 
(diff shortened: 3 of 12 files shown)


src/network/core/packet.cpp
```diff
@@ -264,6 +291,15 @@ void Packet::PrepareToRead()
 {
 	/* Put the position on the right place */
 	this->pos = sizeof(PacketSize);
+
+	if (cs == nullptr || cs->receive_encryption_handler == nullptr) return;
+
+	size_t mac_size = cs->receive_encryption_handler->MACSize();
+	if (this->buffer.size() >= 1 + pos + mac_size) {
+		/* There is at least enough data to have something to decompress. */
+		cs->receive_encryption_handler->Decrypt(std::span(&this->buffer[pos], mac_size), std::span(&this->buffer[pos + mac_size], this->buffer.size() - pos - mac_size));
```
Milek7: Does this means that *any* small packet can be injected into connection and it will be accepted? Doesn't seem right.

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/**
+ * Create a new secret key that's filled with random bytes.
+ * @return The randomly filled key.
+ */
+/* static */ X25519SecretKey X25519SecretKey::CreateRandom()
+{
+	X25519SecretKey secret_key;
+	RandomBytesWithFallback(secret_key);
```
Milek7: `WithFallback` doesn't look appropriate for cryptography.

src/network/core/packet.cpp
```diff
@@ -38,16 +38,24 @@ Packet::Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size)
...
  */
-Packet::Packet(PacketType type, size_t limit) : pos(0), limit(limit), cs(nullptr)
+Packet::Packet(NetworkSocketHandler *cs, PacketType type, size_t limit) : pos(0), limit(limit), cs(cs)
 {
 	/* Allocate space for the the size so we can write that in just before sending the packet. */
-	this->Send_uint16(0);
+	size_t size = sizeof(PacketSize);
+	if (cs != nullptr && cs->send_encryption_handler != nullptr) {
+		/* Allocate some space for the message authentication code of the encryption. */
+		size += cs->send_encryption_handler->MACSize();
```
Milek7, line 54: Shouldn't MTU also be adjusted for that, like in https://github.com/Milek7/OpenTTD/commit/d3653710d35e44c63397601e67c31d2c10f9a0ae#diff-31dd9b11e3c119e4e65cf2b3dc0ab704b71d41728d0e1ad8ecc109cfee907ed7L177?

src/network/core/packet.cpp
Milek7: It also looks like packet type field isn't protected.

src/network/core/packet.cpp
rubidium42: No. My version just reserves `MACSize` bytes after the packet size, just like space is reserved to store the packet size. Both the filling the space for the MAC and updating the packet size is done at a later stage, just before sending the packet..

src/network/core/packet.cpp
rubidium42: > Does this means that _any_ small packet can be injected into connection and it will be accepted? Doesn't seem right.

Yes it does. Though I'm running into another issue while implementing authorized keys, so this needs (way) more thought. The problem is that authentication of authorized key can fail, and then it should retry. However, the connection is already encrypted at the client by that point while the server has not enabled encryption yet because the authentication has not succeeded yet.


> It also looks like packet type field isn't protected.

What do you mean? The packet type is the first byte after the MAC, so it should be included in the encrypted. Or is this about `protected` in the sense of access specifiers?

src/network/network_crypto.cpp
rubidium42: Technically true, but at some point there is a limit what you can do. That function will, under our supported operating systems return cryptographically-strong random.

The major difference between your (libhydrogen) and our implementation, for the supported platforms, is that your implementation `abort()`s/crashes without useful message to the user the whole application and ours falls back to a worse random when for some  reason the OS does not return random data.

Note that our implementation follows the manual of Monocypher for our supported platforms. So it could only be a problem on some esoteric platform, or under very weird circumstance.

src/network/core/packet.cpp
Milek7: > Yes it does.

Perhaps it could enforce that packet is of ERROR type in that case? Though the implicit assumptions that errors are small feels weird.

> The packet type is the first byte after the MAC, so it should be included in the encrypted.

Ah sorry, I didn't notice that.

src/network/core/packet.cpp
Milek7: But it does increase real packet size on the wire. Though that doesn't really matter for TCP and indeed I noticed now that in the meantime MTU for TCP was split and increased anyway, so it's fine.

rubidium42: I'm not understanding your concern with the MTU.

The packet itself limits the size to some limit, be it the old 1460 MTU or 32k. So, the maximum size for packets that do regularly reach it, i.e. the map packets, has not changed at all.
Due to the MAC there will be more data, so the last packet might be bigger and/or there might be more packets but that has nothing to do with having to adjust the MTU.

In the context of your example, I have already increased `this->current->size` by `MACSize` by the point you get to that selected line of code. So the MTU does not need to be amended.

Milek7: I don't have concern with MTU now. (it was under assumption that low 1460 value was there for a specific reason, but that's not relevant anymore)

src/network/network_crypto.h
```diff
@@ -0,0 +1,183 @@
...
+};
+
+/**
+ * Base class for client side cryptographic authentication handlers.
+ */
+class NetworkAuthenticationClientHandler : public NetworkAuthenticationHandler {
+public:
+	/** The processing result of receiving a request. */
+	enum RequestResult {
+		AWAIT_USER_INPUT, ///< We have requested some user input, but must wait on that.
```
TrueBrain: This name feels very generic. But I guess it is really specific: `AWAIT_USER_PASSWORD_INPUT` or something?

src/network/network_crypto.h
```diff
@@ -0,0 +1,183 @@
...
+		IMMEDIATE_RESPONSE, ///< We can immediately respond to the server.
```
TrueBrain: This name took me a bit to understand, but I guess it is meant as counter balance to `AWAIT`. Maybe `READY`? Not sure if that is any better :)

src/network/network_crypto.cpp
```diff
@@ -284,12 +284,6 @@ NetworkAuthenticationServerHandler::ResponseResult X25519AuthenticationHandler::
 	return NetworkAuthenticationServerHandler::AUTHENTICATED;
 }
 
-/* virtual */ NetworkAuthenticationClientHandler::RequestResult X25519PlainClientHandler::ReceiveRequest(struct Packet &p)
-{
-	bool success = this->X25519AuthenticationHandler::ReceiveRequest(p);
-	return success ? NetworkAuthenticationClientHandler::IMMEDIATE_RESPONSE : NetworkAuthenticationClientHandler::INVALID;
-}
```
TrueBrain: Guess this should go one commit up?

Same as the snippet one below.

src/network/core/packet.cpp
```diff
@@ -48,7 +48,14 @@ Packet::Packet(NetworkSocketHandler *cs, size_t limit, size_t initial_read_size)
 Packet::Packet(NetworkSocketHandler *cs, PacketType type, size_t limit) : pos(0), limit(limit), cs(cs)
 {
 	/* Allocate space for the the size so we can write that in just before sending the packet. */
-	this->Send_uint16(0);
+	size_t size = sizeof(PacketSize);
```
TrueBrain: Instead of assuming `sizeof(PacketSize)` is also the encoded byte-length for our packet size, maybe we should add a simple wrapper function that does that for us. Something like `GetLengthByteSize()`.

There are several instances in this change that uses `sizeof(PacketSize)`, meaning: the encoded length of the packet size. Being a bit more explicit about this intention makes reading it a bit easier.

(in other words: I just spend 10 minutes trying to understand where this was coming from, to finally find out `PacketSize` is just a typing alias for `uint16_t` :D )

src/network/core/udp.cpp
```diff
@@ -137,7 +137,10 @@ void NetworkUDPSocketHandler::ReceivePackets()
 				Debug(net, 1, "Received a packet with mismatching size from {}", address.GetAddressAsString());
 				continue;
 			}
-			p.PrepareToRead();
+			if (!p.PrepareToRead()) {
+				Debug(net, 1, "Invalid packet received (too small / MAC validation failed)");
```
TrueBrain: I wonder if `MAC` is useful to the user. Maybe: `too small / encryption error` ? (also for the other instances of this message)

src/network/core/tcp_game.h
```diff
@@ -56,13 +56,19 @@ enum PacketGameType : uint8_t {
 	 * the map and other important data.
 	 */
 
-	/* After the join step, the first is checking NewGRFs. */
+	/* After the join step, the first perform game authentication and enabling encryption. */
+	PACKET_SERVER_AUTH_REQUEST,          ///< The server requests the client to authenticate using a number of methods.
+	PACKET_CLIENT_AUTH_RESPONSE,         ///< The client responds to the authentication request.
+	PACKET_SERVER_ENABLE_ENCRYPTION,     ///< The server requests to enable encryption with the keys of the last \c PACKET_CLIENT_AUTH_RESPONSE..
```
TrueBrain:
```suggestion
	PACKET_SERVER_ENABLE_ENCRYPTION,     ///< The server requests to enable encryption with the keys of the last \c PACKET_CLIENT_AUTH_RESPONSE.
```

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,448 @@
...
+ * @param secret_key The secret key to read/validate/fix.
+ * @param public_key The public key to update.
+ * @return The valid secret key.
+ */
+/* static */ X25519SecretKey X25519AuthorizedKeyClientHandler::GetValidSecretKeyAndUpdatePublicKey(std::string &secret_key, std::string &public_key)
+{
+	X25519SecretKey key{};
+	if (!ConvertHexToBytes(secret_key, key)) {
+		Debug(net, 0, "[crypto] Found invalid secret key, creating a new random key");
+		RandomBytesWithFallback(key);
```
TrueBrain:
```suggestion
		key->CreateRandom();
```

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,448 @@
...
+		secret_key = FormatArrayAsHex(key);
+	}
+
+	public_key = FormatArrayAsHex(key.CreatePublicKey());
+	return key;
+}
+
+/* virtual */ NetworkAuthenticationServerHandler::ResponseResult X25519AuthorizedKeyServerHandler::ReceiveResponse(Packet &p)
+{
+	ResponseResult result= this->X25519AuthenticationHandler::ReceiveResponse(p, {});
```
TrueBrain:
```suggestion
	ResponseResult result = this->X25519AuthenticationHandler::ReceiveResponse(p, {});
```

TrueBrain (commented):
Starting to get in shape nicely :)

From my perspective, what would be easiest, if we PR out the "span" and "add crypto helpers" in another PR. And then have this PR not touch those crypto helpers anymore if possible.

Mainly, I noticed the "add crypto helpers" need a lot of focus, to ensure that is actually correct. But every new push of this PR would require another "focused" look, as GitHub doesn't allow me to see what changed in that commit. So having that in a separate PR would help me, although that PR on its own wouldn't make any sense. Trade-offs :)

src/network/network_crypto.h
rubidium42: Currently it is, indeed, really specific for awaiting a password. But if at some point someone wants to introduce 2FA, where you also have to some on-time-token, then it wouldn't be a password any more but the further handling by the network code would be the same.

With `AWAIT_USER_INPUT` the network handling code at the client is essentially done, and at some point there will be a callback that triggers sending the response. For this the network handling code already passed a callback structure when preparing the authentication, so that arranges everything that's needed.

With `IMMEDIATE_RESPONSE` there is no need to ask the user for something, and it can immediately trigger sending that response.

src/network/network_crypto.h
rubidium42: It is indeed to have something different that `AWAIT`, but just like you I having problems giving it a sensible name. `READY` implies that all work is already done, while it is to instruct that the network code should just send the response instead of doing it via the callback.

"There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors."

TrueBrain: `READY_FOR_RESPONSE`?

I think the "immediate" is mostly what is a bit unclear to me. Like: ANSWER NOW FAST FAST, I got that impression. But it is more meant as: either we are waiting for more user input, or we are ready to go to the next step, right?

And yes, naming is hard :)

src/network/network_server.cpp
```diff
@@ -938,24 +961,40 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::Receive_CLIENT_JOIN(Packet &p)
...
+		case NetworkAuthenticationServerHandler::AUTHENTICATED:
+			break;
 
-	/* Check game password. Allow joining if we cleared the password meanwhile */
-	if (!_settings_client.network.server_password.empty() &&
-			_settings_client.network.server_password.compare(password) != 0) {
-		/* Password is invalid */
-		return this->SendError(NETWORK_ERROR_WRONG_PASSWORD);
+		case NetworkAuthenticationServerHandler::RETRY:
+			return this->SendAuthRequest();
```
TrueBrain: Not sure if `NetworkAuthenticationServerHandler` does this, but should there be a limit on the amount of retries?

rubidium42: Each attempt the failed authentication method is popped from the list at the server. So the number of "retries" is related to the number of configured authentication methods minus one.

Someone could implement a password retry mechanism by just adding the password handler twice.

TrueBrain: To avoid the next person wondering the same, should we call it something else ... `RETRY_NEXT_HANDLER`? `RETRY_NEXT_MECHANISM`? Something like that?

`RETRY_NEXT_METHOD`?

src/network/core/packet.cpp
rubidium42: I added something more explicit for packet size and type.

src/network/core/packet.h
```diff
@@ -40,6 +40,8 @@ typedef uint8_t  PacketType; ///< Identifier for the packet
  *     (year % 4 == 0) and ((year % 100 != 0) or (year % 400 == 0))
  */
 struct Packet {
+	static constexpr size_t EncodedLengthOfPacketSize() { return sizeof(PacketSize); }
+	static constexpr size_t EncodedLengthOfPacketType() { return sizeof(PacketType); }
```
TrueBrain, line 44: Excellent :D Didn't know this many places were using `sizeof()` already. This is much more clear what is going on :)

TrueBrain (dismissed):

TrueBrain (approved):

PR #12303, opened 2024-03-15
Fix #12301: Update ship current order destination tile when relocating buoys

### LINKED ISSUE #12301: [Bug]: Ship becomes lost after moving buoy
### Version of OpenTTD

Openttd 20240315-master-ga7625b8ae0

### Expected result

Either `v->dest_tile` is updated upon relocation, or some other more elaborate method needs to be used to reflect the actual location of the buoy.

### Actual result

This bug is deeper and has nothing to do with the pathfinder. It's about buoy location. The buoy is removed and moved to another tile, this doesn't update the `v->dest_tile` of vehicles.
What happened:
The previous location of the buoy was valid, but then demolished, and rebuilt to a near location, reusing the same `WaypointID` value, but with a different location. The old place was then raised terrain and became waterless.
The pathfinder is given the old location, and the high lvl pathfinder finds a patch with no water, hence the label 0.
Low level pathfinder doesn't even get the chance, it fails right away in high level pathfinder with a size of 0
would probably be given the same erroneous location, since it's also `m_destTile = v->dest_tile;`

### Steps to reproduce

1. Build a buoy.
2. Build a ship.
3. Add order to go to buoy.
4. Start ship.
5. Demolish buoy.
6. Rebuild buoy near the old location, reusing the same WaypointID.
7. Raise terrain on the old location so that it becomes waterless.

Ship is lost!

[ship 2 is lost because buoy was relocated.zip](https://github.com/OpenTTD/OpenTTD/files/14616882/ship.2.is.lost.because.buoy.was.relocated.zip)

SamuXarick:
## Motivation / Problem
#12301 
Relocating a buoy to another tile isn't updating `v->dest_tile`, sending the wrong tile for the pathfinders.

## Description
Fix #12301 Update `v->dest_tile` when relocating buoys.

## Checklist for review

### DIFF
--- a/src/waypoint_cmd.cpp
+++ b/src/waypoint_cmd.cpp
@@ -22,6 +22,7 @@
 #include "window_func.h"
 #include "timer/timer_game_calendar.h"
 #include "vehicle_func.h"
+#include "ship.h"
 #include "string_func.h"
 #include "company_func.h"
 #include "newgrf_station.h"
@@ -487,8 +488,14 @@ CommandCost CmdBuildBuoy(DoCommandFlag flags, TileIndex tile)
 		if (wp == nullptr) {
 			wp = new Waypoint(tile);
 		} else {
-			/* Move existing (recently deleted) buoy to the new location */
-			wp->xy = tile;
+			if (wp->xy != tile) {
+				/* Move existing (recently deleted) buoy to the new location. */
+				wp->xy = tile;
+				for (Ship *v : Ship::Iterate()) {
+					if (!v->current_order.IsType(OT_GOTO_WAYPOINT) || v->current_order.GetDestination() != wp->index) continue;
+					v->SetDestTile(wp->xy);
+				}
+			}
 			InvalidateWindowData(WC_WAYPOINT_VIEW, wp->index);
 		}
 		wp->rect.BeforeAddTile(tile, StationRect::ADD_TRY);
--- a/src/order_cmd.cpp
+++ b/src/order_cmd.cpp
@@ -2189,9 +2189,23 @@ bool ProcessOrders(Vehicle *v)
 	}
 
 	/* If it is unchanged, keep it. */
-	if (order->Equals(v->current_order) && (v->type == VEH_AIRCRAFT || v->dest_tile != 0) &&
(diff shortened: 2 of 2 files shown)


SamuXarick:
I think other order types that aren't these 3: OT_GOTO_STATION, OT_GOTO_DEPOT, OT_GOTO_WAYPOINT, are being improperly handed.

PeterN:
I would expect something to hook into `CmdBuildBuoy`, around `waypoint_cmd.cpp:331`. i.e. actually update a ship's destination if it needs to be changed.

Kuhnovic:
> I would expect something to hook into `CmdBuildBuoy`, around `waypoint_cmd.cpp:331`. i.e. actually update a ship's destination if it needs to be changed.
> 
> Might need cached paths to be invalidated too?

That feels the wrong way around to me. The buoy shouldn't be concerned with whoever is using it as a destination. The ship is heading towards the buoy, so it's the ship's job to find a path to it and keep track of the goal. 

I find the usage of v->dest_tile a bit strange tbh. Destinations tend to be higher level items such as depots, stations or waypoints. These can then be resolved to a destination tile (or multiple tiles) when a PF call is done. If the position of the destination changes then the next PF run will automatically deal with this.

src/waypoint_cmd.cpp
```diff
@@ -327,8 +328,15 @@ CommandCost CmdBuildBuoy(DoCommandFlag flags, TileIndex tile)
...
-			/* Move existing (recently deleted) buoy to the new location */
-			wp->xy = tile;
+			if (wp->xy != tile) {
+				/* Move existing (recently deleted) buoy to the new location. */
+				wp->xy = tile;
+				for (Ship *v : Ship::Iterate()) {
+					if (v->current_order.IsType(OT_GOTO_WAYPOINT)) {
+						v->SetDestTile(wp->xy);
+					}
+				}
```
rubidium42: Please test this PR with ships going to different buoys, and then changing the location of one of the buoys.
As far as I am aware, this is not going to do what you intend it to do in that situation.

SamuXarick: oh, I see

SamuXarick: Solved. Also reduced identation while at it.

Kuhnovic:
Closing this because the solution is backwards, as I decribed in more detail in my earlier comment. I would love to see a proper solution, but that would deserve its own PR.

PR #12304, opened 2024-03-15
Add: track savegame size to report with survey

TrueBrain:
## Motivation / Problem

One of the debates we have from time to time, how big savegames actually are. And the survey answers a lot of questions, but not this specific one.

## Description

Track the size of the last loaded / saved savegame, to give an idea how big savegames are.
For this I use the actual bytes on disk; not the uncompressed format. Although we could also track that, it is far less relevant.

While at it, bundled together all this session information, into a new section called "session". This also updates the survey schema to version 2.

## Checklist for review

### DIFF
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -46,8 +46,8 @@ bool _screen_disable_anim = false;   ///< Disable palette animation (important f
 std::atomic<bool> _exit_game;
 GameMode _game_mode;
 SwitchMode _switch_mode;  ///< The next mainloop command.
-std::chrono::steady_clock::time_point _switch_mode_time; ///< The time when the switch mode was requested.
 PauseMode _pause_mode;
+GameSessionStats _game_session_stats; ///< Statistics about the current session.
 
 static byte _stringwidth_table[FS_END][224]; ///< Cache containing width of often used characters. @see GetCharacterWidth()
 DrawPixelInfo *_cur_dpi;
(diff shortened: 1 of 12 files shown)


src/gfx.cpp
```diff
@@ -46,8 +46,8 @@ bool _screen_disable_anim = false;   ///< Disable palette animation (important f
 std::atomic<bool> _exit_game;
 GameMode _game_mode;
 SwitchMode _switch_mode;  ///< The next mainloop command.
-std::chrono::steady_clock::time_point _switch_mode_time; ///< The time when the switch mode was requested.
 PauseMode _pause_mode;
+GameSessionStats _game_session_stats; ///< Statistics about the current session;
```
2TallTyler:
```suggestion
GameSessionStats _game_session_stats; ///< Statistics about the current session.
```

src/saveload/saveload.cpp
```diff
@@ -2184,14 +2185,18 @@ struct FileReader : LoadFilter {
...
 	}
 
 	size_t Read(byte *buf, size_t size) override
 	{
 		/* We're in the process of shutting down, i.e. in "failure" mode. */
 		if (this->file == nullptr) return 0;
 
-		return fread(buf, 1, size, this->file);
+		size_t res = fread(buf, 1, size, this->file);
+		read_bytes += res;
```
rubidium42: During reading the file we seek through it (see `Reset()`), so you might be counting bytes double. I am not sure whether the amount of bytes counted double is significant enough to care though.
Also, we have a `begin`, so maybe we can just `_game_session_stats.savegame_size = ftell(this->file) - this->begin` with the proper guards for `file` being `nullptr`.

TrueBrain: Good catch! And `ftell` for both looks nicer anyway :D

rubidium42 (approved):

PR #12306, opened 2024-03-16
Codechange: use span to send bytes to Packet and add span recv function

rubidium42:
## Motivation / Problem

With #12300 (encrypted game sockets) a lot more arrays of just bytes have to be written to and read from packets. To prevent having to use `array.data()` and `array.data() + array.size()`, it would be better to transform that `Packet` API to just use `std::span` instead.

## Description

Use `std::span` for `Send_bytes`, returning the remaining to be sent span.
Introduce `Recv_bytes`, which allows writing bytes into a `std::span`.

## Limitations

`Recv_bytes` is technically dead code until #12300 gets merged.

## Checklist for review

### DIFF
--- a/src/network/core/packet.cpp
+++ b/src/network/core/packet.cpp
@@ -176,16 +176,15 @@ void Packet::Send_buffer(const std::vector<byte> &data)
 /**
  * Send as many of the bytes as possible in the packet. This can mean
  * that it is possible that not all bytes are sent. To cope with this
- * the function returns the amount of bytes that were actually sent.
- * @param begin The begin of the buffer to send.
- * @param end   The end of the buffer to send.
- * @return The number of bytes that were added to this packet.
+ * the function returns the span of bytes that were not sent.
+ * @param span The span describing the range of bytes to send.
+ * @return The span of bytes that were not written.
  */
-size_t Packet::Send_bytes(const byte *begin, const byte *end)
+std::span<const byte> Packet::Send_bytes(const std::span<const byte> span)
 {
-	size_t amount = std::min<size_t>(end - begin, this->limit - this->Size());
-	this->buffer.insert(this->buffer.end(), begin, begin + amount);
-	return amount;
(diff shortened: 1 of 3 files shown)


TrueBrain (approved):

src/network/core/packet.cpp
```diff
@@ -370,6 +369,22 @@ std::vector<byte> Packet::Recv_buffer()
 	return data;
 }
 
+/**
+ * Extract at most as many bytes from the packet into the span.
```
TrueBrain: nitpick: this is a hard to read sentence. 

```suggestion
 * Extract at most the length of the span bytes from the packet into the span.
```

?

TrueBrain (changes requested):

TrueBrain (dismissed):

src/network/core/packet.cpp
```diff
@@ -370,6 +369,22 @@ std::vector<byte> Packet::Recv_buffer()
...
 }
 
+/**
+ * Extract at most the length of the span bytes from the packet into the span.
+ * @param span The span to write the bytes to.
+ * @return The number of bytes that were actually read.
+ */
+size_t Packet::Recv_bytes(std::span<byte> span)
+{
+	auto tranfer_to_span = [](std::span<uint8_t> destination, const char *source, size_t amount) {
```
michicc: Is the `uint8_t` versus `byte` in the function parameter intended?

rubidium42: Not really.

TrueBrain: Didn't we want to go for uint8_t instead of byte?

rubidium42: The current way keeps `Packet` consistent with respect to the 'byte' type.

It would be weird if the types of Send and Recv would differ, and I wanted to keep this as small as possible to prevent having to dig myself out of a rabbit hole once again.

Moving to `uint8_t` would be something for another PR.

TrueBrain (approved):

PR #12317, opened 2024-03-16
Codechange: Check cheap station-facility-filter before expensive in-use filter.

PeterN:
## Motivation / Problem

`BuildStationList` has performs a cheap bitmask filter for station facility after an expensive station-in-use filter. 

## Description

Switch the order of these tests around, so that the cheap station facility filter is performed first. This can dramatically improve performance when filtering the station list for station types that can't have shared-owner stations, e.g. rail, bus or truck, as the station-in-use filter does not need to be called at all.

On Wentbourne when filtering for these facilities, the list generation time drops from 50ms with master to under 1ms.

## Checklist for review

### DIFF
--- a/src/station_gui.cpp
+++ b/src/station_gui.cpp
@@ -260,8 +260,8 @@ protected:
 		this->stations_per_cargo_type_no_rating = 0;
 
 		for (const Station *st : Station::Iterate()) {
-			if (st->owner == owner || (st->owner == OWNER_NONE && HasStationInUse(st->index, true, owner))) {
-				if (this->filter.facilities & st->facilities) { // only stations with selected facilities
+			if ((this->filter.facilities & st->facilities) != 0) { // only stations with selected facilities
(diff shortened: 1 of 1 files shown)


src/station_gui.cpp
```diff
@@ -260,8 +260,8 @@ class CompanyStationsWindow : public Window
 		this->stations_per_cargo_type_no_rating = 0;
 
 		for (const Station *st : Station::Iterate()) {
-			if (st->owner == owner || (st->owner == OWNER_NONE && HasStationInUse(st->index, true, owner))) {
-				if (this->filter.facilities & st->facilities) { // only stations with selected facilities
+			if ((this->filter.facilities & st->facilities) != 0) { // only stations with selected facilities
```
TrueBrain, line 263: while at it, mind making this a bit more readable by using `continue` instead of this deep nesting?

TrueBrain (approved):

src/station_gui.cpp
PeterN: Will do as a separate PR I think, to avoid obscuring this simple change.

PR #12320, opened 2024-03-16
Codechange: add set of classes providing authentication and encryption

rubidium42:
## Motivation / Problem

We need a base of crypto principles/helpers for adding:
* logging in without actually sending the password (password-authenticated key exchange)
* logging in without password (authorized keys)
* encrypting the communication

## Description

This adds all the cryptography functions that are required in other places in the code, for the previously mentioned functionality to work.

This has been split off from #12300 in an effort to have a single PR with all the cryptography stuff, and the things that change the network protocol in others. Not all cryptography is in this PR, as the unit tests for encryption/decryption of socket handlers is not included in here. That is due to those requiring the socket handlers to be changed.

When reviewing, start at the top (`@file`) comment of network_crypto.h for a quick overview of how it is supposed to work. For an overview of the X25519 key exchange procedure, take a look at the comment for `X25519AuthenticationHandler` in network_crypto_internal.h

## Limitations

This in itself adds no functionality, just dead code.

## Checklist for review

### DIFF
--- a/src/network/network_crypto.cpp
+++ b/src/network/network_crypto.cpp
@@ -1,0 +1,474 @@
+/*
+ * This file is part of OpenTTD.
+ * OpenTTD is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * OpenTTD is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with OpenTTD. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file network_crypto.cpp Implementation of the network specific cryptography helpers. */
+
+#include "../stdafx.h"
+
+#include "network_crypto_internal.h"
+#include "core/packet.h"
+
+#include "../3rdparty/monocypher/monocypher.h"
+#include "../core/random_func.hpp"
+#include "../debug.h"
+#include "../string_func.h"
+
+#include "../safeguards.h"
+
+/**
+ * Call \c crypto_wipe for all the data in the given span.
+ * @param span The span to cryptographically wipe.
+ */
+static void crypto_wipe(std::span<uint8_t> span)
+{
+	crypto_wipe(span.data(), span.size());
+}
+
+/** Ensure the derived keys do not get leaked when we're done with it. */
+X25519DerivedKeys::~X25519DerivedKeys()
+{
+	crypto_wipe(keys);
+}
+
+/**
+ * Get the key to encrypt or decrypt a message sent from the client to the server.
+ * @return The raw bytes of the key.
+ */
+std::span<const uint8_t> X25519DerivedKeys::ClientToServer() const
+{
+	return std::span(this->keys.data(), X25519_KEY_SIZE);
+}
+
+/**
+ * Get the key to encrypt or decrypt a message sent from the server to the client.
+ * @return The raw bytes of the key.
+ */
+std::span<const uint8_t> X25519DerivedKeys::ServerToClient() const
+{
+	return std::span(this->keys.data() + X25519_KEY_SIZE, X25519_KEY_SIZE);
+}
+
+/**
+ * Perform the actual key exchange.
+ * @param peer_public_key The public key chosen by the other participant of the key exchange.
+ * @param side Whether we are the client or server; used to hash the public key of us and the peer in the right order.
+ * @param our_secret_key The secret key of us.
+ * @param our_public_key The public key of us.
+ * @param extra_payload Extra payload to put into the hash function to create the derived keys.
+ * @return True when the key exchange has succeeded, false when an illegal public key was given.
+ */
+bool X25519DerivedKeys::Exchange(const X25519PublicKey &peer_public_key, X25519KeyExchangeSide side,
+		const X25519SecretKey &our_secret_key, const X25519PublicKey &our_public_key, std::string_view extra_payload)
+{
+	X25519Key shared_secret;
+	crypto_x25519(shared_secret.data(), our_secret_key.data(), peer_public_key.data());
+	if (std::all_of(shared_secret.begin(), shared_secret.end(), [](auto v) { return v == 0; })) {
+		/* A shared secret of all zeros means that the peer tried to force the shared secret to a known constant. */
+		return false;
+	}
+
+	crypto_blake2b_ctx ctx;
+	crypto_blake2b_init(&ctx, this->keys.size());
+	crypto_blake2b_update(&ctx, shared_secret.data(), shared_secret.size());
+	switch (side) {
+		case X25519KeyExchangeSide::SERVER:
+			crypto_blake2b_update(&ctx, our_public_key.data(), our_public_key.size());
+			crypto_blake2b_update(&ctx, peer_public_key.data(), peer_public_key.size());
+			break;
+		case X25519KeyExchangeSide::CLIENT:
+			crypto_blake2b_update(&ctx, peer_public_key.data(), peer_public_key.size());
+			crypto_blake2b_update(&ctx, our_public_key.data(), our_public_key.size());
+			break;
+		default:
+			NOT_REACHED();
+	}
+	crypto_blake2b_update(&ctx, reinterpret_cast<const uint8_t *>(extra_payload.data()), extra_payload.size());
+	crypto_blake2b_final(&ctx, this->keys.data());
+	return true;
+}
+
+/**
+ * Encryption handler implementation for monocypther encryption after a X25519 key exchange.
+ */
+class X25519EncryptionHandler : public NetworkEncryptionHandler {
+private:
+	crypto_aead_ctx context; ///< The actual encryption context.
+
+public:
+	/**
+	 * Create the encryption handler.
+	 * @param key The key used for the encryption.
+	 * @param nonce The nonce used for the encryption.
+	 */
+	X25519EncryptionHandler(const std::span<const uint8_t> key, const X25519Nonce &nonce)
+	{
+		assert(key.size() == X25519_KEY_SIZE);
+		crypto_aead_init_x(&this->context, key.data(), nonce.data());
+	}
+
+	/** Ensure the encryption context is wiped! */
+	~X25519EncryptionHandler()
+	{
+		crypto_wipe(&this->context, sizeof(this->context));
+	}
+
+	size_t MACSize() const override
+	{
+		return X25519_MAC_SIZE;
+	}
+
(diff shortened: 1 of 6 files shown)


src/network/network_crypto.h
```diff
@@ -0,0 +1,282 @@
...
+class NetworkAuthenticationDefaultPasswordProvider : public NetworkAuthenticationPasswordProvider {
+private:
+	const std::string *password; ///< The password to check against.
+public:
+	/**
+	 * Create the provider with the pointer to the password that is to be used. A pointer, so this can handle
+	 * situations where the password gets changed over time.
+	 * @param password The reference to the configured password.
+	 */
+	NetworkAuthenticationDefaultPasswordProvider(const std::string &password) : password(&password) {}
```
TrueBrain, line 130: Just curious: so why not a `string_view`, as that is a "pointer" too?

TrueBrain:
This PR looks absolutely fine to me; I will hold off from an approval to one of the PRs that uses this is reviewed, just to limit the time between this being merged and that. But I really appreciate you making this another PR, makes it far easier to track what changes in this file :)

Also, maybe another developer wants to give it a look, as having this right is pretty valuable :D

src/network/network_crypto.h
rubidium42: Because the password can change during the lifetime of the provider, which is at least as long as the user gets to enter their password before getting booted. In practice these objects will just survive the whole life of the application.

Remember that a `std::string_view` is essentially a pointer to a begin and an end, and `std::string` is a length and a pointer to some allocated memory with the content of the string.

Imagine the password being 'a' and I change it to 'bb'. At best the view given to the provider contains 'b' as the number of characters does not change and you might be lucky that `std::string` did not reallocate. At worst you're reading a deallocated string and segfault.

TrueBrain: tldr; `string_view` is also a length, length of password can change. That is fair :)

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,451 @@
...
+ * Furthermore update the public key so it is always in sync with the private key.
+ * @param secret_key The secret key to read/validate/fix.
+ * @param public_key The public key to update.
+ * @return The valid secret key.
+ */
+/* static */ X25519SecretKey X25519AuthorizedKeyClientHandler::GetValidSecretKeyAndUpdatePublicKey(std::string &secret_key, std::string &public_key)
+{
+	X25519SecretKey key{};
+	if (!ConvertHexToBytes(secret_key, key)) {
+		Debug(net, 0, "[crypto] Found invalid secret key, creating a new random key");
```
TrueBrain:
```suggestion
		if (!secret_key.empty()) {
			Debug(net, 0, "[crypto] Found invalid secret key, creating a new random key");
		} else {
			Debug(net, 3, "[crypto] Creating a new random key");
		}
```

At least, I assume that when the game is first started, the `secret_key` is still empty, and this will populate it for the first time. Which is a good thing. But that shouldn't show the message, I think :)

src/network/network_crypto_internal.h
```diff
@@ -0,0 +1,317 @@
...
+private:
+	X25519SecretKey our_secret_key; ///< The secret key used by us.
+	X25519PublicKey our_public_key; ///< The public key used by us.
+	X25519Nonce nonce; ///< The nonce to prevent replay attacks.
+	X25519DerivedKeys derived_keys; ///< Keys derived from the authentication process.
+
+protected:
+	X25519PublicKey peer_public_key; ///< The public key used by our peer.
+
+	X25519AuthenticationHandler(const X25519SecretKey &secret_key = X25519SecretKey::CreateRandom());
```
TrueBrain: This line has been .. weirding me out. Mostly, I really did not know you could call a function as a default value like this in C++ :P But it also took me a bit of time to find this statement.

But the other thing that kinda bothers me about this line, is who is the owner of the memory?

`CreateRandom()` returns a `X25519SecretKey`. This ctor stores a pointer to that object. But .. who than owns the object? How is the memory still alive? When will it be released?

If I would be to guess, the reason this works is because `X25519SecretKey` is in fact a `std::array`. But it raises a lot of eyebrows.

Shouldn't we just move this out, and not have a default constructor here? I am aware that this means you need to store the `secret_key` in the client-handler on the server-side. But at least that makes more clear who owns the memory?

It might also be that I am just missing something here :)

src/network/network_crypto_internal.h
```diff
@@ -0,0 +1,317 @@
...
+ *    The first of the pair of derived keys is usually used to encrypt client-to-server communcation, and the second of the pair
+ *    is usually used to encrypt server-to-client communication.
+ * 4) a XChaCha20-Poly1305 (authenticated) encryption is performed using:
+ *    - the first of the pair of derived keys as encryption key;
+ *    - a 24 byte nonce;
+ *    - the public key of the client as additional authenticated data.
+ *    - a 8 byte random number as content/message.
+ *
+ * The server initiates the request by sending its public key and a 24 byte nonce that is randomly generated. Normally the side
+ * that sends the encrypted data sends the nonce in their packet, which would the the client on our case. However, there are
```
TrueBrain:
```suggestion
 * that sends the encrypted data sends the nonce in their packet, which would be the client on our case. However, there are
```

src/network/network_crypto_internal.h
```diff
@@ -0,0 +1,317 @@
...
+ * many implementations of clients due to the admin-protocol where this is used, and we cannot guarantee that they generate a
+ * good enough nonce. As such the server sends one instead. The server will create a new set of keys for each session.
+ *
+ * The client receives the request, performs the key exchange, generates the derived keys and then encrypts the message. This
+ * message must contain some content, so it has to be filled with a four random bytes. Once the message has been encrypted, the
```
TrueBrain:
```suggestion
 * message must contain some content, so it has to be filled with a 8 random bytes. Once the message has been encrypted, the
```

At least, you say 8 a bit higher.

src/network/network_crypto_internal.h
rubidium42: Welcome to the magic world of C++ object lifetime.

What happens is that C++ makes a temporary object with the created secret key, then runs the constructor with that and afterwards C++ cleans up that temporary object. Since it's an object on the stack and not the heap, there is no memory management that needs to be done.

See https://godbolt.org/z/oEr3breKn for this magic in action with some output on the (de)constructors etc.

TrueBrain: The trick that happens here, is that although you receive a pointer, you make a copy of the object directly after. So indeed, a lifetime of just the ctor call is sufficient. Still very scary code.

Either way, I am still not sure if it is actually a good idea to have this here. This is really hidden, and really took me a while to find where the server generates his public/secret.

TrueBrain: Guess I would prefer the more explicit route of:

```
        if (this->authentication_handler == nullptr) {
-               this->authentication_handler = NetworkAuthenticationServerHandler::Create(&_password_provider, &_authorized_key_handler);
+               /* Create a new secret key for every connection. */
+               X25519SecretKey secret_key = X25519SecretKey::CreateRandom()
+               this->authentication_handler = NetworkAuthenticationServerHandler::Create(secret_key&, &_password_provider, &_authorized_key_handler);
        }
```

That way it is also documented what the behaviour is, and that it is expected.

rubidium42: The biggest problem I have with that proposal is having to expose the internal crypto guts in the API. 
So basically anything that includes `network/core/core.h` (indirectly) will also get the internal crypto guts included, because the internal details of some secret key needed to be exposed.

If you want to avoid that, you'd need to pass a `std::string` there and if it's empty or invalid regenerate a secret key.

So, could we just add comment to `NetworkAuthenticationServerHandler::Create` that a new public/secret key pair is generated for each authentication method?
By the way, for the client only the authorized key one uses the provided public/secret keypair. The others also generate a new public/secret pair for each method.

src/network/network_crypto.h
```diff
@@ -0,0 +1,282 @@
...
+ * the appropriate \c NetworkAuthenticationServerHandler. This will be used to create a request
+ * for the client, which instantiates a \c NetworkAuthenticationClientHandler to handle that
+ * request.
+ * At the moment there are three types of request: key exchange only, password-authenticated key
+ * exchange (PAKE) and authorized keys. When the request is for a password, the user is asked
+ * for the password via an essentially asynchronous callback from the client handler. For the
+ * other requests no input from the user is needed, and these are immediately ready to generate
+ * the response for the server.
+ *
+ * The server will validate the response resuling in either the user being authenticated or not.
```
TrueBrain:
```suggestion
 * The server will validate the response resulting in either the user being authenticated or not.
```

src/network/network_crypto_internal.h
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * In general this works as follows:
+ * 1) the client and server have or generate a secret and public X25519 key.
+ * 2) the X25519 key exchange is performed at both the client and server, with their own secret key and their peer's public key.
+ * 3) a pair of derived keys is created by BLAKE2b-hashing the following into 64 bytes, in this particular order:
+ *    - the shared secret from the key exchange;
+ *    - the public key of the server;
+ *    - the public key of the client;
+ *    - optional extra payload, e.g. a password in the case of PAKE.
+ *    The first of the pair of derived keys is usually used to encrypt client-to-server communcation, and the second of the pair
```
TrueBrain:
```suggestion
 *    The first of the pair of derived keys is usually used to encrypt client-to-server communication, and the second of the pair
```

src/network/network_crypto.h
```diff
@@ -0,0 +1,282 @@
...
+};
+
+/**
+ * Base class for server side cryptographic authentication handlers.
+ */
+class NetworkAuthenticationServerHandler : public NetworkAuthenticationHandler {
+public:
+	/** The processing result of receiving a response. */
+	enum ResponseResult {
+		AUTHENTICATED, ///< The client was authenticated succesfully.
```
TrueBrain:
```suggestion
		AUTHENTICATED, ///< The client was authenticated successfully.
```

src/network/network_crypto.cpp
```diff
@@ -0,0 +1,455 @@
...
+ */
+bool X25519AuthenticationHandler::SendResponse(Packet &p, std::string_view derived_key_extra_payload)
+{
+	if (!this->derived_keys.Exchange(this->peer_public_key, X25519KeyExchangeSide::CLIENT,
+			this->our_secret_key, this->our_public_key, derived_key_extra_payload)) {
+		Debug(net, 0, "[crypto] Peer sent an illegal public key; authentication aborted.");
+		return false;
+	}
+
+	X25519KeyExchangeMessage message{};
```
TrueBrain: The docs say it encrypts a few random bytes, but these are just zeros?

rubidium42: Now it's filling it with random bytes. It must be something that did not survive the countless refactors that I've done over the time.

TrueBrain (approved):

PR #12322, opened 2024-03-16
Fix #12319, 3a676a5: Some SSE blitters were broken due to ODR violations

### LINKED ISSUE #12319: [Bug]: 32bpp-sse2 blitter is broken in debug build
### Version of OpenTTD

master(0cff40dd19a52ebd1b2435991dab9ebfdacbcec1), Ubuntu 22.04.4 LTS
g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Only affects debug build

### Expected result

Correct rendering.

### Actual result

Screenshot from 2024-03-17 02-49-55

### Steps to reproduce

Run `./openttd -b 32bpp-sse2`

frosch123:
## Motivation / Problem

#12319

## Description

``32bpp_sse_type.h`` and ``32bpp_sse_func.hpp`` are no proper "header" files.
They are "inline" files, which are included by different translation units with different defines for ``SSE_VERSION`` and ``MARGIN_NORMAL_THRESHOLD``.

As such everything has to use internal linkage.

## Limitations

I looked into replacing the defines with template parameters.
Unfortunately ``MARGIN_NORMAL_THRESHOLD`` is redefined to different values, thus this code is very fragile and easy to break.
Readding the ``static`` makes this back-portable.

## Checklist for review

### DIFF
--- a/src/blitter/32bpp_sse_func.hpp
+++ b/src/blitter/32bpp_sse_func.hpp
@@ -15,11 +15,12 @@
  * Be careful when declaring things with external linkage.
  * Use internal linkage instead, i.e. "static".
  */
+#define INTERNAL_LINKAGE static
 
 #ifdef WITH_SSE
 
 GNU_TARGET(SSE_TARGET)
-static inline void InsertFirstUint32(const uint32_t value, __m128i &into)
+INTERNAL_LINKAGE inline void InsertFirstUint32(const uint32_t value, __m128i &into)
 {
 #if (SSE_VERSION >= 4)
 	into = _mm_insert_epi32(into, value, 0);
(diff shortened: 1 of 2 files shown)


src/blitter/32bpp_sse_func.hpp
```diff
@@ -10,10 +10,16 @@
 #ifndef BLITTER_32BPP_SSE_FUNC_HPP
 #define BLITTER_32BPP_SSE_FUNC_HPP
 
+/* ATTENTION
+ * This file is compiled multiple times with different defines for SSE_VERSION and MARGIN_NORMAL_THRESHOLD.
+ * Be careful when declaring things with external linkage.
+ * Use internal linkage instead, i.e. "static".
+ */
+
```
rubidium42, line 19: This is too far away to be in the diff when someone would remove the static. In the commit that introduced the bug I didn't look beyond the diff, so I would not have noticed this message and neither would the reviewer.

frosch123: What's your suggestion? Duplicate in front of every single function?

frosch123: I can also rename the file to ``.inl`` or something.

rubidium42: Maybe something short like `/* Do not remove 'static', see attention above. */` just before each of the `static`s?

frosch123: Now I used another define to obfuscate ``static``, so search/replace won't find it.

rubidium42 (approved):

PR #12323, opened 2024-03-16
Feature: authenticate to the server without sending the password

rubidium42:
## Motivation / Problem

Sending plain unencrypted passwords over the network is bad.

## Description

Uses the crypto methods from #12320 to perform authentication using either password-authenticated key exchange or by authorized keys.

The server login has been moved from after the NewGRF checks to before them, so the order of operations is: client join/version check, server login, NewGRF checks, company login, downloading map, etc.

This is built upon #12320 and is a building stone for #12300.

## Limitations

Only applies to logging into a game. Not for companies, rcon or admin yet. Those will be future PRs.
No console logic has been introduced to add/manage the authorized keys, so they can only be set upon configuration and cannot  be changed in-game (yet). Neither has GUI logic been implemented.

## Checklist for review


src/network/core/config.h
```diff
@@ -91,4 +91,6 @@ static const uint NETWORK_GRF_NAME_LENGTH           =   80;           ///< Maxim
  */
 static const uint NETWORK_MAX_GRF_COUNT             =   255;
 
+static const uint NETWORK_SECRET_KEY_LENGTH = 32 * 2 + 1; ///< The maximum length of the hexadecimal encoded secret keys, in bytes including '\0'.
```
TrueBrain: Should this be based on a value from `network_crypto.h`? On one hand, it is nice that this doesn't need any other header file, on the other hand, they do strongly correlate. Maybe add a reference to it?

src/settings_type.h
```diff
@@ -313,9 +313,12 @@ struct NetworkSettings {
 	std::string server_invite_code_secret;                ///< Secret to proof we got this invite code from the Game Coordinator.
 	std::string server_name;                              ///< name of the server
 	std::string server_password;                          ///< password for joining this server
+	std::vector<std::string> server_authorized_keys; ///< public keys of clients that are authorized to connect to the game
```
TrueBrain: nitpick: please start sentences with capital and end with dot. I know I know, most don't have that. But new entries we have been changing :)

src/table/settings/network_secrets_settings.ini
```diff
@@ -61,6 +61,21 @@ flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_NETWORK_ONLY
...
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
+def      = nullptr
+
+[SDTC_SSTR]
+var      = network.client_public_key
+type     = SLE_STR
+length   = NETWORK_SECRET_KEY_LENGTH
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
+def      = nullptr
+pre_cb   = [](auto) { return false; }
```
TrueBrain, line 80: What is the effect of this `pre_cb`?

src/network/network_client.cpp
```diff
@@ -377,13 +377,14 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::SendNewGRFsOk()
  * Set the game password as requested.
  * @param password The game password.
  */
-NetworkRecvStatus ClientNetworkGameSocketHandler::SendGamePassword(const std::string &password)
+NetworkRecvStatus ClientNetworkGameSocketHandler::SendAuthResponse()
 {
-	Debug(net, 9, "Client::SendGamePassword()");
+	Debug(net, 9, "Client::PACKET_CLIENT_AUTH_RESPONSE()");
```
TrueBrain:
```suggestion
	Debug(net, 9, "Client::SendAuthResponse()");
```

TrueBrain (commented):
Minor stuff; overall it looks really good. Now to test it for myself :)

src/network/network_server.cpp
```diff
@@ -953,24 +965,38 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::Receive_CLIENT_IDENTIFY(Packet
...
-	if (!_settings_client.network.server_password.empty() &&
-			_settings_client.network.server_password.compare(password) != 0) {
-		/* Password is invalid */
-		return this->SendError(NETWORK_ERROR_WRONG_PASSWORD);
+		case NetworkAuthenticationServerHandler::RETRY_NEXT_METHOD:
+			return this->SendAuthRequest();
+
+		case NetworkAuthenticationServerHandler::NOT_AUTHENTICATED:
+		default:
+			return this->SendError(NETWORK_ERROR_WRONG_PASSWORD);
```
TrueBrain: Did some testing, and the one thing that stood out, was this error.

When I set a password on a server, and I enter the wrong one, this is as expected.
When I set a password on a server and add a authorized-keys, and I have the wrong client-public-key and I enter the wrong password, this is as expected.

But, when I only have authorized-keys and set no password, and I have the wrong client-public-key, I also get "wrong password". But I never got the chance to enter a password. So that feels confusing.

Could we add a new error here, something like `NETWORK_ERROR_NO_AUTHORIZED_KEYS` or `NETWORK_ERROR_NOT_ON_ALLOWLIST`, with a string representation on the client for that? So it is a bit more clear that there never was an opportunity to enter a password.

src/settings.cpp
```diff
@@ -1285,6 +1286,7 @@ static void HandleSettingDescs(IniFile &generic_ini, IniFile &private_ini, IniFi
 		proc_list(private_ini, "server_bind_addresses", _network_bind_list);
 		proc_list(private_ini, "servers", _network_host_list);
 		proc_list(private_ini, "bans", _network_ban_list);
+		proc_list(secrets_ini, "server_authorized_keys", _settings_client.network.server_authorized_keys);
```
TrueBrain: Should we put this in `private`, and not in `secret`? It doesn't actually contain secrets. And it is pretty similar to "bans" in that regard.

Bit bikeshedding, I am aware.

src/table/settings/network_secrets_settings.ini
rubidium42: It prevents that you change your public key from the console, as doing that is essentially pointless and might cause further harm down the line when the client and server are in disagreement about the public key of the client.

TrueBrain: Maybe add that as a comment here? As I am sure the next person looking at it would have the same question :)

src/network/core/config.h
rubidium42: They do. I added a comment, and added a `static_assert` in the test case.

src/table/settings/network_secrets_settings.ini
```diff
@@ -61,6 +61,24 @@ flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC | SF_NETWORK_ONLY
 def      = nullptr
 cat      = SC_BASIC
 
+[SDTC_SSTR]
+var      = network.client_secret_key
+type     = SLE_STR
+length   = NETWORK_SECRET_KEY_LENGTH
+flags    = SF_NOT_IN_SAVE | SF_NO_NETWORK_SYNC
+def      = nullptr
+; Prevent the user from setting the public key from the console using 'setting'
```
TrueBrain:
```suggestion
; Prevent the user from setting the secret key from the console using 'setting'
```

TrueBrain (dismissed):

TrueBrain (approved):

PR #12325, opened 2024-03-17
Fix #12302: Allow empty train engines to use an invalid cargo type.

### LINKED ISSUE #12302: [Bug]: #11719 changes the availablity of some vehicles for some GRFs
### Version of OpenTTD

Since #11719 / 2fd9096

### Expected result

Availability of NewGRF vehicles is the same as before

### Actual result

In DACH 0.3.2 (on Bananas), some locomotives which do not carry cargo are now marked as unavailable via the `ei->climates = 0` path in `CalculateRefitMasks`.
It appears that this is due to the vehicle definition not explicitly overriding the cargo type for locomotives which do not carry cargo. (I cannot comment on whether this makes the GRF correct or incorrect).
An example of this is "DB BR 143", engine ID 40.

### Steps to reproduce

See available vehicles for DACH 0.3.2 from bananas before and after #11719.

PeterN:
## Motivation / Problem

As per #12302.

Train engines are able to carry no cargo, and it did not used to matter what their default cargo type is.

Since #11719 it does now matter. If a NewGRF overrides an engine with a cargo type that is not valid for the current climate, and sets the capacity to zero without setting the cargo type to something valid (even though it does not carry cargo), the engine will be disabled.

## Description

Instead, if the cargo type is invalid, check if the cargo capacity is zero and that the cargo type has not been changed from its default value.

The cargo type will be forced to the first available type (usually passengers) instead of the engine being disabled.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -9161,6 +9161,14 @@ static void CalculateRefitMasks()
 				ei->cargo_type = (CargoID)FindFirstBit(ei->refit_mask);
 			}
 		}
+		if (!IsValidCargoID(ei->cargo_type) && e->type == VEH_TRAIN && e->u.rail.railveh_type != RAILVEH_WAGON && e->u.rail.capacity == 0) {
+			/* For train engines which do not carry cargo it does not matter if their cargo type is invalid.
+			 * Fallback to the first available instead, if the cargo type has not been changed (as indicated by
+			 * cargo_label not being CT_INVALID). */
+			if (GetActiveCargoLabel(ei->cargo_label) != CT_INVALID) {
+				ei->cargo_type = static_cast<CargoID>(FindFirstBit(_standard_cargo_mask));
+			}
+		}
 		if (!IsValidCargoID(ei->cargo_type)) ei->climates = 0;
 
 		/* Clear refit_mask for not refittable ships */


src/newgrf.cpp
```diff
@@ -9161,6 +9161,13 @@ static void CalculateRefitMasks()
 				ei->cargo_type = (CargoID)FindFirstBit(ei->refit_mask);
 			}
 		}
+		if (!IsValidCargoID(ei->cargo_type) && e->type == VEH_TRAIN && e->u.rail.railveh_type != RAILVEH_WAGON && e->u.rail.capacity == 0) {
+			/* For train engines which do not carry cargo it does not matter if their cargo type is invalid, fallback to
+			 * the first available instead, if the cargo type has not been changed. */
+			if (GetActiveCargoLabel(ei->cargo_label) != CT_INVALID) {
```
TrueBrain, line 9168: Is this check still needed? You already concluded the cargoID is not valid. Would this ever return anything other than `CT_INVALID`?

src/newgrf.cpp
```diff
@@ -9161,6 +9161,13 @@ static void CalculateRefitMasks()
 				ei->cargo_type = (CargoID)FindFirstBit(ei->refit_mask);
 			}
 		}
+		if (!IsValidCargoID(ei->cargo_type) && e->type == VEH_TRAIN && e->u.rail.railveh_type != RAILVEH_WAGON && e->u.rail.capacity == 0) {
+			/* For train engines which do not carry cargo it does not matter if their cargo type is invalid, fallback to
+			 * the first available instead, if the cargo type has not been changed. */
```
TrueBrain: Ghehe, this is a very long sentence :)

```suggestion
			/* For train engines which do not carry cargo it does not matter if their cargo type is invalid.
			 * Fallback to the first available instead, if the cargo type has not been changed. */
```

src/newgrf.cpp
PeterN: This check is to ensure this change only applies if the default cargo type (stored as a label) has not been changed. 

This value is set to CT_INVALID only when cargo_type has been changed by NewGRF.

PeterN: I've updated the comment to document this slightly better.

TrueBrain (approved):

PR #12326, opened 2024-03-17
Feature: authorized key authentication for rcon

rubidium42:
## Motivation / Problem

Having to type the password each time you are using rcon will get cumbersome. Since we have authorized keys for the game login, which not add that to rcon as well?

## Description

Skim the peer/client's public key during the game authentication phase and store it in the server's socket handler.

When a rcon command is received, first check the public key against the authorized keys for rcon. If that fails, fall back to the normal password validation.

This means that the rcon password is not checked any more when using authorized keys, however you still need to enter something as I did not want to introduce a difficult dance of back-and-forth communication to later request the rcon password when the client's public key was not in the authorized key list. That something can be as simple as just `*`, e.g. `rcon * "say hi"`.

## Limitations

Built on top of #12300 and everything that's built upon, so it cannot be merged directly.

There is no in game (console) management of the authorized keys yet, so it authorized keys can only be set via the configuration file. This management should be a separate PR, which could work the same for authorized keys of game, rcon, companies and the admin port.

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -60,6 +60,7 @@ template SocketList TCPListenHandler<ServerNetworkGameSocketHandler, PACKET_SERV
 
 static NetworkAuthenticationDefaultPasswordProvider _password_provider(_settings_client.network.server_password); ///< Provides the password validation for the game's password.
 static NetworkAuthenticationDefaultAuthorizedKeyHandler _authorized_key_handler(_settings_client.network.server_authorized_keys); ///< Provides the authorized key handling for the game authentication.
+static NetworkAuthenticationDefaultAuthorizedKeyHandler _rcon_authorized_key_handler(_settings_client.network.rcon_authorized_keys); ///< Provides the authorized key validation for rcon.
 
 
 /** Writing a savegame directly to a number of packets. */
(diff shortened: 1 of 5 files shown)


src/network/network_server.cpp
```diff
@@ -989,6 +990,7 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::Receive_CLIENT_AUTH_RESPONSE(P
 	auto authentication_method = this->authentication_handler->GetAuthenticationMethod();
 	switch (this->authentication_handler->ReceiveResponse(p)) {
 		case NetworkAuthenticationServerHandler::AUTHENTICATED:
+			this->peer_public_key = this->authentication_handler->GetPeerPublicKey();
```
TrueBrain: nitpick: would it be cleaner to put this a bit lower next to `receive_encryption_handler` and friends?

The other cases of the case `return` anyway :)

src/settings.cpp
```diff
@@ -1287,6 +1288,7 @@ static void HandleSettingDescs(IniFile &generic_ini, IniFile &private_ini, IniFi
 		proc_list(private_ini, "servers", _network_host_list);
 		proc_list(private_ini, "bans", _network_ban_list);
 		proc_list(private_ini, "server_authorized_keys", _settings_client.network.server_authorized_keys);
+		proc_list(private_ini, "rcon_authorized_keys", _settings_client.network.rcon_authorized_keys);
```
TrueBrain, line 1291: Hmm .. with `rcon_authorized_keys`, I now wonder if we should also name it `client_authorized_keys`, instead of `server_`?

src/settings_type.h
```diff
@@ -315,6 +315,7 @@ struct NetworkSettings {
 	std::string server_password;                          ///< password for joining this server
 	std::vector<std::string> server_authorized_keys; ///< Public keys of clients that are authorized to connect to the game.
 	std::string rcon_password;                            ///< password for rconsole (server side)
+	std::vector<std::string> rcon_authorized_keys; ///< public keys of clients that are authorized to use the rconsole (server side)
```
TrueBrain: nitpick: capitals, dots, etc :)

src/settings.cpp
rubidium42: I've chosen to use `s/password/authorized_keys/` to determine the setting names and not put any more thought than that into it. It'll keep the naming at least consistent within the configuration.

TrueBrain: I guess. Naming is still hard :P

TrueBrain (approved):

PR #12329, opened 2024-03-17
Feature: console command to change authorized keys

rubidium42:
## Motivation / Problem

It would be nice if you were able to add/remove authorized keys while the game is running.

## Description

Add a command that allows you to list the authorized keys, as well as remove/add them for specific types of authorized keys. For adding/removing you may also use `client:id` as shorthand for getting the key of that client.

## Limitations

A lot of convenience things might still be missing.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -1956,6 +1956,101 @@ DEF_CONSOLE_CMD(ConCompanyPassword)
 	return true;
 }
 
+/** All the known authorized keys with their name. */
+static std::vector<std::pair<std::string_view, std::vector<std::string> *>> _console_cmd_authorized_keys{
+	{ "rcon", &_settings_client.network.rcon_authorized_keys },
+	{ "server", &_settings_client.network.server_authorized_keys },
+};
+
+/**
+ * Simple helper to find the location of the given authorized key in the authorized keys.
+ * @param authorized_keys The keys to look through.
+ * @param authorized_key The key to look for.
+ * @return The iterator to the location of the authorized key, or \c authorized_keys.end().
+ */
+static auto FindKey(std::vector<std::string> *authorized_keys, std::string_view authorized_key)
+{
+	return std::find_if(authorized_keys->begin(), authorized_keys->end(), [authorized_key](auto &value) { return StrEqualsIgnoreCase(value, authorized_key); });
+}
+
+DEF_CONSOLE_CMD(ConNetworkAuthorizedKey)
+{
+	if (argc <= 2) {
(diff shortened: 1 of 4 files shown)


src/console_cmds.cpp
```diff
@@ -1955,6 +1955,78 @@ DEF_CONSOLE_CMD(ConCompanyPassword)
...
+			if (iter != authorized_keys->end()) {
+				authorized_keys->erase(iter);
+				IConsolePrint(CC_INFO, "Removed {} from {}.", authorized_key, name);
+			} else {
+				IConsolePrint(CC_WARNING, "Not removed {} from {} as it does not exist.", authorized_key, name);
+			}
+			continue;
+		}
+
+		/* No valid command was given. */
```
TrueBrain: no console output about this fact?

src/console_cmds.cpp
```diff
@@ -1956,6 +1956,101 @@ DEF_CONSOLE_CMD(ConCompanyPassword)
...
+}
+
+DEF_CONSOLE_CMD(ConNetworkAuthorizedKey)
+{
+	if (argc <= 2) {
+		IConsolePrint(CC_HELP, "List and update authorized keys. Usage: 'authorized_key list [type]|add [type] [key]|remove [type] [key]'.");
+		IConsolePrint(CC_HELP, "  list: list all the authorized keys of the given type.");
+		IConsolePrint(CC_HELP, "  add: add the given key to the authorized keys of the given type.");
+		IConsolePrint(CC_HELP, "  remove: remove the given key from the authorized keys of the given type; use 'all' to remove all authorized keys.");
+		IConsolePrint(CC_HELP, "Instead of a key, use 'client:id' to add/remove the key of that given client.");
```
TrueBrain: Hmm .. this "client:id" isn't very clear, that the "client" is a fixed string, but "id" is dynamic.

```suggestion
		IConsolePrint(CC_HELP, "Instead of a key, use 'client:<id>' to add/remove the key of that given client.");
```

Or we can just check if the value is an integer (`strol` or something), and then assume it is a clientid? Dunno.

TrueBrain (approved):

src/console_cmds.cpp
rubidium42: `stoi` and friends will happily try. If the key does start with a digit, then `stoi` returns that digit or a bigger number when there are more digits.

Doesn't mean that checking for a valid number can't be done, but I suspect someone might be wanting to implement something like removing the key at index `<id>` from the list and that's primarily why I added 'client:' in front of it. It also makes it very clear that you mean the client's id.

PR #12330, opened 2024-03-18
Codechange: Use single list for hierarchical group lists.

PeterN:
## Motivation / Problem

Vehicle group lists are displayed on both the vehicle list and company colour schemes windows.

Both are implemented mostly the same way with separate code, using a GUIList for the groups, and a separate vector for the hierarchy indentations.

## Description

Replace both group list implementations (vehicle group list and company colour group list) with a single implementation, using a struct to hold the group and indentation level instead of two separate lists.

Parts that were previously duplicated are now shared.

The use of a single GUIList now allows the use of more iterator functions instead of indexes, and matches how the Engine List hierarchy is handled.

## Checklist for review

### DIFF
--- a/src/group_gui.cpp
+++ b/src/group_gui.cpp
@@ -27,6 +27,7 @@
 #include "company_gui.h"
 #include "gui.h"
 #include "group_cmd.h"
+#include "group_gui.h"
 #include "vehicle_cmd.h"
 #include "gfx_func.h"
 
@@ -35,8 +36,6 @@
 #include "table/sprites.h"
 
 #include "safeguards.h"
-
-typedef GUIList<const Group*> GUIGroupList;
 
 static constexpr NWidgetPart _nested_group_widgets[] = {
 	NWidget(NWID_HORIZONTAL), // Window header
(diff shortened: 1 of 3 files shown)


src/group_gui.cpp
```diff
@@ -111,6 +110,73 @@ static constexpr NWidgetPart _nested_group_widgets[] = {
...
+{
+	for (const auto &item : src) {
+		if (item.group->parent != parent) continue;
+
+		dst.emplace_back(item.group, indent);
+
+		if (fold && item.group->folded) {
+			/* Test if this group has children at all. If not, the folded flag should be cleared to avoid lingering unfold buttons in the list. */
+			GroupID groupid = item.group->index;
+			bool has_children = std::any_of(src.begin(), src.end(), [groupid](const auto &child){ return child.group->parent == groupid; });
```
rubidium42:
```suggestion
			bool has_children = std::any_of(src.begin(), src.end(), [groupid](const auto &child) { return child.group->parent == groupid; });
```

rubidium42 (dismissed):

rubidium42 (approved):

Issue #12333, opened 2024-03-18
[Bug]: Ship pathfinder lost entering dead end after terraforming


Kuhnovic:
### Version of OpenTTD

20240318-master-m107c208d87 on Win10

### Expected result

The ship should either:
- ideally avoid the dead end (but it's currently in the ship cache)
- turn around when it detects the dead end

### Actual result

![image](https://github.com/OpenTTD/OpenTTD/assets/68320206/f350b28b-d0cd-4fe7-977a-ee66016280b8)
The red area just got terraformed. The ship still has the white and yellow track dirs in its cache. It will start a new PF run when the ship is at the yellow trackdir. This PF run will fail because only the forward direction is considered.

### Steps to reproduce

1. Open ship_dead_end.sav
2. Wait for the ship to get lost

[ship_dead_end.zip](https://github.com/OpenTTD/OpenTTD/files/14642137/ship_dead_end.zip)

Kuhnovic:
![openttd_Ed0k1FuTMR](https://github.com/user-attachments/assets/c34d89d6-c10c-47c5-b892-af3fb14fb90c)

The ship happens to run out of cached path as it enters the river. A new PF call is done, and as of #13171 reversal is also included as an option. This turns out to be the best (and only) option, so the ship turns around.

Issue #12231, opened 2024-03-08
[Bug]: CheckShipReverse finding no path prevents ships from choosing one of the tracks available upon reversing


SamuXarick:
### Version of OpenTTD

OpenTTD 20240308-master-g216905ce1f

### Expected result

Ships can choose one of the tracks available upon reversing. They become unstuck, but still lost.

### Actual result

Ships can only do 180 degree turns when reversing. They're stuck and lost.

### Steps to reproduce

[ship stuck at 180 degrees.zip](https://github.com/OpenTTD/OpenTTD/files/14537668/ship.stuck.at.180.degrees.zip)
1. Load the attached savegame
2. Start all ships

PR #12335, opened 2024-03-18
Fix #12228, #12231, #12333: Ship pathfinder always considered reversing

### LINKED ISSUE #12228: [Bug]: CheckShipReverse's returned path restricts ship movement, causing it to get lost
### Version of OpenTTD

OpenTTD 20240307-master-g8fb26612c0

### Expected result

Ship reverses when leaving Grindington Docks, meaning it's allowed to reach region `2,1` and isn't flagged as lost.

`CheckShipReverse` should try to mimic the same path results as `ChooseShipTrack` to maintain behaviour consistency.

### Actual result

Ship doesn't reverse when leaving Grindington Docks, moves forward and gets cornered between the two docks, flagging the ship as lost.

`CheckShipReverse` restricts the path to regions `2,2` and `1,2`, but the ship happens to be at the top corner of region `2,2`, meaning it fails to find a path, defaults to returning false to reversing.

### Steps to reproduce

[ship checking reverse when leaving dock.zip](https://github.com/OpenTTD/OpenTTD/files/14531109/ship.checking.reverse.when.leaving.dock.zip)
1. Load the attached savegame
2. Watch ship when it departs from Grindington Docks.

Kuhnovic:
## Motivation / Problem

Fixes #12228, #12231, #12333

## Description

Except for specific situations such as leaving a dock, the low-level ship pathfinder only considers the forward direction as an option. However, the high-level pathfinder doesn't (and can't) take the ship direction into account, it only knows whether tiles of water are interconnected or not. This means that there are certain situations where there high-level pathfinder returns a path that requires a reversal, but the low-level pathfinder doesn't consider that reversal to be an option.

The solution is to always consider reversal. This also means the CheckShipReverse and ChooseShipTrack functions can be merged together, which takes away a good amount of duplication.

## Limitations

- In theory ships can now turn around whenever they want, even mid-journey between two docks for example. I ran some tests and never saw this happen, but that doesn't mean it couldn't happen
- An unintended consequence of this change is that ships are now able to turn around in long canals / rivers if you decide to skip the order. Previously they had to go all the way to the end of the canal before they could turn around. IMO it's an improvement, but it might be considered "changing the gameplay".

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_ship.cpp
+++ b/src/pathfinder/yapf/yapf_ship.cpp
@@ -21,7 +21,7 @@
 
 constexpr int NUMBER_OR_WATER_REGIONS_LOOKAHEAD = 4;
 constexpr int MAX_SHIP_PF_NODES = (NUMBER_OR_WATER_REGIONS_LOOKAHEAD + 1) * WATER_REGION_NUMBER_OF_TILES * 4; // 4 possible exit dirs per tile.
-
+constexpr int SHIP_REVERSAL_PENALTY = YAPF_TILE_LENGTH * 10;
 constexpr int SHIP_LOST_PATH_LENGTH = 8; // The length of the (aimless) path assigned when a ship is lost.
 
 template <class Types>
@@ -172,6 +172,14 @@ public:
 		return 'w';
 	}
 
+	/** Returns a random trackdir out of a set of trackdirs. */
+	static Trackdir GetRandomTrackdir(TrackdirBits trackdirs)
+	{
+		const int strip_amount = RandomRange(CountBits(trackdirs));
+		for (int s = 0; s < strip_amount; ++s) RemoveFirstTrackdir(&trackdirs);
+		return FindFirstTrackdir(trackdirs);
+	}
+
 	/** Returns a random tile/trackdir that can be reached from the current tile/trackdir, or tile/INVALID_TRACK if none is available. */
 	static std::pair<TileIndex, Trackdir> GetRandomFollowUpTileTrackdir(const Ship *v, TileIndex tile, Trackdir dir)
 	{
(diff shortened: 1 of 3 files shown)


2TallTyler (commented):
Commit checker wants `Fix #12228, Fix #12231, Fix #12333: Ship pathfinder always considered reversing`, I think. 🙂

rubidium42:
Has the problem mentioned at the top of the description been resolved with the force push? If not, maybe "Convert to draft".

Kuhnovic:
Problems should be resolved now. This could really benefit from some AI torture testing :)

src/pathfinder/yapf/yapf_ship.cpp
```diff
@@ -21,7 +21,7 @@
 
 constexpr int NUMBER_OR_WATER_REGIONS_LOOKAHEAD = 4;
 constexpr int MAX_SHIP_PF_NODES = (NUMBER_OR_WATER_REGIONS_LOOKAHEAD + 1) * WATER_REGION_NUMBER_OF_TILES * 4; // 4 possible exit dirs per tile.
-
+constexpr int SHIP_REVERSAL_PENALTY = YAPF_TILE_LENGTH * 10;
```
SamuXarick, line 24:
```suggestion
constexpr int SHIP_REVERSAL_PENALTY = YAPF_TILE_LENGTH * 25;
```
I tested values as low as 1500 and 2000, and they were still causing issues in the tests. 2500 was the lowest value I tested which doesn't cause ships to be going back and forth.

rubidium42: How big is the loop that the ships that are still going in circles are making with the 2000 penalty?
Am I right that those "circles" are mostly up/down or left/right on the map, i.e. over the diagonals?

Might then be that the minimum penalty is related to the maximum distance within a tile, e.g. 32 "half" tiles, with a penalty of 32 * YAPF_TILE_CORNER_LENGTH (71) => 2272. That could explain why 2000 doesn't work and 2500 does. Though this is pure speculation from my part. 
It would be nice if there were some background to why a particular penalty was chosen. As "arbitrary" by "empirical research" might not be the best reason.

Kuhnovic:
There are still issues after these "fixes". I'm going to close this PR and split the changes up. Some of these changes are really nice IMO and relatively safe. The other stuff needs more work.

PR #8391, opened 2020-12-16
RFC: Feature: Public key company authorization

Milek7:
This PR serves more as RFC, as there are many things left to consider which could change implementation significantly

[Libhydrogen](https://github.com/jedisct1/libhydrogen) is included in-tree as cryptography library.
As we don't need full-blown complex TLS crypto library, [Libsodium](https://github.com/jedisct1/libsodium) is usually good choice. However, it is rather big library, which would add another external dependency, grow code size unnecessarily (especially important eg. with Emscripten), etc. Thus libhydrogen is smaller alternative, which could be included directly in-tree avoiding any extra dependencies. [Monocypher](https://monocypher.org/) was also considered but I decided against it as it requires random bytes to be supplied by application, which would require additional platform-specific code on our side.

**How this is implemented now:**
For now in this PR only single pubkey is stored for a company. It works as alternative access method to password, so you can join with matching pubkey OR password. As I think storing multiple pubkeys should be ultimately implemented, GUI changes for current solution are rudimentary.

In `NetworkStartUp`, keypair is read from `BASE_DIR/keypair` in format of `publickey,secretkey` (keys are hex-encoded). If file cannot be parsed or keypair is invalid, new keypair is generated and saved to this file.
Always during joining to server, after NewGRF check and possible game password check server sends `PACKET_SERVER_NEED_KEYAUTH` packet containing challenge with 16 random bytes. Client then replies with `PACKET_CLIENT_KEYAUTH` containing its own public key and signature for given challenge. Server verifies challenge and stores client pubkey in `ClientInfo`, or disconnects client when signature is invalid.
Now, if client tries to join company with matching pubkey, access is allowed without any passwords.
Client can protect company with pubkey sending `PACKET_CLIENT_PROTECT_COMPANY`, containing bool indicating whether to protect or unprotect. As for now only single pubkey is stored, it always stores pubkey of client who sent the packet.
Autoprotecting on company creation can be enabled by `network.company_autoprotect` setting (enabled by default).
As company is indicated to clients as 'protected' whenever there is any protection on the company, and there is no way to know whether pubkey will succeed, clients now always try to join after clicking on `Password` button without password. This is harmless, except of short password prompt flicker in case when pubkey auth succeeds.
Protecting/unprotecting company with pubkeys manually is done in password setting GUI through two additional buttons.

**Implications and further changes**
[...]

PR #12337, opened 2024-03-19
Change: replace company passwords with allowlist

rubidium42:
## Motivation / Problem

Passwords are not necessarily bad, but the fact that they are not stored in the save is causing a lot of complaints in the community.

With recent changes we added authorized key support for joining the game, and with that comes a public key for which we know the client also knows the private key. This public key can easily be used to create an allow list that is saved and as such survives server restarts.

## Description

This removes company passwords and everything associated with that, and replaces that with an allow list of clients that may join a company. This allow list is saved in the saves and as such survives reloading the game (Closes #8391, related to #10999).

Things that got removed:
* auto clean unprotected: all companies are now protected.
* lot of UI for company passwords.
* default company password.
* network_id, game seed and custom password hashing.

Added:
* saving of an allow list.
* the client making the company gets automatically added to the allow list of that company.
* in the client list there is a green "join" button for clients that are not on the allow list for your company, that you can allow to join your company.
In the example below `Tester #1` is on the allow list for company `Unnamed`, and `Howdy` is not.
clients
* in the client/company list, the join button is only shown when you are on the allow list; the server owner is by design always granted access, even though not on the allow list.

## Limitations

This does not:
* provide (UI for) management of allow lists (including removal from allow list).
* provide options to export and import allow lists.
* provide owner/admin vs user for allow list management; when you're allowed in the company, you can allow others.
* provide requesting of access to a company (use chat for that).
* provide option for automatically allowing everyone into companies, i.e. making them essentially unprotected again.
* add moved clients to the allow list automatically

All because these would balloon this PR significantly. The current set of functionality seems to be the minimal viable product, and it already quite big so better not make it any bigger.

I'm not sure whether much of the PR can be split into separate PRs, as that would create a weird Frankenstein-state with both passwords and authorized keys that's not really useful or has loads of dead code.

## Checklist for review

### DIFF
--- a/src/network/core/tcp_admin.h
+++ b/src/network/core/tcp_admin.h
@@ -333,7 +333,7 @@ protected:
 	 * string  Name of the company.
 	 * string  Name of the companies manager.
 	 * uint8_t   Main company colour.
-	 * bool    Company is password protected.
+	 * bool    Company is protected.
 	 * uint32_t  Year the company was inaugurated.
 	 * bool    Company is an AI.
 	 * @param p The packet that was just received.
@@ -347,7 +347,7 @@ protected:
 	 * string  Name of the company.
 	 * string  Name of the companies manager.
 	 * uint8_t   Main company colour.
-	 * bool    Company is password protected.
+	 * bool    Company is protected.
 	 * uint8_t   Quarters of bankruptcy.
 	 * uint8_t   Owner of share 1.
 	 * uint8_t   Owner of share 2.
--- a/src/network/core/tcp_game.h
+++ b/src/network/core/tcp_game.h
@@ -202,7 +202,8 @@ protected:
 	 * Send information about a client:
 	 * uint32_t  ID of the client (always unique on a server. 1 = server, 0 is invalid).
 	 * uint8_t   ID of the company the client is playing as (255 for spectators).
-	 * string  Name of the client.
+	 * string Name of the client.
+	 * string Public key of the client.
 	 * @param p The packet that was just received.
 	 */
 	virtual NetworkRecvStatus Receive_SERVER_CLIENT_INFO(Packet &p);
--- a/src/network/core/tcp_game.h
+++ b/src/network/core/tcp_game.h
@@ -71,10 +71,6 @@ enum PacketGameType : uint8_t {
 	PACKET_SERVER_CHECK_NEWGRFS,         ///< Server sends NewGRF IDs and MD5 checksums for the client to check.
 	PACKET_CLIENT_NEWGRFS_CHECKED,       ///< Client acknowledges that it has all required NewGRFs.
 
-	/* Checking the company passwords. */
-	PACKET_SERVER_NEED_COMPANY_PASSWORD, ///< Server requests the (hashed) company password.
-	PACKET_CLIENT_COMPANY_PASSWORD,      ///< Client sends the (hashed) company password.
-
 	/* The server welcomes the authenticated client and sends information of other clients. */
 	PACKET_SERVER_WELCOME,               ///< Server welcomes you and gives you your #ClientID.
 	PACKET_SERVER_CLIENT_INFO,           ///< Server sends you information about a client.
@@ -119,9 +115,7 @@ enum PacketGameType : uint8_t {
 	PACKET_SERVER_MOVE,                  ///< Server tells everyone that someone is moved to another company.
 
 	/* Configuration updates. */
-	PACKET_CLIENT_SET_PASSWORD,          ///< A client (re)sets its company's password.
 	PACKET_CLIENT_SET_NAME,              ///< A client changes its name.
-	PACKET_SERVER_COMPANY_UPDATE,        ///< Information (password) of a company changed.
 	PACKET_SERVER_CONFIG_UPDATE,         ///< Some network configuration important to the client changed.
 
 	/* A client quitting. */
@@ -226,14 +220,6 @@ protected:
 	virtual NetworkRecvStatus Receive_SERVER_AUTH_REQUEST(Packet &p);
 
 	/**
-	 * Indication to the client that the server needs a company password:
-	 * uint32_t  Generation seed.
-	 * string  Network ID of the server.
-	 * @param p The packet that was just received.
-	 */
-	virtual NetworkRecvStatus Receive_SERVER_NEED_COMPANY_PASSWORD(Packet &p);
-
-	/**
 	 * Send the response to the authentication request:
 	 * 32 * uint8_t Public key of the client.
 	 * 16 * uint8_t Message authentication code.
@@ -251,18 +237,8 @@ protected:
 	virtual NetworkRecvStatus Receive_SERVER_ENABLE_ENCRYPTION(Packet &p);
 
 	/**
-	 * Send a password to the server to authorize
-	 * uint8_t   Password type (see NetworkPasswordType).
-	 * string  The password.
-	 * @param p The packet that was just received.
-	 */
-	virtual NetworkRecvStatus Receive_CLIENT_COMPANY_PASSWORD(Packet &p);
-
-	/**
 	 * The client is joined and ready to receive their map:
 	 * uint32_t  Own client ID.
-	 * uint32_t  Generation seed.
-	 * string  Network ID of the server.
 	 * @param p The packet that was just received.
 	 */
 	virtual NetworkRecvStatus Receive_SERVER_WELCOME(Packet &p);
@@ -403,13 +379,6 @@ protected:
 	virtual NetworkRecvStatus Receive_SERVER_EXTERNAL_CHAT(Packet &p);
 
 	/**
-	 * Set the password for the clients current company:
-	 * string  The password.
-	 * @param p The packet that was just received.
-	 */
-	virtual NetworkRecvStatus Receive_CLIENT_SET_PASSWORD(Packet &p);
-
-	/**
 	 * Gives the client a new name:
 	 * string  New name of the client.
 	 * @param p The packet that was just received.
@@ -498,17 +467,9 @@ protected:
 	/**
 	 * Request the server to move this client into another company:
 	 * uint8_t   ID of the company the client wants to join.
-	 * string  Password, if the company is password protected.
 	 * @param p The packet that was just received.
 	 */
 	virtual NetworkRecvStatus Receive_CLIENT_MOVE(Packet &p);
-
-	/**
-	 * Update the clients knowledge of which company is password protected:
-	 * uint16_t  Bitwise representation of each company
-	 * @param p The packet that was just received.
-	 */
-	virtual NetworkRecvStatus Receive_SERVER_COMPANY_UPDATE(Packet &p);
 
 	/**
 	 * Update the clients knowledge of the max settings:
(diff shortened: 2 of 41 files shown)


DefinitelyNotRau117:
> * clients moved by an admin (rcon) move get added to the allow list of the destination company.


With some degree of probability, a situation may occur when the admin move a player “for a couple of minutes to fix something in someone else’s company”, but then access to the company should be closed.
Will it be possible to “opt out” of being automatically added to the allowlist?

I suggest adding a command parameter to disable auto-adding to the allowlist. Or, what would be more determinant, do not add client to the allowlist by default, add it only when using the parameter. Let's say _rcon 1234 "move 999 15 **a**"_

TrueBrain:
> * in the client list there is a green "join" button for clients that are not on the allow list for your company, that you can allow to join your company.

Mind adding a screenshot about this? Just because I am too lazy to build and test myself :)

src/network/core/tcp_game.h
```diff
@@ -203,6 +203,7 @@ class NetworkGameSocketHandler : public NetworkTCPSocketHandler {
 	 * Send information about a client:
 	 * uint32_t  ID of the client (always unique on a server. 1 = server, 0 is invalid).
 	 * uint8_t   ID of the company the client is playing as (255 for spectators).
+	 * uint16_t  Bitmask of the companies this client can join.
```
TrueBrain: As the allow-list is stored in the savegame later on, isn't this information deducible from the savegame?

I am mainly considering the case where some patch-sets make it possible to have 500 companies in network games. This would be another nail in that idea, as having 500 bits is impractical :)

src/network/core/tcp_game.h
```diff
@@ -502,6 +503,13 @@ class NetworkGameSocketHandler : public NetworkTCPSocketHandler {
 	 */
 	virtual NetworkRecvStatus Receive_CLIENT_MOVE(Packet &p);
 
+	/**
+	 * Request the server to authorize a given client to join our company.
+	 * uint32_t  ID of the client to authorize to join our company.
+	 * @param p The packet that was just received.
+	 */
+	virtual NetworkRecvStatus Receive_CLIENT_COMPANY_AUTHORIZE(Packet &p);
```
TrueBrain: I wonder if this is the best approach; we could also make it a DoCommand, and that way all clients get notified of the change too.

If we would use a DoCommand, the server will need to do filtering on the command ofc. But this isn't something new, we already do that for join-company etc. Basically, it will become comparable with `CMD_COMPANY_CTRL`.

It would also remove the need for `joinable_companies` to exist. And later additions might be easier, as dealing with the Command system from GUIs etc we have plenty of examples for.

Additionally, if using the DoCommand system, and the client makes a savegame at any moment in time, they too have the right access list in their game. As, if I read it correctly, currently it does not. It does receive the initial allow-list, but it is not modified (client-side) after. And for some reason this feels like asking for a bit of trouble.

But in the tldr, it feels there are currently two systems: `joinable_companies` to talk server<->client, and `network_authorized_keys` server-side, which is also sent to the clients on join. It feels like using DoCommand would collapse them into a single system, making it easier to understand.

And now I am curious about your thoughts on this :)

src/network/core/tcp_admin.h
```diff
@@ -347,7 +347,7 @@ class NetworkAdminSocketHandler : public NetworkTCPSocketHandler {
 	 * string  Name of the company.
 	 * string  Name of the companies manager.
 	 * uint8_t   Main company colour.
-	 * bool    Company is password protected.
+	 * bool    Company is password protected (always false since 15 due to removal of passwords).
```
TrueBrain: Maybe it would be better to make it always `true`, to make software using admin port think all companies are now protected? (as they, are just not with a password)

rubidium42:
> With some degree of probability, a situation may occur when the admin move a player “for a couple of minutes to fix something in someone else’s company”, but then access to the company should be closed. Will it be possible to “opt out” of being automatically added to the allowlist?

Not with this PR, but that does not rule out that it won't be made. Just that adding that would balloon the size of this PR even more.


> Mind adding a screenshot about this? Just because I am too lazy to build and test myself :)

Sure.

src/network/core/tcp_game.h
TrueBrain: Btw, for some ideas I wrote in https://gist.github.com/TrueBrain/cd5979cb7ba14b6ac649c74cd9fcd3eb, we would need to get the pubkey of all clients to all clients. So that might be good to use here, as that would also allow the DoCommand to add any other pubkey they like (which for example they got via a copy/paste from a friend).

I know, not for this PR, but just some food for thought.

src/network/core/tcp_game.h
rubidium42: To be brutally honest, I do not think that we should actually be sending the public keys of all clients to all clients as that would make tracking clients by non-servers way too trivial to implement. Within the context of servers there's nothing that can be done about that I fear.

The only change that would be required for that is saving an empty collection of authorized keys when we are sending the network saves compared to saving it just to disk.

src/network/core/tcp_game.h
rubidium42: The pool has a decent bitmap with way more than 500 bits.
Also the stored public key strings are at least 65 * 8 bits (520).

But alas, I do not think we should send the public key with the client to all clients as it's essentially a unique identifier that you might want to share with the server but not with everybody.
It's not that I'd like to share my SSH public key really publicly, I share it only with servers I want to login to.

src/network/core/tcp_admin.h
rubidium42: That a good reason to set it to `true`.
Though it is not protected by a password, which is what the comment says.

I don't care. Anyone else with an opinion?

src/network/core/tcp_game.h
TrueBrain: Hmm, that is not how public keys work :) They are meant to be shared, to be public. For example:

https://github.com/TrueBrain.keys
(empty for me, as I don't have any keys installed on GitHub atm)

You already share publicly your SSH public key (sorry if you didn't know :D ).

But mostly, you already send the public keys of everyone on the allow list to all clients. So that is already done / implied :) But this conversation got split up over multiple comments, let's continue in the other, as there you mention something far more interesting to this conversation :)

src/network/core/tcp_game.h
TrueBrain: Ah, okay, so you want to keep the public keys of clients hidden as you are afraid people would monitor who is where, etc.

When I worked out how multiplayer could look like, many moons ago (https://gist.github.com/TrueBrain/05d5d066efedd656ec7daa05f92fa012), it kinda was the point that public keys became this identity you can share with others, to have friends list, and to find each other. But that indeed explicitly makes you "trackable" (in some sense) on what server you are etc.

I can see both arguments to this story .. on one hand you have user-friendly-ness, on the other hand you have do-not-track. But given the state you have this PR in, and that it is more about the second than the first, I have no problem going with the second (for now at least :P ), and see where it leads us.

But yes, it does require finding a solution to ensure the authorized-keys are not send to the clients; as that would instantly defeat that purpose :D (you can track by downloading the map; which wouldn't be much different than sending the public-keys to everyone).

Personally, I don't share the hesitation of not sharing the pubkey with everyone. And yes, in some sense that means you can be tracked where you are playing. But I am really not worried someone can see on what server I am playing etc .. it is a game after all. So my acceptance for that is a lot higher. That said, it is my personal opinion, and I have no problem going with yours for now :)

src/network/core/tcp_admin.h
TrueBrain: So, change the comment to read: "Company is protected (always false ...)" :P

It is more that tooling that already exists isn't going: omg your server has unprotected companies!!!!, after this patch :) As that isn't true. And the intention of this field was ofc the "protected" part, not the "password" part :)

src/network/core/tcp_game.h
TrueBrain: Another thought, but really not for this PR: one could implement two client-side modes: "Privacy" and "Public".

With the first, you generate a new pubkey for every server you connect to (stable for that server, but never reused with another server), based on invite-code or something.

With the second, you have a single pubkey for every server.

That would also tackle the "do not track" situation, giving the control to the player what level of privacy they would like.

But again, if you ask me, not for this PR; after this explanation of yours, I am find with this approach :) Maybe others have an opinion :P

rubidium42: The command idea is intriguing, but it has some drawbacks to think about:
* just passing authorized keys as string; that works fine, but someone can just keep sending different strings and essentially cause a denial of service attack on servers by either memory exhaustion or blowing up the saves to a point where clients cannot join fast enough.
* using `ClientID`, and resolving the authorized key from there. Now your client is a bit slow, so server executes the command successfully and then sends that the client left. The client will process the "left" message and remove the client, which makes the client not resolve and we have a different state in the server and client. That would mean that moving client moving to commands causes a desync.
* so you would need to send the authorized key and client, add special code for validating those to the server, and then reject or accept that. I'm not sure that's the way to go.

Of course the first case can also be triggered with the other possibilities, but at most MAX_COMPANY times the authorized key for each of the joined clients. As such, it would grow significantly slower as a lot of clients need to join.

Not that I see a really clean and simple way to keep the authorized keys from the save for non locally stored server games.

TrueBrain: The first problem could be prevented by just adding a max to the length of the authorize list. That is not that weird, I would say. Limit of 50 or so.

I am not sure I fully understand your second and third point. I think you are talking about my suggestion to just share pubkeys with clients/servers? As if so, I think it would make most sense to just send pubkeys, and not ClientIDs. As that way, you can also have a client authorize a pubkey of a player not in-game currently. I cannot see any scenario where you also want to send a ClientID there, or the server need to do any validation if they match. I would just keep it simple.

There are a bunch of settings you could add here, ofc. For example, a setting that does validate if it knows the pubkey. But this too is irrelevant to the ClientID. 

If you do want to use the ClientID for this DoCommand, I think that only makes sense in a scenario where you want to keep the pubkey server-side only. But in that case, using DoCommand is a bit tricky, as how do you tell a client that is joining which company it can join?

Either way, I do not really understand what you mean with your second drawback.

- Client A sends DoCommand with ClientID
- Server processes DoCommand, adds pubkey of ClientID to allow-list
- Server broadcasts DoCommand to all other clients
- Server notices client A is too slow, and kicks it
- Client A is now irrelevant, it is disconnected
- Client B...Z receive the DoCommand to add the pubkey to the allow-list
- Client B..Z receive disconnect

I guess you talk about the situation where the last and second-to-last are switched around? Another reason to not use ClientID when you share pubkeys between client/server, and irrelevant if you keep the pubkey server-side-only as the only one who actually cared was the client that is disconnected :)

rubidium42: I have changed it to a command in an effort to keep the allow lists of the companies the same between clients and server

PeterN:
I think I found a bug:

1) Start up a non-dedicated multiplayer server.
2) Switch to spectator.
3) (Try to) Join original company.

image

Any company created after this I can join and leave as expected, just not the initial company.

PeterN (approved):

MuxyDuGoulp:
damnit, no more company password.
Who was complaining ?
In OpenTTD France's Servers, all companies passwords were saved into a password file, and were reloaded if savegame used on the server.
RIP Network id. I was using this item to identify uniquely the server from admin port.
I will study this change ...

PeterN:
Yes. And that is one of the reasons why passwords are gone...

PR #12338, opened 2024-03-19
Codechange: Use std::list for News Items.

PeterN:
## Motivation / Problem

The list of `NewsItem`s is a C-style doubly-linked list, using manual object management with `new`/`delete`, and lots of manual iteration.

## Description

Switch to using a `std::list` instead. Current, forced, and statusbar news items are now referenced by list iterators, to allow list traversal. These are updated (as before) when `DeleteNewsItem` is called.

## Limitations

"Why?" I guess...

## Checklist for review

### DIFF
--- a/src/crashlog.cpp
+++ b/src/crashlog.cpp
@@ -56,11 +56,12 @@ static void SurveyRecentNews(nlohmann::json &json)
 	json = nlohmann::json::array();
 
 	int i = 0;
-	for (NewsItem *news = _latest_news; i < 32 && news != nullptr; news = news->prev, i++) {
-		TimerGameCalendar::YearMonthDay ymd = TimerGameCalendar::ConvertDateToYMD(news->date);
+	for (const auto &news : GetNews()) {
+		TimerGameCalendar::YearMonthDay ymd = TimerGameCalendar::ConvertDateToYMD(news.date);
 		json.push_back(fmt::format("({}-{:02}-{:02}) StringID: {}, Type: {}, Ref1: {}, {}, Ref2: {}, {}",
-		               ymd.year, ymd.month + 1, ymd.day, news->string_id, news->type,
-		               news->reftype1, news->ref1, news->reftype2, news->ref2));
+		               ymd.year, ymd.month + 1, ymd.day, news.string_id, news.type,
+		               news.reftype1, news.ref1, news.reftype2, news.ref2));
+		if (++i > 32) break;
 	}
 }
 
--- a/src/news_gui.cpp
+++ b/src/news_gui.cpp
@@ -21,7 +21,6 @@
 #include "town.h"
 #include "sound_func.h"
 #include "string_func.h"
-#include "dropdown_func.h"
 #include "statusbar_gui.h"
 #include "company_manager_face.h"
 #include "company_func.h"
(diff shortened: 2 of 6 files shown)


src/news_gui.cpp
```diff
@@ -44,25 +44,40 @@
 
 #include "safeguards.h"
 
-const NewsItem *_statusbar_news_item = nullptr;
+static const uint MIN_NEWS_AMOUNT = 30;        ///< preferred minimum amount of news messages.
```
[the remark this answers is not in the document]
PeterN: Well, except as a template parameter, CodeQL...

src/crashlog.cpp
```diff
@@ -55,8 +55,8 @@ static void SurveyRecentNews(nlohmann::json &json)
 {
 	json = nlohmann::json::array();
 
-	int i = 0;
-	for (NewsItem *news = _latest_news; i < 32 && news != nullptr; news = news->prev, i++) {
+	auto [first, last] = GetRecentNews(32);
+	for (auto news = first; news != last; ++news) {
```
rubidium42: Is there a reason to have this function? It's only used once in here, and with this it has to iterate through the news items twice. After all `std::distance` is not a constant operation of `std::list` compared to `std::vector`.

May something inspired by the following would work?
```suggestion
	for (auto news : std::ranges::take_view(_news, 32)) {
```

nielsmh:
Is there a reason to even keep it as a linked list? Could a vector or deque accomplish the same in possibly less awkward ways?

src/crashlog.cpp
PeterN: std::ranges doesn't work with clang 15 :(

PeterN:
> Is there a reason to even keep it as a linked list? Could a vector or deque accomplish the same in possibly less awkward ways?

DeleteNews does delete items through the list rather than just the start or end. Each of those deletions would require moving everything if it s a vector/deque.

src/news_gui.cpp
```diff
@@ -44,25 +43,42 @@
...
-static uint _total_news = 0;             ///< current number of news items
-static NewsItem *_oldest_news = nullptr; ///< head of news items queue
-NewsItem *_latest_news = nullptr;        ///< tail of news items queue
+static NewsContainer _news; ///< List of news, with newest items at the start.
 
 /**
  * Forced news item.
  * Users can force an item by accessing the history or "last message".
  * If the message being shown was forced by the user, a pointer is stored
  * in _forced_news. Otherwise, \a _forced_news variable is nullptr.
```
rubidium42: It's not quite `nullptr` anymore.

src/news_gui.cpp
```diff
@@ -999,7 +971,7 @@ static void RemoveOldNewsItems()
  */
 void ChangeVehicleNews(VehicleID from_index, VehicleID to_index)
 {
-	for (NewsItem *ni = _oldest_news; ni != nullptr; ni = ni->next) {
+	for (auto ni = std::begin(_news); ni != std::end(_news); ++ni) {
```
rubidium42: Can't this be something like the following. Yeah, I know it's a bit of scope creep.
```suggestion
	for (auto ni : _news) {
```

PeterN: Good catch, range-for is preferred in my book.

rubidium42 (approved):

PR #12345, opened 2024-03-21
Codechange: Replace path cache deques with vectors.

PeterN:
## Motivation / Problem

RoadVehicle path cache has two std::deques which is quite awkward to manage, and also quite memory hungry.
Ship path cache use a single std::deque.
std::deque was used to be able to push/pop from either end.

#12275 attempts to partially address RoadVehicle's path cache by hiding the fact it has two deques.

With this PR I address both the split, and the memory usage.

## Description

Instead, change both to use a single `std::vector` each. For `RoadVehicle`s this now contains a `std::pair<Trackdir, TileIndex>` which simplifies usage.

This is now push/popped from the back, so sometimes needs to be reversed.

Also a special-case for ships that was copied across in the original implementation, which no longer exists for ships, has been removed for RVs. As far as I can tell this never made any sense, as an RV destination tile cannot have a path finding choice, so would never appear in the path cache anyway.

| Type | Master | PR |
| ---- | ------ | -- |
| Ship | 640 bytes + | 584 bytes + |
| RoadVehicle | 776 bytes + | 640 bytes + |

+ Extra memory outside the object is used by both deque and vector.

## Limitations

Performance not compared. No regression expected for Ships, but RVs has a little bit of erasing which could perhaps be improved.

~~Saveload code for RoadVehicle is a bit gnarly as std::pair is difficult to saveload, so it always uses two temporary vectors. Maybe using a struct instead would help.~~

## Checklist for review

### DIFF
--- a/src/roadveh.h
+++ b/src/roadveh.h
@@ -81,24 +81,16 @@ static const uint8_t RV_OVERTAKE_TIMEOUT = 35;
 void RoadVehUpdateCache(RoadVehicle *v, bool same_length = false);
 void GetRoadVehSpriteSize(EngineID engine, uint &width, uint &height, int &xoffs, int &yoffs, EngineImageType image_type);
 
-struct RoadVehPathCache {
-	std::deque<Trackdir> td;
-	std::deque<TileIndex> tile;
+/** Element of the RoadVehPathCache. */
+struct RoadVehPathElement {
+	Trackdir trackdir; ///< Trackdir for this element.
+	TileIndex tile; ///< Tile for this element.
 
-	inline bool empty() const { return this->td.empty(); }
-
-	inline size_t size() const
-	{
-		assert(this->td.size() == this->tile.size());
-		return this->td.size();
-	}
-
-	inline void clear()
-	{
-		this->td.clear();
-		this->tile.clear();
-	}
+	constexpr RoadVehPathElement() : trackdir(INVALID_TRACKDIR), tile(INVALID_TILE) {}
+	constexpr RoadVehPathElement(Trackdir trackdir, TileIndex tile) : trackdir(trackdir), tile(tile) {}
 };
+
+using RoadVehPathCache = std::vector<RoadVehPathElement>;
 
 /**
  * Buses, trucks and trams belong to this class.
--- a/src/saveload/vehicle_sl.cpp
+++ b/src/saveload/vehicle_sl.cpp
@@ -12,6 +12,7 @@
 #include "saveload.h"
 #include "compat/vehicle_sl_compat.h"
 
+#include "../debug.h"
 #include "../vehicle_func.h"
 #include "../train.h"
 #include "../roadveh.h"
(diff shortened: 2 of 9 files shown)


src/saveload/vehicle_sl.cpp
```diff
@@ -841,22 +845,56 @@ class SlVehicleRoadVeh : public DefaultSaveLoadHandler<SlVehicleRoadVeh, Vehicle
...
+		if (IsSavegameVersionBefore(SLV_PATH_CACHE_FORMAT)) {
+			/* Path cache is now taken from back instead of front, so needs reversing. */
+			std::reverse(std::begin(rv->path), std::end(rv->path));
+		}
+	}
+
 	void Save(Vehicle *v) const override
 	{
 		if (v->type != VEH_ROAD) return;
+		ClearPathCache();
```
Kuhnovic: I think it makes more sense to call ClearPathCache from JoinPathCache / SplitPathCache directly, since only those two functions deal with the static vectors.

src/saveload/vehicle_sl.cpp
```diff
@@ -841,22 +845,56 @@ class SlVehicleRoadVeh : public DefaultSaveLoadHandler<SlVehicleRoadVeh, Vehicle
...
+			rv->path.emplace_back(rv_path_td[c], rv_path_tile[c]);
+		}
+
+		if (IsSavegameVersionBefore(SLV_PATH_CACHE_FORMAT)) {
+			/* Path cache is now taken from back instead of front, so needs reversing. */
+			std::reverse(std::begin(rv->path), std::end(rv->path));
+		}
+	}
+
 	void Save(Vehicle *v) const override
```
Kuhnovic, line 877: A reference probably makes more sense in this context, a nullptr vehicle doesn't make much sense. But I guess I lot of the code here is C-style passing pointers so in that sense it would be consistent. Whatever you prefer.

Same for the other functions btw.

src/saveload/vehicle_sl.cpp
```diff
@@ -841,22 +845,56 @@ class SlVehicleRoadVeh : public DefaultSaveLoadHandler<SlVehicleRoadVeh, Vehicle
...
+
+	void SplitPathCache(RoadVehicle *rv) const
+	{
+		for (const auto &pair : rv->path) {
+			rv_path_td.push_back(pair.dir);
+			rv_path_tile.push_back(pair.tile);
+		}
+	}
+
+	void JoinPathCache(RoadVehicle *rv) const
```
Kuhnovic: Isn't "merge"a more correct them instead of "join"? Join immediately makes me thing about (string) concatenation, where merging is more like "combine" or "mix".

Not a native English speaker, just my 2ct

src/saveload/vehicle_sl.cpp
PeterN: It's the interface to SaveLoad that uses pointers, I'm not adding to changing anything here.

src/saveload/vehicle_sl.cpp
PeterN: The order of operations is different:

Save is Clear → Split → SlObject.
Load is Clear → SlObject → Join

src/saveload/vehicle_sl.cpp
PeterN: Fair :)

src/saveload/vehicle_sl.cpp
Kuhnovic: Ah yeah, then ignore this :)

PeterN:
I also have an alternative version of this that uses a standard saveload struct list for the road vehicle path cache, which avoids having to split and merge.

However the code for merging is still needed for loading existing saves...

PeterN:
Now updated finally to use the "normal" method to save the vector members together.

Introduces VectorSaveLoadHandler to simplify loading/saving vectors this way, including existing vectors.

PeterN:
As well as the storage size for `std::deque`'s internal data, there is also the buffer size used by `std::deque`. On my system that is 512 bytes. This means that 512 bytes are used for every ship with a cache, and 1024 bytes for every road vehicle with a cache.

With the new `std::vector` code we only ever add a few elements so the size should be fairly small going by `std::vector`'s default allocation strategy.

Here is a demonstration with a large savegame, showing the path caches weighing in at 39.8MiB with master, and 91.3KiB with this PR.

image
image

src/roadveh.h
```diff
@@ -81,25 +81,16 @@ static const uint8_t RV_OVERTAKE_TIMEOUT = 35;
 void RoadVehUpdateCache(RoadVehicle *v, bool same_length = false);
 void GetRoadVehSpriteSize(EngineID engine, uint &width, uint &height, int &xoffs, int &yoffs, EngineImageType image_type);
 
-struct RoadVehPathCache {
-	std::deque<Trackdir> td;
-	std::deque<TileIndex> tile;
+struct RoadVehPath {
```
LordAro: maybe `RoadVehPathNode` ? `RoadVehPath` suggests it's the entire path

also for all the other structs

(also, doxygen?)

Kuhnovic: Node seems to suggest it's a pathfinder node, which it isn't. I would just call it RoadVehPathElement or RoadVehPathEntry, something along those lines

LordAro (approved):

PR #12346, opened 2024-03-21
Fix #7982: Show existing coverage with unambiguous adjacent station

### LINKED ISSUE #7982: Station coverage isn't shown when mousing over a part of station adjacent to other station
## Version of OpenTTD
master-3fcb240f8e4d02aace63a7ba4a96284b3aedef86

## Expected result
Showing a coverage area of a station we're mousing over as it done for any other part of it.

## Actual result
No coverage area is shown.

## Steps to reproduce
When building a station hover a part of existing station adjacent to some other station (with coverage enabled ofc).

Screenshot from 2020-02-07 01-00-34

My guess is that it hides coverage because it thinks there is an ambiguity for station selection  as there is for empty tile.

raddari:
## Motivation / Problem
Fixes #7982.

Hovering a station which is adjacent to both a station with the same ID, and a station with a different ID results in only the hovered station's coverage being shown instead of the combined coverage.

image

## Description
Changes coverage display to check for a station on the hovered tile and another tile in the area containing a station with the same ID.

Convenient test save with a few station layouts:
[Test Stations.sav.tar.gz](https://github.com/OpenTTD/OpenTTD/files/14693877/Test.Stations.sav.tar.gz)

image

## Limitations
This feels like a bit of a hack. I'm sure there's a better way of implementing this, but I think the whole function would need a rewrite.

## Checklist for review

### DIFF
--- a/src/station_gui.cpp
+++ b/src/station_gui.cpp
@@ -92,6 +92,15 @@ void FindStationsAroundSelection()
 	/* Tile area for TileHighlightData */
 	TileArea location(TileVirtXY(_thd.pos.x, _thd.pos.y), _thd.size.x / TILE_SIZE - 1, _thd.size.y / TILE_SIZE - 1);
 
+	/* If the current tile is already a station, then it must be the nearest station. */
+	if (IsTileType(location.tile, MP_STATION) && GetTileOwner(location.tile) == _local_company) {
+		T* st = T::GetByTile(location.tile);
+		if (st != nullptr) {
+			SetViewportCatchmentSpecializedStation<T>(st, true);
+			return;
+		}
+	}
+
 	/* Extended area by one tile */
 	uint x = TileX(location.tile);
 	uint y = TileY(location.tile);


src/station_gui.cpp
```diff
@@ -107,6 +107,11 @@ void FindStationsAroundSelection()
 		if (IsTileType(tile, MP_STATION) && GetTileOwner(tile) == _local_company) {
 			T *st = T::GetByTile(tile);
 			if (st == nullptr) continue;
+			if (IsTileType(location.tile, MP_STATION) && GetStationIndex(location.tile) == GetStationIndex(tile)) {
```
PeterN: I think this can be simplified to just:
```suggestion
			if (location.tile == tile) {
```

PeterN (changes requested):

src/station_gui.cpp
raddari: Fixed. Works as intended with this change

raddari:
> Although thinking about this some more, this won't work if there are two adjacent tiles of different stations before the highlight location, as it will already have stopped the loop.
> 
> However, if location.tile is a station then that can be checked first and the tile area loop itself can be skipped.
> 
> I would suggest undoing the change, and inserting something like this just after `TileArea location` is set up:
> 
> ```c++
> 	/* If the current tile is already station then that has to be the nearest station. */
> 	if (IsTileType(location.tile, MP_STATION) && GetTileOwner(location.tile) == _local_company) {
> 		T *st = T::GetByTile(location.tile);
> 		if (st != nullptr) {
> 			SetViewportCatchmentSpecializedStation<T>(st, true);
> 			return;
> 		}
> 	}
> ```

Yeah, that looks much better than my original idea. I definitely didn't understand how stations worked enough before starting this PR, my bad.

PeterN (approved):

J0anJosep:
I'll port this to my depot branches.
Is it `T* st` or `T *st`?

PeterN:
Whoops, should be `T *st` but I didn't spot that difference. Rebasing should just bring everything in though anyway.

PR #12356, opened 2024-03-22
Cleanup: remove checks for old MSVC versions

rubidium42:
## Motivation / Problem

#12355 triggered me to remove `#if` `_MSC_VER` checks for specific version ranges of MSVC. To support the functions we are using, we require at least 19.29 (some 2019 version), whereas all checks are for versions less than 19.00 (2015).

## Description

Remove the bits of code the preprocessor would remove.

## Checklist for review

### DIFF
--- a/src/os/windows/win32.cpp
+++ b/src/os/windows/win32.cpp
@@ -259,7 +259,7 @@ void CreateConsole()
 		return;
 	}
 
-#if defined(_MSC_VER) && _MSC_VER >= 1900
+#if defined(_MSC_VER)
 	freopen("CONOUT$", "a", stdout);
 	freopen("CONIN$", "r", stdin);
 	freopen("CONOUT$", "a", stderr);
--- a/src/stdafx.h
+++ b/src/stdafx.h
@@ -28,6 +28,12 @@
 
 #if defined(__APPLE__)
 #	include "os/macosx/osx_stdafx.h"
+#else
+/* It seems that we need to include stdint.h before anything else
+ * We need INT64_MAX, which for most systems comes from stdint.h.
+ * For OSX the inclusion is already done in osx_stdafx.h. */
+#	define __STDC_LIMIT_MACROS
+#	include <stdint.h>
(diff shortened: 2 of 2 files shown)


src/os/windows/win32.cpp
```diff
@@ -259,7 +259,7 @@ void CreateConsole()
 		return;
 	}
 
-#if defined(_MSC_VER) && _MSC_VER >= 1900
+#if defined(_MSC_VER)
```
PeterN, line 262: I think this `#if`, and whole `#else ... #endif`  section, could be removed too? This is Windows specific code with a CYGWIN guard.

src/stdafx.h
```diff
@@ -28,6 +28,12 @@
 
 #if defined(__APPLE__)
 #	include "os/macosx/osx_stdafx.h"
+#else
+/* It seems that we need to include stdint.h before anything else
+ * We need INT64_MAX, which for most systems comes from stdint.h.
+ * For OSX the inclusion is already done in osx_stdafx.h. */
+#	define __STDC_LIMIT_MACROS
+#	include <stdint.h>
```
PeterN, line 36: Do we need to `#include <stdint.h>` at all now that we `#include <cstdint>`?

src/os/windows/win32.cpp
rubidium42: A `!`CYGWIN guard

src/stdafx.h
rubidium42: I don't know. I don't want to pull that change into this PR. Feel free to make a separate PR to figure that out though.

src/os/windows/win32.cpp
glx22: IIRC these `freopen` are still needed. And the `#else` part is for MinGW

PeterN: Ah, MinGW strikes, of course.

PeterN (approved):

PR #12358, opened 2024-03-23
Codechange: move 'months_empty' to CompanyProperties

rubidium42:
## Motivation / Problem

For 'autoclean' there is a `struct` with the company password and a `months_empty` field. When removing company passwords (#12337), this `struct` would only contain the months empty field and the vast majority would be accessed using `_network_company_states[c->index].months_empty`, so moving it to `c->months_empty` would simplify that code.

Furthermore the `months_empty` is a `uint16_t` where the maximum thresholds are `240`, so a `uint8_t` would be just fine (with some guarding).

## Description

Put the `months_empty` in `CompanyProperties` where it should not increase the size of the `CompanyProperties` object.

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -961,7 +961,8 @@ NetworkRecvStatus ServerNetworkGameSocketHandler::Receive_CLIENT_IDENTIFY(Packet
 	Debug(desync, 1, "client: {:08x}; {:02x}; {:02x}; {:02x}", TimerGameEconomy::date, TimerGameEconomy::date_fract, (int)ci->client_playas, (int)ci->index);
 
 	/* Make sure companies to which people try to join are not autocleaned */
-	if (Company::IsValidID(playas)) _network_company_states[playas].months_empty = 0;
+	Company *c = Company::GetIfValid(playas);
+	if (c != nullptr) c->months_empty = 0;
 
 	return this->SendNewGRFCheck();
 }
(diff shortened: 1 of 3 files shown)


src/network/network_server.cpp
```diff
@@ -1643,37 +1644,37 @@ static void NetworkAutoCleanCompanies()
...
 	/* Go through all the companies */
-	for (const Company *c : Company::Iterate()) {
+	for (Company *c : Company::Iterate()) {
 		/* Skip the non-active once */
 		if (c->is_ai) continue;
 
 		if (!HasBit(has_clients, c->index)) {
 			/* The company is empty for one month more */
-			_network_company_states[c->index].months_empty++;
+			if (c->months_empty != std::numeric_limits<decltype(c->months_empty)>::max()) c->months_empty++;
```
2TallTyler, line 1653: `std::numeric_limits` is fine, but I wonder if it would look cleaner if it compared against a new `AUTOCLEAN_MAX_MONTHS` constant that would take the place of the hardcoded 240 for the three autoclean settings in `network_settings.ini`. Probably a potato/potato question. 😄

2TallTyler (approved):

src/network/network_server.cpp
rubidium42: Lets assume `autoclean_novehicles` is `AUTOCLEAN_MAX_MONTHS` and we limit `months_empty` to `AUTOCLEAN_MAX_MONTHS`.
`months_empty > autoclean_novehicles` => `AUTOCLEAN_MAX_MONTHS < AUTOCLEAN_MAX_MONTHS` => false, so auto clean will be disabled.
So the threshold would need to become `AUTOCLEAN_MAX_MONTHS + 1` and then it starts to become doubtful whether it's really better.

2TallTyler: Ah, yes. 🙂

PR #12369, opened 2024-03-24
Fix #12387, eabb9db: [NewGRF] Wrong tile offset passed to rail station CB 149 (slope check)

### LINKED ISSUE #12387: [Bug]: Building station on slopes is not posible.
### Version of OpenTTD

v 14RC3, Windows

### Expected result

Station supposed to be build on slope only because it got underground entry.

### Actual result

With 'New Stations' grf there isn't possible to build Bamberg station on slope as it supposed to be and it worked in v13. Instead of that is possible to build on flat surface.

### Steps to reproduce

[Temp.zip](https://github.com/OpenTTD/OpenTTD/files/14773167/Temp.zip)

michicc:
## Motivation / Problem

Trying to build rail station 'Bamberg' from newstats 0.6 on the correct slope fails with a slope check error.

## Description / Limitations

In PR #9852 the cost calculation for stations was changed. This invertedly caused the slope check CB 149 to be called with the wrong tile area so the position offsets passed to the CB was always 0. This was caused by `CheckFlatLandRailStation` being called with single tiles and not the full station area.

While investigation, I've also noticed that road stops have had the same change regarding `CheckFlatLandRoadStop`. This leads to the observable difference that it is now possible to drag-build roadstops over multiple height levels while that was not allowed before #9852.

This PR changes the flat land/slope check to once again iterate over the full tile area.

As road stops have no slope CB, one can argue which drag-build behaviour is actually appropriate.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -952,7 +952,8 @@ static CommandCost CheckFlatLandRailStation(TileIndex tile_cur, TileIndex north_
 
 /**
  * Checks if a road stop can be built at the given tile.
- * @param tile_area Area to check.
+ * @param cur_tile Tile to check.
+ * @param allowed_z Height allowed for the tile. If allowed_z is negative, it will be set to the height of this tile.
  * @param flags Operation to perform.
  * @param invalid_dirs Prohibited directions (set of DiagDirections).
  * @param is_drive_through True if trying to build a drive-through station.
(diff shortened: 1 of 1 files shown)


PeterN (approved):
Ignored white space changes. We are removing a loop that was only ever looping over one tile, and then passing the correct origin tile to the slope check function.

PR #12376, opened 2024-03-25
Fix #11345: Use correct default button value for vehicle service interval setting

### LINKED ISSUE #11345: [Bug]: Vehicle service interval setting can't use Default button in percent mode
### Version of OpenTTD

13.4

### Expected result

interval

The Settings menu supports both time- and percent-based vehicle service intervals.

When in percent mode, the default setting is 50%. This is updated automatically when changing `Service intervals are in percents`. Clicking the Default button should reset this interval to 50.

### Actual result

When using percentage mode, this interval is clamped between 5 and 90. Clicking the Default button on the query text window tries to set it to 150, which fails, and nothing happens.

This seems to be a limitation of our settings system, although I'd point at this overcomplicated reuse of service intervals as the real problem. And I'm about to make it worse in #11341. 😉 

### Steps to reproduce

1. Open the Settings menu.
2. Enable `Service intervals are in percents`.
3. Set the default service interval for any vehicle type to a non-default value.
4. Double-click the value to open the query window.
5. Click the Default button.
6. Note that the setting is not changed to its default.

chengandre:
## Motivation / Problem

Fixes #11345

The Default button reads the value in `sd->def`, which only allows one value to be stored. Thus, it tries to reset to the default value in days, which is not possible when service interval are in other units.

The Default Value Text also reads the value in `sd->def` and always displays the same string, which can be confusing.

## Description

A callback is added to correct the value read from `sd->def`.

The `set_value_dparams_cb` callback is also used to supply the correct value string.

CleanShot 2024-03-25 at 17 43 31

## Checklist for review

### DIFF
--- a/src/settings_internal.h
+++ b/src/settings_internal.h
@@ -169,7 +169,7 @@ struct IntSettingDesc : SettingDesc {
 	 * units or expressed as a percentage.
 	 * @param value The prospective new value for the setting.
 	 */
-	typedef void GetDefaultValueCallback(int32_t &value);
+	typedef int32_t GetDefaultValueCallback();
 
 	template <
 		typename Tdef,
--- a/src/settings_internal.h
+++ b/src/settings_internal.h
@@ -165,9 +165,9 @@ struct IntSettingDesc : SettingDesc {
 	 */
 	typedef void PostChangeCallback(int32_t value);
 	/**
-	 * A callback to set the correct default value, in case it can be measured in time
+	 * A callback to get the correct default value. For example a default that can be measured in time
 	 * units or expressed as a percentage.
-	 * @param value The prospective new value for the setting.
+	 * @return The correct default value for the setting.
 	 */
 	typedef int32_t GetDefaultValueCallback();
 
--- a/src/settings_internal.h
+++ b/src/settings_internal.h
@@ -164,6 +164,12 @@ struct IntSettingDesc : SettingDesc {
 	 * @param The new value for the setting.
 	 */
 	typedef void PostChangeCallback(int32_t value);
+	/**
+	 * A callback to set the correct default value, in case it can be measured in time
+	 * units or expressed as a percentage.
+	 * @param value The prospective new value for the setting.
+	 */
+	typedef void GetDefaultValueCallback(int32_t &value);
 
 	template <
 		typename Tdef,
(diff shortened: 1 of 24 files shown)


src/lang/brazilian_portuguese.txt
```diff
@@ -1833,7 +1833,9 @@ STR_CONFIG_SETTING_SERVINT_AIRCRAFT                             :Intervalo padr
 STR_CONFIG_SETTING_SERVINT_AIRCRAFT_HELPTEXT                    :Definir o intervalo padrão de manutenção para novas aeronaves, se um intervalo de manutenção não for explicitamente configurado para o veículo
 STR_CONFIG_SETTING_SERVINT_SHIPS                                :Intervalo padrão de manutenção para embarcações: {STRING}
 STR_CONFIG_SETTING_SERVINT_SHIPS_HELPTEXT                       :Definir o intervalo padrão de manutenção para novas embarcações, se um intervalo de manutenção não for explicitamente configurado para o veículo
-STR_CONFIG_SETTING_SERVINT_VALUE                                :{COMMA}{NBSP}Dia{P 0 "" s}/Minuto{P 0 "" s}/%
+STR_CONFIG_SETTING_SERVINT_VALUE_DAYS                           :{COMMA}{NBSP}Dia{P 0 "" s}
+STR_CONFIG_SETTING_SERVINT_VALUE_MINUTES                        :{COMMA}{NBSP}Minuto{P 0 "" s}
+STR_CONFIG_SETTING_SERVINT_VALUE_PERCENTAGE                     :{COMMA}{NBSP}%
```
nielsmh: Please don't change language files other than `english.txt` in PRs, the web translator doesn't like it very well.

src/settings_table.cpp
```diff
@@ -116,6 +116,28 @@ static void SettingsValueAbsolute(const IntSettingDesc &sd, uint first_param, in
 	SetDParam(first_param + 1, abs(value));
 }
 
+/** Settings Default Value displays the correct units or as a percentage */
+static void SettingsValueText(const IntSettingDesc &sd, uint first_param, int32_t value)
```
nielsmh: This is a really generic name for a quite specific function. Shouldn't it be named something related to servicing interval since it's only used for those settings?

nielsmh (commented):
You should probably separate the change of adding a default value callback feature from the actually using that feature to determine the default for a setting. I.e. split up the first commit into two.

src/lang/brazilian_portuguese.txt
chengandre: Got it, I'll revert the files.

src/settings_table.cpp
chengandre: Indeed, I'll make it more specific.

src/settings_internal.h
```diff
@@ -164,6 +164,12 @@ struct IntSettingDesc : SettingDesc {
 	 * @param The new value for the setting.
 	 */
 	typedef void PostChangeCallback(int32_t value);
+	/**
+	 * A callback to set the correct default value, in case it can be measured in time
+	 * units or expressed as a percentage.
+	 * @param value The prospective new value for the setting.
+	 */
+	typedef void GetDefaultValueCallback(int32_t &value);
```
rubidium42: I would expect a function called `Get...` to return something. What's the rationale behind using a reference, instead of just returning the correct default given the current state of the (other) settings?
The only reason I can think of to use a reference is to not set the default in some cases, but that kind of defeats the purpose. Especially as it was just set to something before from the default value. Why not completely ignore the static default value and only use the result of this callback?

So, around line 2733 of settings_gui.cpp you'd get:
```
		} else if (sd->get_def_cb != nullptr) {
			value = sd->get_def_cb();
		} else {
			value = sd->def;
		}
```

src/settings_internal.h
```diff
@@ -164,6 +164,12 @@ struct IntSettingDesc : SettingDesc {
 	 * @param The new value for the setting.
 	 */
 	typedef void PostChangeCallback(int32_t value);
+	/**
+	 * A callback to set the correct default value, in case it can be measured in time
+	 * units or expressed as a percentage.
```
rubidium42: This  seems to exclude other use cases given the wording. Might you have meant this as an example?
```suggestion
	 * A callback to set the correct default value. For example a default that can be measured in time
	 * units or expressed as a percentage.
```

src/settings_internal.h
chengandre: Indeed, it makes more sense to just return the correct default value. I've just pushed an update to correct that.

src/settings_internal.h
chengandre: Yes, looks better. Since I've have change the callback to return the correct value, I will also replace "set" with "get".

rubidium42 (approved):

PR #12377, opened 2024-03-25
Fix: do not send chat to clients that have not authorized yet

rubidium42:
## Motivation / Problem

It seems weird that we check whether a client is authorized for almost all things that are being sent to the clients, except for chat messages. Yes, for shutdown the also send that to all clients but that seems way more reasonable as you're telling "I'm going, don't bother going any further" and allowing the client to show "The server shut down" instead of "Connection lost" or similar.

In any case, now you will get broadcast chat messages already when you just connect to get the game info. That seems unwanted to me, so I propose to only send chat messages when the clients have authorized as this is the minimum socket status for most other mass-sent packets. Similarly if you identified, you could already receive direct chats and company chats, even though you're not accepted to the company or game yet.

There are a few cases where it checks `>` instead of `>=` even though it's fairly clear that authorized clients should also receive that information. After all, after authorized you start downloading the map. For example: you are put in wait which is beyond authorized. Then when the wait is over, you will be moved back in authorized and suddenly not disconnection notices any more until the map download actually starts?

## Description

Add status checks to the client iterations for the chat messages and change `> STATUS_AUTHORIZED` to `>= STATUS_AUTHORIZED`.

## Checklist for review

### DIFF
--- a/src/network/network_server.cpp
+++ b/src/network/network_server.cpp
@@ -1316,7 +1316,7 @@ void NetworkServerSendChat(NetworkAction action, DestType desttype, int dest, co
 			} else {
 				/* Else find the client to send the message to */
 				for (NetworkClientSocket *cs : NetworkClientSocket::Iterate()) {
-					if (cs->client_id == (ClientID)dest) {
+					if (cs->client_id == (ClientID)dest && cs->status >= ServerNetworkGameSocketHandler::STATUS_AUTHORIZED) {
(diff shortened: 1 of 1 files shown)


glx22 (commented):
Makes sense to me, but I don't understand enough the network protocol to approve.

TrueBrain (approved):

PR #12386, opened 2024-03-27
Change: Use aspect ratios to set some widget dimensions.

PeterN:
## Motivation / Problem

When using custom font sizes outside the normal range, image-based widgets alongside text-based widgets can end up with very strange proportions.

Image-based widgets automatically scale by interface scale, but this does not help if the font sizes have been changed instead, as they are not scaled. This leads to a messy user-interface and complaints that some buttons are too small.

For example, a widget designed to be 11x12 pixels can end up being 11x24 pixels instead.

(Some players like to use 1x interface scale with much larger fonts than normal, because we gave them the ability to do that.)

image

## Description

This PR adds the ability to specify a target aspect ratio for each widget. This allows for the general shape of a widget to be maintained if the height (or width) of the widget is changed due to be automatically resized during the window layout process.

A widget designed to be 11x12 can then automatically be resized to 22x24, maintaining the shape of the widget.

image

## Limitations

Not all widgets are affected by font size, e.g. scrollbars, so these widgets stay at the selected interface scale, and they can look out of proportion. To that I say increase your interface scale...

Using interface scale as it is intended allows the interface to be scaled consistently and generally within design proportions. This PR attempts only to patch over some of the problems that occur when using font sizes that are outside of the design intentions.

Some 'virtual' widgets cannot easily apply the aspect-ratio system as it is applied at the widget level. e.g. the button on drop down menus is not a real button.

## Checklist for review


src/widget.cpp
```diff
@@ -2564,24 +2564,28 @@ NWidgetLeaf::NWidgetLeaf(WidgetType tp, Colours colour, WidgetID index, uint32_t
 			this->SetFill(0, 0);
 			this->SetMinimalSize(WidgetDimensions::WD_STICKYBOX_WIDTH, WidgetDimensions::WD_CAPTION_HEIGHT);
 			this->SetDataTip(STR_NULL, STR_TOOLTIP_STICKY);
+			this->SetAspect((float)this->min_x / (float)this->min_y);
```
rubidium42: Shouldn't these cast use `static_cast<float>` to prevent future issues?

PeterN: Added an additional SetAspect() method to handle this without adding casts in lots of places.

rubidium42 (approved):

PR #12392, opened 2024-03-28
Doc: Improve the output and documentation of the font command.

merni-ns:
## Motivation / Problem
- The help text of the font command has been outdated since the default font was changed from the sprite font.
- The output format of the bare `font` command is not self-explanatory especially in its two sets of output for each size.
- Some other behaviours (like multiplying the font size by the interface scaling) were undocumented.

## Description
Clarify the documentation and the output format.

## Limitations
Perhaps things could still be more user-friendly.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -2179,10 +2179,14 @@ DEF_CONSOLE_CMD(ConFont)
 		IConsolePrint(CC_HELP, "Manage the fonts configuration.");
 		IConsolePrint(CC_HELP, "Usage 'font'.");
 		IConsolePrint(CC_HELP, "  Print out the fonts configuration.");
-		IConsolePrint(CC_HELP, "Usage 'font [medium|small|large|mono] [<name>] [<size>] [aa|noaa]'.");
+		IConsolePrint(CC_HELP, "  The \"Currently active\" configuration is the one actually in effect (after interface scaling and replacing unavailable fonts).");
+		IConsolePrint(CC_HELP, "  The \"Requested\" configuration is the one requested via console command or config file.");
+		IConsolePrint(CC_HELP, "Usage 'font [medium|small|large|mono] [<font name>] [<size>] [aa|noaa]'.");
 		IConsolePrint(CC_HELP, "  Change the configuration for a font.");
 		IConsolePrint(CC_HELP, "  Omitting an argument will keep the current value.");
-		IConsolePrint(CC_HELP, "  Set <name> to \"\" for the sprite font (size and aa have no effect on sprite font).");
+		IConsolePrint(CC_HELP, "  Set <font name> to \"\" for the default font. Note that <size> and aa/noaa have no effect if the default font is in use, and fixed defaults are used instead.");
+		IConsolePrint(CC_HELP, "  If the sprite font is enabled in Game Options, it is used instead of the default font.");
+		IConsolePrint(CC_HELP, "  The <size> is automatically multiplied by the current interface scaling.");
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


src/console_cmds.cpp
```diff
@@ -2179,10 +2179,14 @@ DEF_CONSOLE_CMD(ConFont)
 		IConsolePrint(CC_HELP, "Manage the fonts configuration.");
 		IConsolePrint(CC_HELP, "Usage 'font'.");
 		IConsolePrint(CC_HELP, "  Print out the fonts configuration.");
+		IConsolePrint(CC_HELP, "  The \"Currently active\" configuration is the one actually in effect (after interface scaling and replacing unavailable fonts).");
+		IConsolePrint(CC_HELP, "  The \"Requested\" configuration is the one requested via console command or config file.");
 		IConsolePrint(CC_HELP, "Usage 'font [medium|small|large|mono] [<name>] [<size>] [aa|noaa]'.");
 		IConsolePrint(CC_HELP, "  Change the configuration for a font.");
 		IConsolePrint(CC_HELP, "  Omitting an argument will keep the current value.");
-		IConsolePrint(CC_HELP, "  Set <name> to \"\" for the sprite font (size and aa have no effect on sprite font).");
+		IConsolePrint(CC_HELP, "  Set <name> to \"\" for the default font, or \"sprite\" for the sprite font.");
```
glx22: I don't think `"sprite"` has an effect (other than trying to load a font named "sprite" which may exist)

src/console_cmds.cpp
```diff
@@ -2179,10 +2179,14 @@ DEF_CONSOLE_CMD(ConFont)
 		IConsolePrint(CC_HELP, "Manage the fonts configuration.");
 		IConsolePrint(CC_HELP, "Usage 'font'.");
 		IConsolePrint(CC_HELP, "  Print out the fonts configuration.");
+		IConsolePrint(CC_HELP, "  The \"Currently active\" configuration is the one actually in effect (after interface scaling and replacing unavailable fonts).");
+		IConsolePrint(CC_HELP, "  The \"Requested\" configuration is the one requested via console command or config file.");
 		IConsolePrint(CC_HELP, "Usage 'font [medium|small|large|mono] [<name>] [<size>] [aa|noaa]'.");
```
ladysadie: Can you change "name" here to something less ambiguous. "Name of font" or "Font path" ??

merni-ns: What other name could it be? As for "path", you generally don't need to specify a path, and if you do specify one, my guess is that the system handles that and not openttd.

ladysadie: I can't think of a simple explanation here so I'm just going to try my best to layout the weird things I've learned when it comes to loading fonts.

The console command won't act on: 
"font medium 12 aa" 
if the default font is being used.  The user has to manually set the font name/path before they're allowed to modify the size/aa settings.  This issue is completely not listed in the help text at all.

When OpenTTD loads the default font it does so by calling "TryLoadDefaultTrueTypeFont" which will search the "baseset" directory.  "baseset" is the home directory for the OpenTTD-* font files. If you try:
"font medium OpenTTD-Sans 16 aa"
It will fail.  Calling SetFont which is what the console command code does will not search the baseset directory.  SetFont uses a different mechanism for loading the font file but, one that does not search the baseset directory.

You can use the generic name of a system font since OpenTTD has some code for dealing with Linux/Windows/etc... system fonts.  In that case a font name will suffice.

So depending on the use-case a font name might work. Unless it's an OpenTTD font in which case you will need to use the absolute path. 

Honestly the font command is a mess. It might be better just to rework the actual code and then update the docs on a much saner command.

merni-ns: > The console command won't act on:
"font medium 12 aa"
if the default font is being used. The user has to manually set the font name/path before they're allowed to modify the size/aa settings. This issue is completely not listed in the help text at all.

It is -- the existing text says `(size and aa have no effect on sprite font)` and the proposed one says `Note that <size> and aa/noaa have no effect if the default font is in use, and fixed defaults are used instead.`

> When OpenTTD loads the default font it does so by calling "TryLoadDefaultTrueTypeFont" which will search the "baseset" directory. "baseset" is the home directory for the OpenTTD-* font files. If you try:
"font medium OpenTTD-Sans 16 aa"
It will fail. [...]
So depending on the use-case a font name might work. Unless it's an OpenTTD font in which case you will need to use the absolute path.

`font medium OpenTTD-Sans 16 aa` fails, but `font medium "OpenTTD Sans" 16 aa` works, at least for me.

merni-ns: > Honestly the font command is a mess. It might be better just to rework the actual code and then update the docs on a much saner command.

Suggestions/PRs for how a "saner" command would work are welcome :) Someone is also working on a font picker in the GUI.

ladysadie: Okay I had to re-read the patch because my memory of it was totally off. I think what you have added is fine.

Yeah that person is me. :P I have a different discord display name.

It may be because I'm an idiot but, my first attempt to use this command I thought "name" was supposed to be a nickname. As if the command wanted me to name my new font settings.  Using "myfont" got me no where. Took me a minute to realize "name" meant name of a font. I'd like to see "name" to be "font name/path" but I can't/won't stop you over this.

rubidium42 (approved):

PR #12399, opened 2024-03-29
Feature: Allow base sounds set to be changed mid-game.

PeterN:
## Motivation / Problem

Baseset sounds cannot be changed while a game is running.

## Description

Allow changing base sounds during the game from the Game Options window.

This is achieved by stopping all currently playing sounds, reloading the `_original_sounds` array with the new baseset, and then updating the 'sound pool' to replace the baseset (non-NewGRF) sound information.

## Checklist for review

### DIFF
--- a/src/mixer.cpp
+++ b/src/mixer.cpp
@@ -35,6 +35,7 @@ struct MixerChannel {
 };
 
 static std::atomic<uint8_t> _active_channels;
+static std::atomic<uint8_t> _stop_channels;
 static MixerChannel _channels[8];
 static uint32_t _play_rate = 11025;
 static uint32_t _max_size = UINT_MAX;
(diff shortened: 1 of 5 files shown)


src/mixer.cpp
```diff
@@ -108,6 +109,16 @@ static void MxCloseChannel(uint8_t channel_index)
 	_active_channels.fetch_and(~(1 << channel_index), std::memory_order_release);
 }
 
+/**
+ * Close all mixer channels.
+ * This signals to the mixer that each channel should be closed even if it has not played all remaining smaples.
```
rubidium42:
```suggestion
 * This signals to the mixer that each channel should be closed even if it has not played all remaining samples.
```

src/sound.cpp
```diff
@@ -233,6 +233,34 @@ void SndCopyToPool()
...
+
+	auto set = BaseSounds::GetSet(index);
+	BaseSounds::ini_set = set->name;
+	BaseSounds::SetSet(set);
+
+	MxCloseAllChannels();
+	InitializeSound();
+
+	/* Replace baseset sounds in the pool with the updated original sounds. This is safe to do as
+	 * any sound still playing holds its own shared_ptr to the sample data. */
```
rubidium42: I can't find any `shared_ptr` related to sound. Was this patch built on top of something else?

rubidium42: Do you actually need to close channels if it's using `shared_ptr` for the sample data? Can't the samples just remain as is. So you'll let the "old" sounds run out and start "new" sounds. Only if it's simpler though.

PeterN: Yes, they can continue playing, it's more a user-expectation thing. If they aren't stopped and carry on playing, it can appear like the change has had no effect.

PeterN: Ah yes, it was. In fact without that other patch this is even safer as the sound data is owned by the mixer channel. I'll update the comment.

rubidium42 (approved):

PR #12401, opened 2024-03-31
Fix fb9d4af: use different nonces for key exchange and stream encryption

rubidium42:
## Motivation / Problem

I kinda assumed that using the same nonce for the key exchange and encryption would be fine. However, that doesn't seem to be the case...

The key exchange uses `crypto_aead_lock` and `crypto_aead_unlock`, whereas the encryption uses `crypto_aead_read` and `crypto_aead_write` which at first glance seem different but lock/unlock internally use read/write. So we cannot reuse the nonce there.

## Description

Differentiate between the nonces:
* a key exchange nonce
* an encryption nonce

And put the encryption nonce in the "enable encryption" packet.

## Checklist for review


TrueBrain:
`crypto_aead_lock` (and unlock) is a wrapper around `crypto_aead_write` to both initialize the crypto aead, and wipe it after. So normally you either use the one or the other, and not both. So you can just initialize an aead context and use `crypto_aead_write`, without ever using `crypto_aead_lock`.

https://github.com/LoupVaillant/Monocypher/blob/master/src/monocypher.c#L2931

The nonce is to ensure that if the session key is the same, something else at least is different. So the combination `(session-key, nonce)` should be unique, ideally, for-ever.
As your session-key is derived from the secret/public key of both parties, if both parties use the same secret/public key combination, the shared-secret would be the same, and as such the session-keys. So we need to send a nonce to both parties, to ensure this uniqueness constraint.

That said, you only need one nonce for server<->client communication, as the shared-secret comes in two parts. One part is used for server->client's session-key, the other part for client->server's. So having one nonce for this, is perfectly fine. In fact, it would be well advised to use a single nonce here, as this is what is commonly used for these kind of implementations. It is also what Monocypher suggests in their examples, although it is a bit hidden in words:

https://monocypher.org/manual/x25519#EXAMPLES
`Generate a pair of shared keys with your secret key and their public key (this can help nonce management for full duplex communications).`

Important to note that when using `crypto_aead_write` the encryption key (session-key) is changed every call, which ensures that every next message also has an unique `(session-key, nonce)`.

----

For the key-exchange you might be right, although I forgot how we did this exactly. If we create a new context with identical session-key, it might be wise to roll an additional nonce there. But ideally this nonce then should be sent every time a key-exchange is done. Just as a reminder: the nonce is not a secret, and can be initiated by both the server and the client.

Either way, just wanted to give you a bit of context so you can make the right call. Hopefully it removes the doubtfulness of your description :)

TrueBrain (approved):

rubidium42:
I have reduced it to one nonce for encryption. Though thinking about it some more... we could also generate 3 keys. One for key exchange, one for client-to-server encryption and one for server-to-client encryption in which case we do not need multiple nonces.

Regarding the authentication; each method creates a new nonce, so we're not reusing nonces there.

rubidium42:
Generating 3 keys with Blake2b isn't going to work as the hash is at most 64 bytes long, and we'd need 96. Staying close to the examples provided by the library is worth more than trying to find some other hash.

So two nonces it is.

PR #12402, opened 2024-03-31
Change: [YAPF] Add a cost for RVs when DriveThroughRoadStops are blocked by other stopped vehicles.

zacc:
## Motivation / Problem
(Context/Previous attempt to solve this problem: https://github.com/OpenTTD/OpenTTD/pull/12290)

The current YAPF algorithm for road vehicles treats the fill amount of a drive through road station as equal regardless of where in the stop the RV is.

The last remaining RV in a RoadStop can be loading and blocking the entrance tile for following vehicles. When the entrance is blocked, RVs will queue behind that loading vehicle causing their unloading to be delayed. 

This is most noticeable on stations with low cargo volume with several vehicles waiting to load.

## Description
Compared to my previous attempt at solving this problem, this version is YAPF only.
It checks each vehicle on the first tile of the route destination. If a vehicle's length is greater or equal to the boundary of the entrance tile then the tile will be considered to be blocked.

The change in cost weighting causes vehicles to re-route to another free tile at the same station.

I have seen a 1-2% improvement in profit on all-RV maps but of course YMMV!

## Limitations
Vehicles running nose-to-tail will still try to enter the same tile and end up queueing. Usually they have both passed the last node (or alternate route path) before the stop entrance so don't have a chance to re-route. I don't think this can't be avoided without adding some kind of reservation system similar to the train system. This queueing can also be mitigated in the player's road and roadstop design.

Potential discussion points:
* A new penalty setting for this?
* Relevance of the existing occupancy weighting and should it stay?

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_road.cpp
+++ b/src/pathfinder/yapf/yapf_road.cpp
@@ -11,6 +11,7 @@
 #include "yapf.hpp"
 #include "yapf_node_road.hpp"
 #include "../../roadstop_base.h"
+#include "../../vehicle_func.h"
 
 #include "../../safeguards.h"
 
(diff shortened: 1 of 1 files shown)


zacc:
@2TallTyler @JGRennison , this is a follow-up to my previous PR #12290 which you commented on. Appreciate any new feedback you have. Thanks.

src/pathfinder/yapf/yapf_road.cpp
```diff
@@ -54,6 +55,41 @@ class CYapfCostRoadT
...
+		DiagDirection dir; ///< The direction the vehicle is travelling.
+	};
+
+	/* Find if the tile entry direction is blocked, preventing an RV from stopping. */
+	static Vehicle *FindBlockingVehiclesInTile(Vehicle *v, void *data)
+	{
+		TileDataHelper *tdh = (TileDataHelper*)data;
+
+		/* Exclude if not a RV, not travelling in the right direction or moving */
+		if (v->type != VEH_ROAD || DirToDiagDir(v->direction) != tdh->dir || !v->IsPrimaryVehicle() || v->cur_speed > 0) return nullptr;
```
JGRennison, line 70: This appears to only detect road vehicles where the front part is on the entrance tile.
In the case where the road stop and the road vehicle are more than one tile long, this may not detect that the stop is blocked at all, even when it is.

zacc: I see, yes a 3 carriage tram could be on the middle of the 2nd stop tile and still blocking the entrance. I had intended to focus on only vehicles that were loading in the destination, but I'll adjust it to consider all vehicles and then only the length of that carriage.

zacc:
I am going to close this because since OpenTTD my benchmarks of this PR vs OpenTTD master branch have changed. Profit in v14 has improved but it is not clear to me yet if any changes to the YAPF were responsible. I'll review the gameplay with v14 first.

PR #12407, opened 2024-04-01
Codechange: Use functions to create common drop down list items, and move definitions.

PeterN:
## Motivation / Problem

Due to header dependencies, drop down list item template definitions end up being included in when compiling many source files. In many cases only the concept of a DropDownList is needed, or only the basic common item types are used.

The drop down item templates also need other headers to be included.

## Description

Add functions to create the common drop down list items, which are used instead of directly creating instances. Because drop down list items inherit from DropDownListItem, only the base class needs to be known to the callers.

This allows the complex drop down list templates to be moved to a separate header which is included in fewer source files, thereby reducing header dependencies a little.

Changing code in dropdown_common_type.h now results in 10 source files being recompiled, instead of 360+

## Limitations

This didn't actually have any effect on compilation time on my system, but does reduce the binary size very slightly, along with reducing the number of exported DropDownListItem-related symbols from 122 to 96.

## Checklist for review


src/dropdown_type.h
```diff
@@ -47,162 +44,13 @@ class DropDownListItem {
...
-using DropDownListStringItem = DropDownString<DropDownListItem>;
-using DropDownListIconItem = DropDownIcon<DropDownString<DropDownListItem>>;
-using DropDownListCheckedItem = DropDownCheck<DropDownString<DropDownListItem>>;
+/* Helper functions for commonly used drop down list items. */
+std::unique_ptr<DropDownListItem> MakeDropDownListDividerItem();
+std::unique_ptr<DropDownListItem> MakeDropDownListStringItem(StringID str, int value, bool masked = false, bool shaded = false);
+std::unique_ptr<DropDownListItem> MakeDropDownListStringItem(const std::string &str, int value, bool masked = false, bool shaded = false);
+std::unique_ptr<DropDownListItem> MakeDropDownListIconItem(SpriteID sprite, PaletteID palette, StringID str, int value, bool masked = false, bool shaded = false);
+std::unique_ptr<DropDownListItem> MakeDropDownListIconItem(const Dimension &dim, SpriteID sprite, PaletteID palette, StringID str, int value, bool masked = false, bool shaded = false);
+std::unique_ptr<DropDownListItem> MakeDropDownListCheckedItem(bool checked, StringID str, int value, bool masked = false, bool shaded = false);
```
rubidium42: Shouldn't "functions" go to the "dropdown_func.h"?

rubidium42 (approved):

PR #12409, opened 2024-04-01
Codechange: Replace Array/FixedSizeArray with std::deque.

PeterN:
## Motivation / Problem

Array/FixedSizeArray is actually a resizeable container that allocates space in chunks and allows resizing without invalidating pointers.

Lots of custom code with... interesting coding style.

There seems to be a size limit, but this doesn't ever appear to be checked.

## Description

Allocating space in chunks and resizing without invalidating pointers is also a behaviour of std::deque, so use that instead.

## Limitations

Whoa there...

Not properly tested, nor performance/memory benchmarked.

These classes are only used deep within YAPF, and there's no unit-tests for this, so it's difficult to assert that behaviour is the same.

## Checklist for review


rubidium42 (approved):
Conceptually fine by me. Haven't tested it either, though the behaviour seems to be the same.

PeterN:
I tried a quick and dirty (and inaccurate) way to test performance, and it's basically testing the performance of the performance test.

On a debug build, I get 500µs for 1,000,000 calls to CreateNewNode(), both in master and with this PR. The other users of SmallArray are called even less. So performance is a non-issue IMHO.

PR #12413, opened 2024-04-02
Codechange: Remove per font AA settings.

ladysadie:
OpenTTD will use the global AA font setting for all fonts from now on.

## Description

Console font command allow users to change aa settings per font while the settings UI only allows AA to be enabled/disabled globally.

This change will resolve that inconsistency by removing the per font aa settings and amending the font console command.  The font console command was expanded with an option to toggle AA globally. AA state is now consistent across the UI and console output. 

## Limitations

There's a check to see if the current blitter supports AA.  It's not obvious to me when the blitter would be in a state where that would fail. That's not a new quirk though.

## Checklist for review


PeterN:
I don't think the font command needs to know about the global antialiasing setting, as the toggle in Game Options window is enough. IMHO.

rubidium42 (approved):

rubidium42 (approved):

PR #12417, opened 2024-04-03
Change: replace liblzma with libarchive for OTTX savegame format

TrueBrain:
## Motivation / Problem

Current context: xz has a backdoor in 5.6.0 which exploits sshd (and only sshd, as far as we currently known). OpenTTD shouldn't be affected at all, but in an abundance of caution GitHub closed down the xz repository while investigating how big the damage actually is.

In result, vcpkg (our package manager) currently cannot install liblzma, meaning we cannot release, neither 14.0 nor any nightly.

There are two solutions here:
1) wait it out, till GitHub (and the community as a whole) wraps up their investigation, and vcpkg resolves the issue so liblzma can be installed again.
2) use an alternative library to load/save lzma (`OTTX`) savegames.

The first can take N time, where N is totally unknown. The hopes are that it will be wrapped up before weeks end, but really, nobody knows.

This implements the second.

## Description

This is a drop-in replacement. Savegames with both can be read by both.

Initially I was more curious how screwed we would be if the result of the investigation would be: you can never trust liblzma again, find another way. In other words, how vendor-locked-in are we actually. Turns out: we aren't. As I had the code to proof that anyway, it was a very small effort to actually wrap it up in a PR.

Do note, I am not saying we should be doing this, as I think we should wait at least a few more days. But just to be prepared, I also wanted to get this ball rolling. To not put all our eggs in a single basket (it has been Easter, please give me this pun ;) ).

PS: in general, libarchive might be a good library to use for our compression, as it supports many other formats. For example, also `zstd`. It would make switching and supporting them all trivial, and could collapse our whole saveload into a much simpler structure. But that is far far out of scope of this PR, as this PR just aims to replace liblzma.

## Limitations

I dropped `tar.xz` support for the TextGui. It was new code (for 14.0) anyway, and I didn't feel like replacing that code. If anyone does so, feel free.
I couldn't find any NewGRF that currently uses `.tar.xz`, but that might be my lack of searching for it. But mostly, I tried to minimize the amount of lines I had to add, as we most likely want to backport this to 14.0 (if we want this at all), where less lines is better.

Emscripten won't support OTTX savegames with this, as it depended on a custom patched liblzma support. It would require some effort to hook libarchive into Emscripten, although it is known it does work.

## Checklist for review

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -53,12 +53,12 @@ jobs:
 
         echo "::group::Install dependencies"
         sudo apt-get install -y --no-install-recommends \
+          libarchive-dev \
           liballegro4-dev \
           libcurl4-openssl-dev \
           libfontconfig-dev \
           libharfbuzz-dev \
           libicu-dev \
-          liblzma-dev \
           liblzo2-dev \
           ${{ inputs.libraries }} \
           zlib1g-dev \
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -47,12 +47,12 @@ jobs:
 
         echo "::group::Install dependencies"
         sudo apt-get install -y --no-install-recommends \
+          libarchive-dev \
           liballegro4-dev \
           libcurl4-openssl-dev \
           libfontconfig-dev \
           libharfbuzz-dev \
           libicu-dev \
-          liblzma-dev \
           liblzo2-dev \
           libsdl2-dev \
           zlib1g-dev \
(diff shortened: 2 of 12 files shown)


TrueBrain:
Meh. The idea was nice. I did a quick look over the code, and I thought I read that it uses 7zip LZMA SDK to (de)compress. I am wrong. It uses liblzma for that. It only also supports 7zip LZMA SDK for identification.

I also tried to use 7zip LZMA SDK, but really, fuck no. That SDK is not an SDK. No real documentation, no examples, just impossible to read code. So I am not going to even attempt that more than I did.

Either way, this PR is pointless for the goal set. Closing up!

jonnyawsom3:
I was going through the repo for fun, checking on the save file and screenshot compression to see if there were any gains to be had. Replacing liblzma with the 7z SDK came to mind since it _should_ be a lot faster, but then I found this PR.
Thanks for looking into it back then, shame it sounds like a nightmare to do.

PR #12420, opened 2024-04-04
Add: support for admin to change company allow lists

rubidium42:
## Motivation / Problem

#12337 only allows clients to add other clients to a company's allow list. However, it does not bring any tools for the server admin to remedy any issues created by bad actors.

## Description

Amend command for adding authorized keys to companies to also support removing authorized keys. Also do some more stringent checks on the received requests.

Add option to the "authorized_key" console command to list/add/remove authorized keys for companies as well.
This effectively meant rewriting the whole function to make it easier to also let this work for companies.

## Limitations

No UI for the less tech-savvy server owners.
No support for clients to remove other clients from companies due to griefing issues.

## Checklist for review

### DIFF
--- a/src/company_cmd.cpp
+++ b/src/company_cmd.cpp
@@ -979,16 +979,46 @@ CommandCost CmdCompanyCtrl(DoCommandFlag flags, CompanyCtrlAction cca, CompanyID
 	return CommandCost();
 }
 
+static bool ExecuteAllowListCtrlAction(CompanyAllowListCtrlAction action, Company *c, const std::string &public_key)
+{
+	switch (action) {
+		case CALCA_ADD:
+			return c->allow_list.Add(public_key);
+
+		case CALCA_REMOVE:
+			return c->allow_list.Remove(public_key);
+
+		default:
+			NOT_REACHED();
+	}
+}
+
 /**
- * Add the given public key to the allow list of this company.
+ * Add or remove the given public key to the allow list of this company.
  * @param flags Operation to perform.
- * @param public_key The public key of the client to add.
+ * @param action The action to perform.
+ * @param public_key The public key of the client to add or remove.
  * @return The cost of this operation or an error.
  */
-CommandCost CmdCompanyAddAllowList(DoCommandFlag flags, const std::string &public_key)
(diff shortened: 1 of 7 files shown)


src/console_cmds.cpp
```diff
@@ -1913,6 +1913,50 @@ static std::vector<std::pair<std::string_view, NetworkAuthorizedKeys *>> _consol
...
+		case CNAKA_ADD:
+			if (authorized_keys->Contains(authorized_key)) {
+				IConsolePrint(CC_WARNING, "Not added {} to {} as it already exists.", authorized_key, name);
+				return;
+			}
+
+			if (company == INVALID_COMPANY) {
+				authorized_keys->Add(authorized_key);
+			} else {
+				Command<CMD_COMPANY_ALLOW_LIST_CTRL>::SendNet(STR_NULL, company, CALCA_ADD, authorized_key);
```
PeterN: Is `SendNet()` correct? We use `Post()` elsewhere in this file.

rubidium42: Yes, `::SendNet` is the correct one in this case. I've added some comments as to why.

rubidium42: > Yes, `::SendNet` is the correct one in this case. I've added some comments as to why.

I was testing it incorrectly. I should've updated `_current_company` instead of `_local_company`. With changing the current company, `::Post` works just fine so I changed it to that.

src/company_cmd.cpp
```diff
@@ -980,15 +980,37 @@ CommandCost CmdCompanyCtrl(DoCommandFlag flags, CompanyCtrlAction cca, CompanyID
...
+			modifier_function = &NetworkAuthorizedKeys::Remove;
+			break;
+
+		default:
+			return CMD_ERROR;
+	}
+
 	if (flags & DC_EXEC) {
-		if (Company::Get(_current_company)->allow_list.Add(public_key)) {
+		if (std::invoke(modifier_function, c->allow_list, public_key)) {
```
PeterN: Seems quite a complex way to choose whether to Add or Remove, are more operations planned?

rubidium42: I have not planned any yet.

It would mean two switch statements; one for validation and one for actually executing the command.

PeterN (approved):

PR #12422, opened 2024-04-04
Fix: Viewport signs assume small font is smaller than normal font.

PeterN:
## Motivation / Problem

If the small font is set to a larger size than the normal font for some reason, viewport signs will be drawn incorrectly as the area marked dirty only considered the normal size font.

Setting the small font to be larger than the normal font is a weird thing to do, but it's possible, so...

image

## Description

Take the largest dimensions instead of assuming FS_NORMAL is larger than FS_SMALL.

image

## Checklist for review

### DIFF
--- a/src/viewport.cpp
+++ b/src/viewport.cpp
@@ -1314,7 +1314,7 @@ void ViewportAddString(const DrawPixelInfo *dpi, ZoomLevel small_from, const Vie
 	int right  = left + dpi->width;
 	int bottom = top + dpi->height;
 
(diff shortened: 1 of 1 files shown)


src/viewport.cpp
```diff
@@ -1490,12 +1489,16 @@ void ViewportSign::MarkDirty(ZoomLevel maxzoom) const
 {
 	Rect zoomlevels[ZOOM_LVL_END];
 
+	/* We don't know which size will be drawn, so mark the largest area dirty. */
+	const uint width = std::max(this->width_normal, this->width_small) / 2 + 1;
```
rubidium42: Maybe name this `half_width`? Reading that left and right are a width away from the centre in the code below feels a bit weird.

rubidium42 (approved):

PR #12425, opened 2024-04-04
Fix: Aircraft crash counter was too low to reach ground.

PeterN:
## Motivation / Problem

Aircraft can float above the ground when crashed as the counter limit to reach the ground is too low.

image

## Description

~~Increase to an arbitrary higher value. When the aircraft reaches the ground the counter already skips ahead, so this does not increase the time it tasks to disappear after reaching the ground.~~

The counter is now reset while falling, and when the ground is reached it skips ahead to continue the normal counter processing.

image

## Limitations

~~This could affect savegames slightly where an aircraft is up to half-way through the original crash counter limit, as it can then continue falling to the ground and will restart the final cleanup timer.~~

## Checklist for review

### DIFF
--- a/src/aircraft_cmd.cpp
+++ b/src/aircraft_cmd.cpp
@@ -1182,9 +1182,11 @@ static bool HandleCrashedAircraft(Aircraft *v)
 	if (v->crashed_counter < 500 && st == nullptr && ((v->crashed_counter % 3) == 0) ) {
 		int z = GetSlopePixelZ(Clamp(v->x_pos, 0, Map::MaxX() * TILE_SIZE), Clamp(v->y_pos, 0, Map::MaxY() * TILE_SIZE));
 		v->z_pos -= 1;
-		if (v->z_pos == z) {
+		if (v->z_pos <= z) {
 			v->crashed_counter = 500;
-			v->z_pos++;
+			v->z_pos = z + 1;
+		} else {
+			v->crashed_counter = 0;
 		}
 	}
 


rubidium42:
By the looks of it, the effects on saves are more  widespread.
* if it's in the waiting to get cleared stage, it will have to wait 4500 extra. After all, it it were at 9999 and this change comes in it has to wait using 14500 instead of 1000.
* if the counter was between 500 and 650 the crash animation will get replayed.

The first one seems the most annoying, the crash animation being played a second time might not be that troublesome.

Though... the point of this change/fix is that it removes the limitation on falling down. Why not reset the counter to 0 when `v->z_pos > z` and set it to 500 when `v->z_pos <= z` (and do `v->z_pos = z + 1` so it's always above ground)? That way the issue should be solved without any of the drawbacks or an arbitrary limitation. Or am I missing something?

rubidium42:
My mentioning that checking of `v->z_pos == z` needed to be amended was for this specific reason:
Aurillac Transport, 2035-02-24
The plane crashed, and I built a mountain at that location, so it would never reach `v->z_pos == z` (without looping). Now the plane is still crashing below the bottom of the map.
[POC.zip](https://github.com/OpenTTD/OpenTTD/files/14888316/POC.zip)

src/aircraft_cmd.cpp
```diff
@@ -1185,6 +1185,8 @@ static bool HandleCrashedAircraft(Aircraft *v)
 		if (v->z_pos == z) {
 			v->crashed_counter = 500;
 			v->z_pos++;
```
rubidium42:
```suggestion
		if (v->z_pos <= z) {
			v->crashed_counter = 500;
			v->z_pos = z + 1;
```

rubidium42 (approved):

PR #12432, opened 2024-04-05
Fix: do not use lengthof() for non C-style arrays

rubidium42:
## Motivation / Problem

Using `lengthof` of C++-containers is all kinds of bad.

## Description

Replace the cases where `lengthof` was used incorrectly with a simple substitute. It's not the proper-proper fix, but that requires migrating more to proper C++-containers and that makes backporting this a lot harder.

Replace `lengthof` with a variant that simply causes compilation failures when a non C-style array is passed. It is the way the locations with problems were found.

## Limitations

Going fully C++ containers is the better way, but baby steps.
This is just trying to do the minimum to have something that can be backported easily.

## Checklist for review


PeterN (approved):
Surprisingly few, tbh.

PR #12435, opened 2024-04-06
Change: Disallow using Action A to load sprites above the baseset unless reserved.

PeterN:
## Motivation / Problem

Using Action A above the baseset is error prone as the sprites are not fixed and can be moved around.

Any NewGRF doing so is likely to break in the future, so force it to break instead.

The NewGRF specifications have always been very clear on this, but it seems authors are not reading the spec. https://newgrf-specs.tt-wiki.net/wiki/ActionA

There are some authors not properly reading the spec, and using Action A when they should be using Action 5. These NewGRFs do appear to work for them, but there's no guarantee they will continue to do so.

## Description

When loading sprites via Action A, check if the sprite ID is either in the permitted range (below SPR_OPENTTD_BASE), or if the NewGRF has reserved some SpriteIDs via GRM reservation.

If these conditions are satisfied, the sprites are loaded at the requested location.

Otherwise the sprites are loaded at the current location so that they have no effect, and a warning debug is output. It might be possible to skip the sprites, but it seems a bit tricky as we are mid-way into the Action A here so the usual way isn't available.

This effectively "breaks" the NewGRF  now, instead of at some possible time in the future when authors have moved on to other things.

## Limitations

I know of at least one NewGRF that is affected by this change,  There may be others, but hopefully anything old enough to be affected and unmaintained will have been broken for a  while already...

There may be some other conditions I haven't considered where Action A is appropriate, but I can't think of any.

### NewGRFs known to be affected by this change

| file | reason | Status |
| ---- | ------ | ------ |
| asasignals-v5p1 | Using Action A to replace Action 5 sprites. | Author informed, has retracted NewGRF: https://github.com/j-n-dev/Asasignals |
| CZTR_Bridges-0.9.0 | Assumes contiguous ranges, and reservation is too small. | Broken before this change, and unlikely to be resolved: https://www.tt-forums.net/viewtopic.php?p=1257714#p1257714 |

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -376,7 +376,7 @@ struct GRFLocation {
 	}
 };
 
-static std::map<GRFLocation, SpriteID> _grm_sprites;
+static std::map<GRFLocation, std::pair<SpriteID, uint16_t>> _grm_sprites;
 typedef std::map<GRFLocation, std::vector<uint8_t>> GRFLineToSpriteOverride;
 static GRFLineToSpriteOverride _grf_line_to_action6_sprite_override;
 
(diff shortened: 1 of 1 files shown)


PeterN:
Example of the GrfMsg that is output:

`[2024-04-06 14:03:53] dbg: [grf:0] [41520001-asasignals-v5p1.tar/asasignals-v5p1/asasignals.grf:30] SpriteReplace: Set 0] Changing 16 sprites, beginning with 5215, above limit of 4896 and not reserved, ignoring.`

src/newgrf.cpp
```diff
@@ -7093,6 +7108,18 @@ static void SpriteReplace(ByteReader *buf)
 			i, num_sprites, first_sprite
 		);
 
+		if (first_sprite + num_sprites >= SPR_OPENTTD_BASE) {
+			/* Outside allowed range, check for GRM sprite reservations. */
+			if (!IsGRMReservedSprite(first_sprite, num_sprites)) {
+				GrfMsg(0, "SpriteReplace: Set {}] Changing {} sprites, beginning with {}, above limit of {} and not reserved, ignoring.",
```
rubidium42: I don't know what the intention of the (unopened) square bracket is, but it's inconsistent. Not sure whether the following is more along the lines of what's wanted though.
```suggestion
				GrfMsg(0, "SpriteReplace: Set {} changing {} sprites, beginning with {}, above limit of {} and not reserved, ignoring.",
```

PeterN: Oops, it's meant to be `[Set {}]` like the message on line 7107.

PeterN:
As mentioned on Discord, this also does not allow using multiple reservations as if they are contiguous. That case is possible to handle by checking for the complete reserved range (instead of each range) if we want to be more lenient.

rubidium42 (approved):

PR #12436, opened 2024-04-06
Codechange: use C++ containers for parsing the settings int lists

rubidium42:
## Motivation / Problem

In #12432 I used a hacky method to get rid of a bad `lengthof` and added a to-do. Now it's time to get rid of that to-do.

## Description

In the past you passed `ParseIntList` a pointer and a number of elements. This works fine for C-style arrays or `std::array`, but it's cumbersome with `std::vector` as you need to preallocate memory in the hope that it's right.
Furthermore some other magic was performed to clamp values to be within a specific range. This could have unexpected side effects, for example `2**32-1` would not be within the bounds of `signed long` and clamped to `2**31-1` before later being converted back to an uint32.

Now `ParseIntList` returns an optional vector with uint32s. Each of the current user is either uint32 (NewGRF parameters, base graphics parameters, resolution) or uint8 (custom music playlist).

Also simplify the code a bit by:
* using `memset` directly on the array, instead of `memset` on a temporary table and then copying values over.
* removing arbitrary 64 element limit.
* using saveload's `WriteValue` instead of our own version.

Finally the resolution's configuration suggests it's a signed integer, but it's stored in `Dimension` which uses `uint`. So, change the configuration to that as well.

## Limitations

`ParseIntList` does not support all `VarType`s, but given its limited use implementing support for that seems pointless. If it's needed at a later point, it shouldn't be too hard to add by going to parse a 64 bit integer. But out of scope for now.

I haven't put much thought into the best performance, as this is only used when reading the configuration file, so any gains would be really minimal over maintainability of the code..

## Checklist for review

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -239,24 +239,20 @@ static size_t LookupManyOfMany(const std::vector<std::string> &many, const char 
 }
 
 /**
- * Parse an integerlist string and set each found value
- * @param p the string to be parsed. Each element in the list is separated by a
- * comma or a space character
- * @param items pointer to the integerlist-array that will be filled with values
- * @param maxitems the maximum number of elements the integerlist-array has
- * @return returns the number of items found, or -1 on an error
+ * Parse a string into a vector of uint32s.
+ * @param p the string to be parsed. Each element in the list is separated by a comma or a space character
+ * @return std::optional with a vector of parsed integers. The optional is empty upon an error.
  */
-template<typename T>
-static int ParseIntList(const char *p, T *items, size_t maxitems)
+static std::optional<std::vector<uint32_t>> ParseIntList(const char *p)
 {
-	size_t n = 0; // number of items read so far
 	bool comma = false; // do we accept comma?
+	std::vector<uint32_t> result;
 
 	while (*p != '\0') {
 		switch (*p) {
 			case ',':
 				/* Do not accept multiple commas between numbers */
-				if (!comma) return -1;
+				if (!comma) return std::nullopt;
 				comma = false;
 				[[fallthrough]];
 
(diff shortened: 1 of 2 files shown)


src/settings.cpp
```diff
@@ -280,52 +275,35 @@ static int ParseIntList(const char *p, T *items, size_t maxitems)
...
 /**
  * Load parsed string-values into an integer-array (intlist)
  * @param str the string that contains the values (and will be parsed)
  * @param array pointer to the integer-arrays that will be filled
- * @param nelems the number of elements the array holds. Maximum is 64 elements
+ * @param nelems the number of elements the array holds.
  * @param type the type of elements the array holds (eg INT8, UINT16, etc.)
  * @return return true on success and false on error
  */
 static bool LoadIntList(const char *str, void *array, int nelems, VarType type)
```
LordAro, line 291: how easy would it be to get rid of the void* out param and casting to char* at the same time?

rubidium42: The `void*` comes from `GetVariableAddress`, which you already pass a `void*` to so arguably that would need to be changed as well which basically entails rewriting the whole variable access bit of the save-load system.

I would expect that touches thousands if not tens of thousands lines of code. It'd probably be worth it in the long run, as we can get rid of defining the types of the variables but alas... a hell of a lot of work.

PeterN:
I would prefer to remove the C-style array pointer and memset, etc...

Possible to pass a `std::span` and use `std::fill` to empty it.

rubidium42:
> I would prefer to remove the C-style array pointer and memset, etc...

So would I! But that's way outside of the scope of this PR.
I do not think that the caller of `LoadIntList` should do a switch on `VarType`, cast the `void*` it got from `GetVariableAddress` to create the appropriate pointer type to then create a `std::span` so you can later `std::fill` it. That seems more way more convoluted than just using `memset`.

PeterN:
Oh it comes from that :(

PeterN (approved):

PR #12437, opened 2024-04-06
Add German town name parts

rbnagm:
I added more German town name parts to add more variety for the generated town names.

Also lower frequency of suffixes. It's questionable whether suffixes as we have them now, are good at all.


PeterN:
This would change the names of existing towns on any savegame that uses German town names, which probably isn't desirable.

rubidium42:
This change with a new game will probably be fine.
However, with existing games the names of towns will change and there is a chance that duplicate town names will occur in some games.

That's because the towns names generated by this are not stored in the save game but rather the seed, and it will retry with different seeds (for a "few" times) to generate unique names. However, now the seeds will generate different names and those could be colliding.

LordAro:
(as such we generally treat town name lists as "fixed" and any attempt to change them should be done via NewGRF rather than changing the base game)

(if the situation regarding town names in saves as described by @rubidium42 ever changes, then we may revisit that)

TrueBrain:
And in case you wonder how, with for example NML (https://github.com/OpenTTD/nml) you can create your own townname GRF, with a few examples on how to do this here:

https://github.com/2TallTyler/town_names

And if you have any questions, please find us on Discord for help, suggestions, etc. Tnx!

PR #12442, opened 2024-04-06
Codechange: Use std::accumulate to determine infrastructure total rail/road pieces.

PeterN:
## Motivation / Problem

Adding up of array contents by indexing using `lengthof` to find the bounds.

## Description

Replace with `std::reduice()` instead.

## Checklist for review

### DIFF
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -42,9 +42,7 @@ struct CompanyInfrastructure {
 	/** Get total sum of all owned track bits. */
 	uint32_t GetRailTotal() const
 	{
-		uint32_t total = 0;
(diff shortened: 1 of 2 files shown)


LordAro:
Based purely on the numbers provided by https://en.cppreference.com/w/cpp/algorithm/reduce, it looks like std::accumulate is faster? No need to go for the fancy std::reduce for something that doesn't need it

michicc:
I'd have used `std::accumulate`, as that is more explicitly named in what it does. For reduce, you have to know that the default operator is addition and not something else. But maybe you want the parallel execution policy for this? 😁

PeterN:
Changed to the older `std::accumulate()` as `std::reduce()` seems too new...

rubidium42 (approved):

PR #12444, opened 2024-04-07
Codechange: use std::size instead of lengthof for town names

rubidium42:
## Motivation / Problem

Use C++ `std::size` over C-style `lengthof` macro in the town names.

## Description

Most of the code in the town name generators do some form of "get random index into some array", where it passed `lengthof` into the function as `int`. This parameter is now changed to `size_t` and it's passing the result of `std::size`.

## Limitations

There are more places where `int` or `uint` could be replaced with `size_t`, but that's beyond the scope for this. After all, this is already big enough as it is.

## Checklist for review

### DIFF
--- a/src/townname.cpp
+++ b/src/townname.cpp
@@ -166,9 +166,9 @@ bool GenerateTownName(Randomizer &randomizer, uint32_t *townnameparts, TownNames
  * @param seed seed
  * @return seed transformed to a number from given range
  */
-static inline uint32_t SeedChance(uint8_t shift_by, int max, uint32_t seed)
+static inline uint32_t SeedChance(uint8_t shift_by, size_t max, uint32_t seed)
 {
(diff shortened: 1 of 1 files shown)


src/townname.cpp
```diff
@@ -166,9 +166,9 @@ bool GenerateTownName(Randomizer &randomizer, uint32_t *townnameparts, TownNames
  * @param seed seed
  * @return seed transformed to a number from given range
  */
-static inline uint32_t SeedChance(uint8_t shift_by, int max, uint32_t seed)
+constexpr uint32_t SeedChance(uint8_t shift_by, size_t max, uint32_t seed)
```
PeterN: Seems weird to make these constexpr, even if it is possible, are they ever used that way?

PeterN (approved):

PR #12445, opened 2024-04-07
Codechange: use single function returning std::span over two functions that return size and begin

rubidium42:
## Motivation / Problem

`cpp_lengthof` is a disaster waiting to happen.

Given `struct A { std::vector<int> b; }`, then `cpp_lengthof(A, b)` returns something like 6, since `sizeof(std::vector)` is 24 and `sizeof(int)` is 4.

With the NewGRF debug UI, `cpp_lengthof` was used to return the size of a (now) `std::array`, and with a separate function the begin of the array was returned. Now we have `std::span`, returning both in a `std::span` would be much simpler.

## Description

Replace the two separate functions with a function returning a `std::span` and by doing so remove half of the `cpp_lengthof` users.

## Limitations

The other half of the `cpp_lengthof` users still exist, but that ought to be done in a separate PR.

## Checklist for review

### DIFF
--- a/src/newgrf_debug_gui.cpp
+++ b/src/newgrf_debug_gui.cpp
@@ -181,25 +181,14 @@ public:
 	}
 
 	/**
-	 * Allows to know the size of the persistent storage.
+	 * Gets the span containing the persistent storage.
 	 * @param index Index of the item.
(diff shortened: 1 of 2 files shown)


PeterN (approved):
More robust and less code, nice.

PR #12449, opened 2024-04-07
Codechange: use range-based for loops and let count be correct count

rubidium42:
## Motivation / Problem

Most video drivers use the `endof` macro to iterate over the mapping of keys, instead of simpler range-based for loops.
Most video drivers have a mapping structure that has a count, which is consistently one too low.

## Description

Replace C-style array loop with range-based for loops.
Let the count of the `VkMapping` structures be an actual count (instead of count - 1), and use `IsInsideBS` instead of a custom version.

## Checklist for review

### DIFF
--- a/src/video/allegro_v.cpp
+++ b/src/video/allegro_v.cpp
@@ -251,8 +251,8 @@ struct AllegroVkMapping {
 	uint8_t map_to;
 };
 
-#define AS(x, z) {x, 0, z}
-#define AM(x, y, z, w) {x, y - x, z}
+#define AS(x, z) {x, 1, z}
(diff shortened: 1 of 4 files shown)


PeterN (approved):

PeterN:
Oops, I did a bad review. With this applied keypresses aren't handled properly. Characters work but special keys are not handled in the console.

PeterN:
```
-if ((uint)(sym->sym - map->vk_from) <= map->vk_count) {
+if (IsInsideBS(sym, map.vk_from, map.vk_count)) {
```

`sym->sym` to `sym` is likely... Why does that compile...

PR #12451, opened 2024-04-07
Codechange: replace last cases of endof with other (safer) constructs

rubidium42:
## Motivation / Problem

`endof` does not play nice with C++ containers. It will return something, but that's likely not correct.

## Description

Basically use `std::end` instead of `endof`.

For newgrf_airport and saveload also explicitly use `std::begin` to be more future proof.

For landscape rewrite the logic so it uses `std::none_of` with `std::begin` and `std::end`, so what the intention of the code is becomes clearer as well.

## Checklist for review

### DIFF
--- a/src/newgrf_airport.h
+++ b/src/newgrf_airport.h
@@ -130,8 +130,8 @@ struct AirportSpec {
 	/** Get the index of this spec. */
 	uint8_t GetIndex() const
 	{
-		assert(this >= specs && this < endof(specs));
-		return (uint8_t)(this - specs);
+		assert(this >= std::begin(specs) && this < std::end(specs));
+		return static_cast<uint8_t>(std::distance(std::cbegin(specs), this));
 	}
 
 	static const AirportSpec dummy; ///< The dummy airport.
--- a/src/landscape.cpp
+++ b/src/landscape.cpp
@@ -933,19 +933,17 @@ static void GenerateTerrain(int type, uint flag)
 static void CreateDesertOrRainForest(uint desert_tropic_line)
 {
 	uint update_freq = Map::Size() / 4;
-	const TileIndexDiffC *data;
 
 	for (TileIndex tile = 0; tile != Map::Size(); ++tile) {
 		if ((tile.base() % update_freq) == 0) IncreaseGeneratingWorldProgress(GWP_LANDSCAPE);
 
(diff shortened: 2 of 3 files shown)


src/newgrf_airport.h
```diff
@@ -130,8 +130,8 @@ struct AirportSpec {
 	/** Get the index of this spec. */
 	uint8_t GetIndex() const
 	{
-		assert(this >= specs && this < endof(specs));
-		return (uint8_t)(this - specs);
+		assert(this >= std::begin(specs) && this < std::end(specs));
+		return static_cast<uint8_t>(std::distance(this, std::cbegin(specs)));
```
LordAro: It's fine either way, but it feels a little odd to have `std::distance(end, start)`

```suggestion
		return static_cast<uint8_t>(std::distance(std::cbegin(specs), this));
```

src/saveload/saveload.cpp
```diff
@@ -2906,19 +2906,19 @@ static SaveOrLoadResult DoLoad(std::shared_ptr<LoadFilter> reader, bool load_che
 	if (_sl.lf->Read((uint8_t*)hdr, sizeof(hdr)) != sizeof(hdr)) SlError(STR_GAME_SAVELOAD_ERROR_FILE_NOT_READABLE);
 
 	/* see if we have any loader for this type. */
-	const SaveLoadFormat *fmt = _saveload_formats;
+	const SaveLoadFormat *fmt = std::begin(_saveload_formats);
 	for (;;) {
 		/* No loader found, treat as version 0 and use LZO format */
-		if (fmt == endof(_saveload_formats)) {
+		if (fmt == std::end(_saveload_formats)) {
```
LordAro: this loop could do with a bit of reorganisation, imo

```
for (_saveload_formats) {
   ...
}

if (not loaded) {
   assert(find(lzo_support));
   try_lzo;
}
```

?

rubidium42: It could indeed be improved, but that would be a massive refactor as it'd be better to move all the logic for choosing the `SaveLoadFormat` into a separate function. That way you can return the format when you found it, instead of extra checks. That would make it better organised, but well beyond the scope for this PR.

PeterN (approved):

PR #12456, opened 2024-04-08
Codechange: use std::span instead of pointer + length

rubidium42:
## Motivation / Problem

Why pass a pointer and length, when you can pass a `std::span`? Especially when there's another parameter between the pointer and length.

## Description

Use `std::span`.

## Checklist for review

### DIFF
--- a/src/industry_gui.cpp
+++ b/src/industry_gui.cpp
@@ -367,16 +367,18 @@ class BuildIndustryWindow : public Window {
 	 * @param prefixstr    String to use for the first item
 	 * @return A formatted raw string
 	 */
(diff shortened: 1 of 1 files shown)


src/industry_gui.cpp
```diff
@@ -367,16 +367,16 @@ class BuildIndustryWindow : public Window {
 	 * @param prefixstr    String to use for the first item
 	 * @return A formatted raw string
 	 */
-	std::string MakeCargoListString(const CargoID *cargolist, const CargoSuffix *cargo_suffix, int cargolistlen, StringID prefixstr) const
+	std::string MakeCargoListString(const std::span<const CargoID> cargolist, const CargoSuffix *cargo_suffix, StringID prefixstr) const
```
PeterN: cargo_suffix can also be a span, and it should be the same length as cargolist, IIRC.

PeterN (approved):

PR #12460, opened 2024-04-08
Fix #12459, f6a88e4: Crashes when deleting news messages.

### LINKED ISSUE #12459: [Crash]: Selling a vehicle after returning to depot
### Version of OpenTTD

since 20240404--gf6a88e40a4 (Ubuntu 20.04)

### Steps to reproduce

1. Build a depot (any depot: road, rail, boat, airport)
2. Build a vehicle
3. Send it out
4. Send vehicle back to depot
5. Sell vehicle

- This seems to have been introduced with commit (https://github.com/OpenTTD/OpenTTD/commit/f6a88e40a4540dbf58f664d0c322b39f6a929169): no crash observed with previous commit (https://github.com/OpenTTD/OpenTTD/commit/08cf106fc613c56ff8628046440e243b5dcd58c7)
- Note: this was observed on OpenTTD compiled on Ubuntu 20.04

### Upload crash files

[crash20240408205710.zip](https://github.com/OpenTTD/OpenTTD/files/14910937/crash20240408205710.zip)

PeterN:
## Motivation / Problem

As per #12459, the game can crash when a news message currently being shown in the ticker or news window is deleted.

The updated logic in f6a88e4 for deleting news messages did things in the wrong order, updating iterators before actually deleting the news message, so promptly trying to show the deleted news again...

## Description

Logic of deleting news messages is now corrected to:

1) If current or statusbar are showing the message to be deleted, move it on the next older item.
2) actually remove the news item from the list.
3) if current or statusbar news were updated, move them on to the next newer item (or close the window/stop the ticker if news is now empty.

## Checklist for review

### DIFF
--- a/src/news_gui.cpp
+++ b/src/news_gui.cpp
@@ -685,12 +685,13 @@ static bool ReadyForNextNewsItem()
 /** Move to the next ticker item */
 static void MoveToNextTickerItem()
 {
-	assert(!std::empty(_news));
-
 	/* There is no status bar, so no reason to show news;
 	 * especially important with the end game screen when
 	 * there is no status bar but possible news. */
 	if (FindWindowById(WC_STATUS_BAR, 0) == nullptr) return;
+
+	/* No news to move to. */
+	if (std::empty(_news)) return;
 
 	/* if we're not at the latest item, then move on */
 	while (_statusbar_news != std::begin(_news)) {
@@ -719,8 +720,6 @@ static void MoveToNextTickerItem()
 /** Move to the next news item */
 static void MoveToNextNewsItem()
 {
-	assert(!std::empty(_news));
-
 	/* There is no status bar, so no reason to show news;
 	 * especially important with the end game screen when
 	 * there is no status bar but possible news. */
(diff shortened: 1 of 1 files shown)


src/news_gui.cpp
```diff
@@ -755,28 +757,37 @@ static void MoveToNextNewsItem()
...
-		/* When we're the current news, go to the previous item first;
+	bool updateCurrentNews = (_forced_news == ni || _current_news == ni);
+	bool updateStatusbarNews = (_statusbar_news == ni);
+
+	if (updateCurrentNews) {
+		/* When we're the current news, go to the next older item first;
 		 * we just possibly made that the last news item. */
-		if (_current_news == ni) _current_news = (_current_news == std::begin(_news)) ? std::end(_news) : std::prev(_current_news);
+		if (_current_news == ni) ++_current_news;
+	}
```
rubidium42: You're checking `_forced_news` for being the item that's going to be removed, but it is kept as a dangling iterator.
Wouldn't it be better to ensure it's not dangling at any point? Maybe just resetting to `std::end` might be enough here? Even though the later `MoveToNextItem` will also do that.

PeterN: Done

Kuhnovic (approved):

PR #12464, opened 2024-04-09
Codechange: Use vector and iterators for NewGRF classes instead of array.

PeterN:
## Motivation / Problem

NewGRF classes have a fixed-size array which is looped through to find valid classes. This is all hidden behind an interface design to permit access only by indices, which ends up making calling code more complicated.

## Description

Replace fixed the arrays with std::vector, expose as read-only spans, and make more use of iterators.

## Limitations

Does not increase any limits... yet.

## Checklist for review

### DIFF
--- a/src/newgrf_class_func.h
+++ b/src/newgrf_class_func.h
@@ -11,35 +11,19 @@
 
 #include "table/strings.h"
 
-/** Instantiate the array. */
-template <typename Tspec, typename Tid, Tid Tmax>
-NewGRFClass<Tspec, Tid, Tmax> NewGRFClass<Tspec, Tid, Tmax>::classes[Tmax];
-
-/** Reset the class, i.e. clear everything. */
-template <typename Tspec, typename Tid, Tid Tmax>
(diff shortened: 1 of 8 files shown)


src/newgrf_class_func.h
```diff
@@ -11,51 +11,35 @@
...
-}
-
 /** Reset the classes, i.e. clear everything. */
 template <typename Tspec, typename Tid, Tid Tmax>
 void NewGRFClass<Tspec, Tid, Tmax>::Reset()
 {
-	for (Tid i = (Tid)0; i < Tmax; i++) {
-		classes[i].ResetClass();
-	}
+	classes.clear();
```
LordAro: NewGRFClass::classes ? Looks a bit like it's either missing `this->` (which it isn't) or it's a function variable otherwise

(and all the other places)

PeterN: It's static.

PeterN: Oh, you meant "Should this be ...?"

rubidium42 (approved):

PR #12468, opened 2024-04-09
Codechange: Pass std::string_view from drivers/blitters instead of char *.

PeterN:
## Motivation / Problem

Driver/blitter functions return C-style `const char *`.

## Description

Return more modern `std::string_vew` instead. Gets rid of 12% of `const char *`s.

## Limitations

`std::string_view` can't be `nullptr`, but can be `empty()`.

## Checklist for review

### DIFF
--- a/src/driver.h
+++ b/src/driver.h
@@ -23,9 +23,9 @@ public:
 	/**
 	 * Start this driver.
 	 * @param parm Parameters passed to the driver.
-	 * @return nullptr if everything went okay, otherwise an error message.
+	 * @return std::nullopt if everything went okay, otherwise an error message.
 	 */
-	virtual const char *Start(const StringList &parm) = 0;
+	virtual std::optional<std::string_view> Start(const StringList &parm) = 0;
 
 	/**
 	 * Stop this driver.
@@ -47,7 +47,7 @@ public:
 	 * Get the name of this driver.
 	 * @return The name of the driver.
 	 */
-	virtual const char *GetName() const = 0;
+	virtual std::string_view GetName() const = 0;
 };
 
 DECLARE_POSTFIX_INCREMENT(Driver::Type)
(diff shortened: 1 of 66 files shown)


src/driver.h
```diff
@@ -25,7 +25,7 @@ class Driver {
 	 * @param parm Parameters passed to the driver.
 	 * @return nullptr if everything went okay, otherwise an error message.
```
rubidium42: This comment is now out of date.

I'm wondering whether we want to use "the string is empty" as "there is no error". Might wrapping the string view in `std::optional` be a clearer API, as the existence of the string means an error?

src/settings_gui.cpp
```diff
@@ -579,7 +579,7 @@ struct GameOptionsWindow : Window {
 				break;
 
 			case WID_GO_VIDEO_DRIVER_INFO:
-				SetDParamStr(0, VideoDriver::GetInstance()->GetInfoString());
+				SetDParamStr(0, std::string{VideoDriver::GetInstance()->GetInfoString()});
```
rubidium42, line 582: Might it be useful to create a `SetDParamStr` that accepts `std::string_view` and then does this wrapping into `std::string`? That way we have one location that misuses this trick, instead of it being spread over the code. It will also make other uses of `std::string_view` easier.

rubidium42 (dismissed):

src/driver.h
glx22: I too would prefer `std::optional` here.

src/settings_gui.cpp
PeterN: If I remember correctly, the issue with that is overloading between `const char *`, `std::string` and `std::string_view` breaks things, although if we always own strings now probably the `const char *` version could be removed.

PeterN:
Now a bit larger as some of the functions called from `Start()` returned `const char *` as well, so these have also been updated as converting trying to convert `nullptr` into `std::string_view` is bad.

rubidium42 (approved):

PR #12475, opened 2024-04-10
Codechange: GetOpt refactor

rubidium42:
## Motivation / Problem

`GetOpt` is the bit of code that handles the option parsing from the command line.
This uses all kinds of C-style constructs, like pointers plus lengths. When we want to use it in a more C++ environment, this quickly becomes troublesome.

## Description

Use designated initializers instead of a load of macros to construct the option data. For OpenTTD itself even fill a vector as writing down complete tables with massive amounts of copied data is not useful; there are three variants of parameters that all have just a short name.

Use a span instead of an array with an end-of-table sentinel for the option data, and use ranged for loops to iterate them.
Use a span instead of an array and count for the (remaining) arguments.
Replace a `goto` with a call to a function.

## Limitations

From the OSX/Unix/Windows mains we should actually be sending a span. However, to make proper use of that for Windows quite a few changes are required, making this PR bigger. So that improvement will be another PR.

There might be bits that are considered C-like, such as using `char*` instead of `std::string` or `std::string_view`. I do not think those changes make the code more readable, especially as we'd need to construct vectors of strings/views from the command line arguments before continuing. It can definitely be done, but for the sake of the reviewer I left that to another PR.

## Checklist for review

### DIFF
--- a/src/misc/getoptdata.cpp
+++ b/src/misc/getoptdata.cpp
@@ -21,15 +21,14 @@
  */
 int GetOptData::GetOpt()
 {
-	char *s = this->cont;
+	const char *s = this->cont;
 	if (s == nullptr) {
-		if (this->numleft == 0) return -1; // No arguments left -> finished.
+		if (this->arguments.empty()) return -1; // No arguments left -> finished.
 
-		s = this->argv[0];
+		s = this->arguments[0];
 		if (*s != '-') return -1; // No leading '-' -> not an option -> finished.
 
-		this->argv++;
-		this->numleft--;
+		this->arguments = this->arguments.subspan(1);
 
 		/* Is it a long option? */
 		for (auto &option : this->options) {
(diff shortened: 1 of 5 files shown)


src/misc/getoptdata.cpp
```diff
@@ -21,63 +21,62 @@
...
+		this->arguments = this->arguments.subspan(1);
 
 		/* Is it a long option? */
-		for (odata = this->options; odata->flags != ODF_END; odata++) {
-			if (odata->longname != nullptr && !strcmp(odata->longname, s)) { // Long options always use the entire argument.
+		for (auto &option : this->options) {
+			if (option.longname != nullptr && !strcmp(option.longname, s)) { // Long options always use the entire argument.
 				this->cont = nullptr;
-				goto set_optval;
+				return GetOpt(option);
```
glx22:
```suggestion
				return this->GetOpt(option);
```

src/misc/getoptdata.cpp
```diff
@@ -21,63 +21,62 @@
...
-					if (odata->flags == ODF_OPTIONAL_VALUE && this->argv[0][0] == '-') return odata->id;
-
-					this->opt = this->argv[0]; // Next argument is the option value.
-					this->argv++;
-					this->numleft--;
-					return odata->id;
-
-				default: NOT_REACHED();
-			}
+			return GetOpt(option);
```
glx22:
```suggestion
			return this->GetOpt(option);
```

glx22 (approved):

PR #12478, opened 2024-04-10
Fix #12477: Use std::filesystem::rename instead of Windows API call.

### LINKED ISSUE #12477: [Crash]: changing setting when running OpenTTD from DropBox
### Version of OpenTTD

Windows 10, x64 14.0 RC3

### Steps to reproduce

I start the game.

 I go into NEW GAME.

I select AI SETTINGS

I click on the arrows that change the MAXIMUM NO OF COMPETITORS.

(it's set to random AI)

I get 

**_Assertion failed at line 2874 of D:\\a\\OpenTTD\\OpenTTD\\src\\window.cpp: HasModalProgress() || IsLocalCompany()"_**

I have a very large newGRF folder, including downloads of tons of AIs.

### Upload crash files

[OpenTTDCrash.zip](https://github.com/OpenTTD/OpenTTD/files/14938367/OpenTTDCrash.zip)

PeterN:
## Motivation / Problem

On Windows we issue a Shell File Operation to rename the ini file on save, as the old `rename()` function does not allow overwriting existing functions. This can cause Shell-type things to happen, and our WndProc can be called again when we do not expect it.

## Description

It is now 2024, and `std::filesystem::rename()` exists, which seems to specify how renaming to an existing filename should behave.

## Limitations

Works on Linux, completely untested on Windows. Not sure how `std::filesystem::rename()` handles Unicode filenames.

## Checklist for review

### DIFF
--- a/src/ini.cpp
+++ b/src/ini.cpp
@@ -22,11 +22,7 @@
 # include <fcntl.h>
 #endif
 
-#ifdef _WIN32
-# include <windows.h>
-# include <shellapi.h>
-# include "core/mem_func.hpp"
-#endif
+#include <filesystem>
 
 #include "safeguards.h"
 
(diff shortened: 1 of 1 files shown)


src/ini.cpp
```diff
@@ -91,30 +93,11 @@ bool IniFile::SaveToDisk(const std::string &filename)
...
-	shfopt.wFunc  = FO_MOVE;
-	shfopt.fFlags = FOF_NOCONFIRMATION | FOF_NOCONFIRMMKDIR | FOF_NOERRORUI | FOF_SILENT;
-	shfopt.pFrom  = tfile_new;
-	shfopt.pTo    = tfilename;
-	SHFileOperation(&shfopt);
-#else
-	if (rename(file_new.c_str(), filename.c_str()) < 0) {
-		Debug(misc, 0, "Renaming {} to {} failed; configuration not saved", file_new, filename);
+	std::error_code ec;
+	std::filesystem::rename(file_new, filename, ec);
```
glx22: Since we use OTTD2FS for the `std::ofstream` (line 58), I guess we should also use it here.

PeterN: std::filesystem::path should be able to handle conversion, but we don't/can't generally use u8string/char8_t.

So I've put OTTD2FS in here, on the assumption it's needed.

glx22:
Works for me on windows, but I was not able to reproduce the crash in master so I can't tell if it's fixed or not.

glx22 (approved):
Managed to reproduce the crash with master, and this PR fixes it.

PR #12479, opened 2024-04-11
Fix: Signature validation did not close its file.

PeterN:
## Motivation / Problem

`_ValidateSignatureFile()` opens a `FILE *`, but never closes it.

## Description

Call `fclose()` after the file is read into memory.

## Checklist for review

### DIFF
--- a/src/signature.cpp
+++ b/src/signature.cpp
@@ -205,6 +205,7 @@ static bool _ValidateSignatureFile(const std::string &filename)
 
 	std::string text(filesize, '\0');
 	size_t len = fread(text.data(), filesize, 1, f);
+	FioFCloseFile(f);
 	if (len != 1) {
 		Debug(misc, 0, "Failed to validate signature: failed to read file: {}", filename);
 		return false;


src/signature.cpp
```diff
@@ -205,6 +205,7 @@ static bool _ValidateSignatureFile(const std::string &filename)
 
 	std::string text(filesize, '\0');
 	size_t len = fread(text.data(), filesize, 1, f);
+	fclose(f);
```
TrueBrain: Oops. Also: `FioFCloseFile`?

PeterN: Can do, but it doesn't do anything different, and we already often call `fclose()` instead of `FioFCloseFile()`.

I have a bigger refactor that cleans everything up, but that's too much for a quick fix.

TrueBrain: Yeah, but most of them didn't just open it with `FioFOpenFile` :P It is often `fopen`/`fclose` or `FioFOpenFile`/`FioFCloseFile`. And I know, it is doing the exact same. Just thinking about "the next person" that looks at this code :P

It all depends on how long it takes for that "bigger" refactor to land :) But if you don't mind terribly, I would just change it into the counter part of what opened the file :)

TrueBrain (approved):

PR #12480, opened 2024-04-11
Codechange: pass command line arguments as std::span to openttd_main

rubidium42:
## Motivation / Problem

Why pass a C-style span when it can be an actual span?
Why create a limited `char*[]` when that can just be a vector?

## Description

Use `std::span` for `openttd_main`.
In the Windows specific code, the `argc/argv` were extracted from a single string. Make this be extracted into a vector of `char*`, which can easily be passed as span.

## Limitations

Not using `std::string` or `std::string_view` as those do not play nice with the current further uses. Now it is a span, changing the span to be a `std::string` or `std::string_view` will become easier, but that's out of scope for this PR.

## Checklist for review

### DIFF
--- a/src/openttd.cpp
+++ b/src/openttd.cpp
@@ -502,11 +502,10 @@ static std::vector<OptionData> CreateOptions()
 
 /**
  * Main entry point for this lovely game.
- * @param argc The number of arguments passed to this game.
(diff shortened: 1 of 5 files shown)


glx22 (approved):
Windows part works for me.

PR #12483, opened 2024-04-12
Codechange: Use std::filesystem::remove/rename in settingsgen.

PeterN:
## Motivation / Problem

Use of `unlink`/`rename` in settingsgen.cpp which requires system-type includes.

## Description

Use `std::filesystem::remove`/`std::filesystem::rename` instead.

`std::filesystem::rename` specifies that renaming to an existing file will replace it, so an additional remove for Windows is not necessary.

## Checklist for review

### DIFF
--- a/src/settingsgen/settingsgen.cpp
+++ b/src/settingsgen/settingsgen.cpp
@@ -15,10 +15,7 @@
 #include "../core/mem_func.hpp"
 #include "../error_func.h"
 
-#if !defined(_WIN32) || defined(__CYGWIN__)
-#include <unistd.h>
-#include <sys/stat.h>
-#endif
+#include <filesystem>
 
 #include "../safeguards.h"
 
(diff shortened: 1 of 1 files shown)


src/settingsgen/settingsgen.cpp
```diff
@@ -493,15 +490,14 @@ int CDECL main(int argc, char *argv[])
 		CopyFile(after_file, fp);
 		fclose(fp);
 
+		std::error_code error_code;
 		if (CompareFiles(tmp_output, output_file)) {
 			/* Files are equal. tmp2.xxx is not needed. */
-			unlink(tmp_output);
+			std::filesystem::remove(tmp_output, error_code);
```
rubidium42, line 496: Should we maybe fail when removing the file fails?

rubidium42 (approved):
With or without failing for being unable to remove the temporary file, it's fine by me.

src/settingsgen/settingsgen.cpp
PeterN: Wondering if it adds anything to do so?

rubidium42: Worst case for not doing it would be potentially causing an error on the next run that the file cannot be overwritten.
On the other hand, should the CI fail due to that? I don't know. That's why I was pondering.

PeterN: I'll leave it out for now, if it is to be checked, the change should probably cover both settingsgen and strgen.

PR #12487, opened 2024-04-12
Codechange: Replace platform-specific calls with std::filesystem::last_write_time.

PeterN:
## Motivation / Problem

fios.cpp has Windows and Unix specific code to determine the modification date of a file.

## Description

Use `std::filesystem::last_write_time` instead. The modification time doesn't ever seem to be displayed, it is only used in sorting, so the values don't need to match their previous values.

## Checklist for review

### DIFF
--- a/src/fios.cpp
+++ b/src/fios.cpp
@@ -21,10 +21,7 @@
 #include "tar_type.h"
 #include <sys/stat.h>
 #include <charconv>
-
-#ifndef _WIN32
-# include <unistd.h>
-#endif /* _WIN32 */
+#include <filesystem>
 
 #include "table/strings.h"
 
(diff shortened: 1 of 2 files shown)


glx22:
Master:
image

This PR:
image

The order is messed up.

PeterN:
So this call just doesn't work on Windows?

rubidium42:
> Using milliseconds causes the uint64_t - uint64_t operation which is then placed into a int to behave differently. Worked around by using seconds instead of milliseconds.

How is converting to seconds make it not have the same issues? Okay, maybe after 49 thousand instead of 49 days, but still not a completely insane timescale to think about. Especially when the solution is relatively trivial.

Before the epoch there are issues because we use a `uint64_t` instead of just `int64_t` which my 64 bit Linux uses. So the type of `mtime` should probably be changed to `int64_t` to prevent any issues with pre-(undefined-)epoch times.
Then milliseconds or seconds will be just fine if you use `ClampTo<int32_t>` like we do when comparing for example engine costs.

rubidium42 (approved):
Works for me with saves from 1969, 1970, 1972, today, 2038, 2039 and 2300.

PR #12498, opened 2024-04-14
Fix #12497: Add workaround for motion_counter being implemented correctly.

### LINKED ISSUE #12497: [Bug]: Vehicle::motion_counter for articulated train engine parts no longer updated due to #12229
### Version of OpenTTD

Since #12229

### Expected result

All parts of an articulated train engine should have their motion_counter updated.

### Actual result

Only the head part now has its motion_counter updated.
This breaks running animations for NewGRFs such as Timberwolf's Trains.

### Steps to reproduce

See Vehicle::motion_counter for an articulated engine

PeterN:
## Motivation / Problem

#12229 stopped updating motion_counter for non-engine parts of trains, and in doing so accidentally followed the spec for NewGRF var 46, which breaks NewGRFs that used to accidentally work, but by the spec should not.

The spec has this to say:
> This variable counts the amount of motion that a vehicle has done.  It is only valid for the first vehicle in a consist (i.e. use VarAction2 type 82!)

## Description

Make var 46 return motion_counter of the first engine, regardless of self or parent scope. This means var 46 is always in sync with the head engine, and avoids further changes to when motion_counter is updated.

This technically re-breaks the spec again, but also makes things that used to work work again.

## Limitations

:shrug: 

## Checklist for review

### DIFF
--- a/src/newgrf_engine.cpp
+++ b/src/newgrf_engine.cpp
@@ -537,7 +537,7 @@ static uint32_t VehicleGetVariable(Vehicle *v, const VehicleScopeResolver *objec
 		}
 
 		case 0x46: // Motion counter
-			return v->motion_counter;
+			return v->First()->motion_counter;
 
 		case 0x47: { // Vehicle cargo info
 			/* Format: ccccwwtt


rubidium42:
I'm a bit conflicted about this. It's indeed nice to "fix" broken NewGRFs, but should that be at the cost of not being able to introduce vehicle specific motion counters at some point over consist specific ones?

PR #12510, opened 2024-04-17
Codechange: Use arc4random_buf on random_func.cpp for OpenBSD

pacoesteban:
## Motivation / Problem

This gets ride of the following warning when compiling for OpenBSD:

```
./src/core/random_func.cpp:137:3:
warning: "No cryptographically-strong random generator available; using a
fallback instead" [-W#warnings]
```

## Description

Simple update of the condition passed to the preprocessor.
We're already using this as a local patch on the OpenBSD ports tree, but would be nice if included here as it's done for other BSDs.

## Checklist for review


TrueBrain (approved):
One can't have enough BSD variant :P

Tnx for upstreaming!

PR #12516, opened 2024-04-17
Fix: Mark vehicle status bars dirty when a vehicle leaves unbunching depot

2TallTyler:
## Motivation / Problem

When a vehicle leaves its unbunching depot, it sets the next departure time for all vehicles in its shared order group. If other vehicles are also trying to leave but have been blocked by a path reservation, their status bar will show `Waiting for free path` until one successfully leaves. Then they switch back to `Waiting to unbunch` until the next departure time.

This change of status bar text doesn't mark the window dirty, so the text can be partially redrawn and jumbled.

## Description

When a vehicle leaves its unbunching depot, dirty the vehicle view windows of all other vehicles in the shared order group.

## Checklist for review

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -2538,6 +2538,7 @@ void Vehicle::LeaveUnbunchingDepot()
 		if (u->vehstatus & (VS_STOPPED | VS_CRASHED)) continue;
 
 		u->depot_unbunching_next_departure = next_departure;
+		SetWindowDirty(WC_VEHICLE_VIEW, u->index);
 	}
 }
 


src/vehicle.cpp
```diff
@@ -2539,6 +2539,8 @@ void Vehicle::LeaveUnbunchingDepot()
 
 		u->depot_unbunching_next_departure = next_departure;
 	}
+
+	SetWindowClassesDirty(WC_VEHICLE_VIEW);
```
rubidium42: Is it really necessary that the view of *all* vehicles in the game need to be dirtied? Isn't it only the vehicle that got unbunched, as the state for the others should not have changed?

Note that it's not necessarily bad to do it for a whole class when there can be multiple, and the window is not shown often. As then trying to find many windows might be more expensive than occasionally redrawing the window even when it's not needed. It's just that above you only dirty the timetable of the current vehicle.

2TallTyler: All vehicles that share orders with this one must be dirtied. I moved the call inside the loop.

We only dirty the timetable of the current vehicles because it's the only one marked "on time" -- other vehicles in the shared-order group are not changed.

PeterN (approved):

PR #12519, opened 2024-04-17
Codechange: use infinite-fast-forward when rerunning command-log

TrueBrain:
## Motivation / Problem

When replaying a desync command log, it uses fast-forward. Just this depends on a setting, and you need to know that a value of zero is special, meaning: AS FAST AS YOU CAN.

## Description

Don't depend on a setting when replaying: go as fast as you possibly can.

PS: this also removes `_ddc_fastforward` from even existing when the defines are not set.

## Checklist for review

### DIFF
--- a/src/network/network.cpp
+++ b/src/network/network.cpp
@@ -37,6 +37,9 @@
 #include "../gfx_func.h"
 #include "../error.h"
 #include "../misc_cmd.h"
+#ifdef DEBUG_DUMP_COMMANDS
+#	include "../fileio_func.h"
+#endif
 #include <charconv>
 #include <sstream>
 #include <iomanip>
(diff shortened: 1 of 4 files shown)


src/network/network.cpp
```diff
@@ -37,15 +37,21 @@
...
 #include "../safeguards.h"
 
 #ifdef DEBUG_DUMP_COMMANDS
-#include "../fileio_func.h"
-/** When running the server till the wait point, run as fast as we can! */
+/** Helper variable to make the dedicated server go fast until the (first) join.
+ * Used to load the desync debug logs, i.e. for reproducing a desync.
+ * There's basically no need to ever enable this, unless you really know what
+ * you are doing, i.e. debugging a desync.
+ * See docs/desync.txt for details. */
```
glx22, line 54: It's been renamed
```suggestion
 * See docs/desync.md for details. */
```

TrueBrain: Will do that in another PR, after this is merged. I just now only moved this block.

glx22 (approved):

PR #12523, opened 2024-04-17
Add: [NewGRF] Vehicle property to copy properties of an existing definiton.

PeterN:
## Motivation / Problem

Variants require relevant engine properties to be set for each variant, resulting in larger source file and NewGRF files. Often most of the properties should be the same. This could be achieved by making the NML tooling duplicate properties, but this does not solve the 'bloat'.

## Description

Add a property to copy properties of an existing engine definition. Any properties altered afterwards will not propagate to the copies. For flexibility, this does NOT also set the variant_id property, which should also be set after using the copy properties property properly.

So this simplifies creation of similar variants, but can also be used for distinct types.

Strings and Action 3-2-1 chains are not copied.

## Limitations

Untested.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1334,6 +1334,16 @@ static ChangeInfoResult RailVehicleChangeInfo(uint engine, int numinfo, int prop
 				SB(ei->callback_mask, 8, 8, buf->ReadByte());
 				break;
 
+			case 0x80: { // Copy existing train properties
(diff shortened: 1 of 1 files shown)


PeterN:
Apparently not useful enough for any interest.

Yozora3:
Any chance for reopening this? This is literally why half of the grf developers I know, including myself, doesn't use variants for liveries

PR #12524, opened 2024-04-17
Codechange: Use std::filesystem::directory_iterator instead of opendir.

PeterN:
## Motivation / Problem

We use `opendir`/`readdir`/`closedir` to scan for files inside directories.

This a C API so requires `char *` strings, and does not exist on Windows, so we have our own custom implementation using the Windows APIs.

C++'s take on this is `std::filesystem::directory_iterator` (which I've previously use for language file scanning), which is cross-platform and basically much nicer to use.

## Description

This changeset make use of `std::filesystem::directory_iterator`, tweaks a couple of still-specific functions (Windows does hidden files differently) and then removes all of the custom opendir/readdir/closedir implemention for Windows.

The first commit is actually #12493, which is simplified with a different call but is included here because the current implementation uses ttd_opendir.

## Limitations

Not actually tested on Windows because I don't use it.

## Checklist for review


src/os/windows/win32.cpp
```diff
@@ -192,27 +80,15 @@ void FiosGetDrives(FileList &file_list)
...
-	sb->st_mode  = (fd->dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)? S_IFDIR : S_IFREG;
-
-	return true;
-}
+	UINT sem = SetErrorMode(SEM_FAILCRITICALERRORS); // Disable 'no-disk' message box.
 
-bool FiosIsHiddenFile(const struct dirent *ent)
-{
-	return (ent->dir->fd.dwFileAttributes & (FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM)) != 0;
+	DWORD attributes = GetFileAttributes(path.string().c_str());
```
glx22:
```suggestion
	DWORD attributes = GetFileAttributes(OTTD2FS(path.string()).c_str());
```

PeterN: It's already in native format straight from the directory_iterator, so don't think that is right.

glx22: `GetFileAttributes()` wants an LPCWSTR (`const wchar_t*`) and `path.string().c_str()` is `const char*`.
```suggestion
	DWORD attributes = GetFileAttributes(path.wstring().c_str());
```
seems better.

glx22 (dismissed):

glx22 (approved):

PR #12535, opened 2024-04-19
Codechange: Use string_view in IniItem/IniGroup/IniFile.

PeterN:
## Motivation / Problem

I'm still learning. A while back I switched the Ini file system from `char *` to `std::string`, not giving any consideration to `std::string_view`. However after reviewing it, I can see there are places where `std::string`s are constructed from `char *`. These temporary strings are then either thrown away, or copied again, and... thrown away.

## Description

Use `std::string_view` in IniItem/IniGroup/IniFile. This avoids making extra copies of strings.

## Checklist for review

### DIFF
--- a/src/ini_load.cpp
+++ b/src/ini_load.cpp
@@ -20,7 +20,7 @@
  * @param parent the group we belong to
  * @param name   the name of the item
  */
-IniItem::IniItem(const std::string &name)
+IniItem::IniItem(std::string_view name)
 {
 	this->name = StrMakeValid(name);
 }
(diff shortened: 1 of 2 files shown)


src/ini_load.cpp
```diff
@@ -29,7 +29,7 @@ IniItem::IniItem(const std::string &name)
  * Replace the current value with another value.
  * @param value the value to replace with.
  */
-void IniItem::SetValue(const std::string_view value)
+void IniItem::SetValue(std::string_view value)
```
glx22, line 32: Why remove the `const`?

PeterN: std::string_view is read only, and passed by value, so const or not does not make a difference.

So I made it consistent.

glx22 (approved):

PR #12541, opened 2024-04-20
Codechange: Replace C-casts in pool functions.

PeterN:
## Motivation / Problem

C-casts in pool functions.

## Description

Replace with `static_cast` or `reinterpret_cast`.

## Checklist for review

### DIFF
--- a/src/core/pool_type.hpp
+++ b/src/core/pool_type.hpp
@@ -78,8 +78,8 @@ private:
  */
 template <class Titem, typename Tindex, size_t Tgrowth_step, size_t Tmax_size, PoolType Tpool_type = PT_NORMAL, bool Tcache = false, bool Tzero = true>
 struct Pool : PoolBase {
-	/* Ensure Tmax_size is within the bounds of Tindex. */
-	static_assert((uint64_t)(Tmax_size - 1) >> 8 * sizeof(Tindex) == 0);
+	/* Ensure the highest possible index, i.e. Tmax_size -1, is within the bounds of Tindex. */
+	static_assert(Tmax_size - 1 <= MAX_UVALUE(Tindex));
 
 	static constexpr size_t MAX_SIZE = Tmax_size; ///< Make template parameter accessible from outside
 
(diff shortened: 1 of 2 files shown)


rubidium42 (dismissed):

src/core/pool_type.hpp
```diff
@@ -79,7 +79,7 @@ struct PoolBase {
 template <class Titem, typename Tindex, size_t Tgrowth_step, size_t Tmax_size, PoolType Tpool_type = PT_NORMAL, bool Tcache = false, bool Tzero = true>
 struct Pool : PoolBase {
 	/* Ensure Tmax_size is within the bounds of Tindex. */
-	static_assert((uint64_t)(Tmax_size - 1) >> 8 * sizeof(Tindex) == 0);
+	static_assert(Tmax_size - 1 <= MAX_UVALUE(Tindex));
```
rubidium42, line 82: Lets assume `Tmax_size = 256, Tindex = uint8_t`. Previously that would have been `255 >> 8 == 0` => `true`. Now that yields `255 <= 255` => `true`.

With `Tmax_size= 257` both will fail. So, it's behaving  as it used to behave. It does bring a question about the comment as 256 is definitely not within the bounds of `uint8_t`.

rubidium42 (dismissed):

src/core/pool_type.hpp
```diff
@@ -79,7 +79,7 @@ struct PoolBase {
 template <class Titem, typename Tindex, size_t Tgrowth_step, size_t Tmax_size, PoolType Tpool_type = PT_NORMAL, bool Tcache = false, bool Tzero = true>
 struct Pool : PoolBase {
 	/* Ensure Tmax_size is within the bounds of Tindex. */
```
rubidium42:
```suggestion
	/* Ensure the highest possible index, i.e. Tmax_size - 1, is within the bounds of Tindex. */
```

src/core/pool_type.hpp
rubidium42: The comment is definitely misleading.

The client and order backup pool 'correctly' use 256 elements with `uint8_t`.
The league table pool does not, as it also allows  256 elements but... has `LeagueTableID INVALID_LEAGUE_TABLE = 0xFF` which means you can create a league table with the invalid id.

rubidium42 (approved):

PR #12545, opened 2024-04-20
Fix: allow only 255 league tables, as 255 is the invalid id sentinel

rubidium42:
## Motivation / Problem

While reviewing #12541, the maximum size check and comment for pool size were not in sync which yielded a search for what is valid.

The result: with the league table pool you can add/allocate 256 elements. These will be numbered 0 to 255. 0xFF (or 255) is also used for as sentinel for an invalid league table pool id and as such should not be allowed to be allocated by the pool.

## Description

Reduce the league table pool size by 1.

## Limitations

Might, in theory, cause some issues with scripts that can't create 256 elements any more. However, they were not allowed to change or use the 256th element afterwards, as that would be flagged as an invalid id.

## Checklist for review

### DIFF
--- a/src/league_base.h
+++ b/src/league_base.h
@@ -20,7 +20,7 @@ bool IsValidLink(Link link);
 typedef Pool<LeagueTableElement, LeagueTableElementID, 64, 64000> LeagueTableElementPool;
 extern LeagueTableElementPool _league_table_element_pool;
 
-typedef Pool<LeagueTable, LeagueTableID, 4, 256> LeagueTablePool;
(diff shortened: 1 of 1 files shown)


PeterN:
If a savegame exists with the affected ID, can it still be loaded?

rubidium42:
> If a savegame exists with the affected ID, can it still be loaded?

No. But I also doubt there are any non-synthetic vanilla saves that have 256 league tables. If there are, then we might consider adding code to handle such cases I'd say.

PeterN (approved):

PR #12547, opened 2024-04-20
Fix 952d111: Houses and industry tiles could accept incorrect cargo.

PeterN:
## Motivation / Problem

Houses and industry tiles could accept incorrect cargo.

## Description

Default cargo label was not cleared (set to CT_INVALID) when using older 3-slot acceptance properties for house and industry tiles.

Missed in #12053 and #12062.

## Limitations

Found accidentally while looking at the code, not via a problem in game, so no test case look at.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -2570,6 +2570,7 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
 					} else {
 						housespec->accepts_cargo[j] = cargo;
 					}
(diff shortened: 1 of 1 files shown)


PeterN:
This is really minor one;
* For houses, this should not be an issue as acceptance level is set to 0 if the cargo is not valid, so it won't actually accept anything.
* For industries, normally industries define cargos in the same set, so it's unlikely to trigger, but not impossible.

rubidium42 (approved):

PR #12549, opened 2024-04-20
Fix 5008706: improved scenario editor tooltips in map generation stages are out of place

rubidium42:
## Motivation / Problem

In commit 5008706 some tooltips were improved, however... these strings were reused in the map generation progress.

This meant you would go through the following stage with odd names:
- Open landscaping menu or generate a new world
- Build or generate towns
- Build or generate industries

## Description

Create new strings, with the old content, within the range of map generation strings. Also reorder the strings, so they table in the code and strings are in the same order.

## Limitations

We could, in theory, remove the table now and just add the generation class to the string instead of looking up via a table. However, that change is more involved and not necessarily back-portable.

## Checklist for review


2TallTyler (approved):
Haha, what a good bug!

PR #12552, opened 2024-04-21
Fix #12551, 8746be8: Reinstate current order test when removing road stop.

### LINKED ISSUE #12551: [Crash]: Assertion failed at line 114 of D:\\a\\OpenTTD\\OpenTTD\\src\\core\\pool_type.hpp: index < this->first_unused
### Version of OpenTTD

Nightly build 20240421,Windows

### Steps to reproduce

I don't know WHAT HAPPENED

### Upload crash files

[OpenTTD.zip](https://github.com/OpenTTD/OpenTTD/files/15050886/OpenTTD.zip)

PeterN:
## Motivation / Problem

#12144 replaced pool iteration with FindVehiclesWithOrder, however the test for current_order being OT_GOTO_STATION was erroneously removed.

## Description

Reinstate the check of current_order.

Likely fixes the cause of #12551, @JGRennison has been able to reproduce.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -2192,7 +2192,7 @@ static CommandCost RemoveRoadStop(TileIndex tile, DoCommandFlag flags, int repla
 			[](const Vehicle *v) { return v->type == VEH_ROAD; },
 			[station_id](const Order *order) { return order->IsType(OT_GOTO_STATION) && order->GetDestination() == station_id; },
 			[station_id, tile](Vehicle *v) {
-				if (v->dest_tile == tile) {
+				if (v->current_order.IsType(OT_GOTO_STATION) && v->dest_tile == tile) {
 					v->SetDestTile(v->GetOrderStationLocation(station_id));
 				}
 			}


PeterN:
This should fix the cause but not the symptoms. I'm not sure if we should do that, or how to do that.

JGRennison:
> This should fix the cause but not the symptoms. I'm not sure if we should do that, or how to do that.

Whilst it's apparently been there for a long time, the last_station_visited clearing behaviour in `GetOrderStationLocation` seems to me to be very surprising in a "Get" function of that type. Nothing about it's name or documented purpose suggests that it should do something like that. It seems a bit of a footgun.
Renaming it to make clearer that it's a destructive operation, or moving the clearing to whichever call site at which the clearing behaviour is intended might be an idea.

PeterN:
Yes, it will also skip orders if the station is invalid for the vehicle. Not really a Get operation :-)

rubidium42 (approved):
I would not bother too much about broken saves for now. If the 12551 issues keep coming after this has been fixed then it might potentially be worth considering trying to implement a workaround for broken saves.

PR #12554, opened 2024-04-21
Feature: [NewGRF] Allow fixed layout up to 256 tiles per NewGRF rail station.

PeterN:
## Motivation / Problem

Station layouts may contain more than 8 different tile types, but due to hysterical reasons, only 8 tiles can be used and stored in the map on build.

Callback 24 exists which allows to choose the fixed layout dynamically but this is still limited to 8.

Using more than this requires using callback 14, which allows more (more than 256 even) but also requires complex calculations for each station tile to be drawn.

The technical limit that originally prevented using more than 8 tile layouts on the map was removed in 2007, in commit 3dd6362bb8.

## Description

Allow using up to 256 tile layouts in property 0E or callback 24, which defines the layout to be saved into the map.

This was originally limited to 8, because station graphics above 8 referred to other station types but that was changed in 2007.

1) More efficient than using callback 14, as that needs to be checked every time a station tile is rendered.
2) The layout does not get changed when the station is changed (this may or may not be desirable!)

Using more than 256 layouts still requires callback 14.

## Limitations

Not actually tested as this requires skill at creating NewGRF stations...

## Checklist for review


frosch123:
My understanding of station tiles was, they are mostly used to control `pylons`, `wires`, `blocked`.
How are these 3 things defined now?

PeterN:
Oh yes. That was the bit I was going to add but forgot about. Whoops!

The only reason for that's what "they are mostly used to control" is because with only 8 layouts there isn't flexibility to do much else.

src/newgrf.cpp
```diff
@@ -2144,6 +2171,13 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 				AddStringForMapping(buf->ReadWord(), [statspec](StringID str) { StationClass::Get(statspec->cls_id)->name = str; });
 				break;
 
+			case 0x1E: { // Extended tile flags (replaces prop 11, 14 and 15)
+				uint16_t tiles = buf->ReadExtendedByte();
```
rubidium42: Should this limit the amount of elements to just 256? Or at least throw a warning of some sort?
After all, how much sense does it make to have thousands of tile flags when only the first 256 can be used?

TheGwyd:
[256 Tile Type Test.zip](https://github.com/user-attachments/files/16444913/256.Tile.Type.Test.zip)
I quickly put together this GRF which should test the extended limit of tile types. This only actually has 200 (100 in each direction), but I figure that should be enough to prove it works. There is a version of the station that uses var 10 and cb 24, and another version of the same station that uses var 41 and cb 14, to compare behaviour

PeterN:
Seems to work?

image

PeterN:
Quick and dirty performance test of a large screenful of 8 64x64 stations, forcing a full screen refresh every tick:

Rendering CB14 stations - ~40ms average
Rendering CB24 stations - ~27ms average

image
image

michicc (approved):
I might have missed some implication regarding the NewGRF specs, but the code itself looks very fine to me.

Please do remember the NewGRF specs tracker page and at least the NFO docs.

PR #12557, opened 2024-04-23
Revert 2408a68910: Remove work around for an MSVC bug from 17 years ago.

PeterN:
## Motivation / Problem

Odd looking code on Windows that checks if a file exists before opening it -- but only at one specific point.

Turns out this is a work around for a bug in an ancient MSVC, but the original commit links to a URL that no longer exists.

## Description

Remove the work around.

## Limitations

Maybe it still applies... but unlikely.

## Checklist for review


TrueBrain (approved):
Break all stuff!

glx22 (approved):
We no longer support VS2005, so it should be fine.

PR #12558, opened 2024-04-23
Change: [RoadStop] For DriveThroughRoadStop, calculate the occupied level based on actual vehicles in the stop in that tick.

zacc:
## Motivation / Problem

Previous attempts at solving this: #12290, #12402.

The `occupied` counter in RoadStop::Entry does not consider that free space in front a road vehicle loading in a road stop, while vacant might be inaccessible.
In turn, `GetOccupied` would under-report the occupied level to YAPF and road vehicles would drive like lemmings into road stops that didn't actually have any available space.

This change causes RoadStop tiles in front of a road vehicle that is loading to be considered to be occupied regardless of whether they have a vehicle on them or not. 

## Description
This uses a `FindVehicleOnPos` function to check vehicles on the tile when `GetOccupied` is called.

The result is slightly more accurate `cost` values in the YAPF, enough to improve the distribution of road vehicles where multiple entries to a RoadStop are available.

## Limitations

This does come at a small cost to Cargo handling and Road vehicle ticks frame rates, but I believe the improvement in YAPF performance to be worth the cost to fix `GetOccupied`.

Articulated RVs will still try to drive in to road stops that don't have enough space for their full length. This PR however sets up the change to test adding cost penalty when the available space is less than the vehicle length, in a future PR.

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_road.cpp
+++ b/src/pathfinder/yapf/yapf_road.cpp
@@ -75,11 +75,11 @@ protected:
 						/* Increase the cost for drive-through road stops */
 						cost += Yapf().PfGetSettings().road_stop_penalty;
 						DiagDirection dir = TrackdirToExitdir(trackdir);
-						if (!RoadStop::IsDriveThroughRoadStopContinuation(tile, tile - TileOffsByDiagDir(dir))) {
-							/* When we're the first road stop in a 'queue' of them we increase
-							 * cost based on the fill percentage of the whole queue. */
+						if (Yapf().PfDetectDestinationTile(tile, trackdir)){
+							/* We've found the destination file so get the occupied
+							 * level from the Entry object. */
 							const RoadStop::Entry *entry = rs->GetEntry(dir);
-							cost += entry->GetOccupied() * Yapf().PfGetSettings().road_stop_occupied_penalty / entry->GetLength();
+							cost += entry->GetOccupied(rs, dir) * Yapf().PfGetSettings().road_stop_occupied_penalty / entry->GetLength();
 						}
 					} else {
 						/* Increase cost for filled road stops */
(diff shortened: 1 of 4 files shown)


src/pathfinder/yapf/yapf_road.cpp
```diff
@@ -75,11 +75,11 @@ class CYapfCostRoadT
 						/* Increase the cost for drive-through road stops */
 						cost += Yapf().PfGetSettings().road_stop_penalty;
 						DiagDirection dir = TrackdirToExitdir(trackdir);
-						if (!RoadStop::IsDriveThroughRoadStopContinuation(tile, tile - TileOffsByDiagDir(dir))) {
-							/* When we're the first road stop in a 'queue' of them we increase
-							 * cost based on the fill percentage of the whole queue. */
+						if (Yapf().PfDetectDestinationTile(tile, trackdir)){
```
zacc, line 78: I modified this if statement so that `GetOccupied` would only be called on destination tiles.
In my testing I did not find calling it for every RoadStop along the route to add value because the cost added on line 76 is already very high and enough of a penalty.

src/roadstop.cpp
```diff
@@ -311,67 +276,88 @@ void RoadStop::Entry::Enter(const RoadVehicle *rv)
...
+/* Find the amount of free space in this tile, measured from the direction of entry. */
+Vehicle *FindTileFreeSpace(Vehicle *v, void *data)
 {
-	RoadStopEntryRebuilderHelper *rserh = (RoadStopEntryRebuilderHelper*)data;
-	/* Not a RV or not in the right direction or crashed :( */
-	if (v->type != VEH_ROAD || DirToDiagDir(v->direction) != rserh->dir || !v->IsPrimaryVehicle() || (v->vehstatus & VS_CRASHED) != 0) return nullptr;
+	TileDataHelper *tdh = (TileDataHelper*)data;
+
+	/* Exclude if not a RV, not travelling in the Trackdir direction */
+	if (v->type != VEH_ROAD || DirToDiagDir(v->direction) != tdh->dir) return nullptr;
```
zacc, line 309: This deliberately does not exclude vehicles that have cur_speed>0. 
Including vehicles that are moving (ie just entering the road stop) I found to have a big effect on helping closely following RVs to redirect to another entry of the RS.

zacc:
I am going to merge the `tile_free_space` logic from this PR into the Rebuild function in https://github.com/OpenTTD/OpenTTD/pull/12290 so that the `FindTileFreeSpace` is only run upon Rebuild rather than every GetOccupied request.

PR #12560, opened 2024-04-23
Codechange: Store validity of water regions in separate vector

Kuhnovic:
## Motivation / Problem

The WaterRegion class used to be (and to some extent still is) a relatively large class. Earlier efforts such as #12278 have already reduced the size significantly, but further improvements can be done. This is one of them.

## Description

Instead of storing the validity of a region as a boolean member, we now store the state for each water region in a std::vector<bool>, which means only a single bit is needed. 

Granted, saving 7 bits per region is not a huge saving, but it should help with data locality. `InvalidateWaterRegion` is called each time a tile is changed, which happens a lot. Having the data stored as contiguous bits instead of a member in a possibly large array should make it more cache-friendly. In theory a 64 byte cache line can contain the state of 512 regions, which is half of a 512x512 map.

_I know that `std::vector<bool>` technically isn't required to allocate contiguous memory, but it should still help even if the data is completely scrambled in memory, which it probably isn't. I'd love to use std::bitset instead but it requires that we know the size at compile time._

## Limitations

The logic becomes a bit more separated, easier to make mistakes.

## Checklist for review

### DIFF
--- a/src/pathfinder/water_regions.cpp
+++ b/src/pathfinder/water_regions.cpp
@@ -51,7 +51,6 @@ class WaterRegion
 private:
 	std::array<TWaterRegionTraversabilityBits, DIAGDIR_END> edge_traversability_bits{};
 	bool has_cross_region_aqueducts = false;
-	bool initialized = false;
 	TWaterRegionPatchLabel number_of_patches = 0; // 0 = no water, 1 = one single patch of water, etc...
 	const OrthogonalTileArea tile_area;
 	std::unique_ptr<TWaterRegionPatchLabelArray> tile_patch_labels; ///< Tile patch labels, this may be nullptr in the following trivial cases: region is invalid, region is only land (0 patches), region is only water (1 patch)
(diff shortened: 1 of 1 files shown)


src/pathfinder/water_regions.cpp
```diff
@@ -228,6 +210,7 @@ class WaterRegion
 };
 
 std::vector<WaterRegion> _water_regions;
+std::vector<bool> _is_water_region_valid;
```
glx22, line 213: It might be better to use something similar to `Pool::used_bitmap`.

PeterN: I was thinking of abstracting out a generic bitmap class, but honestly `std::vector<bool>` is a good place to start. Make it work before optimizing? :)

Kuhnovic: I'd rather stick to the std::vector<bool>. Mostly because it's the most out-of-the-box solution :)

src/pathfinder/water_regions.cpp
```diff
@@ -316,15 +301,21 @@ WaterRegionPatchDesc GetWaterRegionPatchInfo(TileIndex tile)
 void InvalidateWaterRegion(TileIndex tile)
 {
 	if (!IsValidTile(tile)) return;
-	const int water_region_index = GetWaterRegionIndex(tile);
-	_water_regions[water_region_index].Invalidate();
+
+	auto invalidateRegion = [](TileIndex tile) {
```
JGRennison: Other local lambdas in the source use snake case instead of camel. This may be worth revising for consistency.

src/pathfinder/water_regions.cpp
```diff
@@ -409,8 +400,12 @@ void VisitWaterRegionPatchNeighbors(const WaterRegionPatchDesc &water_region_pat
  */
 void AllocateWaterRegions()
 {
+	const int number_of_regions = static_cast<size_t>(GetWaterRegionMapSizeX()) * GetWaterRegionMapSizeY();
```
JGRennison: It's not really clear whether you actually want a size_t or an int here?

Kuhnovic: What is clear is that I needed more sleep when I wrote this

rubidium42 (approved):

PR #12574, opened 2024-04-25
Codechange: Make sort list function lists safer.

PeterN:
## Motivation / Problem

GUIList has a pointer only to the start of each sort/filter func list, which has the potential for UB as it is unable to validate that the selected sort or filter type is in range.

## Description

Pass a std::span to `SetSortFuncs()` and `SetFilterFuncs()` instead and check if the selected type is in range before using it.

## Limitations

Passing a temporary to `SetSortFuncs()` or `SetFilterFuncs()` will lead to bad things still. GUIList could own the list instead.

## Checklist for review

### DIFF
--- a/src/bridge_gui.cpp
+++ b/src/bridge_gui.cpp
@@ -76,7 +76,7 @@ private:
 
 	/* Constants for sorting the bridges */
 	static const StringID sorter_names[];
-	static GUIBridgeList::SortFunction * const sorter_funcs[];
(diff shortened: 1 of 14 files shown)


rubidium42 (approved):
I wouldn't worry about the sort function going away. Every case is currently `static`, and I'd reckon that anything else will quickly cause misbehaviours and/or crashes while developing.

PR #12580, opened 2024-04-26
Fix: Scrollbar scrolls to wrong position for current position.

PeterN:
## Motivation / Problem

Scrollbar::ScrollTowards() incorrectly scrolls downwards if the current position is the same as the request position, so that the position is no longer visible.

image

## Description

Only scroll downwards if the new position is greater than the current position.

image

## Limitations

~~I think I imagined this being reproduceable, but I can't seem to any more...~~

## Checklist for review

### DIFF
--- a/src/widget_type.h
+++ b/src/widget_type.h
@@ -839,7 +839,7 @@ public:
 	 */
 	void ScrollTowards(size_type position)
 	{
-		if (position < this->GetPosition()) {
+		if (position <= this->GetPosition()) {
 			/* scroll up to the item */
 			this->SetPosition(position);
 		} else if (position >= this->GetPosition() + this->GetCapacity()) {


src/widget_type.h
```diff
@@ -839,7 +839,7 @@ class Scrollbar {
 	 */
 	void ScrollTowards(size_type position)
 	{
-		if (position < this->GetPosition()) {
+		if (position <= this->GetPosition()) {
 			/* scroll up to the item */
 			this->SetPosition(position);
 		} else if (position >= this->GetPosition() + this->GetCapacity()) {
```
rubidium42, line 845: I wonder whether you're solving the right issue. Basically you're saying that sometimes you get the behaviour that you request to scroll to position X and it ends up at X + 1, right?
Won't that mean that it's being run with a capacity of 0? In the old case you'd get into this branch and then set position to `position - 0 + 1`. That would match the behaviour you were seeing.

PR #12588, opened 2024-04-28
Codechange: pass options to ShowDropDownMenu using a span

rubidium42:
## Motivation / Problem

Using a custom sentinel for end-of-container management instead of a span.

## Description

This is split into four steps:
1. `static inline` of some string list tables in GUI classes that are only in C++ files, so `std::span` can pick up their length.
2. Changing `ShowDropDownMenu` and `GetStringListWidth` to accept a span, without changing the behaviour for looking for the sentinel. This includes making some shared GUI tables `std::initializer_lists` and returning `std::span` from the `GetVehicleSorterNames` function.
3. Moving quite a few loops for getting the bounding box dimensions of a string list to a new function `GetStringListDimension`, which behaves similarly to `GetStringListWidth` but for both axes. That means, it does allow setting the font size, but it's not actually used.
4. Remove the sentinels and move to ranged for loops instead of sentinel based loops.

## Limitations

Some locations want to index into what has become `std::initializer_lists`. This class has no `operator[]`, so an indirection via `std::data` is performed. An alternative would be using something like `std::vector` instead, but that'd likely create an initializer list to fill the vector, thus having the tables twice in memory. Is there some other solution I'm missing?

This might silently break other PRs that are working with drop down lists.

## Checklist for review


PeterN (approved):
Another one that is probably in my stashes somewhere :-)

src/autoreplace_gui.cpp
```diff
@@ -369,9 +368,7 @@ class ReplaceVehicleWindow : public Window {
 
 			case WID_RV_START_REPLACE: {
 				Dimension d = GetStringBoundingBox(STR_REPLACE_VEHICLES_START);
-				for (int i = 0; _start_replace_dropdown[i] != INVALID_STRING_ID; i++) {
-					d = maxdim(d, GetStringBoundingBox(_start_replace_dropdown[i]));
-				}
+				maxdim(d, GetStringListBoundingBox(_start_replace_dropdown));
```
LordAro, line 371: Isn't there a missing `d =` ?

(Does `maxdim` need a noignore attribute?)

PR #12595, opened 2024-04-29
Feature: Add more filtering and favourites to build object/station/waypoint/roadstop windows.

PeterN:
## Motivation / Problem

Objects, Stations, Waypoints and Roadstops all support selecting add-on content from NewGRF files.

The original design is that you select from a list of classes (up to 15 could be used!), and then pick a type from a
list.

Obviously 15 wasn't enough, so at some point this was extended, and a search filter was added. This filter can only
search from the names of classes, not the names of types within.

These windows have been developed 'ad-hoc' over the years, and although they mostly share the same layout and
functionality, everything is duplicated and no code is shared between them, so changing anything has to be done at least
4 or 5 times...

Last but not least the Build Object window has some interesting design that means the window expands based on NewGRF content, and this can expand to ridiculous proportions.

## Description

Introducing the unified-code build-picker window. This replaces much of the similar code in each of the build windows
with a shared window handler. Helper classes/callbacks are used to determine per-feature specifics, but the general
function is performed by the shared code.

Along with unifying the code behing these windows, this allow news features to be added for all features all at once:

* Additional text search for types. You can now search for within classes for a specific item.
* A new "All" button that 'ignores' the class selection list (but not the class text filter) and displays all types
  together. This makes the type text search even more useful.
* A new "Used" button, which will display only types that have already been placed by the current company. When lots of
  add-on content is loaded, this makes choosing from a more limited subset much easier, as you no longer have to scroll
  through a list of 200+ classes...
* Favourites! Ctrl-click on an item in the list of types to mark it as a favourite. This then appears with other saved
  items under the new "Saved" button. The favourites are saved to a local configuration file (favs.cfg) so are
  remembered after quitting and restarting.

image

image

And this adds less lines of code than is removed...

## Limitations

This is pretty large, but I've built the commits up in a way that can be broken into smaller PRs if required.

Probably some bugs, there's a lot going on!

The way the looping works is kind of a backwards step, but it was intended to be generic. Although all the buildable stuff right now uses NewGRFClass, even within that framework things are not consistent (waypoints, bus vs truck, etc). And there's also the possibility of introducing a build picker for houses, which don't use the NewGRFClass system.

## Checklist for review

### DIFF
--- a/src/newgrf_class.h
+++ b/src/newgrf_class.h
@@ -42,6 +42,9 @@ private:
 	static void InsertDefaults();
 
 public:
+	using spec_type = Tspec;
+	using index_type = Tindex;
+
 	uint32_t global_id; ///< Global ID for class, e.g. 'DFLT', 'WAYP', etc.
 	StringID name;    ///< Name of this class.
 
@@ -67,8 +70,6 @@ public:
 	uint GetSpecCount() const { return static_cast<uint>(this->spec.size()); }
 	/** Get the number of potentially user-available specs within the class. */
 	uint GetUISpecCount() const { return this->ui_count; }
-	int GetUIFromIndex(int index) const;
-	int GetIndexFromUI(int ui_index) const;
 
 	const Tspec *GetSpec(uint index) const;
 
@@ -80,7 +81,6 @@ public:
 	static void Assign(Tspec *spec);
 	static uint GetClassCount();
 	static uint GetUIClassCount();
-	static Tindex GetUIClass(uint index);
 	static NewGRFClass *Get(Tindex class_index);
 
 	static const Tspec *GetByGrf(uint32_t grfid, uint16_t local_id);
--- a/src/newgrf_class.h
+++ b/src/newgrf_class.h
@@ -12,21 +12,31 @@
 
 #include "strings_type.h"
 
+/* Base for each type of NewGRF spec to be used with NewGRFClass. */
+template <typename Tindex>
+struct NewGRFSpecBase {
+	Tindex class_index; ///< Class index of this spec, invalid until class is allocated.
+	uint16_t index; ///< Index within class of this spec, invalid until inserted into class.
+};
 
 /**
  * Struct containing information relating to NewGRF classes for stations and airports.
  */
-template <typename Tspec, typename Tid, Tid Tmax>
+template <typename Tspec, typename Tindex, Tindex Tmax>
 class NewGRFClass {
 private:
+	/* Tspec must be of NewGRFSpecBase<Tindex>. */
+	static_assert(std::is_base_of_v<NewGRFSpecBase<Tindex>, Tspec>);
+
 	uint ui_count = 0; ///< Number of specs in this class potentially available to the user.
+	Tindex index; ///< Index of class within the list of classes.
 	std::vector<Tspec *> spec; ///< List of specifications.
 
 	/**
 	 * The actual classes.
 	 * @note This may be reallocated during initialization so pointers may be invalidated.
 	 */
-	static inline std::vector<NewGRFClass<Tspec, Tid, Tmax>> classes;
+	static inline std::vector<NewGRFClass<Tspec, Tindex, Tmax>> classes;
 
 	/** Initialise the defaults. */
 	static void InsertDefaults();
@@ -48,11 +58,11 @@ public:
 	 * Get read-only span of all classes of this type.
 	 * @return Read-only span of classes.
 	 */
-	static std::span<NewGRFClass<Tspec, Tid, Tmax> const> Classes() { return NewGRFClass::classes; }
+	static std::span<NewGRFClass<Tspec, Tindex, Tmax> const> Classes() { return NewGRFClass::classes; }
 
 	void Insert(Tspec *spec);
 
-	Tid Index() const { return static_cast<Tid>(std::distance(&*std::cbegin(NewGRFClass::classes), this)); }
+	Tindex Index() const { return this->index; }
 	/** Get the number of allocated specs within the class. */
 	uint GetSpecCount() const { return static_cast<uint>(this->spec.size()); }
 	/** Get the number of potentially user-available specs within the class. */
@@ -66,14 +76,14 @@ public:
 	bool IsUIAvailable(uint index) const;
 
 	static void Reset();
-	static Tid Allocate(uint32_t global_id);
+	static Tindex Allocate(uint32_t global_id);
 	static void Assign(Tspec *spec);
 	static uint GetClassCount();
 	static uint GetUIClassCount();
-	static Tid GetUIClass(uint index);
-	static NewGRFClass *Get(Tid cls_id);
+	static Tindex GetUIClass(uint index);
+	static NewGRFClass *Get(Tindex class_index);
 
-	static const Tspec *GetByGrf(uint32_t grfid, uint16_t local_id, int *index);
+	static const Tspec *GetByGrf(uint32_t grfid, uint16_t local_id);
 };
 
 #endif /* NEWGRF_CLASS_H */
--- a/src/newgrf_roadstop.h
+++ b/src/newgrf_roadstop.h
@@ -22,6 +22,9 @@
 /** The maximum amount of roadstops a single GRF is allowed to add */
 static const int NUM_ROADSTOPS_PER_GRF = UINT16_MAX - 1;
 
+static const uint32_t ROADSTOP_CLASS_LABEL_DEFAULT = 'DFLT';
+static const uint32_t ROADSTOP_CLASS_LABEL_WAYPOINT = 'WAYP';
+
 enum RoadStopClassID : uint16_t {
 	ROADSTOP_CLASS_BEGIN = 0, ///< The lowest valid value
 	ROADSTOP_CLASS_DFLT = 0, ///< Default road stop class.
--- a/src/newgrf_roadstop.h
+++ b/src/newgrf_roadstop.h
@@ -181,4 +181,14 @@ int AllocateSpecToRoadStop(const RoadStopSpec *statspec, BaseStation *st, bool e
 void DeallocateSpecFromRoadStop(BaseStation *st, uint8_t specindex);
 void RoadStopUpdateCachedTriggers(BaseStation *st);
 
+/**
+ * Test if a RoadStopClass is the waypoint class.
+ * @param cls RoadStopClass to test.
+ * @return true if the class is the waypoint class.
+ */
+inline bool IsWaypointClass(const RoadStopClass &cls)
+{
+	return cls.global_id == ROADSTOP_CLASS_LABEL_WAYPOINT;
+}
+
 #endif /* NEWGRF_ROADSTATION_H */
--- a/src/newgrf_roadstop.h
+++ b/src/newgrf_roadstop.h
@@ -118,7 +118,7 @@ struct RoadStopResolverObject : public ResolverObject {
 };
 
 /** Road stop specification. */
-struct RoadStopSpec {
+struct RoadStopSpec : NewGRFSpecBase<RoadStopClassID> {
 	/**
 	 * Properties related the the grf file.
 	 * NUM_CARGO real cargo plus three pseudo cargo sprite groups.
@@ -126,7 +126,6 @@ struct RoadStopSpec {
 	 * evaluating callbacks.
 	 */
 	GRFFilePropsBase<NUM_CARGO + 3> grf_prop;
-	RoadStopClassID cls_id;     ///< The class to which this spec belongs.
 	StringID name;              ///< Name of this stop
 
 	RoadStopAvailabilityType stop_type = ROADSTOPTYPE_ALL;
--- a/src/newgrf_station.h
+++ b/src/newgrf_station.h
@@ -181,6 +181,16 @@ using StationClass = NewGRFClass<StationSpec, StationClassID, STAT_CLASS_MAX>;
 
 const StationSpec *GetStationSpec(TileIndex t);
 
+/**
+ * Test if a StationClass is the waypoint class.
+ * @param cls StationClass to test.
+ * @return true if the class is the waypoint class.
+ */
+inline bool IsWaypointClass(const StationClass &cls)
+{
+	return cls.global_id == STATION_CLASS_LABEL_WAYPOINT;
+}
+
 /* Evaluate a tile's position within a station, and return the result a bitstuffed format. */
 uint32_t GetPlatformInfo(Axis axis, uint8_t tile, int platforms, int length, int x, int y, bool centred);
 
--- a/src/newgrf_station.h
+++ b/src/newgrf_station.h
@@ -79,6 +79,9 @@ struct StationResolverObject : public ResolverObject {
 	uint32_t GetDebugID() const override;
 };
 
+static const uint32_t STATION_CLASS_LABEL_DEFAULT = 'DFLT';
+static const uint32_t STATION_CLASS_LABEL_WAYPOINT = 'WAYP';
+
 enum StationClassID : uint16_t {
 	STAT_CLASS_BEGIN = 0, ///< the lowest valid value
 	STAT_CLASS_DFLT = 0, ///< Default station class.
--- a/src/newgrf_station.h
+++ b/src/newgrf_station.h
@@ -108,8 +108,8 @@ enum StationRandomTrigger {
 };
 
 /** Station specification. */
-struct StationSpec {
-	StationSpec() : cls_id(STAT_CLASS_DFLT), name(0),
+struct StationSpec : NewGRFSpecBase<StationClassID> {
+	StationSpec() : name(0),
 		disallowed_platforms(0), disallowed_lengths(0),
 		cargo_threshold(0), cargo_triggers(0),
 		callback_mask(0), flags(0), pylons(0), wires(0), blocked(0),
@@ -121,7 +121,6 @@ struct StationSpec {
 	 * evaluating callbacks.
 	 */
 	GRFFilePropsBase<NUM_CARGO + 3> grf_prop;
-	StationClassID cls_id;     ///< The class to which this spec belongs.
 	StringID name;             ///< Name of this station.
 
 	/**
--- a/src/object_gui.cpp
+++ b/src/object_gui.cpp
@@ -14,10 +14,8 @@
 #include "newgrf_object.h"
 #include "newgrf_text.h"
 #include "object.h"
-#include "querystring_gui.h"
-#include "sortlist_type.h"
-#include "stringfilter_type.h"
-#include "string_func.h"
+#include "picker_gui.h"
+#include "sound_func.h"
 #include "strings_func.h"
 #include "viewport_func.h"
 #include "tilehighlight_func.h"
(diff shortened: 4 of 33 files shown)


PeterN:
This also resolves (by ignoring truncated strings and the entire resizeable object preview concept) the following usability problems:

image

image

PeterN:
The first 3 commits are sort of unrelated and could/should be stand alone, but are certainly required for this PR.

c47f6b7d also opens the path for another requested feature which is multiple waypoint classes. With this whole PR in place, that feature becomes a one-liner to add, but I didn't want to include it here.

src/newgrf_class.h
```diff
@@ -12,21 +12,31 @@
 
 #include "strings_type.h"
 
+/* Base for each type of NewGRF spec to be used with NewGRFClass. */
+template <typename Tindex>
+struct NewGRFSpecBase {
+	Tindex class_index; ///< Class index of thie spec, invalid until class is allocated.
```
2TallTyler:
```suggestion
	Tindex class_index; ///< Class index of this spec, invalid until class is allocated.
```

src/newgrf_roadstop.h
```diff
@@ -181,4 +181,14 @@ int AllocateSpecToRoadStop(const RoadStopSpec *statspec, BaseStation *st, bool e
 void DeallocateSpecFromRoadStop(BaseStation *st, uint8_t specindex);
 void RoadStopUpdateCachedTriggers(BaseStation *st);
 
+/**
+ * Test if a RoadStopClass is the waypoint class.
+ * @param cls RoadStopClass to test.
+ * @return true if the class the waypoint class.
```
2TallTyler:
```suggestion
 * @return true if the class is the waypoint class.
 ```

File: src/newgrf_station.h [hunk 3]
```diff
@@ -181,6 +181,16 @@ using StationClass = NewGRFClass<StationSpec, StationClassID, STAT_CLASS_MAX>;
 
 const StationSpec *GetStationSpec(TileIndex t);
 
+/**
+ * Test if a StationClass is the waypoint class.
+ * @param cls StationClass to test.
+ * @return true if the class the waypoint class.
```
2TallTyler:
```suggestion
 * @return true if the class is the waypoint class.
```

src/picker_gui.h
```diff
@@ -0,0 +1,176 @@
...
+		PFI_CLASS = 1U << 0, ///< Refresh the class list.
+		PFI_TYPE = 1U << 1, ///< Refresh the type list.
+		PFI_POSITION = 1U << 2, ///< Update scroll positions.
+		PFI_VALIDATE = 1U << 3, ///< Validate selected item.
+	};
+
+	static const uint EDITBOX_MAX_SIZE = 16; ///< The maximum number of characters for the filter edit box.
+
+	bool class_picker;
+	bool type_picker;
```
2TallTyler: These confused me when reviewing their implementation, because they are not named like bools — at first I thought you meant to check `if (class_picker != nullptr)`

Maybe `use_class_picker` or `has_class_picker` would make this clearer?

src/picker_gui.cpp
```diff
@@ -0,0 +1,463 @@
...
+		case WID_PW_TYPE_ITEM: {
+			assert(this->GetWidget<NWidgetBase>(widget)->GetParentWidget<NWidgetMatrix>()->GetCurrentElement() < static_cast<int>(this->types.size()));
+			const auto &item = this->types[this->GetWidget<NWidgetBase>(widget)->GetParentWidget<NWidgetMatrix>()->GetCurrentElement()];
+
+			DrawPixelInfo tmp_dpi;
+			Rect ir = r.Shrink(WidgetDimensions::scaled.bevel);
+			if (FillDrawPixelInfo(&tmp_dpi, ir)) {
+				AutoRestoreBackup dpi_backup(_cur_dpi, &tmp_dpi);
+				int x = (ir.Width()  - ScaleSpriteTrad(64)) / 2 + ScaleSpriteTrad(31);
+				int y = (ir.Height() + ScaleSpriteTrad(58)) / 2 - ScaleSpriteTrad(31);
```
2TallTyler: Are there constants that could be used instead of these magic numbers? ("No" is a valid answer. 🙂)

src/road_gui.cpp
```diff
@@ -1093,144 +1093,137 @@ static void ShowRoadDepotPicker(Window *parent)
...
-			pos++;
+		const auto *spec = this->GetSpec(cls_id, id);
+		return IsRoadStopAvailable(spec, roadstoptype == ROADSTOP_BUS ? STATION_BUS : STATION_TRUCK);
+	}
+
+	void DrawType(int x, int y, int cls_id, int id) const override
+	{
+		const auto *spec = this->GetSpec(cls_id, id);
+		if (spec == nullptr) {
+			StationPickerDrawSprite(x, y, roadstoptype == ROADSTOP_BUS ? STATION_BUS : STATION_TRUCK, INVALID_RAILTYPE, _cur_roadtype, _roadstop_gui.orientation);
```
2TallTyler, line 1159: Looks like you copied and pasted this, but should this be `INVALID_ROADTYPE`?

```suggestion
			StationPickerDrawSprite(x, y, roadstoptype == ROADSTOP_BUS ? STATION_BUS : STATION_TRUCK, INVALID_ROADTYPE, _cur_roadtype, _roadstop_gui.orientation);
```

src/picker_gui.cpp
```diff
@@ -352,9 +363,23 @@ void PickerWindow::BuildPickerTypeList()
...
+	if (show_all) {
+		/* Reserve enough space for everything. */
+		int total = 0;
+		for (int class_index : this->classes) total += this->callbacks.GetTypeCount(class_index);
+		this->types.reserve(total);
+		/* Add types in all classes. */
+		for (int class_index : this->classes) {
+			int count = this->callbacks.GetTypeCount(class_index);
+			for (int i = 0; i < count; i++) {
+				if (this->callbacks.GetTypeName(class_index, i) == STR_NULL) continue;
```
2TallTyler: Does this ignore NewGRFs that use an empty string for the type name? (If that's even possible) Or would that be `STR_EMPTY`?

2TallTyler (commented):
This is a fantastic quality of life feature. Code looks generally good to me, and while the GUI diffs are hard to read I know you'll fix any bugs that slip past reviews. 😉

The saved feature is fantastic, although it doesn't work in emscription so I can't test the following questions without checking out and building locally. Am I reading the code correctly that the saved list is independent of what GRFs you have loaded, and the load order? So I can save my favorite stations, for example, and if I have the GRF loaded they'll show up as saved? And if the GRF is not loaded, nothing will happen? (And the saved list won't be wiped?)

I did find one bug in emscription, which is that the selected item in the picker is not updated when I filter by used or saved -- the orientation preview and name string show whatever was selected in the previous unfiltered view.

Sometimes an object in the filtered list is selected... (the wrong one)

selected

...and sometimes not.

unselected

src/widgets/picker_widget.h
```diff
@@ -0,0 +1,35 @@
...
+
+/** Widgets of the #PickerWindow class. */
+enum PickerClassWindowWidgets : WidgetID {
+	WID_PW_START = 1 << 16, ///< Dummy to ensure widgets don't overlap.
+
+	WID_PW_CLASS_SEL, ///< Stack to hide the class picker.
+	WID_PW_CLASS_FILTER, ///< Editbox filter.
+	WID_PW_CLASS_LIST, ///< List of classes.
+	WID_PW_CLASS_SCROLL, ///< Scrollbar for list of classes.
+	WID_PW_CLASS_RESIZE, ///< Class resize handle, if standalone window.
```
2TallTyler: Is this standalone window mode part of this PR, or leftover from a different iteration of the patch? (I remember seeing screenshots where the window could be split up...)

src/road_gui.cpp
PeterN: Nope, this is rabbit hole where StationPickerDrawSprite is used for both rail and road stations.

src/picker_gui.cpp
PeterN: image
I'm going with "should be but out of scope"

src/picker_gui.cpp
PeterN: Changed it to `INVALID_STRING_ID` instead, which I don't think is possible for NewGRFs to use.

src/widgets/picker_widget.h
```diff
@@ -0,0 +1,34 @@
...
+	WID_PW_TYPE_SEL, ///< Stack to hide the type picker.
+	WID_PW_TYPE_FILTER, ///< Text filter.
+	WID_PW_MODE_ALL, ///< Toggle "Show all" filter mode.
+	WID_PW_MODE_USED, ///< Toggle showing only used types.
+	WID_PW_MODE_SAVED, ///< Toggle showing only saved types.
+	WID_PW_TYPE_MATRIX, ///< Matrix with items.
+	WID_PW_TYPE_ITEM, ///< A single item.
+	WID_PW_TYPE_SCROLL, ///< Scrollbar for the matrix.
+	WID_PW_TYPE_NAME, ///< Name of selected item.
+	WID_PW_TYPE_RESIZE, ///< Type resize handle, if standalone window.
```
2TallTyler: Missed a `standalone window` reference. 🙂

PeterN:
> The saved feature is fantastic, although it doesn't work in emscription so I can't test the following questions without checking out and building locally. Am I reading the code correctly that the saved list is independent of what GRFs you have loaded, and the load order? So I can save my favorite stations, for example, and if I have the GRF loaded they'll show up as saved? And if the GRF is not loaded, nothing will happen? (And the saved list won't be wiped?)

The saved list is stored by grfid and localid, so unless the grf version is update, they are stable. They are mapped to the current loaded class and type index when the window is opened (this could be done once on load but I didn't want to add more hooks elsewhere in the code), and items not present are given the fake class and type values of -1, so they are ignored, but still retained for next time.

> I did find one bug in emscription, which is that the selected item in the picker is not updated when I filter by used or saved -- the orientation preview and name string show whatever was selected in the previous unfiltered view.

Probably a case of deciding when it is desirable to refresh the selected item. Sometimes I want to search and if I don't find what I want, clearing it will leave me where I was. But I'll have a fiddle.

PeterN:
Did some updating of my separated window version and found some issues, so have tweaked a few bits here.

* The filter state is now updated even if the class or type pickers aren't visible, as it's simpler to have it correct than to keep checking if they're in use.
* Button lowered states now get updated in OnInvalidate, so that they're synced across split windows, and now there is only one place where they are changed.

src/object_gui.cpp
```diff
@@ -34,208 +34,129 @@
...
+			list.insert(GetPickerItem(spec));
 		}
+	}
 
-		this->object_classes.Filter(this->string_filter);
-		this->object_classes.RebuildDone();
-		this->object_classes.Sort();
+	static ObjectPickerCallbacks instance;
+};
+/* static */ ObjectPickerCallbacks ObjectPickerCallbacks::instance;
```
rubidium42, line 110: Maybe `static inline`?

src/object_gui.cpp
```diff
@@ -246,41 +167,14 @@ class BuildObjectWindow : public Window {
...
-				}
-
-				/* Get the right size for the single widget based on the current spec. */
-				ObjectClass *objclass = ObjectClass::Get(_selected_object_class);
-				const ObjectSpec *spec = objclass->GetSpec(_selected_object_index);
+				/* Get the right amount of buttons based on the current spec. */
+				const ObjectClass *objclass = ObjectClass::Get(_object_gui.sel_class);
+				const ObjectSpec *spec = objclass->GetSpec(_object_gui.sel_type);
+				size.width  = ScaleGUITrad(64) + WidgetDimensions::scaled.fullbevel.Horizontal();
+				size.height = ScaleGUITrad(58) + WidgetDimensions::scaled.fullbevel.Vertical();
```
rubidium42: The 64 and 56 are used later in this class again. Would it make sense to have these as the same named constant?

src/rail_gui.cpp
```diff
@@ -929,65 +929,109 @@ static void HandleStationPlacement(TileIndex start, TileIndex end)
...
+				list.insert({sm.grfid, sm.localidx, sm.spec->class_index, sm.spec->index});
+			}
 		}
-		this->vscroll->SetCount(this->station_classes.size());
-		this->vscroll->ScrollTowards(pos);
 	}
 
+	static StationPickerCallbacks instance;
+};
+/* static */ StationPickerCallbacks StationPickerCallbacks::instance;
```
rubidium42, line 1034: Maybe `static inline`?

src/rail_gui.cpp
```diff
@@ -1235,43 +1154,8 @@ struct BuildRailStationWindow : public PickerWindowBase {
...
-				}
-				size.width = std::max(size.width, d.width + padding.width);
-				break;
-			}
-
 			case WID_BRAS_PLATFORM_DIR_X:
 			case WID_BRAS_PLATFORM_DIR_Y:
-			case WID_BRAS_IMAGE:
 				size.width  = ScaleGUITrad(64) + WidgetDimensions::scaled.fullbevel.Horizontal();
 				size.height = ScaleGUITrad(58) + WidgetDimensions::scaled.fullbevel.Vertical();
```
rubidium42: I've seen these magic numbers for the object GUI as well, might it be a shared named constant?

src/rail_gui.cpp
```diff
@@ -2018,232 +1765,115 @@ static void ShowBuildTrainDepotPicker(Window *parent)
...
 	}
 
-	void OnInvalidateData([[maybe_unused]] int data = 0, [[maybe_unused]] bool gui_scope = true) override
-	{
-		if (!gui_scope) return;
-		this->list.ForceRebuild();
-	}
+	static WaypointPickerCallbacks instance;
+};
+/* static */ WaypointPickerCallbacks WaypointPickerCallbacks::instance;
```
rubidium42, line 1844: `static inline`?

src/station_cmd.cpp
```diff
@@ -1335,8 +1335,10 @@ CommandCost CmdBuildRailStation(DoCommandFlag flags, TileIndex tile_org, RailTyp
 	if (!ValParamRailType(rt) || !IsValidAxis(axis)) return CMD_ERROR;
 
 	/* Check if the given station class is valid */
-	if ((uint)spec_class >= StationClass::GetClassCount() || spec_class == STAT_CLASS_WAYP) return CMD_ERROR;
-	if (spec_index >= StationClass::Get(spec_class)->GetSpecCount()) return CMD_ERROR;
+	if ((uint)spec_class >= StationClass::GetClassCount()) return CMD_ERROR;
```
rubidium42:
```suggestion
	if (static_cast<uint>(spec_class) >= StationClass::GetClassCount()) return CMD_ERROR;
```

src/station_cmd.cpp
```diff
@@ -1946,10 +1948,12 @@ CommandCost CmdBuildRoadStop(DoCommandFlag flags, TileIndex tile, uint8_t width,
 	bool distant_join = (station_to_join != INVALID_STATION);
 
 	/* Check if the given station class is valid */
-	if ((uint)spec_class >= RoadStopClass::GetClassCount() || spec_class == ROADSTOP_CLASS_WAYP) return CMD_ERROR;
-	if (spec_index >= RoadStopClass::Get(spec_class)->GetSpecCount()) return CMD_ERROR;
+	if ((uint)spec_class >= RoadStopClass::GetClassCount()) return CMD_ERROR;
```
rubidium42:
```suggestion
	if (static_cast<uint>(spec_class) >= RoadStopClass::GetClassCount()) return CMD_ERROR;
```

src/waypoint_cmd.cpp
```diff
@@ -177,8 +177,10 @@ CommandCost CmdBuildRailWaypoint(DoCommandFlag flags, TileIndex start_tile, Axis
 {
 	if (!IsValidAxis(axis)) return CMD_ERROR;
 	/* Check if the given station class is valid */
-	if (spec_class != STAT_CLASS_WAYP) return CMD_ERROR;
-	if (spec_index >= StationClass::Get(spec_class)->GetSpecCount()) return CMD_ERROR;
+	if ((uint)spec_class >= StationClass::GetClassCount()) return CMD_ERROR;
```
rubidium42:
```suggestion
	if (static_cast<uint>(spec_class) >= StationClass::GetClassCount()) return CMD_ERROR;
```

src/picker_gui.cpp
```diff
@@ -0,0 +1,644 @@
...
+
+			/* Resizing in X direction only at blob size, but at pixel level in Y. */
+			resize.height = 1;
+			break;
+
+		/* Type picker */
+		case WID_PW_TYPE_ITEM:
+			size.width  = ScaleGUITrad(64) + WidgetDimensions::scaled.fullbevel.Horizontal();
+			size.height = ScaleGUITrad(58) + WidgetDimensions::scaled.fullbevel.Vertical();
+			break;
```
rubidium42: I've seen these magic numbers in the object and station UIs. Are they the same?

src/road_gui.cpp
```diff
@@ -1239,86 +1251,12 @@ struct BuildRoadStationWindow : public PickerWindowBase {
 		this->PickerWindowBase::Close();
```
rubidium42: Might this need to be?
```suggestion
		this->PickerWindow::Close();
```

src/object_gui.cpp
PeterN: They're actually used all over the place. I think resolving that is a job for a separate PR.

src/object_gui.cpp
PeterN: Not possible as it's an instance of itself.

PeterN:
> Hey Peter I'm writing this from the point of view of a player who usually only uses a few add-ons with stations or objects and for whom this feature will not be particularly necessary. I'm not against it, I'm sure it will be very useful for many, but I would have a few remarks and ideas.
> 
> **1. Window size** With these filtering options the window becomes consistently large. Personally, I also use this width (5x4), but I suspect there may be a lot of players who prefer a smaller, less obstructive size (maybe the survey gives some results here?).

Don't confuse "size shown in the screenshot" with "minimum size". The minimum is now 2 preview boxes wide. Granted, it was 1 before.

> **2. "Used"** I don't know if this is really important and useful enough to make it worth increasing the minimum window size for this reason.

See 1. They're basically favourites within a single savegame.

> **3. "Saved" (read: "Favorites")** Maybe this is not the most important window for me, where I would like to have this option, but I like this feature. :) However, I would have two comments here. The first is the title "Saved", which is misleading in my opinion. I think "Favorites" would be the most explicit and obvious. The second comment is the general idea that players can mark their favorites not only 

As per 2, "used" and "saved" are basically both favourites, but "used" is automatic from the current game and saved is ones explicitly marked. That's one reason why I didn't use the word favourites. That and, indeed, it would make the labels too long.

And the rest is... waaaaaaay out of scope.

> **4. One search bar by name** I wonder if this would not be a more aesthetic solution. Next to such a bar there could be a "All", "Search in all" or "Items" button. Currently, the presented graphic shows a selection of default stations on the main list and graphics of all add-on groups. What if another group is selected? I assume the result will be the same. For me this is confusing. Maybe it would be better if, after entering the phrase, the correct stations/objects would be shown on the right, and the groups with these stations/objects would be marked on the left? If you don't enter a phrase, as in your presentation, in my opinion all groups should be selected because elements from all groups are displayed.

The left filter searches class labels, and is useful for players who like to load a lot of different NewGRFs (or play on servers with such.)

The right filter searches within with the type names, and that depends on the class that is selected (or searches All, if that is selected.)

Try the preview the see how it works rather than assuming.

PeterN:
> Why, after selecting "Default" or any other stations and pressing "All", are elements from all groups shown,

Because that's entire the purpose of the All button.
image

> When selecting "Used" or "Saved" it always connects to "All". After you turn off "All" you may not see any items and you must select one of the groups. The question is, couldn't these two options do without "All" and just always show all elements?

The rationale is you may still want to also filter your used/saved lists by class, but selecting All automatically means that just clicking on Used/Saved is at least going to show something by default (if you have anything to list at all.)

src/picker_gui.cpp
```diff
@@ -0,0 +1,644 @@
...
+	this->types.SetFiltering(this->callbacks.type_last_filtering);
+	this->types.SetSortFuncs(_type_sorter_funcs);
+	this->types.SetFilterFuncs(_type_filter_funcs);
+
+	this->FinishInitNested(this->window_number);
+
+	this->InvalidateData(PFI_CLASS | PFI_TYPE | PFI_POSITION | PFI_VALIDATE);
+}
+
+void PickerWindow::Close([[maybe_unused]] int data)
```
rubidium42: I'd wager a bet that `data` is used.
```suggestion
void PickerWindow::Close(int data)
```

src/picker_gui.cpp
```diff
@@ -0,0 +1,644 @@
...
+{
+	this->callbacks.Close(data);
+	this->PickerWindowBase::Close(data);
+}
+
+void PickerWindow::UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize)
```
rubidium42: `padding` and `resize` are used.
```suggestion
void PickerWindow::UpdateWidgetSize(WidgetID widget, Dimension &size, const Dimension &padding, [[maybe_unused]] Dimension &fill, Dimension &resize)
```

src/picker_gui.cpp
```diff
@@ -0,0 +1,644 @@
...
+}
+
+void PickerWindow::OnResize()
+{
+	if (this->has_class_picker) {
+		this->GetScrollbar(WID_PW_CLASS_SCROLL)->SetCapacityFromWidget(this, WID_PW_CLASS_LIST);
+	}
+}
+
+void PickerWindow::OnClick([[maybe_unused]] Point pt, WidgetID widget, [[maybe_unused]] int click_count)
```
rubidium42: `pt` is used.
```suggestion
void PickerWindow::OnClick(Point pt, WidgetID widget, [[maybe_unused]] int click_count)
```

src/picker_gui.cpp
```diff
@@ -0,0 +1,644 @@
...
+				this->InvalidateData(PFI_POSITION);
+			}
+			if (_settings_client.sound.click_beep) SndPlayFx(SND_15_BEEP);
+			CloseWindowById(WC_SELECT_STATION, 0);
+			break;
+		}
+	}
+}
+
+void PickerWindow::OnInvalidateData([[maybe_unused]] int data, [[maybe_unused]] bool gui_scope)
```
rubidium42:
```suggestion
void PickerWindow::OnInvalidateData(int data, bool gui_scope)
```

src/object_gui.cpp
```diff
@@ -34,208 +34,129 @@
...
 	}
 
 	void OnInit() override
 	{
-		this->object_margin = ScaleGUITrad(4);
+		this->GetWidget<NWidgetMatrix>(WID_BO_OBJECT_MATRIX)->SetCount(4);
+		this->PickerWindow::OnInit();
 	}
 
 	void UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize) override
```
rubidium42, line 148:
```suggestion
	void UpdateWidgetSize(WidgetID widget, Dimension &size, const Dimension &padding, Dimension &fill, Dimension &resize) override
```

src/object_gui.cpp
```diff
@@ -407,131 +247,80 @@ class BuildObjectWindow : public Window {
...
-
-		this->GetWidget<NWidgetMatrix>(WID_BO_OBJECT_MATRIX)->SetClicked(view_number);
-		this->GetWidget<NWidgetMatrix>(WID_BO_SELECT_MATRIX)->SetClicked(object_number);
-		this->UpdateSelectSize();
+		this->GetWidget<NWidgetMatrix>(WID_BO_OBJECT_MATRIX)->SetClicked(_object_gui.sel_view);
+		this->UpdateSelectSize(spec);
 		this->SetDirty();
 	}
 
 	void OnInvalidateData([[maybe_unused]] int data = 0, [[maybe_unused]] bool gui_scope = true) override
```
rubidium42, line 285:
```suggestion
	void OnInvalidateData(int data = 0, bool gui_scope = true) override
```

src/object_gui.cpp
```diff
@@ -407,131 +247,80 @@ class BuildObjectWindow : public Window {
...
-		this->vscroll->SetCapacityFromWidget(this, WID_BO_CLASS_LIST);
+		if ((data & PickerWindow::PFI_POSITION) != 0) {
+			const auto objclass = ObjectClass::Get(_object_gui.sel_class);
+			const auto spec = objclass->GetSpec(_object_gui.sel_type);
+			_object_gui.sel_view = std::min<int>(_object_gui.sel_view, spec->views - 1);
+			this->UpdateButtons(spec);
+		}
 	}
 
 	void OnClick([[maybe_unused]] Point pt, WidgetID widget, [[maybe_unused]] int click_count) override
```
rubidium42, line 299:
```suggestion
	void OnClick(Point pt, WidgetID widget, int click_count) override
```

src/picker_gui.h
```diff
@@ -0,0 +1,224 @@
...
+	}
+};
+
+/** Class for PickerClassWindow to collect information and retain state. */
+class PickerCallbacks {
+public:
+	explicit PickerCallbacks(const std::string &ini_group);
+	virtual ~PickerCallbacks();
+
+	virtual void Close([[maybe_unused]] int data) { }
```
rubidium42: I do not think it makes much sense annotating that the parameters are unused in here. Or is it required that both the declaration and definition are in sync? If so, then I've missed a lot. If not, then I've also not added a comment to each of the locations where it should be removed.
```suggestion
	virtual void Close(int data) { }
```

src/rail_gui.cpp
```diff
@@ -1063,138 +1090,35 @@ struct BuildRailStationWindow : public PickerWindowBase {
...
-	 * Select the specified station class.
-	 * @param station_class Station class select.
-	 */
-	void SelectOtherClass(StationClassID station_class)
-	{
-		_railstation.station_class = station_class;
+		this->PickerWindow::Close();
 	}
 
 	void OnInvalidateData([[maybe_unused]] int data = 0, [[maybe_unused]] bool gui_scope = true) override
```
rubidium42, line 1102:
```suggestion
	void OnInvalidateData(int data = 0, bool gui_scope = true) override
```

src/rail_gui.cpp
```diff
@@ -1235,54 +1159,18 @@ struct BuildRailStationWindow : public PickerWindowBase {
 	void UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize) override
```
rubidium42, line 1159:
```suggestion
	void UpdateWidgetSize(WidgetID widget, Dimension &size, const Dimension &padding, Dimension &fill, Dimension &resize) override
```

src/road_gui.cpp
```diff
@@ -1093,232 +1094,169 @@ static void ShowRoadDepotPicker(Window *parent)
...
-		for (uint i = 0; i < _roadstop_gui_settings.roadstop_count; i++) {
-			if (IsRoadStopAvailable(rs_class->GetSpec(i), st)) {
-				_roadstop_gui_settings.roadstop_type = i;
-				break;
-			}
-		}
+		this->PickerWindow::Close();
 	}
 
 	void OnInvalidateData([[maybe_unused]] int data = 0, [[maybe_unused]] bool gui_scope = true) override
```
rubidium42, line 1254:
```suggestion
	void OnInvalidateData(int data = 0, bool gui_scope = true) override
```

src/road_gui.cpp
```diff
@@ -1353,52 +1291,23 @@ struct BuildRoadStationWindow : public PickerWindowBase {
 	void UpdateWidgetSize(WidgetID widget, Dimension &size, [[maybe_unused]] const Dimension &padding, [[maybe_unused]] Dimension &fill, [[maybe_unused]] Dimension &resize) override
```
rubidium42, line 1291:
```suggestion
	void UpdateWidgetSize(WidgetID widget, Dimension &size, const Dimension &padding, Dimension &fill, Dimension &resize) override
```

rubidium42 (dismissed):

PeterN:
Fixed an issue with the SZSP plane breaking the build tram station window when no custom road stops are loaded. SZSP_VERTICAL or SZSP_HORIZONTAL needs to be set depending the parent container type. It was correct for the rail waypoint window before.

2TallTyler (approved):

PR #12596, opened 2024-04-30
Fix 2d27e8e: Update numpad keycodes for SDL2, making it usable

infinifen:
## Motivation / Problem

SDL2 changed the order of the numpad keycodes. In SDL1, they were ordered 0-9, in SDL2 they now go 1-9, then 0 at the end. This was not accounted for when constructing the SDL keycode mappings and due to the way the AM macro works it ended up binding exactly 0 keys, making the numpad unusable under SDL2 regardless of hotkeys.cfg contents.
## Description

The problem is solved by listing the key mappings individually.

## Limitations

I believe this solves the problem in all scenarios and haven't found any corner cases with the solution.

## Checklist for review

### DIFF
--- a/src/video/sdl2_v.cpp
+++ b/src/video/sdl2_v.cpp
@@ -282,8 +282,15 @@ static const SDLVkMapping _vk_mapping[] = {
 	AM_UP(SDLK_F1, SDLK_F12, WKC_F1, WKC_F12),
 
 	/* Numeric part. */
-	/* In SDL2 the order of the numpad keycodes changed so that the smallest value is SDLK_KP_1 and the largest SDLK_KP_0, we need to bind 1-9 and then 0 separately. */
-	AM(SDLK_KP_1, SDLK_KP_9, '1', '9'),
+	AS(SDLK_KP_1,        '1'),
+	AS(SDLK_KP_2,        '2'),
+	AS(SDLK_KP_3,        '3'),
+	AS(SDLK_KP_4,        '4'),
+	AS(SDLK_KP_5,        '5'),
+	AS(SDLK_KP_6,        '6'),
+	AS(SDLK_KP_7,        '7'),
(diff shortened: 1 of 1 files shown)


src/video/sdl2_v.cpp
```diff
@@ -282,7 +282,9 @@ static const SDLVkMapping _vk_mapping[] = {
 	AM_UP(SDLK_F1, SDLK_F12, WKC_F1, WKC_F12),
 
 	/* Numeric part. */
-	AM(SDLK_KP_0, SDLK_KP_9, '0', '9'),
+	/* In SDL2 the order of the numpad keycodes changed so that the smallest value is SDLK_KP_1 and the largest SDLK_KP_0, we need to bind 1-9 and then 0 separately. */
+	AM(SDLK_KP_1, SDLK_KP_9, '1', '9'),
+	AS(SDLK_KP_0,        '0'),
```
TrueBrain:
```suggestion
	/* SDL2 goes from SDLK_KP_1, ... SDLK_KP_9, SDLK_KP_0, where we go from 0, 1, ... 9. */
	AM(SDLK_KP_1, SDLK_KP_9, '1', '9'),
	AS(SDLK_KP_0,        '0'),
```

Or something. But your current comment describes a change, rather than the situation. And that means it only makes sense in the context of this PR, but not when you look at the file in 10 years :) I know, a bit of a nuance, but I know how frustrating this kind of code can be in 10 years :P

Btw, nice catch, tnx for fixing :)

PeterN: Given the order might change... should we not just list each one individually?

infinifen: I could do that, and then it would also remove the need for that comment.

TrueBrain (approved):

PR #12597, opened 2024-04-30
Codechange: use one RAII method to manage FILE*

rubidium42:
## Motivation / Problem

Preferably we'd be doing RAII with FILE*, so we do not need to manually manage them getting closed properly in all corner cases of a function.

However, OpenTTD has already sprouted two methods. One via `std::unique_ptr` and one variant where a new object is constructed on the stack to manage closing an already existing FILE*. The latter is close to RAII, but not quite... so I rather get rid of it and use the `std::unique_ptr` variant. Especially as reassigning something to the FILE* does not close the previous version, or could create a double `fclose` which the `std::unique_ptr` does not allow.

## Description

The `std::unique_ptr` variant requires template parameters for the 'deleter', i.e. the thing that releases the resources. To make this less bothersome a type alias `AutoCloseFile` is created and used at all current locations.
The `FileCloser`, the extra object to manage the lifetime of another object, is replaced by `AutoCloseFile`.
Finally rename `FileDeleter` to `FileCloser`, as it's more likely that you'd search for 'FileCloser' for some helper that automatically closes files than a deleter.

## Limitations

Using `std::unique_ptr` you'd need `.get()` to access the `FILE*`, which is somewhat ugly.

This does not add more places where `AutoCloseFile` is used, but a longer term goal should be that no more raw `fclose` calls are made on FILE*s.

## Checklist for review

### DIFF
--- a/src/dedicated.cpp
+++ b/src/dedicated.cpp
@@ -12,7 +12,7 @@
 #include "debug.h"
 
 std::string _log_file; ///< File to reroute output of a forked OpenTTD to
-std::unique_ptr<FILE, FileDeleter> _log_fd; ///< File to reroute output of a forked OpenTTD to
+AutoCloseFile _log_fd; ///< File to reroute output of a forked OpenTTD to
 
 #if defined(UNIX)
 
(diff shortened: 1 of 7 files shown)


PeterN:
Ah, because I got sidetracked... my version of this: https://github.com/OpenTTD/OpenTTD/compare/master...PeterN:raii-filehandle

rubidium42:
Peter's version is much further ahead than this is.

PR #12599, opened 2024-04-30
Feature: Engine expiry setting to keep owned engines available for purchase

2TallTyler:
## Motivation / Problem

@andythenorth and others want to roleplay with their favorite trains forever, yet continue the passage of time.

We have "vehicles never expire" for this, but it's too big a hammer with a GRF like Iron Horse.

## Description

Extend the vehicle expiry setting to add a third option: Engines that are owned by any company remain available for purchase, while unowned vehicles expire normally.

While I'm renaming things, this setting is about engine expiry, not vehicle expiry. 😃 

## Limitations

Iterating through all vehicles every time `CalcEngineReliability()` is called feels expensive. We should probably cache a list of currently-owned engines and update it using a timer, once per economy month (about a minute). I could use some help determining the proper way to do this. 🙂 

## Checklist for review

### DIFF
--- a/src/engine_type.h
+++ b/src/engine_type.h
@@ -195,6 +195,15 @@ enum EngineNameContext : uint8_t {
 	AutoreplaceVehicleInUse = 0x22, ///< Name is show in the autoreplace window 'Vehicles in use' panel.
 };
 
+/**
+ * Possible values of the "engine_expiry" setting.
+ */
+enum EngineExpiryMode : uint8_t {
+	Off = 0,   ///< Engines expire normally.
+	Never = 1, ///< Engines never expire.
+	Owned = 2, ///< Engines which are currently owned by any company never expire. Unowned vehicles expire normally.
+};
+
 /** Combine an engine ID and a name context to an engine name dparam. */
 inline uint64_t PackEngineNameDParam(EngineID engine_id, EngineNameContext context, uint32_t extra_data = 0)
 {
--- a/src/table/settings/game_settings.ini
+++ b/src/table/settings/game_settings.ini
@@ -10,6 +10,7 @@
 
 [pre-amble]
 static constexpr std::initializer_list<const char*> _roadsides{"left", "right"};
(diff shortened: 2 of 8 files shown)


src/table/settings/game_settings.ini
```diff
@@ -243,8 +243,20 @@ strval   = STR_CONFIG_SETTING_NONE
...
+type     = SLE_UINT8
+from     = SLV_ENGINE_EXPIRY_OWNED
+flags    = SF_GUI_DROPDOWN | SF_NO_NETWORK
+def      = EngineExpiryMode::Off
+min      = EngineExpiryMode::Off
+max      = EngineExpiryMode::Owned
 str      = STR_CONFIG_SETTING_NEVER_EXPIRE_VEHICLES
 strhelp  = STR_CONFIG_SETTING_NEVER_EXPIRE_VEHICLES_HELPTEXT
+strval   = STR_CONFIG_SETTING_NEVER_EXPIRE_VEHICLES_OFF
+cat      = SC_BASIC
```
TrueBrain, line 261: And the hopefully predictable question: why not make it store the enum name, instead of a number? Makes the configuration a bit more self-explaining.

frosch123:
I do not quite understand from the diff, how this works.

What happens when the last vehicle is sold?
* If the engine is past phase 3: does the engine immediately become unavailable?
* If the engine is in phase 3: does the engine immediately drop in reliability?

2TallTyler:
The behavior of selling the last vehicle is identical to what currently happens if `Vehicles never expire` is disabled. The specifics of that behavior are a bit beyond my understanding. 🙂

src/engine_type.h
```diff
@@ -195,6 +195,15 @@ enum EngineNameContext : uint8_t {
 	AutoreplaceVehicleInUse = 0x22, ///< Name is show in the autoreplace window 'Vehicles in use' panel.
 };
 
+/**
+ * Possible values of the "engine_expiry" setting.
+ */
+enum EngineExpiryMode : uint8_t {
```
Rito13, line 201: Now it could and probably should be an `enum class`, right?

PR #12607, opened 2024-05-01
Codechange: Use vector for airport tile layouts.

PeterN:
## Motivation / Problem

AirportSpec contains some pointers, and pointers to pointers, along with counts of the number of items that they point to.

These pointers point to either the original airport layouts or custom airport layouts, which may or may not need to be freed afterwards.

The layout of the data is awkward to understand as well, being stored in separate arrays instead of an array of structs.

## Description

Simplify AirportSpec data by storing layout information together in a vector, instead of separate arrays.

Adds the struct `AirportTileLayout` which holds both the `AirportTileTable` data and the rotation, keeping this data together.

The depot data is actually not changeable as it depends on the FSM, so this pointer is removed and replaced with a span as it never needs to be copied or changed.

This removes manual memory management and separate count members.

The default layouts will be copied instead of always referring to the originals.

## Checklist for review

### DIFF
--- a/src/airport_gui.cpp
+++ b/src/airport_gui.cpp
@@ -277,7 +277,7 @@ public:
 			const AirportSpec *as = ac->GetSpec(_selected_airport_index);
 			if (as->IsAvailable()) {
 				/* Ensure the airport layout is valid. */
-				_selected_airport_layout = Clamp(_selected_airport_layout, 0, as->num_table - 1);
+				_selected_airport_layout = Clamp(_selected_airport_layout, 0, static_cast<uint8_t>(as->layouts.size() - 1));
 				selectFirstAirport = false;
 				this->UpdateSelectSize();
 			}
@@ -306,7 +306,7 @@ public:
 					StringID string = GetAirportTextCallback(as, _selected_airport_layout, CBID_AIRPORT_LAYOUT_NAME);
 					if (string != STR_UNDEFINED) {
 						SetDParam(0, string);
-					} else if (as->num_table > 1) {
+					} else if (as->layouts.size() > 1) {
 						SetDParam(0, STR_STATION_BUILD_AIRPORT_LAYOUT_NAME);
 						SetDParam(1, _selected_airport_layout + 1);
 					}
@@ -348,7 +348,7 @@ public:
 				for (int i = 0; i < NUM_AIRPORTS; i++) {
 					const AirportSpec *as = AirportSpec::Get(i);
 					if (!as->enabled) continue;
-					for (uint8_t layout = 0; layout < as->num_table; layout++) {
+					for (uint8_t layout = 0; layout < static_cast<uint8_t>(as->layouts.size()); layout++) {
 						SpriteID sprite = GetCustomAirportSprite(as, layout);
 						if (sprite != 0) {
 							Dimension d = GetSpriteSize(sprite);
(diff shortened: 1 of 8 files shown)


src/newgrf.cpp
```diff
@@ -3942,89 +3916,68 @@ static ChangeInfoResult AirportChangeInfo(uint airport, int numinfo, int prop, B
...
+								GrfMsg(2, "AirportChangeInfo: Attempt to use airport tile {} with airport id {}, not yet defined. Ignoring.", local_tile_id, airport + i);
 							} else {
-								as->size_x = std::max<uint8_t>(as->size_x, att[k].ti.x + 1);
-								as->size_y = std::max<uint8_t>(as->size_y, att[k].ti.y + 1);
+								/* Declared as been valid, can be used */
+								tile.gfx = tempid;
 							}
+						} else if (tile.gfx == 0xFF) {
+							tile.ti.x = (int8_t)GB(tile.ti.x, 0, 8);
+							tile.ti.y = (int8_t)GB(tile.ti.y, 0, 8);
```
rubidium42:
```suggestion
							tile.ti.x = static_cast<int8_t>(GB(tile.ti.x, 0, 8));
							tile.ti.y = static_cast<int8_t>(GB(tile.ti.y, 0, 8));
```

src/newgrf.cpp
```diff
@@ -3942,89 +3916,68 @@ static ChangeInfoResult AirportChangeInfo(uint airport, int numinfo, int prop, B
...
-								uint16_t tempid = _airporttile_mngr.GetID(local_tile_id, _cur.grffile->grfid);
+					for (;;) {
+						auto &tile = layout.tiles.emplace_back();
+						tile.ti.x = buf->ReadByte();
+						tile.ti.y = buf->ReadByte();
+						if (tile.ti.x == 0 && tile.ti.y == 0x80) {
+							/* Convert terminator to our own. */
+							tile.ti.x = -0x80;
+							tile.ti.y = 0;
+							tile.gfx = 0;
```
rubidium42, line 3944: Not for this PR, but if the tile table is now a vector or span then the termination sentinel does not need to be saved in OpenTTD's state.

src/newgrf.cpp
LordAro: I'm confused by what this line is actually doing. Seems like just doing `tile.ti.x & 0xFF` ?

src/airport_gui.cpp
```diff
@@ -348,7 +348,7 @@ class BuildAirportWindow : public PickerWindowBase {
 				for (int i = 0; i < NUM_AIRPORTS; i++) {
 					const AirportSpec *as = AirportSpec::Get(i);
 					if (!as->enabled) continue;
-					for (uint8_t layout = 0; layout < as->num_table; layout++) {
+					for (uint8_t layout = 0; layout < static_cast<uint8_t>(as->layouts.size()); layout++) {
```
LordAro, line 351: would it be nicer to change `layout` to be a size_t, rather than have the cast? Might need some signature changes as well I suppose

src/newgrf.cpp
PeterN: It's converting an unsigned 8 bit value to a signed 8 bit value.. This allows the airport layout to "check" for clear tiles outside of the north edges.

src/newgrf.cpp
PeterN: Agreed, I figured this already touches enough :)

src/airport_gui.cpp
PeterN: That does lead to a bit of a rabbit hole, when there's no need for for a size_t field further down.

rubidium42 (approved):

PR #12608, opened 2024-05-01
Codechange: Ensure SDLK mappings stay in the expected order.

PeterN:
## Motivation / Problem

#12596 found an issue with SDLK mappings that we did not know about, but we can make the compiler tell us.

## Description

Add a constexpr constructor that ensures at compile-time that the source SDLK range matches the target range.

Therefore, it is no longer possible (fingers-crossed...) for the key ranges to be incorrect.

## Checklist for review


rubidium42 (approved):
Something similar can also be done for Allegro and Windows.
For MacOS the mapping is fully 1:1.

PR #12611, opened 2024-05-02
Feature: Add font picker to the game settings.

ladysadie:
## Motivation / Problem

Motivation for this is to allow the user to resize fonts in the game UI.  This is an issue for players on high resolution screens where the small font is unreadable. 

## Description

I added a new tab to the game options dialog. (Pictured below for reference:)

image

I moved the two font related options from GUI to this tab.  Then added a section for each font. Each font has a slider to adjust the font size. Each font is rendered on a button that when clicked launches a font loading dialog pictured below:

image

The slider allows users to scale up/down fonts for readability.  The font chooser should hopefully help players who use languages that aren't completely supported by OpenTTD's base fonts.

For the Future:
*  I think removing font console command makes sense as it's no longer needed.
* I'd like to expand on the font chooser loading box. It could be smarter about looking at System directories for fonts.
* When the player clicks a font while on the font chooser dialog it would be cool to see that font rendered in the preview panel.  Right now there's just a text label describing where that is used in the game client.

## Limitations
Only looks for TTF fonts. (If this is a problem I could fix it.)

The sliders have limits for how small/large each font can be set.  I think this makes sense. I don't really see a point in allowing the user to set a font to be 96.  You can size font that high in office documents but I don't see a need for that in the game client. Or going the other way font size of 4 would just be unreadable. 

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -2233,7 +2233,7 @@ DEF_CONSOLE_CMD(ConFont)
 		FontCacheSubSetting *setting = GetFontCacheSubSetting(fs);
 		/* Make sure all non sprite fonts are loaded. */
 		if (!setting->font.empty() && !fc->HasParent()) {
-			InitFontCache(fs == FS_MONO);
+			InitFontCache();
 			fc = FontCache::Get(fs);
 		}
 		IConsolePrint(CC_DEFAULT, "{} font:", FontSizeToName(fs));
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -38,23 +38,24 @@ extern std::string _config_file;
 extern std::string _highscore_file;
 
 static const char * const _subdirs[] = {
-	"",
-	"save" PATHSEP,
-	"save" PATHSEP "autosave" PATHSEP,
-	"scenario" PATHSEP,
-	"scenario" PATHSEP "heightmap" PATHSEP,
-	"gm" PATHSEP,
-	"data" PATHSEP,
-	"baseset" PATHSEP,
-	"newgrf" PATHSEP,
-	"lang" PATHSEP,
-	"ai" PATHSEP,
-	"ai" PATHSEP "library" PATHSEP,
-	"game" PATHSEP,
-	"game" PATHSEP "library" PATHSEP,
-	"screenshot" PATHSEP,
-	"social_integration" PATHSEP,
+	"",                                     ///< BASE_DIR
+	"save" PATHSEP,                         ///< SAVE_DIR
+	"save" PATHSEP "autosave" PATHSEP,      ///< AUTOSAVE_DIR
+	"scenario" PATHSEP,                     ///< SCENARIO_DIR
+	"scenario" PATHSEP "heightmap" PATHSEP, ///< HEIGHTMAP_DIR
+	"gm" PATHSEP,                           ///< OLD_GM_DIR
+	"data" PATHSEP,                         ///< OLD_DATA_DIR
+	"baseset" PATHSEP,                      ///< BASESET_DIR
+	"newgrf" PATHSEP,                       ///< NEWGRF_DIR
+	"lang" PATHSEP,                         ///< LANG_DIR
+	"ai" PATHSEP,                           ///< AI_DIR
+	"ai" PATHSEP "library" PATHSEP,         ///< AI_LIBRARY_DIR
+	"game" PATHSEP,                         ///< GAME_DIR
+	"game" PATHSEP "library" PATHSEP,       ///< GAME_LIBRARY_DIR
+	"screenshot" PATHSEP,                   ///< SCREENSHOT_DIR
+	"social_integration" PATHSEP,           ///< SOCIAL_INTEGRATION_DIR
 };
+// This should line up with the "Subdirectory" enum defined in fileio_type.h
 static_assert(lengthof(_subdirs) == NUM_SUBDIRS);
 
 /**
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -54,6 +54,7 @@ static const char * const _subdirs[] = {
 	"game" PATHSEP "library" PATHSEP,       ///< GAME_LIBRARY_DIR
 	"screenshot" PATHSEP,                   ///< SCREENSHOT_DIR
 	"social_integration" PATHSEP,           ///< SOCIAL_INTEGRATION_DIR
+	"baseset" PATHSEP,                      ///< FONT_DIR
 };
 // This should line up with the "Subdirectory" enum defined in fileio_type.h
 static_assert(lengthof(_subdirs) == NUM_SUBDIRS);
--- a/src/fileio_type.h
+++ b/src/fileio_type.h
@@ -109,24 +109,24 @@ inline DetailedFileType GetDetailedFileType(FiosType fios_type)
  * The different kinds of subdirectories OpenTTD uses
  */
 enum Subdirectory {
-	BASE_DIR,      ///< Base directory for all subdirectories
-	SAVE_DIR,      ///< Base directory for all savegames
-	AUTOSAVE_DIR,  ///< Subdirectory of save for autosaves
-	SCENARIO_DIR,  ///< Base directory for all scenarios
-	HEIGHTMAP_DIR, ///< Subdirectory of scenario for heightmaps
-	OLD_GM_DIR,    ///< Old subdirectory for the music
-	OLD_DATA_DIR,  ///< Old subdirectory for the data.
-	BASESET_DIR,   ///< Subdirectory for all base data (base sets, intro game)
-	NEWGRF_DIR,    ///< Subdirectory for all NewGRFs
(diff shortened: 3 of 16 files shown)


PeterN:
Requiring the user to search for a font by path is not very friendly. It would be better to ask the OS what fonts are available and list those, not using the normal 'fios' window. Don't involve directories at all, users do not expected this.

TheMowgliMan:
> Requiring the user to search for a font by path is not very friendly. It would be better to ask the OS what fonts are available and list those, not using the normal 'fios' window. Don't involve directories at all, users do not expected this.

It might be nice to be able to open a standard file picker for people who install fonts in non-standard locations (for example, if thry dont know where the system default location is). Id normally say its unneccesary to support that, but given its already implemented it might as well stick around.

ladysadie:
> Requiring the user to search for a font by path is not very friendly. It would be better to ask the OS what fonts are available and list those, not using the normal 'fios' window. Don't involve directories at all, users do not expected this.

I think I remember you saying you wrote some code for doing that.

I'm not against the idea.  I was trying to get the smallest function change. I'd rather do it as a follow-on PR as expansion of this basic code.  Would that be acceptable?

PeterN:
Not really. Trying to join two different ways of picking a font into one UI seems to be overly complex and unnecessary.

Q. How do I pick `Verdana Bold` or `Iosevka Curly Slab Bold` with a filesystem selector?
A. I can't, the file needs to be opened to found out what fonts are contained within it.

On Linux, FontConfig handles finding fonts and listing what's in them, and Windows and macOS have native APIs to list fonts.

glx22:
On windows (at least, don't really know about other APIs), it's possible to request only fonts compatible with current "language".

LordAro:
The font picker definitely should be improved to use native APIs, but I don't see it as a huge blocker as we'll need the fallback for systems that don't support fontconfig/native API anyway (and for manual font picking regardless? Or maybe that can just be relegated to config file editing only)

LordAro:
However, with the font picker as it is now, I'd get rid of the right panel and just have the window title say which font is being selected

ladysadie:
> Not really. Trying to join two different ways of picking a font into one UI seems to be overly complex and unnecessary.
> 
> Q. How do I pick `Verdana Bold` or `Iosevka Curly Slab Bold` with a filesystem selector? A. I can't, the file needs to be opened to found out what fonts are contained within it.
> 
> On Linux, FontConfig handles finding fonts and listing what's in them, and Windows and macOS have native APIs to list fonts.

I was thinking you would use those api's to source the list of available fonts.  Then display that list in the filterbox instead of the directory listing?  Until the various OS'es get that support just displaying the directories works?

I realized that there should at some point be some code to handle bold/regular/etc but I haven't spent any time yet researching the topic.

ladysadie:
> On windows (at least, don't really know about other APIs), it's possible to request only fonts compatible with current "language".

Okay. Something to keep in mind about the windows fonts.  This does beg the question: what happens if they differ? (Windows is using one language and the game another?) 

I would assume the filtering should use the language set in OpenTTD?

ladysadie:
> However, with the font picker as it is now, I'd get rid of the right panel and just have the window title say which font is being selected

Okay I'll try implementing that and get back with a screenshot.

Another thing that occurred to me is having a button some where to just "reset" to the default OpenTTD ttf font is something I should also add.

PeterN:
Using the fios window unnecessarily complicates the the fios system and the font selector.

Fonts are not files, they just happen to be contained in them. As glx22 mentions, the APIs are able to tell us what languages they support, along with all the different font variants contained within a single file.

I don't think there's any need for a fallback using a file picker window. Outside of Windows and mac OS, if you do not have font-config these days, you are not running a regular desktop OS—and I think manual configuration like already exists should be fine.

TheMowgliMan:
> Not really. Trying to join two different ways of picking a font into one UI seems to be overly complex and unnecessary.
> 
> Q. How do I pick `Verdana Bold` or `Iosevka Curly Slab Bold` with a filesystem selector? A. I can't, the file needs to be opened to found out what fonts are contained within it.
> 
> On Linux, FontConfig handles finding fonts and listing what's in them, and Windows and macOS have native APIs to list fonts.

A s,mall browse button as many apps use for this sort of application would suffice and not be overly intrusive.

ladysadie:
> However, with the font picker as it is now, I'd get rid of the right panel and just have the window title say which font is being selected

image

Done.

ladysadie:
> Using the fios window unnecessarily complicates the the fios system and the font selector.
> 
> Fonts are not files, they just happen to be contained in them. As glx22 mentions, the APIs are able to tell us what languages they support, along with all the different font variants contained within a single file.
> 
> I don't think there's any need for a fallback using a file picker window. Outside of Windows and mac OS, if you do not have font-config these days, you are not running a regular desktop OS—and I think manual configuration like already exists should be fine.

I don't feel like the FIOS was negatively impacted by this commit.

I've never used those apis. I skimmed them.  I would make a PR for each OS starting with Linux. What would the others have in the meantime?

ladysadie:
> I like this feature, but isn't it overkill to create a separate tab for it? :) Maybe it would be better to add a button under the "Graphics" tab in the Interface section that would open an additional window with font settings? Font settings v1
> 
> Or add the ability to scroll the window and have a separate section for font settings in the same tab? This would also solve the problem of inaccessible settings if someone uses the interface magnification too high. Font settings v2
> 
> Ultimately, this is not the most important thing. What is more important is how you choose the font. I think a drop down list with or without font preview would be much better than selecting via path.

I am solidly in the land of "can't please everyone" with the UI design.

Total no on the scroll bar design. I don't like it.

Adding a sub-window is a maybe? I would probably stick the button that launches it on the "General Tab" next to the language box.

Is there something specific about the new tab design you don't like? Clicking a button to open a font settings windows implies a second click to close that window.  Adding a tab is one click to focus the tab but same number of clicks to close the font options along with the rest of game options dialog. Convince me the second click is worth it?

Something I've noticed is the "Prefer Sprites" button  disables all of the font settings.  Not sure how to feel about that right now.

ladysadie:
> > Is there something specific about the new tab design you don't like? Clicking a button to open a font settings windows implies a second click to close that window. Adding a tab is one click to focus the tab but same number of clicks to close the font options along with the rest of game options dialog. Convince me the second click is worth it?
> 
> The thing is, I like this design. :) I like things that are easily accessible (I almost always and everywhere find that single clicking is better than double), the way of resizing is ok, the preview is even more than I would like, the way of choosing the font... I would prefer to choose from a drop down list, but if it is problematic, it is still much better than current methods. However, when it comes to my mockups, yes, the one with the scroll bar is a rather poor solution, but I presented it for reference. The main point is that the font doesn't seem to be an element of particular importance that it should have a dedicated tab. This is a setting that players are unlikely to change very often. Besides, if it weren't for the illegibility of the small font, I personally wouldn't see the point in changing anything. Now that the font is scalable, I don't think players need to have these settings so exposed. As for the location, for me it is clearly an element of graphics and interface. Currently, its settings are there and it seems to be the most appropriate place for me.



> > Is there something specific about the new tab design you don't like? Clicking a button to open a font settings windows implies a second click to close that window. Adding a tab is one click to focus the tab but same number of clicks to close the font options along with the rest of game options dialog. Convince me the second click is worth it?
> 
> The thing is, I like this design. :) I like things that are easily accessible (I almost always and everywhere find that single clicking is better than double), the way of resizing is ok, the preview is even more than I would like, the way of choosing the font... I would prefer to choose from a drop down list, but if it is problematic, it is still much better than current methods. However, when it comes to my mockups, yes, the one with the scroll bar is a rather poor solution, but I presented it for reference. The main point is that the font doesn't seem to be an element of particular importance that it should have a dedicated tab. This is a setting that players are unlikely to change very often. Besides, if it weren't for the illegibility of the small font, I personally wouldn't see the point in changing anything. Now that the font is scalable, I don't think players need to have these settings so exposed. As for the location, for me it is clearly an element of graphics and interface. Currently, its settings are there and it seems to be the most appropriate place for me.

If we both like the design does it really need to change?

How can we determine which settings are important enough to get a "tab"? I'm only one person but the whole game options dialog is a "get it right" then never touch it again.  I set my language. Scaled the interface up.  Then scaled up the small/medium fonts to make them easy to see on my screen.  No more squinting, no more lifting the laptop up to my face to read map legends.  I don't open the Game options anymore when I'm playing.

Having a tab does have the advantage of being more discoverable for the players though. At the cost of space on the tab bar.  Currently seems like a fine trade-off to me.  I don't see the tabs as crowded or illegible. 

Scaling up the font with the new TTF fonts is nice. Certainly an improvement over the fixed size sprite fonts.  It did fall short of making the small font readable.  Which was the motivation for this change. 

I've started work on adding the code to use fontconfig on Linux.  Not sure how long that's going to take. That also doesn't help Windows or Mac players though.

ladysadie:
Given the discussion that we had on the OpenTTD developer channel I'm to refactor this code and split it up.  So for now I can close this.

PR #12618, opened 2024-05-03
Change: Treat extended waypoint classes as 'WAYP' class.

PeterN:
## Motivation / Problem

Authors have asked for more custom waypoint classes, as lumping all waypoints into just one class gives a very big list.

My plan on the easiest way to achieve this is to treat all classes with the MSB as 0xFF as a waypoint class.

WIth #12595, this is almost implemented. However, to use the feature requires support from NewGRFs, and if they add this support then they will not work properly in 14.x

## Description

So this PR is intended to provide a little "forward-compatibility" for 14.x, by translating new waypoint classes to the standard 'WAYP' class.

This maps the extended custom waypoint clases (where the first byte of the label is 0xFF) to the standard 'WAYP' label, allowing use of them before the feature exists.

## Limitations

This also assumes that more-waypoint-classes plan with 0xFF as MSB is reasonable.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1937,8 +1937,9 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 				}
 
 				/* Swap classid because we read it in BE meaning WAYP or DFLT */
-				uint32_t classid = buf->ReadDWord();
-				statspec->cls_id = StationClass::Allocate(BSWAP32(classid));
+				uint32_t classid = BSWAP32(buf->ReadDWord());
+				if (GB(classid, 24, 8) == 0xFF) classid = 'WAYP';
+				statspec->cls_id = StationClass::Allocate(classid);
 				break;
 			}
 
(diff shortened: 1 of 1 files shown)


bootmii:
I feel like this isn't the best time to do this, and I'd prefer this to be in 14.2 over 14.1. But I trust @PeterN's judgment. I'm sorry I was rude to you and called this PR "stupid".

2TallTyler:
To clarify the "why" a bit from our Discord discussion:

The purpose of this future-proofing is to allow GRF authors to make waypoints that have classes, which show up in classes in nightlies (and eventually 15.0+), but also work in older versions of OpenTTD like 14.1. That way, they can make classes without having to wait a whole year for the next release.

I very much like the idea, but let's not hold up 14.1 while we decide on this. We can ship a 14.2 if we decide to do this.

rubidium42:
I guess this requires retargetting the 14 branch now.

PeterN:
Probably, a backport from master would need 1) changing for 14.x and 2) reverting again.

rubidium42 (approved):

PeterN:
Seems a bit pointless at this point.

PR #12619, opened 2024-05-03
Codechange: Use vectors instead of CallocT/free for cache checks.

PeterN:
## Motivation / Problem

calloc/free is used in `CheckCaches()` to allocate temporary storage to compare caches.

This is allocated and free for each vehicle chain, and the length of the memory to be allocated has to be predetermined.

## Description

Use vectors instead. These are allocated automatically as needed and allow reuse of the allocated memory for each vehicle chain.

This removes manual memory allocation with calloc/free.

## Checklist for review

### DIFF
--- a/src/cachecheck.cpp
+++ b/src/cachecheck.cpp
@@ -79,34 +79,29 @@ void CheckCaches()
 		rs->GetEntry(DIAGDIR_NW)->CheckIntegrity(rs);
 	}
 
+	std::vector<NewGRFCache> grf_cache;
+	std::vector<VehicleCache> veh_cache;
+	std::vector<GroundVehicleCache> gro_cache;
+	std::vector<TrainCache> tra_cache;
+
 	for (Vehicle *v : Vehicle::Iterate()) {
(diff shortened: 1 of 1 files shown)


src/cachecheck.cpp
```diff
@@ -79,29 +79,26 @@ void CheckCaches()
...
 
+	std::vector<NewGRFCache> grf_cache;
+	std::vector<VehicleCache> veh_cache;
+	std::vector<GroundVehicleCache> gro_cache;
+	std::vector<TrainCache> tra_cache;
+
 	for (Vehicle *v : Vehicle::Iterate()) {
 		if (v != v->First() || v->vehstatus & VS_CRASHED || !v->IsPrimaryVehicle()) continue;
 
 		uint length = 0;
```
rubidium42: Shouldn't this move to line 117? And can't the `length++` at line 106 be removed?

PeterN: Yes. I just hadn't spotted this in the excitement of removing the first loop :-)

rubidium42 (approved):

PR #12635, opened 2024-05-07
Fix: always allow setting company settings, company/president name/face

rubidium42:
## Motivation / Problem

I got triggered into defining why some functions were using `Command::SendNet` and not `Command::Post`. This spiraled into a rabbit hole with this as result.

There is a nice feature that synchronises the client settings upon setting up the company. Before this, those commands would not be executed when no-actions-while-paused is set. This means that, silently and depending on the server configuration, your wished for configuration might not be there.

Similarly there is the president's face that's being set while creating a new company and setting of the president/company name upon creation, when no-actions-while-paused is set.

So, just allow these operations also while paused to get a uniform experience when joining. To keep the UI somewhat consistent, apply this "freedom" also to the other bits set from the company UI; specifically company name and company colour.

## Description

There are two `::SendNet` calls that do not need to be and they are made `::Post`.

For the remaining `::SendNet` calls I have added some documentation why they are not `::Post`, with added benefit that you'll now find those as well when searching for `::Post`.

Finally I added some more commands to the "company settings"-pause group, and made that pause group essentially "always allowed".

## Checklist for review


src/network/network_server.cpp
```diff
@@ -2157,6 +2157,11 @@ void NetworkServerNewCompany(const Company *c, NetworkClientInfo *ci)
 		/* ci is nullptr when replaying, or for AIs. In neither case there is a client. */
 		ci->client_playas = c->index;
 		NetworkUpdateClientInfo(ci->client_id);
+
+		/* This function is called from a command, but is only called for the server.
+		 * The client information is managed out-of-band from the commands, so to not have a
+		 * different state/president/company name in the different clients, we need to
+		 * circumvent the normal ::Post logic and go directly to sending the command. */
```
2TallTyler:
```suggestion
		/*
		 * This function is called from a command, but is only called for the server.
		 * The client information is managed out-of-band from the commands, so to not have a
		 * different state/president/company name in the different clients, we need to
		 * circumvent the normal ::Post logic and go directly to sending the command.
		 */
```

src/settings.cpp
```diff
@@ -1790,6 +1790,9 @@ void SyncCompanySettings()
 		const SettingDesc *sd = GetSettingDesc(desc);
 		uint32_t old_value = (uint32_t)sd->AsIntSetting()->Read(old_object);
 		uint32_t new_value = (uint32_t)sd->AsIntSetting()->Read(new_object);
+		/* This is called from a command, and since it contains local configuration information
+		 * that the rest of the clients do not know about, we need to circumvent the normal ::Post
+		 * local command validation and immediately send the command to the server. */
```
```suggestion
		/*
		 * This is called from a command, and since it contains local configuration information
		 * that the rest of the clients do not know about, we need to circumvent the normal ::Post
		 * local command validation and immediately send the command to the server.
		 */
```

src/network/network_server.cpp
PeterN: Why? We do that for Doxygen, but this isn't Doxygen.

2TallTyler: Oh, I've been doing it everywhere. 🤷 

`CODINGSTYLE.md` has an example using a third style, so I guess there are no rules for regular comments. 🙂 
```
/* Unoptimized code:
 * foo is not touched inside the loop!
 */
```

PeterN (approved):

PR #12638, opened 2024-05-07
Codechange: use std::optional<std::string> over char * for text query result

rubidium42:
## Motivation / Problem

`OnQueryTextFinished` gets a `char *` which in many places gets passed into `StrEmpty`.

The `char *` fulfils multiple meanings; when it's `nullptr` it means the window was closed and there is no string, when the string is empty (`str[0] == '\0'`) this denotes the default value (also functionally; clearing the company name and then pressing OK will give you the default name), and then anything non-empty is considered valid data.

Since we're moving towards a more C++-styled codebase, convert this `char *` into `std::optional<std::string>` instead.

## Description

Convert the parameter type of `OnQueryTextFinished`.

## Limitations

This introduces quite a few `->c_str()`s because many places use `atoi` and friends. Replacing `atoi` with a more modern variant is beyond the scope of this PR.

## Checklist for review

### DIFF
--- a/src/cheat_gui.cpp
+++ b/src/cheat_gui.cpp
@@ -404,14 +404,14 @@ struct CheatWindow : Window {
 		this->SetDirty();
 	}
 
-	void OnQueryTextFinished(char *str) override
+	void OnQueryTextFinished(std::optional<std::string> str) override
 	{
 		/* Was 'cancel' pressed or nothing entered? */
-		if (str == nullptr || StrEmpty(str)) return;
+		if (!str.has_value() || str->empty()) return;
 
 		const CheatEntry *ce = &_cheats_ui[clicked_widget];
 		int oldvalue = (int32_t)ReadValue(ce->variable, ce->type);
-		int value = atoi(str);
+		int value = atoi(str->c_str());
 		*ce->been_used = true;
 		value = ce->proc(value, value - oldvalue);
 
--- a/src/build_vehicle_gui.cpp
+++ b/src/build_vehicle_gui.cpp
@@ -1853,11 +1853,11 @@ struct BuildVehicleWindow : Window {
 		}
 	}
 
-	void OnQueryTextFinished(char *str) override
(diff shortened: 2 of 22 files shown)


src/cheat_gui.cpp
```diff
@@ -404,14 +404,14 @@ struct CheatWindow : Window {
 		this->SetDirty();
 	}
 
-	void OnQueryTextFinished(char *str) override
+	void OnQueryTextFinished(std::optional<std::string> str) override
```
LordAro, line 407: A few more const-ings? I don't think these are modified anywhere within the OnQuery* functions?

Not sure about `&` either, presumably the std::optional is already avoiding a copy?

PeterN: std::optional is not avoiding a copy.

rubidium42: `std::optional` doesn't allow references.

Then there is the choice whether we want to make everything `const`, or do we want the string to be `std::move`able? There are three places where we could use `std::move` as data is written directly to variables. In all other cases it's either `->c_str()` or passed as a `const std::string &` into a command.

PeterN: Who owned the string when it was `char *` form? `std::string_view` might be appropriate, but then again is also potentially as dangerous as `char *` in this context.

rubidium42: `std::string_view` requires changing many command signatures as well as move `std::string_view` much deeper into the settings. That would rabbit hole quite quickly.

LordAro:
>Replacing atoi with a more modern variant is beyond the scope of this PR.

Why?

rubidium42:
> > Replacing atoi with a more modern variant is beyond the scope of this PR.
> 
> Why?

Because `std::stoi` throws an exception that needs to be managed. Using that would also imply having proper error handling. Adding that error handling is not bad, but it balloons this PR and it would likely be better to have a PR/commit on a case-by-case basis.

LordAro (approved):

PR #12641, opened 2024-05-07
Fix: Allow changing size of default OpenTTD font.

PeterN:
## Motivation / Problem

Size configuration for default font was ignored as a different code path to load the font was followed.

Size of default font could only be changed after manually setting the configured font to the default font name.

## Description

Resolved by removing this additional path and conditionally selecting the default font.

This allows the normal font size selection to occur, and reduces some special-casing for the default font.

## Checklist for review

### DIFF
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -139,50 +139,57 @@ void SetFont(FontSize fontsize, const std::string &font, uint size)
 
 #ifdef WITH_FREETYPE
 extern void LoadFreeTypeFont(FontSize fs);
-extern void LoadFreeTypeFont(FontSize fs, const std::string &file_name, uint size);
 extern void UninitFreeType();
 #elif defined(_WIN32)
 extern void LoadWin32Font(FontSize fs);
-extern void LoadWin32Font(FontSize fs, const std::string &file_name, uint size);
 #elif defined(WITH_COCOA)
 extern void LoadCoreTextFont(FontSize fs);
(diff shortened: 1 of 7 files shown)


src/fontcache.cpp
```diff
@@ -139,52 +139,59 @@ void SetFont(FontSize fontsize, const std::string &font, uint size)
...
+/**
+ * Get font to use for a given font size.
+ * @param fs Font size.
+ * @return If configured, the font name to use, or the path of the default TrueType font if sprites are not preferred.
+ */
+std::string GetFontCacheFontName(FontSize fs)
+{
+	const FontCacheSubSetting *settings = GetFontCacheSubSetting(fs);
+	if (!settings->font.empty()) return settings->font;
+	if (_fcsettings.prefer_sprite) return {};
```
ladysadie, line 191: Shouldn't the prefer sprite flag override any font name having been set? Wouldn't have a font name set here break the prefer sprites checkbox?

glx22: prefer sprite only affects default font, if you set a font name the checkbox should be disabled.

ladysadie: That is how it works today.  I didn't know that. 

I am not sure about disabling the checkbox though.  That seems to be unrelated to this review though.

rubidium42 (approved):

ladysadie:
I guess I don't understand.

Are 6px/10px not good sizes or is the code that determines what the minimum height is giving inflated sizes back?

PR #12645, opened 2024-05-08
Cleanup: Remove MinGW Windows XP support workaround.

PeterN:
## Motivation / Problem

The first few defines in `stdafx.h` are there to allow building with MinGW to support Windows XP+ instead of Window 7+.

However our lowest supported version is Windows 7 now anyway.

## Description

Remove obsolete defines for Windows XP support.

## Limitations

Works with MinGW on Linux, untested on real Windows.

## Checklist for review

### DIFF
--- a/src/stdafx.h
+++ b/src/stdafx.h
@@ -10,15 +10,6 @@
 #ifndef STDAFX_H
 #define STDAFX_H
 
-#if defined(_WIN32)
-	/* MinGW defaults to Windows 7 if none of these are set, and they must be set before any MinGW header is included */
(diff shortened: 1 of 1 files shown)


PeterN:
Ah ha, this may just be there to disable Windows 7+ bits that we don't need, rather than for actually supporting Windows XP?

PeterN:
Misinterpretation I think. Although it mentions Windows XP, this is not specifically about supporting XP, just setting a known API level to use.

glx22:
The comment about MinGW is also misleading, these defines without the comment used to be in a _MSC_VER section, before #9135.

PR #12646, opened 2024-05-08
Change: Use default NewGRF cargo translation table.

PeterN:
## Motivation / Problem

When loading NewGRFs without cargo translation table, we have to check to see if we need to perform a bitnum lookup or just treat the slot as a climate dependent cargo type.

This requires multiple special cases, and always assumes that cargo slots are in their original positions, which is not always the case.

This can result in unexpected behaviour when mixing old sets (Ikarus Buses) with modern sets (FIRS 5):

image

## Description

Instead of falling back to bitnum lookup or climate-dependent cargo types, install a default cargo translation table that performs either of these functions instead.

A NewGRF can still install its own translation table and there will be no change in this case.

This allows better mapping of climate-dependent or bitnum cargo slots, falling back to INVALID_CARGO if they are not defined, and reduces special-casing.

Because everything is now mapped via label, this should allow cargo types to be moved to different slots (as is permitted for default vehicles/houses/industries and anything that installs its own translation table.)

This gives better results mixing old (Ikarus) and new (FIRS 5):

image

## Limitations

Well... I think this breaks everything using pre-cargo-label cargo sets, but it's entirely possible they are already broken in 14.x. (Seem to be broken in at least 1.11.2, so probably nothing to be concerned with...)

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1153,15 +1153,10 @@ static ChangeInfoResult RailVehicleChangeInfo(uint engine, int numinfo, int prop
 				if (ctype == 0xFF) {
 					/* 0xFF is specified as 'use first refittable' */
 					ei->cargo_type = INVALID_CARGO;
-				} else if (_cur.grffile->grf_version >= 8) {
+				} else {
 					/* Use translated cargo. Might result in INVALID_CARGO (first refittable), if cargo is not defined. */
 					ei->cargo_type = GetCargoTranslation(ctype, _cur.grffile);
-				} else if (ctype < NUM_CARGO) {
-					/* Use untranslated cargo. */
-					ei->cargo_type = ctype;
-				} else {
-					ei->cargo_type = INVALID_CARGO;
-					GrfMsg(2, "RailVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
+					if (ei->cargo_type == INVALID_CARGO) GrfMsg(2, "RailVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
 				}
 				ei->cargo_label = CT_INVALID;
 				break;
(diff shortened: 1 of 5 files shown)


src/newgrf.cpp
```diff
@@ -1410,15 +1405,10 @@ static ChangeInfoResult RoadVehicleChangeInfo(uint engine, int numinfo, int prop
...
+				} else {
 					/* Use translated cargo. Might result in INVALID_CARGO (first refittable), if cargo is not defined. */
 					ei->cargo_type = GetCargoTranslation(ctype, _cur.grffile);
-				} else if (ctype < NUM_CARGO) {
-					/* Use untranslated cargo. */
-					ei->cargo_type = ctype;
-				} else {
-					ei->cargo_type = INVALID_CARGO;
-					GrfMsg(2, "RailVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
+					if (ei->cargo_type == INVALID_CARGO) GrfMsg(2, "RailVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
```
glx22:
```suggestion
					if (ei->cargo_type == INVALID_CARGO) GrfMsg(2, "RoadVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
```

src/newgrf.cpp
```diff
@@ -1606,15 +1596,10 @@ static ChangeInfoResult ShipVehicleChangeInfo(uint engine, int numinfo, int prop
...
+				} else {
 					/* Use translated cargo. Might result in INVALID_CARGO (first refittable), if cargo is not defined. */
 					ei->cargo_type = GetCargoTranslation(ctype, _cur.grffile);
-				} else if (ctype < NUM_CARGO) {
-					/* Use untranslated cargo. */
-					ei->cargo_type = ctype;
-				} else {
-					ei->cargo_type = INVALID_CARGO;
-					GrfMsg(2, "ShipVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
+					if (ei->cargo_type == INVALID_CARGO) GrfMsg(2, "RailVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
```
glx22:
```suggestion
					if (ei->cargo_type == INVALID_CARGO) GrfMsg(2, "ShipVehicleChangeInfo: Invalid cargo type {}, using first refittable", ctype);
```

PeterN:
This goes against the spec when it comes to very old NewGRFs that deal with custom cargoes before labels were a thing.

Not sure how many of those exist, nor how many are still used, as compatibility is already hit & miss.

glx22:
Test build available at https://cdn.openttd.org/openttd-branches/pr12646/20240710-pr12646-gbc7f2af8a6/

PeterN:
I've done some scanning of bananas-available vehicle NewGRFs and I believe that this will have no negative impact on the vast majority of sets. Certainly many more sets will be fixed than broken.

I have another change in the pipeline to allow installing a cargo translation table when overriding a NewGRF which should allow anything found to be broken to be fixed. That depends on this PR because it needs the translation table to be applied regardless of GRF version.

michicc:
I like the idea. Code looks fine to me, but I don't really know the somewhat... convoluted ways of cargo handling in NewGRF well enough to notice any potential issues.

2TallTyler (approved):
Nobody has spoken against this despite your many invitations to do so, so it's time to forever hold our peace. 

Or at least wait for you to fix any unexpected bugs, which is fine too. 😃

PR #12650, opened 2024-05-09
Fix #12648: Ensure all uses of std::filesystem::path use native encoding.

### LINKED ISSUE #12648: [Crash]: If the directory name contains multibyte characters, openTTD will crash.
### Version of OpenTTD

14.1,windows11

### Steps to reproduce

1.Install openTTD in the multibyte character folder.
2.Start openTTD.

In countries that use multibyte characters (e.g., Japan), user folders and documents folders may be in multibyte characters.
Basically, even if a documents folder looks like it is in multibyte characters, the actual path is often in single-byte characters. 
However, if the documents folder is in OneDrive, the actual path is also likely to be multibyte characters.
If only the documents folder is multibyte characters, it will crash when trying to download newGRF.

### Upload crash files

[crash20240509071908.zip](https://github.com/OpenTTD/OpenTTD/files/15259119/crash20240509071908.zip)

PeterN:
## Motivation / Problem

std::filesystem::path does not expect UTF-8 strings, and it seems may produce invalid results when done so, so they must be converted to native format first (and back to utf-8 after.)

## Description

Ensure strings passed to std::filesystem::path go through OTTD2FS conversion, and back to UTF-8 with FS2OTTD after.

Sadly std::filesystem::path will implicitly convert a std::string, so it's difficult to catch all case.

This is not the correct way to deal with this situation. C++20 would like us to use std::u8string everything, but that is not really feasible. So this is a quick & dirty fix without adding too much code.

Maybe fixes #12648 .

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -489,7 +489,7 @@ bool TarScanner::AddFile(const std::string &filename, size_t, [[maybe_unused]] c
 
 	_tar_list[this->subdir][filename] = std::string{};
 
-	std::string filename_base = std::filesystem::path(filename).filename().string();
+	std::string filename_base = FS2OTTD(std::filesystem::path(OTTD2FS(filename)).filename());
 	SimplifyFileName(filename_base);
 
 	TarHeader th;
--- a/src/game/game_text.cpp
+++ b/src/game/game_text.cpp
@@ -386,7 +386,7 @@ void ReconsiderGameScriptLanguage()
 {
 	if (_current_data == nullptr) return;
 
-	std::string language = _current_language->file.stem().string();
+	std::string language = FS2OTTD(_current_language->file.stem());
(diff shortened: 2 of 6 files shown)


glx22:
[openttd-20240509-pr12650-g613cc82705-windows-win64.zip](https://github.com/OpenTTD/OpenTTD/files/15266643/openttd-20240509-pr12650-g613cc82705-windows-win64.zip)

koutya0125:
I tried it and it started up fine.
But, I have not been able to verify that newGRF can be successfully downloaded when the document folder is in multibyte characters, since changing the documents folder requires editing in the registry.
Is it possible to change files that are normally stored in the documents folder (e.g. newGRF) to any folder?

PeterN:
Not sure if this fixes all problems but it seems to go some way to help.

rubidium42 (approved):

PR #12652, opened 2024-05-10
Codechange: Use vector/string instead of malloc when reading music cat file data.

PeterN:
## Motivation / Problem

`GetMusicCatEntryName()` and `GetMusicCatEntryData()` both allocate memory with malloc, which needs to be freed by the caller.

## Description

Instead return the data as `std:;string` or `std::vector`, so no manual cleaning up is necessary.

## Checklist for review

### DIFF
--- a/src/music.cpp
+++ b/src/music.cpp
@@ -8,6 +8,7 @@
 /** @file music.cpp The songs that OpenTTD knows. */
 
 #include "stdafx.h"
+#include "string_func.h"
 
 
 /** The type of set we're replacing */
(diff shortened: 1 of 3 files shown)


src/music.cpp
```diff
@@ -138,13 +133,12 @@ bool MusicSet::FillSetDetails(const IniFile &ini, const std::string &path, const
 				/* Song has a CAT file index, assume it's MPS MIDI format */
 				this->songinfo[i].filetype = MTT_MPSMIDI;
 				this->songinfo[i].cat_index = atoi(item->value->c_str());
-				char *songname = GetMusicCatEntryName(filename, this->songinfo[i].cat_index);
-				if (songname == nullptr) {
+				auto songname = GetMusicCatEntryName(filename, this->songinfo[i].cat_index);
+				if (songname.has_value()) {
```
rubidium42:
```suggestion
				if (!songname.has_value()) {
```

src/music.cpp
```diff
@@ -22,55 +22,50 @@
...
-	return nullptr;
+	if (entrynum >= entry_count) return std::nullopt;
+
+	file.SeekTo(entrynum * 8, SEEK_SET);
+	file.SeekTo(file.ReadDword(), SEEK_SET);
+	uint8_t namelen = file.ReadByte();
+
+	std::string name(namelen, '\0');
+	file.ReadBlock(name.data(), namelen);
+	return name;
```
rubidium42: Probably not a big issue, but if `name` contains a `'\0'` somewhere internally you'd have a '\0' in the middle of a string. Previously that would be the end of the string, regardless of the extra space you allocated.

Next to that we should always make sure that the strings coming from the outside are valid UTF-8. So it should have been done here, but sadly wasn't.
```suggestion
	return StrMakeValid(name);
```

rubidium42 (approved):

PR #12653, opened 2024-05-11
Add: [NewGRF] Extended custom waypoint classes.

PeterN:
## Motivation / Problem

Authors have asked for more custom waypoint classes, as lumping all waypoints into just one class gives a very big list.

The simplest way to achieve this, instead of adding more properties, is to allow more classes to be treated as waypoints.

## Description

With this PR, station class labels with the first byte set to 0xFF will be treated in the same way as the 'WAYP' class. Also applies to road stop waypoints, although those are not supported yet.

This allows waypoints to be split into categories just like stations, instead of all being lumped together.

image

Modified version of UK Waypoints for testing:

[UK_Waypoints-1.zip](https://github.com/OpenTTD/OpenTTD/files/15281355/UK_Waypoints-1.zip)

## Limitations

My review of all the custom station and roadstop NewGRF files I have available showed that they all use only ASCII characters for the class label. Therefore, the 0xFF marker was chosen because it minimises the chance of conflicts with existing NewGRFs... but it's not impossible.

## Checklist for review

### DIFF
--- a/src/newgrf_roadstop.h
+++ b/src/newgrf_roadstop.h
@@ -188,7 +188,7 @@ void RoadStopUpdateCachedTriggers(BaseStation *st);
  */
 inline bool IsWaypointClass(const RoadStopClass &cls)
 {
-	return cls.global_id == ROADSTOP_CLASS_LABEL_WAYPOINT;
+	return cls.global_id == ROADSTOP_CLASS_LABEL_WAYPOINT || GB(cls.global_id, 24, 8) == UINT8_MAX;
 }
 
 #endif /* NEWGRF_ROADSTATION_H */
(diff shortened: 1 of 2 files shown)


rubidium42 (approved):

2TallTyler:
How does this interact with #12618? Do we need both?

PeterN:
> How does this interact with #12618? Do we need both?

#12618 would need to be applied only to 14.x (or applied to master and then reverted again)

2TallTyler:
Could we just apply this PR to 14.2, or does it rely on something that's not in 14.x?

PeterN:
Nope, 14.x does not have `IsWaypointClass()` nor the new class/type picker—the old waypoint UI in 14.x can only show 'WAYP'.

PR #12657, opened 2024-05-11
Feature: Add the ability to clone an area with structures

StanislavShamrai:
## Motivation / Problem

@StanislavShamrai wants to automate a process by creating a new object using existing logic.

## Description

You can build larger paths by simply copying and pasting some of the existing intersections and paths.
You can rotate an area by inserting it at a different angle.

## Limitations

This was created for railroad tracks, tunnels, and bridges, but the same behavior may be added for other types of tracks and structures in the future.

## Checklist for review

example:
[clone-area.webm](https://github.com/OpenTTD/OpenTTD/assets/85177194/0f189ceb-4ed9-45de-b121-86794a60e8a9)


src/terraform_gui.cpp
```diff
@@ -295,6 +325,19 @@ struct TerraformToolbarWindow : Window {
...
+	{
+		if (value) {
+			this->EnableWidget(WID_TT_CLONE_AREA_PASTE);
+			this->RaiseWidget(WID_TT_CLONE_AREA_PASTE);
+			this->SetWidgetDirty(WID_TT_CLONE_AREA_PASTE);
+			return;
+		}
+		this->DisableWidget(WID_TT_CLONE_AREA_PASTE);
+		this->RaiseWidget(WID_TT_CLONE_AREA_PASTE);
+		this->SetWidgetDirty(WID_TT_CLONE_AREA_PASTE);
```
PeterN:
```suggestion
this->SetWidgetDisabledState(WID_TT_CLONE_AREA_PASTE, !value);
this->RaiseWidget(WID_TT_CLONE_AREA_PASTE);
this->SetWidgetDirty(WID_TT_CLONE_AREA_PASTE);
```

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,561 @@
...
+					return TRACK_RIGHT;
+					break;
+				case TRACK_LEFT:
+					return TRACK_LOWER;
+					break;
+				case TRACK_RIGHT:
+					return TRACK_UPPER;
+					break;
+				default: break;
+			}
```
PeterN:
```suggestion
switch (track) {
    case TRACK_UPPER: return TRACK_LEFT;
    case TRACK_LOWER: return TRACK_RIGHT;
    case TRACK_LEFT: return TRACK_LOWER;
    case TRACK_RIGHT: return TRACK_UPPER;
    default: break;
}
```

And similar below

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,561 @@
...
+			rotated.y = - point.y;
+			break;
+		case DIAGDIRDIFF_90RIGHT:
+			rotated.x = point.y;
+			rotated.y = - point.x;
+			break;
+		default:
+			return point;
+	}
+	return rotated;
```
PeterN:
```suggestion
switch (angle) {
    case DIAGDIRDIFF_90LEFT: return {-point.y, point.x};
    case DIAGDIRDIFF_REVERSE: return {-point.x, -point.y};
    case DIAGDIRDIFF_90RIGHT: return {point.y, -point.x};
    default: return point;
}
```

PeterN (commented):
Seems well written, though maybe a bit repetitive (I guess that's the point :D)

src/clone_area_cmd.cpp
StanislavShamrai: It wasn't a good idea
Warning: narrowing conversion of '-(int)point.TileIndexDiffC::y' from 'int' to 'int16_t' {aka 'short int'} [-Wnarrowing]

PeterN:
The list of commits looks a bit messy, did you mean to squash it again?

Additionally there are a lot of seemingly unrelated changes into table/sprites.h which might be better in a separate PR?

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+#include "road.h"
+#include "road_internal.h"
+#include "depot_base.h"
+#include "tunnelbridge_map.h"
+#include "sound_func.h"
+#include "rail_cmd.h"
+#include "tunnelbridge_cmd.h"
+#include "station_cmd.h"
+#include "terraform_cmd.h"
+
```
2TallTyler, line 26: Every file should have `#include "safeguards.h"` as its last include.

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+TileIndex selected_tile;
+TileIndex selected_start_tile;
+bool selected_diagonal;
```
2TallTyler, line 29: Globals should be avoided if possible, and if used, need an underscore prefix, like `TileIndex _selected_tile;`.

I wonder, can these be stored as properties of `TerraformToolbarWindow`, like `last_user_action`? They'd then be passed as command parameters.

The copied area would be lost when the terraform toolbar is closed, although that is already the case (possibly just because the button is disabled when the toolbar is initialized).

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+	}
+}
+
+/**
+ * Adjusting the position of the rotated point, since the tile has four corners.
+ *
+ * @param rotated Rotated point setting.
+ * @param angle Angle of rotation.
+ */
+TileIndexDiffC FixAfterRotate(TileIndexDiffC rotated, DiagDirDiff angle) {
```
2TallTyler, line 52: This is a bit messy as a separate function. I would include it as part of `Rotate()` and use a bool parameter to use the additional logic only when required. The name of the parameter, and the comments, should explain why this logic is only sometimes required.

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+}
+
+/**
+ * Rotate Track by an angle using the formula for rotating a point on a plane.
+ *
+ * @param track Rotating track.
+ * @param angle Rotate an angle.
+ */
+Track Rotate(Track track, DiagDirDiff angle) {
+	switch (angle) {
```
2TallTyler, line 77: This nesting is hard to read, especially since each branch of the top-level switch returns. I would get rid of the top-level switch and use a series of `if` statements for the angles (no `if else` or `else`, the returns make that unnecessary).

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+/**
+ * Rotate Axis by an angle using the formula for rotating a point on a plane.
+ *
+ * @param axis Rotating axis.
+ * @param angle Rotate an angle.
+ */
+Axis Rotate(Axis axis, DiagDirDiff angle) {
+	if (angle == DIAGDIRDIFF_90LEFT || angle == DIAGDIRDIFF_90RIGHT) {
+		return (Axis)(axis ^ 1);
```
2TallTyler:
```suggestion
		return static_cast<Axis>(axis ^ 1);
```

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,533 @@
...
+ */
+std::tuple<CommandCost, Money, TileIndex> CmdCloneAreaPasteProperty(DoCommandFlag flags, TileIndex tile, TileIndex area_start, [[maybe_unused]] bool diagonal)
+{
+	if (area_start >= Map::Size()) return { CMD_ERROR, 0, INVALID_TILE };
+
+	CommandCost cost(EXPENSES_CONSTRUCTION);
+	CommandCost last_error(INVALID_STRING_ID);
+	bool had_success = false;
+	bool auto_remove_signals = true;
+	bool signals_copy = true;
```
2TallTyler, line 356: Neither of these are ever changed, they are always true. Unfinished work?

src/command_type.h
```diff
@@ -317,6 +317,8 @@ enum Commands : uint16_t {
 	CMD_STORY_PAGE_BUTTON,            ///< selection via story page button
 
 	CMD_LEVEL_LAND,                   ///< level land
+	CMD_CLONE_AREA_COPY,              ///< Clone land
+	CMD_CLONE_AREA_PASTE,             ///< Clone land
```
2TallTyler:
```suggestion
	CMD_CLONE_AREA_COPY,              ///< copy land
	CMD_CLONE_AREA_PASTE,             ///< paste land
```

src/lang/english.txt
```diff
@@ -4971,6 +4973,8 @@ STR_ERROR_TREE_PLANT_LIMIT_REACHED                              :{WHITE}... tree
 STR_ERROR_NAME_MUST_BE_UNIQUE                                   :{WHITE}Name must be unique
 STR_ERROR_GENERIC_OBJECT_IN_THE_WAY                             :{WHITE}{1:STRING} in the way
 STR_ERROR_NOT_ALLOWED_WHILE_PAUSED                              :{WHITE}Not allowed while paused
+STR_ERROR_CAN_T_CLONE_AREA_COPY                                 :{WHITE}Can't Copy this area...
+STR_ERROR_CAN_T_CLONE_AREA_PASTE                                :{WHITE}Can't Paste this area...
```
```suggestion
STR_ERROR_CAN_T_CLONE_AREA_COPY                                 :{WHITE}Can't copy this area...
STR_ERROR_CAN_T_CLONE_AREA_PASTE                                :{WHITE}Can't paste this area...
```

src/lang/english.txt
```diff
@@ -2966,6 +2966,8 @@ STR_LANDSCAPING_TOOLTIP_LOWER_A_CORNER_OF_LAND                  :{BLACK}Lower a
 STR_LANDSCAPING_TOOLTIP_RAISE_A_CORNER_OF_LAND                  :{BLACK}Raise a corner of land. Click+Drag to raise the first selected corner and level the selected area to the new corner height. Ctrl+Click+Drag to select the area diagonally. Also press Shift to show cost estimate only
 STR_LANDSCAPING_LEVEL_LAND_TOOLTIP                              :{BLACK}Level an area of land to the height of the first selected corner. Ctrl+Click+Drag to select the area diagonally. Also press Shift to show cost estimate only
 STR_LANDSCAPING_TOOLTIP_PURCHASE_LAND                           :{BLACK}Purchase land for future use. Ctrl+Click+Drag to select the area diagonally. Also press Shift to show cost estimate only
+STR_LANDSCAPING_TOOLTIP_CLONE_AREA_COPY                         :{BLACK}Copy Area
```
```suggestion
STR_LANDSCAPING_TOOLTIP_CLONE_AREA_COPY                         :{BLACK}Copy an area of rail infrastructure and land. Ctrl+Click+Drag to select the area diagonally. Also press Shift to show cost estimate only
```

2TallTyler (commented):
Thanks for the contribution! We see players request this regularly, and this looks like a well-structured approach to providing the feature. Some thoughts and questions:

* This feature only affects railway infrastructure: roads, objects, etc., are not copied. Why not?
* After selecting an area to be copied, it would be very helpful to show a preview of the size of the area which will be pasted, and in which orientation. Maybe even show a "ghost" of the infrastructure to be built, if that's possible.
* After a lot of experimentation, I discovered that dragging an area to paste controls rotation. This is not obvious and would be incompatible with the preview suggested above. I would suggest a new mini toolbar (like how the Signals button is a mini toolbar for the Rail toolbar) that has rotation buttons.
* Pasting an area that affects tiles outside the game world frequently (but not always) causes a crash.
* It doesn't cost anything to clone an area, even when building new structures.

src/clone_area_cmd.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+/**
+ * Insert structures into the selected area on the map.
+ * @param flags operation to perform
+ * @param tile end tile of rail conversion drag
+ * @param area_start start tile of drag
+ * @param diagonal build diagonally or not.
+ * @return the cost of this operation or an error
+ */
+std::tuple<CommandCost, Money, TileIndex> CmdCloneAreaPasteProperty(DoCommandFlag flags, TileIndex tile, TileIndex area_start, [[maybe_unused]] bool diagonal)
```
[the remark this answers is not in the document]
2TallTyler: I agree with this warning. More comments would be very helpful. 😃

src/clone_area_cmd.cpp
StanislavShamrai: there is a lot of code from the existing code base
I don't have time to improve every line of code

StanislavShamrai:
there is a lot of code from the existing code base
I don't have time to improve every line of code
I may come back to this code later, but if you want to make any changes, that's up to you.

Rito13:
Any thoughts on that?

PR #12661, opened 2024-05-11
Feature: Allow manually placing town buildings in scenario editor.

PeterN:
## Motivation / Problem

Very little manual control over where houses are placed in towns. This is mostly to be expected as the game is meant to manage towns itself, however some manual control in the Scenario Editor (and perhaps a Sandbox mode) might be useful.

## Description

* Add a network command to manually place houses. (Not strictly necessary for Scenario Editor but very useful for a sandbox mode.)
* Allow Houses to be resolved without a Town and Tile.
* Add a Town House builder to the Scenario Editor. This is accessed from the Landscaping toolbar as there is no town toolbar.

Once placed these houses behave like any other and can be removed by players and towns.

Uses the unified picker system, so also supports used/saved favourites. As town building don't have class labels, town zones are use to imitate them.

Houses sometimes have up to 4 variations, selected psuedo randomly based on tile. The House Selection window does not provide a way to choose between these, as there is no manual control, so in order to represent them in the UI it will cycle through each variation every 2.5 seconds.

"Some highly advanced town planning"
image

More of the same, with NewGRFs...
image

## Limitations

This is limited to Scenario Editor only. No sandbox mode as been implemented as the Sandbox Settings window is still using the old cheats system.

Once that is changed to use normal settings, not much effort is required to enable this.

## Checklist for review

### DIFF
--- a/src/newgrf_house.cpp
+++ b/src/newgrf_house.cpp
@@ -105,11 +105,14 @@ void ResetHouses()
  */
 HouseResolverObject::HouseResolverObject(HouseID house_id, TileIndex tile, Town *town,
 		CallbackID callback, uint32_t param1, uint32_t param2,
-		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers)
+		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers, int view)
 	: ResolverObject(GetHouseSpecGrf(house_id), callback, param1, param2),
-	house_scope(*this, house_id, tile, town, not_yet_constructed, initial_random_bits, watched_cargo_triggers),
+	house_scope(*this, house_id, tile, town, not_yet_constructed, initial_random_bits, watched_cargo_triggers, view),
 	town_scope(*this, town, not_yet_constructed) // Don't access StorePSA if house is not yet constructed.
 {
+	/* Tile must be valid and a house tile, unless not yet constructed in which case it may also be INVALID_TILE. */
+	assert((IsValidTile(tile) && (not_yet_constructed || IsTileType(tile, MP_HOUSE))) || (not_yet_constructed && tile == INVALID_TILE));
+
 	this->root_spritegroup = HouseSpec::Get(house_id)->grf_prop.spritegroup[0];
 }
 
@@ -210,14 +213,12 @@ void DecreaseBuildingCount(Town *t, HouseID house_id)
 /* virtual */ uint32_t HouseScopeResolver::GetRandomBits() const
 {
 	/* Note: Towns build houses over houses. So during construction checks 'tile' may be a valid but unrelated house. */
-	assert(IsValidTile(this->tile) && (this->not_yet_constructed || IsTileType(this->tile, MP_HOUSE)));
 	return this->not_yet_constructed ? this->initial_random_bits : GetHouseRandomBits(this->tile);
 }
 
 /* virtual */ uint32_t HouseScopeResolver::GetTriggers() const
 {
 	/* Note: Towns build houses over houses. So during construction checks 'tile' may be a valid but unrelated house. */
-	assert(IsValidTile(this->tile) && (this->not_yet_constructed || IsTileType(this->tile, MP_HOUSE)));
 	return this->not_yet_constructed ? 0 : GetHouseTriggers(this->tile);
 }
 
@@ -362,6 +363,32 @@ static uint32_t GetDistanceFromNearbyHouse(uint8_t parameter, TileIndex tile, Ho
  */
 /* virtual */ uint32_t HouseScopeResolver::GetVariable(uint8_t variable, [[maybe_unused]] uint32_t parameter, bool *available) const
 {
+	if (this->tile == INVALID_TILE) {
+		/* House does not yet exist, nor is it being planned to exist. Provide some default values intead. */
+		switch (variable) {
+			case 0x40: return TOWN_HOUSE_COMPLETED | this->view << 2; /* Construction stage. */
+			case 0x41: return 0;
+			case 0x42: return 0;
+			case 0x43: return 0;
+			case 0x44: return 0;
+			case 0x45: return _generating_world ? 1 : 0;
+			case 0x46: return 0;
+			case 0x47: return 0;
+			case 0x60: return 0;
+			case 0x61: return 0;
+			case 0x62: return 0;
+			case 0x63: return 0;
+			case 0x64: return 0;
+			case 0x65: return 0;
+			case 0x66: return 0xFFFFFFFF; /* Class and ID of nearby house. */
+			case 0x67: return 0;
+		}
+
+		Debug(grf, 1, "Unhandled house variable 0x{:X}", variable);
+		*available = false;
+		return UINT_MAX;
+	}
+
 	switch (variable) {
 		/* Construction stage. */
 		case 0x40: return (IsTileType(this->tile, MP_HOUSE) ? GetHouseBuildingStage(this->tile) : 0) | TileHash2Bit(TileX(this->tile), TileY(this->tile)) << 2;
@@ -481,12 +508,10 @@ static uint32_t GetDistanceFromNearbyHouse(uint8_t parameter, TileIndex tile, Ho
 }
 
 uint16_t GetHouseCallback(CallbackID callback, uint32_t param1, uint32_t param2, HouseID house_id, Town *town, TileIndex tile,
-		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers)
+		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers, int view)
 {
-	assert(IsValidTile(tile) && (not_yet_constructed || IsTileType(tile, MP_HOUSE)));
-
 	HouseResolverObject object(house_id, tile, town, callback, param1, param2,
-			not_yet_constructed, initial_random_bits, watched_cargo_triggers);
+			not_yet_constructed, initial_random_bits, watched_cargo_triggers, view);
 	return object.ResolveCallback();
 }
 
--- a/src/newgrf_house.cpp
+++ b/src/newgrf_house.cpp
@@ -173,6 +173,15 @@ void InitializeBuildingCounts()
 }
 
 /**
+ * Get read-only span of total HouseID building counts.
+ * @return span of HouseID building counts.
+ */
+std::span<const uint> GetBuildingHouseIDCounts()
+{
+	return _building_counts.id_count;
+}
+
+/**
  * IncreaseBuildingCount()
  * Increase the count of a building when it has been added by a town.
  * @param t The town that the building is being built in
--- a/src/terraform_gui.cpp
+++ b/src/terraform_gui.cpp
@@ -12,6 +12,7 @@
 #include "clear_map.h"
 #include "company_func.h"
 #include "company_base.h"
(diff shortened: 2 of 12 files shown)


src/town_cmd.cpp
```diff
@@ -2665,6 +2665,31 @@ static bool CheckTownBuild2x2House(TileIndex *tile, Town *t, int maxz, bool nosl
 	return false;
 }
 
+static void BuildTownHouse(Town *t, TileIndex tile, const HouseSpec *hs, HouseID house, uint8_t random_bits)
```
2TallTyler, line 2676: Missing Doxygen, and maybe should have a different name from the old `BuildTownHouse()`... Maybe the other should be renamed `TrySpawnTownHouse()` or similar?

src/town_cmd.cpp
```diff
@@ -2784,36 +2809,68 @@ static bool BuildTownHouse(Town *t, TileIndex tile)
...
 
-		MakeTownHouse(tile, t, construction_counter, construction_stage, house, random_bits);
-		UpdateTownRadius(t);
-		UpdateTownGrowthRate(t);
+	if (static_cast<size_t>(house) >= HouseSpec::Specs().size()) return CMD_ERROR;
+	const HouseSpec *hs = HouseSpec::Get(house);
+	if (!hs->enabled) return CMD_ERROR;
 
-		return true;
+	if (TimerGameCalendar::year < hs->min_year || TimerGameCalendar::year > hs->max_year) return CMD_ERROR;
```
2TallTyler, line 2840: A lot of these checks are duplicated from BuildTownHouse(). Could they be refactored into a shared `bool CheckTownHouseBuild()` or something?

src/newgrf_house.cpp
```diff
@@ -105,11 +105,14 @@ void ResetHouses()
...
 HouseResolverObject::HouseResolverObject(HouseID house_id, TileIndex tile, Town *town,
 		CallbackID callback, uint32_t param1, uint32_t param2,
-		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers)
+		bool not_yet_constructed, uint8_t initial_random_bits, CargoTypes watched_cargo_triggers, int view)
 	: ResolverObject(GetHouseSpecGrf(house_id), callback, param1, param2),
-	house_scope(*this, house_id, tile, town, not_yet_constructed, initial_random_bits, watched_cargo_triggers),
+	house_scope(*this, house_id, tile, town, not_yet_constructed, initial_random_bits, watched_cargo_triggers, view),
 	town_scope(*this, town, not_yet_constructed) // Don't access StorePSA if house is not yet constructed.
 {
+	/* Tile must be valid and a house tile, unless not yet construct in which case it may also be INVALID_TILE. */
```
2TallTyler:
```suggestion
	/* Tile must be valid and a house tile, unless not yet constructed in which case it may also be INVALID_TILE. */
```

src/terraform_gui.cpp
```diff
@@ -475,6 +476,8 @@ static constexpr NWidgetPart _nested_scen_edit_land_gen_widgets[] = {
 			EndContainer(),
 			NWidget(WWT_PUSHIMGBTN, COLOUR_GREY, WID_ETT_PLACE_OBJECT), SetMinimalSize(23, 22),
 										SetFill(0, 1), SetDataTip(SPR_IMG_TRANSMITTER, STR_SCENEDIT_TOOLBAR_PLACE_OBJECT),
+			NWidget(WWT_PUSHIMGBTN, COLOUR_DARK_GREEN, WID_ETT_PLACE_HOUSE), SetMinimalSize(22, 22),
+					SetFill(0, 1), SetDataTip(SPR_IMG_TOWN, STR_SCENEDIT_TOOLBAR_PLACE_OBJECT),
```
2TallTyler:
```suggestion
					SetFill(0, 1), SetDataTip(SPR_IMG_TOWN, STR_SCENEDIT_TOOLBAR_PLACE_HOUSE),
```

src/town_cmd.cpp
```diff
@@ -4061,3 +4062,98 @@ extern const TileTypeProcs _tile_type_town_procs = {
...
+ * @param house_id House ID to draw.
+ * @param view The house's 'view'.
+ */
+void DrawHouseInGUI(int x, int y, HouseID house_id, int view)
+{
+	auto draw = [](int x, int y, HouseID house_id, int view) {
+		if (house_id >= NEW_HOUSE_OFFSET) {
+			/* Houses don't necessarily need new graphics. If they don't have a
+			* spritegroup associated with them, then the sprite for the substitute
+			* house id is drawn instead. */
```
2TallTyler:
```suggestion
			 * spritegroup associated with them, then the sprite for the substitute
			 * house id is drawn instead. */
```

src/town_cmd.cpp
```diff
@@ -4061,3 +4062,98 @@ extern const TileTypeProcs _tile_type_town_procs = {
...
+
+/**
+ * Draw representation of a house tile for GUI purposes.
+ * @param x Position x of image.
+ * @param y Position y of image.
+ * @param spec House spec to draw.
+ * @param house_id House ID to draw.
+ * @param view The house's 'view'.
+ */
+void DrawNewHouseTileInGUI(int x, int y, const HouseSpec *spec, HouseID house_id, int view)
```
2TallTyler: Should `DrawNewHouseTileInGUI()` and `DrawHouseInGUI()` be in `town_gui.cpp` instead of `town_cmd.cpp`, since they're related to the user interface and not commands?

src/terraform_gui.cpp
```diff
@@ -475,6 +476,8 @@ static constexpr NWidgetPart _nested_scen_edit_land_gen_widgets[] = {
 			EndContainer(),
 			NWidget(WWT_PUSHIMGBTN, COLOUR_GREY, WID_ETT_PLACE_OBJECT), SetMinimalSize(23, 22),
 										SetFill(0, 1), SetDataTip(SPR_IMG_TRANSMITTER, STR_SCENEDIT_TOOLBAR_PLACE_OBJECT),
+			NWidget(WWT_PUSHIMGBTN, COLOUR_DARK_GREEN, WID_ETT_PLACE_HOUSE), SetMinimalSize(22, 22),
```
2TallTyler:
```suggestion
			NWidget(WWT_PUSHIMGBTN, COLOUR_GREY, WID_ETT_PLACE_HOUSE), SetMinimalSize(22, 22),
```

2TallTyler (commented):
This is a great feature. A few thoughts:

* Adding this to Scenario Editor first and then adding it to normal gameplay as a later PR makes sense.
* Adding the button to the object toolbar works, but another option (which might make more sense) is adding it to the Town button dropdown. In SE there is no dropdown since there's only one option, but in-game you long-click the Town button to open a dropdown containing `Town directory` and `Found town`. This is where I would expect to find `Build houses`.

src/town_cmd.cpp
PeterN: They are very similar but with a few differences.

Some of the tests aren't performed because we let the user do more than towns.
TryBuildTownHouse just continues its loop, by CmdPlaceHouse needs to return error strings.

I did consider making TryBuildTownHouse issue a Command<CMD_PLACE_HOUSE>::Do. but that complicates CmdPlaceHouse more -- e.g TryBuildTownHouse only evaluates the slope and some other things once, not every attempt to build.

src/town_cmd.cpp
```diff
@@ -2784,36 +2817,68 @@ static bool TryBuildTownHouse(Town *t, TileIndex tile)
...
+
+	int maxz = GetTileMaxZ(tile);
+
+	/* Make sure there is no slope? */
+	bool noslope = (hs->building_flags & TILE_NOT_SLOPED) != 0;
+	if (noslope && slope != SLOPE_FLAT) return_cmd_error(STR_ERROR_FLAT_LAND_REQUIRED);
+
+	TileIndex origin_tile = tile;
+	if (hs->building_flags & TILE_SIZE_2x2) {
+		if (!CheckTownBuild2x2House(&tile, t, maxz, noslope)) return_cmd_error(STR_ERROR_CAN_T_CLEAR_THIS_AREA);
```
2TallTyler: Should this (and the others nearby) return the error string `STR_ERROR_CAN_T_BUILD_HOUSE` instead?

src/town_gui.cpp
```diff
@@ -1304,3 +1307,319 @@ void InitializeTownGui()
...
+		Hotkey('F', "focus_filter_box", PCWHK_FOCUS_FILTER_BOX),
+	}};
+};
+
+/** Nested widget definition for the build NewGRF rail waypoint window */
+static constexpr NWidgetPart _nested_build_house_widgets[] = {
+	NWidget(NWID_HORIZONTAL),
+		NWidget(WWT_CLOSEBOX, COLOUR_DARK_GREEN),
+		NWidget(WWT_CAPTION, COLOUR_DARK_GREEN), SetDataTip(STR_HOUSE_PICKER_CAPTION, STR_TOOLTIP_WINDOW_TITLE_DRAG_THIS),
+		NWidget(WWT_DEFSIZEBOX, COLOUR_DARK_GREEN),
```
2TallTyler, line 1606: Other pickers have pin and shade buttons, but these have gone missing.

2TallTyler (commented):
I'd still prefer to see this accessed from a dropdown in the town button of the main toolbar, but if you feel strongly otherwise I'm open to having my mind changed. 😄

src/town_cmd.cpp
PeterN: No, it's a sub-error, although actually it was an incorrect string for a sub-error.

So I've changed this instead to actually call CMD_CLEAR for each tile to get a decent error message.

src/town_gui.cpp
PeterN: Only objects have the pin (people apparently like the object picker to stay open regardless of what toolbar is active).

Waypoint doesn't have the shadebox either, which is were I copied these widgets from, but I'll add it.

PeterN:
> I'd still prefer to see this accessed from a dropdown in the town button of the main toolbar, but if you feel strongly otherwise I'm open to having my mind changed. 😄

I've moved it to a toolbar menu now so you can try it. Although after doing so I noticed that JGRPP has a house picker in the same spot as it was, heh...

src/town_cmd.cpp
```diff
@@ -2665,14 +2665,47 @@ static bool CheckTownBuild2x2House(TileIndex *tile, Town *t, int maxz, bool nosl
 	return false;
 }
 
+/**
+ * Build a house at this tile.
+ * @param t The town the house will belong to.
+ * @param tile The tile to try building on.
+ * @param hs The @a HouseSpec of the house.
+ * @param house The @a HouseID of the house.
+ * @param random_bits The random data to associated with the house.
```
2TallTyler:
```suggestion
 * @param random_bits The random data to be associated with the house.
```

src/town_gui.cpp
2TallTyler: Ah, didn't notice the pin difference. That makes sense since only one transport type toolbar is visible at once -- if you have the road toolbar open and click on airports, that toolbar opens and the road toolbar is closed.

That said, houses don't have a transport type. Given that the house placer will likely be used for decorating cities similar to objects, I think it's appropriate for it to be pinnable. I can't speak to others' play styles, but I would personally like to pin it. 🙂

2TallTyler:
Oh yes, I like the toolbar placement much better than in the landscaping toolbar. 😃

2TallTyler (approved):

PR #12667, opened 2024-05-12
Codechange: use std::vector instead of malloced array

rubidium42:
## Motivation / Problem

Manual memory management in the binary heap used by YAPF.

## Description

Just use a `std::vector` as back end. Also replace `uint` with `size_t` where applicable.

## Limitations

This implementation does not actually use element 0 to keep the other logic much simpler (fewer potential off-by-one errors). I left the logic the same, as changing the logic will impact logic in other places of the code due to for example indices shifting.

I have tried replacing this implementation with `std::priority_queue`, but that is not feasible as for the current path finding purposes we want to be able to remove elements from the queue to insert another node with a lower cost. Or because the (previously best) node is only removed from the queue after a cycle of finding open nodes has been done.

## Checklist for review

### DIFF
--- a/src/misc/binaryheap.hpp
+++ b/src/misc/binaryheap.hpp
@@ -10,8 +10,6 @@
 #ifndef BINARYHEAP_HPP
 #define BINARYHEAP_HPP
 
-#include "../core/alloc_func.hpp"
-
 /** Enable it if you suspect binary heap doesn't work well */
 #define BINARYHEAP_CHECK 0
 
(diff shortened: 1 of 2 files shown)


src/misc/binaryheap.hpp
```diff
@@ -50,27 +52,18 @@
...
-	uint capacity; ///< Maximum number of items the heap can hold
-	T **data;      ///< The pointer to the heap item pointers
+	size_t items = 0; ///< Number of valid items in the heap
+	std::vector<T *> data; ///< The pointer to the heap item pointers
 
 public:
 	/**
 	 * Create a binary heap.
-	 * @param max_items The limit of the heap
+	 * @param initial_size The initial size of the heap
```
PeterN: Not its actual size, just reserved.

rubidium42: Renamed it to `initial_capacity`. After all, it was not a limit because it would just reallocate if more space was needed.

PeterN (approved):
Not tested but looks good to me.

PR #12671, opened 2024-05-12
Doc: Mention Release type

Berbe:
## Motivation / Problem
Enhance compiling docs

## Description
Follow-up of an IRC conversation

## Checklist for review

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -104,7 +104,9 @@ builds.
 
 - `-DCMAKE_BUILD_TYPE=RelWithDebInfo`: build a release build. This is
    significantly faster than a debug build, but has far less useful information
-   in case of a crash.
+   in case of a crash. Binaries still contain symbols, aka "unstripped".
+- `-DCMAKE_BUILD_TYPE=Release`: Same as previous option, without symbols,
+   aka "stripped".
 - `-DOPTION_DEDICATED=ON`: build OpenTTD without a GUI. Useful if you are
    running a headless server, as it requires less libraries to operate.
 - `-DOPTION_USE_ASSERTS=OFF`: disable asserts. Use with care, as assert


COMPILING.md
```diff
@@ -104,7 +104,9 @@ builds.
 
 - `-DCMAKE_BUILD_TYPE=RelWithDebInfo`: build a release build. This is
    significantly faster than a debug build, but has far less useful information
-   in case of a crash.
+   in case of a crash. Binaries still contain symbols, aka "unstripped".
+- `-DCMAKE_BUILD_TYPE=Release`: Same as previous option, without symbols,
+   aka "stripped".
```
TrueBrain, line 109: The target group of this document is people that want to self-compile OpenTTD. In most of these cases, `RelWithDebInfo` really is the better choice. `Release` is only useful if you actually want to publish your release, for which this document is not a good fit.

I am  not in favour of giving people options in documents like this, as it requires them to understand context. What are "symbols"? What is "stripped"? Most people wouldn't know nor care. The original text, most likely intentional, stayed away from these wording, and pointed out what the result was: faster, but less information in case of a crash.

Even so, `RelWithDebInfo` really is the better pick. In short: I would suggest that we don't make this change. Curious about other developer's opinions.

glx22: I tend to agree, mentioning `Release` here is not helpful, it's part of the "known" cmake options.
We mention `RelWithDebInfo` because it makes sense as default value is `Debug` and is dead slow.

PeterN: I also agree...

Berbe: > I tend to agree, mentioning `Release` here is not helpful, it's part of the "known" cmake options. We mention `RelWithDebInfo` because it makes sense as default value is `Debug` and is dead slow.

Based on the "well-known" logic, it makes no sense to mention `RelWithDebInfo` either, as it is mentioned in [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html).
If `RelWithDebInfo` is mentioned, that piece of argument does not work for `Release`.

Moreover, compiling is not merely intended for debug builds, but might also make sense for people intending to produce release builds, not relying on thrid-party compilation.

However, if you want to choose merely one option (mentioning both might be too much information), `RelWithDebInfo` looks less interesting than `Release`.

TrueBrain:
Sorry we didn't get back on a decision on this PR for this long. That is on us.

As mentioned in our comments, we prefer `RelWithDebInfo`, and although I can understand from your perspective `Release` is more interesting, it is not the direction we want to go in.

In short, closing this PR, but I do appreciate the thinking along and trying to improve the documentation :)

PR #12683, opened 2024-05-14
Fix: Timetable precision

felixprigge:
## Motivation / Problem

When using days as display style for the timetable, two lines with the same total timetable duration may run out of sync. The internal precision in the timetable calculation is done in seconds even if the display mode is days. The total duration can be one second (half a day) off without noticing due to rounding to full days in the GUI. Unless the client switches display style to seconds or ticks and checks all timetables.
Timetable1
Timetable2

## Description

First, when using days as display style for timetables, set the internal precision also to days (74 ticks). This way the total duration will be exactly what is shown in the timetable. 
Second, for savegames where this problem already occurred and for cases, where the client manually altered the seconds or ticks of the timetable, perform a check if the displayed total duration is the exact duration. If not, insert an "approx." before the days/seconds in the GUI so the client knows that the actual duration differs slightly from the presented. The client can then correct the seconds or ticks if desired.
Timetable3

## Limitations

The expression "approx." only appears for the total duration so far and not for each individual order to keep strings short and the timetable window narrow. This could still be added if desired. I would suggest defining a new "approx." string and insert it into existing strings if needed, rather than creating more duplicates.

## Checklist for review

### DIFF
--- a/src/timetable_cmd.cpp
+++ b/src/timetable_cmd.cpp
@@ -519,7 +519,8 @@ void UpdateVehicleTimetable(Vehicle *v, bool travelling)
 	/* Before modifying waiting times, check whether we want to preserve bigger ones. */
 	if (!real_current_order->IsType(OT_CONDITIONAL) &&
 			(travelling || time_taken > real_current_order->GetWaitTime() || remeasure_wait_time)) {
-		/* Round up to the smallest unit of time commonly shown in the GUI (seconds) to avoid confusion.
+		/* Round up to the unit currently shown in the GUI for days and seconds.
+		 * Round up to seconds if currently used display style is ticks.
 		 * Players timetabling in Ticks can adjust later.
 		 * For trains/aircraft multiple movement cycles are done in one
 		 * tick. This makes it possible to leave the station and process
@@ -527,7 +528,8 @@ void UpdateVehicleTimetable(Vehicle *v, bool travelling)
 		 * the timetable entry like is done for road vehicles/ships.
 		 * Thus always make sure at least one tick is used between the
 		 * processing of different orders when filling the timetable. */
-		uint time_to_set = CeilDiv(std::max(time_taken, 1), Ticks::TICKS_PER_SECOND) * Ticks::TICKS_PER_SECOND;
+		uint factor = _settings_client.gui.timetable_mode == TimetableMode::Days ? Ticks::DAY_TICKS : Ticks::TICKS_PER_SECOND;
+		uint time_to_set = CeilDiv(std::max(time_taken, 1), factor) * factor;
 
 		if (travelling && (autofilling || !real_current_order->IsTravelTimetabled())) {
 			ChangeTimetable(v, v->cur_real_order_index, time_to_set, MTF_TRAVEL_TIME, autofilling);
--- a/src/timetable_gui.cpp
+++ b/src/timetable_gui.cpp
@@ -565,7 +565,13 @@ struct TimetableWindow : Window {
 		TimerGameTick::Ticks total_time = v->orders != nullptr ? v->orders->GetTimetableDurationIncomplete() : 0;
 		if (total_time != 0) {
 			SetTimetableParams(0, 1, total_time);
-			DrawString(tr, v->orders->IsCompleteTimetable() ? STR_TIMETABLE_TOTAL_TIME : STR_TIMETABLE_TOTAL_TIME_INCOMPLETE);
+			if (!v->orders->IsCompleteTimetable()) {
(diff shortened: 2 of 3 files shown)


PeterN:
I've not looked too in depth but I don't think 'approximately' needs to be abbreviated here.

Also, we require rebase instead of merge to keep up to date :)

flowprint:
Rebase, ok. Thank for mentioning :)

src/timetable_gui.cpp
```diff
@@ -561,11 +561,29 @@ struct TimetableWindow : Window {
 	{
 		const Vehicle *v = this->vehicle;
 		Rect tr = r.Shrink(WidgetDimensions::scaled.framerect);
+		TimerGameTick::Ticks preci;
+		switch (_settings_client.gui.timetable_mode) {
+			case TimetableMode::Days :
```
LordAro: no spaces before case `:` please

src/timetable_gui.cpp
```diff
@@ -561,11 +561,29 @@ struct TimetableWindow : Window {
...
+			case TimetableMode::Ticks :
+				preci = 1;
+				break;
+		}
 
 		TimerGameTick::Ticks total_time = v->orders != nullptr ? v->orders->GetTimetableDurationIncomplete() : 0;
 		if (total_time != 0) {
 			SetTimetableParams(0, 1, total_time);
-			DrawString(tr, v->orders->IsCompleteTimetable() ? STR_TIMETABLE_TOTAL_TIME : STR_TIMETABLE_TOTAL_TIME_INCOMPLETE);
+			if (!(v->orders->IsCompleteTimetable())) {
```
LordAro:
```suggestion
			if (!v->orders->IsCompleteTimetable()) {
```

src/timetable_gui.cpp
```diff
@@ -561,11 +561,29 @@ struct TimetableWindow : Window {
 	{
 		const Vehicle *v = this->vehicle;
 		Rect tr = r.Shrink(WidgetDimensions::scaled.framerect);
+		TimerGameTick::Ticks preci;
```
LordAro: I don't think we need to worry about saving four characters from `precision`

src/timetable_cmd.cpp
```diff
@@ -519,16 +519,22 @@ void UpdateVehicleTimetable(Vehicle *v, bool travelling)
...
 		 * Thus always make sure at least one tick is used between the
 		 * processing of different orders when filling the timetable. */
-		uint time_to_set = CeilDiv(std::max(time_taken, 1), Ticks::TICKS_PER_SECOND) * Ticks::TICKS_PER_SECOND;
-
+		uint time_to_set;
+		if (_settings_client.gui.timetable_mode == TimetableMode::Days) {
+			time_to_set = CeilDiv(std::max(time_taken, 1), Ticks::DAY_TICKS) * Ticks::DAY_TICKS;
+		} else {
+			time_to_set = CeilDiv(std::max(time_taken, 1), Ticks::TICKS_PER_SECOND) * Ticks::TICKS_PER_SECOND;
+		}
```
LordAro: Feels slightly clunky to me. Maybe:

```suggestion
		uint factor = _settings_client.gui.timetable_mode == TimetableMode::Days ? Ticks::DAY_TICKS : Ticks::TICKS_PER_SECOND;
		uint time_to_set = CeilDiv(std::max(time_taken, 1), factor) * factor;
```

?

src/timetable_gui.cpp
felixprigge: Thank you for your review, LordAro. I have a quick question. Is it a good habit to initialize 'precision' as 'TimerGameTick::Ticks' here or would 'uint' be totally fine?

2TallTyler:
Commit checker is failing for trailing whitespace.

flowprint:
I tried to figure this out but can't find any trailing whitespace in the timetable_cmp.cpp around line 537. Thank you for your patience.

2TallTyler:
Try squashing your commits with an interactive rebase, and force-push, to see if that makes commit checker happy. It looks like you've fixed something in your latest commit, but the commit checker considers each commit separately.

felixprigge:
@2TallTyler Sorry for being afk for so long. I will try this weekend. The sqashing is not possible online inside github, right?

2TallTyler:
Correct, you'll need to do it locally and force-push to GitHub.

LordAro (approved):
Fwiw, rebasing is only necessary if there's a conflict - the branch doesn't need to be fully up to date to be merged

But then it is a good way of poking us, so...

felixprigge:
haha, oh sorry. I was wondering if it actually is required to be up do date at every moment. So I just tried every day.
Didn't want to annoy you guys, next time I know better. Thank you!

PR #12690, opened 2024-05-16
Feature: Add font resizing sliders to the game options UI.

ladysadie:
## Motivation / Problem

The OpentTTD's small font (FS_SMALL)  can be illegible on high resolution displays even after the interface scaling is applied.

## Description

Sliders have been added to allow the player to easily adjust the font sizes of the UI.  The prefer sprite and AA options were moved from the graphics tab to the new font since they're a font related setting.

The new UI is pictured below:
image

## Checklist for review

### DIFF
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -171,6 +171,12 @@ void ResizeFont(FontSize font_size, uint size)
 	if (_save_config) SaveToConfig();
 }
 
+/**
+ * Called to change a font or font size used by OpenTTD.
+ * @param font_size The FontSize(enum not pixel size) of the font to change.
+ * @param font The font name to use for this font.
+ * @param size The size(pixel size) to use for this font.
+ */
 void SetFont(FontSize font_size, const std::string &font, uint size)
 {
 	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -98,9 +98,32 @@ bool GetFontAAState()
 	return _fcsettings.global_aa;
 }
 
-void SetFont(FontSize fontsize, const std::string &font, uint size)
+/**
+ * Prints the font name information. Output is:
+ *
+ * The provided desc.
+ * (FontSize Enum) : Actual Font=(string name) Setting=(string name)
+ *  etc for all the font sizes.
+ *
+ * Actual font is what's currently loaded and in use. Setting is what is recorded in the config file.
+ *
+ * @param desc - A string used to give context for this debug print.
+ */
+void DebugPrintFontSettings(const std::string &desc)
 {
-	FontCacheSubSetting *setting = GetFontCacheSubSetting(fontsize);
+	Debug(fontcache, 3, "{}", desc);
+
+	for (FontSize fs = FS_BEGIN; fs < FS_END; fs++) {
+		FontCache *loaded_font = FontCache::Get(fs);
+		FontCacheSubSetting *setting = GetFontCacheSubSetting(fs);
+
+		Debug(fontcache, 3, "  {}: Actual Font=\"{}\" Setting=\"{}\"", FontSizeToName(fs), (loaded_font == NULL) ? "NULL" : loaded_font->GetFontName(), setting->font);
+	}
+}
+
+void SetFont(FontSize font_size, const std::string &font, uint size)
+{
+	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
 	bool changed = false;
 
 	if (setting->font != font) {
@@ -115,11 +138,11 @@ void SetFont(FontSize fontsize, const std::string &font, uint size)
 
 	if (!changed) return;
 
-	if (fontsize != FS_MONO) {
+	if (font_size != FS_MONO) {
 		/* Try to reload only the modified font. */
 		FontCacheSettings backup = _fcsettings;
 		for (FontSize fs = FS_BEGIN; fs < FS_END; fs++) {
-			if (fs == fontsize) continue;
+			if (fs == font_size) continue;
 			FontCache *fc = FontCache::Get(fs);
 			GetFontCacheSubSetting(fs)->font = fc->HasParent() ? fc->GetFontName() : "";
 		}
@@ -233,6 +256,8 @@ void InitFontCache(bool monospace)
 		LoadCoreTextFont(fs);
 #endif
 	}
+
+	DebugPrintFontSettings("End of initFontCache()");
 }
 
 /**
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -55,6 +55,11 @@ int FontCache::GetDefaultFontHeight(FontSize fs)
 	return _default_font_height[fs];
 }
 
+void FontCache::SetFontSize([[maybe_unused]] int pixels)
+{
+	Debug(fontcache, 0, "Font {} isn't resizable.", this->GetFontName());
+}
+
 /**
  * Get the font name of a given font size.
  * @param fs The font size to look up.
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -137,7 +137,7 @@ extern void LoadWin32Font(FontSize fs);
 extern void LoadCoreTextFont(FontSize fs);
 #endif
 
-static void LoadFontHelper(FontSize fs)
+static void LoadFontHelper([[maybe_unused]] FontSize fs)
 {
 #ifdef WITH_FREETYPE
 	LoadFreeTypeFont(fs);
@@ -215,7 +215,10 @@ bool IsDefaultFont(const FontCacheSubSetting &setting)
 uint GetFontCacheFontSize(FontSize fs)
 {
 	const FontCacheSubSetting &setting = *GetFontCacheSubSetting(fs);
-	return IsDefaultFont(setting) ? FontCache::GetDefaultFontHeight(fs) : setting.size;
+
+	if (IsDefaultFont(setting) && setting.size == 0) return FontCache::GetDefaultFontHeight(fs);
+
+	return setting.size;
 }
 
 /**
@@ -223,7 +226,7 @@ uint GetFontCacheFontSize(FontSize fs)
  * @param fs Font size.
  * @return Name of default font file.
  */
-static std::string GetDefaultTruetypeFont(FontSize fs)
+[[maybe_unused]] static std::string GetDefaultTruetypeFont(FontSize fs)
 {
 	switch (fs) {
 		case FS_NORMAL: return "OpenTTD-Sans.ttf";
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -28,6 +28,8 @@ static const int _default_font_ascender[FS_END] = { 8, 5, 15,  8};
 
 FontCacheSettings _fcsettings;
 
+static void LoadFontHelper(FontSize fs);
+
 /**
  * Create a new font cache.
  * @param fs The size of the font.
@@ -121,6 +123,26 @@ void DebugPrintFontSettings(const std::string &desc)
 	}
 }
 
+#ifdef WITH_FREETYPE
+extern void LoadFreeTypeFont(FontSize fs);
+extern void UninitFreeType();
+#elif defined(_WIN32)
+extern void LoadWin32Font(FontSize fs);
+#elif defined(WITH_COCOA)
+extern void LoadCoreTextFont(FontSize fs);
+#endif
+
+static void LoadFontHelper(FontSize fs)
+{
+#ifdef WITH_FREETYPE
+	LoadFreeTypeFont(fs);
+#elif defined(_WIN32)
+	LoadWin32Font(fs);
+#elif defined(WITH_COCOA)
+	LoadCoreTextFont(fs);
+#endif
+}
+
 void SetFont(FontSize font_size, const std::string &font, uint size)
 {
 	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
@@ -159,15 +181,6 @@ void SetFont(FontSize font_size, const std::string &font, uint size)
 	if (_save_config) SaveToConfig();
 }
 
-#ifdef WITH_FREETYPE
-extern void LoadFreeTypeFont(FontSize fs);
-extern void UninitFreeType();
-#elif defined(_WIN32)
-extern void LoadWin32Font(FontSize fs);
-#elif defined(WITH_COCOA)
-extern void LoadCoreTextFont(FontSize fs);
-#endif
-
 /**
  * Test if a font setting uses the default font.
  * @return true iff the font is not configured and no fallback font data is present.
@@ -188,7 +201,6 @@ uint GetFontCacheFontSize(FontSize fs)
 	return IsDefaultFont(setting) ? FontCache::GetDefaultFontHeight(fs) : setting.size;
 }
 
-#if defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA)
 /**
  * Get name of default font file for a given font size.
  * @param fs Font size.
@@ -204,7 +216,6 @@ static std::string GetDefaultTruetypeFont(FontSize fs)
 		default: NOT_REACHED();
 	}
 }
-#endif /* defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA) */
 
 /**
  * Get path of default font file for a given font size.
@@ -248,13 +259,8 @@ void InitFontCache(bool monospace)
 		FontCache *fc = FontCache::Get(fs);
 		if (fc->HasParent()) delete fc;
 
-#ifdef WITH_FREETYPE
-		LoadFreeTypeFont(fs);
-#elif defined(_WIN32)
-		LoadWin32Font(fs);
-#elif defined(WITH_COCOA)
-		LoadCoreTextFont(fs);
-#endif
+		LoadFontHelper(fs);
+
 	}
 
 	DebugPrintFontSettings("End of initFontCache()");
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -148,6 +148,29 @@ static void LoadFontHelper(FontSize fs)
 #endif
 }
 
+/**
+ * Called to change the size setting of a font in OpenTTD.
+ * @param font_size FontSize(enum not pixel size) of the font to change.
+ * @param size The new pixel size to use for this font.
+ */
+void ResizeFont(FontSize font_size, uint size)
+{
+	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
+
+	if (setting->size == size) return;
+
+	setting->size = size;
+
+	FontCache *loaded_font = FontCache::Get(font_size);
+	loaded_font->SetFontSize(size);
+
+	LoadStringWidthTable();
+	UpdateAllVirtCoords();
+	ReInitAllWindows(true);
+
+	if (_save_config) SaveToConfig();
+}
+
 void SetFont(FontSize font_size, const std::string &font, uint size)
 {
 	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -160,19 +160,7 @@ void SetFont(FontSize font_size, const std::string &font, uint size)
 
 	if (!changed) return;
 
-	if (font_size != FS_MONO) {
-		/* Try to reload only the modified font. */
-		FontCacheSettings backup = _fcsettings;
-		for (FontSize fs = FS_BEGIN; fs < FS_END; fs++) {
-			if (fs == font_size) continue;
-			FontCache *fc = FontCache::Get(fs);
-			GetFontCacheSubSetting(fs)->font = fc->HasParent() ? fc->GetFontName() : "";
-		}
-		CheckForMissingGlyphs();
-		_fcsettings = backup;
-	} else {
-		InitFontCache(true);
-	}
+	CheckForMissingGlyphs();
 
 	LoadStringWidthTable();
 	UpdateAllVirtCoords();
@@ -247,20 +235,20 @@ std::string GetFontCacheFontName(FontSize fs)
 
 /**
  * (Re)initialize the font cache related things, i.e. load the non-sprite fonts.
- * @param monospace Whether to initialise the monospace or regular fonts.
  */
-void InitFontCache(bool monospace)
+void InitFontCache()
 {
 	FontCache::InitializeFontCaches();
 
 	for (FontSize fs = FS_BEGIN; fs < FS_END; fs++) {
-		if (monospace != (fs == FS_MONO)) continue;
 
 		FontCache *fc = FontCache::Get(fs);
-		if (fc->HasParent()) delete fc;
+
+		if (fc->HasParent()) {
+			delete fc;
+		}
 
 		LoadFontHelper(fs);
-
 	}
 
 	DebugPrintFontSettings("End of initFontCache()");
--- a/src/fontcache.cpp
+++ b/src/fontcache.cpp
@@ -199,9 +199,10 @@ void SetFont(FontSize font_size, const std::string &font, uint size)
 
 /**
  * Test if a font setting uses the default font.
+ * @param setting The font setting to check.
  * @return true iff the font is not configured and no fallback font data is present.
  */
-static bool IsDefaultFont(const FontCacheSubSetting &setting)
+bool IsDefaultFont(const FontCacheSubSetting &setting)
 {
 	return setting.font.empty() && setting.os_handle == nullptr;
 }
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -232,6 +232,7 @@ std::string GetFontCacheFontName(FontSize fs);
 void DebugPrintFontSettings(const std::string &desc);
 bool GetFontAAState();
 void InitFontCache();
+void ResizeFont(FontSize font_size, uint size);
 void SetFont(FontSize fontsize, const std::string &font, uint size);
 void UninitFontCache();
 
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -208,6 +208,8 @@ struct FontCacheSettings {
 	bool global_aa;             ///< Whether to anti alias all font sizes.
 };
 
+const int DEFAULT_FONT_MAX_HEIGHT = 40; ///< The maximum height allowed by font sliders.
+
 extern FontCacheSettings _fcsettings;
 
 /**
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -65,6 +65,8 @@ public:
 	 */
 	virtual int GetFontSize() const { return this->height; }
 
+	virtual void SetFontSize([[maybe_unused]] int pixels);
+
 	/**
 	 * Map a SpriteID to the key
 	 * @param key The key to map to.
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -232,6 +232,7 @@ std::string GetFontCacheFontName(FontSize fs);
 void DebugPrintFontSettings(const std::string &desc);
 bool GetFontAAState();
 void InitFontCache();
+bool IsDefaultFont(const FontCacheSubSetting &setting);
 void ResizeFont(FontSize font_size, uint size);
 void SetFont(FontSize fontsize, const std::string &font, uint size);
 void UninitFontCache();
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -226,6 +226,9 @@ inline FontCacheSubSetting *GetFontCacheSubSetting(FontSize fs)
 
 uint GetFontCacheFontSize(FontSize fs);
 std::string GetFontCacheFontName(FontSize fs);
+
+void DebugPrintFontSettings(const std::string &desc);
+
 void InitFontCache(bool monospace);
 void UninitFontCache();
 
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -228,11 +228,9 @@ uint GetFontCacheFontSize(FontSize fs);
 std::string GetFontCacheFontName(FontSize fs);
 
 void DebugPrintFontSettings(const std::string &desc);
-
-void InitFontCache(bool monospace);
-void UninitFontCache();
-
 bool GetFontAAState();
+void InitFontCache(bool monospace);
 void SetFont(FontSize fontsize, const std::string &font, uint size);
+void UninitFontCache();
 
 #endif /* FONTCACHE_H */
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -229,7 +229,7 @@ std::string GetFontCacheFontName(FontSize fs);
 
 void DebugPrintFontSettings(const std::string &desc);
 bool GetFontAAState();
-void InitFontCache(bool monospace);
+void InitFontCache();
 void SetFont(FontSize fontsize, const std::string &font, uint size);
 void UninitFontCache();
 
--- a/src/fontcache.h
+++ b/src/fontcache.h
@@ -228,11 +228,10 @@ inline FontCacheSubSetting *GetFontCacheSubSetting(FontSize fs)
 	}
 }
 
-uint GetFontCacheFontSize(FontSize fs);
-std::string GetFontCacheFontName(FontSize fs);
-
 void DebugPrintFontSettings(const std::string &desc);
 bool GetFontAAState();
+std::string GetFontCacheFontName(FontSize fs);
+uint GetFontCacheFontSize(FontSize fs);
 void InitFontCache();
 bool IsDefaultFont(const FontCacheSubSetting &setting);
 void ResizeFont(FontSize font_size, uint size);
--- a/src/fontcache/freetypefontcache.cpp
+++ b/src/fontcache/freetypefontcache.cpp
@@ -32,8 +32,6 @@
 class FreeTypeFontCache : public TrueTypeFontCache {
 private:
 	FT_Face face;  ///< The font face associated with this font.
-
-	void SetFontSize(int pixels);
 	const Sprite *InternalGetGlyph(GlyphID key, bool aa) override;
 
 public:
@@ -44,6 +42,7 @@ public:
 	std::string GetFontName() override { return fmt::format("{}, {}", face->family_name, face->style_name); }
 	bool IsBuiltInFont() override { return false; }
 	const void *GetOSHandle() override { return &face; }
+	void SetFontSize(int pixels) override;
 };
 
 FT_Library _library = nullptr;
@@ -64,6 +63,8 @@ FreeTypeFontCache::FreeTypeFontCache(FontSize fs, FT_Face face, int pixels) : Tr
 
 void FreeTypeFontCache::SetFontSize(int pixels)
 {
+	this->req_size = pixels;
+
 	if (pixels == 0) {
 		/* Try to determine a good height based on the minimal height recommended by the font. */
 		int scaled_height = ScaleGUITrad(FontCache::GetDefaultFontHeight(this->fs));
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -951,6 +951,8 @@ STR_GAME_OPTIONS_TAB_GENERAL                                    :General
 STR_GAME_OPTIONS_TAB_GENERAL_TT                                 :{BLACK}Choose general settings
 STR_GAME_OPTIONS_TAB_GRAPHICS                                   :Graphics
 STR_GAME_OPTIONS_TAB_GRAPHICS_TT                                :{BLACK}Choose graphics settings
+STR_GAME_OPTIONS_TAB_FONTS                                      :Fonts
+STR_GAME_OPTIONS_TAB_FONTS_TT                                   :{BLACK}Choose font settings
 STR_GAME_OPTIONS_TAB_SOUND                                      :Sound
 STR_GAME_OPTIONS_TAB_SOUND_TT                                   :{BLACK}Choose sound and music settings
 STR_GAME_OPTIONS_TAB_SOCIAL                                     :Social
@@ -1053,8 +1055,21 @@ STR_GAME_OPTIONS_GUI_SCALE_AUTO_TOOLTIP                         :{BLACK}Check th
 STR_GAME_OPTIONS_GUI_SCALE_BEVELS                               :{BLACK}Scale bevels
 STR_GAME_OPTIONS_GUI_SCALE_BEVELS_TOOLTIP                       :{BLACK}Check this box to scale bevels by interface size
 
+STR_GAME_OPTIONS_SMALL_FONT_FRAME                               :{BLACK}Small font
+STR_GAME_OPTIONS_SMALL_FONT_TOOLTIP                             :{BLACK}Used in map legends and on graphs.
+STR_GAME_OPTIONS_SMALL_FONT_SIZE_SLIDER_TOOLTIP                 :{BLACK}Drag to change the small font size
+STR_GAME_OPTIONS_MEDIUM_FONT_FRAME                              :{BLACK}Medium font
+STR_GAME_OPTIONS_MEDIUM_FONT_TOOLTIP                            :{BLACK}Most text is displayed using this font.
+STR_GAME_OPTIONS_MEDIUM_FONT_SIZE_SLIDER_TOOLTIP                :{BLACK}Drag to change the medium font size
+STR_GAME_OPTIONS_LARGE_FONT_FRAME                               :{BLACK}Large font
+STR_GAME_OPTIONS_LARGE_FONT_TOOLTIP                             :{BLACK}Used to display News headlines.
+STR_GAME_OPTIONS_LARGE_FONT_SIZE_SLIDER_TOOLTIP                 :{BLACK}Drag to change the large font size
+STR_GAME_OPTIONS_MONOSPACED_FONT_FRAME                          :{BLACK}Monospaced font
+STR_GAME_OPTIONS_MONOSPACED_FONT_TOOLTIP                        :{BLACK}Used to display the Readme and other documents.
+STR_GAME_OPTIONS_MONOSPACED_FONT_SIZE_SLIDER_TOOLTIP            :{BLACK}Drag to change the monospaced font size
+STR_GAME_OPTIONS_FONT_SIZE                                      :{BLACK}Size:
 STR_GAME_OPTIONS_GUI_FONT_SPRITE                                :{BLACK}Use traditional sprite font
-STR_GAME_OPTIONS_GUI_FONT_SPRITE_TOOLTIP                        :{BLACK}Check this box if you prefer to use the traditional fixed-size sprite font
+STR_GAME_OPTIONS_GUI_FONT_SPRITE_TOOLTIP                        :{BLACK}Check this box if you prefer to use the traditional fixed-size sprite font instead of the default OpenTTD TrueType font
 STR_GAME_OPTIONS_GUI_FONT_AA                                    :{BLACK}Anti-alias fonts
 STR_GAME_OPTIONS_GUI_FONT_AA_TOOLTIP                            :{BLACK}Check this box to anti-alias resizable fonts
 
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -767,9 +767,9 @@ struct GameOptionsWindow : Window {
 				this->SetWidgetDisabledState(WID_GO_GUI_FONT_AA, _fcsettings.prefer_sprite);
 				this->SetDirty();
 
-				InitFontCache(false);
-				InitFontCache(true);
 				ClearFontCache();
+				InitFontCache();
+
 				CheckForMissingGlyphs();
 				SetupWidgetDimensions();
 				UpdateAllVirtCoords();
--- a/src/settings_gui.cpp
+++ b/src/settings_gui.cpp
@@ -707,7 +707,10 @@ struct GameOptionsWindow : Window {
 
 	void SetTab(WidgetID widget)
 	{
-		this->SetWidgetsLoweredState(false, WID_GO_TAB_GENERAL, WID_GO_TAB_GRAPHICS, WID_GO_TAB_FONTS, WID_GO_TAB_SOUND, WID_GO_TAB_SOCIAL);
+		this->SetWidgetsLoweredState(false, WID_GO_TAB_GENERAL, WID_GO_TAB_GRAPHICS, WID_GO_TAB_SOUND, WID_GO_TAB_SOCIAL);
+#ifdef HAS_TRUETYPE_FONT
+		this->SetWidgetsLoweredState(false, WID_GO_TAB_FONTS);
+#endif
 		this->LowerWidget(widget);
 		GameOptionsWindow::active_tab = widget;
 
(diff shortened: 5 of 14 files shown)


ladysadie:
Should I add the header guards for HAS_TRUETYPE_FONT around the font tab UI code? I considered doing it. Wasn't sure if the "juice was worth the squeeze" for it though.

src/fontcache/freetypefontcache.cpp
```diff
@@ -113,7 +113,7 @@ void FreeTypeFontCache::SetFontSize(int pixels)
 	}
 }
 
-static FT_Error LoadFont(FontSize fs, FT_Face face, const char *font_name, uint size)
```
PeterN: How is a static function inherited? This parameter is used in the Debug call.

src/fileio.cpp
```diff
@@ -37,23 +37,24 @@ extern std::string _config_file;
...
+	"newgrf" PATHSEP,                       ///< NEWGRF_DIR
+	"lang" PATHSEP,                         ///< LANG_DIR
+	"ai" PATHSEP,                           ///< AI_DIR
+	"ai" PATHSEP "library" PATHSEP,         ///< AI_LIBRARY_DIR
+	"game" PATHSEP,                         ///< GAME_DIR
+	"game" PATHSEP "library" PATHSEP,       ///< GAME_LIBRARY_DIR
+	"screenshot" PATHSEP,                   ///< SCREENSHOT_DIR
+	"social_integration" PATHSEP,           ///< SOCIAL_INTEGRATION_DIR
 };
+// This should line up with the "Subdirectory" enum defined in fileio_type.h
```
PeterN: This is all unrelated, cut this out and make a separate PR. We also prefer not to align things this way any more.

src/fileio_type.h
```diff
@@ -106,42 +109,42 @@ inline DetailedFileType GetDetailedFileType(FiosType fios_type)
...
+	AI_DIR,                 ///< Subdirectory for all %AI files
+	AI_LIBRARY_DIR,         ///< Subdirectory for all %AI libraries
+	GAME_DIR,               ///< Subdirectory for all game scripts
+	GAME_LIBRARY_DIR,       ///< Subdirectory for all GS libraries
+	SCREENSHOT_DIR,         ///< Subdirectory for all screenshots
 	SOCIAL_INTEGRATION_DIR, ///< Subdirectory for all social integration plugins
-	NUM_SUBDIRS,   ///< Number of subdirectories
-	NO_DIRECTORY,  ///< A path without any base directory
+	NUM_SUBDIRS,            ///< Number of subdirectories
+	NO_DIRECTORY,           ///< A path without any base directory
```
PeterN: Unrelated.

src/fileio_type.h
```diff
@@ -106,42 +109,42 @@ inline DetailedFileType GetDetailedFileType(FiosType fios_type)
...
-	SP_APPLICATION_BUNDLE_DIR,     ///< Search within the application bundle
-	SP_AUTODOWNLOAD_DIR,           ///< Search within the autodownload directory
-	SP_AUTODOWNLOAD_PERSONAL_DIR,  ///< Search within the autodownload directory located in the personal directory
+	SP_PERSONAL_DIR,                  ///< Search in the personal directory
+	SP_SHARED_DIR,                    ///< Search in the shared directory, like 'Shared Files' under Windows
+	SP_BINARY_DIR,                    ///< Search in the directory where the binary resides
+	SP_INSTALLATION_DIR,              ///< Search in the installation directory
+	SP_APPLICATION_BUNDLE_DIR,        ///< Search within the application bundle
+	SP_AUTODOWNLOAD_DIR,              ///< Search within the autodownload directory
+	SP_AUTODOWNLOAD_PERSONAL_DIR,     ///< Search within the autodownload directory located in the personal directory
```
PeterN: Unrelated.

src/fileio_type.h
```diff
@@ -40,6 +40,9 @@ enum DetailedFileType {
 	DFT_FIOS_DIR,    ///< A directory entry.
 	DFT_FIOS_DIRECT, ///< Direct filename.
 
+	/* Ending variable used for compile time check of this array against _fios_colours in fiose_gui.cpp */
+	DFT_END,
+
```
PeterN: Unrelated, separate PR.

src/fios_gui.cpp
```diff
@@ -238,7 +238,8 @@ static const TextColour _fios_colours[] = {
 	TC_DARK_GREEN,   // DFT_FIOS_DIR
 	TC_ORANGE,       // DFT_FIOS_DIRECT
 };
-
+/* This should align with the DetailedFileType enum defined in fileio_type.h */
+static_assert (lengthof(_fios_colours) == DFT_END);
```
PeterN: Unrelated, separate PR.

src/fontcache.cpp
```diff
@@ -213,6 +225,8 @@ void InitFontCache(bool monospace)
 		LoadCoreTextFont(fs);
 #endif
 	}
+
+	DebugPrintFontSettings("End of initFontCache()");
```
PeterN, line 292: Seems unrelated to the actual task, and not really necessary.

There are coding-style issues in your Debug function too, space comes before `*` or `&`.

src/fontcache.cpp
```diff
@@ -110,6 +113,40 @@ void DebugPrintFontSettings(const std::string& desc)
 	}
 }
 
+#ifdef WITH_FREETYPE
+extern void LoadFreeTypeFont(FontSize fs);
+extern void LoadFreeTypeFont(FontSize fs, const std::string &file_name, uint size);
```
PeterN: Does not exist.

src/fontcache.cpp
```diff
@@ -110,6 +113,40 @@ void DebugPrintFontSettings(const std::string& desc)
...
+extern void UninitFreeType();
+#elif defined(_WIN32)
+extern void LoadWin32Font(FontSize fs);
+extern void LoadWin32Font(FontSize fs, const std::string &file_name, uint size);
```
PeterN: Does not exist.

src/fontcache.cpp
```diff
@@ -110,6 +113,40 @@ void DebugPrintFontSettings(const std::string& desc)
...
+#elif defined(WITH_COCOA)
+extern void LoadCoreTextFont(FontSize fs);
+extern void LoadCoreTextFont(FontSize fs, const std::string &file_name, uint size);
```
PeterN: Does not exist.

src/fontcache.cpp
```diff
@@ -110,6 +113,40 @@ void DebugPrintFontSettings(const std::string& desc)
...
+
+static void LoadFontHelper(FontSize fs)
+{
+#ifdef WITH_FREETYPE
+		LoadFreeTypeFont(fs);
+#elif defined(_WIN32)
+		LoadWin32Font(fs);
+#elif defined(WITH_COCOA)
+		LoadCoreTextFont(fs);
+#endif
```
PeterN: Too much indentation.

src/fontcache.cpp
```diff
@@ -110,6 +113,40 @@ void DebugPrintFontSettings(const std::string& desc)
...
+static void LoadFontHelper(FontSize fs, const std::string &file_name, uint size)
+{
+#ifdef WITH_FREETYPE
+		LoadFreeTypeFont(fs, file_name, size);
+#elif defined(_WIN32)
+		LoadWin32Font(fs, file_name, size);
+#elif defined(WITH_COCOA)
+		LoadCoreTextFont(fs, file_name, size);
+#endif
+}
```
PeterN: Calls functions that don't exist, so does not need to be added.

src/fontcache.cpp
```diff
@@ -173,7 +200,21 @@ static std::string GetDefaultTruetypeFont(FontSize fs)
...
+	std::string full_font = FioFindFullPath(BASESET_DIR, font_name);
+	if (!full_font.empty()) {
+		if (size == 0) {
+			size = FontCache::GetDefaultFontHeight(fs);
+		}
+
+		LoadFontHelper(fs, full_font, size);
+	}
 #endif /* defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA) */
+}
```
PeterN: This was removed in master, does not need to be added back again.

src/fontcache.cpp
```diff
@@ -217,13 +258,12 @@ void InitFontCache(bool monospace)
...
-		LoadWin32Font(fs);
-#elif defined(WITH_COCOA)
-		LoadCoreTextFont(fs);
-#endif
+
+		if (!_fcsettings.prefer_sprite && GetFontCacheSubSetting(fs)->font.empty()) {
+			TryLoadDefaultTrueTypeFont(fs);
+		} else {
+			LoadFontHelper(fs);
+		}
```
PeterN: More readding removing code.

Just `LoadFontHelper(fs);`

src/fontcache.h
```diff
@@ -227,11 +227,9 @@ inline FontCacheSubSetting *GetFontCacheSubSetting(FontSize fs)
...
 
 void DebugPrintFontSettings(const std::string& desc);
-
-void InitFontCache(bool monospace);
-void UninitFontCache();
-
 bool GetFontAAState();
+void InitFontCache(bool monospace);
 void SetFont(FontSize fontsize, const std::string &font, uint size);
+void UninitFontCache();
```
PeterN: Why shuffle these?

src/strings.cpp
```diff
@@ -2293,7 +2293,7 @@ void CheckForMissingGlyphs(bool base_font, MissingGlyphSearcher *searcher)
 			/* Our fallback font does miss characters too, so keep the
 			 * user chosen font as that is more likely to be any good than
 			 * the wild guess we made */
-			InitFontCache(searcher->Monospace());
+			InitFontCache();
```
PeterN, line 2293: When searching for missing glyphs in the monospace font, this will now reinitialise all the font caches, not just monospace.

src/fontcache.cpp
```diff
@@ -147,6 +147,30 @@ static void LoadFontHelper(FontSize fs, const std::string &file_name, uint size)
...
+	} else {
+		LoadFontHelper(font_size);
+	}
+
+	LoadStringWidthTable();
+	UpdateAllVirtCoords();
+	ReInitAllWindows(true);
+
+	if (_save_config) SaveToConfig();
+}
```
PeterN: I think this will replace the FontCache for font_size without destroying the existing FontCache.

Resizing a font shouldn't need to load fonts again, probably you can expose SetFontSize() which all scaleable FontCaches implement.

src/fontcache.cpp
```diff
@@ -29,8 +29,7 @@ static const int _default_font_ascender[FS_END] = { 8, 5, 15,  8};
 FontCacheSettings _fcsettings;
 
 static void LoadFontHelper(FontSize fs);
-static void LoadFontHelper(FontSize fs, const std::string &file_name, uint size);
-static void TryLoadDefaultTrueTypeFont([[maybe_unused]] FontSize fs, uint size);
```
PeterN: Don't add it.

src/fontcache.cpp
```diff
@@ -115,17 +114,14 @@ void DebugPrintFontSettings(const std::string& desc)
...
 extern void UninitFreeType();
 #elif defined(_WIN32)
 extern void LoadWin32Font(FontSize fs);
-extern void LoadWin32Font(FontSize fs, const std::string &file_name, uint size);
 #elif defined(WITH_COCOA)
 extern void LoadCoreTextFont(FontSize fs);
-extern void LoadCoreTextFont(FontSize fs, const std::string &file_name, uint size);
 #endif
 
-static void LoadFontHelper(FontSize fs)
```
PeterN: Don't add it.

src/fontcache.cpp
```diff
@@ -136,17 +132,6 @@ static void LoadFontHelper(FontSize fs)
...
-static void LoadFontHelper(FontSize fs, const std::string &file_name, uint size)
-{
-#ifdef WITH_FREETYPE
-		LoadFreeTypeFont(fs, file_name, size);
-#elif defined(_WIN32)
-		LoadWin32Font(fs, file_name, size);
-#elif defined(WITH_COCOA)
-		LoadCoreTextFont(fs, file_name, size);
-#endif
-}
```
PeterN: Don't add it.

src/fontcache.cpp
```diff
@@ -213,21 +198,6 @@ static std::string GetDefaultTruetypeFont(FontSize fs)
...
-	if (!full_font.empty()) {
-		if (size == 0) {
-			size = FontCache::GetDefaultFontHeight(fs);
-		}
-
-		LoadFontHelper(fs, full_font, size);
-	}
-#endif /* defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA) */
-}
-
```
PeterN: Don't add it.

src/fontcache.cpp
```diff
@@ -243,6 +213,25 @@ static std::string GetDefaultTruetypeFontFile([[maybe_unused]] FontSize fs)
...
+			setting->size = FontCache::GetDefaultFontHeight(fs);
+		}
+
+		setting->font = full_font;
+		LoadFontHelper(fs);
+		setting->font = "";
+	}
+#endif /* defined(WITH_FREETYPE) || defined(_WIN32) || defined(WITH_COCOA) */
+}
+
```
PeterN: No need to add this, Load*Font() already handles the empty configuration case.

src/settings_gui.cpp
```diff
@@ -175,6 +175,32 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+static const int _max_small_font_size = 20;
+
+static const int _min_medium_font_size = 8;
+static const int _max_medium_font_size = 22;
+
+static const int _min_large_font_size = 18;
+static const int _max_large_font_size = 32;
+
+static const int _min_monospaced_font_size = 8;
+static const int _max_monospaced_font_size = 22;
```
PeterN: Using something like

```
FontCache::GetDefaultFontHeight(fs) * smallest_factor
FontCache::GetDefaultFontHeight(fs) * largest_factor
```

would be better than these arbitrary hard coded values.

src/widgets/settings_widget.h
```diff
@@ -12,21 +12,21 @@
...
+	WID_GO_TAB_SELECTION,                    ///< Background of the tab selection.
+	WID_GO_CURRENCY_DROPDOWN,                ///< Currency dropdown.
+	WID_GO_DISTANCE_DROPDOWN,                ///< Measuring unit dropdown.
+	WID_GO_AUTOSAVE_DROPDOWN,                ///< Dropdown to say how often to autosave.
+	WID_GO_LANG_DROPDOWN,                    ///< Language dropdown.
+	WID_GO_RESOLUTION_DROPDOWN,              ///< Dropdown for the resolution.
+	WID_GO_FULLSCREEN_BUTTON,                ///< Toggle fullscreen.
+	WID_GO_GUI_SCALE,                        ///< GUI Scale slider.
+	WID_GO_GUI_SCALE_AUTO,                   ///< Autodetect GUI scale button.
+	WID_GO_GUI_SCALE_BEVEL_BUTTON,           ///< Toggle for chunky bevels.
```
PeterN: No need for this.

src/widgets/settings_widget.h
```diff
@@ -39,54 +39,54 @@ enum GameOptionsWidgets : WidgetID {
...
+	WID_GO_REFRESH_RATE_DROPDOWN,            ///< Dropdown for all available refresh rates.
+	WID_GO_VIDEO_DRIVER_INFO,                ///< Label showing details about the current video driver.
+	WID_GO_SURVEY_SEL,                       ///< Selection to hide survey if no JSON library is compiled in.
+	WID_GO_SURVEY_PARTICIPATE_BUTTON,        ///< Toggle for participating in the automated survey.
+	WID_GO_SURVEY_LINK_BUTTON,               ///< Button to open browser to go to the survey website.
+	WID_GO_SURVEY_PREVIEW_BUTTON,            ///< Button to open a preview window with the survey results
+	WID_GO_SOCIAL_PLUGINS,                   ///< Main widget handling the social plugins.
+	WID_GO_SOCIAL_PLUGIN_TITLE,              ///< Title of the frame of the social plugin.
+	WID_GO_SOCIAL_PLUGIN_PLATFORM,           ///< Platform of the social plugin.
+	WID_GO_SOCIAL_PLUGIN_STATE,              ///< State of the social plugin.
```
PeterN: No need for this.

src/widgets/settings_widget.h
```diff
@@ -39,54 +39,54 @@ enum GameOptionsWidgets : WidgetID {
...
+	WID_GS_OPTIONSPANEL,      ///< Panel widget containing the option lists.
+	WID_GS_SCROLLBAR,         ///< Scrollbar.
+	WID_GS_HELP_TEXT,         ///< Information area to display help text of the selected option.
+	WID_GS_EXPAND_ALL,        ///< Expand all button.
+	WID_GS_COLLAPSE_ALL,      ///< Collapse all button.
+	WID_GS_RESET_ALL,         ///< Reset all button.
+	WID_GS_RESTRICT_CATEGORY, ///< Label upfront to the category drop-down box to restrict the list of settings to show
+	WID_GS_RESTRICT_TYPE,     ///< Label upfront to the type drop-down box to restrict the list of settings to show
+	WID_GS_RESTRICT_DROPDOWN, ///< The drop down box to restrict the list of settings
+	WID_GS_TYPE_DROPDOWN,     ///< The drop down box to choose client/game/company/all settings
```
PeterN: No need for this.

PeterN (commented):
Probably enough comments for now

src/fontcache.cpp
ladysadie: I see now that you changed the underlying functions in that PR you did.

src/fontcache/freetypefontcache.cpp
ladysadie: I think I just was looking at the wrong file when trying to fix an earlier compiler warning. I can't get a warning from this line in any of the builds I tried. (Desktop.release or Dedicated).

I'll drop it.

src/fontcache.cpp
ladysadie: I fixed the style issues.

I used this function a lot during the development of the original font chooser. I'm still using it while I work out the fontconfig code.  I imagine anyone else working on this code might also find it useful.

src/settings_gui.cpp
ladysadie: Replace one set of hard coded numbers with a different set of hard coded numbers?

In effect it would be a trade off  between tying these numbers to the list in fontcache against the extra code for that. Is that a trade off you think is worth making?

Would each font have it's own factor scales here?  Medium and mono sort of share sizing.

src/fontcache.h
ladysadie: I alphabetized them. Is there a consistent scheme for ordering things like this? I don't see anything mentioned in the style guide.

src/fileio.cpp
ladysadie: I think this change made more sense when this PR used a file chooser.  I'll split it out.

What's the new preferred comment alignment scheme?

src/fileio_type.h
ladysadie: You don't need to repeat a comment like this. It all comes from the same commit and I can extrapolate. :)

src/widgets/settings_widget.h
ladysadie: What should the style be? The file itself uses an inconsistent spacing. I also didn't see anything in the coding style doc  about it.

src/strings.cpp
ladysadie: That's correct. Is there a reason for the monospaced font to be special cased here?

src/fontcache.cpp
ladysadie: Fixed.

src/fontcache.cpp
ladysadie: Fixed.

src/fontcache.cpp
ladysadie: Done.

src/fileio.cpp
2TallTyler: The new alignment is generally "don't try to align anything," instead of trying to be consistent and making PRs bigger when things need to be realigned, as in this case. 🙂

In other words:
- If the section is already consistent, follow the pattern if you can without changing existing lines
- If the section isn't consistent, only add a single space and don't try to align anything
- If you're making a new block, only add a single space

ladysadie: Thanks.

src/fontcache.h
```diff
@@ -225,10 +227,12 @@ inline FontCacheSubSetting *GetFontCacheSubSetting(FontSize fs)
 }
 
 std::string GetFontCacheFontName(FontSize fs);
-void InitFontCache(bool monospace);
-void UninitFontCache();
 
+void DebugPrintFontSettings(const std::string& desc);
```
LordAro:
```suggestion
void DebugPrintFontSettings(const std::string &desc);
```

src/fontcache.h
```diff
@@ -65,6 +65,8 @@ class FontCache {
 	 */
 	virtual int GetFontSize() const { return this->height; }
 
+	virtual void SetFontSize([[maybe_unused]]int pixels);
```
LordAro:
```suggestion
	virtual void SetFontSize([[maybe_unused]] int pixels);
```

a few other places too

src/settings_gui.cpp
```diff
@@ -175,6 +175,32 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+
+static const int FONT_SLIDER_MARK_COUNT = 15; // Labeled at every even number. 8 Labeled marks and 7 unlabeled ones.
+
+static StringID FontSizeMarkFunction([[maybe_unused]] int mark_count, int mark, int value)
+{
+	// Skip the odd marks :
+	if ((mark & 1) == 1) {
```
LordAro, line 194:
```suggestion
	if (mark % 2 == 1) {
```

No need to be fancy, can probably remove the comment as a result

src/settings_gui.cpp
```diff
@@ -175,6 +175,32 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+
+static const int _min_large_font_size = 16;
+static const int _max_large_font_size = 30;
+
+static const int _min_monospaced_font_size = 8;
+static const int _max_monospaced_font_size = 22;
+
+static const int FONT_SLIDER_MARK_COUNT = 15; // Labeled at every even number. 8 Labeled marks and 7 unlabeled ones.
+
+static StringID FontSizeMarkFunction([[maybe_unused]] int mark_count, int mark, int value)
```
LordAro, line 192: function could do with some doxygen documentation

src/settings_gui.cpp
```diff
@@ -890,14 +997,27 @@ struct GameOptionsWindow : Window {
...
 	void OnMouseLoop() override
 	{
-		if (_left_button_down || this->gui_scale == _gui_scale) return;
+		if (_left_button_down) {
+			return;
+		}
 
-		_gui_scale_cfg = this->gui_scale;
+		if (this->gui_scale != _gui_scale)
+		{
```
LordAro:
```suggestion
		if (this->gui_scale != _gui_scale) {
```

src/settings_gui.cpp
```diff
@@ -890,14 +997,27 @@ struct GameOptionsWindow : Window {
...
+			}
+		} else if (static_cast<uint>(this->small_font_size) != _fcsettings.small.size) {
+			ResizeFont(FS_SMALL, static_cast<uint>(this->small_font_size));
+		} else if (static_cast<uint>(this->medium_font_size) != _fcsettings.medium.size) {
+			ResizeFont(FS_NORMAL, static_cast<uint>(this->medium_font_size));
+		} else if (static_cast<uint>(this->large_font_size) != _fcsettings.large.size) {
+			ResizeFont(FS_LARGE, static_cast<uint>(this->large_font_size));
+		} else if (static_cast<uint>(this->monospaced_font_size) != _fcsettings.mono.size) {
+			ResizeFont(FS_MONO, static_cast<uint>(this->monospaced_font_size));
 		}
```
LordAro: I suspect it's not actually possible to trigger more than one of these conditions at once, but maybe it would work better conceptually if it wasn't just a massive if/else if chain?

src/settings_gui.cpp
```diff
@@ -1124,6 +1236,66 @@ static constexpr NWidgetPart _nested_game_options_widgets[] = {
 					EndContainer(),
 				EndContainer(),
 			EndContainer(),
+			/* End graphics tab */
+
+			/* Fonts tab*/
```
LordAro:
```suggestion
			/* Fonts tab */
```

LordAro (commented):
just some thoughts on scrolling through the code. The stuff with the constants still needs resolving too

src/fontcache.h
ladysadie: I found 1 other place to fix.

src/settings_gui.cpp
ladysadie: It shouldn't be possible.

I can simplify the blocks here by just making them a bunch of unrelated if's since that's effectively how they operate.  Is that better?

ladysadie:
> just some thoughts on scrolling through the code. The stuff with the constants still needs resolving too

Okay. I thought there was a some that determined the minimum height  font a font could be? I wasn't able to find it with a few grep searches. I'll keep looking.  I could use that information and then the sizes could be dynamic. But I need to find that code...

src/settings_gui.cpp
ladysadie: I'm reverting this change back to the original because of the code warning.

src/settings_gui.cpp
ladysadie: The constants are no more.

src/fontcache.cpp
```diff
@@ -98,9 +98,21 @@ bool GetFontAAState()
 	return _fcsettings.global_aa;
 }
 
-void SetFont(FontSize fontsize, const std::string &font, uint size)
+void DebugPrintFontSettings(const std::string &desc)
```
2TallTyler, line 119: Missing doxygen.

src/fontcache.cpp
```diff
@@ -68,7 +70,6 @@ std::string FontCache::GetName(FontSize fs)
 	}
 }
 
-
```
2TallTyler: Leave this newline alone, it didn't hurt you. 😉

src/fontcache.cpp
```diff
@@ -79,7 +93,6 @@ int GetCharacterHeight(FontSize size)
 	return FontCache::Get(size)->GetHeight();
 }
 
-
```
2TallTyler: This one too.

src/fontcache.cpp
```diff
@@ -23,11 +23,20 @@
 #include "safeguards.h"
 
 /** Default heights for the different sizes of fonts. */
-static const int _default_font_height[FS_END]   = {10, 6, 18, 10};
+static const int _default_font_height[FS_END]   = {10, 6, 16, 10};
```
2TallTyler: Why change the default font size? Seems out of scope for this PR.

src/fontcache.cpp
```diff
@@ -23,11 +23,20 @@
...
 static const int _default_font_ascender[FS_END] = { 8, 5, 15,  8};
 
+const int _default_font_sane_max_height_limit = 40;
```
2TallTyler: Wrong formatting for a constant (you've formatted it as a global variable that can change), and could be less wordy.

```suggestion
const int DEFAULT_FONT_MAX_HEIGHT = 40; ///< The maximum height of the default font.
```

src/fontcache.cpp
```diff
@@ -23,11 +23,20 @@
...
+const int _default_font_min_max_gap = 14;
```
2TallTyler: As above, wrong formatting and needs doxygen. Maybe a rename too, unless the doxygen clears up the conflict of `min_max`. 😛

src/fontcache.cpp
```diff
@@ -23,11 +23,20 @@
...
+// Because of the math that builds the font size sliders _default_font_min_max_gap needs to be an even number:
```
2TallTyler: We use `/* Regular comment. */` and `a = b; // End of line comment`.

```suggestion
/* Because of the math that builds the font size sliders _default_font_min_max_gap needs to be an even number. */
```

src/fontcache.cpp
```diff
@@ -23,11 +23,20 @@
...
+static_assert(_default_font_min_max_gap % 2 == 0);
+
+
+
```
2TallTyler: Too many newlines.

src/fontcache.cpp
```diff
@@ -53,6 +62,12 @@ int FontCache::GetDefaultFontHeight(FontSize fs)
 	return _default_font_height[fs];
 }
 
+void FontCache::SetFontSize([[maybe_unused]] int pixels)
+{
+	Debug(fontcache, 0, "Font {} isn't resizable.", this->GetFontName());
+	return;
```
2TallTyler: No need to write `return`, the function just ends. 😉

src/fontcache.cpp
```diff
@@ -98,9 +111,61 @@ bool GetFontAAState()
...
+#ifdef WITH_FREETYPE
+	LoadFreeTypeFont(fs);
+#elif defined(_WIN32)
+	LoadWin32Font(fs);
+#elif defined(WITH_COCOA)
+	LoadCoreTextFont(fs);
+#endif
+}
+
+void ResizeFont(FontSize font_size, uint size)
```
2TallTyler, line 156: Needs doxygen.

src/fontcache.cpp
```diff
@@ -98,9 +111,61 @@ bool GetFontAAState()
...
 {
-	FontCacheSubSetting *setting = GetFontCacheSubSetting(fontsize);
+	FontCacheSubSetting *setting = GetFontCacheSubSetting(font_size);
+
+	if (setting->size == size) {
+		return;
+	}
```
2TallTyler:
```suggestion
	if (setting->size == size) return;
```

src/fontcache.cpp
```diff
@@ -98,9 +111,61 @@ bool GetFontAAState()
...
+	loaded_font->SetFontSize(size);
+
+	LoadStringWidthTable();
+	UpdateAllVirtCoords();
+	ReInitAllWindows(true);
+
+	if (_save_config) SaveToConfig();
+}
+
+void SetFont(FontSize font_size, const std::string &font, uint size)
```
2TallTyler, line 180: Needs doxygen, while you're here.

src/fontcache.cpp
```diff
@@ -162,16 +206,15 @@ static bool IsDefaultFont(const FontCacheSubSetting &setting)
 uint GetFontCacheFontSize(FontSize fs)
 {
 	const FontCacheSubSetting &setting = *GetFontCacheSubSetting(fs);
-	return IsDefaultFont(setting) ? FontCache::GetDefaultFontHeight(fs) : setting.size;
+	return (IsDefaultFont(setting) && setting.size == 0) ? FontCache::GetDefaultFontHeight(fs) : setting.size;
```
2TallTyler: I would split this ternary.

```suggestion
	if (IsDefaultFont(setting) && setting.size == 0)  return FontCache::GetDefaultFontHeight(fs);
	return setting.size;
```

src/fontcache.h
```diff
@@ -206,6 +208,8 @@ struct FontCacheSettings {
 	bool global_aa;             ///< Whether to anti alias all font sizes.
 };
 
+extern const int _default_font_min_max_gap;
```
2TallTyler: These should just be defined here in the header instead of linking them as externals.

src/fontcache/freetypefontcache.cpp
```diff
@@ -239,7 +241,9 @@ const Sprite *FreeTypeFontCache::InternalGetGlyph(GlyphID key, bool aa)
 	uint height = std::max(1U, (uint)slot->bitmap.rows  + shadow);
 
 	/* Limit glyph size to prevent overflows later on. */
-	if (width > MAX_GLYPH_DIM || height > MAX_GLYPH_DIM) UserError("Font glyph is too large");
+	if (width > MAX_GLYPH_DIM || height > MAX_GLYPH_DIM) {
+		UserError("Font glyph is too large");
+	}
```
2TallTyler: Unnecessary change, can be removed to keep the diff small(er).

src/lang/english.txt
```diff
@@ -1053,8 +1055,21 @@ STR_GAME_OPTIONS_GUI_SCALE_AUTO_TOOLTIP                         :{BLACK}Check th
 STR_GAME_OPTIONS_GUI_SCALE_BEVELS                               :{BLACK}Scale bevels
 STR_GAME_OPTIONS_GUI_SCALE_BEVELS_TOOLTIP                       :{BLACK}Check this box to scale bevels by interface size
 
+STR_GAME_OPTIONS_SMALL_FONT_FRAME                               :{BLACK}Small font
+STR_GAME_OPTIONS_SMALL_FONT_TOOLTIP                             :{BLACK}Used in map legends and on graphs.
+STR_GAME_OPTIONS_SMALL_FONT_SIZE_SLIDER_TOOLTIP                 :{BLACK}Drag to change the small size
```
```suggestion
STR_GAME_OPTIONS_SMALL_FONT_SIZE_SLIDER_TOOLTIP                 :{BLACK}Drag to change the small font size
```

src/settings_gui.cpp
```diff
@@ -175,6 +175,62 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+ * Calculates the starting mark value for a font size slider. Taking into account the allowed minimum and the need to have
+ * FONT_SLIDER_MARK_COUNT marks. (The max allowed font size value is currently the same for all fonts)
+ *
+ * @param current_size The current size of the font in question.
+ * @param minimum_size The allowed minimum value for this font size.
+ * @return An int value that serves as the minimum value that will be on a slider.
+ */
+static int CalculateFontSliderMin(int current_size, int minimum_size)
+{
+	// This works fine unless the scale is towards the high end of the allowable range:
```
2TallTyler: Wrong comment format (should be `/* comment*/`).

src/settings_gui.cpp
```diff
@@ -175,6 +175,62 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+	int first_check = std::max(current_size - FONT_HALF_GAP, minimum_size);
```
2TallTyler: This is a bit oddly named. Is there a more descriptive name?

src/settings_gui.cpp
```diff
@@ -175,6 +175,62 @@ static StringID VolumeMarkFunc(int, int mark, int value)
...
+ * Calculates the ending mark value for a font size slider. Taking into account the allowed maximum and the need to have
+ * FONT_SLIDER_MARK_COUNT marks. (The max allowed font size value is currently the same for all fonts)
+ *
+ * @param current_size The current size of the font in question.
+ * @param minimum_size The allowed minimum value for this font size.
+ * @return An int value that serves as the maximum value that will be on a slider.
+ */
+static int CalculateFontSliderMax(int current_size, int minimum_size)
+{
+	// This works fine unless the scale is towards the low end of the allowable range:
```
2TallTyler: More comments with the wrong format. Same note on `first_check` -- maybe something about checking the max height?

src/settings_gui.cpp
```diff
@@ -365,6 +436,24 @@ struct GameOptionsWindow : Window {
 		this->reload = false;
 		this->gui_scale = _gui_scale;
 
+		// Fonts sizes could be zero here. The font loading code will choose a non-zero value but then it scales it making it useless here.
```
2TallTyler: Wrong comment format, and it would be better to explain why this code exists instead of explaining what it doesn't do. The negation is confusing. 🙂

src/settings_gui.cpp
```diff
@@ -175,6 +175,62 @@ static StringID VolumeMarkFunc(int, int mark, int value)
 	return STR_GAME_OPTIONS_VOLUME_MARK;
 }
 
+static const int FONT_HALF_GAP = _default_font_min_max_gap / 2;
+static const int FONT_SLIDER_MARK_COUNT = _default_font_min_max_gap + 1; // Labeled at every other number. 8 Labeled marks and 7 unlabeled ones.
+
+/**
+ * Called to get the mark string for the font size sliders. Labels every other number starting from the min_*_font_size
```
2TallTyler:
```suggestion
 * Callback to get the mark string for the font size sliders. Labels every other number starting from the min_*_font_size
```

src/settings_gui.cpp
```diff
@@ -759,32 +887,34 @@ struct GameOptionsWindow : Window {
...
-#ifdef HAS_TRUETYPE_FONT
-			case WID_GO_GUI_FONT_SPRITE:
+			case WID_GO_FONT_SPRITE:
 				_fcsettings.prefer_sprite = !_fcsettings.prefer_sprite;
 
-				this->SetWidgetLoweredState(WID_GO_GUI_FONT_SPRITE, _fcsettings.prefer_sprite);
-				this->SetWidgetDisabledState(WID_GO_GUI_FONT_AA, _fcsettings.prefer_sprite);
+				this->SetWidgetLoweredState(WID_GO_FONT_SPRITE, _fcsettings.prefer_sprite);
+				this->SetWidgetDisabledState(WID_GO_FONT_AA, _fcsettings.prefer_sprite);
+				this->SetWidgetDisabledState(WID_GO_FONT_SMALL_FONT_SIZE_SLIDER, _fcsettings.prefer_sprite && GetFontCacheSubSetting(FS_SMALL)->font.empty());
```
2TallTyler: Question: Why does the font need to be empty to disable this? Wouldn't the "Prefer sprite font" be enough?

src/settings_gui.cpp
```diff
@@ -890,14 +1049,32 @@ struct GameOptionsWindow : Window {
 
 	void OnMouseLoop() override
 	{
-		if (_left_button_down || this->gui_scale == _gui_scale) return;
+		if (_left_button_down) {
+			// Nothing will have changed if the mouse button isn't down, so bail out.
+			return;
+		}
```
2TallTyler:
```suggestion
		if (_left_button_down) return; // Nothing will have changed if the mouse button isn't down, so bail out.
```

src/settings_gui.cpp
```diff
@@ -1124,6 +1293,66 @@ static constexpr NWidgetPart _nested_game_options_widgets[] = {
...
+				EndContainer(),
+
+				/* Small font: */
+				NWidget(WWT_FRAME, COLOUR_GREY), SetDataTip(STR_GAME_OPTIONS_SMALL_FONT_FRAME, STR_NULL), SetPIP(0, WidgetDimensions::unscaled.vsep_wide, 0),
+					NWidget(NWID_HORIZONTAL), SetPIP(0, WidgetDimensions::unscaled.hsep_wide, 0),
+						NWidget(WWT_LABEL, COLOUR_GREY, WID_GO_FONT_SMALL_FONT_LABEL), SetDataTip(STR_JUST_RAW_STRING, STR_GAME_OPTIONS_SMALL_FONT_TOOLTIP), SetFill(1, 0), SetTextStyle(TC_BLACK, FS_SMALL), SetMinimalTextLines(1, 6, FS_SMALL), SetAlignment(SA_CENTER),
+					EndContainer(),
+					NWidget(NWID_HORIZONTAL), SetPIP(0, WidgetDimensions::unscaled.hsep_wide, 0),
+						NWidget(WWT_TEXT, COLOUR_GREY, WID_GO_TEXT_FONT_SIZE), SetMinimalSize(0, 12), SetDataTip(STR_GAME_OPTIONS_FONT_SIZE, STR_NULL),
+						NWidget(WWT_EMPTY, COLOUR_GREY, WID_GO_FONT_SMALL_FONT_SIZE_SLIDER), SetMinimalSize(67, 0), SetMinimalTextLines(1, 12 + WidgetDimensions::unscaled.vsep_normal, FS_NORMAL), SetFill(1, 0), SetResize(1, 0), SetDataTip(0x0, STR_GAME_OPTIONS_SMALL_FONT_SIZE_SLIDER_TOOLTIP),
```
2TallTyler: Some of these lines are pretty long and should be broken up. Line continuations have two tabs more indentation than the line they continue.

src/fontcache.cpp
ladysadie: I'll change it.  2 lines above this is:
"_default_font_height[FS_END]"
which is what I used as an example when I wrote those lines.  Is that line improperly formatted?

src/fontcache.cpp
ladysadie: I guess I don't understand what you mean about the min_max conflict?

I fixed the formatting though.

src/settings_gui.cpp
ladysadie: Earlier I made the mistake of assuming that the  prefer sprites checkbox should override all current font settings.  I think it was Peter that corrected me either in chat or in one of these reviews.

The prefer sprites checkbox should only affect the default OpenTTD fonts but, not any other fonts that the user has manually set. Checking for empty font used to be a good way to check for default fonts.  There's a new function for that I could use instead now. I'll update it to call that function.

src/fontcache.h
```diff
@@ -206,6 +208,11 @@ struct FontCacheSettings {
 	bool global_aa;             ///< Whether to anti alias all font sizes.
 };
 
+const int DEFAULT_FONT_MAX_HEIGHT = 40; ///< The maximum height allowed by font sliders.
+const int DEFAULT_FONT_MIN_MAX_GAP = 14; ///< This controls how many steps will be on any slider. The gap between the low end of the slider and the high end.
```
2TallTyler: Would this be `DEFAULT_FONT_SLIDER_STEPS` perhaps?

src/fontcache.h
```diff
@@ -206,6 +208,11 @@ struct FontCacheSettings {
...
+/** Because of the math that builds the font size sliders DEFAULT_FONT_MIN_MAX_GAP needs to be an even number: */
```
2TallTyler: `/** comment */` is for doxygen, `/* comment */` is for regular comments.

```suggestion
/* Because of the math that builds the font size sliders DEFAULT_FONT_MIN_MAX_GAP needs to be an even number: */
```

src/fontcache.cpp
```diff
@@ -162,16 +221,20 @@ static bool IsDefaultFont(const FontCacheSubSetting &setting)
 uint GetFontCacheFontSize(FontSize fs)
 {
 	const FontCacheSubSetting &setting = *GetFontCacheSubSetting(fs);
-	return IsDefaultFont(setting) ? FontCache::GetDefaultFontHeight(fs) : setting.size;
+
+	if (IsDefaultFont(setting) && setting.size == 0) {
+		return FontCache::GetDefaultFontHeight(fs);
+	} else  {
+		return setting.size;
+	}
```
2TallTyler: An early return means you don't need an `else`.

```suggestion
	if (IsDefaultFont(setting) && setting.size == 0) return FontCache::GetDefaultFontHeight(fs);
	return setting.size;
```

src/fontcache/freetypefontcache.cpp
```diff
@@ -80,6 +81,7 @@ void FreeTypeFontCache::SetFontSize(int pixels)
 	} else {
 		pixels = ScaleGUITrad(pixels);
 	}
+
```
2TallTyler: Random newline snuck in here. 😉

src/settings_gui.cpp
```diff
@@ -365,6 +436,24 @@ struct GameOptionsWindow : Window {
 		this->reload = false;
 		this->gui_scale = _gui_scale;
 
+		/** Retrieve the current fonts settings. This requires some bounds checking as fonts sizes could be zero here. */
```
2TallTyler:
```suggestion
		/* Retrieve the current fonts settings. This requires some bounds checking as fonts sizes could be zero here. */
```

src/settings_gui.cpp
2TallTyler: Aha, thanks for the explanation. 😃

src/settings_gui.cpp
```diff
@@ -890,14 +1049,30 @@ struct GameOptionsWindow : Window {
 
 	void OnMouseLoop() override
 	{
-		if (_left_button_down || this->gui_scale == _gui_scale) return;
+		/** Nothing will have changed if the mouse button isn't down, so bail out. */
+		if (_left_button_down) { return; }
```
2TallTyler:
```suggestion
		if (_left_button_down) return;
```

src/settings_gui.cpp
```diff
@@ -890,14 +1049,30 @@ struct GameOptionsWindow : Window {
 
 	void OnMouseLoop() override
 	{
-		if (_left_button_down || this->gui_scale == _gui_scale) return;
+		/** Nothing will have changed if the mouse button isn't down, so bail out. */
```
2TallTyler:
```suggestion
		/* Nothing will have changed if the mouse button isn't down, so bail out. */
```

src/fontcache.cpp
ladysadie: I'm aware. I did it for readability.  I'll change it.

src/fontcache.h
ladysadie: There's a FONT_SLIDER_MARK_COUNT in the settings_gui.cpp file.  Honestly since this variable is just used to construct that one it's kind of redundant. Removed.

src/fontcache.h
PeterN: They're usually grouped for a reason Init/Uninit being related. Don't randomly alphabetize things.

src/strings.cpp
PeterN: Yes, monospace font searching is only used when loading a text file. None of the language files make use of the monospace font.

Conversely, the normal font sizes aren't relevant when loading a text file.

This is also why the monospace font loading is normally deferred until a text file is loaded.

2TallTyler:
Trying to open the Font tab in the preview deployment crashes the game. It probably should not do that. 😛

src/settings_gui.cpp
```diff
@@ -583,16 +707,22 @@ struct GameOptionsWindow : Window {
...
 			case WID_GO_TAB_GENERAL: pane = 0; break;
 			case WID_GO_TAB_GRAPHICS: pane = 1; break;
+#ifdef HAS_TRUETYPE_FONT
+			case WID_GO_TAB_FONTS: pane = 2; break;
+			case WID_GO_TAB_SOUND: pane = 3; break;
+			case WID_GO_TAB_SOCIAL: pane = 4; break;
+#else
 			case WID_GO_TAB_SOUND: pane = 2; break;
 			case WID_GO_TAB_SOCIAL: pane = 3; break;
+#endif
```
PeterN, line 725: It might be better to include a dummy pane if truetype fonts are disabled, instead of renumbering them.

Or just hide the button and leave the pane alone.

PeterN, line 728: You don't hide any widgets when HAS_TRUETYPE_FONT is not set, so this is wrong.

The simplest approach would be to just hide WID_GO_TAB_FONTS and leave the rest alone.

ladysadie: I messed up.  Anyways I've tracked down all of the uses of the font tab and guarded them as well.  I had to hack a desktop build disabling Freetype font support but I tested it and everything seems fine now.  I'll check the preview build when it's done though.

ladysadie: The tab isn't visible on the preview build and the other tabs still work fine. Seems weird but as far as I know the preview build builds without freetype font support so this is kinda weird but "works as intended".

src/strings.cpp
ladysadie: > Conversely, the normal font sizes aren't relevant when loading a text file.

Some font (looks like the FS_NORMAL font) is used in the title bars of the display windows.  So this isn't true.

My understanding of your point is that the old code defers the monospaced font loading and for some cases that's better.  One less case in the future as the Monospaced font will need to be loaded for the font settings tab.

My counter argument is that the font code isn't in the critical path.  Optimization is nice but unneeded. The font code is already badly optimized with repeated calls to InitFontCache and CheckForMissingGlyphs.  Simplifying this block of code is worth the cost.

ladysadie:
I feel like trying to get this done again.

I see the base code has changed a bit so I'll have to merge my old work with the new code.  I still have my trial fontconfig code. I think my plan will be to implement the slider code for all platforms and then a font picker for the Linux build. I can't do any Mac or Windows development.

PR #12697, opened 2024-05-18
Fix: Social plugins not working on macOS

yeah-its-gloria:
## Motivation / Problem
Applications running on the Hardened Runtime get skillchecked by System Integrity Protection when loading dynamic libraries that are signed with certificates belonging to a different Team ID.
Apple describes this as preventing code injection.

This happens particularly in the case of Steam, whose dynamic library for Steamworks is signed by Valve, not us, therefore triggering the protection mechanism and preventing `dlopen` from functioning.
Steam also relies on environment variables to inject the in game overlay, which is also blocked for the same reason.

This causes something like this to be output to stdout:
```
[2024-05-18 11:40:44] dbg: [driver:0] Could not change to foreground application. Error -50
dlopen failed trying to load: [...]/steamclient.dylib
with error:
dlopen([...]/steamclient.dylib, 0x0002): tried: '[...]/steamclient.dylib' (code signature in [...] '[...]/steamclient.dylib' not valid for use in >process: mapping process and mapped file (non-platform) have different Team IDs), [more of the same but for different paths])

[S_API] SteamAPI_Init(): Failed to load module '[...]/steamclient.dylib'
[2024-05-18 11:40:44] dbg: [misc:0] [Social Integration: steam/libsteam-social.dylib] Failed to initialize
```

This is Bad. However, Steamworks has a solution, which is adding entitlements to the application that allow it to bypass these limits set out by the Hardened Runtime.

## Description
Adds entitlements needed to load arbitrary plugins correctly (Steam particularly) during codesigning.

## Limitations
This could possibly allow bad actors to perform code injection on OpenTTD with malicious plugins. ~~Kinda goes against "social" if you ask me.~~

## Checklist for review

### DIFF
--- a/os/macosx/openttd.entitlements
+++ b/os/macosx/openttd.entitlements
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!-- As requested in https://partner.steamgames.com/doc/store/application/platforms, allows arbitrary plugin loads on hardened runtime -->
+	<key>com.apple.security.cs.allow-dyld-environment-variables</key>
+	<true/>
+	<key>com.apple.security.cs.disable-library-validation</key>
+	<true/>
+</dict>
+</plist>
--- a/cmake/PackageBundle.cmake
+++ b/cmake/PackageBundle.cmake
@@ -4,6 +4,7 @@ set(CPACK_BUNDLE_NAME "OpenTTD")
 set(CPACK_BUNDLE_ICON "${CMAKE_SOURCE_DIR}/os/macosx/openttd.icns")
 set(CPACK_BUNDLE_PLIST "${CMAKE_CURRENT_BINARY_DIR}/Info.plist")
 set(CPACK_DMG_BACKGROUND_IMAGE "${CMAKE_SOURCE_DIR}/os/macosx/splash.png")
+set(CPACK_BUNDLE_APPLE_ENTITLEMENTS "${CMAKE_SOURCE_DIR}/os/macosx/openttd.entitlements")
 set(CPACK_DMG_FORMAT "UDBZ")
 
 # Create a temporary Info.plist.in, where we will fill in the version via


glx22:
Test build [here](https://cdn.openttd.org/openttd-branches/pr12697/20240518-pr12697-g43b881be25/)

os/macosx/openttd.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.cs.allow-dyld-environment-variables</key>
+	<true/>
+	<key>com.apple.security.cs.disable-library-validation</key>
```
TrueBrain, line 8: We already add some entitlements, but in a different place:

https://github.com/OpenTTD/OpenTTD/blob/master/.github/workflows/release-macos.yml#L133

Possibly a simpler solution would be to add `-o library` there. Or at least synchronize the two, so we don't have two places doing somewhat the same work.

os/macosx/openttd.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.cs.allow-dyld-environment-variables</key>
```
TrueBrain, line 6: This lacks information why this entry is here (as for the next), so it is easily forgotten in any refactor. I would suggest to add an XML comment here, referencing the Steamwork documentation for example.

src/video/cocoa/cocoa_wnd.mm
```diff
@@ -261,6 +261,14 @@ - (void)unregisterObserver
...
 	[ [ NSNotificationCenter defaultCenter ] removeObserver:self ];
 }
+
+/**
+ * Indicate to AppKit that we don't interfere with its secure coding.
+ */
+- (BOOL)applicationSupportsSecureRestorableState
+{
+	return YES;
+}
```
TrueBrain: This function (and its comment) are very vague and non-descriptive. The motivation also doesn't mention why it should exist. This means it is not unlikely someone removes it somewhere in the future, as it "seems to have no effect".

Would you mind given a more clear indication why this is needed, and what it is doing? Either in the motivation of the PR, or if possible, in the comment?

When I read the documentation, it seems unrelated to this PR, and has nothing to do with the social plugins? But I might be misunderstanding the very few words Apple's documentation uses to describe this function. Hence my remark here :)

os/macosx/openttd.entitlements
yeah-its-gloria: We need both for proper Steam support, and the DYLD environment flags entitlement isn't supported by options as far as I know.

src/video/cocoa/cocoa_wnd.mm
yeah-its-gloria: Since macOS 14 Sonoma, macOS logs a warning about not explicitly stating whether secure state recording is supported by the app.
By default, it logs a warning and returns true.
This code does the same without the warning, and since it's so small and changes nothing of value (OpenTTD encodes no state of its own to restore from to begin with), it felt more fitting to include here than to make a separate PR.

TrueBrain: Please don't sneak other change in PRs; it makes a nightmare for us when something breaks because of it. When you do a git blame on this line, you will find this PR, with no mention in the description about it what so ever. That would be incredibly hard to figure out why it was added and what it tried to solve.

So yeah, even for these small things, please just create a tiny small PR :) Future-us will thank you.

PS: this is a lesson we learnt the hard way over the last 20 years :D Especially when it comes to MacOS changes ..... :P

glx22:
Including the quote from discord, containing some info that should have been in PR description
>the app fails to boot like this
>```
>[2024-05-18 11:40:44] dbg: [driver:0] Could not change to foreground application. Error -50
>dlopen failed trying to load: [...]/steamclient.dylib
>with error:
>dlopen([...]/steamclient.dylib, 0x0002): tried: '[...]/steamclient.dylib' (code signature in [...] '[...]/steamclient.dylib' not valid for use in >process: mapping process and mapped file (non-platform) have different Team IDs), [more of the same but for different paths])
>
>[S_API] SteamAPI_Init(): Failed to load module '[...]/steamclient.dylib'
>[2024-05-18 11:40:44] dbg: [misc:0] [Social Integration: steam/libsteam-social.dylib] Failed to initialize
>```
>
>https://partner.steamgames.com/doc/store/application/platforms Steamworks makes a note about having to add entitlements, are these present in the Steam version?

TrueBrain (approved):

PR #8789, opened 2021-03-01
Fix: [OSX] Hide dock when entering fullscreen

orudge:
## Motivation / Problem

As per the last comment in #8038, the dock still appears if you hover over the bottom of the screen when in fullscreen mode.

## Description

When switching into fullscreen mode, we hide both the menu and the dock programmatically. They are unhidden again when exiting full screen.

## Limitations

None that I'm aware of.

## Checklist for review

Some things are not automated, and forgotten often. This list is a reminder for the reviewers.
* The bug fix is important enough to be backported? (label: 'backport requested')
* This PR affects the save game format? (label 'savegame upgrade')
* This PR affects the GS/AI API? (label 'needs review: Script API')
    * ai_changelog.hpp, gs_changelog.hpp need updating.
    * The compatibility wrappers (compat_*.nut) need updating.
* This PR affects the NewGRF API? (label 'needs review: NewGRF')
    * newgrf_debug_data.h may need updating.
    * [PR must be added to API tracker](https://wiki.openttd.org/en/Development/NewGRF/Specification%20Status)

PR #12699, opened 2024-05-19
Fix #10044: Cursor being poorly managed on macOS in fullscreen mode

### LINKED ISSUE #10044: [Bug]: System cursor rendered on top of game cursor on macOS
### Version of OpenTTD

12.2, macOS Monterey 12.5

### Expected result

The game should always hide the system's cursor.

### Actual result

Sometimes the system's cursor is rendered over the game's cursor.

### Steps to reproduce

1. Enter Full Screen
2. Move cursor to the bottom of the screen, and move up directly (do not trigger the Dock)
3. The system's cursor is rendered over the game's cursor

yeah-its-gloria:
## Motivation / Problem
As reported in #10044, macOS doesn't reliably switch between the in game and system cursor when going over the Menu Bar or Dock.

## Description
This bug appears to be caused by the fact these are forcefully hidden, despite this being uncommon behavior (I mostly went through my Steam library, testing BAD END THEATER, Factorio and Terraria, all games that don't fully hide the cursor. None of these games hid the Menu Bar or Dock fully, instead letting the OS decide, which relies on user settings).
As such, this changes the settings used for fullscreen mode on macOS to instead automatically hide things if possible, instead of requiring them to be hidden, like how most apps and a lot of games do.

Closes #10044.

## Limitations
PR #8789, where the change was introduced for the most part, forcefully hid the Menu Bar particularly to prevent it rolling over when pressing a button at the top.
Testing this on my own MacBook Air, I failed to replicate that problem, although I have no idea if this might be caused by the fact it reserves a lot of space at the top of the screen.
I do not have an external monitor with me to test this at the minute, and I don't even know if this behavior is different because that author was running macOS Catalina and I'm running Sonoma.
Even despite this mostly fixing the bug, at times I managed to still make the cursor look wrong, perhaps pointing to an issue related to how it's updated, although with everything acting more alike how other games act in fullscreen, this bug happened a lot less frequently.

Frankly, this might be a problem with screen safe zones instead of the menu bar rolling over, perhaps at some point it might be worth adding a setting to add some spacing between the screen edges and the buttons up top.

## Checklist for review

### DIFF
--- a/src/video/cocoa/cocoa_v.mm
+++ b/src/video/cocoa/cocoa_v.mm
@@ -206,9 +206,6 @@ bool VideoDriver_Cocoa::ToggleFullscreen(bool full_screen)
 	if ([ this->window respondsToSelector:@selector(toggleFullScreen:) ]) {
 		[ this->window performSelector:@selector(toggleFullScreen:) withObject:this->window ];
 
-		/* Hide the menu bar and the dock */
-		[ NSMenu setMenuBarVisible:!full_screen ];
-
 		this->UpdateVideoModes();
 		InvalidateWindowClassesData(WC_GAME_OPTIONS, 3);
 		return true;
--- a/src/video/cocoa/cocoa_wnd.mm
+++ b/src/video/cocoa/cocoa_wnd.mm
@@ -1288,7 +1288,7 @@ void CocoaDialog(const char *title, const char *message, const char *buttonLabel
 /** Presentation options to use for full screen mode. */
 - (NSApplicationPresentationOptions)window:(NSWindow *)window willUseFullScreenPresentationOptions:(NSApplicationPresentationOptions)proposedOptions
 {
-	return NSApplicationPresentationFullScreen | NSApplicationPresentationHideMenuBar | NSApplicationPresentationHideDock;
+	return NSApplicationPresentationFullScreen | NSApplicationPresentationAutoHideMenuBar;
 }
 
 @end


glx22:
Proper commit message format is `Fix #10044: ...`

2TallTyler:
@Bouke, you opened the issue that this fixes. Any thoughts?

Bouke:
I tried reproducing the issue on 12.2, 13.1, 14.1 and c5d43b433b9e4b377b7dab39c098e240d8175502. I have set my Dock to automatically hide: image. In neither of the versions I was able to exactly reproduce my bug report: moving the mouse to the bottom of the screen doesn't trigger showing the Dock. However I have found two more ways to trigger the bug in 12.2, 13.1 and 14.1:

1. On a display with notch: move the mouse to where the menu bar would be, next to the notch.
2. Trigger the screenshot utility by pressing ⌘⇧4 and create screenshot/cancel.

This patch fixes 1, but not 2. This could indicate that there's additional ways to trigger the bug.

I have found that this patch also changes behaviour on external displays. On the primary display, the space normally occupied by the menu bar isn't usable by an application. It is either rendered next to the notch, or it is black depending on the setting for "automatically show and hide the menu bar". Because that space is always reserved regardless of containing the menu bar, I have it set to never hide: image. Then when I take an application fullscreen on an external monitor, the menubar is rendered on a black background.

OpenTTD 12.2, 13.1 and 14.1 however don't consider this setting and there's no menubar regardless of this setting (a bug in itself?). On external displays, this means that OpenTTD uses the full height of the display and no space is reserved for the menu bar.  However with this patch, the content of OpenTTD is pushed down; there is space reserved for the menubar but there's nothing shown there.

I think there's two ways to look at this:

- games in 'full screen' should really be 'full screen' and be allowed to override/ignore the above mentioned setting; this is what happens in 12.2, 13.1 and 14.1. On external displays it uses the full display, on the notched primary display it is pushed below the notch.
- 'full screen' should still abide the above mentioned setting and render the menubar.

However with this patch we get neither: there is space reserved for the menu bar (reducing vertical height), but it is not doing anything by being a black space.

c5d43b433b9e4b377b7dab39c098e240d8175502:
Screenshot 2024-06-16 at 09 36 21
14.1:
Screenshot 2024-06-16 at 09 37 52

PR #12702, opened 2024-05-20
Codechange: Use std::filesystem::exists instead of access.

PeterN:
## Motivation / Problem

Use of C-style API along with 'hidden' define for Windows for testing if file exists with `access()` or `_taccess()`.

## Description

Use platform-independent `std::filesystem::exists()` instead.

## Checklist for review

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -16,7 +16,6 @@
 #include "tar_type.h"
 #ifdef _WIN32
 #include <windows.h>
-# define access _taccess
 #elif defined(__HAIKU__)
 #include <Path.h>
 #include <storage/FindDirectory.h>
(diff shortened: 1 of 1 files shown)


src/fileio.cpp
```diff
@@ -135,7 +134,7 @@ bool FioCheckFileExists(const std::string &filename, Subdirectory subdir)
  */
 bool FileExists(const std::string &filename)
 {
-	return access(OTTD2FS(filename).c_str(), 0) == 0;
+	return std::filesystem::exists(OTTD2FS(filename));
```
rubidium42: Do we want an exception being thrown "on underlying OS API errors"? So, essentially crash the game?

Or do we want to get the non-throwing overload and return `false` when things go wrong?

rubidium42 (approved):

PR #12707, opened 2024-05-21
Fix #12391: Fixed console save message timings and altered when they appear.

### LINKED ISSUE #12391: [Bug]: Console message "map successfully saved" appears before save actually completes
### Version of OpenTTD

14.0-RC3, windows 10

### Expected result

«saving map» message remains until the saving is completed
«map successfully saved» appears only after the actual completion of the save

### Actual result

The message «map successfully saved» appears right after the command, while the Zzz icon and « * * SAVING GAME * * » are still visible
изображение

### Steps to reproduce

1. Load a map whose saving time will be quite long (for example 4k² with trees)
2. Open console and enter save <name>, press enter

TrevorShelton:
## Motivation / Problem

Issue #12391

## Description

Each of the saving messages on the console appear at the correct times now. The started saving message appears when starting a save, and the two outcome messages of saving are at the end now. A couple of functions needed to be altered to take the filename for this. Since these are put in the save process as a whole now you can see them in the console even if not saving via console command.

## Limitations

Save filters don't appear to use a filename, so they use default which is set to have the name string empty. 

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -21,7 +21,6 @@
  */
 
 #include "../stdafx.h"
-#include "../console_func.h"
 #include "../debug.h"
 #include "../station_base.h"
 #include "../thread.h"
@@ -2832,7 +2831,7 @@ static SaveOrLoadResult SaveFileToDisk(bool threaded, const std::string &filenam
 
 		if (threaded) SetAsyncSaveFinish(SaveFileDone);
 
-		IConsolePrint(CC_INFO, "Saved successfully as '{}'.", filename);
+		Debug(sl, 1, "Map saved as '{}'.", filename);
 
 		return SL_OK;
 	} catch (...) {
@@ -2854,7 +2853,6 @@ static SaveOrLoadResult SaveFileToDisk(bool threaded, const std::string &filenam
 			asfp();
 		}
 
-		IConsolePrint(CC_ERROR, "Saving map failed.");
 		return SL_ERROR;
 	}
 }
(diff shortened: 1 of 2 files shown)


src/saveload/saveload.cpp
```diff
@@ -20,6 +20,7 @@
  * </ol>
  */
 
+#include "../console_func.h"
 #include "../stdafx.h"
```
rubidium42: This `stdafx.h` include must always be the first include.

rubidium42 (commented):
I wonder how useful this specific change is. I would reckon that the command is mostly for use via `rcon` or the admin connection. Yes, you can use it in single player from the console as well, but I'd reckon then you'll usually just use the UI.

In both `rcon` and admin connection cases, *only* text that is printed while the command is executed will be returned to the caller. So the "Map successfully saved" message will not be returned to the caller. Even worse, it might be sent to a completely different caller.

Finally, with this solution you need to guard `IConsolePrint` and friends for data race conditions due to that function being called twice at the same time. Once from the thread performing the save and once from the 'main' thread running some other command.

I think the amount of work to get all of that correct vastly outweighs just changing "Map successfully saved to '{}'." to "Saving map to '{}'." and removing the "Saving map..." string as that's not really useful in any case.

You could consider adding `Debug(sl, 1, ...` messages where you added `IConsolePrint` messages, so at least with the admin connection and for dedicated servers you get the chance to enable getting that feedback. With `rcon` I'm not seeing a feasible way to get the feedback there, and for the situation you are running from the UI you already get the error message or a sign in the bottom toolbar.

src/saveload/saveload.cpp
TrevorShelton: Is that why Linux GCC-Dedicated was complaining?

src/saveload/saveload.cpp
```diff
@@ -2814,7 +2814,7 @@ static void SaveFileError()
  * We have written the whole game into memory, _memory_savegame, now find
  * and appropriate compressor and start writing to file.
  */
-static SaveOrLoadResult SaveFileToDisk(bool threaded)
+static SaveOrLoadResult SaveFileToDisk(bool threaded, const std::string &filename)
```
LordAro, line 2817: It feels very confusing to me that filename is only used to display the debug message, it's not actually "used" for generating the actual file. Not sure if there's anything that can be done about that without a much more significant refactor though. Especially when there's a default parameter on `DoSave` that's just empty.

Also, missing doxygen

PR #12710, opened 2024-05-22
Fix d6ccfdb: GetCharPosition failed for RTL text.

PeterN:
## Motivation / Problem

Changes in d6ccfdb assumed that character indexes only incremented, which is not true for RTL text. An early exit if the current index is higher than the index to find meant that for RTL runs, the position was never found.

This makes editing RTL text 'impossible' — it is simpler to clear and start again.

## Description

Make GetCharPosition() find the glyph position by searching for the nearest match, without an early exit.

## Limitations

This does not completely fix RTL editing, that requires more changes to come.

## Checklist for review

### DIFF
--- a/src/gfx_layout.cpp
+++ b/src/gfx_layout.cpp
@@ -11,6 +11,7 @@
 #include "core/math_func.hpp"
 #include "gfx_layout.h"
 #include "string_func.h"
+#include "strings_func.h"
 #include "debug.h"
 
 #include "table/control_codes.h"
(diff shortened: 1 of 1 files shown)


src/gfx_layout.cpp
```diff
@@ -245,33 +246,32 @@ Point Layouter::GetCharPosition(std::string_view::const_iterator ch) const
...
-			if (it == charmap.end()) break;
-
-			/* We just passed our character, it's probably a ligature, use the last found position. */
-			if (static_cast<size_t>(*it) > index) return *position;
-			++position;
+			/* No exact match, but maybe this is the best match. */
+			if (cur_index < index && (best_index < cur_index || best_index == SIZE_MAX)) {
+				best_index = cur_index;
+				position = &*itp;
+			}
```
rubidium42, line 277: It's not immediately clear to me what is considered the best match. Is it the position just before the selected ligature/character from the original string?

PeterN: Added a code comment to hopefully explain this better

rubidium42 (approved):

PR #12711, opened 2024-05-22
Fix: Reverse left/right keypress when editing RTL text.

PeterN:
## Motivation / Problem

When editing RTL text, the left cursor key moves right, and the right cursor key moves left.

Pressing left moves to the previous logical character position, which for RTL text is character to the right of the current character. Therefore, pressing left should increment instead of decrement the character position to move left, and vice versa.

## Description

Swap direction when pressing left or right cursor keys when editing RTL text, so that pressing left moves left.

## Limitations

When editing mixed RTL/LTR text, the cursor direction will change direction depending on where the cursor is. This is not intuitive, but it is how other software (Google Chrome!) behaves, so it's not exactly unacceptable.

I believe changing this to always move consistently requires a different approach to cursor movement which would not easily be backportable.

## Checklist for review

### DIFF
--- a/src/textbuf.cpp
+++ b/src/textbuf.cpp
@@ -323,6 +323,40 @@ void Textbuf::UpdateMarkedText()
 }
 
 /**
+ * Move to previous character position.
+ * @param what Move ITER_CHARACTER or ITER_WORD.
+ * @return true iff able to move.
+ */
+bool Textbuf::MovePrev(StringIterator::IterType what)
+{
+	if (this->caretpos == 0) return false;
+
+	size_t pos = this->char_iter->Prev(what);
+	if (pos == StringIterator::END) return true;
+
+	this->caretpos = static_cast<uint16_t>(pos);
+	this->UpdateCaretPosition();
+	return true;
+}
+
+/**
+ * Move to next character position.
+ * @param what Move ITER_CHARACTER or ITER_WORD.
+ * @return true iff able to move.
+ */
+bool Textbuf::MoveNext(StringIterator::IterType what)
+{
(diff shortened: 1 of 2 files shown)


src/textbuf.cpp
```diff
@@ -322,6 +322,40 @@ void Textbuf::UpdateMarkedText()
...
+ * @return true iff able to move.
+ */
+bool Textbuf::MovePrev(StringIterator::IterType what)
+{
+	if (this->caretpos == 0) return false;
+
+	size_t pos = this->char_iter->Prev(what);
+	if (pos == StringIterator::END) return true;
+
+	this->caretpos = (uint16_t)pos;
```
rubidium42:
```suggestion
	this->caretpos = static_cast<uint16_t>(pos);
```

src/textbuf.cpp
```diff
@@ -322,6 +322,40 @@ void Textbuf::UpdateMarkedText()
...
+ * @return true iff able to move.
+ */
+bool Textbuf::MoveNext(StringIterator::IterType what)
+{
+	if (this->caretpos >= this->bytes - 1) return false;
+
+	size_t pos = this->char_iter->Next(what);
+	if (pos == StringIterator::END) return true;
+
+	this->caretpos = (uint16_t)pos;
```
rubidium42:
```suggestion
	this->caretpos = static_cast<uint16_t>(pos);
```

PeterN:
Reworked it slightly as `(keycode & WKC_CTRL) ? ...` broke our codestyle.

rubidium42 (approved):

PR #6848, opened 2018-07-04
Feature : New Trees Placement Algorithm "Forest".

SylvainDevidal:
It creates new trees only in existing forest, preserving meadows.

PR #12712, opened 2024-05-23
Feature: "improved" trees now only spread in forests

Moth-Tolias:
## Motivation / Problem
when tree spreading is on trees rapidly swarm every tile on the map, making them little more than noise that gets tuned out.

## Description
resuscitation of #6848.

## Limitations
i had a nightmare of a time bringing this patch back. i think i've done it right, but if there _are_ any residual glaring issues they are undoubtedly my own.

## Checklist for review

### DIFF
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -99,11 +99,14 @@ static bool IsNearbyForest(TileIndex tile)
 	/* Count the trees around the clear tile to determine if it's near a forest */
 	for (TileIndex t : TileArea(tile).Expand(FOREST_SEARCH_RADIUS)) {
 		if (IsTileType(t, MP_TREES)) {
-			planted_tile_count++;
+			if (++planted_tile_count >= FOREST_THRESHOLD)
+			{
+				return true;
+			}
 		}
 	}
 
-	return (planted_tile_count >= FOREST_THRESHOLD);
+	return false;
 }
 
 /**
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -41,6 +41,7 @@ enum TreePlacer {
 	TP_NONE,     ///< No tree placer algorithm
 	TP_ORIGINAL, ///< The original algorithm
 	TP_IMPROVED, ///< A 'improved' algorithm
+	TP_FOREST,   ///< An algorithm trying to make only forests
 };
 
 /** Where to place trees while in-game? */
@@ -78,6 +79,30 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
 
 		default: return false;
 	}
+}
+
+/**
+* Tests if a tile is near an already existing forest.
+* Trees won't spread if not placed near some other trees.
+*
+* @param tile the tile of interest
+* @return true if the tile is near a forest
+*/
+static bool IsNearbyForest(TileIndex tile)
+{
+	uint planted_tile_count = 0;
+
+	for (int x = -2; x <= 2; x++) {
+		for (int y = -2; y <= 2; y++) {
+			TileIndex vincity = TileAddWrap(tile, x, y);
+			if (vincity != INVALID_TILE &&
+					IsTileType(vincity, MP_TREES)) {
+				planted_tile_count++;
+			}
+		}
+	}
+
+	return (planted_tile_count >= 6);
 }
 
 /**
@@ -257,7 +282,10 @@ void PlaceTreesRandomly()
 
 		if (CanPlantTreesOnTile(tile, true)) {
 			PlaceTree(tile, r);
-			if (_settings_game.game_creation.tree_placer != TP_IMPROVED) continue;
+			if (_settings_game.game_creation.tree_placer != TP_IMPROVED &&
+					_settings_game.game_creation.tree_placer != TP_FOREST) {
+				continue;
+			}
 
 			/* Place a number of trees based on the tile height.
 			 *  This gives a cool effect of multiple trees close together.
@@ -358,6 +386,7 @@ void GenerateTrees()
 
 	switch (_settings_game.game_creation.tree_placer) {
 		case TP_ORIGINAL: i = _settings_game.game_creation.landscape == LT_ARCTIC ? 15 : 6; break;
+		case TP_FOREST: /* FALL THROUGH */
 		case TP_IMPROVED: i = _settings_game.game_creation.landscape == LT_ARCTIC ?  4 : 2; break;
 		default: NOT_REACHED();
 	}
@@ -370,6 +399,8 @@ void GenerateTrees()
 	SetGeneratingWorldProgress(GWP_TREE, total);
 
 	if (num_groups != 0) PlaceTreeGroups(num_groups);
+
+	if (_settings_game.game_creation.tree_placer == TP_FOREST) return;
 
 	for (; i != 0; i--) {
 		PlaceTreesRandomly();
@@ -758,6 +789,9 @@ static void TileLoop_Trees(TileIndex tile)
 						/* Don't plant trees, if ground was freshly cleared */
 						if (IsTileType(tile, MP_CLEAR) && GetClearGround(tile) == CLEAR_GRASS && GetClearDensity(tile) != 3) return;
 
+						/* Plants trees only near existing forests */
+						if (_settings_game.game_creation.tree_placer == TP_FOREST && !IsNearbyForest(tile)) return;
+
 						PlantTreesOnTile(tile, treetype, 0, 0);
 
 						break;
@@ -859,6 +893,9 @@ void OnTick_Trees()
 	r = Random();
 	tile = RandomTileSeed(r);
 	if (CanPlantTreesOnTile(tile, false) && (tree = GetRandomTreeType(tile, GB(r, 24, 8))) != TREE_INVALID) {
+		/* Plants trees only near existing forests */
+		if (_settings_game.game_creation.tree_placer == TP_FOREST && !IsNearbyForest(tile)) return;
+
 		PlantTreesOnTile(tile, tree, 0, 0);
 	}
 }
--- a/src/tree_cmd.cpp
+++ b/src/tree_cmd.cpp
@@ -91,6 +91,10 @@ static bool IsNearbyForest(TileIndex tile)
 {
 	uint planted_tile_count = 0;
 
+	// An already planted tilte can always be planted again
+	if (IsTileType(tile, MP_TREES)) return true;
+
+	// Count the trees arround the clear tile to determine if it's near a forest
 	for (int x = -2; x <= 2; x++) {
 		for (int y = -2; y <= 2; y++) {
 			TileIndex vincity = TileAddWrap(tile, x, y);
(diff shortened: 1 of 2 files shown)


src/tree_cmd.cpp
```diff
@@ -80,6 +81,34 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
...
+* Trees won't spread if not placed near some other trees.
+*
+* @param tile the tile of interest
+* @return true if the tile is near a forest
+*/
+static bool IsNearbyForest(TileIndex tile)
+{
+	uint planted_tile_count = 0;
+
+	// An already planted tilte can always be planted again
```
```suggestion
	/* An already planted tile can always be planted again. */
```

src/tree_cmd.cpp
```diff
@@ -80,6 +81,34 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
...
+	if (IsTileType(tile, MP_TREES)) return true;
+
+	// Count the trees arround the clear tile to determine if it's near a forest
```
```suggestion
	/* Count the trees around the clear tile to determine if it's near a forest */
```

src/tree_cmd.cpp
```diff
@@ -80,6 +81,34 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
...
+		for (int y = -2; y <= 2; y++) {
+			TileIndex vincity = TileAddWrap(tile, x, y);
+			if (vincity != INVALID_TILE &&
+				IsTileType(vincity, MP_TREES)) {
+				planted_tile_count++;
+			}
+		}
+	}
+
+	return (planted_tile_count >= 6);
```
2TallTyler: Magic number should be a constant.

src/tree_cmd.cpp
```diff
@@ -80,6 +81,34 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
...
+*/
+static bool IsNearbyForest(TileIndex tile)
+{
+	uint planted_tile_count = 0;
+
+	// An already planted tilte can always be planted again
+	if (IsTileType(tile, MP_TREES)) return true;
+
+	// Count the trees arround the clear tile to determine if it's near a forest
+	for (int x = -2; x <= 2; x++) {
```
2TallTyler: You might use `CircularTileSearch()` for this?

src/tree_cmd.cpp
```diff
@@ -799,6 +848,11 @@ static void TileLoop_Trees(TileIndex tile)
 						}
 						break;
 				}
+
+				// When using improved trees, when a "alone" tree is dead, a new one is planted immediately
```
```suggestion
				/* When using improved trees, when a "alone" tree is dead, a new one is planted immediately. */
```

src/tree_cmd.cpp
```diff
@@ -778,6 +821,12 @@ static void TileLoop_Trees(TileIndex tile)
 				AddTreeCount(tile, -1);
 				SetTreeGrowth(tile, 3);
 			} else {
+				// Backups the type of tree if using improved trees
```
```suggestion
				/* Backups the type of tree if using improved trees */
```

src/tree_cmd.cpp
```diff
@@ -249,6 +288,7 @@ void PlaceTreesRandomly()
 
 	i = Map::ScaleBySize(DEFAULT_TREE_STEPS);
 	if (_game_mode == GM_EDITOR) i /= EDITOR_TREE_DIV;
+	if (_settings_game.game_creation.tree_placer == TP_IMPROVED) i /= IMPROVED_TREES_DIV;
```
2TallTyler: Needs a comment to explain that we plant a limited number of "alone trees" outside existing forests.

src/tree_cmd.cpp
```diff
@@ -193,11 +224,19 @@ static void PlaceTreeGroups(uint num_groups)
 			int x = GB(r, 0, 5) - 16;
 			int y = GB(r, 8, 5) - 16;
 			uint dist = abs(x) + abs(y);
+			uint max_dist = MAX_DISTANCE_FROM_CENTER;
+
+			if (_settings_newgame.game_creation.tree_placer == TP_IMPROVED) {
```
2TallTyler: Needs a comment explaining why the distance is different when planting forests, I think.

2TallTyler (commented):
I like this feature a lot, but think it should be a new option instead of a drastic change to the existing "Improved" option. I see in the commit history that this was originally the case. Speaking of commit history, it should be [squashed using an interactive rebase](https://matthewsetter.com/git-interactive-rebase/).

Moth-Tolias:
oh, whoops. my screen hadn't updated before i hit the the rebase button, so i didn't see the review. i'll see about that now.

src/tree_cmd.cpp
Moth-Tolias: it looks like that only returns a boolean rather than an integer, unfortunately.

2TallTyler: `if CircularTileSearch(IsTreeCallback) planted_tile_count++;` ?

src/tree_cmd.cpp
Moth-Tolias: "alone trees" seems a rather odd phrase. if you have any recommendations for a better one i would be interested to hear them.

src/tree_cmd.cpp
Moth-Tolias: I'm not actually sure myself. shall i remove this part?

Moth-Tolias:
> I like this feature a lot, but think it should be a new option instead of a drastic change to the existing "Improved" option. I see in the commit history that this was originally the case.

there are, in my opinion, two separate changes in this patch - the change to tree placement at worldgen, and the change to tree spreading during gameplay. the actual tree spreading should probably go in the tree spreading option, but perhaps the worldgen placement should be kept as the "improved" option?

src/tree_cmd.cpp
Moth-Tolias: wouldn't that hit the first tree every time?

2TallTyler: Oh, er, yes. I think you can pass a reference to CicularTileSearch that you can update. I'd suggest looking at other uses to see if anything is applicable.

2TallTyler:
If they are two separate changes, I'd suggest splitting them into separate PRs. Easier to review that way. 🙂

Moth-Tolias:
I'll see what i can do.

Moth-Tolias:
i've split the worldgen changes into #12719. this pr now only has the spreading code.

tabytac:
I know I may be a bit late to this, but I wanted to mention the "Perfect" tree placer setting, which was added in [2021 here](https://github.com/JGRennison/OpenTTD-patches/pull/278). This setting seems to eliminate the shape-based tree placer altogether, resulting in a more sparse distribution of trees at low altitudes that increases with elevation. For reference, here is the same map seed with only the tree setting changed in JGRpp 0.58.3, as shown in the pictures below.

I don't think this should replace the improved setting proposed in this PR, but I wanted to bring it to attention in case anyone interested in this PR finds it useful or thinks it could be added as a "sparse" or something option in the dropdown in a future PR.

openttd_2024-05-26_09 31 16

openttd_2024-05-26_09 31 43

Moth-Tolias:
> "Perfect" tree placer setting

interesting! if these current prs get merged i'd certainly like to investigate that. though i don't agree at all with VaccumBreather's assertion that tile/coordinate based indexing inevitably results in banding and should never be used - there are plenty of decent hash functions that can and should be used for randomization outside of security-critical contexts.

src/tree_cmd.cpp
PeterN: Set up a `TileArea` and iterate with that?

Also tiles can have 1 to 3 trees on them, not sure if this should be taken into account.

src/tree_cmd.cpp
```diff
@@ -799,6 +837,11 @@ static void TileLoop_Trees(TileIndex tile)
 						}
 						break;
 				}
+
+				/* When using improved trees, when a "alone" tree is dead, a new one is planted immediately. */
+				if (_settings_game.game_creation.tree_placer == TP_IMPROVED && !IsNearbyForest(tile)) {
+					PlantTreesOnTile(tile, treetype, 0, 0);
+				}
```
PeterN: Maybe this case should be tested before converting the tree to a clear tile, to avoid unnecessary work.

PeterN (commented):
I like the general principle, but the results are a bit stark, with clear diamond shapes caused by the square forest test.

It appears that there are NO trees outside a forest, which looks odd. I would suggest a low random chance of a tree spawning anyway.

I'm speculating here, but this introduces a lot of tile iteration for tree growth during the tile loop, so could have quite a negative impact on game performance on large maps.

Moth-Tolias:
> I like the general principle, but the results are a bit stark, with clear diamond shapes caused by the square forest test.
> 
> It appears that there are NO trees outside a forest, which looks odd. I would suggest a low random chance of a tree spawning anyway.

the original pr had this comment as well. i figured it should logically be its own change: #12719 

> I'm speculating here, but this introduces a lot of tile iteration for tree growth during the tile loop, so could have quite a negative impact on game performance on large maps.

my intent was to make this forest spreading a new spreading setting rather than relying on the "improved" placement, but i'm realizing that doing so would involve changing the spreading options from 

- no spreading
- spread only in rainforests
- spread everywhere

to

- no spreading
- spread only in rainforests
- spread only in forests
- spread only in forests that are in rainforests
- spread everywhere

which is a bit more complexity than i was bargaining for. i'd like your feedback before i go ahead with it.

src/tree_cmd.cpp
```diff
@@ -80,6 +82,30 @@ static bool CanPlantTreesOnTile(TileIndex tile, bool allow_desert)
...
+	if (IsTileType(tile, MP_TREES)) return true;
+
+	/* Count the trees around the clear tile to determine if it's near a forest */
+	for (TileIndex t : TileArea(tile).Expand(FOREST_SEARCH_RADIUS)) {
+		if (IsTileType(t, MP_TREES)) {
+			planted_tile_count++;
+		}
+	}
+
+	return (planted_tile_count >= FOREST_THRESHOLD);
```
PeterN: The threshold is only 6, but you iterate for all 25 tiles.

You can early return true after incrementing `planted_tile_count` if it's reached the threshold, and then just return false at the end of the function.

This will avoid unnecessarily checking more tiles than necessary.

src/tree_cmd.cpp
```diff
@@ -99,11 +99,14 @@ static bool IsNearbyForest(TileIndex tile)
 	/* Count the trees around the clear tile to determine if it's near a forest */
 	for (TileIndex t : TileArea(tile).Expand(FOREST_SEARCH_RADIUS)) {
 		if (IsTileType(t, MP_TREES)) {
-			planted_tile_count++;
+			if (++planted_tile_count >= FOREST_THRESHOLD)
+			{
+				return true;
+			}
 		}
```
PeterN:
```suggestion
		if (!IsTileType(t, MP_TREES)) continue;
		++planted_tile_count;
		if (planted_tile_count >= FOREST_THRESHOLD) return true;
```

1) early continue.
2) prefer to avoid side effects in if-conditions.
3) unnecessary bracing.

PR #12714, opened 2024-05-24
Add: Setting to disable news message for old vehicles

merni-ns:
## Motivation / Problem
When playing without breakdowns (which is [quite common](https://survey.openttd.org/summaries/2024/wk20/14.1#game.settings.difficulty.vehicle_breakdowns)), vehicles getting old matters less. However, news messages about this still keep appearing, and it is not possible to disable them without also disabling other useful messages like lost, stuck, negative income, etc.

## Description
Add a setting to disable the news messages about vehicles getting old.

## Limitations
The whole set of news settings are a bit of a mess -- there are settings to control whole news types (off/summary/full), then settings to control a few specific triggers. A better system of filtering might be nice.

## Checklist for review

### DIFF
--- a/src/lang/english.txt
+++ b/src/lang/english.txt
@@ -1488,6 +1488,9 @@ STR_CONFIG_SETTING_WARN_INCOME_LESS                             :Warn if a vehic
 STR_CONFIG_SETTING_WARN_INCOME_LESS_HELPTEXT                    :When enabled, a news message gets sent when a vehicle has not made any profit within a year
 STR_CONFIG_SETTING_WARN_INCOME_LESS_HELPTEXT_PERIOD             :When enabled, a news message gets sent when a vehicle has not made any profit within a period
 
+STR_CONFIG_SETTING_WARN_OLD_VEHICLE                             :Warn if a vehicle is getting old: {STRING2}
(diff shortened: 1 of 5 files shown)


glx22:
Vehicle age matters for performance rating IIRC.

2TallTyler:
Vehicle age is also used for variable running costs in Timberwolf's Trains (old vehicles get very expensive).

2TallTyler:
Code looks fine to me and keeping as a separate setting (instead of automatically changing the behavior when breakdowns are off) makes the most sense to me.

If you test it (limitations says you have not) to ensure it works, I will approve. 😉

2TallTyler (approved):

DefinitelyNotRau117:
Hmm. It would be nice to add a second level for this setting, something like "allow priority messages from vehicles from NewGRF", and support for such messages, preferably separately for each vehicle - this would work well for xUSSR, because not all trains there change their characteristics upon reaching a certain age.

In this case, a compromise can be reached, when standard vehicles do not receive messages about obsolescence, but only "special" ones from newgrfs.

LordAro (approved):

PR #12718, opened 2024-05-25
Codechange: Replace FILE * with FileHandle class.

PeterN:
## Motivation / Problem

User of `FILE *` are expected to clean up manually. There is the `FileCloser` class which wraps a `FILE *`, but this has to be manually instantiated. There is also `std::unique_ptr<FILE, FileDeleter>`, but this also relies on wrapping a `FILE *`.

## Description

Replace `FileCloser` with `FileHandle` class. This still wraps a `FILE *`, but the wrapping is handled internally. To enforce the RAII concept, `FileHandle` cannot have a `nullptr` `FILE *`, and the static methods Open/OpenNative are used to return a `std::optional<FileHandle>`

`FileHandle` uses `std::unique<FILE, FileDeleter>` internally so that it also doesn't have to manually manage the file pointer, and so that copying a FileHandle is automatically deleted.

`FioFOpenFile()` also now returns a `std::optional<FileHandle>`, and all other direct uses of `fopen()` are replaced with `FioFOpenFile()`

This standardises all file access and removes the need for manual clean up.

## Limitations

This is still just wrapping a C API, but ifstream/ofstream seem to have some differences that make conversion a bit awkward.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -1130,15 +1130,14 @@ DEF_CONSOLE_CMD(ConExec)
 
 	if (argc < 2) return false;
 
-	FILE *script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
+	auto script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
 
-	if (script_file == nullptr) {
+	if (!script_file.has_value()) {
 		if (argc == 2 || atoi(argv[2]) != 0) IConsolePrint(CC_ERROR, "Script file '{}' not found.", argv[1]);
 		return true;
 	}
 
 	if (_script_current_depth == 11) {
-		FioFCloseFile(script_file);
 		IConsolePrint(CC_ERROR, "Maximum 'exec' depth reached; script A is calling script B is calling script C ... more than 10 times.");
 		return true;
 	}
@@ -1147,7 +1146,7 @@ DEF_CONSOLE_CMD(ConExec)
 	uint script_depth = _script_current_depth;
 
 	char cmdline[ICON_CMDLN_SIZE];
-	while (fgets(cmdline, sizeof(cmdline), script_file) != nullptr) {
+	while (fgets(cmdline, sizeof(cmdline), *script_file) != nullptr) {
 		/* Remove newline characters from the executing script */
 		for (char *cmdptr = cmdline; *cmdptr != '\0'; cmdptr++) {
 			if (*cmdptr == '\n' || *cmdptr == '\r') {
@@ -1163,12 +1162,11 @@ DEF_CONSOLE_CMD(ConExec)
 		if (_script_current_depth == script_depth - 1) break;
 	}
 
-	if (ferror(script_file)) {
+	if (ferror(*script_file) != 0) {
 		IConsolePrint(CC_ERROR, "Encountered error while trying to read from script file '{}'.", argv[1]);
 	}
 
 	if (_script_current_depth == script_depth) _script_current_depth--;
-	FioFCloseFile(script_file);
 	return true;
 }
 
@@ -1222,7 +1220,7 @@ extern void ShowFramerateWindow();
 
 DEF_CONSOLE_CMD(ConScript)
 {
-	extern FILE *_iconsole_output_file;
+	extern std::optional<FileHandle> _iconsole_output_file;
 
 	if (argc == 0) {
 		IConsolePrint(CC_HELP, "Start or stop logging console output to a file. Usage: 'script <filename>'.");
@@ -1233,8 +1231,8 @@ DEF_CONSOLE_CMD(ConScript)
 	if (!CloseConsoleLogIfActive()) {
 		if (argc < 2) return false;
 
-		_iconsole_output_file = fopen(argv[1], "ab");
-		if (_iconsole_output_file == nullptr) {
+		_iconsole_output_file = FileHandle::Open(argv[1], "ab");
+		if (!_iconsole_output_file.has_value()) {
 			IConsolePrint(CC_ERROR, "Could not open console log file '{}'.", argv[1]);
 		} else {
 			IConsolePrint(CC_INFO, "Console log output started to '{}'.", argv[1]);
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -120,11 +120,8 @@ static void FillValidSearchPaths(bool only_local_path)
  */
 bool FioCheckFileExists(const std::string &filename, Subdirectory subdir)
 {
-	FILE *f = FioFOpenFile(filename, "rb", subdir);
-	if (f == nullptr) return false;
-
-	FioFCloseFile(f);
-	return true;
+	auto f = FioFOpenFile(filename, "rb", subdir);
+	return f.has_value();
 }
 
 /**
@@ -136,14 +133,6 @@ bool FileExists(const std::string &filename)
 {
 	std::error_code ec;
 	return std::filesystem::exists(OTTD2FS(filename), ec);
-}
-
-/**
- * Close a file in a safe way.
- */
-void FioFCloseFile(FILE *f)
-{
-	fclose(f);
 }
 
 /**
@@ -191,7 +180,7 @@ std::string FioFindDirectory(Subdirectory subdir)
 	return _personal_dir;
 }
 
-static FILE *FioFOpenFileSp(const std::string &filename, const char *mode, Searchpath sp, Subdirectory subdir, size_t *filesize)
+static std::optional<FileHandle> FioFOpenFileSp(const std::string &filename, const char *mode, Searchpath sp, Subdirectory subdir, size_t *filesize)
 {
 #if defined(_WIN32)
 	/* fopen is implemented as a define with ellipses for
@@ -201,7 +190,6 @@ static FILE *FioFOpenFileSp(const std::string &filename, const char *mode, Searc
 	wchar_t Lmode[5];
 	MultiByteToWideChar(CP_ACP, 0, mode, -1, Lmode, static_cast<int>(std::size(Lmode)));
 #endif
-	FILE *f = nullptr;
 	std::string buf;
 
 	if (subdir == NO_DIRECTORY) {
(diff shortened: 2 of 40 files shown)


src/music/dmusic.cpp
```diff
@@ -429,14 +429,13 @@ bool DLSFile::ReadDLSWaveList(FILE *f, DWORD list_length)
...
-bool DLSFile::LoadFile(const wchar_t *file)
+bool DLSFile::LoadFile(const std::filesystem::path &file)
 {
 	Debug(driver, 2, "DMusic: Try to load DLS file {}", FS2OTTD(file));
 
-	FILE *f = _wfopen(file, L"rb");
-	if (f == nullptr) return false;
-
-	FileCloser f_scope(f);
+	auto of = FileHandle::OpenNative(file, "rb");
```
LordAro: Why is OpenNative (or _wfopen) required here? Or rather, I see the difference between the two but don't follow why this is the only needed usage of it

src/console_cmds.cpp
```diff
@@ -1112,15 +1112,14 @@ DEF_CONSOLE_CMD(ConExec)
 
 	if (argc < 2) return false;
 
-	FILE *script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
+	auto script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
 
-	if (script_file == nullptr) {
+	if (!script_file.has_value()) {
```
rubidium42, line 1135: This PR is inconsistent whether to use `x` or `x.has_value()` for boolean expressions. I'm seeing both constructs being used. Is that deliberate?

src/fileio_type.h
```diff
@@ -148,4 +148,31 @@ enum Searchpath : unsigned {
 
 DECLARE_POSTFIX_INCREMENT(Searchpath)
 
+class FileHandle {
+public:
+	static std::optional<FileHandle> OpenNative(const std::filesystem::path &filename, const std::string &mode);
+	static std::optional<FileHandle> Open(const std::string &filename, const std::string &mode);
+
+	inline void Close() { f.reset(); }
```
rubidium42:
```suggestion
	inline void Close() { this->f.reset(); }
```

src/fileio_type.h
```diff
@@ -148,4 +148,31 @@ enum Searchpath : unsigned {
...
+
+	inline operator FILE *()
+	{
+		assert(f != nullptr);
```
rubidium42:
```suggestion
		assert(this->f != nullptr);
```

src/heightmap.cpp
```diff
@@ -265,40 +259,34 @@ static void ReadHeightmapBMPImageData(uint8_t *map, BmpInfo *info, BmpData *data
...
 static bool ReadHeightmapBMP(const char *filename, uint *x, uint *y, uint8_t **map)
 {
-	FILE *f;
+	auto f = FioFOpenFile(filename, "rb", HEIGHTMAP_DIR);
+	if (!f.has_value()) {
+		ShowErrorMessage(STR_ERROR_BMPMAP, STR_ERROR_PNGMAP_FILE_NOT_FOUND, WL_ERROR);
+		return false;
+	}
+
+	BmpBuffer buffer( std::move(*f));
```
rubidium42:
```suggestion
	BmpBuffer buffer(std::move(*f));
```

src/music/dmusic.cpp
rubidium42: Because paths have two encodings. One is our default UTF-8, and the other is operating system native which under Windows is of course `wchar_t` and UTF-16.

Since the path is gathered from the registry and uses Windows specific path functions, the calling code already has the (Windows) native path, so there's little point in converting it to `UTF-8` and then back to its native encoding.

src/saveload/oldloader.cpp
```diff
@@ -65,7 +65,7 @@ static uint8_t ReadByteFromFile(LoadgameState *ls)
 	if (ls->buffer_cur >= ls->buffer_count) {
 
 		/* Read some new bytes from the file */
-		int count = (int)fread(ls->buffer, 1, BUFFER_SIZE, ls->file);
+		int count = (int)fread(ls->buffer, 1, BUFFER_SIZE, *ls->file);
```
rubidium42:
```suggestion
		int count = static_cast<int>(fread(ls->buffer, 1, BUFFER_SIZE, *ls->file));
```

src/script/squirrel.cpp
```diff
@@ -622,40 +622,37 @@
...
 	}
 
-	if (file == nullptr) {
+	if (!file) {
 		return sq_throwerror(vm, "cannot open the file");
 	}
 	unsigned short bom = 0;
 	if (size >= 2) {
-		[[maybe_unused]] size_t sr = fread(&bom, 1, sizeof(bom), file);
+		[[maybe_unused]] size_t sr = fread(&bom, 1, sizeof(bom), *file);
```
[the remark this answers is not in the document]
rubidium42:
```suggestion
		if (fread(&bom, 1, sizeof(bom), *file) != 1) return sq_throwerror(vm, "cannot read the file");
```

rubidium42 (commented):
Is there something `safeguard-y` that can be done to just prevent calling `fopen`/`fclose` in the rest of the code base?

Not necessarily needed for this PR, though.

src/console_cmds.cpp
PeterN: I intended to use `has_value()` consistently, but clearly I failed in places.

src/script/squirrel.cpp
```diff
@@ -622,40 +622,37 @@ SQRESULT Squirrel::LoadFile(HSQUIRRELVM vm, const std::string &filename, SQBool
...
 	}
 
-	if (file == nullptr) {
+	if (!file.has_value()) {
 		return sq_throwerror(vm, "cannot open the file");
 	}
 	unsigned short bom = 0;
 	if (size >= 2) {
-		[[maybe_unused]] size_t sr = fread(&bom, 1, sizeof(bom), file);
+		if (fread(&bom, 1, sizeof(bom), *file) != 1) return sq_throwerror(vm, "cannot read the file");;
```
rubidium42: This might solve the regression-test failure. I haven't tested it though.
```suggestion
		if (fread(&bom, 1, sizeof(bom), *file) != sizeof(bom)) return sq_throwerror(vm, "cannot read the file");;
```

src/fileio.cpp
```diff
@@ -1040,20 +1024,18 @@ void SanitizeFilename(std::string &filename)
  */
 std::unique_ptr<char[]> ReadFileToMem(const std::string &filename, size_t &lenp, size_t maxsize)
 {
-	FILE *in = fopen(filename.c_str(), "rb");
-	if (in == nullptr) return nullptr;
-
-	FileCloser fc(in);
+	auto in = FileHandle::Open(filename, "rb");
+	if (!in) return nullptr;
```
rubidium42:
```suggestion
	if (!in.has_value()) return nullptr;
```

rubidium42 (commented):
Looks okay, except for one missed `.has_value()`.

Someone probably needs to shoehorn-merge this due to the remaining CodeQL warnings, which I already dismissed.

rubidium42 (dismissed):

rubidium42:
Stupid MSVC :( 
Now it's complaining about the same thing in the next function, when it didn't before... smells very whack-a-moley. Though I wouldn't know how to solve this in another way, except creating a new function to extract bytes from a larger integer, which feels its own kind of duplicate.

PeterN:
Yeah, just the act of putting `#include <filesystem>` in stdafx.h is causing this. https://github.com/PeterN/OpenTTD/pull/6

PeterN:
Got rid of the `#include <filesystem>` change, got rid of `FileHandle::OpenNative()`, and converted those callers to use UTF8 encoding instead.

src/fileio.cpp
```diff
@@ -480,12 +466,13 @@ bool TarScanner::AddFile(const std::string &filename, size_t, [[maybe_unused]] c
...
 	if (it != _tar_list[this->subdir].end()) return false;
 
-	FILE *f = fopen(filename.c_str(), "rb");
+	auto of = FileHandle::Open(filename, "rb");
 	/* Although the file has been found there can be
 	 * a number of reasons we cannot open the file.
 	 * Most common case is when we simply have not
 	 * been given read access. */
-	if (f == nullptr) return false;
+	if (!of) return false;
```
rubidium42: `of` is the optional FileHandle, right?
```suggestion
	if (!of.has_value()) return false;
```

src/fileio.cpp
```diff
@@ -584,14 +570,13 @@ bool TarScanner::AddFile(const std::string &filename, size_t, [[maybe_unused]] c
...
 			Debug(misc, 0, "The file '{}' can't be read as a valid tar-file", filename);
-			fclose(f);
 			return false;
 		}
 		pos += skip;
 	}
 
 	Debug(misc, 4, "Found tar '{}' with {} new files", filename, num);
-	fclose(f);
+	f.Close();
```
rubidium42: Can't this be left out? Won't it be closed after the `return true` below, just like the `return false` cases above?

src/network/network_content.cpp
```diff
@@ -524,7 +522,7 @@ bool ClientNetworkContentSocketHandler::BeforeDownload()
 	if (this->curInfo->filesize != 0) {
 		/* The filesize is > 0, so we are going to download it */
 		std::string filename = GetFullFilename(this->curInfo, true);
-		if (filename.empty() || (this->curFile = fopen(filename.c_str(), "wb")) == nullptr) {
+		if (filename.empty() || !(this->curFile = FileHandle::Open(filename, "wb"))) {
```
rubidium42:
```suggestion
		if (filename.empty() || !(this->curFile = FileHandle::Open(filename, "wb")).has_value()) {
```

src/console_cmds.cpp
```diff
@@ -1233,8 +1231,8 @@ DEF_CONSOLE_CMD(ConScript)
 	if (!CloseConsoleLogIfActive()) {
 		if (argc < 2) return false;
 
-		_iconsole_output_file = fopen(argv[1], "ab");
-		if (_iconsole_output_file == nullptr) {
+		_iconsole_output_file = FileHandle::Open(argv[1], "ab");
+		if (_iconsole_output_file.has_value()) {
```
rubidium42: I think the logic got inverted here for some reason.
```suggestion
		if (!_iconsole_output_file.has_value()) {
```

rubidium42 (approved):

PR #12724, opened 2024-05-27
Codechange: Use std::unordered_map for SpriteFontCache's glyph map.

PeterN:
## Motivation / Problem

SpriteFontCache uses a multi-layer look up table to map characters to SpriteIDs which is manually managed with pointers, malloc and free.

TrueTypeFontCache has a similar structure for mapping glyphs to raw Sprites.

## Description

~~Add a custom SparseArray container. This is in fontcache.h as that's where it's used.~~

~~Replaces a C-style malloc'd array and malloc'd contents with a vector and unique_ptr.~~

~~Keeps the implementation details mostly within the SparseArray class, and no manual memory clean up is necessary.~~

~~This does not yet touch TrueTypeFontCache's glyph map, as that needs extra handling.~~

Replace with std::unordered_map.

Replaces a C-style malloc'd array and malloc'd contents, and no manual memory clean up is necessary.

## Limitations

~~This could be a std::map, however that is "very sparse", and I measured performance at at least 10x slower than this container.~~

## Checklist for review

### DIFF
--- a/src/fontcache/spritefontcache.cpp
+++ b/src/fontcache/spritefontcache.cpp
@@ -35,7 +35,7 @@ static int ScaleFontTrad(int value)
  * Create a new sprite font cache.
  * @param fs The font size to create the cache for.
  */
-SpriteFontCache::SpriteFontCache(FontSize fs) : FontCache(fs), glyph_to_spriteid_map(nullptr)
+SpriteFontCache::SpriteFontCache(FontSize fs) : FontCache(fs)
 {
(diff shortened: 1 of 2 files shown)


rubidium42:
Is `std::unordered_map` also 10x slower? Or is there something else why `std::unordered_map` would not work?

LordAro:
Ah, but how does memory usage compare? :)

rubidium42 (approved):

PR #12725, opened 2024-05-27
Codechange: Use std::unordered_map for storing TrueTypeFontCache's GlyphEntry map.

PeterN:
## Motivation / Problem

Similar to #12724, TrueTypeFontCache has yet-another-implementation of a 'sparse' array using blocks of malloc'd arrays.

## Description

Replace with manually managed pointers and arrays with a `std::unordered_map`.

GlyphEntry previously shared sprite data for unknown glyphs, and freeing this memory was specially handled in the clean up loops when cleaning the glyph map. This is not simple to replicate with a destructor (as the destructor is always called, not just for glyph map entries), so I elected to not share the memory instead.

This required refactoring how sprite memory is allocated using a stateful helper instead of function pointers. This first change is the same as in #11495.

## Checklist for review

### DIFF
--- a/src/blitter/32bpp_anim_sse4.hpp
+++ b/src/blitter/32bpp_anim_sse4.hpp
@@ -39,7 +39,7 @@ public:
 	template <BlitterMode mode, Blitter_32bppSSE_Base::ReadMode read_mode, Blitter_32bppSSE_Base::BlockType bt_last, bool translucent, bool animated>
 	void Draw(const Blitter::BlitterParams *bp, ZoomLevel zoom);
 	void Draw(Blitter::BlitterParams *bp, BlitterMode mode, ZoomLevel zoom) override;
-	Sprite *Encode(const SpriteLoader::SpriteCollection &sprite, AllocatorProc *allocator) override {
+	Sprite *Encode(const SpriteLoader::SpriteCollection &sprite, SpriteAllocator &allocator) override {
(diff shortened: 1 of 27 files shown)


PeterN:
Hmm, I can probably remove the first part for now and keep using malloc/free (but removing the deduplication for simplicity)

rubidium42 (commented):
Why use a `std::vector`? Wouldn't a `std::unique_ptr` suffice? After all, the size doesn't seem to be used.

With `std::vector` `GlyphEntry` is 16/32 bytes, with `std::unique_ptr` it's 8/16 bytes (32/64 bit MSVC/GCC).

I was also thinking quite a bit about whether shared pointers could be used instead of removing the deduplication of the question mark sprite. Though that would only seriously benefit someone who has loaded a language-font combination that has almost no matching glyphs, and then it would start paying off. In the normal case there'd be barely any question mark sprites, so essentially should have no overhead for it to be worth it.

I think that might be even a better reason for axing the deduplication than "it's not easy to do it with the structure I've made now". After all, the complexity only starts paying dividends when there are many missing characters. That's not something I would have optimized for.

PeterN:
Awesome, the entire de-duplication stuff is pointless because at this point key is always 0 anyway. And we should really just be using the glyph at position 0 instead of substituting with `?`, as it is the dimensions of GlyphID 0 that the layouter is using, not the glyph we've substituted.

rubidium42 (approved):

PR #12742, opened 2024-06-01
Codechange: Update each town label dimensions once during map gen.

PeterN:
## Motivation / Problem

When debugging layouting I noticed that during map generation that each successive increase in a town's population causes the town's sign to be updated, even though it is never visible yet.

If town population is visible in town signs, then each population change also causes the town sign to be relayouted with the new population. This not particularly expensive but given it happens multiple times per town, does add up.

## Description

Update each town label dimension only once during town generation. This avoids re-layouting the label for each change in population as the town is grown.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -79,6 +79,8 @@ void RebuildTownKdtree()
 	_town_kdtree.Build(townids.begin(), townids.end());
 }
 
+/** Set if a town is being generated. */
+static bool _generating_town = false;
 
 /**
  * Check if a town 'owns' a bridge.
@@ -442,6 +444,8 @@ void ClearAllTownCachedNames()
 static void ChangePopulation(Town *t, int mod)
 {
 	t->cache.population += mod;
+	if (_generating_town) [[unlikely]] return;
+
 	InvalidateWindowData(WC_TOWN_VIEW, t->index); // Cargo requirements may appear/vanish for small populations
 	if (_settings_client.gui.population_in_label) t->UpdateVirtCoord();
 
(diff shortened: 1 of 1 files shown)


src/town_cmd.cpp
```diff
@@ -2036,6 +2039,9 @@ static void DoCreateTown(Town *t, TileIndex tile, uint32_t townnameparts, TownSi
 		GrowTown(t);
 	} while (--i);
 
+	t->UpdateVirtCoord();
```
2TallTyler, line 2042: Moving this does not change the state of `_generating_town` when these lines are called, unless I'm reading something wrong.

src/town_cmd.cpp
```diff
@@ -442,6 +444,8 @@ void ClearAllTownCachedNames()
 static void ChangePopulation(Town *t, int mod)
 {
 	t->cache.population += mod;
+	if (_generating_town) [[unlikely]] return;
```
2TallTyler, line 447: Could this use the existing global `_generating_world` instead of creating a new global?

src/town_cmd.cpp
PeterN: These are moved so that it's done after the town growth is completed instead of before. `_generating_town` is tested only in `ChangePopulation()`, not these two calls.

2TallTyler: Oh, I see -- this does the update that is skipped during `ChangePopulation()`.

src/town_cmd.cpp
PeterN: World generation has multiple stages, with town generation near the start. Town population is also updated in the later stages when we 'pre-roll' game ticks for a while, and there we want the signs to be updated.

2TallTyler (approved):

PR #12750, opened 2024-06-03
Add: [Script] GSController::TriggerSave to trigger saves from game scripts

michalc:
## Motivation / Problem

The change in https://github.com/OpenTTD/OpenTTD/pull/10655, released as part of OpenTTD 14, changed how auto saving works from every X amount of game time, to every X amount of real time. This is an improvement for normal play, but there are cases where saving every X amount of game time is useful. For example when developing and comparing AIs, or using OpenTTD to conduct experiments such as via https://github.com/michalc/OpenTTDLab which extracts data from savegame files.

## Description

This change addresses this by adding the GSController::TriggerSave function to the GameScript API that allows game scripts to trigger a save, and with a specific file name. Because game script have access to the in-game date through GSDate.GetCurrentDate, and can sleep arbitrary amounts of ticks, it means they can approximate the pre OpenTTD 14 behaviour of saving every X amount of game time.

It also means that via game scripts, games can be saved immediately after game start, which wasn't possible with the previous autosave behaviour.

The name of the function is not Save to avoid conflict and confusion between the existing Save function that game scripts can implement to add custom data to savegame files.

## Limitations

None known

## Checklist for review

### DIFF
--- a/src/script/api/game/game_controller.hpp.sq
+++ b/src/script/api/game/game_controller.hpp.sq
@@ -21,6 +21,7 @@ void SQGSController_Register(Squirrel *engine)
 	SQGSController.DefSQStaticMethod(engine, &ScriptController::Break,             "Break",             2, ".s");
 	SQGSController.DefSQStaticMethod(engine, &ScriptController::GetSetting,        "GetSetting",        2, ".s");
 	SQGSController.DefSQStaticMethod(engine, &ScriptController::GetVersion,        "GetVersion",        1, ".");
+	SQGSController.DefSQStaticMethod(engine, &ScriptController::TriggerSave,       "TriggerSave",       2, ".s");
 	SQGSController.DefSQStaticMethod(engine, &ScriptController::Print,             "Print",             3, ".bs");
 
 	SQGSController.PostRegister(engine);
--- a/src/script/api/game_changelog.hpp
+++ b/src/script/api/game_changelog.hpp
@@ -17,6 +17,9 @@
  *
  * This version is not yet released. The following changes are not set in stone yet.
  *
+ * API additions:
+ * \li GSController::TriggerSave
+ *
  * \b 14.0
  *
  * API additions:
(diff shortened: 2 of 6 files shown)


TrueBrain:
GameScript is not intended or designed for this kind of functionality. Although I understand your usecase, I also realise it is rather niche and won't be used by many.

As such, a better place for functionality like this is either a custom patch, or some patch on top of the console scripting. Or some improvement to the rcon. Or admin port. All three are designed for this kind of functionality.

But GameScript is intended to steer the game (as in, the simulation itself) a user is playing in a certain direction. Giving GS control to these kind of functionalities, that are outside the simulation itself, is asking for a lot of trouble.

For example, a GS author might misunderstand this functionality, and fill the disk of its users with tons of savegames. Or leave it as debug statement by accident. The troubles with this are endless. And it also opens the floodgates with tons of other things like this.. savegames ... insert-any-other-console-command .. 

 Sorry to be the bearer of bad news :( but the GS isn't and shouldn't be the solution to any and all trouble :) Hope you understand :)

(Ps: this is my view, maybe other devs disagree with me)

James103:
As it currently stands, a Game Script could be made and uploaded to BaNaNaS right now which, in a build containing this PR, causes the game to try to save every tick, or even multiple times a tick, with random filenames, causing the game to slow down, the client's hard drive to fill rapidly, and a portion of the CPU to be at 100% usage.

At the very least, `GSController::TriggerSave` should cause the Game Script to sleep for the remainder of the tick every time it is executed (like all built-in functions that trigger commands).

michalc:
This was discussed at https://discord.com/channels/142724111502802944/1008473233844097104/1247278039927361546 and essentially yes because of the abuse possibilities that @TrueBrain and @James103 you mention, going for another way.

Best contender so far is some sort of console command that can schedule a console script, e.g. `trigger bla.scr on-month-change` and in that script it can trigger a save, and then re-schedule itself for the next month. Or something like that...

This actually should work well with OpenTTDLab: it already uses a `game_start.scr` console script https://github.com/michalc/OpenTTDLab/blob/793682558d23d68cc9ea99cb971eac37eaa8d8c7/openttdlab.py#L301, and so hopefully it should be a simple matter to add another line to it.

PR #12752, opened 2024-06-04
Change: Do not automatically connect double depots with track

steve-goldman:
## Motivation / Problem

Some players who build rail depots that face each other across a single track, also known as "double depots", prefer the depots not to be connected. The connecting track is constructed automatically when the second depot is constructed, so the player needs to remove the track each time.  This is tedious and error prone for the player.

image

## Description

Change to game behavior: the double depots no longer get connected with track automatically.

image

## Checklist for review

### DIFF
--- a/src/rail_gui.cpp
+++ b/src/rail_gui.cpp
@@ -148,7 +148,12 @@ void CcRailDepot(Commands, const CommandCost &result, TileIndex tile, RailType, 
 
 	if (IsTileType(tile, MP_RAILWAY)) {
 		PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir], _place_depot_extra_track[dir]);
-		PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir + 4], _place_depot_extra_track[dir + 4]);
+
+		/* Don't place the rail straight out of the depot of there is another depot across from it. */
+		Tile double_depot_tile = tile + TileOffsByDiagDir(dir);
+		bool is_double_depot = IsValidTile(double_depot_tile) && IsRailDepotTile(double_depot_tile);
+		if (!is_double_depot) PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir + 4], _place_depot_extra_track[dir + 4]);
+
 		PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir + 8], _place_depot_extra_track[dir + 8]);
 	}
 }


LordAro:
Nice. However, doesn't this also affect the scenario where there's another rail, rather than a depot?

LordAro:
I'd be tempted to look explicitly for a depot and skip it only in that case

And then probably just get rid of the setting.

steve-goldman:
> I'd be tempted to look explicitly for a depot and skip it only in that case
> 
> And then probably just get rid of the setting.

Thanks for the suggestion.  I updated the commit to consider the narrower double depot situation.

I left it as a setting (see updated screenshot in description above) as my instinct is that changing it could surprise some people.

Again, still happy to iterate.

src/rail_gui.cpp
```diff
@@ -145,10 +145,11 @@ void CcRailDepot(Commands, const CommandCost &result, TileIndex tile, RailType,
 	if (!_settings_client.gui.persistent_buildingtools) ResetObjectToPlace();
 
 	tile += TileOffsByDiagDir(dir);
+	bool is_double_depot = IsValidTile(tile) ? IsRailDepotTile(tile + TileOffsByDiagDir(dir)) : false;
```
2TallTyler: This tile validity check doesn't make sense. Shouldn't it check whether `tile + TileOffsByDiagDir(dir)` is a valid tile?

2TallTyler (commented):
I second the suggestion to remove the setting and let this be the new behavior. We can always add a setting if people complain.

OpenTTD has too many settings as it is, so we are careful not to add more without good reason. 😃

src/rail_gui.cpp
steve-goldman: Thanks for the feedback...

I was trying to use the same convention that that this function seemed to use for the `tile` param: Initially we assume `tile` is valid, then we change `tile` to be the tile adjacent to the depot, which might not be on the map, but we call `IsTileType` on it without checking for validity.  I took this to mean that a tile adjacent to a valid tile might not be valid, but it is still legal in the sense that we can safely pass it to functions such as `IsTileType`.

My change concerns the *next* tile over.  We could safely call `IsRailDepotTile` on that tile if we establish that the tile adjacent to it is valid.

Anyway, that was my thinking.  I agree that we could equivalently get this done by calling `IsValidTile` on that second tile, and the code might be clearer.

```cpp
	tile += TileOffsByDiagDir(dir);
	Tile next_tile = tile + TileOffsByDiagDir(dir);
	bool is_double_depot = IsValidTile(next_tile) && IsRailDepotTile(next_tile);
```

steve-goldman:
> I second the suggestion to remove the setting and let this be the new behavior. We can always add a setting if people complain.
> 
> OpenTTD has too many settings as it is, so we are careful not to add more without good reason. 😃

I updated the commit.  The new setting is gone.  Also localized and clarified the business logic based on your other feedback.

src/rail_gui.cpp
```diff
@@ -148,7 +148,11 @@ void CcRailDepot(Commands, const CommandCost &result, TileIndex tile, RailType,
 
 	if (IsTileType(tile, MP_RAILWAY)) {
 		PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir], _place_depot_extra_track[dir]);
-		PlaceExtraDepotRail(tile, _place_depot_extra_dir[dir + 4], _place_depot_extra_track[dir + 4]);
+
```
2TallTyler, line 151: This could use a brief comment explaining what's going on here, like `/* Don't connect to a double depot on the other side of the track. */`

Otherwise, looks good to me!

steve-goldman: Updated, thanks.

2TallTyler (approved):

PR #12754, opened 2024-06-04
Bribing possible if other co has excl. rights

bjornwarmedal:
Fix: A bribe can void an Exclusive Transport Rights contract, but it was not possible to bribe a town that already has an outstanding opinion of you.

## Motivation / Problem

If corporation A buys exclusive transport rights in SomeTown another corporation can void that contract with a successful bribe to the local authority. But if corporation B has an Outstanding relationship with SomeTown it can not bribe the local authority. Perfectly reasonable if the bribe only affected said relationship, but causes a lockout from affecting the contract.

## Description

This fixes so that bribing is possible if:
- The bribery setting is enabled AND
- You have a lower than Outstanding rating in the town OR
  - Someone else currently has an Exclusive Transport Rights contract with the town.

## Checklist for review

### DIFF
--- a/src/lang/french.txt
+++ b/src/lang/french.txt
@@ -1499,7 +1499,7 @@ STR_CONFIG_SETTING_TIMEKEEPING_UNITS_CALENDAR                   :Calendrier
 STR_CONFIG_SETTING_TIMEKEEPING_UNITS_WALLCLOCK                  :Pendule
 
 STR_CONFIG_SETTING_MINUTES_PER_YEAR                             :Minutes par année{NBSP}: {STRING}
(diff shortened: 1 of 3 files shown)


src/town_cmd.cpp
```diff
@@ -3576,7 +3576,7 @@ TownActions GetMaskOfTownActions(CompanyID cid, const Town *t)
 			const TownActions cur = (TownActions)(1 << i);
 
 			/* Is the company not able to bribe ? */
-			if (cur == TACT_BRIBE && (!_settings_game.economy.bribe || t->ratings[cid] >= RATING_BRIBE_MAXIMUM)) continue;
+			if (cur == TACT_BRIBE && (!_settings_game.economy.bribe || (t->ratings[cid] >= RATING_BRIBE_MAXIMUM && (t->exclusivity == _current_company || t->exclusivity == INVALID_COMPANY)))) continue;
```
LordAro: 4 `))))` is probably indicating that the condition is getting too complex. Split into multiple ifs or variables?

LordAro (commented):
Change seems reasonable on the face of it.

Commit message needs fixing though, and should be a `Change: ...`

PR #12760, opened 2024-06-06
Change: Position caret at correct position on glyph for RTL languages.

PeterN:
## Motivation / Problem

Caret position is positioned based on the left-coordinate of the current character. This is correct for LTR languages, but not for RTL

## Description

Set caret position to the right instead of left position of a glyph when the current language is RTL.

This requires altering all the layouters to store both the left and right positions, which is somewhat invasive as the right position information is only used for the caret position.

## Limitations

I have not tested the Windows or mac OS changes.

## Checklist for review

### DIFF
--- a/src/gfx_layout_fallback.cpp
+++ b/src/gfx_layout_fallback.cpp
@@ -49,10 +49,10 @@ public:
 		FallbackVisualRun(Font *font, const char32_t *chars, int glyph_count, int char_offset, int x);
 		const Font *GetFont() const override { return this->font; }
 		int GetGlyphCount() const override { return static_cast<int>(this->glyphs.size()); }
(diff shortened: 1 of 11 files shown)


src/gfx_layout_fallback.cpp
```diff
@@ -118,23 +118,22 @@ FallbackParagraphLayout::FallbackVisualRun::FallbackVisualRun(Font *font, const
 	this->glyph_to_char.reserve(char_count);
 
 	/* Positions contains the location of the begin of each of the glyphs, and the end of the last one. */
```
rubidium42: I guess this comment needs to updating.

PeterN: Hmm, I don't think it's even necessary, it was there to explain the +1.

rubidium42 (approved):

PR #12761, opened 2024-06-06
Add: [Console] schedule command to execute a script file later

michalc:
## Motivation / Problem

 The change in https://github.com/OpenTTD/OpenTTD/pull/10655, released as part of OpenTTD 14, changed how auto saving works from every X amount of game time, to every X amount of real time. This is an improvement for normal play, but there are cases where saving every X amount of game time is useful. For example when developing and comparing AIs, or using OpenTTD to conduct experiments such as via https://github.com/michalc/OpenTTDLab which extracts data from savegame files.

## Description

This change adds a general "schedule" console command that makes it possible for a console script to be scheduled for the future, where that script can, for example, contain a "save" command to save the game. That script can also then reschedule itself or another script, and thus making it possible to re-create the pre-OpenTTD 14 behaviour of saving the game every X amount of game time. (Possibly with a whole "suite" of game scripts that call each other in a chain, each of them hard coded with a different filename to allow all the savegames to be saved with different names, but that's a doable detail)

The current change only allows scheduling "on-next-calendar-month", since that is my specific use case, but the syntax is hopefully friendly to this being extended in the future. And by scheduling any console script, it is much more general and so hopefully useful than just for saving games.

There was some discussion on what this should be called at https://discord.com/channels/142724111502802944/1008473233844097104/1247278039927361546 (as well as some issues with an alternatively proposed solution using game scripts at https://github.com/OpenTTD/OpenTTD/pull/12750). Instead of "schedule", "trigger" was suggested for something to only be called once, but after writing code and output to the user, it didn't make it immediately clear that "trigger" means "once at some point in the future", and seemed too close to "exec". Also I think "schedule" can be used _both_ for one-off or repeated events in the future (in English it's common to say to "schedule an appointment", even if it's just the one).

Or code-wise:

schedule on-next-calendar-month my.scr (the only current behaviour)
schedule every-calendar-month my.scr (possible future behaviour?)
schedule "* 1 * *" my.scr (possible future cron-like behaviour?)

Possible future extensions could include some mechanism for cancelling something that has been scheduled, and I have no reason to think that mechanism would be different for something scheduled to repeat or just as a one-off, so I think also it makes sense to that we would have the one command for both.

## Limitations

The new "schedule" command:

- Only allows a single script to be scheduled - any (correct) usage of the "schedule" command will overwrite the previous, but hopefully the console messages make this clear
- Only allows a script to be scheduled once at the start of next month (but  this could be extended if useful).
- Does not offer a way to unschedule, but hopefully since the script can only be scheduled to run once in the future, and relatively close in the future (at the start of next month), this shouldn't be too much of an annoying missing feature.

## Checklist for review

### DIFF
--- a/src/console_cmds.cpp
+++ b/src/console_cmds.cpp
@@ -29,7 +29,7 @@
 #include "strings_func.h"
 #include "viewport_func.h"
 #include "window_func.h"
-#include "timer/timer_game_calendar.h"
+#include "timer/timer.h"
 #include "company_func.h"
 #include "gamelog.h"
 #include "ai/ai.hpp"
@@ -52,6 +52,24 @@
 
 /* scriptfile handling */
 static uint _script_current_depth; ///< Depth of scripts running (used to abort execution when #ConReturn is encountered).
+
+/* Scheduled execution handling. */
+static std::string _scheduled_monthly_script; ///< Script scheduled to execute by the 'schedule' console command (empty if no script is scheduled).
+
+/** Timer that runs every month of game time for the 'schedule' console command. */
+static IntervalTimer<TimerGameCalendar> _scheduled_monthly_timer = {{TimerGameCalendar::MONTH, TimerGameCalendar::Priority::NONE}, [](auto) {
+	if (_scheduled_monthly_script.empty()) {
+		return;
+	}
+
+	/* Clear the schedule before rather than after the script to allow the script to itself call
+	 * schedule without it getting immediately cleared. */
+	const std::string filename = _scheduled_monthly_script;
+	_scheduled_monthly_script.clear();
+
+	IConsolePrint(CC_DEFAULT, "Executing scheduled script file '{}'...", filename);
+	IConsoleCmdExec(std::string("exec") + " " + filename);
+}};
 
 /** File list storage for the console, for caching the last 'ls' command. */
 class ConsoleFileList : public FileList {
@@ -1151,6 +1169,35 @@ DEF_CONSOLE_CMD(ConExec)
 
 	if (_script_current_depth == script_depth) _script_current_depth--;
 	FioFCloseFile(script_file);
+	return true;
+}
+
+DEF_CONSOLE_CMD(ConSchedule)
+{
+	if (argc < 3 || std::string_view(argv[1]) != "on-next-calendar-month") {
+		IConsolePrint(CC_HELP, "Schedule a local script to execute later. Usage: 'schedule on-next-calendar-month <script>'.");
+		return true;
+	}
+
+	/* Check if the file exists. It might still go away later, but helpful to show an error now. */
+	if (!FioCheckFileExists(argv[2], BASE_DIR)) {
+		IConsolePrint(CC_ERROR, "Script file '{}' not found.", argv[2]);
+		return true;
+	}
+
+	/* We only support a single script scheduled, so we tell the user what's happening if there was already one. */
+	const std::string_view filename = std::string_view(argv[2]);
+	if (!_scheduled_monthly_script.empty() && filename == _scheduled_monthly_script) {
+		IConsolePrint(CC_INFO, "Script file '{}' was already scheduled to execute at the start of next calendar month.", filename);
+	} else if (!_scheduled_monthly_script.empty() && filename != _scheduled_monthly_script) {
+		IConsolePrint(CC_INFO, "Script file '{}' scheduled to execute at the start of next calendar month, replacing the previously scheduled script file '{}'.", filename, _scheduled_monthly_script);
+	} else {
+		IConsolePrint(CC_INFO, "Script file '{}' scheduled to execute at the start of next calendar month.", filename);
+	}
+
+	/* Store the filename to be used by _schedule_timer on the start of next calendar month. */
+	_scheduled_monthly_script = filename;
+
 	return true;
 }
 
@@ -2721,6 +2768,7 @@ void IConsoleStdLibRegister()
 	IConsole::CmdRegister("echo",                    ConEcho);
 	IConsole::CmdRegister("echoc",                   ConEchoC);
 	IConsole::CmdRegister("exec",                    ConExec);
+	IConsole::CmdRegister("schedule",                ConSchedule);
 	IConsole::CmdRegister("exit",                    ConExit);
 	IConsole::CmdRegister("part",                    ConPart);
 	IConsole::CmdRegister("help",                    ConHelp);


src/console_cmds.cpp
```diff
@@ -53,6 +53,20 @@
 /* scriptfile handling */
 static uint _script_current_depth; ///< Depth of scripts running (used to abort execution when #ConReturn is encountered).
 
+/* Scheduled execution handling */
+static std::string _scheduled_filename;
```
PeterN: `_scheduled_script` would be better. Also add Doxygen `///< comment`.

src/console_cmds.cpp
```diff
@@ -53,6 +53,20 @@
...
+static IntervalTimer<TimerGameCalendar> _schedule_timer = {{TimerGameCalendar::MONTH, TimerGameCalendar::Priority::NONE}, [](auto) {
+	if (_scheduled_filename != "") {
+		/* Clear the schedule before rather than after the script to allow the script to itself call
+		 * schedule without it getting immediately cleared */
+		const std::string filename = _scheduled_filename;
+		_scheduled_filename = "";
+
+		IConsolePrint(CC_DEFAULT, "Executing scheduled script file '{}'...", filename);
+		IConsoleCmdExec(std::string("exec") + " " + filename);
+	}
```
PeterN: Use `empty()` and `clear()` instead of `""`.
Prefer early return instead of nesting, i.e. return if the script is not set.

src/console_cmds.cpp
```diff
@@ -1154,6 +1168,37 @@ DEF_CONSOLE_CMD(ConExec)
...
+	/* Check if the file exists. It might still go away later, but helpful to show an error now */
+	FILE *script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
+	if (script_file == nullptr) {
+		IConsolePrint(CC_ERROR, "Script file '{}' not found.", argv[1]);
+		return true;
+	}
+	FioFCloseFile(script_file);
+
+	/* We only support a single script scheduled, so we tell the user what's happening if there was already one */
+	const std::string filename = std::string(argv[1]);
```
PeterN: Use `std::string_view` here to avoid making a copy.

src/console_cmds.cpp
```diff
@@ -1154,6 +1168,37 @@ DEF_CONSOLE_CMD(ConExec)
...
+	if (_scheduled_filename != "" and filename == _scheduled_filename) {
```
PeterN: Avoid `""` again. C++ uses `&&`, not `and`.

src/console_cmds.cpp
```diff
@@ -1154,6 +1168,37 @@ DEF_CONSOLE_CMD(ConExec)
...
+		return true;
+	}
+
+	/* Check if the file exists. It might still go away later, but helpful to show an error now */
+	FILE *script_file = FioFOpenFile(argv[1], "r", BASE_DIR);
+	if (script_file == nullptr) {
+		IConsolePrint(CC_ERROR, "Script file '{}' not found.", argv[1]);
+		return true;
+	}
+	FioFCloseFile(script_file);
```
PeterN: Use `FioCheckFileExists()`?

PeterN (dismissed):

src/console_cmds.cpp
michalc: Ah I was in Python mode!

src/console_cmds.cpp
```diff
@@ -53,6 +53,22 @@
...
 static uint _script_current_depth; ///< Depth of scripts running (used to abort execution when #ConReturn is encountered).
 
+/** Scheduled execution handling */
+static std::string _scheduled_script; ///< Script scheduled to execute by the 'schedule' console command (empty if no script is scheduled)
+static IntervalTimer<TimerGameCalendar> _schedule_timer = {{TimerGameCalendar::MONTH, TimerGameCalendar::Priority::NONE}, [](auto) {
+	if (_scheduled_script.empty()) {
+		return;
+	}
+
+	/** Clear the schedule before rather than after the script to allow the script to itself call
```
PeterN: Just `/* xxx */` for regular comments, `/** xxx */` is for Doxygen.

michalc: Oh!

michalc: Like now?

michalc: (I'm unsure about the top level `Scheduled execution handling`. I see https://github.com/OpenTTD/OpenTTD/blob/d7eb29d2922e10cd1f2fa8d7bba8069931e0a3f6/src/company_gui.cpp#L55 seems to be something fairly similar that does use the Doxygen format)

PeterN: Yeah, the top level one shouldn't be Doxygen-style either. You can Doxygen comment the _schedule_timer variable though.

michalc: Like now?

michalc:
> "On next month", is that the next calendar month or the next economy month (one minute)? Make sure that's clearly described in the in-game help.

Ah it is calendar month (I think!)... maybe the option should be "on-next-calendar-month"...

src/console_cmds.cpp
```diff
@@ -1154,6 +1172,35 @@ DEF_CONSOLE_CMD(ConExec)
...
+	}
+
+	/* We only support a single script scheduled, so we tell the user what's happening if there was already one */
+	const std::string_view filename = std::string_view(argv[1]);
+	if (!_scheduled_script.empty() && filename == _scheduled_script) {
+		IConsolePrint(CC_INFO, "Script file '{}' was already scheduled to execute at the start of next month.", filename);
+	} else if (!_scheduled_script.empty() && filename != _scheduled_script) {
+		IConsolePrint(CC_INFO, "Script file '{}' scheduled to execute at the start of next month, replacing the previously scheduled script file '{}'.", filename, _scheduled_script);
+	} else {
+		IConsolePrint(CC_INFO, "Script file '{}' scheduled to execute at the start of next month.", filename);
```
James103: Since the schedule option runs a script on the next calendar month, would it be helpful to replace "start of next month" with "start of next **calendar** month" in these lines?

michalc: Sounds good and done - although with "the", so "start of the next calendar month" - not exactly sure why, but it sounds less clunky to me.

michalc:
> Ah it is calendar month (I think!)... maybe the option should be "on-next-calendar-month"

Have done this

src/console_cmds.cpp
michalc: (Er... so I forgot to push. _Now_ done...)

michalc: Actually... I think since it's relative to _now_, maybe it shouldn't have the "the"...

michalc: Have changed it to "start of next calendar month" as suggested

src/console_cmds.cpp
```diff
@@ -1154,6 +1172,35 @@ DEF_CONSOLE_CMD(ConExec)
 	return true;
 }
 
+DEF_CONSOLE_CMD(ConSchedule)
+{
+	if (argc < 3 || std::string_view(argv[2]) != "on-next-calendar-month") {
```
TrueBrain: It might be better to have `on-next-calendar-month` first, as `<script>` can be lengthy. Often makes things a bit easier to read :)

What do you think?

src/console_cmds.cpp
```diff
@@ -53,6 +53,24 @@
 /* scriptfile handling */
 static uint _script_current_depth; ///< Depth of scripts running (used to abort execution when #ConReturn is encountered).
 
+/* Scheduled execution handling */
+static std::string _scheduled_script; ///< Script scheduled to execute by the 'schedule' console command (empty if no script is scheduled)
```
TrueBrain:
```suggestion
static std::string _scheduled_monthly_script; ///< Script scheduled to execute by the 'schedule' console command (empty if no script is scheduled)
```

Or something? I dunno .. but it might help to have it more clear. Or it might not :P

TrueBrain (commented):
It looks really nice :) Some very small nitpicks!

src/console_cmds.cpp
```diff
@@ -1154,6 +1172,35 @@ DEF_CONSOLE_CMD(ConExec)
...
 }
 
+DEF_CONSOLE_CMD(ConSchedule)
+{
+	if (argc < 3 || std::string_view(argv[2]) != "on-next-calendar-month") {
+		IConsolePrint(CC_HELP, "Schedule a local script to execute later. Usage: 'schedule <script> on-next-calendar-month'.");
+		return true;
+	}
+
+	/* Check if the file exists. It might still go away later, but helpful to show an error now */
```
TrueBrain:
```suggestion
	/* Check if the file exists. It might still go away later, but helpful to show an error now. */
```

VERY small nitpick. Hold true for all other comments too. Sentences end with a dot. :D

src/console_cmds.cpp
```diff
@@ -53,6 +53,24 @@
 /* scriptfile handling */
 static uint _script_current_depth; ///< Depth of scripts running (used to abort execution when #ConReturn is encountered).
 
+/* Scheduled execution handling. */
+static std::string _scheduled_monthly_script; ///< Script scheduled to execute by the 'schedule' console command (empty if no script is scheduled).
+
+/** Timer that runs every month of game time for the 'schedule' console command. */
+static IntervalTimer<TimerGameCalendar> _schedule_timer = {{TimerGameCalendar::MONTH, TimerGameCalendar::Priority::NONE}, [](auto) {
```
TrueBrain:
```suggestion
static IntervalTimer<TimerGameCalendar> _scheduled_monthly_timer = {{TimerGameCalendar::MONTH, TimerGameCalendar::Priority::NONE}, [](auto) {
```

Sorry, should have mentioned that: while at it, mind also renaming `_schedule_timer`? :)

src/console_cmds.cpp
michalc: Have to admit I don't know what would be best - both seem so reasonable to me! So have changed to have `on-next-calendar-month` first

src/console_cmds.cpp
michalc: Now changed

src/console_cmds.cpp
michalc: Have added the dots

(There is one comment that is maybe a title? "Scheduled execution handling"... it does have a dot but maybe it shouldn't since it's not really a sentence?)

src/console_cmds.cpp
michalc: All good - renamed this too

michalc:
@TrueBrain Thanks for the review and comments! I have hopefully addressed them all

src/console_cmds.cpp
TrueBrain: In general we just say: always add a dot, to remove the discussion whether it is a title or not. But it is all good. If we get 90% dots, I succeeded in my job :D

TrueBrain (approved):
Tnx a lot for this work! Looks really nice. I especially appreciate the clear comments in the code; made it easy to follow :)

PR #12767, opened 2024-06-08
An RFC / POC for adding more companies

JustLinuxUser:
## Motivation / Problem
I really liked the Spiffing Brit's video about massive multiplayer in OpenTTD, and I would really like for something like this to be possible in stock OpenTTD. That's why I decided to work on this.
The point of this PR is to change the max number of companies from 15 to 240

This PR is not finished. Rather this PR is here to ask about a potential of this ever getting merged, and what would be the requirements for a merge. 

## Description

I solved the issue of adding more companies by changing the tile_map, to add m9, and store all the owner information in there.
I also had to change constants like OWNER_NONE, OWNER_TOWN, etc, And I had to change the company bitfield to be a C++ bitfield, and add save/load support for it.

## Limitations / Known issues
I have been slowly finding and fixing bugs introduced by this change, but still not sure how to solve some problems.
- The color changing behavior does not work (the ui is just greyed out) when the number of companies is more then 15.
Possible solution: allow for changing a company color regardless if it's occupied
- The newgrf_town.cpp is broken, and I don't know how to fix it
- The cargo flow legend is broken, the ui is just too big, but putting a scroll bar or showing only existing companies + scrollbar should fix it
- This also deletes randomness of the color selection, so the colors are the same every time
- When creating a new server, the maximum number of digits you can put in the maximum companies field is 99, so to get to the current PR's maximum of 240 I need to click 141 times
- It wouldn't be too much harder to make this patch with ~ 1000 maximum companies, space in the m9 allows it. I would just need to change the Owner enum to be a 16 bit integer. But it would be harder to make this change later, because a new savegame version and some conversion logic would need to be introduced

## Checklist for review
- The pr needs a savegame upgrade
- The PR affects newgrf and brakes it, must be fixed (need help)


nielsmh:
I'm changing this to a Draft PR since, as I understand it, it's still in experimental stage.

Keep in mind that we'll be asking you to clean up the commit history and fix the commit messages to conform to the project standards, before possibly merging in the future.

PR #12770, opened 2024-06-10
Add: Toggle for showing industry names in small map

EdmondChuiHW:
# Motivation / Problem

* For features or gameplay changes:
    * What was the motivation to develop this feature?
      * Came across a fellow player who is visually impaired who'd like to see industry names in the small map similar to town names: [Reddit](https://web.archive.org/web/20240610032237/https://old.reddit.com/r/openttd/comments/1d9v3a9/im_visually_impaired_i_like_i_can_scale_up_the_ui/)
    * Does this address any problem with the gameplay or interface?
      * Interface only
    * Which group of players do you think would enjoy this feature?
      * All players, especially visually impaired like OP, or players with mild colour-blindness like myself

# Description

## What does this feature do?
A new "Show industry names" toggle button and a tooltip are added to the industry view in the small map:
screenshot of the toggle button and tooltip

When toggled on, industry names are drawn below each tile area of the industry:
screenshot of small map displayed with industry names

When an industry legend is selected, the names on the map follow the enabled/disabled statuses:
screenshot of small map displayed with some industry names, as toggled in the industries legend

When an industry legend is hovered over, each industry name blinks together with the industry tile area:
animated image showing blinking behaviour

When toggled off (the default), there's no difference to the behaviour today:
screenshot showing a small map without industry names when the new toggle is off

When switched to other views that isn't "Industries", the toggle and the industry names are not displayed (except the ones mentioned under Limitations):
screenshot showing a small map in 'Vehicles' view

## How does it improve/solve the situation described under 'motivation'?

> The main goals of the official branch are:
> […]
> Improve the user interface
> […]

# Limitations

* Is the problem solved in all scenarios?
  * Yes. Although when used together with "show town names", the text labels appears crowded with no visual distinction between industry and town names:
  * a screenshot showing the small map with both industry and town names
* Is this feature complete? Are there things that could be added in the future?
  * Maybe showing the specific name of each industry? e.g. "Slawood Factory", "Lothill Farm", etc. That might be a solution to showing industry and town names together.
* Are there things that are intentionally left out?
  * The text colour could potentially follow the industry colour; tho industries such as Forest in a temperate map would be hard to read as they'd show up as low-contrast green-on-green.
* Do you know of a bug or corner case that does not work?
  * Yes. I need help hiding the button in other views. Current implementation just disables the button in cargo flow and land owner views (and could appear 'stuck' in the lowered state under irrelevant views). Please see the inline comment in the code
  * screenshot showing a disabled 'show industry names' toggle button in land owner view
  * screenshot showing a disabled and lowered 'show industry names' toggle button in cargo flow view
  * Update: resolved. Button is now only shown in industry map type. See https://github.com/OpenTTD/OpenTTD/pull/12770#discussion_r1650212639

# Checklist for review

  * N/A
  * Yes. [I want to add a new string.](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md#i-want-to-add-a-new-string)
    > Add the new string somewhere in english.txt, where it fits with the neighbouring strings. Don't touch the translations, even if you can speak some of the languages.
  * No
  * No
  * No

### DIFF
--- a/src/smallmap_gui.cpp
+++ b/src/smallmap_gui.cpp
@@ -172,6 +172,8 @@ static LegendAndColour _legend_linkstats[NUM_CARGO + lengthof(_linkstat_colours_
 static LegendAndColour _legend_from_industries[NUM_INDUSTRYTYPES + 1];
 /** For connecting industry type to position in industries list(small map legend) */
 static uint _industry_to_list_pos[NUM_INDUSTRYTYPES];
+/** The string bounding box width for each industry type in the smallmap */
+static uint16_t _industry_to_name_string_width[NUM_INDUSTRYTYPES];
 /** Show heightmap in industry and owner mode of smallmap window. */
 static bool _smallmap_show_heightmap = false;
 /** Highlight a specific industry type */
@@ -635,6 +637,7 @@ protected:
 
 	static SmallMapType map_type; ///< Currently displayed legends.
 	static bool show_towns;       ///< Display town names in the smallmap.
+	static bool show_ind_names;   ///< Display industry names in the smallmap.
 	static int map_height_limit;  ///< Currently used/cached map height limit.
 
 	static const uint INDUSTRY_MIN_NUMBER_OF_COLUMNS = 2; ///< Minimal number of columns in the #WID_SM_LEGEND widget for the #SMT_INDUSTRY legend.
@@ -829,7 +832,7 @@ protected:
 
 		if (map_type == SMT_LINKSTATS) this->overlay->SetDirty();
 		if (map_type != SMT_INDUSTRY) this->BreakIndustryChainLink();
-		this->SetDirty();
+		this->ReInit();
 	}
 
 	/**
@@ -984,13 +987,13 @@ protected:
 	 * Adds town names to the smallmap.
 	 * @param dpi the part of the smallmap to be drawn into
 	 */
-	void DrawTowns(const DrawPixelInfo *dpi) const
+	void DrawTowns(const DrawPixelInfo *dpi, const int vertical_padding) const
 	{
 		for (const Town *t : Town::Iterate()) {
 			/* Remap the town coordinate */
 			Point pt = this->RemapTile(TileX(t->xy), TileY(t->xy));
 			int x = pt.x - this->subscroll - (t->cache.sign.width_small >> 1);
-			int y = pt.y;
+			int y = pt.y + vertical_padding;
 
 			/* Check if the town sign is within bounds */
 			if (x + t->cache.sign.width_small > dpi->left &&
@@ -1000,6 +1003,45 @@ protected:
 				/* And draw it. */
 				SetDParam(0, t->index);
 				DrawString(x, x + t->cache.sign.width_small, y, STR_SMALLMAP_TOWN);
+			}
+		}
+	}
+
+	/**
+	 * Adds industry names to the smallmap.
+	 * @param dpi the part of the smallmap to be drawn into
+	 */
+	void DrawIndustryNames(const DrawPixelInfo *dpi, const int vertical_padding) const
+	{
+		if (this->map_type != SMT_INDUSTRY) return;
+
+		for (const Industry *i : Industry::Iterate()) {
+			const LegendAndColour &tbl = _legend_from_industries[_industry_to_list_pos[i->type]];
+			if (!tbl.show_on_map) continue;
+
+			/* Industry names blink together with their blobs in the smallmap. */
+			const bool is_blinking = i->type == _smallmap_industry_highlight && !_smallmap_industry_highlight_state;
+			if (is_blinking) continue;
+
+			if (_industry_to_name_string_width[i->type] == 0) {
+				_industry_to_name_string_width[i->type] = GetStringBoundingBox(tbl.legend, FS_SMALL).width;
+			}
+			const uint16_t &legend_text_width = _industry_to_name_string_width[i->type];
+
+			/* Remap the industry coordinate */
+			const TileIndex &tile = i->location.GetCenterTile();
+			const Point pt = this->RemapTile(TileX(tile), TileY(tile));
+			const int x = pt.x - this->subscroll - (legend_text_width / 2);
+			const int y = pt.y + vertical_padding;
+
+			/* Check if the industry name is within bounds */
+			if (x + legend_text_width > dpi->left &&
+					x < dpi->left + dpi->width &&
+					y + GetCharacterHeight(FS_SMALL) > dpi->top &&
+					y < dpi->top + dpi->height) {
+
+				/* And draw it. */
+				DrawString(x, x + legend_text_width, y, tbl.legend, TC_WHITE, SA_LEFT, false, FS_SMALL);
 			}
 		}
 	}
@@ -1064,8 +1106,13 @@ protected:
 		/* Draw link stat overlay */
 		if (this->map_type == SMT_LINKSTATS) this->overlay->Draw(dpi);
 
+		const int map_labels_vertical_padding = ScaleGUITrad(2);
+
 		/* Draw town names */
-		if (this->show_towns) this->DrawTowns(dpi);
+		if (this->show_towns) this->DrawTowns(dpi, map_labels_vertical_padding);
+
+		/* Draw industry names */
+		if (this->show_ind_names) this->DrawIndustryNames(dpi, map_labels_vertical_padding);
 
 		/* Draw map indicators */
 		this->DrawMapIndicators();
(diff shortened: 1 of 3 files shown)


src/smallmap_gui.cpp
```diff
@@ -1004,6 +1005,43 @@ class SmallMapWindow : public Window {
 		}
 	}
 
+	/**
+	 * Adds industry names to the smallmap.
+	 * @param dpi the part of the smallmap to be drawn into
+	 */
+	void DrawIndustryNames(const DrawPixelInfo *dpi) const
```
EdmondChuiHW: Structure follows the existing `DrawTowns(…)` method:

https://github.com/OpenTTD/OpenTTD/blob/e35680d0cbf8fe8e7c4b55729a45e9b0257f09fa/src/smallmap_gui.cpp#L988-L1006

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1005,43 @@ class SmallMapWindow : public Window {
...
+
+			/* Check if the industry name is within bounds */
+			if (x + legend_text_width > dpi->left &&
+					x < dpi->left + dpi->width &&
+					y + GetCharacterHeight(FS_SMALL) > dpi->top &&
+					y < dpi->top + dpi->height) {
+
+				/* And draw it. */
+				SetDParam(0, i->index);
+				DrawString(x, x + legend_text_width, y, tbl.legend, TC_WHITE, SA_LEFT, false, FS_SMALL);
```
EdmondChuiHW, line 1044: Not sure if I'm using `DrawString(…)` correctly here.

The same call in `DrawTowns(…)` doesn't pass `FS_SMALL` or `TC_WHITE`; but if I omit those params here, texts are drawn with normal size in blue

Edit: deleted the unused `SetDParam`

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1005,43 @@ class SmallMapWindow : public Window {
...
+		if (this->map_type != SMT_INDUSTRY) return;
+
+		for (const Industry *i : Industry::Iterate()) {
+			const LegendAndColour &tbl = _legend_from_industries[_industry_to_list_pos[i->type]];
+			if (!tbl.show_on_map) continue;
+
+			const bool is_blinking = i->type == _smallmap_industry_highlight && !_smallmap_industry_highlight_state;
+			if (is_blinking) continue;
+
+			const uint8_t legend_text_width = GetStringBoundingBox(tbl.legend, FS_SMALL).width;
```
EdmondChuiHW: Should I cache this somewhere? e.g. `_industry_to_legend_text_width` similar to `_industry_to_list_pos`:

https://github.com/OpenTTD/OpenTTD/blob/e35680d0cbf8fe8e7c4b55729a45e9b0257f09fa/src/smallmap_gui.cpp#L173-L174

Edit: added a cache with lazy calculation

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1005,43 @@ class SmallMapWindow : public Window {
...
+	 * Adds industry names to the smallmap.
+	 * @param dpi the part of the smallmap to be drawn into
+	 */
+	void DrawIndustryNames(const DrawPixelInfo *dpi) const
+	{
+		if (this->map_type != SMT_INDUSTRY) return;
+
+		for (const Industry *i : Industry::Iterate()) {
+			const LegendAndColour &tbl = _legend_from_industries[_industry_to_list_pos[i->type]];
+			if (!tbl.show_on_map) continue;
```
EdmondChuiHW, line 1020: Bit inefficient to iterate thru all industries before checking the legend `show_on_map` flag first.

However, I couldn't find something like this: `Industry::IterateByType(IndustryType)`

```
for (int i = 0; i != _smallmap_industry_count; i++) {
	const LegendAndColour &tbl = _legend_from_industries[i];
	for (const Industry *ind : Industry::IterateByType(tbl.type)) {
		…
	}
}
```

Edit: existing impl uses `::Iterate()` for `Company`, `Town`, and `Vehicle` so I'm assuming it's fine to do for `Industry` too

src/smallmap_gui.cpp
```diff
@@ -1225,40 +1265,46 @@ class SmallMapWindow : public Window {
...
 				disable_all_tooltip = STR_NULL;
+				disable_industry_names = true;
 				plane = 1;
 				break;
 		}
 
 		this->GetWidget<NWidgetCore>(WID_SM_LEGEND)->SetDataTip(STR_NULL, legend_tooltip);
 		this->GetWidget<NWidgetCore>(WID_SM_ENABLE_ALL)->SetDataTip(STR_SMALLMAP_ENABLE_ALL, enable_all_tooltip);
 		this->GetWidget<NWidgetCore>(WID_SM_DISABLE_ALL)->SetDataTip(STR_SMALLMAP_DISABLE_ALL, disable_all_tooltip);
+		this->GetWidget<NWidgetCore>(WID_SM_SHOW_IND_NAMES)->SetDisabled(disable_industry_names);
```
EdmondChuiHW: As mentioned under Limitations, I believe the button should only be shown with `SMT_INDUSTRY` map type (not just disabled).

From the existing code, it looks like a `NWidgetStacked` container is the typical way to show/hide something.

I tried wrapping the toggle with a `NWID_SELECTION`, but it looks like `NWID_SELECTION` is not supposed to be stacked (crashed):

```
NWidget(NWID_SELECTION, INVALID_COLOUR, WID_SM_SHOW_IND_NAMES_CONTAINER),
  NWidget(WWT_TEXTBTN, COLOUR_BROWN, WID_SM_SHOW_IND_NAMES), SetDataTip(STR_SMALLMAP_SHOW_INDUSTRY_NAMES, STR_SMALLMAP_TOOLTIP_SHOW_INDUSTRY_NAMES),
EndContainer(),
```

I tried using `NWID_HORIZONTAL` too but other usages don't use a `WID_` with it.

What would be the best way to show/hide the toggle button?

src/smallmap_gui.cpp
```diff
@@ -1685,6 +1732,14 @@ class SmallMapWindow : public Window {
...
 				break;
 
+			case WID_SM_SHOW_IND_NAMES: // Toggle industry names
+				this->show_industry_names = !this->show_industry_names;
+				this->SetWidgetLoweredState(WID_SM_SHOW_IND_NAMES, this->show_industry_names);
+
+				this->SetDirty();
+				if (_settings_client.sound.click_beep) SndPlayFx(SND_15_BEEP);
+				break;
+
```
EdmondChuiHW: Follows the existing `WID_SM_TOGGLETOWNNAME`:

https://github.com/OpenTTD/OpenTTD/blob/e35680d0cbf8fe8e7c4b55729a45e9b0257f09fa/src/smallmap_gui.cpp#L1727-L1733

EdmondChuiHW:
Hi team! First time contributing. I followed the guidelines as best as I could. Please advise if I missed anything, especially around code style. Thanks!

P.S. The preview link GitHub Action is super cool!!

src/widgets/smallmap_widget.h
```diff
@@ -32,6 +32,7 @@ enum SmallMapWidgets : WidgetID {
 	WID_SM_ENABLE_ALL,     ///< Button to enable display of all legend entries.
 	WID_SM_DISABLE_ALL,    ///< Button to disable display of all legend entries.
 	WID_SM_SHOW_HEIGHT,    ///< Show heightmap toggle button.
+	WID_SM_SHOW_IND_NAMES, ///< Show indsutry names toggle button.
```
2TallTyler:
```suggestion
	WID_SM_SHOW_IND_NAMES, ///< Show industry names toggle button.
```

src/smallmap_gui.cpp
2TallTyler: I agree that the button should be hidden in other map types. You are on the right track despite the crash: see `WID_BV_BUILD_SEL` in `build_vehicle_gui.cpp`. 🙂

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1009,45 @@ class SmallMapWindow : public Window {
...
+			if (!tbl.show_on_map) continue;
+
+			const bool is_blinking = i->type == _smallmap_industry_highlight && !_smallmap_industry_highlight_state;
+			if (is_blinking) continue;
+
+			if (_industry_to_name_string_width[i->type] == 0) {
+				_industry_to_name_string_width[i->type] = GetStringBoundingBox(tbl.legend, FS_SMALL).width;
+			}
+			const uint8_t &legend_text_width = _industry_to_name_string_width[i->type];
+			const int vertical_padding = ScaleGUITrad(4);
```
2TallTyler: Why not use the same vertical positioning as town names?

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1009,45 @@ class SmallMapWindow : public Window {
...
+	 */
+	void DrawIndustryNames(const DrawPixelInfo *dpi) const
+	{
+		if (this->map_type != SMT_INDUSTRY) return;
+
+		for (const Industry *i : Industry::Iterate()) {
+			const LegendAndColour &tbl = _legend_from_industries[_industry_to_list_pos[i->type]];
+			if (!tbl.show_on_map) continue;
+
+			const bool is_blinking = i->type == _smallmap_industry_highlight && !_smallmap_industry_highlight_state;
```
2TallTyler, line 1023: Might be nice to have a comment here, `/* Industry names blink with their blob on the smallmap. */` or similar.

src/smallmap_gui.cpp
```diff
@@ -829,7 +834,7 @@ class SmallMapWindow : public Window {
 
 		if (map_type == SMT_LINKSTATS) this->overlay->SetDirty();
 		if (map_type != SMT_INDUSTRY) this->BreakIndustryChainLink();
-		this->SetDirty();
+		this->ReInit();
```
EdmondChuiHW, line 835: @2TallTyler Thanks for the review and the pointer to `build_vehicle_gui.cpp`!

Glad to know the `NWID_SELECTION` and `SZSP_NONE` combo is the standard ways to show/hide things.

I initially tried that route after looking at your PR for showing/hiding block signals #11897. But ran into a weird rendering issue where the now-void space isn't resized properly:

screenshot showing an empty space rendering incorrectly

So I dug around other call sites and eventually found that I need to use `ReInit()` when using `StackedZeroSizePlanes`:

https://github.com/OpenTTD/OpenTTD/blob/e802c1a97fac5bfad5ff460ea8dfd37d24571602/src/widget_type.h#L495-L498

Let me know if `ReInit` should be avoided here, or if there's a more scoped-down method like `ResizeWindow` (tho that's private).

src/smallmap_gui.cpp
EdmondChuiHW: Town names use zero padding and it looks OK since we don't really need to see the exact roads/boundaries of a town. For industries, zero padding feels a bit too crowded to make out the colour of the blobs:

# Industry names
screenshot of industry names with zero padding

# Industry and town names
screenshot of industry and town names with zero padding

So I discretionarily picked `4`, which looks good on my screen (and the preview too imo).

Let me know if there's a design system/philosophy for padding.

Alternatively, we could match paddings for both town and industry names for consistency. Here's how it looks if we use `2` for both:

# Town names
screenshot of town names with 2 padding

# Industry names
screenshot of industry names with 2 padding

# Industry and town names
screenshot of industry and town names with zero padding

Update: I quite like `2` for both. Latest commit [f15ba9b](https://github.com/OpenTTD/OpenTTD/compare/8bde7dd3ac224e5e92e2a3a32cc89da50bc12241..f15ba9b159e7ac247df75d2dfa354f40963d2dae) now has this

src/widgets/smallmap_widget.h
```diff
@@ -32,6 +32,8 @@ enum SmallMapWidgets : WidgetID {
 	WID_SM_ENABLE_ALL,     ///< Button to enable display of all legend entries.
 	WID_SM_DISABLE_ALL,    ///< Button to disable display of all legend entries.
 	WID_SM_SHOW_HEIGHT,    ///< Show heightmap toggle button.
+	WID_SM_SHOW_IND_NAMES, ///< Show industry names toggle button.
+	WID_SM_SHOW_IND_NAMES_SEL, ///< Container for the 'show industry names' button, which can be hidden.
```
EdmondChuiHW, line 36: Adding a single space instead of shifting the alignment of other comments (in compliance of #12780)

src/smallmap_gui.cpp
2TallTyler: If `ReInit` is used elsewhere, it's good enough for here! I don't imagine it will be clicked very often, so it's cheap. 🙂

2TallTyler (dismissed):

2TallTyler:
The next step for this PR is to wait for another developer to review and merge, or request changes that I missed. You do not need to keep it updated, unless there is a merge conflict. 🙂

EdmondChuiHW:
> The next step for this PR is to wait for another developer to review and merge, or request changes that I missed. You do not need to keep it updated, unless there is a merge conflict. 🙂 

Thank you for the clarification! Thought I needed to/could rebase and merge myself. Sorry for any email spam

src/smallmap_gui.cpp
```diff
@@ -1004,6 +1009,45 @@ class SmallMapWindow : public Window {
...
+			if (!tbl.show_on_map) continue;
+
+			/* Industry names blink together with their blobs in the smallmap. */
+			const bool is_blinking = i->type == _smallmap_industry_highlight && !_smallmap_industry_highlight_state;
+			if (is_blinking) continue;
+
+			if (_industry_to_name_string_width[i->type] == 0) {
+				_industry_to_name_string_width[i->type] = GetStringBoundingBox(tbl.legend, FS_SMALL).width;
+			}
+			const uint8_t &legend_text_width = _industry_to_name_string_width[i->type];
```
rubidium42: Isn't 255 pixels width a bit too small? If the user selects 5x interface size, that would only be 51 pixels in the normal size. That seems quite narrow for an industry name.

EdmondChuiHW: @rubidium42 Good catch! The default industries come in just under 255 at 5x:

screenshot of small map showing industry names at 5x interface scale

screenshot of small map showing industry name label widths at 5x interface scale

1x for comparison:

screenshot of small map showing industry names at 1x interface scale

screenshot of small map showing industry name label widths at 1x interface scale

This is indeed too tight for other custom industries with longer names. Updated from `uint8_t` to `uint16_t` now.

While investigating this, I found a bug where the cached widths are not updated when changing the interface scale. Updated to reset the cache in `OnInit()` now.

Since we now call `ReInit()` each time we switch the map type https://github.com/OpenTTD/OpenTTD/pull/12770#discussion_r1650212639, this does mean the cached widths are recalculated each time the industry view is selected. The cache would only be used while toggling individual industry legends.

The cache-busting could be more precise by adding a "zoom changed" virtual method from `Window`. However, the added complexity of maintaining a new method/API in the common `Window` class is not worth it for this single use case imo.

https://github.com/OpenTTD/OpenTTD/blob/9936b3728abdde9e8bfc792767753f112981b4d4/src/window.cpp#L3304-L3312

2TallTyler (approved):

PR #12772, opened 2024-06-11
Codechange: Use function-static instance of WindowDesc list.

PeterN:
## Motivation / Problem

The WindowDesc list `_window_descs` is a pointer to a vector with manual initialisation with `new` on first use, as long as it's used correctly.

This was done to ensure correct static initialisation order.

## Description

Instead, use a function-static instance. This can only be accessed by calling the `GetWindowDescs()`, which automatically initialises the vector on first use.

This ensures the list cannot be accessed incorrectly and avoids manual life-time management.

## Checklist for review

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -93,10 +93,14 @@ bool _mouse_hovering;      ///< The mouse is hovering over the same point.
 SpecialMouseMode _special_mouse_mode; ///< Mode of the mouse.
 
 /**
- * List of all WindowDescs.
- * This is a pointer to ensure initialisation order with the various static WindowDesc instances.
+ * Get list of all WindowDescs.
+ * @return List of all WindowDescs.
  */
-std::vector<WindowDesc*> *_window_descs = nullptr;
+std::vector<WindowDesc *> &GetWindowDescs()
+{
+	static std::vector<WindowDesc *> *s_window_descs = new std::vector<WindowDesc *>();
+	return *s_window_descs;
+}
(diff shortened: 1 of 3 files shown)


PeterN:
Bah, this solves the SIOF, but leaves de-initialisation undefined.

src/window.cpp
```diff
@@ -93,10 +93,14 @@ bool _mouse_hovering;      ///< The mouse is hovering over the same point.
...
 /**
- * List of all WindowDescs.
- * This is a pointer to ensure initialisation order with the various static WindowDesc instances.
+ * Get list of all WindowDescs.
+ * @return List of all WindowDescs.
  */
-std::vector<WindowDesc*> *_window_descs = nullptr;
+std::vector<WindowDesc *> &GetWindowDescs()
+{
+	static std::vector<WindowDesc *> *s_window_descs = new std::vector<WindowDesc *>();
```
rubidium42, line 101: Why do we allocate the vector at all? Can't it be just `static std::vector<WindowDesc *> s_window_descs;`?
Or is that due to the SIOF thing? But static local variables have a pretty well-defined behaviour, so I would reckon it is not be affected by SIOF.

PeterN: That fixes SIOF, but as far as I know not de-initialisation order.

LordAro: Maybe worth adding a comment to clarify that's why it's a pointer?

PR #12777, opened 2024-06-11
Update cmake/Endian.cmake to work with newer versions of CMake (>=3.2…

### LINKED ISSUE #12732: [Bug]: /cmake/Endian.cmake uses deprecated endiannes test.
### Version of OpenTTD

commit 51e170a

### Expected result

The test should be successfull. 
Use `CMAKE_<LANG>_BYTE_ORDER` instead: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_BYTE_ORDER.html#variable:CMAKE_%3CLANG%3E_BYTE_ORDER

### Actual result

Function `add_endian_definition` in file 

> /cmake/Endian.cmake

 doesn't work properly with CMake version 3.28.0 and causes compilation to fail, probably due to use of deprecated endianness test: https://cmake.org/cmake/help/latest/module/TestBigEndian.html

### Steps to reproduce

1. Compile in Visual Studio 17,9

catlukemich:
Fixes #12732

### DIFF
--- a/cmake/Endian.cmake
+++ b/cmake/Endian.cmake
@@ -14,7 +14,7 @@ function(add_endian_definition)
         endif()
 
     else()
-        if(CMAKE_<LANG>_BYTE_ORDER EQUAL BIG_ENDIAN)
+        if(CMAKE_CXX_BYTE_ORDER EQUAL BIG_ENDIAN)
             add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
         else()
             add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
--- a/cmake/Endian.cmake
+++ b/cmake/Endian.cmake
@@ -14,7 +14,7 @@ function(add_endian_definition)
         endif()
 
     else()
-        if(CMAKE_CXX_BYTE_ORDER EQUAL BIG_ENDIAN)
+        if(CMAKE_CXX_BYTE_ORDER EQUAL BIG_ENDIAN OR CMAKE_OBJCXX_BYTE_ORDER EQUAL BIG_ENDIAN)
             add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
         else()
             add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
--- a/cmake/Endian.cmake
+++ b/cmake/Endian.cmake
@@ -3,12 +3,23 @@
 # add_endian_definition()
 #
 function(add_endian_definition)
-    include(TestBigEndian)
-    TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
+    
+    if(CMAKE_VERSION VERSION_LESS "3.20.0") 
+        include(TestBigEndian)
+        TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
+        if(IS_BIG_ENDIAN)
+            add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
+        else()
+            add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
+        endif()
 
-    if(IS_BIG_ENDIAN)
-        add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
     else()
-        add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
+        if(CMAKE_<LANG>_BYTE_ORDER EQUAL BIG_ENDIAN)
+            add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
+        else()
+            add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
+        endif()
+
     endif()
+
 endfunction()


cmake/Endian.cmake
```diff
@@ -3,12 +3,23 @@
...
+            add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
+        else()
+            add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
+        endif()
 
-    if(IS_BIG_ENDIAN)
-        add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
     else()
-        add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
+        if(CMAKE_<LANG>_BYTE_ORDER EQUAL BIG_ENDIAN)
```
LordAro: yeah I don't think that's going to work

LordAro:
Please don't blank the PR template, it's there for a reason

cmake/Endian.cmake
catlukemich: Works for me.

LordAro: I promise it doesn't. `<LANG>` is a placeholder in the documentation.

https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_BYTE_ORDER.html

glx22: `CMAKE_<LANG>_BYTE_ORDER` is a generic placeholder for documentation, you want to use `CMAKE_CXX_BYTE_ORDER` (and `CMAKE_OBJCXX_BYTE_ORDER` for .mm files)

PeterN: > Works for me.

If it 'works' for you it's because you do not have a BIG_ENDIAN machine, so it doesn't matter if the condition is always false.

rubidium42:
With cmake 3.29.2 on PowerPC G3:
```
grep -R ENDIAN build-*/src/strgen/       
build-master/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_BIG_ENDIAN -DUNIX
build-pr12777/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_LITTLE_ENDIAN -DUNIX
```

So sadly this change does not work for big endian platforms.

lukasz1985:
> With cmake 3.29.2 on PowerPC G3:
> 
> ```
> grep -R ENDIAN build-*/src/strgen/       
> build-master/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_BIG_ENDIAN -DUNIX
> build-pr12777/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_LITTLE_ENDIAN -DUNIX
> ```
> 
> So sadly this change does not work for big endian platforms.

I'm not aware of every single platform the OpenTTD is supposed to be compliant with. This at least (in my best hope) fixes the issue related to building under Visual Studio, as mentioned in bug #12732. The conditional statements used in commit makes this PR not touch the before established endianness test implementation. If you ask me: it's better than nothing and if the solution works for that certain issue it can be called a fix. It's not up to me if the PR is accepted or not though did my best to spare some other folks the pain of debugging CMake scripts.

rubidium42:
> If you ask me: it's better than nothing and if the solution works for that certain issue it can be called a fix.

No, it's not better than nothing. Your PR suggests that it implements Endian checks (with CMake >= 3.20), just in a different way to before. However, the first attempt I ran with a big endian platform did say "it's little endian".

I know an even better solution to your problem. Just replace the whole content of the `add_endian_definition` function with `add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)`. It will work with your MSVC, and it will be equally broken for big endian platforms but the implementation will not pretend that it supports big endian platforms.


What about documenting that you require CMake version 3.16 (inclusive) up to 3.20 (exclusive). It solves the deprecation issue, so it's an equally valid fix for the problem. 

Don't you think? If you don't think it's a valid fix, in what way is it different? As this PR already implicitly adds this requirement for big endian platforms (because it does not work for CMake >= 3.20), so it would be much easier to keep the version requirements the same over all platforms than to differentiate them.


What could help you with debugging this, is to "invert" the logic so big endian is the else branch:
```
    if(CMAKE_CXX_BYTE_ORDER EQUAL LITTLE_ENDIAN OR CMAKE_OBJCXX_BYTE_ORDER EQUAL LITTLE_ENDIAN)
        add_definitions(-DTTD_ENDIAN=TTD_LITTLE_ENDIAN)
    else()
        add_definitions(-DTTD_ENDIAN=TTD_BIG_ENDIAN)
    endif()
```
With this it becomes more obvious that the condition is broken.

lukasz1985:
> No, it's not better than nothing. Your PR suggests that it implements Endian checks (with CMake >= 3.20), just in a different way to before. However, the first attempt I ran with a big endian platform did say "it's little endian".

I will still argue with that. Solving a problem for some cases is better than not solving it at all. It may be irrelevant if you target for perfection at first attempt, but then consider this: if you had gallstones would you like to suffer (or face the death, should I say directly) from the complications, or get you gallbladder removed and live further down the road life with limited consumption of fats and alcohol, but still live? That was a drastic example but same goes with software - we fix bugs to our best knowledge and put efforts to make things better over time.

Speaking of which - I think, there shouldn't be a replacement as your suggestion stands for. I think you're suggesting a degradation of code (not the functionality though). I'm not sure how the endian check works on different machines  - I'm making assumptions that CMake itself and the underlying functions called from`add_endian_definition` don't lie and focus on a single point of the system (at least at the moment).
For future contributions I suggest it stays as it is or strictly speaking - retain the abilty to perform the checks. I guess it should be somewhere mentioned that the endian check doesn't work as supposed to (maybe in another issue, maybe in the code within a commnent, maybe both.. maybe even somewhere else for new contributors convenience), but the code is ready for the upcoming solutions. 
This of course with the assumption that the OpenTTD developers policy is to keep the compatibilty with big endian machines. Otherwise the best solution would be to just remove the function in it's whole, because, why pretend anything at all? Of course, this would involve refactoring code elsewhere that's dependent on this function.

> What about documenting that you require CMake version 3.16 (inclusive) up to 3.20 (exclusive). It solves the deprecation issue, so it's an equally valid fix for the problem.

And, if I understood you correctly, and you want to restrict the build system to the versions you mentioned - this excludes the newer versions of CMake (and older too) - it will complicate things for newcomers (mainly, I guess) such as me. 

And one more other thing - perhaps the issue lies here and we're digging in a one big misunderstanding:

You "spit" (sorry for the phrase ;) this out before:
```
grep -R ENDIAN build-*/src/strgen/       
build-master/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_BIG_ENDIAN -DUNIX
build-pr12777/src/strgen/CMakeFiles/strgen.dir/flags.make:CXX_DEFINES = -DSTRGEN -DTTD_ENDIAN=TTD_LITTLE_ENDIAN -DUNIX
```

Are you sure that's from a build involving the latest commit of this PR? Because the first one(commit) was a complete misunderstanding too.

LordAro:
Alright, I think we're done here. Issue remains open.

lukasz1985:
Alright, I think we're done here. Issue remains open.> ere. Issue rema

That's how much you care about the community and potential contributors. 

Have a nice day.

ahyangyi:
Hey, I noticed that other projects use constructs such as
```
CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN"
```
instead of 
```
CMAKE_CXX_BYTE_ORDER EQUAL BIG_ENDIAN
```

And a quick Google gives this explanation:
> To compare strings use STREQUAL. EQUAL is for numbers.

Perhaps this is worth another try after a quick fix?

PR #12778, opened 2024-06-11
Codechange: Use std::endian instead of TTD_ENDIAN defines.

PeterN:
## Motivation / Problem

Tricks and defines to work out what the endianness is.

## Description

Use constexpr `std::endian::native` comparisons instead.

`union Colour` is a bit scary as it's not quite so simple to shuffle members around. I went for three separate implementations with the correct one aliased with `using Colour =`...

This removes the need for CMake endian testing and our own TTD_ENDIAN defines.

## Limitations

Probably needs testing to ensure it does do the the right thing on various little-and big-endian platforms.

## Checklist for review

### DIFF
--- a/src/screenshot.cpp
+++ b/src/screenshot.cpp
@@ -71,7 +71,7 @@ struct ScreenshotFormat {
 	ScreenshotHandlerProc *proc; ///< Function for writing the screenshot.
 };
 
-#define MKCOLOUR(x)         TO_LE32X(x)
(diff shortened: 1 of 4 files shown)


PeterN:
I've removed the awkward non-trivial bits for this PR, so we can at least make a start on deprecating TTD_ENDIAN.

rubidium42 (approved):

PR #12781, opened 2024-06-11
Codechange: Compiler check for _fios_colours/DetailFileType

ladysadie:
## Motivation / Problem
There's no static enforcement of the DetailedFileType enum and the fios_colors array despite the usage of the enum values as array indices for the fios_colors array.

## Description

Add a static assert to ensure the enum and array are kept in sync.

## Checklist for review

### DIFF
--- a/src/fileio_type.h
+++ b/src/fileio_type.h
@@ -40,6 +40,8 @@ enum DetailedFileType {
 	DFT_FIOS_DIR,    ///< A directory entry.
 	DFT_FIOS_DIRECT, ///< Direct filename.
 
+	DFT_END,         ///< End of this enum. Supports a compile time size check against _fios_colours in fios_gui.cpp
+
 	DFT_INVALID = 255, ///< Unknown or invalid file.
 };
 
--- a/src/fios_gui.cpp
+++ b/src/fios_gui.cpp
@@ -238,7 +238,8 @@ static const TextColour _fios_colours[] = {
 	TC_DARK_GREEN,   // DFT_FIOS_DIR
 	TC_ORANGE,       // DFT_FIOS_DIRECT
 };
-
+/* This should align with the DetailedFileType enum defined in fileio_type.h */
+static_assert(std::size(_fios_colours) == DFT_END);
 
 /**
  * Sort the collected list save games prior to displaying it in the save/load gui.


src/fileio_type.h
```diff
@@ -40,6 +40,9 @@ enum DetailedFileType {
 	DFT_FIOS_DIR,    ///< A directory entry.
 	DFT_FIOS_DIRECT, ///< Direct filename.
 
+	/* Ending variable used for compile time check of this array against _fios_colours in fiose_gui.cpp */
+	DFT_END,
```
2TallTyler: This sort of specific comment seems like an invitation to be outdated someday. I'd suggest something more generic, formatted as an inline Doxygen comment.

```suggestion
	DFT_END, ///< End-of-array marker.
```

2TallTyler:
The `Add:` commit prefix is for player-facing features. This would be a `Codechange:`, I think.

src/fios_gui.cpp
```diff
@@ -238,7 +238,8 @@ static const TextColour _fios_colours[] = {
 	TC_DARK_GREEN,   // DFT_FIOS_DIR
 	TC_ORANGE,       // DFT_FIOS_DIRECT
 };
-
+/* This should align with the DetailedFileType enum defined in fileio_type.h */
+static_assert (lengthof(_fios_colours) == DFT_END);
```
rubidium42:
```suggestion
static_assert(std::size(_fios_colours) == DFT_END);
```
There shouldn't be a space, and we're working out phasing out the custom `lengthof` macro over the standard `std::size` function.

src/fios_gui.cpp
```diff
@@ -238,7 +238,8 @@ static const TextColour _fios_colours[] = {
 	TC_DARK_GREEN,   // DFT_FIOS_DIR
 	TC_ORANGE,       // DFT_FIOS_DIRECT
 };
-
+/* This should align with the DetailedFileType enum defined in fileio_type.h */
+static_assert (std::size(_fios_colours) == DFT_END);
```
2TallTyler:
```suggestion
static_assert(std::size(_fios_colours) == DFT_END);
```

src/fileio_type.h
```diff
@@ -40,6 +40,8 @@ enum DetailedFileType {
 	DFT_FIOS_DIR,    ///< A directory entry.
 	DFT_FIOS_DIRECT, ///< Direct filename.
 
+	DFT_END,         ///< End of this enum. Supports a compile time size check against _fios_colours in fios_gui.cpp
```
2TallTyler, line 43:
```suggestion
	DFT_END,         ///< End of this enum.
```

src/fios_gui.cpp
ladysadie: Sigh. How did I miss that a second time?

src/fileio_type.h
ladysadie: "End of this enum."  isn't a useful comment.  It's an enum called DFT_END. The value's name is self descriptive.

The longer form comment I wrote explains what the value is for. Why it exists.

2TallTyler (approved):

PR #12785, opened 2024-06-14
Codechange: Split MakeNWidget to improve readability.

PeterN:
## Motivation / Problem

In #12779, CodeQL complained that MakeNWidget does not have enough comments. I reviewed the function and determined that comments wouldn't really help.

But I determined that the large switch block is actually handling two separate situations. For the first turn through the loop, it is creating an NWidget object, and for the remaining goes it applies NWIdgetPart attributes to that NWidget object. If it encounters a non-attributte NWidgetPart it will bail out, to be called again.

## Description

To make this clearer, I've split MakeNWidget() into two stages, widget-creation and attribute-applying.

The first thing MakeNWidget needs to do is to create the NWidget. This is handled by a new function that does only this, which is simpler as a result.

Then, like before, we loop through remaining NWidgetPart attributes until it until we reach a non-attribute NWidgetPart. This loop passes each attribute NWidgetPart to a function to process, which only handles these attributes,

Some WidgetTypes are shuffled around to make the new IsAttributeWidgetPartType() function simpler.

This split in NWidget process creation should make it easier to understand, read and extend, without just adding comments that probably wouldn't really help.

## Checklist for review

### DIFF
--- a/src/widget.cpp
+++ b/src/widget.cpp
@@ -2957,192 +2957,189 @@ bool NWidgetLeaf::ButtonHit(const Point &pt)
 /* == Conversion code from NWidgetPart array to NWidgetBase* tree == */
 
 /**
+ * Test if (an NWidgetPart) WidgetType is an attribute widget part type.
+ * @param tp WidgetType to test.
+ * @return True iff WidgetType is an attribute widget.
+ */
+static bool IsAttributeWidgetPartType(WidgetType tp)
+{
+	return tp > WPT_ATTRIBUTE_BEGIN && tp < WPT_ATTRIBUTE_END;
+}
+
+/**
+ * Apply an attribute NWidgetPart to an NWidget.
+ * @param nwid Attribute NWidgetPart
+ * @param dest NWidget to apply attribute to.
+ * @pre NWidgetPart must be an attribute NWidgetPart.
+ */
+static void ApplyNWidgetPartAttribute(const NWidgetPart &nwid, NWidgetBase *dest)
+{
+	switch (nwid.type) {
+		case WPT_RESIZE: {
+			NWidgetResizeBase *nwrb = dynamic_cast<NWidgetResizeBase *>(dest);
+			if (nwrb == nullptr) [[unlikely]] throw std::runtime_error("WPT_RESIZE requires NWidgetResizeBase");
+			assert(nwid.u.xy.x >= 0 && nwid.u.xy.y >= 0);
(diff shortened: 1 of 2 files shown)


src/widget_type.h
```diff
@@ -87,6 +87,7 @@ enum WidgetType {
 	NWID_CUSTOM,          ///< General Custom widget.
 
 	/* Nested widget part types. */
+	WPT_ATTRIBUTE_BEGIN, ///< Begin marked for attribute NWidgetPart types.
```
2TallTyler:
```suggestion
	WPT_ATTRIBUTE_BEGIN, ///< Begin marker for attribute NWidgetPart types.
```

src/widget.cpp
```diff
@@ -2957,192 +2957,189 @@ bool NWidgetLeaf::ButtonHit(const Point &pt)
...
- * Construct a NWidgetBase object from a #NWidget function, and apply all
- * settings that follow it, until encountering a #EndContainer, another
- * #NWidget, or the end of the parts array.
- *
- * @param nwid_begin Iterator to beginning of nested widget parts.
- * @param nwid_end Iterator to ending of nested widget parts.
- * @param dest  Address of pointer to use for returning the composed widget.
- * @param fill_dest Fill the composed widget with child widgets.
- * @return Iterator to remaining nested widget parts.
+ * Test if (an NWidgetPart) WidgetType is an attribute widget part type.
```
rubidium42, line 2960:
```suggestion
 * Test if (a NWidgetPart) WidgetType is an attribute widget part type.
```

src/widget.cpp
```diff
@@ -2957,192 +2957,189 @@ bool NWidgetLeaf::ButtonHit(const Point &pt)
...
-				break;
-
-			case NWID_SELECTION: {
-				if (dest != nullptr) return nwid_begin;
-				dest = std::make_unique<NWidgetStacked>(nwid_begin->u.widget.index);
-				*fill_dest = true;
-				break;
-			}
+/**
+ * Make NWidget from an NWidgetPart.
```
rubidium42, line 3078:
```suggestion
 * Make NWidget from a NWidgetPart.
```

rubidium42 (approved):

src/widget.cpp
PeterN: Hmm, so I pronounce this as "en-widget-part", which therefore requires "an" in English.

rubidium42: Fine by me

PR #12790, opened 2024-06-15
Add: Interface and classes to list system fonts. (WIP)

PeterN:
## Motivation / Problem

Font listing!

## Description

Implement an interface to list system fonts instead of having a user search by path and filename.

This adds an interface to list system fonts and implements it for FontConfig, Windows and Mac OS.
A console command is added as a way to test this interface.

There are two versions of this interface:

* The first has two methods, one to list all families, and another to list all styles in that families. This matches how the Windows API behaves.
* These second lists all families and styles in one go and remembers them, and these are then used to list families and styles. 

This matches the FontConfig and Mac OS APIs better.

The first interface can be used by skipping the last two commits. The last two commits change this to the second interface and would need to be squashed. Which interface should be used is an exercise for later...

## Checklist for review

### DIFF
--- a/src/os/unix/font_unix.cpp
+++ b/src/os/unix/font_unix.cpp
@@ -196,15 +196,14 @@ bool SetFallbackFont(FontCacheSettings *settings, const std::string &language_is
  */
 class FontConfigFontSearcher : public FontSearcher {
 public:
-	std::vector<std::string> ListFamilies(const std::string &language_isocode, int winlangid) override;
-	std::vector<FontFamily> ListStyles(const std::string &language_isocode, int winlangid, std::string_view font_family) override;
+	void UpdateCachedFonts(const std::string &language_isocode, int winlangid) override;
 };
 
-std::vector<std::string> FontConfigFontSearcher::ListFamilies(const std::string &language_isocode, int)
+void FontConfigFontSearcher::UpdateCachedFonts(const std::string &language_isocode, int)
 {
-	std::vector<std::string> families;
+	this->cached_fonts.clear();
 
-	if (!FcInit()) return families;
+	if (!FcInit()) return;
 
 	FcConfig *fc_instance = FcConfigReference(nullptr);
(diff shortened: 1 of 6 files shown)


src/os/unix/font_unix.cpp
```diff
@@ -39,6 +39,19 @@ static std::tuple<std::string, std::string> SplitFontFamilyAndStyle(std::string_
...
 }
 
+/**
+ * Get language string for FontConfig pattern matching.
+ * @param language_isocode Language's ISO code.
+ * @return Language code for FontConfig.
+ */
+static std::string GetFontConfigLanguage(const std::string &language_isocode)
+{
+	/* Fontconfig doesn't handle full language isocodes, only the part
```
ladysadie, line 49: I'm not exactly sure what you mean by this.  The fontconfig library will parse "en_GB" without any issues.  I don't see any fonts when I use that as the language set because all of the fonts on my system use "en". 

Scrolling through the list shows me that fonts having a  "xx-xx" language code is very rare.

michicc:
General thought: I think the API should fit the usage in OTTD. Whether this is a good match for some OS or some other isn't very important, me thinks.

src/os/unix/font_unix.cpp
PeterN:
```
$ fc-list :lang=en | wc -l
1883
$ fc-list :lang=en_GB | wc -l
0
```

PeterN:
> General thought: I think the API should fit the usage in OTTD. Whether this is a good match for some OS or some other isn't very important, me thinks.

Agreed, but I don't know what that usage is yet :)

src/os/unix/font_unix.cpp
ladysadie: The comment is misleading. No issues with parsing that text.  Just none of the fonts use it.

PeterN: The comment does not mention anything about parsing.

PR #12797, opened 2024-06-18
Add: Vehicle window zoom with mouse wheel

steve-goldman:
## Motivation / Problem

Some players would like to be able to zoom in and out in vehicle windows.

## Description

Make the mouse wheel control the zoom level in the vehicle window in the same way that it controls it for the main and extra viewport windows.

output

## Checklist for review


steve-goldman:
If there is appetite for this change, we could consider mouse wheel zoom for these as well:

image

image

steve-goldman:
Also, semi-related...

In the main viewport while following a vehicle, I have found it frustrating that changing zoom breaks you out of follow mode.

We could consider changing that, or doing it with ctrl+mousewheel.

follow

audigex:
Defaulting to not breaking follow mode seems like the best approach. There is already a control to break follow mode and I can't imagine many people are using zoom as a shortcut for that

src/main_gui.cpp
```diff
@@ -87,10 +87,11 @@ void CcPlaySound_EXPLOSION(Commands, const CommandCost &result, TileIndex tile)
  * Zooms a viewport in a window in or out.
  * @param how Zooming direction.
  * @param w   Window owning the viewport.
+ * @param stop_following true to stop following a vehicle if the window had been following one
```
2TallTyler:
```suggestion
 * @param stop_following Should we stop following the vehicle in the viewport?
```

src/vehicle_gui.cpp
```diff
@@ -3317,6 +3317,13 @@ struct VehicleViewWindow : Window {
 		}
 	}
 
+	void OnMouseWheel(int wheel) override
+	{
+		if (_settings_client.gui.scrollwheel_scrolling != 2) {
```
2TallTyler: If you're up for it, changing this magic number into an enum used in the .ini file and everywhere it's checked, would be some nice tech debt to resolve. Look at how `ViewportScrollMode` is defined in `settings_type.h` for reference. This would be a separate commit in this PR, preceding the current commit.

If not, no worries, but it hurts a little bit to see new magic numbers being added to the codebase. 😄

2TallTyler (commented):
Useful feature, thanks for the PR.

I'm not sure why Check Annotations CI is failing -- normally they show in the Files Changed tab. Hopefully a force-push to re-run CI will resolve it.

I am open to adding this feature to other viewport windows, and to changing following behavior (I agree with @audigex), but those should be separate PRs. 🙂

src/main_gui.cpp
steve-goldman: Updated as such in the latest commit.

src/vehicle_gui.cpp
steve-goldman: I think this is done as of the latest forced push.

I also added a commit to use the `ViewportScrollMode` enum for `scroll_mode` in the ini file.... seemed like that was missing.

2TallTyler (approved):

PR #12807, opened 2024-06-25
Fix #10028: Ensure stopped RVs do not prevent overtakes

### LINKED ISSUE #10028: [Bug]: RV overtaking fails if multiple vehicles follow closely

sevii1:
## Motivation / Problem

RV overtaking would fail in certain circumstances if multiple vehicles were driving closely, usually if the stopped RV was between tiles.

Closes #10028.

## Description

Added an extra condition in the check in EnumFindVehBlockingOvertake in roadveh_cmd.cpp that checks to see if the vehicle is currently attempting to overtake but was blocked by a stopped vehicle stuck between two tiles, forcing its speed to be 0, which causes the overtake to occur anyway.

## Limitations

The road vehicle may sometimes overtake in a scenario where it shouldn't, such as if another car is traveling in the opposite direction, causing a potential accident. I can't really pinpoint what exactly causes that to happen and didn't have too much time to test it, but that is one notable edge case.

## Checklist for review

### DIFF
--- a/src/roadveh_cmd.cpp
+++ b/src/roadveh_cmd.cpp
@@ -772,7 +772,17 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
 {
 	const OvertakeData *od = (OvertakeData*)data;
 
-	return (v->type == VEH_ROAD && v->First() == v && v != od->u && v != od->v) ? v : nullptr;
+	/**
+	* first ensure that the vehicle we are looking at is a road vehicle
+	* also ensure that the vehicle we are looking at is neither of the vehicles stored in OvertakeData
+	* 
+	* A block can only happen if more than two vehicles are involved, so we can skip past this if there are only two
+	*/
+	if ((v->type == VEH_ROAD && v->First() == v) && (v != od->u && v != od->v)) {
+		/* ensure that third vehicle overtaking does not get stuck in place by allowing it to overtake */
+		return (((RoadVehicle *)v)->overtaking && v->cur_speed == 0) ? v : nullptr;
+	}
+	return nullptr;
 }
 
 /**


2TallTyler:
The commit checker is failing for trailing whitespace in your comment, and because the commit message doesn't match the standard format. The PR title is correct, so just make the commit message match. 😉 

Also, have you seen #12708? It needs testing, but maybe that's something you'd be interested in.

src/roadveh_cmd.cpp
```diff
@@ -772,7 +772,17 @@ static Vehicle *EnumFindVehBlockingOvertake(Vehicle *v, void *data)
...
+	* first ensure that the vehicle we are looking at is a road vehicle
+	* also ensure that the vehicle we are looking at is neither of the vehicles stored in OvertakeData
+	* 
+	* A block can only happen if more than two vehicles are involved, so we can skip past this if there are only two
+	*/
+	if ((v->type == VEH_ROAD && v->First() == v) && (v != od->u && v != od->v)) {
+		/* ensure that third vehicle overtaking does not get stuck in place by allowing it to overtake */
+		return (((RoadVehicle *)v)->overtaking && v->cur_speed == 0) ? v : nullptr;
+	}
+	return nullptr;
```
rubidium42, line 785:
```suggestion
	/**
	 * First ensure that the vehicle we are looking at is a road vehicle,
	 * Also ensure that the vehicle we are looking at is neither of the vehicles stored in OvertakeData.
	 *
	 * A block can only happen if more than two vehicles are involved, so we can skip past this if there are only two.
	 */
	if ((v->type == VEH_ROAD && v->First() == v) && (v != od->u && v != od->v)) {
		/* Ensure that third vehicle overtaking does not get stuck in place by allowing it to overtake. */
		return (static_cast<RoadVehicle *>(v)->overtaking && v->cur_speed == 0) ? v : nullptr;
	}
	return nullptr;
```

PR #12808, opened 2024-06-25
Change: Changing zoom no longer stops following vehicle

steve-goldman:
## Motivation / Problem

Players would like to be able to change zoom while following a vehicle in the main viewport.  The current behavior stops following when you change zoom.

follow

## Description

Change the behavior of the game so that changing zoom does not stop following.

follow3

## Checklist for review

### DIFF
--- a/src/main_gui.cpp
+++ b/src/main_gui.cpp
@@ -87,11 +87,10 @@ void CcPlaySound_EXPLOSION(Commands, const CommandCost &result, TileIndex tile)
  * Zooms a viewport in a window in or out.
  * @param how Zooming direction.
  * @param w   Window owning the viewport.
- * @param stop_following Should we stop following the vehicle in the viewport?
  * @return Returns \c true if zooming step could be done, \c false if further zooming is not possible.
  * @note No button handling or what so ever is done.
  */
-bool DoZoomInOutWindow(ZoomStateChange how, Window *w, bool stop_following)
+bool DoZoomInOutWindow(ZoomStateChange how, Window *w)
 {
 	Viewport *vp;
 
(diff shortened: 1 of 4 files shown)


2TallTyler:
I like this PR, but we need to consider how the player can break out of following mode.

Mouse movement still works, but arrow keys and edge scrolling do not. I suggest allowing these to break out of following mode, to compensate for zoom no longer having an effect.

steve-goldman:
> I like this PR, but we need to consider how the player can break out of following mode.
> 
> Mouse movement still works, but arrow keys and edge scrolling do not. I suggest allowing these to break out of following mode, to compensate for zoom no longer having an effect.

Thanks for the feedback.  I agree that changing key scrolling and autoscrolling to break out of follow mode makes more sense than the current behavior.

I added some commits.  Let me know what you think?

2TallTyler (dismissed):
Code looks good, and it works!

PeterN (dismissed):
Given it's now possible to follow a vehicle and zoom out fully, why can I not start following a vehicle unless I'm zoomed in to at least normal zoom?

steve-goldman:
> Given it's now possible to follow a vehicle and zoom out fully, why can I not start following a vehicle unless I'm zoomed in to at least normal zoom?

Good question... I hadn't noticed.  Conditioning on the current zoom appears to go back to [when the feature was introduced in 2007](https://github.com/OpenTTD/OpenTTD/commit/705f0ae90c0b17f2d051e1705569f54f687d76c7).

I'd be inclined to label this as a bug, since the outcome of double-clicking does not match the tooltip description, at some zoom levels.

steve-goldman:
> > Given it's now possible to follow a vehicle and zoom out fully, why can I not start following a vehicle unless I'm zoomed in to at least normal zoom?
> 
> Good question... I hadn't noticed. Conditioning on the current zoom appears to go back to [when the feature was introduced in 2007](https://github.com/OpenTTD/OpenTTD/commit/705f0ae90c0b17f2d051e1705569f54f687d76c7).
> 
> I'd be inclined to label this as a bug, since the outcome of double-clicking does not match the tooltip description, at some zoom levels.

@PeterN, I added a commit to change this.  Thanks.

2TallTyler (approved):
Thanks for the quick fix!

steve-goldman:
> Thanks for the quick fix!

Happy to help.  Thanks to you guys for running the project so smoothly.

TrueBrain:
Really nice change/fix :) Tnx!

PR #12824, opened 2024-06-27
Feature: Set application name for SDL2

v1993:
This is used for e.g. screensaver inhibitor description.

## Motivation / Problem

This is how OpenTTD preventing sleep currently looks like in KDE Plasma:

image

## Description

This sets OpenTTD as the application name, which SDL2 then properly passes to system:

image

## Checklist for review

### DIFF
--- a/src/video/sdl2_v.cpp
+++ b/src/video/sdl2_v.cpp
@@ -523,6 +523,8 @@ static std::optional<std::string_view> InitializeSDL()
 	/* Check if the video-driver is already initialized. */
 	if (SDL_WasInit(SDL_INIT_VIDEO) != 0) return std::nullopt;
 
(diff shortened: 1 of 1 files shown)


LordAro:
Seems sensible. Should probably be an "Add" rather than a Feature though.

The existing SetHint is in `VideoDriver_SDL_Base::Start`, any particular reason why the new one can't be there?

rubidium42:
I agree that `VideoDriver_SDL_Base::Start` seems a more sensible location.

Besides that, `SDL_HINT_APP_NAME` has not been part of SDL2 since the begin (it's not even 3 years old). So it would be nice if the hinting was `#ifdef` guarded, just like is done for `SDL_HINT_MOUSE_AUTO_CAPTURE`.

PR #12835, opened 2024-07-01
Fix #12832: Sanitise strings from NewGRF before logging them in Debug output.

### LINKED ISSUE #12832: [Bug]: Several NewGRFs get disabled on load if OpenTTD is started with `-dgrf8`
### Version of OpenTTD

8c7cf3bc75a47c06b5ba4e6a4fbe200d2802f77f on Windows 10, 64-bit (via Steam)

### Expected result

Starting OpenTTD with `-dgrf8` should allow all NewGRFs (that can be used without passing any CLI arguments) to be used in a given save game.

### Actual result

When starting OpenTTD with `-dgrf8`, some NewGRFs are disabled when used in a new game or while loading a saved game. This can cause crashes or unexpected behavior similarly to loading a savegame with missing NewGRFs.

In all such cases so far, the NewGRFs return a fatal error: "Read past end of psuedo-sprite (sprite N)," where N is different for each NewGRF.

NewGRFs affected include:
* Dutch Landmarks Objects
* av8
* City Objects
* Vactrain Set
* FIRS
* WIRES - Welcome Innovation Regarding Electricity Shipment
* ... and many more

### Steps to reproduce

1. Start OpenTTD with `-dgrf8`.
2. Load a savegame containing NewGRFs, or start a new game with NewGRFs.
3. Depending on the NewGRF configuration, some NewGRFs may be disabled.

PeterN:
## Motivation / Problem

As per #12832, loading NewGRFs may 'crash' on Windows when grf debugging is enabled if non-UTF8 strings are loaded from NewGRF files.

## Description

Call `StrMakeValid()` when outputting raw string during load from NewGRF.

`TranslateTTDPatchCodes()` does exist which will convert the string but that does a lot more.

## Limitations

Not actually tested on Windows.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -2891,14 +2891,14 @@ static ChangeInfoResult GlobalVarChangeInfo(uint gvid, int numinfo, int prop, By
 					if (prop == 0x13) {
 						map.openttd_id = lang->GetGenderIndex(name.data());
 						if (map.openttd_id >= MAX_NUM_GENDERS) {
-							GrfMsg(1, "GlobalVarChangeInfo: Gender name {} is not known, ignoring", name);
+							GrfMsg(1, "GlobalVarChangeInfo: Gender name {} is not known, ignoring", StrMakeValid(name));
 						} else {
 							_cur.grffile->language_map[curidx].gender_map.push_back(map);
 						}
 					} else {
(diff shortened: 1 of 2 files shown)


LordAro:
Presumably we're not worried about the performance penalty of adding extra calls into Debug function parameters that might never be output anywhere?

LordAro (approved):

PR #12846, opened 2024-07-06
Codechange: Pass span to StrValid instead of first and last - 1.

PeterN:
## Motivation / Problem

`StrValid()` requires passing the start and end (-1) of the buffer to validate, and callers use the `lastof()` macro to do so.

## Description

Pass `std::span` to `StrValid()` instead. This allows the compiler to do the work for us.

`std::span` is used instead of `std::string_view` as this is only used for fixed-length buffers.

This removes some callers of `lastof()`

## Checklist for review

### DIFF
--- a/src/string.cpp
+++ b/src/string.cpp
@@ -226,32 +226,35 @@ std::string StrMakeValid(std::string_view str, StringValidationSettings settings
 /**
  * Checks whether the given string is valid, i.e. contains only
  * valid (printable) characters and is properly terminated.
- * @param str  The string to validate.
- * @param last The last character of the string, i.e. the string
- *             must be terminated here or earlier.
+ * @note std::span is used instead of std::string_view as we are validating fixed-length string buffers, and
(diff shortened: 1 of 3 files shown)


src/string.cpp
```diff
@@ -226,32 +226,33 @@ std::string StrMakeValid(std::string_view str, StringValidationSettings settings
 /**
  * Checks whether the given string is valid, i.e. contains only
  * valid (printable) characters and is properly terminated.
- * @param str  The string to validate.
- * @param last The last character of the string, i.e. the string
- *             must be terminated here or earlier.
+ * @note std::span is used instead of std::string_view as this is only used for validating fixed-length string buffers.
```
rubidium42: What's maybe even more important why `std::string_view` shouldn't be used because its constructor is going to essentially call `strlen`, which assumes there is a `'\0'`. Exactly one of those things that this function checks to ensure the validity of the passed C-style string.

michicc (approved):

PR #12849, opened 2024-07-09
Codechange: Use find_if to get default writeable saveload format.

PeterN:
## Motivation / Problem

`GetSaveloadFormat()` uses a loop to search for the last writeable saveload format, using `lastof()` to start at the end, and manually decrementing in a loop. This loop assumes there is always a writeable saveload format and does not test for bounds of the `_saveload_formats` array.

## Description

Replace the search with a call to `std::find_if` using reverse iterators to start at the end. While at it, call `SlError()` if no writeable saveload format exists (very unlikely) and replace the return pointer with reference, as the saveload format must exist.

This removes the last use of `lastof()`.

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -2680,14 +2680,15 @@ static const SaveLoadFormat _saveload_formats[] = {
  * uncompressed, or another type
  * @param full_name Name of the savegame format. If empty it picks the first available one
  * @param compression_level Output for telling what compression level we want.
- * @return Pointer to SaveLoadFormat struct giving all characteristics of this type of savegame
+ * @return Reference to SaveLoadFormat struct giving all characteristics of this type of savegame
  */
-static const SaveLoadFormat *GetSavegameFormat(const std::string &full_name, uint8_t *compression_level)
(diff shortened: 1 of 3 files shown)


rubidium42 (approved):
Have you missed the `lastof` in gfx.cpp, or did you make a conscious choice to leave that as-is?

src/saveload/saveload.cpp
```diff
@@ -2680,14 +2680,15 @@ static const SaveLoadFormat _saveload_formats[] = {
  * uncompressed, or another type
  * @param full_name Name of the savegame format. If empty it picks the first available one
  * @param compression_level Output for telling what compression level we want.
- * @return Pointer to SaveLoadFormat struct giving all characteristics of this type of savegame
+ * @return Reference to SaveLoadFormat struct giving all characteristics of this type of savegame
  */
-static const SaveLoadFormat *GetSavegameFormat(const std::string &full_name, uint8_t *compression_level)
+static const SaveLoadFormat &GetSavegameFormat(const std::string &full_name, uint8_t *compression_level)
```
glx22, line 2685: `compression_level` could be a reference, as the code seems to assume it's never `nullptr`

PR #12852, opened 2024-07-10
Codechange: Use range-for and std::size with _plural_forms.

PeterN:
## Motivation / Problem

Use of the `lengthof()` macro, along with indexed array access, to find the maximal plural count.

## Description

Use range-for and `std::size` with the `_plural_forms` loop instead.

This removes indexed array access and use of `lengthof()` macro.

## Checklist for review

### DIFF
--- a/src/strgen/strgen.cpp
+++ b/src/strgen/strgen.cpp
@@ -250,8 +250,8 @@ struct HeaderFileWriter : HeaderWriter, FileWriter {
 	{
 		/* Find the plural form with the most amount of cases. */
 		int max_plural_forms = 0;
-		for (uint i = 0; i < lengthof(_plural_forms); i++) {
-			max_plural_forms = std::max(max_plural_forms, _plural_forms[i].plural_count);
+		for (const auto &pf : _plural_forms) {
+			max_plural_forms = std::max(max_plural_forms, pf.plural_count);
 		}
 
 		fmt::print(this->output_stream,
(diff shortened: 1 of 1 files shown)


src/strgen/strgen.cpp
```diff
@@ -250,8 +250,8 @@ struct HeaderFileWriter : HeaderWriter, FileWriter {
 	{
 		/* Find the plural form with the most amount of cases. */
 		int max_plural_forms = 0;
-		for (uint i = 0; i < lengthof(_plural_forms); i++) {
-			max_plural_forms = std::max(max_plural_forms, _plural_forms[i].plural_count);
+		for (const auto &pf : _plural_forms) {
+			max_plural_forms = std::max(max_plural_forms, pf.plural_count);
 		}
```
LordAro, line 255: `std::max_element` ? Might be more complicated than it's worth

PeterN: `std::max_element` would work if it wasn't the `plural_count` member that we're after.

As such it would need a custom comparator, and indeed would look more complicated.

LordAro (approved):

src/strgen/strgen.cpp
PeterN: Though

```c++
int max_plural_forms = std::max_element(std::begin(_plural_forms), std::end(_plural_forms), [](const auto &a, const auto &b) { return a.plural_count < b.plural_count; })->plural_count;
```

would do it.

LordAro: Like I say, may be more complicated than it's worth

PR #12853, opened 2024-07-10
Codechange: Replace default overrides arrays with std::pair and use range-for.

PeterN:
## Motivation / Problem

Indexed array access and use of `lengthof()` when setting up default grf overrides.

## Description

Replace default grf overrides arrays with array of `std::pair` and use range-for to iterate.

This avoids indexed array access and use of `lengthof()`.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -10075,13 +10075,13 @@ void LoadNewGRF(uint load_index, uint num_baseset)
 		}
 
 		if (stage == GLS_RESERVE) {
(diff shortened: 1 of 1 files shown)


rubidium42 (dismissed):
Would it make sense to `BSWAP32` the (static) table? Or can't we because it's not `constexpr`? It would prevent having to swap these for each loaded NewGRF.

rubidium42 (approved):

PR #12857, opened 2024-07-11
Change: Path signals now show green on junction-less tracks by default

michicc:
## Motivation / Problem

It is well known that path signals are slower. After all, they show red all the time, and red is very much the colour for stop and not go in western cultures. Ergo, red signals are slow signals.

## Description

Make path signals "faster" by making path signals that point into a junction-less track segment show green by default.

This mostly matches how real-world automatic block signalling works. Straight runs of track without any switches are usually signalled in a way that does not require manual action from the signal man and default to green for free tracks. Junctions that would be manually signalled stay default-red with this PR, also matching real-world.

This is purely a visual change and has no effect on pathfinding or path reservations.

Note: PR title != commit message.

## Limitations

Signal state is not recalculated on load, so signals will stay red until a train causes a signal update to occur.

Path signals that are reserved from behind are turned red, as handling already reserved path signals with the normal signal algorithm is prone to race conditions if train A reserves a path and train B enters the block before train A. To simplify this, the state if already reserved path signals will not be updated by the signal handler.

## Checklist for review

### DIFF
--- a/src/pathfinder/yapf/yapf_rail.cpp
+++ b/src/pathfinder/yapf/yapf_rail.cpp
@@ -57,6 +57,8 @@ private:
 	Trackdir  m_res_fail_td;      ///< The trackdir where the reservation failed
 	TileIndex m_origin_tile;      ///< Tile our reservation will originate from
 
+	std::vector<std::pair<TileIndex, Trackdir>> m_signals_set_to_red; ///< List of signals turned red during a path reservation.
+
 	bool FindSafePositionProc(TileIndex tile, Trackdir td)
 	{
 		if (IsSafeWaitingPosition(Yapf().GetVehicle(), tile, td, true, !TrackFollower::Allow90degTurns())) {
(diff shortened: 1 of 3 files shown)


michicc:
https://github.com/OpenTTD/OpenTTD/assets/1447653/fa159af5-1ec1-4e79-838c-db03cc2cdfeb

andythenorth:
"This is excellent" 😄

michicc:
Now with proper handling of manual train reversing.

JGRennison:
This works very nicely in unidirectional track layouts.
When a single piece of plain line has entrances/signals at both ends having both signals be green at the same time can look a little odd though. This leads to cases where signals turn from green to red just as a train approaches them, which could break the illusion a bit.

Silly layout:
image

(I have something vaguely along the lines of this feature in my branch, two green signals facing into the same bit of track is not OK in the context of that feature. I'm inclined to lean that way for this as well.)

2TallTyler:
It's still possible to have a green signal that gets dropped to red by a train in the opposite direction.

green

I wonder if it makes sense to only allow green signals on one-way tracks, that have a one-way path signal between the signal in question and any junctions.

junction

Thoughts?

michicc:
> It's still possible to have a green signal that gets dropped to red by a train in the opposite direction.

Yes, and this, for the fully generic case, basically not solvable. Only turning the signals green on one-way segments means proving the absence of any signal allowing entry into the segment from anywhere. In extreme, proving the absence means a full graph walk.
And somehow I don't think that a full walk of all rail tiles on the map for every single signal change is feasible. Admittedly, the worst-case network is unlikely to occur, but never underestimate our player-base 🤣 

A bit more serious: Right now the signal exploration mostly stops at  the first signal, but your suggestion would mean walking a lot more signal segments each time.

2TallTyler (approved):
Yep, that makes sense. In that case, since testing and code looks good... #YOLO! 😄

PR #12858, opened 2024-07-14
Fix #12856, e17c82e: Updating network settings does not invalidate data

### LINKED ISSUE #12856: [Bug]: New Company button missing while server is full and Max Companies setting changes
### Version of OpenTTD

OpenTTD 14.1

### Expected result

When server is full of companies, Online Players Window updates with correct New Company button when `network.max_companies` increases 

### Actual result

When server is full of companies, Online Players Window will not add "New Company" button when `network.max_companies` increases - and retains "Join Company" - and it's out of sync with the company name list.

### Steps to reproduce

1. Create a dedicated server, set max companies to 1
2. Join game from a client, and create a company.  The server should have "max companies" and there is no longer a "New Company" button in `Online Players`. 
3. Join Spectators. THere should be just one empty company.
4. Keep open the Online Players window
5. Change the Max Companies setting to 2
6. See that there is no "New Company " button.   Closing the window and re-opening it will fix the issue.
image

ivpravdin:
## Motivation / Problem

Fix #12856. When a server changes `network.max_companies` setting, "(New company)" appears in the client list but "+" button does not. This happens because the window is refreshed by the data is not invalidated, therefore it is not re-computed. 

This bug also allows clients to create new companies if the server decreased the `network.max_companies` to equal or below the number of companies in the game. If the client did not refresh their client list window, they are still able to create a new company even if it will put the number of created companies above `network.max_companies`.

## Description

This bug was fixed by executing `InvalidateWindowData` for the client list window for both the server and the client when network.max_companies is changed.

## Limitations

No limitations were identified.

### DIFF
--- a/src/network/network_client.cpp
+++ b/src/network/network_client.cpp
@@ -1188,7 +1188,8 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::Receive_SERVER_CONFIG_UPDATE(P
 
 	_network_server_max_companies = p.Recv_uint8();
 	_network_server_name = p.Recv_string(NETWORK_NAME_LENGTH);
-	SetWindowClassesDirty(WC_CLIENT_LIST);
+
+	InvalidateWindowData(WC_CLIENT_LIST, 0);
 
 	Debug(net, 9, "Client::Receive_SERVER_CONFIG_UPDATE(): max_companies={}", _network_server_max_companies);
 
--- a/src/settings_table.cpp
+++ b/src/settings_table.cpp
@@ -572,9 +572,10 @@ static void UpdateClientConfigValues()
 {
 	NetworkServerUpdateGameInfo();
 
+	InvalidateWindowData(WC_CLIENT_LIST, 0);
+
 	if (_network_server) {
 		NetworkServerSendConfigUpdate();
-		SetWindowClassesDirty(WC_CLIENT_LIST);
 	}
 }
 


src/network/network_client.cpp
```diff
@@ -1188,6 +1188,8 @@ NetworkRecvStatus ClientNetworkGameSocketHandler::Receive_SERVER_CONFIG_UPDATE(P
 
 	_network_server_max_companies = p.Recv_uint8();
 	_network_server_name = p.Recv_string(NETWORK_NAME_LENGTH);
+
+	InvalidateWindowData(WC_CLIENT_LIST, 0);
```
rubidium42, line 1192: `InvalidateWindowData` also marks the window dirty, so the `SetWindowClassesDirty` can be removed.

src/settings_table.cpp
```diff
@@ -572,9 +572,11 @@ static void UpdateClientConfigValues()
 {
 	NetworkServerUpdateGameInfo();
 
+	InvalidateWindowData(WC_CLIENT_LIST, 0);
```
rubidium42, line 575: `InvalidateWindowData` also marks the window dirty, so the `SetWindowClassesDirty` can be removed all together.

src/network/network_client.cpp
ivpravdin: Thanks, I've updated it

src/settings_table.cpp
ivpravdin: I updated

rubidium42 (approved):

PR #12861, opened 2024-07-16
Add: [Script] Include number of victims in ScriptEventVehicleCrashed

glx22:
## Motivation / Problem
A script author wanted to determine the number of victims after a crash.

Doing it inside the script is possible, but the actual number of victims is randomised when the crash happens, and the vehicle may disappear before script has time to start its calculation.

## Description
Since number of victims is known when the Crash event is fired, add it to the event data.

## Checklist for review

### DIFF
--- a/src/disaster_vehicle.cpp
+++ b/src/disaster_vehicle.cpp
@@ -384,13 +384,13 @@ static bool DisasterTick_Ufo(DisasterVehicle *v)
 		if (z <= u->z_pos && (u->vehstatus & VS_HIDDEN) == 0) {
 			v->age++;
 			if (u->crashed_ctr == 0) {
-				u->Crash();
+				uint victims = u->Crash();
 				u->disaster_vehicle = INVALID_VEHICLE;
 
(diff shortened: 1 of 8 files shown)


src/disaster_vehicle.cpp
```diff
@@ -384,13 +384,13 @@ static bool DisasterTick_Ufo(DisasterVehicle *v)
 		if (z <= u->z_pos && (u->vehstatus & VS_HIDDEN) == 0) {
 			v->age++;
 			if (u->crashed_ctr == 0) {
-				u->Crash();
+				uint pass = u->Crash();
```
rubidium42: It might be better if it were to be just `victims`, as it also includes the driver.
The same holds for the other uses of `pass` and/or `num` in this PR.

rubidium42 (dismissed):

src/disaster_vehicle.cpp
glx22: Variables renamed where appropriate.

rubidium42 (approved):

PR #12868, opened 2024-07-17
Codechange: Simplify IsCompanyBuildableVehicleType.

J0anJosep:
## Motivation / Problem

None really.

## Description

IsCompanyBuildableVehicleType used a switch that can be replaced by a single comparison.

## Limitations

None

## Checklist for review

### DIFF
--- a/src/vehicle_func.h
+++ b/src/vehicle_func.h
@@ -90,15 +90,7 @@ Direction GetDirectionTowards(const Vehicle *v, int x, int y);
  */
 inline bool IsCompanyBuildableVehicleType(VehicleType type)
 {
-	switch (type) {
-		case VEH_TRAIN:
-		case VEH_ROAD:
-		case VEH_SHIP:
-		case VEH_AIRCRAFT:
-			return true;
-
-		default: return false;
-	}
+	return type < VEH_COMPANY_END;
 }
 
 /**


PeterN:
While it might be less code (so simpler), it isn't clearer in my opinion.

J0anJosep:
> While it might be less code (so simpler), it isn't clearer in my opinion.

Just looking into the VehicleType enum makes it clear enough.

Maybe what is not clear is the name of the function. It is equivalent to `IsValidCompanyVehicleType`, which would fit better, as there are a lot of IsValid* functions, which usually use a simple comparison like this and not switches.

There are also some uses of `type < VEH_COMPANY_END` or something similar in several loops, which are way clearer than using `IsCompanyBuildableVehicleType(type)`.

2TallTyler:
`IsValidCompanyVehicleType()` does make more sense for how this is used.

I suppose my question here is: Is this a change for change's sake or is this refactoring ahead of a future PR?

J0anJosep:
> `IsValidCompanyVehicleType()` does make more sense for how this is used.
> 
> I suppose my question here is: Is this a change for change's sake or is this refactoring ahead of a future PR?

Neither of the two. It is mostly to keep code consistent, as many other functions use this approach, and simple.

The `IsPlayerBuildableVehicleType` function was introduced in 2007 and the `VEH_COMPANY_END` constant was added in 2011. When `IsPlayerBuildableVehicleType` was introduced the constant didn't exist, so it was written with a switch. But had `VEH_COMPANY_END` been there, it would have used the constant, which is the natural way to do this.

2TallTyler (approved):
Fine by me, I like consistency. 🤷

PR #12873, opened 2024-07-20
Fix #12860, 2bada59: Timetable does not show vehicles as early

### LINKED ISSUE #12860: [Bug]: Timetable status doesn't say if a vehicle is early
### Version of OpenTTD

Nightly (eeed824)

### Expected result

When a vehicle arrives early, its status should say "N days early".

### Actual result

The vehicle says "running on time".

### Steps to reproduce

1. Create a vehicle traveling between two stations and autofill its timetable.
2. Change one running time to a larger number, say from 9 days to 30 days.
3. When it arrives at a station, it will wait until its departure time, but the status will say "running on time" and not "N days early".

ivpravdin:
## Motivation / Problem

Fix #12860. Timetable does not show vehicles as early, only as on time. If the vehicle's travel time is greater than the actual time it takes to reach the destination, the vehicle will arrive early but the timetable will display it as "running on time". 

## Description

This bug was fixed by changing VehicleIsAboveLatenessThreshold() to only accept how many ticks the vehicle is late, not the entire vehicle instance. This allows to pass an absolute value of number of ticks and therefore determine whether the vehicle is on time or late/early.

## Limitations

Limitation (unlikely): if in the future vehicle's late status will have to be determined by anything else other than number of ticks, new parameters will have to be added to VehicleIsAboveLatenessThreshold().

### DIFF
--- a/src/timetable_gui.cpp
+++ b/src/timetable_gui.cpp
@@ -86,20 +86,20 @@ static inline TimerGameTick::Ticks TicksPerTimetableUnit()
 }
 
 /**
- * Determine if a vehicle should be shown as late, depending on the timetable display setting.
- * @param v The vehicle in question.
+ * Determine if a vehicle should be shown as late or early, using a threshold depending on the timetable display setting.
+ * @param ticks The number of ticks that the vehicle is late or early.
  * @param round_to_day When using ticks, if we should round up to the nearest day.
- * @return True if the vehicle is later than the threshold.
+ * @return True if the vehicle is outside the "on time" threshold, either early or late.
  */
-bool VehicleIsAboveLatenessThreshold(const Vehicle *v, bool round_to_day)
+bool VehicleIsAboveLatenessThreshold(TimerGameTick::Ticks ticks, bool round_to_day)
 {
 	switch (_settings_client.gui.timetable_mode) {
 		case TimetableMode::Days:
-			return v->lateness_counter > Ticks::DAY_TICKS;
+			return ticks > Ticks::DAY_TICKS;
 		case TimetableMode::Seconds:
-			return v->lateness_counter > Ticks::TICKS_PER_SECOND;
+			return ticks > Ticks::TICKS_PER_SECOND;
 		case TimetableMode::Ticks:
-			return v->lateness_counter > (round_to_day ? Ticks::DAY_TICKS : 0);
+			return ticks > (round_to_day ? Ticks::DAY_TICKS : 0);
 		default:
 			NOT_REACHED();
 	}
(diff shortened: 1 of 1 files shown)


src/timetable_gui.cpp
```diff
@@ -87,19 +87,19 @@ static inline TimerGameTick::Ticks TicksPerTimetableUnit()
...
  * @param round_to_day When using ticks, if we should round up to the nearest day.
  * @return True if the vehicle is later than the threshold.
  */
-bool VehicleIsAboveLatenessThreshold(const Vehicle *v, bool round_to_day)
+bool VehicleIsAboveLatenessThreshold(TimerGameTick::Ticks ticks, bool round_to_day)
 {
 	switch (_settings_client.gui.timetable_mode) {
 		case TimetableMode::Days:
-			return v->lateness_counter > Ticks::DAY_TICKS;
+			return ticks > Ticks::DAY_TICKS;
```
2TallTyler, line 98: Thanks for the contribution!

Why change what is passed to the function, if all that's needed is to take the absolute value of the lateness counter?

If I'm not mistaken, this fix could be confined to `VehicleIsAboveLatenessThreshold()` by simply changing each switch case as below.

```suggestion
			return abs(v->lateness_counter) > Ticks::DAY_TICKS;
```

src/timetable_gui.cpp
```diff
@@ -87,19 +87,19 @@ static inline TimerGameTick::Ticks TicksPerTimetableUnit()
 
 /**
  * Determine if a vehicle should be shown as late, depending on the timetable display setting.
- * @param v The vehicle in question.
+ * @param ticks How many ticks the vehicle is late.
  * @param round_to_day When using ticks, if we should round up to the nearest day.
  * @return True if the vehicle is later than the threshold.
```
2TallTyler:
```suggestion
 * @return True if the vehicle is outside the "on time" threshold, either early or late.
```

src/timetable_gui.cpp
```diff
@@ -87,19 +87,19 @@ static inline TimerGameTick::Ticks TicksPerTimetableUnit()
 
 /**
  * Determine if a vehicle should be shown as late, depending on the timetable display setting.
```
2TallTyler:
```suggestion
 * Determine if a vehicle should be shown as late or early, using a threshold depending on the timetable display setting.
```

src/timetable_gui.cpp
ivpravdin: Hi, thank you for your suggestion. The problem here is that is that in this line:
`504: bool show_late = this->show_expected && VehicleIsAboveLatenessThreshold(v->lateness_counter, true);` it is necessary to actually determine whether the vehicle is late (and not early). With you suggestion `show_late` would be true for early vehicles as well. I've tried it the way you suggested and that's the issue I ran into. I think it makes sense to make it possible to determine if vehicle is both later/early or just late.

src/timetable_gui.cpp
```diff
@@ -87,19 +87,19 @@ static inline TimerGameTick::Ticks TicksPerTimetableUnit()
...
- * @param v The vehicle in question.
+ * @param ticks How many ticks the vehicle is late.
```
2TallTyler:
```suggestion
 * @param ticks The number of ticks that the vehicle is late or early.
```

src/timetable_gui.cpp
2TallTyler: Ah, thanks for the explanation. You are correct. 🙂 

Do you mind applying my comment suggestions? Then this is ready for an approval. 😃

2TallTyler (approved):

2TallTyler:
Checks are failing due to #12881, nothing you can do until that is fixed. 🙂

PR #12878, opened 2024-07-22
Add: [Script] ScriptEventCompanyRename

bjornwarmedal:
## Motivation / Problem

I have a ranking board of current companies in the storybook. When a company gets renamed I need that to be updated in the storybook.

## Description

I've added a script event type that fires off when a company is renamed.

## Limitations

None that I know of.

## Checklist for review


LordAro:
If you want to be extra, you could add a company rename to the regression AI script and check the event appears...

src/script/api/script_event.hpp
```diff
@@ -32,6 +32,7 @@ class ScriptEvent : public ScriptObject {
 		ET_SUBSIDY_EXPIRED,
 		ET_ENGINE_PREVIEW,
 		ET_COMPANY_NEW,
+		ET_COMPANY_RENAMED,
```
glx22: Inserting it here should not cause issues, but it might be better to put it at the end of the enum.

PeterN:
I mentioned this is in Discord, but better here.

In terms of API additions, does it make sense to add one event for just one thing that can be renamed? What about the manager name or perhaps a town name, etc?

rubidium42:
> In terms of API additions, does it make sense to add one event for just one thing that can be renamed? What about the manager name or perhaps a town name, etc?

I would have said that the name is not really important, as it can be easily fetched with the string system and then it's up-to-date, until it was mentioned that leader boards might also contain companies that do not exist any more. 

In such a case I can imagine that you just store the names of the companies and want to update those. Alternatively, you complicate the string building by having either a {COMPANY} for existing companies and {STRING} for dead companies... but then you would need to pass the (old) company name in the events that notify about the removal of the company. That sounds way more complicated and fragile than just the string names.

So I would suggest that it makes sense for things that might be in something like a leader board, but not for practically irremovable things like towns. Practically I'd say, when doing company name also do manager name. I wouldn't see much use in town names. For industry names I can't see how to do it correctly as those names can have translated parts ({TOWN} woods etc), so I would even strongly dissuade someone from implementing that.

bjornwarmedal:
> I mentioned this is in Discord, but better here.
> 
> In terms of API additions, does it make sense to add one event for just one thing that can be renamed? What about the manager name or perhaps a town name, etc?

I can't see any other multi-purpose events like that. Personally I'd love to see a company manager rename event, but rather as a separated event. Adding two events to the event handler in a script feels easier and neater than having to parse one event to find out which of two different "sub-events" it signals.

bjornwarmedal:
> I would have said that the name is not really important, as it can be easily fetched with the string system and then it's up-to-date, until it was mentioned that leader boards might also contain companies that do not exist any more.

This is the two things I aim to resolve: partly being able to keep "lost" companies in the list, but also knowing *when* to update the list with a new name. The event doesn't signal what the new name is, only which company has changed name. Because, as you say, the name is easily available. Having an event for when it changes is just nicer than periodically polling all the companies by ID to update their names.

bjornwarmedal:
> If you want to be extra, you could add a company rename to the regression AI script and check the event appears...

I'll look at this, but would it be okay if I do that as a separate PR? Since you consider it an extra perk :)

glx22:
Not sure the new event is testable in regression, it will be fired but impossible to catch it because it's not sent to the company triggering it.

bjornwarmedal:
> Not sure the new event is testable in regression, it will be fired but impossible to catch it because it's not sent to the company triggering it.

Well I guess I won't spend time on that then :laughing:

rubidium42:
While preparing a PR for an event when a president gets renamed, I found out that this PR still needs an update of the changelog. I also think that the event keeping the new name might be beneficial for cases where the company goes bankrupt or is removed before the event is processed.
Next to that I'm questioning whether AI's should get an event when a company/president name changes. If so, I'm challenging whether the AI should be notified about their own company/president name changing (e.g. a human player that has joined the AI could change), and then it can also be used in regression.

All in all, I'd suggest waiting a bit for comments on #13208 and model this and the president name variant in the same manner.

bjornwarmedal:
I've attempted to make it testable in regression now, but I honestly don't know what exactly makes it possible.

I've also rewritten it to follow the same format as [#13208], including using CompanyID instead of Owner, and including the new company name in the event.

The event also fires when a company is named at first construction; something I'm wondering if it should really use the rename company workflow (but that's not a discussion for this PR).

PeterN:
You need to update the expected regression test output files as well.

rubidium42 (approved):

PR #12879, opened 2024-07-23
Change: [NewGRF] Install translation tables into overridden NewGRF.

PeterN:
## Motivation / Problem

A NewGRF can override another NewGRF and change properties of the vehicles defined with in it.

This does allow cargo types to be individually modified, however this needs to be done for each vehicle, and cannot affect cargo-specific Action3s.

## Description

With this PR, when a NewGRF overrides another, any translation table that the overriding NewGRF installs will also be installed in the target file.

This allows the overridden NewGRF to make use of a cargo or rail/road type translation table without directly modifying the original file.

As the cargo translation table applies to existing cargo type properties and to cargo-specific Action3s, this allows all cargo properties to be 'fixed' in a much simpler manner.

## Limitations

To actually work properly this PR depends on #12646, otherwise cargo translation tables may not be applied due to the grf_version of the target NewGRF.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -602,6 +602,20 @@ static void SetNewGRFOverride(uint32_t source_grfid, uint32_t target_grfid)
 }
 
 /**
+ * Get overridden GRF for current GRF if present.
+ * @return Overridden GRFFile if present, or nullptr.
+ */
+static GRFFile *GetCurrentGRFOverride()
+{
+	auto found = _grf_id_overrides.find(_cur.grffile->grfid);
+	if (found != std::end(_grf_id_overrides)) {
+		GRFFile *grffile = GetFileByGRFID(found->second);
+		if (grffile != nullptr) return grffile;
+	}
+	return nullptr;
+}
+
+/**
  * Returns the engine associated to a certain internal_id, resp. allocates it.
  * @param file NewGRF that wants to change the engine.
  * @param type Vehicle type.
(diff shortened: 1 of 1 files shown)


PeterN:
As a short example, the following minimal NML will install a cargo translation table override into 'Ikarus Set 6' to ensure that cargo slot 0 is passengers:

```
grf {
    grfid: "PNC\02";
    name: string(STR_IKARUS_NAME);
    desc: string(STR_IKARUS_DESC);
    version: 1;
    min_compatible_version: 1;
}

engine_override("\4D\56\00\00");

cargotable {
    PASS
}
```

(Note that this particular example isn't actually useful as the default translation table change of #12646 will fix this particular NewGRF already.)

src/newgrf.cpp
```diff
@@ -2676,24 +2690,33 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
...
 	}
 
+	std::vector<T> &translation_table = gettable(*_cur.grffile);
 	translation_table.clear();
 	translation_table.reserve(numinfo);
 	for (int i = 0; i < numinfo; i++) {
 		translation_table.push_back(T(BSWAP32(buf.ReadDWord())));
 	}
 
+	GRFFile *override = GetCurrentGRFOverride();
```
LordAro: `override` as a variable name is a bit ick (woe, context unaware syntax highlighting)

src/newgrf.cpp
```diff
@@ -2676,24 +2690,33 @@ static ChangeInfoResult TownHouseChangeInfo(uint hid, int numinfo, int prop, Byt
...
  * @param numinfo Number of subsequent IDs to change the property for.
  * @param buf The property value.
- * @param[in,out] translation_table Storage location for the translation table.
+ * @param gettable Function to get storage for the translation table.
  * @param name Name of the table for debug output.
  * @return ChangeInfoResult.
  */
 template <typename T>
-static ChangeInfoResult LoadTranslationTable(uint gvid, int numinfo, ByteReader &buf, std::vector<T> &translation_table, const char *name)
+static ChangeInfoResult LoadTranslationTable(uint gvid, int numinfo, ByteReader &buf, std::function<std::vector<T> &(GRFFile &)> gettable, std::string_view name)
```
LordAro: historically we've avoided std::function because of various overheads. Guess template parameters are out of the question with lambdas? Maybe passing some sort of enum value as a template param and then filtering with an if constexpr or something?

PeterN: Probably the overhead doesn't matter here given it's only used on load and only for the lifetime for the call. However given it's already a templated function, I simply turned the function into a templated argument instead.

rubidium42 (approved):

PR #12882, opened 2024-07-28
Fix: remove the invalid UTF-8 character (U+200B) from the traditional chinese translation

### LINKED ISSUE #12881: [Bug]: Fails to compile.  FATAL: Unwanted UTF-8 character U+200B in sequence
### Version of OpenTTD

22e525dafef6273fad35c4358b942eb19ee94034

### Expected result

Should compile without error.

### Actual result

Error during compile:
```
[354/845] Compiling language traditional_chinese
FAILED: lang/traditional_chinese.lng /home/builder/build/src/build/lang/traditional_chinese.lng 
cd /home/builder/build/src/build/src/lang && /usr/bin/cmake -E make_directory /home/builder/build/src/build/lang && /home/builder/build/src/build/src/strgen/strgen -s /home/builder/build/src/openttd-git/src/lang -d /home/builder/build/src/build/lang /home/builder/build/src/openttd-git/src/lang/traditional_chinese.txt
/home/builder/build/src/openttd-git/src/lang/traditional_chinese.txt:3311: FATAL: Unwanted UTF-8 character U+200B in sequence '{BLACK}選擇遊戲版圖內最高峰的海拔，​​以海平面為基準計算'
[375/845] Generating table/strings.h
ninja: build stopped: subcommand failed.
```

### Steps to reproduce

Try to compile.

TrueBrain:
## Motivation / Problem

Fixes #12881.

For a few days now you can't build the nightly or source of OpenTTD, as a translator added an unicode character (ZWSP) to a translation. The translator is not to blame, eints should not have allowed it; but here we are.

Several developers noticed it, but nobody seems to have created the one-line yet. Must be the weather :)

## Description

I had to remove two characters in VSCode; not sure what they were exactly, as that wasn't showing. Hopefully I removed the correct characters. OpenTTD builds again at least :) I just have no clue if/how I altered the string.

Removing the string is not possible; eints doesn't allow reverting a string back to its past form, and this string was newly translated. So that would just mean eints would reintroduce it next push. By making a change, eints will pick it up next update, meaning also eints will see this new string. At least, that is what the [documentation](https://github.com/OpenTTD/OpenTTD/blob/master/docs/eints.md) seems to indicate.

And yes, eints should also be fixed to not allow this character in the first place, but that is for another day / PR / developer.

## Checklist for review

### DIFF
--- a/src/lang/traditional_chinese.txt
+++ b/src/lang/traditional_chinese.txt
@@ -3308,7 +3308,7 @@ STR_MAPGEN_DATE_TOOLTIP                                         :{BLACK}選擇�
 STR_MAPGEN_NUMBER_OF_INDUSTRIES                                 :{BLACK}工業數量：
 STR_MAPGEN_NUMBER_OF_INDUSTRIES_TOOLTIP                         :{BLACK}選擇工業密度，或輸入自訂數字
 STR_MAPGEN_HEIGHTMAP_HEIGHT                                     :{BLACK}最高峰：
-STR_MAPGEN_HEIGHTMAP_HEIGHT_TOOLTIP                             :{BLACK}選擇遊戲版圖內最高峰的海拔，​​以海平面為基準計算
+STR_MAPGEN_HEIGHTMAP_HEIGHT_TOOLTIP                             :{BLACK}選擇遊戲版圖內最高峰的海拔，以海平面為基準計算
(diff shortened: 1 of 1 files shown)


michicc (approved):

nielsmh:
I could imagine that the character was intended to be a linebreak hint. I don't think our current text rendering has any explicit way to request line breaks in specific places for writing systems that don't customarily use spaces, other than adding a hard line break?

PR #12890, opened 2024-08-03
Change: [NewGRF] Place all 8 bits of station tile layout in var 40/41/46/47/49.

PeterN:
## Motivation / Problem

#12554 allows station tile IDs up to 255 to be placed statically, but var 40/41/46/47/49 only includes the lower 4 bits of it.

This also applies to var 10 (used during CB 24), where the station doesn't exist.

Fortunately there is enough space in these vars to increase to 8 bits, which allows the full range of station tile IDs to be reported.

## Description

This change uses 8 bits of the return value for station tile ID instead of 4 bits.

## Checklist for review

### DIFF
--- a/src/newgrf_station.cpp
+++ b/src/newgrf_station.cpp
@@ -125,7 +125,7 @@ uint32_t GetPlatformInfo(Axis axis, uint8_t tile, int platforms, int length, int
 	}
 	SB(retval, 16, 4, std::min(15, length));
 	SB(retval, 20, 4, std::min(15, platforms));
-	SB(retval, 24, 4, tile);
+	SB(retval, 24, 8, tile);
 
 	return retval;
 }


TheGwyd:
Reworked the extra tile type GRF to use CB24 to set the layouts of different tiles to a unique value depending on their relative position in the newly build part. CB14 then tests the tile type using the high byte of var41 to select the tile layout. There are 3 stations in this set. One uses CB24 only. The second uses CB24 and CB14 to achieve theoretically the same result as the first station. The third is the same as the second except that it tests only the high 4 bits of var41 to ensure that CB14 is doing something. Source code included so you can check my workings
[256tiletype.zip](https://github.com/user-attachments/files/16487020/256tiletype.zip)

rubidium42 (approved):
Requires an update of https://newgrf-specs.tt-wiki.net/wiki/VariationalAction2/Stations#Platform_info_.2840.2C_41.2C_46.2C_47.2C_49.29 I think.

Not sure if the NewGRF spec tracker needs to be updated as this is a returned value, but there might be something in NML that exposes this? I haven't actively looked though.

PR #12894, opened 2024-08-05
Change: Sort list of cargoes in industry view and fund industry windows.

PeterN:
## Motivation / Problem

Industries lists in the Industry View window and the Fund Industry window are listed in the order they are defined in the industry accepts or produced data.

This means they are not sorted alphabetically (like in other places) and depend only on the definition.

Authors can try sort them but this does not work with different languages.

image

## Description

When displaying accepted and produced cargo lists, sort in the appropriate sorted cargo order (passengers first, then mail, then other cargo types)

image

## Limitations

Pitchforks.

## Checklist for review

### DIFF
--- a/src/industry_gui.cpp
+++ b/src/industry_gui.cpp
@@ -377,8 +377,15 @@ class BuildIndustryWindow : public Window {
 		size_t numcargo = 0;
 		size_t firstcargo = 0;
 
-		for (size_t j = 0; j < cargolist.size(); j++) {
-			if (!IsValidCargoID(cargolist[j])) continue;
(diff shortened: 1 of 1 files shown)


andythenorth:
> But these are not random orders, but deliberate. In FIRS, AXIS, SPI, ECS and probably other sets, these are orders of importance of resources and how they affect the production level 

For FIRS, this is true only in some cases, and generally only by coincidence.

I requested the change, for FIRS, so let's assume I know what my intent was?

Moth-Tolias:
if andy wants his own cargoes to be sorted alphabetically then that's his choice to do so, but i would like to retain the ability to order mine as i see fit. i don't see a way to do that with this change.

PR #12898, opened 2024-08-13
Codechange: Store custom station layouts in a map instead of nested vectors.

PeterN:
## Motivation / Problem

Custom station layouts are stored in 3 layers of nested vectors, one for platforms, one for lengths, and finally the layout data itself.

This requires work to manage as each additional length of platform count requires resizing vectors, and searching for a specific size needs check if all 3 layers are in range and have data.

## Description

Replace with an `unordered_map` of vectors. The map key is the platforms and length combined, each of which is a `uint8_t`, so the key is a `uint16_t`.

This simplifies allocation and searching for layouts. Allocation is automatic by key, and searching for a specific size only needs to find the key.

## Limitations

I did consider making the key a `std::pair<uint8_t, uint8_t>` but this isn't hashable by default.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -2042,14 +2042,14 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 
 					if (length == 0 || number == 0) break;
 
-					if (statspec->layouts.size() < length) statspec->layouts.resize(length);
-					if (statspec->layouts[length - 1].size() < number) statspec->layouts[length - 1].resize(number);
+					const uint8_t *buf_layout = buf.ReadBytes(length * number);
 
-					const uint8_t *layout = buf.ReadBytes(length * number);
-					statspec->layouts[length - 1][number - 1].assign(layout, layout + length * number);
+					/* Create entry in layouts and assign the layout to it. */
+					auto &layout = statspec->layouts[GetStationLayoutKey(number, length)];
+					layout.assign(buf_layout, buf_layout + length * number);
 
 					/* Ensure the first bit, axis, is zero. The rest of the value is validated during rendering, as we don't know the range yet. */
-					for (auto &tile : statspec->layouts[length - 1][number - 1]) {
+					for (auto &tile : layout) {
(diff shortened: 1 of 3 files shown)


src/newgrf.cpp
```diff
@@ -2042,14 +2042,14 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
...
 					if (length == 0 || number == 0) break;
 
-					if (statspec->layouts.size() < length) statspec->layouts.resize(length);
-					if (statspec->layouts[length - 1].size() < number) statspec->layouts[length - 1].resize(number);
-
 					const uint8_t *layout = buf.ReadBytes(length * number);
-					statspec->layouts[length - 1][number - 1].assign(layout, layout + length * number);
+
+					/* Create entry in layouts and assign the layout to it. */
+					auto &l = statspec->layouts[(length << 8) | number];
```
LordAro: better variable name?

src/newgrf_station.h
```diff
@@ -178,7 +178,7 @@ struct StationSpec : NewGRFSpecBase<StationClassID> {
 	 * These can be sparsely populated, and the upper limit is not defined but
 	 * limited to 255.
 	 */
-	std::vector<std::vector<std::vector<uint8_t>>> layouts;
+	std::unordered_map<uint16_t, std::vector<uint8_t>> layouts;
```
LordAro, line 174: wonder if it would be overkill to add some getter/setters to make it clear what the key is

(also the doc comment could do with an update)

PeterN: Added a function to determine the key. Mostly lots of lines due to doxygen.

rubidium42 (approved):
Is a next PR going to be that `GetStationLayout` returns (or passes) a `std::vector` for the layout? That could make it even simpler I'd guess, as you can then just return the found vector.

PeterN:
> Is a next PR going to be that `GetStationLayout` returns (or passes) a `std::vector` for the layout? That could make it even simpler I'd guess, as you can then just return the found vector.

It wasn't on my radar but it could be done.

Issue #12902, opened 2024-08-18
[Crash]: Crash while adding cars to trains


cetinbas:
### Version of OpenTTD

14.1, Windows 11

### Steps to reproduce

Sent trains 1 by 1 to depot to add more coal cars. Worked for 4-5 trains initially. 

### Upload crash files

[crash20240818090427.zip](https://github.com/user-attachments/files/16647810/crash20240818090427.zip)

PeterN:
Another "WndProcGdi needs to be re-entrant" issue.

glx22:
``` 	[Cadre en ligne] openttd.exe!google_breakpad::ExceptionHandler::WriteMinidump() Ligne 747	C++
 	openttd.exe!google_breakpad::ExceptionHandler::WriteMinidump(const std::wstring & dump_path, bool(*)(const wchar_t *, const wchar_t *, void *, _EXCEPTION_POINTERS *, MDRawAssertionInfo *, bool) callback, void * callback_context, _MINIDUMP_TYPE dump_type) Ligne 773	C++
 	openttd.exe!CrashLogWindows::WriteCrashDump() Ligne 113	C++
 	[Cadre en ligne] openttd.exe!std::_Func_class<bool>::operator()() Ligne 854	C++
 	openttd.exe!CrashLogWindows::TryExecute(std::basic_string_view<char,std::char_traits<char>> section_name, std::function<bool __cdecl(void)> && func) Ligne 124	C++
 	openttd.exe!CrashLog::MakeCrashLog() Ligne 294	C++
 	openttd.exe!ExceptionHandler(_EXCEPTION_POINTERS * ep) Ligne 341	C++
 	[Code externe]	
>	openttd.exe!VehicleViewWindow::VehicleViewWindow(WindowDesc * desc, int window_number) Ligne 2991	C++
 	openttd.exe!AllocateWindowDescFront<VehicleViewWindow>(WindowDesc * desc, int window_number, bool return_existing) Ligne 1013	C++
 	[Cadre en ligne] openttd.exe!DepotWindow::DepotClick(int) Ligne 572	C++
 	openttd.exe!DepotWindow::OnClick(Point pt, int widget, int click_count) Ligne 764	C++
 	openttd.exe!DispatchLeftClickEvent(Window * w, int x, int y, int click_count) Ligne 733	C++
 	openttd.exe!MouseLoop(MouseClick click, int mousewheel) Ligne 2844	C++
[...]

PR #12907, opened 2024-08-21
Fix: DepotWindow::GetVehicleFromDepotWndPt not refreshing vehicle list

JGRennison:
## Motivation / Problem

DepotWindow::GetVehicleFromDepotWndPt does not refresh the vehicle lists when generate_list is set.

Currently, the only path for refreshing the lists is in DepotWindow::OnPaint, however window repainting is done after input handling such as clicks which can result in calls to GetVehicleFromDepotWndPt.

Calls to GetVehicleFromDepotWndPt via input events can therefore use an out of date vehicle list potentially containing stale pointers.
This seems a plausible failure path for #12902.

## Description

GetVehicleFromDepotWndPt should refresh the list itself if generate_list is set.
Calling ReInit due to unit number width changes should still only be done via OnPaint.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/depot_gui.cpp
+++ b/src/depot_gui.cpp
@@ -258,6 +258,7 @@ struct DepotWindow : Window {
 	VehicleID vehicle_over; ///< Rail vehicle over which another one is dragged, \c INVALID_VEHICLE if none.
 	VehicleType type;
 	bool generate_list;
+	bool check_unitnumber_digits;
 	WidgetID hovered_widget; ///< Index of the widget being hovered during drag/drop. -1 if no drag is in progress.
 	VehicleList vehicle_list;
 	VehicleList wagon_list;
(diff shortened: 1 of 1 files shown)


src/depot_gui.cpp
```diff
@@ -715,6 +719,16 @@ struct DepotWindow : Window {
...
+			this->check_unitnumber_digits = true;
+		}
+	}
+
+	void OnPaint() override
+	{
+		this->RefreshVehicleList();
+
+		if (this->check_unitnumber_digits) {
+			this->check_unitnumber_digits = false;
```
Berbe, line 731: If another call to `RefreshVehicleList()` happens right after this line, the problem trying to be prevented reappears.

This looks very much like a mutually-exclusive access is required: this calls for a mutex, semaphores, or any other lock system on the vehicles list, doesn't it?

PeterN: OpenTTD is not multithreaded so that does not seem relevant.

PeterN:
Working on a list that is different than displayed, it may then be possible that something happens to a different vehicle than intended -- effectively the position pointed to is invalidated.

It may be better to make this work with VehicleID instead, which can be validated.

rubidium42 (approved):

PR #12908, opened 2024-08-21
Fix #12901: Savegame format of company allow lists

### LINKED ISSUE #12901: [Bug]: Savegame storage of CompanyProperties::allow_list results in invalid table format
### Version of OpenTTD

Since #12337

### Expected result

Savegame chunks which use the new table format meet the format requirements

### Actual result

Savegames where CompanyProperties::allow_list is non-empty can't be parsed by savegame reader tools like https://github.com/TrueBrain/OpenTTD-savegame-reader, because the field as saved does not use the required format.
If the field is not present when loading, the discard unknown table field mechanism does not work with this field when it is non-empty.

### Steps to reproduce

Create a savegame which has an allow public key (e.g. start a new multiplayer server).
Look at the savegame with https://github.com/TrueBrain/OpenTTD-savegame-reader or an equivalent tool.

JGRennison:
## Motivation / Problem

#12901: Savegame format of company allow lists .

## Description

Use sub-struct list for company allow lists.
Disallow saving of invalid string container table fields.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/saveload/saveload.cpp
+++ b/src/saveload/saveload.cpp
@@ -1246,17 +1246,13 @@ public:
 	 */
 	static size_t SlCalcLen(const void *storage, VarType conv, SaveLoadType cmd = SL_VAR)
 	{
-		assert(cmd == SL_VAR || cmd == SL_REF || cmd == SL_STDSTR);
+		assert(cmd == SL_VAR || cmd == SL_REF);
 
 		const SlStorageT *list = static_cast<const SlStorageT *>(storage);
 
-		size_t type_size = SlGetArrayLength(list->size());
-		if constexpr (std::is_same_v<std::vector<std::string>, SlStorageT> || std::is_same_v<std::deque<std::string>, SlStorageT>) {
-			return std::accumulate(list->begin(), list->end(), type_size, [](size_t sum, const std::string &str) { return sum + SlCalcStdStringLen(&str); });
-		} else {
-			size_t item_size = SlCalcConvFileLen(cmd == SL_VAR ? conv : (VarType)SLE_FILE_U32);
-			return list->size() * item_size + type_size;
-		}
+		int type_size = SlGetArrayLength(list->size());
+		int item_size = SlCalcConvFileLen(cmd == SL_VAR ? conv : (VarType)SLE_FILE_U32);
+		return list->size() * item_size + type_size;
 	}
 
 	static void SlSaveLoadMember(SaveLoadType cmd, Tvar *item, VarType conv)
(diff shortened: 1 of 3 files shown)


PeterN:
I'm confused as to what makes a `SL_...VECTOR` of `SLE_STR` special and not valid.

JGRennison:
In the new table format, a string is a length-prefixed sequence of chars.
i.e. the field type is SLE_FILE_STRING | SLE_FILE_HAS_LENGTH_FIELD.

The total length of the field is therefore the size of the length prefix itself plus the length value.

It is not valid to put a second length prefix in front, but still label the field as SLE_FILE_STRING | SLE_FILE_HAS_LENGTH_FIELD.
In #12337 the outer length prefix is not actually a length prefix, but a number of sub-items, it does not match the number of bytes which follow in the field.
As the field is still advertised as a string with one length prefix, readers are entitled to incorrectly read this length prefix as the number of chars in the string.

Potentially the table format could be extended to support fields with nested length prefixes, with a new field type for this, but it seems more pragmatic to use SLE_FILE_STRUCT | SLE_FILE_HAS_LENGTH_FIELD along with a string field in the sub-table, which seems the intended way that the format is meant to be used.

PeterN:
Ah, so both `SL_...VECTOR` and `SLE_STR` are using the same flag `SLE_FILE_HAS_LENGTH_FIELD` in the same place, which only works because `_company_desc` knows what the record is.

src/saveload/saveload.cpp
```diff
@@ -1368,7 +1364,6 @@ static inline size_t SlCalcDequeLen(const void *deque, VarType conv)
 		case SLE_VAR_U32: return SlStorageHelper<std::deque, uint32_t>::SlCalcLen(deque, conv);
 		case SLE_VAR_I64: return SlStorageHelper<std::deque, int64_t>::SlCalcLen(deque, conv);
 		case SLE_VAR_U64: return SlStorageHelper<std::deque, uint64_t>::SlCalcLen(deque, conv);
-		case SLE_VAR_STR: return SlStorageHelper<std::deque, std::string>::SlCalcLen(deque, conv, SL_STDSTR);
```
rubidium42, line 1371: Might be useful to retain the `case` with documentation that no  variable-sized elements may be in the `deque`, and similar to the variant in `SlCalcVectorLen`.
Maybe something along the lines of: `case SLE_VAR_STR: // Variable-sized elements require special handling for the save game to follow the specifications. Use a struct list instead.` should "scare" us from doing it again. And it would make it much easier to prevent such things in future reviews.

rubidium42 (approved):

PR #12918, opened 2024-08-23
Fix #12914: Fix use of invalidated pointer in viewport drawer.

### LINKED ISSUE #12914: [Bug]: AddChildSpriteToFoundation can cause a use after free via ViewportDrawer::last_child
### Version of OpenTTD

master

### Expected result

AddChildSpriteToFoundation can't cause a use-after-free via ViewportDrawer::last_child

### Actual result

AddChildSpriteToFoundation saves the current value of ViewportDrawer::last_child which points into the ViewportDrawer::child_screen_sprites_to_draw std::vector, then calls AddChildSpriteScreen which appends to this vector.
AddChildSpriteScreen handles updates ViewportDrawer::last_child, so it is still valid when ViewportDrawer::child_screen_sprites_to_draw reallocates.

However AddChildSpriteToFoundation then restores the old value of ViewportDrawer::last_child which is a dangling pointer if ViewportDrawer::child_screen_sprites_to_draw was reallocated.

Subsequent calls to AddChildSpriteScreen can then cause a use after free via this dangling pointer.

### Steps to reproduce

See above

PeterN:
## Motivation / Problem

As per #12914, use of invalid dangling pointer due to vector reallocation in viewport drawer.

## Description

~~Store destination list and index within the list, instead of a direct pointer, to find child index to update.~~

Use index of last child instead of pointer to update next_child element.

In case there is no child sprite yet, the most recent parent sprite's first_child is updated instead.

## Limitations

I haven't been able to prove that child sprites are only even added to the most recent parent sprite.

Switching to std::deque is probably less intrusive but not sure about iteration performance.

## Checklist for review

### DIFF
--- a/src/viewport.cpp
+++ b/src/viewport.cpp
@@ -161,6 +161,9 @@ typedef std::vector<StringSpriteToDraw> StringSpriteToDrawVector;
 typedef std::vector<ParentSpriteToDraw> ParentSpriteToDrawVector;
 typedef std::vector<ChildScreenSpriteToDraw> ChildScreenSpriteToDrawVector;
 
+constexpr int LAST_CHILD_NONE = -1; ///< There is no last_child to fill.
+constexpr int LAST_CHILD_PARENT = -2; ///< Fill last_child of the most recent parent sprite.
+
 /** Data structure storing rendering information */
 struct ViewportDrawer {
 	DrawPixelInfo dpi;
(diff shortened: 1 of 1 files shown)


src/viewport.cpp
```diff
@@ -612,7 +610,7 @@ void OffsetGroundSprite(int x, int y)
 	}
 
 	/* _vd.last_child == nullptr if foundation sprite was clipped by the viewport bounds */
```
glx22: This comment is outdated.

src/viewport.cpp
```diff
@@ -831,15 +829,20 @@ void AddChildSpriteScreen(SpriteID image, PaletteID pal, int x, int y, bool tran
...
 
 	/* make the sprites transparent with the right palette */
 	if (transparent) {
 		SetBit(image, PALETTE_MODIFIER_TRANSPARENT);
 		pal = PALETTE_TO_TRANSPARENT;
 	}
 
-	*_vd.last_child = (uint)_vd.child_screen_sprites_to_draw.size();
+	int32_t child_id = static_cast<int32_t>(_vd.child_screen_sprites_to_draw.size());
+	if (_vd.last_child >= 0) {
```
rubidium42: I'm a bit conflicted about this line. On its own it looks very reasonable to guard indexing into an array with a non-negative value. However, `LAST_CHILD_PARENT` is only set and never read, and here the "meaning" seems to be `if (_vd.last_child != LAST_CHILD_PARENT)`. That'd make me think... how is `LAST_CHILD_NONE` then handled as that's a negative value, just to discover that if that's the case you leave the function a couple of lines earlier.

So yeah, not sure. Either question where `LAST_CHILD_PARENT` is read or question why `LAST_CHILD_NONE` is not checked for in this `if`.

rubidium42 (dismissed):

src/viewport.cpp
PeterN: Decided to go with `!= LAST_CHILD_PARENT` as that is the actual intention and it took me a moment to register that already.

If it was for bounds checking then `>= 0` is not a sufficient check, and it would not have the correct result either.

rubidium42 (approved):

PR #12919, opened 2024-08-24
Codechange: [NewGRF] Use reference to unique_ptr instead of .get()

PeterN:
## Motivation / Problem

Unnecessary use of unique_ptr .get()

## Description

Use reference to unique_ptr instead of .get() when loading NewGRF spec data.

This allows assignment to reference when creating a new spec instead of setting and refetching.

## Checklist for review

### DIFF
--- a/src/newgrf.cpp
+++ b/src/newgrf.cpp
@@ -1932,7 +1932,7 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 	if (_cur.grffile->stations.size() < stid + numinfo) _cur.grffile->stations.resize(stid + numinfo);
 
 	for (int i = 0; i < numinfo; i++) {
-		StationSpec *statspec = _cur.grffile->stations[stid + i].get();
+		auto &statspec = _cur.grffile->stations[stid + i];
 
 		/* Check that the station we are modifying is defined. */
 		if (statspec == nullptr && prop != 0x08) {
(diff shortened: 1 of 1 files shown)


src/newgrf.cpp
```diff
@@ -2173,7 +2172,7 @@ static ChangeInfoResult StationChangeInfo(uint stid, int numinfo, int prop, Byte
 				break;
 
 			case 0x1D: // Station Class name
-				AddStringForMapping(buf.ReadWord(), [statspec](StringID str) { StationClass::Get(statspec->class_index)->name = str; });
+				AddStringForMapping(buf.ReadWord(), [statspec = statspec.get()](StringID str) { StationClass::Get(statspec->class_index)->name = str; });
```
rubidium42, line 2175: I do not know whether there is a reason why you wouldn't want to pass the reference by-reference here as well. Just wondering, as I'm  far from sure that it's actually the better solution. Maybe some C++ guru can give their few bits of currency ;)
```suggestion
				AddStringForMapping(buf.ReadWord(), [&statspec](StringID str) { StationClass::Get(statspec->class_index)->name = str; });
```

PeterN: AddStringForMapping saves the lambda for later. The reference only exists for the lifetime of this function, whereas the mapping is done later. Therefore the lambda needs to have its own copy of the pointer.

rubidium42 (approved):

PR #12923, opened 2024-08-31
Codechange: Use std::variant for string parameters and prevent mixing string/numeric parameters.

PeterN:
## Motivation / Problem

`StringParameter` contains both a `uint64_t` value, and a `std::unique_ptr<std::string>`. Which one is read from depends on whether the caller requested a number or a string, and reading from the wrong type does not warn.

`StringParameterData` is similar class to `StringParameter`, but not quite the same. Creating or restoring a backup of string parameter is quite involved.

## Description

This PR changes `StringParameter` to use a `std::variant` to hold either a `uint64_t` or `std::string`. This is aliased to `StringParameterData` for ease of use.

There is no longer any need for the `StringParameterBackup` class.

Backup/restore of string parameter is simplified as std::variant itself takes care of the details.

Reading a numeric parameter as a string, or a string parameter as a number, is now treated as a string error instead of silently using zero.

## Limitations

We should've used `std::variant` originally when string parameters were revamped, but apparently we were hesitant... although we also handled raw `char *` back then too, which has since been removed which simplifies things.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -191,7 +191,7 @@ void CopyOutDParam(std::vector<StringParameterBackup> &backup, size_t num)
  * @param backup The backup to check against.
  * @return True when the parameters have changed, otherwise false.
  */
-bool HaveDParamChanged(const std::vector<StringParameterBackup> &backup)
+bool HaveDParamChanged(const std::span<const StringParameterBackup> backup)
 {
 	bool changed = false;
 	for (size_t i = 0; !changed && i < backup.size(); i++) {
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -76,9 +76,9 @@ void StringParameters::PrepareForNextRun()
  * Get the next parameter from our parameters.
  * This updates the offset, so the next time this is called the next parameter
  * will be read.
- * @return The pointer to the next parameter.
+ * @return The next parameter.
  */
-StringParameter *StringParameters::GetNextParameterPointer()
+const StringParameter &StringParameters::GetNextParameterReference()
 {
 	assert(this->next_type == 0 || (SCC_CONTROL_START <= this->next_type && this->next_type <= SCC_CONTROL_END));
 	if (this->offset >= this->parameters.size()) {
@@ -92,7 +92,7 @@ StringParameter *StringParameters::GetNextParameterPointer()
 	}
 	param.type = this->next_type;
 	this->next_type = 0;
-	return &param;
+	return param;
 }
 
 
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -113,7 +113,7 @@ void SetDParam(size_t n, uint64_t v)
  */
 uint64_t GetDParam(size_t n)
 {
-	return _global_string_params.GetParam(n);
+	return std::get<uint64_t>(_global_string_params.GetParam(n));
 }
 
 /**
@@ -156,15 +156,10 @@ void SetDParamMaxDigits(size_t n, uint count, FontSize size)
  * Copy the parameters from the backup into the global string parameter array.
  * @param backup The backup to copy from.
  */
-void CopyInDParam(const std::span<const StringParameterBackup> backup)
+void CopyInDParam(const std::span<const StringParameterData> backup)
 {
 	for (size_t i = 0; i < backup.size(); i++) {
-		auto &value = backup[i];
-		if (value.string.has_value()) {
-			_global_string_params.SetParam(i, value.string.value());
-		} else {
-			_global_string_params.SetParam(i, value.data);
-		}
+		_global_string_params.SetParam(i, backup[i]);
 	}
 }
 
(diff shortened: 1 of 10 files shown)


src/strings_internal.h
```diff
@@ -92,8 +91,10 @@ class StringParameters {
 	template <typename T>
 	T GetNextParameter()
 	{
-		auto ptr = GetNextParameterPointer();
-		return static_cast<T>(ptr->data);
+		const auto &param = GetNextParameterReference();
+		/* Check that we hold a uint64_t. std::get_if is used as std::get would check again. */
+		if (std::holds_alternative<uint64_t>(param.data)) return static_cast<T>(*std::get_if<uint64_t>(&param.data));
+		throw std::runtime_error("Attempt to read string paarameter as integer");
```
glx22:
```suggestion
		throw std::runtime_error("Attempt to read string parameter as integer");
```

src/strings_internal.h
```diff
@@ -104,8 +105,10 @@ class StringParameters {
 	 */
 	const char *GetNextParameterString()
 	{
-		auto ptr = GetNextParameterPointer();
-		return ptr->string != nullptr ? ptr->string->c_str() : nullptr;
+		const auto &param = GetNextParameterReference();
+		/* Check that we hold a string. std::get_if is used as std::get would check again. */
+		if (std::holds_alternative<std::string>(param.data)) return std::get_if<std::string>(&param.data)->c_str();
+		throw std::runtime_error("Attempt to read integer paarameter as string");
```
glx22:
```suggestion
		throw std::runtime_error("Attempt to read integer parameter as string");
```

src/strings_internal.h
```diff
@@ -92,8 +91,10 @@ class StringParameters {
 	template <typename T>
 	T GetNextParameter()
 	{
-		auto ptr = GetNextParameterPointer();
-		return static_cast<T>(ptr->data);
+		const auto &param = GetNextParameterReference();
+		/* Check that we hold a uint64_t. std::get_if is used as std::get would check again. */
+		if (std::holds_alternative<uint64_t>(param.data)) return static_cast<T>(*std::get_if<uint64_t>(&param.data));
+		throw std::runtime_error("Attempt to read string parameter as integer");
```
rubidium42: Doesn't `std::get_if` implicitly check again, because *if* the type is wrong it will return `nullptr`. As such, it would be doing some checks to determine whether to return `nullptr`. Maybe something like the following would be simpler:
```suggestion
		uint64_t *data = std::get_if<uint64_t>(&param.data);
		if (data != nullptr) return static_cast<T>(*data);
		throw std::runtime_error("Attempt to read string parameter as integer");
```

src/strings_internal.h
```diff
@@ -104,8 +105,10 @@ class StringParameters {
 	 */
 	const char *GetNextParameterString()
 	{
-		auto ptr = GetNextParameterPointer();
-		return ptr->string != nullptr ? ptr->string->c_str() : nullptr;
+		const auto &param = GetNextParameterReference();
+		/* Check that we hold a string. std::get_if is used as std::get would check again. */
+		if (std::holds_alternative<std::string>(param.data)) return std::get_if<std::string>(&param.data)->c_str();
+		throw std::runtime_error("Attempt to read integer parameter as string");
```
rubidium42: Same idea as the integer variant.

src/strings.cpp
```diff
@@ -271,14 +254,24 @@ void GetStringWithArgs(StringBuilder &builder, StringID string, StringParameters
 	switch (tab) {
 		case TEXT_TAB_TOWN:
 			if (index >= 0xC0 && !game_script) {
-				GetSpecialTownNameString(builder, index - 0xC0, args.GetNextParameter<uint32_t>());
+				try {
+					GetSpecialTownNameString(builder, index - 0xC0, args.GetNextParameter<uint32_t>());
+				} catch (std::runtime_error &e) {
```
rubidium42:
```suggestion
				} catch (const std::runtime_error &e) {
```

src/strings.cpp
```diff
@@ -271,14 +254,24 @@ void GetStringWithArgs(StringBuilder &builder, StringID string, StringParameters
...
 				return;
 			}
 			break;
 
 		case TEXT_TAB_SPECIAL:
 			if (index >= 0xE4 && !game_script) {
-				GetSpecialNameString(builder, index - 0xE4, args);
+				try {
+					GetSpecialNameString(builder, index - 0xE4, args);
+				} catch (std::runtime_error &e) {
```
rubidium42:
```suggestion
				} catch (const std::runtime_error &e) {
```

src/strings.cpp
```diff
@@ -191,18 +181,11 @@ void CopyOutDParam(std::vector<StringParameterBackup> &backup, size_t num)
...
 	for (size_t i = 0; !changed && i < backup.size(); i++) {
-		bool global_has_string = _global_string_params.GetParamStr(i) != nullptr;
-		if (global_has_string != backup[i].string.has_value()) return true;
-
-		if (global_has_string) {
-			changed = backup[i].string.value() != _global_string_params.GetParamStr(i);
-		} else {
-			changed = backup[i].data != _global_string_params.GetParam(i);
-		}
+		changed = backup[i] != _global_string_params.GetParam(i);
```
rubidium42: The whole `changed` variable seems a bit pointless as we're essentially short cutting anyway. Not sure what I might've been thinking when making the function as it was though...
```suggestion
		if (backup[i] != _global_string_params.GetParam(i)) return true;
```
Remove `bool changed` and check in the loop, and change the `return` after the loop to just return `false`?

rubidium42 (approved):

PR #12938, opened 2024-09-14
Add: Overlay cargo icon in vehicle/depot list when holding shift+ctrl.

PeterN:
## Motivation / Problem

Dunno, I felt like doing something else...

## Description

Overlay cargo icons on top of vehicles in vehicle and depot lists. Overlay icons are shown only when shift+ctrl are pressed.

image
image
image

## Limitations

It's probably silly and pointless.

~~Drawing an outline uses a 256-entry palette remap, similar to how text colours are done, to draw all non-transparent pixels as a single colour. This won't work for 32bpp sprites.~~

Now just uses PALETTE_ALL_BLACK which doesn't need any new palette code.

## Checklist for review

### DIFF
--- a/src/depot_gui.cpp
+++ b/src/depot_gui.cpp
@@ -974,6 +974,15 @@ struct DepotWindow : Window {
 		}
 	}
 
+	bool last_overlay_state;
+	void OnMouseLoop() override
+	{
+		if (last_overlay_state != ShowCargoIconOverlay()) {
+			last_overlay_state = ShowCargoIconOverlay();
+			this->SetDirty();
+		}
+	}
+
 	void OnMouseDrag(Point pt, WidgetID widget) override
 	{
 		if (this->sel == INVALID_VEHICLE) return;
--- a/src/ship_gui.cpp
+++ b/src/ship_gui.cpp
@@ -44,6 +44,7 @@ void DrawShipImage(const Vehicle *v, const Rect &r, VehicleID selection, EngineI
 	int y = ScaleSpriteTrad(-1) + CenterBounds(r.top, r.bottom, 0);
 
 	seq.Draw(x, y, GetVehiclePalette(v), false);
+	if (v->cargo_cap > 0) DrawCargoIconOverlay(x, y, v->cargo_type);
 
 	if (v->index == selection) {
 		x += x_offs;
--- a/src/vehicle_gui.cpp
+++ b/src/vehicle_gui.cpp
@@ -329,6 +329,82 @@ static bool CargoFilter(const GUIVehicleGroup *vehgroup, const CargoID cid)
 	return false;
 }
 
+/**
+ * Test if cargo icon overlays should be drawn.
+ * @returns true iff cargo icon overlays should be drawn.
+ */
+bool ShowCargoIconOverlay()
+{
+	return _shift_pressed && _ctrl_pressed;
+}
+
+/**
+ * Add a cargo icon to the list of overlays.
+ * @param overlays List of overlays.
+ * @param x Horizontal position.
+ * @param width Width available.
+ * @param v Vehicle to add.
+ */
+void AddCargoIconOverlay(std::vector<CargoIconOverlay> &overlays, int x, int width, const Vehicle *v)
+{
+	bool rtl = _current_text_dir == TD_RTL;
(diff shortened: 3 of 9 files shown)


LordAro:
Is there anything existing that does something distinct with ctrl+shift ? (rather than a combination of ctrl behaviour + shift behaviour)

PeterN:
No, shift only ever had two uses, estimating cost, and speeding up keyboard viewport scrolling. There's very few tests for `_shift_pressed` in the code.

PeterN:
With black outlines instead. Advantage of black is that PALETTE_ALL_BLACK already exists and works with 32bpp sprites.

image
image

2TallTyler:
I like the look of the black outlines better, and they're more consistent with our visual style as they match text shadows.

src/depot_gui.cpp
```diff
@@ -974,6 +974,15 @@ struct DepotWindow : Window {
 		}
 	}
 
+	bool last_state;
+	void OnMouseLoop() override
+	{
+		if (last_state != (_shift_pressed & _ctrl_pressed)) {
```
2TallTyler: These shouldn't be bitwise operators, right?

```suggestion
		if (last_state != (_shift_pressed && _ctrl_pressed)) {
```

src/depot_gui.cpp
```diff
@@ -974,6 +974,15 @@ struct DepotWindow : Window {
 		}
 	}
 
+	bool last_state;
```
2TallTyler: Could probably use a comment somewhere in here (either on `last_state` or inside `OnMouseLoop()` explaining why we mark the depot GUI dirty when shift and control are pressed or released.

src/ship_gui.cpp
```diff
@@ -44,6 +44,7 @@ void DrawShipImage(const Vehicle *v, const Rect &r, VehicleID selection, EngineI
 	int y = ScaleSpriteTrad(-1) + CenterBounds(r.top, r.bottom, 0);
 
 	seq.Draw(x, y, GetVehiclePalette(v), false);
+	if (v->cargo_cap > 0) DrawCargoIconOverlay(x, y, v->cargo_type);
```
2TallTyler, line 47: Why not pass a full reference to `v` to `DrawCargoIconOverlay()` and check the cargo capacity there as an early return?

src/vehicle_gui.cpp
```diff
@@ -329,6 +329,82 @@ static bool CargoFilter(const GUIVehicleGroup *vehgroup, const CargoID cid)
...
+		/* Add new overlay slot. */
+		overlays.emplace_back(rtl ? x - width : x, rtl ? x : x + width, v->cargo_type, v->cargo_cap);
+	} else {
+		/* This is an articulated part with the same cargo type, adjust left or right of last overlay slot. */
+		if (rtl) {
+			overlays.back().left -= width;
+		} else {
+			overlays.back().right += width;
+		}
+		overlays.back().cargo_cap += v->cargo_cap;
```
2TallTyler, line 361: If the vehicle's capacity is 0, the overlay doesn't get drawn. Why add this unused overlay to the vector at all? 

If we need to track it (I assume it's related to articulated vehicles), why store the actual cargo capacity over a bool to mark the segment as not having an icon? (I might be over-optimizing here.)

PeterN: It is indeed added so that if an articulated part is present and is the same cargo type we can extend the slot from the first part of the articulated vehicle.

While it is possible to test if a part has an articulated part, and then add or not add accordingly, and then maybe remove an entry if it turns out the articulated vehicle has no cargo... the code becomes quite a bit more complex.

The overlays vector is marked static to avoid so that the reserved memory is reused and does not need fresh allocations every time.

The capacity could just be stored as a bool, but simply adding capacity saves me from caring about conditionals until it comes to drawing. Potentially less branch prediction happening, but not really my primary thought...

2TallTyler: Works for me!

src/ship_gui.cpp
PeterN: Then `DrawCargoIconOverlays()` wouldn't be able to call it.

2TallTyler (approved):
Sure, why not?

PR #12943, opened 2024-09-16
Fix #12940: Use specific error when overbuilding station on signals

### LINKED ISSUE #12940: [Bug]: When station building is blocked by the presence of a signal, the wrong error text is shown
### Version of OpenTTD

20240915

### Expected result

The error text "Must remove signals first" is shown.

### Actual result

The error text "Must remove railway track first" is shown.

### Steps to reproduce

1. build a railway track and place a signal.
2. attempt to overbuild the track with a station, in such a way that it fails only because of the presence of the signal.
3. read the error message.
Wrindingstone Transport, 8th Sep 1993

2TallTyler:
## Motivation / Problem

Per #12940, when overbuilding a station on rail with signals, the error "Must remove railway track first" is shown instead of the expected "Must remove signals first".

## Description

Check for signals and throw the existing error message if found.

Because `HasSignals()` assumes the tile is a plain rail tile, I refactored to avoid an extra call to `IsPlainRailTile()`, and was able to clarify the comments while I was at it.

Closes #12940.

## Checklist for review

### DIFF
--- a/src/station_cmd.cpp
+++ b/src/station_cmd.cpp
@@ -916,34 +916,32 @@ static CommandCost CheckFlatLandRailStation(TileIndex tile_cur, TileIndex north_
 			}
 		}
 	} else {
-		/* Rail type is only valid when building a railway station; if station to
-			* build isn't a rail station it's INVALID_RAILTYPE. */
-		if (rt != INVALID_RAILTYPE &&
-				IsPlainRailTile(tile_cur) && !HasSignals(tile_cur) &&
-				HasPowerOnRail(GetRailType(tile_cur), rt)) {
-			/* Allow overbuilding if the tile:
-				*  - has rail, but no signals
-				*  - it has exactly one track
-				*  - the track is in line with the station
-				*  - the current rail type has power on the to-be-built type (e.g. convert normal rail to el rail)
-				*/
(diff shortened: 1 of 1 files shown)


PikkaBird:
I'm wondering if the signal check shouldn't be moved further down, so that if there's a more significant problem (like incompatible track types, or the wrong station orientation) it will still complain about the track first.

It's probably a rare situation, but it feels like complaining specifically about the signal when there's also another issue blocking construction is presenting the player with a red herring.

PikkaBird:
> Wouldn't it be better

Sure, perhaps the signal could be auto-removed if the appropriate setting is turned on, but that's a different PR. Maybe you could start it?

(While you're at it, I'm sure a setting to demolish stations one tile at a time, like roadstops, instead of the demolish tool deleting all rail station tiles at once, would be much appreciated).

krysclarke:
> 
> When I want to build an additional track (junction), the game doesn't check if I'm using the right type, it doesn't pester me with messages about impossible track combinations every time. Recently, I don't have to constantly fight warnings when I want to build a new track (junction) in a place where there is a signal. The game doesn't warn me about it, it just removes it. And that was a great change! :)
> 
It does check, unless you turn on the setting "Automatically remove signals during rail construction":
image

LordAro (approved):

PR #12948, opened 2024-09-17
Fix: Crash when pressing arrow keys in bootstrap mode

JGRennison:
## Motivation / Problem

A crash occurs when pressing the arrow keys in bootstrap mode due to #12808.
GetMainWindow() is only usable when a main window is present.

## Description

Fix the above.

## Limitations

N/A

## Checklist for review

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -2761,8 +2761,10 @@ static void HandleKeyScrolling()
 	if (_dirkeys && !EditBoxInGlobalFocus()) {
 		int factor = _shift_pressed ? 50 : 10;
 
-		/* Key scrolling stops following a vehicle. */
-		GetMainWindow()->viewport->follow_vehicle = INVALID_VEHICLE;
+		if (_game_mode != GM_MENU && _game_mode != GM_BOOTSTRAP) {
+			/* Key scrolling stops following a vehicle. */
+			GetMainWindow()->viewport->follow_vehicle = INVALID_VEHICLE;
+		}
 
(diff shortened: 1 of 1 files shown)


rubidium42 (approved):

src/window.cpp
```diff
@@ -2761,8 +2761,10 @@ static void HandleKeyScrolling()
 	if (_dirkeys && !EditBoxInGlobalFocus()) {
 		int factor = _shift_pressed ? 50 : 10;
 
-		/* Key scrolling stops following a vehicle. */
-		GetMainWindow()->viewport->follow_vehicle = INVALID_VEHICLE;
+		if (_game_mode != GM_MENU && _game_mode != GM_BOOTSTRAP) {
```
2TallTyler, line 2764: The arrow keys already do nothing when we are in the menu. (Only moving the mouse stops camera following).

JGRennison: Yes, the contents of the if are not required for the GM_MENU and GM_BOOTSTRAP modes.

2TallTyler: Right, my point was that this check isn't required since GM_MENU is already handled.

JGRennison: The title game panning does not use `follow_vehicle`. It is not required to reset it. This has symmetry with the corresponding condition in ScrollMainViewport.

An equivalent condition of `_game_mode == GM_NORMAL || _game_mode == GM_EDITOR` would be better from first principles, but this is inconsistent with the existing conditions of this type.

2TallTyler: Aha, thanks. I guess I should have asked why you added the check instead of arguing against it. 😅

PR #12950, opened 2024-09-18
Fix fcf8693c4f: Remove 100 byte limit for parsing string command parameters

PeterN:
## Motivation / Problem

Since fcf8693c4f a plural command with long parameters has prevent language compilation.

This length restriction is due to using a fixed sized buffer when parsing string commands, and is not a language file binary format limitation.

## Description

Parse the command and parameters into a `std::string` instead of a fixed-size C-style array. This allows the buffer to grow. Because this is a build tool I've not given any consideration to buffer reallocations etc.

This allows longer parameters to be used in plural and gender commands.

Each individual word list parameter is now limited to 253 bytes, allowing for a trailing NUL and leaving 0xFF reserved. This is now explicitly checked to ensure the string parameters are not too long, as this was not needed with the fixed length parsing buffer.

## Checklist for review

### DIFF
--- a/src/strgen/strgen_base.cpp
+++ b/src/strgen/strgen_base.cpp
@@ -31,8 +31,7 @@ int _cur_line;                        ///< The current line we're parsing in the
 int _errors, _warnings, _show_todo;
 LanguagePackHeader _lang;             ///< Header information about a language.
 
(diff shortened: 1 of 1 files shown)


michicc (approved):
Can't see anything wrong.

PR #12956, opened 2024-09-21
Fix: ICUParagraphLayout line wrapping in the case where the overflow run is not the start run, and the start run had a partial offset

JGRennison:
## Motivation / Problem

Incorrect line wrapping in ICUParagraphLayout in the case where the overflow run is not the start run, and the start run had a partial offset.
This is due to partial_offset being applied to the overflow run even when it did not refer to that run.

image

## Description

Fix the above.

Result:
image

Test save:
(You will need to neuter StrMakeValid to stop the colour codes being removed on load).
[ICUParagraphLayout Ltd., 1990-02-10.sav.zip](https://github.com/user-attachments/files/17085580/ICUParagraphLayout.Ltd.1990-02-10.sav.zip)

## Limitations

N/A

## Checklist for review


PeterN:
Also fixes the this case I know of...
image
image

PeterN (approved):

PR #12960, opened 2024-09-22
Fix #12957: 1-in-X proportion of towns being cities is unnecessarily random.

### LINKED ISSUE #12957: [Bug]: "Proportion of towns that will become cities" is unnecessarily random
### Version of OpenTTD

20240922

### Expected result

By default, OpenTTD generates "1 in 4" towns as cities; towns which have an initial size boost and faster growth. It also sets a flag on the town, and It is not unusual for NewGRF authors to use that flag to differentiate between "cities" and "towns" for house and industry newgrf behaviour. 

Ideally, the proportion in the setting should be strictly adhered to; eg,  at 1 in 4, every fourth town generated is given the city flag.

### Actual result

However, rather than being strictly "1 in 4", a random probability of "city"hood is applied to each town. On large maps, with hundreds of towns, this randomness is hardly significant; but on smaller maps, with only a few towns, it becomes impossible for either GRF authors or players to anticipate the availability and proportion of cities on a randomly generated map.

### Steps to reproduce

On a map with 16 towns (an appropriate number for a 256x256 or 256x512 map) and the city setting at the default 1 in 4, a few generations produced results as varied as 9/16 - more than double the expected ratio - and 2/16 cities. With 8 towns, it only took a few generations to produce a map with zero cities, rather than the expected two.

Unnamed, 7th Jan 1950

PeterN:
## Motivation / Problem

Using an actual random chance to generate cities means that there may be far less or far more than 'expected' towns being made as cities.

While this is the point of randomness, the wording of the setting is "Proporation of towns that will become cities: 1 in X" and does not mention randomness at all.

## Description

Instead use a simple modulus test to ensure that actually 1 in X towns become cities.

This is adjusted by a small random offset so that 1-in-4 cities can mean e.g. 2 or 3 cities if there are 10 towns, instead of always 3.

## Checklist for review

### DIFF
--- a/src/town_cmd.cpp
+++ b/src/town_cmd.cpp
@@ -2395,11 +2395,15 @@ bool GenerateTowns(TownLayout layout)
 		town_names.insert(town->GetCachedName());
 	}
 
+	/* Randomised offset for city status. This means with e.g. 1-in-4 towns being cities, a map with 10 towns
+	 * may have 2 or 3 cities, instead of always 3. */
+	uint city_random_offset = Random() % _settings_game.economy.larger_towns;
+
 	/* First attempt will be made at creating the suggested number of towns.
 	 * Note that this is really a suggested value, not a required one.
 	 * We would not like the system to lock up just because the user wanted 100 cities on a 64*64 map, would we? */
 	do {
-		bool city = (_settings_game.economy.larger_towns != 0 && Chance16(1, _settings_game.economy.larger_towns));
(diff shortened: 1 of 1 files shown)


TrueBrain (approved):
Less random is more boring! But I totally agree, rolling the dice per town is just silly. And the "offset" gives some nice variation, mostly for small town-counts.

I did wonder if there wasn't anything else cool or clever we can do here, but nothing spots to mind. So sure, why not :)

PR #12967, opened 2024-09-28
Codechange: [YAPF] Implement a tie breaker for Binary Heap

SamuXarick:
## Motivation / Problem
While doing path finding with multiple origins and then repeating the search with the origin chosen as the shortest path, it would at times result in a different path.

After some investigation, I noticed Binary Heap was removing the top item and reorganizing the queue so that the order they were first added would no longer be maintained.

## Description
Added a counter variable to keep track of the insertion order.
Changed the data vector to store pairs of (item, counter).
Updated the comparisons to consider the counter when priorities are equal.
Include method modified to include the counter in the pair when adding new items.

## Limitations
None that I can tell so far.

## Checklist for review

### DIFF
--- a/src/misc/binaryheap.hpp
+++ b/src/misc/binaryheap.hpp
@@ -53,7 +53,8 @@ template <class T>
 class CBinaryHeapT {
 private:
 	size_t items = 0; ///< Number of valid items in the heap
-	std::vector<T *> data; ///< The pointer to the heap item pointers
+	size_t counter = 0; ///< Counter to maintain insertion order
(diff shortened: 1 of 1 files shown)


Kuhnovic:
Maintaining the insertion order in the binary heap is not a requirement for the pathfinder to work. It doesn't matter in which order nodes-with-equal-costs are processed. It *does* matter that the binary heap does this in a way that is consistent across all platforms / clients, otherwise desyncs will occur. That doesn't seem to be the case AFAIK.

You have to explain your use case, otherwise this is just a change for the sake of change.

PR #12969, opened 2024-09-29
Add: AppendStringInPlace() to append translated string ID into existing string.

PeterN:
## Motivation / Problem

In a few places we build up strings outside of `FormatString()`, and append the result of `GetString()` to an existing `std::string`. Doing so involves another string copy.

## Description

Add GetStringInPlace() to place translated string ID into an existing string.

This allows avoiding a string copy when building these strings.

## Checklist for review

### DIFF
--- a/src/strings.cpp
+++ b/src/strings.cpp
@@ -323,6 +323,19 @@ std::string GetString(StringID string)
 }
 
 /**
+ * Resolve the given StringID and append in place into an existing std::string with all the associated
+ * DParam lookups and formatting.
+ * @param result The std::string to place the translated string.
+ * @param string The unique identifier of the translatable string.
+ */
+void AppendStringInPlace(std::string &result, StringID string)
+{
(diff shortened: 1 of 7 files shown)


src/strings.cpp
```diff
@@ -322,6 +322,19 @@ std::string GetString(StringID string)
 	return GetStringWithArgs(string, _global_string_params);
 }
 
+/**
+ * Resolve the given StringID and place into an existing std::string with all the associated
+ * DParam lookups and formatting.
+ * @param result The std::string to place the translated string.
+ * @param string The unique identifier of the translatable string.
+ */
+void GetStringInPlace(std::string &result, StringID string)
```
rubidium42: Given the name and comment I would assume that this function also "clears" the string, but that's not what it's supposed to do looking at the uses.
Maybe something along the lines of `AppendStringInPlace`?

rubidium42 (approved):

PR #7589, opened 2019-05-12
Change: Always apply inflation from 1920 to 2090, no matter the game start year.

michicc:
## Motivation / Problem

With inflation dependent on the game start year, the cost of a new vehicle in the year X can be £100, £10000, or £1000000. This makes balancing costs for NewGRF sets mostly impossible.

Authors deal with mostly in two ways: Either completely ignore inflation and only balance for inflation off. Or balance against inflation starting at a specific year.

## Description

As NewGRF sets are not balanced for a variable inflation start anyway, this PR fixes the inflation period to a known value so there is a consistent target to balance against.

## Limitations

As income is also affected by inflation, the starting difficulty increases with the starting year. With well balanced sets, this should be managable.

PR #12974, opened 2024-10-02
Fix #8932: Changed Max Loan in Settings From Currency Value to Percentage

### LINKED ISSUE #8932: Maximum initial loan amount does not match Game Settings
## Version of OpenTTD
1.11.0 (On Steam)

## Expected result
Maximum initial loan that is able to be taken should be limited to the amount in settings. Should be restricted with error message.

## Actual result
Player is able to Loan money up to an amount larger than in Game Settings. 

## Steps to reproduce
1. Start new game
2. Check "Maximum Initial Loan" value in Game Settings (e.g £300,000)
3. Use the Finances screen to borrow money. Once additional borrowing amount of £300,000 is reached (on top of the initial £150,000), the game will allow you to borrow more, in the screenshot a value of £550,000 is possible (Should be £150,000 + £300,000 = £450,000?)

OpenTTD_Loan

AviationGamerX:
## Motivation / Problem

As described in issue #8932, the value of maximum allowed loan in the settings wasn't the actual maximum loan in the game, since, with inflation enabled, that value can be different. There was a need to make it easier for the players to decide on a still value and not be surprised by a different value than what they set in the settings.
This issue seem to have appeared after the inflation change in #7589 

## Description

The solution , as [suggested](https://github.com/OpenTTD/OpenTTD/issues/8932#issuecomment-812704199) by @LordAro ,  was to have the setting allow the player to set the maximum value as a percentage instead of a currency value. Which is a better and simpler way for the player to decide to what extent they wish to increase the maximum loan in regards to whatever value the game's default is. And removes complication to why the value changes with inflation on.

## Limitations

- The setting only allows changing the percentage in intervals of 100%, and does not allow any percentages in between (someone can change this to allow more specific percentages or intervals)
- The limit of percentage allowed in the setting is 6000000 (approximate to the normal currency limit)

Note:
- 100% is currently equal to 300,000 British pounds 

## Checklist for review

### DIFF
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -195,7 +195,7 @@ struct Company : CompanyProperties, CompanyPool::PoolItem<&_company_pool> {
 */
 inline uint32_t GetMaxLoanFromPercentage()
 {
- 	return ((_settings_game.difficulty.max_loan_percentage / 100) * 300000) * GetCurrency().rate; 
+ 	return ((_settings_game.difficulty.max_loan_percentage / 100) * (300000 * GetCurrency().rate)); 
 }
 
 
--- a/src/company_base.h
+++ b/src/company_base.h
@@ -17,6 +17,7 @@
 #include "timer/timer_game_economy.h"
 #include "settings_type.h"
 #include "group.h"
+#include "currency.h"
 
 static const Money COMPANY_MAX_LOAN_DEFAULT = INT64_MIN;
 
@@ -188,6 +189,15 @@ struct Company : CompanyProperties, CompanyPool::PoolItem<&_company_pool> {
 
 	static void PostDestructor(size_t index);
 };
+
+/**
+* Function to get the max loan from the saved percentage in the settings (inline makes it avoid many unwanted errors)
+*/
+inline uint32_t GetMaxLoanFromPercentage()
+{
+ 	return ((_settings_game.difficulty.max_loan_percentage / 100) * 300000) * GetCurrency().rate; 
+}
+
 
 Money CalculateCompanyValue(const Company *c, bool including_loan = true);
 Money CalculateHostileTakeoverValue(const Company *c);
--- a/src/economy.cpp
+++ b/src/economy.cpp
@@ -33,6 +33,7 @@
 #include "sound_func.h"
 #include "autoreplace_func.h"
 #include "company_gui.h"
+#include "company_base.h"
 #include "signs_base.h"
 #include "subsidy_base.h"
 #include "subsidy_func.h"
@@ -762,7 +763,7 @@ bool AddInflation(bool check_year)
 void RecomputePrices()
 {
 	/* Setup maximum loan as a rounded down multiple of LOAN_INTERVAL. */
-	_economy.max_loan = ((uint64_t)_settings_game.difficulty.max_loan * _economy.inflation_prices >> 16) / LOAN_INTERVAL * LOAN_INTERVAL;
+	_economy.max_loan = ((uint64_t)GetMaxLoanFromPercentage() * _economy.inflation_prices >> 16) / LOAN_INTERVAL * LOAN_INTERVAL;
 
 	/* Setup price bases */
 	for (Price i = PR_BEGIN; i < PR_END; i++) {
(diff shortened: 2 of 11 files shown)


ldpl:
I don't think this is a good solution to anything. Problem in #8932 is it's hard to understand what value to use for the setting to get max loan of x pounds in the game when inflation is enabled. This just makes it hard to set max loan value period. For example, for a long time "hardest" max loan was 100k gbp, what do I need to set with this per to get that, 33.3333333%?. Can't even get 33 it seems, it just rounds to 100% and increasing precision to tens of percents won't help it either.

AviationGamerX:
What would you suggest as a good idea instead? Should we warn the player when creating a new game that the max loan will change since inflation is on? Maybe give the player the calculated new max loan as he is choosing the year?
Either that or we can set 100% to 100k, and allow 10% intervals, which would act like 10k intervals, and have the default percentage set to 300%.
Which would be a better solution?

ldpl:
Well, there already is a kind of warning in the setting description... 
IMO if this is to be treated as a bug at all and removing/ignoring inflation nonsense is not a option then best solution would be changing/adjusting how inflation affects max loan so it actually matches at the start of the game.

AviationGamerX:
> @ldpl Well, there already is a kind of warning in the setting description...

Indeed, but that one isn't very noticable, and can be skipped by most players, leading to the confusion that started the whole issue

> @ldpl  IMO if this is to be treated as a bug at all and removing/ignoring inflation nonsense is not an option then best solution would be changing/adjusting how inflation affects max loan so it actually matches at the start of the game. 

That would not be ideal, when inflation increases, usually so do loans, if you make it a strict maximum limit whatever the year is, it wouldn't be ideal if you are taking loans. The idea of percentage was to ease the process and make one value for all cases


> @LC-Zorg  According to the description: 100% = £10,000 / $20,000 / €20,000 / 60,000 PLN / 1,100,000 DKK / 5,000,000 p / ¥22,000,000

My bad forgot to mention this, I added that detail in the limitations, it's 300,000 for 100%

> @LC-Zorg I think that an appropriate description of the function would be enough. Including a clear statement that inflation can significantly increase the level of difficulty. It would also be possible to include information that would directly say how many times the costs will increase over the entire period. Another important piece of information is to indicate the connection between the loan interest rate and the level of inflation.

It would be smart indeed to have the information of both the notice of changing max loan, and how inflation changes in the hint-box of the inflation option, could also be a solution forward.

> @LC-Zorg  Reversing the change that caused this problem, especially since it is not the only one.

That would be a new issue then, you'll have to include all the problems it causes, and see what the others think.

ldpl:
> > @ldpl best solution would be changing/adjusting how inflation affects max loan so it actually matches at the start of the game.
> 
> @LC-Zorg But this is not a problem and there is no need for a solution here, because the value of the loan doesn't change - you can still buy and do the same as playing without inflation. The problem is that after setting the loan to e.g. 10,000, after starting the game, the player sees e.g. 8,000,000

You're contradicting yourself here. Player sees different value when starting the game because it gets adjusted for inflation. What I'm saying is that said adjustment should be changed so that value actually matches at the start of the game.

> @LC-Zorg Reversing the change that caused this problem, especially since it is not the only one.

And reversing is also a way to change/adjust how inflation works. It's just the way I put it allows for more options than simply reversing.

> @AviationGamerX  That would not be ideal, when inflation increases, usually so do loans, if you make it a strict maximum limit whatever the year is, it wouldn't be ideal if you are taking loans. The idea of percentage was to ease the process and make one value for all cases

I'm only speaking of the max loan value at the start of the game, it can increase as the game progresses. Iirc before #7589 it worked like this for ages and never bothered anyone.

What I don't like the most about percentage approach is that it makes things worse for a perfectly working case with disabled inflation. Even if you make "default" 100k-ish like LordAro initially suggested it's still confusing as you're essentially calling 1000 pounds a "percent" now for some reason.

AviationGamerX:
Well I will try to go with the solution of 100% = 100,000, and intervals of 10% = 10,000.
And work on the failed builds, whether or not it's a good idea, I will finish it and see then how it goes.

